PR #29537, opened 2025-08-11
ENH: np.unique: support hash based unique for complex dtype

math-hiyoko:
## Description
This PR introduces hash-based uniqueness extraction support for complex types in NumPy's np.unique function.

## Benchmark Results
The benchmark demonstrates significant performance improvement from the new implementation.

```bash
$ spin bench -t Unique.time_unique_value -c
```

| Change   | Before [989e9a6a] <main>   | After [4ff52ca6] <feature/#28363>   |   Ratio | Benchmark (Parameter)                                                               |
|----------|----------------------------|-------------------------------------|---------|-------------------------------------------------------------------------------------|
| -        | 5.92±0.08μs                | 4.75±0.06μs                         |    0.8  | bench_lib.Unique.time_unique_values(200, 10.0, 20.0, <class 'numpy.complex128'>)    |
| -        | 12.2±0.1ms                 | 8.63±0.1ms                          |    0.7  | bench_lib.Unique.time_unique_values(200000, 10.0, 20.0, <class 'numpy.complex128'>) |
| -        | 5.31±0.3μs                 | 2.95±0.04μs                         |    0.56 | bench_lib.Unique.time_unique_values(200, 10.0, 0.2, <class 'numpy.complex128'>)     |
| -        | 5.03±0.02ms                | 2.71±0.03ms                         |    0.54 | bench_lib.Unique.time_unique_values(200000, 90.0, 20.0, <class 'numpy.complex128'>) |
| -        | 5.57±0.04μs                | 2.62±0.05μs                         |    0.47 | bench_lib.Unique.time_unique_values(200, 90.0, 20.0, <class 'numpy.complex128'>)    |
| -        | 5.42±0.07μs                | 2.04±0.03μs                         |    0.38 | bench_lib.Unique.time_unique_values(200, 90.0, 0.2, <class 'numpy.complex128'>)     |
| -        | 9.91±0.1ms                 | 2.33±0.02ms                         |    0.23 | bench_lib.Unique.time_unique_values(200000, 10.0, 0.2, <class 'numpy.complex128'>)  |
| -        | 5.02±0.08ms                | 882±4μs                             |    0.18 | bench_lib.Unique.time_unique_values(200000, 90.0, 0.2, <class 'numpy.complex128'>)  |

close #28363


ngoldbaum:
Very cool! When you finish this up for review it'd be nice to see some before/after comparisons on some benchmarks.

math-hiyoko:
Now it's ready for being reviewed.

ngoldbaum:
I added the triage review label so hopefully someone at the next triage meeting on Wedesnday will volunteer to look closer at this. Unfortunately I can't take that on this month - too many things going on!

ngoldbaum:
I'm not sure what the state of the `unique` benchmarks is in *our* benchmarks, but I think it would help to add new benchmarks for this case to our benchmark suite.

It's probably worth benchmarking several sizes of inputs, not just really huge arrays.

ngoldbaum:
@charris is curious how you're handling NaN. What happens if there's more than one distinct NaN value in the array.

math-hiyoko:
> @charris is curious how you're handling NaN. What happens if there's more than one distinct NaN value in the array.

I handle this by referencing the equal_nan flag inside both the hash and equality functions used in the unordered_set. If equal_nan=True, all NaN values are treated as identical (they hash to the same value and compare equal). Otherwise, NaNs are treated as distinct and will not collapse into a single value.

```c++
size_t hash_maybenan(const T *value, npy_bool equal_nan) {
    if (equal_nan && isnan(*value)) {
        return 0;
    }
    return hash_nonan(value, equal_nan);
}

int equal_maybenan(const T *lhs, const T *rhs, npy_bool equal_nan) {
    int lhs_isnan = isnan(*lhs);
    int rhs_isnan = isnan(*rhs);
    if (lhs_isnan && rhs_isnan) {
        return equal_nan;
    }
    if (lhs_isnan || rhs_isnan) {
        return false;
    }
    return iseq_nonan(*lhs, *rhs);
}
```

math-hiyoko:
> I'm not sure what the state of the unique benchmarks is in our benchmarks, but I think it would help to add new benchmarks for this case to our benchmark suite.
> 
> It's probably worth benchmarking several sizes of inputs, not just really huge arrays.

We do already have a unique benchmark in:
[benchmarks/benchmarks/bench_lib.py#L119–154](https://github.com/numpy/numpy/blob/71eebaf8513dfd6ed40b0c689702d81e3f6f2527/benchmarks/benchmarks/bench_lib.py#L119-L154).
Right now it only targets float arrays and a limited set of sizes. I think we should broaden coverage—test larger sizes, add complex dtypes and StringDType.

Should I do it in this PR? or open another PR?

ngoldbaum:
> Should I do it in this PR? or open another PR?

Do you mind opening another PR? That should be mergeable quickly hopefully and then you can add a new benchmark here.

numpy/lib/tests/test_arraysetops.py
```diff
@@ -1226,8 +1231,14 @@ def test_unique_array_api_functions(self):
...
             assert len(res_unique_array_api) == len(res_unique)
+            if not isinstance(res_unique_array_api, tuple):
+                res_unique_array_api = (res_unique_array_api,)
+            if not isinstance(res_unique, tuple):
+                res_unique = (res_unique,)
+
             for actual, expected in zip(res_unique_array_api, res_unique):
-                assert_array_equal(actual, expected)
+                # Order of output is not guaranteed
+                assert_equal(np.sort(actual), np.sort(expected))
```
seberg, line 1248: Can we make sure we have tests for both NaNs and `-0.0` vs `0.0`.  For NaN, test with `np.nan` and `-np.nan` (I hope `-np.nan` is enough, but it stays in python world).

numpy/_core/src/multiarray/unique.cpp
```diff
@@ -32,11 +35,116 @@ FinalAction<F> finally(F f) {
...
+template <typename S, typename T, S (*real)(T), S (*imag)(T)>
+size_t hash_complex(const T *value, npy_bool equal_nan) {
+    S value_real = real(*value);
+    S value_imag = imag(*value);
+    int isnan = npy_isnan_wrapper<S>(value_real) || npy_isnan_wrapper<S>(value_imag);
+    if (equal_nan && isnan) {
+        return 0;
+    }
+
+    return hash_nonan<S>(&value_real, equal_nan) ^ hash_nonan<S>(&value_imag, equal_nan);
```
inakleinbottle: Mentioned at the triage meeting: using xor for hash combine might lead to some hash collisions. Since NumPy already uses fmv-hash, could we instead feed the raw bytes into this function instead? This retain the hash integrity of hash.

numpy/lib/tests/test_arraysetops.py
math-hiyoko: I’ve added test cases for both 0.0 vs -0.0 and for np.nan vs -np.nan
https://github.com/math-hiyoko/numpy/blob/d968a8ffbf87c1c206a128f3efff1ef266c36164/numpy/lib/tests/test_arraysetops.py#L1207-L1213
```python
        arr = np.array(
            [
                np.nan, 1.0, 0.0, 4.0, -np.nan,
                -0.0, 1.0, 3.0, 4.0, np.nan,
                5.0, -0.0, 1.0, -np.nan, 0.0,
            ],
        )
```

numpy/_core/src/multiarray/unique.cpp
math-hiyoko: Switched complex hashing to FNV-1a over the raw (real, imag) bytes (via npy_fnv1a) to avoid XOR-combine collisions
https://github.com/math-hiyoko/numpy/blob/d968a8ffbf87c1c206a128f3efff1ef266c36164/numpy/_core/src/multiarray/unique.cpp#L64-L75
```c++
template <typename S, typename T, S (*real)(T), S (*imag)(T)>
size_t hash_complex(const T *value, npy_bool equal_nan) {
    S value_real = real(*value);
    S value_imag = imag(*value);
    int isnan = npy_isnan_wrapper<S>(value_real) || npy_isnan_wrapper<S>(value_imag);
    if (equal_nan && isnan) {
        return 0;
    }

    S pair[2] = {value_real, value_imag};
    return npy_fnv1a(pair, 2 * sizeof(S));
}
```

inakleinbottle: I saw you changed the code once again after this to memcpy from the value_real and value_imag into the pair buffer. Was this to solve a specific problem? It is standard compliant to cast an inline array (implicitly `S*`) to a `void*` directly via a `static_cast` (or just letting it decay naturally), but this wouldn't work with a const array. 

The char buffer might be needed if S is not trivially constructible, but shouldn't be needed if S is a float type (incl. long double).

math-hiyoko: The reason I switched this code was that the tests are currently failing in CI.
https://github.com/numpy/numpy/actions/runs/17482904180/job/49656895573

I was experimenting with different approaches to see if that might resolve the issue (still failing).

math-hiyoko:
Following the merge of https://github.com/numpy/numpy/pull/29621 (expanded benchmark coverage), I ran a before/after comparison against main and this branch.

Comparison: Before = 87e208c9 (main) vs After = 6c6822a6 (feature/#28363)

| Change   | Before [87e208c9] <main>   | After [6c6822a6] <feature/#28363>   |   Ratio | Benchmark (Parameter)                                                               |
|----------|----------------------------|-------------------------------------|---------|-------------------------------------------------------------------------------------|
| +        | 509±8μs                    | 1.93±0.03ms                         |    3.78 | bench_lib.Unique.time_unique_values(200000, 90.0, 20.0, <class 'numpy.float64'>)    |
| +        | 3.14±0.01ms                | 6.08±0.02ms                         |    1.94 | bench_lib.Unique.time_unique_values(200000, 10.0, 20.0, <class 'numpy.float64'>)    |
| +        | 377±10μs                   | 655±1μs                             |    1.74 | bench_lib.Unique.time_unique_values(200000, 90.0, 0.2, <class 'numpy.float64'>)     |
| +        | 5.30±0.02μs                | 6.22±0.04μs                         |    1.17 | bench_lib.Unique.time_unique_values(200, 10.0, 20.0, <class 'numpy.complex128'>)    |
| +        | 1.11±0.01ms                | 1.27±0.01ms                         |    1.14 | bench_lib.Unique.time_unique_values(200000, 10.0, 0.2, <class 'numpy.float64'>)     |
| -        | 4.83±0.02μs                | 4.12±0.03μs                         |    0.85 | bench_lib.Unique.time_unique_values(200, 10.0, 0.2, <class 'numpy.complex128'>)     |
| -        | 4.10±0.09ms                | 2.96±0.04ms                         |    0.72 | bench_lib.Unique.time_unique_values(200000, 90.0, 20.0, <class 'numpy.complex128'>) |
| -        | 5.54±0.05μs                | 3.64±0.1μs                          |    0.66 | bench_lib.Unique.time_unique_values(200, 10.0, 20.0, <class 'numpy.float64'>)       |
| -        | 3.82±0.04μs                | 2.34±0.02μs                         |    0.61 | bench_lib.Unique.time_unique_values(200, 90.0, 20.0, <class 'numpy.float64'>)       |
| -        | 4.91±0.05μs                | 2.71±0.03μs                         |    0.55 | bench_lib.Unique.time_unique_values(200, 90.0, 20.0, <class 'numpy.complex128'>)    |
| -        | 3.87±0.02μs                | 1.98±0.02μs                         |    0.51 | bench_lib.Unique.time_unique_values(200, 10.0, 0.2, <class 'numpy.float64'>)        |
| -        | 3.84±0.05μs                | 1.81±0.05μs                         |    0.47 | bench_lib.Unique.time_unique_values(200, 90.0, 0.2, <class 'numpy.float64'>)        |
| -        | 5.06±0.05μs                | 2.12±0.01μs                         |    0.42 | bench_lib.Unique.time_unique_values(200, 90.0, 0.2, <class 'numpy.complex128'>)     |
| -        | 9.11±0.04ms                | 3.72±0.03ms                         |    0.41 | bench_lib.Unique.time_unique_values(200000, 10.0, 0.2, <class 'numpy.complex128'>)  |
| -        | 4.14±0.09ms                | 1.12±0.01ms                         |    0.27 | bench_lib.Unique.time_unique_values(200000, 90.0, 0.2, <class 'numpy.complex128'>)  |

- For complex arrays the hash-based path yields consistent speedups (down to 0.27×–0.72× the baseline runtime).
- For float arrays there are regressions at a number of sizes/mixes. This likely reflects how fast NumPy’s sort-based path already is for floats.

Given the mixed results, I suggest scoping this PR to complex dtypes for now (where the wins are clear), and leaving float on the existing sort-based path. We can revisit float later with more targeted strategies

ngoldbaum:
> Given the mixed results, I suggest scoping this PR to complex dtypes for now (where the wins are clear), and leaving float on the existing sort-based path. We can revisit float later with more targeted strategies

That makes sense to me. We agreed at the last triage meeting that while this is a sort-of niche use-case, there are definitely times when someone wants to get the unique set of complex floating point values in an array (e.g. image processing tasks for a simulation of a quantum system) and it's worth doing this given the speedups for complex numbers.

doc/release/upcoming_changes/29537.change.rst
```diff
@@ -0,0 +1,7 @@
+``unique_values`` for float and complex dtypes may return unsorted data
```
ngoldbaum: it's only complex now, right? also other mentions of `float` in the release notes below

ngoldbaum (commented):
@inakleinbottle any chance I can get you to look this over with an eye towards C++ correctness?

numpy/_core/src/multiarray/unique.cpp
math-hiyoko: Some floating types (notably npy_longdouble) only use a subset of the bits in their storage.
For example, on some ABIs an 80-bit extended precision value lives in a 16-byte slot, and the remaining bytes/bitfields are unspecified. Those unused/padding bits can be indeterminate, so feeding `npy_fnv1a(pair, 2 * sizeof(S))` with the raw object representation may yield different hashes for numerically equal values (depending on how those slack bits happen to be initialized), and that can show up as non-determinism in CI.

ngoldbaum: That makes sense. Adding a special case for longdouble could work. Clomplex256 is pretty niche! I think in the long run we also want to point people at numpy_quaddtype.

math-hiyoko: I’ve switched the combine step to multiply by the FNV “magic prime” before mixing in the imaginary part, to avoid collisions
```c++
template <typename S, typename T, S (*real)(T), S (*imag)(T)>
size_t hash_complex(const T *value, npy_bool equal_nan) {
    S value_real = real(*value);
    S value_imag = imag(*value);
    int hasnan = npy_isnan(value_real) || npy_isnan(value_imag);
    if (equal_nan && hasnan) {
        return 0;
    }

    size_t hval = std::hash<S>{}(value_real);
    #if NPY_SIZEOF_SIZE_T == 8
        /* multiply by the 64 bit FNV magic prime */
        /* hval *= 0x100000001b3ULL; */
        hval += (hval << 1) + (hval << 4) + (hval << 5) +
		        (hval << 7) + (hval << 8) + (hval << 40);
    #else /* NPY_SIZEOF_SIZE_T == 4 */
        /* multiply by the 32 bit FNV magic prime */
        /* hval *= 0x01000193; */
        hval += (hval<<1) + (hval<<4) + (hval<<7) + (hval<<8) + (hval<<24);
    #endif
    hval ^= std::hash<S>{}(value_imag);

    return hval;
}
```

doc/release/upcoming_changes/29537.change.rst
math-hiyoko: I have fixed the release note and the performance note

math-hiyoko:
I’ve addressed all review comments and updated the PR title and discussion accordingly.

math-hiyoko:
Ping for review

math-hiyoko:
@ngoldbaum 
Another ping for review.
When you get a chance, could you please take a look?

ngoldbaum:
Hi, sorry for not responding - I'm traveling this week and next week. I have this on my to-do list for when I get back.

numpy/_core/src/multiarray/unique.cpp
```diff
@@ -33,11 +36,92 @@ FinalAction<F> finally(F f) {
...
+    #if NPY_SIZEOF_SIZE_T == 8
+        /* multiply by the 64 bit FNV magic prime */
+        /* hval *= 0x100000001b3ULL; */
+        hval += (hval << 1) + (hval << 4) + (hval << 5) +
+		        (hval << 7) + (hval << 8) + (hval << 40);
+    #else /* NPY_SIZEOF_SIZE_T == 4 */
+        /* multiply by the 32 bit FNV magic prime */
+        /* hval *= 0x01000193; */
+        hval += (hval<<1) + (hval<<4) + (hval<<7) + (hval<<8) + (hval<<24);
+    #endif
```
ngoldbaum: This is wrong. The FNV algorithm operates on *bytes* of data. This code is operating on size_t sized ints. 

We want you to hash the bytes of the complex value. Very schematically, without actually compiling any code, take this worth a grain of salt, but something like this:

```
unsigned char* value_bytes = (unsigned char *)&value
size_t hash = npy_fnv1a(value_bytes, sizeof(T))
```

ngoldbaum (changes requested):
I noticed the incorrect hashing algorithm and didn't look at the rest of the code.

numpy/_core/src/multiarray/unique.cpp
math-hiyoko: Thanks for the review.

As I noted earlier, hashing the raw object representation doesn’t work on some dtypes(notably npy_longdouble / complex256).
https://github.com/numpy/numpy/pull/29537#discussion_r2340626617

Is there a reliable way to obtain a byte sequence containing only the meaningful bits (excluding unused or padding bits) from a value?

ngoldbaum: I think you'd have to adjust my suggestion using `sizeof` to account for the padding for types that use that. For complex256 you'd have to do the same thing twice on the real and imaginary parts. I think you'd have to "manually" do that using explicit overloads for types that need it.

ngoldbaum:
I see you debugging in CI here. That's a pretty expensive (in terms of energy usage) way to debug. Maybe you'll find the docs I recently wrote about using native debuggers with NumPy useful? https://numpy.org/devdocs/dev/development_advanced_debugging.html#c-debuggers

math-hiyoko:
Thank you for the suggestion, but the bug did not reproduce in my local environment even in debug build

math-hiyoko:
@ngoldbaum 
As described in npy_math_private.h, some floating-point formats include junk bits inside their binary representation, which means that simply hashing the sizeof representation cannot work reliably.
To address this, I used the macros defined in that file to extract and pack the mantissa, exponent, and sign components separately into a buffer, and then computed the hash from that buffer.
This approach successfully avoids the undefined-bit issue.
https://github.com/numpy/numpy/blob/989e9a6a83a0c1bf11dd5280086694324eac1f47/numpy/_core/src/npymath/npy_math_private.h#L254-L261

ngoldbaum:
Awesome! I'll take a close look at this soon and will make sure we ship this in 2.4.0.

In the meantime, would you mind re-running the benchmarks to make sure there are no big regressions with the new implementation? Or maybe it's faster? Also to make sure the numbers in the release note are still accurate.

math-hiyoko:
I already run benchmarks and ensured there is no big regressions.
The result table in the PR comment https://github.com/numpy/numpy/pull/29537#issue-3308416157 is updated to the latest commit.

numpy/_core/src/multiarray/unique.cpp
```diff
@@ -34,11 +36,148 @@ FinalAction<F> finally(F f) {
...
+    // Some floating-point complex dtypes (e.g., npy_complex256) include undefined or
+    // unused bits in their binary representation
+    // (see: https://github.com/numpy/numpy/blob/main/numpy/_core/src/npymath/npy_math_private.h#L254-L261).
+    // Because hashing the raw bit pattern would make the hash depend on those
+    // undefined bits, we extract the mantissa, exponent, and sign components
+    // explicitly and pack them into a buffer to ensure the hash is well-defined.
+    #if defined(HAVE_LDOUBLE_IBM_DOUBLE_DOUBLE_BE) || defined(HAVE_LDOUBLE_IBM_DOUBLE_DOUBLE_LE)
+    constexpr size_t SIZEOF_BUFFER = NPY_SIZEOF_CLONGDOUBLE;
+    const unsigned char* buffer = reinterpret_cast<const unsigned char*>(value);
+    #else
```
ngoldbaum: The approach below looks correct to me. My only comment is that you might be able to add an `#else if` block for platforms where longdouble and double are identical, avoiding the possibly slightly more expensive but standards-compliant version below for 80-bit longdouble on Linux. I don't know offhand if we already have preprocessor macros for platforms where longdouble is double already defined. If not, then this is fine as-is and no need for you to figure that out to merge this PR.

ngoldbaum (approved):
ping @seberg and @inakleinbottle in case either of you would like to give this a once-over and hit the merge button if you don't spot any issues

numpy/_core/src/multiarray/unique.cpp
math-hiyoko: In practice, long double representations that contain junk bits only appear in the following cases:
• HAVE_LDOUBLE_INTEL_EXTENDED_12_BYTES_LE
• HAVE_LDOUBLE_INTEL_EXTENDED_16_BYTES_LE
• HAVE_LDOUBLE_MOTOROLA_EXTENDED_12_BYTES_BE

Therefore, the high-cost special handling is actually required only for these cases.
For all other platforms,
`const unsigned char* buffer = reinterpret_cast<const unsigned char*>(value);`
should be sufficient.

Actually, macros such as HAVE_LDOUBLE_IEEE_DOUBLE_BE indicate that long double has the same representation as double in big-endian format.

https://github.com/numpy/numpy/blob/989e9a6a83a0c1bf11dd5280086694324eac1f47/numpy/_core/src/npymath/npy_math_private.h#L178-L290

ngoldbaum:
Alright, I’d like to have this merged in with a little time before 2.4.0 comes out. I doubt anyone else will look closely before the release and I already held this up enough already.

Thanks for your diligence here @math-hiyoko.

PR #29539, opened 2025-08-11
BUG: left bit shift undefined behavior

### LINKED ISSUE #29524: BUG: resolve undefined behavior issues currently suppressed by UBSAN CI
This is a tracking issue for resolving the [UndefinedBehaviorSanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) identified issues in numpy. This tracking issue will be closed when the files/modules currently being suppressed for testing with UBSAN have been fixed. Then the suppression for each file/module can be removed from `tools/ci/ubsan_suppressions_x86_64.txt` or `tools/ci/ubsan_suppressions_arm64.txt` and the `compiler_sanitizers.yml` workflow will still run successfully in CI. Some suggested potential fixes are documented in the suppressions files, and repeated here.

To determine where the issue is in the underlying file, you should build numpy with:
```
python -m spin build -- -Db_sanitize=undefined -Db_lundef=false
```

and run tests with:
```
UBSAN_OPTIONS=halt_on_error=1:suppressions=/tools/ci/ubsan_suppressions_x86_64.txt \
spin test -- -v -s --timeout=600 --durations=10
```
and remove the directive line from the suppressions file while testing.

## arm64

- [ ] `shift-base:_core/src/umath/_rational_tests.c` - use `INT_MIN` constant
- [ ] `shift-base:_core/src/npymath/npy_math_internal.h` - check for overflow if signed

add null check before loop:
- [ ] `pointer-overflow:_core/src/common/simd/neon/memory.h`
- [ ] `pointer-overflow:_core/src/multiarray/datetime_busdaycal.c`
- [ ] `pointer-overflow:_core/src/multiarray/nditer_templ.c`
- [ ] `pointer-overflow:_core/src/multiarray/nditer_constr.c`
- [ ] `pointer-overflow:_core/src/umath/loops_arithm_fp.dispatch.c.src`
- [ ] `pointer-overflow:_core/src/umath/loops_unary.dispatch.c.src`
- [ ] `pointer-overflow:_core/src/umath/loops_unary_complex.dispatch.c.src`
- [ ] `pointer-overflow:_core/src/umath/loops_unary_fp_le.dispatch.c.src`
- [ ] `pointer-overflow:_core/src/umath/string_buffer.h`
- [ ] `pointer-overflow:linalg/umath_linalg.cpp`
- [ ] `pointer-overflow:numpy/random/bit_generator.pyx.c`

- [ ] `float-cast-overflow:_core/src/multiarray/lowlevel_strided_loops.c.src`

`_core/src/` and `numpy/random/` had many modules with `call to function through pointer to incorrect function type` errors in UBSAN. Some are documented here, but more would need to be identified to disable the general suppressions on `function:_core/src/*` and `function:numpy/random/*`
- [ ] `function:_core/src/common/cblasfunc.c`
- [ ] `function:_core/src/common/npy_argparse.c`
- [ ] `function:_core/src/multiarray/number.c`
- [ ] `function:_core/src/multiarray/ctors.c`
- [ ] `function:_core/src/multiarray/convert_datatype.c`
- [ ] `function:_core/src/multiarray/dtype_transfer.c`
- [ ] `function:_core/src/multiarray/dtype_traversal.c`
- [ ] `function:_core/src/multiarray/getset.c`
- [ ] `function:_core/src/multiarray/scalarapi.c`
- [ ] `function:_core/src/multiarray/scalartypes.c.src`

## x86-64

replace left bit shift with `LLONG_MIN` constant:
- [x] `shift-base:_core/src/common/simd/sse/arithmetic.h`
- [x] `shift-base:_core/src/common/simd/avx2/arithmetic.h`

use INT_MIN constant:
- [ ] `shift-base:_core/src/umath/_rational_tests.c`

check for overflow if signed:
- [ ] `shift-base:_core/src/npymath/npy_math_internal.h`

check that pointer is not null before calling function:
- [ ] `nonnull-attribute:_core/src/multiarray/array_coercion.c`
- [ ] `nonnull-attribute:_core/src/multiarray/ctors.c`
- [ ] `nonnull-attribute:_core/src/multiarray/datetime_busdaycal.c`
- [ ] `nonnull-attribute:_core/src/multiarray/scalarapi.c`
- [ ] `nonnull-attribute:_core/src/multiarray/calculation.c`

null check before loop:
- [ ] `pointer-overflow:_core/src/multiarray/nditer_templ.c`

m-clare:
- Addresses UBSAN warnings for `sse/arithmetic.h` and `avx2/arithmetic.h` 
- adjusts variable names for operations (`sigb` to `sbit`) to match those in `sse/operations.h` `avx2/operations.h` for the same operation with a vector argument (see snippet below).
- Removes UBSAN suppressions for these files, `compiler_sanitizers.yml` workflow should succeed without errors.
- partially closes #29524 

From `sse/operations.h` L34:
```
NPY_FINLINE __m128i npyv_shr_s64(__m128i a, int c)
{
    const __m128i sbit = npyv_setall_s64(0x8000000000000000);
    const __m128i cv   = _mm_cvtsi32_si128(c);
    __m128i r = _mm_srl_epi64(_mm_add_epi64(a, sbit), cv);
    return _mm_sub_epi64(r, _mm_srl_epi64(sbit, cv));
}
```

The pattern throughout these implementations is to use the hex representation of the value, which I've done here. Not sure if it would be a better option to use the `limit` library instead.

### DIFF
--- a/numpy/_core/src/common/simd/avx2/arithmetic.h
+++ b/numpy/_core/src/common/simd/avx2/arithmetic.h
@@ -215,9 +215,9 @@ NPY_FINLINE npyv_s64 npyv_divc_s64(npyv_s64 a, const npyv_s64x3 divisor)
     // q               = (a + mulhi) >> sh
     __m256i q          = _mm256_add_epi64(a, mulhi);
     // emulate arithmetic right shift
-    const __m256i sigb = npyv_setall_s64(1ULL << 63);
-            q          = _mm256_srl_epi64(_mm256_add_epi64(q, sigb), shf1);
-            q          = _mm256_sub_epi64(q, _mm256_srl_epi64(sigb, shf1));
+    const __m256i sbit = npyv_setall_s64(0x8000000000000000);
+            q          = _mm256_srl_epi64(_mm256_add_epi64(q, sbit), shf1);
+            q          = _mm256_sub_epi64(q, _mm256_srl_epi64(sbit, shf1));
     // q               = q - XSIGN(a)
     // trunc(a/d)      = (q ^ dsign) - dsign
             q          = _mm256_sub_epi64(q, asign);
--- a/numpy/_core/src/common/simd/avx2/arithmetic.h
+++ b/numpy/_core/src/common/simd/avx2/arithmetic.h
@@ -215,7 +215,7 @@ NPY_FINLINE npyv_s64 npyv_divc_s64(npyv_s64 a, const npyv_s64x3 divisor)
     // q               = (a + mulhi) >> sh
     __m256i q          = _mm256_add_epi64(a, mulhi);
     // emulate arithmetic right shift
-    const __m256i sigb = npyv_setall_s64(1LL << 63);
+    const __m256i sigb = npyv_setall_s64(1ULL << 63);
             q          = _mm256_srl_epi64(_mm256_add_epi64(q, sigb), shf1);
             q          = _mm256_sub_epi64(q, _mm256_srl_epi64(sigb, shf1));
     // q               = q - XSIGN(a)
--- a/numpy/_core/src/common/simd/sse/arithmetic.h
+++ b/numpy/_core/src/common/simd/sse/arithmetic.h
@@ -251,9 +251,9 @@ NPY_FINLINE npyv_s64 npyv_divc_s64(npyv_s64 a, const npyv_s64x3 divisor)
     // q               = (a + mulhi) >> sh
     __m128i q          = _mm_add_epi64(a, mulhi);
     // emulate arithmetic right shift
-    const __m128i sigb = npyv_setall_s64(1ULL << 63);
-            q          = _mm_srl_epi64(_mm_add_epi64(q, sigb), divisor.val[1]);
-            q          = _mm_sub_epi64(q, _mm_srl_epi64(sigb, divisor.val[1]));
+    const __m128i sbit = npyv_setall_s64(0x8000000000000000);
+            q          = _mm_srl_epi64(_mm_add_epi64(q, sbit), divisor.val[1]);
+            q          = _mm_sub_epi64(q, _mm_srl_epi64(sbit, divisor.val[1]));
     // q               = q - XSIGN(a)
     // trunc(a/d)      = (q ^ dsign) - dsign
             q          = _mm_sub_epi64(q, asign);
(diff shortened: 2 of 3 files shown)


ngoldbaum:
I re-triggered the failing CI job, which was unrelated to this PR.

ngoldbaum:
Ping @seiko2plus - this PR touches code you originally wrote back in 2021.

numpy/_core/src/common/simd/avx2/arithmetic.h
```diff
@@ -215,9 +215,9 @@ NPY_FINLINE npyv_s64 npyv_divc_s64(npyv_s64 a, const npyv_s64x3 divisor)
     // q               = (a + mulhi) >> sh
     __m256i q          = _mm256_add_epi64(a, mulhi);
     // emulate arithmetic right shift
-    const __m256i sigb = npyv_setall_s64(1LL << 63);
-            q          = _mm256_srl_epi64(_mm256_add_epi64(q, sigb), shf1);
-            q          = _mm256_sub_epi64(q, _mm256_srl_epi64(sigb, shf1));
+    const __m256i sbit = npyv_setall_s64(0x8000000000000000);
```
seberg, line 218: I suppose it is relatively clear from the `sbit` context, but I wonder if we can at least keep a
```suggestion
    const __m256i sbit = npyv_setall_s64(0x8000000000000000 /* 1 << 63 */);
```
(since I don't have a better idea, I assume `1ULL << 63` doesn't warn, but then we would also need to cast back to long long probably, so dunno that is better)

seberg (approved):
I am fine with just putting it in, I do think the `<<` is clearer to read, but this is so targeted, that I guess it just doesn't matter.

numpy/_core/src/common/simd/avx2/arithmetic.h
m-clare: I did try it out with `1ULL << 63` and it removed the warning, but was trying to be consistent with other bit shift operations.

ngoldbaum: Sebastian's suggestion just adds a comment, I think it's fine to add. Probably also worth adding in `sse/operators.h` as well:

```
sse/operators.h
36:    const __m128i sbit = npyv_setall_s64(0x8000000000000000);
220:        const __m128i sbit = npyv_setall_s64(0x8000000000000000);
```

There's also this in `sse/arithmetic.h`:

```
sse/arithmetic.h
254:    const __m128i sigb = npyv_setall_s64(1LL << 63);
```

Although maybe this one is fine because it's a signed long long?

m-clare: Nope, that one also gets flagged by UBSAN. I also have a fix in for that in this PR.

ngoldbaum: Ah of course, got a little tunnel-vision there.

Let's merge this as-is, we can clean up the constants to be formatted nicely later. Fixing the UB is way more important.

ngoldbaum:
Thanks @m-clare!

PR #29540, opened 2025-08-11
CI: run some wheel build jobs by default, and clean up the rest

rgommers:
Now that we have https://github.com/numpy/numpy-release, we can remove most wheel build jobs from this repo and at the same time improve the default set of jobs running in CI by adding some wheel build jobs to the mix (none got run by default before this PR).

Up for discussion:
- Exactly what do we run here? My first stab at it was 1 wheel per platform, minus the ones where we don't vendor OpenBLAS.
- Do we run the `full` or `fast` test suite?
- Are there other jobs we can clean up now because of overlap (like the `linux_musl.yml` one removed here)
- Still TBD: how exactly do we deal with some remaining overlap between files in the `numpy` and `numpy-release` repos? xref https://github.com/numpy/numpy-release/issues/6 which @mattip just opened


andyfaff:
The `linux_musl.yml` file  was supposed to be the place for examining musl, it just so happened to be built/installed through a wheel. I still think it makes sense to keep a musl linux run in this file, but we probably don't need to send it through a wheel first. We may find out at a later stage that we want to amend some options for such a test run.

What pain points do the full/fast runs tend to highlight? We should probably run full on numpy-releases (that gets triggered less), so we can probably run fast in the wheels here, and save some full runs for builds that don't go through wheel building.

rgommers:
> What pain points do the full/fast runs tend to highlight? 

I just had a look at what's decorated with `mark.slow`, and it's not that much - mostly `f2py` tests, I/O tests, some other compile tests, and static typing tests (which won't run by default anyway). So the extra value of running them on more then a handful of platforms is pretty low.

> I still think it makes sense to keep a musl linux run in this file,

I really don't think it adds any value. It made a lot of sense when we did not run wheel build jobs on PRs, but when we run a `musllinux` x86-64 and aarch64 wheel build job already, the extra job is 100% duplicate. 

Other jobs that can go, for the same reason, are:
- the Windows x86-64 and arm64 ones that build with `scipy-openblas64` in `windows.yml`
- the plain Linux one on `ubuntu-22.04-arm` in `linux.yml`

If we remove all those, that still leaves us with four extra jobs after this PR, which seems acceptable for the win of testing wheel builds by default. 

> Still TBD: how exactly do we deal with some remaining overlap between files in the `numpy` and `numpy-release` repos?

Still thinking about this. Less duplication is good for maintainability. We need to have at least some files in this repo so we can do wheel builds. We don't need the licensing related stuff here, because we're not distributing wheels from this repo anymore, and there's a benefit to removing it (e.g., no license checkers will notice LGPL etc. in this repo and raise flags by default).

The main things that leaves are:
1. `cibw_before_build.sh`, `cibw_test_command.sh` and `repair_windows.sh`
2. the `pyproject.toml` cibuildwheel config
3. requirements files

I'm inclined to say for (1) that we avoid duplication and use the files in this repo from `numpy-release`. For (2) the configs are slightly different (e.g., `pyodide` is added), but could be reused. For (3) it's best kept separate with a check on no mismatches probably, and ensuring the `numpy-release` ones are minimal; the ones in this repo are very messy with many unnecessary build and test dependencies added.

rgommers:
This is ready now. I went with the `full` test suite for now, since the extra time is only a couple of minutes per job and I'd like to avoid testing the `slow` tests on niche architectures in `numpy-release` only for now.

tools/wheels/cibw_before_build.sh
```diff
@@ -26,40 +13,40 @@ elif [ -z $INSTALL_OPENBLAS ]; then
...
+    rm -rf $pkgconf_path
+    mkdir -p $pkgconf_path
+    python -m pip install -r $PROJECT_DIR/requirements/ci_requirements.txt
+    python -c "import scipy_${OPENBLAS}; print(scipy_${OPENBLAS}.get_pkg_config())" > $pkgconf_path/scipy-openblas.pc
+
+    # Copy scipy-openblas DLL's to a fixed location so we can point delvewheel
+    # at it in `repair_windows.sh` (needed only on Windows because of the lack
+    # of RPATH support).
+    if [[ $RUNNER_OS == "Windows" ]]; then
+        python <<EOF
```
andyfaff, line 42: This bit is a little complicated to follow for me because the indentation is changing.

tools/wheels/LICENSE_linux.txt
andyfaff, line 1: When the licences are removed we're losing a record of the code that's being bundled (OpenBLAS, etc).
pyproject.toml says 

> The licenses for these vendored components are dynamically included in the build process for PyPI wheels.

I can see these files are now in numpy-release, but isn't it important to have a record of them here somewhere. Are there going to be improvements in how we bundle these extra licences in the wheel build process?

pyproject.toml
```diff
@@ -202,45 +201,33 @@ manylinux-aarch64-image = "manylinux_2_28"
...
 [tool.cibuildwheel.linux.environment]
-# RUNNER_OS is a GitHub Actions specific env var; define it here so it works on Cirrus CI too
+# RUNNER_OS is a GitHub Actions specific env var; define it here so it's
+# defined when running cibuildwheel locally
 RUNNER_OS="Linux"
 # /project will be the $PWD equivalent inside the docker used to build the wheel
 PKG_CONFIG_PATH="/project/.openblas"
-LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/project/.openblas/lib"
-
-[tool.cibuildwheel.macos]
```
andyfaff, line 222: So these entries are being removed because we only need them in numpy-release now?

.github/workflows/linux_musl.yml
andyfaff, line 1: Is the replacement for this the run in wheels.yml?

.github/workflows/linux.yml
```diff
@@ -157,41 +157,6 @@ jobs:
       env:
         PYTHONOPTIMIZE: 2
 
-
-  aarch64_test:
-    needs: [smoke_test]
```
andyfaff, line 162: Is the replacement for this the run in wheels.yml?

.github/workflows/wheels.yml
```diff
@@ -85,34 +35,13 @@ jobs:
           - [ubuntu-22.04-arm, manylinux_aarch64, ""]
           - [ubuntu-22.04-arm, musllinux_aarch64, ""]
           - [macos-13, macosx_x86_64, openblas]
-          - [macos-13, macosx_x86_64, accelerate]
```
andyfaff, line 88: Did you want to remove all macos-13 runs here?

mattip: Should we move them to azure where there are still native macos x86_64 runners?

tools/wheels/LICENSE_linux.txt
rgommers: I think it's actually somewhat important to _not_ have GPL licenses in this repo, because there is nothing inside this repo that's GPL, LGPL or otherwise copyleft.

I'd really like to conceptually separate NumPy the project and how its sources are licensed - which matters to anyone building/redistributing `numpy` - and wheels on PyPI, which are only one way of distributing `numpy`, and a quite unusual way at that. If you get `numpy` from any other source than PyPI, it will not contain copyleft components.

> Are there going to be improvements in how we bundle these extra licences in the wheel build process?

Hopefully yes. See https://github.com/numpy/numpy-release/issues/11

.github/workflows/linux.yml
rgommers: Yes. This job doesn't nothing special being testing Linux on aarch64 with OpenBLAS. The wheel build does that too.

.github/workflows/linux_musl.yml
rgommers: Yes indeed, there are two `musllinux` jobs, one for x86-64 and one for aarch64, in `wheels.yml`.

.github/workflows/wheels.yml
rgommers: Can we do that separately based on what comes out of the discussion at gh-29728?

I haven't yet looked into what Azure plans to do with their platform support. Also, I hope we'll get rid of Azure completely ....

rgommers: For this one, my thinking was that there's another job that tests Accelerate LP64 and ILP64 at once (see [here](https://github.com/numpy/numpy/blob/83402f7a43363e8041b8f297fdd4f76a7fa9fea1/.github/workflows/macos.yml#L142-L151), plus this particular job is quite slow, so I wanted to remove this one as duplicate.

pyproject.toml
rgommers: No we still have macOS wheel builds here in `wheels.yml`. I actually improved how this bundling works, so messing with `DYLD_LIBRARY_PATH` is no longer necessary. If you look at https://github.com/numpy/numpy-release/blob/main/cibuildwheel.toml, you won't find a `macos` entry there either.

tools/wheels/cibw_before_build.sh
rgommers: Yeah, this took me a while. The old version was super confusing, it was unclear when reading the code that `PKG_CONFIG_PATH` was not an environment variable.

The diff is hard to read; just looking at the old and new versions side by side should be helpful. What it does isn't changing IIRC, the new version should just be easier to understand.

mattip: +1 from me, as the previous version's author

.github/workflows/windows.yml
```diff
@@ -85,80 +62,7 @@ jobs:
       run: |
         spin test -- --timeout=600 --durations=10
 
-  python64bit_openblas_winarm64:
```
andyfaff, line 88: I think this is covered by the wheels.yml run

andyfaff:
I'm happy for this to be merged. I'll do so tomorrow morning if it's not done by then.

mattip:
Thanks @rgommers

rgommers:
Thanks for the reviews! Let me look at the last open comment about the `python64bit_openblas_winarm64` job as a potential follow-up.

.github/workflows/windows.yml
rgommers: Oh never mind, I thought this needed to be looked into more, but it's only a "yep, seems right" comment. All good here then.

Issue #29544, opened 2025-08-12
DOC: examples in reference documentation are not run through doctest


mattip:
### Issue with current documentation:

In #21070, we discussed the use of various strategies to doctest documentation. The issue was not resolved. In the mean time, `spin` has grown an option to check RST documentation, and running it on the reference documentation results in errors:

<details><summary>`spin check-tutorials doc/source/reference/`</summary>
```
FAILED ../../../../../doc/source/reference/array_api.rst::array_api.rst
FAILED ../../../../../doc/source/reference/arrays.classes.rst::arrays.classes.rst
FAILED ../../../../../doc/source/reference/arrays.datetime.rst::arrays.datetime.rst
FAILED ../../../../../doc/source/reference/arrays.nditer.cython.rst::arrays.nditer.cython.rst
FAILED ../../../../../doc/source/reference/arrays.promotion.rst::arrays.promotion.rst
FAILED ../../../../../doc/source/reference/constants.rst::constants.rst
FAILED ../../../../../doc/source/reference/distutils.rst::distutils.rst
FAILED ../../../../../doc/source/reference/c-api/coremath.rst::coremath.rst
```

</details>

Additionally, `refguidecheck --help` still shows the `--rst` and `--doctests` options, but they no longer exist as command line arguments

### Idea or request for content:

- Fix the `--help` string of refguide_check` to reflect its true use (checking that `__all__` is correct).
- Fix or mark-as-skip the reference documentation examples
- Rework the doctest test in circleCI to test the reference documentation https://github.com/numpy/numpy/blob/3049f4ea53e083d34461679629a515c7bf053049/.circleci/config.yml#L94-L101

ev-br:
Expanding from https://github.com/numpy/numpy/pull/29534#issuecomment-3172553287: 

Currently, `$ spin check-tutorials` only checks the contents of `doc/dource/user/*rst`. There are several other directories next to it though:

```
~/repos/numpy/doc/source$ ll -d */
drwxrwxr-x 2 br br 4096 июн  4 16:38 building//
drwxrwxr-x 5 br br 4096 авг 10 12:39 dev//
drwxrwxr-x 6 br br 4096 июн  4 16:38 f2py//
drwxrwxr-x 7 br br 4096 авг 10 12:39 reference//
drwxrwxr-x 2 br br 4096 авг 10 12:39 release//
drwxrwxr-x 5 br br 4096 авг 10 12:39 _static//
[...]

PR #29547, opened 2025-08-12
DOC: doctest ReST sources in `doc/source/reference`

ev-br:
Prompted by https://github.com/numpy/numpy/issues/29544 : Add doctesting of ReST sources in `doc/source/reference`, which were never doctested IIUC.

This is a minimal fix. Several larger-scale potential improvements are discussed in gh-29544.

### DIFF
--- a/doc/source/reference/arrays.promotion.rst
+++ b/doc/source/reference/arrays.promotion.rst
@@ -114,9 +114,20 @@ cannot be coerced to the NumPy dtype, an error is raised:
   Traceback (most recent call last):
   ...
   OverflowError: Python int too large to convert to C long
+
+  For floating-point values, however, NumPy returns an infinity and emits a warning:
+
   >>> np.float32(1) + 1e300
+  Traceback (most recent call last):
+  ...
+  RuntimeWarning: overflow encountered in cast
+
+  >>> import warnings
+  >>> with warnings.catch_warnings():
+  ...     warnings.simplefilter('ignore')
+  ...     x = np.float32(1) + 1e300
+  >>> x
   np.float32(inf)
-  ... RuntimeWarning: overflow encountered in cast
 
 Second, since the Python float or integer precision is always ignored, a low
 precision NumPy scalar will keep using its lower precision unless explicitly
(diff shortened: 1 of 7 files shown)


mattip:
LGTM, I guess we know it is working since there is a failing test
```
=================================== FAILURES ===================================
___________________________ [doctest] array_api.rst ____________________________
039 
040 
041 Entry point
042 ===========
043 
044 NumPy installs an `entry point <https://packaging.python.org/en/latest/specifications/entry-points/>`__
045 that can be used for discovery purposes::
046 
047     >>> from importlib.metadata import entry_points
048     >>> entry_points(group='array_api', name='numpy')
Expected:
    (EntryPoint(name='numpy', value='numpy', group='array_api'),)
Got:
    [EntryPoint(name='numpy', value='numpy', group='array_api')]

/home/circleci/repo/doc/source/reference/array_api.rst:48: DocTestFailure
```

Using `[skip actions][skip azp][skip cirrus]` would save some CI cycles

ev-br:
A proof by a CI failure, as it were :-).

Funny enough, the `EntryPoint` thing which fails on CI fails locally for me otherwise. So the type depends on versions of something. Am changing it back to whatever passes on CI.

> Using [skip actions][skip azp][skip cirrus] would save some CI cycles

Yup, thanks. I keep forgetting the exact syntax in different projects. Maybe it's something to add to the PR template? For me (based on statistics of one!) it'd help to have a link to the relevant section of the docs.

doc/source/reference/arrays.promotion.rst
```diff
@@ -114,9 +114,20 @@ cannot be coerced to the NumPy dtype, an error is raised:
   Traceback (most recent call last):
   ...
   OverflowError: Python int too large to convert to C long
+
+  For floating-point values, however, NumPy returns an infinity and emits a warning:
+
   >>> np.float32(1) + 1e300
+  Traceback (most recent call last):
+  ...
+  RuntimeWarning: overflow encountered in cast
```
seberg, line 123: Hmmm, this seems a tad annoying?  Can we change warning settings for doctests to *not* raise?  I think just printing the warning as it was is nicer, because that is what users see!

seberg (commented):
LGTM, nice to add things and fixup the docs.

But I wonder what your opion is on those warnings, or if we can't find an easy fix?
I love running tests with all warnings being turned to errors, but for doc tests this is actually making the documentation worse to bend it to an environment nobody uses in practice.

doc/source/reference/arrays.promotion.rst
mattip: I don't see a way in doctest to ignore the traceback of a warning.

seberg: We don't have to disable the traceback.  We have to modify `conftest.py` so that this warning is shown normally instead of being raised.

The problem is that the doctests are run in the testing environment which defaults to raising all warnings (at least in development mode).
The only reason this docs didn't pass the doctests is that the doctests don't match what a normal user would be running.

PR #29531, opened 2025-08-07
TYP: Type default values in stubs in `numpy/ma`

MarcoGorelli:
Making some progress towards https://github.com/numpy/numpy/issues/28428

Similar PR in pandas: https://github.com/pandas-dev/pandas-stubs/pull/1293

I've done this based on work started in https://gist.github.com/yangdanny97/170f82ee5389584f8b6292bc4ea9c24d, we're looking at open-sourcing a reusable tool to do this automatically where possible:
- if a stub file uses `= ...`
- and the corresponding defintion has a simple default
- then fill the `...` in

The ones in this PR, I checked manually, and they look correct to me

PR #29548, opened 2025-08-12
MAINT: Use double quotes (ruff rule `Q`) (only on `.pyi` files)

MarcoGorelli:
Double-quotes came up in a review: https://github.com/numpy/numpy/pull/29531#discussion_r2267800014

If it's a preferred style, then I'd suggest it be automated

The only manual change I did here was to add `Q` to `ruff.toml`, everything else was just `spin lint --fix`


mattip:
I prefer to play loose and free with either single- or double- quotes, but will play along if that is the community decision. It seems a bit disruptive to reformat everything though.

jorenham:
I'm all for enforcing this, but I can imagine that this 348 file diff could cause merge conflicts in many of the open PR's. So maybe we can limit this to e.g. `.pyi` or certain directories for now?

MarcoGorelli:
Sure, point taken, thanks - I've updated so it only applies to the `.pyi` files

jorenham (approved):

jorenham:
> I prefer to play loose and free with either single- or double- quotes

Out of curiosity; any particular reason for that?

mattip:
> Out of curiosity; any particular reason for that?

Double quotes are harder to type (two key presses) and somehow visually heavier and less appealing. So I usually use single quotes when it is a single word or character and double quotes when I truly want to make a statement.

jorenham:
> Double quotes are harder to type (two key presses) and somehow visually heavier and less appealing. So I usually use single quotes when it is a single word or character and double quotes when I truly want to make a statement.

Yea I guess that makes sense. I'm also fine with using `'` quotes everywhere. I suppose that enforcing one or the other is more about consistency, which can help reduce diff sizes, and avoids you from having to *think* about which one to use. So theoretically it doesn't matter which one we choose. 

But note that the `"` choice is a lot more common (in my experience), probably because of `black`. So for those used to the black style (like me), the chance of accidentally using the wrong quotes is higher if we go with the less common `'`, which could lead to more CI usage. But who knows, perhaps that won't even be noticeable in practice 🤷🏻.

charris:
I have a slight preference for single quotes, but have been trying to move to double quotes because of Black. That said, it is easy enough to run ruff to fix things up.

MarcoGorelli:
> it is easy enough to run ruff to fix things up

Yes, every project I contribute to has some expectation that you'll run `make lint` or `spin lint` or `pre-commit run -a` or some other command that does these automated things.

In other projects I just type single quotes (as @mattip observed, they're easier to type) and then let the auto-formatter sort them out according to the project's preferences

I'd appreciate it if this rule could at least be enabled for the stubs anyway, so that quotes can be consistent and neither reviewers nor contributors need think about them 🙏

seberg:
Seems like we can just put this in? FWIW, would be nice to add the ruff rule, but not usre how easy it is just for the stubs.
(I don't mind just doing more/full auto-formatting, we break all PRs once of course, although you could probably `ruff --fix` and then merge most.)

MarcoGorelli:
Thanks!

> but not usre how easy it is just for the stubs.

That's what this PR does, it enables the rule just for the stubs 😄 Check the changes to `ruff.toml`

jorenham:
And in it goes; thanks Marco :)

seberg:
Hah sorrysomehow I looked and missed the toml changes...

mattip:
> ... quotes can be consistent and neither reviewers nor contributors need think about them 🙏

Quotes can be inconsistent and neither contributors nor reviewers need to think about them as long as there is no syntax error. We could add a note to the development process to state "use whatever quotes you want, quoting will not be something reviewers will complain about". IMO it is a shame black ever went down the route of trying to standardize that.

PR #29549, opened 2025-08-12
BUG: SIMD Neon undefined behavior - pointer overflow

m-clare:
- Addresses UBSAN warnings for `simd/neon/memory.h` by adding bounds checking on stride size - (related to #29524)
- To resolve this, I looked at some of the other SIMD architectures, and used the same kind of assertion as that implemented in `avx2/memory.h` (also in `avx512`)

```
NPY_FINLINE npyv_u32 npyv_loadn_u32(const npy_uint32 *ptr, npy_intp stride)
{
    assert(llabs(stride) <= NPY_SIMD_MAXLOAD_STRIDE32);
    const __m256i steps = _mm256_setr_epi32(0, 1, 2, 3, 4, 5, 6, 7);
    const __m256i idx = _mm256_mullo_epi32(_mm256_set1_epi32((int)stride), steps);
    return _mm256_i32gather_epi32((const int*)ptr, idx, 4);
}
```

While looking at other architecture options for the same function `npyv_loadn_u32`, I noticed that `sse`, `lsx`, `vec` do not have a bounds check on the stride, but were not flagged by UBSAN (they all use array indexing instead of pointer arithmetic). For example for `sse/memory.h` L79:
```
NPY_FINLINE npyv_s32 npyv_loadn_s32(const npy_int32 *ptr, npy_intp stride)
{
    __m128i a = _mm_cvtsi32_si128(*ptr);
#ifdef NPY_HAVE_SSE41
    a = _mm_insert_epi32(a, ptr[stride],   1);
    a = _mm_insert_epi32(a, ptr[stride*2], 2);
    a = _mm_insert_epi32(a, ptr[stride*3], 3);
#else
    __m128i a1 = _mm_cvtsi32_si128(ptr[stride]);
    __m128i a2 = _mm_cvtsi32_si128(ptr[stride*2]);
    __m128i a3 = _mm_cvtsi32_si128(ptr[stride*3]);
    a = _mm_unpacklo_epi32(a, a1);
    a = _mm_unpacklo_epi64(a, _mm_unpacklo_epi32(a2, a3));
#endif
    return a;
}
```
Should these other implementations also include bounds checks?

### DIFF
--- a/numpy/_core/src/common/simd/neon/memory.h
+++ b/numpy/_core/src/common/simd/neon/memory.h
@@ -52,6 +52,7 @@ NPYV_IMPL_NEON_MEM(f64, double)
  ***************************/
 NPY_FINLINE npyv_s32 npyv_loadn_s32(const npy_int32 *ptr, npy_intp stride)
 {
+    assert(llabs(stride) <= NPY_SIMD_MAXLOAD_STRIDE32);
     int32x4_t a = vdupq_n_s32(0);
     a = vld1q_lane_s32((const int32_t*)ptr,            a, 0);
     a = vld1q_lane_s32((const int32_t*)ptr + stride,   a, 1);
@@ -75,6 +76,7 @@ NPY_FINLINE npyv_f32 npyv_loadn_f32(const float *ptr, npy_intp stride)
 //// 64
 NPY_FINLINE npyv_s64 npyv_loadn_s64(const npy_int64 *ptr, npy_intp stride)
 {
+    assert(llabs(stride) <= NPY_SIMD_MAXLOAD_STRIDE64);
     return vcombine_s64(
         vld1_s64((const int64_t*)ptr), vld1_s64((const int64_t*)ptr + stride)
     );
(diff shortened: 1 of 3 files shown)


seberg:
@seiko2plus would know a bit better.  I don't quite understand what UBSANs problem with this is. There isn't an actual pointer overflow here, just some theoretical possibility (i.e. UBSAN wasn't able to infer/figure out that the value is always within range)?

Anyway, let's add the assert ideally with a comment to UBSAN.  This code must assume that all data it is asked to load is valid to load!

But I don't think we should define `NPY_SIMD_MAXLOAD_STRIDE64`, since that would mean there is a limitation of what can be loaded, which isn't the case (and that should even add a runtime check).
(I suppose you could always define it to something that can clearly load everything and use that in the definitions.)

m-clare:
@seberg Alternatively, if UBSAN is being overzealous here, I can just mark the suppression as one that should remain in the file.

seberg:
True also fine, but I am happy with the assert solution.  I just don't think adding the `#define` is quite ideal as that carries a meaning beyond what is needed here.

PR #29322, opened 2025-07-04
MAINT: remove internal uses of assert_warns and suppress_warnings

ngoldbaum:
Towards fixing #29293, also see https://github.com/scipy/scipy/pull/23275 where I did this in SciPy.

The only nontrivial changes (IMO, anyway) are in `numpy/ma/tests/test_core.py`.

PR #29550, opened 2025-08-12
DEP: Deprecate NumPy warning control utilities

### LINKED ISSUE #29293: ENH: make np.testing.suppress_warnings thread-safe
Python 3.14 makes the `warnings` module thread-safe. While testing whether we can globally allow tests that use `warnings` to run on 3.14 under pytest-run-parallel, I discovered that SciPy is making heavy use of `np.testing.supress_warnings`, and it isn't thread-safe:

```
PARALLEL FAILED scipy/cluster/tests/test_vq.py::TestKMeans::test_kmeans_lost_cluster[numpy] - AttributeError: 'suppress_warnings' object has no attribute '_orig_show'
```

(from logs of a SciPy test run)

The relevant SciPy test is here:

https://github.com/scipy/scipy/blob/8f51c63d384f0e9c966355362f982ea7ec2de1cc/scipy/cluster/tests/test_vq.py#L276

I'm not totally sure why the failure is happening - I would think each test instance would have its own private context manager instance, but the error I'm seeing makes it seem like threads are sharing the context manager.

In any case, it's clearly thread-unsafe and relies on manipulating global state stored in the `warnings` module:

https://github.com/numpy/numpy/blob/e740470a94bf7b15a5dd981d76a802876a1de56c/numpy/testing/_private/utils.py#L2400-L2402

The whole thing should probably be implemented using a context variable instead of manipulating / relying on global state on the `warnings` module.

ngoldbaum:
fixes #29293

Deprecates the following public API:

* `np.testing.assert_warns`
* `np.testing.suppress_warnings`

These solve problems that no longer exist in modern Python versions. As of Python 3.14, where warnings can be made thread-safe, they're actively harmful because the implementation is not thread-safe. We should deprecate these helpers and point people at the standard library or pytest.

`assert_warns` and `suppress_warnings` used to be used heavily in NumPy and SciPy, but https://github.com/numpy/numpy/pull/29322/ cleaned up NumPy and https://github.com/scipy/scipy/pull/23275 cleaned up SciPy. I don't think there are any substantial uses in active open source codebases besides NumPy and SciPy, judging by Github code searches.

I marked this for 2.5 but maybe we should backport it to the 2.4 branch?

I had to adjust some of the tests for `suppress_warnings` slightly, because now a new warning is getting generated.

I also used the `skip_file_prefixes` keyword for `warnings.warn`, which is only available on 3.12 and newer. Do I need to care about 3.11 support to make the warning context nice? There's no single `stacklevel` I can set.

An earlier version of this PR also deprecated `clear_and_suppress_warnings` and `assert_no_warnings`, but I ran into some issues getting the tests working again with the deprecation and I don't actually need to deprecate them, so I punted.

### DIFF
--- a/numpy/_core/tests/test_deprecations.py
+++ b/numpy/_core/tests/test_deprecations.py
@@ -516,3 +516,36 @@ class TestFlatiterIndexingFloatIndex(_DeprecationTestCase):
             arr.flat[[1.]] = 10
 
         self.assert_deprecated(assign_to_index)
+
+
+class TestWarningUtilityDeprecations(_DeprecationTestCase):
+    # Deprecation in NumPy 2.5
+    message = r"NumPy warning suppression and assertion utilities are deprecated."
+
+    def test_assert_warns_deprecated(self):
+        def use_assert_warns():
+            with np.testing.assert_warns(RuntimeWarning):
+                warnings.warn("foo", RuntimeWarning)
+
+        self.assert_deprecated(use_assert_warns)
+
+    def test_assert_no_warns_deprecated(self):
+        def use_assert_no_warns():
+            with np.testing.assert_no_warnings():
+                pass
+
+        self.assert_deprecated(use_assert_no_warns)
+
+    def test_suppress_warnings_deprecated(self):
+        def use_suppress_warnings():
+            with np.testing.suppress_warnings() as sup:
+                sup.filter(RuntimeWarning, 'invalid value encountered in divide')
+
+        self.assert_deprecated(use_suppress_warnings)
+
+    def test_clear_and_catch_warnings_deprecated(self):
+        def use_clear_and_catch_warnings():
+            with np.testing.clear_and_catch_warnings():
+                pass
+
+        self.assert_deprecated(use_clear_and_catch_warnings)
--- a/numpy/_core/tests/test_deprecations.py
+++ b/numpy/_core/tests/test_deprecations.py
@@ -536,9 +536,3 @@ class TestWarningUtilityDeprecations(_DeprecationTestCase):
 
         self.assert_deprecated(use_suppress_warnings)
 
-    def test_clear_and_catch_warnings_deprecated(self):
-        def use_clear_and_catch_warnings():
-            with np.testing.clear_and_catch_warnings():
-                pass
-
-        self.assert_deprecated(use_clear_and_catch_warnings)
--- a/numpy/_core/tests/test_deprecations.py
+++ b/numpy/_core/tests/test_deprecations.py
@@ -535,4 +535,3 @@ class TestWarningUtilityDeprecations(_DeprecationTestCase):
                 sup.filter(RuntimeWarning, 'invalid value encountered in divide')
 
         self.assert_deprecated(use_suppress_warnings)
-
--- a/numpy/_core/tests/test_deprecations.py
+++ b/numpy/_core/tests/test_deprecations.py
@@ -529,13 +529,6 @@ class TestWarningUtilityDeprecations(_DeprecationTestCase):
 
         self.assert_deprecated(use_assert_warns)
 
-    def test_assert_no_warns_deprecated(self):
-        def use_assert_no_warns():
-            with np.testing.assert_no_warnings():
-                pass
-
-        self.assert_deprecated(use_assert_no_warns)
-
     def test_suppress_warnings_deprecated(self):
         def use_suppress_warnings():
             with np.testing.suppress_warnings() as sup:
--- a/numpy/_core/tests/test_deprecations.py
+++ b/numpy/_core/tests/test_deprecations.py
@@ -519,7 +519,7 @@ class TestFlatiterIndexingFloatIndex(_DeprecationTestCase):
 
 
 class TestWarningUtilityDeprecations(_DeprecationTestCase):
-    # Deprecation in NumPy 2.5
+    # Deprecation in NumPy 2.5, 2025-08
     message = r"NumPy warning suppression and assertion utilities are deprecated."
 
     def test_assert_warns_deprecated(self):
--- a/numpy/_core/tests/test_deprecations.py
+++ b/numpy/_core/tests/test_deprecations.py
@@ -525,7 +525,7 @@ class TestWarningUtilityDeprecations(_DeprecationTestCase):
     def test_assert_warns_deprecated(self):
         def use_assert_warns():
             with np.testing.assert_warns(RuntimeWarning):
-                warnings.warn("foo", RuntimeWarning)
+                warnings.warn("foo", RuntimeWarning, stacklevel=1)
 
         self.assert_deprecated(use_assert_warns)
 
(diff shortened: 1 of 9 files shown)


numpy/testing/_private/utils.py
jorenham, line 1: Could you also mark these functions as `@deprecated` in the [stubs](https://github.com/numpy/numpy/blob/73d45140f2f1acd94f01cb7dd02352296ded7940/numpy/testing/_private/utils.pyi)? See https://github.com/numpy/numpy/blob/73d45140f2f1acd94f01cb7dd02352296ded7940/numpy/fft/helper.pyi#L1-L14 for an example.

jorenham (approved):

numpy/testing/_private/utils.py
```diff
@@ -56,6 +56,10 @@ class KnownFailureException(Exception):
 
 NUMPY_ROOT = pathlib.Path(np.__file__).parent
 
+WARNING_KWARGS = {}
+if sys.version_info >= (3, 12):
+    WARNING_KWARGS['skip_file_prefixes'] = (__file__, contextlib.__file__)
```
ngoldbaum: this is a little gross - do we still care about Python 3.11 support?

mattip: I think we should only drop it after we release 2.4

numpy/tests/test_warnings.py
```diff
@@ -53,6 +53,7 @@ def visit_Call(self, node):
             args = {kw.arg for kw in node.keywords}
             if "stacklevel" in args:
                 return
+            breakpoint()
```
tylerjereddy: maybe this is on purpose, but it would be the first such case in the entire code base, so maybe a leftover?

numpy/typing/tests/data/pass/ndarray_misc.py
```diff
@@ -190,11 +192,11 @@ class IntSubClass(npt.NDArray[np.intp]): ...
...
+with pytest.warns(DeprecationWarning):
     ctypes_obj.get_data()  # type: ignore[deprecated]  # pyright: ignore[reportDeprecated]
-with np.testing.assert_warns(DeprecationWarning):
+with pytest.warns(DeprecationWarning):
     ctypes_obj.get_shape()  # type: ignore[deprecated]  # pyright: ignore[reportDeprecated]
-with np.testing.assert_warns(DeprecationWarning):
+with pytest.warns(DeprecationWarning):
     ctypes_obj.get_strides()  # type: ignore[deprecated]  # pyright: ignore[reportDeprecated]
-with np.testing.assert_warns(DeprecationWarning):
+with pytest.warns(DeprecationWarning):
```
tylerjereddy: there's also `with pytest.deprecated_call()`, probably doesn't matter

numpy/testing/_private/utils.py
seberg: If you hate it too much, I don't care to just not do it here...  But since it may help someone figure out what to change quicker in their tests, seems fine.

numpy/tests/test_warnings.py
seberg:
```suggestion
```
I'll just apply the deletion.

seberg (approved):
LGTM, thanks.  Yeah, I hope both isn't used too much, with pytest being long there.

I remember seeing `suppress_warnings()` in the wild in places where I didn't want to see it, but I suspect that is largely fixed (and I never wanted to see it there ;)).

But no backporting.  This clearly isn't a bug fix, adding deprecation warnings and thus breaking someones CI in a bug fix release isn't acceptable (at least not without a clear argument that this is such a big deal that breaking peoples CI when they don't expect it is clearly worth it.).

numpy/testing/_private/utils.py
ngoldbaum: We'll drop Python 3.11 support before this code gets released to users, so this can get cleaned up along with other code that checks for `sys.version_info >= (3, 12)`. So I'll leave this as-is, I think.

charris: > We'll drop Python 3.11 support before this code gets released 

We will drop Python 3.11 in 2.5.

ngoldbaum: If this is merged in `main` right now, it won't be released until NumPy 2.5, right? That's what I was trying to say.

charris: 2.4 is the next release, we will branch sometime in November and release in December, 2.5 will be released next June.

ngoldbaum: Oh darn. Well in that case I'll take a look to see if there's a more complicated way to get nice warning contexts on 3.11. I suspect there isn't because the call stack might or might not include contextlib depending on the usage pattern. If that's the case I'll leave this as-is.

ngoldbaum: It looks like duplicating the logic and using `stacklevel` does work, so I went with that. See latest commit.

numpy/_core/tests/test_deprecations.py
```diff
@@ -516,3 +516,22 @@ def assign_to_index():
             arr.flat[[1.]] = 10
 
         self.assert_deprecated(assign_to_index)
+
+
+class TestWarningUtilityDeprecations(_DeprecationTestCase):
+    # Deprecation in NumPy 2.5, 2025-08
```
seberg:
```suggestion
    # Deprecation in NumPy 2.4, 2025-08
```

numpy/testing/_private/utils.py
```diff
@@ -2036,6 +2041,10 @@ def assert_warns(warning_class, *args, **kwargs):
     >>> ret = np.testing.assert_warns(DeprecationWarning, deprecated_func, 4)
     >>> assert ret == 16
     """
+    warnings.warn(
+        "NumPy warning suppression and assertion utilities are deprecated. "
+        "Use warnings.catch_warnings, warnings.filterwarnings, pytest.warns, "
+        "or pytest.filterwarnings instead.", DeprecationWarning, stacklevel=2)
```
seberg:
```suggestion
        "or pytest.filterwarnings instead. (Deprecated NumPy 2.4)",
        DeprecationWarning, stacklevel=2)
```

numpy/testing/_private/utils.py
```diff
@@ -2348,7 +2362,12 @@ def some_function():
             # do something which causes a warning in np.ma.core
             pass
     """
-    def __init__(self, forwarding_rule="always"):
+    def __init__(self, forwarding_rule="always", _warn=True):
+        if _warn:
+            warnings.warn(
+                "NumPy warning suppression and assertion utilities are deprecated. "
+                "Use warnings.catch_warnings, warnings.filterwarnings, pytest.warns, "
+                "or pytest.filterwarnings instead.", DeprecationWarning, stacklevel=2)
```
seberg:
```suggestion
                "or pytest.filterwarnings instead. (Deprecated NumPy 2.4)",
                DeprecationWarning, stacklevel=2)
```

seberg (approved):
Hah, good Chuck noticed it said 2.5.  Would be nice to also include the version in the message.
And thanks for fixing the stacklevel, that new file skipping feature is neat, but since it works stacklevel seems nicer!

Otherwise, please just go ahead and self-merge.

ngoldbaum:
Test failures are transient network issues. Merging because the last commit just changed some error messages and is syntactically correct.

PR #29551, opened 2025-08-12
BUG: resolve invalid grep with env neutral script

### LINKED ISSUE #29538: BUG: grep: Invalid option -- P when running `spin lint` on macOS
### Describe the issue:

When running spin lint on macOS, the command fails with the following error.
It seems that the grep command on macOS (BSD grep) does not support the -P option (Perl-compatible regex), which is available on GNU grep. This causes the linting step to break out of the box.

https://github.com/Lightning-AI/pytorch-lightning/issues/19938 is a similar issue.

```python
$ spin lint
```

### Error message:

```shell
Running Ruff Check...
All checks passed!

Running C API borrow-reference linter...
grep: invalid option -- P
usage: grep [-abcdDEFGHhIiJLlMmnOopqRSsUVvwXxZz] [-A num] [-B num] [-C[num]]
        [-e pattern] [-f file] [--binary-files=value] [--color=when]
        [--context[=num]] [--directories=action] [--label] [--line-buffered]
        [--null] [pattern] [file ...]
grep: invalid option -- P
usage: grep [-abcdDEFGHhIiJLlMmnOopqRSsUVvwXxZz] [-A num] [-B num] [-C[num]]
        [-e pattern] [-f file] [--binary-files=value] [--color=when]
        [--context[=num]] [--directories=action] [--label] [--line-buffered]
        [--null] [pattern] [file ...]
...
```

commit hash: [622f874](https://github.com/numpy/numpy/commit/622f874ac343f9d2df39806e49881e3115060be5)

### Runtime Environment:

```bash
$ grep --version
grep (BSD grep, GNU compatible) 2.6.0-FreeBSD
```

### Context for the issue:

_No response_

lvllvl:
Attempts to resolve #29538


tools/ci/check_c_api_usage.py
```diff
@@ -0,0 +1,211 @@
...
+    "PySequence_Fast"
+)
+
+# Match any function as a standalone C identifier: (?<!\w)(NAME)(?!\w)
+FUNC_RX = re.compile(r"(?<!\w)(?:"
+                     + "|".join(map(re.escape, SUSPICIOUS_FUNCS))
+                     + r")(?!\w)")
+
+NOQA_OK = "noqa: borrowed-ref OK"
+NOQA_MANUAL = "noqa: borrowed-ref - manual fix needed"
```
tylerjereddy: I suppose it is minor, but some of these module-level variables get used directly in functions without passing them in, which is a bit global-like. I wonder if it might make sense to pass them through. I guess it matters less for non-library code--I might be slightly tempted to just pass them through `main` and to the function that uses them from there.

tools/ci/check_c_api_usage.py
```diff
@@ -0,0 +1,211 @@
...
+    )
+    args = ap.parse_args(argv)
+
+    root = Path(args.root)
+    if not root.exists():
+        print(f"error: root '{root}' does not exist", file=sys.stderr)
+        return 2
+
+    exts = set(args.ext) if args.ext else set(DEFAULT_EXTS)
+    excludes = set(args.exclude) if args.exclude else set(DEFAULT_EXCLUDES)
```
tylerjereddy: these `DEFAULT_..` vars are already `set`s, right?

tools/ci/check_c_api_usage.py
```diff
@@ -0,0 +1,211 @@
...
+
+    # Output file (mirrors your shell behavior)
+    tmpdir = Path(".tmp")
+    tmpdir.mkdir(exist_ok=True)
+    fd, outpath = tempfile.mkstemp(
+        prefix="c_api_usage_report.",
+        suffix=".txt",
+        dir=tmpdir
+        )
+    os.close(fd)
```
tylerjereddy: I suppose we could try to use a context manager to avoid the manual resource handling

tools/ci/check_c_api_usage.py
```diff
@@ -0,0 +1,211 @@
...
+        "--root",
+        default="numpy",
+        type=str,
+        help="Root directory to scan (default: numpy)"
+        )
+    ap.add_argument(
+        "--ext",
+        action="append",
+        default=None,
+        help="File extension(s) to include (repeatable). Defaults to .c,.h,.c.src,.cpp",
```
tylerjereddy: I wonder if it would make sense to just substitute the variable in directly here to avoid hardcoding the defaults in two separate locations, i.e.,

```diff
--- a/tools/ci/check_c_api_usage.py
+++ b/tools/ci/check_c_api_usage.py
@@ -148,7 +148,7 @@ def main(argv: list[str] | None = None) -> int:
         "--ext",
         action="append",
         default=None,
-        help="File extension(s) to include (repeatable). Defaults to .c,.h,.c.src,.cpp",
+        help=f"File extension(s) to include (repeatable). Defaults to {DEFAULT_EXTS}",
```

tools/linter.py
```diff
@@ -50,9 +50,9 @@ def run_check_c_api(self) -> tuple[int, str]:
         # Running borrowed ref checker
         print("Running C API borrow-reference linter...")
         borrowed_ref_script = os.path.join(self.repository_root, "tools", "ci",
-                                           "check_c_api_usage.sh")
+                                           "check_c_api_usage.py")
         borrowed_res = subprocess.run(
-            ["bash", borrowed_ref_script],
+            ["python3", borrowed_ref_script],
```
tylerjereddy: Would it be slightly safer to use `sys.executable` here? I'm reminded of a recent discussion (https://github.com/scipy/scipy/issues/23334) about preferring that in general with `spin`.

tools/ci/check_c_api_usage.py
```diff
@@ -0,0 +1,211 @@
+#!/usr/bin/env python3
+from __future__ import annotations
```
tylerjereddy, line 2: Is this just needed in case someone is using an older Python version than we normally support outside of linting?

tylerjereddy (commented):
This does seem to fix the issue described in the ticket for me locally on (ARM) Mac.

I added a few small comments in case they help.

tools/ci/check_c_api_usage.py
lvllvl: Yeah, I think intiially the issue popped up because I didn't think about a certain environment where this script may be used. So I thought maybe this would be helpful.

tools/ci/check_c_api_usage.py
lvllvl: ok I'll move these global variables to the main function.

tools/ci/check_c_api_usage.py
lvllvl: yes, I'll adjust these as well. Currently these are global vars but I'll bring them into main.

lvllvl: yes, I'll adjust these as well. Currently these are global vars but I'll bring them into main.

tools/ci/check_c_api_usage.py
lvllvl: ok making this change.

tools/ci/check_c_api_usage.py
lvllvl: I am open to removing this. I don't know if my initial assumption is correct or appropriate for the codebase. But lmk what your preference is here.

tools/linter.py
lvllvl: ok added this `sys.executable` in the script

lvllvl:
Hi @tylerjereddy, I made a few changes based on your comments. Please let me know if any other changes are needed before I can merge this PR.

lvllvl:
hi @tylerjereddy , I just wanted to follow up. do you think this PR needs any other changes? I was able to make changes based on your feedback. Please let me know what you think.

tools/ci/check_c_api_usage.py
```diff
@@ -0,0 +1,222 @@
...
+
+    if args.ext:
+        exts = {e if e.startswith(".") else f".{e}" for e in args.ext}
+    else:
+        exts = set(default_exts)
+    excludes = set(args.exclude) if args.exclude else set(default_excludes)
+
+    root = Path(args.root)
+    if not root.exists():
+        print(f"error: root '{root}' does not exist", file=sys.stderr)
```
djeada, line 185: hi, just something that caught my eye: why are there print statements in production code? isn't logging preferred?

tools/ci/check_c_api_usage.py
```diff
@@ -0,0 +1,222 @@
...
+    )
+    func_rx = build_func_rx(suspicious_funcs)
+    noqa_markers = (
+        "noqa: borrowed-ref OK",
+        "noqa: borrowed-ref - manual fix needed"
+        )
+    default_exts = {".c", ".h", ".c.src", ".cpp"}
+    default_excludes = {"pythoncapi-compat"}
+
+    ap = argparse.ArgumentParser(description="Borrow-ref C API linter (Python).")
```
djeada, line 144: neat: I would add `--dry-run` or `--quiet` flag. For CI use, usually all that is needed is the exit status.

tools/ci/check_c_api_usage.py
```diff
@@ -0,0 +1,222 @@
...
+        prefix="c_api_usage_report.",
+        suffix=".txt",
+        dir=tmpdir,
+        mode="w+",
+        encoding="utf-8",
+        delete=False,
+        ) as out:
+        report_path = Path(out.name)
+        out.write("Running Suspicious C API usage report workflow...\n\n")
+        for p in files:
```
djeada: i would replace the sequential `for p in files:` loop with parallel execution, e.g., using `concurrent.futures.ThreadPoolExecutor` to run `scan_file` on each file concurrently

tools/ci/check_c_api_usage.py
lvllvl: ok I added a `--quiet` flag

tools/ci/check_c_api_usage.py
lvllvl: yeah that's a good point. I just updated this and will push shortly

ngoldbaum:
Sorry for the long delay in looking at this. I think this looks great at a high level. I'm just going to put this PR through its paces a little locally by intentionally introducing some issues that the linter should catch. If that works I'll go ahead and merge it.

tools/ci/check_c_api_usage.py
```diff
@@ -0,0 +1,280 @@
...
+    with tempfile.NamedTemporaryFile(
+        prefix="c_api_usage_report.",
+        suffix=".txt",
+        dir=tmpdir,
+        mode="w+",
+        encoding="utf-8",
+        delete=False,
+        ) as out:
+        report_path = Path(out.name)
+        out.write("Running Suspicious C API usage report workflow...\n\n")
```
ngoldbaum, line 237: The terminal output for `spin lint` when I run this locally mixes direct writes to stdout and logging. Why use both? IMO it overcomplicates this to use logging so I'd stick to the pattern you have here for all the reporting.

tools/ci/check_c_api_usage.py
```diff
@@ -0,0 +1,280 @@
...
+                           p for p in files}
+            for fut in as_completed(fut_to_file):
+                try:
+                    all_hits.extend(fut.result())
+                except Exception as e:
+                    log.warning("Failed to scan %s: %s", fut_to_file[fut], e)
+    # Sort for deterministic output: by path, then line number
+    all_hits.sort(key=lambda t: (t[2], t[1]))
+
+    with tempfile.NamedTemporaryFile(
```
ngoldbaum, line 228: IMO you should only try to create this file if the linter failed, otherwise successful linter runs have noise in their output about writing this file to disk, but the linter ran successfully so it's very unlikely that anyone cares.

ngoldbaum (commented):
Seems to work. I have some quibbles about the output of the script that I'd like addressed before merging. Again, sorry for the delay in looking at this.

ngoldbaum:
Also if the linter fails IMO you should write a summary line at the very end saying the borrowed reference checker failed. Right now it's a little ambiguous if the checker failed or not, since the last message is printed with an INFO log level:

```
goldbaum at Nathans-MBP in ~/Documents/numpy on BUG-29538!
± spin lint
Running Ruff Check...
All checks passed!

Running C API borrow-reference linter...
INFO: Scanning 549 C/C++ source files...
Running Suspicious C API usage report workflow...

Found suspicious call to PyDict_GetItemWithError in file: numpy/_core/src/multiarray/convert_datatype.c
 -> numpy/_core/src/multiarray/convert_datatype.c:150a:    res = PyDict_GetItemWithError(NPY_DT_SLOTS(from)->castingimpls, (PyObject *)to);
Recommendation:
If this use is intentional and safe, add '// noqa: borrowed-ref OK' on the same line to silence this warning.
Otherwise, consider replacing the call with a thread-safe API function.

INFO: Report written to: /Users/goldbaum/Documents/numpy/.tmp/c_api_usage_report.e6h1cod_.txt

goldbaum at Nathans-MBP in ~/Documents/numpy on BUG-29538!
± echo $?
1
```

Here I'm using a linter error I manually inserted into the codebase.

djeada (commented):
lgtm, thanks

lvllvl:
> Also if the linter fails IMO you should write a summary line at the very end saying the borrowed reference checker failed. Right now it's a little ambiguous if the checker failed or not, since the last message is printed with an INFO log level:
> 
> ```
> goldbaum at Nathans-MBP in ~/Documents/numpy on BUG-29538!
> ± spin lint
> Running Ruff Check...
> All checks passed!
> 
> Running C API borrow-reference linter...
> INFO: Scanning 549 C/C++ source files...
> Running Suspicious C API usage report workflow...
> 
> Found suspicious call to PyDict_GetItemWithError in file: numpy/_core/src/multiarray/convert_datatype.c
>  -> numpy/_core/src/multiarray/convert_datatype.c:150a:    res = PyDict_GetItemWithError(NPY_DT_SLOTS(from)->castingimpls, (PyObject *)to);
> Recommendation:
> If this use is intentional and safe, add '// noqa: borrowed-ref OK' on the same line to silence this warning.
> Otherwise, consider replacing the call with a thread-safe API function.
> 
> INFO: Report written to: /Users/goldbaum/Documents/numpy/.tmp/c_api_usage_report.e6h1cod_.txt
> 
> goldbaum at Nathans-MBP in ~/Documents/numpy on BUG-29538!
> ± echo $?
> 1
> ```
> 
> Here I'm using a linter error I manually inserted into the codebase.

Ok I was able to add a failure message at the end here

tools/ci/check_c_api_usage.py
lvllvl: ok so if I understood you correctly, you want the script to reflect the pattern I used before, which was mostly printing to the console. I reverted back to not using logging in this script. I have a log file on failure, but that's about it.

tools/ci/check_c_api_usage.py
lvllvl: Ok made the adjustment to only create this log file if and only if `all_hits` has any findings (i.e., failures).

ngoldbaum:
Thanks for your persistence on this!

lvllvl:
> Thanks for your persistence on this!

Thanks for the feedback everyone

PR #29555, opened 2025-08-13
BUG: Fix metadata not roundtripping when pickling datetime

### LINKED ISSUE #29344: BUG: Pickling/unpickling datetime64 adds dtype metadata
### Describe the issue:

I have an array with dtype datetime64. The dtype.metadata is None. After serializing the array with pickle dump/load, the dtype.metadata is some object "mappingproxy({})".

This is a problem because trying to np.save the array with dtype.metadata produces a UserWarning.

dtype.metadata appears to be some kind of undocumented feature(???) I've never heard of before now.

```python
import numpy as np
import pickle
import io

a = np.asarray([np.datetime64('2000-01-01')])
print(f'{a = }')
print(f'{a.dtype = }')
print(f'{a.dtype.metadata = }')

b = pickle.loads(pickle.dumps(a))
print(f'{b = }')
print(f'{b.dtype = }')
print(f'{b.dtype.metadata = }')

print('Trying to save a')
f = io.BytesIO()
np.save(f, a)

print('Trying to save b')
f = io.BytesIO()
np.save(f, b)
```

### Error message:

```shell
a = array(['2000-01-01'], dtype='datetime64[D]')
a.dtype = dtype('<M8[D]')
a.dtype.metadata = None
b = array(['2000-01-01'], dtype='datetime64[D]')
b.dtype = dtype('<M8[D]')
b.dtype.metadata = mappingproxy({})
Trying to save a
Trying to save b
<...>\site-packages\numpy\lib\_format_impl.py:393: UserWarning: metadata on a dtype is not saved to an npy/npz. Use another format (such as pickle) to store it.
d['descr'] = dtype_to_descr(array.dtype)
```

Numpy version
2.3.1

Python version
3.13.5 | packaged by conda-forge | (main, Jun 16 2025, 08:20:19)

### Runtime Environment:

_No response_

### Context for the issue:

I'm trying to do some multiprocessing with numpy arrays. The multiprocessing.Pool.map() uses pickle for serialization, which led to this issue.

It's pretty surprising that a visible UserWarning is produced because of some internal magic undocumented metadata silliness.

MaanasArora:
Fixes #29344.

It seems that for datetime dtypes, `arraydescr_reduce` was initializing a new mapping instead of setting to `Py_None`, and then `arraydescr_setstate` just read it in. I set the metadata to and then checked for `Py_None` and refactored some of the logic to be clearer hopefully. 

Adding a test. Thanks for reviewing!

### DIFF
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -2879,7 +2879,6 @@ arraydescr_setstate(_PyArray_LegacyDescr *self, PyObject *args)
     char endian;
     PyObject *endian_obj;
     PyObject *subarray, *fields, *names = NULL, *metadata=NULL;
-    PyObject *old_metadata, *new_metadata;
     int incref_names = 1;
     int int_dtypeflags = 0;
     npy_uint64 dtypeflags;
@@ -3200,6 +3199,7 @@ arraydescr_setstate(_PyArray_LegacyDescr *self, PyObject *args)
         self->flags = _descr_find_object((PyArray_Descr *)self);
     }
 
+    PyObject *old_metadata, *new_metadata;
     if (PyDataType_ISDATETIME(self)) {
         PyArray_DatetimeMetaData temp_dt_data;
 
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -2878,6 +2878,7 @@ arraydescr_setstate(_PyArray_LegacyDescr *self, PyObject *args)
     char endian;
     PyObject *endian_obj;
     PyObject *subarray, *fields, *names = NULL, *metadata=NULL;
+    PyObject *old_metadata, *new_metadata;
     int incref_names = 1;
     int int_dtypeflags = 0;
     npy_uint64 dtypeflags;
@@ -3198,9 +3199,6 @@ arraydescr_setstate(_PyArray_LegacyDescr *self, PyObject *args)
         self->flags = _descr_find_object((PyArray_Descr *)self);
     }
 
-    PyObject *old_metadata, *new_metadata;
-    old_metadata = self->metadata;
-
     if (PyDataType_ISDATETIME(self)) {
         PyArray_DatetimeMetaData temp_dt_data;
 
@@ -3227,6 +3225,7 @@ arraydescr_setstate(_PyArray_LegacyDescr *self, PyObject *args)
         new_metadata = metadata;
     }
 
+    old_metadata = self->metadata;
     /*
      * We have a borrowed reference to metadata so no need
      * to alter reference count when throwing away Py_None.
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -3198,16 +3198,10 @@ arraydescr_setstate(_PyArray_LegacyDescr *self, PyObject *args)
         self->flags = _descr_find_object((PyArray_Descr *)self);
     }
 
-    /*
-     * We have a borrowed reference to metadata so no need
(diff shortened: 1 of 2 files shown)


MaanasArora:
Added test! Confirmed that it fails on main.

numpy/_core/tests/test_datetime.py
```diff
@@ -889,6 +889,15 @@ def test_pickle(self):
...
             assert_equal(pickle.loads(pkl), np.dtype('>M8[us]'))
 
+        # check that dtype metadata round-trips when none
+        # gh-29555
+        dt = np.dtype('>M8[us]')
+        assert dt.metadata is None
+        for proto in range(2, pickle.HIGHEST_PROTOCOL + 1):
+            res = pickle.loads(pickle.dumps(dt, protocol=proto))
+            assert_equal(res, dt)
+            assert res.metadata is None
```
tylerjereddy, line 899: Since this test already has 7 other assertions above, any objection to just making it its own regression test:

```diff
--- a/numpy/_core/tests/test_datetime.py
+++ b/numpy/_core/tests/test_datetime.py
@@ -889,8 +889,8 @@ def test_pickle(self):
                 b"I1\nI1\nI1\ntp7\ntp8\ntp9\nb."
             assert_equal(pickle.loads(pkl), np.dtype('>M8[us]'))
 
+    def test_gh_29555(self):
         # check that dtype metadata round-trips when none
-        # gh-29555
         dt = np.dtype('>M8[us]')
```

Beyond that, the regression test does indeed seem to fail before/pass after the patch in my hands locally.

numpy/_core/src/multiarray/descriptor.c
```diff
@@ -2671,8 +2671,9 @@ _get_pickleabletype_from_datetime_metadata(PyArray_Descr *dtype)
     if (dtype->metadata != NULL) {
         Py_INCREF(dtype->metadata);
         PyTuple_SET_ITEM(ret, 0, dtype->metadata);
-    } else {
-        PyTuple_SET_ITEM(ret, 0, PyDict_New());
+    }
+    else {
+        PyTuple_SET_ITEM(ret, 0, Py_None);
```
seberg, line 2676: This steals a reference to None!  Otherwise LGTM, though.

numpy/_core/src/multiarray/descriptor.c
```diff
@@ -2877,6 +2878,7 @@ arraydescr_setstate(_PyArray_LegacyDescr *self, PyObject *args)
     char endian;
     PyObject *endian_obj;
     PyObject *subarray, *fields, *names = NULL, *metadata=NULL;
+    PyObject *old_metadata, *new_metadata;
```
seberg: Would probably move these down, but doesn't matter (and yeah, everything else is here so).

seberg (commented):
The None reference is a bug but we can also just apply a suggestion even.  Tylers test suggestion seems like a nice to have.

numpy/_core/tests/test_datetime.py
MaanasArora: That seems nice, applied, thanks!

numpy/_core/src/multiarray/descriptor.c
MaanasArora: Oh yes, sorry - fixed. Thanks!

numpy/_core/src/multiarray/descriptor.c
MaanasArora: Yes, they seem more local state - I moved them down to be clearer.

seberg:
Thanks @MaanasArora, let's put this in.

This has been around forever, but I can't really see that not including an empty metadatadict in such a niche use-case can break anything, so I think we could backport.

PR #29557, opened 2025-08-13
DOC: Document datetime and timedelta to python's object type conversion

devmt04:
The need to document `datetime` and `timedelta` to Python object type conversions arose during a discussion in issue #29516 

NumPy follows a strict type conversion protocol when converting datetime64 or timedelta64 to standard Python objects. The rules are as follows:

Conversion of datetime64 to Python objects

- Not-a-time is returned as None.
- For days or coarser, returns a datetime.date.
- For microseconds or coarser, returns a datetime.datetime.
- For units finer than microseconds, returns an integer.

Conversion of timedelta64 to Python objects

- Not-a-time is returned as None.
- For microseconds or coarser, returns a datetime.timedelta.
- For Y/M/B (nonlinear units), generic units and units finer than microseconds, returns an integer.

Currently, this behavior is not clearly documented, which can lead to ambiguity for end users, forcing them to infer the conversion rules through experimentation.

### DIFF
--- a/doc/source/reference/arrays.datetime.rst
+++ b/doc/source/reference/arrays.datetime.rst
@@ -313,18 +313,18 @@ us / μs    microsecond      +/- 2.9e5 years         [290301 BC, 294241 AD]
 Converting datetime and timedelta to Python Object
 ==================================================
 
-NumPy follows a strict protocol when converting ``np.datetime64`` and/or ``np.timedelta64`` to Python Objects (e.g., ``tuple``, ``list``, ``datetime.datetime``). 
+NumPy follows a strict protocol when converting `datetime64` and/or `timedelta64` to Python Objects (e.g., `tuple`, `list`, `datetime.datetime`). 
 
 The protocol is described in the following table:
 
 ================================ ================================= ==================================
-         Input Type                         for ``datetime64``                    for ``timedelta64``
+         Input Type                         for `datetime64`               for `timedelta64`
 ================================ ================================= ==================================
-              ``NaT``                           ``None``                               ``None``
-     Finer than Microseconds                    ``int``                                ``int``
-     Microseconds or Coarser              ``datetime.datetime``                 ``datetime.timedelta``
-         Days or Coarser                    ``datetime.date``                   ``datetime.timedelta``
-Non-linear(Y/M) and genric units            ``datetime.date``                          ``int``
+              `NaT`                             `None`                           `None`
+     Finer than Microseconds                    `int`                            `int`
+     Microseconds or Coarser              `datetime.datetime`              `datetime.timedelta`
+         Days or Coarser                    `datetime.date`                `datetime.timedelta`
+Non-linear(Y/M) and genric units            `datetime.date`                      `int`
 ================================ ================================= ==================================
 
 .. admonition:: Example
--- a/doc/source/reference/arrays.datetime.rst
+++ b/doc/source/reference/arrays.datetime.rst
@@ -309,6 +309,24 @@ us / μs    microsecond      +/- 2.9e5 years         [290301 BC, 294241 AD]
    as      attosecond       +/- 9.2 seconds         [  1969 AD,   1970 AD]
 ======== ================ ======================= ==========================
 
+Converting datetime and timedelta to Python Object
+==================================================
+
+NumPy follows a strict protocol when converting datetime and timedelta to Python Objects (e.g., tuple, list, datetime.datetime).
+
+For conversion of datetime to a Python Object:
+
+- For days or coarser, returns a datetime.date.
+- For days or coarser, returns a datetime.date.
+- For microseconds or coarser, returns a datetime.datetime.
+- For units finer than microseconds, returns an integer.
+
+For conversion of timedelta to Python Object
+
+- Not-a-time is returned as the string "NaT".
+- For microseconds or coarser, returns a datetime.timedelta.
+- For units finer than microseconds, returns an integer.
+
 Business day functionality
 ==========================
 
--- a/doc/source/reference/arrays.datetime.rst
+++ b/doc/source/reference/arrays.datetime.rst
@@ -320,11 +320,12 @@ The protocol is described in the following table:
 ================================ ================================= ==================================
          Input Type                         for `datetime64`               for `timedelta64`
 ================================ ================================= ==================================
-              ``NaT``                           ``None``                           ``None``
-     Finer than Microseconds                    ``int``                            ``int``
-     Microseconds or Coarser              `datetime.datetime`              `datetime.timedelta`
-         Days or Coarser                    `datetime.date`                `datetime.timedelta`
-Non-linear(Y/M) and genric units            `datetime.date`                      ``int``
+          ``NaT``                             ``None``                           ``None``
+        ns/ps/fs/as                           ``int``                            ``int``
+        μs/ms/s/m/h                      `datetime.datetime`               `datetime.timedelta`
+      D/W (Linear units)                   `datetime.date`                 `datetime.timedelta` 
+    Y/M (Non-linear units)                 `datetime.date`                        ``int``
+        Generic units                      `datetime.date`                        ``int``
 ================================ ================================= ==================================
 
 .. admonition:: Example
@@ -333,20 +334,39 @@ Non-linear(Y/M) and genric units            `datetime.date`                     
 
     >>> import numpy as np
 
-    >>> type(np.datetime64('NaT').astype(object))
+    >>> type(np.datetime64('NaT').item())
     <class 'NoneType'>
 
-    >>> type(np.timedelta64('NaT').astype(object))
+    >>> type(np.timedelta64('NaT').item())
     <class 'NoneType'>
 
-    >>> type(np.timedelta64(123, 'ns').astype(object))
+    >>> type(np.timedelta64(123, 'ns').item())
     <class 'int'>
 
-    >>> type(np.datetime64('2025-01-01T12:00:00.123456').astype(object))
+    >>> type(np.datetime64('2025-01-01T12:00:00.123456').item())
     <class 'datetime.datetime'>
 
-    >>> type(np.timedelta64(10, 'D').astype(object))
+    >>> type(np.timedelta64(10, 'D').item())
     <class 'datetime.timedelta'>
+
+
+In the casea where conversion of `datetime64` and/or `timedelta64` is done against Python types like ``int``, ``float``, and ``str`` the corresponding return types will be ``np.str_``, ``np.int64`` and ``np.float64``.
+
+
+.. admonition:: Example
+
+  .. try_examples::
+
+    >>> import numpy as np
+    
+    >>> type(np.timedelta64(1, 'D').astype(int))
+    <class 'numpy.int64'>
+
+    >>> type(np.datetime64('2025-01-01T12:00:00.123456').astype(float))
+    <class 'numpy.float64'>
+
+    >>> type(np.timedelta64(123, 'ns').astype(str))
+    <class 'numpy.str_'>
 
 
 Business day functionality
--- a/doc/source/reference/arrays.datetime.rst
+++ b/doc/source/reference/arrays.datetime.rst
@@ -316,7 +316,7 @@ NumPy follows a strict protocol when converting datetime and timedelta to Python
 
 For conversion of datetime to a Python Object:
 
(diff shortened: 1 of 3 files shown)


doc/source/reference/arrays.datetime.rst
```diff
@@ -309,6 +309,24 @@ us / μs    microsecond      +/- 2.9e5 years         [290301 BC, 294241 AD]
...
+- Not-a-time is returned as the string "NaT".
+- For days or coarser, returns a datetime.date.
+- For microseconds or coarser, returns a datetime.datetime.
+- For units finer than microseconds, returns an integer.
+
+For conversion of timedelta to Python Object
+
+- Not-a-time is returned as the string "NaT".
+- For microseconds or coarser, returns a datetime.timedelta.
+- For units finer than microseconds, returns an integer.
```
jorenham: months, years, and unit-less timedelta's also return integers:

```pycon
>>> import numpy as np
>>> np.timedelta64(1, "M").item()
1
>>> np.timedelta64(1, "Y").item()
1
>>> np.timedelta64(1).item()
1
```

devmt04: Solved.

doc/source/reference/arrays.datetime.rst
```diff
@@ -309,6 +309,26 @@ us / μs    microsecond      +/- 2.9e5 years         [290301 BC, 294241 AD]
...
+- Not-a-time is returned as None.
+- For days or coarser, returns a datetime.date.
+- For microseconds or coarser, returns a datetime.datetime.
+- For units finer than microseconds, returns an integer.
+
+For conversion of timedelta to Python Object
+
+- Not-a-time is returned as None.
+- For microseconds or coarser, returns a datetime.timedelta.
+- For Y/M/B (nonlinear units), generic units and units finer than microseconds, returns an integer.
```
ngoldbaum: It might be easier to read and digest this as a table. I'd also like to see some code examples to explicitly show what happens. The nice part about code examples is they get run as part of our test suite.

Instead of "Not-a-time" I think we usually refer to it as `NaT`.

ngoldbaum: I think also adding some text this page explaining what NaT is would be helpful. We use it in a few places above, but we never explicitly explain it's the datetime equivalent of NaN and that NaT is short for not-a-time.

devmt04: > but we never explicitly explain it's the datetime equivalent of NaN and that NaT is short for not-a-time.

I think we should explain NaT explicitly under title "Datetime units", as we already had given an overview of it under "Basic datetimes".

devmt04: Also I have noticed a unknown non-linear timedelta type "B" mentioned in C code comments, is that a thing in NumPy? As I found no reference of it in docs.

```c
For conversion of timedelta to Python Object

- Not-a-time is returned as None.
- For microseconds or coarser, returns a datetime.timedelta.
- For Y/M/B (nonlinear units), generic units and units finer than microseconds, returns an integer.
```

All I am aware of is that "B" is one of the Period aliases in Pandas. Is this a mistake?

devmt04: I have committed some changes assuming  that  "B" unit is kinda a mistake here.

doc/source/reference/arrays.datetime.rst
```diff
@@ -309,6 +309,46 @@ us / μs    microsecond      +/- 2.9e5 years         [290301 BC, 294241 AD]
...
+==================================================
+
+NumPy follows a strict protocol when converting datetime64 and/or timedelta64 to Python Objects (e.g., tuple, list, datetime.datetime). 
+
+The protocol is described in the following table:
+
+================================ ================================= ==================================
+         Input Type                         for datetime64                    for timedelta64
+================================ ================================= ==================================
+              NaT                               None                               None
```
ngoldbaum:
```suggestion
            ``NaT``                           ``None``                           ``None``
```

similarly for all the other python types in the heading and the rows below as well as the text. You don't need to do it in the code examples because those are already interpreted as Python code.

ngoldbaum (commented):
While you're at it, I still think adding a new section that describes what NaT is is worth doing. But this is fine to merge without that. It looks much better with the table and code examples!

Maybe @rossbar is interested in looking this over from a documentation editing point of view?

doc/source/reference/arrays.datetime.rst
devmt04: Done! Thanks for a constructive review.

doc/source/reference/arrays.datetime.rst
```diff
@@ -309,6 +309,46 @@ us / μs    microsecond      +/- 2.9e5 years         [290301 BC, 294241 AD]
...
+The protocol is described in the following table:
+
+================================ ================================= ==================================
+         Input Type                         for `datetime64`               for `timedelta64`
+================================ ================================= ==================================
+              ``NaT``                           ``None``                           ``None``
+     Finer than Microseconds                    ``int``                            ``int``
+     Microseconds or Coarser              `datetime.datetime`              `datetime.timedelta`
+         Days or Coarser                    `datetime.date`                `datetime.timedelta`
+Non-linear(Y/M) and genric units            `datetime.date`                      ``int``
```
tylerjereddy: typo `genric` I think

tylerjereddy (commented):
Seems useful to me as well, I learned some stuff--I think I found one small typo when reading it over.

doc/source/reference/arrays.datetime.rst
jorenham: Personally I'd prefer to have the unit explicitly listed in the leftmost column, or example as `ns, ps, fs, as` for the `int` case.

But I suppose that doing both is also an option, i.e. add an additional column between "Input Type" and "for `datetime64`" labelled as "Unit code" or something.

jorenham: Maybe it's just me, but when I read this, it took me a second or two to figure out what "finer" and "coarser" meant in this context.
So maybe we could call it "shorter" and "longer" instead?

doc/source/reference/arrays.datetime.rst
```diff
@@ -309,6 +309,46 @@ us / μs    microsecond      +/- 2.9e5 years         [290301 BC, 294241 AD]
    as      attosecond       +/- 9.2 seconds         [  1969 AD,   1970 AD]
 ======== ================ ======================= ==========================
 
+
+Converting datetime and timedelta to Python Object
+==================================================
+
+NumPy follows a strict protocol when converting `datetime64` and/or `timedelta64` to Python Objects (e.g., ``tuple``, ``list``, `datetime.datetime`).
```
jorenham, line 316: The cases where `.item()` returns a `tuple` or `list` are pretty rare in practice, so if you need examples, then I'd go with more common python types like `float` or `str`.

numpy/_core/src/multiarray/_datetime.h
```diff
@@ -252,9 +253,9 @@ convert_datetime_to_pyobject(npy_datetime dt, PyArray_DatetimeMetaData *meta);
 /*
  * Converts a timedelta into a PyObject *.
  *
- * Not-a-time is returned as the string "NaT".
+ * NaT (Not-a-time) is returned as None.
  * For microseconds or coarser, returns a datetime.timedelta.
- * For units finer than microseconds, returns an integer.
+ * For Y/M (nonlinear units), generic units and units finer than microseconds, returns an integer.
```
jorenham: If I were in your shoes, I'd write something like

```suggestion
 * Returns an integer for Y/M (nonlinear units), ns/ps/fs/as (units shorter than microseconds), and generic units.
```

doc/source/reference/arrays.datetime.rst
```diff
@@ -309,6 +309,46 @@ us / μs    microsecond      +/- 2.9e5 years         [290301 BC, 294241 AD]
...
+    <class 'NoneType'>
+
+    >>> type(np.timedelta64(123, 'ns').astype(object))
+    <class 'int'>
+
+    >>> type(np.datetime64('2025-01-01T12:00:00.123456').astype(object))
+    <class 'datetime.datetime'>
+
+    >>> type(np.timedelta64(10, 'D').astype(object))
+    <class 'datetime.timedelta'>
```
jorenham:
```suggestion
    >>> type(np.datetime64('NaT').item())
    <class 'NoneType'>

    >>> type(np.timedelta64('NaT').item())
    <class 'NoneType'>

    >>> type(np.timedelta64(123, 'ns').item())
    <class 'int'>

    >>> type(np.datetime64('2025-01-01T12:00:00.123456').item())
    <class 'datetime.datetime'>

    >>> type(np.timedelta64(10, 'D').item())
    <class 'datetime.timedelta'>
```

Besides being a bit shorter, `.item()` also has better static type annotations. In fact, it's even theoretically impossible to properly annotate `astype(object)`, since `object` is the most general type. So I guess my point is that it's best to avoid it, or at least, if you care about type checking.

numpy/_core/src/multiarray/_datetime.h
devmt04: Yeah, I think explicitly mentioning the units will be more helpful.

doc/source/reference/arrays.datetime.rst
devmt04: > meant in this context.

I guess, terms like "coarser" are used widely, specially for date and time , in scientific journals, but I am not sure.

devmt04: I think we should just explicitly mention the units (like ns/ps/fs/as ) instead of using terms like "coarser" or "finer"

doc/source/reference/arrays.datetime.rst
devmt04: I just get to know that for python types like `str`,  `int`,  `float` it returns `np.str_`, `np.int64` and `np.float64`. I guess we should stick with examples like `tuple`, `list`, and `datetime.datetime` and explicitly mention the behavior with types like `str`, `int`, and `float`.

devmt04: > for python types like `str`, `int`, `float` it returns `np.str_`, `np.int64` and `np.float64`.

Can this be seen as a bug? Since it should be expected to convert them to Python objects.

doc/source/reference/arrays.datetime.rst
jorenham: > I guess, terms like "coarser" are used widely, specially for date and time , in scientific journals, but I am not sure.

I'd also say the same thing to the authors of those journals :P

doc/source/reference/arrays.datetime.rst
```diff
@@ -309,6 +309,66 @@ us / μs    microsecond      +/- 2.9e5 years         [290301 BC, 294241 AD]
...
+
+================================ ================================= ==================================
+         Input Type                         for `datetime64`               for `timedelta64`
+================================ ================================= ==================================
+          ``NaT``                             ``None``                           ``None``
+        ns/ps/fs/as                           ``int``                            ``int``
+        μs/ms/s/m/h                      `datetime.datetime`               `datetime.timedelta`
+      D/W (Linear units)                   `datetime.date`                 `datetime.timedelta` 
+    Y/M (Non-linear units)                 `datetime.date`                        ``int``
+        Generic units                      `datetime.date`                        ``int``
```
jorenham: nitpick: the alignment of `int` in the last column is off

doc/source/reference/arrays.datetime.rst
```diff
@@ -309,6 +309,66 @@ us / μs    microsecond      +/- 2.9e5 years         [290301 BC, 294241 AD]
...
+    <class 'int'>
+
+    >>> type(np.datetime64('2025-01-01T12:00:00.123456').item())
+    <class 'datetime.datetime'>
+
+    >>> type(np.timedelta64(10, 'D').item())
+    <class 'datetime.timedelta'>
+
+
+In the casea where conversion of `datetime64` and/or `timedelta64` is done against Python types like ``int``, ``float``, and ``str`` the corresponding return types will be ``np.str_``, ``np.int64`` and ``np.float64``.
```
jorenham:
```suggestion
In the case where conversion of `datetime64` and/or `timedelta64` is done against Python types like ``int``, ``float``, and ``str`` the corresponding return types will be ``np.str_``, ``np.int64`` and ``np.float64``.
```

mattip:
Putting this in as an improvement, we can incrementally improve the nuances in further PRs.

PR #29564, opened 2025-08-14
TYP: use ``TypeAliasType`` for ``ArrayLike`` and ``DTypeLike`` on py312+

jorenham:
The main selling point is that it helps reduce introspection spaghetti.

```pycon
>>> import numpy.typing as npt
>>> def f(a: npt.ArrayLike, dtype: npt.DTypeLike) -> None: ...
... 
>>> help(f)
```

Before:

```py
f(
    a: Union[collections.abc.Buffer, numpy._typing._array_like._SupportsArray[numpy.dtype[Any]], numpy._typing._nested_sequence._NestedSequence[numpy._typing._array_like._SupportsArray[numpy.dtype[Any]]], complex, bytes, str, numpy._typing._nested_sequence._NestedSequence[complex | bytes | str]],
    dtype: Union[type[Any], numpy.dtype[Any], numpy._typing._dtype_like._SupportsDType[numpy.dtype[Any]], tuple[Any, Any], list[Any], numpy._typing._dtype_like._DTypeDict, str, NoneType]
) -> None
```

After:

```py
f(a: ArrayLike, dtype: DTypeLike) -> None
```

This can also help despaghettify type-checker output in some cases.

In case you're wondering about what those `typing.TypeAliasType` things are; they're what you get when you use the new Python 3.12+ [PEP 695](https://peps.python.org/pep-0695/#generic-type-alias) `type _ = ...` syntax.
They're a bit different from `typing.TypeAlias` because they actually wrap the alias, which is accessible through its `__value__` attribute.  I don't expect this to compat issues though, since there's usually no reason to dissect these aliases at runtime. 
Static type-checkers don't treat them any differently.

### DIFF
--- a/numpy/_typing/__init__.py
+++ b/numpy/_typing/__init__.py
@@ -1,7 +1,8 @@
 """Private counterpart of ``numpy.typing``."""
 
+import sys
+
 from ._array_like import (
-    ArrayLike as ArrayLike,
     NDArray as NDArray,
     _ArrayLike as _ArrayLike,
     _ArrayLikeAnyString_co as _ArrayLikeAnyString_co,
@@ -84,7 +85,6 @@ from ._char_codes import (
 
 #
 from ._dtype_like import (
-    DTypeLike as DTypeLike,
     _DTypeLike as _DTypeLike,
     _DTypeLikeBool as _DTypeLikeBool,
     _DTypeLikeBytes as _DTypeLikeBytes,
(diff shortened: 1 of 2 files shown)


jorenham:
Let's first merge #29611, so we can see what mypy_primer has to say about this.

jorenham:
no mypy_primer diff, so that's good news :)
https://github.com/numpy/numpy/actions/runs/17143128503/job/48634273161?pr=29564

charris:
Thanks Joren.

PR #29566, opened 2025-08-15
ENH, SIMD: Optimize the argmax/argmin implementation based on Highway…

ixgbe:

The output/results can be found in the attached file.
[spin_bench_argmax_argmin_rvv.txt](https://github.com/user-attachments/files/21782585/spin_bench_argmax_argmin_rvv.txt)


numpy/_core/src/multiarray/argfunc.dispatch.cpp
```diff
@@ -0,0 +1,522 @@
...
+#define MIN(a,b) (((a)<(b))?(a):(b))
+
+namespace {
+using namespace np::simd;
+
+template <typename T>
+struct OpGt {
+#if NPY_HWY
+    template <typename V, typename = std::enable_if_t<kSupportLane<T>>>
+    HWY_INLINE HWY_ATTR auto operator()(const V &a, const V &b) const {
```
seiko2plus:
```suggestion
    template <class X = T, typename = std::enable_if_t<kSupportLane<X>>>
    NPY_FINLINE Mask<T> operator()(const Vec<T> &a, const Vec<T> &b) const { 
```
Avoid `auto`, `typename V` and instead explicitly declare the vector type based on lane type for type safety and readability. Also, no need for `HWY_ATTR` compiler flags are passed via compiler arguments. Use `NPY_FINLINE` for all inlineable functions instead of `HWY_INLINE`; it doesn't require Highway, so it can still be used when Highway is disabled.

numpy/_core/src/multiarray/argfunc.dispatch.cpp
```diff
@@ -0,0 +1,522 @@
...
+static HWY_INLINE HWY_ATTR npy_intp
+simd_argfunc_small(T *ip, npy_intp len)
+{
+    //static_assert(kMaxLanes<uint8_t> <= 64,
+    //    "the following 8/16-bit argmax kernel isn't applicable for larger SIMD");
+    /* TODO: add special loop for large SIMD width.
+       i.e avoid unroll by x4 should be numerically safe till 2048-bit SIMD width
+       or maybe expand the indices to 32|64-bit vectors(slower).  */
+
+    using UnsignedT = std::conditional_t<sizeof(T) == 1, uint8_t, uint16_t>;
```
seiko2plus:
```suggestion
    using U = std::make_unsigned_t<T>;
```

numpy/_core/src/multiarray/argfunc.dispatch.cpp
```diff
@@ -0,0 +1,522 @@
...
+    constexpr npy_intp idx_max = (sizeof(T) == 1) ? NPY_MAX_UINT8 : NPY_MAX_UINT16;
+
+    Op op_func;
+    T s_acc = *ip;
+    npy_intp ret_idx = 0, i = 0;
+
+    HWY_LANES_CONSTEXPR int vstep = Lanes<T>();
+    const int wstep = vstep*4;
+    const auto vindices_0 = hn::Iota(_Tag<UnsignedT>(), UnsignedT(0));
```
seiko2plus:
```suggestion
    const Vec<U> vindices_0 = Iota(U{0});
```
Expand the wrapper and avoid `auto` for type safety.

PR #29567, opened 2025-08-15
ENH: (``linalg``) add weighted Gram matrix computation

MengAiDev:
ENH: Add weighted_gram_matrix function to numpy.linalg

Implement weighted_gram_matrix function in numpy/linalg/_linalg.py to
compute weighted Gram matrices efficiently. Add corresponding test cases
in numpy/linalg/tests/test_linalg.py to verify correctness. Update
.gitignore to exclude .env/ directory from version control.

See #29559


.gitignore
```diff
@@ -133,3 +133,5 @@ tools/swig/test/Vector_wrap.cxx
 tools/swig/test/Array.py
 .openblas
 numpy/_distributor_init_local.py
+
+.env/
```
mattip: These kinds of additions are better handled in your global gitignore. See  the "Which file to place a pattern" section in the [gitignore documentation](https://git-scm.com/docs/gitignore).

numpy/linalg/_linalg.py
```diff
@@ -3659,3 +3659,73 @@ def vecdot(x1, x2, /, *, axis=-1):
...
+    >>> # Without weights
+    >>> np.linalg.weighted_gram_matrix(X)
+    array([[ 35,  44],
+           [ 44,  56]])
+
+    """
+    X = asanyarray(X)
+
+    if X.ndim != 2:
+        raise ValueError("X must be a 2-D array")
```
mattip: This could be generalized to higher dimensions, right?

mattip (commented):
Nice, especially the tests of many of the corner cases. I am not sure we are ready for a Pr to implement this, as the issue has two one-line implementations in the comments, and a fast cython-based implementation is in [tabmat.sandwich](https://tabmat.readthedocs.io/en/latest/api.html#tabmat.MatrixBase.sandwich) with specializations for sparse and categorical matrices. Before this gets much further, I think it would be nice to see some benchmarks of this approach vs. tabmat for some larger matrices.

MengAiDev:
Here is a benchmark:
```python

#!/usr/bin/env python3
"""
Benchmark script to compare NumPy's weighted_gram_matrix implementation 
with tabmat's sandwich product.
"""

import time
import numpy as np
import tabmat


def benchmark_numpy_implementation(X, weights=None, iterations=5):
    """Benchmark NumPy's weighted_gram_matrix implementation."""
    times = []
    for _ in range(iterations):
        start = time.perf_counter()
        result = np.linalg.weighted_gram_matrix(X, weights=weights)
        end = time.perf_counter()
        times.append(end - start)
    return np.mean(times), result


def benchmark_tabmat_implementation(X, weights=None, iterations=5):
    """Benchmark tabmat's sandwich product implementation."""
    # Convert to tabmat format
    if isinstance(X, np.ndarray):
        X_tabmat = tabmat.DenseMatrix(X)
    else:
        X_tabmat = X
    
    times = []
    for _ in range(iterations):
        start = time.perf_counter()
        if weights is not None:
            result = X_tabmat.sandwich(weights)
        else:
            # For no weights, use sandwich with ones
            result = X_tabmat.sandwich(np.ones(X.shape[0]))
        end = time.perf_counter()
        times.append(end - start)
    return np.mean(times), result


def run_benchmarks():
    """Run benchmarks for different matrix sizes."""
    print("Running benchmarks for weighted Gram matrix computation...")
    print("=" * 60)
    
    # Test cases with different sizes
    test_cases = [
        {"name": "Small (100x50)", "shape": (100, 50)},
        {"name": "Medium (1000x500)", "shape": (1000, 500)},
        {"name": "Large (5000x2000)", "shape": (5000, 2000)},
    ]
    
    results = []
    
    for case in test_cases:
        print(f"\nTesting {case['name']} matrix...")
        print("-" * 30)
        
        # Generate test data
        np.random.seed(42)  # For reproducibility
        X = np.random.rand(*case['shape'])
        weights = np.random.rand(case['shape'][0])
        
        # Benchmark NumPy implementation
        numpy_time, numpy_result = benchmark_numpy_implementation(X, weights)
        print(f"NumPy implementation: {numpy_time:.6f} seconds")
        
        # Benchmark tabmat implementation
        tabmat_time, tabmat_result = benchmark_tabmat_implementation(X, weights)
        print(f"Tabmat implementation: {tabmat_time:.6f} seconds")
        
        # Calculate speedup
        speedup = numpy_time / tabmat_time if tabmat_time > 0 else float('inf')
        print(f"Speedup (NumPy/tabmat): {speedup:.2f}x")
        
        # Verify results are similar (within tolerance)
        diff = np.abs(numpy_result - tabmat_result).max()
        print(f"Max difference between implementations: {diff:.2e}")
        
        results.append({
            "case": case["name"],
            "numpy_time": numpy_time,
            "tabmat_time": tabmat_time,
            "speedup": speedup,
            "max_diff": diff
        })
    
    # Print summary
    print("\n" + "=" * 60)
    print("BENCHMARK SUMMARY")
    print("=" * 60)
    print(f"{'Case':<15} {'NumPy (s)':<12} {'Tabmat (s)':<12} {'Speedup':<10} {'Max Diff':<10}")
    print("-" * 60)
    
    for result in results:
        print(f"{result['case']:<15} {result['numpy_time']:<12.6f} {result['tabmat_time']:<12.6f} "
              f"{result['speedup']:<10.2f}x {result['max_diff']:<10.2e}")


if __name__ == "__main__":
    run_benchmarks()

```

MengAiDev:
I update it because when handle small matrixs, numpy is faster, otherwise, tabmat is faster

.gitignore
```diff
@@ -132,4 +132,4 @@ tools/swig/test/Vector.py
 tools/swig/test/Vector_wrap.cxx
 tools/swig/test/Array.py
 .openblas
-numpy/_distributor_init_local.py
```
jorenham: revert please

jorenham (changes requested):
Like has been said before, this PR is a bit premature, as this should be discussed in the mailing list first. 
But if we assume that there'll be a consensus in favor of this, then  this would also require the stubs to be updated. Specifically in https://github.com/numpy/numpy/blob/8b89d0f127eb76c8f84c14b4814a18814f3bdec7/numpy/linalg/linalg.pyi and https://github.com/numpy/numpy/blob/main/numpy/linalg/_linalg.pyi

MengAiDev:
It seems the gitignore is the same as this commit: https://github.com/numpy/numpy/commit/0c203ed4d6ccfb44efef41e87fe5819dbeb7d822

jorenham:
> It seems the gitignore is the same as this commit: [0c203ed](https://github.com/numpy/numpy/commit/0c203ed4d6ccfb44efef41e87fe5819dbeb7d822)

well, almost; you're missing a trailing newline

MengAiDev:
I have update the `.pyi` files !

numpy/linalg/__init__.pyi
```diff
@@ -66,6 +67,7 @@ __all__ = [
     "trace",
     "vecdot",
     "vector_norm",
+    "weighted_gram_matrix"
```
jorenham:
```suggestion
    "weighted_gram_matrix",
```

numpy/linalg/_linalg.pyi
```diff
@@ -496,3 +497,9 @@ def matmul(
     x1: _ArrayLikeComplex_co,
     x2: _ArrayLikeComplex_co,
 ) -> NDArray[complexfloating]: ...
+
+def weighted_gram_matrix(
+    X: array_like,
+    *,
+    weights: array_like | None = ...
+) -> ndarray: ...
```
jorenham: The `array_like` from the docs isn't really a thing in the stubs (kinda confusing, I know). And because I don't want to make this a complicated typing thing by getting fancy with overloads and generics and such, I suppose that something like this will suffice (for now):

```suggestion
def weighted_gram_matrix(
    X: _ArrayLikeComplex_co,
    *,
    weights: _ArrayLikeComplex_co | None = None,
) -> NDArray[Any]: ...
```

numpy/linalg/_linalg.py
```diff
@@ -3659,3 +3659,115 @@ def vecdot(x1, x2, /, *, axis=-1):
...
+        raise ValueError("X must be at least a 2-D array")
+
+    if weights is None:
+        # Standard Gram matrix X.T @ X
+        # For arrays with more than 2 dimensions, transpose the last two axes
+        if X.ndim > 2:
+            # Use swapaxes to transpose the last two dimensions
+            return _core_matmul(swapaxes(X, -2, -1), X)
+        else:
+            return _core_matmul(X.T, X)
```
jorenham: I'm not sure if this matters in performance, but it could also be written as:

```suggestion
        # Standard Gram matrix X.T @ X over the last two axes
        return _core_matmul(swapaxes(X, -2, -1), X)
```

numpy/linalg/_linalg.py
```diff
@@ -3659,3 +3659,115 @@ def vecdot(x1, x2, /, *, axis=-1):
...
+
+    weights = asanyarray(weights)
+
+    if weights.ndim < 1:
+        raise ValueError("weights must be at least a 1-D array")
```
jorenham: this check seems redundant

numpy/linalg/_linalg.py
```diff
@@ -3659,3 +3659,115 @@ def vecdot(x1, x2, /, *, axis=-1):
...
+
+    # Use the optimized computation: (X.T * weights) @ X
+    # This avoids creating the full diagonal matrix diag(weights)
+    # For arrays with more than 2 dimensions, transpose the last two axes
+    if X.ndim > 2:
+        # Reshape weights to match the transposed X for broadcasting
+        weights_reshaped = weights[..., None, :]
+        return _core_matmul(swapaxes(X, -2, -1) * weights_reshaped, X)
+    else:
+        return _core_matmul((X.T * weights[..., None, :]), X)
```
jorenham: Same story as two comments up

```suggestion
    # Apply the optimized computation (X.T * weights) @ X over the last two axes
    # This avoids creating the full diagonal matrix diag(weights)
    return _core_matmul(swapaxes(X, -2, -1) * weights[..., None, :], X)
```

numpy/linalg/_linalg.py
```diff
@@ -3659,3 +3659,115 @@ def vecdot(x1, x2, /, *, axis=-1):
...
+    >>> weights_3d = np.array([[1, 2, 3], [4, 5, 6]])  # Shape (2, 3)
+    >>> G = np.linalg.weighted_gram_matrix(X_3d, weights=weights_3d)
+    >>> G.shape
+    (2, 4, 4)
+    """
+
+    X = asanyarray(X)
+
+    if X.ndim < 2:
+        raise ValueError("X must be at least a 2-D array")
```
jorenham: A `LinAlgError` seems more appropriate

jorenham: It might also help to include the actual `X.ndim` in the error message, similar to how e.g. `np.linalg.det(np.array([]))` reports 

```
LinAlgError: 1-dimensional array given. Array must be at least two-dimensional
```

numpy/linalg/tests/test_linalg.py
jorenham, line 1: It's better to use a `rng = np.random.default(42)`, instead of the *legacy*  `np.random.seed(42)`.

MengAiDev:
Is this because of my PR? https://github.com/numpy/numpy/actions/runs/17027897302/job/48266415776?pr=29567

jorenham:
> Is this because of my PR? [numpy/numpy/actions/runs/17027897302/job/48266415776?pr=29567](https://github.com/numpy/numpy/actions/runs/17027897302/job/48266415776?pr=29567)

nah don't worry about it

PR #29569, opened 2025-08-15
ENH: Add ndmax parameter to np.array to control recursion depth

kibitzing:
Implementaion of #29499

## Description

This PR introduces a new parameter, `ndmax`, to the `np.array` function. This parameter allows users to explicitly limit the maximum number of dimensions that NumPy will create when converting nested sequences (like lists of lists) into an array.

To ensure backward compatibility, the default behavior remains unchanged. When `ndmax` is not specified, the function falls back to its current heuristic. This means it will create as many dimensions as possible from nested sequences, up to the compile-time limit defined by `NPY_MAXDIMS`.

---

### Motivation and Context

As discussed in issue #29499, the current behavior of `np.array` can lead to unexpected results when `dtype=object` is specified. For example:

* A list of lists with **non-uniform lengths** correctly creates a 1D array of `list` objects.
* A list of lists with **uniform lengths** is implicitly promoted to a 2D `ndarray`, even with `dtype=object`.

This inconsistency arises because, as @rkern explained, NumPy has a strong heuristic that assumes uniformly nested sequences are intended to be multi-dimensional arrays. While this is often the desired outcome, it overrides the user's explicit intent in cases where a 1D array of `list` objects is needed, regardless of their shape. This can cause subtle bugs, particularly in data processing pipelines for machine learning.

The existing workarounds, such as pre-allocating with `np.empty` and then assigning or using `np.fromiter`, are effective but less intuitive than a direct parameter. The addition of `ndmax` provides a clear and explicit way to control this behavior directly within the most commonly used array creation function.

---

### How to use `ndmax`

The `ndmax` parameter provides fine-grained control over the array creation process.

* Current Behavior (Unexpected Promotion):
```python
# Even with dtype=object, this creates a 2D NumPy array
b = np.array([[1, 2, 3], [4, 5, 6]], dtype=object)
print(b.shape)      # (2, 3)
print(type(b[0]))   # <class 'numpy.ndarray'>
```

* New Behavior with `ndmax=1` (Desired Outcome):

```python
# With ndmax=1, NumPy stops at 1 dimension, preserving inner lists
b_new = np.array([[1, 2, 3], [4, 5, 6]], dtype=object, ndmax=1)
print(b_new.shape)      # (2,)
print(type(b_new[0]))   # <class 'list'>
```

This aligns the behavior for uniform lists with that of non-uniform lists when the user's intent is to create an array of objects. 

---

### Implementation Details

The implementation was made straightforward by leveraging the pre-existing `max_depth` parameter in the core C-API function, `PyArray_FromAny_int`. The new Python-level `ndmax` argument is passed directly to this internal parameter.

This approach ensures that the new feature is built upon NumPy's established and tested array creation logic, minimizing risk and maintaining internal consistency.

### DIFF
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -1508,6 +1508,10 @@ PyArray_FromAny(PyObject *op, PyArray_Descr *newtype, int min_depth,
         return NULL;
     }
 
+    if (max_depth == 0 || max_depth > NPY_MAXDIMS) {
+        max_depth = NPY_MAXDIMS;
+    }
+
     int was_scalar;
     PyObject* ret =  PyArray_FromAny_int(
             op, dt_info.descr, dt_info.dtype,
@@ -1544,10 +1548,6 @@ PyArray_FromAny_int(PyObject *op, PyArray_Descr *in_descr,
     coercion_cache_obj *cache = NULL;
     int ndim = 0;
     npy_intp dims[NPY_MAXDIMS];
-
-    if (max_depth == 0 || max_depth > NPY_MAXDIMS) {
-        max_depth = NPY_MAXDIMS;
-    }
 
     if (context != NULL) {
         PyErr_SetString(PyExc_RuntimeError, "'context' must be NULL");
@@ -1800,6 +1800,10 @@ PyArray_CheckFromAny(PyObject *op, PyArray_Descr *descr, int min_depth,
         Py_XDECREF(dt_info.descr);
         Py_XDECREF(dt_info.dtype);
         return NULL;
+    }
+
+    if (max_depth == 0 || max_depth > NPY_MAXDIMS) {
+        max_depth = NPY_MAXDIMS;
     }
 
     PyObject* ret =  PyArray_CheckFromAny_int(
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -1508,6 +1508,12 @@ PyArray_FromAny(PyObject *op, PyArray_Descr *newtype, int min_depth,
         return NULL;
     }
 
+    /*
+     * The internal implementation treats 0 as actually wanting a zero-dimensional
+     * array, but the API for this function has typically treated it as
+     * "anything is fine", so convert here.
+     * TODO: should we use another value as a placeholder instead?
+     */
     if (max_depth == 0 || max_depth > NPY_MAXDIMS) {
         max_depth = NPY_MAXDIMS;
     }
@@ -1802,6 +1808,7 @@ PyArray_CheckFromAny(PyObject *op, PyArray_Descr *descr, int min_depth,
         return NULL;
     }
 
+    /* See comment in PyArray_FromAny for rationale */
     if (max_depth == 0 || max_depth > NPY_MAXDIMS) {
         max_depth = NPY_MAXDIMS;
     }
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -1545,6 +1545,10 @@ PyArray_FromAny_int(PyObject *op, PyArray_Descr *in_descr,
     int ndim = 0;
     npy_intp dims[NPY_MAXDIMS];
 
+    if (max_depth == 0 || max_depth > NPY_MAXDIMS) {
+        max_depth = NPY_MAXDIMS;
+    }
+
     if (context != NULL) {
         PyErr_SetString(PyExc_RuntimeError, "'context' must be NULL");
         return NULL;
@@ -1563,7 +1567,7 @@ PyArray_FromAny_int(PyObject *op, PyArray_Descr *in_descr,
     Py_BEGIN_CRITICAL_SECTION(op);
 
     ndim = PyArray_DiscoverDTypeAndShape(
-            op, NPY_MAXDIMS, dims, &cache, in_DType, in_descr, &dtype,
+            op, max_depth, dims, &cache, in_DType, in_descr, &dtype,
             copy, &was_copied_by__array__);
 
     if (ndim < 0) {
@@ -1583,7 +1587,7 @@ PyArray_FromAny_int(PyObject *op, PyArray_Descr *in_descr,
         npy_free_coercion_cache(cache);
         goto cleanup;
     }
-    if (max_depth != 0 && ndim > max_depth) {
+    if (ndim > max_depth && (in_DType == NULL || in_DType->type_num != NPY_OBJECT)) {
         PyErr_SetString(PyExc_ValueError,
                 "object too deep for desired array");
         npy_free_coercion_cache(cache);
--- a/numpy/_core/src/multiarray/multiarraymodule.c
+++ b/numpy/_core/src/multiarray/multiarraymodule.c
@@ -1572,7 +1572,7 @@ _array_fromobject_generic(
 
     if (ndmin > ndmax) {
         PyErr_Format(PyExc_ValueError,
-                "ndmin must be <= ndmax (=%d)", ndmax);
+                "ndmin must be <= ndmax (%d)", ndmax);
         goto finish;
     }
     /* fast exit if simple call */
--- a/numpy/_core/src/multiarray/multiarraymodule.c
+++ b/numpy/_core/src/multiarray/multiarraymodule.c
@@ -1749,11 +1749,7 @@ array_array(PyObject *NPY_UNUSED(ignored),
     }
 
     if (ndmax > NPY_MAXDIMS || ndmax < 0) {
-        if (ndmax > NPY_MAXDIMS) {
-            PyErr_Format(PyExc_ValueError, "ndmax must be <= NPY_MAXDIMS (=%d)", NPY_MAXDIMS);
-        } else {
-            PyErr_Format(PyExc_ValueError, "ndmax must be >= 0");
-        }
+        PyErr_Format(PyExc_ValueError, "ndmax must be in the range [0, NPY_MAXDIMS (%d)] ", NPY_MAXDIMS);
         Py_XDECREF(dt_info.descr);
         Py_XDECREF(dt_info.dtype);
         return NULL;
--- a/numpy/_core/src/multiarray/multiarraymodule.c
+++ b/numpy/_core/src/multiarray/multiarraymodule.c
@@ -1748,11 +1748,11 @@ array_array(PyObject *NPY_UNUSED(ignored),
         op = args[0];
     }
 
-    if (ndmax > NPY_MAXDIMS || ndmax <= 0) {
+    if (ndmax > NPY_MAXDIMS || ndmax < 0) {
         if (ndmax > NPY_MAXDIMS) {
             PyErr_Format(PyExc_ValueError, "ndmax must be <= NPY_MAXDIMS (=%d)", NPY_MAXDIMS);
         } else {
-            PyErr_Format(PyExc_ValueError, "ndmax must be > 0");
+            PyErr_Format(PyExc_ValueError, "ndmax must be >= 0");
         }
         Py_XDECREF(dt_info.descr);
         Py_XDECREF(dt_info.dtype);
--- a/numpy/_core/src/multiarray/multiarraymodule.c
+++ b/numpy/_core/src/multiarray/multiarraymodule.c
@@ -1748,10 +1748,12 @@ array_array(PyObject *NPY_UNUSED(ignored),
         op = args[0];
     }
 
-    if (ndmax > NPY_MAXDIMS) {
-        PyErr_Format(PyExc_ValueError,
-            "ndmax bigger than allowable number of dimensions "
-            "NPY_MAXDIMS (=%d)", NPY_MAXDIMS);
+    if (ndmax > NPY_MAXDIMS || ndmax <= 0) {
+        if (ndmax > NPY_MAXDIMS) {
+            PyErr_Format(PyExc_ValueError, "ndmax must be <= NPY_MAXDIMS (=%d)", NPY_MAXDIMS);
+        } else {
+            PyErr_Format(PyExc_ValueError, "ndmax must be > 0");
+        }
         Py_XDECREF(dt_info.descr);
         Py_XDECREF(dt_info.dtype);
         return NULL;
--- a/numpy/_core/src/multiarray/multiarraymodule.c
+++ b/numpy/_core/src/multiarray/multiarraymodule.c
@@ -1560,7 +1560,7 @@ _prepend_ones(PyArrayObject *arr, int nd, int ndmin, NPY_ORDER order)
 static inline PyObject *
 _array_fromobject_generic(
         PyObject *op, PyArray_Descr *in_descr, PyArray_DTypeMeta *in_DType,
-        NPY_COPYMODE copy, NPY_ORDER order, npy_bool subok, int ndmin)
+        NPY_COPYMODE copy, NPY_ORDER order, npy_bool subok, int ndmin, int ndmax)
 {
     PyArrayObject *oparr = NULL, *ret = NULL;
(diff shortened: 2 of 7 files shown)


numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -1747,8 +1748,19 @@ array_array(PyObject *NPY_UNUSED(ignored),
         op = args[0];
     }
 
+    if (ndmax > NPY_MAXDIMS || ndmax <= 0) {
```
mhvk: So here I'd also check `ndmax < 0` rather than `<= 0`.

numpy/_core/src/multiarray/ctors.c
```diff
@@ -1545,6 +1545,10 @@ PyArray_FromAny_int(PyObject *op, PyArray_Descr *in_descr,
     int ndim = 0;
     npy_intp dims[NPY_MAXDIMS];
 
+    if (max_depth == 0 || max_depth > NPY_MAXDIMS) {
```
mhvk: To support actually using `max_depth = 0` as a real option, the check here would need to be removed.

But that means adjusting code that calls it. The main user is `PyArray_FromAny` (in `ctors.c` as well), which would then need this `if` statement (indeed, perhaps the whole statement should be moved there; it is not unreasonable to assume that for internal calls the values are known to be good).

One other user is `PyArray_CheckFromAny_int` (also in`ctors.c`), though logically there also perhaps `PyArray_CheckFromAny` is the one that should be adjusted.

The final user seems to be `scalartypes.c.src` -- there one could just replace the 0 with `NPY_MAX_DIMS`.

mhvk (commented):
@kibitzing - thanks for the PR. I think this is rather nice, especially that you found all it really takes is exposing existing functionality!

However, to me at least there is one annoyance: it would seem to me that there is no reason not to support `np.array(..., ndmax=0)` as meaning that the user wants an array with `ndim=0` (i.e., a scalar array). 

Unfortunately, this clashes with the numpy C API, which has `PyArray_FromAny` and friends interpret `max_depth=0` as arbitrary (it is actually not documented as such, but since it is used like that throughout the code base, it surely is used outside of numpy as well).

But I think we can make it work by adjusting `PyArray_FromAny` to change 0 to `NPY_MAXDIMS` and then have `PyArray_FromAny_int` not do any checks. See in-line comments.

What do you think?

kibitzing:
Hello @mhvk,
Thank you for the excellent feedback and for proposing a clear solution!

I also considered supporting `ndmax=0` for scalar creation. I initially decided against it, partly because of the C-API's convention that you mentioned, and partly because I was cautious about modifying such a core piece of code.

However, the detailed guidance you've provided makes the path forward very clear, so I'm happy to get to work on implementing it.

numpy/_core/src/multiarray/multiarraymodule.c
kibitzing: Thank you for the suggestion, I've changed as you suggested.

numpy/_core/src/multiarray/ctors.c
kibitzing: Thank you for the clear guidance.

As you suggested, I've refactored the logic so `PyArray_FromAny_int` can treat `max_depth=0` as a request for a 0-D array. 
The responsibility for handling the "no limit" case (by converting `0` to `NPY_MAXDIMS`) has been moved to its callers:
- The check was added to `PyArray_FromAny` and `PyArray_CheckFromAny`.
- Call sites in `scalartypes.c.src` and `array_converter_new` were updated to use `NPY_MAXDIMS` explicitly.

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -1747,8 +1748,19 @@ array_array(PyObject *NPY_UNUSED(ignored),
         op = args[0];
     }
 
+    if (ndmax > NPY_MAXDIMS || ndmax < 0) {
+        if (ndmax > NPY_MAXDIMS) {
+            PyErr_Format(PyExc_ValueError, "ndmax must be <= NPY_MAXDIMS (=%d)", NPY_MAXDIMS);
```
mhvk: Again, I wouldn't ask if there wasn't another reason to push a further commit, but to have error path code interrupt the flow as little as possible, I'd tend to have just one `PyErr_Format`, with (e.g.) "must have 0 <= ndmax <= NPY_MAXDIMS (=%d)".

numpy/_core/src/multiarray/ctors.c
```diff
@@ -1508,6 +1508,10 @@ PyArray_FromAny(PyObject *op, PyArray_Descr *newtype, int min_depth,
         return NULL;
     }
 
+    if (max_depth == 0 || max_depth > NPY_MAXDIMS) {
```
mhvk, line 1517: This would not be worth doing if there was nothing else, but since we need to do the docs, maybe worth putting in a comment, say along the lines of
```
/*
 * The` internal implementation treats 0 as actually wanting a zero-dimensional
 * array, but the API for this function has typically treated it as "anything is fine",
 * so convert. TODO: should we use another value as a placeholder instead?
 */
```

mhvk (commented):
@kibitzing - this looks great. I was about to approve, but then I realized we should make sure this is actually documented. So, could you add a description in `_add_newdocs.py` (search for `'array'` - l.806; you'll need a`.. versionadded:...` in there, see examples in that file).

Furthermore, personally, I feel this is worth a little what's-new entry - there are many who have run into this issue! (We've got some code in astropy that we can now simplify!) For this you need to add a fragment in `doc/release/upcoming_changes` - see the `README.rst` in that directory).

Since that means another CI run anyway, I also put two absoltely nitpicky comments in-line, which you might as well do too... (but feel free to disagree and ignore).

Let me also ping @mattip and @rkern, in case they want to have a look as well (though in the end this is so much a case of just exposing what was effectively in place already that I'm quite happy to just get it in myself too).

numpy/_core/src/multiarray/multiarraymodule.c
kibitzing: Thank you for the feedback!
Updated to use a single `PyErr_Format` with the combined bounds check message.

numpy/_core/src/multiarray/ctors.c
kibitzing: Added the comment explaining the 0 value handling, thanks for the suggestion!

kibitzing:
Hello @mhvk,
Thanks for the thorough review! I've addressed all the points:
- Added documentation and examples in `_add_newdocs.py` with the `.. versionadded::` directive
- Created a news fragment in `doc/release/upcoming_changes` for the what's new entry (happy to hear it will help simplify things for you!)
- Applied the inline suggestions. Thanks for spotting those, they definitely improve the code quality

Please let me know if there's anything else you'd like me to adjust!

mhvk (approved):
Super, this now looks great! Let's get it in. Thanks very much for solving a long-standing annoyance in an elegant way!

kibitzing:
Hello @mhvk,
Thank you so much for your insightful, thoughtful and detailed feedback throughout the process! 
I've learned a lot. It was a real pleasure working together on this!

mhvk:
Indeed, a nice collaboration!

numpy/_core/src/multiarray/ctors.c
seberg: From a Pytohn perspective supporting 0 is desired though, while from the C one it is tricky. I wonder if we can't make it work (even if not in C -- the internal function may have to differ).

seberg: Sorry for the noise... this is the public function and you didn't change the internal one, nice :).

kibitzing: No problem at all! Thanks for taking a look, @seberg :)

numpy/_core/src/multiarray/ctors.c
```diff
@@ -1583,7 +1593,7 @@ PyArray_FromAny_int(PyObject *op, PyArray_Descr *in_descr,
         npy_free_coercion_cache(cache);
         goto cleanup;
     }
-    if (max_depth != 0 && ndim > max_depth) {
+    if (ndim > max_depth && (in_DType == NULL || in_DType->type_num != NPY_OBJECT)) {
```
tylerjereddy, line 1596: Reverting this change per diff below the fold resolves gh-29721 for me.

<details>

```diff
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -1593,7 +1593,7 @@ PyArray_FromAny_int(PyObject *op, PyArray_Descr *in_descr,
         npy_free_coercion_cache(cache);
         goto cleanup;
     }
-    if (ndim > max_depth && (in_DType == NULL || in_DType->type_num != NPY_OBJECT)) {
+    if (max_depth != 0 && ndim > max_depth) {
         PyErr_SetString(PyExc_ValueError,
                 "object too deep for desired array");
         npy_free_coercion_cache(cache);
```

</details>

Application of that patch also passes the full NumPy testsuite locally on x86_64 Linux, so I kind of wonder if this is really justified, or at least if there is some way to keep the same behavior for error handling.

mhvk: @tylerjereddy - thanks for the sleuthing! Is there a test case that mimics what happens in scipy? I ask in part because I wonder if it is the re-addition of `maxdepth != 0` that makes things work, or the removal of the `in_DType` check. I'm also confused that the test cases we added still pass after reverting this - I thought this was all quite logically needed... Let me ping @kibitzing as well.

kibitzing: Thanks both for the discussion!

To give a bit of context first: before this PR, `max_depth=0` was not a user-facing value. It was used internally as a signal meaning "no limit" or "arbitrary". That’s why we originally had the `max_depth != 0` check.

After this PR, however, we explicitly allow users to set `max_depth`, and `0` is now a valid input value. With the previous condition, tests likely passed only because we normalized all `max_depth=0` values to `NPY_MAXDIMS` (64). The actual `ndim` comparison was already handled in the preceding line based on `max_depth`.

My guess is that the failure you are seeing arises from a call to `PyArray_FromAny_int` with `max_depth=0`. On my side, I verified that the tests do not break even without the `max_depth != 0` condition (tested on macOS).

It would be very helpful to have a minimal reproducer extracted from SciPy. That would let us determine whether we need to refine the condition (e.g., separate the `max_depth` logic from the object-DType logic).

mattip: `max_depth` is the 4th argument in `PyArray_FromAny`. We have one such call inside NumPy behind a define https://github.com/numpy/numpy/blob/104dcc3eafcf1591e680a2b3af6f69f465c912fe/numpy/_core/include/numpy/ndarrayobject.h#L86-L87

And there are two in scipy
```
$ git grep PyArray_FromAny
scipy/spatial/src/distance_pybind.cpp:    auto* arr = PyArray_FromAny(obj.ptr(), descr, 0, 0, flags, nullptr);
scipy/spatial/src/distance_pybind.cpp:    auto* arr = PyArray_FromAny(obj.ptr(), nullptr, 0, 0, flags, nullptr);
```

mattip: There are many uses of `PyArray_FromObject` in scipy, where the [last argument is max_depth](https://numpy.org/devdocs/reference/c-api/array.html#c.PyArray_FromObject) and the value is `0`

```
$ git grep "PyArray_FromObject.*0)"
scipy/signal/_correlate_nd.cc:    ax = (PyArrayObject *)PyArray_FromObject(x, typenum, 0, 0);
scipy/signal/_correlate_nd.cc:    ay = (PyArrayObject *)PyArray_FromObject(y, typenum, 0, 0);
scipy/signal/_correlate_nd.cc:    aout = (PyArrayObject *)PyArray_FromObject(out, typenum, 0, 0);
scipy/signal/_lfilter.cc:    arX = (PyArrayObject *) PyArray_FromObject(X, typenum, 0, 0);
scipy/signal/_sigtoolsmodule.cc:        afill = (PyArrayObject *)PyArray_FromObject(fill_value, typenum, 0, 0);

```

mhvk: @mattip - those call the public version, but for that function there was meant to be no change. Indeed, in there `0` is translated into `maxdepth` before calling the `PyArray_FromArray_int`, which is where the clause we are discussing here was changed. Hence, this all still leaves me a bit confused about why the change here caused problems. I suspect it is the object check... @tylerjereddy - are you still setup to check if just doing
```
if (ndim > max_depth) {
```
solves the scipy problems too? (Sorry, am in the midst of teaching prep so unable to test myself right now...)

tylerjereddy:
@mhvk Yes, that patch resolves both SciPy release branch test failures on x86_64 Linux--I just checked.

That said, it looks like Ralf agrees this is a pretty minor matter, so I'm just going to skip those tests for now--we're just exchanging one error class for another as far as the user is concerned.

numpy/_core/src/multiarray/ctors.c
mhvk: Good to hear that works. Though I'd still like to understand why the rest is not necessary for numpy -- if it really isn't, then I think we should just remove it still.

tylerjereddy:
Although on Windows it looks like sometimes the error doesn't happen at all... yuck, anyway I'm going to suppress the test downstream for now I think

numpy/_core/src/multiarray/ctors.c
kibitzing: I now understand this issue properly, and I agree this is indeed a bug.

When I originally replaced:
```c
if (max_depth != 0 && ndim > max_depth) {
```
with:
```c
if (ndim > max_depth && (in_DType == NULL || in_DType->type_num != NPY_OBJECT))
```

I focused only on the meaning of `max_depth != 0` as "max_depth was explicitly set by the user" and tried to use `(in_DType == NULL || in_DType->type_num != NPY_OBJECT)` as a proxy for this condition. My reasoning was that since `max_depth` option is only allowed for `object` dtypes, an `object` dtype would indicate that `max_depth` was set externally.

However, this was incorrect logic. The original `max_depth != 0` condition wasn't primarily about "don't raise an error if max_depth was set externally." Rather, it was about avoiding false positives: since `max_depth` defaults to 0 when not set, the condition `ndim > max_depth` would always be true and always raise an error without this guard.

I also confused this with the actual behavior I implemented where objects deeper than `max_depth` get stored as nested objects. But if that behavior had occurred, it would have been handled in `PyArray_DiscoverDTypeAndShape`, and the resulting `ndim` would already be bounded by `max_depth`. Therefore, no additional restriction should be needed.

The `(in_DType == NULL || in_DType->type_num != NPY_OBJECT)` condition is not only redundant but actually problematic. It can suppress "too deep" errors in cases where it should actually be raised, as @tylerjereddy observed. (Thank you for catching this!)

kibitzing: I said `ndim` should be bounded by `max_depth`, but `ndim > max_depth` still happens when recursive inference is bypassed - specifically when a `np.array` is passed as input to `np.array()`. In such cases, `ndim` is determined directly from the input array's dimensions without inference, so `ndim > max_depth` can still occur.
For example, if `dtype=object`, the input is a `np.array`, and `max_depth` is set smaller than the input array's dimensions, we should raise an error but currently don't due to the object dtype condition. This could lead to unexpected bugs downstream.
Therefore, I believe we should remove the object dtype condition and simply use:
```C
if (ndim > max_depth) {
```
This would properly handle all cases while maintaining the intended behavior.

mhvk: @kibitzing - that all makes sense. Would you be able to make a quick PR that fixes this?

kibitzing: @mhvk Done! I created a [PR](https://github.com/numpy/numpy/pull/29745) that removes the object dtype check and adds a regression test.

Issue #28247, opened 2025-01-29
BUG: order="A" doesn't do anything?


crusaderky:
### Describe the issue:

From the [documentation of asarray](https://numpy.org/devdocs/reference/generated/numpy.asarray.html):

>  ‘A’ (any) means ‘F’ if a is Fortran contiguous, ‘C’ otherwise

However, `np.asarray(..., order="A")` doesn't seem to do anything, making it redundant with `order='K'`.

### Reproduce the code example:

```python
>>> x = np.broadcast_to(np.zeros(()), (2, ))
>>> np.asarray(x, order="A").flags
  C_CONTIGUOUS : False
  F_CONTIGUOUS : False

>>> x = np.zeros(2)[::-1]
>>> np.asarray(x, order="A").flags
  C_CONTIGUOUS : False
  F_CONTIGUOUS : False

>>> x = np.zeros(4)[::2]
>>> np.asarray(x, order="A").flags
  C_CONTIGUOUS : False
  F_CONTIGUOUS : False
```

### Error message:

```shell

```

### Python and NumPy Versions:

python 3.10
numpy 2.2.1

seberg:
I think it actually is different *if* a copy is made (I won't argue this makes sense!).  But mainly `A` is just much older than `K`.

The documentation(?) issue I suppose is that `"C"` and `"F"` actually ensure a contiguous result, but `"A"` doesn't.  (FWIW, even for `K` it _isn't_ clear that you don't want a single chunk of memory in principle, except numpy doesn't really flag that.)

guilhermep21:
Is that a good first issue?

seberg:
It's not a bug, maybe a documentation issue, which a PR is welcome for.

nbidler:
I've picked this up for a 'first issue' and while I found the [location of the docs in the files](https://github.com/numpy/numpy/blob/51a72bcf2e2fc7c44e17ba293d30296da5ec7cd7/numpy/_core/_add_newdocs.py), I have spent three weeks trying to trace the chain of custody from `asarray()` call down to the C-level implementation and having trouble using `pdb` or `gdb` to trace things, so I am just going to alter the docs directly to reflect this.

To be clear, I also duplicated the behavior described above, but have not been able to trace where the actual handling of the 'order' arg occurs.

mattip:
[...]

PR #29571, opened 2025-08-15
DOC: clarify numpy.asarray parameter 'order' meanings

nbidler:
See #28247. Changed description of options for parameter `order` - made explicit that 'C' and 'F' give contiguous results, and 'A' will give results equivalent to 'F' if the input is Fortran-contiguous OR non-contiguous, but if the desired output is _non-contiguous Fortran-style_ then the user also needs to set `copy` to `True`.


numpy/_core/_add_newdocs.py
```diff
@@ -943,12 +943,12 @@
...
-        'C' row-major (C-style),
-        'F' column-major (Fortran-style) memory representation.
-        'A' (any) means 'F' if `a` is Fortran contiguous, 'C' otherwise
-        'K' (keep) preserve input order
-        Defaults to 'K'.
+        The memory layout of the output ndarray. The output of 'A' and 'K' depend on the memory layout of input a.
+        'C' gives a row-major layout (C-style),
+        'F' gives a column-major layout (Fortran-style).
+        Both 'C' and 'F' ensure a contiguous result.
+        'A' (any) is equivalent to 'F' if input a is non-contiguous or Fortran-contiguous, otherwise, it is equivalent to 'C'. If the desired output is a non-contiguous Fortran-style ndarray, the user should set ``copy`` to ``True`` to ensure this.
```
seberg: Can we please just add a minimal thing that `'A'`, like `'K'`, will not copy to ensure a contiguous output?
Yeah, maybe still confusing, so mentioning the opposite that 'C' and 'F' will copy if needed is maybe clarifying.

numpy/_core/_add_newdocs.py
```diff
@@ -943,12 +943,12 @@
     dtype : data-type, optional
         By default, the data-type is inferred from the input data.
     order : {'C', 'F', 'A', 'K'}, optional
-        Memory layout.  'A' and 'K' depend on the order of input array a.
-        'C' row-major (C-style),
-        'F' column-major (Fortran-style) memory representation.
-        'A' (any) means 'F' if `a` is Fortran contiguous, 'C' otherwise
-        'K' (keep) preserve input order
-        Defaults to 'K'.
+        The memory layout of the output ndarray. The output of 'A' and 'K' depend on the memory layout of input a.
```
jorenham: How about simplifying this to

```suggestion
        The memory layout of the output.
```

and move the explanation of the input-dependence of `A` and `K` to the individual explanations of `A` and `K` below?

Issue #27018, opened 2024-07-23
BUG: f2py cannot find .mod files automatically for the new meson backend


1234zou:
### Describe the issue:

I'm trying to compile my Fortran code using `f2py`. For python\>=3.12, the meson backend is now the default. But meson/f2py cannot automatically find the generated .mod files in the current directory. These .mod files could be automatically found for older versions of python and distutils backend. By the way, gcc/gfortran in this machine is 4.8.5.

Thanks for any kind help and suggestion!

Best,
Jingxiang

### Reproduce the code example:
A simple example is shown using two Fortran files and one Python script:

here is a.f90
```
module para
 implicit none
 integer, parameter :: n = 5
end module para
```

b.f90
```
subroutine prt_n
 use para, only: n
 implicit none
 write(6,'(A,I2)') 'n=', n
end subroutine prt_n
```

Compiling commands are
```
gfortran a.f90 -c
f2py -c a.o b.f90 -m test
```
The first command would generate `a.o` and `para.mod` files in the current directory. `para.mod` can be automatically detected for the distutils backend, but not for the meson backend.

test.py
```python
from test import prt_n
prt_n()
```
Assuming that the test.xxx.so is generated (which not True for the meson backend), the result would be `n= 5`.

### Error message:
After running `f2py -c a.o b.f90 -m test`, the error message is
```
Found ninja-1.10.2 at /public/home/jxzou/software/anaconda3/envs/py312/bin/ninja
INFO: autodetecting backend as ninja
INFO: calculating backend command to run: /public/home/jxzou/software/anaconda3/envs/py312/bin/ninja -C /public/home/jxzou/tmp/tmpgljbyueg/bbdir
ninja: Entering directory `/public/home/jxzou/tmp/tmpgljbyueg/bbdir'
[3/6] Compiling Fortran object test.cpython-312-x86_64-linux-gnu.so.p/b.f90.o
FAILED: test.cpython-312-x86_64-linux-gnu.so.p/b.f90.o
[...]

PR #29572, opened 2025-08-16
ENH: enable processing object file for f2py meson backend

### LINKED ISSUE #28191: BUG: f2py undefined symbol for the new meson backend
### Describe the issue:

This is a continued problem of a previous numpy f2py [issue](https://github.com/numpy/numpy/issues/27018). The error information is different and I think it deserves opening a new issue. After adding `-I$(pwd)` to `f2py`, there is no compilation error, but there is an importing error `undefined symbol`. The following test is performed using Python 3.12.8 and gcc 8.5.0.

Thanks for any kind help and suggestion!

Best,
Jingxiang

A simple example is shown using two Fortran files and one Python script:
a.f90
```
module para
 implicit none
 integer :: n
end module para
```
b.f90
```
subroutine prt_n
 use para, only: n
 implicit none
 n = 5
 write(6,'(A,I2)') 'n=', n
end subroutine prt_n
```
Compiling commands are
```
gfortran a.f90 -c
f2py -c a.o b.f90 -m test -I$(pwd)
```
The file `test.cpython-312-x86_64-linux-gnu.so` is generated.

### Error message:
Start python and it can be found
```shell
>>>from test import prt_n
ImportError: /public/home/jxzou/wfu/pbc_pm_test/test.cpython-312-x86_64-linux-gnu.so: undefined symbol: __para_MOD_n
```
which means that the integer variable `n` seems not be used during compiling.

Python 3.12.8, numpy 2.2.1

### Runtime Environment:
```
[{'numpy_version': '2.2.1',
  'python': '3.12.8 | packaged by Anaconda, Inc. | (main, Dec 11 2024, '
            '16:31:09) [GCC 11.2.0]',
  'uname': uname_result(system='Linux', node='mu012', release='4.18.0-372.64.1.kos5.x86_64', version='#1 SMP Tue Aug 8 20:23:39 EDT 2023', machine='x86_64')},
                      'found': ['SSSE3',
                                'SSE41',
                                'POPCNT',
                                'SSE42',
                                'AVX',
                                'F16C',
                                'FMA3',
                                'AVX2',
                                'AVX512F',
                                'AVX512CD',
                                'AVX512_SKX',
                                'AVX512_CLX',
                                'AVX512_CNL',
                                'AVX512_ICL'],
                      'not_found': ['AVX512_KNL', 'AVX512_KNM']}},
 {'filepath': '/public/home/jxzou/software/anaconda3/envs/mokit-py312/lib/libmkl_rt.so.2',
  'internal_api': 'mkl',
  'num_threads': 64,
  'prefix': 'libmkl_rt',
  'threading_layer': 'intel',
  'user_api': 'blas',
  'version': '2023.1-Product'},
 {'filepath': '/public/home/jxzou/software/anaconda3/envs/mokit-py312/lib/libiomp5.so',
  'internal_api': 'openmp',
  'num_threads': 64,
  'prefix': 'libiomp',
  'user_api': 'openmp',
  'version': None}]
```

### Context for the issue:

We are developers of the open source package [MOKIT](https://gitlab.com/jxzou/mokit). By using f2py, we offer lots of convenient and efficient Python APIs to users in the computational chemistry field. We are grateful for your numpy f2py functionalities. And we are still looking for a solution to move to new meson backend. Thank you.

jeanwsr:
This PR is going to fix #28191, in which we just realized that all `.o` file in `f2py -c a.o b.f90 -m test` is ignored by the meson backend. This feature was smoothly supported in the disutils backend before. 

Given that the name of object file is already scanned by [f2py](https://github.com/numpy/numpy/blob/4e00e4df15528323f6d76bbd40157b582861af86/numpy/f2py/f2py2e.py#L713), it is not difficult to recover this feature.
Basically, I add `objects: []` in meson.build.template and create `objects_substitution`, `_prepare_objects` functions for passing and copying the object file. I'm new to contributing to f2py, all feedback is welcome.

Another question is: what if the `.o` file has a module? Currently my choice is not to copy the `.mod` file, since the directory of them can be specified by something like `-I$(PWD)` , which is a workaround mentioned in #27018. Please feel free to discuss any other treatment on the `.mod` files.

### DIFF
--- a/numpy/f2py/tests/test_f2py2e.py
+++ b/numpy/f2py/tests/test_f2py2e.py
@@ -673,6 +673,25 @@ def test_inclheader(capfd, hello_world_f90, monkeypatch):
             assert "#include <stdbool.h>" in ocmr
             assert "#include <stdio.h>" in ocmr
 
+@pytest.mark.skipif((platform.system() != 'Linux'), reason='Compiler required')
+def test_cli_obj(capfd, hello_world_f90, monkeypatch):
+    """Ensures that the extra object can be specified when using meson backend
+    """
+    ipath = Path(hello_world_f90)
+    mname = "blah"
+    odir = "tttmp"
+    obj = "extra.o"
+    monkeypatch.setattr(sys, "argv",
+                        f'f2py --backend meson --build-dir {odir} -m {mname} -c {obj} {ipath}'.split())
+
+    with util.switchdir(ipath.parent):
+        Path(obj).touch()
+        compiler_check_f2pycli()
+        with Path(f"{odir}/meson.build").open() as mesonbuild:
+            mbld = mesonbuild.read()
+            assert "objects:" in mbld
+            assert f"'''{obj}'''" in mbld
+
 
 def test_inclpath():
     """Add to the include directories
--- a/numpy/f2py/_backends/_meson.py
+++ b/numpy/f2py/_backends/_meson.py
@@ -50,6 +50,7 @@ class MesonTemplate:
         self.pipeline = [
             self.initialize_template,
             self.sources_substitution,
+            self.objects_substitution,
             self.deps_substitution,
             self.include_substitution,
             self.libraries_substitution,
@@ -77,6 +78,11 @@ class MesonTemplate:
     def sources_substitution(self) -> None:
         self.substitutions["source_list"] = ",\n".join(
             [f"{self.indent}'''{source}'''," for source in self.sources]
+        )
+
+    def objects_substitution(self) -> None:
+        self.substitutions["obj_list"] = ",\n".join(
+            [f"{self.indent}'''{obj}'''," for obj in self.objects]
         )
 
     def deps_substitution(self) -> None:
@@ -186,6 +192,7 @@ class MesonBackend(Backend):
 
     def compile(self) -> None:
         self.sources = _prepare_sources(self.modulename, self.sources, self.build_dir)
+        _prepare_objects(self.modulename, self.extra_objects, self.build_dir)
         self.write_meson_build(self.build_dir)
         self.run_meson(self.build_dir)
         self._move_exec_to_root(self.build_dir)
@@ -216,6 +223,12 @@ def _prepare_sources(mname, sources, bdir):
     ]
     return extended_sources
 
+def _prepare_objects(mname, objects, bdir):
+    Path(bdir).mkdir(parents=True, exist_ok=True)
+    # Copy objects
+    for obj in objects:
+        if Path(obj).exists() and Path(obj).is_file():
+            shutil.copy(obj, bdir)
 
 def _get_flags(fc_flags):
     flag_values = []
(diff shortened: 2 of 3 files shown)


numpy/f2py/tests/test_f2py2e.py
```diff
@@ -673,6 +673,26 @@ def test_inclheader(capfd, hello_world_f90, monkeypatch):
             assert "#include <stdbool.h>" in ocmr
             assert "#include <stdio.h>" in ocmr
 
+@pytest.mark.skipif((platform.system() != 'Linux'), reason='Compiler required')
```
HaoZeke, line 676: I think specializing the skipif should fix the false positive test failure.

jeanwsr: Fixed.

numpy/f2py/tests/test_f2py2e.py
```diff
@@ -673,6 +673,26 @@ def test_inclheader(capfd, hello_world_f90, monkeypatch):
...
+    mname = "blah"
+    odir = "tttmp"
+    obj = "extra.o"
+    monkeypatch.setattr(sys, "argv",
+                        f'f2py --backend meson --build-dir {odir} -m {mname} -c {obj} {ipath}'.split())
+
+    with util.switchdir(ipath.parent):
+        Path(obj).touch()
+        compiler_check_f2pycli()
+        # out, _ = capfd.readouterr()
```
HaoZeke:
```suggestion
```

HaoZeke (approved):
Minor nit aside this is good to go, thanks @jeanwsr

I think it makes sense to not modify the mod file handling here

HaoZeke:
@jeanwsr, one last thing, could you [rewrite the commit history](https://numpy.org/doc/stable/dev/development_workflow.html#rewriting-commit-history) to use the [prefixes defined in the documentation](https://numpy.org/doc/stable/dev/development_workflow.html#writing-the-commit-message)? The relevant ones would be ENH: MAINT and TST

jeanwsr:
The test failure seems unrelated.

HaoZeke:
Thanks for seeing this through @jeanwsr

steog88:
I have the same problem, so I tried the latest numpy hoping that it would work, but it did not. Here I see the PR is merged, but the latest numpy releases do not contain the fix yet (if I am not mistaken, I do not see the new lines even in the 2.4.0.dev0). Do you know when it will be released?

jeanwsr:
> I have the same problem, so I tried the latest numpy hoping that it would work, but it did not. Here I see the PR is merged, but the latest numpy releases do not contain the fix yet (if I am not mistaken, I do not see the new lines even in the 2.4.0.dev0). Do you know when it will be released?

It will be in 2.4.0rc1, then 2.4.0 (next month i think). Before that one can manually patch the local numpy to apply the changes (compiling numpy is not needed).

PR #29244, opened 2025-06-20
DEP: Deprecate setting the dtype of an exact numpy.ndarray

eendebakpt:
Partly addresses https://github.com/numpy/numpy/issues/28800. A continuation of #28901.

We only deprecate setting the dtype of an exact array. For masked and record arrays it will take some more refactoring to avoid setting the dtype. We skip warnings for pypy as we cannot rely on reference counting to avoid warnings when the dtype is changed via `array.view(new_dtype)`.

PR #28901, opened 2025-05-04
DEP: Deprecate setting the strides and dtype of a numpy array

eendebakpt:
See https://github.com/numpy/numpy/issues/28800

Issue #28800, opened 2025-04-23
TSK: deprecate setting `strides`, `dtype`, and `shape` on arrays.


ngoldbaum:
As seen in https://github.com/numpy/numpy/issues/28727, mutating arrays like this is dangerous in surprising ways, particularly on the free-threaded build.

We should guide people towards safer alternatives that create new views or copies of arrays instead of mutating.

We should first try deprecating `strides` and `dtype`, then `shape`, since we expect that setting `shape` is used the most (before `np.reshape` it was the only way to do it).

To do the actual deprecation, first someone needs to add the warning and deal with any fallout in the NumPy tests. Make sure the deprecation warnings indicate clearly how to update affected code. It would also be good to look at some downstream libraries like SciPy, matplotlib, pandas, and scikit-learn to see how much noise these deprecations cause in downstream tests.

If it isn't too bad and the fixes are straightforward, we should be good to go.

seberg:
> (before np.reshape it was the only way to do it).

Tiny correction:  Reshape always existed, but raising an error when it failed with `copy=False` is new (i.e. knowing that the result is a view).  Although, the copy was made even on error before.

But yeah, shape assignment was/is pretty common (I have seen it in educational materials, maybe it still is).

jorenham:
Yes please! THe no. 1 rule of typing is that *types don't change*, and `dtype` and `shape` are part of the `ndarray` *type*.

ricardoV94:
It may be useful to make `as_strided` faster. It's rather slow compared to transpose + reshape (for the cases where it is identical), even though it should be doing the same thing as those

taldcroft:
From a user and astropy-dev perspective, I have always conceptualized `shape` as being a lightweight attribute that can be changed freely and merely informs the array how to interpret the `item` in `__getitem__`. 

I suspect that deprecating `shape` as a mutable attribute will impact a lot of user code. E.g. in astropy tables it would always be preferable to change the shape in-place.
```
>>> from astropy.table import Table
>>> t = Table({"a": [1, 2, 3, 4]})
>>> t["a"].shape = (2, 2)  # In-place, column id unchanged
>>> t["a"] = np.reshape(t["a"], (2, 2))  # Bunch of Python code to validate and replace column, 4 times slower
```
This is not the end of the world, but I just want to note the impact.

One naive question I have is how changing the `shape` attribute is more thread-unsafe than mutating the array data itself.

ngoldbaum:
[...]

PR #29575, opened 2025-08-16
DEP: Deprecate setting the dtype of an numpy.ndarray

eendebakpt:
Partly addresses https://github.com/numpy/numpy/issues/28800. A continuation of https://github.com/numpy/numpy/pull/28901 and https://github.com/numpy/numpy/pull/29244.

We deprecate setting the dtype of an numpy arrays. For masked and record arrays no warnings are generated, as the dtype usage is a bit more complex (in a followup PR we can enable deprecation warnings by modifying the dtype setter logic). We skip warnings for pypy as we cannot rely on reference counting to avoid warnings when the dtype is changed via `array.view(new_dtype)`.


ngoldbaum (commented):
Have you looked at all at how noisy this is downstream? It's a little hard to search for but I suspect this pattern is used quite a bit. It would probably help us to understand what downstream usage looks like to try running some downstream test suites with a patched numpy to see how noisy it is.

SciPy is probably a good choice, maybe astropy too because they implement an ndarray subclass. I'd skip pandas because its test suite takes a very long time to run outside of github actions.

eendebakpt:
> Have you looked at all at how noisy this is downstream? It's a little hard to search for but I suspect this pattern is used quite a bit. It would probably help us to understand what downstream usage looks like to try running some downstream test suites with a patched numpy to see how noisy it is.
> 
> SciPy is probably a good choice, maybe astropy too because they implement an ndarray subclass. I'd skip pandas because its test suite takes a very long time to run outside of github actions.


Astropy indeed uses setting the dtype and shape. I created a ticket and a PR https://github.com/astropy/astropy/pull/18562. Lets see how this is picked up before merging this PR.

eendebakpt:
In astropy the only deprecated cases left are cases where we expect to use the new `ndarray._set_dtype` from this PR. So I think this PR is ready.

Alternatively, we split this PR to first merge the `_set_dtype` and merge the rest a bit later (when downstream has had some time to adapt).

mattip:
Since there is a PR ready to go, I don't think it is worth the effort to split this. 

Grepping around in pandas, it seems the only uses are in `self.dtype = ` in non-ndarray class `__init__` methods. Maybe I am optimistic.

doc/release/upcoming_changes/29244.deprecation.rst
```diff
@@ -0,0 +1,7 @@
+Setting the ``dtype`` attribute is deprecated
+---------------------------------------------
+Setting the dtype attribute is now deprecated since mutating
+an array is unsafe if an array is shared, especially by multiple
+threads.  As an alternative, you can create a view with a new dtype
+via `a.view(dtype=new_dtype)`.
+
```
eendebakpt:
```suggestion
via `array.view(dtype=new_dtype)`.
```

numpy/_core/src/multiarray/getset.c
```diff
@@ -522,6 +517,43 @@ array_descr_set(PyArrayObject *self, PyObject *arg, void *NPY_UNUSED(ignored))
     return -1;
 }
 
+static int
+non_unique_reference(PyObject *lhs)
+{
+    // Return 1 if we have a guaranteed non-unique reference
+    // If 0 the object can be unique or non-unique
```
eendebakpt:
```suggestion
    // When 0 is returned, the object can be unique or non-unique
```

numpy/_core/tests/test_deprecations.py
```diff
@@ -11,7 +11,7 @@
 import numpy as np
 import numpy._core._struct_ufunc_tests as struct_ufunc
 from numpy._core._multiarray_tests import fromstring_null_term_c_api  # noqa: F401
-from numpy.testing import assert_raises
+from numpy.testing import IS_PYPY, assert_raises, temppath
```
eendebakpt:
```suggestion
from numpy.testing import IS_PYPY, assert_raises
```

numpy/_core/src/multiarray/getset.c
```diff
@@ -495,6 +490,43 @@ array_descr_set(PyArrayObject *self, PyObject *arg, void *NPY_UNUSED(ignored))
...
+        PyErr_SetString(PyExc_AttributeError,
+                "Cannot delete array dtype");
+        return -1;
+    }
+
+    if (non_unique_reference((PyObject *)self)) {
+         // this will not emit deprecation warnings for all cases, but for most it will
+         // we skip unique references, so that we will not get a deprecation warning
+         // when array.view(new_dtype) is called
+         /* DEPRECATED 2025-06-20, NumPy 2.4 */
```
eendebakpt:
```suggestion
         /* DEPRECATED 2026-02-06, NumPy 2.5 */
```

ngoldbaum:
I think this needs a mailing list ping to be mergeable.

eendebakpt:
Reference to thread on mailing list archive: https://mail.python.org/archives/list/numpy-discussion@python.org/thread/FLSZFJ2TDLXHMFAUIYNM43VZQ7T73LZR/

ngoldbaum (approved):

ngoldbaum:
Thanks for your patience here @eendebakpt. I think we've done our due diligence.

@seberg, can I nominate you to hit the merge button if you agree?

jorenham:
this needs a rebase

seberg:
Sorry for missing this.  Let's put it in, thanks everyone.

While not the best, if you are hit by this and views don't work, you can use `._set_dtype()`, but please do leave a comment/issue with the use-case, so we know that it is used and also understand the why a bit.

neutrinoceros:
> if you are hit by this and views don't work, you can use `._set_dtype()`, but please do leave a comment/issue with the use-case,

hi there. I just used it in astropy to adapt a class that inherits from `numpy.ndarray`, assuming that's okay
https://github.com/astropy/astropy/pull/19540

seberg:
> I just used it in astropy to adapt a class that inherits from numpy.ndarray, assuming that's okay

I think the question here is whether there is a reason that you cannot follow the NumPy deprecation here for the `Column` class?

EDIT: I.e. can't you almost as easily fix the `yt` code to avoid it instead.

neutrinoceros:
Oh I *do* plan to follow the deprecation, but I can't introduce a deprecation in a patch release, so I'll only do it for astropy 8.0.0, while the patch I linked is meant to be backported as is to 7.2.x

> EDIT: I.e. can't you almost as easily fix the yt code to avoid it instead.

meaning ? feel free to join me in the convo downstream if you have guidance. Thanks !

seberg:
I am mostly wondering if you can't just "blame NumPy", i.e. this deprecation (from an `astropy` perspective) is related to the users NumPy version and not an astropy (minor) version "introducing it" (even if you react to it)?
The warning (even if astropy makes it more specific) could say it was introduced by NumPy 2.5?  (Maybe this is weird?!)

From a yt perspective, I guess the question is how important these path that magically convert integer arrays to float arrays of the same size are for users.
Admittedly, that seems a bit tricky to judge (and if there is an easy workaround for the user, the NumPy warning wouldn't make that very clear).

I suppose, my first feeling is that yt using `_set_dtype` (possibly with a more specific `DeprecationWarning` that again might blame NumPy) might be the better spot for this fix.  But I am happy to trust you either way.  Thanks for the info!  The yt use-case probably does need `_set_dtype()`, if -- hopefully -- just for transitioning.

---

@eendebakpt in case you want to follow up quickly, I missed that the warning says "2.4" rather than "2.5".

neutrinoceros:
> The warning (even if astropy makes it more specific) could say it was introduced by NumPy 2.5? (Maybe this is weird?!)

We've seen all kinds of complaints when upgrading numpy "breaks" astropy by introducing warnings (because many projects test with warnings treated as errors), so I'm reallyjust  taking the path of least resistance, but I don't disagree with you, personally.

>  in case you want to follow up quickly, I missed that the warning says "2.4" rather than "2.5".

already fixed at #31186 ;-)

rgommers:
@eendebakpt there is a single instance that I think you missed, and it's showing up as a warning in wheel build jobs now:

https://github.com/numpy/numpy/blob/8ed4764b28bc09db8cdc90169747b8776b68c2fa/numpy/_core/tests/test_records.py#L274

If that was by accident, would you be able to fix that up?

PR #29576, opened 2025-08-17
BUG: allow MaskedArray min/max with dtype=object

### LINKED ISSUE #9103: MaskedArray min/max methods do not work for dtype object
MaskedArray's min/max methods will raise an exception if its dtype is object. However ndarray min/max methods will work just fine. For example,

```
>>> test = np.asarray([1,2,3], dtype='object')
>>> test.min()
1

>>> test = np.ma.masked_array([1,2,3], dtype='object', mask=[True, False, False])
>>> test.min()
AttributeError: 'int' object has no attribute 'view'
```

These methods are useful for cases when having to use Python objects as dtype because your data contains BigInteger and BigDouble (i.e. int and Decimal respectively). 

The issue seems to lie in this line in ``np.ma.core.min``,

`result = self.filled(fill_value).min(axis=axis, out=out, **kwargs).view(type(self))`

The view call will fail when dtype is object, because min(...) will return e.g. int or Decimal, rather than some type of ndarray.

benburrill:
Fixes #9103.  There are actually 2 bugs here:
- The `AttributeError` is fixed by checking if the result of min/max is an `ndarray` before calling `.view()`.
- The `TypeError` (which is what you actually get when you run the [#9103](https://github.com/numpy/numpy/issues/9103#issue-228130721) reproducer, even though it is reported as the AttributeError) is fixed by adding a universal infinity filler object which (preferentially) compares greater than all other objects other than itself.  Previously, the string `'?'` was used as the default filler for object min/max, which was wrong even for the case of an array of string objects.

In the process of fixing this bug, I also found several other related issues by searching for uses of `.view()`.  I was initially planning to fix those in this PR, but I think some of them may require a bit more discussion.  I plan to open a new issue about that later.

### DIFF
--- a/numpy/ma/core.py
+++ b/numpy/ma/core.py
@@ -191,6 +191,39 @@ for v in ["Y", "M", "W", "D", "h", "m", "s", "ms", "us", "ns", "ps",
     default_filler["M8[" + v + "]"] = np.datetime64("NaT", v)
     default_filler["m8[" + v + "]"] = np.timedelta64("NaT", v)
 
+
+class _UniversalInfinity:
+    def __init__(self, neg=False):
+        self.neg = neg
+
+    def __le__(self, other):
+        return self.neg
+
+    def __lt__(self, other):
+        return self.neg and self != other
+
+    def __ge__(self, other):
+        return not self.neg
+
+    def __gt__(self, other):
+        return not self.neg and self != other
+
+    def __eq__(self, other):
+        return isinstance(other, _UniversalInfinity) and self.neg == other.neg
+
+    def __hash__(self):
+        return hash(self.neg)
+
+    def __neg__(self):
+        return _UniversalInfinity(not self.neg)
+
+    def __pos__(self):
+        return self
+
+    def __repr__(self):
+        return "-uinf" if self.neg else "uinf"
+
+
 float_types_list = [np.half, np.single, np.double, np.longdouble,
                     np.csingle, np.cdouble, np.clongdouble]
 
@@ -211,6 +244,8 @@ for sctype in ntypes.sctypeDict.values():
         min_val, max_val = info.min, info.max
     elif scalar_dtype.kind == "b":
         min_val, max_val = 0, 1
+    elif scalar_dtype.kind == "O":
+        min_val, max_val = -_UniversalInfinity(), _UniversalInfinity()
     else:
         min_val, max_val = None, None
 
--- a/numpy/ma/core.py
+++ b/numpy/ma/core.py
@@ -5942,8 +5942,9 @@ class MaskedArray(ndarray):
         # No explicit output
         if out is None:
             result = self.filled(fill_value).min(
-                axis=axis, out=out, **kwargs).view(type(self))
-            if result.ndim:
+                axis=axis, out=out, **kwargs)
+            if isinstance(result, ndarray) and result.ndim:
+                result = result.view(type(self))
                 # Set the mask
                 result.__setmask__(newmask)
                 # Get rid of Infs
@@ -6047,8 +6048,9 @@ class MaskedArray(ndarray):
         # No explicit output
         if out is None:
             result = self.filled(fill_value).max(
-                axis=axis, out=out, **kwargs).view(type(self))
-            if result.ndim:
+                axis=axis, out=out, **kwargs)
+            if isinstance(result, ndarray) and result.ndim:
+                result = result.view(type(self))
                 # Set the mask
                 result.__setmask__(newmask)
                 # Get rid of Infs
--- a/numpy/ma/tests/test_core.py
+++ b/numpy/ma/tests/test_core.py
@@ -1453,9 +1453,17 @@ class TestMaskedArrayArithmetic:
             assert masked_array([cmax, 0], mask=[0, 1]).min() == cmax
 
     def test_minmax_object_dtype(self):
-        a = masked_array([1, 2, 3], dtype=object)
-        assert a.min() == 1
-        assert a.max() == 3
+        one = 1
+        three = 3
+        a = masked_array([one, 2, three], dtype=object)
+        assert a.min() is one
+        assert a.max() is three
+
+        one_ao = np.array(1)
+        three_ao = np.array(3)
+        aoao = masked_array([one_ao, np.array(2), three_ao], dtype=object)
+        assert aoao.min() is one_ao
+        assert aoao.max() is three_ao
 
     @pytest.mark.parametrize("dtype", "bBiIqQ")
     @pytest.mark.parametrize("mask", [
--- a/numpy/ma/tests/test_core.py
+++ b/numpy/ma/tests/test_core.py
@@ -1452,12 +1452,13 @@ class TestMaskedArrayArithmetic:
             assert masked_array([-cmax, 0], mask=[0, 1]).max() == -cmax
             assert masked_array([cmax, 0], mask=[0, 1]).min() == cmax
 
+    @pytest.mark.parametrize("mask", [nomask, [False, True, False]])
     @pytest.mark.parametrize("arg1, arg2, arg3", [
         (1, 2, 3),
         (np.array(1), np.array(2), np.array(3)),
     ])
-    def test_minmax_object_dtype(self, arg1, arg2, arg3):
-        a = masked_array([arg1, arg2, arg3], dtype=object)
+    def test_minmax_object_dtype(self, arg1, arg2, arg3, mask):
+        a = masked_array([arg1, arg2, arg3], mask=mask, dtype=object)
         assert a.min() is arg1
         assert a.max() is arg3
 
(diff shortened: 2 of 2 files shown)


numpy/ma/tests/test_core.py
```diff
@@ -1452,6 +1452,19 @@ def test_minmax_dtypes(self):
...
+        three = 3
+        a = masked_array([one, 2, three], dtype=object)
+        assert a.min() is one
+        assert a.max() is three
+
+        one_ao = np.array(1)
+        three_ao = np.array(3)
+        aoao = masked_array([one_ao, np.array(2), three_ao], dtype=object)
+        assert aoao.min() is one_ao
+        assert aoao.max() is three_ao
```
tylerjereddy: Minor point, but you can make this more concise and more granular in error reporting via parametrization:

<details>

```diff
--- a/numpy/ma/tests/test_core.py
+++ b/numpy/ma/tests/test_core.py
@@ -1452,18 +1452,14 @@ def test_minmax_dtypes(self):
             assert masked_array([-cmax, 0], mask=[0, 1]).max() == -cmax
             assert masked_array([cmax, 0], mask=[0, 1]).min() == cmax
 
-    def test_minmax_object_dtype(self):
-        one = 1
-        three = 3
-        a = masked_array([one, 2, three], dtype=object)
-        assert a.min() is one
-        assert a.max() is three
-
-        one_ao = np.array(1)
-        three_ao = np.array(3)
-        aoao = masked_array([one_ao, np.array(2), three_ao], dtype=object)
-        assert aoao.min() is one_ao
-        assert aoao.max() is three_ao
+    @pytest.mark.parametrize("arg1, arg2, arg3", [
+        (1, 2, 3),
+        (np.array(1), np.array(2), np.array(3)),
+    ])
+    def test_minmax_object_dtype(self, arg1, arg2, arg3):
+        a = masked_array([arg1, arg2, arg3], dtype=object)
+        assert a.min() is arg1
+        assert a.max() is arg3
```

</details>

numpy/ma/core.py
```diff
@@ -5942,8 +5942,9 @@ def min(self, axis=None, out=None, fill_value=None, keepdims=np._NoValue):
         # No explicit output
         if out is None:
             result = self.filled(fill_value).min(
-                axis=axis, out=out, **kwargs).view(type(self))
-            if result.ndim:
+                axis=axis, out=out, **kwargs)
```
tylerjereddy, line 5980: I'm a little confused here I think. When I write a new regression test that more closely matches the original reproducer, like below, it errors out here.

```diff
--- a/numpy/ma/tests/test_core.py
+++ b/numpy/ma/tests/test_core.py
@@ -1461,6 +1461,11 @@ def test_minmax_object_dtype(self, arg1, arg2, arg3):
         assert a.min() is arg1
         assert a.max() is arg3
 
+
+    def test_gh_9103(self):
+        arr = masked_array([1, 2, 3], dtype='object', mask=[True, False, False])
+        assert arr.min() == 2
```

<details>

<summary> Traceback </summary>

```
_____________________________________________________________________________ TestMaskedArrayArithmetic.test_gh_9103 ______________________________________________________________________________

self = <numpy.ma.tests.test_core.TestMaskedArrayArithmetic object at 0x1033739b0>

    def test_gh_9103(self):
        arr = masked_array([1, 2, 3], dtype='object', mask=[True, False, False])
>       assert arr.min() == 2

arr        = masked_array(data=[--, 2, 3],
             mask=[ True, False, False],
       fill_value=np.str_('?'),
            dtype=object)
self       = <numpy.ma.tests.test_core.TestMaskedArrayArithmetic object at 0x1033739b0>

numpy/ma/tests/test_core.py:1467: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
numpy/ma/core.py:5944: in min
    result = self.filled(fill_value).min(
        _mask      = array([ True, False, False])
        axis       = None
        fill_value = None
        keepdims   = <no value>
        kwargs     = {}
        newmask    = np.False_
        out        = None
        self       = masked_array(data=[--, 2, 3],
             mask=[ True, False, False],
       fill_value=np.str_('?'),
            dtype=object)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

a = array(['?', 2, 3], dtype=object), axis = None, out = None, keepdims = False, initial = <no value>, where = True

    def _amin(a, axis=None, out=None, keepdims=False,
              initial=_NoValue, where=True):
>       return umr_minimum(a, axis, None, out, keepdims, initial, where)
E       TypeError: '<=' not supported between instances of 'str' and 'int'

a          = array(['?', 2, 3], dtype=object)
axis       = None
initial    = <no value>
keepdims   = False
out        = None
where      = True

numpy/_core/_methods.py:45: TypeError

```

</details>

benburrill: That's a great catch, I totally neglected to even test the masked array with an actual mask!  I would say that is a separate bug from the one I'm trying to fix though.  Actually it seems like it may be a regression that has occurred in the 8 years since #9103 was reported.  Even without my change, in current versions of numpy, the #9103 reproducer now gives that TypeError rather than the AttributeError.  Of course you can work around this by setting an appropriate `fill_value` other than '?' (and arguably you should have a `fill_value` that matches the actual data you're storing in the array).  But I'm not sure why it worked at all in 2017.  I'll look into it a bit more.

benburrill: Hmm... even with the contemporary numpy 1.13.0 from 2017, I'm seeing the `TypeError` rather than the `AttributeError` on the #9103 reproducer.  I only get the `AttributeError` if I remove the mask or set a `fill_value`.  I'd test older versions, but I'm running into some segfaults.

tylerjereddy:
I haven't thought too much more about this, but if the discussion above is suggesting that this PR doesn't fix the reproducer in the original ticket, it might be sensible to at least not have this PR set to automatically close/address the ticket, etc.

Then the scope might be clearer/a bit less confusing--it may very well be just fine to address the separate issue here without closing that ticket.

benburrill:
That's reasonable.  I changed it to no longer close #9103.

numpy/ma/core.py
```diff
@@ -5942,8 +5942,9 @@ def min(self, axis=None, out=None, fill_value=None, keepdims=np._NoValue):
...
+            if isinstance(result, ndarray) and result.ndim:
```
benburrill, line 5981: There is an (extreme) edge case I am aware of here that my code fails to address: ideally, I think that

```python
x = np.array([1, 2])
assert np.fromiter([x], dtype=object).view(np.ma.MaskedArray).min() is x
```

should pass, but with my change, x will get erroneously converted to a masked array when returned by `min`.

This is identical to the old behavior, so it's not necessarily a problem, but it is a failure of my assumptions: initially when working on this PR, I thought I only needed to worry about arrays containing 0-D array objects, since arrays of higher dimension are not orderable (and so you'd just get an error when calling `min()`, as you should).  But of course I forgot to consider that `min()` will succeed regardless if there is only one item.  Also, as another similar edge case, higher dimensional arrays actually **are** orderable if they themselves only contain one item (not just if they are 0-D).

I'm not sure if this is actually worth fixing though.  Thoughts?

numpy/ma/core.py
benburrill: I've been trying to think if there is any "good" way to fix the `TypeError` issue, besides the obvious way of creating a universal infinity object that compares greater to any other object regardless of type[^1], and putting that in `min_filler`, but really I don't think there is (short of completely reimplementing min).  That said, it should be a very easy fix to create such an object, so I'd be willing to add that to the PR if you think it is an acceptable solution.

[^1]: ...as long as those other objects aren't also playing the same game of overriding comparison operators for all other types (eg numpy array broadcasting) -- which is part of why I don't consider it a "good" solution even though it should work for most cases.

benburrill: I went ahead with my proposal to add a universal infinity filler to fix the TypeError issue.  Marking as resolved, assuming you consider this to be an acceptable solution.

benburrill:
I have added my proposed fix for the `TypeError` issue to this PR, so I have changed it back to close #9103.

benburrill:
I would appreciate additional feedback, especially regarding the introduction of the `_UniversalInfinity` filler object.  Let me know if you want to see more tests, etc.

PR #27784, opened 2024-11-17
BUG: ``sinc``: fix underflow for float16 

lucascolley:
While reviewing https://github.com/data-apis/array-api-extra/pull/20, @jakirkham suggested these small improvements to the implementation of `sinc`.

If the change from `1.0e-20` to `np.finfo(x.dtype).smallest_normal` is classed as breaking, we can leave that out.

PR #29580, opened 2025-08-17
BUG: Fix sinc for custom objects

fjosw:
This PR introduces a simple fix for the sinc function which stopped working for minimal custom objects. The problem was introduced in https://github.com/numpy/numpy/pull/27784 / numpy 2.3 and is caused by an explicit check for the kind of the associated dtype in order to determine an appropriate eps. This PR makes the check more robust.

Minimal breaking example:
```python
import numpy as np

class MyObject:

    def __init__(self, value):
        self.value = value

    def __rmul__(self, y):
        return self.value * y

number = MyObject(2)
print(np.sinc(number))
```

Fails with
```
  File ".../numpy_repro.py", line 12, in <module>
    print(np.sinc(number))
          ^^^^^^^^^^^^^^^
  File ".../python3.12/site-packages/numpy/lib/_function_base_impl.py", line 3852, in sinc
    eps = np.finfo(x.dtype).eps if x.dtype.kind == "f" else 1e-20
                                   ^^^^^^^
AttributeError: 'float' object has no attribute 'dtype'
```

### DIFF
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -3856,8 +3856,10 @@ def sinc(x):
     """
     x = np.asanyarray(x)
     x = pi * x
-    # Hope that 1e-20 is sufficient for objects...
-    eps = np.finfo(x.dtype).eps if x.dtype.kind == "f" else 1e-20
+    if hasattr(x, "dtype") and getattr(x.dtype, "kind", None) == "f":
+        eps = np.finfo(x.dtype).eps
+    else:
+        eps = 1e-20
     y = where(x, x, eps)
     return sin(y) / y
 


numpy/lib/_function_base_impl.py
```diff
@@ -3856,8 +3856,10 @@ def sinc(x):
     """
     x = np.asanyarray(x)
     x = pi * x
-    # Hope that 1e-20 is sufficient for objects...
-    eps = np.finfo(x.dtype).eps if x.dtype.kind == "f" else 1e-20
+    if hasattr(x, "dtype") and getattr(x.dtype, "kind", None) == "f":
+        eps = np.finfo(x.dtype).eps
+    else:
+        eps = 1e-20
```
tylerjereddy, line 3862: I suppose if it is worth fixing it is worth adding a regression test inspired by your reproducer like below the fold.

<details>

```diff
--- a/numpy/lib/tests/test_function_base.py
+++ b/numpy/lib/tests/test_function_base.py
@@ -2363,6 +2363,19 @@ def test_masked(self):
 
 class TestSinc:
 
+    def test_sinc_obj(self):
+        class MyObject:
+
+            def __init__(self, value):
+                self.value = value
+
+            def __rmul__(self, y):
+                return self.value * y
+
+        number = MyObject(2)
+        assert_allclose(np.sinc(number), np.sinc(2))
```

</details>

But maybe best to wait for feedback from a more active dev before putting more time into it.

jorenham:
This way of using `sinc` is not (and never has been) documented. Before 2.3, these `__rmul__` objects were accepted by accident; not by design. It's a clever hack, but it's not a feature. 

If we were to actively restore support for this hack, and add regression tests for it, then we're effectively adding official support for it. That means that would also have to document it, and modify the typing stubs to reflect this. But because I haven't seen a convincing use-case, and am not able to think of one either, I don't think that we should. And even if we wouldn't document it, and just keep it something like an "informal feature", then we risk setting a precedent for hacks like these.

So long story short; I don't think we should let `sinc` such `__rmul__` objects. It already accepts array-likes, which are already flexible enough in my view.

fjosw:
Thanks a lot for your feedback, I agree that this is a bit of hack and not really worth being officially supported.

I still think that the current way of determining `eps` in the `sinc` function is not clean. Playing a bit more with `sinc` and objects I observed that
```python
np.sinc(np.array(1.0, dtype=object))
```
fails with the unintuitive
```python
  File ".../python3.12/site-packages/numpy/lib/_function_base_impl.py", line 3852, in sinc
    eps = np.finfo(x.dtype).eps if x.dtype.kind == "f" else 1e-20
                                   ^^^^^^^
AttributeError: 'float' object has no attribute 'dtype'
```
while I would expect this behaviour
```python
>>> np.sinc(np.array([1.0], dtype=object))
AttributeError: 'float' object has no attribute 'sin'

  File ".../python3.12/site-packages/numpy/lib/_function_base_impl.py", line 3854, in sinc
    return sin(y) / y
           ^^^^^^
TypeError: loop of ufunc does not support argument 0 of type float which has no callable sin method
```

Not a big issue, though, so feel free to close the PR 👍

seberg:
We should fix this, but by using `np.multiply(x, pi, out=...)` to ensure an array result.  That said, that might break other objects that don't implement `__array_ufunc__`, so dunno if one should backport that.

EIDT: Although, dunno, there is some risk in backporting that, but this might be a bit annoying regression also.

numpy/lib/_function_base_impl.py
seberg: Yeah, we should definitely add a test. I do still wonder if we can find a slightly nicer work-around for backporting, but can't think of one.

seberg:
@fjosw stumbled on this, 2.4 is basically out now. So if you are interested in picking it up again. I think we could do that `out=...` now which would fix what you saw there (if not also fine).

PR #29591, opened 2025-08-19
TYP: Fix reverse operators typing for objects with __array__ and __array_ufunc__

MengAiDev:
Fixes #29486

## Description
This PR addresses an issue with static typing of reverse operators in NumPy's `ndarray` class. When using objects that implement both `__array__` and `__array_ufunc__` (like pandas Series) in reverse operations with NumPy arrays, the static typing incorrectly reports the result as an `ndarray` instead of the appropriate type from the left-hand side object.

The solution implements a Protocol-based approach that adds overloads to the reverse operators (`__radd__`, `__rsub__`, `__rmul__`, etc.) in the `ndarray` class. These overloads check if the left-hand side operand implements the corresponding forward operator and, if so, delegate to that operator, returning its result type.

### DIFF
--- a/numpy/__init__.pyi
+++ b/numpy/__init__.pyi
@@ -153,6 +153,7 @@ from numpy._typing._callable import (
     _FloatMod,
     _FloatDivMod,
     _NumberOp,
+    _CanAdd,
     _ComparisonOpLT,
     _ComparisonOpLE,
     _ComparisonOpGT,
@@ -207,6 +208,7 @@ from typing import (
     SupportsIndex,
     TypeAlias,
     TypedDict,
+    TypeVar,
     final,
     overload,
     type_check_only,
@@ -2909,6 +2911,9 @@ class ndarray(_ArrayOrScalarCommon, Generic[_ShapeT_co, _DTypeT_co]):
     def __add__(self: NDArray[Any], other: _ArrayLikeObject_co, /) -> Any: ...
 
     # Keep in sync with `MaskedArray.__radd__`
+    # First try to delegate to the left-hand side if it implements __add__
+    @overload
+    def __radd__(self, other: _CanAdd[Self, _T], /) -> _T: ...
     @overload  # signature equivalent to __add__
     def __radd__(self: NDArray[_NumberT], other: int | np.bool, /) -> ndarray[_ShapeT_co, dtype[_NumberT]]: ...
     @overload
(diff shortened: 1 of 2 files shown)


numpy/__init__.pyi
```diff
@@ -207,6 +208,7 @@ from typing import (
     SupportsIndex,
     TypeAlias,
     TypedDict,
+    TypeVar,
```
jorenham: The `TypeVar` should be imported from `typing_extensions` because we use the PEP 696 `default` kwarg, which requires Python 3.13+

numpy/_typing/_callable.pyi
```diff
@@ -42,6 +42,7 @@ from ._scalars import _BoolLike_co, _IntLike_co, _NumberLike_co
 
 _T1 = TypeVar("_T1")
 _T2 = TypeVar("_T2")
+_T = TypeVar("_T")
```
jorenham: unused

numpy/_typing/_callable.pyi
```diff
@@ -360,3 +361,9 @@ class _ComparisonOpGE(Protocol[_T1_contra, _T2_contra]):
     def __call__(self, other: _NestedSequence[_SupportsGT], /) -> NDArray[np.bool]: ...
     @overload
     def __call__(self, other: _SupportsGT, /) -> np.bool: ...
+
+@type_check_only
+class _CanAdd(Protocol[_T1_contra, _T2_co]):
+    """Protocol for objects that can perform addition operations."""
+    @overload
```
jorenham: only one method signature, so this should be removed

```suggestion
```

jorenham (changes requested):

numpy/__init__.pyi
jorenham, line 1: I can see how you might want to use `_CanAdd`  for `__radd__`, but it doesn't make much sense to also use that for other reflected binops like `__rpow__`

jorenham:
Thanks for this PR; it's always good to see interest in improving NumPy's type annotations.

But keep in mind that every commit you push notifies us, and that running the CI isn't free. So please run the mypy tests locally first (`spin mypy`) before you push any changes.

Also note that this is a *very* difficult problem to solve, and I'm still not sure what the right approach is, even though I do this for a living. Don't take this the wrong way, but based on your changes so far, it looks to me like you're not an expert in writing typing stubs. I'm a bit worried that the time that your generously investing in this might go to waste, due to its complexity.

So perhaps it would be more efficient (and more fun) to work on a different numpy typing issue? Some options that come to mind are #29595, or annotating one of the untyped `numpy.ma` functions.

jorenham:
Considering the mypy_primer output and the reasons from my previous comment, I'm going to close this.

PR #29592, opened 2025-08-19
ENH: Extract shape validation logic in _convert_from_tuple

richardsmythe:
## Summary
Refactors the `_convert_from_tuple` function in `descriptor.c` by extracting shape validation logic into a helper function `_validate_shape_dims`. This improves code organization and readability while preserving all existing behavior and error messages.

## Motivation
I'm refactoring the _convert_from_tuple function to improve code maintainability and readability by extracting validation logic into a helper function. This long function currently handles multiple responsibilities in a single ~150-line function, making it difficult to understand, test, and maintain. I propose breaking it down into smaller, single-purpose functions that are easier to reason about and modify.

## Changes Made
- **Extracted `_validate_shape_dims` helper function** from `_convert_from_tuple`
- **Preserved exact error messages**: "dimension smaller then zero" and "does not fit into a C int"
- **Added comprehensive tests** in `test_dtype.py` to ensure validation logic works correctly

## Testing
- Added `test_validate_shape_dims_helper_function()` as well as a test
- All existing tests pass (46,487 passed, 3 pre-existing unrelated failures)
- Test coverage: 93% on test_dtype.py
- Verified no behavioral changes or regressions

## Files Changed
- `numpy/_core/src/multiarray/descriptor.c`: Refactored shape validation logic
- `numpy/_core/tests/test_dtype.py`: Added test that verifies errors are returned as before. Existing tests cover edge cases.

Fixes no specific issue - this is a code quality improvement.

### DIFF
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -240,6 +240,27 @@ is_datetime_typestr(char const *type, Py_ssize_t len)
     return 0;
 }
 
+/* Helper function to validate shape dimensions for fixed-type tuple */
+static int
+_validate_shape_dims(PyArray_Dims *shape)
+{
+    for (int i = 0; i < shape->len; i++) {
+        if (shape->ptr[i] < 0) {
+            PyErr_SetString(PyExc_ValueError,
+                            "invalid shape in fixed-type tuple: "
+                            "dimension smaller then zero.");
+            return -1;
+        }
+        if (shape->ptr[i] > NPY_MAX_INT) {
+            PyErr_SetString(PyExc_ValueError,
+                            "invalid shape in fixed-type tuple: "
+                            "dimension does not fit into a C int.");
+            return -1;
+        }
+    }
+    return 0;
+}
+
 static PyArray_Descr *
 _convert_from_tuple(PyObject *obj, int align)
 {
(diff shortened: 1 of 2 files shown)


ngoldbaum:
It's not clear to me: what motivated this change? IMO refactoring code to be in a function just for the sake of it doesn't help readability. Are there other places we want to re-use this logic?

Also the test is very complicated - what new things is it testing that weren't tested before?

charris:
>It's not clear to me: what motivated this change? 

I wonder about AI, it tends to do such things.

richardsmythe:
> It's not clear to me: what motivated this change? IMO refactoring code to be in a function just for the sake of it doesn't help readability. Are there other places we want to re-use this logic?
> 
> Also the test is very complicated - what new things is it testing that weren't tested before?

Thanks for the feedback.

The motivation is that the original _convert_from_tuple function was handling two responsibilities, converting tuple format to dtype structure and validating shape dimensions . I think that extracting the validation logic into _validate_shape_dims, we separate these concerns, making the main function easier to understand and the validation logic more explicit, instead of the previous inline loop with multiple conditions. Also it uses the same goto fail pattern as per the rest of the function.

You're absolutely right about the test complexity. I've simplified it to focus only on what I changed, making sure the refactored validation preserves the exact original error behaviour for the two validation cases. The existing tests cover the edge cases so it makes sense to simplify the test. 

Thanks for taking the time to review this.

ngoldbaum:
IMO the new C function doesn't add anything, I'd prefer to leave it as-is. 

> I think that extracting the validation logic into _validate_shape_dims, we separate these concerns, making the main function easier to understand and the validation logic more explicit, instead of the previous inline loop with multiple conditions. Also it uses the same goto fail pattern as per the rest of the function.

But now whoever is reading this code needs to scroll 100 lines up to see what the new function you added does. If we wanted to re-use this logic elsewhere or if this change was motivated by some kind of functional change to NumPy or a future bugfix I might be more inclined to accept this.

The new test is fine, I guess?

By the way, if you're using an LLM to draft your responses or to edit the numpy codebase, we'd appreciate it if you disclose that.

richardsmythe:
> IMO the new C function doesn't add anything, I'd prefer to leave it as-is.
> 
> > I think that extracting the validation logic into _validate_shape_dims, we separate these concerns, making the main function easier to understand and the validation logic more explicit, instead of the previous inline loop with multiple conditions. Also it uses the same goto fail pattern as per the rest of the function.
> 
> But now whoever is reading this code needs to scroll 100 lines up to see what the new function you added does. If we wanted to re-use this logic elsewhere or if this change was motivated by some kind of functional change to NumPy or a future bugfix I might be more inclined to accept this.
> 
> The new test is fine, I guess?
> 
> By the way, if you're using an LLM to draft your responses or to edit the numpy codebase, we'd appreciate it if you disclose that.

No problem, understood. I have closed the PR.

Issue #29552, opened 2025-08-12
TST: Make test suite more thread-safe


bwhitt7:
Hello! I'm currently interning at Quansight, with @ngoldbaum as my mentor. We are working on making NumPy's test suite more thread safe in an effort to make NumPy have better support for free-threaded Python.

When running NumPy's test suite with [pytest-run-parallel](https://github.com/Quansight-Labs/pytest-run-parallel), there are numerous failures, with many thread-unsafe methods and modules causing tests to fail for various reasons (such as AssertionErrors due to global state being modified between tests, or the most common: AttributeErrors due to pytest's xunit setup and teardown methods being thread-unsafe). I have identified all these thread-unsafe tests, and have begun fixing some of these issues.

The thread-unsafe tests are documented [in this spreadsheet](https://docs.google.com/spreadsheets/d/1mOlsjDz7S4dXWTA1R6HHi5Tilis7XOJeLuNTwZkKxAU/edit?usp=sharing), as well as my progress on fixing them in my own fork. Feel free to take a look, and feel free to modify the spreadsheet with your own findings if you decide to tackle some of these.

### Progress

- [x] Update classic pytest xunit setup and teardown methods into regular methods or fixtures. These methods are thread-unsafe, often not running before the tests and resulting in AttributeErrors. This is the large majority of failures found, and will likely expose more thread safety issues once fixed.
- [x] Some tests require a lot of memory, and will crash when ran on multiple threads. Marking them with pytest.mark.parallel_threads(n) to only run on _n_ threads should help.
- [x] Update tests that use np.random.seed() to use Generator or an instance of RandomState instead. np.random is global state, which often results in AssertionErrors, with random number generation not acting as expected.
- [x] For tests that use tmp_path and tmpdir, ensure files generated by tests on parallel threads are unique. Could utilize uuids? pytest-run-parallel is currently looking into patching tmp_path to be thread safe.
- [x] Fix up tests that race to modify parameterized values, potentially use a fixture or method generator instead.
- [x] Fix or mark tests that use pytest fixtures that are thread-unsafe (capfd, recwarn).
- [x] Mark or remove unittest.mock.patch.
- [x] Fix tests that seem to have thread safety issues with warnings.
- [x] Mark tests that have unreliable results with hypothesis.
- [x] Mark tests that use sys.getrefcount() and produce unexpected results.
- [x] Mark tests that use assert_no_gc_cycle.
[...]

PR #29596, opened 2025-08-19

bwhitt7:
This PR is related to #29552, with the goal of making NumPy's testing suite more thread safe. This PR is "testing the waters" with a test file that had a lot of simple xunit setup methods. All setup methods were replaced with methods that are manually called in each test. This makes the tests more thread safe, since xunit setup methods work strangely with threads. An example of a change:

```python
# TestComparisons
def setup_method(self): # setup method to be replaced
    self.A = np.array([['abc', 'abcc', '123'],
                       ['789', 'abc', 'xyz']]).view(np.char.chararray)
    self.B = np.array([['efg', 'efg', '123  '],
                       ['051', 'efgg', 'tuv']]).view(np.char.chararray)

    def test_not_equal(self):
        assert_array_equal((self.A != self.B), # calling self to access variables
                           [[True, True, False], [True, True, True]])
```

to

```python
# TestComparisons
def _create_array_a(self): # "creation" methods used instead
    return np.array([['abc', 'abcc', '123'],
                        ['789', 'abc', 'xyz']]).view(np.char.chararray)

def _create_array_b(self):
    return np.array([['efg', 'efg', '123  '],
                        ['051', 'efgg', 'tuv']]).view(np.char.chararray)

def test_not_equal(self):
    (A, B) = (self._create_array_a(), self._create_array_b()) # calling creation methods to set up variables
    assert_array_equal((A != B),
                       [[True, True, False], [True, True, True]])
```

With this PR, you should be able to call the tests from this file with [pytest-run-parallel](github.com/Quansight-Labs/pytest-run-parallel) and have all tests successfully run in parallel threads (and also run normally without pytest-run-parallel).

### DIFF
--- a/numpy/_core/tests/test_defchararray.py
+++ b/numpy/_core/tests/test_defchararray.py
@@ -84,7 +84,6 @@ class TestBasic:
         assert_equal(A.itemsize, 4)
         assert_(issubclass(A.dtype.type, np.str_))
 
-
 class TestVecString:
     def test_non_existent_method(self):
 
@@ -135,7 +134,6 @@ class TestVecString:
 
         assert_raises(ValueError, fail)
 
-
 class TestWhitespace:
     def test1(self):
         A = np.array([['abc ', '123  '],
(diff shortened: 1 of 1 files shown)


numpy/_core/tests/test_defchararray.py
```diff
@@ -136,121 +137,130 @@ def fail():
...
-                           ['789', 'abc', 'xyz']]).view(np.char.chararray)
-        self.B = np.array([['efg', 'efg', '123  '],
-                           ['051', 'efgg', 'tuv']]).view(np.char.chararray)
+    def _create_array_a(self):
+        return np.array([['abc', 'abcc', '123'],
+                            ['789', 'abc', 'xyz']]).view(np.char.chararray)
+
+    def _create_array_b(self):
+        return np.array([['efg', 'efg', '123  '],
+                            ['051', 'efgg', 'tuv']]).view(np.char.chararray)
```
ngoldbaum: IMO this should be 

```
def _create_arrays(self):
    return (
        np.array([['abc', 'abcc', '123'], ['789', 'abc', 'xyz']]).view(np.char.chararray),
        np.array([['efg', 'efg', '123  '], ['051', 'efgg', 'tuv']]).view(np.char.chararray)
    )
```

Since in all cases this is what you want anyway, it's a little clearer just with one function call.

numpy/_core/tests/test_defchararray.py
```diff
@@ -136,121 +137,130 @@ def fail():
...
-        # Array with longer strings, > MEMCHR_CUT_OFF in code.
-        self.C = (np.array(['ABCDEFGHIJKLMNOPQRSTUVWXYZ',
-                            '01234567890123456789012345'])
-                  .view(np.char.chararray))
 
     def test_len(self):
-        assert_(issubclass(np.char.str_len(self.A).dtype.type, np.integer))
-        assert_array_equal(np.char.str_len(self.A), [[5, 0], [5, 9], [12, 5]])
-        assert_array_equal(np.char.str_len(self.B), [[3, 0], [5, 9], [12, 5]])
+        (A, B) = (self._create_array_a(), self._create_array_b())
```
ngoldbaum: If a test only uses `A`, you could do `(A, _) = self._create_arrays()` to discard `B` without binding it to a variable. By convention in Python a variable named `_` is used for this purpose.

jorenham: also, no need for parentheses here (on both sides)

numpy/_core/tests/test_defchararray.py
bwhitt7: I was considering that, however some classes inherit from TestComparisons (TestComparisonsMixed1 and TestComparisonsMixed2) and give their own values for the A or B arrays. I figured keeping the creation methods separate would work best here so they would only need to overwrite the one they are changing, instead of redeclaring both arrays. How if that would be fine, I can definitely change this to be for both arrays.

numpy/_core/tests/test_defchararray.py
```diff
@@ -84,6 +84,7 @@ def test_from_unicode(self):
         assert_equal(A.itemsize, 4)
         assert_(issubclass(A.dtype.type, np.str_))
 
+
```
jorenham:
```suggestion
```

jorenham (commented):
Thanks for this PR. 

These changes caused several array definitions to not be aligned anymore, could you fix those?

numpy/_core/tests/test_defchararray.py
ngoldbaum: In that case maybe you could have a slightly less wordy API? Something like:

```
def A(self):
    return np.array([['abc', 'abcc', '123'], ['789', 'abc', 'xyz']]).view(np.char.chararray)
```

And then later use it like `A = self.A()`. That way you can keep them separate.

bwhitt7:
Hopefully this push should address your suggestions @ngoldbaum @jorenham

numpy/_core/tests/test_defchararray.py
```diff
@@ -134,123 +134,126 @@ def fail():
...
-                           ['789', 'abc', 'xyz']]).view(np.char.chararray)
-        self.B = np.array([['efg', 'efg', '123  '],
-                           ['051', 'efgg', 'tuv']]).view(np.char.chararray)
+    def A(self):
+        return np.array([['abc', 'abcc', '123'],
+                            ['789', 'abc', 'xyz']]).view(np.char.chararray)
+
+    def B(self):
+        return np.array([['efg', 'efg', '123  '],
+                            ['051', 'efgg', 'tuv']]).view(np.char.chararray)
```
ngoldbaum:
```suggestion
    def A(self):
        return np.array([['abc', 'abcc', '123'],
                         ['789', 'abc', 'xyz']]).view(np.char.chararray)

    def B(self):
        return np.array([['efg', 'efg', '123  '],
                         ['051', 'efgg', 'tuv']]).view(np.char.chararray)
```

Let's make these formatted a little more nicely

numpy/_core/tests/test_defchararray.py
```diff
@@ -134,123 +134,126 @@ def fail():
...
-        # assert_array_equal(self.A.count('\0'), [[0, 0], [0, 0], [1, 0]])
-        assert_array_equal(self.A.count('a', 0, 2), [[1, 0], [0, 0], [0, 0]])
-        assert_array_equal(self.B.count('a'), [[0, 0], [0, 1], [0, 0]])
-        assert_array_equal(self.B.count('123'), [[0, 0], [1, 0], [1, 0]])
-        # assert_array_equal(self.B.count('\0'), [[0, 0], [0, 0], [1, 0]])
+        # assert_array_equal(A.count('\0'), [[0, 0], [0, 0], [1, 0]])
+        assert_array_equal(A.count('a', 0, 2), [[1, 0], [0, 0], [0, 0]])
+        assert_array_equal(B.count('a'), [[0, 0], [0, 1], [0, 0]])
+        assert_array_equal(B.count('123'), [[0, 0], [1, 0], [1, 0]])
+        # assert_array_equal(B.count('\0'), [[0, 0], [0, 0], [1, 0]])
```
ngoldbaum: While we're at it, let's go ahead and delete these commented out tests

```suggestion
        assert_array_equal(A.count('a', 0, 2), [[1, 0], [0, 0], [0, 0]])
        assert_array_equal(B.count('a'), [[0, 0], [0, 1], [0, 0]])
        assert_array_equal(B.count('123'), [[0, 0], [1, 0], [1, 0]])
```

numpy/_core/tests/test_defchararray.py
```diff
@@ -270,103 +273,119 @@ def test_find(self, dtype, encode):
                            [[1, -1], [-1, -1], [-1, -1]])
         assert_array_equal(A.find([encode('1'), encode('P')]),
                            [[-1, -1], [0, -1], [0, 1]])
-        C = self.C.astype(dtype)
+        C = (np.array(['ABCDEFGHIJKLMNOPQRSTUVWXYZ',
+                            '01234567890123456789012345'])
```
ngoldbaum: Let's make this line up too

```suggestion
        C = (np.array(['ABCDEFGHIJKLMNOPQRSTUVWXYZ',
                       '01234567890123456789012345'])
```

ngoldbaum (commented):
Looking good! Just a couple small suggestions.

bwhitt7:
@ngoldbaum Made the changes! Found a few other spots where the code didn't lineup and tried fixing it.

ngoldbaum:
Thanks @bwhitt7!

PR #29598, opened 2025-08-20
BUG: fix for evaluation of random_f and random_standard_cauchy in distributions.c

zuhu2195:
C program does not guarantee the order of evaluation of its operands, as a result if we have a function that returns f1()/f2(),
here f2() may get evaluated first. consider the following program,

```
#include <stdio.h>

/* Dummy bit generator state */
typedef struct {
    int call;
} bitgen_t;

/* Deterministic "random_standard_normal" function */
double random_standard_normal(bitgen_t *bitgen_state) {
    if (bitgen_state->call == 0) {
        bitgen_state->call++;
        return 8.0;   /* first fixed value */
    } else {
        return 2.0;   /* second fixed value */
    }
}

/* Function returning ratio of two standard normals */
double random_ratio(bitgen_t *bitgen_state) {
    return random_standard_normal(bitgen_state) /
           random_standard_normal(bitgen_state);
}

int main(void) {
    bitgen_t state = {0}; /* no randomness */

    double val = random_ratio(&state);
    printf("Deterministic ratio value: %f\n", val);

    return 0;
}
```

here the function random_standard_normal uses bitgen_state struct to evaluate the first and second call, but since C guarantees no order the first call might be f2() which can cause return of 8.0 in f2() and 2.0 in f1() clearly the result returned will be 2.0/8.0 which if 1/4, however if f1() evaluates first result returned will be 8.0/2.0 which is 4.

### DIFF
--- a/numpy/random/src/distributions/distributions.c
+++ b/numpy/random/src/distributions/distributions.c
@@ -470,7 +470,7 @@ double random_chisquare(bitgen_t *bitgen_state, double df) {
 double random_f(bitgen_t *bitgen_state, double dfnum, double dfden) {
   double subexpr1 = random_chisquare(bitgen_state, dfnum) * dfden;
   double subexpr2 = random_chisquare(bitgen_state, dfden) * dfnum;
-  return subexpr1/subexpr2;
+  return subexpr1 / subexpr2;
 }
 
 double random_standard_cauchy(bitgen_t *bitgen_state) {
--- a/numpy/random/src/distributions/distributions.c
+++ b/numpy/random/src/distributions/distributions.c
@@ -476,7 +476,7 @@ double random_f(bitgen_t *bitgen_state, double dfnum, double dfden) {
 double random_standard_cauchy(bitgen_t *bitgen_state) {
   double subexpr1 = random_standard_normal(bitgen_state);
   double subexpr2 = random_standard_normal(bitgen_state);
-  return subexpr1/subexpr2;
+  return subexpr1 / subexpr2;
 }
 
 double random_pareto(bitgen_t *bitgen_state, double a) {
--- a/numpy/random/src/distributions/distributions.c
+++ b/numpy/random/src/distributions/distributions.c
@@ -468,12 +468,15 @@ double random_chisquare(bitgen_t *bitgen_state, double df) {
 }
 
 double random_f(bitgen_t *bitgen_state, double dfnum, double dfden) {
-  return ((random_chisquare(bitgen_state, dfnum) * dfden) /
-          (random_chisquare(bitgen_state, dfden) * dfnum));
+  double subexpr1 = random_chisquare(bitgen_state, dfnum) * dfden;
+  double subexpr2 = random_chisquare(bitgen_state, dfden) * dfnum;
+  return subexpr1/subexpr2;
 }
 
 double random_standard_cauchy(bitgen_t *bitgen_state) {
-  return random_standard_normal(bitgen_state) / random_standard_normal(bitgen_state);
+  double subexpr1 = random_standard_normal(bitgen_state);
+  double subexpr2 = random_standard_normal(bitgen_state);
+  return subexpr1/subexpr2;
 }
 
 double random_pareto(bitgen_t *bitgen_state, double a) {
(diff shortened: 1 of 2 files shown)


mattip:
Makes sense, as [the standard is clear](https://en.cppreference.com/w/c/language/eval_order.html) that order of evaluation is arbitrary. In practice, it would seem to be consistent, but this PR corrects a possible defect.

WarrenWeckesser (approved):
To be clear: the potentially different compiler-dependent behaviors do not mean that the stream of random variates will be incorrect.  Whether the numerator or denominator is evaluated first does not affect the correctness of the calculation.  The behavior being compiler-dependent means that the streams of variates generated by code compiled with different compilers might be different.  But the different streams are both "correct" (in as much as both are numerical approximations of the underlying random process).

Having said that, I agree that it is a good idea to fix the ambiguous order of evaluation.  While in general we can't *guarantee* identical streams across different platforms and compilers, it seems reasonable to remove potential sources of difference when we can.

We already have legacy implementations for the `f` and `standard_cauchy` calculations, so this change only affects the `Generator` methods.

zuhu2195:
The correctness of calculation can be impacted if there is an argument whose state changes from one call to another within the same function. Like bitgen_state in random_f(). I suggest we fix legacy_f and legacy_standard_cauchy as well.

zuhu2195:
I have added another commit for legacy_f and legacy_standard_cauchy, this keeps both the functions in sync and consistent.

WarrenWeckesser:
> The correctness of calculation can be impacted if there is an argument whose state changes from one call to another within the same function. Like bitgen_state in random_f().

The *result* is certainly affected, but in this case either order of evaluation generates a correct random variate.

>  I suggest we fix legacy_f and legacy_standard_cauchy as well.

Generally we don't touch the legacy code; it is considered "frozen". Please revert your changes to `legacy_f` and `legacy_standard_cauchy`.  (Sorry, I should have mentioned that in my previous comment about the legacy code.)

zuhu2195:
ok, will revert it. My platform is NonStop OS and these functions(legacy_f and legacy_standard_cauchy) also seem to affect the output of np.random.f in my case(giving non standard output). Will go with a local fix for legacy functions.

mattip:
We care that the output of `random_f` and `random_standard_cauchy` represent the underlying random distributions. We make efforts to also make the results reproducible across different platforms and compilers, but that is not guaranteed and non-reproducibility does not make the results incorrect.

numpy/random/src/distributions/distributions.c
```diff
@@ -468,12 +468,15 @@ double random_chisquare(bitgen_t *bitgen_state, double df) {
 }
 
 double random_f(bitgen_t *bitgen_state, double dfnum, double dfden) {
-  return ((random_chisquare(bitgen_state, dfnum) * dfden) /
-          (random_chisquare(bitgen_state, dfden) * dfnum));
+  double subexpr1 = random_chisquare(bitgen_state, dfnum) * dfden;
+  double subexpr2 = random_chisquare(bitgen_state, dfden) * dfnum;
+  return subexpr1/subexpr2;
```
jorenham: Maybe we should give them some breathing room

```suggestion
  return subexpr1 / subexpr2;
```

numpy/random/src/distributions/distributions.c
```diff
@@ -468,12 +468,15 @@ double random_chisquare(bitgen_t *bitgen_state, double df) {
...
 }
 
 double random_standard_cauchy(bitgen_t *bitgen_state) {
-  return random_standard_normal(bitgen_state) / random_standard_normal(bitgen_state);
+  double subexpr1 = random_standard_normal(bitgen_state);
+  double subexpr2 = random_standard_normal(bitgen_state);
+  return subexpr1/subexpr2;
```
jorenham:
```suggestion
  return subexpr1 / subexpr2;
```

WarrenWeckesser (approved):

WarrenWeckesser:
Thanks @zuhu2195.

The change is straightforward, but I'll still give it another day or so before merging (unless another dev hits the green button before then).

WarrenWeckesser:
Merged, thanks @zuhu2195.

PR #29601, opened 2025-08-20
DOC: fix for f2py migrating-to-meson page

jeanwsr:
- format fix for underscore, which does not render correctly
- remove one duplicate variable


melissawm (approved):
lgtm, thanks @jeanwsr !

PR #29605, opened 2025-08-20

bwhitt7:
This PR is related to #29552 and a continuation of #29596, with the goal of making NumPy's testing suite more thread safe. This changes the setup methods for 4 more testing files in numpy/_core, as well as moving some variables that were only used once to be inside the test they're used in instead of declared in a method.

With this PR, you should be able to call the tests from these files with [pytest-run-parallel](https://github.com/numpy/numpy/pull/github.com/Quansight-Labs/pytest-run-parallel) and have all tests successfully run in parallel threads (and also run normally without pytest-run-parallel).


ngoldbaum (approved):
LGTM, it's all only mechanical changes.

ngoldbaum:
Thanks @bwhitt7!

PR #29606, opened 2025-08-21
DOC: clarify numpy.asarray parameter 'order'

nbidler:
See #28247. Made explicit that 'C' and 'F' can copy input to ensure output format, 'A' does not ensure contiguous output. Kept 'K' the same because all the stride analysis/flag-handling behavior should be extraneous to the end user.


melissawm (approved):
Looks much clearer to me, but I'll leave to @seberg to merge. Cheers!

numpy/_core/_add_newdocs.py
```diff
@@ -967,12 +967,12 @@
...
-        'C' row-major (C-style),
-        'F' column-major (Fortran-style) memory representation.
-        'A' (any) means 'F' if `a` is Fortran contiguous, 'C' otherwise
-        'K' (keep) preserve input order
-        Defaults to 'K'.
+        The memory layout of the output.
+        'C' gives a row-major layout (C-style),
+        'F' gives a column-major layout (Fortran-style).
+        'C' and 'F' will copy if needed to ensure the output format.
+        'A' (any) is equivalent to 'F' if input a is non-contiguous or Fortran-contiguous, otherwise, it is equivalent to 'C'. Unlike 'C' or 'F', 'A' does not ensure that the result is contiguous.
```
seberg: Yeah, seems good, can you break the long line though?

Also, can you see which other places have the identical comment, we should change them all (where applicable, which might be some exceptions but not sure).

nbidler: Will do. 
1. formality question: Is it okay to change it on my repo directly (since it's just replacing ' ' with '\n') or should I do the whole process again with "change locally, spin build, spin test, spin lint, push to personal repo, create new pull request?" Or is there a procedure to alter a repo that had a push request already made?
2. Did you want me to hit up anywhere that had 'order' with 'A' and 'K' args, or anywhere that had the whole `Fortran contiguous, ‘C’ otherwise ‘K’ (keep) preserve input order Defaults` string (which has line-breaks but no punctuation)?

nbidler: also, importantly, does that include [the ufunc optional arguments](https://numpy.org/doc/stable/reference/ufuncs.html)? I ask because they are very different and I don't want to change it lightly.

nbidler:
for clarity's sake, I am going through the docs and have so far found [Ravel](https://numpy.org/doc/stable/reference/generated/numpy.ravel.html), [Flatten](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.flatten.html), [asanyarray](https://numpy.org/doc/stable/reference/generated/numpy.asanyarray.html), [asarray_chkfinite](https://numpy.org/doc/stable/reference/generated/numpy.asarray_chkfinite.html), and [the universal functions optional keyword arguments](https://numpy.org/doc/stable/reference/ufuncs.html) and I am currently trying to trace their behavior or trying to figure out how to make a non-contiguous array in python to pass into the functions.

Edit1: `ravel` probably isn't relevant (the output is always a 1D array, ergo, both C/Fortran and contiguous again), `flatten` same thing, `asanyarray` has `asarray`-like results so I'm going to double-check the code to make sure it shares behaviors, `asarray_chkfinite` returns an ndarray so I'm also going to double-check that as well.

Edit2: using [this little script ](https://github.com/nbidler/asarray-similar_tests)for testing, just to document what I was checking.

Edit3: `asanyarray` falls into `_array_fromobject_generic` (if not an error) just like `asarray`, and `asarray_chkfinite` just calls `asarray` outright, so I will change those descriptions.

numpy/_core/_add_newdocs.py
seberg: No, it doesn't matter there, because theynever return views.

nbidler:
Just to end with a comment: Changed description of 'order' in `asanyarray` and `asarray_chkfinite`

nbidler:
Didn't mean to close the first time, first time was out "do i close and re-make with the new fixes?" - or do I wait for a maintainer to act?

nbidler:
@seberg do I wait for you to approve changes or do I close this PR and make a new one with all the requested changes? It's my first time doing a PR and I'm not sure about the permissions here.

nbidler:
I guess I need to make a new PR to get this merged?

nbidler:
made a [new PR](https://github.com/numpy/numpy/pull/29655) for clarity

PR #29611, opened 2025-08-21
CI: more specific mypy_primer ``on:`` paths

jorenham:
mypy_primer will now also run on changes to `.py` files in `numpy/_typing` and `numpy/typing`, and will ignore changes in the `numpy/typing/tests`.


charris:
Thanks Joren.

PR #29612, opened 2025-08-21
CI: replace comment-hider action in mypy_primer workflow

jorenham:
[kanga333/comment-hider](https://github.com/kanga333/comment-hider) does not seem to be maintained and is still using node 16, which is deprecated. 

[int128/hide-comment-action](https://github.com/int128/hide-comment-action) looks somewhat more active, offerst the same relevant functionality, and uses node 20.


charris:
Thanks Joren.

PR #29613, opened 2025-08-22
CI: only test pyodide

andyfaff:
addresses item in #29462, which is the build and upload of wheels for pyodide. https://github.com/numpy/numpy-release/pull/7 enables the build and upload of those wheels.

I left the emscripten workflow in there, so regular tests are done for each PR. If necessary I can remove altogether.

Note that the numpy-release PR is not quite mergeable, it needs one of the following options:

- allow the use of the `pypa/cibuildwheel` action
- the use of Python 3.11/3.14 as the system python
- or an upstream bug in `cibuildwheel` or `actions/setup-python` to be fixed. (https://github.com/pypa/cibuildwheel/issues/2566)

### DIFF
--- a/.github/workflows/emscripten.yml
+++ b/.github/workflows/emscripten.yml
@@ -5,25 +5,6 @@ on:
     branches:
       - main
       - maintenance/**
-  # Note: this workflow gets triggered on the same schedule as the
-  # wheels.yml workflow to upload WASM wheels to Anaconda.org.
-  schedule:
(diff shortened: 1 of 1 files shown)


rgommers (approved):
Thanks @andyfaff!

> I left the emscripten workflow in there, so regular tests are done for each PR. If necessary I can remove altogether.

I think this choice makes sense. It's good to have one job that tests whether we introduce new issues for WASM yes or no.

PR #29616, opened 2025-08-22

bwhitt7:
This PR is related to #29552, with this being the third PR. This PR has the goal of making NumPy's testing suite more thread safe. This introduces changes to the setup methods for 2 more testing files in numpy/_core.

test_overrides.py includes unittest.mock.Mock, a thread-unsafe functionality that is caught by pytest-run-parallel, so it is not completely thread-safe yet, but all of its setup methods have been replaced.

### DIFF
--- a/numpy/_core/tests/test_numeric.py
+++ b/numpy/_core/tests/test_numeric.py
@@ -747,7 +747,7 @@ class TestBoolScalar:
 
 
 class TestBoolArray:
-    def _create_arrays(self):
+    def _create_bool_arrays(self):
         # offset for simd tests
         t = np.array([True] * 41, dtype=bool)[1::]
         f = np.array([False] * 41, dtype=bool)[1::]
@@ -761,7 +761,7 @@ class TestBoolArray:
         return t, f, o, nm, im
 
     def test_all_any(self):
-        t, f, _, nm, im = self._create_arrays()
+        t, f, _, nm, im = self._create_bool_arrays()
         assert_(t.all())
         assert_(t.any())
         assert_(not f.all())
@@ -789,7 +789,7 @@ class TestBoolArray:
             assert_(not np.all(e), msg=f"{i!r}")
 
     def test_logical_not_abs(self):
-        t, f, o, nm, im = self._create_arrays()
+        t, f, o, nm, im = self._create_bool_arrays()
         assert_array_equal(~t, f)
         assert_array_equal(np.abs(~t), f)
         assert_array_equal(np.abs(~f), t)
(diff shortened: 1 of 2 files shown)


numpy/_core/tests/test_overrides.py
```diff
@@ -646,16 +649,17 @@ def test_nep35_functions_as_array_functions(self,):
...
-        my_func = getattr(self.MyArray, function)
+        my_func = getattr(MyArray, function)
 
         if numpy_ref is True:
             ref = np.array(1)
         else:
-            ref = self.MyArray.array()
-
+            ref = MyArray.array()
+        print(ref)
```
djeada: neat: probably should be removed

numpy/_core/tests/test_numeric.py
```diff
@@ -747,27 +747,29 @@ def test_bitwise_xor(self):
 
 
 class TestBoolArray:
-    def setup_method(self):
+    def _create_arrays(self):
```
djeada: since we already use pytest wouldn't be better to turn it to a fixture?

also more descriptive names, like `bool_arrays` are usually preferred

numpy/_core/tests/test_numeric.py
```diff
@@ -787,118 +789,103 @@ def test_all_any(self):
...
+        assert_array_equal(~np.abs(f), t)
+        assert_array_equal(~np.abs(t), f)
+        assert_array_equal(np.abs(~nm), im)
+        np.logical_not(t, out=o)
+        assert_array_equal(o, f)
+        np.abs(t, out=o)
+        assert_array_equal(o, t)
 
     def test_logical_and_or_xor(self):
-        assert_array_equal(self.t | self.t, self.t)
```
djeada, line 803: here as well as in many other tests, there are repeated checks with slightly different data

i think that more standard way would be to use fixtures:

```
@pytest.mark.parametrize("op,a,b,expected", [
    (lambda x, y: x | y, t, t, t),
    (lambda x, y: x | y, f, f, f),
    (lambda x, y: x | y, t, f, t),
    # ...
])
```

numpy/_core/tests/test_overrides.py
bwhitt7: Whoops, missed that!

numpy/_core/tests/test_numeric.py
bwhitt7: I can definitely rename it to something more descriptive.

In the tracking issue #29552, we decided that fixtures may not be the best solution. Part of it is them being slightly harder to read compared to these "setup" methods, another part is that fixtures aren't guaranteed to be thread safe. At most, fixtures are only created once per test, so if you run a test on multiple threads, any mutations to the fixture value will be shared between threads.

numpy/_core/tests/test_numeric.py
ngoldbaum: I think this would be fine to do but is orthogonal to what Britney is working on so isn't necessary to merge this

ngoldbaum (approved):
Looks good, thanks @bwhitt7!

PR #29623, opened 2025-08-25
BUG: Added manual checks for self-nested sequences during array creation

### LINKED ISSUE #29620: BUG: np.array from self-referencing list fills RAM
### Describe the issue:

Creating a numpy array from a self referencing list does not raise an error, but fills the ram endlessly.

```python
import numpy as np

l = []
l.append(l)
l.append(l)
np.array(l)
```

### Error message:

```shell

```

2.3.2
3.13.7 (main, Aug 15 2025, 12:34:02) [GCC 15.2.1 20250813]

### Runtime Environment:

[{'numpy_version': '2.3.2',
  'python': '3.13.7 (main, Aug 15 2025, 12:34:02) [GCC 15.2.1 20250813]',
  'uname': uname_result(system='Linux', node='my-computer', release='6.16.1-arch1-1', version='#1 SMP PREEMPT_DYNAMIC Fri, 15 Aug 2025 16:04:43 +0000', machine='x86_64')},
                      'found': ['SSSE3',
                                'SSE41',
                                'POPCNT',
                                'SSE42',
                                'AVX',
                                'F16C',
                                'FMA3',
                                'AVX2'],
                      'not_found': ['AVX512F',
                                    'AVX512CD',
                                    'AVX512_KNL',
                                    'AVX512_KNM',
                                    'AVX512_SKX',
                                    'AVX512_CLX',
                                    'AVX512_CNL',
                                    'AVX512_ICL',
                                    'AVX512_SPR']}},
 {'architecture': 'Haswell',
  'filepath': '/tmp/python_venv/lib/python3.13/site-packages/numpy.libs/libscipy_openblas64_-8fb3d286.so',
  'internal_api': 'openblas',
  'num_threads': 32,
  'user_api': 'blas',
  'version': '0.3.30'}]

### Context for the issue:

_No response_

devmt04:
Previously, when creating a NumPy array with self-nested sequences, it causes a quick RAM exhaustion, as also described in https://github.com/numpy/numpy/issues/29620, all due to the cyclic recursions.

With these changes, when creating an array with sequences, a check for self-nesting is performed first, which prevents any further cyclic recursions.

Before:

```python
>>> import numpy as np

>>> l = []
>>> l.append(l)
>>> l.append(l)
>>> np.array(l) # cause a hang here
```

Now:

```python
>>> np.array(l)
  File "<python-input-4>", line 1, in <module>
    np.array(l)
    ~~~~~~~~^^^
ValueError: NumPy array cannot be created using self-nested sequences.
```

closes #29620

### DIFF
--- a/numpy/_core/src/multiarray/array_coercion.c
+++ b/numpy/_core/src/multiarray/array_coercion.c
@@ -1197,6 +1197,16 @@ PyArray_DiscoverDTypeAndShape_Recursive(
         copy = -1;
     }
 
+    if(is_sequence){
+        if (PySequence_Check(objects[0])){
+            if(objects[0] == obj){
+                  PyErr_SetString(PyExc_ValueError,
+                            "NumPy array cannot be created using cyclic self-nested sequence.");
+                return -1;
+            }
+        }
+    }
+
     /* Recursive call for each sequence item */
     for (Py_ssize_t i = 0; i < size; i++) {
         max_dims = PyArray_DiscoverDTypeAndShape_Recursive(
--- a/numpy/_core/src/multiarray/array_coercion.c
+++ b/numpy/_core/src/multiarray/array_coercion.c
@@ -1197,9 +1197,10 @@ PyArray_DiscoverDTypeAndShape_Recursive(
         copy = -1;
     }
 
-    if(is_sequence){
-        if (PySequence_Check(objects[0])){
-            if(objects[0] == obj){
+    /* Check for self-nested sequences to prevent RAM exhaustion */
+    if (is_sequence) {
+        if (PySequence_Check(objects[0])) {
+            if (objects[0] == obj) {
                   PyErr_SetString(PyExc_ValueError,
                             "NumPy array cannot be created using self-nested sequences.");
                 return -1;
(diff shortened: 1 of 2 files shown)


jorenham:
I believe that https://github.com/numpy/numpy/pull/29569 also addresses this, but I might be wrong

numpy/_core/src/multiarray/array_coercion.c
```diff
@@ -1197,6 +1197,17 @@ PyArray_DiscoverDTypeAndShape_Recursive(
         copy = -1;
     }
 
+    /* Check for self-nested sequences to prevent RAM exhaustion */
+    if (is_sequence) {
+        if (PySequence_Check(objects[0])) {
+            if (objects[0] == obj) {
```
jorenham, line 1203: this still leaves

```py
l = [[]]
l[0].append(l)
l[0].append(l)
np.array(l)
```

and 62 other cases like these

jorenham: (not that that's necessarily a problem, it's just something to be aware of)

devmt04: Yeah, I got u, this addresses only one specific and most common case of self-nesting. Thanks !

Issue #29624, opened 2025-08-25
BUG: Precision loss with np.power starting with numpy 2.3


dangillet:
### Describe the issue:

Since numpy 2.3, `np.power` provides a different result than `np.float_power`.

I'm aware of issues #26002 and #23523. But using `NPY_DISABLE_CPU_FEATURES` does not seem to change the result.

### Reproduce the code example:

```python
import numpy as np

result = np.power(np.float64(141322), np.float64(-1.0))
# np.float64(7.0760391163442356e-06)

# View the result bit representation
np.binary_repr(result.view(np.uint64), width=64)
# '0011111011011101101011011101011011010010010111101000001000000000'


result = np.float_power(np.float64(141322), np.float64(-1.0))
# np.float64(7.076039116344235e-06)

# View the result bit representation
np.binary_repr(result.view(np.uint64), width=64)
# '0011111011011101101011011101011011010010010111101000000111111111'


# Compare with python arithmetic

result = 141322 ** -1.0
# 7.076039116344235e-06

# View the result bit representation
import struct
''.join('{:0>8b}'.format(c) for c in struct.pack('!d', result))
# '0011111011011101101011011101011011010010010111101000000111111111'
```

`numpy 2.2.6` does not have this behaviour change. All results are like the Python result.

### Python and NumPy Versions:

```
3.11.13 (main, Jun  4 2025, 08:57:30) [GCC 13.3.0]
2.3.2
```

### Runtime Environment:

```
[{'numpy_version': '2.3.2',
  'python': '3.11.13 (main, Jun  4 2025, 08:57:30) [GCC 13.3.0]',
  'uname': uname_result(system='Linux', node='F0251-linux', release='6.8.0-78-generic', version='#78-Ubuntu SMP PREEMPT_DYNAMIC Tue Aug 12 11:34:18 UTC 2025', machine='x86_64')},
 {'simd_extensions': {'baseline': ['SSE', 'SSE2', 'SSE3'],
                      'found': ['SSSE3',
                                'SSE41',
                                'POPCNT',
                                'SSE42',
                                'AVX',
                                'F16C',
                                'FMA3',
                                'AVX2'],
                      'not_found': ['AVX512F',
                                    'AVX512CD',
                                    'AVX512_KNL',
                                    'AVX512_KNM',
                                    'AVX512_SKX',
                                    'AVX512_CLX',
                                    'AVX512_CNL',
                                    'AVX512_ICL',
                                    'AVX512_SPR']}},
[...]

PR #29626, opened 2025-08-25
BUG: Fix precision loss with np.power for negative exponents (gh-29624)

SumitkCodes:
This PR fixes a regression introduced in NumPy 2.3 where `np.power` produced different results than `np.float_power` for negative exponents, specifically `x ** -1.0`.

**Fixes:** #29624

### Problem
Starting with NumPy 2.3, `np.power` was using a fast path optimization for `x ** -1.0` that called direct division (`1.0 / x`) instead of the standard power function (`pow(x, -1.0)`). This caused precision differences:

### Solution
- Removed the problematic fast path optimization for negative exponents
- Now uses standard `pow()` function for consistent precision
- Added comprehensive tests to prevent future regressions

### Changes
- `loops_umath_fp.dispatch.c.src`: Remove fast path for `x ** -1.0`
- `test_umath.py`: Add test for precision consistency
- `test_regression.py`: Add regression test

This ensures reproducible results for scientific computing applications that depend on exact bit-level consistency.

### DIFF
--- a/numpy/_core/src/umath/loops_umath_fp.dispatch.c.src
+++ b/numpy/_core/src/umath/loops_umath_fp.dispatch.c.src
@@ -155,10 +155,7 @@ NPY_NO_EXPORT void NPY_CPU_DISPATCH_CURFX(@TYPE@_@func@)
         int fastop_found = 1;
         BINARY_LOOP_SLIDING {
             const @type@ in1 = *(@type@ *)ip1;
-            if (in2 == -1.0) {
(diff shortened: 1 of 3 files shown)


WarrenWeckesser:
I have some reservations about this change.  In case anyone is about to merge, please hold off for a bit, while I do some investigation.

seberg:
Thanks for  jumping at an issue quickly, but this wasn't an issue in the end, so closing.

PR #29627, opened 2025-08-25
BLD: Fix PPC64LE build failure with -mcpu=power9 flag

SumitkCodes:
Fixes #29622

**Problem:**
When `CFLAGS='-mcpu=power9'` is set on PPC64LE systems, NumPy build fails with:
**Root Cause:**
 VSX2 targets were missing `-mvsx` flag
VSX2 targets incorrectly used VSX3 intrinsics even when compiled for VSX2
**Solution:**
1. Add `-mvsx` flag to VSX2 configuration in `meson_cpu/ppc64/meson.build`
2. Add `-mvsx` flag to VSX2 flags in `numpy/distutils/ccompiler_opt.py`  
3. Guard VSX3 intrinsics with `NPY__CPU_TARGET_VSX3` in `numpy/_core/src/common/half.hpp`

### DIFF
--- a/meson_cpu/ppc64/meson.build
+++ b/meson_cpu/ppc64/meson.build
@@ -13,7 +13,7 @@ if compiler_id == 'clang'
   VSX.update(args: ['-mvsx', '-maltivec'])
 endif
 VSX2 = mod_features.new(
-  'VSX2', 2, implies: VSX, args: {'val': '-mcpu=power8', 'match': '.*vsx'},
+  'VSX2', 2, implies: VSX, args: ['-mcpu=power8', '-mvsx'],
   detect: {'val': 'VSX2', 'match': 'VSX'},
   test_code: files(source_root + '/numpy/distutils/checks/cpu_vsx2.c')[0],
 )
--- a/numpy/_core/src/common/half.hpp
+++ b/numpy/_core/src/common/half.hpp
@@ -30,7 +30,7 @@ class Half final {
     #if defined(NPY_HAVE_FP16)
         __m128 mf = _mm_load_ss(&f);
         bits_ = _mm_extract_epi16(_mm_cvtps_ph(mf, _MM_FROUND_TO_NEAREST_INT), 0);
-    #elif defined(NPY_HAVE_VSX3) && defined(NPY_HAVE_VSX_ASM)
+    #elif defined(NPY_HAVE_VSX3) && defined(NPY_HAVE_VSX_ASM) && defined(NPY__CPU_TARGET_VSX3)
         __vector float vf32 = vec_splats(f);
         __vector unsigned short vf16;
         __asm__ __volatile__ ("xvcvsphp %x0,%x1" : "=wa" (vf16) : "wa" (vf32));
@@ -55,7 +55,7 @@ class Half final {
     #if defined(NPY_HAVE_AVX512FP16)
         __m128d md = _mm_load_sd(&f);
         bits_ = _mm_extract_epi16(_mm_castph_si128(_mm_cvtpd_ph(md)), 0);
-    #elif defined(NPY_HAVE_VSX3) && defined(NPY_HAVE_VSX3_HALF_DOUBLE)
+    #elif defined(NPY_HAVE_VSX3) && defined(NPY_HAVE_VSX3_HALF_DOUBLE) && defined(NPY__CPU_TARGET_VSX3)
         __asm__ __volatile__ ("xscvdphp %x0,%x1" : "=wa" (bits_) : "wa" (f));
     #elif defined(__ARM_FP16_FORMAT_IEEE)
         __fp16 f16 = __fp16(f);
@@ -72,9 +72,9 @@ class Half final {
         float ret;
         _mm_store_ss(&ret, _mm_cvtph_ps(_mm_cvtsi32_si128(bits_)));
         return ret;
-    #elif defined(NPY_HAVE_VSX3) && defined(vec_extract_fp_from_shorth)
+    #elif defined(NPY_HAVE_VSX3) && defined(vec_extract_fp_from_shorth) && defined(NPY__CPU_TARGET_VSX3)
         return vec_extract(vec_extract_fp_from_shorth(vec_splats(bits_)), 0);
-    #elif defined(NPY_HAVE_VSX3) && defined(NPY_HAVE_VSX_ASM)
+    #elif defined(NPY_HAVE_VSX3) && defined(NPY_HAVE_VSX_ASM) && defined(NPY__CPU_TARGET_VSX3)
         __vector float vf32;
         __asm__ __volatile__("xvcvhpsp %x0,%x1"
                              : "=wa"(vf32)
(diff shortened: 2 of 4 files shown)


mattip:
There are some failing tests around compiler options.

SumitkCodes:
@mattip Working on it

SumitkCodes:
@mattip can you check the changes now

mattip:
LGTM. It would be nice if @seiko2plus could chime in.

numpy/_core/src/common/half.hpp
```diff
@@ -30,7 +30,7 @@ class Half final {
     #if defined(NPY_HAVE_FP16)
         __m128 mf = _mm_load_ss(&f);
         bits_ = _mm_extract_epi16(_mm_cvtps_ph(mf, _MM_FROUND_TO_NEAREST_INT), 0);
-    #elif defined(NPY_HAVE_VSX3) && defined(NPY_HAVE_VSX_ASM)
+    #elif defined(NPY_HAVE_VSX3) && defined(NPY_HAVE_VSX_ASM) && defined(NPY__CPU_TARGET_VSX3)
```
rgommers, line 33: As explained in https://numpy.org/devdocs/reference/simd/how-it-works.html#generating-the-main-configuration-header, these `NPY__CPU_TARGET_` defines shouldn't be used. Why is this change necessary?

meson_cpu/ppc64/meson.build
```diff
@@ -13,7 +13,7 @@ if compiler_id == 'clang'
   VSX.update(args: ['-mvsx', '-maltivec'])
 endif
 VSX2 = mod_features.new(
-  'VSX2', 2, implies: VSX, args: {'val': '-mcpu=power8', 'match': '.*vsx'},
+  'VSX2', 2, implies: VSX, args: ['-mcpu=power8', '-mvsx'],
```
rgommers, line 16: `-mvsx` is already defined in the `VSX` feature, and right in this line we have `implies: VSX`. So this looks like it'll start inserting duplicate flags.

Also, can you explain why `'match': '.*vsx'` needs to be dropped?

rgommers (commented):
A few questions. The change does solve a reported issue, but I'm not sure that everything is correct, so some more explanations would be helpful.

numpy/_core/src/common/half.hpp
SumitkCodes: The issue is that VSX2 targets (compiled with -mcpu=power8 -mvsx) were trying to use VSX3 intrinsics, causing compilation errors.
NPY_HAVE_VSX3 tells us the project supports VSX3, but NPY__CPU_TARGET_VSX3 tells us we're currently compiling a VSX3-specific target. We need both to prevent VSX2 targets from using VSX3 intrinsics.
This isn't runtime dispatch it's preventing compilation errors during the build. The macro is only defined when compiling VSX3 targets so it's the right tool for this job.

meson_cpu/ppc64/meson.build
SumitkCodes: I removed the 'match': '.*vsx' thinking the pattern matching wasn't working properly for VSX2 targets
The build error suggests VSX2 targets aren't getting the right flags
but maybe the real issue is elsewhere in the inheritance chain. Should I revert this change?

seiko2plus: Dropping 'match': '.*vsx'. It's only needed here alongside the other targets. 
The `match` key is a regex used to filter implied flags. Normally, the compiler 
should imply this by default, which is why it’s explicitly defined here.

numpy/_core/src/common/half.hpp
seiko2plus: `NPY__CPU_TARGET_VSX3` is a private #definition and should not be used here. 
Filtering out `VSX` from the regex of `VSX3` and `VSX4` should fix the build issue.

`NPY__CPU_TARGET_VSX3` was originally used internally with distutils for 
dispatchable sources, serving as a helper to define implied feature macros. 
Since moving to Meson, we explicitly pass implied feature #definitions instead.

seiko2plus: So on other words, adding `&& defined(NPY__CPU_TARGET_VSX3)` will always disables this branch.

charris:
@seiko2plus  Ping.

erikarn:
hi! any chance we could have this sorted out soon? I'm doing P9 and soon some P10 stuff in Linux + FreeBSD bare metal and pseries guest VMs and i'd like to use numpy and some other tooling on it.

is there anything i/we can do to get this sorted out?

Thanks!

mgorny:
Oh, we've missed an anniversary.

meson_cpu/ppc64/meson.build
eli-schwartz: I cannot say what effect `implies: VSX` has since the cpu features module is still "pending upstream integration, but ultimately it does not in fact seem to have the effect of adding the implied flags. :/

eli-schwartz: In fact, nowhere in the entire logs do I see -mvsx *anywhere*, not even in translation units that get built as "VSX". Instead, they get built with -mcpu=power8. I don't know what this is supposed to imply.

eli-schwartz (commented):
I just hit this in gentoo packaging CI. Given:

```
CPU Optimization Options
  baseline:
    Requested : 
    Enabled   : 
  dispatch:
    Requested : MAX
    Enabled   : VSX VSX2 VSX3 VSX4
```

It fails to build (i.e. `USE="cpudetection" CPU_FLAGS_PPC="" emerge numpy` disables all baseline, but enables cpudispatch).

The generated command line looks like e.g. this (full log at https://paste.gentoo.zip/8foqtRIX):

```
powerpc64le-unknown-linux-gnu-gcc -Inumpy/_core/libargfunc.dispatch.h_VSX2.a.p -Inumpy/_core -I../numpy-2.5.1/numpy/_core -Inumpy/_core/include -I../numpy-2.5.1/numpy/_core/include -I../numpy-2.5.1/numpy/_core/src/common -I../numpy-2.5.1/numpy/_core/src/multiarray -I../numpy-2.5.1/numpy/_core/src/multiarray/stringdtype -I../numpy-2.5.1/numpy/_core/src/npymath -I../numpy-2.5.1/numpy/_core/src/umath -I/usr/include/python3.14 -I/var/tmp/portage/dev-python/numpy-2.5.1/work/numpy-2.5.1-python3_14/meson_cpu -fdiagnostics-color=always -Wall -Winvalid-pch -std=c11 -fno-strict-aliasing -O2 -pipe -fno-strict-aliasing -DNDEBUG -fPIC -DNPY_INTERNAL_BUILD -DHAVE_NPY_CONFIG_H -D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE=1 -D_LARGEFILE64_SOURCE=1 -O3 -DNPY_HAVE_VSX2 -DHWY_COMPILE_ONLY_STATIC -DHWY_DISABLE_ATTR -mcpu=power8 -DNPY_MTARGETS_CURRENT=VSX2 -MD -MQ numpy/_core/libargfunc.dispatch.h_VSX2.a.p/meson-generated_argfunc.dispatch.c.o -MF numpy/_core/libargfunc.dispatch.h_VSX2.a.p/meson-generated_argfunc.dispatch.c.o.d -o numpy/_core/libargfunc.dispatch.h_VSX2.a.p/meson-generated_argfunc.dispatch.c.o -c numpy/_core/libargfunc.dispatch.h_VSX2.a.p/argfunc.dispatch.c
```

No -mvsx in sight.


Previously, I see:

```
Generating multi-targets for "_umath_tests.dispatch.h" 
  Enabled targets: VSX3, VSX2, VSX, baseline
Generating multi-targets for "argfunc.dispatch.h" 
  Enabled targets: VSX2, baseline
```

PR #29628, opened 2025-08-26

bwhitt7:
This PR is related to #29552, with this being the fourth PR. This PR has the goal of making NumPy's testing suite more thread safe. This introduces changes to the setup methods for 2 more testing files, one in numpy/lib and one in numpy/ma. Pytest's classic xunit setup and teardown methods are thread unsafe, typically not running before the tests do, leading to AttributionErrors. This PR changes these setup method to basic "creation" methods that are called within the tests that need them.

For a few tests in test_core, the setup methods also utilizes np.seterr(). This was modified to be an autouse fixture.

For both files, while all of the thread-unsafe setup methods have been replaced, there are still some thread safety issues with a few tests (either due to usage of tmp_path, stdout, or something else), so you can't run them all with parallel threads just yet.

### DIFF
--- a/numpy/lib/tests/test_recfunctions.py
+++ b/numpy/lib/tests/test_recfunctions.py
@@ -1,3 +1,4 @@
+import pytest
 
 import numpy as np
 import numpy.ma as ma
@@ -31,19 +32,14 @@ zip_dtype = np.lib.recfunctions._zip_dtype
 
 class TestRecFunctions:
     # Misc tests
-
-    def setup_method(self):
+    def test_zip_descr(self):
+        # Test zip_descr
         x = np.array([1, 2, ])
         y = np.array([10, 20, 30])
         z = np.array([('A', 1.), ('B', 2.)],
                      dtype=[('A', '|S3'), ('B', float)])
         w = np.array([(1, (2, 3.0)), (4, (5, 6.0))],
                      dtype=[('a', int), ('b', [('ba', float), ('bb', int)])])
-        self.data = (w, x, y, z)
-
-    def test_zip_descr(self):
-        # Test zip_descr
-        (w, x, y, z) = self.data
 
         # Std array
         test = zip_descr((x, x), flatten=True)
@@ -240,6 +236,7 @@ class TestRecFunctions:
         dt = np.dtype((np.record, dt))
         assert_(repack_fields(dt).type is np.record)
 
+    @pytest.mark.thread_unsafe
     def test_structured_to_unstructured(self, tmp_path):
         a = np.zeros(4, dtype=[('a', 'i4'), ('b', 'f4,u2'), ('c', 'f4', 2)])
         out = structured_to_unstructured(a)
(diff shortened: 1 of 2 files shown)


numpy/lib/tests/test_recfunctions.py
```diff
@@ -833,7 +825,7 @@ def test_join(self):
         #                   dtype=[('a', int), ('b', int),
         #                          ('c', int), ('d', int)])
         #assert_equal(test, control)
-
+        a, b = self._create_arrays()
```
ngoldbaum: can you move this back to where it was? Just in case someone ever cares to fix the broken test. IMO better to not break it or delete it.

numpy/ma/tests/test_core.py
```diff
@@ -2590,16 +2590,17 @@ def test_fillvalue_bytes_or_str(self):
...
-        np.seterr(divide='ignore', invalid='ignore')
 
-    def teardown_method(self):
-        np.seterr(**self.err_status)
+    @pytest.fixture(autouse=True)
+    def err_status(self):
+        err = np.geterr()
+        np.seterr(divide='ignore', invalid='ignore')
+        yield err
+        np.seterr(**err)
```
ngoldbaum, line 2603: Why is this getting defined at class scope over and over again? It seems to me there should be a way to avoid doing that.

numpy/ma/tests/test_core.py
```diff
@@ -4416,17 +4433,6 @@ def test_varstd(self):
...
-
-    def setup_method(self):
-        x = np.array([1., 1., 1., -2., pi / 2.0, 4., 5., -10., 10., 1., 2., 3.])
-        y = np.array([5., 0., 3., 2., -1., -4., 0., -10., 10., 1., 0., 3.])
-        m1 = [1, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0]
-        m2 = [0, 0, 1, 0, 0, 1, 1, 0, 0, 0, 0, 1]
-        xm = masked_array(x, mask=m1)
-        ym = masked_array(y, mask=m2)
-        xm.set_fill_value(1e+20)
-        self.info = (xm, ym)
```
ngoldbaum, line 4428: Was all this just completely unused?

ngoldbaum (commented):
Looking good, although I have a few comments and questions

numpy/ma/tests/test_core.py
bwhitt7: I think it's only used for two classes. I could set it so its scoped to the class so that it doesn't need to run for every test, just once for the class, since it's just setting how errors are handled.

numpy/ma/tests/test_core.py
bwhitt7: Yes, wasn't used for any of the tests and no other tests inherit from this class, so I figured it was safe to delete. It was likely copy-and-pasted from another class since this set of arrays is used elsewhere.

ngoldbaum:
Thanks @bwhitt7!

PR #29629, opened 2025-08-26
TYP: replace scalar type ``__init__`` with ``__new__``

jorenham:
ported from numpy/numtype#692

---

Long story short; this is why:

```pycon
>>> np.int8.__init__.__qualname__
'object.__init__'
>>> np.int8.__new__.__qualname__
'int8.__new__'
```


jorenham:
> Diff from [mypy_primer](https://github.com/hauntsaninja/mypy_primer?rgh-link-date=2025-08-27T00%3A22%3A17Z), showing the effect of this PR on type check results on a corpus of open source code:
> 
> ```diff
> xarray (https://github.com/pydata/xarray)
> + xarray/namedarray/dtypes.py: note: In function "maybe_promote":
> + xarray/namedarray/dtypes.py:86: error: Too many arguments for "generic"  [call-arg]
> ```

Relevant code: https://github.com/pydata/xarray/blob/98732e76481bdf5c497a030657cff32db134e27d/xarray/namedarray/dtypes.py#L48-L86

This new primer error is a true positive. They're calling `dtype_out.type(fill_value)` with `dtype_out: np.dtype[np.generic]`. Type-checkers will interpret that as `np.generic(fill_value)`.
It's easy to fix by using `np.dtype` instead of `np.dtype[np.generic]` in the function signature. A generic-typing approach could also work, but would be more involved.

charris:
Thanks Joren. IIUC, it is up to xarray to fix the problem.

jorenham:
> IIUC, it is up to xarray to fix the problem.

That's right

PR #29630, opened 2025-08-27
TYP: fix slightly incorrect ``memoryview`` type argument in ``ScalarType``

jorenham:
ported from numpy/numtype#693

---

In static-typing land, `memoryview` is a generic type with a type parameter that defaults to `int`. For example `x: memoryview` is equivalent to `x: memoryview[int]`, and type-checkers treat them in exactly the same way.
But the `memoryview` **type** in the `numpy.ScalarType` tuple can be used to construct instances that aren't necessarily of `int`. So its type argument must be set to `Any`, because if omitted, it'll default to `int`.


mattip:
Thanks @jorenham

PR #29370, opened 2025-07-13
ENH: Allow subscript access for `np.bool` by adding `__class_getitem__`

riku-sakamoto:
Fixes #29247

This PR adds support for subscript access to `np.bool` by implementing a `__class_getitem__` method.

## Implementation Details

- Implements `booleantype_class_getitem_abc`, following the pattern used for other scalar types.
- Adds `__class_getitem__` to the `np.bool` type.

@ngoldbaum 

I created new PR.
Would you mind reviewing this when you get a chance? Thank you!

PR #29631, opened 2025-08-27
TYP: Make ``datetime64`` a generic type at runtime

jorenham:
In the stubs, `datetime64` is a generic type with an optional type parameter. But due to the lack of `__class_getitem__`, something like `np.datetime64[int]` would fail at runtime, even though it's valid in static typing land.

This changes `datetime64` so that it's can also be used as generic type at runtime by adding a `__class_getitem__` method that returns a `types.GenericAliasType` (yea, the naming is horrible...).

Before:

```pycon
In [1]: np.__version__
Out[1]: '2.3.2'

In [2]: np.datetime64[int]
Cell In[2], line 1
----> 1 np.datetime64[int]

TypeError: type 'numpy.datetime64' is not subscriptable
```

After:

```pycon
In [1]: np.__version__
Out[1]: '2.4.0.dev0+git20250825.47430c7'

In [2]: np.datetime64[int]
Out[2]: numpy.datetime64[int]
```

(I'm getting pretty good at pretending to know C, eh?)

---

related to #29370

### DIFF
--- a/numpy/_core/src/multiarray/scalartypes.c.src
+++ b/numpy/_core/src/multiarray/scalartypes.c.src
@@ -2853,11 +2853,15 @@ static PyMethodDef numbertype_methods[] = {
     {NULL, NULL, 0, NULL}  /* sentinel */
 };
 
-static PyMethodDef booleantype_methods[] = {
-        /* for typing */
-        {"__class_getitem__", Py_GenericAlias, METH_CLASS | METH_O, NULL},
-        {NULL, NULL, 0, NULL} /* sentinel */
+/**begin repeat
+ * #name = boolean,datetime#
+ */
+static PyMethodDef @name@type_methods[] = {
+    /* for typing */
+    {"__class_getitem__", Py_GenericAlias, METH_CLASS | METH_O, NULL},
+    {NULL, NULL, 0, NULL} /* sentinel */
 };
+/**end repeat**/
 
 /**begin repeat
  * #name = cfloat,clongdouble#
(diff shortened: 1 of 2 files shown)


numpy/_core/src/multiarray/scalartypes.c.src
```diff
@@ -4564,6 +4568,7 @@ initialize_numeric_types(void)
 
     /**end repeat**/
 
+    PyDatetimeArrType_Type.tp_methods = datetimetype_methods;
```
mattip, line 4571: Nice.
I don't know much about the `__class_getitem__` protocol implementation, don't the other dtypes need them as well? Why are `datetime64` and `bool` special?

jorenham: In case of `np.bool` it can be useful to be able to distinguish between `np.True_: np.bool[Literal[True]]` and `np.False_: np.bool[Literal[False]]` like that. 
For `np.datetime64` the generic type parameter tells what `.item()` will return. So `np.datetime64[None]` is the type of NaT, `np.datetime64[int]` have units below ms, and likewise there are `np.datetime64[datetime.datetime]` and `np.datetime64[datetime.date]`.
For the other *conrete* scalar types, at least the numerical ones, there are no such usecases at the moment.
So in short: If you write `np.int8[Any]` you'll see a squiggly. If you write `np.bool[Any]` or `np.datetime64[Any]`, you don't.

mattip:
Thanks @jorenham

PR #29638, opened 2025-08-28
DOC: Document scalar.astype() always returns array, copy=False ignored

meet-vasita:
Add note in scalars documentation explaining astype behavior difference
Clarify that scalar.astype() always returns ndarray, never scalar
Document that copy=False parameter has no effect for scalars
Helps resolve user confusion about inconsistent behavior vs arrays

### DIFF
--- a/doc/source/reference/arrays.scalars.rst
+++ b/doc/source/reference/arrays.scalars.rst
@@ -479,15 +479,27 @@ that the same hardware flags are set and used to interpret the results
 as for :ref:`ufunc <ufuncs>`, so that the error state used for ufuncs
 also carries over to the math on array scalars.
 
+**Important Note about astype():**
+
+.. note::
+   The :meth:`~generic.astype` method behaves differently for scalars than for arrays:
+   
+   - **Always returns an ndarray**, never a scalar, regardless of input type
+   - **The copy parameter has no effect** - a copy is always made when converting scalar to array
+   - This is different from :meth:`numpy.ndarray.astype` which can return the same object when ``copy=False``
+
+The exceptions to the above rules are given below:
+
 The exceptions to the above rules are given below:
 
 .. autosummary::
    :toctree: generated/
 
-   generic.__array__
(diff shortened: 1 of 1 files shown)


meet-vasita:
Im Trying to solve this ''BUG: np.astype(copy=False) copies for scalars #29455''

seberg:
Thank you, but I think we should probably consider this a bug, although I could imagine it being common enough that I am not 100% sure about it (and it might even be good to go via a deprecation, even though `copy=False` is very new).

I wonder what others think about this?

meet-vasita:
Thanks for the feedback! I see your point about this being more of a bug than just a documentation issue. I’m happy to adjust the PR depending on what the maintainers decide (e.g., updating docs, working on a deprecation path, or helping with a fix). Please let me know the preferred next step.

ngoldbaum:
I agree it doesn't make much sense to document this bad behavior. We should just fix it.

Also for future documentation-only PRs, consider adding `[skip cirrus] [skip actions] [skip azp]` to your commit message. That will only run the CI on circleci, where we do the doc builds and other doctests.

meet-vasita:
Thanks for the clarification! I understand now that it makes more sense to fix the behavior instead of documenting it.

eendebakpt:
I have a branch that addresses part of the issue https://github.com/numpy/numpy/compare/main...eendebakpt:numpy:astype. It relaxes the documentation a bit and (for the common case) avoids the copy.

@meet-vasita If you are interested, feel free to convert the branch into a PR.

meet-vasita:
Thanks for sharing this! 🙏 I’d be happy to pick up your branch and convert it into a PR. I’ll test it locally and update the docs/tests as needed

seberg:
Brought it up today, not that we had much discussion, but I think (probably we have that on the other PR), I am OK with it returning scalars, since we also have a function after all (`asarray` and even `astype`).

I think I'll close this PR, since I think we should really just fix it.

PR #29640, opened 2025-08-29
ENH: Add extended sorting APIs (first version)

charris:
This PR adds new sorting APIs that can be used for extending our current sort types. The new APIs dispense with the *kind* argument in the legacy interface with three new boolean arguments:

- stable -- whether the sort should stable, default is False,
- descending -- whether to sort in descending order, default is False,
- nanfirst -- whether NaNs sort to the last or first, default is False.

The main difference is that sorts are now chosen by functionality instead of by algorithm.

The functions added are:

- (API)PyArray_SortExt,
- (ndarray method)sortext,
- (function)sortext,
- (API)PyArray_ArgSortExt,
- (ndarray method)argsortext,
- (function)argsortext.

Both the legacy and new methods and functions call through the API functions, so those are the two functions that need changes depending on how we want to implement adding new sort algorithms. The legacy APIs were left for downstream compatibility.

Thought was given to just modifying the legacy methods and functions to optionally use the new arguments, but documenting that made such a complicated mess that I decided it was unusable in practice. It would confuse the heck out of users.

One change that has happened under the covers is that heapsort is no longer publically accessible. It is used by introsort, so the code remains, but in practice its only virtue was that it was both inplace and always ``nlog(n)``. Introsort has the same virtue and is usually faster, so it is used in its place. I doubt anyone will notice.

I note that the partioning/selection functions have not been modified, we may want to do that a some point.


doc/source/reference/c-api/array.rst
```diff
@@ -2319,6 +2319,40 @@ Item selection and manipulation
     structured array, create a new data-type with a different order of names
     and construct a view of the array with that new data-type.
 
+.. c:function:: PyObject* PyArray_SortExt(PyArrayObject* self, int axis, bool stable, bool descending, bool nanfirst)
+
+    .. versionadded:: 2.24.0
```
jorenham, line 2324:
```suggestion
    .. versionadded:: 2.4.0
```

doc/source/reference/c-api/array.rst
```diff
@@ -2319,6 +2319,40 @@ Item selection and manipulation
...
+    the first field is equal will use the second field and so on. To alter the
+    sort order of a structured array, create a new data-type with a different
+    order of names and construct a view of the array with that new data-type.
+
+    Raises a *TypeError* if the required algorithm is not available.
+
+
+.. c:function:: PyObject* PyArray_ArgSortExt(PyArrayObject* self, int axis, int stable, int descending, bool nanfirst)
+
+    .. versionadded:: 2.24.0
```
jorenham, line 2340:
```suggestion
    .. versionadded:: 2.4.0
```

numpy/_core/__init__.pyi
jorenham, line 1: I think you updated the wrong `__init__.pyi` here, since this is the one in `_core`

jorenham (commented):
Wow, you actually managed to do this in a day; impressive.

I like the stubs for the new functions. It might be nice to also add  some for their sibling `ndarray` methods.
Also, unlike the dynamic `__all__` in `numpy/__init__.py`, the one in `__init__.pyi` is a big ol' list of exports. So if the idea is to export the `argsortext` and `sortext`, then they'd have to be added to that `__all__` by hand.

doc/source/reference/c-api/array.rst
charris: Oops.

numpy/_core/__init__.pyi
charris: Where should it go? I was kind of guessing.

charris:
I am open to the modifying the function signatures relative to keyword/keyword only. Keyword only would work fine for the new parameters.

numpy/_core/__init__.pyi
jorenham: one directory up, so the `numpy/__init__.pyi`

jorenham:
> I am open to the modifying the function signatures relative to keyword/keyword only. Keyword only would work fine for the new parameters.

After it's released, it's a lot more difficult to make parameters keyword only than the other way around. So the signature that's the least difficult to later change would look something like

```
def sortext(a, /, *, axis=-1, order=None,
            stable=None, descending=None, nanfirst=None): ...
```

... but I can imagine that some would consider that to be a bit too restrictive, especially the `axis=-1`.

Personally, I don't have a strong opinion &mdash; none of the parameters after `a` affect the output type, so the `/` and `*` won't make it easier to annotate or something.

numpy/_core/__init__.pyi
charris: Hmm, `sort` and `argsort` seem to be in both.

jorenham: but `sortext` and `argsortext` not, right? or aren't they supposed to be public api?

charris: I was thinking both need to be consistent.

seberg:
> I am open to the modifying the function signatures relative to keyword/keyword only. Keyword only would work fine for the new parameters.

This seems necessary to me.  I am -1 on adding a new function/method, you need to extend the existing signature (at which point kwarg only is probably the only useful thing anyway), rather than duplicating it.
I am also not in favor of including the `nanfirst` (or a different name) parameter at all in Python.  It makes sense to C to plan ahead, because adding a `PyArray_SortEx2` is annoying, but in Python it is just confusing.

The Python API thus requires no change at all.  I don't really care myself for adding the C-side functions to the new signature, but I also don't mind it.

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -1599,6 +1600,69 @@ PyArray_Sort(PyArrayObject *op, int axis, NPY_SORTKIND which)
...
+        }
+    }
+
+    if (sort == NULL) {
+        PyErr_SetString(PyExc_TypeError,
+                        "no current sort function meets the requirements");
+        return -1;
+    }
+    return _new_sortlike(op, axis, sort, NULL, NULL, 0);
+}
```
seberg, line 1663: One other question: For the sake of not keeping the old version a stub forever, can we make the old version just call this one?  (Yes, I guess it effectively means deactivating the algorithm choice completely, but I guess that was where we were going anyway.)

seberg:
To me an interesting point that this helps make is the question of whether we should keep using `NPY_SORTKIND` as a meaningful user choice for algorithms?
The other PR initially suggested to just keep using it, possibly allowing to extend it easily in the future.

But, I could see an approach here, where we mostly rely on `strict=True/False`.  Then now (or in the future), we also allow `NPY_SORTKIND` (or something similar), but see it more as a hint that can be ignored (maybe with a warning).  That way these algorithms could stay accessible, but we can focus on just stable/unstable.

charris:
> NPY_SORTKIND as a meaningful user choice for algorithms?

What makes this PR big is adding the method, and I don' t much like using `sortext` instead of `sort`.  If we don't want to just extend `kind`, we could add keyword only arguments. the trick is to make it clear that the user uses one or the other. That was actually my first approach here. If we can decide how to clearly document that there are two mutually exclusive interfaces? The default for kind at the C level is actually -1, so we could reinterpret that as "use keywords". So in addition to
`quicksort`, `heapsort`, `mergesort`, we would have `keywords` and make that the default. We already do that for the `stable` keyword, this would just make it explicit. Note that at the API level, the keywords in the newer functions are combined to make an index for switch statements, essentially a three bit binary number. We could make that the interface so that the call value is essentially `STABLE | DESCENDING | NANFIRST`. That sort of interface is actually rather CPYTHONIC.

seberg:
> If we can decide how to clearly document that there are two mutually exclusive interfaces?

We already have that with `stable=` and `kind=` which are exclusive.  That isn't actually documented much, which I suspect isn't too bad since the errors are clear enough -- and typing stubs also even (`kind=` should have a very brief note that it can't be used with the others).
Another thing is that we should try to de-emphasize `kind=`, because we prefer the other API (that may include deprecating it at least as a positional argument and making it the last keyword only argument).
(The last thing is, that the `Notes` section is rather big on algorithms which map to `kind=`, it would be nice to set that into context of `stable=True/False` a bit)

But what I am actually not 100% clear on: Are you sure to move to a future where we may remove algorithms that are not available via `stable=True/False`?  Because if the answer is "no", then I think we should consider other paths (such as keeping `kind=` around, or keeping it around as a hint that may be ignored except for most builtin types).

> That sort of interface is actually rather CPYTHONIC

Good point. If we add new public user-facing C-API, a request flag API sounds like a great idea.  Request flags actually do allow future flag additions (i.e. no worries about NaNs for now).

charris:
> We already have that with stable= and kind= which are exclusive

Yes, we also have
```
        _NPY_SORT_UNDEFINED=-1,
        NPY_QUICKSORT=0,
        NPY_HEAPSORT=1,
        NPY_MERGESORT=2,
        NPY_STABLESORT=2,
```
I'm thinking we should add `NPY_KEYWORD=-1`, which is how `_NPY_SORT_UNDEFINED` is already used, and make it a default. No one using the default will even notice. We can add `NPY_SORTKEYS`
```
    NPY_STABLE = 1
    NPY_DESCENDING = 2,
    NPY_NANFIRST = 4
```

seberg:
Sure, we can re-name `_NPY_SORT_UNDEFINED`, I might just suggest removing the underscore if we need the `-1` level to be public API.

The question for me is whether long-term, we are OK with removing the more specific `kind=` choices, then?
If not, then `NPY_SORTKIND` remains a valid thing to pass, and the API you are prosing here should probably be `PyArray_SortEx(arr, axis, STABLE|DESCENDING[|NAN_PRECEDENCE], NPY_SORTKIND kind)` (using the flags idea) then passing `NPY_SORT_UNDEFINED=-1` as the last argument would be typical.
(But it would also be OK to pass something else, except that we would raise e.g. for quicksort+stable. And we may want to either ignore the request or warn if we don't have a quicksort for that dtype.)

If we say `kind=` will become unavailable in the (very) long term (including for builtin sorts), then we only need `NPY_SORTKIND` for the old `PyArray_Sort(..., sort_kind)` and there we would just translate it to `stable=True/False`.
In that case, we don't need to change anything at all.  We may need `_NPY_SORT_UNDEFINED` for the internal main sorting function but not as public API.

charris:
> We can add NPY_SORTKEYS

Or maybe `NPY_SORTFLAGS`

> where we may remove algorithms that are not available via stable=True/False

Not sure what you mean here, but we can certainly remove algorithms that aren't used. Note that both radixsort and heapsort are used internally, the first as the stable sort for small integers, the second internally in introsort (quicksort).

charris:
> might just suggest removing the underscore if we need the -1 level to be public API.

I would like a name that is more self documenting, `NPY_KEYWORD` is a defined behavior, it means we will use keywords.

charris:
Note that I am not set on this PR, it serves a basis for discussion, but there is nothing sacred about a day's work. I will happily close it and make another if we are clear on what we want things to look like.

charris:
So I am thinking the way to go is to add new API functions, like
`PyArray_SortEx(self, axis, kind, flags)` . If kind is not `NPY_SORTKEYWORD`, then `flags` is ignored, otherwise `flags` determines the sort. The default flags of `0` will have the same behavior as the current defaults. This makes it easy for the legacy `PyArray_Sort` to call through the new API. If we later add a slot function, it can also be called with `flags`.  Note than any function can be use for the default as long as it is not descending or nanfirst, it may or not be stable.

`SortEx` is better than `SortExt`, which has the slightly confusing `tExt` in it.

Maybe `NPY_USEFLAGS` instead of `NPY_SORTKEYWORD`?

mhvk:
Not sure I've got a really informed opinion yet, but
1. I agree with @seberg that on the python side ideally we just add keywords rather than new functions/methods. So, `ndarray.[arg]sort(kind=None, order=None, *, stable=True, ascending=True[, nanfirst=True])`. 
2. I like the idea of slowly deprecating `kind` by doing, say, `.sort(*args, order=None, stable=True, ..., kind=np._NoValue)` and warning if `*args` is set but use it to set `kind`, `order` as needed, etc.
3. For the C api, there seems to be no choice but to introduce new functions, since the `argsort` version has no ability to control anything. My tendency would be to go with something like `PyArray_SortExt(PyArrayObject* self, int axis, NPY_SORT_FLAGS flags)`, i.e., do not have the `kind` at all but rather work the various kinds into the flags. This could be just `flags | (NPY_SORTKIND+1) << 16` but maybe it would be better to make more functional flags that inform the algorithm used, like  `NPY_SORT_LOWMEM`. An advantage would be that the C code would never error - if for a given choice of `stable` and `ascending/descending`, there is only one algorithm, `LOWMEM` is automatically satisfied. It does mean mixing up requirements and requests, though.

seberg:
Yeah, hints like lowmem was the direction I was thinking. If we are satisfied with hints then all is good to me: we can remove the sort kind completely which is nice.
Hints can always be added quite easily both here and in the DType API middle layer, so we can decide if we care to add them now (to keep all existing algorithms reachable) or not (and remove some alorithm implementations soon at least).

I trust Chuck about the decision of having such hints now to keep algorithm choices available.

(Small note: I think we should deprecate `kind` likely unless we re-use it as a hint. But it would be nice to make sure downstream can switch to `stable=` also onold NumPy versions - I am not sure that is the case yet.)

charris:
I've written another version which does things a bit differently. I'll put it up with only the C files in a bit to make it easier to review, documentation and the np.functions can follow later. It uses both `kind` and `flags` (request flag API), so as to have complete legacy compatibility. And yes, the Python side uses three keywords if `kind` is not used, with `kind` having precedence. All the tests pass, no new methods are added, and the legacy C-API calls through the new one.

charris:
See https://github.com/numpy/numpy/pull/29642.

seberg:
> so as to have complete legacy compatibility

I would like to avoid this and make a decision (at least how it's supposed to look long term).
Either we say that the legacy stuff is unnecessary and can be mapped to just stable/unstable (because nobody needs to pick algorithms in practice).
Or we say that the legacy choice is important enough to keep and we morph "kind" into a "hint" (or keep it).

That way, the long term API will allow `stable=True, kind="low-memory"` and even now we can allow `stable=True, kind="mergesort"`.

I am not happy with an either/or API (either kind or flags), because we'll end up with two separate systems (without a plan on how it might go away) and I find it very weird if `kind="quicksort"` is considered OK, but `kind="quicksort", descending=True` is impossible.

charris:
The legacy compatibility is mostly for the C-API and so we can specify 'heapsort'. Personally, I be happy to dispose of 'heapsort', I don't think is serves any purpose except internally. We can deprecate it easily enough. It is in fact easy to deprecate `kind` if we wish to make the transition to pure keyword arguments. But if we are going to go the deprecation route, it is good to have an the keyword option in place. 

> stable=True, kind="mergesort".

Actually, no. You can do `kind="stable"`, but you cannot use both keywords and kind, it raises an error. I don't know when that went in, but it isn't new. It probably came in when the `stable` keyword was added. The new API keeps that, you can use one or the other interface, but not both.

Long term, I would be happy to get rid of `kind`, but I think we need to keep the `sort` methods to avoid terrible downstream churn. Folks who use the defaults won't need to change anything, and those who use `kind=` can change it to
keywords, we just need to give them instructions, omit `kind` and replace it with the following:

* `kind='quicksort' -> default.
* `kind='heapsort' -> default.
* `kind='mergesort' -> stable=1
* `kind='stable' -> stable=1.

And after a while we can just raise remove 'kind'. The C-API seems little used, but there is really no replacement for downstream users if we remove it and don't make the new functions public. I kept the 'kind' keyword in the new C-API just so the legacy functions could call 'heapsort' through it. I suggest we keep it private for the moment, deprecate 'heapsort', and then we can remove it and make the new C-API functions public if we wish.

seberg:
I know the current state, but we can change that to allow mixing where it makes sense!

We do not need to keep the heapsort *implementation*, though, just the kwarg/C-API for a (long) while. Since using a different sort should give equivalent (enough) results.

So that sounds like a +1 to only two sort algorithm choices (stable/unstable). We won't be tied to that forever (except going back and forth is annoying for users), but if we are happy now it simplifies things.

Should bring it once up but that sounds fine to me. I am just really not sure how much the choice still matters performance wise for specific use-cases (like almost sorted arrays?), if its a lot, someone may be upset...

charris:
> We do not need to keep the heapsort implementation, 

We do, introsort (quicksort) calls it when pure quicksorting gets stuck.  That is why we don't need it in the public api.

timsort probably does better with almost sorted arrays, in fact, timsort (mergesort) is faster on some hardware than quicksort.  My interpretation of `stable` is that stability is guaranteed, not it cannot be used for the default. Radix sort is stable, and much faster than quicksort, and should probably be the default for small integers.

seberg:
Yeah, the question is just if some user will be upset if they can't direct NumPy to the best algorithm (upset in a meaningful way, I don't care about 10% performance).

It sounds like "no". So I think we should effectively remove kind completely and just re-map it to stable/unstable for now for simplicity. I like that this came outof this PR.

mhvk:
Agreed that `kind` can be removed in the new `flags`. We can always use some of the bits as a "request" map for speed or memory optimizations.

seberg:
>  We can always use some of the bits as a "request" map for speed or memory optimizations.

I guess the reason why I am circling a bit here is that I would like to not wobble back and forth.  If a current user will be upset because `kind="..."` stops working and that prompts us to introduce new bits as a "request" map, then I would be happier to introduce them right now (and use them to keep existing functionality available).

That is just to avoid an awkward: If NumPy <2.4 use `kind="quicksort"` if `numpy>2.6` request="low-memory" otherwise "NumPy removed this feature :(".

(If we started from scratch, it would be clear to not worry. But as is, I would like the old API to just call the new API which means losing some capabilities right now.)

mhvk:
I think for now we are only talking about not having `kind` in the new C API. If we get complaints that the new C API doesn't allow one to select X, Y, or Z, it seems to me we can add it at that point. And there wouldn't be too much wrong with having a standard translation between `kind` and `flag` bits on the python side, if it turns out to be necessary (or even pass on `kind` as part of the flags).

In other words, having a new C API that omits `kind` would seem to go in the direction that we want to go -- removing `kind` altogether -- yet leave the option open to course correct without much hassle if that turns out to be a bad idea.

Aside: I think we should move the discussion to #29642!

seberg:
> I think for now we are only talking about not having kind in the new C API

Yeah, I am more interested in how this informs the API for DTypes. And I would like us to dog feetbthat. So if we keep kind around anywhere, I want it there too in some form.

PR #29641, opened 2025-08-29

bwhitt7:
This PR is related to #29552. This PR has the goal of making NumPy's testing suite more thread safe. This introduces changes to the setup methods for 5 more testing files, 4 in numpy/ma and one in numpy/matrixlib. Pytest's classic xunit setup and teardown methods are thread unsafe, typically not running before the tests do, leading to AttributionErrors. This PR changes these setup method to basic "creation" methods that are called within the tests that need them.

For test_old_ma and test_subclassing, while all of the thread-unsafe setup methods have been replaced, there are still some thread safety issues with a few tests (either due to usage of global state or stdout), so you can't run them all with parallel threads just yet.

### DIFF
--- a/numpy/ma/tests/test_extras.py
+++ b/numpy/ma/tests/test_extras.py
@@ -1314,11 +1314,11 @@ class TestMedian:
 
 class TestCov:
 
-    def setup_method(self):
-        self.data = array(np.random.rand(12))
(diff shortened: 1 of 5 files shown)


ngoldbaum (approved):
Thank you!

PR #29642, opened 2025-08-30
ENH: Add extended sorting APIs

charris:
This PR adds new sorting APIs that can be used for extending our current
sort types. The new C-APIs do not have the *which* argument of the
legacy C-APIs, instead they have *flags* request flag API. In this
transition, heapsort can no longer be called directly, introsort
(quicksort) is called instead. The sort/argsort methods get new keyword
arguments used to set the flags when *kind* is not given, so *flags* is
invisible at the Python level. The new keywords are:

- stable -- whether the sort should stable, default is False,
- descending -- whether to sort in descending order, default is False,
- nanfirst -- whether NaNs sort to the last or first, default is False.

The main difference here is that the keywords select by functionality,
whereas *kind* selects by algorithm. Note that descending and nanfirst
are not yet implemented, so an error is raised if either is specified.

The added C-API functions are:

- (API)PyArray_SortEx,
- (API)PyArray_ArgSortEx,

Those are the two functions that need changes depending on how we want
to implement adding new sort algorithms.  They should be pretty easy to
adapt to whatever we choose to do.

The legacy C-API functions, PyArray_Sort and PyArray_ArgSort, have been
modified to call through the new C-API functions. In order to do so
the *kind* function has been encoded in the *flags* like so:

- "quicksort" - default (0)
- "heapsort"  - default (0)
- "mergesort" - stable  (1)

I note that the partitioning/selection functions have not been modified,
we may want to do that a some point.

### DIFF
--- a/doc/release/upcoming_changes/29642.c_api.rst
+++ b/doc/release/upcoming_changes/29642.c_api.rst
@@ -11,4 +11,4 @@ needed. The new names of interest are:
 * NPY_SORT_NANFIRST -- NaNs sort to the beginning
 
 The semantic change is that NPY_HEAPSORT is mapped to NPY_QUICKSORT when used.
-
+Note that NPY_SORT_DESCENDING and NPY_SORT_NANFIRST have yet to be implemented.
--- a/doc/release/upcoming_changes/29642.c_api.rst
+++ b/doc/release/upcoming_changes/29642.c_api.rst
@@ -8,7 +8,6 @@ needed. The new names of interest are:
 * NPY_SORT_DEFAULT -- default sort (same value as NPY_QUICKSORT)
 * NPY_SORT_STABLE  -- the sort must be stable (same value as NPY_MERGESORT)
 * NPY_SORT_DESCENDING -- the sort must be descending
-* NPY_SORT_NANFIRST -- NaNs sort to the beginning
 
 The semantic change is that NPY_HEAPSORT is mapped to NPY_QUICKSORT when used.
-Note that NPY_SORT_DESCENDING and NPY_SORT_NANFIRST have yet to be implemented.
+Note that NPY_SORT_DESCENDING is not yet implemented.
--- a/doc/release/upcoming_changes/29642.c_api.rst
+++ b/doc/release/upcoming_changes/29642.c_api.rst
@@ -1,0 +1,14 @@
+The NPY_SORTKIND enum has been enhanced with new variables
+----------------------------------------------------------
+This is of interest if you are using ``PyArray_Sort`` or ``PyArray_ArgSort``.
+We have changed the semantics of the old names in the NPY_SORTKIND enum and
+added new ones. The changes are backward compatible, and no recompilation is
+needed. The new names of interest are:
+
+* NPY_SORT_DEFAULT -- default sort (same value as NPY_QUICKSORT)
+* NPY_SORT_STABLE  -- the sort must be stable (same value as NPY_MERGESORT)
+* NPY_SORT_DESCENDING -- the sort must be descending
+* NPY_SORT_NANFIRST -- NaNs sort to the beginning
+
+The semantic change is that NPY_HEAPSORT is mapped to NPY_QUICKSORT when used.
+
--- a/doc/release/upcoming_changes/29642.change.rst
+++ b/doc/release/upcoming_changes/29642.change.rst
@@ -1,0 +1,4 @@
+Sorting ``kind='heapsort'`` now maps to ``kind='quicksort'``
+------------------------------------------------------------
+It is unlikely that this change will be noticed, but if you do see a change in
+execution time or unstable argsort order, this is likely the cause.
--- a/doc/release/upcoming_changes/29642.change.rst
+++ b/doc/release/upcoming_changes/29642.change.rst
@@ -1,4 +1,7 @@
 Sorting ``kind='heapsort'`` now maps to ``kind='quicksort'``
 ------------------------------------------------------------
 It is unlikely that this change will be noticed, but if you do see a change in
-execution time or unstable argsort order, this is likely the cause.
+execution time or unstable argsort order, that is likely the cause. Please let
+us know if there is a performance regression. Congratulate us if it is
+improved :)
+
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -2303,29 +2303,36 @@ Item selection and manipulation
 
 .. c:function:: PyObject* PyArray_Sort(PyArrayObject* self, int axis, NPY_SORTKIND kind)
 
-    Equivalent to :meth:`ndarray.sort<numpy.ndarray.sort>` (*self*, *axis*,
-    *kind*).  Return an array with the items of *self* sorted along *axis*. The
-    array is sorted using an algorithm whose properties are specified by
-    *kind*, an integer/enum specifying the reguirements of the sorting
-    algorithm used. If *self* ->descr is a data-type with fields defined, then
-    self->descr->names is used to determine the sort order. A comparison where
-    the first field is equal will use the second field and so on. To alter the
-    sort order of a structured array, create a new data-type with a different
-    order of names and construct a view of the array with that new data-type.
+    Return an array with the items of ``self`` sorted along ``axis``. The array
+    is sorted using an algorithm whose properties are specified by the value of
+    ``kind``, an integer/enum specifying the reguirements of the sorting
+    algorithm used. If ``self* ->descr`` is a data-type with fields defined,
+    then ``self->descr->names`` is used to determine the sort order. A comparison
+    where the first field is equal will use the second field and so on. To
+    alter the sort order of a structured array, create a new data-type with a
+    different order of names and construct a view of the array with that new
+    data-type.
 
+    This is the C level function called by the ndarray method
+    :meth:`ndarray.sort<numpy.ndarray.sort>`, though with a different meaning
+    of ``kind`` -- see ``NPY_SORTKIND`` below.
 
 .. c:function:: PyObject* PyArray_ArgSort(PyArrayObject* self, int axis, NPY_SORTKIND kind)
 
-    Equivalent to :meth:`ndarray.argsort<numpy.ndarray.argsort>` (*self*,
-    *axis*, *kind*).  Return an array of indices such that selection of these
-    indices along the given ``axis`` would return a sorted version of *self*.
-    The array is sorted using an algorithm whose properties are specified by
-    *kind*, an integer/enum specifying the reguirements of the sorting
-    algorithm used. If *self* ->descr is a data-type with fields defined, then
-    self->descr->names is used to determine the sort order. A comparison where
-    the first field is equal will use the second field and so on. To alter the
-    sort order of a structured array, create a new data-type with a different
-    order of names and construct a view of the array with that new data-type.
+    Return an array of indices such that selection of these indices along the
+    given ``axis`` would return a sorted version of ``self``.  The array is
+    sorted using an algorithm whose properties are specified by ``kind``, an
+    integer/enum specifying the reguirements of the sorting algorithm used. If
+    ``self->descr`` is a data-type with fields defined, then
+    ``self->descr->names`` is used to determine the sort order. A comparison
+    where the first field is equal will use the second field and so on. To
+    alter the sort order of a structured array, create a new data-type with a
+    different order of names and construct a view of the array with that new
+    data-type.
+
+    This is the C level function called by the ndarray method
+    :meth:`ndarray.argsort<numpy.ndarray.argsort>`, though with a different
+    meaning of ``kind`` -- see ``NPY_SORTKIND`` below.
 
 .. c:function:: PyObject* PyArray_LexSort(PyObject* sort_keys, int axis)
 
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -4381,15 +4381,6 @@ Enumerated Types
         This functionality is not yet implemented for any of the NumPy types
         and cannot yet be set from the Python interface.
 
-    .. c:enumerator:: NPY_SORT_NANFIRST
-
-        (Requirement) Specifies that the sort must sort NaNs to
-        the beginning, e.g. NaNs compare less than any other value for the
-        type. This is only relevant if the type has NaN equivalents.
-        This functionality is not yet implemented for any of the NumPy types
-        and cannot yet be set from the Python interface. It may change in the
-        future.
-
 .. c:enum:: NPY_SCALARKIND
 
     A special variable type indicating the number of "kinds" of
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -2303,21 +2303,29 @@ Item selection and manipulation
 
 .. c:function:: PyObject* PyArray_Sort(PyArrayObject* self, int axis, NPY_SORTKIND kind)
 
-    Equivalent to :meth:`ndarray.sort<numpy.ndarray.sort>` (*self*, *axis*, *kind*).
-    Return an array with the items of *self* sorted along *axis*. The array
-    is sorted using the algorithm denoted by *kind*, which is an integer/enum pointing
-    to the type of sorting algorithms used.
+    Equivalent to :meth:`ndarray.sort<numpy.ndarray.sort>` (*self*, *axis*,
+    *kind*).  Return an array with the items of *self* sorted along *axis*. The
+    array is sorted using an algorithm whose properties are specified by
+    *kind*, an integer/enum specifying the reguirements of the sorting
+    algorithm used. If *self* ->descr is a data-type with fields defined, then
+    self->descr->names is used to determine the sort order. A comparison where
+    the first field is equal will use the second field and so on. To alter the
+    sort order of a structured array, create a new data-type with a different
+    order of names and construct a view of the array with that new data-type.
 
-.. c:function:: PyObject* PyArray_ArgSort(PyArrayObject* self, int axis)
 
-    Equivalent to :meth:`ndarray.argsort<numpy.ndarray.argsort>` (*self*, *axis*).
-    Return an array of indices such that selection of these indices
-    along the given ``axis`` would return a sorted version of *self*. If *self* ->descr
-    is a data-type with fields defined, then self->descr->names is used
-    to determine the sort order. A comparison where the first field is equal
-    will use the second field and so on. To alter the sort order of a
-    structured array, create a new data-type with a different order of names
-    and construct a view of the array with that new data-type.
+.. c:function:: PyObject* PyArray_ArgSort(PyArrayObject* self, int axis, NPY_SORTKIND kind)
+
+    Equivalent to :meth:`ndarray.argsort<numpy.ndarray.argsort>` (*self*,
+    *axis*, *kind*).  Return an array of indices such that selection of these
+    indices along the given ``axis`` would return a sorted version of *self*.
+    The array is sorted using an algorithm whose properties are specified by
+    *kind*, an integer/enum specifying the reguirements of the sorting
+    algorithm used. If *self* ->descr is a data-type with fields defined, then
+    self->descr->names is used to determine the sort order. A comparison where
+    the first field is equal will use the second field and so on. To alter the
+    sort order of a structured array, create a new data-type with a different
+    order of names and construct a view of the array with that new data-type.
 
 .. c:function:: PyObject* PyArray_LexSort(PyObject* sort_keys, int axis)
 
@@ -4321,7 +4329,11 @@ Enumerated Types
 .. c:enum:: NPY_SORTKIND
 
     A special variable-type which can take on different values to indicate
-    the sorting algorithm being used.
+    the sorting algorithm being used. These algorithm types have not been
+    treated strictly for some time, but rather treated as stable/not stable.
+    In NumPy 2.4 they are replaced by requirements (see below), but done in a
+    backwards compatible way. These values will continue to work, except that
+    that NPY_HEAPSORT will do the same thing as NPY_QUICKSORT.
 
     .. c:enumerator:: NPY_QUICKSORT
 
@@ -4335,11 +4347,41 @@ Enumerated Types
 
     .. c:enumerator:: NPY_NSORTS
 
-       Defined to be the number of sorts. It is fixed at three by the need for
-       backwards compatibility, and consequently :c:data:`NPY_MERGESORT` and
-       :c:data:`NPY_STABLESORT` are aliased to each other and may refer to one
-       of several stable sorting algorithms depending on the data type.
+        Defined to be the number of sorts. It is fixed at three by the need for
+        backwards compatibility, and consequently :c:data:`NPY_MERGESORT` and
+        :c:data:`NPY_STABLESORT` are aliased to each other and may refer to one
+        of several stable sorting algorithms depending on the data type.
 
+    In NumPy 2.4 the algorithm names are replaced by requirements. You can still use
+    the old values, a recompile is not needed, but they are reinterpreted such that
+
+    * NPY_QUICKSORT and NPY_HEAPSORT -> NPY_SORT_DEFAULT
+    * NPY_MERGESORT and NPY_STABLE -> NPY_SORT_STABLE
+
+    .. c:enumerator:: NPY_SORT_DEFAULT
+
+        The default sort for the type. For the NumPy builtin types it may be
+        stable or not, but will be ascending and sort NaN types to the end. It
+        is usually chosen for speed and/or low memory.
+
+    .. c:enumerator:: NPY_SORT_STABLE
+
+        (Requirement) Specifies that the sort must be stable.
+
+    .. c:enumerator:: NPY_SORT_DESCENDING
+
+        (Requirement) Specifies that the sort must be in descending order.
+        This functionality is not yet implemented for any of the NumPy types
+        and cannot yet be set from the Python interface.
+
+    .. c:enumerator:: NPY_SORT_NANFIRST
+
+        (Requirement) Specifies that the sort must sort NaNs to
+        the beginning, e.g. NaNs compare less than any other value for the
+        type. This is only relevant if the type has NaN equivalents.
+        This functionality is not yet implemented for any of the NumPy types
+        and cannot yet be set from the Python interface. It may change in the
+        future.
 
 .. c:enum:: NPY_SCALARKIND
 
--- a/numpy/_core/include/numpy/ndarraytypes.h
+++ b/numpy/_core/include/numpy/ndarraytypes.h
@@ -169,12 +169,11 @@ enum NPY_TYPECHAR {
  *
  * Updated in NumPy 2.4
  *
- * Updated with new names denoting requirements rather than the algorithm. All
- * the previous values are reused in a way that should be downstream
- * compatible, but the actual algorithms used may be different than before. The
- * new approach should be more flexible and easier to update. The idea is that
- * NPY_SORT_STABLE | NPY_SORT_DESCENDING | NPY_SORT_NANFIRST should provide an
- * index.
+ * Updated with new names denoting requirements rather than specifying a
+ * particular algorithm. All the previous values are reused in a way that
+ * should be downstream compatible, but the actual algorithms used may be
+ * different than before. The new approach should be more flexible and easier
+ * to update.
  * 
  * Names with a leading underscore are private, and should only be used
  * internally by NumPy.
--- a/numpy/_core/include/numpy/ndarraytypes.h
+++ b/numpy/_core/include/numpy/ndarraytypes.h
@@ -193,7 +193,6 @@ typedef enum {
         NPY_SORT_DEFAULT = 0,
         NPY_SORT_STABLE = 2,
         NPY_SORT_DESCENDING = 4,
-        NPY_SORT_NANFIRST = 8,
 } NPY_SORTKIND;
 #define NPY_NSORTS (NPY_STABLESORT + 1)
 
--- a/numpy/_core/include/numpy/ndarraytypes.h
+++ b/numpy/_core/include/numpy/ndarraytypes.h
@@ -162,18 +162,39 @@ enum NPY_TYPECHAR {
 };
 
 /*
- * Changing this may break Numpy API compatibility
- * due to changing offsets in PyArray_ArrFuncs, so be
- * careful. Here we have reused the mergesort slot for
- * any kind of stable sort, the actual implementation will
- * depend on the data type.
+ * Changing this may break Numpy API compatibility due to changing offsets in
+ * PyArray_ArrFuncs, so be careful. Here we have reused the mergesort slot for
+ * any kind of stable sort, the actual implementation will depend on the data
+ * type.
+ *
+ * Updated in NumPy 2.4
+ *
+ * Updated with new names denoting requirements rather than the algorithm. All
+ * the previous values are reused in a way that should be downstream
+ * compatible, but the actual algorithms used may be different than before. The
+ * new approach should be more flexible and easier to update. The idea is that
+ * NPY_SORT_STABLE | NPY_SORT_DESCENDING | NPY_SORT_NANFIRST should provide an
+ * index.
+ * 
+ * Names with a leading underscore are private, and should only be used
+ * internally by NumPy.
+ *
+ * NPY_NSORTS remains the same for backwards compatibility, it should not be
+ * changed.
  */
+
 typedef enum {
-        _NPY_SORT_UNDEFINED=-1,
-        NPY_QUICKSORT=0,
-        NPY_HEAPSORT=1,
-        NPY_MERGESORT=2,
-        NPY_STABLESORT=2,
+        _NPY_SORT_UNDEFINED = -1,
+        NPY_QUICKSORT = 0,
+        NPY_HEAPSORT = 1,
+        NPY_MERGESORT = 2,
+        NPY_STABLESORT = 2,
+        // new style names 
+        _NPY_SORT_HEAPSORT = 1,
+        NPY_SORT_DEFAULT = 0,
+        NPY_SORT_STABLE = 2,
+        NPY_SORT_DESCENDING = 4,
+        NPY_SORT_NANFIRST = 8,
 } NPY_SORTKIND;
 #define NPY_NSORTS (NPY_STABLESORT + 1)
 
--- a/numpy/_core/src/multiarray/item_selection.c
+++ b/numpy/_core/src/multiarray/item_selection.c
@@ -1549,56 +1549,6 @@ fail:
 }
 
 
-/*NUMPY_API
- * Sort an array in-place
- */
-NPY_NO_EXPORT int
-PyArray_Sort(PyArrayObject *op, int axis, NPY_SORTKIND which)
-{
-    PyArray_SortFunc *sort = NULL;
-    int n = PyArray_NDIM(op);
-
-    if (check_and_adjust_axis(&axis, n) < 0) {
-        return -1;
-    }
-
-    if (PyArray_FailUnlessWriteable(op, "sort array") < 0) {
-        return -1;
-    }
-
-    if (which < 0 || which >= NPY_NSORTS) {
-        PyErr_SetString(PyExc_ValueError, "not a valid sort kind");
-        return -1;
-    }
-
-    sort = PyDataType_GetArrFuncs(PyArray_DESCR(op))->sort[which];
-
-    if (sort == NULL) {
-        if (PyDataType_GetArrFuncs(PyArray_DESCR(op))->compare) {
-            switch (which) {
-                default:
-                case NPY_QUICKSORT:
-                    sort = npy_quicksort;
-                    break;
-                case NPY_HEAPSORT:
-                    sort = npy_heapsort;
-                    break;
-                case NPY_STABLESORT:
-                    sort = npy_timsort;
-                    break;
-            }
-        }
-        else {
-            PyErr_SetString(PyExc_TypeError,
-                            "type does not have compare function");
-            return -1;
-        }
-    }
-
-    return _new_sortlike(op, axis, sort, NULL, NULL, 0);
-}
-
-
 /*
  * make kth array positive, ravel and sort it
  */
@@ -1708,52 +1658,6 @@ PyArray_Partition(PyArrayObject *op, PyArrayObject * ktharray, int axis,
 
     Py_DECREF(kthrvl);
 
-    return ret;
-}
-
-
-/*NUMPY_API
- * ArgSort an array
- */
-NPY_NO_EXPORT PyObject *
-PyArray_ArgSort(PyArrayObject *op, int axis, NPY_SORTKIND which)
-{
-    PyArrayObject *op2;
-    PyArray_ArgSortFunc *argsort = NULL;
-    PyObject *ret;
-
-    argsort = PyDataType_GetArrFuncs(PyArray_DESCR(op))->argsort[which];
-
-    if (argsort == NULL) {
-        if (PyDataType_GetArrFuncs(PyArray_DESCR(op))->compare) {
-            switch (which) {
-                default:
-                case NPY_QUICKSORT:
-                    argsort = npy_aquicksort;
-                    break;
-                case NPY_HEAPSORT:
-                    argsort = npy_aheapsort;
-                    break;
-                case NPY_STABLESORT:
-                    argsort = npy_atimsort;
-                    break;
-            }
-        }
-        else {
-            PyErr_SetString(PyExc_TypeError,
-                            "type does not have compare function");
-            return NULL;
-        }
-    }
-
-    op2 = (PyArrayObject *)PyArray_CheckAxis(op, &axis, 0);
-    if (op2 == NULL) {
-        return NULL;
-    }
-
-    ret = _new_argsortlike(op2, axis, argsort, NULL, NULL, 0);
-
-    Py_DECREF(op2);
     return ret;
 }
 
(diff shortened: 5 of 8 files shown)


charris:
Whoa, that is one of the ugliest diffs I've ever seen. The actual code just adds new functions and deletes the content of old ones. I suggest looking at the files direct.

charris:
Note that I use some static arrays inside the functions which are only written to when they are initialized.  The initialization may not be thread safe, although they should be in C++11. They can be purely local without much cost. I used tables  so that minimal changes would be needed if we add function slots one way or another. They contain variables that are not available until link time, but then, so do the standard dtypes. I'm not clear if function names can be considered constant at compile time,  If so, I could simply pull them out of the functions and initialize them in the usual way.

I think it can be done, but I not sure.

charris:
OK, `const myfunc foo[] = {func1, func2};` works.

charris:
Cleaned things up a bit. @seberg You wanted to remove the legacy `PyArray_Sort` and `PyArray_ArgSort` from the public API. This is a chance to avoid exposing the new functions in API. I suppose we can always add them later if someone wants them.

mhvk:
I think I'd still suggest that the new C-API does not have sort kind at all, and that the old functions just translate. Alternatively, for now, make the new functions private and have the public `SortEx` not have the kind argument, i.e., both old and new call into the private ones.

charris:
> I think I'd still suggest that the new C-API does not have sort kind at all

Internally, I've been using the fourth flag bit to mark using "kind" as a table index. We could translate the kind values into flag bits:

kind == 0 -> 0  (default)
kind == 1 -> 8 (new bit, heapsort)
kind == 2 -> 1 (stable)

or maybe

kind == 0 -> 8   (default)
kind == 1 -> 9   (new bit, heapsort)
kind == 2 -> 10 (stable)

i.e., if the fourth bit is set, use the lower bits to hold the flag value. I chose the fourth bit to reserve eight values of the keywords, but we could pretty reserve more space. I expect we will have a minimum of 32 bits to work with, flags is an int type. So we can get rid of `kind` in the C-API. Getting rid of it in the public method API will take deprecation. For that matter,  we might could use a different bit to mark argsorts, but maybe that is going too far.

Maybe we should just make the flags variable uint32.

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -3154,3 +3058,164 @@ PyArray_MultiIndexSetItem(PyArrayObject *self, const npy_intp *multi_index,
...
+/* Table of generic sort functions for use in PyArray_SortEx*/
+static PyArray_SortFunc* const generic_sort_table[] = {npy_quicksort,
+                                                       npy_heapsort,
+                                                       npy_timsort};
+
+/*NUMPY_API
+ * Sort an array in-place with extended parameters
+ */
+NPY_NO_EXPORT int
+PyArray_SortEx(PyArrayObject *op, int axis, NPY_SORTKIND which, int flags)
```
mhvk: Now looking a bit more closely, I see how you use `which` to infer flags. My sense would still be not to have `which` in the new C API call sequence at all, but, say, store it in the top 4 bits of `uint32 flags`, so one would have
```
which = ((int32)flags >> 28);
if (which == NPY_USEFLAGS) {
    switch (flags & 0x0fffffff) {
        ...
    /* no case 8 */
} else {
    sort = sort_table[which];
}
```
And then call appropriately elsewhere.

In the header file, we should then reserve the appropriate bits.

Alternatively, it is perhaps easier to just rename the functions you have here `Py_Array_Sort_int(...)` and then define
```
NPY_NO_EXPORT int
PyArray_SortEx(op, axis, flags) {
    return PyArray_SortEX_int(op, axis, NPY_USEFLAGS, flags);
}
``` 

Yet another alternative would be to remove the `which` part of the logic here altogether, and move it to the old functions. But that would seem to lead to quite a lot of code duplication.

charris:
The way I see things at the moment.

Decided: 

- Methods `sort` and `argsort` keep their current APIs for back compatibility (done)
- Remove the `which` argument from the new C-API functions, use only the `flags`. (done)
- Make the `flags` argument explicitly uint32. (todo)

To be Decided:

- Remap `kind` losing `heapsort` , 
- Make public the new C-API,
- Deprecate `kind` ,

I would also argue for reserving the lower 8 bits of of the flags for sort selection, and the remaining bits for hints.
That is just to avoid future complications in implementations. That can be handled through documentation.

mhvk:
Agreed on the "decided". For those to be decided, I'd say "remap, make public, and deprecate" - though the deprecation could be done in a follow-up PR.

charris:
I went ahead and made the modifications. It isn't possible to set the enum type until C23, so to be safe I just left it an enum, which will change size if needed, but the size will be platform specific. I think that is safe.

numpy/_core/include/numpy/ndarraytypes.h
```diff
@@ -169,14 +169,25 @@ enum NPY_TYPECHAR {
  * depend on the data type.
  */
 typedef enum {
-        _NPY_SORT_UNDEFINED=-1,
+        NPY_SORT_USEFLAGS=-1,
```
mhvk: Maybe want to mention in the comment that this enum is intended to be deprecated eventually?

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -3154,3 +3058,166 @@ PyArray_MultiIndexSetItem(PyArrayObject *self, const npy_intp *multi_index,
...
+    if (check_and_adjust_axis(&axis, PyArray_NDIM(op)) < 0) {
+        return -1;
+    }
+
+    if (PyArray_FailUnlessWriteable(op, "sort array") < 0) {
+        return -1;
+    }
+
+    sort_table = PyDataType_GetArrFuncs(PyArray_DESCR(op))->sort;
+    switch (flags) {
```
mhvk, line 3089: Hmm, we should make sure that any iteration is correct.... Here, if people pass in `NPY_SORT_STABLE | NPY_SORT_DESCENDING`, they get `quicksort`, which is neither stable nor descending... For now, just have the default be `NULL`? (so one falls through to the generic function, and from there to an error; may need to adjust error messages so that it is clear what was wrong).

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -3154,3 +3058,166 @@ PyArray_MultiIndexSetItem(PyArrayObject *self, const npy_intp *multi_index,
...
+NPY_NO_EXPORT PyObject *
+PyArray_ArgSortEx(PyArrayObject *op, int axis, NPY_SORTFLAGS flags)
+{
+    PyArrayObject *op2;
+    PyObject *ret;
+    PyArray_ArgSortFunc **argsort_table = NULL;
+    PyArray_ArgSortFunc *argsort = NULL;
+
+    // Look for type specific functions
+    argsort_table = PyDataType_GetArrFuncs(PyArray_DESCR(op))->argsort;
```
mhvk, line 3147: Same as for regular sort.

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -3154,3 +3058,166 @@ PyArray_MultiIndexSetItem(PyArrayObject *self, const npy_intp *multi_index,
...
+ */
+NPY_NO_EXPORT int
+PyArray_Sort(PyArrayObject *op, int axis, NPY_SORTKIND which)
+{
+    NPY_SORTFLAGS flags = 0;
+
+    // remap which to flags, note that NPY_HEAPSORT is effectively
+    // remapped to NPY_QUICKSORT.
+    switch (which) {
+        case NPY_STABLESORT:
```
mhvk: Should also have `NPY_MERGESORT` here, right? For the C API, we cannot be sure that it won't be used.

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -3154,3 +3058,166 @@ PyArray_MultiIndexSetItem(PyArrayObject *self, const npy_intp *multi_index,
...
+ */
+NPY_NO_EXPORT PyObject *
+PyArray_ArgSort(PyArrayObject *op, int axis, NPY_SORTKIND which)
+{
+    NPY_SORTFLAGS flags = 0;
+
+    // remap which to flags, note that NPY_HEAPSORT is effectively
+    // remapped to NPY_QUICKSORT.
+    switch (which) {
+        case NPY_STABLESORT:
```
mhvk: Add `NPY_MERGESORT` also here.

numpy/_core/src/multiarray/methods.c
```diff
@@ -1250,30 +1249,63 @@ array_sort(PyArrayObject *self,
...
+        int keywords_used = (stable != -1) | (descending != -1) | (nanfirst != -1);
+        if (keywords_used) {
+            PyErr_SetString(PyExc_ValueError,
+                    "`kind` and keyword parameters can't be provided at "
+                    "the same time. Use only one of them.");
+            return NULL;
+        }
+        // remap sortkind to flags, NPY_HEAPSORT is effectively mapped to NPY_QUICKSORT.
+        switch (sortkind) {
+            case NPY_STABLESORT:
```
mhvk: I think here `NPY_MERGESORT` indeed cannot happen. Maybe add a comment just in case.

charris:
We can postpone adding documentation and keywords at the Python Level until we actually add the new functionality.

numpy/_core/src/multiarray/methods.c
```diff
@@ -1393,28 +1415,63 @@ array_partition(PyArrayObject *self,
...
             "$stable", &PyArray_OptionalBoolConverter, &stable,
+            "$descending", &PyArray_OptionalBoolConverter, &descending,
+            "$nanfirst", &PyArray_OptionalBoolConverter, &nanfirst,
             NULL, NULL, NULL) < 0) {
         return NULL;
     }
-    if (order == Py_None) {
-        order = NULL;
+
+    if (sortkind == NPY_SORT_USEFLAGS) {
```
mhvk: It may be worth splitting out this sortkind checking into a separate function, since it is duplicated between sort and argsort. Something like
```
int get_sort_flags(NPY_SORTKIND sortkind, int stable, int descending, int nanfirst)
```
where the return argument is flags (or -1 if an error occurred. (Maybe better with `flags` passed in?)

numpy/_core/src/multiarray/methods.c
```diff
@@ -1393,28 +1415,63 @@ array_partition(PyArrayObject *self,
...
+                break;
+            case NPY_HEAPSORT:
+            case NPY_QUICKSORT:
+            default:
+                flags = 0;
+        }
     }
+
+    // Reorder field names if required.
+    order = (order != Py_None)? order: NULL;
```
mhvk, line 1452: The same may hold for the order part, but maybe best for follow-up

mhvk (commented):
Some initial comments, probably too early, since really they are mostly about making sure that the new C API gets fully exercised. (Maybe one way is to start with just the stable flag in this commit, and introduce `descending` and `nanfirst` in following ones, together with their support. But probably all premature...)

mhvk:
> We can postpone adding documentation and keywords at the Python Level until we actually add the new functionality.

indeed, though whatever we expose at the C level should be correct (even if it means erroring on `NPY_SORT_DESCENDING`)

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -3154,3 +3058,166 @@ PyArray_MultiIndexSetItem(PyArrayObject *self, const npy_intp *multi_index,
...
+    if (PyArray_FailUnlessWriteable(op, "sort array") < 0) {
+        return -1;
+    }
+
+    sort_table = PyDataType_GetArrFuncs(PyArray_DESCR(op))->sort;
+    switch (flags) {
+        case NPY_SORT_STABLE:
+            sort = sort_table[NPY_STABLESORT];
+            break;
+        default:
```
mhvk, line 3096: I guess this really should be `case 0:` and the default should just fall through.

charris: I figured that the compiler would take care of it. Case statements are reconfigured as binary search under the covers.  But I can see your point.

numpy/_core/src/multiarray/item_selection.c
charris: mergesort and stablesort are the same thing in the enum.  And mergesort isn't actually mergesort, it is either timsort or radix sort depending on the type.

numpy/_core/src/multiarray/item_selection.c
mhvk: Hmm, I never even thought of the compiler for `case NPY_SORT_STABLE` doing the equivalent of `(flags & NPY_SORT_STABLE) == NPY_SORT_STABLE`... that's quite a bit of magic...

numpy/_core/src/multiarray/methods.c
charris: There are a couple of such duplications, I thought to leave it for a later cleanup if desired. See also the legacy C-API functions. And wait until after the new slot is added, which might take over that whole bit and delete this code.

numpy/_core/src/multiarray/item_selection.c
mhvk: Ah, should have seen that. Ignore this one then!

mhvk: Actually, maybe add to the comment you have already?

numpy/_core/include/numpy/ndarraytypes.h
charris: If we leave a comment here, it should be searchable. Hmm, maybe
```
\\ TODO: remove after sort kind deprecation expires.
```

numpy/_core/src/multiarray/methods.c
mhvk: Fair enough!

numpy/_core/src/multiarray/item_selection.c
charris: Should be Fixed.

numpy/_core/src/multiarray/methods.c
```diff
@@ -1250,30 +1249,64 @@ array_sort(PyArrayObject *self,
...
+                    "`kind` and keyword parameters can't be provided at "
+                    "the same time. Use only one of them.");
+            return NULL;
+        }
+        // remap sortkind to flags, NPY_HEAPSORT is effectively mapped to NPY_QUICKSORT.
+        switch (sortkind) {
+            default:
+            case NPY_HEAPSORT:
+            case NPY_QUICKSORT:
+                flags = 0;
```
mhvk: Needs a `break` after, no? 

I'd also swap the order, so default is part of the end (also above, in `item_selection.c`, and below), but no big deal.

mhvk (commented):
looks good now, spotted just one missing `break`. Mostly now seems a question of how to do follow-up: as is, we introduce C API and (undocumented) python API for options that error. Do we want that? E.g., @seberg suggested to not have `nanfirst` yet. 

On the other hand, things are now in place to add that!

numpy/_core/src/multiarray/item_selection.c
charris: I used NPY_MERGESORT instead, which might make more sense.

numpy/_core/src/multiarray/methods.c
charris: I've just used NPY_MERGESORT instead, which might make it clearer.

numpy/_core/src/multiarray/methods.c
charris: I used all the different possibilities, so the default is not longer needed. The range is already checked by the sortkind converter.

charris:
All the keywords are there if you want to test them:
```
In [1]: np.ones(10).sort(descending=1)
Cell In[1], line 1
----> 1 np.ones(10).sort(descending=1)

TypeError: no current sort function meets the requirements
```
They just won't be documented. I suppose we could print out the flag values to make it clearer.

seberg:
> Do we want that? E.g., @seberg suggested to not have nanfirst yet.

Yes, there is no reason in Python to add a kwarg early, it is just confusing to have it.  On C the story may be a bit different (although, not much because `NPY_TARGET_VERSION` allows us to decide whether or not to expose a new flag!  We could skip adding the flags even there -- although I don't care about that!).

---

I am curious now if we can't avoid the new function and just change the definition of `which` in `PyArray_Sort()`?  (Eventually deprecated if e.g. "heapsort" is passed.  Would leave some bits unused in the new sort flags, but that seems OK.)
Not sure how one would best write it, though.  Maybe a `union` of both enums (and make sure the enums don't collide in flags)?  Or just document it as "legacy flag maps to stable/unstable".

charris:
> I am curious now if we can't avoid the new function

Yes, we could. But you didn't want to extend  SORTKIND :) And to be fair, that would have been a kludge.

I rather like the new functions and the flags interface, I think things worked out well. The only real difference is heapsort not being public, and I've thought it unneeded since introsort went in. From the beginning it was the slowest sort, taking about 2x longer than quicksort, and quicksort has gotten faster. 

The flag interface is nicely extensible in the future,  we have been abusing SORTKIND for many years as new sorting functions have come along: introsort, radix sort,  timsort, SIMD quicksort, etc.  We don't want to make users choose among those options, let the experts do it.

All we have do to get rid of `nanfirst` is comment out a line 
```
//           "$nanfirst", &PyArray_OptionalBoolConverter, &nanfirst,
```

The parser then raises an error:

```
In [1]: np.ones(10).sort(nanfirst=1)
Cell In[1], line 1
----> 1 np.ones(10).sort(nanfirst=1)

TypeError: sort() got an unexpected keyword argument 'nanfirst'
```

charris:
Note that all user dtypes need is a compare function, they just put NULLs in the current sorting slots. We only need the current sorting slots for type specific builtin sorts. To add descending sorts, the quick and dirty solution is to add generic sorting routines and add them to the new tables which are introduced here along with the new interfaces. We only need an extended dtype if we want them to be builtin. We can start with the generic sorts and use them for timing comparison, which will tell us if it is worth going to type specific sorts. 

I need to see how runtime dispatching works, I haven't figured that out yet.

seberg:
> the quick and dirty solution is to add generic sorting routines and add them to the new tables which are introduced here along with the new interfaces. We only need an extended dtype if we want them to be builtin.

Yeah, but I happen to care a lot about this, that if we add `descending` we do it in a way that can be exposed to user DTypes, because we can't focus on the API decisions.
(But this is the other PR and I think most decisions are actually already getting close -- but API is hard and it's necessary to have multiple eyes on it.  The biggest API complexity is maybe not missing the chance of fixing more of the API at the same time.)

I don't see that we can expect generic sorts to have decent performance for `descending`, if that was the case, we wouldn't use Raghuveers or highway sorts (which we should plug in for `descending` as well if possible). 

> But you didn't want to extend SORTKIND

Maybe, I may have changed my opinion. But there were two different things here: One is keeping the actual algorithm choice (we are doing the opposite now, _always_ removing that choice) and the other is absorbing the new sort option flags into a single larger enum.

Now that we decided to narrow the algorithm choice to stable/unstable. I think avoiding `PyArray_SortEx()` is probably nice (I am not 100% yet, because I am not sure if that ends up with some monster of an enum).  Just having `PyArray_SortEx` is not that pretty and a bit confusing as well (which one should I use?).

> I need to see how runtime dispatching works, I haven't figured that out yet.

Isn't that what the other PR tried to solve -- and while it should morph with recent discussion, mostly did solve?

charris:
> that ends up with some monster of an enum)

The current values are orthogonal, so three enum entries gives 8 choices. I don't expect the enum to grow much, what will grow are the switch statements.

EDIT: I expect the other PR, if implemented with either a table or function, to get passed a flag, which is just a number. If implemented with a table the number is an index into the table. In short, I expect the switch statements to go away, essentially moved down to the new interface.

charris:
For instance, `descending && stable`  becomes an option. What would grow exponentially is SORTKIND. I think the `flag` option is preferable and doesn't impact the other PR, except it becomes simpler IMHO.

seberg:
> For instance, descending && stable becomes an option. What would grow exponentially is SORTKIND

I am not sure I follow, I am suggesting to add the _flags_ to the kind, or alternative change your sort flags to:
```
typedef enum {
        NPY_SORT_DEFAULT = 0,
        NPY_SORT_STABLE = 1,
        NPY_SORT_DESCENDING = 4,  // no collision wwith NPY_SORTKIND
        NPY_SORT_NANFIRST = 8,
} NPY_SORTFLAGS;
```
and then make the signature be:
```
typedef union {
    NPY_SORTKIND old_style;
    NPY_SORTKIND new_style;
} NPY_ANY_SORT_KIND;

PyArrayObject *
PyArray_Sort(PyArrayObject *arr, int axis, NPY_ANY_SORT_KIND which);
```
which remains fully ABI (and I think also C++ API) compatible.  The only "issue" is that `NPY_SORTFLAGS` has a hole at value 2, which seems unproblematic IMO.

(But we could also keep using a single enum, and just try to deprecate the other names very slowly -- starting with just a compiler warning.)

mhvk:
It may be worth thinking through @seberg's suggestion of just keeping the old C API functions, but (somewhat) reinterpreting `sortkind`. There seem to be two options:
1. Merging `sortkind` and `flags` for C API purposes. I don't think it would be particularly hard to implement: basically, reserve the lowest 4 (or even just 2)  bits for `sortkind` and add the flags starting at the higher bits, i.e., `sortkind = kludge & 0x7; flags = kludge >> 4` and then have the code as is (one would have `_NPY_SORT_UNDEFINED = 0x07`).
2. Reinterpret `sortkind` as flags: bit 2 (mergesort, stable sort) becomes a requirement for stable sorting (`NPY_SORT_STABLE = 2`) and bit 1 a request for low memory (`NPY_SORT_LOWMEM = 1`, which would currently be ignored if `NPY_SORT_STABLE` is set too). In this case `NPY_SORT_DESCENDING = 4` and `NPY_SORT_NANFIRST = 8`.

p.s. On the sort-function provision side, rather than a table indexed by something constructed from flags or so, might it not be easiest to pass on flags to the user function? I think with `flags` rather than `sortkind` that becomes a reasonable option. E.g., if we also define a `NPY_SORT_REQUIREMENTS_MASK` that includes only flags that are requirements, not requests, then the function could do `flags & NPY_SORT_REQUIREMENTS & ~my_sort_abilities` to see whether it can handle them. If its return values could be `fail, success, not-implemented`, one could even allow a fall-back to the generic sort via comparison.

charris:
> Merging sortkind and flags for C API purposes.

Sure, we could do that, just set the flag and call the new functions :) That is essentially the same as we do now, SortEx becomes a helper function, which we either keep separate or copy into the existing functions. We would also need to  change the SORTKIND parser. We could in fact make it return a flag, but that might be too sneaky :)

OTOH, I think making that change obfuscates the code. We will be doing the same thing, just in a more complicated way by mixing two functionalities together.

mhvk:
@charris - but what about my second option? It is not really merging the two, just reinterpreting the existing sortkind as `NPY_SORT_LOWMEM` and `NPY_SORT_STABLE`. It actually does not feel all that hacky, apart from mixing required and requested flags in a slightly non-obvious way.

charris:
The second option would be SORTKIND:

`NPY_SORT_DEFAULT = NPY_QUICKSORT = NPY_HEAPSORT = 0` 
`NPY_SORT_STABLE = NPY_MERGESORT = 1`
`NPY_SORT_DESCENDING = 2`

As long as everyone uses the enum names instead of actual numbers (which is the whole point of using enums), that would work. Then SortEx becomes Sort.

charris:
The problem is mixing `kind` with the keyword arguments. bit that is handled in the `sort` method.

mhvk:
> The problem is mixing `kind` with the keyword arguments. bit that is handled in the `sort` method.

But that would only be a problem on the python side - where the solution you have now works fine.

charris:
@mhvk The main problem with changing the numbering is that it isn't backward compatible, users would need to recompile. So the best way to do things would be to reserve the first three values for current numbering. I suppose we could even split it between positive and negative numbers.

mhvk:
I'm confused: I thought the numbering *is* backwards compatible in the way I suggested. Specifically, if we have the following flags,
```
        NPY_SORT_DEFAULT = 0,
        NPY_SORT_STABLE = 2,
        NPY_SORT_DESCENDING = 4,
        NPY_SORT_NANFIRST = 8,
        NPY_SORT_LOWMEM = 1,
```
then one would have a good match with the kinds:
```
        NPY_QUICKSORT = 0,  -> no flags, fine
        NPY_HEAPSORT = 1, -> would equal NPY_SORT_LOWMEM, which is reasonable (but would have no effect)
        NPY_MERGESORT = 2, -> NPY_SORT_STABLE, obvious
        NPY_STABLESORT = 2, -> same
```
Of course,  `NPY_SORT_LOWMEM | NPY_SORT_STABLE = 3` does not currently exist, so that can default to ignoring `LOWMEM` (which is a request, not a requirement; in my suggested follow-up scheme, one would have `NPY_SORT_REQUIREMENT_MASK = 0xFFFFFFFD` -- keeping bits reserved as requirements until we decide otherwise).

p.s. Of course, for now it may be easier to just have a flag `_NPY_SORT_RESERVED = 1` -- the low memory one just came up because heapsort does not need a work-space (nor does `quicksort`, of course, but this allows us to replace that with something that does need some memory if needed).

charris:
That makes sense. I assume we are still omitting heapsort, right? At some point we will need to decide what "low memory" means. Radix sort uses some extra memory, but not that much, at least for small integers (I need to check how it is implemented for larger integers).

mhvk:
Yes, I'd still omit `heapsort` - just using that what makes it stand out is low memory, even if `quicksort` does just as well.

I think `LOWMEM` would mean a request to optimize that, i.e., given the choice between multiple algorithms, pick the one that uses least memory. Presumably, there could be a `SPEED` too, which would pick the fastest algorithm even if it doubled memory use. Though, really, whether we end up needing any of these flags is not clear! My real point was that we could treat the current `sortkind` as the combination of two flags that happen to make some kind of sense.

charris:
> easier to just have a flag _NPY_SORT_RESERVED = 1 -

I think it best not to regard the heapsort slot as reserved unless we deprecate it first, I even have some qualms about using it for "low memory", as it might be best to put that in the "requests" area for consistency. That said, I can't think of a case where the default doesn't have the lowest memory requirements, but I suppose if we are already remapping heapsort to quicksort, it won't matter if we remap it to something else.

One advantage of reusing SORTKIND is that we don't need to worry about rewriting the other sorting like functions that use the current interface. 

So I'm OK with this proposal.

Another question we need to address at some point is the [array-api standard](https://data-apis.org/array-api/latest/API_specification/sorting_functions.html); I assume that the `stable` keyword was added to support the standard, but I am not sure if we also need to support the default. It looks like we do need to support `descending` as well. @rgommers Can you explain the implications here?

mhvk:
> as it might be best to put that in the "requests" area for consistency

My own sense is that given that we use an `enum`, we do not really have to worry about reserving particular blocks of bits for particular purposes.  (Plus, possibly the only request we ever have is low memory...)

Anyway, it sounds like the outlines of a plan!

charris:
@mhvk I've pushed the new proposal. Nothing is documented yet.

charris:
Note that I've commented out using any keywords except 'stable', we can uncomment them when they are needed.

mhvk (commented):
I feel this worked out nicely, but have not a very good sense of how it fits into the larger picture of dtypes with sort functions. If @seberg thinks this will work, then what remains for this PR would be to adjust the C API documentation.

charris:
> but have not a very good sense of how it fits into the larger picture of dtypes with sort functions

It isn't *needed* for that PR, but things are easier when just one thing at a time needs changes. The information in the proposed context structure is now in the PyArray* functions, so they can do much of the work of selecting the functions.
The functions selected should all be specific to a given functionality and require no more arguments.

The new function/table pointer in the proposed new dtype slot should now take an index. The generic functions themselves only need a comparison function and so only need two variants: ascending/descending, the same is true for the builtin functions. The problem with the old slots is that they only had room for one variant: stable, unstable. Fixing that is now easier because there are only four variants (ignoring nan order),  selected functions with the desired functionality and that function choice is made elsewhere. Elsewhere is where SIMD dispatching takes place, for instance. It is just another sorting function among many.

mhvk:
OK, thanks. I think it indeed helps a great deal to do this separately from a discussion on how to implement the dtype slots, because I think what we ended up with here seems sensible -- while I'm far from sure that a table of functions is the right implementation for the new slots! (For one thing, you need to pass on flags already for the "requested" flags, so might as well pass all; for another, even the generic functions may need to do something for `nanfirst` - or the comparisons have to take that into account. Anyway, that should be discussed elsewhere!)

charris:
Curiously, NPY_SORTKIND isn't documented anywhere except except for a short listing  in the PyArray_Sort documentation.  Documenting NPY_SORTKIND is all the documentation needed for this PR. The release note might contain a bit more on the dropping of heapsort case somebody notices a change and wants to know why.

seberg:
> I feel this worked out nicely, but have not a very good sense of how it fits into the larger picture of dtypes with sort functions.

We are very flexible there.  For me the important part of this PR is dropping heapsort and defining the flags, which is nice.
There are decisions to be made about how to do the generic sort. I think Chuck had some good ideas mentioned here, but the discussion really belongs on the other PR.
(What makes the other PR unwieldy unfortunately is changing the sort signature, but I think that is worthwhile.)

I was worried about planning for Nan precedence/first option, because it is a request that cannot be ignored. But as things are, I toned that down.  I think I don't care anymore, because I am not sure we'll ever do it, and the more I think about it, the more I think there reasonable paths to add it without provisioning now.
(e.g. rename `NPY_DT_sort_slots` to `NPY_DT_sort_slots_with_nan` and that rename alone changes things so that we know that NaNs are supported by that DType.)

numpy/_core/include/numpy/ndarraytypes.h
```diff
@@ -162,18 +162,39 @@ enum NPY_TYPECHAR {
...
+ * PyArray_ArrFuncs, so be careful. Here we have reused the mergesort slot for
+ * any kind of stable sort, the actual implementation will depend on the data
+ * type.
+ *
+ * Updated in NumPy 2.4
+ *
+ * Updated with new names denoting requirements rather than the algorithm. All
+ * the previous values are reused in a way that should be downstream
+ * compatible, but the actual algorithms used may be different than before. The
+ * new approach should be more flexible and easier to update. The idea is that
```
mhvk: No big deal but would delete the sentence "The idea is ..." - the implementation should not matter here (I may well end up thinking that this is the right approach, and now see how indexing is sort-of baked in to the present API, but not having seeing any implementation, I still feel the logical solution would be a single slot for a function with a flags argument...).

doc/source/reference/c-api/array.rst
```diff
@@ -2303,21 +2303,29 @@ Item selection and manipulation
...
-
-    Equivalent to :meth:`ndarray.argsort<numpy.ndarray.argsort>` (*self*, *axis*).
-    Return an array of indices such that selection of these indices
-    along the given ``axis`` would return a sorted version of *self*. If *self* ->descr
-    is a data-type with fields defined, then self->descr->names is used
-    to determine the sort order. A comparison where the first field is equal
-    will use the second field and so on. To alter the sort order of a
-    structured array, create a new data-type with a different order of names
-    and construct a view of the array with that new data-type.
+    Equivalent to :meth:`ndarray.sort<numpy.ndarray.sort>` (*self*, *axis*,
```
mhvk: This was already not quite true any more, and will soon be even less so; I'd tend to start with what the function actually does (so, "Return an array ... " and then at the very end say something like
```
This function is the implementation for :meth:`ndarray.sort<numpy.ndarray.sort>`,
though with slightly different meaning of ``kind`` -- see ``NPY_SORTKIND`` below.
```

(and similar for argsort)

numpy/_core/src/multiarray/methods.c
```diff
@@ -1250,30 +1249,56 @@ array_sort(PyArrayObject *self,
...
+    int descending = -1;
+    int nanfirst = -1;
     NPY_PREPARE_ARGPARSER;
 
     if (npy_parse_arguments("sort", args, len_args, kwnames,
             "|axis", &PyArray_PythonPyIntFromInt, &axis,
             "|kind", &PyArray_SortkindConverter, &sortkind,
             "|order", NULL, &order,
             "$stable", &PyArray_OptionalBoolConverter, &stable,
+//            "$descending", &PyArray_OptionalBoolConverter, &descending,
```
mhvk, line 1265: I'd tend to comment out also the definitions/uses, but obviously no big deal.

doc/release/upcoming_changes/29642.c_api.rst
```diff
@@ -0,0 +1,14 @@
...
+We have changed the semantics of the old names in the NPY_SORTKIND enum and
+added new ones. The changes are backward compatible, and no recompilation is
+needed. The new names of interest are:
+
+* NPY_SORT_DEFAULT -- default sort (same value as NPY_QUICKSORT)
+* NPY_SORT_STABLE  -- the sort must be stable (same value as NPY_MERGESORT)
+* NPY_SORT_DESCENDING -- the sort must be descending
+* NPY_SORT_NANFIRST -- NaNs sort to the beginning
+
+The semantic change is that NPY_HEAPSORT is mapped to NPY_QUICKSORT when used.
```
mhvk, line 12: Should this mention that as of yet descending and nanfirst are not actually supported by `PyArray_[Arg]Sort`?  Can always delete the sentence in the follow-up PRs...

doc/release/upcoming_changes/29642.change.rst
```diff
@@ -0,0 +1,4 @@
+Sorting ``kind='heapsort'`` now maps to ``kind='quicksort'``
+------------------------------------------------------------
+It is unlikely that this change will be noticed, but if you do see a change in
+execution time or unstable argsort order, this is likely the cause.
```
mhvk: Maybe add "Please let us know if this causes a performance regression."?

mhvk (commented):
Last set of (very) nitpicky comments -- I think it is over to @seberg with this!

numpy/_core/src/multiarray/methods.c
charris: We want this. When we get around to implementing it we just need to uncomment the corresponding line here. The rest of code using `descending` and `nanfirst` has been left intact.

numpy/_core/src/multiarray/methods.c
```diff
@@ -1393,28 +1408,56 @@ array_partition(PyArrayObject *self,
...
+    PyObject *res;
+    PyObject *order = NULL;
+    PyArray_Descr *saved = NULL;
+    PyArray_Descr *newd;
     NPY_SORTKIND sortkind = _NPY_SORT_UNDEFINED;
-    PyObject *order = NULL, *res;
-    PyArray_Descr *newd, *saved=NULL;
     int stable = -1;
+    int descending = -1;
+    int nanfirst = -1;
```
seberg: This doesn't need the optional bool converter, since the default is well defined in both cases.  The optional one is just for when `None`/not passed has a special meaning.
(Which is the case for `stable`, because it needs to distinguish not passed for the error.)

I would also be happy to just delete the NaN thing here, but either way. I have been drifting away from wanting it much.  And I think if we have it then `nan_placement="first"` might be clearer than a boolean thing in Python.

(Anyway, it doesn't really matter of course as it is all commented out anyway.)

doc/release/upcoming_changes/29642.c_api.rst
seberg: Yeah, we could even just omit it for now, but I don't care much.

What I think we definitely need here, is make sure that the C-side raises an error when the new flags are used (or, comment them out also in C for now).

charris:
>  is make sure that the C-side raises an error when the new flags are used

You get a  TypeError: "no current sort function meets the requirements"

doc/release/upcoming_changes/29642.c_api.rst
mhvk: The code as written will raise an error if `DESCENDING` or `NANFIRST` is passed in, since there is no sort defined for those.

seberg: Ah, right, I didn't realize that the old code already had the necessary check here.

numpy/_core/src/multiarray/methods.c
charris: We can go further, with heapsort gone, the `kind` argument has the same functionality as the `stable` keyword, we can deprecate it! Internally, we use it to set the `stable` default when it is given. Then all the keywords can be treated the same way. No reason not to allow both arguments any more.

seberg: Yes, but I would say to hold off with deprecating it, unfortunately.
I think `stable=` was added in NumPy 2.0, and it would be a pain if projects like sklearn can't transition to `stable=` for *all* NumPy versions they support (i.e. avoid the need for a numpy version specific code).

(Unless `stable=` was actually added earlier.)

charris: If we allow both  'kind` and keywords, `kind` would precedence in the case (kind="stable", stable=False), the only difference with current behavior would be no error message, and we could still do that but it would be different message.

seberg: I am happy to allow mixing both, at least if we raise an error when they mismatch.  But I am not following what you are suggesting.
Users can't pass `stable=True` yet, on all versions, so I don't want to ask them to replace `kind=...` yet (even if we allow `kind="stable"` it forces double-change: `kind="mergesort"` -> `kind="stable"` -> `stable=True`.

charris: kind="stable" is an alias for kind="mergesort", no double-change needed, except maybe in documentation. Something like:  "Use `stable=True` in place of  `kind='mergesort'` or `kind='stable'` ".

We can leave that for a discussion in a different PR, or perhaps an issue. I think the only thing to be decided is the removal of `NPY_NANFIRST` from the enum. If we are unlikely to use it, it is easier to add it when needed than to remove it once it is in. But we should leave space for more requirements if we add new 'hint' keywords in the future.

charris: I'd rather a follow on PR if we want to enable both `kind` and keywords. I think the important thing for this PR is to have the C interface defined.

seberg: Right. FWIW, the nan parameter matters more for the DType API layer, where new hints can be added nicely (if the user expects it), but requirements cannot.

We could add the bit-mask that Marten suggested.  I am starting to think: let's forget about it.  I think in practice we likely only need to define that a `descending` sort *also* should return the NaNs last.
And if we have to do slightly uglier things in the case that we need more I am OK with that.

For hints, we could prepare for it, but we can also do something like:
```
#define NPY_SORT_LOWMEM (PyArray_RuntimeVersion > 2_4 ? FLAG : 0)
```
a bit ugly, but the user won't notice.

charris:
What do you want done before this goes in?

numpy/_core/src/multiarray/methods.c
charris: > I think stable= was added in NumPy 2.0

Yep, #25437.

mhvk:
I'm OK with it as is, but think given the discussion it might be best not to advertise `NANFIRST` in any way -- easiest would be just to remove it for now. I do think it is fine to leave `descending` in, even if it is not yet supported, since the array API requires that eventually we implement it. But that the way that one is added also already sets an obvious template for adding further options later.

But again, I don't feel strongly, so maybe @seberg can pipe in too.

p.s. Overall, a bit of a tortuous path, but I think a very nice result!

numpy/_core/src/multiarray/methods.c
charris: For hints, we could add something like "NPY_SORT_FIRST_HINT =  16," that would put the requirements in the first four bits of NPY_SORTKIND, with the first bit unused, e.g,, three possible requirements. You get an index by
```
index  = (sortkind & 0xF) >> 1. 
```
Since it often leads to future troubles to start with too few bits, "NPY_SORT_FIRST_HINT =  1 << 8," might be a better choice.  Then the first eight bits would be reserved.

charris:
> only need to define that a descending sort also should return the NaNs last.

If there is no nanfirst keyword, we should say that all alternatives sort NaNs to the end.
If there is a nanfirst keyword, we should say that all alternatives sort NaNs to the end unless `nanfirst=1`.

Not sure where that should go, maybe in the extending dtypes documentation.

charris:
I have removed all references to nanfirst.
I left the `descending` converter alone, as it seems a converter is wanted and the current converter looks exactly right
```
PyArray_BoolConverter(PyObject *object, npy_bool *val)
{
    int bool_val = PyObject_IsTrue(object);
    if (bool_val == -1) {
        return NPY_FAIL;
    }
    *val = (npy_bool)bool_val;
    return NPY_SUCCEED;
}
```

mhvk (approved):
Looks all good to me now!

Maybe left is a comment by @serberg (https://github.com/numpy/numpy/pull/29642#discussion_r2318219969):
> This doesn't need the optional bool converter, since the default is well defined in both cases. The optional one is just for when None/not passed has a special meaning.

(I think `OptionalBoolConverter` is for the case that one can pass in `None` as well.)

charris:
> I think OptionalBoolConverter is for the case that one can pass in None as well.)

Why do you need `None` in parsing a keyword only argument `$stable`? The parser isn't called if the argument isn't present. If we use builtin formats and parse it as 'i':
```
    case 'i': {/* signed int */
        int *p = va_arg(*p_va, int *);
        long ival;
        if (float_argument_error(arg))
            RETURN_ERR_OCCURRED;
        ival = PyLong_AsLong(arg);
        if (ival == -1 && PyErr_Occurred())
            RETURN_ERR_OCCURRED;
        else if (ival > INT_MAX) {
            PyErr_SetString(PyExc_OverflowError,
                            "signed integer is greater than maximum");
            RETURN_ERR_OCCURRED;
        }
        else if (ival < INT_MIN) {
            PyErr_SetString(PyExc_OverflowError,
                            "signed integer is less than minimum");
            RETURN_ERR_OCCURRED;
        }
        else
            *p = ival;
        break;
    }

```
which is not simpler than calling the `__bool__` dunder, which doesn't need to do all those other checks. We could make a new converter that checked for an integer type first, (`PyLong_Check`), if you want to avoid accepting all types that define `__bool__`. Can't do better than that, Python doesn't have a real `bool` type.

mhvk:
I think just use `PyArray_BoolConverter`

charris:
> I think just use PyArray_BoolConverter

Ah, it's already written, is it. Which raises the opposite question, why isn't it used  for `stable`? Lets try that.

charris:
Well, that answers that question, the original was in fact correct, `PyArray_OptionalBoolConverter` *does* need to be used for both 'stable' and 'descending'.

@mhvk it was suggested by @seberg that I not use  `PyArray_OptionalBoolConverter` for the new options,
by which I assume he meant `PyArray_BoolConverter` instead, which I had forgotten existed.

seberg:
> the original was in fact correct, PyArray_OptionalBoolConverter does need to be used for both 'stable' and 'descending'.

I am not following.  You don't really _need_ it for stable but it is has some sense because of the `-1` default when it's not passed.
What it gives you is that `kind="stable", stable=None` is allowed, i.e. `stable=None` behaves identical to it not being passed.

But for `descending=` you don't need it, it's signature is not `stable=None/np._NoValue`, but `descending=False`.  It has a well defined default and `descending=None` has the same meaning as `desecnding=False` and that will always be true.
*There will never be later code checking for `descending == -1` in the C-code here.*

charris:
> But for descending= you don't need it

I tested that and it didn't work. It compiles, and runs without errors, but not correctly. Which is to say, sortkind is not correctly constructed. The function signature is defined in the parser call, and both `"|stable"` and `"|descending"` are keyword only, so it makes sense that they need the same function. Note that using the alternate function to parse stable gives hundreds of test errors.

I think the vectorcall logic goes this way.  All the possible variables must be processed. If it is determined that there is no corresponding value in the args list, the question is "what to do"? This is determined on our end, and I think we have decided to return  `None` in that case, so the Optional parsing function is needed to  ignore that case and do nothing.
The straight boolean parser will only work correctly  for strictly positional arguments, where the wanted value is always in the args list or an error is raised if the args list is too short.

Note that the two functions also have different return types, the straight converter returns to `npy_bool` (unsigned char), while the optional converter returns to `int`. That could cause problems on big endian architectures if the return variables don't have the correct type. 

@ngoldbaum I think you implemented the fastcall processing, can you confirm my understanding?

seberg:
> That could cause problems on big endian architectures if the return variables don't have the correct type.

I implemented fastcall, but it has nothing to do with things here, it behaves exactly the same as the builtin parser.
Maybe you forgot to change the variable to `npy_bool descending = NPY_FALSE;`?

> This is determined on our end, and I think we have decided to return None in that case

When the argument is optional, you always have to set the default.  I would still avoid `OptionalBoolConverter` because it converts `None` to `False` rather than "keep default unchanged".

charris:
Summary: As long as we allow `kind` or keywords, but not both, we need to use the optional parser so that we can tell when a keyword is used.

The problem with not using the non-optional converter is that we assume that it doesn't affect the current value of the variable (which is what online documentation suggests), which is incorrect. It sets the variable to either 0 ot 1. Setting the variable -1 before the call serves no purpose. So the non-optional version can only be used when 0 is the default value. The code can be fixed to work with the non-optional parser, but will not work as long as we want to enforce use `kind`, or use keywords, but not both.

I don't think we need to keep the one or the other choice now that heapsort is gone, but as long as we do, we need to use the optional parser. If we allow both `kind` and keywords, then we can do;
```
int stable = -1;  -> npy_bool stable = 0; 
```
and  give  the keyword priority over `kind` in the sortkind construction.

seberg:
> Summary: As long as we allow kind or keywords, but not both, we need to use the optional parser so that we can tell when a keyword is used.

I suppose I think there is no reason to reject `sort(kind="mergesort", descending=False)`.  But fine, if you want `kind="mergesort", descending=None` to work but `kind="mergesort", descending=False` to fail, then yes we need the Optional version.
(If I went back in time, I might actually start arguing the same direction for `stable=` also -- to not use the Optional one.)

EDIt: anyway, I don't care much, just do the version you prefer.

charris:
Summary: I'll stick with the current implementation.

Using a split is an easy rule to remember, as opposed to having different behaviors for different keywords. I'll go for that.

charris:
Thanks all. And especial thanks to  @mhvk  for his close reading and suggestions.

mhvk:
Looking forward to the next steps!

PR #29643, opened 2025-08-30
MAINT: Improve astype performance and update documentation

meet-vasita:
This pull request improves the behavior of astype by avoiding unnecessary copies when the requirements are already satisfied. I also updated the documentation to make it clear that in some cases the input array may be returned instead of always creating a new one. The main goal is to make astype more efficient and easier to understand.

Thanks to @eendebakpt for the initial work on this branch. 
I’ve picked it up to convert into a PR and will address any feedback, update docs, and add tests as needed.

### DIFF
--- a/numpy/_core/src/multiarray/scalartypes.c.src
+++ b/numpy/_core/src/multiarray/scalartypes.c.src
@@ -2131,7 +2131,7 @@ gentype_byteswap(PyObject *self, PyObject *args, PyObject *kwds)
  *
  * #name = take, getfield, put, repeat, tofile, mean, trace, diagonal, clip,
  *         std, var, sum, cumsum, prod, cumprod, compress, sort, argsort,
- *         round, argmax, argmin, max, min, any, all, astype, resize,
+ *         round, argmax, argmin, max, min, any, all, resize,
  *         reshape, choose, tostring, tobytes, copy, searchsorted, view,
  *         flatten, ravel, squeeze#
  */
@@ -2142,6 +2142,22 @@ gentype_@name@(PyObject *self, PyObject *args, PyObject *kwds)
 }
 /**end repeat**/
 
+static PyObject *
+gentype_astype(PyObject *self, PyObject *args, PyObject *kwds)
+{
+    if ((kwds == NULL) && (PyTuple_Size(args)==1)) {
+        // no keyword arguments, single argument is the dtype
+        PyArray_Descr *dtype = PyArray_DescrFromScalar(self);
+        assert(dtype != NULL);
+        int same_dtype = ( (PyObject *)dtype == PyTuple_GET_ITEM(args, 0));
+        Py_DECREF(dtype);
+        if (same_dtype) {
+            // dtypes are equal, scalars are immutable so no copy is needed
+            return Py_NewRef(self);
+        }
+    }
+    return gentype_generic_method(self, args, kwds, "astype");
+}
 
 /**begin repeat
  * #name = integer, floating#
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -3202,7 +3202,7 @@ add_newdoc('numpy._core.multiarray', 'ndarray', ('astype',
     copy : bool, optional
         By default, astype always returns a newly allocated array. If this
         is set to false, and the `dtype`, `order`, and `subok`
-        requirements are satisfied, the input array is returned instead
+        requirements are satisfied, the input array may be returned instead
         of a copy.
 
     Returns


numpy/_core/src/multiarray/scalartypes.c.src
```diff
@@ -2142,6 +2142,22 @@ gentype_@name@(PyObject *self, PyObject *args, PyObject *kwds)
 }
 /**end repeat**/
 
+static PyObject *
+gentype_astype(PyObject *self, PyObject *args, PyObject *kwds)
+{
+    if ((kwds == NULL) && (PyTuple_Size(args)==1)) {
```
devmt04, line 2148: Why `kwds==NULL`? Aren't keywords arguments are already set and always passed?

meet-vasita: the kwds argument can actually be NULL when no keyword arguments are given (for example, when someone calls just astype(x)). It only becomes a dictionary when keywords are used, like astype(dtype=x).

That’s why I added the kwds == NULL check to safely handle the case where no keywords are passed.

devmt04: > calls just astype(x))

but some keywords arguments are set by default. See,
`ndarray.astype(dtype, order='K', casting='unsafe', subok=True, copy=True)`

Is this still makes kwds `NULL` ?

mattip:
> This pull request improves the behavior of astype by avoiding unnecessary copies when the requirements are already satisfied

Was this discussed as a desired change in behaviour? I think some users may be expecting `c = np.astype(b, b.dtype)` to return a copy.

meet-vasita:
Thanks for the feedback! My goal with this change was mainly to improve performance by skipping unnecessary copies when the dtype is already the same. I can see your point though, some users might expect astype(b, b.dtype) to always create a fresh copy, even if nothing changes.

I’m happy to adjust depending on what the team feels is the right behavior. Do you think this optimization should move forward (with clear documentation), or should it be discussed more broadly before changing the current behavior?

eendebakpt:
> > This pull request improves the behavior of astype by avoiding unnecessary copies when the requirements are already satisfied
> 
> Was this discussed as a desired change in behaviour? I think some users may be expecting `c = np.astype(b, b.dtype)` to return a copy.

I suggested this behavior on slack and no objections have been raised (but there was not much discussion either). The behaviour would not be ok for numpy arrays, but for scalars it is perhaps ok since the scalars are immutable.

meet-vasita:
That makes sense, especially since scalars are immutable. I’m glad to hear this was already suggested and discussed on Slack.

seberg:
> The behaviour would not be ok for numpy arrays, but for scalars it is perhaps ok since the scalars are immutable

Yes, the immutability should make it OK, I agree.  Note that void scalars are (unfortunately) mutable.
The scalar behavior is a bit funky, in that it returns scalars which means it can't preserve all dtype requests (because scalars don't have byte-order or metadata), but that doesn't change here.

`==` is a bit of a sad check, but of course it is a valid fast-path and `PyArray_EquivTypes` wouldn't be correct.

I guess it all comes down to: I am OK with this, but structured dtypes need a guard at least.

---

One could actually speed this up for all cases by avoiding intermediate arrays completely, although I am not sure how unwieldy/long it'll be.

eendebakpt:
> Yes, the immutability should make it OK, I agree. Note that void scalars are (unfortunately) mutable. The scalar behavior is a bit 

Here is a minimal example:
```
import numpy as np
print(np.__version__)

# try a structured dtype
x = np.array([('Rex'), ('Fido')], dtype=[('name', 'U10') ])
v = x[1]
np.isscalar(v) # True
print(v) # ('Fido',)
v.abc = 10 # AttributeError: 'numpy.void' object has no attribute 'abc'

v[0] = 1 # works! so v is mutable!
print(v) # ('1',)

isinstance(v, np.void) # True
```

@seberg Is (the C equivalent of) `isinstance(v, np.void) ` the right guard?

eendebakpt:
@meet-vasita In https://github.com/numpy/numpy/pull/29656 I implemented the special casing for the void scalars. (not sure yet it covers all the cases, but something along these lines you can use here)

meet-vasita:
Thanks @eendebakpt, that’s very helpful! 🙌 I’ll take a look at #29656 and incorporate a similar approach here.

numpy/_core/src/multiarray/scalartypes.c.src
```diff
@@ -2142,6 +2142,22 @@ gentype_@name@(PyObject *self, PyObject *args, PyObject *kwds)
...
+        assert(dtype != NULL);
+        int same_dtype = ( (PyObject *)dtype == PyTuple_GET_ITEM(args, 0));
+        Py_DECREF(dtype);
+        if (same_dtype) {
+            // dtypes are equal, scalars are immutable so no copy is needed
+            return Py_NewRef(self);
+        }
+    }
+    return gentype_generic_method(self, args, kwds, "astype");
+}
```
tylerjereddy, line 2160: Our current `asv` benchmarks seem performance neutral on this: `asv continuous -E virtualenv -e -b "time_astype" main fix-astype`

Even if I hack something to convert from the same type the result seems neutral:

<details>

```diff
--- a/benchmarks/benchmarks/bench_ufunc.py
+++ b/benchmarks/benchmarks/bench_ufunc.py
@@ -309,14 +309,16 @@ class NDArrayAsType(Benchmark):
     timeout = 10
 
     def setup(self, typeconv):
-        if typeconv[0] == typeconv[1]:
-            raise NotImplementedError(
-                    "Skipping test for converting to the same dtype")
         self.xarg = get_squares_().get(typeconv[0])
+        self.start = self.xarg.astype(typeconv[0])
 
     def time_astype(self, typeconv):
         self.xarg.astype(typeconv[1])
 
+    def time_astype_same(self, typeconv):
+        # same type
+        self.start.astype(typeconv[0])

```
 
Is there a simple incantation I can use to verify the measured improvement in performance?
</details>

tylerjereddy: I tried "scalar" too (below the fold), still performance neutral, maybe I misunderstand.

<details>

```diff
--- a/benchmarks/benchmarks/bench_ufunc.py
+++ b/benchmarks/benchmarks/bench_ufunc.py
@@ -309,14 +309,16 @@ class NDArrayAsType(Benchmark):
     timeout = 10
 
     def setup(self, typeconv):
-        if typeconv[0] == typeconv[1]:
-            raise NotImplementedError(
-                    "Skipping test for converting to the same dtype")
         self.xarg = get_squares_().get(typeconv[0])
+        self.start = np.array(5).astype(typeconv[0])
 
     def time_astype(self, typeconv):
         self.xarg.astype(typeconv[1])
 
+    def time_astype_same(self, typeconv):
+        # same type
+        self.start.astype(typeconv[0])
+
```

</details>

tylerjereddy: ah, I guess that's not really scalar either

tylerjereddy: Even hacking to scalar I didn't see a change >= 10 % in `asv` locally:

<details>

```diff
--- a/benchmarks/benchmarks/bench_ufunc.py
+++ b/benchmarks/benchmarks/bench_ufunc.py
@@ -309,14 +309,16 @@ class NDArrayAsType(Benchmark):
     timeout = 10
 
     def setup(self, typeconv):
-        if typeconv[0] == typeconv[1]:
-            raise NotImplementedError(
-                    "Skipping test for converting to the same dtype")
         self.xarg = get_squares_().get(typeconv[0])
+        self.start = np.float64(5.)
 
     def time_astype(self, typeconv):
         self.xarg.astype(typeconv[1])
 
+    def time_astype_same(self, typeconv):
+        # same type
+        self.start.astype(np.float64)
+

```

</details>

eendebakpt: I tried this locally:
```
from .common import Benchmark
import numpy as np


class ScalarAstype(Benchmark):
    def setup(self):
        self.start = np.float64(5.)
        self.dtype = np.float64(15.).dtype
        self.dtype_other = np.float32(15.).dtype

    def time_astype(self):
        self.start.astype(self.dtype)

    def time_astype_other(self):
        self.start.astype(self.dtype_other)

    def time_astype_v2(self):
        self.start.astype(np.float64)
```
With the PR the performance of `time_astype` improves by a factor 10, but `time_astype_v2` stays roughly the same.

Maybe we can improve so that  `np.float64` is also picked by the fast path, but not sure it is worthwhile.

seberg: > Maybe we can improve so that  np.float64 is also picked by the fast path, but not sure it is worthwhile.

You could just call `PyArray_DescrConverter` (or what it was) in this function already.  Since calling it a second time later should be basically instant probably.
Maybe that actually makes sense, since writing `astype("float64")` is pretty common, passing in a dtype is actually probably rare!

eendebakpt:
@meet-vasita Do you want to continue with the work in this PR? If not, I will create a new PR with the suggestion from Sebastian.

meet-vasita:
Hi @eendebakpt, you can go ahead create a new suggested PR

eendebakpt:
Work has continued in https://github.com/numpy/numpy/pull/32379, closing here.

PR #29648, opened 2025-09-01
TST: delete global env_setup fixture

ngoldbaum:
Towards https://github.com/numpy/numpy/issues/29552. Ping @bwhitt7.

We'd like to use `pytest-run-parallel` more but the `monkeypatch` fixture gets detected by the plugin as a thread-unsafe fixture.

### DIFF
--- a/numpy/conftest.py
+++ b/numpy/conftest.py
@@ -147,14 +147,6 @@ def check_fpu_mode(request):
 def add_np(doctest_namespace):
     doctest_namespace['np'] = numpy
 
-@pytest.fixture(autouse=True, scope="session")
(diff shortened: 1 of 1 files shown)


seberg:
> Is this still needed anymore?

Probably not, I think we should just try removing it.  This was probably added at a time where dict iteration order was undefined, which is a time long gone by.

EDIT: Or maybe doctests/very specific tests, but I think then we can deal with it there as we notice issues.

seberg:
Well, since all tests pass, let's put it in and see if someone actually notices :). Thanks Nathan!

PR #29649, opened 2025-09-01
BUG: avoid thread-unsafe refcount check in temp elision

ngoldbaum:
Checking `Py_REFCNT == 1` is thread-unsafe on the free-threaded build, since refcounts are stored and loaded in free-threading using atomic operations.

Python 3.14 added [`PyUnstable_Object_IsUniquelyReferenced`](https://docs.python.org/3.14/c-api/object.html#c.PyUnstable_Object_IsUniquelyReferenced) to replace static refcount checks. I added a backport to pythoncapi-compat in https://github.com/python/pythoncapi-compat/pull/149, which allows me to use the function unconditionally and get correct behavior on 3.13t as well. This relies on some private implementation details on 3.13t, but all that is done in the compat header.

### DIFF
--- a/numpy/_core/src/multiarray/temp_elide.c
+++ b/numpy/_core/src/multiarray/temp_elide.c
@@ -5,6 +5,7 @@
 #include <Python.h>
 
 #include "npy_config.h"
+#include "npy_pycompat.h"
 #include "numpy/arrayobject.h"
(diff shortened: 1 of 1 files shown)


charris:
Thanks Nathan.

PR #29655, opened 2025-09-02
DOC: clarify numpy.asarray, numpy.asanyarray, numpy.asarray_chkfinite parameter 'order' 

nbidler:
See issue [28247](https://github.com/numpy/numpy/issues/28247). Made explicit that 'C' and 'F' can copy input to ensure output format, 'A' does not ensure contiguous output. Kept 'K' the same because all the stride analysis/flag-handling behavior should be extraneous to the end user. See PR [29606](https://github.com/numpy/numpy/pull/29606). Member [melissawm](https://github.com/melissawm) approved initially, then Member @seberg requested "can you see which other places have the identical comment, we should change them all (where applicable" so additional changes were made to the 'order' parameter of `asanyarray` and `asarray_chkfinite ` because their descriptions were identical to `asarray` and behavior was very similar to `asarray`. This is a new PR that includes initial changes and the requested additions.


mattip:
Thanks @nbidler

PR #29656, opened 2025-09-02
ENH: Improve performance of numpy scalar __copy__ and __deepcopy__

eendebakpt:
The numpy scalars are immutable (expect for subclasses of np.void), so copy or deepcopy can return the same object. This improves performance, since the current implementation converts to array, copies and converts back to scalar.

### DIFF
--- a/numpy/_core/tests/test_multiarray.py
+++ b/numpy/_core/tests/test_multiarray.py
@@ -2470,7 +2470,7 @@ class TestMethods:
         with pytest.raises(AssertionError):
             assert_array_equal(a, b)
 
-    def test__deepcopy___void_scalar(self, dtype):
+    def test__deepcopy___void_scalar(self):
         # see comments in gh-29643
         value = np.void('Rex', dtype=[('name', 'U10')])
         value_deepcopy = value.__deepcopy__(None)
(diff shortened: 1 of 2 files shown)


meet-vasita:
Thanks @eendebakpt! I see #29656 is running through CI now.

seberg (approved):
Thanks, somehow missed this, but I think this is good.  Are you up for adding a test that we are actually returning the identity as expected for the other dtypes?  But also looks good without.

numpy/_core/tests/test_multiarray.py
```diff
@@ -2470,12 +2470,19 @@ def test__deepcopy__(self, dtype):
...
+        # see comments in gh-29643
+        value = np.void('Rex', dtype=[('name', 'U10')])
+        value_deepcopy = value.__deepcopy__(None)
+        value[0] = None
+        assert value_deepcopy[0] == 'Rex'
+
+    @pytest.mark.parametrize("dtype", [np.int64, np.float32, np.float64])
+    def test__deepcopy__scalar(self):
+        # test optimization from gh-29656
+        value = dtype(1.1)
```
```suggestion
    @pytest.mark.parametrize("sctype", [np.int64, np.float32, np.float64])
    def test__deepcopy__scalar(self, sctype):
        # test optimization from gh-29656
        value = sctype(1.1)
```

numpy/_core/tests/test_multiarray.py
```diff
@@ -2470,6 +2470,20 @@ def test__deepcopy__(self, dtype):
         with pytest.raises(AssertionError):
             assert_array_equal(a, b)
 
+    def test__deepcopy___void_scalar(self, dtype):
```
```suggestion
    def test__deepcopy___void_scalar(self):
```

PR #28622, opened 2025-04-01
ENH, SIMD: Initial implementation of Highway wrapper

seiko2plus:
A thin wrapper over Google's Highway SIMD library to simplify its interface. This commit provides the implementation of that wrapper, consisting of:

- simd.hpp: Main header defining the SIMD namespaces and configuration
- simd.inc.hpp: Template header included multiple times with different namespaces

The wrapper eliminates Highway's class tags by:
- Using lane types directly which can be deduced from arguments
- Leveraging namespaces (np::simd and np::simd128) for different register widths

A README is included to guide usage and document design decisions.

PR #29658, opened 2025-09-03
MAINT: Optimize the logical implementation for RISC-V based on Highway Wrapper

ixgbe:
Test command: spin bench --compare HEAD~1 -t bench_ufunc.LogicalSpecific

Add class LogicalSpecific in benchmarks/benchmarks/bench_ufunc.py for this test on Banana Pi BPI-F3

```
class LogicalSpecific(Benchmark):
    timeout = 10
    params = ['bool']
    param_names = ['dtype']

    def setup(self, dtype):
        np.seterr(all='ignore')
        self.logical_and_fn = np.logical_and
        self.logical_or_fn = np.logical_or
        self.logical_not_fn = np.logical_not
        self.absolute_fn = np.absolute

        from .common import get_square_
        arr1 = get_square_(dtype)
        arr2 = get_square_(dtype)

        self.data_and = (arr1, arr2)
        self.data_or  = (arr1, arr2)
        self.data_not = (arr1,)
        self.data_abs = (arr1,)

    def time_logical_and(self, dtype):
        self.logical_and_fn(*self.data_and)

    def time_logical_or(self, dtype):
        self.logical_or_fn(*self.data_or)

    def time_logical_not(self, dtype):
        self.logical_not_fn(*self.data_not)

    def time_absolute(self, dtype):
        self.absolute_fn(*self.data_abs)

```
(numpy_venv) root@yangwang:/home/yangwang/numpy# git log -2
commit e5446650f2f29ac2c21e459835d9b04baa1b131d (HEAD -> main)
Author: Wang Yang <yangwang@iscas.ac.cn>
Date:   Fri Aug 29 15:39:43 2025 +0800

ENH, SIMD: Optimize the logical implementation based on Highway Wrapper

commit b112b4e4cc12ae8a0968b26dcaa53ba032fced29
Author: Wang Yang <yangwang@iscas.ac.cn>
Date:   Fri Aug 29 15:38:43 2025 +0800

TST:Add logical test

(numpy_venv) root@yangwang:/home/yangwang/numpy#
(numpy_venv) root@yangwang:/home/yangwang/numpy# spin bench --compare HEAD~1 -t bench_ufunc.LogicalSpecific
| Change   | Before [b112b4e4] <main~1>   | After [e5446650] <main>   |   Ratio | Benchmark (Parameter)                                |
|----------|------------------------------|---------------------------|---------|------------------------------------------------------|
| -        | 92.2±0.5μs                   | 43.4±8μs                  |    0.47 | bench_ufunc.LogicalSpecific.time_logical_not('bool') |
| -        | 90.3±0.9μs                   | 41.9±8μs                  |    0.46 | bench_ufunc.LogicalSpecific.time_absolute('bool')    |
| -        | 128±0.3μs                    | 56.8±10μs                 |    0.44 | bench_ufunc.LogicalSpecific.time_logical_or('bool')  |
| -        | 127±1μs                      | 55.1±10μs                 |    0.43 | bench_ufunc.LogicalSpecific.time_logical_and('bool') |

### DIFF
--- a/numpy/_core/src/umath/loops_logical.dispatch.cpp
+++ b/numpy/_core/src/umath/loops_logical.dispatch.cpp
@@ -4,17 +4,16 @@
 #include "lowlevel_strided_loops.h"
 #include "fast_loop_macros.h"
 #include <functional>
-
+#include "simd/simd.hpp"
 #include <hwy/highway.h>
-namespace hn = hwy::HWY_NAMESPACE;
 
 struct logical_and_t {};
 struct logical_or_t {};
 struct absolute_t {};
 struct logical_not_t {};
 
-const hn::ScalableTag<uint8_t> u8;
-using vec_u8 = hn::Vec<decltype(u8)>;
+namespace {
+using namespace np::simd;
 
 /*******************************************************************************
  ** Defining the SIMD kernels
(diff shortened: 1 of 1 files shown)


mattip:
For someone not fluent in the current code base around the highway wrappers, could you explain where the speed-up comes from? Most of the changes look like cleanups or code reorganization.

mattip:
> Add class LogicalSpecific 

Could you add the benchmark? Preferably in a separate PR so it can be part of the code base and enable comparison to this PR

ixgbe:
> For someone not fluent in the current code base around the highway wrappers, could you explain where the speed-up comes from? Most of the changes look like cleanups or code reorganization.



> For someone not fluent in the current code base around the highway wrappers, could you explain where the speed-up comes from? Most of the changes look like cleanups or code reorganization.

Before my changes, SIMD acceleration on RISC-V was controlled by the NPY_SIMD macro, which is not defined for the RISC-V architecture. This means that no SIMD acceleration was enabled at all on RISC-V prior to this modification.

However, thanks to the groundwork laid in [https://github.com/numpy/numpy/pull/28622 ](https://github.com/numpy/numpy/pull/28622), Highway support is now available on RISC-V through the NPY_HWY macro. My changes build on that foundation, allowing NumPy to leverage Highway's optimized SIMD kernels on RISC-V platforms, where previously no vectorization was taking place.

In other words, this isn't just a refactor — it enables real performance acceleration on RISC-V for the first time by unlocking Highway-based vectorization.

ixgbe:
> Could you add the benchmark? Preferably in a separate PR so it can be part of the code base and enable comparison to this PR

Good point — I'll submit a follow-up PR adding benchmarks for these operations using ASV. That way we can measure the impact of this change and reuse the benchmarks going forward.

mattip:
> it enables real performance acceleration on RISC-V 

Ahh, got it. So what looks cosmetic is actually meaningful for RISC-V.


> PR #29676 

Sorry, I should have checked before asking for more work. Don't we already have ufunc benchmarks via `time_ufunc_types` (i.e. `spin bench -q -t time_ufunc_types`)? Although it seems there is no selector to benchmark the specific dtype, this seems to run the benchmarks on all the `get_squares_()` dtypes: int*, float* and complex128.

ixgbe:
> > it enables real performance acceleration on RISC-V
> 
> Ahh, got it. So what looks cosmetic is actually meaningful for RISC-V.
> 
> > PR #29676
> 
> Sorry, I should have checked before asking for more work. Don't we already have ufunc benchmarks via `time_ufunc_types` (i.e. `spin bench -q -t time_ufunc_types`)? Although it seems there is no selector to benchmark the specific dtype, this seems to run the benchmarks on all the `get_squares_()` dtypes: int*, float* and complex128.

Actually, I initially used:

`spin bench --compare HEAD~1 -t bench_ufunc.UFunc`

to evaluate performance. However, I ran into an issue: this benchmark runs across all data types (int, float, complex, etc.), and the reported result is an aggregated total time.

The problem is that improvements on a specific dtype (e.g., bool) can be masked by the overall sum, especially when other dtypes show no change or even slight regressions. Since asv continuous --factor 1.05 treats any speedup below 5% as noise or insignificant, a meaningful optimization on boolean ufuncs might not show up as a noticeable improvement in the total time.

To get a more sensitive and focused measurement, I created a dedicated benchmark class that isolates operations on bool arrays. This allows us to:

Clearly observe performance changes for logical ufuncs
Detect smaller but meaningful improvements
Establish a baseline for future optimizations (e.g., SIMD acceleration on RISC-V)
This way, we won’t miss architecture-specific gains just because they’re diluted in a global aggregate.


image
image

mattip:
Thanks @ixgbe

PR #29661, opened 2025-09-03
ENH: Add where= parameter support to np.ptp

KAVYANSHTYAGI:
### **PR Description**

This PR extends `numpy.ptp` (peak-to-peak: max - min) to support the `where=` keyword argument, consistent with other NumPy reduction functions such as `sum`, `min`, and `max`.

#### **Changes Made**

* Updated `_methods._ptp` to forward `where` into `umr_maximum` and `umr_minimum`.
* Extended public API in `fromnumeric.py` to accept `where=np._NoValue`.
* Updated docstrings to document the new `where` parameter.
* Updated type hints in `fromnumeric.pyi` to include `where`.
* Added unit tests in `test_numeric.py` (`test_ptp_where`) to validate functionality.

#### **Behavior**

* `np.ptp` now honors masks passed via `where`, enabling selective element inclusion in peak-to-peak computation.
* For example:

  ```python
  >>> import numpy as np
  >>> a = np.array([1, 5, 7, 0])
  >>> where = [True, False, True, False]
  >>> np.ptp(a, where=where)
  6
  ```
* Edge cases (e.g., all-False masks) follow the identity rules of `maximum` (`-inf`) and `minimum` (`+inf`), resulting in `-inf` for floating-point arrays.

#### **Testing**

* Added direct test for masked computation.
* Verified broadcasting behavior and compatibility with existing parameters (`axis`, `out`, `keepdims`).

#### **Documentation**

* Added `where` parameter description to the `ptp` docstring.
* `versionadded` tag should be included to mark this feature.

---

This aligns `ptp` with other NumPy reductions and improves API consistency.

---
Fix #29675

### DIFF
--- a/numpy/_core/fromnumeric.py
+++ b/numpy/_core/fromnumeric.py
@@ -2953,12 +2953,12 @@ def cumsum(a, axis=None, dtype=None, out=None):
     return _wrapfunc(a, 'cumsum', axis=axis, dtype=dtype, out=out)
 
 
-def _ptp_dispatcher(a, axis=None, out=None, keepdims=None):
+def _ptp_dispatcher(a, axis=None, out=None, keepdims=None, where=None):
     return (a, out)
 
 
 @array_function_dispatch(_ptp_dispatcher)
-def ptp(a, axis=None, out=None, keepdims=np._NoValue):
+def ptp(a, axis=None, out=None, keepdims=np._NoValue, where=np._NoValue):
     """
     Range of values (maximum - minimum) along an axis.
 
@@ -2997,6 +2997,10 @@ def ptp(a, axis=None, out=None, keepdims=np._NoValue):
         `ndarray`, however any non-default value will be.  If the
         sub-class' method does not implement `keepdims` any
         exceptions will be raised.
+
+    where : array_like of bool, optional
+        Elements to include in the peak-to-peak. See
+        `~numpy.ufunc.reduce` for details.
 
     Returns
     -------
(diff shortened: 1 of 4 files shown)


mattip:
Before diving in to fix the failures, we should discuss the necessity for this change. Is there an open issue that describes the real world use case for adding the `where` argument to ptp?

KAVYANSHTYAGI (commented):
@mattip, I have opened an issue in which the whole scenario is elaborated. Thank you

numpy/_core/fromnumeric.py
```diff
@@ -2953,12 +2953,12 @@ def cumsum(a, axis=None, dtype=None, out=None):
...
 
 
-def _ptp_dispatcher(a, axis=None, out=None, keepdims=None):
+def _ptp_dispatcher(a, axis=None, out=None, keepdims=None, where=None):
     return (a, out)
 
 
 @array_function_dispatch(_ptp_dispatcher)
-def ptp(a, axis=None, out=None, keepdims=np._NoValue):
+def ptp(a, axis=None, out=None, keepdims=np._NoValue, where=np._NoValue):
```
seberg, line 2961: This should be kwarg only.

FWIW, I don't care much to do this, but I agree with Matti in that I am curious where this type of issues come from.  It feels like a lot more recently and context matters.

For example, if an automatic tool says: oh this might match better if it has a `where=`, then we'll never see the end of it.
OTOH, if you have an *actual* use, that is a reason to just do it (when there is little downside otherwise).

(The matter would be different if a maintainer spearheaded this so that we have dedicated review bandwidth.)

charris:
Need rebase.

PR #29662, opened 2025-09-03
BLD: Add missing <unordered_map> include

tobbi:
Fixes homebrew builds, see https://github.com/orgs/Homebrew/discussions/6386#discussioncomment-14293523


seberg (approved):
Looks clearly right, thanks!  Confusing that none of our builds noticed this.

PR #29666, opened 2025-09-03

bwhitt7:
This PR is related to #29552. This PR has the goal of making NumPy's testing suite more thread safe. This introduces changes to the setup methods for 2 more testing files, 1 in numpy/_core and 1 in numpy/testing. Pytest's classic xunit setup and teardown methods are thread unsafe, typically not running before the tests do, leading to AttributionErrors. This PR changes these setup method to basic "creation" methods that are called within the tests that need them, or for test_utils, creating class methods that are overridden when needed.

Notability, this PR introduces changes to RNG in the test suite. np.random.seed is thread-unsafe, so utilizing the new Generator will give us thead-safe RNG. In tests where the results of the RNG is important (such as tests under TestDot in test_multiarray.py), we can instead use RandomState, which is a thread safe implementation of the legacy random generator. Eventually it'll probably be a good idea to rewrite these tests with the new default RNG.

Both files still have some thread-unsafe functionality, such as the usage of temporary files in test_multiarray.py.

### DIFF
--- a/numpy/_core/tests/test_multiarray.py
+++ b/numpy/_core/tests/test_multiarray.py
@@ -6837,7 +6837,7 @@ class TestDot:
     N = 7
 
     def _create_data(self):
-        rng = np.random.default_rng(128)
+        rng = np.random.RandomState(128)
         A = rng.random((4, 2))
         b1 = rng.random((2, 1))
         b2 = rng.random(2)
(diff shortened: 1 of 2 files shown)


numpy/_core/tests/test_multiarray.py
```diff
@@ -6052,39 +6064,44 @@ def test_mmap_close(self):
...
-        self.b = a[::2, ::2]
-        self.a0 = a0
-        self.b0 = a0[::2, ::2]
+        b = a[::2, ::2]
+        return a, b
+
+    def _create_arrays0(self):
+        a0 = np.arange(20.0).reshape((4, 5))
+        b0 = a0[::2, ::2]
+        return a0, b0
```
ngoldbaum: this is only called once below so you can delete this kinda awkward function and inline its content in `test__array__`. But I'd construct `a0` and `b0` based on `a` and `b` as the `setup_method` used to, it's a little clearer what the intent is.

numpy/_core/tests/test_multiarray.py
```diff
@@ -6808,61 +6834,63 @@ def test_vdot_uncontiguous(self):
...
-        self.A = np.random.rand(4, 2)
-        self.b1 = np.random.rand(2, 1)
-        self.b2 = np.random.rand(2)
-        self.b3 = np.random.rand(1, 2)
-        self.b4 = np.random.rand(4)
-        self.N = 7
+    N = 7
+
+    def _create_data(self):
+        rng = np.random.RandomState(128)
```
ngoldbaum, line 6839: neat, I'm glad this was simple to change

ngoldbaum (commented):
Just a couple comments. The change to `test_utils` makes sense, seems like the most straightforward way to avoid the xunit setups in those.

numpy/_core/tests/test_multiarray.py
bwhitt7: Hmmm, technically `a` is supposed to be more based on `a0` in this case, such as how `a` changes its writable flag to False while `a0` doesn't modify that value. Maybe it's supposed to show that `a` changing that flag doesn't affect `a0`? Would it make sense to have `a0` explicitly set its flag to `True` in `test__array__`, or just redeclare all the values to be similar to how they were originally constructed?

```python
def test___array__(self):
        a, b = self._create_arrays()
        a0 = a.reshape((4, 5))
        a0.flags.writeable = True
        b0 = a0[::2, ::2]
        c = a.flat.__array__()
        ...
```

or

```python
def test___array__(self):
        a0 = np.arange(20.0)
        a = a0.reshape(4, 5)
        a0 = a0.reshape((4, 5))
        a.flags.writeable = False
        b = a[::2, ::2]
        b0 = a0[::2, ::2]
        c = a.flat.__array__()
        ...
```

Both result in the test passing, but I don't know which would be more "true" to the original intent of the test.

ngoldbaum: Let's go with option number 2. I don't think it actually matters, but that's closer to how this test was originally written.

bwhitt7:
@ngoldbaum Made the changes.

ngoldbaum:
I restarted the tests that were getting instantly cancelled yesterday and everything is green now.

Thanks @bwhitt7!

PR #29671, opened 2025-09-04
TST: Replace test_smoke xunit setup with methods

bwhitt7:
This PR is related to #29552. This PR has the goal of making NumPy's testing suite more thread safe. This introduces changes to the setup methods for numpy/random/tests/test_smoke.py. Pytest's classic xunit setup and teardown methods are thread unsafe, typically not running before the tests do, leading to AttributionErrors. Additionally, or test_smoke.py, defining the Generators at class level will lead to issues with parallel threads messing with the same Generator and creating unexpected results. This PR changes these setup method to basic "creation" methods that are called within the tests that need them.

### DIFF
--- a/numpy/random/tests/test_smoke.py
+++ b/numpy/random/tests/test_smoke.py
@@ -8,12 +8,8 @@ import numpy as np
 from numpy.random import MT19937, PCG64, PCG64DXSM, SFC64, Generator, Philox
 from numpy.testing import assert_, assert_array_equal, assert_equal
 
-
-@pytest.fixture(scope='module',
-                params=(np.bool, np.int8, np.int16, np.int32, np.int64,
-                        np.uint8, np.uint16, np.uint32, np.uint64))
-def dtype(request):
-    return request.param
+DTYPES_BOOL_INT_UINT = (np.bool, np.int8, np.int16, np.int32, np.int64,
+                        np.uint8, np.uint16, np.uint32, np.uint64)
 
 
 def params_0(f):
@@ -99,32 +95,22 @@ class RNGData:
     advance: int
     seed: list[int]
     rg: Generator
-    initial_state: dict
     seed_vector_bits: int
-    seed_error: Exception
 
 
 class RNG:
     @classmethod
-    def _create_gen(cls):
+    def _create_rng(cls):
         # Overridden in test classes. Place holder to silence IDE noise
         bit_generator = PCG64
         advance = None
         seed = [12345]
         rg = Generator(bit_generator(*seed))
-        initial_state = rg.bit_generator.state
         seed_vector_bits = 64
-        seed_error = TypeError
-        return RNGData(
-            bit_generator, advance, seed, rg,
-            initial_state, seed_vector_bits, seed_error
-        )
-
-    def _reset_state(self, data):
-        data.rg.bit_generator.state = data.initial_state
+        return RNGData(bit_generator, advance, seed, rg, seed_vector_bits)
 
     def test_init(self):
-        data = self._create_gen()
+        data = self._create_rng()
         data.rg = Generator(data.bit_generator())
         state = data.rg.bit_generator.state
         data.rg.standard_normal(1)
@@ -134,7 +120,7 @@ class RNG:
         assert_(comp_state(state, new_state))
 
     def test_advance(self):
-        data = self._create_gen()
+        data = self._create_rng()
         state = data.rg.bit_generator.state
         if hasattr(data.rg.bit_generator, 'advance'):
             data.rg.bit_generator.advance(data.advance)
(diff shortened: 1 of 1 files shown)


numpy/random/tests/test_smoke.py
```diff
@@ -94,28 +94,30 @@ def warmup(rg, n=None):
...
-        cls.initial_state = cls.rg.bit_generator.state
-        cls.seed_vector_bits = 64
-        cls._extra_setup()
+        bit_generator = PCG64
+        advance = None
+        seed = [12345]
+        rg = Generator(bit_generator(*seed))
+        initial_state = rg.bit_generator.state
+        seed_vector_bits = 64
+        return bit_generator, advance, seed, rg, initial_state, seed_vector_bits
```
ngoldbaum: IMO, it makes more sense in this case, with six return values, to return a dict. But actually given the way the tests are currently written, it would be a smaller change to return a namedtuple:

```
# added to module imports if it's not already there
from collections import namedtuple


RNGTestData = namedtuple('RNGTestData', ['bit_generator', 'advance', 'seed', 'rg', 'initial_state', 'seed_vector_bits'])
return RNGTestData(bit_generator=PCG64, advance=None, seed=[12345], initial_state, seed_vector_bits)
```

And then later in a test, you can do things like:

```
data = self._create_data()
state = data.bit_generator.state
```

ngoldbaum: Actually, I think we can use a dataclass here instead of a namedtuple, which might be a little nicer.

bwhitt7: Cool idea! Just implemented it.

numpy/random/tests/test_smoke.py
```diff
@@ -708,102 +781,127 @@ def test_integers_broadcast_errors(self, dtype):
...
+        advance = 2**63 + 2**31 + 2**15 + 1
+        seed = [12345]
+        rg = np.random.default_rng(*seed)
+        initial_state = rg.bit_generator.state
+        seed_vector_bits = 64
+        seed_error = TypeError
+        return RNGData(
+            bit_generator, advance, seed, rg,
+            initial_state, seed_vector_bits, seed_error
+        )
```
ngoldbaum: how about this instead:
```suggestion
        return RNGData(
            bit_generator = PCG64,
            advance = 2**63 + 2**31 + 2**15 + 1,
            seed = [12345],
            rg = np.random.default_rng(*seed),
            initial_state = rg.bit_generator.state,
            seed_vector_bits = 64,
            seed_error = TypeError,
        )
```

You can make a similar edit for the other `RNGData` you create above.

ngoldbaum:
Ping @rkern - would you mind taking a look at this? I'm pretty sure that the changes to remove state from the test class and force each test to explicitly create an RNG are correct but it'd be good to have eyes on this from someone who has touched this code in the past.

numpy/random/tests/test_smoke.py
```diff
@@ -92,403 +93,469 @@ def warmup(rg, n=None):
...
+        data = self._create_gen()
+        vals = data.rg.weibull(1.0, 10)
         assert_(len(vals) == 10)
 
     def test_zipf(self):
-        vals = self.rg.zipf(10, 10)
+        data = self._create_gen()
+        vec_1d = np.arange(2.0, 102.0)
+        vec_2d = np.arange(2.0, 102.0)[None, :]
+        mat = np.arange(2.0, 102.0, 0.01).reshape((100, 100))
```
ngoldbaum, line 437: This is the inlined content of `_extra_setup` and this matches what was in that function. Just commenting because it requires a bit of scrolling to confirm.

rkern:
I believe these changes are valid. I think the original `setup_class` was a mistake; these should have been `setup_method`s (i.e. called once before each test method invocation) since `self.rg` is frequently modified in each test (it just so happens to not cause a problem in our test suite because most of these tests are not about the precise details of the stream, just invariants that should be true no matter where you are in the stream, but nonetheless, test isolation is desirable). These modifications are in line with `setup_method` semantics, which I think are fine. Having `_create_rng` be a `@classmethod` is unnecessary (but also not harmful).

That said, I am not *particularly* a fan of having to avoid this standard feature of `pytest/unittest` to support a separate tool, though. It seems like we are incurring a ton of churn and a new heavy restriction on how we write tests. Have all the options to solve this in `pytest-run-parallel` been exhausted? Would it be possible to dispatch a (suitably deep, a la `dill` or `cloudpickle`) clone of the collected test item to each worker thread work? Then at least `setup_method` or `TestCase.setUp` would work. Coordinating `setup_module/setup_class/TestCase.setUpClass` could be trickier, I imagine.

ngoldbaum:
> That said, I am not particularly a fan of having to avoid this standard feature of pytest/unittest to support a separate tool, though. It seems like we are incurring a ton of churn and a new heavy restriction on how we write tests. Have all the options to solve this in pytest-run-parallel been exhausted? Would it be possible to dispatch a (suitably deep, a la dill or cloudpickle) clone of the collected test item to each worker thread work? Then at least setup_method or TestCase.setUp would work. Coordinating setup_module/setup_class/TestCase.setUpClass could be trickier, I imagine.

Ultimately this comes down to https://github.com/Quansight-Labs/pytest-run-parallel/issues/14. My understanding after going through the pytest docs is that fixing this requires re-implementing the full pytest testing protocol. Right now, we only monkeypatch the actual test execution.

We could also use fixtures, but then the test instances would share the fixtures too for the same reason. See https://github.com/numpy/numpy/issues/29552#issuecomment-3187030766 for more from @rgommers about how we came up with this approach.

I understand the desire to avoid churn. That said, IMO the explicit per-test setup is clearer than the xunit-style setup. It's more verbose than pytest fixtures, but it's also less magical.

The churn isn't for nothing - we also get the benefit of testing large chunks of the NumPy API in thread pools to elucidate any global state that we might have missed in the first pass to add free-threaded support. We've already discovered some reports of data races from TSan that we want to follow up on before expanding the TSan CI test coverage.

rkern:
Did you end up looking at the `pytest_runtest_protocol` ([per your comment](https://github.com/Quansight-Labs/pytest-run-parallel/issues/14#issuecomment-3164809815))? That would seem to be the more robust place to hook in. It probably wouldn't solve every permutation, but would probably handle a good chunk.

ngoldbaum:
That’s what I was referring to here:

> My understanding after going through the pytest docs is that fixing this requires re-implementing the full pytest testing protocol.

We’ve been focusing on other priorities (Python 3.14 support, lately) and I haven’t sat down to try to implement it because my impression from the pytest docs is we’d have to re-implement a decent subset of pytest itself.

Numpy is also the first project we’ve come across that uses xunit style tests so heavily. This issue causes less churn in other projects that rely on function-scope fixtures that don’t get cleaned up until after all the threads are done running tests.

rgommers:
I commented in https://github.com/Quansight-Labs/pytest-run-parallel/issues/14 on trying to explore `pytest_runtest_protocol`. Giving both the large amount of work and the effectively deprecated status of that protocol, it is unlikely to be a short-term change that can be made though.

> We could also use fixtures, but then the test instances would share the fixtures too for the same reason. See [#29552 (comment)](https://github.com/numpy/numpy/issues/29552#issuecomment-3187030766) for more from @rgommers about how we came up with this approach.

We had some discussion about this in the community meeting yesterday, with varying opinions. There was a question about whether we have any guidelines for this - we do ([here](https://numpy.org/devdocs/reference/testing.html#easier-setup-and-teardown-functions-methods), but it's brief and only says "*Setup and teardown functions to functions and methods are known as “fixtures”, and they should be used sparingly*" without explaining why. The main reason is fixtures, both the `setup_*` ones and the `pytest`-specific dependency injection ones, are _hard to debug and discover_. It's effectively non-local control flow or monkeypatching, modifying some state globally or at the module/class level. From basic principles, this is strictly worse than doing things in a functional style when both styles are possible with a similar amount of code. 

To give a simple example here from this same file:
- Problem 1: There are exactly 3 tests in that file that are parametrized, however it is hard to see that from reading the code. Even if you do spot the fixture near the top of the file, it's quite hard to find which tests use that fixture (a regular search in your editor for `dtype` will not help you).
- Problem 2: If you comment out that fixture to see what starts failing, the failures contain confusing errors like `Invalid integer data type 'T'`. Where did that come from? A _second_ fixture named `dtype` at a more global level in `numpy/conftest.py`. You won't find that unless you've been bitten by fixtures before and remember to look in `conftest.py`.
    - Request: @ngoldbaum can you please try to clean up those fixtures in `conftest.py`? That code is wrong on multiple levels - none of the four fixtures there should be there, or at the very least not have such nonspecific unsearchable names as well as have good code comments on why they're there.

Here is a simple way to make that code more idiomatic and easier to understand is, in the same number of lines of code:

<details>
<summary>Diff for smoke_test.py using parametrize()</summary>

```diff
diff --git a/numpy/random/tests/test_smoke.py b/numpy/random/tests/test_smoke.py
index 6f07443f79..02976f4121 100644
--- a/numpy/random/tests/test_smoke.py
+++ b/numpy/random/tests/test_smoke.py
@@ -8,12 +8,8 @@
 from numpy.testing import assert_, assert_array_equal, assert_equal
 
 
-@pytest.fixture(scope='module',
-                params=(np.bool, np.int8, np.int16, np.int32, np.int64,
-                        np.uint8, np.uint16, np.uint32, np.uint64))
-def dtype(request):
-    return request.param
-
+DTYPES_BOOL_INT_UINT = (np.bool, np.int8, np.int16, np.int32, np.int64,
+                        np.uint8, np.uint16, np.uint32, np.uint64)
 
 def params_0(f):
     val = f()
@@ -653,6 +649,7 @@ def test_output_fill_error(self):
         with pytest.raises(ValueError):
             rg.standard_gamma(1.0, out=existing[::3])
 
+    @pytest.mark.parametrize("dtype", DTYPES_BOOL_INT_UINT)
     def test_integers_broadcast(self, dtype):
         if dtype == np.bool:
             upper = 2
@@ -686,6 +683,7 @@ def test_integers_broadcast(self, dtype):
         b = self.rg.integers([upper] * 10, dtype=dtype)
         assert_equal(a, b)
 
+    @pytest.mark.parametrize("dtype", DTYPES_BOOL_INT_UINT)
     def test_integers_numpy(self, dtype):
         high = np.array([1])
         low = np.array([0])
@@ -699,6 +697,7 @@ def test_integers_numpy(self, dtype):
         out = self.rg.integers(low, high[0], dtype=dtype)
         assert out.shape == (1,)
 
+    @pytest.mark.parametrize("dtype", DTYPES_BOOL_INT_UINT)
     def test_integers_broadcast_errors(self, dtype):
         if dtype == np.bool:
             upper = 2
```

</details>

Another example where this PR already does improve the code is for the `_extra_setup` method. That is a classmethod called from `setup_method`, and that call has to be repeated an extra 6 times in the `setup_method` of subclasses. All that for a few variables used in a single test, which should have just been defined locally in that test. We'd never accept such oddly written code in a new PR, it's only there because that code was added in bulk and hence not reviewed line by line like we do for normal-sized PRs. 

* * *

Back to what to do next in this PR, and after that. I have several suggestions:
- We should update our testing guidelines a bit, because it doesn't explain what's wrong with fixtures and why to limit their use to the minimum necessary. I can take this action.
- @ngoldbaum I think it'd be good to start investigating what a redesign would take, and opening a conversation with `pytest` maintainers about what to do since they plan to replace `pytest_runtest_protocol`. Just that talking and sketching out a plan shouldn't take a huge amount of time I'd think
- @rkern I suggest accepting these changes in principle, after some further improvements (see below). Because yes it's some churn, but (a) it'll be quite helpful both for improving thread-safety, (b) no one objected to the plan to make these changes for weeks, and 8 PRs along these lines have already been merged (and they're an improvement on average), and (c) it will also make the test code better: 
    - I went through the remaining `setup_method` usages, and the vast majority can be replaced by something better. Sometimes it's used just to set a property on some subclass, that could be a plain attribute. There's a lot of `geterr`/`seterr` that predates the ability of using `np.errstate` as a context manager. There's manual temporary file creation, which is one case where it is appropriate to use a `pytest` builtin fixture. And there's also a number of places where it's simply unnecessary (`self.a = np.arange(10)` isn't a good `setup_method` body). 
    - For the minority of cases where it's appropriate to keep using `setup_method`, those can be marked with `pytest.mark.thread_unsafe`.
- @bwhitt7 it'd be great to try and make this PR and the test code in it as clean as possible. Suggestions:
    - Adopt my diff for using `parametrize` from above.
    - `seed_error` isn't actually used at all for testing AFAICT, so you can delete it everywhere
    - `_reset_state` should not be a method; it's only used in a single test (`test_integers_broadcast`), so define it locally as a function within that test.
        - Related: `initial_state` can be removed from the dataclass and all test subclasses as well, since it's always defined as `rg.bit_generator.state` and only used inside `_reset_state`. 
        - At that point `return RNGData(...)` fits on a single line rather than four lines - yet more boilerplate removal.
    - Use `rg = self._create_gen().rg` as the first line of the test where possible. In most tests only `data.rg` is used, and the `data.` is very repetitive for no reason. The end result will be a cleaner diff with just a lot of `self.` removals.
    - At that point I think this PR will look like a clear win; please have another look through for other opportunities of making things simpler and more functional.

numpy/random/tests/test_smoke.py
rgommers: This suggestion doesn't actually work, since both `bit_generator` and `seed` are defined and then used on next lines. I made a different suggestion in my long comment.

ngoldbaum:
> can you please try to clean up those fixtures in conftest.py? That code is wrong on multiple levels - none of the four fixtures there should be there, or at the very least not have such nonspecific unsearchable names as well as have good code comments on why they're there.

You're right, this needs a cleanup. To give a little context, the fixtures originally came from the stringdtype prototype. After noticing that I was often skipping corner cases, I found it useful to combine fixtures and very quickly build of a test matrix of common parameters I can use for every test I wrote. That is one feature of fixtures that's really nice, although you're right it comes at the price of making the parameters of test functions be nonlocal and require knowledge of pytest trivia to understand.

Eventually, the fixtures moved out of `test_stringdtype.py`, where the generic names made more sense, and into the common `conftest.py` because we wanted to move one or two of the stringdtype tests into the multithreading tests. We should have flagged the bad names at that point but this was missed. Thanks for catching it now.

I'll see if it's possible to clean this up.

rkern:
> * (b) no one objected to the plan to make these changes for weeks, and 8 PRs along these lines have already been merged

I'd be happier about this point if this plan had been the outcome of a discussion on the mailing list making it clear that there was going to be some hefty new constraints on writing tests going forward. All I see is the originating Github issue and then mostly within-Quansight interactions on the subsequent PRs. I didn't become alerted to this until I was explicitly pinged for `np.random` reasons. We don't have explicit policies about this (like we do for public API changes), but I do suggest that this is a policy change that rises to the same level and should have hit the mailing list rather than just a Github issue.

I'll give my retroactive consent given the factors that have finally been explained here, but let's be real clear that it's retroactive, and the silence until now isn't evidence that consent was given ahead of time.

@bwhitt7, to be clear, you've done great, detailed work that I appreciate. We will be getting a lot out of this.

ngoldbaum:
> but I do suggest that this is a policy change that rises to the same level and should have hit the mailing list rather than just a Github issue.

Feedback received. We did bring this up at a community meeting, but you're right, we should have pinged the mailing list before starting work. I'll circle back with @bwhitt7 so we can write a message on the mailing list to alert anyone else who missed this.

Thanks for your practicality here despite my lapse in not following community norms about the mailing list.

rkern:
Thank you! I do appreciate it. Let's chalk this up to testing our asynchronous free-threaded project communication protocols and finding some race conditions that all work out in eventual consistency in the end. :-)

bwhitt7:
@rgommers Thanks for the suggestions! Went ahead and implemented them. I renamed the creation method to _create_rng since I felt like that was more clear. There was also a few times where bit_generator was the only value being used, so I sometimes did bit_generator = self._create_rng().bit_generator here too.

@rkern You're 100% right about the mailing list, I'll work with Nathan to submit something there. Thank you for your help and feedback so far!

ngoldbaum (approved):
All LGTM!

rgommers (approved):
Thanks @bwhitt7, this looks good to me now. Thanks @rkern and @ngoldbaum as well, looks like we arrived at a good path forward. In it goes.

PR #29672, opened 2025-09-05
MAINT: Update pythoncapi-compat from main.

charris:
Best to keep active maintenance branches in sync.


charris:
The error looks unrelated.

PR #29678, opened 2025-09-05
BUG, BLD: Correct regex for ppc64 VSX3/VSX4 feature detection

Sanjaykumar030:
This PR fixes the regex patterns used to detect VSX3 and VSX4 CPU features
in the ppc64 Meson build scripts. Previously, the regex used a character
class:

    '.*[mcpu=|vsx].*'

This incorrectly matched individual characters instead of the intended
substrings "mcpu=" or "vsx".

Changes made:
- Updated the regex to use a non-capturing group with alternation:

      '.*(?:mcpu=|vsx).*'

- No changes were made to feature hierarchy, test code, extra tests, or
  compiler flags.

This ensures that the build system correctly detects VSX3/VSX4 features
on both GCC and Clang compilers on ppc64 platforms.

---

#### Before:
- VSX3/VSX4 detection could fail or mis-detect features because of
  improper regex.

#### After:
- Regex properly matches either "mcpu=" or "vsx" as intended.
- Extra tests and compiler flags remain unchanged.

### DIFF
--- a/meson_cpu/ppc64/meson.build
+++ b/meson_cpu/ppc64/meson.build
@@ -23,7 +23,7 @@ if host_machine.endian() == 'little'
   VSX.update(implies: VSX2)
 endif
 VSX3 = mod_features.new(
-  'VSX3', 3, implies: VSX2, args: {'val': '-mcpu=power9', 'match': '.*[mcpu=|vsx].*'},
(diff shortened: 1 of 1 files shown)


charris (approved):

Sanjaykumar030:
Thank you, @charris, for the review, and the merge! I appreciate the opportunity to contribute.

PR #29692, opened 2025-09-05
TST: Replace test_deprecations setup/teardown with context manager

bwhitt7:
This PR is related to #29552. This PR has the goal of making NumPy's testing suite more thread safe. This introduces changes to the setup methods for `numpy/_core/tests/test_deprecations.py`. Pytest's classic xunit setup and teardown methods are thread unsafe, typically not running before the tests do, leading to AttributionErrors.

With the help of @ngoldbaum , we determined that replacing the setup method with a context manage would be the way to go. Another option would be using a pytest fixture, but they only run once per test. Multiple threads would share the same fixture, leading to assertion errors with the number of warnings captured.

### DIFF
--- a/numpy/_core/tests/test_deprecations.py
+++ b/numpy/_core/tests/test_deprecations.py
@@ -20,22 +20,20 @@ class _DeprecationTestCase:
     message = ''
     warning_cls = DeprecationWarning
 
-    def setup_method(self):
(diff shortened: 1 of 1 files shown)


ngoldbaum:
Thanks @bwhitt7!

PR #29693, opened 2025-09-05
TST: xfail test_kind::test_quad_precision on AIX/PPC

kkwli:
On AIX, cmake's `platform.machine()` does not return a meaningful string that can be used to determine the architecture. This PR is to add the check of `platform.processor()` to xfail the test.

### DIFF
--- a/numpy/f2py/tests/test_kind.py
+++ b/numpy/f2py/tests/test_kind.py
@@ -10,7 +10,7 @@ from numpy.f2py.crackfortran import (
 
 from . import util
 
-IS_PPC = platform.machine().lower().startswith("ppc") or platform.processor().lower().startswith("powerpc")
+IS_PPC_OR_AIX = platform.machine().lower().startswith("ppc") or platform.system() == 'AIX'
 
 class TestKind(util.F2PyTest):
     sources = [util.getpath("tests", "src", "kind", "foo.f90")]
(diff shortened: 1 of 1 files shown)


numpy/f2py/tests/test_kind.py
```diff
@@ -37,7 +38,7 @@ def test_real(self):
                 i
             ), f"selectedrealkind({i}): expected {selected_real_kind(i)!r} but got {selectedrealkind(i)!r}"
 
-    @pytest.mark.xfail(platform.machine().lower().startswith("ppc"),
+    @pytest.mark.xfail(IS_PPC,
```
tylerjereddy: > cmake's platform.machine()

What does CMake have to do with NumPy in this context? It is being used to compile CPython somewhere or? I thought CPython was still using an older build system.

Also, if the root of the problem is `AIX` OS itself, would it make sense to just check for it directly with `sys.platform`, which works just fine to pick up `aix`? https://docs.python.org/3/library/sys.html#sys.platform

Can you provide some context--you're working on an AIX machine and this test failed?

kkwli: Thanks for the review. My apologies. I meant python's `platform.machine()`. On my AIX machine, `platform.machine()` returns `00FA13034C00` which is not useful. Anyway, checking `sys.platform == 'aix'` or `platform.system() == 'AIX'` is fine for this purpose.

> Can you provide some context--you're working on an AIX machine and this test failed?

Yes, I am working on AIX and trying to use flang in numpy.f2py. On AIX, Fortran's `selected_real_kind(32)` returns -1. This test is already xfailed for ppc platform.

ngoldbaum (approved):
Makes sense!

PR #29695, opened 2025-09-06
TYP: fix ``np.bool`` method declarations

jorenham:
This partially addresses the incorrect way that we've been defining the methods of scalar types. Currently, this accidentally woirks in mypy, but there are good reasons for why it shouldn't. But as it currently stands, mypy can't address this, because that would cause problems for many libraries that use numpy.

Unless mypy_primer complains, then I'd like to backjport this, so that mypy doesn't have to wait ~5 months or so.

See https://github.com/python/mypy/pull/15993#issuecomment-3240220599 for context.

Note that the integer and floating scalar-type also have this issue. I'll address those in separate PR's (this weekend, hopefully).


charris:
Thanks Joren.

PR #29697, opened 2025-09-06
BUG: Correct ambiguous logic for `s390x` CPU feature detection

Sanjaykumar030:
This PR continues the work from #29678 to improve CPU feature detection in the Meson build system, specifically for the `s390x` architecture.

Problem:
The current detection for `VXE` and `VXE2` is ambiguous:
- VXE2: The regex 'VX.*' is overly greedy and could match any string containing “VX”.
- VXE: The simple match for 'VX' could incorrectly enable VXE on systems that only support VX.

These ambiguities may cause NumPy to compile with unsupported CPU instructions, risking Illegal Instruction crashes or performance issues.

Solution:
This PR replaces the fragile patterns with precise regular expressions using word boundaries (`\b`):
- `VXE → '\\bvxe\\b'`
- `VXE2 → '\\bvxe2\\b'`

No changes are made to the feature hierarchy, compiler flags, or test code.

### DIFF
--- a/meson_cpu/s390x/meson.build
+++ b/meson_cpu/s390x/meson.build
@@ -7,12 +7,12 @@ VX = mod_features.new(
 )
 VXE = mod_features.new(
   'VXE', 2, implies: VX, args: {'val': '-march=arch12', 'match': '-march=.*'},
-  detect: {'val': 'VXE', 'match': 'VX'},
+  detect: {'val': 'VXE', 'match': '\\bvxe\\b'},
   test_code: files(source_root + '/numpy/distutils/checks/cpu_vxe.c')[0],
 )
 VXE2 = mod_features.new(
   'VXE2', 3, implies: VXE, args: {'val': '-march=arch13', 'match': '-march=.*'},
-  detect: {'val': 'VXE2', 'match': 'VX.*'},
+  detect: {'val': 'VXE2', 'match': '\\bvxe2\\b'},
   test_code: files(source_root + '/numpy/distutils/checks/cpu_vxe2.c')[0],
 )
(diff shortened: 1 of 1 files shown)


Sanjaykumar030:
Hi @jorenham,

This PR refines the s390x CPU feature detection by replacing ambiguous VXE and VXE2 regex patterns with precise word-boundary matches. All tests pass, and there are no changes to compiler flags or feature hierarchy. Could you please review when you have a moment?

Thanks!

charris:
Thanks @Sanjaykumar030 .

charris:
I'm am curious about the change to lower case. Is the match independent of case?

Sanjaykumar030:
Thanks for the question, @charris. The new regex patterns are case-sensitive, chosen to align with the lowercase feature strings reported on s390x. This keeps the behavior consistent with the existing build logic and avoids any breaking changes.

jorenham:
@charris You might want to have a look at [this PR](https://github.com/scipy/scipy-stubs/pull/862#pullrequestreview-3194501912) by the same author.

seberg:
Just went through my emails, @ngoldbaum can you have a brief look at this?  I don't know how this matching works, but given Joren's comment, it is unclear to me that the author does.

Sanjaykumar030:
Great if any other maintainer is double-checking for safety. For clarity, here’s a comparison of old vs new behavior:

CPU info: vx  
OLD -> VX=True, VXE=False, VXE2=False  
NEW -> VX=True, VXE=False, VXE2=False  

CPU info: vxe  
OLD -> VX=True, VXE=True, VXE2=False  
NEW -> VX=False, VXE=True, VXE2=False  

CPU info: vxenabled  
OLD -> VX=True  
NEW -> no match

PR #23399, opened 2023-03-15
ENH: float64 sin/cos using Numpy intrinsics

Mousius:
This takes the [sin](https://github.com/ARM-software/optimized-routines/blob/master/math/v_sin.c) and [cos](https://github.com/ARM-software/optimized-routines/blob/master/math/v_cos.c) algorithms from Optimized Routines under MIT license, and converts them to Numpy intrinsics.

The routines are within the ULP boundaries of other vectorised math routines (<4ULP). The routines reduce performance in some special cases but improves normal cases. Comparing to the SVML implementation, these routines are more performant in special cases, we're therefore safe to assume the performance is acceptable for AArch64 as well.

| performance ratio (lower is better)  | benchmark |
| ----  | ---- |
| 1.8   | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc	'cos'>	4	2	'd') |
| 1.79  | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc	'cos'>	4	4	'd') |
| 1.77  | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc	'cos'>	4	1	'd') |
| 1.74  | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc	'cos'>	2	2	'd') |
| 1.74  | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc	'cos'>	2	4	'd') |
| 1.72  | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc	'cos'>	2	1	'd') |
| 1.6   | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc	'cos'>	1	2	'd') |
| 1.6   | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc	'cos'>	1	4	'd') |
| 1.56  | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc	'cos'>	1	1	'd') |
| 1.42  | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc	'sin'>	2	2	'd') |
| 1.41  | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc	'sin'>	2	4	'd') |
| 1.37  | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc	'sin'>	2	1	'd') |
[...]

PR #29699, opened 2025-09-06
ENH: Add SIMD sin/cos implementation with numpy-simd-routines

seiko2plus:
numpy-simd-routines added as subrepo in meson subprojects
  directory and the current FP configuration is static, ~1ulp used for double-precision
  ~4ulp for single-precision with handling floating-point errors,
  special-cases extended precision for large arguments,
  subnormals are enabled by default too.

  numpy-simd-routines supports all SIMD extensions that are supported
  by Google Highway including non-FMA extensions and is fully independent
  from libm to guarantee unified results across all compilers and
  platforms.

  Note: that there was no SIMD optimization enabled for sin/cos
  for double-precision before, only single-precision.

# Benchmark

## X86 Platform
The following benchmark was tested against GCC 14.3.0 on an x86 CPU (Ryzen 7 7700X).

<details>
<summary>Environment</summary>

```Bash
> uname -a
Linux seiko-pc 6.12.60 #1-NixOS SMP PREEMPT_DYNAMIC Mon Dec  1 10:43:41 UTC 2025 x86_64 GNU/Linux
> gcc --version
gcc (GCC) 14.3.0
Copyright (C) 2024 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
> python --version
Python 3.12.12
> lscpu
Architecture:                x86_64
  CPU op-mode(s):            32-bit, 64-bit
  Address sizes:             48 bits physical, 48 bits virtual
  Byte Order:                Little Endian
CPU(s):                      16
  On-line CPU(s) list:       0-15
Vendor ID:                   AuthenticAMD
  Model name:                AMD Ryzen 7 7700X 8-Core Processor
    CPU family:              25
    Model:                   97
    Thread(s) per core:      2
    Core(s) per socket:      8
    Socket(s):               1
    Stepping:                2
    Frequency boost:         enabled
    CPU(s) scaling MHz:      72%
    CPU max MHz:             5573.0000
    CPU min MHz:             545.0000
    BogoMIPS:                8982.62
    Flags:                   fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx mmxext fxsr_opt pdpe1gb rdtscp lm constant_tsc rep_good amd_lbr_v2 nopl 
                             xtopology nonstop_tsc cpuid extd_apicid aperfmperf rapl pni pclmulqdq monitor ssse3 fma cx16 sse4_1 sse4_2 movbe popcnt aes xsave avx f16c rdrand lahf_lm cmp_legacy svm extapic cr8_leg
                             acy abm sse4a misalignsse 3dnowprefetch osvw ibs skinit wdt tce topoext perfctr_core perfctr_nb bpext perfctr_llc mwaitx cpb cat_l3 cdp_l3 hw_pstate ssbd mba perfmon_v2 ibrs ibpb stibp
                              ibrs_enhanced vmmcall fsgsbase bmi1 avx2 smep bmi2 erms invpcid cqm rdt_a avx512f avx512dq rdseed adx smap avx512ifma clflushopt clwb avx512cd sha_ni avx512bw avx512vl xsaveopt xsavec
                              xgetbv1 xsaves cqm_llc cqm_occup_llc cqm_mbm_total cqm_mbm_local user_shstk avx512_bf16 clzero irperf xsaveerptr rdpru wbnoinvd cppc arat npt lbrv svm_lock nrip_save tsc_scale vmcb_cl
                             ean flushbyasid decodeassists pausefilter pfthreshold avic vgif x2avic v_spec_ctrl vnmi avx512vbmi umip pku ospke avx512_vbmi2 gfni vaes vpclmulqdq avx512_vnni avx512_bitalg avx512_vpo
                             pcntdq rdpid overflow_recov succor smca fsrm flush_l1d amd_lbr_pmc_freeze
```
</details>

<details>
<summary>x86-64-v4</summary>

```
export NPY_DISABLE_CPU_FEATURES=""
spin bench --compare parent/main --cpu-affinity 7 -t "\(<ufunc '(cos|sin)'>"
```

| Change   | Before [b70fc774] <brings_npsr~2>   | After [c5fc8423] <brings_npsr>   |   Ratio | Benchmark (Parameter)                                                   |
|----------|-------------------------------------|----------------------------------|---------|-------------------------------------------------------------------------|
| -        | 689±6μs                             | 640±30μs                         |    0.93 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 4, 1, 'f')        |
| -        | 683±0.9μs                           | 602±2μs                          |    0.88 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 1, 2, 'f')        |
| -        | 682±1μs                             | 603±1μs                          |    0.88 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 4, 1, 'f')        |
| -        | 695±0.7μs                           | 604±1μs                          |    0.87 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 1, 2, 'f')        |
| -        | 354±0.7μs                           | 231±0.4μs                        |    0.65 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 1, 1, 'f')        |
| -        | 363±0.2μs                           | 234±4μs                          |    0.64 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 1, 1, 'f')        |
| -        | 1.04±0.02ms                         | 632±20μs                         |    0.61 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 4, 2, 'f')        |
| -        | 1.02±0ms                            | 612±10μs                         |    0.6  | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 4, 2, 'f')        |
| -        | 1.32±0.01ms                         | 795±2μs                          |    0.6  | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 1, 2, 'f') |
| -        | 1.72±0.03ms                         | 806±3μs                          |    0.47 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 4, 1, 'f') |
| -        | 1.44±0.02ms                         | 603±0.3μs                        |    0.42 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 1, 2, 'f') |
| -        | 5.05±0.1ms                          | 2.07±0.01ms                      |    0.41 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 4, 2, 'd') |
| -        | 4.98±0.01ms                         | 1.88±0.03ms                      |    0.38 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 4, 1, 'd') |
| -        | 5.95±0.02ms                         | 2.26±0.07ms                      |    0.38 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 4, 2, 'd') |
| -        | 2.18±0.05ms                         | 792±1μs                          |    0.36 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 4, 2, 'f') |
| -        | 1.78±0.03ms                         | 608±4μs                          |    0.34 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 4, 1, 'f') |
| -        | 5.65±0.02ms                         | 1.87±0.2ms                       |    0.33 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 4, 1, 'd')        |
| -        | 5.65±0.01ms                         | 1.87±0.01ms                      |    0.33 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 4, 2, 'd')        |
| -        | 5.77±0.01ms                         | 1.90±0.01ms                      |    0.33 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 4, 2, 'd')        |
| -        | 6.00±0.04ms                         | 1.98±0.02ms                      |    0.33 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 4, 1, 'd') |
| -        | 1.36±0.01ms                         | 443±2μs                          |    0.32 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 1, 1, 'f') |
| -        | 5.75±0ms                            | 1.68±0.04ms                      |    0.29 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 4, 1, 'd')        |
| -        | 4.97±0ms                            | 1.46±0.04ms                      |    0.29 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 1, 2, 'd') |
| -        | 2.42±0.2ms                          | 609±2μs                          |    0.25 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 4, 2, 'f') |
| -        | 5.62±0.01ms                         | 1.25±0.02ms                      |    0.22 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 1, 2, 'd')        |
| -        | 5.72±0ms                            | 1.25±0.01ms                      |    0.22 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 1, 2, 'd')        |
| -        | 4.97±0ms                            | 1.08±0ms                         |    0.22 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 1, 1, 'd') |
| -        | 7.08±0.01ms                         | 1.58±0.02ms                      |    0.22 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 1, 2, 'd') |
| -        | 7.07±0ms                            | 1.20±0.01ms                      |    0.17 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 1, 1, 'd') |
| -        | 1.49±0ms                            | 233±10μs                         |    0.16 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 1, 1, 'f') |
| -        | 5.61±0ms                            | 828±0.4μs                        |    0.15 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 1, 1, 'd')        |
| -        | 5.73±0ms                            | 859±8μs                          |    0.15 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 1, 1, 'd')        |

</details>

<details>
<summary>x86-64-v3</summary>

```
export NPY_DISABLE_CPU_FEATURES="X86_V4"
spin bench --compare parent/main --cpu-affinity 7 -t "\(<ufunc '(cos|sin)'>"
```

| Change   | Before [b70fc774] <brings_npsr~2>   | After [c5fc8423] <brings_npsr>   |   Ratio | Benchmark (Parameter)                                                   |
|----------|-------------------------------------|----------------------------------|---------|-------------------------------------------------------------------------|
| +        | 844±1μs                             | 939±50μs                         |    1.11 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 1, 2, 'f')        |
| -        | 4.98±0ms                            | 4.57±0.1ms                       |    0.92 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 1, 2, 'd') |
| -        | 4.98±0ms                            | 4.20±0.01ms                      |    0.84 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 1, 1, 'd') |
| -        | 6.03±0.06ms                         | 5.00±0.06ms                      |    0.83 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 4, 2, 'd') |
| -        | 6.06±0.06ms                         | 4.78±0.04ms                      |    0.79 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 4, 1, 'd') |
| -        | 1.37±0.01ms                         | 913±40μs                         |    0.67 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 4, 1, 'f')        |
| -        | 1.37±0.01ms                         | 901±10μs                         |    0.66 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 4, 1, 'f')        |
| -        | 1.45±0ms                            | 941±2μs                          |    0.65 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 4, 2, 'f')        |
| -        | 830±0.3μs                           | 511±2μs                          |    0.62 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 1, 1, 'f')        |
| -        | 7.07±0.01ms                         | 4.35±0.1ms                       |    0.61 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 1, 2, 'd') |
| -        | 844±2μs                             | 496±5μs                          |    0.59 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 1, 1, 'f')        |
| -        | 1.46±0.01ms                         | 864±9μs                          |    0.59 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 4, 2, 'f')        |
| -        | 3.21±0.03ms                         | 1.83±0ms                         |    0.57 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 1, 2, 'f') |
| -        | 7.08±0.01ms                         | 3.98±0.06ms                      |    0.56 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 1, 1, 'd') |
| -        | 3.63±0.09ms                         | 1.84±0.01ms                      |    0.51 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 4, 2, 'f') |
| -        | 5.65±0.01ms                         | 2.82±0.1ms                       |    0.5  | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 4, 2, 'd')        |
| -        | 5.64±0ms                            | 2.74±0.3ms                       |    0.49 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 4, 1, 'd')        |
| -        | 5.75±0ms                            | 2.76±0.1ms                       |    0.48 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 4, 1, 'd')        |
| -        | 3.00±0.09ms                         | 1.44±0ms                         |    0.48 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 1, 1, 'f') |
| -        | 3.78±0.1ms                          | 1.82±0.02ms                      |    0.48 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 4, 1, 'f') |
| -        | 5.77±0.01ms                         | 2.68±0.02ms                      |    0.46 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 4, 2, 'd')        |
| -        | 5.73±0ms                            | 2.11±0.03ms                      |    0.37 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 1, 2, 'd')        |
| -        | 5.62±0ms                            | 2.03±0.01ms                      |    0.36 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 1, 2, 'd')        |
| -        | 5.62±0.02ms                         | 1.67±0ms                         |    0.3  | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 1, 1, 'd')        |
| -        | 5.72±0ms                            | 1.73±0ms                         |    0.3  | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 1, 1, 'd')        |
| -        | 3.23±0.02ms                         | 866±10μs                         |    0.27 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 1, 2, 'f') |
| -        | 3.48±0.1ms                          | 856±3μs                          |    0.25 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 4, 2, 'f') |
| -        | 3.68±0.06ms                         | 851±2μs                          |    0.23 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 4, 1, 'f') |
| -        | 3.03±0.08ms                         | 497±4μs                          |    0.16 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 1, 1, 'f') |

</details>

<details>
<summary>x86-64-v2</summary>

```
export NPY_DISABLE_CPU_FEATURES="X86_V4 X86_V3"
spin bench --compare parent/main --cpu-affinity 7 -t "\(<ufunc '(cos|sin)'>"
```

| Change   | Before [b70fc774] <brings_npsr~2>   | After [c5fc8423] <brings_npsr>   |   Ratio | Benchmark (Parameter)                                                   |
|----------|-------------------------------------|----------------------------------|---------|-------------------------------------------------------------------------|
| +        | 4.98±0ms                            | 7.04±0.1ms                       |    1.41 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 4, 2, 'd') |
| +        | 2.12±0ms                            | 2.88±0.05ms                      |    1.36 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 4, 1, 'f') |
| +        | 4.97±0.01ms                         | 6.68±0.02ms                      |    1.34 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 4, 1, 'd') |
| +        | 2.12±0ms                            | 2.84±0.01ms                      |    1.34 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 4, 2, 'f') |
| +        | 2.13±0ms                            | 2.82±0ms                         |    1.32 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 1, 2, 'f') |
| +        | 4.98±0.01ms                         | 6.19±0.05ms                      |    1.24 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 1, 2, 'd') |
| +        | 4.97±0ms                            | 5.77±0.05ms                      |    1.16 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 1, 1, 'd') |
| +        | 5.96±0.03ms                         | 6.90±0.2ms                       |    1.16 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 4, 2, 'd') |
| +        | 2.13±0ms                            | 2.45±0ms                         |    1.15 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 1, 1, 'f') |
| +        | 5.95±0.06ms                         | 6.75±0.03ms                      |    1.13 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 4, 1, 'd') |
| -        | 7.07±0.01ms                         | 6.31±0.02ms                      |    0.89 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 1, 2, 'd') |
| -        | 7.08±0ms                            | 5.86±0.03ms                      |    0.83 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 1, 1, 'd') |
| -        | 5.64±0.01ms                         | 4.47±0.3ms                       |    0.79 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 4, 2, 'd')        |
| -        | 5.76±0.01ms                         | 4.15±0.1ms                       |    0.72 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 4, 2, 'd')        |
| -        | 5.64±0.02ms                         | 3.94±0.06ms                      |    0.7  | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 4, 1, 'd')        |
| -        | 5.76±0.01ms                         | 4.01±0.1ms                       |    0.7  | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 4, 1, 'd')        |
| -        | 2.01±0.01ms                         | 1.36±0.02ms                      |    0.67 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 4, 1, 'f') |
| -        | 2.02±0.02ms                         | 1.35±0.03ms                      |    0.67 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 4, 2, 'f') |
| -        | 2.02±0ms                            | 1.30±0ms                         |    0.64 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 1, 2, 'f') |
| -        | 5.62±0.01ms                         | 3.54±0.08ms                      |    0.63 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 1, 2, 'd')        |
| -        | 5.73±0.01ms                         | 3.58±0.06ms                      |    0.62 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 1, 2, 'd')        |
| -        | 5.62±0.01ms                         | 3.11±0.07ms                      |    0.55 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 1, 1, 'd')        |
| -        | 5.72±0.01ms                         | 3.12±0.05ms                      |    0.54 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 1, 1, 'd')        |
| -        | 2.01±0ms                            | 926±3μs                          |    0.46 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 1, 1, 'f') |
| -        | 3.29±0ms                            | 1.45±0.1ms                       |    0.44 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 4, 2, 'f')        |
| -        | 3.31±0ms                            | 1.41±0.04ms                      |    0.42 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 4, 2, 'f')        |
| -        | 3.31±0ms                            | 1.36±0.01ms                      |    0.41 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 4, 1, 'f')        |
| -        | 3.28±0ms                            | 1.30±0ms                         |    0.4  | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 1, 2, 'f')        |
| -        | 3.29±0ms                            | 1.32±0.03ms                      |    0.4  | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 4, 1, 'f')        |
| -        | 3.30±0ms                            | 1.32±0.03ms                      |    0.4  | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 1, 2, 'f')        |
| -        | 3.30±0ms                            | 947±4μs                          |    0.29 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 1, 1, 'f')        |
| -        | 3.28±0ms                            | 925±2μs                          |    0.28 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 1, 1, 'f')        |

</details>

#### Binary size change: _multiarray_umath.cpython-312-x86_64-linux-gnu.so

|  stripped |  size (bytes) |
|------------|--------------|
| **before** | 9,895,936    |
| **after**  | 10,014,720   |
| **diff**   | **+118,784** |

### DIFF
--- a/numpy/_core/src/common/simd/simd.hpp
+++ b/numpy/_core/src/common/simd/simd.hpp
@@ -76,17 +76,25 @@ namespace sr = npsr::HWY_NAMESPACE;
 /// Default precision configrations for NumPy SIMD Routines
 // TODO: Allow user to configure the default precision at build time.
 namespace detail {
-using PreciseHigh = decltype(npsr::Precise{});
-using PreciseLow = decltype(npsr::Precise{npsr::kLowAccuracy});
+// `npsr::Precise` is a meta-RAII class responsible for managing the floating-point environment at runtime
+// and holding compile-time configurations for intrinsic floating-point behavior.
+// Default configurations are set to the highest IEEE compliance intent.
+// For more information, see:
+// https://github.com/numpy/numpy-simd-routines/blob/1acd14571b9f58072d1f7f17886945000d78b56a/npsr/precise.h#L93
+// However, when float64 SIMD is not available or disabled, we define a dummy type to avoid
+// clearing floating-point exceptions within its destructor.
 struct PresiceDummy {};
+
 template <typename T>
 struct PreciseByType {};
 template <>
-struct PreciseByType<float> { using Type = PreciseLow; };
+struct PreciseByType<float> {
+    using Type = decltype(npsr::Precise{npsr::kLowAccuracy});
+};
 template <>
 struct PreciseByType<double> {
 #if NPY_HWY_F64
-    using Type = PreciseHigh;
+    using Type = decltype(npsr::Precise{});
 #else
     // If float64 SIMD isn’t available, use a dummy type.
     // The scalar path will run, but `Type` must still be defined.
@@ -96,9 +104,12 @@ struct PreciseByType<double> {
 #endif
 };
 } // namespace detail
+
+/// `npsr::Precise<...>` for the given type `T`.
+/// If `T` is `float`, it uses the low-accuracy configuration by default.
+/// If `T` is `double`, it uses the high-accuracy configuration by default(if float64 SIMD is available).
 template <typename T>
 using Precise = typename detail::PreciseByType<T>::Type;
-
 #endif
 #include "simd.inc.hpp"
 }  // namespace simd
--- a/numpy/_core/src/common/simd/simd.hpp
+++ b/numpy/_core/src/common/simd/simd.hpp
@@ -1,6 +1,5 @@
 #ifndef NUMPY__CORE_SRC_COMMON_SIMD_SIMD_HPP_
 #define NUMPY__CORE_SRC_COMMON_SIMD_SIMD_HPP_
-
 /**
  * This header provides a thin wrapper over Google's Highway SIMD library.
  *
@@ -19,7 +18,9 @@
  */
 #ifndef NPY_DISABLE_OPTIMIZATION
 #include <hwy/highway.h>
-
+#include <npsr/npsr.h>
+#include <type_traits>
+#include <limits>
 /**
  * We avoid using Highway scalar operations for the following reasons:
  *
@@ -67,6 +68,37 @@ namespace hn = hwy::HWY_NAMESPACE;
 // internally used by the template header
 template <typename TLane>
 using _Tag = hn::ScalableTag<TLane>;
+
+/// NumPy SIMD Routines namespace alias
+/// npsr is tag free by design so we only include it within main namespace (np::simd)
+namespace sr = npsr::HWY_NAMESPACE;
+
+/// Default precision configrations for NumPy SIMD Routines
+// TODO: Allow user to configure the default precision at build time.
+namespace detail {
+using PreciseHigh = decltype(npsr::Precise{});
+using PreciseLow = decltype(npsr::Precise{npsr::kLowAccuracy});
+struct PresiceDummy {};
+template <typename T>
+struct PreciseByType {};
+template <>
+struct PreciseByType<float> { using Type = PreciseLow; };
+template <>
+struct PreciseByType<double> {
+#if NPY_HWY_F64
+    using Type = PreciseHigh;
+#else
+    // If float64 SIMD isn’t available, use a dummy type.
+    // The scalar path will run, but `Type` must still be defined.
+    // The dummy is never passed; it only satisfies interfaces.
+    // This also avoids spurious FP exceptions during RAII.
+    using Type = PresiceDummy;
+#endif
+};
+} // namespace detail
+template <typename T>
+using Precise = typename detail::PreciseByType<T>::Type;
+
 #endif
 #include "simd.inc.hpp"
 }  // namespace simd
--- a/numpy/_core/src/umath/loops_trigonometric.dispatch.cpp
+++ b/numpy/_core/src/umath/loops_trigonometric.dispatch.cpp
@@ -1,299 +1,119 @@
-#include "fast_loop_macros.h"
-#include "loops.h"
-#include "loops_utils.h"
+#include "numpy/npy_common.h"  // npy_intp
+#include "loops.h"        // forward declarations
+#include "loops_utils.h"  // is_mem_overlap
 
-#include "simd/simd.h"
-#include "simd/simd.hpp" 
-#include <hwy/highway.h>
+#include "simd/simd.hpp"  // Highway & NumPy SIMD Routines
+#include <type_traits>    // std::enable_if_t
+#include <algorithm>      // std::min
+#include <cmath>          // std::sin, std::cos
 
-namespace hn = hwy::HWY_NAMESPACE;
+// annonymous namespace for ODR since this source compiled multiple times
+// based on the dispatch targets
+namespace {
+using namespace np::simd;
 
-/*
- * Vectorized approximate sine/cosine algorithms: The following code is a
- * vectorized version of the algorithm presented here:
- * https://stackoverflow.com/questions/30463616/payne-hanek-algorithm-implementation-in-c/30465751#30465751
- * (1) Load data in registers and generate mask for elements that are within
- * range [-71476.0625f, 71476.0625f] for cosine and [-117435.992f, 117435.992f]
- * for sine.
- * (2) For elements within range, perform range reduction using
- * Cody-Waite's method: x* = x - y*PI/2, where y = rint(x*2/PI). x* \in [-PI/4,
- * PI/4].
- * (3) Map cos(x) to (+/-)sine or (+/-)cosine of x* based on the
- * quadrant k = int(y).
- * (4) For elements outside that range, Cody-Waite
- * reduction performs poorly leading to catastrophic cancellation. We compute
- * cosine by calling glibc in a scalar fashion.
- * (5) Vectorized implementation
- * has a max ULP of 1.49 and performs at least 5-7x(x86) - 2.5-3x(Power) -
- * 1-2x(Arm) faster than scalar implementations when magnitude of all elements
- * in the array < 71476.0625f (117435.992f for sine).  Worst case performance
- * is when all the elements are large leading to about 1-2% reduction in
- * performance.
- * TODO: use vectorized version of Payne-Hanek style reduction for large
- * elements or when there's no native FUSED support instead of fallback to libc
- */
-
-#if NPY_SIMD_FMA3  // native support
-typedef enum {
-    SIMD_COMPUTE_SIN,
-    SIMD_COMPUTE_COS
-} SIMD_TRIG_OP;
-
-const hn::ScalableTag<float> f32;
-const hn::ScalableTag<int32_t> s32;
-using vec_f32 = hn::Vec<decltype(f32)>;
-using vec_s32 = hn::Vec<decltype(s32)>;
-using opmask_t = hn::Mask<decltype(f32)>;
-
-HWY_INLINE HWY_ATTR vec_f32
-simd_range_reduction_f32(vec_f32 &x, vec_f32 &y, const vec_f32 &c1,
-                         const vec_f32 &c2, const vec_f32 &c3)
-{
-    vec_f32 reduced_x = hn::MulAdd(y, c1, x);
-    reduced_x = hn::MulAdd(y, c2, reduced_x);
-    reduced_x = hn::MulAdd(y, c3, reduced_x);
-    return reduced_x;
-}
-
-HWY_INLINE HWY_ATTR vec_f32
-simd_cosine_poly_f32(vec_f32 &x2)
-{
-    const vec_f32 invf8 = hn::Set(f32, 0x1.98e616p-16f);
-    const vec_f32 invf6 = hn::Set(f32, -0x1.6c06dcp-10f);
-    const vec_f32 invf4 = hn::Set(f32, 0x1.55553cp-05f);
-    const vec_f32 invf2 = hn::Set(f32, -0x1.000000p-01f);
-    const vec_f32 invf0 = hn::Set(f32, 0x1.000000p+00f);
-
-    vec_f32 r = hn::MulAdd(invf8, x2, invf6);
-    r = hn::MulAdd(r, x2, invf4);
-    r = hn::MulAdd(r, x2, invf2);
-    r = hn::MulAdd(r, x2, invf0);
-    return r;
-}
-/*
- * Approximate sine algorithm for x \in [-PI/4, PI/4]
- * Maximum ULP across all 32-bit floats = 0.647
- * Polynomial approximation based on unpublished work by T. Myklebust
- */
-HWY_INLINE HWY_ATTR vec_f32
-simd_sine_poly_f32(vec_f32 &x, vec_f32 &x2)
-{
-    const vec_f32 invf9 = hn::Set(f32, 0x1.7d3bbcp-19f);
-    const vec_f32 invf7 = hn::Set(f32, -0x1.a06bbap-13f);
-    const vec_f32 invf5 = hn::Set(f32, 0x1.11119ap-07f);
-    const vec_f32 invf3 = hn::Set(f32, -0x1.555556p-03f);
-
-    vec_f32 r = hn::MulAdd(invf9, x2, invf7);
-    r = hn::MulAdd(r, x2, invf5);
-    r = hn::MulAdd(r, x2, invf3);
-    r = hn::MulAdd(r, x2, hn::Zero(f32));
-    r = hn::MulAdd(r, x, x);
-    return r;
-}
-
-static void HWY_ATTR SIMD_MSVC_NOINLINE
-simd_sincos_f32(const float *src, npy_intp ssrc, float *dst, npy_intp sdst,
-                npy_intp len, SIMD_TRIG_OP trig_op)
-{
-    // Load up frequently used constants
-    const vec_f32 zerosf = hn::Zero(f32);
-    const vec_s32 ones = hn::Set(s32, 1);
-    const vec_s32 twos = hn::Set(s32, 2);
-    const vec_f32 two_over_pi = hn::Set(f32, 0x1.45f306p-1f);
-    const vec_f32 codyw_pio2_highf = hn::Set(f32, -0x1.921fb0p+00f);
-    const vec_f32 codyw_pio2_medf = hn::Set(f32, -0x1.5110b4p-22f);
-    const vec_f32 codyw_pio2_lowf = hn::Set(f32, -0x1.846988p-48f);
-    const vec_f32 rint_cvt_magic = hn::Set(f32, 0x1.800000p+23f);
-    // Cody-Waite's range
-    float max_codi = 117435.992f;
-    if (trig_op == SIMD_COMPUTE_COS) {
-        max_codi = 71476.0625f;
-    }
-    const vec_f32 max_cody = hn::Set(f32, max_codi);
-
-    const int lanes = hn::Lanes(f32);
-    const vec_s32 src_index = hn::Mul(hn::Iota(s32, 0), hn::Set(s32, ssrc));
-    const vec_s32 dst_index = hn::Mul(hn::Iota(s32, 0), hn::Set(s32, sdst));
-
-    for (; len > 0; len -= lanes, src += ssrc * lanes, dst += sdst * lanes) {
-        vec_f32 x_in;
-        if (ssrc == 1) {
-            x_in = hn::LoadN(f32, src, len);
-        }
-        else {
-            x_in = hn::GatherIndexN(f32, src, src_index, len);
-        }
-        opmask_t nnan_mask = hn::Not(hn::IsNaN(x_in));
-        // Eliminate NaN to avoid FP invalid exception
-        x_in = hn::IfThenElse(nnan_mask, x_in, zerosf);
-        opmask_t simd_mask = hn::Le(hn::Abs(x_in), max_cody);
-        /*
-         * For elements outside of this range, Cody-Waite's range reduction
-         * becomes inaccurate and we will call libc to compute cosine for
-         * these numbers
-         */
-        if (!hn::AllFalse(f32, simd_mask)) {
-            vec_f32 x = hn::IfThenElse(hn::And(nnan_mask, simd_mask), x_in,
-                                       zerosf);
-
-            vec_f32 quadrant = hn::Mul(x, two_over_pi);
-            // round to nearest, -0.0f -> +0.0f, and |a| must be <= 0x1.0p+22
-            quadrant = hn::Add(quadrant, rint_cvt_magic);
-            quadrant = hn::Sub(quadrant, rint_cvt_magic);
-
-            // Cody-Waite's range reduction algorithm
-            vec_f32 reduced_x =
-                    simd_range_reduction_f32(x, quadrant, codyw_pio2_highf,
-                                             codyw_pio2_medf, codyw_pio2_lowf);
-            vec_f32 reduced_x2 = hn::Mul(reduced_x, reduced_x);
-
-            // compute cosine and sine
-            vec_f32 cos = simd_cosine_poly_f32(reduced_x2);
-            vec_f32 sin = simd_sine_poly_f32(reduced_x, reduced_x2);
-
-            vec_s32 iquadrant = hn::NearestInt(quadrant);
-            if (trig_op == SIMD_COMPUTE_COS) {
-                iquadrant = hn::Add(iquadrant, ones);
-            }
-            // blend sin and cos based on the quadrant
-            opmask_t sine_mask = hn::RebindMask(
-                    f32, hn::Eq(hn::And(iquadrant, ones), hn::Zero(s32)));
-            cos = hn::IfThenElse(sine_mask, sin, cos);
-
-            // multiply by -1 for appropriate elements
-            opmask_t negate_mask = hn::RebindMask(
-                    f32, hn::Eq(hn::And(iquadrant, twos), twos));
-            cos = hn::MaskedSubOr(cos, negate_mask, zerosf, cos);
-            cos = hn::IfThenElse(nnan_mask, cos, hn::Set(f32, NPY_NANF));
-
-            if (sdst == 1) {
-                hn::StoreN(cos, f32, dst, len);
-            }
-            else {
-                hn::ScatterIndexN(cos, f32, dst, dst_index, len);
-            }
-        }
-        if (!hn::AllTrue(f32, simd_mask)) {
-            static_assert(hn::MaxLanes(f32) <= 64,
-                          "The following fallback is not applicable for "
-                          "SIMD widths larger than 2048 bits, or for scalable "
-                          "SIMD in general.");
-            npy_uint64 simd_maski;
-            hn::StoreMaskBits(f32, simd_mask, (uint8_t *)&simd_maski);
-#if HWY_IS_BIG_ENDIAN
-            static_assert(hn::MaxLanes(f32) <= 8,
-                          "This conversion is not supported for SIMD widths "
-                          "larger than 256 bits.");
-            simd_maski = ((uint8_t *)&simd_maski)[0];
+template <typename T>
+struct OPsin {
+    using LaneType = T;
+    NPY_FINLINE T operator()(T x) const { return std::sin(x); }
+#if NPY_HWY
+    template <class X = T, typename = std::enable_if_t<kSupportLane<X>>>
+    NPY_FINLINE Vec<T> operator()(const Vec<T>& x) {
+      return sr::Sin(prec, x);
+    }  
+    Precise<T> prec;
 #endif
-            float NPY_DECL_ALIGNED(NPY_SIMD_WIDTH) ip_fback[hn::MaxLanes(f32)];
-            hn::Store(x_in, f32, ip_fback);
+};
 
-            // process elements using libc for large elements
-            if (trig_op == SIMD_COMPUTE_COS) {
-                for (unsigned i = 0; i < hn::Lanes(f32); ++i) {
-                    if ((simd_maski >> i) & 1) {
-                        continue;
-                    }
-                    dst[sdst * i] = npy_cosf(ip_fback[i]);
-                }
-            }
-            else {
-                for (unsigned i = 0; i < hn::Lanes(f32); ++i) {
-                    if ((simd_maski >> i) & 1) {
-                        continue;
-                    }
-                    dst[sdst * i] = npy_sinf(ip_fback[i]);
-                }
-            }
+template <typename T>
+struct OPcos {
+    using LaneType = T;
+    NPY_FINLINE T operator()(T x) const { return std::cos(x); }
+#if NPY_HWY
+    template <class X = T, typename = std::enable_if_t<kSupportLane<X>>>
+    NPY_FINLINE Vec<T> operator()(const Vec<T>& x) {
+      return sr::Cos(prec, x);
+    }  
+    Precise<T> prec;
+#endif
+};
+
+template <typename OP, typename T = typename OP::LaneType>
+NPY_NOINLINE void
+UnaryContig(OP &op, const T *src, T *dst, npy_intp len)
+{
+#if NPY_HWY
+    if constexpr (kSupportLane<T>) {
+        const npy_intp nlanes = Lanes<T>();
+        for (; len > 0; len -= nlanes, src += nlanes, dst += nlanes) {
+            Vec<T> x = LoadN(src, len);
+            Vec<T> r = op(x);
+            StoreN(r, dst, len);
         }
-        npyv_cleanup();
+    }
+#else
+    if constexpr (0) {
+    }
+#endif
+    else {
+        for (const T *end = src + len; src < end; ++src, ++dst) {
+            *dst = op(*src);
+        }
+    }
+};
+
+template <typename OP, typename T = typename OP::LaneType>
+NPY_FINLINE void
+UnaryOP(OP &op, char **args, npy_intp const *dimensions, npy_intp const *steps)
+{
+    npy_intp len = dimensions[0];
+    npy_intp sin = steps[0], sout = steps[1];
+    char *in = args[0], *out = args[1];
+    // We are dealing with large math kernels allowing inlining
+    // or specialized implementations for non-contiguous, overlapping cases, or even
+    // providing a separate iteration for contiguous tailing would easily bloat the
+    // binary size.
+    // Especially if we may need to support runtime dispatching for precision control.
+    // If you really think that providing such specialization for these kinds of
+    // kernels will increase performance, please make sure to benchmark it first and
+    // compare the binary size before and after the change.
+    // Note: the intrinsics of NumPy SIMD routines are inlined by default.
+    if (NPY_UNLIKELY(is_mem_overlap(args[0], sin, args[1], sout, len) ||
+                     sin != sizeof(T) || sout != sizeof(T))) {
+        // this for non-contiguous or overlapping case
+        constexpr npy_intp kBufferLen = 4096 / sizeof(T);
+#if NPY_HWY
+        HWY_ALIGN
+#endif
+        T buffer[kBufferLen];
+        for (npy_intp processed = 0; processed < len;) {
+            npy_intp chunk = std::min(kBufferLen, len - processed);
+            for (npy_intp i = 0; i < chunk; ++i, in += sin) {
+                buffer[i] = *reinterpret_cast<const T *>(in);
+            }
+            UnaryContig(op, buffer, buffer, chunk);
+            for (npy_intp i = 0; i < chunk; ++i, out += sout) {
+                *reinterpret_cast<T *>(out) = buffer[i];
+            }
+            processed += chunk;
+        }
+    }
+    else {
+        UnaryContig(op, (T *)in, (T *)out, len);
     }
 }
-#endif  // NPY_SIMD_FMA3
+}  // namespace
 
-/* Disable SIMD code sin/cos f64 and revert to libm: see
- * https://mail.python.org/archives/list/numpy-discussion@python.org/thread/C6EYZZSR4EWGVKHAZXLE7IBILRMNVK7L/
- * for detailed discussion on this*/
-#define DISPATCH_DOUBLE_FUNC(func)                                          \
-    NPY_NO_EXPORT void NPY_CPU_DISPATCH_CURFX(DOUBLE_##func)(               \
+#define UMATH_UNARY_SR_IMPL(TYPE, type, op)                                 \
+    NPY_NO_EXPORT void NPY_CPU_DISPATCH_CURFX(TYPE##_##op)(                 \
             char **args, npy_intp const *dimensions, npy_intp const *steps, \
             void *NPY_UNUSED(data))                                         \
     {                                                                       \
-        UNARY_LOOP                                                          \
-        {                                                                   \
-            const npy_double in1 = *(npy_double *)ip1;                      \
-            *(npy_double *)op1 = npy_##func(in1);                           \
-        }                                                                   \
+        OP##op<type> math_op;                                               \
+        UnaryOP(math_op, args, dimensions, steps);                          \
     }
 
-DISPATCH_DOUBLE_FUNC(sin)
-DISPATCH_DOUBLE_FUNC(cos)
-
-NPY_NO_EXPORT void
-NPY_CPU_DISPATCH_CURFX(FLOAT_sin)(char **args, npy_intp const *dimensions,
-                                  npy_intp const *steps,
-                                  void *NPY_UNUSED(data))
-{
-#if NPY_SIMD_FMA3
-    npy_intp len = dimensions[0];
-
-    if (is_mem_overlap(args[0], steps[0], args[1], steps[1], len) ||
-        !npyv_loadable_stride_f32(steps[0]) ||
-        !npyv_storable_stride_f32(steps[1])) {
-        UNARY_LOOP
-        {
-            simd_sincos_f32((npy_float *)ip1, 1, (npy_float *)op1, 1, 1,
-                            SIMD_COMPUTE_SIN);
-        }
-    }
-    else {
-        const npy_float *src = (npy_float *)args[0];
-        npy_float *dst = (npy_float *)args[1];
-        const npy_intp ssrc = steps[0] / sizeof(npy_float);
-        const npy_intp sdst = steps[1] / sizeof(npy_float);
-
-        simd_sincos_f32(src, ssrc, dst, sdst, len, SIMD_COMPUTE_SIN);
-    }
-#else
-    UNARY_LOOP
-    {
-        const npy_float in1 = *(npy_float *)ip1;
-        *(npy_float *)op1 = npy_sinf(in1);
-    }
-#endif
-}
-
-NPY_NO_EXPORT void
-NPY_CPU_DISPATCH_CURFX(FLOAT_cos)(char **args, npy_intp const *dimensions,
-                                  npy_intp const *steps,
-                                  void *NPY_UNUSED(data))
-{
-#if NPY_SIMD_FMA3
-    npy_intp len = dimensions[0];
-
-    if (is_mem_overlap(args[0], steps[0], args[1], steps[1], len) ||
-        !npyv_loadable_stride_f32(steps[0]) ||
-        !npyv_storable_stride_f32(steps[1])) {
-        UNARY_LOOP
-        {
-            simd_sincos_f32((npy_float *)ip1, 1, (npy_float *)op1, 1, 1,
-                            SIMD_COMPUTE_COS);
-        }
-    }
-    else {
-        const npy_float *src = (npy_float *)args[0];
-        npy_float *dst = (npy_float *)args[1];
-        const npy_intp ssrc = steps[0] / sizeof(npy_float);
-        const npy_intp sdst = steps[1] / sizeof(npy_float);
-
-        simd_sincos_f32(src, ssrc, dst, sdst, len, SIMD_COMPUTE_COS);
-    }
-#else
-    UNARY_LOOP
-    {
-        const npy_float in1 = *(npy_float *)ip1;
-        *(npy_float *)op1 = npy_cosf(in1);
-    }
-#endif
-}
+UMATH_UNARY_SR_IMPL(FLOAT, float, sin)
+UMATH_UNARY_SR_IMPL(DOUBLE, double, sin)
+UMATH_UNARY_SR_IMPL(FLOAT, float, cos)
+UMATH_UNARY_SR_IMPL(DOUBLE, double, cos)
--- a/numpy/_core/tests/test_umath.py
+++ b/numpy/_core/tests/test_umath.py
@@ -1649,6 +1649,54 @@ class TestSpecialFloats:
         callable(x[::stride], out=x[:M])
         assert_equal(x[:M], y)
 
+    @pytest.mark.parametrize("func", [np.sin, np.cos])
+    @pytest.mark.parametrize("dtype", ["f", "d"])
+    def test_trig_facts(self, dtype, func):
+        atol, max_ulp = (1e-6, 3) if dtype == "f" else (1e-15, 1)
+        test_data = {
+            np.sin: [
+                (0, 0),
+                (np.pi, 0),
+                (-np.pi, 0),
+                (2 * np.pi, 0),
+                (-2 * np.pi, 0),
+                (np.pi / 2, 1),
+                (-np.pi / 2, -1),
+                (3 * np.pi / 2, -1),
+                (-3 * np.pi / 2, 1),
+                (np.pi / 6, 0.5),
+                (5 * np.pi / 6, 0.5),
+                (-np.pi / 6, -0.5),
+                (-5 * np.pi / 6, -0.5),
+            ],
+            np.cos: [
+                (0, 1),
+                (np.pi, -1),
+                (-np.pi, -1),
+                (2 * np.pi, 1),
+                (-2 * np.pi, 1),
+                (np.pi / 2, 0),
+                (-np.pi / 2, 0),
+                (3 * np.pi / 2, 0),
+                (-3 * np.pi / 2, 0),
+                (np.pi / 3, 0.5),
+                (5 * np.pi / 3, 0.5),
+                (-np.pi / 3, 0.5),
+                (-5 * np.pi / 3, 0.5),
+            ],
+        }[func]
+        for s_x, s_expected in test_data:
+            x = np.array([s_x] * (128 + 1), dtype=dtype)
+            result = func(x)
+            # Check that all results are identical, to catch issues with
+            # vectorization
+            assert np.all(result == result[0])
+            expected = np.array([s_expected] * (128 + 1), dtype=dtype)
+            if s_expected == 0:
+                assert_allclose(result, expected, atol=atol, rtol=0)
+            else:
+                assert_array_max_ulp(result, expected, maxulp=max_ulp)
+
     @pytest.mark.parametrize('dt', ['e', 'f', 'd', 'g'])
     def test_sqrt_values(self, dt):
         with np.errstate(all='ignore'):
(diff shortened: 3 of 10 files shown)


seiko2plus:
I have updated the benchmark to use GCC tests instead of Clang, as GCC is commonly the default compiler for wheels on Linux. At the moment, I only have access to bare-metal x86 hardware; Perhaps I should try cloud services—e.g., AWS—for testing on other architectures. The ufunc inner implementation is optimized for size as much as possible and only increases `+118,784B` (stripped) on x86/gcc.

cc: @charris, @rgommers, @mattip, @seberg, @r-devulap, @Mousius

seiko2plus:
Ah, regarding the reverted #23399, as discussed in the [thread/C6EYZZSR4EWGVKHAZXLE7IBILRMNVK7L/](https://mail.python.org/archives/list/numpy-discussion@python.org/thread/C6EYZZSR4EWGVKHAZXLE7IBILRMNVK7L/). The default precision for f64 is set up to 1ULP, including for large arguments, with the ability to add a build-time option to change this behavior later:

https://github.com/numpy/numpy/blob/67d02748ce5421bb10400b0d6334295069cf7b56/numpy/_core/src/common/simd/simd.hpp#L79-L100

cc: @mhvk

rgommers:
> At the moment, I only have access to bare-metal x86 hardware; Perhaps I should try cloud services—e.g., AWS—for testing on other architectures. 

No more GCC compile farm access? AWS seems okay of course. If you want me to run this on macOS arm64 (M1), I can.

> The ufunc inner implementation is optimized for size as much as possible and only increases +118,784B (stripped) on x86/gcc.

That's impressively small, glad to see that.

seberg:
@seiko2plus thanks a lot for this hard work!  I don't want to discuss it on the PR here, but maybe you can send a very brief mail summarizing precision changes for the float32 version? (I am slightly worried might get into the same dance as with the 64bit version and SVML before, so should at least increase awareness.)

numpy/_core/src/umath/loops_trigonometric.dispatch.cpp
```diff
@@ -1,299 +1,119 @@
...
+    // providing a separate iteration for contiguous tailing would easily bloat the
+    // binary size.
+    // Especially if we may need to support runtime dispatching for precision control.
+    // If you really think that providing such specialization for these kinds of
+    // kernels will increase performance, please make sure to benchmark it first and
+    // compare the binary size before and after the change.
+    // Note: the intrinsics of NumPy SIMD routines are inlined by default.
+    if (NPY_UNLIKELY(is_mem_overlap(args[0], sin, args[1], sout, len) ||
+                     sin != sizeof(T) || sout != sizeof(T))) {
+        // this for non-contiguous or overlapping case
```
mhvk, line 83: I always have trouble knowing exactly what one can ask the iterator, but is it not possible to set some flag that ensures that for non-contiguous or overlapping data, a copy is made to a contiguous buffer? I ask since if so we do not have to do this inside the loop (with similar duplicated code presumably elsewhere).

numpy/_core/src/common/simd/simd.hpp
```diff
@@ -67,6 +68,37 @@ namespace hn = hwy::HWY_NAMESPACE;
 // internaly used by the template header
 template <typename TLane>
 using _Tag = hn::ScalableTag<TLane>;
+
+/// NumPy SIMD Routines namespace alias
+/// npsr is tag free by design so we only include it within main namespace (np::simd)
+namespace sr = npsr::HWY_NAMESPACE;
+
+/// Default precision configrations for NumPy SIMD Routines
```
mhvk, line 76: Might be good here to tell what low and high imply (or at least where to look it up)

numpy/_core/tests/test_umath.py
```diff
@@ -2097,12 +2097,15 @@ def test_sincos_float32(self):
...
+        # been set in the precision mode
+        maxulp_f32 = 2 if _ufunc_has_fma(np.sin, np.float32) else 3
+        assert_array_max_ulp(np.sin(x_f32), np.float32(np.sin(x_f64)), maxulp=maxulp_f32)
+        assert_array_max_ulp(np.cos(x_f32), np.float32(np.cos(x_f64)), maxulp=maxulp_f32)
         # test aliasing(issue #17761)
         tx_f32 = x_f32.copy()
-        assert_array_max_ulp(np.sin(x_f32, out=x_f32), np.float32(np.sin(x_f64)), maxulp=2)
-        assert_array_max_ulp(np.cos(tx_f32, out=tx_f32), np.float32(np.cos(x_f64)), maxulp=2)
+        assert_array_max_ulp(np.sin(x_f32, out=x_f32), np.float32(np.sin(x_f64)), maxulp=maxulp_f32)
+        assert_array_max_ulp(np.cos(tx_f32, out=tx_f32), np.float32(np.cos(x_f64)), maxulp=maxulp_f32)
```
mhvk, line 2279: It maybe good to add an explicit test that `cos(0)==1` for all float, as it is the thing that threw our tests off in a previous iteration (and I think it is something you now explicitly capture).

mhvk (commented):
@seiko2plus - looks very impressive! A few small comments inline, with that about the iterator perhaps the most important (if it is possible; can be for follow-up). I also made a small comment over at your new npsr routines.

numpy/_core/src/umath/loops_trigonometric.dispatch.cpp
seberg: This is already the case I believe. But these paths get used by `.accumulate` unfortunately. Although, it may be that we can simplify the check based on that knowledge?

EDIt: Sorry to be clear, I _think_ that is the case, I am not 100% sure.

mhvk: I don't think it can be used by `.accumulate` since this ufunc has `nin=1`... If a flag is indeed passed, maybe it is worth replacing this with an assert on the stride and see if anything breaks? Logically, if we are passing the iterator a flag that we require contiguous input and output, then we should never get anything else...

seberg: Ah, there is no flag to enforce a contiguous loop, but we could probably add one, if we need this more I think that makes sense. The buffer machinery will have a lot of overhead unfortunately, but overall it is still better likely.

For _overlap_, I believe the rule should be (or am I missing something?!):
* If memory overlap exists it must be exactly the same memory (i.e. args[0] == args[1] and steps identical).
* _Except_ for `.accumulate` which you are right, doesn't apply here.

mhvk: Ah, if there is no "needs contiguous" flag, it is out of scope here. But I think it does make sense to add the option to request buffering to a contiguous and aligned chunk to the iterator. I would think the overhead cannot be that much worse than it is here, since one would already be inside the iterator anyway typically for non-contiguous data. And it would certainly be nice to keep the underlying loops simple...

seberg: >  I would think the overhead cannot be that much worse than it is here

Well, the buffering has a "huge" amount of one time overhead. For very large arrays, yes the overhead may well be lower in practice (or at least nicer, as it's repeated fewer times).

mhvk: @seberg - you mentioned there is no flag, but is one needed? If `sin`/`cos` are defined as an array method, would things not work automatically if one filled only the `contiguous_loop` slot? Or is that at the moment only treated as something that will be used if present if data are contiguous.

mhvk: > > I would think the overhead cannot be that much worse than it is here
> 
> Well, the buffering has a "huge" amount of one time overhead. For very large arrays, yes the overhead may well be lower in practice (or at least nicer, as it's repeated fewer times).

The other thing is that if we centralize this, there will be more of an incentive to speed it up.  Isn't the buffer pre-allocated by the compiler here? It would make more sense to have just one scratch buffer in the iterator, perhaps using your mechanism that allocates if it is too small.

seberg: @mhvk no not yet, the contiguous loop is an optimization only, only with that flag or a "non-contig to contig wrapper" could it be more. Dunno what is better, centralizing or not, I mostly like it because it removes unnecessary repitition both in code and binary, whether it is slower or faster.
(but yeah, we shouldn't worry about that here.)

mhvk: So, basically it would be some way to tell the ufunc (in this case) that the `NPY_ITER_CONTIG` flag should be passed on to the iterator...  Anyway, let's take the discussion out of this PR -- see #30413.

Mousius:
Generally seems better overall, testing on some metal AWS instances:

<details>

<summary>c8g.metal</summary>

```
| Change   | Before [4900421c] <main>   | After [2f990392]    |   Ratio | Benchmark (Parameter)                                                             |
|----------|----------------------------|---------------------|---------|-----------------------------------------------------------------------------------|
| +        | 3.89±0.02ms                | 5.61±0.02ms         |    1.44 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 4, 2, 'd')           |
| +        | 3.91±0ms                   | 5.56±0.02ms         |    1.42 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 4, 2, 'd')           |
| +        | 3.86±0ms                   | 5.29±0.03ms         |    1.37 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 4, 1, 'd')           |
| +        | 3.88±0.01ms                | 5.24±0.01ms         |    1.35 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 4, 1, 'd')           |
| +        | 3.85±0.01ms                | 5.00±0.01ms         |    1.3  | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 1, 2, 'd')           |
| +        | 3.90±0ms                   | 4.96±0.01ms         |    1.27 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 1, 2, 'd')           |
| +        | 385±10μs                   | 430±8μs             |    1.12 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'conjugate'> (0), 1, 2, 'e')        |
| +        | 391±5μs                    | 431±10μs            |    1.1  | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'conjugate'> (0), 4, 2, 'e')        |
| +        | 389±7μs                    | 429±4μs             |    1.1  | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'conjugate'> (1), 4, 2, 'e') |
| +        | 392±7μs                    | 427±10μs            |    1.09 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'conjugate'> (1), 4, 2, 'e')        |
| +        | 384±10μs                   | 420±10μs            |    1.09 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'conjugate'> (1), 1, 2, 'e') |
| +        | 2.13±0.03ms                | 2.32±0.04ms         |    1.09 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'exp2'>, 1, 2, 'd')          |
| +        | 383±6μs                    | 412±8μs             |    1.08 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'conjugate'> (1), 4, 1, 'e') |
| +        | 375±6μs                    | 400±7μs             |    1.07 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'conjugate'> (1), 1, 1, 'e')        |
| +        | 1.19±0.02ms                | 1.26±0.03ms         |    1.06 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'reciprocal'>, 1, 2, 'd')           |
| +        | 2.07±0ms                   | 2.18±0.02ms         |    1.05 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'exp'>, 4, 2, 'f')           |
| +        | 421±5μs                    | 443±5μs             |    1.05 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'positive'>, 4, 1, 'e')      |
| -        | 5.95±0.01ms                | 5.58±0.02ms         |    0.94 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 4, 2, 'd')                  |
| -        | 50.1±0.4μs                 | 47.2±0.8μs          |    0.94 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'absolute'>, 1, 1, 'e')      |
| -        | 2.92±0.01ms                | 2.72±0.02ms         |    0.93 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'log'>, 1, 1, 'f')           |
| -        | 2.38±0ms                   | 2.17±0ms            |    0.91 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 1, 2, 'f')                  |
| -        | 2.37±0.01ms                | 2.15±0ms            |    0.91 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 1, 2, 'f')                  |
| -        | 2.37±0ms                   | 2.14±0ms            |    0.9  | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 4, 1, 'f')                  |
| -        | 2.34±0ms                   | 2.12±0ms            |    0.9  | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 4, 1, 'f')                  |
| -        | 5.87±0.01ms                | 5.24±0.01ms         |    0.89 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 4, 1, 'd')                  |
| -        | 1.61±0ms                   | 1.43±0ms            |    0.89 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'deg2rad'>, 1, 1, 'e')              |
| -        | 1.61±0ms                   | 1.43±0ms            |    0.89 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'deg2rad'>, 1, 2, 'e')              |
| -        | 1.61±0ms                   | 1.43±0ms            |    0.89 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'deg2rad'>, 4, 1, 'e')              |
| -        | 1.61±0ms                   | 1.44±0ms            |    0.89 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'deg2rad'>, 4, 2, 'e')              |
| -        | 1.61±0ms                   | 1.43±0ms            |    0.89 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'radians'>, 1, 1, 'e')              |
| -        | 1.61±0ms                   | 1.43±0ms            |    0.89 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'radians'>, 1, 2, 'e')              |
| -        | 1.61±0ms                   | 1.43±0ms            |    0.89 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'radians'>, 4, 1, 'e')              |
| -        | 1.61±0ms                   | 1.44±0ms            |    0.89 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'radians'>, 4, 2, 'e')              |
| -        | 1.61±0ms                   | 1.43±0ms            |    0.89 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'deg2rad'>, 1, 1, 'e')       |
| -        | 1.61±0ms                   | 1.43±0ms            |    0.89 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'deg2rad'>, 1, 2, 'e')       |
| -        | 1.61±0ms                   | 1.43±0ms            |    0.89 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'deg2rad'>, 4, 1, 'e')       |
| -        | 1.61±0ms                   | 1.44±0ms            |    0.89 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'deg2rad'>, 4, 2, 'e')       |
| -        | 2.40±0.01ms                | 2.13±0.06ms         |    0.89 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'exp'>, 4, 1, 'd')           |
| -        | 1.61±0ms                   | 1.43±0ms            |    0.89 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'radians'>, 1, 1, 'e')       |
| -        | 1.61±0ms                   | 1.43±0ms            |    0.89 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'radians'>, 1, 2, 'e')       |
| -        | 1.61±0ms                   | 1.43±0ms            |    0.89 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'radians'>, 4, 1, 'e')       |
| -        | 1.61±0ms                   | 1.44±0ms            |    0.89 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'radians'>, 4, 2, 'e')       |
| -        | 5.86±0.04ms                | 4.96±0.01ms         |    0.85 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 1, 2, 'd')                  |
| -        | 7.08±0.03ms                | 5.61±0.02ms         |    0.79 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 4, 2, 'd')                  |
| -        | 2.77±0ms                   | 2.15±0ms            |    0.78 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 4, 2, 'f')                  |
| -        | 2.78±0ms                   | 2.13±0ms            |    0.77 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 4, 2, 'f')                  |
| -        | 7.04±0.02ms                | 5.28±0.01ms         |    0.75 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 4, 1, 'd')                  |
| -        | 7.01±0.02ms                | 4.99±0ms            |    0.71 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 1, 2, 'd')                  |
| -        | 3.04±0ms                   | 2.15±0ms            |    0.71 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 1, 2, 'f')           |
| -        | 3.00±0ms                   | 2.12±0ms            |    0.71 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 4, 1, 'f')           |
| -        | 3.09±0ms                   | 2.17±0ms            |    0.7  | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 1, 2, 'f')           |
| -        | 3.05±0ms                   | 2.14±0ms            |    0.7  | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 4, 1, 'f')           |
| -        | 5.84±0.01ms                | 3.96±0ms            |    0.68 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 1, 1, 'd')                  |
| -        | 3.49±0ms                   | 2.13±0.01ms         |    0.61 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 4, 2, 'f')           |
| -        | 3.54±0ms                   | 2.14±0ms            |    0.6  | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 4, 2, 'f')           |
| -        | 1.95±0ms                   | 1.16±0ms            |    0.59 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 1, 1, 'f')                  |
| -        | 6.99±0.02ms                | 3.99±0ms            |    0.57 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 1, 1, 'd')                  |
| -        | 2.11±0ms                   | 1.17±0ms            |    0.55 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 1, 1, 'f')                  |
| -        | 17.6±2ms                   | 9.11±1ms            |    0.52 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'tanh'>, 4, 2, 'd')                 |
| -        | 2.67±0ms                   | 1.17±0ms            |    0.44 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 1, 1, 'f')           |
| -        | 2.63±0ms                   | 1.16±0ms            |    0.44 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 1, 1, 'f')           |
```

</details>

<details>

<summary>c7g.metal</summary>

```| Change   | Before [4900421c] <main>   | After [2f990392]    |   Ratio | Benchmark (Parameter)                                                   |
|----------|----------------------------|---------------------|---------|-------------------------------------------------------------------------|
| +        | 4.27±0ms                   | 5.98±0.02ms         |    1.4  | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 4, 2, 'd') |
| +        | 4.59±0.01ms                | 6.05±0.02ms         |    1.32 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 4, 2, 'd') |
| +        | 4.26±0.01ms                | 5.27±0.01ms         |    1.24 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 4, 1, 'd') |
| +        | 4.30±0ms                   | 5.02±0.02ms         |    1.17 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 1, 2, 'd') |
| +        | 4.58±0ms                   | 5.36±0.01ms         |    1.17 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 4, 1, 'd') |
| +        | 4.60±0ms                   | 5.08±0ms            |    1.1  | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 1, 2, 'd') |
| -        | 4.30±0ms                   | 4.08±0ms            |    0.95 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 1, 1, 'd') |
| -        | 4.60±0.01ms                | 4.17±0ms            |    0.91 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 1, 1, 'd') |
| -        | 7.28±0.01ms                | 5.95±0.01ms         |    0.82 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 4, 2, 'd')        |
| -        | 7.21±0.02ms                | 5.27±0.01ms         |    0.73 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 4, 1, 'd')        |
| -        | 7.15±0.04ms                | 4.98±0ms            |    0.7  | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 1, 2, 'd')        |
| -        | 8.80±0.03ms                | 6.03±0.01ms         |    0.69 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 4, 2, 'd')        |
| -        | 8.62±0.03ms                | 5.36±0.01ms         |    0.62 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 4, 1, 'd')        |
| -        | 2.97±0ms                   | 1.83±0ms            |    0.62 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 4, 1, 'f')        |
| -        | 8.62±0.02ms                | 5.09±0.01ms         |    0.59 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 1, 2, 'd')        |
| -        | 3.12±0.01ms                | 1.82±0ms            |    0.58 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 1, 2, 'f')        |
| -        | 7.15±0.03ms                | 4.08±0ms            |    0.57 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 1, 1, 'd')        |
| -        | 3.20±0ms                   | 1.83±0ms            |    0.57 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 1, 2, 'f')        |
| -        | 3.20±0.01ms                | 1.84±0ms            |    0.57 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 4, 1, 'f')        |
| -        | 3.76±0ms                   | 1.85±0ms            |    0.49 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 4, 2, 'f')        |
| -        | 8.59±0.09ms                | 4.18±0ms            |    0.49 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 1, 1, 'd')        |
| -        | 3.74±0.01ms                | 1.84±0ms            |    0.49 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 4, 2, 'f')        |
| -        | 3.80±0ms                   | 1.82±0ms            |    0.48 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 1, 2, 'f') |
| -        | 3.84±0ms                   | 1.83±0ms            |    0.48 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 4, 1, 'f') |
| -        | 3.90±0ms                   | 1.84±0ms            |    0.47 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 1, 2, 'f') |
| -        | 3.96±0ms                   | 1.85±0.01ms         |    0.47 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 4, 1, 'f') |
| -        | 4.39±0ms                   | 1.84±0.01ms         |    0.42 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 4, 2, 'f') |
| -        | 2.55±0ms                   | 1.04±0ms            |    0.41 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sin'>, 1, 1, 'f')        |
| -        | 4.50±0.01ms                | 1.86±0.01ms         |    0.41 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 4, 2, 'f') |
| -        | 2.69±0.01ms                | 1.05±0.01ms         |    0.39 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 1, 1, 'f')        |
| -        | 3.26±0ms                   | 1.04±0.01ms         |    0.32 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sin'>, 1, 1, 'f') |
| -        | 3.33±0ms                   | 1.04±0.01ms         |    0.31 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 1, 1, 'f') |
```

</details>

Happy to merge as-is to help @seiko2plus keep moving 😸

mhvk:
Looks like the only slowdowns are `UnaryFPSpecial` for double -- is that because the data in that case include "weird" numbers (like denormal). If so, this indeed seems a great improvement.

seberg:
I don't really remember the state of this PR (it is tagged for 2.5).  Should we push for it?

One serious worry I have is still this part:
> ~4ulp for single-precision with handling floating-point errors,
which to me seems very optimistic that nobody cares (and we had discussed it also in some meeting to a similar result). Yes many (maybe most) float32 users may not care about precision, but that doesn't mean that no-one will be bitten by it unfortunately.
My opinion is that we really need a "fast-math" kind of switch for this type of thing, and otherwise stay roughly within e.g. glibc precision.

Mousius:
@seberg I think the *current* FP32 routines are also 4ULP, so this doesn't change anything and we can move forwards with it. I'm in agreement with you that we should give users the ability to opt-in/out of lower precision though.

numpy/_core/tests/test_umath.py
```diff
@@ -2202,19 +2251,30 @@ def test_sincos_float32(self):
...
+            np.float32(np.cos(x_f64)),
+            maxulp=maxulp_f32
+        )
         # test aliasing(issue #17761)
         tx_f32 = x_f32.copy()
         assert_array_max_ulp(
             np.sin(x_f32, out=x_f32),
             np.float32(np.sin(x_f64)),
-            maxulp=2,
+            maxulp=maxulp_f32,
```
seberg, line 2272: @Mousius if there is no big change then we don't have to worry about it, but this looks a bit like there may be one?  (But maybe not, fma not being available sounds possibly niche these days, but not sure.)

Mousius: My assumption with this is that the expectation was always 4ULP, but some of the routines would perform better than that.

seiko2plus:
My apologies for the delayed response.

@rgommers,

> No more GCC compile farm access? AWS seems okay of course. If you want me to run this on macOS arm64 (M1), I can.

My concerns were mainly for `SVE`, which is not supported by `M1`. I was expecting a performance gain similar to `AVX2`, but we need to dig more, I suppose. However, the current gain is acceptable. It seems AWS is the only option I have.

@seberg,

> thanks a lot for this hard work! I don't want to discuss it on the PR here, but maybe you can send a very brief mail summarizing precision changes for the float32 version? (I am slightly worried might get into the same dance as with the 64bit version and SVML before, so should at least increase awareness.)

I sent an email a long time ago:
https://mail.python.org/archives/list/numpy-discussion@python.org/thread/YIU5TZOENCRQ7GUZRKXYMUK2EIYWJINE/
And I have clarified the default precision of ~4ulp for `fp32` and ~1ulp for `fp64`.

One thing worth noting here: we already ship an enabled SIMD kernel for `float32`, and it is largely equivalent to the low-precision profile in this PR. The two differences are that the existing kernel falls back to `libm` for large inputs and lacks support for non-FMA SIMD extensions.


> I don't really remember the state  of this PR (it is tagged for 2.5). Should we push for it?

The build was totally broken due to the `meson` subproject functionality not working as it's supposed to due to a diverged rebase, which is resolved by numpy/meson#27, #31633.

New CI errors are caused by the LUT optimization patch numpy/numpy-simd-routines#7; (my bad) I will prepare a fix for it.


>  looks very impressive! A few small comments inline, with that about the iterator perhaps the most important (if it is possible; can be for follow-up). I also made a small comment over at your new npsr routines.

@mhvk,
Thank you, I have covered your suggestions in a separated commit:
 https://github.com/numpy/numpy/pull/29699/commits/c981beb27b26d51e913992edaeb235ac5285c447

@Mousius,
> Generally seems better overall, testing on some metal AWS instances:
> | -        | 5.84±0.01ms                | 3.96±0ms            |    0.68 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'cos'>, 1, 1, 'd')                  |

Not good to me or at least not as I was expecting. Just wondering, I will dig into it. We should have a similar gain to `X86_V3`, but definitely not within this PR.

@mhvk,
> Looks like the only slowdowns are UnaryFPSpecial for double -- is that because the data in that case include "weird" numbers (like denormal). If so, this indeed seems a great improvement.

`inf`/`nans` too, the performance gain is not good overall (compared to x86) even for the happy paths, but it's acceptable for now.

numpy/_core/tests/test_umath.py
seiko2plus: `maxulp_f32` is `~3` ulp for SIMD extensions without native FMA support, which I think is acceptable. If we wanted to tighten that, we could add an extra correction step — but I'd prefer to do that only after we have a proper test suite in place (we need a large dataset to validate edge cases reliably).

As a quick workaround, if the relaxed tolerance is really a concern, we can temporarily force the high-precision profile for those extensions until the correction step is implemented.

seiko2plus: Ah, also I suppose we agreed that the low-prec profile can even go further to `~4ulp`.

seberg: > Ah, also I suppose we agreed that the low-prec profile can even go further to ~4ulp.

This was the reason why I commented: I am not sure that this was ever quite agreed upon. And if it was, I suspect that this was before the recent fallout and discussion around lower fp16 precision.
https://mail.python.org/archives/list/numpy-discussion@python.org/thread/QLR57SMUPQVMGKXWQYR5CSE6JBPK43IG/#4GWQFDFYG7HCBHQEX5K4VP7DDGN25ZNZ

Having noticeably lower precision than typical math library/current code seems like something that we need to be very careful about to me.

numpy/_core/src/common/simd/simd.hpp
seiko2plus: done

numpy/_core/tests/test_umath.py
Mousius: Yeash, the conversation around making it all toggle-able was a long time ago. 

Just to clarify - do you want to block this PR @seberg? I don't see it as a change in behaviour.

seberg: I commented on the PR mainly based on the PR description.  We had serious issues with float64 being less precise, and whatever I discuss/think about it, I am not convinced we can just say that float32 you can cut corners without worry.  The same argument was made with float16 and we had to revert it.

So blocking this PR?  No, because you are both promising that there is in practice no precision change.  But blocking anything that makes precision worse and argues it is OK because 4 ULP are OK for float32: yes.

After the whole float16 thing *and* the float64 precision change issue and with people noticing different precision across architectures, I have serious concerns about saying that being a bit less precise than normal C-code/glibc code is OK.
(That isn't to say that we have to match it _exactly_, but I have always been nervous about going noticeably less precise and I feel like we have been bitten quite often when being optimistic about it.)

Or maybe in other words: I am not happy to keep being optimistic about this, if we really want a much faster but less precise version, I think we should go through the trouble of implementing the choice.

PR #29706, opened 2025-09-07
TYP: fix ``np.number`` and ``np.*integer`` method declaration

jorenham:
Follow-up of #29695, partially addressing https://github.com/python/mypy/pull/15993#issuecomment-3240220599

This only leaves `floating`, `timedelta64`, and `datetime64`.


jorenham:
The mypy_primer diffs might look problematic at first glance, but they're actually good news. 

In case of jax and pandas-stubs, it allows them to remove some `# type: ignore` comments, which suggests that this fixes something that used to be broken.

The ffreqtrade diff applies to a single line, and is caused by mypy now reporting a different error code (`operator` instead of `arg-type`).
It's similar for colour, where the error code changed from `call-overload` to `operator`, which affects two lines.
Changes in error codes typically don't affect CI, so they probably won't even notice it.

numpy/_typing/_callable.pyi
```diff
@@ -11,24 +11,12 @@ See the `Mypy documentation`_ on protocols for more details.
...
-    float64,
-    floating,
-    generic,
-    int_,
-    integer,
-    number,
-    signedinteger,
-    unsignedinteger,
-)
+from numpy import complex128, complexfloating, float64, floating, integer
```
charris, line 15: The integer types are moved?

jorenham: ruff removed those because they weren't being used anymore

charris:
Thanks Joren.

Issue #29144, opened 2025-06-08
ENH: Add reverse argument to ufunc.accumulate, cumsum, nancumsum, cumprod, nancumprod


carlosgmartin:
### Proposed new feature or change:

[Mailing list post](https://mail.python.org/archives/list/numpy-discussion@python.org/thread/7DWQVFLHLCBBY2YIP7HL64UTDKHAQPSN/)

Add a `reverse` argument to accumulating functions that, when true, causes the accumulation to be performed in reverse. Examples:

- [ufunc.accumulate](https://numpy.org/doc/stable/reference/generated/numpy.ufunc.accumulate.html)
- [cumsum](https://numpy.org/doc/stable/reference/generated/numpy.cumsum.html)
- [nancumsum](https://numpy.org/doc/stable/reference/generated/numpy.nancumsum.html)
- [cumprod](https://numpy.org/doc/stable/reference/generated/numpy.cumprod.html)
- [nancumprod](https://numpy.org/doc/stable/reference/generated/numpy.nancumprod.html)

Currently, this can be done via two [flips](https://numpy.org/doc/stable/reference/generated/numpy.flip.html), before and after the operation. But it might be more efficient to do it directly at the low level.

PR #29712, opened 2025-09-08
ENH: add reverse to cumsum/cumprod/nancumsum/nancumprod

viraatdas:
## Description

This PR implements a `reverse` keyword-only parameter for the following functions:

- `numpy.cumsum`
- `numpy.cumprod`
- `numpy.nancumsum`
- `numpy.nancumprod`

When `reverse=True`, accumulation proceeds right-to-left along the specified axis.  
It is equivalent to:

```python
np.flip(fn(np.flip(a, axis=axis), axis=axis), axis=axis)
```

but implemented using negative-stride views to avoid extra copies.
If `out` is provided, we accumulate into a flipped view of the output array.

### Rationale

- Addresses [gh-29144](https://github.com/numpy/numpy/issues/29144).
- A flip-based approach is efficient, requires no C-API changes, and preserves subclass behavior.
- Scope is limited to high-level APIs (`numpy.cumsum`, etc.) and does **not** modify `ndarray.cumsum`/`cumprod` or `ufunc.accumulate`.  
  If desired, `ufunc.accumulate(reverse=...)` can follow in a later PR.

---

### API Changes

- **New argument:** `reverse=False` (keyword-only).
- Supported for both regular and NaN-aware variants.
- Works with `axis=None`, `dtype=`, and `out=`.

---

### Examples

```python
>>> import numpy as np
>>> x = np.array([1, 2, 3, 4])
>>> np.cumsum(x)
array([ 1,  3,  6, 10])
>>> np.cumsum(x, reverse=True)
array([10,  9,  7,  4])

>>> np.cumprod(x, reverse=True)
array([24, 24, 12, 4])

>>> y = np.array([1.0, np.nan, 2.0])
>>> np.nancumsum(y, reverse=True)
array([3., 2., 2.])

>>> # Works with multi-dimensional arrays
>>> z = np.arange(6).reshape(2, 3)
>>> np.cumsum(z, axis=1, reverse=True)
array([[ 3,  3,  2],
       [12,  9,  5]])
```

## Tests
- Added new test files covering:
- 1D and multi-D arrays
- `axis=None` behavior
- `dtype` overrides
- `out` parameter
- NaN propagation for `nancumsum` and `nancumprod`
- Empty arrays

All new tests pass locally.

### DIFF
--- a/numpy/_core/fromnumeric.py
+++ b/numpy/_core/fromnumeric.py
@@ -2870,12 +2870,12 @@ def cumulative_sum(x, /, *, axis=None, dtype=None, out=None,
     return _cumulative_func(x, um.add, axis, dtype, out, include_initial)
 
 
-def _cumsum_dispatcher(a, axis=None, dtype=None, out=None):
+def _cumsum_dispatcher(a, axis=None, dtype=None, out=None, *, reverse=False):
     return (a, out)
 
 
 @array_function_dispatch(_cumsum_dispatcher)
-def cumsum(a, axis=None, dtype=None, out=None):
+def cumsum(a, axis=None, dtype=None, out=None, *, reverse=False):
     """
     Return the cumulative sum of the elements along a given axis.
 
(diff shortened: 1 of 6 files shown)


mattip:
The proposal had no discussion. Usually new kwargs or functionality is discussed in the issue and a concensus is reached that it is a good thing before submitting a PR.

seberg:
Thanks for the PR but I'll close it given the age, the fact that we haven't quite decided on it and that I think that if we do it, we should probably push it down into C since that is probably even the smaller change (no duplication in all of the Python entrypoints).

PR #29713, opened 2025-09-08
MAINT: update spin to 0.14 in requirements files

ngoldbaum:
I can't find a previous pull request or issue for this. Not sure if there's some reason why we're still pinned to 0.13. Let's see what the CI says.

### DIFF
--- a/requirements/build_requirements.txt
+++ b/requirements/build_requirements.txt
@@ -1,5 +1,5 @@
 meson-python>=0.13.1
 Cython>=3.0.6
 ninja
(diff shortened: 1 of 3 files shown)


rgommers (approved):
Thanks @ngoldbaum. No particular reason for the 0.13 pin AFAIK; we did need a pin earlier because `spin` wasn't super stable yet, but I think development has slowed down. If 0.15.0 is fully backwards compatible, we may consider changing the `==` to `>=`.

PR #29714, opened 2025-09-08
TST: update test_regression::test_gph25784

kkwli:
- restrict Windows only
- use an option that is supported by all compilers

Closes [#29681](https://github.com/numpy/numpy/issues/29681)


ngoldbaum:
Thanks. I'll merge this once CI finishes.

It would be neat to convert one of our existing CI jobs that uses gfortran to instead use flang. I think it's worth adding some CI coverage for flang now that it's apparently mature enough to be used with f2py!

PR #29715, opened 2025-09-08
BUG: Fix `dtype` refcount in `__array__`

### LINKED ISSUE #29707: BUG: Reference counting problem on dtypes with `__array__`
### Describe the issue:

The following code crashes during process shutdown with the error:
```
*** Reference count error detected: an attempt was made to deallocate the dtype 5 (i) ***
```

I am only able to reproduce this under Python 3.12 on Mac and Linux, not Python 3.13 or 3.14.

```python
import numpy as np

class MyArray:
  def __init__(self, val):
    self.val = val

  def __array__(self, dtype=None, copy=None):
    return self.val.__array__(dtype=dtype, copy=copy)

x = np.array(-1, dtype=np.int32)
x = MyArray(x)
np.array(x)
```

### Error message:

```shell
*** Reference count error detected: an attempt was made to deallocate the dtype 5 (i) ***
```

2.3.2
3.12.10 (main, Apr  9 2025, 04:03:51) [Clang 20.1.0 ]

### Runtime Environment:

[{'numpy_version': '2.3.2',
  'python': '3.12.10 (main, Apr  9 2025, 04:03:51) [Clang 20.1.0 ]',
  'uname': uname_result(system='Linux', node='redacted', release='6.12.32-redacted-amd64', version='#1 SMP PREEMPT_DYNAMIC Debian 6.12.32-redacted (2025-06-26)', machine='x86_64')},
                      'found': ['SSSE3',
                                'SSE41',
                                'POPCNT',
                                'SSE42',
                                'AVX',
                                'F16C',
                                'FMA3',
                                'AVX2'],
                      'not_found': ['AVX512F',
                                    'AVX512CD',
                                    'AVX512_KNL',
                                    'AVX512_KNM',
                                    'AVX512_SKX',
                                    'AVX512_CLX',
                                    'AVX512_CNL',
                                    'AVX512_ICL',
                                    'AVX512_SPR']}},
 {'architecture': 'Haswell',
  'filepath': '/.../home/phawkins/venv/py312/lib/python3.12/site-packages/numpy.libs/libscipy_openblas64_-8fb3d286.so',
  'internal_api': 'openblas',
  'num_threads': 64,
  'user_api': 'blas',
  'version': '0.3.30'}]

### Context for the issue:

Distilled out of a reference counting error found while working on JAX.

mtsokol:
This PR fixes a refcount bug raised in https://github.com/numpy/numpy/issues/29707 and adds a test which fails on `main` branch for linux, x86.

### DIFF
--- a/numpy/_core/src/multiarray/methods.c
+++ b/numpy/_core/src/multiarray/methods.c
@@ -980,7 +980,6 @@ array_getarray(PyArrayObject *self, PyObject *args, PyObject *kwds)
             return ret;
         } else { // copy == NPY_COPY_NEVER
             PyErr_SetString(PyExc_ValueError, npy_no_copy_err_msg);
-            // On error release a strong reference.
             Py_DECREF(newtype);
             Py_DECREF(self);
             return NULL;
--- a/numpy/_core/src/multiarray/methods.c
+++ b/numpy/_core/src/multiarray/methods.c
@@ -963,6 +963,7 @@ array_getarray(PyArrayObject *self, PyObject *args, PyObject *kwds)
         if (newtype == NULL) {
             newtype = PyArray_DESCR(self);
         }
+        Py_INCREF(newtype);
         ret = PyArray_CastToType(self, newtype, 0);
         Py_DECREF(self);
         return ret;
@@ -971,6 +972,7 @@ array_getarray(PyArrayObject *self, PyObject *args, PyObject *kwds)
             return (PyObject *)self;
         }
         if (copy == NPY_COPY_IF_NEEDED) {
+            Py_INCREF(newtype);
             ret = PyArray_CastToType(self, newtype, 0);
             Py_DECREF(self);
             return ret;
--- a/numpy/_core/src/multiarray/methods.c
+++ b/numpy/_core/src/multiarray/methods.c
@@ -962,17 +962,19 @@ array_getarray(PyArrayObject *self, PyObject *args, PyObject *kwds)
     if (copy == NPY_COPY_ALWAYS) {
         if (newtype == NULL) {
             newtype = PyArray_DESCR(self);
+            Py_INCREF(newtype);
         }
-        Py_INCREF(newtype);
         ret = PyArray_CastToType(self, newtype, 0);
         Py_DECREF(self);
         return ret;
     } else { // copy == NPY_COPY_IF_NEEDED || copy == NPY_COPY_NEVER
-        if (newtype == NULL || PyArray_EquivTypes(PyArray_DESCR(self), newtype)) {
+        if (newtype == NULL) {
+            return (PyObject *)self;
+        } else if (PyArray_EquivTypes(PyArray_DESCR(self), newtype)) {
+            Py_DECREF(newtype);
             return (PyObject *)self;
         }
         if (copy == NPY_COPY_IF_NEEDED) {
-            Py_INCREF(newtype);
             ret = PyArray_CastToType(self, newtype, 0);
             Py_DECREF(self);
             return ret;
--- a/numpy/_core/src/multiarray/methods.c
+++ b/numpy/_core/src/multiarray/methods.c
@@ -962,6 +962,8 @@ array_getarray(PyArrayObject *self, PyObject *args, PyObject *kwds)
     if (copy == NPY_COPY_ALWAYS) {
         if (newtype == NULL) {
             newtype = PyArray_DESCR(self);
+            // Take a new strong reference to compensate for
+            // PyArray_CastToType stealing a reference to newtype.
             Py_INCREF(newtype);
         }
         ret = PyArray_CastToType(self, newtype, 0);
@@ -969,6 +971,8 @@ array_getarray(PyArrayObject *self, PyObject *args, PyObject *kwds)
         return ret;
     } else { // copy == NPY_COPY_IF_NEEDED || copy == NPY_COPY_NEVER
         if (newtype == NULL || PyArray_EquivTypes(PyArray_DESCR(self), newtype)) {
+            // If newtype isn't NULL then release a strong reference
+            // introduced by PyArray_DescrConverter2 in arg parsing.
             Py_XDECREF(newtype);
             return (PyObject *)self;
         }
--- a/numpy/_core/src/multiarray/methods.c
+++ b/numpy/_core/src/multiarray/methods.c
@@ -968,10 +968,8 @@ array_getarray(PyArrayObject *self, PyObject *args, PyObject *kwds)
         Py_DECREF(self);
         return ret;
     } else { // copy == NPY_COPY_IF_NEEDED || copy == NPY_COPY_NEVER
-        if (newtype == NULL) {
-            return (PyObject *)self;
-        } else if (PyArray_EquivTypes(PyArray_DESCR(self), newtype)) {
-            Py_DECREF(newtype);
+        if (newtype == NULL || PyArray_EquivTypes(PyArray_DESCR(self), newtype)) {
+            Py_XDECREF(newtype);
             return (PyObject *)self;
         }
         if (copy == NPY_COPY_IF_NEEDED) {
--- a/numpy/_core/src/multiarray/methods.c
+++ b/numpy/_core/src/multiarray/methods.c
@@ -951,6 +951,7 @@ array_getarray(PyArrayObject *self, PyObject *args, PyObject *kwds)
                 (PyObject *)self
         );
         if (new == NULL) {
+            Py_XDECREF(newtype);
             return NULL;
         }
         self = new;
@@ -982,6 +983,9 @@ array_getarray(PyArrayObject *self, PyObject *args, PyObject *kwds)
             return ret;
         } else { // copy == NPY_COPY_NEVER
             PyErr_SetString(PyExc_ValueError, npy_no_copy_err_msg);
+            // On error release a strong reference introduced by
+            // PyArray_DescrConverter2 in arg parsing.
+            Py_DECREF(newtype);
             Py_DECREF(self);
             return NULL;
         }
--- a/numpy/_core/src/multiarray/methods.c
+++ b/numpy/_core/src/multiarray/methods.c
@@ -934,6 +934,11 @@ array_getarray(PyArrayObject *self, PyObject *args, PyObject *kwds)
         return NULL;
     }
 
+    if (newtype == NULL) {
+        newtype = PyArray_DESCR(self);
+        Py_INCREF(newtype);  // newtype is owned.
+    }
+
     /* convert to PyArray_Type */
     if (!PyArray_CheckExact(self)) {
         PyArrayObject *new;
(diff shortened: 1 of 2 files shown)


ngoldbaum (dismissed):
LGTM but I'll let someone else look this over before merging

numpy/_core/src/multiarray/methods.c
```diff
@@ -963,6 +963,7 @@ array_getarray(PyArrayObject *self, PyObject *args, PyObject *kwds)
         if (newtype == NULL) {
             newtype = PyArray_DESCR(self);
         }
+        Py_INCREF(newtype);
```
benburrill: What about the case when you pass an explicit dtype?

With this change, the following test fails:
```python
dt2 = np.dtype(np.float32)
old_refcount2 = sys.getrefcount(dt2)
np.array(MyArray(), dtype=dt2)
assert_equal(old_refcount2, sys.getrefcount(dt2))
```

For this case, the following seems to fix it:

```suggestion
        if (newtype == NULL) {
            newtype = PyArray_DESCR(self);
            Py_INCREF(newtype);
        }
```

numpy/_core/src/multiarray/methods.c
```diff
@@ -962,17 +962,19 @@ array_getarray(PyArrayObject *self, PyObject *args, PyObject *kwds)
...
-        Py_INCREF(newtype);
         ret = PyArray_CastToType(self, newtype, 0);
         Py_DECREF(self);
         return ret;
     } else { // copy == NPY_COPY_IF_NEEDED || copy == NPY_COPY_NEVER
-        if (newtype == NULL || PyArray_EquivTypes(PyArray_DESCR(self), newtype)) {
+        if (newtype == NULL) {
+            return (PyObject *)self;
+        } else if (PyArray_EquivTypes(PyArray_DESCR(self), newtype)) {
+            Py_DECREF(newtype);
```
ngoldbaum:
```suggestion
        if (newtype == NULL || PyArray_EquivTypes(PyArray_DESCR(self), newtype)) {
            Py_XDECREF(newtype);
```

mtsokol: Done!

numpy/_core/src/multiarray/methods.c
mtsokol: Thanks for pointing it out! Indeed this path wasn't tested - fixed it.

numpy/_core/tests/test_api.py
```diff
@@ -91,12 +91,36 @@ def test_array_array():
...
     assert_equal(bytes(np.array(o).data), bytes(a.data))
 
-    # test array
+    # test __array__
     def custom__array__(self, dtype=None, copy=None):
         return np.array(100.0, dtype=dtype, copy=copy)
 
     o = type("o", (object,), {"__array__": custom__array__})()
     assert_equal(np.array(o, dtype=np.float64), np.array(100.0, np.float64))
+    if HAS_REFCOUNT:
```
benburrill: Although other tests in this same file also check HAS_REFCOUNT with `if`, I think it might be better to use a skipif test instead, so the test doesn't pass silently.  See, for example: https://github.com/numpy/numpy/blob/83402f7a43363e8041b8f297fdd4f76a7fa9fea1/numpy/_core/tests/test_arrayprint.py#L548

ngoldbaum: In practice that means that rest of the test gets skipped on pypy, which I don't think we want.

benburrill: What I meant was splitting it into a separate test with skipif, not marking the whole test as skipif.

mtsokol: Makes sense - I separated it into a new refcount test.

ngoldbaum:
@benburrill thanks for the code review help!

The reference counting here is tricky - I'd like @seberg to look this over. He's on vacation this week so might not look at it until next week.

mtsokol:
Sure, let's wait for another pair of eyes. In general, `PyArg_ParseTupleAndKeywords` increments refcount for `newtype`, and `PyArray_CastToType` steals it - therefore the standard path where we pass `dtype` doesn't alter refcount.

When we don't pass it, and have `copy=True`, we need an additional `Py_INCREF(newtype);` to even out `PyArray_CastToType`.

When we pass it (and is the same as array's) and have `copy=None`, we have one refcount extra from `PyArg_ParseTupleAndKeywords` parsing, ergo we need an additional `Py_XDECREF(newtype);` for that path to even out missing stealing call (which is only present later, when `newtype` is present and differs from array's dtype).

The test covers: `dtype(none | same | different) x copy(true | false)` (6) scenarios.

ngoldbaum:
I think it would help if you added an edited-down subset of that last comment as a comment in the source code. Probably some of it makes more sense as a comment in the test and some of it makes sense as a comment in the `__array__` implementation.

benburrill:
> In general, `PyArg_ParseTupleAndKeywords` increments refcount for `newtype`

Just to be slightly more precise, `PyArray_DescrConverter2` is (as I understand it) what's actually responsible for incrementing the refcount for `newtype`, not `PyArg_ParseTupleAndKeywords` itself.

mtsokol:
Correct, `PyArray_DescrConverter2` converter function which is passed to `PyArg_ParseTupleAndKeywords`.

benburrill:
Yeah, I just thought the wording could potentially be a bit misleading to readers (if you're planning to put something similar in a comment in the source code), since although

> In general, `PyArg_ParseTupleAndKeywords` increments refcount for `newtype`

is true for this specific code,

> In general, `PyArg_ParseTupleAndKeywords` increments refcount

is not true.

numpy/_core/tests/test_api.py
```diff
@@ -157,6 +157,36 @@ def custom__array__(self, dtype=None, copy=None):
...
+def test___array___refcount():
+    class MyArray:
+        def __init__(self, dtype):
+            self.val = np.array(-1, dtype=dtype)
+
+        def __array__(self, dtype=None, copy=None):
+            return self.val.__array__(dtype=dtype, copy=copy)
+
+    # test all possible scenarios:
+    # dtype(none | same | different) x copy(true | false)
```
benburrill: I just realized you haven't fully tested the COPY_NEVER case.  Really, you should be testing `copy(True | False | None)`.  Currently, you're actually testing `copy(True | None)`.

This test fails:
```python
    np.array(MyArray(dt), dtype=dt2, copy=False)
assert_equal(old_refcount2, sys.getrefcount(dt2))
```

Adding `Py_DECREF(newtype);` to the `npy_no_copy_err_msg` branch fixes it.

I'm also suspicious about the `if (!PyArray_CheckExact(self))` branch (I think at the very least there should be `Py_XDECREF(newtype)` before the `return NULL;`, and possibly more than that), but I'm not totally sure how to test it.

mtsokol: Right, `copy=False` when dtypes are different was a path missed in the test - thanks!

About `PyArray_CheckExact`: `Py_XDECREF` was clearly missing there, but I think it's already an esoteric path, as `PyArray_NewFromDescrAndBase` needs to fail and set an exception.

numpy/_core/src/multiarray/methods.c
```diff
@@ -962,12 +963,18 @@ array_getarray(PyArrayObject *self, PyObject *args, PyObject *kwds)
     if (copy == NPY_COPY_ALWAYS) {
         if (newtype == NULL) {
             newtype = PyArray_DESCR(self);
+            // Take a new strong reference to compensate for
+            // PyArray_CastToType stealing a reference to newtype.
+            Py_INCREF(newtype);
```
seberg:
```suggestion
            Py_INCREF(newtype);  // newtype is owned.
```
Maybe. I don't like the overly long comments, TBH.

We could add a comment on all `CastToType` calls that it steals the reference, though.

numpy/_core/src/multiarray/methods.c
```diff
@@ -962,12 +963,18 @@ array_getarray(PyArrayObject *self, PyObject *args, PyObject *kwds)
...
         }
         ret = PyArray_CastToType(self, newtype, 0);
         Py_DECREF(self);
         return ret;
     } else { // copy == NPY_COPY_IF_NEEDED || copy == NPY_COPY_NEVER
         if (newtype == NULL || PyArray_EquivTypes(PyArray_DESCR(self), newtype)) {
+            // If newtype isn't NULL then release a strong reference
+            // introduced by PyArray_DescrConverter2 in arg parsing.
```
seberg:
```suggestion
```
I disagree that it is helpful to add it here when it's not everywhere else.

seberg (approved):
LGTM, although I think the new comments are more verbose than helpful.  Half the paths (at least the error paths) have no comment.  What *is* useful, IMO, is that `CastToType` steals the reference (because reference stealing is confusing API basically).

If we want to really clarify this all, I think the right thing would be to simplify the code instead...
* Swap `self` and `new` naming.  Because `new` is clearly a new reference ;), while `self` is *suprisingly* a new reference.
* One could move `newtype == NULL` logic to the top directly after the argument parsing, that way it get's much clearer why the `Incref` is needed.  (at the cost of an unnecessary incref/decref.)
* The best way to avoid piling up `DECREF()` before the return would be to just make it `goto finish;` that has the logic.  But since that requires adding `newtype = NULL;  // Stolen by CastToType` (or another incref), I am not sure it's nicer.  It would be clearer though.

Anyway, besides thinking that the comments are not useful except as comments during review, this looks right to me.

numpy/_core/src/multiarray/methods.c
mtsokol: Done! Also added `CastToType` comments.

numpy/_core/src/multiarray/methods.c
mtsokol: Removed!

mtsokol:
@seberg thanks for feedback! I found moving `newtype == NULL` check to the top really useful - now it's clearer how we count refs here. I decided skip the other two points if it's OK.

numpy/_core/src/multiarray/methods.c
```diff
@@ -960,22 +966,22 @@ array_getarray(PyArrayObject *self, PyObject *args, PyObject *kwds)
...
             return (PyObject *)self;
         }
         if (copy == NPY_COPY_IF_NEEDED) {
-            ret = PyArray_CastToType(self, newtype, 0);
+            ret = PyArray_CastToType(self, newtype, 0);  // steals newtype reference.
             Py_DECREF(self);
             return ret;
         } else { // copy == NPY_COPY_NEVER
             PyErr_SetString(PyExc_ValueError, npy_no_copy_err_msg);
+            // On error release a strong reference.
```
seberg:
```suggestion
```

seberg:
Nice, thanks!  Yeah, I didn't care about doing all of those things, more of a rambling that in my opinion code reorganization would do more good than long comments :).

PR #29718, opened 2025-09-08
TST: not to include the LONGDOUBLE test on AIX

kkwli:
On AIX, `double` and `long double` have the same size. This patch is
update the test to exclude `LONGDOUBLE` test.


mattip:
Thanks @kkwli

PR #29729, opened 2025-09-09
TST: Fix np.random thread test failures

bwhitt7:
This PR is related to #29552, specifically seeking to address `np.random` thread safe issues. Calls to legacy `np.random` affects global state, and occasionally this results in test failures when running tests in parallel threads, especially when tests are expecting specific results for a specific seed (`np.random.seed` is a good indicator of this). This can be fixed by making these global state calls into local state calls using `Generator` or `RandomState` instances.

For now, most of these tests have been converted to use `np.random.RandomState` instances, since these test files are generally testing the legacy RandomState. Notably `numpy/random/test/test_random.py` and `numpy/random/test/test_regression.py` have not been touched since I figured they should continue to test the global `np.random` calls and not be run in parallel. If folks think it's okay to also convert these files, I can definitely do so.

### DIFF
--- a/numpy/random/tests/test_random.py
+++ b/numpy/random/tests/test_random.py
@@ -1062,9 +1062,6 @@ class TestBroadcast:
     # correctly when presented with non-scalar arguments
     seed = 123456789
 
-    def setSeed(self):
-        return random.RandomState(self.seed)
-
     # TODO: Include test for randint once it can broadcast
     # Can steal the test written in PR #6938
 
@@ -1075,11 +1072,11 @@ class TestBroadcast:
                             0.53413660089041659,
                             0.50955303552646702])
 
-        rng = self.setSeed()
+        rng = random.RandomState(self.seed)
         actual = rng.uniform(low * 3, high)
         assert_array_almost_equal(actual, desired, decimal=14)
 
-        rng = self.setSeed()
+        rng = random.RandomState(self.seed)
         actual = rng.uniform(low, high * 3)
         assert_array_almost_equal(actual, desired, decimal=14)
 
@@ -1091,12 +1088,12 @@ class TestBroadcast:
                             2.1283977976520019,
                             1.8417114045748335])
 
-        rng = self.setSeed()
+        rng = random.RandomState(self.seed)
         actual = rng.normal(loc * 3, scale)
         assert_array_almost_equal(actual, desired, decimal=14)
         assert_raises(ValueError, rng.normal, loc * 3, bad_scale)
 
-        rng = self.setSeed()
+        rng = random.RandomState(self.seed)
         actual = rng.normal(loc, scale * 3)
         assert_array_almost_equal(actual, desired, decimal=14)
         assert_raises(ValueError, rng.normal, loc, bad_scale * 3)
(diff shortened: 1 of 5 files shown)


numpy/random/tests/test_randomstate.py
```diff
@@ -1320,134 +1319,127 @@ class TestBroadcast:
     seed = 123456789
 
     def set_seed(self):
-        random.seed(self.seed)
+        return random.RandomState(self.seed)
```
rkern: The name of this method is no longer appropriate. Feel free to inline `random.RandomState(self.seed)` everywhere in this class (or rename this method if it's easier).

numpy/random/tests/test_randomstate.py
```diff
@@ -364,20 +364,20 @@ def test_repeatability(self):
                'uint8':  '001aac3a5acb935a9b186cbe14a1ca064b8bb2dd0b045d48abeacf74d0203404'}  # noqa: E501
 
         for dt in self.itype[1:]:
-            random.seed(1234)
+            rng = random.RandomState(1234)
 
             # view as little endian for hash
             if sys.byteorder == 'little':
-                val = self.rfunc(0, 6, size=1000, dtype=dt)
+                val = rng.randint(0, 6, size=1000, dtype=dt)
```
rkern, line 373: There are more `self.rfunc` references in this class that still need to be removed. Delete the class attribute as it's no longer necessary.

rkern (commented):
A couple of minor things. Other than that, the changes look good.

AFAICT, all of the tests in `test_random.py` could be converted to use explicit `RandomState` instances; I don't think they test anything about the `np.random.*` global method aliases separately. The only test in `test_regression.py` that actually tests global state functionality specifically is `test_call_within_randomstate` (it checks whether a specific `RandomState` method call on a separate instance might affect the global state). The rest are safe to convert. I think many of the `test_random.py` tests are duplicated by `test_randomstate.py`, but I don't think we need a separate copy that tests the `np.random.*` aliases explicitly, at least not in terms of their detailed functionality. IMO, at least.

bwhitt7:
@rkern Thank you for the feedback! I went ahead and committed the changes you suggested, as well as making changes to `test_random.py` and `test_regressions.py`. For now I'm mainly converting the tests that caused thread safety issues, but I can go back to these tests at some point and convert the rest of of the global `np.random.*` calls to use `RandomState` instances instead if folks think that's best.

numpy/random/tests/test_random.py
```diff
@@ -1060,137 +1063,130 @@ class TestBroadcast:
     seed = 123456789
 
     def setSeed(self):
```
ngoldbaum: I feel like this should be renamed to e.g. `create_rng`. But also it's a one-liner and doesn't really need to be a function at all IMO.

ngoldbaum (approved):
I read through everything and this is 100% a mechanical change that looks correct to me. Just one nitpick where I think we don't want to leave it as it was anymore because of how you've changed it.

numpy/random/tests/test_random.py
bwhitt7: Ah, I made this change in `test_randomstate` and forgot to do the same in `test_random`, thanks for the catch!

ngoldbaum:
Thanks @bwhitt7!

PR #29733, opened 2025-09-10
TYP: fix method declarations in ``floating``, ``timedelta64``, and ``datetime64``

jorenham:
Follow-up of #29695 and #29706, resolving https://github.com/python/mypy/pull/15993#issuecomment-3240220599


jorenham:
I like the mypy_primer diffs.

For freqtrade it's safe to ignore, because it's just showing that the reported error code changed for a single line.
For `pandas-stubs` and [`scipy-stubs`](https://github.com/scipy/scipy-stubs/)[^1] this is good news, since they can remove a bunch of `# type ignore`s with this.

[^1]: Wow what an amazing package! I wonder who wrote it... 😏

charris:
Thanks Joren.

PR #28925, opened 2025-05-08
DEP: Deprecate setting the strides attribute of a numpy array

eendebakpt:
Addresses part of https://github.com/numpy/numpy/issues/28800.

Changes are factored out of https://github.com/numpy/numpy/pull/28901.

PR #29734, opened 2025-09-10
TYP: fix ``ndarray.strides`` decorator order

jorenham:
The `@deprecated` decorator was added in https://github.com/numpy/numpy/pull/28925, but mypy did not like the order:

```
Only supported top decorator is @strides.setter (misc)
```


charris:
Thanks Joren.

PR #29736, opened 2025-09-10
TYP: sort out some of the ``# type: ignore`` comments in ``__init__.pyi``

jorenham:
This makes it a bit ~easier~ less difficult to spot the squigglies that we should actually care about, because it's a bit of a "boy who cried wolf" situation right now.


charris:
Thanks Joren.

PR #29740, opened 2025-09-11
TST: disable overflow exception test of numpy.power on AIX

kkwli:
`numpy.power()` does not raise the overflow exception with numpy.float64 type.
It can be illustrated in the following code.
```python
import numpy as np
#fi = np.finfo(np.float64)
dtyp = np.dtype(np.float64).type
with np.errstate(all='raise'):
  try:
    x = np.power(dtyp(2), dtyp(2048))   # 2**fi.nexp == 2048
    print(f'No fpe error ({x})')
  except FloatingPointError as exc:
    print(f'FloatingPointError')
```
Output on AIX:
```
No fpe error (inf)
```
Output on Linux/PowerPC and Linux/x86:
```
FloatingPointError
```
This patch is to guard the `self.assert_raises_fpe(overflow, np.power, ftype(2), ftype(2**fi.nexp))` test to skip it on AIX.

### DIFF
--- a/numpy/_core/tests/test_numeric.py
+++ b/numpy/_core/tests/test_numeric.py
@@ -1033,8 +1033,11 @@ class TestFloatExceptions:
                                    lambda a, b: a + b, ft_max, ft_max * ft_eps)
             self.assert_raises_fpe(overflow,
                                    lambda a, b: a - b, -ft_max, ft_max * ft_eps)
-            self.assert_raises_fpe(overflow,
-                                   np.power, ftype(2), ftype(2**fi.nexp))
+            # On AIX, pow() with double does not raise the overflow exception,
+            # it returns inf. Long double is the same as double.
+            if sys.platform != 'aix' or typecode not in 'dDgG':
+                self.assert_raises_fpe(overflow,
+                                       np.power, ftype(2), ftype(2**fi.nexp))
             self.assert_raises_fpe(divbyzero,
                                    lambda a, b: a / b, ftype(1), ftype(0))
             self.assert_raises_fpe(


numpy/_core/tests/test_numeric.py
```diff
@@ -1033,8 +1033,11 @@ def test_floating_exceptions(self, typecode):
                                    lambda a, b: a + b, ft_max, ft_max * ft_eps)
             self.assert_raises_fpe(overflow,
                                    lambda a, b: a - b, -ft_max, ft_max * ft_eps)
-            self.assert_raises_fpe(overflow,
-                                   np.power, ftype(2), ftype(2**fi.nexp))
+            # On AIX, pow() with double does not raise the overflow exception,
+            # it returns inf. Long double is the same as double.
+            if sys.platform != 'aix' or typecode not in 'dDgG':
+                self.assert_raises_fpe(overflow,
+                                       np.power, ftype(2), ftype(2**fi.nexp))
```
ngoldbaum, line 1040: Is this papering over a bug in our floating point exception detection or is this just a fundamental limitation of the AIX math library?

kkwli: Sorry for the slow response. The `assert_raises_fpe` method expects an exception but there is no exception raised in this case (`np.power(ftype(2), ftype(2**fi.nexp))`). In my opinion, the detection is performing as designed. 

For this particular case (i.e. `pow(2., 2048.)` in C), the `errno` is set to `ERANGE` instead and no exception is raised on AIX. I do not know if we can detect `errno` with `np.power(ftype(2), ftype(2**fi.nexp))` in order to special case on AIX. Any suggestion is welcome.

ngoldbaum: > In my opinion, the detection is performing as designed.

What motivates that opinion? It seems to work differently on all other platforms. Is there a piece of documentation we can link to?

> For this particular case (i.e. pow(2., 2048.) in C), the errno is set to ERANGE instead and no exception is raised on AIX. I do not know if we can detect errno with np.power(ftype(2), ftype(2**fi.nexp)) in order to special case on AIX. Any suggestion is welcome.

Ah, thanks, this is nice context. Here's where the "guts" of the scalar power operation are defined:

https://github.com/numpy/numpy/blob/5f800455a0fb86d4fd33cd93336996e4b00f0414/numpy/_core/src/umath/scalarmath.c.src#L1524-L1565

You can see it uses `npy_get_floatstatus_barrier` to check the FPE register. If you try to use e.g. a C debugger, is no FPE getting set?

I think errno and erange is a separate issue, numpy doesn't use the C standard library pow function to implement the power operator. We do our own FPE checking.

I think @seberg has more context about how NumPy's FPE checking works than I do. Pinging in case he's interested.

seberg: Hmmm, I wasn't aware that errno may be used for math errors?  In practice I didn't realize that happened.  NumPy always uses `fenv.h` only, I think.

[this](https://en.cppreference.com/w/c/numeric/math/math_errhandling.html) says:
> If the implementation supports IEEE floating-point arithmetic (IEC 60559), math_errhandling & MATH_ERREXCEPT is required to be non-zero.
so I am not sure that AIX is just not fully IEEE compliant and we shouldn't care, or whether errno can actually be translated in any reasonable form (I suspect no, but not sure).

FWIW, NumPy cares about these, but a lot just don't so ignoring seems OK to me. Spurious ones are more problematic (e.g. from bad SIMD code).

But maybe I am missing something?

ngoldbaum: > But maybe I am missing something?

No, I just like having a principled reason for adding test skips. I agree that it's probably fine to add this, but it'd be nice to have a comment explaining why.

kkwli: @ngoldbaum Thanks for the suggestion. I cannot set the breakpoint at `npy_get_floatstatus_barrier`. I already built the numpy with `... CFLAGS=' -g -O0' ... -Dbuildtype=debug`. I also insert an `assert` in `double_ctype_power`, it seems that it is not called.

ngoldbaum: I'm sorry, I pointed you at the wrong spot in NumPy internals. My fault for not actually trying my suggestion.

I made a slightly modified version of your test script:

<details>

```python
import os
import signal

import numpy as np

PID = os.getpid()

def do_nothing(*args):
    pass

signal.signal(signal.SIGUSR1, do_nothing)

os.kill(PID, signal.SIGUSR1)

#fi = np.finfo(np.float64)
dtyp = np.dtype(np.float64).type
with np.errstate(all='raise'):
  try:
    x = np.power(dtyp(2), dtyp(2048))   # 2**fi.nexp == 2048
    print(f'No fpe error ({x})')
  except FloatingPointError as exc:
    print(f'FloatingPointError')
```

</details>

This follows a tip I recently added to the docs to make it easier to set braekpoints that might get triggered while NumPy is initializing: https://numpy.org/devdocs/dev/development_advanced_debugging.html#running-a-test-script

When I try on my ARM Mac, I see two calls to `npy_get_floatstatus_barrier` after the numpy import finishes. Here are the stack traces:

```
(lldb) break set --name npy_get_floatstatus_barrier
Breakpoint 1: 2 locations.
(lldb) c
Process 41962 resuming
Process 41962 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
    frame #0: 0x0000000101bfe360 _multiarray_umath.cpython-314t-darwin.so`npy_get_floatstatus_barrier(param="") at ieee754.c.src:359:20
   356
   357 	int npy_get_floatstatus_barrier(char* param)
   358 	{
-> 359 	    int fpstatus = fetestexcept(FE_DIVBYZERO | FE_OVERFLOW |
   360 	                                FE_UNDERFLOW | FE_INVALID);
   361 	    /*
   362 	     * By using a volatile, the compiler cannot reorder this call
Target 0: (python3.14) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
  * frame #0: 0x0000000101bfe360 _multiarray_umath.cpython-314t-darwin.so`npy_get_floatstatus_barrier(param="") at ieee754.c.src:359:20
    frame #1: 0x0000000101bfe300 _multiarray_umath.cpython-314t-darwin.so`npy_clear_floatstatus_barrier(param="") at ieee754.c.src:377:20
    frame #2: 0x0000000101bb0348 _multiarray_umath.cpython-314t-darwin.so`try_trivial_single_output_loop(context=0x000000016fdfd3a8, op=0x000000016fdfd840, order=NPY_KEEPORDER, errormask=1170) at ufunc_object.c:969:9
    frame #3: 0x0000000101bacdb0 _multiarray_umath.cpython-314t-darwin.so`PyUFunc_GenericFunctionInternal(ufunc=0x000004356e623d00, ufuncimpl=0x000004356e4edc80, operation_descrs=0x000000016fdfd878, op=0x000000016fdfd840, casting=NPY_SAME_KIND_CASTING, order=NPY_KEEPORDER, wheremask=0x0000000000000000) at ufunc_object.c:2241:26
    frame #4: 0x0000000101baa9ac _multiarray_umath.cpython-314t-darwin.so`ufunc_generic_fastcall(ufunc=0x000004356e623d00, args='0x16fdfd7f8', len_args=2, kwnames='0x16fdfd7e8', outer='\0') at ufunc_object.c:4555:18
    frame #5: 0x0000000101ba8174 _multiarray_umath.cpython-314t-darwin.so`ufunc_generic_vectorcall(ufunc='0x16fdfd928', args='0x16fdfd920', len_args=9223372036854775810, kwnames='0x16fdfd910') at ufunc_object.c:4619:12
    frame #6: 0x00000001007a5c00 libpython3.14t.dylib`PyObject_Vectorcall + 88
    frame #7: 0x00000001008f7f24 libpython3.14t.dylib`_PyEval_EvalFrameDefault + 9052
    frame #8: 0x00000001008f5640 libpython3.14t.dylib`PyEval_EvalCode + 604
    frame #9: 0x0000000100981320 libpython3.14t.dylib`run_mod + 624
    frame #10: 0x000000010097dde4 libpython3.14t.dylib`_PyRun_SimpleFileObject + 912
    frame #11: 0x000000010097d6ac libpython3.14t.dylib`_PyRun_AnyFileObject + 160
    frame #12: 0x00000001009abdfc libpython3.14t.dylib`pymain_run_file + 332
    frame #13: 0x00000001009ab4c8 libpython3.14t.dylib`Py_RunMain + 1912
    frame #14: 0x00000001009ab81c libpython3.14t.dylib`pymain_main + 324
    frame #15: 0x00000001009ab8bc libpython3.14t.dylib`Py_BytesMain + 40
    frame #16: 0x000000019c152b98 dyld`start + 6076
(lldb) c
Process 41962 resuming
Process 41962 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
    frame #0: 0x0000000101bfe360 _multiarray_umath.cpython-314t-darwin.so`npy_get_floatstatus_barrier(param="power") at ieee754.c.src:359:20
   356
   357 	int npy_get_floatstatus_barrier(char* param)
   358 	{
-> 359 	    int fpstatus = fetestexcept(FE_DIVBYZERO | FE_OVERFLOW |
   360 	                                FE_UNDERFLOW | FE_INVALID);
   361 	    /*
   362 	     * By using a volatile, the compiler cannot reorder this call
Target 0: (python3.14) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
  * frame #0: 0x0000000101bfe360 _multiarray_umath.cpython-314t-darwin.so`npy_get_floatstatus_barrier(param="power") at ieee754.c.src:359:20
    frame #1: 0x0000000101ba3504 _multiarray_umath.cpython-314t-darwin.so`_check_ufunc_fperr(errmask=1170, ufunc_name="power") at extobj.c:534:13
    frame #2: 0x0000000101bb0438 _multiarray_umath.cpython-314t-darwin.so`try_trivial_single_output_loop(context=0x000000016fdfd3a8, op=0x000000016fdfd840, order=NPY_KEEPORDER, errormask=1170) at ufunc_object.c:991:15
    frame #3: 0x0000000101bacdb0 _multiarray_umath.cpython-314t-darwin.so`PyUFunc_GenericFunctionInternal(ufunc=0x000004356e623d00, ufuncimpl=0x000004356e4edc80, operation_descrs=0x000000016fdfd878, op=0x000000016fdfd840, casting=NPY_SAME_KIND_CASTING, order=NPY_KEEPORDER, wheremask=0x0000000000000000) at ufunc_object.c:2241:26
    frame #4: 0x0000000101baa9ac _multiarray_umath.cpython-314t-darwin.so`ufunc_generic_fastcall(ufunc=0x000004356e623d00, args='0x16fdfd7f8', len_args=2, kwnames='0x16fdfd7e8', outer='\0') at ufunc_object.c:4555:18
    frame #5: 0x0000000101ba8174 _multiarray_umath.cpython-314t-darwin.so`ufunc_generic_vectorcall(ufunc='0x16fdfd928', args='0x16fdfd920', len_args=9223372036854775810, kwnames='0x16fdfd910') at ufunc_object.c:4619:12
    frame #6: 0x00000001007a5c00 libpython3.14t.dylib`PyObject_Vectorcall + 88
    frame #7: 0x00000001008f7f24 libpython3.14t.dylib`_PyEval_EvalFrameDefault + 9052
    frame #8: 0x00000001008f5640 libpython3.14t.dylib`PyEval_EvalCode + 604
    frame #9: 0x0000000100981320 libpython3.14t.dylib`run_mod + 624
    frame #10: 0x000000010097dde4 libpython3.14t.dylib`_PyRun_SimpleFileObject + 912
    frame #11: 0x000000010097d6ac libpython3.14t.dylib`_PyRun_AnyFileObject + 160
    frame #12: 0x00000001009abdfc libpython3.14t.dylib`pymain_run_file + 332
    frame #13: 0x00000001009ab4c8 libpython3.14t.dylib`Py_RunMain + 1912
    frame #14: 0x00000001009ab81c libpython3.14t.dylib`pymain_main + 324
    frame #15: 0x00000001009ab8bc libpython3.14t.dylib`Py_BytesMain + 40
    frame #16: 0x000000019c152b98 dyld`start + 6076
(lldb) c
Process 41962 resuming
FloatingPointError
```

So I think you actually want to step through the execution of `try_trivial_single_output_loop` to see what's going on with the FPE detection.

kkwli: Apologies for the long delay. Other projects got in the way! Thanks a lot for the instructions to debug the code. 

I have no luck in building a debuggable library even though I supply `CFLAGS='-O0 -g'` and `-Dbuildtype=debug` in the build.
```
(gdb) info sharedlibrary
From                To                  Syms Read   Shared Object Library
...
0x0900000000000700  0x090000000048b221  Yes (*)     /usr/lib/libc.a(shr_64.o)
0x090000001651b438  0x09000000167ef4ac  No          /home/kli/numpy/build-install/usr/lib64/python3.12/site-packages/numpy/linalg/_umath_linalg.cpython-312.so
...
(*): Shared library is missing debugging information.
```
Anyway, you a few print statements to get the idea of what the code path is. I think operation (`pow`) is performed in https://github.com/numpy/numpy/blob/dd6fdef027db9fae59ab1fb03b25865ac512073b/numpy/_core/src/umath/ufunc_object.c#L975
Since both `flags` and `res` are zero, it executes https://github.com/numpy/numpy/blob/dd6fdef027db9fae59ab1fb03b25865ac512073b/numpy/_core/src/umath/ufunc_object.c#L988-L991
Inside `_check_ufunc_fperr`, `npy_get_floatstatus_barrier` is called. In `npy_get_floatstatus_barrier`, I check that `fetestexcept` returns `0`. If `strided_loop` really performs `pow` routine, the behavior is consistent with what we expect on AIX (i.e. `pow(2.0, 2048)` does not raise an exception).

kkwli: This C code illustrates the particular behavior on AIX.
```c
#include <stdio.h>
#include <math.h>
#include <fenv.h>
#include <errno.h>
#include <string.h>

#pragma STDC FENV_ACCESS ON

int main() {
  double result = 1.0;
  double base = 2.0;
  int exponent = 2048;

  errno = 0;
  feclearexcept(FE_ALL_EXCEPT); // Clear all exceptions

// Case 1
  result = pow(base, exponent);
// Case 2
//  for (int i=0; i<2048; i++)
//    result *= base;

  if (fetestexcept(FE_OVERFLOW)) {
    printf("Overflow occurred!\n");
  } else {
    printf("Result: %f\n", result);
  }

  printf("errno=%d ... %s\n", errno, strerror(errno));
  return 0;
}
```
Output:
```
Result: INF
errno=34 ... Result too large
```

ngoldbaum: OK, I'm now sufficiently convinced we're not just papering over a bug. Thanks for looking into this!

kkwli: @ngoldbaum Thanks.

PR #29696, opened 2025-09-06
DOC: Improve the format of a math operation

star1327p:
Improve the format of a math operation in:
https://numpy.org/devdocs/dev/internals.code-explanations.html

Improved version:
image

Original version:
image

PR #29742, opened 2025-09-11
DOC: Improve math format in accumulate 

star1327p:
Improve math format in accumulate 
https://numpy.org/devdocs/dev/internals.code-explanations.html

Closed #29696 and reopened a new PR.

### DIFF
--- a/doc/source/dev/internals.code-explanations.rst
+++ b/doc/source/dev/internals.code-explanations.rst
@@ -563,7 +563,10 @@ way, the operation performed is
 .. math::
    :nowrap:
 
-   $$o_k = \begin{cases} & i[0] & k=0 \\\ & i[k] \text{ <op> }o_{k-1} & k=1,\ldots, N \text{.} \end{cases}$$
+   \begin{align*}
+   o & =  \quad i[0] \quad  &  \\
+   o & =  i[k]\textrm{ <op> }o & k=1,\ldots, N.
+   \end{align*}
 
 where :math:`N+1` is the number of elements in the input, :math:`i`,
 :math:`o` is the output, and :math:`i[k]` is the
--- a/doc/source/dev/internals.code-explanations.rst
+++ b/doc/source/dev/internals.code-explanations.rst
@@ -563,10 +563,7 @@ way, the operation performed is
 .. math::
    :nowrap:
 
-   \begin{align*}
-   o & =  \quad i[0] \quad  &  \\
-   o & =  i[k]\textrm{ <op> }o & k=1,\ldots, N.
-   \end{align*}
+   $$o_k = \begin{cases} & i[0] & k=0 \\\ & i[k] \text{ <op> }o_{k-1} & k=1,\ldots, N \text{.} \end{cases}$$
 
 where :math:`N+1` is the number of elements in the input, :math:`i`,
 :math:`o` is the output, and :math:`i[k]` is the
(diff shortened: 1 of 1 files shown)


jorenham:
[current](https://numpy.org/doc/stable/dev/internals.code-explanations.html#accumulate):

image

[this PR](https://output.circle-artifacts.com/output/job/69e163bb-fd3b-4907-b8ab-a5b1cacc2a6d/artifacts/0/doc/build/html/dev/internals.code-explanations.html#accumulate):

image

I can't say I like the alignment here. I think it'd be better without the `\quad`s

jorenham:
How about something like this? 

```math
   o_k = \begin{cases}
     & i[0]  & \text{if } k=0  \\
     & i[k] \text{ <op> }o_{k-1} & \text{if } k=1,\ldots, N \text{.}
   \end{cases}
```

star1327p:
@jorenham Thank you for the suggestion! I have made the changes, and hope this works!

mattip:
Close/reopen to retry failed CI setup

doc/source/dev/internals.code-explanations.rst
```diff
@@ -563,10 +563,7 @@ way, the operation performed is
 .. math::
    :nowrap:
 
-   \begin{align*}
-   o & = & i[0] \\
-   o & = & i[k]\textrm{<op>}o\quad k=1\ldots N
-   \end{align*}
+   $$o_k = \begin{cases} & i[0] & k=0 \\\ & i[k] \text{ <op> }o_{k-1} & k=1,\ldots, N \text{.} \end{cases}$$
```
mattip: Why did you change `o` to be a vector? It is a single value in a reduce. That is the point of reduction. It is initialized to the value of `i[0]` and then subsequent iterations update it `o = i[k] <op> o`.

doc/source/dev/internals.code-explanations.rst
```diff
@@ -599,10 +596,7 @@ the current output pointer. Thus, the operation performed is
 .. math::
    :nowrap:
 
-   \begin{align*}
-   o[0] & = & i[0] \\
-   o[k] & = & i[k]\textrm{<op>}o[k-1]\quad k=1\ldots N.
-   \end{align*}
+   $$o_k = \begin{cases} & i[0] & k=0 \\\ & i[k] \text{ <op> }o_{k-1} & k=1,\ldots, N \text{.} \end{cases}$$
```
mattip, line 602: This now mixes subscript indexing and square-bracket indexing. Our documentation uses square brackets when describing ndarray indexing, so please revert that change. I am not sure the `cases` notation is appropriate here, since we are iterating over `i` and `o`, the first iteration is special.

mattip (commented):
I think your original motivation was to simply move the `i[0]` inwards a bit on the first line. Maybe for a first PR that is all you should try to do, and not change other things.

mattip:
I see @jorenham's suggestion above also mixes subscripting and bracket-indexing. Sorry to step on his toes.

star1327p:
@mattip Okay, I have applied the original changes in #29696. This time I remember to add `[skip actions][skip azp][skip cirrus]` in the commit to save CI resources. Hope this works!

jorenham:
> I see @jorenham's suggestion above also mixes subscripting and bracket-indexing. Sorry to step on his toes.

I figured that since it's a mathematical formulation, then the $=$ sign means "equals" and not "(re)assignment", so it'd require an index. E.g. if you write $a = a + 1$ that would imply that $a - a = 1$ and therefore $0 = 1$.
I agree that without an index it's probably clearer to most readers, but then this mathematical formulation might not be the best fit, and we should probably use pseudocode instead.

mattip:
Pseudocode is probably a better fit indeed. The formulations here do not accurately reflect the true C code, which uses `op.identity`. In [ufunc.reduce](https://numpy.org/devdocs/reference/generated/numpy.ufunc.accumulate.html#numpy-ufunc-accumulate) and [ufunc.accumulate](https://numpy.org/devdocs/reference/generated/numpy.ufunc.accumulate.html#numpy-ufunc-accumulate) there is a more accurate description of the functionality. The thing to make clear here is the way the same loops are used internally, only the step for the output is different.

mattip:
Closing, @star1327p if you wish to continue and refactor the equation as pseudocode that would be great.

Issue #29721, opened 2025-09-09
BUG (?): backwards compatibility of ndmax changes?


tylerjereddy:
I bisected NumPy against the SciPy `maintenance/1.16.x` branch and found some newly-occurring test failures are the result of the commit below the fold.

<details>

```
4e00e4df15528323f6d76bbd40157b582861af86 is the first bad commit
commit 4e00e4df15528323f6d76bbd40157b582861af86
Author: kibitzing <kang-jingu@naver.com>
Date:   Sat Aug 16 21:20:52 2025 +0900

    ENH: Add ndmax parameter to np.array to control recursion depth (#29569)
    
    * ENH: add ndmax parameter to np.array
    
    * ENH: validate ndmax argument is positive
    
    * TST: add ndmax tests for array creation
    
    * ENH: allow np.array with ndmax=0 to create 0-D array
    
    * MNT: simplify ndmax validation and error message
    
    * MNT: improve consistency in error message formatting
    
    * DOC: add comment explaining legacy behavior of max_depth=0
    
    * TST: update tests to reflect new ndmax validation and error message
    
    * DOC: add documentation and examples for np.array ndmax parameter
    
    * DOC: add release note for numpy.array ndmax parameter

 doc/release/upcoming_changes/29569.new_feature.rst | 27 ++++++++
 numpy/_core/_add_newdocs.py                        | 26 +++++++-
 numpy/_core/src/multiarray/array_converter.c       |  2 +-
 numpy/_core/src/multiarray/ctors.c                 | 19 +++++-
 numpy/_core/src/multiarray/multiarraymodule.c      | 28 ++++++---
 numpy/_core/src/multiarray/scalartypes.c.src       |  2 +-
 numpy/_core/tests/test_multiarray.py               | 73 ++++++++++++++++++++++
 7 files changed, 162 insertions(+), 15 deletions(-)
 create mode 100644 doc/release/upcoming_changes/29569.new_feature.rst
```

</details>

That's from gh-29569, which touches C code, and SciPy also has some tricky NumPy C API business going on in the test failures at https://github.com/scipy/scipy/pull/23543. Could we get a solution that is backwards compatible, even if the conclusion is that SciPy is misbehaving here?

I'll try to dissect a bit deeper in `_lfilter.cc` maybe--it may be that we're using an argument incorrectly, but even so, it fully backwards compatible down to bare metal it shouldn't break.

mattip:
[...]

PR #29745, opened 2025-09-12
BUG: Fix max_depth validation condition in PyArray_FromAny_int

kibitzing:
Fix for issue #29721 reported by @tylerjereddy after analysis and confirmation that this is indeed a bug.

When `ndim > max_depth`, an error should have been raised, but the unnecessary check `(in_DType == NULL || in_DType->type_num != NPY_OBJECT)` prevented the proper error from being raised, causing different errors to occur downstream instead. Therefore, this condition has been removed.

Detailed discussion can be found at: https://github.com/numpy/numpy/pull/29569/files#r2340732107

### DIFF
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -1593,7 +1593,7 @@ PyArray_FromAny_int(PyObject *op, PyArray_Descr *in_descr,
         npy_free_coercion_cache(cache);
         goto cleanup;
     }
-    if (ndim > max_depth && (in_DType == NULL || in_DType->type_num != NPY_OBJECT)) {
+    if (ndim > max_depth) {
         PyErr_SetString(PyExc_ValueError,
                 "object too deep for desired array");
         npy_free_coercion_cache(cache);
--- a/numpy/_core/tests/test_multiarray.py
+++ b/numpy/_core/tests/test_multiarray.py
@@ -1354,6 +1354,12 @@ class TestCreation:
         with pytest.raises(ValueError, match="ndmax must be in the range"):
             np.array(data, ndmax=65)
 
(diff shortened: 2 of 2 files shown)


mattip:
Before this PR the new test does not raise. It ignores the `ndmax=1` and returns an object array with shape `(2,3)`, which does not respect the `ndmax` argument. I guess a case could be made that we could wrap the input ndarray input into a new, 1d ndarray of `dtype=object` and shape `(1,)`, but that seems like a very odd use case. Let's go with this fix, and see if anyone complains.

mattip:
Thanks @kibitzing

kibitzing:
Thanks for the thorough review, @mattip !
As you said, with the current implementation, when a `np.array` is input and `ndmax` is smaller than the dim of the array, it raises an error instead of silently ignoring `ndmax`.
If this becomes an issue, I'd be glad to revisit and implement the approach you suggested.

mhvk:
Thanks both! Had I been asked, I would not have known what `np.array(np.array([[1,2,3],[4,5,6]]), dtype=object, ndmax=1)` should return: an object array scalar, or an object array with shape `(2,)` (as it would do for just the nested list), or indeed error. Refusing the temptation to guess and erroring seems appropriate here!

seberg:
Agreed an error is the only righ thing.  If the argument would be `ndims=1` then I suppose one could argue to just treat the nested array as a list.
With `ndmax` though that can't be cleanly defined and I agree to not guess.

But that is nothing to do.  I suppose if `ndmin` _and_ `ndmax` are used one could allow such logic.  I am not sure what happens right now in the `ndmin` case, TBH.  I suspect it might just add an additional dimension (which is just as wrong, but would have been the case since forever presumably).

Issue #26701, opened 2024-06-15
BUG: `clongdouble(str)` to parses as `float64` and loses precision in string conversion.


WarrenWeckesser:
### Describe the issue:

When given a string, it looks like `np.clongdouble()` is parsing the numeric values as `float64` and losing precision in the conversion.  For example, in the following, I expect `np.longdouble(s)` and `np.clongdouble(s).real` to give the same result:

```
In [73]: s = "0.333333333333333333333333"

In [74]: np.longdouble(s)
Out[74]: np.longdouble('0.33333333333333333334')

In [75]: np.clongdouble(s).real
Out[75]: np.longdouble('0.33333333333333331483')

``` 
Instead, `clongdouble()` appears to be parsing to `float64`, c.f.

```
In [78]: np.longdouble(np.float64(s))
Out[78]: np.longdouble('0.33333333333333331483')
```

This is on a Linux machine where the `long double` type is 80 bit extended precision.

### Reproduce the code example:

```python
See above.
```

### Error message:

```shell
No message, just incorrect result.
```

### Python and NumPy Versions:

2.1.0.dev0+git20240613.787ff0f
3.12.3 (main, Apr 25 2024, 11:11:06) [GCC 11.4.0]

### Runtime Environment:

[{'numpy_version': '2.1.0.dev0+git20240613.787ff0f',
  'python': '3.12.3 (main, Apr 25 2024, 11:11:06) [GCC 11.4.0]',
  'uname': uname_result(system='Linux', node='pop-os', release='6.8.0-76060800daily20240311-generic', version='#202403110203~1715181801~22.04~aba43ee SMP PREEMPT_DYNAMIC Wed M', machine='x86_64')},
 {'simd_extensions': {'baseline': ['SSE', 'SSE2', 'SSE3'],
                      'found': ['SSSE3',
                                'SSE41',
                                'POPCNT',
                                'SSE42',
                                'AVX',
                                'F16C',
                                'FMA3',
                                'AVX2'],
                      'not_found': ['AVX512F',
                                    'AVX512CD',
                                    'AVX512_KNL',
                                    'AVX512_KNM',
                                    'AVX512_SKX',
                                    'AVX512_CLX',
                                    'AVX512_CNL',
                                    'AVX512_ICL']}},
 {'architecture': 'Zen',
  'filepath': '/usr/lib/x86_64-linux-gnu/openblas-pthread/libopenblasp-r0.3.20.so',
  'internal_api': 'openblas',
[...]

PR #29746, opened 2025-09-12
BUG: clongdouble(str) loses precision during string conversion

MyUserNameWasTakenLinux:
This pull request addresses [#26701](https://github.com/numpy/numpy/issues/26701). Converting to a `np.clongdouble` from a string would result in a loss of precision. This is because the conversion process involves Python's complex type as an intermediate, which has less precision than `np.clongdouble`. The pull request adds a test which fails on the `main` branch on linux, x86.

### DIFF
--- a/numpy/_core/src/multiarray/arraytypes.c.src
+++ b/numpy/_core/src/multiarray/arraytypes.c.src
@@ -446,11 +446,11 @@ static PyObject *
 
 /**begin repeat
  *
- * #NAME = CFLOAT, CDOUBLE, CLONGDOUBLE#
- * #type = npy_cfloat, npy_cdouble, npy_clongdouble#
- * #ftype = npy_float, npy_double, npy_longdouble#
- * #kind = CFloat, CDouble, CLongDouble#
- * #suffix = f, , l#
+ * #NAME = CFLOAT, CDOUBLE#
+ * #type = npy_cfloat, npy_cdouble#
+ * #ftype = npy_float, npy_double#
+ * #kind = CFloat, CDouble#
+ * #suffix = f, #
  */
 NPY_NO_EXPORT int
 @NAME@_setitem(PyArray_Descr *descr, PyObject *op, char *ov)
@@ -528,6 +528,7 @@ NPY_NO_EXPORT int
 }
 
 /**end repeat**/
+
 
 static inline npy_longdouble
 string_to_long_double(PyObject*op)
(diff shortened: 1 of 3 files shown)


ngoldbaum:
Neat!

Just a head's up - longdouble is a bit of an unloved feature. That said, this seems reasonable to support.

Some suggestions:

* I doubt the test you added exercises all the code paths in the code you added. Can you try to make sure your tests cover all the new code you added? Error cases are often poorly tested, for example.
* While we're at it, we could fix the casts between the string DTypes and the longdouble dtype - IIRC the casts have the same problem.

I wonder if the linalg test that is failing is relying on the old broken conversion.

Ping @swayaminsync - you might be interested in this. Swayam has been working on a [quad precision DType](https://github.com/numpy/numpy-user-dtypes/tree/main/quaddtype) we're hoping will ultimately allow us to deprecate longdouble support (or at least the float128 alias) in NumPy proper. The fact that longdouble has a subtly different meaning under different architectures is a constant source of bug reports and most users won't need it at all.

WarrenWeckesser:
Not a full review, but here are a couple issues that should be easy to fix:

* This will not handle an input such as `'0.5j'` correctly; it returns `0.5+0j`.
* It incorrectly accepts an input such as `'0.5 plate of shrimp'`.  It prematurely returns successfully in that case.

WarrenWeckesser:
Again not a full review (sorry), but it looks like the failures are because the new `CLONGDOUBLE_setitem()` is not correctly handling all its possible inputs.  It might be given, for example, a Python `float`, a Python `complex`, a NumPy `complex64`, a NumPy `complex128`, a NumPy `float64`, a NumPy `longdouble`, etc.  I don't know the NumPy C API well enough off the top of my head to quickly suggest the minimal set of type checks and conversion functions that would handle all the cases that will be handed to `CLONGDOUBLE_setitem()`.

numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -620,6 +621,189 @@ LONGDOUBLE_setitem(PyObject *op, void *ov, void *vap)
...
+            PyErr_Format(PyExc_ValueError,
+                        "invalid literal for complex long double: %s",
+                        s);
+            Py_XDECREF(b);
+            return temp;
+        }
+
+        const char *p = end;
+        if (*p != '+' && *p != '-') {
+            if(*p == 'j') {
```
benburrill:
```suggestion
            if (*p == 'j' || *p == 'J') {
```

numpy/_core/src/multiarray/arraytypes.c.src
benburrill, line 1: Probably should allow parentheses ([as CPython, and hence `np.cdouble` does](https://github.com/python/cpython/blob/805e3368d6d07e58430654d1365283924fdf4143/Objects/complexobject.c#L947-L1068)).  Python likes to put parentheses in complex reprs, so I think it is important to handle this case at least, even if you're dropping support for some of the more esoteric bits of the CPython parser, like `1-j` and `1_000`.

numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -620,6 +621,189 @@ LONGDOUBLE_setitem(PyObject *op, void *ov, void *vap)
...
+        b = PyUnicode_AsUTF8String(op);
+        if(!b) {
+            return temp;
+        }
+    }
+    else {
+        b = op;
+        Py_XINCREF(b);
+    }
+    s = PyBytes_AsString(b);
```
benburrill, line 657: Most of these type checks and conversions are redundant because `string_to_long_cdouble` is only ever passed bytes or unicode objects it seems.  Why not just have `string_to_long_cdouble(char *s)`?

numpy/_core/src/multiarray/arraytypes.c.src
MyUserNameWasTakenLinux: Thanks for catching that

numpy/_core/src/multiarray/arraytypes.c.src
MyUserNameWasTakenLinux: You're right the type check `PyLong_Check(op) && !PyBool_Check(op)` is dead code, I had it there mainly because I was using `string_to_long_double` and `LONGDOUBLE_setitem` as templates. I think the function signature should still take a `PyObject*` to keep it consistent with `string_to_long_double(PyObject *op)`.

benburrill: Yeah that makes sense, I agree it should be consistent with `string_to_long_double`.  There's maybe a case to be made for changing `string_to_long_double` as well, but that's probably out of scope for this PR, so you can mark this as resolved.

MyUserNameWasTakenLinux: Thanks

numpy/_core/src/multiarray/arraytypes.c.src
MyUserNameWasTakenLinux: The parser can handle parentheses now, though I'm not sure if dropping support for `1-j` or `1_000` would be okay.

benburrill: For the case of `1_000`, although it's technically a regression, `np.longdouble` doesn't support underscores either, so I think it's perfectly reasonable for `np.clongdouble` not to support them anymore.  As for `1-j`, CPython says it's for backwards compatibility and may be removed in the future.  So in my opinion, probably fine to drop as well.  But I do think it would be good to get the opinion of a numpy maintainer.

MyUserNameWasTakenLinux:
Currently, I'm relying on Python's complex type to deal with Python types as input to `CLONGDOUBLE_setitem(PyObject *op)`. But this causes a loss in precision when converting from a Python `int` to `np.clongdouble`. I tried to use `npy_longdouble_from_PyLong` to handle the case of converting from a Python int to `np.clongdouble`, but that change causes `test_nep50_weak_integers_with_inexact` to fail.

benburrill:
That test specifically excludes `longdouble` (notice it tests "dDG", but not "g"), which has the same int->string issue, so it would probably fine to just change the test to exclude "G" as well.  That's not an ideal solution, I think the test should be rewritten so that if `npy_longdouble_from_PyLong` is changed in the future to go directly from int->longdouble, the test should either test what it was supposed to test, or xpass or something, but doing the same thing for `clongdouble` as is done for `longdouble` seems fine to me.

<sup>It's kinda funny because IIRC running the numpy tests is what was used to determine how small of a limit they could get away with for integer string conversion, and yet the chosen limit is still is a problem even for the numpy tests...</sup>

MyUserNameWasTakenLinux:
I've added a few tests that cover some of the conversion cases. Please let me know if there is anything you'd like me to change.

Something I don't understand is why `x[0] = y` from
```python   
def test_from_zerodim_array(self):
        x = np.zeros(1, dtype=np.clongdouble)
        py_val = 2.0 + 3.0j
        y = np.array(np.clongdouble(py_val))
        x[0] = y
        assert np.allclose(complex(x[0]), py_val)
```
doesn't pass through the `PyArray_IsZeroDim(op)` branch. Instead it goes through `PyArray_IsScalar(op, CLongDouble)`.  Is it possible that the `PyArray_IsZeroDim` branch is never taken?

ngoldbaum:
> Is it possible that the PyArray_IsZeroDim branch is never taken?

Maybe? You could try putting a breakpoint inside that branch with a C debugger and see if the breakpoint ever gets hit. Although if the tests don't cover that branch then this exercise wouldn't prove anythig - maybe test coverage is bad? But if it is covered then that might give you more context if you look at the call stack.

I recently added docs for using C debuggers with NumPy's internals, you can try it out: https://numpy.org/devdocs/dev/development_advanced_debugging.html#c-debuggers

numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -620,6 +621,206 @@ LONGDOUBLE_setitem(PyObject *op, void *ov, void *vap)
...
+    if (s) {
+        npy_bool has_brackets = NPY_FALSE;
+        start = s;
+
+        while (isspace((unsigned char) *s)) {
```
seberg, line 662: Hmmm, we'll lose the ability to parse certain strings in principle that Python parses (e.g. non-breaking space).  But `clongdouble` currently doesn't parse these either.  So maybe OK.

numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -620,6 +621,206 @@ LONGDOUBLE_setitem(PyObject *op, void *ov, void *vap)
...
+            }
+            else {
+                if (*p != '\0') {
+                    goto parse_failure;
+                }
+                else {
+                    Py_XDECREF(b);
+                    return temp;
+                }   
+            }
```
seberg, line 711: This is a huge repeated block, and not very clear that it is the finalization and will never continue below.  Even a goto would seem better, I think just put what is after this into the if instead.

numpy/_core/tests/test_scalarmath.py
```diff
@@ -1069,6 +1069,57 @@ def test_longdouble_complex():
...
+        py_val = 42
+        x[0] = py_val
+        assert x[0].real == np.longdouble(py_val)
+        assert x[0].imag == 0
+
+    @pytest.mark.parametrize("bad_str", [
+        "not a number",
+        "(2+2j"  # Missing closing brackets
+        "2+5j )"  # Missing opening brackets
+        "3+4k"
```
seberg: I think this could have more tests, e.g. `3j+3`, `3+-1j`?  Also maybe unicode characters for example (but that is different test maybe).

numpy/_core/src/multiarray/arraytypes.c.src
seberg: Supporting it seems easy enough, I would lean towards just keeping it if it's simple and adding a test that will fail if Python ever breaks it.
But if you don't, at least add a test for it raising!

seberg (commented):
Overall, looks reasonable, I think a few more tests would be good, this type of code is deceptive.  But maybe Nathan or Warren have a look through again, then I am happy (with that code duplication removed, it's a lot of lines and it confused me).

MyUserNameWasTakenLinux:
Sorry for the long delay, I've removed the code duplication and added some tests for invalid strings.

ngoldbaum (approved):
Sorry for letting this sit for so long! I'll also add a milestone while will help us avoid forgetting.

PR #29749, opened 2025-09-13
TYP: mypy 1.18.1

jorenham:
https://mypy.readthedocs.io/en/stable/changelog.html#mypy-1-18

It's should be significantly faster now (~0.67x in case of scipy-stubs), and solves some issues that used to lead to some incorrectly inferred types in our type-tests (e.g. `np.vstack`)


charris:
Thanks Joren.

PR #29208, opened 2025-06-13
ENH: improve Timsort with powersort merge-policy

moritz-gross:
## our contribution

This PR implements [Powersort](https://en.wikipedia.org/wiki/Powersort), a sorting algorithm based on the currently used Timsort that has already been implemented in [CPython](https://github.com/python/cpython/issues/78742) and [PyPy](https://foss.heptapod.net/pypy/pypy/-/blob/branch/default/rpython/rlib/listsort.py).

We benchmarked the implementation on [collected data](https://powersort-competition.github.io/PowersortCompetitionWebsite/#/aboutA) and archieved a strong improvement (in particular on large instances with length over 1 million):
plot of benchmark results 

Powersort is a drop-in replacement with the same functionality as Timsort, which we tested with `spin test -m full` (though only `spin test -- -k "test_multiarray"` should be relevant).

## open questions
- So far, the only implemented data type is [numeric sorts](https://github.com/numpy/numpy/blob/aa4f974c2f9905ac05b395a35bf14ea4d0fb9499/numpy/_core/src/npysort/timsort.cpp#L92), i.e. _string_ and _generic_ are not implemented yet, as well as _argsort_ respectively. Should we add this in this PR?

PR #29751, opened 2025-09-13
ENH: implement powersort merge-policy for argsort

moritz-gross:
This pull request follows up on the [previous accepted pull request](https://github.com/numpy/numpy/pull/29208) for numeric **_inplace_** sorting. Here we bring Powersort for numeric **_argsort_** as well.

On the same [sample data](https://powersort-competition.github.io/PowersortCompetitionWebsite/#/aboutA) as before, we analyzed the improvement of Powersort over Timsort, both for the existing inplace-sorting as well as for argsort. The improvement for argsort is slightly weaker but more concentrated. Slow instances show the biggest change: for the biggest quartile, Timsort on average is 10.7% slower for inplace sorting and 5.7% for argsorting.

For both Timsort and Powersort, argsort is consistently slower than inplace-sorting, by an average factor of 1.80 and 1.84 respectively. This behavior was also already suggested in the review to the last pull request. It is also a likely explanation for the weaker yet more concentrated improvement, as the merge order has less share of the overall computation due to the additional level of indirection.

image

### DIFF
--- a/numpy/_core/src/npysort/timsort.cpp
+++ b/numpy/_core/src/npysort/timsort.cpp
@@ -782,59 +782,24 @@ amerge_at_(type *arr, npy_intp *tosort, const run *stack, const npy_intp at,
 
 template <typename Tag, typename type>
 static int
-atry_collapse_(type *arr, npy_intp *tosort, run *stack, npy_intp *stack_ptr,
-               buffer_intp *buffer)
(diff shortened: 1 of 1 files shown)


jorenham:
What data did you use to test this?

charris:
Thanks @moritz-gross .

PR #29753, opened 2025-09-14
DOC: Fix typo in absolute_beginners.rst

phoenixstudiodz:
Fix a typo in the documentation (match the illustration).

image

```
data = np.array([1.0, 2.0, 3.0])
>>> data.max()
2.0 # this should be 3.0
>>> data.min()
1.0
>>> data.sum()
3.0 # this should be 6.0
```

### DIFF
--- a/doc/source/user/absolute_beginners.rst
+++ b/doc/source/user/absolute_beginners.rst
@@ -863,12 +863,13 @@ NumPy also performs aggregation functions. In addition to ``min``, ``max``, and
 result of multiplying the elements together, ``std`` to get the standard
 deviation, and more. ::
 
+  >>> data = np.array([1, 2, 3])
   >>> data.max()
-  2.0
+  3
   >>> data.min()
-  1.0
+  1
   >>> data.sum()
-  3.0
+  6
 
 .. image:: images/np_aggregation.png
 


doc/source/user/absolute_beginners.rst
```diff
@@ -864,11 +864,11 @@ result of multiplying the elements together, ``std`` to get the standard
 deviation, and more. ::
 
   >>> data.max()
```
mattip, line 867: On line 838, in the section about Broadcasting, data is redefined `data = np.array([1.0, 2.0])`. So the text is correct, the image is wrong. I think the easiest thing would be to redefine `data` here to match the image (which will restore it to the same values as in the "Indexing and slicing" section):

```suggestion
  >>> data = np.array([1, 2, 3])
  
  >>> data.max()
```

Then at some point someone could regenerate the image in the "Broadcasting" section and change the text so that the value of `data` remains consistent throughout this page. When they do that, it would be nice to add a `.. note` how they created the image.

melissawm:
Docs build is failing because of the output:

```
868   >>> data.max()
Expected:
    3.0
Got:
    np.int64(3)
```

melissawm (approved):

melissawm:
Thanks, @phoenixstudiodz !

PR #29754, opened 2025-09-14
MAINT: pin asv<0.6.5 

mattip:
ASV had a recent release, and it seems to have broken something in NumPy. I cannot reproduce locally, but maybe I am not being exact enough. In any case, let's pin this until someone can figure out what is broken.


mattip:
[CI is passing](https://github.com/numpy/numpy/actions/runs/17713063344/job/50334583798?pr=29754).

charris:
Thanks Matti.

PR #29756, opened 2025-09-14
BLD/CI: build macOS x86-64 wheel on macos-14 as a cross build

### LINKED ISSUE #29728: CI: macos-13 runner image deprecated, drops macOS x86-64 support
We'll need to stop using this image in CI very soon, because GHA will start doing its normal deprecation thing with intermittently making the runners unavailable to alert users to the deprecation.

I [posted on the mailing list](https://mail.python.org/archives/list/numpy-discussion@python.org/thread/OLNBDCRVVGXIXG5G3RUQDHEIYDHYV2DO/) about potentially dropping support for x86-64 wheels on PyPI.

rgommers:
The macos-13 image is deprecated, meaning GitHub Actions no longer offers x86-64 runners. Hence we need to cross compile. Cibuildwheel runs the tests under Rosetta2, and that works fine once we skip a couple of problematic cases.

A few of the changes in the macOS section were simple cleanups to bring things back in sync with the `numpy-release` repo. E.g., the F77/F90 env vars were distutils-specific and no longer used, and the `DYLD_LIBRARY_PATH` usage was also unnecessary.

Closes gh-29728

_Opened as Draft PR because the `.pc` file handling needs cleaning up and it's not yet 100% clear if `cibuildwheel` won't change how this works (xref https://github.com/pypa/cibuildwheel/issues/2592)._

### DIFF
--- a/tools/ci/scipy-openblas-macos-x86-64.pc
+++ b/tools/ci/scipy-openblas-macos-x86-64.pc
@@ -1,0 +1,12 @@
+libdir=/Users/runner/work/numpy/numpy/host-env/scipy_openblas64/lib
+includedir=/Users/runner/work/numpy/numpy/host-env/scipy_openblas64/include
+openblas_config=Dont-trust-this-because-hardcoded-pcfile
+version=0.3.30
+extralib=-lm -lpthread -lgfortran -lquadmath -L${libdir} -lscipy_openblas64_
+Name: openblas
+Description: OpenBLAS is an optimized BLAS library based on GotoBLAS2 1.13 BSD version
+Version: ${version}
+URL: https://github.com/xianyi/OpenBLAS
+Libs: ${libdir}/libscipy_openblas64_.dylib -Wl,-rpath,${libdir}
+Libs.private: ${extralib}
+Cflags: -I${includedir} -DBLAS_SYMBOL_PREFIX=scipy_ -DBLAS_SYMBOL_SUFFIX=64_ -DHAVE_BLAS_ILP64 -DOPENBLAS_ILP64_NAMING_SCHEME
--- a/tools/wheels/cibw_before_build.sh
+++ b/tools/wheels/cibw_before_build.sh
@@ -32,8 +32,17 @@ if [[ "$INSTALL_OPENBLAS" = "true" ]] ; then
     echo pkgconf_path is $pkgconf_path, OPENBLAS is ${OPENBLAS}
     rm -rf $pkgconf_path
     mkdir -p $pkgconf_path
-    python -m pip install -r $PROJECT_DIR/requirements/ci_requirements.txt
-    python -c "import scipy_${OPENBLAS}; print(scipy_${OPENBLAS}.get_pkg_config())" > $pkgconf_path/scipy-openblas.pc
+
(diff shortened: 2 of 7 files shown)


tools/wheels/cibw_before_build.sh
```diff
@@ -32,8 +32,17 @@ if [[ "$INSTALL_OPENBLAS" = "true" ]] ; then
...
     rm -rf $pkgconf_path
     mkdir -p $pkgconf_path
-    python -m pip install -r $PROJECT_DIR/requirements/ci_requirements.txt
-    python -c "import scipy_${OPENBLAS}; print(scipy_${OPENBLAS}.get_pkg_config())" > $pkgconf_path/scipy-openblas.pc
+
+    if [[ $CIBW_ARCHS_MACOS == "x86_64" ]]; then
+        # We're cross compiling; the before-build hook isn't using Rosetta (see cibuildwheel#2592)
+        mkdir host-env
+        python -m pip install -r $PROJECT_DIR/requirements/ci_requirements.txt --platform macosx_10_13_x86_64 --only-binary :all: -U --target ./host-env
+	# Use a handwritten .pc file, because we can't run the cross Python to generate it
```
rgommers, line 40: tab instead of spaces - needs fixing

tools/ci/scipy-openblas-macos-x86-64.pc
```diff
@@ -0,0 +1,12 @@
+libdir=/Users/runner/work/numpy/numpy/host-env/scipy_openblas64/lib
```
rgommers, line 1: This file should be generated by a script. The problem with `scipy-openblas` is that it has a function to generate the `.pc` file, but it relies on `dlopen`-ing the shared library, which won't work for cross builds.

andyfaff:
@rgommers please accept my apology, I thought it would be easier to do that this. cibuildwheel takes care of this for me for my personal project, but that doesn't use meson/etc and is a lot simpler.

rgommers:
@andyfaff no worries at all, no need to apologize. If things "just work" it's natural to assume it will do so for `numpy` too. It took me a little while to even understand what `cibuildwheel` is doing, since its docs don't explain it and I thought it was doing a native emulated build rather than a cross build. Either way, we'd have some troubles with it though. And SciPy will too - I think a cross build is actually nicer for SciPy.

agriyakhetarpal:
@rgommers, in https://github.com/pypa/cibuildwheel/pull/2601 I've changed the images to use the new `macos-15-intel` ones, so that we can perform native macOS Intel builds for some more time. I hope that helps!

P.S. Probably by the time GitHub deprecates support for these images (a couple of years, perhaps), support for the platform in general would be at its twilight. I suppose the cross-compilation experiment here is still worth trying, of course – for other platform/architecture combos, if not macOS ARM ➡️ macOS Intel.

andyfaff:
MacOS Intel runners on 15, awesome. Didn't know about that.

andyfaff:
@rgommers, I think this can be closed as the macos-15-intel runners should be able to build the wheel. #29886 is effectively a (native) superset of this PR which should be merged instead of this.

I think we should stop making `macosx-x86_64` when the macos-15-intel image is retired.

rgommers:
Agreed - great news on the new runner images, so we can close this.

PR #29761, opened 2025-09-16
BUG: Stable ``ScalarType`` ordering

jorenham:
The key used to sort `np.ScalarType` is not unique (e.g. `int64` and `longlong` on linux both have key `('i', '8')`). To remove duplicates, `set` is used, whose iteration order depends on the hashes of the containing scalartype classes. So in case of ties, the hash would be the tiebreaker. The problem is that these hashes are arbitrary, leading to differences in `ScalarType` order between platforms, python versions, machines, etc.

In case you're wondering why I care: Stubtest is currently reporting errors for `ScalarType` for certain platform and python versions: https://github.com/numpy/numtype/actions/runs/17778048694/job/50530549124?pr=706. And I'm sure you can imagine that it wasn't easy to figure out what was going on here (especially if you consider the stubtest error spaghetti output, but I digress). Hence the pedantic "backport candidate" label ;)

### DIFF
--- a/numpy/_core/tests/test_numerictypes.py
+++ b/numpy/_core/tests/test_numerictypes.py
@@ -616,6 +616,35 @@ class TestScalarTypeNames:
         assert np.dtype(t.__name__).type is t
 
 
+class TestScalarTypeOrder:
+    @pytest.mark.parametrize(('a', 'b'), [
+        # signedinteger
+        (np.byte, np.short),
+        (np.short, np.intc),
+        (np.intc, np.long),
+        (np.long, np.longlong),
+        # unsignedinteger
+        (np.ubyte, np.ushort),
+        (np.ushort, np.uintc),
+        (np.uintc, np.ulong),
+        (np.ulong, np.ulonglong),
+        # floating
+        (np.half, np.single),
+        (np.single, np.double),
+        (np.double, np.longdouble),
+        # complexfloating
+        (np.csingle, np.cdouble),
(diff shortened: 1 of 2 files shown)


ngoldbaum:
Can we have a test that asserts the hard-coded sorted order?

numpy/_core/tests/test_numerictypes.py
```diff
@@ -616,6 +616,35 @@ def test_names_are_undersood_by_dtype(self, t):
...
+        (np.single, np.double),
+        (np.double, np.longdouble),
+        # complexfloating
+        (np.csingle, np.cdouble),
+        (np.cdouble, np.clongdouble),
+        # flexible
+        (np.bytes_, np.str_),
+        (np.str_, np.void),
+        # bouncy castles
+        (np.datetime64, np.timedelta64),
```
ngoldbaum, line 642: Total quibble but I think you can write this with just a single static list and then generate these tuples with e.g. `itertools.pairwise`.

ngoldbaum (approved):

numpy/_core/tests/test_numerictypes.py
jorenham: I wanted to do that at first. But I then decided against that because I didn't want to set the (IMO arbitrary) "kind" group ordering in stone (bool < complexfloating < np.floating < np.signedinteger < bouncy castles < object < flexible < unsignedinteger).

charris:
Thanks Joren.

PR #29737, opened 2025-09-10
ENH, API: New sorting slots for DType API

MaanasArora:
Supersedes #28516 after @charris defined the new C-API for `PyArray_Sort` in #29642.

I have kept this light so far, with a sorting table (struct) for the DType API, using the new signature with the context. The new design is untested so far because without `sort_compare`, using it directly in say StringDType would be difficult, but happy to move this to parity if this looks right so far considering Chuck's comments. Thank you!

PR #29763, opened 2025-09-17
Feat: Adding support of `np.finfo` to the DType C API

### LINKED ISSUE #27231: ENH: `np.finfo` support for Numpy User DTypes

Require the support for integrating User DTypes with `numpy.finfo` as currently it raises the ValueError
```
In [2]: np.finfo(QuadPrecDType)
Cell In[2], line 1
----> 1 np.finfo(QuadPrecDType)

File ~/Desktop/numpy-user-dtypes/quaddtype/.venv/lib/python3.11/site-packages/numpy/_core/getlimits.py:519, in finfo.__new__(cls, dtype)
    517     dtype = newdtype
    518 if not issubclass(dtype, numeric.inexact):
--> 519     raise ValueError("data type %r not inexact" % (dtype))
    520 obj = cls._finfo_cache.get(dtype)
    521 if obj is not None:

ValueError: data type <class 'numpy.object_'> not inexact
```

SwayamInSync:
closes #27231

This PR contibutes as follows:
- Adds a new slot ` NPY_DT_get_finfo`, allowing user to define the custom finfo object which can be used by the `np.finfo`
- Adds 2 `arraydescr_methods` 
    - `_get_finfo`: This available the finfo object coming from slot
    - `_is_user_dtype` This checks whether the dtype is user-defined or not
    
Subclassing inexact isn't an issue here, as it can be done by subclassing the scalar type to `PyFloatingArrType_Type`

 ## Usage
 ```cpp
static PyObject *
custom_get_dtype_info(PyArray_Descr *descr, NPY_DTypeInfo_Type info_type)
{    
   switch (info_type) {
       case NPY_DTYPE_INFO_FLOAT:
           {
               PyObject *finfo_dict = PyDict_New();
               if (!finfo_dict) return NULL;
               
               PyDict_SetItemString(finfo_dict, "precision", PyLong_FromLong(34));
               PyDict_SetItemString(finfo_dict, "bits", PyLong_FromLong(128));
               //  more fields
               return finfo_dict;
           }
       case NPY_DTYPE_INFO_INTEGER:
           // Not implemented yet, could add iinfo support later
           PyErr_SetString(PyExc_NotImplementedError, 
                          "Integer info not implemented for this dtype");
           return NULL;
       case NPY_DTYPE_INFO_GENERIC:
           // Not implemented yet, could add generic info later
           PyErr_SetString(PyExc_NotImplementedError, 
                          "Generic info not implemented for this dtype");
           return NULL;
       default:
           PyErr_SetString(PyExc_ValueError, "Unknown dtype info type");
           return NULL;
   }
}

static PyType_Slot CustomDType_Slots[] = {
       {NPY_DT_get_dtype_info, &custom_get_dtype_info},
       {0, NULL}
}
```

### DIFF
--- a/numpy/_core/include/numpy/dtype_api.h
+++ b/numpy/_core/include/numpy/dtype_api.h
@@ -379,7 +379,7 @@ typedef int (PyArrayMethod_PromoterFunction)(PyObject *ufunc,
 #define NPY_DT_get_clear_loop 9
 #define NPY_DT_get_fill_zero_loop 10
 #define NPY_DT_finalize_descr 11
-#define NPY_DT_get_finfo 12
+#define NPY_DT_get_dtype_info 12
 
 // These PyArray_ArrFunc slots will be deprecated and replaced eventually
 // getitem and setitem can be defined as a performance optimization;
@@ -491,10 +491,27 @@ typedef int(PyArrayDTypeMeta_SetItem)(PyArray_Descr *, PyObject *, char *);
 typedef PyObject *(PyArrayDTypeMeta_GetItem)(PyArray_Descr *, char *);
 
 /*
- * Function to compute finfo for a custom dtype.
- * Should return a Python object with finfo attributes (precision, eps, max, etc.)
- * or NULL to fall back to default behavior.
+ * Enum for different types of dtype information that can be requested.
  */
-typedef PyObject *(PyArrayDTypeMeta_GetFinfo)(PyArray_Descr *);
+typedef enum {
+    NPY_DTYPE_INFO_FLOAT = 0,    /* For finfo() - floating point parameters */
+    NPY_DTYPE_INFO_INTEGER = 1,  /* For iinfo() - integer parameters (not implemented) */
+    NPY_DTYPE_INFO_GENERIC = 2   /* For generic dtype info (not implemented) */
+} NPY_DTYPE_INFO_TYPE;
+
+/*
+ * Function to compute dtype information for custom dtypes.
+ * 
+ * @param descr: The dtype descriptor
+ * @param info_type: The type of information requested (float, integer, etc.)
+ * 
+ * Should return a Python object with appropriate attributes based on info_type:
+ * - For NPY_DTYPE_INFO_FLOAT: finfo attributes (precision, eps, max, etc.)
+ * - For NPY_DTYPE_INFO_INTEGER: iinfo attributes (min, max, etc.) - not implemented yet
+ * - For NPY_DTYPE_INFO_GENERIC: generic dtype info - not implemented yet
+ * 
+ * Returns NULL to fall back to default behavior or on error.
+ */
+typedef PyObject *(PyArrayDTypeMeta_GetDTypeInfo)(PyArray_Descr *, NPY_DTYPE_INFO_TYPE);
 
 #endif  /* NUMPY_CORE_INCLUDE_NUMPY___DTYPE_API_H_ */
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by SwayamInSync** [comment]
cc: @seberg @ngoldbaum

File: numpy/_core/include/numpy/dtype_api.h [hunk 1]
```diff
@@ -477,4 +478,11 @@ typedef PyArray_Descr *(PyArrayDTypeMeta_FinalizeDescriptor)(PyArray_Descr *dtyp
 typedef int(PyArrayDTypeMeta_SetItem)(PyArray_Descr *, PyObject *, char *);
 typedef PyObject *(PyArrayDTypeMeta_GetItem)(PyArray_Descr *, char *);
 
+/*
+ * Function to compute finfo for a custom dtype.
+ * Should return a Python object with finfo attributes (precision, eps, max, etc.)
+ * or NULL to fall back to default behavior.
+ */
+typedef PyObject *(PyArrayDTypeMeta_GetFinfo)(PyArray_Descr *);
```
ngoldbaum: Before this is merged it's gonna need new docs and a release note - see #29737 for a recent worked example. It'd also be great if you could look that over and give feedback if you have any.

ngoldbaum (commented):
Nice, this looks reasonable. Maybe we should call it `GetDTypeInfo` and have it accept an enum that could be integer, float, generic, but leave everything except for float unimplemented for now?

Just trying to think of if someone wants to do the same thing for info about integers or other numeric types that aren't necessarily floats.

numpy/_core/include/numpy/dtype_api.h
SwayamInSync: documentation writing 🥲
Will go through this as well

SwayamInSync:
@ngoldbaum made the changes to also take an enum input.
Also updated the PR's Usage section

SwayamInSync:
I think spin is causing issue here as well with click's new release
Same issue on scipy: https://github.com/scipy/scipy/issues/23642
PR fix: https://github.com/scipy/scipy/pull/23646

SwayamInSync:
@ngoldbaum new commits didn't change the code, only added docs so you can take a look and let me know if any changes required

ngoldbaum:
https://github.com/scientific-python/spin/issues/293 for the spin issues. Hopefully we'll have a new release up shortly with a fix.

I think the discussion on the mailing list is more leaning towards not adding a new DType slot and instead doing it in a different way, probably via some kind of registration like Marten brought up. Couldn't we just add a Python API hook that a DType can call during initialization to register itself as a floating point dtype that has information it wants to add to `finfo`? I don't think it's necessary to write a full ABC (probably would have to be in C) to get that set up, but if you're up for a bigger project then I think it would be really interesting and useful to more fully working out exactly how to set up a hierarchy of numeric dtype classes with suitable properties and whatnot that people can subclass to automatically register with NumPy and then appear as a "fully supported" dtype in the public NumPy API.

So, sorry that there might need to be a second revision on this. That's one reason it's a good idea to gather feedback on a new feature before doing the work.

SwayamInSync:
Hey no issue @ngoldbaum (I will just put this in a different branch, just for sake of bookkeeping)

> I think it would be really interesting and useful to more fully working out exactly how to set up a hierarchy of numeric dtype classes with suitable properties and whatnot that people can subclass to automatically register with NumPy and then appear as a "fully supported" dtype in the public NumPy API.

Yup this can one shot other issues we've dealing and I indeed want to tackle this, I will plan thoroughly and draft a PR + announcing on mailing list.

For now just rounding this to 
- Adding `finfo` class to the python interface of user-defined dtype 
- NumPy's `getlimits.py` (need to update) will use it, if the attribute is available and `dtype.type` is subclassing `np.inexact`

SwayamInSync:
closing this, will do a new PR simple fix

PR #29766, opened 2025-09-18
ENH, SIMD: Optimize the arithm fp implementation based on Highway Wra…

ixgbe:
Test command: spin bench --compare HEAD~1 -t bench_ufunc.ADD_SUB_MUL_DIV -t bench_ufunc.CONJUGATE_SQUARE_COMPLEX

Add class ADD_SUB_MUL_DIV and CONJUGATE_SQUARE_COMPLEX in benchmarks/benchmarks/bench_ufunc.py for this test on Banana Pi BPI-F3

```
class ADD_SUB_MUL_DIV(Benchmark):
    timeout = 10
    params = ['float32', 'float64', 'complex64', 'complex128']
    param_names = ['dtype']

    def setup(self, dtype):
        np.seterr(all='ignore')
        self.add_fn = np.add
        self.subtract_fn = np.subtract
        self.multiply_fn = np.multiply
        self.divide_fn = np.divide

        from .common import get_square_
        arr1 = get_square_(dtype)
        arr2 = get_square_(dtype).copy()

        self.data_add = (arr1, arr2)
        self.data_subtract = (arr1, arr2)
        self.data_multiply = (arr1, arr2)
        self.data_divide = (arr1, arr2)

    def time_add(self, dtype):
        self.add_fn(*self.data_add)

    def time_subtract(self, dtype):
        self.subtract_fn(*self.data_subtract)

    def time_multiply(self, dtype):
        self.multiply_fn(*self.data_multiply)

    def time_divide(self, dtype):
        self.divide_fn(*self.data_divide)


class CONJUGATE_SQUARE_COMPLEX(Benchmark):
    timeout = 10
    params = ['complex64', 'complex128']
    param_names = ['dtype']

    def setup(self, dtype):
        np.seterr(all='ignore')
        self.conjugate_fn = np.conjugate
        self.square_fn = np.square

        from .common import get_square_
        arr1 = get_square_(dtype)

        self.data_conjugate = (arr1,)
        self.data_square = (arr1,)

    def time_conjugate(self, dtype):
        self.conjugate_fn(*self.data_conjugate)

    def time_square(self, dtype):
        self.square_fn(*self.data_square)

```

(numpy_venv) root@yangwang:/home/yangwang/numpy# spin bench --compare HEAD~1 -t bench_ufunc.ADD_SUB_MUL_DIV -t bench_ufunc.CONJUGATE_SQUARE_COMPLEX
| Change   | Before [404b35ed] <main~1>   | After [ecab0fcc] <main>   |   Ratio | Benchmark (Parameter)                                          |
|----------|------------------------------|---------------------------|---------|----------------------------------------------------------------|
| -        | 165±4μs                      | 155±2μs                   |    0.94 | bench_ufunc.ADD_SUB_MUL_DIV.time_add('float64')                |
| -        | 247±5μs                      | 219±4μs                   |    0.89 | bench_ufunc.CONJUGATE_SQUARE_COMPLEX.time_square('complex128') |
| -        | 177±10μs                     | 157±4μs                   |    0.88 | bench_ufunc.ADD_SUB_MUL_DIV.time_multiply('float64')           |
| -        | 273±10μs                     | 233±4μs                   |    0.85 | bench_ufunc.ADD_SUB_MUL_DIV.time_add('complex128')             |
| -        | 311±20μs                     | 259±7μs                   |    0.83 | bench_ufunc.ADD_SUB_MUL_DIV.time_subtract('complex128')        |
| -        | 206±4μs                      | 169±4μs                   |    0.82 | bench_ufunc.ADD_SUB_MUL_DIV.time_subtract('float64')           |
| -        | 123±6μs                      | 99.0±7μs                  |    0.8  | bench_ufunc.ADD_SUB_MUL_DIV.time_add('float32')                |
| -        | 244±20μs                     | 195±4μs                   |    0.8  | bench_ufunc.ADD_SUB_MUL_DIV.time_multiply('complex64')         |
| -        | 243±2μs                      | 177±8μs                   |    0.73 | bench_ufunc.ADD_SUB_MUL_DIV.time_subtract('complex64')         |
| -        | 226±8μs                      | 163±4μs                   |    0.72 | bench_ufunc.ADD_SUB_MUL_DIV.time_add('complex64')              |
| -        | 138±5μs                      | 98.2±7μs                  |    0.71 | bench_ufunc.ADD_SUB_MUL_DIV.time_subtract('float32')           |
| -        | 226±8μs                      | 153±9μs                   |    0.68 | bench_ufunc.CONJUGATE_SQUARE_COMPLEX.time_square('complex64')  |
| -        | 378±20μs                     | 235±9μs                   |    0.62 | bench_ufunc.ADD_SUB_MUL_DIV.time_divide('float64')             |
| -        | 241±0.6μs                    | 139±4μs                   |    0.58 | bench_ufunc.ADD_SUB_MUL_DIV.time_divide('float32')             |


numpy/_core/src/umath/loops_arithm_fp.dispatch.cpp
```diff
@@ -0,0 +1,766 @@
+#include "simd/simd.h"
+#include "loops_utils.h"
+#include "loops.h"
+#include "simd/simd.hpp"
+#include <hwy/highway.h>
```
seiko2plus:
```suggestion
```
Disabling optimizations at build time should still let NumPy build without Google Highway and wrapper already handles this.

numpy/_core/src/umath/loops_arithm_fp.dispatch.cpp
```diff
@@ -0,0 +1,766 @@
...
+#include <limits>   // Required for std::numeric_limits
+
+namespace {
+using namespace np::simd;
+
+template <typename T>
+struct OpAdd {
+#if NPY_HWY
+    template <typename V, typename = std::enable_if_t<kSupportLane<T>>>
+    HWY_INLINE HWY_ATTR auto operator()(const V &a, const V &b) const {
```
seiko2plus:
```suggestion
    NPY_FINLINE V operator()(const V &a, const V &b) const {
```
No need for `HWY_ATTR` here. We already use our own source-based CPU dispatcher. Even if we switch to Highway’s dispatcher (or another custom one), it would be through `#pragma` or Highway’s `BEGIN/AFTER` macros, which handle CPU target attributes for the whole definition scope.

numpy/_core/src/umath/loops_arithm_fp.dispatch.cpp
```diff
@@ -0,0 +1,766 @@
...
+        return hn::Add(a, b);
+    }
+#endif
+    HWY_INLINE T operator()(T a, T b) const {
+        return a + b; 
+    }
+    static constexpr int PW = 1;
+    static constexpr bool is_div = false;
+    static constexpr bool is_mul = false;
```
seiko2plus:
```suggestion
```
You can just test the operation class itself with `std::is_same_v`.

numpy/_core/src/umath/loops_arithm_fp.dispatch.cpp
```diff
@@ -0,0 +1,766 @@
...
+
+template <typename T>
+struct OpAdd {
+#if NPY_HWY
+    template <typename V, typename = std::enable_if_t<kSupportLane<T>>>
+    HWY_INLINE HWY_ATTR auto operator()(const V &a, const V &b) const {
+        return hn::Add(a, b);
+    }
+#endif
+    HWY_INLINE T operator()(T a, T b) const {
```
seiko2plus:
```suggestion
    NPY_FINLINE T operator()(T a, T b) const {
```
Use `NPY_FINLINE` instead of `HWY_INLINE` for the same reason mentioned above.

numpy/_core/src/umath/loops_arithm_fp.dispatch.cpp
```diff
@@ -0,0 +1,766 @@
...
+real_single_double_func(char **args, npy_intp const *dimensions, npy_intp const *steps)
+{
+    Op op_func;
+    npy_intp len = dimensions[0];
+    char *src0 = args[0], *src1 = args[1], *dst = args[2];
+    npy_intp ssrc0 = steps[0], ssrc1 = steps[1], sdst = steps[2];
+
+    // reduce
+    if (ssrc0 == 0 && ssrc0 == sdst && src0 == dst) {
+        if constexpr (Op::PW) {
```
seiko2plus:
```suggestion
        if constexpr (std::same_v<Op, OpAdd<T>>) {
```
Like this

numpy/_core/src/umath/loops_arithm_fp.dispatch.cpp
```diff
@@ -0,0 +1,766 @@
...
+                for (; len >= lstep; len -= lstep, src0 += wstep, dst += wstep) {
+                    auto a0 = LoadU<T>((const T*)src0);
+                    auto a1 = LoadU<T>((const T*)(src0 + vstep));
+                    auto r0 = op_func(a0, b);
+                    auto r1 = op_func(a1, b);
+                    StoreU<T>(r0, (T*)dst);
+                    StoreU<T>(r1, (T*)(dst + vstep));
+                }
+                for (; len > 0; len -= hstep, src0 += vstep, dst += vstep) {
+                    if constexpr (Op::is_mul) {
```
seiko2plus:
```suggestion
                    if constexpr (std::is_same_v<Op, OpMul<T>>) {
```

Just rely on the class type instead.

numpy/_core/src/umath/loops_arithm_fp.dispatch.cpp
```diff
@@ -0,0 +1,766 @@
...
+     *
+     * - The precision may vary between the emulated SIMD and scalar division due to
+     *   non-uniform branches (non-contiguous) in the code, leading to precision
+     *   inconsistencies.
+     *
+     * - Considering the necessity of multiple Newton-Raphson iterations, the performance
+     *   gain may not sufficiently offset these drawbacks.
+     */
+
+#if NPY_HWY
```
seiko2plus, line 134:
```suggestion
#if NPY_HWY
     using V = Vec<T>;
```
Use `V` instead of `auto` for readability.

numpy/_core/src/umath/loops_arithm_fp.dispatch.cpp
```diff
@@ -0,0 +1,766 @@
...
+                    auto b1 = LoadU<T>((const T*)(src1 + vstep));
+                    auto r0 = op_func(a0, b0);
+                    auto r1 = op_func(a1, b1);
+                    StoreU<T>(r0, (T*)dst);
+                    StoreU<T>(r1, (T*)(dst + vstep));
+                }
+                for (; len > 0; len -= hstep, src0 += vstep, src1 += vstep, dst += vstep) {
+                    if constexpr(Op::is_div){
+                        auto a = hn::LoadNOr(Set<T>(1.0), _Tag<T>(), (const T*)src0, len);
+                        auto b = hn::LoadNOr(Set<T>(1.0), _Tag<T>(), (const T*)src1, len);
```
seiko2plus:
```suggestion
                        V a = LoadNOr(Set<T>(1.0), t_src0, len);
                        V b = LoadNOr(Set<T>(1.0), t_src1, len);
```
Expand the wrapper instead, and define it under:
https://github.com/numpy/numpy/blob/5f800455a0fb86d4fd33cd93336996e4b00f0414/numpy/_core/src/common/simd/simd.inc.hpp#L47-L53

numpy/_core/src/umath/loops_arithm_fp.dispatch.cpp
```diff
@@ -0,0 +1,766 @@
...
+                for (; len >= lstep; len -= lstep, src0 += wstep, src1 += wstep, dst += wstep) {
+                    auto a0 = LoadU<T>((const T*)src0);
+                    auto a1 = LoadU<T>((const T*)(src0 + vstep));
+                    auto b0 = LoadU<T>((const T*)src1);
+                    auto b1 = LoadU<T>((const T*)(src1 + vstep));
+                    auto r0 = op_func(a0, b0);
+                    auto r1 = op_func(a1, b1);
+                    StoreU<T>(r0, (T*)dst);
+                    StoreU<T>(r1, (T*)(dst + vstep));
+                }
```
seiko2plus:
```suggestion
            const T *t_src0 = reinterpret_cast<const T*>(src0);
            const T *t_src1 = reinterpret_cast<const T*>(src1);
            T *t_dst = reinterpret_cast<T*>(dst);
            HWY_LANES_CONSTEXPR int vstep = Lanes<T>();
            const int wstep = vstep * 2;
            // lots of specializations, to squeeze out max performance
            if (ssrc0 == sizeof(T) && ssrc0 == ssrc1 && ssrc0 == sdst) {
                for (; len >= wstep; len -= wstep, t_src0 += wstep, t_src1 += wstep, t_dst += wstep) {
                    V a0 = LoadU(t_src0);
                    V a1 = LoadU(t_src0 + vstep);
                    V b0 = LoadU(t_src1);
                    V b1 = LoadU(t_src1 + vstep);
                    V r0 = op_func(a0, b0);
                    V r1 = op_func(a1, b1);
                    StoreU(r0, t_dst);
                    StoreU(r1, t_dst + vstep);
                }
```
Iterate over the pointer type instead of bytes — much simpler.

numpy/_core/src/umath/loops_arithm_fp.dispatch.cpp
```diff
@@ -0,0 +1,766 @@
...
+                    if constexpr(Op::is_div){
+                        auto a = hn::LoadNOr(Set<T>(1.0), _Tag<T>(), (const T*)src0, len);
+                        auto b = hn::LoadNOr(Set<T>(1.0), _Tag<T>(), (const T*)src1, len);
+                        auto r = op_func(a, b);
+                        hn::StoreN(r, _Tag<T>(), (T*)dst, len);
+                     } else {
+                        auto a = hn::LoadN(_Tag<T>(), (const T*)src0, len);
+                        auto b = hn::LoadN(_Tag<T>(), (const T*)src1, len);
+                        auto r = op_func(a, b);
+                        hn::StoreN(r, _Tag<T>(), (T*)dst, len);
```
seiko2plus:
```suggestion
                        StoreN(r, dst, len);
                     } else {
                        V a = LoadN(t_src0, len);
                        V b = LoadN(t_src1, len);
                        V r = op_func(a, b);
                        StoreN(r, t_dst, len);
```
use pointer type based, avoid `auto` and expand the wrapper

numpy/_core/src/umath/loops_arithm_fp.dispatch.cpp
```diff
@@ -0,0 +1,766 @@
...
+
+            HWY_LANES_CONSTEXPR int vstep = Lanes<T>();
+            const int wstep = vstep * 2;
+            const int hstep = vstep / 2;
+
+            // lots of specializations, to squeeze out max performance
+            // contiguous
+            if (ssrc0 == 2 && ssrc0 == ssrc1 && ssrc0 == sdst) {
+                for (; len >= vstep; len -= vstep, src0 += wstep, src1 += wstep, dst += wstep) {
+                    auto a0 = LoadU<T>(src0);
```
seiko2plus:
```suggestion
                    V a0 = LoadU(src0);
```
Avoid auto. `LoadU` should detect the pointer type, so no need for explicit deduction.

numpy/_core/src/umath/loops_arithm_fp.dispatch.cpp
```diff
@@ -0,0 +1,766 @@
...
+                for (; len > 0; len -= hstep, src0 += vstep, src1 += vstep, dst += vstep) {
+                    auto a = hn::LoadN(_Tag<T>(), src0, len*2);
+                    auto b = hn::LoadN(_Tag<T>(), src1, len*2);
+                    auto r = op_func(a, b);
+                    hn::StoreN(r, _Tag<T>(), dst, len*2);
+                }
+            }
+            // scalar 0
+            else if (ssrc0 == 0) {
+                auto a = hn::OddEven(Set<T>(src0[1]), Set<T>(src0[0]));
```
seiko2plus:
```suggestion
                V a = hn::OddEven(Set(src0[1]), Set(src0[0]));
```
Same reasoning as `LoadU`.

numpy/_core/src/umath/loops_arithm_fp.dispatch.cpp
```diff
@@ -0,0 +1,766 @@
...
+                    i0 = hn::OddEven(Add(i0, Set<D>(1)), i0);
+                    i1 = hn::OddEven(Add(i1, Set<D>(1)), i1);
+                    for (; len >= vstep; len -= vstep, src1 += ssrc1*vstep, dst += sdst*vstep) {
+                        auto b0 = hn::GatherIndex(_Tag<T>(), src1, i0);
+                        auto b1 = hn::GatherIndex(_Tag<T>(), src1 + ssrc1*hstep, i0);
+                        auto r0 = op_func(a, b0);
+                        auto r1 = op_func(a, b1);
+                        hn::ScatterIndex(r0, _Tag<T>(), dst, i1);
+                        hn::ScatterIndex(r1, _Tag<T>(), dst + sdst*hstep, i1);
+                    }
```
seiko2plus:
```suggestion
```
All the kernels based on gather/scatter for complex datatypes may need to be reimplemented, since they’re very likely to cause massive performance regressions.

Here’s why:
`npyv_loadn2_*` and `npyv_storen2_*` are paired, element-wise non-contiguous loaders/stores. With complex datatypes, we’re supposed to load both elements in a single operation using the same stride. Gather/scatter instead performs two separate loads per item. For example for doubles — universal intrinsics can do a single 128-bit load, while Highway ends up doing two loads plus a permute.

My assumptions/suggestions:

1. On SIMD-capable archs (x86/ARM), the compiler might optimize away some extra loads, but regressions are still very likely. You may try to confirm with benchmarks, e.g. `spin bench --compare parent/main -t "mul|add|sub"`.

2. Double-check Highway — last time I looked, it didn’t implement paired non-contiguous load/store, but worth confirming.

3. Look at universal intrinsics’ `storen2/loadn2` implementations, then reimplement them on top of Highway. You can define them inside the Highway wrapper for now and later consider upstreaming.

seiko2plus (changes requested):
Hey, thank you for working on this. I went through the changes and left some comments mainly around readability, and potential performance issues with non-contig load/store complex datatypes.

numpy/_core/src/umath/loops_arithm_fp.dispatch.cpp
ixgbe: In the Highway project, GatherIndex and ScatterIndex operations have already been implemented in the following platform backends: **hwy/ops/arm_sve-inl.h, hwy/ops/x86_512-inl.h, hwy/ops/x86_256-inl.h, and hwy/ops/rvv-inl.h**.
These operations enable efficient paired non-contiguous load and store through index-based memory access, allowing users to access scattered memory locations via index arrays.

seiko2plus (changes requested):
LGTM so far, except that handling non-contiguous operations for complex data types—previously explained in [this discussion](https://github.com/numpy/numpy/pull/29766#discussion_r2362473314) —is still unresolved. Local tests show regressions of up to `4x` on `x86`. 

Highway’s gather/scatter was not suppose to handle paired non-contiguous accesses, however it can be used as-is for sizeless SIMD registers with 64-bit lanes, since the maximum supported element width is 64 bits. For 32-bit lanes, you can use a reinterpret cast `(hwy::*::BitCast)` to double the lane width (×2). Make sure the offset size in bytes matches the original lane size—for example, when reinterpreting `float32` as `float64`, the offset scale must remain `4` bytes, not `8`.

Handling negative strides matters: use a signed type (in C++ `ptrdiff_t`, which is equivalent to NumPy’s `npy_intp`). If supporting signed offsets is tricky on certain ISAs (e.g., RISC-V extensions), it’s acceptable to disable negative-stride support only for those specific extensions via a `constexpr` switch and fallback to scalar loop. However, it should not be disabled generically across all SIMD extensions, especially for the currently supported architectures.

ixgbe:
> LGTM so far, except that handling non-contiguous operations for complex data types—previously explained in [this discussion](https://github.com/numpy/numpy/pull/29766#discussion_r2362473314) —is still unresolved. Local tests show regressions of up to `4x` on `x86`.
> 
> Highway’s gather/scatter was not suppose to handle paired non-contiguous accesses, however it can be used as-is for sizeless SIMD registers with 64-bit lanes, since the maximum supported element width is 64 bits. For 32-bit lanes, you can use a reinterpret cast `(hwy::*::BitCast)` to double the lane width (×2). Make sure the offset size in bytes matches the original lane size—for example, when reinterpreting `float32` as `float64`, the offset scale must remain `4` bytes, not `8`.
>
 Thanks for the suggestion! However, the 64-bit BitCast approach has alignment requirements:
- ✓ Works only when stride is even
- ✗ Odd strides cause misaligned access
> Handling negative strides matters: use a signed type (in C++ `ptrdiff_t`, which is equivalent to NumPy’s `npy_intp`). If supporting signed offsets is tricky on certain ISAs (e.g., RISC-V extensions), it’s acceptable to disable negative-stride support only for those specific extensions via a `constexpr` switch and fallback to scalar loop. However, it should not be disabled generically across all SIMD extensions, especially for the currently supported architectures.

PR #29129, opened 2025-06-05
ENH: add a casting option 'same_value' and use it in np.astype

mattip:
Users have asked for an option to raise an error when casting overflows. This is hard to do with the NEP-50 based casting, which only offers blanket rules for casting one dtype to another, without taking the actual values into account. This PR adds a new `same_value` argument to the `casting` kwarg (implemented in `PyArray_CastingConverter`), and extends the casting loop functions to raise a `ValueError` if `same_value` casting and the value would be changed by the cast. So far this is only implemented for `ndarray.astype`, i.e. `np.array([1000]).astype(np.int8, casting='same_value')` will now raise an error.

Performance: since the PR touches the assignment deep in the inner loop, I checked early on that it does not impact performance when not using `same_value` casting. The loop pseudo code now looks like this, and when compiled with `-O3` (via a gcc pragma specific to the loop), it ~seems the compiler is smart enough to make the `if` condition not impact performance.~ seems there is a small performance impact for some casting functions
```
int same_value_casting = <check some condition>
type2 dst_value; type1 src_value;
while (<condition>) {
    <setup dst_value, src_value>
    dst_value = (cast)src_value;
    if (same_value_casting) {
        <do some checking>;
    }
    <iterate>;
}

```

Protection: This PR only changes `ndarray.astype`. Future PRs may implement more. I tried to guard the places that use `PyArray_CastingConverter` to raise an error if `'same_value'` is passed in:
- `array_datetime_as_string` (by disallowing `'same_value'` in `can_cast_datetime64_units`)
- `array_copyto`
- `array_concatenate` (by disallowing `'same_value'` in `PyArray_ConcatenateInto`)
- `array_einsum` (by disallowing `'same_value`' in `PyArray_EinsteinSum`)
- `ufunc_generic_fastcall` 

`'same_value'` is allowed in
- `array_astype` (the whole goal of this PR)
- `array_can_cast_safely`
- `npyiter_init` (I am pretty sure that is OK?)
- `NpyIter_NestedIters` (I am pretty sure that is OK?)

Testing: I added tests for ndarray.astype() covering all combinations of built-in types, and also some tests for properly blocking `casting='same_value'`.

TODO:
- [x] disallow `astype(..., casting='same_value')` with `datetime64` or user-defined dtypes?
- [ ] rerun benchmarks against `main`

PR #29767, opened 2025-09-18
DOC: add another mention of 'same_value'  

mattip:
Follow-up to #29129 

Add a mention of `same_value` casting in the beginner type documentation. Also rearrange the text to discuss casting at the end rather than in the middle of the section.

### DIFF
--- a/doc/source/user/basics.types.rst
+++ b/doc/source/user/basics.types.rst
@@ -35,18 +35,6 @@ See :ref:`arrays.dtypes.constructing` for more information about specifying and
 constructing data type objects, including how to specify parameters like the
 byte order.
 
-To convert the type of an array, use the .astype() method. For example: ::
-
-    >>> z.astype(np.float64)                 #doctest: +NORMALIZE_WHITESPACE
-    array([0.,  1.,  2.])
-
-Note that, above, we could have used the *Python* float object as a dtype
-instead of `numpy.float64`.  NumPy knows that
-:class:`int` refers to `numpy.int_`, :class:`bool` means
-`numpy.bool`, that :class:`float` is `numpy.float64` and
-:class:`complex` is `numpy.complex128`.  The other data-types do not have
-Python equivalents.
-
 To determine the type of an array, look at the dtype attribute::
 
     >>> z.dtype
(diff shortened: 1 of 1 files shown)


doc/source/user/basics.types.rst
```diff
@@ -66,6 +54,28 @@ properties of the type, such as whether it is an integer::
     >>> np.issubdtype(d, np.floating)
     False
 
+To convert the type of an array, use the .astype() method. For example: ::
+
```
charris, line 58: You just need `::` here, it leaves a single `:` behind.

mattip: There are many examples of `: ::` in the documentation, maybe we should clean them all up at once?

<details><summary> `git grep ": ::" doc/source` </summary>

```
doc/source/glossary.rst:       Trailing slices can be omitted: ::
doc/source/user/basics.creation.rst:example: ::
doc/source/user/basics.indexing.rst:than dimensions, one gets a subdimensional array. For example: ::
doc/source/user/basics.indexing.rst:element being returned. That is: ::
doc/source/user/basics.indexing.rst:permitted to assign a constant to a slice: ::
doc/source/user/basics.indexing.rst:or an array of the right size: ::
doc/source/user/basics.indexing.rst:exceptions (assigning complex to floats or ints): ::
doc/source/user/basics.indexing.rst:example is often surprising to people: ::
doc/source/user/basics.indexing.rst:in Python. For example: ::
doc/source/user/basics.indexing.rst:object: ::
doc/source/user/basics.indexing.rst:converted to an array as a list would be. As an example: ::
doc/source/user/basics.types.rst:To convert the type of an array, use the .astype() method. For example: ::
doc/source/user/basics.types.rst::ref:`overflow-errors`): ::
doc/source/user/misc.rst:Note: cannot use equality to test NaNs. E.g.: ::
doc/source/user/misc.rst:Other related special value functions: ::
doc/source/user/misc.rst:from the results: ::
doc/source/user/misc.rst:     attribute: ::

```

</details>

mattip: Fixed

charris:
Looking at the published documentation, both versions work, but plain old `::` is the cleaner way to do it.

charris:
Thanks Matti.

PR #29768, opened 2025-09-19
BUG: Fix pocketfft umath strides for powerpc compatibility

barracuda156:
See: https://github.com/numpy/numpy/issues/29758

### DIFF
--- a/numpy/fft/_pocketfft_umath.cpp
+++ b/numpy/fft/_pocketfft_umath.cpp
@@ -32,7 +32,7 @@
 template<PyUFuncGenericFunction cpp_ufunc>
 static void
 wrap_legacy_cpp_ufunc(char **args, npy_intp const *dimensions,
-                      ptrdiff_t const *steps, void *func)
+                      npy_intp const *steps, void *func)
 {
     NPY_ALLOW_C_API_DEF
     try {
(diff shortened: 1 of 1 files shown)


numpy/fft/_pocketfft_umath.cpp
```diff
@@ -233,8 +233,9 @@ irfft_loop(char **args, npy_intp const *dimensions, npy_intp const *steps, void
     size_t nin = (size_t)dimensions[1], nout = (size_t)dimensions[2];
     ptrdiff_t step_in = steps[3], step_out = steps[4];
 
+#ifndef POCKETFFT_NO_VECTORS
     size_t npts_in = nout / 2 + 1;
-
+#endif
```
seberg: Would be nicer to just move it into the existing `#ifndef` below, but a small nit, but thanks for fixing the warning!

seberg (approved):
Looks good, thanks.  `ptrdiff_t` vs. `npy_intp` usage differs a bit between functions now, but it doesn't matter.

EDIT: To be clear, the test failures are unrelated, there are some general problems that will be fixed very soon hopefully (maybe it already is).

Unfortunately, tests are failing.  It would be nice if you can move the `npts_in` just to keep things slightly simpler.

barracuda156:
@seberg Do we need to change something here at the moment re test failures?

seberg:
It would be nice to move that one thing, but otherwise should be all good.  We can re-trigger CI manually (or close/reopen PR) when things are resolved.

numpy/fft/_pocketfft_umath.cpp
mhvk: Agreed, while you are at it, please do move it inside (it should have been there in the first place, as it is not used outside).

mhvk (commented):
Thanks for catching this -- as @seberg noted in the issue, the reason I used `ptrdiff_t` here was that that was what `pocketfft` expected. I vaguely recall getting compiler warnings without it, but if this works now without explicit casts, then all fine by me.

mattip (approved):
LGTM pending CI success

barracuda156:
@seberg Thank you for helping here, and sorry for a delay, I did not have much time this weekend.

Issue #27698, opened 2024-11-04
Add Android Support


rbebb:
### Proposed new feature or change:

Now that Android is a supported platform in Python 3.13, it would be great to have numpy support given that it’s the foundation of many other Python libraries. Tools that let you run Python on Android (like Chaquopy) currently have to build packages like numpy manually and create custom patches.

mattip:
I had to dig around a bit for context: it seems Chaquopy is the system used to build some python libraries with c-extensions for beeware, which is the goto authority on packaging python for ios and android. As far as I can tell, those platforms cannot install wheels and are using a different packaging system for NumPy.

What would android support look like? What do you call "numpy support"? Is there a way to emulate android on github CI so we can test the eventual build?

It seems the patch you pointed to still uses setuptools, so the Chaquopy build would need to move to meson in order for us to merge the patch. It would be good to add such a build to CI, maybe as an additional job on the wheel build, to make sure we don't break something.

rgommers:
> It would be good to add such a build to CI, maybe as an additional job on the wheel build, to make sure we don't break something.

+1 a CI job would be fine if that seems helpful. It'd have a similar status as for Pyodide.

mhsmith:
The Chaquopy build tool does produce wheels – we currently host them [here](https://chaquo.com/pypi-13.1/), and the Chaquopy Gradle plugin (either directly or via Briefcase) can build them into Android apps with a slightly patched version of pip. The equivalent for iOS is [this tool](https://github.com/beeware/mobile-forge) and [this repository](https://anaconda.org/beeware/repo).

Now that Android and iOS are officially supported in Python itself, we're planning to contribute support for these platforms to more general packaging tools such as cibuildwheel. The plan is that these mobile-specific packaging tools and repositories will eventually become obsolete, and projects will be able to build Android and iOS wheels with their existing workflows, and publish them on PyPI along with all the other platforms.

Testing on Android and iOS emulators is also possible on GitHub Actions, and we're already doing it in the Toga project. To see what this looks like, open [one of our recent CI runs](https://github.com/beeware/toga/actions/workflows/ci.yml?query=branch%3Amain) and look at the "testbed" jobs.

@freakboy3742 is leading the work on this. We're not yet ready to recommend it for production use, but we'll post a comment here once that changes.

freakboy3742:
A quick +1 to everything @mhsmith said here. 

[...]

PR #29770, opened 2025-09-19
BLD: cross-compilation for android using cibuildwheel

kfchou:
Related to: #27698.

These configs enable you to run `cibuildwheel --platform android` to build android-compatible `.whl` files.

You can use [`briefcase`](https://briefcase.readthedocs.io/en/stable/) to specify these wheels when you want to build a python-based android app that uses `numpy`. See [discussion](https://github.com/beeware/briefcase/discussions/2472#discussioncomment-14366117) for details.

**Testing**
I tested these wheels using a demo `briefcase` app, importing numpy and get it to print out mathematical operations. 

Questions for reviewers:
* Should I include the demo `briefcase` app for testing? Not sure how I should do that
* This build requires python 3.13. I'm not sure how to specify this requirement.


andyfaff:
A discussion is needed before we add wheel support for the Android platform. These discussions take place on numpy-discussion@scipy.org, as well as the regular online meetings (advertised on the mailing list).

Whilst Android is obviously quite popular we need to know that finite resources (developer time, CI time) is going spent effectively. i.e. we need to know that demand will be there.

Probably the first step would be to add a CI entry, rather than distributing a wheel.

andyfaff:
Ok, so the original issue said all of that and more. We'd probably be looking for a CI entry for regular test runs, not building a wheel.

tools/wheels/cibw_before_build_android.sh
```diff
@@ -0,0 +1,451 @@
...
+cpp_args = ['-DHAVE_SIN=1', '-DHAVE_COS=1', '-DHAVE_TAN=1', '-DHAVE_SINH=1', '-DHAVE_COSH=1', '-DHAVE_TANH=1', '-DHAVE_ASIN=1', '-DHAVE_ACOS=1', '-DHAVE_ATAN=1', '-DHAVE_ATAN2=1', '-DHAVE_EXP=1', '-DHAVE_LOG=1', '-DHAVE_LOG10=1', '-DHAVE_SQRT=1', '-DHAVE_FLOOR=1', '-DHAVE_CEIL=1', '-DHAVE_FABS=1', '-DHAVE_POW=1', '-DHAVE_FMOD=1', '-DHAVE_LDEXP=1', '-DHAVE_FREXP=1', '-DHAVE_MODF=1', '-DHAVE_COPYSIGN=1', '-DHAVE_FINITE=1', '-DHAVE_ISINF=1', '-DHAVE_ISNAN=1', '-DHAVE_ISFINITE=1', '-DHAVE_EXPM1=1', '-DHAVE_LOG1P=1', '-DHAVE_ASINH=1', '-DHAVE_ACOSH=1', '-DHAVE_ATANH=1', '-DHAVE_RINT=1', '-DHAVE_TRUNC=1', '-DHAVE_EXP2=1', '-DHAVE_LOG2=1', '-DHAVE_HYPOT=1', '-DHAVE_CBRT=1', '-DHAVE_NEXTAFTER=1', '-DHAVE_CSIN=1', '-DHAVE_CCOS=1', '-DHAVE_CTAN=1', '-DHAVE_CSINH=1', '-DHAVE_CCOSH=1', '-DHAVE_CTANH=1', '-DHAVE_CASIN=1', '-DHAVE_CACOS=1', '-DHAVE_CATAN=1', '-DHAVE_CASINH=1', '-DHAVE_CACOSH=1', '-DHAVE_CATANH=1', '-DHAVE_CEXP=1', '-DHAVE_CLOG=1', '-DHAVE_CPOW=1', '-DHAVE_CSQRT=1', '-DHAVE_CABS=1', '-DHAVE_CARG=1', '-DHAVE_CIMAG=1', '-DHAVE_CREAL=1', '-DHAVE_CONJ=1', '-DHAVE_STRTOLL=1', '-DHAVE_STRTOULL=1', '-DHAVE_STRTOLD=1']
+EOF
```
mattip, line 289: This should all end up in one of the config.h files, not as command line flags

mattip:
> Should I include the demo briefcase app for testing? Not sure how I should do that

We do have aarch64 testing base on github `ubuntu-22.04-arm` runners. Is there a way to leverage those to test android? I think testing would be a better first step than just wheel building.

tools/wheels/cibw_before_build_android.sh
```diff
@@ -0,0 +1,451 @@
+#!/bin/bash
+set -e
+set -x
+
+# This script is designed to be called by cibuildwheecho "--- END WORKAROUND ---"
```
rgommers, line 5: Accidental edit at the end of this sentence

tools/wheels/cibw_before_build_android.sh
```diff
@@ -0,0 +1,451 @@
...
+
+# Get the target architecture from cibuildwheel's environment variables
+# (Now set by our workaround above if cibuildwheel didn't set it)
+echo "--- Architecture Configuration ---"
+echo "CIBW_ARCHS: ${CIBW_ARCHS:-'(not set)'}"
+echo "CIBW_ARCHS_ANDROID: ${CIBW_ARCHS_ANDROID:-'(not set)'}"
+
+# Check for architecture in standard cibuildwheel environment variablesbuilding for Android.
```
rgommers, line 13: typo in "variablesbuilding"

tools/wheels/cibw_before_build_android.sh
```diff
@@ -0,0 +1,451 @@
...
+# Since cibuildwheel 3.x has built-in Android support, we mainly need to generate 
+# the correct Meson cross file.
+
+# WORKAROUND FOR CIBUILDWHEEL ISSUE:
+# cibuildwheel documentation states that CIBW_ARCHS should be available for Android builds,
+# but cibuildwheel 3.1.4 does NOT set these environment variables for Android.
+# This appears to be a bug or incomplete implementation in cibuildwheel.
```
rgommers, line 20: It'd be good to fix that upstream in `cibuildwheel` and add a link here to that issue or PR.

tools/wheels/cibw_before_build_android.sh
```diff
@@ -0,0 +1,451 @@
...
+# 
+# References:
+# - Documentation: https://cibuildwheel.pypa.io/en/stable/options/#archs
+# - Issue should be reported to: https://github.com/pypa/cibuildwheel/issues
+#
+# As a workaround, we detect the architecture from various sources:
+# 1. Check if CIBW_ARCHS is already set (future-proofing)
+# 2. Look for architecture hints in environment variables
+# 3. Check process command line for architecture indicators
```
rgommers, line 29: You already set this in `pyproject.toml`, so three different ways of handling it here seems like overkill?

tools/wheels/cibw_before_build_android.sh
```diff
@@ -0,0 +1,451 @@
...
+else
+    # This should not happen now that we set CIBW_ARCHS in the workaround above
+    echo "ERROR: Architecture detection failed even after workaround!"
+    echo "This indicates a serious issue with the detection logic."
+    exit 1
+fi
+
+echo "--- Building for architecture: $TARGET_ARCH ---"
+
+# Set the target API level from cibuildwheel's environment variable
```
rgommers, line 113: Also already handled in `pyproject.toml`

tools/wheels/cibw_before_build_android.sh
```diff
@@ -0,0 +1,451 @@
...
+    *)
+        echo "Unsupported architecture: $TARGET_ARCH"
+        exit 1
+        ;;
+esac
+
+echo "Mapped to Android: $ANDROID_ARCH, Meson: $MESON_CPU_FAMILY/$MESON_CPU"
+
+# Generate a dynamic cross file for the current architecture
+CROSS_FILE="/tmp/android_${ANDROID_ARCH}.meson.cross"
```
rgommers, line 154: This doesn't match the path in `pyproject.toml`. Inserting the architecture name isn't necessary, and hard to encode statically in `pyproject.toml`.

tools/wheels/cibw_before_build_android.sh
```diff
@@ -0,0 +1,451 @@
...
+# via environment variables after this script runs, but we override
+# them here to use API $API_LEVEL toolchain.
+#
+
+[binaries]
+# Use wrapper scripts that filter out \$(BLDLIBRARY) from linking commands
+c = '/tmp/android-clang-wrapper.sh'
+cpp = '/tmp/android-clang++-wrapper.sh'
+ar = '/home/kfchou/.cache/briefcase/tools/android_sdk/ndk/27.2.12479018/toolchains/llvm/prebuilt/linux-x86_64/bin/llvm-ar'
+strip = '/home/kfchou/.cache/briefcase/tools/android_sdk/ndk/27.2.12479018/toolchains/llvm/prebuilt/linux-x86_64/bin/llvm-strip'
```
rgommers, line 176: (1) Paths in this cross file should be a template that's filled out correctly at build time,
(2) It looks like this belongs in a separate file, rather than inlined into a bash script. `ci/cross_android.ini` (or `.ini.templ`) would be fine

tools/wheels/cibw_before_build_android.sh
```diff
@@ -0,0 +1,451 @@
...
+has_function_carg = true
+has_function_cimag = true
+has_function_creal = true
+has_function_conj = true
+has_function_strtoll = true
+has_function_strtoull = true
+has_function_strtold = true
+
+# Skip BLAS/LAPACK for Android to avoid complex math library dependencies
+allow-noblas = true
```
rgommers, line 270: This goes in `[tool.cibuildwheel.android]` in `pyproject.toml` normally. There's already an example for 32-bit Windows there.

tools/wheels/cibw_before_build_android.sh
```diff
@@ -0,0 +1,451 @@
...
+# Specify long double format for ARM64 Android to avoid cross-compilation test run
+# Android ARM64 uses IEEE 754 quad precision (128-bit)
+longdouble_format = 'IEEE_QUAD_LE'
+
+[built-in options]
+# Android Python extensions need proper library linking for runtime resolution
+# This ensures symbols like PyExc_ValueError and math functions are available
+# Note: On Android, math functions are part of libc, not a separate libm
+c_link_args = ['-lpython3.13']
+cpp_link_args = ['-lpython3.13']
```
rgommers, line 281: If this isn't handled correctly in Meson, that's where this should be fixed. That also avoids hardcoding the interpreter version here.

tools/wheels/cibw_before_build_android.sh
rgommers: Not even that, if these checks actually fail, they can either be fixed in `numpy/_core/meson.build` or a simple skip for the checks can be added there. That goes for `has_function_sin` et al. higher up too.

tools/wheels/cibw_before_build_android.sh
```diff
@@ -0,0 +1,451 @@
...
+echo "Using cibuildwheel's built-in Android toolchain"
+echo "Cross file: /tmp/android_current.meson.cross"
+
+# Debug Python configuration to understand the BLDLIBRARY issue
+echo "--- Debugging Python configuration ---"
+python -c "
+import sysconfig
+import os
+
+# Check BLDLIBRARY configuration
```
rgommers, line 308: Left-over debug code from here onwards?

tools/wheels/repair_android_wheel.sh
```diff
@@ -0,0 +1,142 @@
+#!/bin/bash
```
rgommers, line 1: This looks like it should be either a separate tool or part of Briefcase. It's not specific to `numpy`, just a generic wheel repair utility.

rgommers (commented):
Thanks for working on this @kfchou! I left some more detailed comments; I'd say the overall goal is to make the custom code we need to carry for this as small and `numpy`-specific as possible, and having one CI job that tests that building an Android wheel actually works as intended.

> Should I include the demo briefcase app for testing? Not sure how I should do that

No, that is undesirable. The right approach is running the test suite (or a relevant subset of it, depending on how show things run) under an emulator. On gh-27698 it was pointed out that this is possible.

> This build requires python 3.13. I'm not sure how to specify this requirement.

The general config files (`pyproject.toml`, `cibw_before*`, `repair_*`) should not make such an assumption, since they should work for any Python version >=3.13. In the CI job to test this cross build, 3.13 can be picked. 

https://github.com/beeware/toga has examples of CI jobs that may be helpful here.

tools/wheels/cibw_before_build_android.sh
kfchou: This is handled by
```
# Create a symlink with a predictable name for the config
ln -sf "$CROSS_FILE" "/tmp/android_current.meson.cross"
```
where `/tmp/android_current.meson.cross` is the file referenced in `pyproject.toml`.

perhaps this is overkill in production but it was helpful during development

rgommers: Thanks for the context. I'd prefer to simplify this; fewer moving pieces is easier to understand.

kfchou:
Thanks for your feedback everyone.  @rgommers, I'm slowly working through your comments.

Admittedly, most of the this work is done by directing copilot/sonnet since I'm not experience with Meson or C.

mhsmith:
> We do have aarch64 testing base on github `ubuntu-22.04-arm` runners. Is there a way to leverage those to test android? 

Unfortunately I found that none of GitHub's ARM runners can run the Android emulator ([1](https://github.com/pypa/cibuildwheel/pull/2349#issuecomment-2937749871), [2](https://github.com/pypa/cibuildwheel/pull/2349#issuecomment-2940950298)). They can only run it on Linux x86_64.

>I think testing would be a better first step than just wheel building.

I would strongly recommend doing this through cibuildwheel, as it automaticaly handles all the details of setting up Android build and test environments, starting an emulator, etc. However, on Android it does require the test command to be a `python -c` or `python -m` command, so the shell script currently in the cibuildwheel section of pyproject.toml wouldn't work here.

tools/wheels/cibw_before_build_android.sh
mhsmith: I think this is a misunderstanding: cibuildwheel reads the CIBW variables as one possible means of configuration, it doesn't set them itself.

tools/wheels/cibw_before_build_android.sh
```diff
@@ -0,0 +1,451 @@
...
+print('LIBS:', repr(libs))
+
+# Check library directory
+libdir = sysconfig.get_config_var('LIBDIR')
+print('LIBDIR:', repr(libdir))
+"
+
+# Try different approaches to fix BLDLIBRARY issue
+# The issue is meson-python is inserting literal '$(BLDLIBRARY)' in link commands
+# For Android, we don't want to link Python extensions to libpython at all
```
mhsmith, line 330: We absolutely do want to link them to libpython, but if it's using the Python .pc file to achieve this, then this may be affected by https://github.com/python/cpython/issues/138800.

mhsmith:
This PR has now been overtaken by #30412, so I suggest closing it.

mhsmith:
@rgommers: This can now be closed.

PR #29774, opened 2025-09-20
BUG: np.setbufsize should raise ValueError for negative input

### LINKED ISSUE #29651: np.setbufsize accepts negative values silently instead of raising an error
`numpy.setbufsize` currently accepts negative values for size, but does nothing silently.

```python
import numpy as np

print("Default bufsize:", np.getbufsize())  # 8192

with np.errstate():
    prev = np.setbufsize(-1024)   # negative size, meaningless
    print("Previous bufsize:", prev)          # 8192
    print("Current bufsize:", np.getbufsize())  # still 8192, unchanged

# automatically restored
print("Restored bufsize:", np.getbufsize())  # 8192
```

I know that negative buffer sizes have no real meaning. Should this be treated the same as other invalid sizes (e.g., very small values), and raise an error instead of silently doing nothing?

cibarbia05:
### Closes #29651

This PR fixes an issue where `np.setbufsize` silently accepted negative
values. Passing a negative value now raises a `ValueError` with a clear
error message.

### Changes
- Added a check in `np.setbufsize` to raise `ValueError` when `size < 0`.
- Added a new unit test `test_negative_value_raises` in
  `numpy/_core/tests/test_umath.py`.

### Example
```python
>>> import numpy as np
>>> np.setbufsize(-5)
    ...
ValueError: buffer size must be non-negative

### DIFF
--- a/numpy/_core/_ufunc_config.py
+++ b/numpy/_core/_ufunc_config.py
@@ -187,6 +187,9 @@ def setbufsize(size):
     8192
 
     """
+    if size < 0:
+        raise ValueError("buffer size must be non-negative")
+        
     old = _get_extobj_dict()["bufsize"]
     extobj = _make_extobj(bufsize=size)
     _extobj_contextvar.set(extobj)
--- a/numpy/_core/_ufunc_config.py
+++ b/numpy/_core/_ufunc_config.py
@@ -189,7 +189,6 @@ def setbufsize(size):
     """
     if size < 0:
         raise ValueError("buffer size must be non-negative")
-        
     old = _get_extobj_dict()["bufsize"]
     extobj = _make_extobj(bufsize=size)
     _extobj_contextvar.set(extobj)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by mattip** [comment]
Something broke the build. I am seeing this in various CI steps that re-use the meson build to run other spin steps
```
Invoking `build` prior to running tests:
$ /opt/hostedtoolcache/Python/3.11.13/x64/bin/python vendored-meson/meson/meson.py compile -j Sentinel.UNSET -C build
```

Where is this `Sentinel.UNSET` coming from? There is a click problem I see elsewhere, something about sentinel values, did we not pin the click version in the requirements?

**Comment by WarrenWeckesser** [comment]
The corresponding scipy issue is https://github.com/scipy/scipy/issues/23642.

A solution would be to ensure that `spin` 0.15 is used.  (`click` is a transitive dependency coming from `spin`.)

**Comment by charris** [comment]
close/reopen

File: numpy/_core/_ufunc_config.py [hunk 1]
```diff
@@ -187,6 +187,9 @@ def setbufsize(size):
     8192
 
     """
+    if size < 0:
+        raise ValueError("buffer size must be non-negative")
+
```
mattip: Linting failure
```suggestion

```

mattip:
Thanks @cibarbia05

cibarbia05:
> Thanks @cibarbia05

Thanks for the review and merge! Glad to contribute.

PR #29776, opened 2025-09-21
TYP: fix and improve ``{f,i}info`` stubs in ``_core.getlimits``

jorenham:
This is a port of numpy/numtype#125 with some modifications for the sake of backwards compatibility.

The reason for moving these classes to the `_core.getlimits` stubs is because that's also where they're defined at runtime. It also helps make `__init__.pyi` less monolithic.


jorenham:
woops; forgot about the fallback overload

charris:
close/reopen.

charris:
Thanks Joren.

PR #29777, opened 2025-09-21
BLD: Upgrade spin requirement to version 0.15

WarrenWeckesser:
`spin` 0.15 pins its `click` dependency to avoid the newly released `click` version 8.3.0.

### DIFF
--- a/.github/workflows/compiler_sanitizers.yml
+++ b/.github/workflows/compiler_sanitizers.yml
@@ -90,6 +90,9 @@ jobs:
       - name: Uninstall pytest-xdist (conflicts with TSAN)
         run: pip uninstall -y pytest-xdist
 
+      - name: Upgrade spin (gh-29777)
+        run: pip install -U spin
+
       - name: Build NumPy with ThreadSanitizer
         run: python -m spin build -- -Db_sanitize=thread
 
--- a/environment.yml
+++ b/environment.yml
@@ -16,7 +16,7 @@ dependencies:
   - ninja
   - pkg-config
   - meson-python
-  - spin==0.13
+  - spin==0.15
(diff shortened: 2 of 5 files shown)


WarrenWeckesser:
The `clang_TSAN` job from `compiler_sanitizers.yml` appears to use a container where all the Python packages are pre-installed.  I don't know how and when this container is updated, so I'm not sure what is in there.  Apparently it gets the newly released `click` 8.3.0, but it does not have the even newer `spin` 0.15.  I could attempt to hack `compiler_sanitizers.yml` to upgrade `spin` before it is used to build numpy (probably right after `pytest-xdist` is removed), but I don't know if that is the "correct" solution for this job.

*Edit:* Let's give it a shot...

charris:
@ngoldbaum I see that you have opened other issues at https://github.com/nascheme/cpython_sanity/pkgs/container/numpy-tsan, seems we need another.

> Edit: Let's give it a shot...

We can ignore/disable it for a while.

ngoldbaum:
> We can ignore/disable it for a while.

Go ahead and disable it, I’ll reenable it later this week with a fix.

charris:
Looks like @WarrenWeckesser fix works.

ngoldbaum:
Yeah that works. The other thing that might be more sustainable is to use the `cpython_tsan` image and then install numpy’s build nd test dependencies from the requirements file like we do everywhere else. The only difference that `numpy_tsan` has is that it includes a numpy build and I think whoever added this used that to bootstrap the numpy build environment, but that leads to issues like this.

charris:
Let's get this in to fix the CI failures,  we can tweak it later. Thanks Warren.

Issue #19511, opened 2021-07-18
ENH: Fix double-double smallest normal number


seberg:
See this and the following discussion, we should likely fix the double-double value for `np.finfo(np.longdouble).smallest_normal`. @wkschwartz, lets continue discussion here.

I might be missing some context here, so forgive me if I'm confusing matters. It appears that you are able to compute the smallest positive subnormal 𝜎, and a few lines below you define `macheps`, so I take it you are able to compute the machine epsilon 𝜀. From there, you can in fact compute the smallest positive normal 𝜔 as 𝜎 ⨸ 𝜀, where ⨸ means division in the relevant floating point system (which is what the `/` operator does in C and Python). In Python terms, you can do approximately the following, but you'll need to move up the definition of `macheps` from below.

```Python
    # Leave the same value for the smallest subnormal as double
    smallest_subnormal_dd = ld(nextafter(0., 1.))
    smallest_normal_dd = smallest_subnormal / macheps
```

_Originally posted by @wkschwartz in https://github.com/numpy/numpy/pull/18536#discussion_r670442245_

wkschwartz:
Copying my follow up to the OP discussion below. Like I wrote, I'll explain more later with cites (but can't right now). 

-------------
> We have those two main choices: just go with double precision (the old value) or go with the smallest number that has a 106bit (implicit) mantissa (and above which all numbers do!)

Upon further reflection the correct answer is the latter. I can explain more later but for now the basic idea is that the distinction between subnormal and normal is all about guaranteed minimum precision. Subnormals don't satisfy the main axiom of FP in numerical analysis that _relative_ errors are bounded by half a machine epsilon. Normals do maintain that invariant. So setting the dividing line between them based on precision is the only useful way to go.

wkschwartz:
The following "standard model" of floating-point (FP) arithmetic in numerical analysis is used in Higham (2002, §2.2) and Trefethen and Bau (1997, Ch. 13):

> The set of FPs is a subset of the real numbers. Let _u_ be a small number called the _unit roundoff_. To each real arithmetic operation ∗ ∈ {+, −, ×, ÷} belongs a corresponding FP operation ⊛ such that, for all FPs _a_ and _b_, there exists _ε_ with |_ε_| ≤ _u_ and _a_ ⊛ _b_ = (_a_ ∗ _b_) (1 + _ε_).

[...]

PR #29778, opened 2025-09-21
WIP: Reorganize finfo and add new constant slot

seberg:
Reworking of finfo and addition of a C-side constant grabbing function.

Right now, deleted a lot of the old stuff simply, but it may make sense to move some of it to the tests it removes the warning/bad value for certain semi-well defined values of double-double:  I think that is fine (with a release note), just take what C says...  The fact is, it came about because we didn't want to put something clearly wrong and nobody figured out what is right.  If C says something, we can just assume that is right.

Since this is supposed to create scalars, I somewhat like to use getitem, but adding the dummy one has some fallouts (a few void passes actually need the array).
There is also some other weirder fallouts still, so this needs figuring out (or maybe just creating the scalar another way).


charris:
close/reopen

numpy/_core/src/multiarray/dtypemeta.c
```diff
@@ -118,6 +118,43 @@ use_new_as_default(PyArray_DTypeMeta *self)
...
+legacy_fallback_getitem(PyArray_Descr *descr, char *data)
+{
+    PyArrayObject_fields arr_fields = {
+        .flags = NPY_ARRAY_WRITEABLE,  /* assume array is not behaved. */
+    };
+    Py_SET_TYPE(&arr_fields, &PyArray_Type);
+    Py_SET_REFCNT(&arr_fields, 1);
+    arr_fields.descr = descr;
+    return PyDataType_GetArrFuncs(descr)->getitem(data, &arr_fields);
+}
```
seberg, line 155: @SwayamInSync you probably got that, but should just delete this part and all changes that look related to `getitem` or `setitem` for now (this might be right, but I broke something and we don't really need it -- the problem is in part structured and in part probably that NumPy dtypes return python objects for `getitem`, we actually need two getitems or just keep a special case around for them).

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -4327,6 +4327,86 @@ normalize_axis_index(PyObject *NPY_UNUSED(self),
...
+        {"eps", NPY_CONSTANT_finfo_eps, 0},
+        {"smallest_normal", NPY_CONSTANT_finfo_smallest_normal, 0},
+        {"smallest_subnormal", NPY_CONSTANT_finfo_smallest_subnormal, 0},
+        {"nmant", NPY_CONSTANT_finfo_nmant, 1},
+        {"minexp", NPY_CONSTANT_finfo_min_exp, 1},
+        {"maxexp", NPY_CONSTANT_finfo_max_exp, 1},
+        {"precision", NPY_CONSTANT_finfo_decimal_digits, 1},
+        {NULL, -1},
+    };
+
```
seberg, line 4367: I forgot cleanup here which is _theoretically_ necessary.  I think the easiest workaround will be to create an empty array that is large enough to hold all of the constants and then fill it in (you can assume contiguity so it's easy).

Otherwise you need to clean up the element each iteration, but if we use an array for now then this is much simpler.

Fixing getitem/setitem and removing that looks worthwhile to me, but we don't have to do it yet.

numpy/_core/src/multiarray/dtypemeta.c
```diff
@@ -1173,6 +1211,9 @@ dtypemeta_wrap_legacy_descriptor(
     dt_slots->is_known_scalar_type = python_builtins_are_known_scalar_types;
     dt_slots->common_dtype = default_builtin_common_dtype;
     dt_slots->common_instance = NULL;
+    // These may be overwritten, but if not provide fallback via array struct hack:
+    dt_slots->setitem = legacy_fallback_setitem;
+    dt_slots->getitem = legacy_fallback_getitem;
     dt_slots->ensure_canonical = ensure_native_byteorder;
     dt_slots->get_fill_zero_loop = NULL;
     dt_slots->finalize_descr = NULL;
```
seberg, line 1219: Missed this, need to set `get_constant` to the default here too to avoid a lingering bug.

seberg (commented):
Small comments for @SwayamInSync on where I stopped but realized how to continue.

seberg:
Doing this will also close gh-19511.  Could have a small release note for that, maybe.
(I suppose there is a chance some compiler comes around and misses these at which point we may have to add it back, but we do require C11, so they should be defined.)

SwayamInSync:
I see, thanks for the head start @seberg

seberg:
Closing, as gh-29836 should supersede this.

PR #29780, opened 2025-09-22
BUG: Fix assert in nditer buffer setup

### LINKED ISSUE #29690: BUG: Crash in np.unravel_index if debug assertions enabled.
### Describe the issue:

In a debug build of NumPy:
```
In [1]: import numpy as np

In [2]: np.unravel_index(np.array([[1, 0, 1, 0]], dtype=np.uint32), (4,))
python3: ../numpy/_core/src/multiarray/nditer_constr.c:2081: npyiter_find_buffering_setup: Assertion `(axisdata->shape == 1) ? (prev_strides[iop] == strides[iop]) : 1' failed.
```

```python
import numpy as np
np.unravel_index(np.array([[1, 0, 1, 0]], dtype=np.uint32), (4,))
```

### Error message:

```shell
See above.
```

2.4.0.dev0+git20250905.366c0a3
3.12.10 (main, Apr  9 2025, 04:03:51) [Clang 20.1.0 ]

This also reproduces in 2.3.2.

### Runtime Environment:

[{'numpy_version': '2.4.0.dev0+git20250905.366c0a3',
  'python': '3.12.10 (main, Apr  9 2025, 04:03:51) [Clang 20.1.0 ]',
  'uname': uname_result(system='Linux', node='redacted', release='6.12.32-redacted-amd64', version='#1 SMP PREEMPT_DYNAMIC Debian 6.12.32-redacted (2025-06-26)', machine='x86_64')},
 {'simd_extensions': {'baseline': [], 'found': [], 'not_found': []}},
 {'architecture': 'Zen',
  'filepath': '/usr/lib/x86_64-linux-gnu/openblas-pthread/libopenblasp-r0.3.29.so',
  'internal_api': 'openblas',
  'num_threads': 128,
  'prefix': 'libopenblas',
  'user_api': 'blas',
  'version': '0.3.29'}]

### Context for the issue:

Found while debugging a crash in the JAX test suite.

seberg:
When using a buffered iteration, there was a new assert to check that strides were set up nicely to ensure that we are not missing dimension coalescing.

However, when the strides are set up to track an index, then they were still set up with a 0 for length 1.
I _think_ that is just unnecessary (i.e. the assert is correct to point it out).  But, let's do this first for backporting at least.

Closes gh-29690

### DIFF
--- a/numpy/_core/src/multiarray/nditer_constr.c
+++ b/numpy/_core/src/multiarray/nditer_constr.c
@@ -2077,8 +2077,14 @@ npyiter_find_buffering_setup(NpyIter *iter, npy_intp buffersize)
         npy_intp *strides = NAD_STRIDES(axisdata);
 
         for (int iop = 0; iop < nop; iop++) {
-            /* Check that we set things up nicely (if shape is ever 1) */
-            assert((axisdata->shape == 1) ? (prev_strides[iop] == strides[iop]) : 1);
+            /*
+             * Check that we set things up nicely so strides coalesc.  Except
+             * for index operands, which currently disrupts coalescing.
+             * NOTE(seberg): presumably `npyiter_compute_index_strides` should
+             * not set the strides to 0, but this was safer for backporting.
+             */
+            assert((axisdata->shape != 1) || (prev_strides[iop] == strides[iop])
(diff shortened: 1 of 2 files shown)


mattip:
Thanks @seberg

PR #29784, opened 2025-09-22
TYP: fix np.number and np.*integer method declaration

charris:
Backport of #29706.

Follow-up of #29695, partially addressing https://github.com/python/mypy/pull/15993#issuecomment-3240220599

This only leaves `floating`, `timedelta64`, and `datetime64`.


jorenham:
This apparently needs #29629

jorenham:
Primer diffs looks scarier than it is (see https://github.com/numpy/numpy/pull/29706#issuecomment-3263968978 for details), so this should be good now

PR #29278, opened 2025-06-26
TYP: fix overloads where `out: _ArrayT` was typed as being the default

MarcoGorelli:
There's a few places where `out: _ArrayT = ...` is used, even though the default for `out` is `None` (not `_ArrayT`)

PR #29785, opened 2025-09-22
TYP: mypy 1.18.1

charris:
Backport of #29749.

https://mypy.readthedocs.io/en/stable/changelog.html#mypy-1-18

It's should be significantly faster now (~0.67x in case of scipy-stubs), and solves some issues that used to lead to some incorrectly inferred types in our type-tests (e.g. `np.vstack`)

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -25,7 +25,7 @@ dependencies:
   - hypothesis
   # For type annotations
   - typing_extensions>=4.5.0
-  - mypy=1.16.1
+  - mypy=1.18.1
   - orjson  # makes mypy faster
   # For building docs
   - sphinx>=4.5.0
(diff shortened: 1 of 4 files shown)


jorenham:
This also includes #29278, which is what's causing the mypy_primer diff

charris:
I added #29278 to get rid of an mypy error. Is there a better way?
```
FAILED numpy/typing/tests/test_typing.py::test_reveal[array_constructors] - Failed: numpy/typing/tests/data/reveal/array_constructors.pyi:56 - reveal mismatch:

    error:
    Expression is of type "ndarray[tuple[Any, ...], dtype[Any]]", not "Any" 
    [assert-type]
        assert_type(np.concatenate([A, A]), Any)  # pyright correctly infers t...
        ^
```

jorenham:
> I added #29278 to get rid of an mypy error. Is there a better way?

The only relevant bit of #29278 is the change in `numpy/typing/tests/data/reveal/array_constructors.pyi:56`. Given the primer output, it's probably better to keep #29278 out of this, and only copy that specific change. (The reported `ndarray[tuple[Any, ...], dtype[Any]]` is the same as `npt.NDArray[Any]`)

And I see that you removed the problematic ignore comments in `numpy/typing/tests/data/reveal/testing.pyi` already, which is exactly what I would've done.

PR #29786, opened 2025-09-22
DOC: Add “How to diagnose numerical issues (precision, overflow, NaNs)”

Username46786:
Adds a new How-to page that helps users quickly triage precision loss,
overflow/underflow, and NaN/Inf issues, with links to errstate, seterr,
isclose/allclose, finfo/spacing, and sum accuracy notes.

New file: doc/source/user/how-to-numerical-issues.rst

ToC: added to user/howtos_index.rst
This aligns with NEP-44’s emphasis on practical how-tos. No API changes.

PR #29787, opened 2025-09-22
DOC: Add how-to-numerical-issues to howtos_index.rst

Username46786:
Follow-up to #29786.

Adds the new page `how-to-numerical-issues.rst` to the How-tos table of contents
(`doc/source/user/howtos_index.rst`) so that it appears in the rendered docs
navigation.

This was mentioned in the original PR description but not committed, and was
caught in review. Thanks for the reminder!


jorenham:
this should be in #29786

Issue #15320, opened 2020-01-12
TST: TestClip.test_NaT_propagation xfails unexpectedly


Kai-Striega:
The test TestClip.test_NaT_propagation xfails due to errors in the test, rather than the test conditions not being met. This is similar to #15319, where the parameter `arr` is expected then `a` is used by the test. However there are several more reasons for the test to fail prematurely:

1. Passing `nat` to `np.clip` is deprecated and raises a DepcrecationWarning
2. Cannot cast ufunc 'clip' input 1 from dtype('float64') to dtype('<m8') with casting rule 'same_kind'

I believe that these were changed previously and missed due to the test being expected to fail.

### Reproducing code example:
```
$ python runtests.py -v -t numpy/core/tests/test_numeric.py::TestClip::test_NaT_propagation -- --runxfail
Building, see build.log...
Build OK
NumPy version 1.19.0.dev0+b757fb3
NumPy relaxed strides checking option: True
============================= test session starts ==============================
platform linux -- Python 3.7.5, pytest-5.3.2, py-1.8.1, pluggy-0.13.1
rootdir: /mnt/c/Users/Kai/Develop/numpy-git, inifile: pytest.ini
plugins: forked-1.1.3, xdist-1.31.0
collected 1 item

numpy/core/tests/test_numeric.py F                                       [100%]

=================================== FAILURES ===================================
_______________ TestClip.test_NaT_propagation[arr0-amin0-amax0] ________________

self = <numpy.core.tests.test_numeric.TestClip object at 0x7f9b0cfde840>
arr = array([1, 1, 1, 1, 1, 1, 1, 1, 1, 1], dtype=timedelta64)
amin = numpy.timedelta64('NaT')
amax = array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0], dtype=int32)

    @pytest.mark.xfail(reason="propagation doesn't match spec")
    @pytest.mark.parametrize("arr, amin, amax", [
        (np.array([1] * 10, dtype='m8'),
         np.timedelta64('NaT'),
         np.zeros(10, dtype=np.int32)),
    ])
    def test_NaT_propagation(self, arr, amin, amax):
        # NOTE: the expected function spec doesn't
        # propagate NaT, but clip() now does
>       expected = np.minimum(np.maximum(a, amin), amax)
E       NameError: name 'a' is not defined

amax       = array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0], dtype=int32)
[...]

PR #29802, opened 2025-09-24
MAINT: Remove xfail and deprecation filter from a test.

WarrenWeckesser:
See https://github.com/numpy/numpy/issues/15320.

That issue is from numpy 1.19.


seberg:
Thanks, lgtm.  I guess we fixed NaT propagation for maximum/minimum at some point.

PR #28782, opened 2025-04-20
ENH: np.linalg.inv: Allow disabling error when one matrix is singular in a stack

math-hiyoko:
# Description
This PR introduces a new optional parameter noerr (default: False) to the numpy.linalg.inv function. The purpose of this parameter is to allow users to compute inverses of multiple matrices simultaneously (stacked matrices with shape (m, n, n)), without raising a LinAlgError if at least one matrix is singular. #27035 

# Example
```PYTHON
>>> import numpy as np
>>> a = np.array([
...    [[1, 0], [0, 1]],   # invertible
...    [[0, 0], [0, 0]],   # singular
... ])
>>> np.linalg.inv(matrices, noerr=True)
array([[[ 1.,  0.],
        [ 0.,  1.]],

       [[nan, nan],
        [nan, nan]]])
```

close #27035

PR #29811, opened 2025-09-25
BUG: linalg: emit a MemoryError on a malloc failure

ev-br:
Otherwise, a malloc failure in `init_gesv(...)` is not acted upon, and the python return value is silently wrong---`inv(non_zero_array)` returns an array of zeros.

On main, using the script under the fold, 
- limiting the memory to 30mb emits `numpy._core._exceptions._ArrayMemoryError: Unable to allocate 191. MiB for an array with shape (5000, 5000) and data type float64` from somewhere else in the gufunc machinery;
- limiting the memory to 300mb makes `res` be an array of all zeros, no error.

Note that I don't know what I'm doing with `NPY_ALLOW_C_API_DEF`  etc ---here I simply parrot the macros from elsewhere in this source file, e.g. https://github.com/numpy/numpy/blob/main/numpy/linalg/umath_linalg.cpp#L1202
The same issue seems present in other linalg functions, so if this fix looks reasonable, I'll extend to other functions.

<details>

```
import os
import psutil     # need to `pip install psutil`
import resource
import numpy as np


def set_limit(buffer):
     """
     Set a limit on process virtual memory space.
     """
     process = psutil.Process(os.getpid())
     memory_info = process.memory_info()
     used = memory_info.vms # current virtual memory space
     old_limit = resource.getrlimit(resource.RLIMIT_AS)
     new_limit = (used + buffer, old_limit[1]) # set new
     resource.setrlimit(resource.RLIMIT_AS, new_limit)


np.random.seed(123467)
arr = np.random.rand(25_000_000).reshape(5000, 5000)

set_limit(300 * 1024 * 1024) # ~300MB

res = np.linalg.inv(arr)
assert not (res == 0).all()
```

</details>

### DIFF
--- a/numpy/linalg/umath_linalg.cpp
+++ b/numpy/linalg/umath_linalg.cpp
@@ -1853,6 +1853,12 @@ using ftyp = fortran_type_t<typ>;
         END_OUTER_LOOP
 
         release_gesv(&params);
+    } else {
+        /* malloc failed in init_gesv() */
+        NPY_ALLOW_C_API_DEF
+        NPY_ALLOW_C_API;
+        PyErr_NoMemory();
+        NPY_DISABLE_C_API;
     }
 
     set_fp_invalid_or_clear(error_occurred);
--- a/numpy/linalg/umath_linalg.cpp
+++ b/numpy/linalg/umath_linalg.cpp
@@ -8,8 +8,6 @@
 #define PY_SSIZE_T_CLEAN
 #include <Python.h>
 
-#include <iostream>
-
 #define NPY_TARGET_VERSION NPY_2_1_API_VERSION
 #define NPY_NO_DEPRECATED_API NPY_API_VERSION
 #include "numpy/arrayobject.h"
--- a/numpy/linalg/umath_linalg.cpp
+++ b/numpy/linalg/umath_linalg.cpp
@@ -8,6 +8,8 @@
 #define PY_SSIZE_T_CLEAN
 #include <Python.h>
 
+#include <iostream>
+
 #define NPY_TARGET_VERSION NPY_2_1_API_VERSION
 #define NPY_NO_DEPRECATED_API NPY_API_VERSION
 #include "numpy/arrayobject.h"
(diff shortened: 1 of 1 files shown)


seberg:
Using `npy_gil_error` might be nice, but not helpful without changing more in the file probably.

Thanks, this is right, but TBH, a bit half hearted.  The same function is called in multiple places, but only one is fixed.
I would say: move the error setting into the `init_gesv` and also don't omit `init_evd` (maybe more?), then audit the actual usages if they set error again we can remove it.
(I.e. set error at the source, for "allocations" that might be a bit different but considering that we just repeat things otherwise...)

ev-br:
Thanks---yes, the initial attempt was exploratory, to double-check that the general direction makes sense. Now that you confirmed it (thanks!), the updated PR extends this check across `umath_linalg`. 

The error message is not as nice as it is for other memory failures which are reported through `errstate` (cf https://github.com/numpy/numpy/pull/28782#issuecomment-2829969890), but it's better than a wrong answer anyway.

numpy/linalg/umath_linalg.cpp
```diff
@@ -8,6 +8,8 @@
 #define PY_SSIZE_T_CLEAN
 #include <Python.h>
 
+#include <iostream>
+
```
seberg:
```suggestion
```
I don't think you need no iostream :).

numpy/linalg/umath_linalg.cpp
```diff
@@ -448,6 +450,15 @@ set_fp_invalid_or_clear(int error_occurred)
...
 }
 
+static inline void
+report_no_memory()
+{
+    NPY_ALLOW_C_API_DEF
+    NPY_ALLOW_C_API;
+    PyErr_NoMemory();
+    NPY_DISABLE_C_API;
+}
```
seberg, line 458: Let's go with this, `npy_gil_error`, got enough other places.

seberg (approved):
Thanks @ev-br!  Awesome, this looks pretty now *and* a cleanup :).

seberg:
Seems simple and enough and very safe to backport as well.

numpy/linalg/umath_linalg.cpp
ev-br: Ouch print debugging leftovers

PR #29812, opened 2025-09-25
BLD: refactor to avoid 'unused function' warnings'

mattip:
Closes #29805

In #29129 these functions were added but not properly masked off for builds that do not use them.


charris:
Thanks Matti.

Issue #29561, opened 2025-08-14
BUG: `np.log(arr, where=(arr > 0.0))` ignoring `where` optional parameter.


J0wsper:
### Describe the issue:

When using the `np.log` function and setting the `where` optional parameter to only include values greater than `0.0` on a 1-dimensional array, the function can simply ignore it. This can either result in less-than-zero values being squashed to 0.0 or having their sign flipped. Interestingly, the presence of `print` statements seem to influence this behavior. Additionally, printing unrelated variables can influence this behavior.

### Reproduce the code example:

```python
import numpy as np

# Flipping sign
vals = np.empty((5,))
vals.fill(-np.finfo("float32").max)
print(vals)
log_vals = np.log(vals, where=(vals > 0.0))
print(log_vals)
# Expected output: [-3.40282347e+38 -3.40282347e+38 -3.40282347e+38 -3.40282347e+38 -3.40282347e+38]
# Actual output: [3.40282347e+38 3.40282347e+38 3.40282347e+38 3.40282347e+38 3.40282347e+38]

# Squashing to 0.0.
# NOTE: This will only squash to 0 if you run this script independently of the previous script.
vals2 = np.empty((5,))
vals2.fill(-np.finfo("float32").max)
log_vals2 = np.log(vals2, where=(vals2 > 0.0))
print(log_vals2)
# Expected output: [-3.40282347e+38 -3.40282347e+38 -3.40282347e+38 -3.40282347e+38 -3.40282347e+38]
# Actual output: non-deterministic small values near 0.0
```

### Error message:

```shell

```

### Python and NumPy Versions:

`python`: 3.12.11
`numpy`: 2.2.6

### Runtime Environment:

```
[{'numpy_version': '2.2.6',
  'python': '3.12.11 (main, Jul 11 2025, 22:43:48) [Clang 20.1.4 ]',
  'uname': uname_result(system='Linux', node='spacetee', release='6.15.6-1-default', version='#1 SMP PREEMPT_DYNAMIC Fri Jul 11 05:06:35 UTC 2025 (a69e09e)', machine='x86_64')},
 {'simd_extensions': {'baseline': ['SSE', 'SSE2', 'SSE3'],
                      'found': ['SSSE3',
                                'SSE41',
                                'POPCNT',
                                'SSE42',
                                'AVX',
                                'F16C',
                                'FMA3',
                                'AVX2'],
                      'not_found': ['AVX512F',
                                    'AVX512CD',
[...]

Issue #17192, opened 2020-08-31
Default behavior of ufunc returns uninitialized values


tbenst:
First of all, thank you so much for the work on Numpy--an incredible contribution to science and progress.

### Reproducing code example:

Numpy has excellent support for functional programming patterns, and the hundreds of `np.*` functions I regularly use act as pure functions with no side effects.

The current behavior of `np.negative` breaks this pattern, and is not deterministic:

```python
>>> import numpy as np
>>> np.negative(np.arange(5), where=np.array([0,0,1,1,0],dtype=bool))
array([     94871006724704,        171798691860,                  -2,
                        -3, 4285579895102272626])
>>> np.negative(np.arange(5), where=np.array([0,0,1,1,0],dtype=bool))
array([ 94871006750048, 140541164903200,              -2,              -3,
                    48])
>>> x = np.arange(5)
>>> np.negative(x,out=x, where=np.array([0,0,1,1,0],dtype=bool))
array([ 0,  1, -2, -3,  4])
```

Thus, np.negative does not act as a mathematical function, which is confusing and error-prone. It's hard to imagine that a user would ever want uninitialized values in the returned values. 

 A quick search on github reveals that there are thousands (if not hundreds of thousands) of instances where code introduces uninitialized values due to calls to `np.negative`: https://github.com/search?q=%22np.negative%22&type=Code

Based on reading #7158, I believe this issue applies to all `ufunc`. I understand in https://github.com/numpy/numpy/pull/11086 that the documentation was improved, but I wanted to open this issue to highlight both the danger of uninitialized values being returned, as well as how frequently in the wild these uninitialized values are silently corrupting data.

### Numpy/Python version information:

```python
>>> import sys, numpy; print(numpy.__version__, sys.version)
1.19.1 3.8.5 (default, Aug  5 2020, 08:36:46) 
[GCC 7.3.0]
```

mattip:
The performance price would be a `memcpy` of the contents of (one of) the input to the newly-created output, possibly broadcast to match the output shape. What would the default value be for binary operations like `add`, or `matmul`, or for reduce operations like `sum` ?

eric-wieser:
> A quick search on github reveals that there are thousands (if not hundreds of thousands) of instances where code introduces uninitialized values due to calls to `np.negative`: https://github.com/search?q=%22np.negative%22&type=Code

I don't really believe this analysis. Only when `where` is specified can there be a danger here, and even then there are safe ways to use `where`.

[...]

PR #29813, opened 2025-09-25
ENH: add warning when calling ufunc with 'where' and without 'out'

### LINKED ISSUE #29804: BUG: Weird memory corruption bug related to `np.divide`
### Describe the issue:

The following code fragment is somehow accessing uninitialized memory when it's run with the print statements left out. This took me by surprise, and it caused errors in an RL algorithm I am working on. I was just sketching it out using Numpy. I'll try using `nan_to_num` to get around the divide by 0 errors instead.

```python
import numpy as np

expected_values_nom_values = np.array([
    [0,0,0],
    [0,0,0],
    [0,0,0],
],dtype=np.float32)

expected_values_den_values = np.array([
    [0,0,0],
    [0,0,0],
    [0,0,0],
],dtype=np.float32)

def expected_values():
    a,b = expected_values_nom_values[0,:], expected_values_den_values[0,:]
    # print(f"a: {a}")
    # print(f"b: {b}")
    return np.divide(a, b, where=(b != 0))

print(expected_values())
```

### Error message:

```shell
Here is what happens when I run it in the terminal


(mrakgr) mrakgr@Marko:~/Spiral-s-ML-Library$ python /home/mrakgr/Spiral-s-ML-Library/hd-cfr-python/bug.py
[1.8983535e+03 4.5584239e-41 3.9749163e-31]
(mrakgr) mrakgr@Marko:~/Spiral-s-ML-Library$ python /home/mrakgr/Spiral-s-ML-Library/hd-cfr-python/bug.py
[-1.039874e-09  4.579023e-41  5.763683e-10]
(mrakgr) mrakgr@Marko:~/Spiral-s-ML-Library$ python /home/mrakgr/Spiral-s-ML-Library/hd-cfr-python/bug.py
[1.8641260e-37 4.5823861e-41 4.0987058e-20]
(mrakgr) mrakgr@Marko:~/Spiral-s-ML-Library$ python /home/mrakgr/Spiral-s-ML-Library/hd-cfr-python/bug.py
[-9.5116059e+30  4.5835071e-41  2.2177525e-15]
(mrakgr) mrakgr@Marko:~/Spiral-s-ML-Library$ python /home/mrakgr/Spiral-s-ML-Library/hd-cfr-python/bug.py
[-7.6582234e+23  4.5606660e-41  1.4948110e-20]
(mrakgr) mrakgr@Marko:~/Spiral-s-ML-Library$ python /home/mrakgr/Spiral-s-ML-Library/hd-cfr-python/bug.py
[-1.4026275e+29  4.5724369e-41  1.6299158e-21]
(mrakgr) mrakgr@Marko:~/Spiral-s-ML-Library$ python /home/mrakgr/Spiral-s-ML-Library/hd-cfr-python/bug.py
[1.6858093e-10 4.5591246e-41 5.0520033e-10]


Here is what happens when I run it with the print statements uncommented:


(mrakgr) mrakgr@Marko:~/Spiral-s-ML-Library$ python /home/mrakgr/Spiral-s-ML-Library/hd-cfr-python/bug.py
a: [0. 0. 0.]
b: [0. 0. 0.]
[0. 0. 0.]
(mrakgr) mrakgr@Marko:~/Spiral-s-ML-Library$ python /home/mrakgr/Spiral-s-ML-Library/hd-cfr-python/bug.py
a: [0. 0. 0.]
b: [0. 0. 0.]
[0. 0. 0.]
(mrakgr) mrakgr@Marko:~/Spiral-s-ML-Library$ python /home/mrakgr/Spiral-s-ML-Library/hd-cfr-python/bug.py
a: [0. 0. 0.]
b: [0. 0. 0.]
[0. 0. 0.]
```

2.3.3

### Runtime Environment:

[{'numpy_version': '2.3.3',
  'python': '3.12.3 (main, Aug 14 2025, 17:47:21) [GCC 13.3.0]',
  'uname': uname_result(system='Linux', node='Marko', release='5.15.167.4-microsoft-standard-WSL2', version='#1 SMP Tue Nov 5 00:21:55 UTC 2024', machine='x86_64')},
                      'found': ['SSSE3',
                                'SSE41',
                                'POPCNT',
                                'SSE42',
                                'AVX',
                                'F16C',
                                'FMA3',
                                'AVX2',
                                'AVX512F',
                                'AVX512CD',
                                'AVX512_SKX',
                                'AVX512_CLX',
                                'AVX512_CNL',
                                'AVX512_ICL'],
                      'not_found': ['AVX512_KNL', 'AVX512_KNM', 'AVX512_SPR']}},
 {'architecture': 'SkylakeX',
  'filepath': '/home/mrakgr/.venv/lib/python3.12/site-packages/numpy.libs/libscipy_openblas64_-8fb3d286.so',
  'internal_api': 'openblas',
  'num_threads': 16,
  'user_api': 'blas',
  'version': '0.3.30'}]

### Context for the issue:

_No response_

mattip:
Closes #29804, related to # #17192, #29561 and maybe more

Users naively add a `where` argument without a proper `out` argument, and are surprised when the `False` values of the where mask are uninitialized. This PR adds a warning. The warning can be surpressed by using `where=mask, out=None`. That came in handy in some obscure tests.

### DIFF
--- a/numpy/_core/src/umath/ufunc_object.c
+++ b/numpy/_core/src/umath/ufunc_object.c
@@ -4486,6 +4486,15 @@ ufunc_generic_fastcall(PyUFuncObject *ufunc,
         return override;
     }
 
+    /* Warn if "where" is used without "out", issue 29561 */
+    if ((where_obj != NULL) && (full_args.out == NULL) && (out_obj != Py_None)) {
+        if (PyErr_Warn(PyExc_RuntimeWarning,
+                "'where' used without 'out', expect unitialized memory in output. "
+                "If this is intentional, use out=None.") < 0) {
+            goto fail;
+        }
+    }
+
     if (outer) {
         /* Outer uses special preparation of inputs (expand dims) */
         PyObject *new_in = prepare_input_arguments_for_outer(full_args.in, ufunc);
--- a/numpy/_core/src/umath/ufunc_object.c
+++ b/numpy/_core/src/umath/ufunc_object.c
@@ -4487,7 +4487,7 @@ ufunc_generic_fastcall(PyUFuncObject *ufunc,
     }
 
     /* Warn if "where" is used without "out", issue 29561 */
-    if ((where_obj != NULL) && (full_args.out == NULL) && (out_obj != Py_None)) {
+    if ((where_obj != NULL) && (full_args.out == NULL) && (out_obj == NULL)) {
         if (PyErr_Warn(PyExc_UserWarning,
                 "'where' used without 'out', expect unitialized memory in output. "
                 "If this is intentional, use out=None.") < 0) {
--- a/numpy/_core/src/umath/ufunc_object.c
+++ b/numpy/_core/src/umath/ufunc_object.c
@@ -4488,7 +4488,7 @@ ufunc_generic_fastcall(PyUFuncObject *ufunc,
 
     /* Warn if "where" is used without "out", issue 29561 */
     if ((where_obj != NULL) && (full_args.out == NULL) && (out_obj != Py_None)) {
-        if (PyErr_Warn(PyExc_RuntimeWarning,
+        if (PyErr_Warn(PyExc_UserWarning,
                 "'where' used without 'out', expect unitialized memory in output. "
                 "If this is intentional, use out=None.") < 0) {
             goto fail;
--- a/numpy/_core/tests/test_ufunc.py
+++ b/numpy/_core/tests/test_ufunc.py
@@ -1743,11 +1743,13 @@ class TestUfunc:
     def test_where_warns(self):
         a = np.arange(7)
         mask = a % 2 == 0
-        with warnings.catch_warnings(record=True) as w_where:
-            warnings.simplefilter('always')
-            result = np.add(a, a, where=mask)
-        assert len(w_where) == 1
-        assert "'where'" in str(w_where[0])
+        with pytest.warns(UserWarning, match="'where' used without 'out'"):
+            result1 = np.add(a, a, where=mask)
+        # Does not warn
+        result2 = np.add(a, a, where=mask, out=None)
+        # Sanity check
+        assert np.all(result1[::2] == [0, 4, 8, 12])
+        assert np.all(result2[::2] == [0, 4, 8, 12])
 
     @staticmethod
     def identityless_reduce_arrs():
--- a/numpy/_core/tests/test_ufunc.py
+++ b/numpy/_core/tests/test_ufunc.py
@@ -1710,9 +1710,6 @@ class TestUfunc:
         assert_equal(a, [[0, 27], [14, 5]])
 
     def test_where_param_buffer_output(self):
-        # This test is temporarily skipped because it requires
-        # adding masking features to the nditer to work properly
-
         # With casting on output
         a = np.ones(10, np.int64)
         b = np.ones(10, np.int64)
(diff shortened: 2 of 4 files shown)


numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4486,6 +4486,15 @@ ufunc_generic_fastcall(PyUFuncObject *ufunc,
         return override;
     }
 
+    /* Warn if "where" is used without "out", issue 29561 */
+    if ((where_obj != NULL) && (full_args.out == NULL) && (out_obj != Py_None)) {
+        if (PyErr_Warn(PyExc_RuntimeWarning,
```
seberg: I would go with a `UserWarning`.

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4486,6 +4486,15 @@ ufunc_generic_fastcall(PyUFuncObject *ufunc,
         return override;
     }
 
+    /* Warn if "where" is used without "out", issue 29561 */
+    if ((where_obj != NULL) && (full_args.out == NULL) && (out_obj != Py_None)) {
```
```suggestion
    if ((where_obj != NULL) && (full_args.out == NULL) && (out_obj == NULL)) {
```
Seems a bit clearer to me?

numpy/_core/tests/test_ufunc.py
```diff
@@ -1743,6 +1740,15 @@ def test_where_with_broadcasting(self):
         assert_array_equal((a[where] < b_where), out[where].astype(bool))
         assert not out[~where].any()  # outside mask, out remains all 0
 
+    def test_where_warns(self):
+        a = np.arange(7)
+        mask = a % 2 == 0
+        with warnings.catch_warnings(record=True) as w_where:
```
seberg: Please use `with pytest.warns()`, also removes the need for that `'where' in str(...)`

numpy/_core/tests/test_umath.py
```diff
@@ -3734,6 +3735,7 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
                     else:
                         kwargs["where"] = kwargs["where"][0]
 
+                print(f"{kwargs=}")
```
seberg:
```suggestion
```

seberg (approved):
Thanks!  I assume this is rare enough that introducing a warning is not a problem and it is very simple to remove the warning after all.

So I am in favor of trying this, approving, but the comments should be (mostly) addressed.

Test changes look good, the only ones I would worry about would be `__array_wrap__` related, and those don't seem touched.

mhvk (commented):
Yes, nice; we really should have done this earlier...

numpy/_core/src/umath/ufunc_object.c
mattip: It seems clearer to me to compare to the desired `Py_None` than to rely on logic

seberg: I suppose if you see it this way, but maybe we should make it:
```
!out_is_passed_by_position && out_obj == NULL
```

that makes it clear: out was not passed explicitly.  But more importantly, there is a concrete reason to avoid `!= None` and that is:
```
np.frexp(4., out=(None, None), where=False)
```

mattip: OK, let's go with your suggestion

mhvk: Also need to avoid a case where `out` is an actual array!

mhvk (approved):
Nice! (I think my last comment was wrong, though)

andyfaff:
Is it possible to add a stacklevel value to the warning, because it'd be nice to see the warnings caller.

seberg:
The warning already has the best possible stacklevel since it's in C (so no need to add one).  But of course sometimes ufuncs are called indirectly (e.g. mean/var maybe).

I am not sure if there is a nice way to add `skip_file_prefixes` in C, although I guess one could bother to just call into Python to do it (warnings are slow anyway, I guess).
(This would also have to be Python 3.12 only as it is a new feature)

PR #28774, opened 2025-04-19
TYP: fix `count_nonzero` signature

charris:
Backport of #28750.

This fixes the following valid usage in https://github.com/data-apis/array-api-compat/blob/16978e63b744185bc4ced4677c822468e6ddebd2/array_api_compat/numpy/_aliases.py#L126-L127 from being falsely rejected:

```py
def count_nonzero(
    x: Array,
    axis: int | tuple[int, ...] | None = None,
    keepdims: py_bool = False,
) -> Array:
    result = np.count_nonzero(x, axis=axis, keepdims=keepdims)
```

Issue #17562, opened 2020-10-14
`count_nonzero` can return either a builtin `int` or numpy's `intp`


BvB93:
The return type of [`np.count_nonzero()`](https://numpy.org/doc/stable/reference/generated/numpy.count_nonzero.html) is currently somewhat inconsistent:
* A builtin integer is returned if `axis is None and not keepdims`.
* An `np.intp` (or an array) is returned otherwise.

The former case is handled by the an identically named function in the `np.core._multiarray_umath` 
module (which apparently always returns an `int`) while the latter is effectively a wrapper around 
`np.ndarray.sum()` with its dtype explicitly set to `np.intp`:

https://github.com/numpy/numpy/blob/4ccfbe69b7a8c511d262a9759e576bb87ec119cc/numpy/core/numeric.py#L484-L495

### Reproducing code example:
MacOS 10.15.6; Python 3.8.5; NumPy 1.20.0.dev0+eb2c751

```python
In [1]: import numpy as np

In [2]: ar = np.arange(10)

In [3]: np.count_nonzero(ar).__class__
Out[3]: int

In [4]: np.count_nonzero(ar, axis=0).__class__
Out[4]: numpy.int64  # i.e. np.intp
```

mattip:
> The former case is handled by the an identically named function in the `np.core._multiarray_umath` module (which apparently always returns an int)

It returns a `np.int64` AFAICT. Did something change at some point since this issue was opened?

WarrenWeckesser:
The change was https://github.com/numpy/numpy/pull/28615/commits/ae8b43d6647d0a8f6bf2490f5ecfdfaab4e548e4.  It is mentioned in the 2.3.0 release notes: https://numpy.org/doc/stable/release/2.3.0-notes.html#changes (fourth bullet point).

mattip:
Cool, thanks. Then we can close this?

WarrenWeckesser:
Fixed by https://github.com/numpy/numpy/commit/ae8b43d6647d0a8f6bf2490f5ecfdfaab4e548e4.

PR #29815, opened 2025-09-25
DOC: Update docstring for `count_nonzero`

penguinpee:
The return type of the function was changed in 2.3.0, but the docstring
remained the same (gh-28774).

Addresses gh-17562
(addressed by the change to `count_nonzero` but not yet closed)

### DIFF
--- a/numpy/_core/numeric.py
+++ b/numpy/_core/numeric.py
@@ -523,11 +523,11 @@ def count_nonzero(a, axis=None, *, keepdims=False):
     --------
     >>> import numpy as np
     >>> np.count_nonzero(np.eye(4))
-    4
+    np.int64(4)
     >>> a = np.array([[0, 1, 7, 0],
     ...               [3, 0, 2, 19]])
     >>> np.count_nonzero(a)
-    5
+    np.int64(5)
     >>> np.count_nonzero(a, axis=0)
     array([1, 1, 2, 1])
     >>> np.count_nonzero(a, axis=1)


mattip:
Grepping around in the docstrings via vim's `vimgrep "+y/Returns\n.*\n.*:.*array of int/ numpy/**/*.py`, I see we have a mix of language:
- `left_shift` has "array of int" when it can return a numpy scalar int or a ndarray of int
- `argmax`, `argmin` have "ndarray of ints" when they can return a numpy scalar int or a ndarray of int
- `searchsorted`, `count_nonzero` have "int or array of ints" when they can return a numpy scalar or a ndarray of int
- Reductions and accumulations has "ndarray" but mention scalars in the text.

There may be more?

I think it is OK to leave the text as-is. Personally I don't like the `np.intp` reference and find the current text preferable. It is consistent with other similar usages.

penguinpee:
Well, for `count_nonzero` the return type was explicitly changed to be _always_ `np.intp`. It used to be `int` or `np.intp`.

I think at least the example snippet should be changed to reflect the actual output since the change.

seberg:
> Personally I don't like the np.intp reference and find the current text preferable. It is consistent with other similar usages.

Agreed, we return a NumPy `int`, which is an `intp`/`int64` (assuming 64bit system) here.  Changing the `>>>` is good (although there will be many similar places), but I would leave `int or array of int`.  The only weird thing is that `int` might be a Python int, but we rarely return Python integers.

penguinpee:
Okay. I reverted to `int or array of int` for what the function returns.

> The only weird thing is that int might be a Python int, but we rarely return Python integers.

Well, in this particular case it used to be one or the other as mentioned before. So, I thought being explicit in what the function returns might be helpful. But if `int` in NumPy docs is generally understood to be a NumPy `int` (aka `intp`) I'm fine with leaving it as is. With the change in the code snippet it can at least be inferred.

mattip:
Thanks @penguinpee

PR #29816, opened 2025-09-26
TST: Mark thread-unsafe tests

bwhitt7:
This PR has the goal of making it so the entire NumPy test suite can be run under `pytest-run-parallel` ([link](github.com/Quansight-Labs/pytest-run-parallel])). It utilizes `pytest-run-parallel`'s `thread_unsafe` marker to make it so thread-unsafe tests will be ran in a single thread. You can check out my work and goals concerning the test suite's thread safety in this issue: #29552

Previously I have submitted PRs that refactor the test suite to make it more thread safe. However, not all tests can be made thread safe, with some tests using thread-unsafe features in `pytest`, NumPy, or even Python itself. Some of these issues are fixable, but for now, this PR marks the rest of these tests as thread-unsafe until they may be fixed.

`pytest-run-parallel` is currently not included with the test_requirements, so you will need to install it yourself using `pip install pytest-run-parallel` (if there is interest to include it in the pip requirements, I can definitely add that). If you have installed it previously, make sure to update it since it had a recent release that makes the `tmp_path` and `tmpdir` fixtures thread-safe. I have made some edits to `conftest.py` to make it so `pytest-run-parallel` isn't required.

You can then test out this branch by running the test suite with `spin test -- --parallel-threads=auto --skip-thread-unsafe=true`. I encourage folks to test out this branch, since it's very possible I missed some tests that fail in different environments and conditions (I'm currently running everything under a WSL Ubuntu environment). Ideally this should be ran with `python 3.14t` due to some thread-safety issues that aren't marked here (such as issues with `warnings`) being fixed in that version.

ping @ngoldbaum @lysnikolaou


ngoldbaum:
To simplify this, maybe it makes sense to have a submodule-wide thread-unsafe marker for the f2py tests? I think you can do that by creating a `conftest.py` in the root of the submodule and adding the marker there.

While we probably do care about testing libraries built using f2py in multithreaded scenarios, I don't think we care much about building f2py extensions simultaneously in threads, and that's what most of the f2py tests are focused on.

numpy/_core/tests/test_arrayprint.py
```diff
@@ -546,6 +547,7 @@ def test_any_text(self, text):
         assert_equal(result, expected_repr)
 
     @pytest.mark.skipif(not HAS_REFCOUNT, reason="Python lacks refcounts")
+    @pytest.mark.thread_unsafe(reason="sys.getrefcounts is thread-unsafe")
```
ngoldbaum: It's not so much that `getrefcounts` is thread-unsafe, it's that the test depends on global state. In this case in the garbage collector, because it temporarily disables the gc for the whole process, including other threads running this test.

numpy/_core/tests/test_arrayprint.py
```diff
@@ -533,6 +533,7 @@ def test_nested_array_repr(self):
         )
 
     @given(hynp.from_dtype(np.dtype("U")))
+    @pytest.mark.thread_unsafe(reason="gives unreliable results w/ hypothesis")
```
ngoldbaum: Hypothesis is supposed to be thread-safe. I'd like to understand this a little better.

numpy/_core/tests/test_cpu_features.py
```diff
@@ -124,6 +124,7 @@ def load_flags_auxv(self):
         " therefore this test class cannot be properly executed."
     ),
 )
+@pytest.mark.thread_unsafe(reason="setup & tmp_path_factory threads-unsafe, modifies environment variables")
```
ngoldbaum, line 127: Didn't we fix most of these issues? That said, the environment variable mutation isn't safe.

numpy/_core/tests/test_custom_dtypes.py
```diff
@@ -298,6 +299,7 @@ def test_flatiter_index(self, index):
         np.testing.assert_array_equal(
             arr.view(np.float64), arr2.view(np.float64))
 
+@pytest.mark.thread_unsafe(reason="_ScaledFloatTestDType setup is thread-unsafe")
```
ngoldbaum: Let's file an issue for this problem that we can link to from these messages. That way if anyone comes along and decides to fix it, they can easily update these tests too.

numpy/_core/tests/test_deprecations.py
```diff
@@ -508,6 +508,7 @@ def assign_to_index():
         self.assert_deprecated(assign_to_index)
 
 
+@pytest.mark.thread_unsafe(reason="is deprecated due to being thread-unsafe")
```
ngoldbaum:
```suggestion
@pytest.mark.thread_unsafe(reason="warning control utilities are deprecated due to being thread-unsafe")
```

numpy/_core/tests/test_dtype.py
```diff
@@ -1916,6 +1917,7 @@ def test_pairs(self, pair):
 
 class TestUserDType:
     @pytest.mark.leaks_references(reason="dynamically creates custom dtype.")
+    @pytest.mark.thread_unsafe(reason="crashes when GIL disabled")
```
ngoldbaum: I think the crash is because legacy user dtype setup isn't thread-safe? I think we should explain why the crash is happening in this message.

numpy/_core/tests/test_multiarray.py
```diff
@@ -5577,6 +5577,7 @@ def test_invalid_axis(self):  # gh-7528
         x = np.linspace(0., 1., 42 * 3).reshape(42, 3)
         assert_raises(AxisError, np.lexsort, x, axis=2)
 
+@pytest.mark.thread_unsafe(reason="tmp_filename is thread-unsafe")
```
ngoldbaum: IMO we should just make the fixture thread-safe in numpy. You can do that by adding random data or unique thread-local data (e.g. the thread ID) to the temporary path.

ngoldbaum (commented):
Still not fully finished reviewing but I figured I'd send in the comments I have so far.

rgommers:
> To simplify this, maybe it makes sense to have a submodule-wide thread-unsafe marker for the f2py tests? I think you can do that by creating a `conftest.py` in the root of the submodule and adding the marker there.

Agreed, that is a good idea. The marker should not go in a `conftest.py` but in `__init__.py`, there are already a couple for `f2py`: https://github.com/numpy/numpy/blob/main/numpy/f2py/tests/__init__.py

mhvk:
Small fly-by comment: to avoid regressions, I would suggest to do your parallel tests in one of the CI runs (maybe repurpose one, surely we have too large a number already!). It will help with future fixing of those tests as well.

ngoldbaum:
CI is the followup after this!

numpy/_core/tests/test_multiarray.py
```diff
@@ -5829,6 +5830,7 @@ def test_fromfile_offset(self, tmp_filename):
                     sep=",", offset=1)
 
     @pytest.mark.skipif(IS_PYPY, reason="bug in PyPy's PyNumber_AsSsize_t")
+    @pytest.mark.thread_unsafe(reason="os.dup thread-unsafe?")
```
ngoldbaum: this test is thread-unsafe because of monkeypatching, not because of os.dup

numpy/_core/tests/test_multiarray.py
```diff
@@ -6375,6 +6378,7 @@ def test_multifield_indexing_view(self):
         v[:] = (4, 5)
         assert_equal(a[0].item(), (4, 1, 5))
 
+
```
ngoldbaum: Please delete the whitespace-only changes in this PR, to make it a little easier to review.

numpy/_core/tests/test_multiarray.py
```diff
@@ -8492,6 +8497,7 @@ def test_padding(self):
             x = np.array([(1,), (2,)], dtype={'f0': (int, j)})
             self._check_roundtrip(x)
 
+    @pytest.mark.thread_unsafe(reason="sys.getrefcounts is thread-unsafe")
```
ngoldbaum: It's not so much that getrefcounts is thread-unsafe, it's because testing for refcounts on a global object in a multithreaded test isn't guaranteed to always give the same results because of order of operation. A better reason would be "test result depends on the reference count of a global object"

numpy/_core/tests/test_multiarray.py
```diff
@@ -8613,6 +8619,7 @@ class foo(ctypes.Structure):
         assert_equal(arr['a'], 3)
 
     @pytest.mark.parametrize("obj", [np.ones(3), np.ones(1, dtype="i,i")[()]])
+    @pytest.mark.thread_unsafe(reason="_multiarray_tests is thread-unsafe?")
```
ngoldbaum: It looks like this test uses memoryview, which is thread-unsafe.

numpy/_core/tests/test_multiarray.py
```diff
@@ -9009,6 +9016,7 @@ def __array_interface__(self):
         # This fails due to going into the buffer protocol path
         (f, {'data': None, 'shape': ()}, TypeError),
         ])
+    @pytest.mark.thread_unsafe(reason="setup is thread_unsafe, sys.getrefcounts is thread-unsafe")
```
ngoldbaum: Didn't we fix all the thread-unsafe setups? It also looks like this is getting the reference count of a global, which is thread-unsafe.

I think I'm going to stop here. I'd like you to go through the rest of these messages and make sure that they're all as up-to-date as you can make them, given the work you've already done. I'd also like you to double-check the reasons you're using. Feel free to check with Lysandros if you want to get a double-check on your reasoning.

bwhitt7:
Thank you for the reviews @ngoldbaum ! I'll take a look through these and fix things up tomorrow.

bwhitt7:
> Agreed, that is a good idea. The marker should not go in a `conftest.py` but in `__init__.py`, there are already a couple for `f2py`: https://github.com/numpy/numpy/blob/main/numpy/f2py/tests/__init__.py

@rgommers I'm taking a look at how to use markers for modules, and it seems to be a slightly different process from using something like pytest.skip. Looks like theres a [pytestmark variable](https://docs.pytest.org/en/stable/example/markers.html#marking-whole-classes-or-modules) that pytest will check for, but it does involve importing this variable into every test file. Do you know of an alternative solution to this? Otherwise I can go forward with this.

rgommers:
Ah, didn't quite see that difference coming. If a one-liner like `pytestmark = [pytest.mark.thread_unsafe]` doesn't work at the `__init__.py` level, then adding that line at the top of each file doesn't seem terrible. Not sure if Nathan's `conftest.py` does work, that's also an option then (fewer lines of code, but an extra config file - it's a toss-up).

bwhitt7:
@rgommers Looks like you can mark up the module using conftest.py. This is how I did it.
```python
import pytest

def pytest_itemcollected(item):
    item.add_marker(pytest.mark.thread_unsafe(reason="f2py is thread-unsafe"))
```
It is an extra file, but its just a few lines of code vs. adding something to every test file in the f2py module.

bwhitt7:
Pushed some changes, mainly rewording some reasons as per Nathan's suggestions, and moving the f2py markers to the new `conftest.py` file. I can definitely change this and instead put `pytestmark` in each f2py test file if folks think that's a better solution.

There are some tests I marked that can be fixed to be more thread-safe (such as some setup methods sticking around that had other thread safety issues), so I may make a separate PR to handle those.

numpy/_core/tests/test_cpu_features.py
lysnikolaou: It looks like this is starting processes with setting different environment variables which are global for the class. It could be made thread safe but probably not worth the effort.

numpy/_core/tests/test_regression.py
```diff
@@ -2386,6 +2386,7 @@ def test_dtype_scalar_squeeze(self):
                 assert_equal(squeezed, scvalue)
                 assert_equal(type(squeezed), type(scvalue))
 
+    @pytest.mark.thread_unsafe(reason="immortal string, sys.getrefcounts is thread-unsafe")
```
lysnikolaou: Is this string immortalized/interned by NumPy? Otherwise I don't think it should be immortal, right?

numpy/_core/tests/test_ufunc.py
```diff
@@ -1762,6 +1762,7 @@ def identityless_reduce_arrs():
 
     @pytest.mark.parametrize("a", identityless_reduce_arrs())
     @pytest.mark.parametrize("pos", [(1, 0, 0), (0, 1, 0), (0, 0, 1)])
+    @pytest.mark.thread_unsafe(reason="races to modify parameters")
```
lysnikolaou: Maybe we can copy `a` into a new object before modifying/using it here? The test doesn't seem to depend on the object itself. Ditto for the other "races to modify parameters" test (not sure if that's also relevant there).

numpy/f2py/tests/conftest.py
```diff
@@ -0,0 +1,5 @@
+import pytest
+
+
+def pytest_itemcollected(item):
```
lysnikolaou: I think we should be wrapping this with `@pytest.hookimpl(tryfirst=True)` to make sure it's executed before pytest-run-parallel's `pytest_itemcollected` is run. I'm not too familiar with how `pluggy` orders hooks, but I think it could theoretically re-order them, if we don't explicitly say `tryfirst=True`.

numpy/lib/tests/test__datasource.py
```diff
@@ -88,6 +88,7 @@ def invalid_httpfile():
     return http_fakefile
 
 
+@pytest.mark.thread_unsafe(reason="mkdtemp thread-unsafe")
```
lysnikolaou, line 91: `mkdtemp` could be make thread safe here by adding the thread id as a prefix/suffix. However, since all tests use `self.ds`, that's global state which makes it harder to make the tests thread safe. Should probably give that as the reason.

Also, if `datasource.DataSource` is not thread safe, that should probably be documented.

numpy/lib/tests/test_io.py
```diff
@@ -193,6 +193,7 @@ def test_format_2_0(self):
             self.check_roundtrips(a)
 
 
+@pytest.mark.thread_unsafe(reason="some tests fail from RountripTest, file race?")
```
lysnikolaou: I can't reproduce this and by quickly looking over what `roundtrip` does, it should't be failing. Can you provide more info on how this fails?

numpy/lib/tests/test_loadtxt.py
```diff
@@ -74,13 +74,15 @@ def mixed_types_structured():
 
 
 @pytest.mark.parametrize('skiprows', [0, 1, 2, 3])
+@pytest.mark.thread_unsafe(reason="mixed_types_structured fixture is thread-unsafe?")
```
lysnikolaou: I can't reproduce this either and it doesn't look like this should be failing either. Can you give us more info about what the error might be?

numpy/ma/tests/test_core.py
```diff
@@ -1046,6 +1046,7 @@ def test_mvoid_iter(self):
         # w/ mask
         assert_equal(list(a[1]), [masked, 4])
 
+    @pytest.mark.thread_unsafe(reason="masked_print_option thread unsafe?")
```
lysnikolaou: Yeah, `masked_print_option.set_display("-X-")` looks like global state.

numpy/polynomial/tests/test_printing.py
```diff
@@ -246,6 +246,7 @@ def test_linewidth_printoption(self, lw, tgt):
                 assert_(len(line) < lw)
 
 
+@pytest.mark.thread_unsafe(reason="set_default_printstyle() is thread-unsafe?")
```
lysnikolaou: Yup, looks like global state again.

numpy/matrixlib/tests/test_matrix_linalg.py
```diff
@@ -81,6 +83,7 @@ class TestDetMatrix(DetCases, MatrixTestCase):
     pass
 
 
+@pytest.mark.thread_unsafe(reason="residuals not calculated properly for square tests")
```
lysnikolaou: This looks like a real bug maybe? We probably need to dive a bit deeper. Certainly warrants an issue.

numpy/random/tests/test_generator_mt19937.py
```diff
@@ -1266,6 +1266,7 @@ def test_dirichlet_small_alpha(self):
         assert_array_almost_equal(actual, expected, decimal=15)
 
     @pytest.mark.slow
+    @pytest.mark.thread_unsafe(reason="crashes")
```
lysnikolaou: This should probably be documented as well if it's not.

lysnikolaou (commented):
Thanks for this @bwhitt7! I had a first look at this and identified some places, where things can be improved (maybe).

numpy/lib/tests/test_loadtxt.py
bwhitt7: https://gist.github.com/bwhitt7/2e679f124e7268b1d9a5c3309a217914
This is the error I get when running `test_loadtxt.py`. I marked `mixed_types_structured` being the issue since both failing tests had that fixture. Not sure if that's the source, especially since I'm having trouble figuring out where this error is coming from. Looks like `np.loadtxt` isn't working properly here.

I'm running with 3.14t-dev on WSL with `spin test -- numpy/lib/test/test_loadtxt.py --parallel-threads=auto`

numpy/lib/tests/test_io.py
bwhitt7: https://gist.github.com/bwhitt7/6991209b4d5806b07da360fdeb31aa21
Here's the error I'm getting. It might be an issue with temporary files, with the usage of `NamedTemporaryFile` in `RoundtipTest`. I'm planning on making a separate PR that will address the leftover temporary file issues, so I'll see if I can tackle that.

Strange that you can't reproduce it though. I do know that this tests are very flaky, failing and succeeding seemingly at random.

numpy/_core/tests/test_ufunc.py
bwhitt7: Will do! Seems to have done the trick. Should I include that fix in this PR or a separate one?

numpy/_core/tests/test_regression.py
bwhitt7: Hmmm, re-running this test, it seems I can't reproduce the error I got. It should be fine then and hopefully thread-safe.

numpy/_core/tests/test_custom_dtypes.py
bwhitt7: Made the issue here https://github.com/numpy/numpy/issues/29850

numpy/random/tests/test_generator_mt19937.py
bwhitt7: I'm thinking this is probably a memory issue with crashing, since this test is marked as slow. I'll need to see how much memory this test consumes.

numpy/f2py/tests/conftest.py
bwhitt7: Will do!

numpy/_core/tests/test_multiarray.py
bwhitt7: This is honestly a fairly big fix, especially since the fixture is parameterized (so if we remove the fixture, that means putting the parameterize values on every test that used tmp_filename), and our fixture patches don't feed into other fixtures. I'll see what I can do here to make the fixture thread safe however. Regardless it probably warrants it's own PR.

numpy/matrixlib/tests/test_matrix_linalg.py
bwhitt7: Here's the issue https://github.com/numpy/numpy/issues/29851

rgommers:
> an file. I can definitely change this and instead put

This seems fine. You do have to add that new file to `meson.build` so it actually gets installed, otherwise this will behave differently when running the tests in-tree vs. on an installed package.

bwhitt7:
>  You do have to add that new file to `meson.build`

@rgommers Sure! Do you have some resources I can look at to show I could do something like this? I see there's some meson.build files in the root folders of NumPy but I don't think any of them directly look at the test files in f2py.

lysnikolaou:
> You do have to add that new file to `meson.build` so it actually gets installed, otherwise this will behave differently when running the tests in-tree vs. on an installed package.

It looks like this is handled already by https://github.com/numpy/numpy/blob/ffe55df43f77e4fee273957fa936f455a70c2d2a/numpy/meson.build#L365-L370

Right?

numpy/lib/tests/test_io.py
lysnikolaou: Aah, it's because of `self.arr` and `self.arr_reloaded`. Maybe `RoundtripTest.roundtrip` could return them instead of setting instance variables? Yeah, weird that I couldn't reproduce this. On a closer look, it seems like an obvious race that should be triggering fairly often.

numpy/matrixlib/tests/test_matrix_linalg.py
lysnikolaou: Great, thanks! Maybe instead of marking it as `thread_unsafe` we should be xfailing it instead if `num_parallel_threads > 1`?

numpy/_core/tests/test_multiarray.py
bwhitt7: Tried fixing it in this PR https://github.com/numpy/numpy/pull/29858

numpy/_core/tests/test_multiarray.py
bwhitt7: Ah it was less thread-unsafe setup, and this sort of class-wide usage of this Foo class that the `self.f` variable sets up. The test is a bit confusing to me, but it seems to be trying to modify the class in some way? With interfaces? It feeds `self.f` into the test twice too hehe. I'll see if I can fix this, but it'll be thread-unsafe anyways.

numpy/_core/tests/test_ufunc.py
bwhitt7: Made a PR for this here https://github.com/numpy/numpy/pull/29867

numpy/lib/tests/test__datasource.py
bwhitt7: `datasource.Datasource` looks to be thread safe once the temporary file issue is fixed. Put these changes in this PR https://github.com/numpy/numpy/pull/29858

numpy/lib/tests/test_io.py
bwhitt7: Yep, that was it, thank you! Added those edits to https://github.com/numpy/numpy/pull/29867

rgommers:
Ah yes it is - good catch

numpy/matrixlib/tests/test_matrix_linalg.py
bwhitt7: Ah that could work. Though I'm a bit hesitant to do that since these tests don't always fail when ran in parallel (whoops, forgot to mention that in the issue I made). Sometimes they do, sometimes they don't. Is that still fine to mark with an xfail?

numpy/_core/tests/test_arrayprint.py
ngoldbaum: @bwhitt7 and I chatted about this, she's going to file a hypothesis issue and update this message with a link to it.

numpy/_core/tests/test_dtype.py
```diff
@@ -1316,6 +1316,7 @@ def test_object_flag_not_inherited(self):
 
     @pytest.mark.slow
     @hypothesis.given(dtype=hynp.nested_dtypes())
+    @pytest.mark.thread_unsafe(reason="hynp.nested_dtypes thread unsafe?")
```
ngoldbaum: needs a link to a hypothesis issue

numpy/_core/tests/test_dtype.py
```diff
@@ -1324,6 +1325,7 @@ def test_make_canonical_hypothesis(self, dtype):
         assert np.can_cast(two_arg_result, canonical, casting="no")
 
     @pytest.mark.slow
+    @pytest.mark.thread_unsafe(reason="gives unreliable results w/ hypothesis")
```
ngoldbaum: needs a link to a hypothesis issue

numpy/_core/tests/test_finfo.py
```diff
@@ -69,6 +69,7 @@ def float64_ma():
     'machep', 'negep', 'minexp', 'maxexp', 'nmant', 'iexp',
     'eps', 'epsneg', 'precision', 'resolution'
 ])
+@pytest.mark.thread_unsafe(reason="request.getfixturevalue thread-unsafe?")
```
ngoldbaum: Maybe phrase this as "Complex fixture setup is thread-unsafe" and link to https://github.com/pytest-dev/pytest/issues/13768.

numpy/_core/tests/test_multiarray.py
ngoldbaum: We'll merge that PR then update this PR to remove these markers.

numpy/_core/tests/test_multiarray.py
```diff
@@ -9013,6 +9018,9 @@ def __array_interface__(self):
         # This fails due to going into the buffer protocol path
         (f, {'data': None, 'shape': ()}, TypeError),
         ])
+    @pytest.mark.thread_unsafe(
+        reason="interface testing is thread-unsafe, test result depends on the reference count of a global object"
```
ngoldbaum:
```suggestion
        reason="test result depends on the reference count of a global object"
```

ngoldbaum: unless you were trying to say something with the first part - it was too unclear for me to understand, if so.

numpy/_core/tests/test_multiarray.py
```diff
@@ -9701,6 +9709,7 @@ def test_ctypes_is_available(self):
         assert_equal(ctypes, test_arr.ctypes._ctypes)
         assert_equal(tuple(test_arr.ctypes.shape), (2, 3))
 
+    @pytest.mark.thread_unsafe(reason="modifies global state")
```
ngoldbaum:
```suggestion
    @pytest.mark.thread_unsafe(reason="modifies global module state")
```

numpy/_core/tests/test_multithreading.py
```diff
@@ -12,6 +12,7 @@
 if IS_WASM:
     pytest.skip(allow_module_level=True, reason="no threading support in wasm")
 
+pytestmark = pytest.mark.thread_unsafe(reason="multithreading on parallel threads")
```
ngoldbaum:
```suggestion
pytestmark = pytest.mark.thread_unsafe(reason="Tests in this module are already explicitly multi-threaded")
```

numpy/_core/tests/test_nditer.py
```diff
@@ -3415,6 +3416,7 @@ def test_arbitrary_number_of_ops_error():
         np.nested_iters(args, [[0], []])
 
 
+@pytest.mark.thread_unsafe(reason="capfd is thread-unsafe?")
```
ngoldbaum:
```suggestion
@pytest.mark.thread_unsafe(reason="capfd is thread-unsafe")
```

No need to be uncertain - it's thread-unsafe. It's reading from stdout and stderr and there's only one of each stream globally.

numpy/_core/tests/test_scalarmath.py
```diff
@@ -871,6 +871,7 @@ def recursionlimit(n):
 @given(sampled_from(objecty_things),
        sampled_from(binary_operators_for_scalar_ints),
        sampled_from(types + [rational]))
+@pytest.mark.thread_unsafe(reason="gives unreliable results w/ hypothesis")
```
ngoldbaum: needs a hypothesis issue

numpy/_core/tests/test_scalarmath.py
```diff
@@ -882,6 +883,7 @@ def test_operator_object_left(o, op, type_):
...
```
ngoldbaum: needs a hypothesis issue

numpy/_core/tests/test_ufunc.py
```diff
@@ -3247,6 +3248,7 @@ def test_resolve_dtypes_reduction_errors(self):
 
     @pytest.mark.skipif(not hasattr(ct, "pythonapi"),
             reason="`ctypes.pythonapi` required for capsule unpacking.")
+    @pytest.mark.thread_unsafe(reason="ctype global state")
```
ngoldbaum:
```suggestion
    @pytest.mark.thread_unsafe(reason="modifies global object in the ctypes API")
```

numpy/lib/tests/test__iotools.py
```diff
@@ -200,6 +202,10 @@ def test_missing(self):
         except ValueError:
             pass
 
+    @pytest.mark.thread_unsafe(
+        reason=("numpy/lib/tests/test_io.py::TestFromTxt::"
+                "test_dtype_with_object_no_converter "
+                "fails due to thread-unsafe try/finally"))
```
ngoldbaum:
```suggestion
        reason=("monkeypatches StringConverter"))
```

numpy/lib/tests/test_function_base.py
```diff
@@ -3961,6 +3964,7 @@ def test_quantile_monotonic(self, method):
         quantile = np.quantile([0., 1., 2., 3.], p0, method=method)
         assert_equal(np.sort(quantile), quantile)
 
+    @pytest.mark.thread_unsafe(reason="gives unreliable results w/ hypothesis")
```
ngoldbaum: needs a hypothesis issue

numpy/lib/tests/test_loadtxt.py
ngoldbaum: Let's file a NumPy issue about this, with instructions on how to trigger the issue. Then you can reference the issue in the thread-unsafe reason.

numpy/lib/tests/test_loadtxt.py
```diff
@@ -74,13 +74,15 @@ def mixed_types_structured():
...
 def test_structured_dtype_and_skiprows_no_empty_lines(
         skiprows, mixed_types_structured):
     data, dtype, expected = mixed_types_structured
     a = np.loadtxt(data, dtype=dtype, delimiter=";", skiprows=skiprows)
     assert_array_equal(a, expected[skiprows:])
 
 
+@pytest.mark.thread_unsafe(reason="mixed_types_structured fixture is thread-unsafe?")
```
ngoldbaum: ditto here with a link to a numpy issue and deleting the question mark.

numpy/lib/tests/test_nanfunctions.py
```diff
@@ -723,6 +723,7 @@ def test_ddof(self):
                 res = nf(_ndat, axis=1, ddof=ddof)
                 assert_almost_equal(res, tgt)
 
+    @pytest.mark.thread_unsafe(reason="pytest.recwarn is thread-unsafe")
```
ngoldbaum: is it? Even on 3.14t or 3.14 with context-aware warnings turned on? I'd be a little surprised.

numpy/tests/test_ctypeslib.py
```diff
@@ -302,6 +305,7 @@ def test_scalar(self):
         ct = np.ctypeslib.as_ctypes_type(dt)
         assert_equal(ct, ctypes.c_uint16)
 
+    @pytest.mark.thread_unsafe(reason="some sort of data race?")
```
ngoldbaum: The uncertain reasons in this file need a NumPy issue(s) as well.

numpy/tests/test_numpy_config.py
```diff
@@ -21,6 +21,7 @@ class TestNumPyConfigs:
     ]
 
     @patch("numpy.__config__._check_pyyaml")
+    @pytest.mark.thread_unsafe(reason="unittest.mock.patch thread-unsafe?")
```
ngoldbaum:
```suggestion
    @pytest.mark.thread_unsafe(reason="unittest.mock.patch thread-unsafe")
```

No need to be uncertain, it's unsafe. This is monkeypatching.

numpy/typing/tests/test_isfile.py
```diff
@@ -25,6 +27,7 @@
     FILES += [ROOT / "distutils" / "__init__.pyi"]
 
 
+@pytest.mark.thread_unsafe(reason="os.path has a thread-safety bug")
```
ngoldbaum: This nerd-sniped me pretty good just now. I managed to make a reproducer that doesn't use NumPy at all, see https://github.com/python/cpython/issues/140054. You can link to that issue in the reason here.

numpy/tests/test_ctypeslib.py
bwhitt7: https://github.com/numpy/numpy/issues/29943 Made the issue here

numpy/lib/tests/test_loadtxt.py
bwhitt7: Realized this can be easily fixed hehe, made a PR here https://github.com/numpy/numpy/pull/29942

numpy/lib/tests/test_nanfunctions.py
bwhitt7: Yep still appears to be thread-unsafe. Here's a gist of the error it produces: https://gist.github.com/bwhitt7/eb576be5e47d027a8664871aff28d914

numpy/f2py/tests/conftest.py
```diff
@@ -0,0 +1,6 @@
+import pytest
+
+
+@pytest.hookimpl(tryfirst=True)
+def pytest_itemcollected(item):
+    item.add_marker(pytest.mark.thread_unsafe(reason="f2py is thread-unsafe"))
```
ngoldbaum: Let's soften this a little and just say that the tests aren't thread-safe.

numpy/lib/tests/test_nanfunctions.py
ngoldbaum: We're fixing this another way in https://github.com/numpy/numpy/pull/29957. I also reported https://github.com/Quansight-Labs/pytest-run-parallel/issues/135.

numpy/polynomial/tests/test_printing.py
```diff
@@ -259,6 +260,7 @@ def test_set_default_printoptions():
         poly.set_default_printstyle('invalid_input')
 
 
+@pytest.mark.thread_unsafe(reason="set_default_printstyle() is thread-unsafe")
```
ngoldbaum: Here and in the other one in this file: it's not that it's thread-unsafe but that the "default" printstyle is global state. Using the term "global state" is more consistent with the reasons you use elsewhere and also is less scary-sounding.

numpy/random/tests/test_extending.py
```diff
@@ -57,6 +57,7 @@
 @pytest.mark.skipif(sysconfig.get_platform() == 'win-arm64',
                     reason='Meson unable to find MSVC linker on win-arm64')
 @pytest.mark.slow
+@pytest.mark.thread_unsafe(reason="crashes sometimes during meson runs")
```
ngoldbaum: This is a little too vague for my taste. I'd say that "building cython code in a subprocess doesn't make sense to do in many threads and sometimes crashes"

numpy/testing/tests/test_utils.py
```diff
@@ -1875,6 +1879,7 @@ def test_suppress_warnings_type():
 @pytest.mark.filterwarnings(
     "ignore:.*NumPy warning suppression and assertion utilities are deprecated"
     ".*:DeprecationWarning")
+@pytest.mark.thread_unsafe(reason="uses deprecated thread-unsafe warnings")
```
ngoldbaum: "warnings control utilities" instead of just "warnings", here and in the other identical reasons in this file.

numpy/tests/test_numpy_config.py
```diff
@@ -21,6 +21,7 @@ class TestNumPyConfigs:
     ]
 
     @patch("numpy.__config__._check_pyyaml")
+    @pytest.mark.thread_unsafe(reason="unittest.mock.patch thread-unsafe")
```
ngoldbaum:
```suggestion
    @pytest.mark.thread_unsafe(reason="unittest.mock.patch updates global state")
```

numpy/typing/tests/test_isfile.py
```diff
@@ -25,6 +27,9 @@
     FILES += [ROOT / "distutils" / "__init__.pyi"]
 
 
+@pytest.mark.thread_unsafe(
+    reason="os.path has a thread-safety bug (python/cpython#140054)"
```
ngoldbaum: It looks like this will be fixed in 3.14.1, so you could also add a note that we expect this to only be a problem in 3.14.0.

ngoldbaum (approved):
Awesome. I have a bunch more nitpicks but this is mergeable as-is. Thanks so much for your diligence here. Being careful to leave behind good reasons in these messages will make it much easier for future contributors to understand what these markers mean. Hopefully also we'll fix the ones that are indicating bugs in NumPy!

bwhitt7:
Looks like there's some problem with the new `conftest.py` file messing with the full test job. Likely need to do an extra step to get the file built with NumPy.

numpy/conftest.py
```diff
@@ -225,3 +242,9 @@ def warnings_errors_and_rng(test=None):
         'numpy/random/_examples',
         'numpy/f2py/_backends/_distutils.py',
     ]
+
+def pytest_collection_modifyitems(config, items):
+    for item in items:
+        if Path(item.fspath).parent == Path(__file__).parent / 'f2py' / 'tests':
+            item.add_marker(pytest.mark.thread_unsafe(
+                reason="f2py tests are not thread-unsafe"))
```
ngoldbaum: oops, can you fix my bad grammar here?

bwhitt7: Whoops didn't see that! Will do

ngoldbaum:
In it goes, thanks @bwhitt7

PR #29819, opened 2025-09-27
ENH: Add fast path in ufuncs for numerical scalars.

mhvk:
It has always been a bit annoying that `math.sin` is so much faster than `np.sin` on a scalar.

This PR adds a fast check for numerical scalars, improving the speed by about a factor of 5--6, with essentially no cost for arrays.  At the moment, just for single-input, single-output ufuncs, with only the input passed in. The trial assumes that the output has the same type as the input (if that fails, the regular path is taken).

Sample speed test:
```
%timeit np.sin(1.5)
103 ns ± 3.32 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
# on main
688 ns ± 12.4 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
%timeit math.sin(1.5)
38.3 ns ± 4.77 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

Note that I did some trials to find out where most time (690 ns on my laptop when plugged in) is spent in the regular path (tested by simply returning the input argument at various places):
- 36 ns at start, OK
- 39 ns -> +3 ns for memory allocation
- 77ns -> +38 ns for turning args into full_args.in tuple
- 94 ns -> +14 ns for override check
- ! 344 ns -> +250 ns for `convert_ufunc_arguments`
- 369 ns -> +15 ns for `promote_and_get_ufuncimpl`
- 400 ns -> +31 ns for `resolve_descriptors`
- ! 621 ns -> +221 ns for `PyUFunc_GenericFunctionInternal`
- 651 ns -> +30 ns for wrap-check and return

So, about 40% of the time is spent in `convert_ufunc_arguments`, mostly in turning the scalars into arrays - `PyArray_FromAny` perhaps could be sped up a bit (time needed is similar to that of `np.array(scalar)`), though I think the time is dominated by actually creating an array structure in the first place. 

But there would seem to be some space for improvements at the 10% level also for regular arrays by, e.g., not creating the `full_args.in` tuple until it is actually needed (for overrides). There may also be little bits to gain in `PyUFunc_GenericFunctionInternal`.

### DIFF
--- a/doc/release/upcoming_changes/29819.improvement.rst
+++ b/doc/release/upcoming_changes/29819.improvement.rst
@@ -1,0 +1,6 @@
+Performance increase for scalar calculations
+--------------------------------------------
+The speed of calculations on scalars has been improved by about a factor 6 for
+ufuncs that take only one input (like ``np.sin(scalar)``), reducing the speed
+difference from their ``math`` equivalents from a factor 19 to 3 (the speed
+for arrays is left unchanged).
--- a/numpy/_core/src/umath/ufunc_object.c
+++ b/numpy/_core/src/umath/ufunc_object.c
@@ -66,6 +66,7 @@
 #include "npy_static_data.h"
 #include "multiarraymodule.h"
 #include "number.h"
+#include "scalartypes.h"  // for is_anyscalar_exact and scalar_value
 
 /********** PRINTF DEBUG TRACING **************/
 #define NPY_UF_DBG_TRACING 0
@@ -3721,7 +3722,7 @@ PyUFunc_GenericReduction(PyUFuncObject *ufunc,
     if (ret == NULL) {
         goto fail;
     }
-    
+
     Py_XDECREF(out);
 
     Py_DECREF(signature[0]);
@@ -4270,6 +4271,135 @@ replace_with_wrapped_result_and_return(PyUFuncObject *ufunc,
     return NULL;
 }
 
+/*
+ * Check whether the input object is a known scalar and whether the ufunc has
+ * a suitable inner loop for it, which takes and returns the data type of the
+ * input (this function is not called if output or any other argument was given).
+ * If a loop was found, call it and store the result.
+ *
+ * Returns -2 if a short-cut is not possible, 0 on success and -1 on error.
+ */
+static int
+try_trivial_scalar_call(
+    PyUFuncObject *ufunc, PyObject *const obj, PyObject **result)
+{
+    assert(ufunc->nin == 1 && ufunc->nout == 1 && !ufunc->core_enabled);
+    npy_clongdouble cin, cout;  // aligned storage, using longest type.
+    char *in = (char *)&cin, *out = (char *)&cout;
+    char *data[] = {in, out};
+    int ret = -2;
+    PyArray_Descr *dt;
+    /*
+     * For supported input, get input pointer and descriptor. Otherwise, bail.
+     */
+    if (obj == Py_False || obj == Py_True) {
+        *(npy_bool *)in = (obj == Py_True);
+        dt = PyArray_DescrFromType(NPY_BOOL);
+    }
+    else if (PyFloat_CheckExact(obj)) {
+        *(double *)in = PyFloat_AS_DOUBLE(obj);
+        dt = PyArray_DescrFromType(NPY_FLOAT64);
+    }
+    else if (PyLong_CheckExact(obj)) {
+        int overflow;
+        npy_intp val = PyLong_AsLongAndOverflow(obj, &overflow);
+        if (overflow) {
+            return -2;  // bail, main code perhaps deals with this.
+        }
+        if (error_converting(val)) {
+            return -1;  // should never happen; pass on it if does.
+        }
+        *(npy_intp *)in = val;
+        dt = PyArray_DescrFromType(NPY_INTP);
+    }
+    else if (PyComplex_CheckExact(obj)) {
+        Py_complex oop = PyComplex_AsCComplex(obj);
+        if (error_converting(oop.real)) {
+            return -1;  // should never happen; pass on it if does.
+        }
+        *(double *)in = oop.real;
+        *(double *)(in+sizeof(double)) = oop.imag;
+        dt = PyArray_DescrFromType(NPY_COMPLEX128);
+    }
+    else if (is_anyscalar_exact(obj)) {
+        dt = PyArray_DescrFromScalar(obj);
+        if (!PyDataType_ISNUMBER(dt)) {
+            goto bail;
+        }
+        data[0] = scalar_value(obj, dt);
+    }
+    else {
+        return -2;
+    }
+    /*
+     * Check the ufunc supports our descriptor, bailing (return -2) if not.
+     */
+    // Try getting info from the (private) cache.  Fall back if not found,
+    // so that the the dtype gets registered and things will work next time.
+    PyArray_DTypeMeta *op_dtypes[2] = {NPY_DTYPE(dt), NULL};
+    PyObject *info = PyArrayIdentityHash_GetItem(  // borrowed reference.
+        (PyArrayIdentityHash *)ufunc->_dispatch_cache,
+        (PyObject **)op_dtypes);
+    if (info == NULL) {
+        goto bail;
+    }
+    // Check actual dtype is correct (can be wrong with promotion).
+    PyObject *all_dtypes = PyTuple_GET_ITEM(info, 0);
+    if ((PyTuple_GET_ITEM(all_dtypes, 0) != (PyObject *)NPY_DTYPE(dt)) ||
+        (PyTuple_GET_ITEM(all_dtypes, 1) != (PyObject *)NPY_DTYPE(dt))) {
+        goto bail;
+    }
+    // Get method, bailing if not an arraymethod (e.g., a promotor).
+    PyArrayMethodObject *method = (PyArrayMethodObject *)PyTuple_GET_ITEM(info, 1);
+    if (!PyObject_TypeCheck(method, &PyArrayMethod_Type)) {
+        goto bail;
+    }
+    // Get loop, requiring that the output and input dtype are the same.
+    PyArrayMethod_Context context;
+    PyArray_Descr *descrs[2] = {dt, dt};
+    NPY_context_init(&context, descrs);
+    context.caller = (PyObject *)ufunc;
+    context.method = method;
+    npy_intp strides[2] = {0, 0};  // 0 ensures scalar math, not SIMD for half.
+    PyArrayMethod_StridedLoop *strided_loop;
+    NpyAuxData *auxdata = NULL;
+    NPY_ARRAYMETHOD_FLAGS flags = 0;
+    if (method->get_strided_loop(&context, 1, 0, strides,
+                                 &strided_loop, &auxdata, &flags) < 0) {
+        ret = -1;  // Should not happen, so raise error if it does anyway.
+        goto bail;
+    }
+    /*
+     * Call loop with single element, checking floating point errors.
+     */
+    if (!(flags & NPY_METH_NO_FLOATINGPOINT_ERRORS)) {
+        npy_clear_floatstatus();
+    }
+    npy_intp n = 1;
+    ret = strided_loop(&context, data, &n, strides, auxdata);
+    NPY_AUXDATA_FREE(auxdata);
+    if (ret == 0) {
+        if (!(flags & NPY_METH_NO_FLOATINGPOINT_ERRORS)) {
+            // Check for any unmasked floating point errors (note: faster
+            // than _check_ufunc_fperr as one doesn't need mask up front).
+            int fpe_errors = npy_get_floatstatus();
+            if (fpe_errors) {
+                if (PyUFunc_GiveFloatingpointErrors(
+                        ufunc_get_name_cstr(ufunc), fpe_errors) < 0) {
+                    ret = -1;  // Real error, falling back would not help.
+                    goto bail;
+                }
+            }
+        }
+        *result = PyArray_Scalar(out, dt, NULL);
+        if (*result == NULL) {
+            ret = -1;  // Real error (should never happen).
+        }
+    }
+  bail:
+    Py_DECREF(dt);
+    return ret;
+}
 
 /*
  * Main ufunc call implementation.
@@ -4288,6 +4418,14 @@ ufunc_generic_fastcall(PyUFuncObject *ufunc,
     int errval;
     int nin = ufunc->nin, nout = ufunc->nout, nop = ufunc->nargs;
 
+    if (len_args == 1 && kwnames == NULL && !PyArray_Check(args[0])
+            && nin == 1 && nout == 1 && !ufunc->core_enabled) {
+        // Possibly scalar input, try the fast path, falling back on failure.
+        PyObject *result = NULL;
+        if (try_trivial_scalar_call(ufunc, args[0], &result) != -2) {
+            return result;
+        }
+    }
     /* All following variables are cleared in the `fail` error path */
     ufunc_full_args full_args = {NULL, NULL};
     PyArrayObject *wheremask = NULL;
@@ -4301,7 +4439,7 @@ ufunc_generic_fastcall(PyUFuncObject *ufunc,
         return NULL;
     }
     memset(scratch_objs, 0, sizeof(void *) * (nop * 4 + 2));
-    
+
     PyArray_DTypeMeta **signature = (PyArray_DTypeMeta **)scratch_objs;
     PyArrayObject **operands = (PyArrayObject **)(signature + nop);
     PyArray_DTypeMeta **operand_DTypes = (PyArray_DTypeMeta **)(operands + nop + 1);
@@ -4366,17 +4504,17 @@ ufunc_generic_fastcall(PyUFuncObject *ufunc,
         /* Extra positional args but no keywords */
         /* DEPRECATED NumPy 2.4, 2025-08 */
         if ((PyObject *)ufunc == n_ops.maximum || (PyObject *)ufunc == n_ops.minimum) {
-            
+
             if (DEPRECATE(
                 "Passing more than 2 positional arguments to np.maximum and np.minimum "
-                "is deprecated. If you meant to use the third argument as an output, " 
+                "is deprecated. If you meant to use the third argument as an output, "
                 "use the `out` keyword argument instead. If you hoped to work with "
                 "more than 2 inputs, combine them into a single array and get the extrema "
                 "for the relevant axis.") < 0) {
                 return NULL;
             }
         }
-        
+
         if (all_none) {
             Py_SETREF(full_args.out, NULL);
         }


charris:
The error is unrelated.

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4270,6 +4271,116 @@ replace_with_wrapped_result_and_return(PyUFuncObject *ufunc,
...
+ * If a loop was found, call it and store the result.
+ *
+ * Returns -2 if a short-cut is not possible, 0 on success and -1 on error.
+ */
+static int
+try_trivial_scalar_call(
+    PyUFuncObject *ufunc, PyObject *const obj, PyObject **result)
+{
+    assert(ufunc->nin == 1 && ufunc->nout == 1 && !ufunc->core_enabled);
+    char in[sizeof(npy_clongdouble)], out[sizeof(npy_clongdouble)];  // storage
```
seberg: Need to type this as `maxalign_t` or so to guarantee it's aligned.

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4270,6 +4271,116 @@ replace_with_wrapped_result_and_return(PyUFuncObject *ufunc,
...
+        (PyArrayIdentityHash *)ufunc->_dispatch_cache,
+        (PyObject **)op_dtypes);
+    if (info == NULL) {
+        goto fail;
+    }
+    // Try getting method, falling back if not an arraymethod (e.g., a promotor)
+    PyArrayMethodObject *method = (PyArrayMethodObject *)PyTuple_GET_ITEM(info, 1);
+    if (!PyObject_TypeCheck(method, &PyArrayMethod_Type)) {
+        goto fail;
+    }
```
seberg: We should have a helper for this, no?  Also `info` being `NULL` _might_ be OK, I think (i.e. integer input for a ufunc that only supports floats needs promotion and may not be already cached).
Did you integrate this fast path here because it was a speed difference?  (I am sure it is, but I wonder if it would be worth 10%.

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4270,6 +4271,116 @@ replace_with_wrapped_result_and_return(PyUFuncObject *ufunc,
...
+    // Try getting loop, for speed assuming output and input dtype are the same.
```
seberg: At least, I think we need to check that the DType types all match up.  But is it so bad to call `resolve_descriptors` and bail here?

EDIT: Maybe you can fast-path this if the default `resolve_descriptors` is used, in which case we can just use the default descriptor for the result (or check that the type matches for it).

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4270,6 +4271,116 @@ replace_with_wrapped_result_and_return(PyUFuncObject *ufunc,
...
+    // Call loop with single element.
+    if (!(flags & NPY_METH_NO_FLOATINGPOINT_ERRORS)) {
+        npy_clear_floatstatus();
+    }
+    npy_intp n = 1;
+    ret = strided_loop(&context, data, &n, strides, auxdata);
+    NPY_AUXDATA_FREE(auxdata);
+    if (ret == 0) {
+        if (!(flags & NPY_METH_NO_FLOATINGPOINT_ERRORS)
+                && npy_get_floatstatus() != 0) {
```
seberg: Why not just call `_check_ufunc_fperr`? I can't really see that doing this here makes a meaningful difference speed wise? (Maybe this comes from the scalar path?)
If we really think it does, the question is if we can't tell the compiler to inline the function.

seberg (commented):
Yeah, also one of those long standing things...  I suppose this is rather simple, how much does it to the simplest array paths?

This is of course (and has to be) specific to NumPy scalars and 1 in/1 out ufuncs.

While we need to do a few more defensive guesses here I think, which will slow it down slightly, it is simple enough that it seems worthwhile to just do this.

I am really curious how far we can get without a fully special path here, e.g. by trying to make creation of (simple) 0-D arrays very fast (without that, one would have to delay array creation, since channeling non-arrays into the iterator wouldn't be fun, unfortunately.)

numpy/_core/src/umath/ufunc_object.c
mhvk: Yes, I integrated things since for pure float input it all adds up quickly... And the idea is to fall back as fast as possible for anything that needs promotion.

numpy/_core/src/umath/ufunc_object.c
mhvk: I cannot use `resolve_descriptors` because that takes `PyArrayObject *operands[]` and the thing I really need to avoid is turning the scalar into an array. The logic here was that if a loop with the same dtype for in and out exists, it is pretty much guaranteed to be the right one. And if it does not exist, then we just fall back - it is unlikely to be a frequently used single-input ufunc!

numpy/_core/src/umath/ufunc_object.c
mhvk: Yes, it did make a meaningful difference for the pure float case. The problem is that `_check_ufunc_pferr` needs the error mask, and it takes time to get that. Logically, one should not bother extracting the error mask if no error has occurred, but changing that is a bit beyond this PR.

numpy/_core/src/umath/ufunc_object.c
seberg: That's fine, but `info` being `NULL` is completely OK, you should return -2 (if that's what it was).

numpy/_core/src/umath/ufunc_object.c
seberg: > because that takes `PyArrayObject *operands[]`

No, it takes descriptors.

EDIT: But, forgot to add:  The point is even if you gamble on it being all right, the loop could be `Int8DType->Float64DType` in principle.  We do need to guard at least against that.

numpy/_core/src/umath/ufunc_object.c
seberg: Ah, right.  I had mis-looked.  Yeah, ufuncs always get that because the buffer size is stored there as well, but that would add 10-20ns...
Could maybe use `PyUFunc_GiveFloatingpointErrors` to simplify it slightly, though.

numpy/_core/src/umath/ufunc_object.c
mhvk: Not quite sure how to use `maxalign`... Would something like this work?
```
    npy_clongdouble cin, cout;  // aligned storage, using longest type.
    char *in = (char *)&cin;
    char *data[] = {in, (char *)&out};
```

mhvk:
> Yeah, also one of those long standing things... I suppose this is rather simple, how much does it to the simplest array paths?

As far as I could tell, negligible impact on arrays, because of the `!PyArray_Check(args[0])` in front of the call (note that that one should have `len_args == 1` as the very first check - I've got that committed, but will wait pushing until I've been able to answer your comments.
 
> This is of course (and has to be) specific to NumPy scalars and 1 in/1 out ufuncs.

And to python floats!  For 2in/1out it is much less problematic since most of those are called indirectly, and operators can avoid the trouble.
 
> While we need to do a few more defensive guesses here I think, which will slow it down slightly, it is simple enough that it seems worthwhile to just do this.

I tried hard to think what could go wrong with what I did, bailing in the face of uncertainty. Let me know if I missed something!

> I am really curious how far we can get without a fully special path here, e.g. by trying to make creation of (simple) 0-D arrays very fast (without that, one would have to delay array creation, since channeling non-arrays into the iterator wouldn't be fun, unfortunately.)

Agreed, we should make array creation faster. But just doing `np.empty(())` is already slow. It seems hard to believe one could reduce it ~10 fold (which would be needed to become competitive).

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4288,6 +4399,14 @@ ufunc_generic_fastcall(PyUFuncObject *ufunc,
     int errval;
     int nin = ufunc->nin, nout = ufunc->nout, nop = ufunc->nargs;
 
+    if (!PyArray_Check(args[0]) && nin == 1 && nout == 1
```
mhvk: This should be `len_args == 1 && kwnames == NULL && !PyArray_Check(args[0])` (~with no `kwnames`,~ wrong, those do matter). Will push together with other changes later.

numpy/_core/src/umath/ufunc_object.c
seberg: Yeah, I think that is used elsewhere also.  `max_align_t` would be the type otherwise, may need an include, though.  But I don't see it used yet, so I suspect we always assume longdouble works.

numpy/_core/src/umath/ufunc_object.c
mhvk: But that's what I do on 4333 - maybe the `goto fail` is misleading here - it just means to decref the descriptor and return `ret` which is `-2` (bail) by default.  I can rename `fail` to `bail`...

numpy/_core/src/umath/ufunc_object.c
mhvk: But if the real loop is `int->float`, presumably `int->int` simply won't exist, and then we bail and do the right thing the slow way (again, you may have been misled by my `goto fail`...).

numpy/_core/src/umath/ufunc_object.c
mhvk: Yes, let me use that - once the status indicates something has happened, we don't have to worry about every ns any more.

numpy/_core/src/umath/ufunc_object.c
seberg: Ah, yeah... I didn't realize that part with many -2 returns before.  Maybe `goto fail` is misleading indeed.  (FWIW, I just saw that you have at least one `PyErr_Clear()`, I really would like to avoid it, it's always a weird thing to do IMO.)

seberg:
> It seems hard to believe one could reduce it ~10 fold

Yeah, not quite that much probably.  One could have a free-list, which might be very fast.  But we may have to check for a custom allocator being set, and that alone might actually be significant in this context (compared to what you have now).

> I tried hard to think what could go wrong with what I did, bailing in the face of uncertainty. Let me know if I missed something!

Just the two promotion/resolving steps:
* I think you didn't check that the loop DTypes match.  That is both input and output DType are what you expect it to be.  That is just an identity check maybe, but it is crucial.
* Skipping `resolve_dtypes` should be OKish, but if it is just ~5% slowdown, I might prefer not to (maybe a check for the default one can recover parts also).  You could imagine a ufunc that does something more, but it's hard to imagine for builtin numeric ones (it would have to be a `byteswap()` ufunc, but that is a useless one).

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4270,6 +4271,116 @@ replace_with_wrapped_result_and_return(PyUFuncObject *ufunc,
...
+        *(npy_bool *)in = (obj == Py_True);
+        dt = PyArray_DescrFromType(NPY_BOOL);
+    }
+    else if (PyFloat_CheckExact(obj)) {
+        *(double *)in = PyFloat_AS_DOUBLE(obj);
+        dt = PyArray_DescrFromType(NPY_FLOAT64);
+    }
+    else if (PyLong_CheckExact(obj)) {
+        if (PyLong_AsInt64(obj, (int64_t *)in) < 0) {
+            PyErr_Clear();
```
mhvk: @seberg - just as an example: here I clear the error and bail, so that the main routine can have a go. This can lead to different errors. E.g.,
```
np.log(2**100)
# TypeError: loop of ufunc does not support argument 0 of type int which has no callable sin method
```
while if we let this stay in place one would get,
```
OverflowError: Python int too large to convert to C long
```
And, really, I'm doing the wrong thing here: if somewhere down the line we decide that such long integers should just be converted to double.

Anyway, my logic with the `PyErr_Clear()` was to ensure we don't change behaviour.

seberg: There is a `WithOverflow` version I think.  That way you can detect the overflow without having to deal with an actual error being set.  The scalar-math code should use that (I would think).
(Pretty much any use of `PyErr_Clear()` is fishy in the end, this is maybe less fishy, but also a place where a nice alternative exists).

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4270,6 +4271,116 @@ replace_with_wrapped_result_and_return(PyUFuncObject *ufunc,
...
+    // so that the the dtype gets registered and things will work next time.
+    PyArray_DTypeMeta *op_dtypes[2] = {NPY_DTYPE(dt), NULL};
+    PyObject *info = PyArrayIdentityHash_GetItem(
+        (PyArrayIdentityHash *)ufunc->_dispatch_cache,
+        (PyObject **)op_dtypes);
+    if (info == NULL) {
+        goto fail;
+    }
+    // Try getting method, falling back if not an arraymethod (e.g., a promotor)
+    PyArrayMethodObject *method = (PyArrayMethodObject *)PyTuple_GET_ITEM(info, 1);
```
mhvk: > I think you didn't check that the loop DTypes match. That is both input and output DType are what you expect it to be. That is just an identity check maybe, but it is crucial.

Do you mean checking the dtypes on the method here? Can `get_strided_loop` succeed if there is a mismatch?

numpy/_core/src/umath/ufunc_object.c
mhvk: Unfortunately, the alternative is not *that* nice: from https://docs.python.org/3/c-api/long.html, I see that `PyLong_AsLongAndOverflow` can still return `overflow=0` and `res=-1` with error set, so I'd still have to check that (possibly, other errors can only occur with non-exact long, but not sure I can count on that...). Plus now I lost the nice numpy `Int64` macro, and have to deal with `long` conversion.

seberg: Scalar math uses this:
```
        int overflow;
        long val = PyLong_AsLongAndOverflow(value, &overflow);
        if (overflow) {
            /* handle as if "unsafe" */
            return -2;
        }
        if (error_converting(val)) {
            return -1;  /* should not be possible */
        }
```
that seems fine?!  (Yes loses very large ints on windows, but OK...)
Not sure what else you can do, there are ideally two results here, normal errors and criticial errors, so it doesn't get much shorter?  (Not that I think a real, criticial error _should_ happen, but...)

numpy/_core/src/umath/ufunc_object.c
seberg: Yes, the first entry are the dtypes as returned by promotion.  Assuming `signature=` is not used, those are the _actual_ DTypes that belong to the `ArrayMethod` registration and we *promise* that those are the DTypes passed in to `resolve_descriptors` the return of which we promise to pass to `get_strided_loop`.

Anyway, the point is that there is no guarantee that no promotion happened, that is: `PyTuple_GET_ITEM(PyTuple_GET_ITEM(info, 0), 0) != NPY_DTYPE(descr)`.
If that is the case, you might find a loop that belongs to a totally different dtype, and `get_strided_loop` is perfectly allowed to ignore that you passed it a wrong one.
This probably all kinda works out for the legacy loops, because they go back into a long search, but...

(EDIT: I am not sure that this split in dtype-tuple and ArrayMethod is all that useful in the end.  In part it came from a desire to allow cleanup when someone deleted a `DType`.  But Python is lacking the mechanisms to actually implement all but trivial cleanups, so I think we have to just accept `DTypes` are immortal, which seems OK to me.
Maybe we *should* move it onto the `ArrayMethod` itself -- we could, it's pretty much opaque.  Right now we probably use the same one for all legacy dtypes, but that can be done via abstract dtypes and duplicating isn't _that_ bad maybe)

numpy/_core/src/umath/ufunc_object.c
mhvk: > (Not that I think a real, criticial error _should_ happen, but...)

I can do that, but my worry is that the real ufunc might just use an `object` loop and not fail. 

But then, now it is like `complex`, really this error should not happen, and maybe if it does, it is good it gets noticed. I'll go with your suggestion.

mhvk:
OK, I pushed the changes so far. 

I can try how much extra time `resolve_descriptors` takes. I think I was confused about it: I thought you meant the function inside `ufunc_object` (which does take `operands[]` that I don't have), but now realize you probably meant `method->resolve_descriptors`. Anyway, got to teach soon, so better focus on that now. Will look at it later.

numpy/_core/src/umath/ufunc_object.c
mhvk: @seberg - returning to this speed-up for ufuncs on scalars: I think I have addressed all your earlier comments, would you be willing to have another look? (I only left this one open in case what I did is not what you meant...)

doc/release/upcoming_changes/29819.improvement.rst
```diff
@@ -0,0 +1,4 @@
+The speed of calculations on scalars has been improved by about a factor 6 for
```
seberg, line 3: Either add a bullet point or a heading :).  (Maybe the setup should be smart enough for this, but it isn't)

numpy/_core/src/umath/ufunc_object.c
seberg: This is what I mean.  I'll go through it now, may have to let this sink in a bit.
Too bad with all the efforts of speeding up the allocation, but in the end this seems worthwhile.

In practice, I suspect you could just use the output DType here as in `if DType->singleton and PyDataType_NUMBER(DType->singleton)` assume all else is good.

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4270,6 +4271,137 @@ replace_with_wrapped_result_and_return(PyUFuncObject *ufunc,
     return NULL;
 }
 
+/*
+ * Check whether the input object is a known scalar and whether the ufunc has
+ * a suitable inner loop for it, which takes and returns the data type of the
+ * input (this function is not called if output or any other argument was given).
+ * If a loop was found, call it and store the result.
+ *
+ * Returns -2 if a short-cut is not possible, 0 on success and -1 on error.
```
seberg, line 4280: N.B. with a clean slate I would use 1 for success and 0 for "not applicable" (matches `PyObject_GetOptionalAttr ` etc.).
But there is no point unless we fix the other `try_` functions here.

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4270,6 +4271,137 @@ replace_with_wrapped_result_and_return(PyUFuncObject *ufunc,
...
+    if (info == NULL) {
+        goto bail;
+    }
+    // Check actual dtype is correct (can be wrong with promotion).
+    PyObject *all_dtypes = PyTuple_GET_ITEM(info, 0);
+    if ((PyTuple_GET_ITEM(all_dtypes, 0) != (PyObject *)NPY_DTYPE(dt)) ||
+        (PyTuple_GET_ITEM(all_dtypes, 1) != (PyObject *)NPY_DTYPE(dt))) {
+        goto bail;
+    }
+    // Get method, balling if not an arraymethod (e.g., a promotor).
```
```suggestion
    // Get method, bailing if not an arraymethod (e.g., a promotor).
```

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4270,6 +4271,137 @@ replace_with_wrapped_result_and_return(PyUFuncObject *ufunc,
...
+    context.caller = (PyObject *)ufunc;
+    context.method = method;
+    npy_intp strides[2] = {0, 0};  // 0 ensures scalar math, not SIMD for half.
+    PyArrayMethod_StridedLoop *strided_loop;
+    NpyAuxData *auxdata = NULL;
+    NPY_ARRAYMETHOD_FLAGS flags = 0;
+    if (method->get_strided_loop(&context, 1, 0, strides,
+                                 &strided_loop, &auxdata, &flags) < 0) {
+        // Clear error, may just be because of wrong assumption.
+        PyErr_Clear();
```
```suggestion
```
Got to remove this. We really can't rely on this ever erroring.  The reason this works is solely for the old-style ufuncs which have to do a second dance of loop lookup (which is a waste but can fail gracefully).

But...  I assume that the DType check above will make this check unnecessary. (We are still gambling on `resolve_descriptors`, but that is a pretty small gamble for these dtypes).

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4270,6 +4271,137 @@ replace_with_wrapped_result_and_return(PyUFuncObject *ufunc,
...
+    if (!PyObject_TypeCheck(method, &PyArrayMethod_Type)) {
+        goto bail;
+    }
+    // Get loop, requiring that the output and input dtype are the same.
+    PyArrayMethod_Context context;
+    PyArray_Descr *descrs[2] = {dt, dt};
+    NPY_context_init(&context, descrs);
+    context.caller = (PyObject *)ufunc;
+    context.method = method;
+    npy_intp strides[2] = {0, 0};  // 0 ensures scalar math, not SIMD for half.
```
seberg, line 4363: Do the strides really matter here? Slightly worrying :).  (FWIW, I use either `elsize` or `0` for these often.  Mainly because we might actually have contiguous special loops in theory, while scalar ones are rare.)

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4270,6 +4271,137 @@ replace_with_wrapped_result_and_return(PyUFuncObject *ufunc,
...
+    else {
+        return -2;
+    }
+    /*
+     * Check the ufunc supports our descriptor, bailing (return -2) if not.
+     */
+    // Try getting info from the (private) cache.  Fall back if not found,
+    // so that the the dtype gets registered and things will work next time.
+    PyArray_DTypeMeta *op_dtypes[2] = {NPY_DTYPE(dt), NULL};
+    PyObject *info = PyArrayIdentityHash_GetItem(  // borrowed reference.
```
seberg, line 4340: I suspect we may need locking here for free-threading.  (Maybe should change this to not use borrowed references to make it more robust about that.)

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4270,6 +4271,137 @@ replace_with_wrapped_result_and_return(PyUFuncObject *ufunc,
...
+                }
+            }
+        }
+        *result = PyArray_Scalar(out, dt, NULL);
+        if (*result == NULL) {
+            ret = -1;  // Real error (should never happen).
+        }
+    }
+  bail:
+  fail:
```
seberg: If have both fail and bail, they should actually include the `ret = -2` and `ret = -1` :).
(Just one label is enough, maybe finish?)

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4270,6 +4271,137 @@ replace_with_wrapped_result_and_return(PyUFuncObject *ufunc,
...
+        int overflow;
+        long val = PyLong_AsLongAndOverflow(obj, &overflow);
+        if (overflow) {
+            return -2;  // bail, main code perhaps deals with this.
+        }
+        if (error_converting(val)) {
+            return -1;  // should never happen; pass on it if does.
+        }
+        *(npy_int64 *)in = (npy_int64)val;
+        dt = PyArray_DescrFromType(NPY_INT64);
```
seberg: This needs to be `INTP` (I think we also have DEFAULT_INT maybe if you prefer). Fortunately, we can assume `INTP >= LONG` though, I think.

(Interesting that the tests didn't notice...)

seberg (commented):
Thanks, a few smaller comments.  I suppose we should rarely add overhead (mainly in functions that always promote currently).
But if we promote and go the slow path (i.e. scalar in), the added overhead will be overshadowed by a lot.

Still a bit curious about just using `resolve_descriptors` (maybe with a check whether it is the default one).  But now with the `DType` check I think this is OK.
(I am also not 100% sure that there might not be some problem there.)

So overall, I suspect we should actually just do it, it's simple enough compared to the speedup.

This might actually propel us very fairly close to the scalarmath binary operators speed wise, but I guess those are unfortunately a bit more special...

doc/release/upcoming_changes/29819.improvement.rst
mhvk: Done.

numpy/_core/src/umath/ufunc_object.c
mhvk: Agreed, that also matches what `PyArray_ResolveWriteBackIfCopy` does and feels "good".

Can try to do that for all at some point do that, but indeed best not here.

numpy/_core/src/umath/ufunc_object.c
mhvk: Thanks for catching that! I guess the conversion to `npy_int64` made it safe, but indeed the return type would be incorrect. Note that there are not that many one-input, one-output integer ufuncs, so it may just never get tested with python ints...

Anyway, done! (went with `npy_long` and `NPY_LONG` to make clear it is the same thing).

numpy/_core/src/umath/ufunc_object.c
mhvk: Are you sure? I'm only reading it. I think this would only be a problem if there is no locking on the writing side (and atomic swap for insertion).

numpy/_core/src/umath/ufunc_object.c
mhvk: Done.

numpy/_core/src/umath/ufunc_object.c
mhvk: OK, let me just return -1 and change the comment to say that this should not happen.
...
At least locally, all the tests pass.

numpy/_core/src/umath/ufunc_object.c
mhvk: I removed `fail` - any time that is used, we also set `ret = -1`, so `goto bail` is clear enough there.

numpy/_core/src/umath/ufunc_object.c
mhvk: I had `{elsize, elsize}` before, and that failed for `half`, which entered some `SIMD` loop. But I cannot reproduce it right now, perhaps because I'm in my office on a computer that's too old to hit the relevant loop. I'll try on my laptop at home to see if it actually errored.

Regardless, the loop selection is via stride, so by passing in 0, one hits the unary loop, which is faster: https://github.com/numpy/numpy/blob/d28ba9b8e3ffb12e74ef8931ce8f7da480596df8/numpy/_core/src/umath/loops_half.dispatch.c.src#L71-L95
So, in that sense the comment is correct...

mhvk (commented):
Thanks for the detailed review. All comments addressed, I think!

mhvk:
> This might actually propel us very fairly close to the scalarmath binary operators speed wise, but I guess those are unfortunately a bit more special...

I did think about the binary operators, but, as you say, they already have a special path, and typically one will do `scalar1 + scalar2`, so it is less important that `np.add(scalar1, scalar2)` is very fast too. They will hopefully get the benefit from storing on the instance, though.

numpy/_core/src/umath/ufunc_object.c
mhvk: > In practice, I suspect you could just use the output DType here as in `if DType->singleton and PyDataType_NUMBER(DType->singleton)` assume all else is good.

Didn't see this reply earlier: I'm a bit confused by this; should I still add something like it before or after getting the method?

mhvk:
> Too bad with all the efforts of speeding up the allocation, but in the end this seems worthwhile.

The allocation stuff will still help a lot with things like `np.mul(small_array, scalar)`, etc.

numpy/_core/src/umath/ufunc_object.c
seberg: Ah, i think that is right.  We basically assume we are not replacing entries, which should be safe enough.  I think the normal ufuncs actually try a lookup also there.

numpy/_core/src/umath/ufunc_object.c
seberg: Hmm, yeah, I guess ufuncs would use the 0 path. I don't really like the idea of relying on 0 strides to get a different precision, which makes me hope that the comment ages badly, but maybe it is actually very realistic.

numpy/_core/src/umath/ufunc_object.c
mhvk: > Ah, i think that is right. We basically assume we are not replacing entries, which should be safe enough. I think the normal ufuncs actually try a lookup also there.

Yes, I suspect the reason I have this is that I just traced the normal path and copied things as I needed them...

numpy/_core/src/umath/ufunc_object.c
seberg: Small thing, but `NPY_LONG` _isn't_ the default dtype, it would be wrong on windows.  Using `FromLong` is fine as it is just smaller (to my knowledge) but the actual value will be an `npy_intp` and the dtype `NPY_INTP`.

mhvk: OK, sorry, I thought it made sense given the `FromLong`, but I guess sadly it doesn't. Anyway, changed to `npy_intp` in all cases.

mhvk:
circle ci/build failure is due to a timeout error

numpy/_core/src/umath/ufunc_object.c
seberg: Yeah, I guess you took this from scalars and that code predated the `long -> intp` move in NumPy 2.
We should probably `#define PyLong_AsSize_tAndOverflow` depending on the size of `Py_ssize_t/intp` (same size) and then use that in both places.
But again, not here!

seberg:
Thanks, I think I'll try to add a check for `resolve_descriptors` being the default one (or maybe even call it) -- the weird resolve with scalars doesn't matter for unary I think.

I would take a <10% performance hit there to be on the safe side of things and that might also unlock supporting a few more cases (but I am not suggesting to necessarily do that now).

mhvk:
@seberg - your comment just above suggested you still wanted to make a small change. What would that be? Or maybe OK to just go with what I have here?

seberg:
Sorry, forgot about this.  I pushed a commit that adds it, it was a bit more unwieldy than I hoped.  I think the "scalar" special path doesn't matter for unary ufuncs.  (I did rebase main so unfortunately, it's just a pull.)

For me, it pushes things from 77ns to 82ns for the `np.sin(1.5)` example.  It does make `np.signbit` fast, not that I think it particularly matters.
But, overall, I would prefer to add this since it is <10% overhead and does fix potential subtle issues.

mhvk (commented):
@seberg - Thanks!  But I'm still confused why we need `resolve_descriptors` when the goal is not to be general but just to make a common case much faster (i.e., it is fine to bail on `signbit`!)

I looked back at [NEP 43](https://numpy.org/neps/nep-0043-extensible-ufuncs.html) to try to understand why we would need `resolve_descriptors` and it states (omitting strings):
> - In general, whenever an output DType is parametric the parameters have to be found (resolved).
>  - When they are not parametric, a default implementation is provided which fills in the default dtype instances (ensuring for example native byte order).

Neither would seem relevant here - we only accept standard python and numpy scalars, so are we not guaranteed to deal with a non-parametric dtype with native byte order?

I do see that my original check that the output had to have the same dtype as the input would fail for `signbit`, but I'm still confused how it could fail for, say, `np.sin`, i.e., how can `resolve_descriptors` be relevant?

mhvk:
p.s. The core dumps seem related...

seberg:
Ack, yeah, the crash is certainly related.

Yeah, maybe it is rather unnecessary, it would make timedelta work (at least in principle) and would bring us closer to allowing more scalars, but that requires some restructure anyway.

You are right it seems unnecessary, I don't love that there is no guarantee nobody writes a `byteswap_input` ufunc, but that is pretty unlikely in practice.  So yeah... probably should just undo this again (and at best add a small comment).

mhvk:
> I don't love that there is no guarantee nobody writes a `byteswap_input` ufunc

I'd be happy to guard against that, as long as it is fast (maybe, `ufunc->types != NULL`?)

But it also seems fine not to guard against the hypothetical, when it is so difficult to come up with an example where it might be problematic.

seberg:
Yeah, let's try what you had then. I do think allowing a different output dtype makes sense (because it should be pretty much free), and I also think adding the resolve-dtypes call is likely good if just for silly safety.
But... I can't disagree that it shouldn't really matter in practice.

(Just one last note because I don't think I said it: Unary ufuncs don't do special scalar promotions, which means the int/float handling here should be correct.)

hawkinsp:
I think this PR introduced a race condition that we see in JAX's tsan CI (https://github.com/jax-ml/jax/actions/runs/19400717531):

```
WARNING: ThreadSanitizer: data race (pid=95179)
  Write of size 8 at 0x7fffb4060f78 by thread T3 (mutexes: read M0, write M1):
    #0 PyArrayIdentityHash_SetItem /__w/jax/jax/numpy/.mesonpy-tk47p_zd/../numpy/_core/src/common/npy_hashtable.cpp:224:20 (_multiarray_umath.cpython-314t-x86_64-linux-gnu.so+0x41dbbb) (BuildId: fe065841b00019c85589d862249e0c30081a9a3f)
    #1 promote_and_get_info_and_ufuncimpl(_tagPyUFuncObject*, tagPyArrayObject* const*, PyArray_DTypeMeta**, PyArray_DTypeMeta**, unsigned char) /__w/jax/jax/numpy/.mesonpy-tk47p_zd/../numpy/_core/src/umath/dispatching.cpp:871:17 (_multiarray_umath.cpython-314t-x86_64-linux-gnu.so+0x435463) (BuildId: fe065841b00019c85589d862249e0c30081a9a3f)
    #2 promote_and_get_info_and_ufuncimpl_with_locking(_tagPyUFuncObject*, tagPyArrayObject* const*, PyArray_DTypeMeta**, PyArray_DTypeMeta**, unsigned char) /__w/jax/jax/numpy/.mesonpy-tk47p_zd/../numpy/_core/src/umath/dispatching.cpp:1004:12 (_multiarray_umath.cpython-314t-x86_64-linux-gnu.so+0x432fd9) (BuildId: fe065841b00019c85589d862249e0c30081a9a3f)
    #3 promote_and_get_ufuncimpl /__w/jax/jax/numpy/.mesonpy-tk47p_zd/../numpy/_core/src/umath/dispatching.cpp:1101:22 (_multiarray_umath.cpython-314t-x86_64-linux-gnu.so+0x432fd9)
    #4 ufunc_generic_fastcall /__w/jax/jax/numpy/.mesonpy-tk47p_zd/../numpy/_core/src/umath/ufunc_object.c:4683:38 (_multiarray_umath.cpython-314t-x86_64-linux-gnu.so+0x4404df) (BuildId: fe065841b00019c85589d862249e0c30081a9a3f)
...

  Previous read of size 8 at 0x7fffb4060f78 by thread T2 (mutexes: read M0):
    #0 find_item(PyArrayIdentityHash const*, _object* const*) /__w/jax/jax/numpy/.mesonpy-tk47p_zd/../numpy/_core/src/common/npy_hashtable.cpp:79:13 (_multiarray_umath.cpython-314t-x86_64-linux-gnu.so+0x41ddd0) (BuildId: fe065841b00019c85589d862249e0c30081a9a3f)
    #1 PyArrayIdentityHash_GetItem /__w/jax/jax/numpy/.mesonpy-tk47p_zd/../numpy/_core/src/common/npy_hashtable.cpp:240:21 (_multiarray_umath.cpython-314t-x86_64-linux-gnu.so+0x41ddd0)
    #2 try_trivial_scalar_call /__w/jax/jax/numpy/.mesonpy-tk47p_zd/../numpy/_core/src/umath/ufunc_object.c:4340:22 (_multiarray_umath.cpython-314t-x86_64-linux-gnu.so+0x43f606) (BuildId: fe065841b00019c85589d862249e0c30081a9a3f)
    #3 ufunc_generic_fastcall /__w/jax/jax/numpy/.mesonpy-tk47p_zd/../numpy/_core/src/umath/ufunc_object.c:4425:13 (_multiarray_umath.cpython-314t-x86_64-linux-gnu.so+0x43f606)
    #4 ufunc_generic_vectorcall /__w/jax/jax/numpy/.mesonpy-tk47p_zd/../numpy/_core/src/umath/ufunc_object.c:4766:12 (_multiarray_umath.cpython-314t-x86_64-linux-gnu.so+0x43d60e) (BuildId: fe065841b00019c85589d862249e0c30081a9a3f)
    #5 _PyObject_VectorcallTstate /__w/jax/jax/cpython/./Include/internal/pycore_call.h:169:11 (python3.14+0x1fa17a) (BuildId: 010258c9a6c23915fd3f27ed01befca5ea57f252)
```

I suspect that's enough for you to figure it out but if you like I can make a better repro.

mhvk:
@hawkinsp - Ouch, that seems clearly a mistake here.

Looking through your trace, the data race occurs when the new code reads the ufunc implementation cache, presumably for the case that the identify has is not yet defined, as the race is with a `SetItem` inside `promote_and_get_ufuncimpl` in `dispatching.cpp`. I think the problem is that I simply cribbed my bit of code from `promote_and_get_info_and_ufuncimpl`, while I should have added a case for `Py_GIL_DISABLED` following the start of `promote_and_get_info_and_ufuncimpl_with_locking`.

Obviously, not difficult in principle to fix, though seeing this, it may be better to think through how to make this safer generally. For instance, should `PyArrayIdentityHash_GetItem` not take care of this if the GIL is disabled? After all, it is the cache itself which carries the mutex, so it would seem logical to do the locking directly in the code that uses it. Looking at `npy_hashtable`, it would seem relatively simple: just have separate versions for `*_GetItem` and `*_SetItem` with and without GIL.  If that works, one can get rid of `promote_and_get_ufuncimpl_with_locking`.

I am happy to try that (though not before Thursday), but one problem is that I'm not sure how to write a good test case...  @ngoldbaum, since you were the one that introduced the `_with_locking` variant, could you advice?

seberg:
I think we can just add the same locking code?  For testing, I think this test covers it:
https://github.com/numpy/numpy/blob/0432a855aa750b13de43625e02b87d0580290bb7/numpy/_core/tests/test_multithreading.py#L28
it just needs parametrization to also cover the scalar path probably.

As for the error checking issue (not discussed here). You should be able to make the `PyErr_Occurred()` conditional on the legacy wrapping, not sure how slow it actually is (i.e. whether that is worthwhile). I think the ufunc code may currently be doing it uncondtionally.

May look at it later too, both seem like pretty straight forward follow-ups.

mhvk:
@seberg - yes, agreed it is not too hard. Main question would seem to be whether to add it in my new fast path, or whether to move things more generally to where it is actually needed, i.e., locking in `npy_hashtable` itself, error checking inside the legacy loop wrapper. But having a quick local fix now and then refactoring would be a good solution too.

p.s. Unlikely to get to this myself until Thursday, so certainly happy if anyone else has a fix beforehand!

seberg:
> error checking inside the legacy loop wrapper

I have a PR for that one, and no that doesn't work, because you don't have the GIL there.  The second one is less urgent probably, but let's see.

mhvk:
> > error checking inside the legacy loop wrapper
> 
> I have a PR for that one, and no that doesn't work, because you don't have the GIL there. 

Ah, yes, I now saw the PR. Since it apparently has very little cost, definitely the easiest to go with a local check. (Conceptually, I'd still like to push it in the legacy loop wrappers even if it is fairly costly there, but am saying that without really knowing how much hassle it would be...)

ngoldbaum:
@mhvk let me know if you want me to take a crack at this. I can probably fix this on funded time for Quansight's free-threading project.

mhvk:
@ngoldbaum - I think it would be faster for you than for me, so definitely fine for you to do it. I'm very unlikely to have time myself before Friday.

PR #29823, opened 2025-09-28
ENH: cleanup warning generation and unmark xfailed tests

mattip:
~@mhvk noticed that~ (Edit: it was [someone else](https://github.com/numpy/numpy/pull/29813#issuecomment-3342253909), sorry Martin) I used `PyErr_Warn()` which was [deprecated in python 2.5](https://docs.python.org/3/whatsnew/2.5.html#build-and-c-api-changes) but not removed, so I went ahead and changed all the instances explicitly instead of relying [on the macro](https://github.com/python/cpython/blob/1c5b28405a77df68d40991051b4c0769e728bf76/Include/cpython/warnings.h#L19) 
```
// DEPRECATED: Use PyErr_WarnEx() instead.
#define PyErr_Warn(category, msg) PyErr_WarnEx((category), (msg), 1)
```

When running tests, I noticed two tests that XPASS (i.e. are expected to fail but pass successfully) so I removed the marker.

### DIFF
--- a/numpy/_core/src/umath/ufunc_object.c
+++ b/numpy/_core/src/umath/ufunc_object.c
@@ -4488,9 +4488,9 @@ ufunc_generic_fastcall(PyUFuncObject *ufunc,
 
     /* Warn if "where" is used without "out", issue 29561 */
     if ((where_obj != NULL) && (full_args.out == NULL) && (out_obj == NULL)) {
-        if (PyErr_Warn(PyExc_UserWarning,
+        if (PyErr_WarnEx(PyExc_UserWarning,
                 "'where' used without 'out', expect unitialized memory in output. "
-                "If this is intentional, use out=None.") < 0) {
+                "If this is intentional, use out=None.", 1) < 0) {
             goto fail;
         }
     }
--- a/numpy/_core/src/common/npy_longdouble.c
+++ b/numpy/_core/src/common/npy_longdouble.c
@@ -144,8 +144,8 @@ npy_longdouble_from_PyLong(PyObject *long_obj) {
     result = NumPyOS_ascii_strtold(cstr, &end);
     if (errno == ERANGE) {
         /* strtold returns INFINITY of the correct sign. */
-        if (PyErr_Warn(PyExc_RuntimeWarning,
-                "overflow encountered in conversion from python long") < 0) {
+        if (PyErr_WarnEx(PyExc_RuntimeWarning,
+                "overflow encountered in conversion from python long", 1) < 0) {
             goto fail;
         }
     }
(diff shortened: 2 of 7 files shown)


mattip:
A qemu docker build is failing
```
docker run --rm --privileged loongcr.lcpu.dev/multiarch/archlinux
Unable to find image 'loongcr.lcpu.dev/multiarch/archlinux:latest' locally
docker: Error response from daemon: received unexpected HTTP status: 502 Bad Gateway
```

@pengxu originally laid this out in #27856. Any idea what is going wrong?
This is from

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4488,9 +4488,9 @@ ufunc_generic_fastcall(PyUFuncObject *ufunc,
 
     /* Warn if "where" is used without "out", issue 29561 */
     if ((where_obj != NULL) && (full_args.out == NULL) && (out_obj == NULL)) {
-        if (PyErr_Warn(PyExc_UserWarning,
+        if (PyErr_WarnEx(PyExc_UserWarning,
                 "'where' used without 'out', expect unitialized memory in output. "
-                "If this is intentional, use out=None.") < 0) {
+                "If this is intentional, use out=None.", 1) < 0) {
```
mhvk, line 4493: Do you want to make this 2, given the request in https://github.com/numpy/numpy/pull/29813#issuecomment-3342253909? Also makes sense to keep this just a straight-up replacement PR.

mhvk (approved):
No worries about the wrong "m" - but since I was here and it seemed easy, I thought I might as well review.

numpy/_core/src/umath/ufunc_object.c
mattip: > Do you want to make this 2 ...

I think 1 is correct. Given a test file /tmp/test_warning.py
```
import numpy as np

def func(a):
    return np.add(a, a, where=a<10)

def call_func(a):
    func(a)

a = np.arange(1000)
call_func(a)
```

using 2 produces
```
$ python /tmp/test_warning.py 
/tmp/test_warning.py:7: UserWarning: 'where' used without 'out', expect unitialized memory in output. If this is intentional, use out=None.
  func(a)
```

and 1 produces
```
$ python /tmp/test_warning.py 
/tmp/test_warning.py:4: UserWarning: 'where' used without 'out', expect unitialized memory in output. If this is intentional, use out=None.
  return np.add(a, a, where=a<10)
```

> Also makes sense to keep this just a straight-up replacement PR

I view both commits as cleanup, since they really do not change anything

charris:
Thanks Matti.

numpy/_core/src/umath/ufunc_object.c
mhvk: Yes, all makes sense, nice!

PR #29827, opened 2025-09-28
ENH: add Array API compatibility to ``polynomial.polyval``

### LINKED ISSUE #29680: BUG: polynomial evaluation functions assume numpy array and work inconsistently with other array types
### Describe the issue:

The following code works with numpy arrays, lists and tuples but breaks on `jax.numpy` arrays for example:

https://github.com/numpy/numpy/blob/f675dbb5c4fa78869afca042e805b6962a0e7690/numpy/polynomial/polynomial.py#L748-L751

```python
This results in inconsistent behavior:


>>> from numpy.polynomial import polynomial as poly
>>> b = jax.numpy.array(
       [[1, 2, 3],
       [2, 3, 4]]
)
>>> poly.polyval(np.array(b),np.array(b), tensor=True)
array([[[ 3.,  5.,  7.],
        [ 5.,  7.,  9.]],

       [[ 5.,  8., 11.],
        [ 8., 11., 14.]],

       [[ 7., 11., 15.],
        [11., 15., 19.]]])
>>> poly.polyval(b,b, tensor=True)
Array([[ 3.,  8., 15.],
       [ 5., 11., 19.]], dtype=float32)
```

### Error message:

```shell

```

```python
>>> import sys, numpy; print(numpy.__version__); print(sys.version)
2.3.2
3.13.7 (v3.13.7:bcee1c32211, Aug 14 2025, 19:10:51) [Clang 16.0.0 (clang-1600.0.26.6)]
```

### Runtime Environment:

_No response_

### Context for the issue:

It may be a better idea to check for the attribute `.ndim` that is used afterwards or the attribute `__array__` to identify an Array-Like.

I would be willing to make a Pull request.

samruddhibaviskar11:
Closes: #29680

## Problem

`numpy.polynomial.polyval` currently checks explicitly for `np.ndarray`.

As a result:
- ✅ NumPy arrays, lists, and tuples work
- ❌ JAX, CuPy, Dask, and other Array API–compatible arrays are treated as scalars → inconsistent/broken behavior

### Example (from issue #29680)

```python
from numpy.polynomial import polynomial as poly
import jax.numpy as jnp

b = jnp.array([[1, 2, 3], [2, 3, 4]])
poly.polyval(np.array(b), np.array(b), tensor=True).shape
# (3, 2, 3)
poly.polyval(b, b, tensor=True).shape
# (2, 3)  # inconsistent!
Fix
Minimal, backward-compatible change:

python
Copy code
if isinstance(x, (tuple, list)):
    x = np.asarray(x)

nd = getattr(x, "ndim", None)
if tensor and isinstance(nd, int) and nd > 0:
    c = c.reshape(c.shape + (1,) * nd)
Keeps tuple/list → np.asarray conversion

Uses .ndim instead of isinstance(x, np.ndarray)

Ensures Array API–compatible arrays (JAX, CuPy, etc.) broadcast consistently

Leaves scalars and polynomial objects untouched

Tests
Added regression and edge-case tests in numpy/polynomial/tests/test_polynomial.py:
✅ JAX arrays with tensor=True (via pytest.importorskip)
✅ CuPy arrays with tensor=True
✅ Scalar behavior preserved (polyval(2, [1, 2, 3]) == 17)
✅ Polynomial objects preserved as scalars
✅ 0-D arrays handled correctly
✅ Empty arrays handled correctly
✅ Object dtype arrays still work
✅ tensor=True vs tensor=False semantics verified

Thanks to @seberg, @charris, and @FelixBenning for the helpful discussion and guidance.

Future scope
Extend the same safe pattern to other polynomial functions (polyval2d, polyval3d, polygrid2d, etc.) in follow-up PRs.

### DIFF
--- a/numpy/polynomial/polynomial.py
+++ b/numpy/polynomial/polynomial.py
@@ -745,17 +745,25 @@ def polyval(x, c, tensor=True):
     if c.dtype.char in '?bBhHiIlLqQpP':
         # astype fails with NA
         c = c + 0.0
+
+    # Convert lists/tuples to arrays, leave array-likes and objects untouched
     if isinstance(x, (tuple, list)):
         x = np.asarray(x)
+
+    # Safely check for array-like objects exposing .ndim
     nd = getattr(x, "ndim", -1)
-    if tensor and nd > 0:
+
+    # Only reshape for tensor=True (outer-product evaluation)
+    # Prevents double-reshaping inside polyval2d/polyval3d
+    if tensor and nd > 0 and c.ndim == 1:
         c = c.reshape(c.shape + (1,) * nd)
+    # Handle tensor=False for multi-dimensional coefficient arrays
+    # (preserves Array API consistency and per-column evaluation)
     elif not tensor and c.ndim > 1:
-        # Evaluate column-wise for 2D+ coefficient arrays
         return np.array([
-        polyval(x, c[:, i], tensor=False)
-        for i in range(c.shape[1])
-    ])
+            polyval(x, c[:, i], tensor=False)
+            for i in range(c.shape[1])
+        ])
     c0 = c[-1] + x * 0
     for i in range(2, len(c) + 1):
         c0 = c[-i] + c0 * x
--- a/numpy/polynomial/polynomial.py
+++ b/numpy/polynomial/polynomial.py
@@ -747,10 +747,15 @@ def polyval(x, c, tensor=True):
         c = c + 0.0
     if isinstance(x, (tuple, list)):
         x = np.asarray(x)
-    nd = getattr(x, "ndim", None)
-    if tensor and isinstance(nd, int) and nd > 0:
+    nd = getattr(x, "ndim", -1)
+    if tensor and nd > 0:
         c = c.reshape(c.shape + (1,) * nd)
-
+    elif not tensor and c.ndim > 1:
+        # Evaluate column-wise for 2D+ coefficient arrays
+        return np.array([
+        polyval(x, c[:, i], tensor=False)
+        for i in range(c.shape[1])
+    ])
     c0 = c[-1] + x * 0
     for i in range(2, len(c) + 1):
         c0 = c[-i] + c0 * x
--- a/numpy/polynomial/polynomial.py
+++ b/numpy/polynomial/polynomial.py
@@ -747,8 +747,9 @@ def polyval(x, c, tensor=True):
         c = c + 0.0
     if isinstance(x, (tuple, list)):
         x = np.asarray(x)
-    if isinstance(x, np.ndarray) and tensor:
-        c = c.reshape(c.shape + (1,) * x.ndim)
+    nd = getattr(x, "ndim", None)
+    if tensor and isinstance(nd, int) and nd > 0:
+        c = c.reshape(c.shape + (1,) * nd)
 
     c0 = c[-1] + x * 0
     for i in range(2, len(c) + 1):
--- a/numpy/polynomial/tests/test_polynomial.py
+++ b/numpy/polynomial/tests/test_polynomial.py
@@ -690,3 +690,59 @@ def test_polygrid2d_array_function_hook():
     c = ArrayFunctionInterceptor()
     result = np.polynomial.polynomial.polygrid2d(x, y, c)
     assert result == "intercepted"
+
+def test_polyval_with_jax_array_tensor(self):
+    jnp = pytest.importorskip("jax.numpy")
+    b = jnp.array([[1, 2, 3], [2, 3, 4]])
+    c = jnp.array(b)
+    # tensor=True should broadcast like NumPy
+    np_result = poly.polyval(np.array(b), np.array(b), tensor=True)
+    jax_result = poly.polyval(b, c, tensor=True)
+    assert np.allclose(np_result, np.array(jax_result), atol=1e-12)
+
+def test_polyval_with_cupy_array_tensor(self):
+    cp = pytest.importorskip("cupy")
+    b = cp.array([[1, 2], [3, 4]])
+    c = cp.array([1., 2., 3.])
+    # tensor=True with CuPy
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: numpy/polynomial/polynomial.py [hunk 1]
```diff
@@ -747,8 +747,9 @@ def polyval(x, c, tensor=True):
         c = c + 0.0
     if isinstance(x, (tuple, list)):
         x = np.asarray(x)
-    if isinstance(x, np.ndarray) and tensor:
-        c = c.reshape(c.shape + (1,) * x.ndim)
+    nd = getattr(x, "ndim", None)
+    if tensor and isinstance(nd, int) and nd > 0:
```
mattip: Drive-by simplification
```suggestion
    nd = getattr(x, "ndim", -1)
    if tensor and nd > 0:
```

samruddhibaviskar11: Thanks for the suggestion! I’ve simplified the code as suggested.

ev-br:
What this PR does is different from what is typically meant under Array API compatibility. In this PR, `polyval`

1. returns a numpy array for inputs from a different namespace; typically Array APi is xp-array in xp-array out
2. allows mixed inputs while typically you want to error out for `polyal(x=numpy_array, c=torch_tensor)`
3. will fail with an unhelpful message for inputs on a non-CPU device. For instance, if `x` is a torch GPU tensor, it has the `ndim` attribute but `np.array(c)` fails:  it will otherwise have to transfer data between the device and the host.

In SciPy, we sometimes use a related pattern: convert to numpy, compute, convert back --- for functions which use custom C kernels (e.g. `scipy.ndimage`). These kernels are necessarily CPU-only, thus simply cannot work for GPU arrays, and scipy has to somehow delegate the work to somewhere which has matching CUDA kernels (e.g. `cupy`).

`numpy.polynomial` however, does not carry dedicated C kernels, and can in principle be made array namespace agnostic.

A typical pattern is along the lines of 

```
def polyval(x, c, ...):
     xp = array_namespace(x, c)     # from array_api_compat, fails on a namespace mismatch
     x = xp.asarray(x)
     c = xp.asarray(c)

     .... proceed with xp-namespaced arrays
```
 
Downstream, we would typically use `array_namespace` function from the `array_api_compat` package. It might be awkward to rely on `array_api_compat` in numpy itself, so probably one can use `x.__array_namespace__()` instead, as recommended in https://data-apis.org/array-api/latest/purpose_and_scope.html#how-to-adopt-this-api

samruddhibaviskar11:
> What this PR does is different from what is typically meant under Array API compatibility. In this PR, `polyval`
> 
> 1. returns a numpy array for inputs from a different namespace; typically Array APi is xp-array in xp-array out
> 2. allows mixed inputs while typically you want to error out for `polyal(x=numpy_array, c=torch_tensor)`
> 3. will fail with an unhelpful message for inputs on a non-CPU device. For instance, if `x` is a torch GPU tensor, it has the `ndim` attribute but `np.array(c)` fails:  it will otherwise have to transfer data between the device and the host.
> 
> In SciPy, we sometimes use a related pattern: convert to numpy, compute, convert back --- for functions which use custom C kernels (e.g. `scipy.ndimage`). These kernels are necessarily CPU-only, thus simply cannot work for GPU arrays, and scipy has to somehow delegate the work to somewhere which has matching CUDA kernels (e.g. `cupy`).
> 
> `numpy.polynomial` however, does not carry dedicated C kernels, and can in principle be made array namespace agnostic.
> 
> A typical pattern is along the lines of
> 
> ```
> def polyval(x, c, ...):
>      xp = array_namespace(x, c)     # from array_api_compat, fails on a namespace mismatch
>      x = xp.asarray(x)
>      c = xp.asarray(c)
> 
>      .... proceed with xp-namespaced arrays
> ```
> 
> Downstream, we would typically use `array_namespace` function from the `array_api_compat` package. It might be awkward to rely on `array_api_compat` in numpy itself, so probably one can use `x.__array_namespace__()` instead, as recommended in https://data-apis.org/array-api/latest/purpose_and_scope.html#how-to-adopt-this-api

Thanks @ev-br! You're right — this isn't proper Array API compatibility.

I'd like to implement it correctly using the xp = array_namespace(x, c) pattern you described. A few questions:

• Since NumPy can't depend on array_api_compat, should I use x.__array_namespace__() directly or implement a minimal helper?
• For scope — just polyval first, or include other polynomial functions as well?

I'll ensure backward compatibility (lists → NumPy, scalars/polynomial objects preserved) and add tests for JAX/CuPy/PyTorch, including GPU tensors.

ev-br:
> For scope — just polyval first, or include other polynomial functions as well?

Long-term, upstream will certainly want the whole package. How to get there depends on the person(s) doing the work (you?)

> Since NumPy can't depend on array_api_compat, should I use x.array_namespace() directly or implement a minimal helper?

Realistically, you'd need `array_api_compat` for `array_namespace`, `array_api_extra` for `atleast_nd` and `array-api-strict` for testing. All of these packages depend on numpy. Thus I personally suspect the most productive path is the second bullet from https://github.com/numpy/numpy/issues/29680#issuecomment-3342656197 --- split `np.polynomial` to a separate package under the `numpy` umbrella. Note I'm not a numpy dev though, and it really is a question to numpy devs.

samruddhibaviskar11:
Thanks @ev-br — that’s very helpful context.

I see two possible paths forward based on your suggestion and the earlier discussion:

Option 1: Split numpy.polynomial into a separate package (e.g. numpy-polynomial) for full Array API support.
This would avoid circular dependencies and make it easier to rely on array_api_compat / array_api_extra, similar to what SciPy has done with its vendored _polyutils.

Option 2: Keep it in NumPy for now and implement a minimal fix using __array_namespace__() directly —
fixing the current JAX/CuPy behavior while preserving backward compatibility (lists → NumPy, scalars and Polynomial objects unchanged).

@mattip @seberg @charris @jorenham  — what would you prefer?
I’m happy to proceed either way, just want to align with NumPy’s long-term direction.

PR #29836, opened 2025-09-30
ENH, FEAT: Reorganize finfo and add new constant slot

### LINKED ISSUE #27231: ENH: `np.finfo` support for Numpy User DTypes

SwayamInSync:
closes #27231 

## Summary

This pull request removes the legacy `MachAr` class and related files from NumPy, streamlining the codebase and updating the internal dtype API to support querying numerical constants directly from dtypes. The changes modernize how machine limits and constants are handled, moving away from Python-side introspection to a more robust and extensible C API.

**Removal of legacy Python code:**

* Removed the entire `numpy/_core/_machar.py` file, which contained the deprecated `MachAr` class for floating-point machine parameter introspection.
* Removed the associated type stub file `numpy/_core/_machar.pyi` and its references from the build configuration, fully eliminating Python-side support for `MachAr`. 
* Cleaned up imports in `numpy/_core/__init__.py` to remove `_machar`, reflecting the removal of the module.
* Removed initialization of legacy limits in `numpy/__init__.py` as they are no longer needed with the new approach.

**Enhancements to dtype API for numerical constants:**

* Added new constant IDs (e.g., `NPY_CONSTANT_zero`, `NPY_CONSTANT_one`, `NPY_CONSTANT_maximum_finite`, etc.) and a new `PyArrayDTypeMeta_GetConstant` function to the dtype API, enabling direct querying of numerical constants from dtypes in C. 
* Updated internal offset definitions and macros in `numpy/_core/include/numpy/dtype_api.h` to accommodate the new constants and API changes.

**Miscellaneous codebase improvements:**

* Added `#include <float.h>` in `numpy/_core/src/multiarray/arraytypes.c.src` to ensure access to floating-point limits in C.
* Included the updated `numpy/dtype_api.h` header in `arraytypes.c.src` to support the new constant querying API.


numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -4327,6 +4327,101 @@ normalize_axis_index(PyObject *NPY_UNUSED(self),
...
 
+static PyObject *
+_populate_finfo_constants(PyObject *NPY_UNUSED(self), PyObject *args)
+{
+    if (PyTuple_Size(args) != 2) {
+        PyErr_SetString(PyExc_TypeError, "Expected 2 arguments");
+        return NULL;
+    }
+    PyObject *finfo = PyTuple_GetItem(args, 0);
+    if (finfo == NULL || finfo == Py_None) {
```
seberg, line 4338: Can't be `NULL`, but doesn't matter.  I might not bother, it'll just say "can't set attributes of NoneType" without this, but happy either way.

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -4327,6 +4327,101 @@ normalize_axis_index(PyObject *NPY_UNUSED(self),
...
+        {"maxexp", NPY_CONSTANT_finfo_max_exp, 1},
+        {"precision", NPY_CONSTANT_finfo_decimal_digits, 1},
+        {NULL, -1},
+    };
+
+    for (int i = 0; finfo_constants[i].name != NULL; i++) 
+    {
+        PyObject *value_obj;
+        if (!finfo_constants[i].is_int) {
+            memset(buffer_data, 0, PyArray_DESCR(buffer_array)->elsize);
```
seberg: Rather than memsetting here, I would suggest to create a large enough array and just write to `buffer_data[i * elsize]` or `buffer_data += elsize` in the loop.

The reason is that technically memsetting could be incorrect if the dtype includes references such as object dtype (I admit, hard to imagine for an inexact dtype, but still).

numpy/_core/include/numpy/dtype_api.h
```diff
@@ -479,6 +480,42 @@ typedef PyArray_Descr *(PyArrayDTypeMeta_EnsureCanonical)(PyArray_Descr *dtype);
...
+#define NPY_CONSTANT_nan 8
+#define NPY_CONSTANT_finfo_radix 9
+#define NPY_CONSTANT_finfo_eps 10
+#define NPY_CONSTANT_finfo_smallest_normal 11
+#define NPY_CONSTANT_finfo_smallest_subnormal 12
+/* Constants that are always of integer type, value is `npy_intp/Py_ssize_t` */
+#define NPY_CONSTANT_finfo_nmant 13
+#define NPY_CONSTANT_finfo_min_exp 14
+#define NPY_CONSTANT_finfo_max_exp 15
+#define NPY_CONSTANT_finfo_decimal_digits 16
```
seberg: Would be nice to see opinions about the choice of constant here, and also the constants derived in Python.

numpy/_core/src/multiarray/dtypemeta.c
```diff
@@ -1165,17 +1178,21 @@ dtypemeta_wrap_legacy_descriptor(
...
-            nonparametric_discover_descr_from_pyobject);
+        nonparametric_discover_descr_from_pyobject);
     dt_slots->is_known_scalar_type = python_builtins_are_known_scalar_types;
     dt_slots->common_dtype = default_builtin_common_dtype;
     dt_slots->common_instance = NULL;
     dt_slots->ensure_canonical = ensure_native_byteorder;
     dt_slots->get_fill_zero_loop = NULL;
     dt_slots->finalize_descr = NULL;
+    dt_slots->setitem = legacy_setitem_using_DType;
+    dt_slots->getitem = legacy_getitem_using_DType;
```
seberg: Isn't this what breaks CI?  I think you need to remove this again (leave it at `NULL`), might as well delete the `legacy_setitem_using_DType` function also.

numpy/_core/src/multiarray/dtypemeta.c
```diff
@@ -1165,17 +1178,21 @@ dtypemeta_wrap_legacy_descriptor(
     dtype_class->scalar_type = descr->typeobj;
     dtype_class->type_num = descr->type_num;
     dt_slots->f = *arr_funcs;
+    dt_slots->setitem = arr_funcs->setitem;
+    dt_slots->getitem = arr_funcs->getitem;
```
seberg:
```suggestion
```
This isn't right, just delete this again, it was a failed try... it's more ocmplicated.

seberg (commented):
Thanks a few small comments.  It seems like at least one run fails around querying the subnormal information from the macros.
Not sure what to do about that, I suppose either hard-code or maybe use `nextafter` for them.
(if we do that, it might be nice to see if we can find out whether subnormals work in that step, no idea what `nextafter` does.  OTOH, the typical thing is FTZ mode, which I suspect still allows manual creation with `nextafter`... so it's really a user problem and we would need an `ftz` attribute on `finfo` rather than changing what `smallest_subnormal` gives.)

numpy/_core/src/multiarray/multiarraymodule.c
SwayamInSync: Ah yes in principle, for such cases memsetting could've lost the data pointer but refernce count remains. 
Thanks for catching this.

numpy/_core/src/multiarray/dtypemeta.c
SwayamInSync: The `default_funcs` structure at line 165 still uses these legacy wrapper functions, we can set them NULL here but keep those functions?

seberg: Yes, can probably set to `NULL` also, or just remove the change, since it really is unrelated (probably NULLed in an above memset already).

SwayamInSync:
> Thanks a few small comments. It seems like at least one run fails around querying the subnormal information from the macros. Not sure what to do about that, I suppose either hard-code or maybe use `nextafter` for them. (if we do that, it might be nice to see if we can find out whether subnormals work in that step, no idea what `nextafter` does. OTOH, the typical thing is FTZ mode, which I suspect still allows manual creation with `nextafter`... so it's really a user problem and we would need an `ftz` attribute on `finfo` rather than changing what `smallest_subnormal` gives.)

I tried modifying the template to use nextafter when macro is not defined, only hardocded the16 bit (it was going 0)

edit: wait, maybe I can do better, let me fix all the other error cases first

SwayamInSync:
Any reason why for double-double format tiny was set to NAN instead of `DBL_MIN`?
```
float_dd_ma = MachArLike(ld,
                         machep=-105,
                         negep=-106,
                         minexp=-1022,
                         maxexp=1024,
                         it=105,
                         ...
                         eps=exp2(ld(-105)),  # This is 2^-105!
                         epsneg=exp2(ld(-106)),
                         huge=nextafter(ld(inf), ld(0)),
                         tiny=nan,  # Set to NaN for double-double!
                         smallest_subnormal=nextafter(0., 1.))
```

seberg:
> Any reason why for double-double format tiny was set to NAN

IIRC it was some weird/wrong value, then things were changed and nobody wanted to figure out the right value, so it was set to NaN because that seemed still better than being nonsense.
I _think_ there is a difficulty that the tiny definition is a bit trickier, because the precision of double-double changes over the full range.
Anyway, have a brief look at gh-19511 and then set it to something reasonable if you find a value (I would love to use the one from the headers, but dunno if even those are unreliable...).

SwayamInSync:
>I would love to use the one from the headers, but dunno if even those are unreliable

I think `tiny` might be taken from headers (as the underflow checks are passed) but the `huge` value will need to be what is set in machar above ` nextafter(ld(inf), ld(0))`

Let me confirm this and post here whatever suits the best

SwayamInSync:
Only override the `EPSILON-2^-105` to  and `MAX_FINITE=npy_nextafterl(NPY_INFINITY, 0.0L)` seems to worked for native ppc64le

SwayamInSync:
For this particular issue of testing array_repr
https://github.com/numpy/numpy/blob/50204c8ef6bd66176341bc6d29c97ad94a8a3b68/numpy/_core/tests/test_longdouble.py#L288-L294

platforms where `long double == double (float64)` earlier it is only getting passed because of the difference in data type namings and default precision value of 8 for printing

https://github.com/numpy/numpy/blob/50204c8ef6bd66176341bc6d29c97ad94a8a3b68/numpy/_core/printoptions.py#L18

## Windows Old behaviour (before refactoring)
Resulting in 2 arrays as following
```python
In [1]: import numpy as np

In [2]: np.__version__
Out[2]: '2.3.3'

In [3]: LD_INFO = np.finfo(np.longdouble)

In [4]: o = 1 + LD_INFO.eps

In [5]: a = np.array([o])

In [6]: b = np.array([1], dtype=np.longdouble)

In [7]: a, b
Out[7]: (array([1.]), array([1.], dtype=float64))
```
As you can see the only difference is in dtype leading to that test's assert to fail leading to passing the test

## In this new behaviour
```python
In [1]: import numpy as np

In [2]: np.__version__
Out[2]: '2.4.0.dev0+git20250930.d8508a2'

In [3]: LD_INFO = np.finfo(np.longdouble)

In [4]: o = 1 + LD_INFO.eps

In [5]: a = np.array([o])

In [6]: b = np.array([1], dtype=np.longdouble)

In [7]: a, b
Out[7]: (array([1.], dtype=float64), array([1.], dtype=float64))
```

So since both get dispatched from the same dtype, results in failing the test, I am not sure whether testing this was the intention behind this test. Anyways the hack to retain the original behaviour is as follows

inside `getlimits.py`
```python
# On platforms where longdouble is the same size as double (e.g., Windows),
# use double descriptor to populate constants for backward compatibility.
# The old MachArLike code would match the float64 signature on such platforms
# and return float64 scalars.
if (self.dtype.type == ntypes.longdouble and
    self.dtype.itemsize == numeric.dtype(ntypes.double).itemsize):
     self.dtype = populate_dtype
     populate_dtype = numeric.dtype(ntypes.double)
else:
     populate_dtype = self.dtype
# Fills in all constants defined directly on the dtype (in C)
_populate_finfo_constants(self, populate_dtype)
```

This override to use `float64` which is as per here the implied dtype and might not need to explicitly put in array representations
https://github.com/numpy/numpy/blob/50204c8ef6bd66176341bc6d29c97ad94a8a3b68/numpy/_core/arrayprint.py#L1520-L1562

@seberg what you suggest? keeping the current way (which retain the past behaviour) or need to change the array print behaviour by increasing the precision to respectively used dtype and ignorning the name of dtype (Not sure how useful, maybe can do in a different PR)?

SwayamInSync:
The Qemu/loongarch64 docker image is not even building so considering that out for now. This PR is ready for the review

numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -4318,10 +4320,209 @@ PyArray_DescrFromType(int type)
...
+            val = 1;
+            break;
+        case NPY_CONSTANT_minimum_finite:
+            val = @NEG_MAX@;
+            break;
+        case NPY_CONSTANT_maximum_finite:
+            #if defined(HAVE_LDOUBLE_IBM_DOUBLE_DOUBLE_LE) || \
+                defined(HAVE_LDOUBLE_IBM_DOUBLE_DOUBLE_BE)
+                /* For IBM double-double, use nextafter(inf, 0) to get the true
+                 * maximum representable value (matches old MachArLike behavior) */
```
seberg, line 4447: Interesting, the headers really gave a different value?  (ridiculous but OK...)

numpy/_core/getlimits.py
```diff
@@ -548,75 +225,118 @@ def __new__(cls, dtype):
...
+        # On platforms where longdouble is the same size as double (e.g., Windows),
+        # use double descriptor to populate constants for backward compatibility.
+        # The old MachArLike code would match the float64 signature on such platforms
+        # and return float64 scalars and use float64 as the dtype.
+        if (self.dtype.type == ntypes.longdouble and
+            self.dtype.itemsize == numeric.dtype(ntypes.double).itemsize):
+            populate_dtype = numeric.dtype(ntypes.double)
+            self.dtype = populate_dtype  # Also update self.dtype for consistency
+        else:
+            populate_dtype = self.dtype
```
seberg: @SwayamInSync since you asked on this one.  If changing the test to just compare the `str(a[0])` on the scalar works (I _think_ it should), then I would do that (scalars should print with full required precision).
But if that isn't straight forward, I am happy to defer this... It was a weird artifact of the old approach!

(FWIW, it may very well be that there was a time when the test did something reasonable, just not in the past 10 years :).)

numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -4318,10 +4320,209 @@ PyArray_DescrFromType(int type)
...
+/*
+ * On PPC64 systems with IBM double-double format, LDBL_EPSILON from the C
+ * standard library is incorrect. IBM double-double is a pair of IEEE binary64
+ * values (not a true IEEE quad), giving ~106 bits of mantissa precision (53+53).
+ * The correct epsilon is 2^-105, but glibc incorrectly returns 2^-1074 (DBL_TRUE_MIN).
+ */
+#if defined(HAVE_LDOUBLE_IBM_DOUBLE_DOUBLE_LE) || \
+    defined(HAVE_LDOUBLE_IBM_DOUBLE_DOUBLE_BE)
+    #undef LDBL_EPSILON
+    #define LDBL_EPSILON 2.46519032881566189191e-32L  /* 2^-105 */
```
seberg: I wonder what `npy_nextafterl(1L, 2L)` gives on these systems.  FWIW, I think there may be two different double-double versions.  Normalized and not normalized.
We care about normalized (which I think makes sense even for unnormalized for eps, IMO!).
But if you think of non-normalized, then `1 + TINY` exists and in that sense eps is 1^-1074.

(Long story short for: I think I agree with disagreeing with the headers, that seems like the _useful_ "eps" even if it might not be `npy_nextafterl(1L, 2L)`!)

numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -4318,10 +4320,209 @@ PyArray_DescrFromType(int type)
...
+                #else
+                    val = npy_nextafter(0.0, 1.0);
+                #endif
+            }
+            else {  // long double
+                #ifdef LDBL_TRUE_MIN
+                    val = LDBL_TRUE_MIN;
+                #else
+                    val = npy_nextafterl(0.0L, 1.0L);
+                #endif
```
seberg: I might suggest shortening it here, e.g. by using:
```
#ifndef DBL_TRUE_MIN
#define DBL_TRUE_MIN npy_nextafter(0.0, 1.0)
#endif
```
etc. before this loop.  I think `npy_half` was just a uint16, so you can put assign the bit-value directly.

numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -4318,10 +4320,209 @@ PyArray_DescrFromType(int type)
...
+            val = @ABB@_MAX;
+            break;
+        case NPY_CONSTANT_inf:
+            val = (@type@)NPY_INFINITYF;
+            break;
+        case NPY_CONSTANT_nan:
+            val = (@type@)NPY_NANF;
+            break;
+        case NPY_CONSTANT_finfo_radix:
+            *(npy_intp *)ptr = @RADIX@;
```
seberg: Ah, you changed the radix to an integer?  Sounds fine, I had put it as a non-int because it seemed useful to derive other values via `radix` math.
(Honestly, it's always 2 anyway, so not sure if it is important but if we use it, let's keep it, who knows :))

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -4327,6 +4327,110 @@ normalize_axis_index(PyObject *NPY_UNUSED(self),
...
+        {"precision", NPY_CONSTANT_finfo_decimal_digits, 1},
+        {NULL, -1},
+    };
+
+    int n_float_constants = 0;
+    for (int i = 0; finfo_constants[i].name != NULL; i++) {
+        if (!finfo_constants[i].is_int) {
+            n_float_constants++;
+        }
+    }
```
seberg, line 4371: Before we do this, just hard-code the length of finfo_constants.  NULL termination is neat if we don't need a length, but not now :).

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -4327,6 +4327,110 @@ normalize_axis_index(PyObject *NPY_UNUSED(self),
...
+        }
+        int res = PyObject_SetAttrString(finfo, finfo_constants[i].name, value_obj);
+        Py_DECREF(value_obj);
+        if (res < 0) {
+            goto fail;
+        }
+    }
+
+  Py_DECREF(buffer_array);
+  Py_RETURN_NONE;
```
seberg:
```suggestion
    Py_RETURN_NONE;
```
(and above)

seberg (commented):
Just leaving as comments. I like this, how happy are you with the approach?

Did you have a look at the derived values I had already created?  The other question is whether we should rescue some of the old parts (much reduced) and slam them into tests?

Need to have another pass, but wanted to give these comments.  But also, I think it's basically done.
(With this change, I assume there may be weird platform fallouts eventually, but that is something to deal with when it happens.)

numpy/_core/getlimits.py
SwayamInSync: Yeah but the test's name was `test_array_repr` and if we modified it to check scalar :)
We have another option to skip it if `longdouble == double`

numpy/_core/src/multiarray/arraytypes.c.src
SwayamInSync: ` 2^-1074` IIRC this was the value coming from the header

numpy/_core/getlimits.py
seberg: You would have to bump the precision (use printoptions) for array, then its fine with str (to drop the dtype). But also fine or easier to do scalars (or both) and change the test name. They use the same repr, but scalars should always print with sufficient precision.

numpy/_core/include/numpy/dtype_api.h
```diff
@@ -479,6 +480,42 @@ typedef PyArray_Descr *(PyArrayDTypeMeta_EnsureCanonical)(PyArray_Descr *dtype);
...
+ * These are also used to expose .finfo and .iinfo for example.
+ */
+/* Numerical constants */
+#define NPY_CONSTANT_zero 1
+#define NPY_CONSTANT_one 2
+#define NPY_CONSTANT_all_bits_set 3
+#define NPY_CONSTANT_maximum_finite 4
+#define NPY_CONSTANT_minimum_finite 5
+#define NPY_CONSTANT_inf 6
+#define NPY_CONSTANT_ninf 7
```
juntyr, line 494: Could this use a clearer name, e.g. `neg_inf`

numpy/_core/include/numpy/dtype_api.h
```diff
@@ -479,6 +480,42 @@ typedef PyArray_Descr *(PyArrayDTypeMeta_EnsureCanonical)(PyArray_Descr *dtype);
...
+#define NPY_CONSTANT_finfo_eps 10
+#define NPY_CONSTANT_finfo_smallest_normal 11
+#define NPY_CONSTANT_finfo_smallest_subnormal 12
+/* Constants that are always of integer type, value is `npy_intp/Py_ssize_t` */
+#define NPY_CONSTANT_finfo_nmant 13
+#define NPY_CONSTANT_finfo_min_exp 14
+#define NPY_CONSTANT_finfo_max_exp 15
+#define NPY_CONSTANT_finfo_decimal_digits 16
+
+/* It may make sense to continue with other constants here, e.g. pi, etc? */
```
juntyr, line 506: It would be great to have precision-dependent definitions of pi and e

seberg: Yes, and I think this is a good way (I once thought we might have a constants dtype or a ufunc, but I prefer this for being light-weight a ufunc may still be a solution for scipy if they ever want such a thing...).

But, I would suggest to make that a second PR and focus on the finfo here (I think even iinfo isn't included, and we should definitely follow up on that as well).

numpy/_core/include/numpy/dtype_api.h
seberg: Sounds good, I know I added it, but I guess we could even remove it and define it in Python as `-inf`.  I think we removed `np.ninf` after all.  (But I don't mind either way).

SwayamInSync:
@seberg you can have a round 2 now

>  how happy are you with the approach?

I like this one, we are more closer to standards. Also for IBM DD format tiny coming from headers seems to pass the test so maybe we close the issue https://github.com/numpy/numpy/issues/19511 you mentioned

> Did you have a look at the derived values I had already created

I checked and fixed 2 of them `nexp`  and `iexp`

> The other question is whether we should rescue some of the old parts (much reduced) and slam them into tests?

you mean verification on computed values with the old machar computed ones?

numpy/_core/include/numpy/dtype_api.h
seberg: Might as well still change this I guess.

numpy/_core/include/numpy/dtype_api.h
```diff
@@ -479,6 +480,42 @@ typedef PyArray_Descr *(PyArrayDTypeMeta_EnsureCanonical)(PyArray_Descr *dtype);
...
+#define NPY_CONSTANT_maximum_finite 4
+#define NPY_CONSTANT_minimum_finite 5
+#define NPY_CONSTANT_inf 6
+#define NPY_CONSTANT_ninf 7
+#define NPY_CONSTANT_nan 8
+#define NPY_CONSTANT_finfo_radix 9
+#define NPY_CONSTANT_finfo_eps 10
+#define NPY_CONSTANT_finfo_smallest_normal 11
+#define NPY_CONSTANT_finfo_smallest_subnormal 12
+/* Constants that are always of integer type, value is `npy_intp/Py_ssize_t` */
```
seberg, line 500: Since you changed radix to be integral, can you re-order it to be here maybe?

Actually, thinking about it, I think we should tag the integer ones to make it clearer that they are different, i.e. make them all `(1<<16 + 0,1,...,N)`.

If we ever add a Python side "get constant" API, it would be necessary to be able to know the return type with a simple check.

numpy/_core/src/multiarray/multiarraymodule.c
seberg: I still feel like just deleting this for simplicity.  I.e. just make the integer be left as untouched in the array (making it slightly too large).  
But doesn't matter, probably a matter of opinion.

SwayamInSync:
I think after this refactor constraints like these on `dtype.kind` can be modified to check whether dtype provides.eps or not instead of kind checking

https://github.com/numpy/numpy/blob/17d5deda759e3a3ce79d88966ce69fdb323f623b/numpy/lib/_function_base_impl.py#L3863

Edit: On a second thought, its fine

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -4327,6 +4327,110 @@ normalize_axis_index(PyObject *NPY_UNUSED(self),
...
+
+    PyArrayObject *buffer_array = NULL;
+    char *buffer_data = NULL;
+    npy_intp dims[1] = {n_float_constants};
+
+    Py_INCREF(descr);
+    buffer_array = (PyArrayObject *)PyArray_NewFromDescr(&PyArray_Type,
+            descr, 1, dims, NULL, NULL, 0, NULL);
+    if (buffer_array == NULL) {
+        Py_DECREF(descr);
```
seberg:
```suggestion
```
A function that steals should also steal on error.

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -4327,6 +4327,110 @@ normalize_axis_index(PyObject *NPY_UNUSED(self),
...
+            goto fail;
+        }
+        int res = PyObject_SetAttrString(finfo, finfo_constants[i].name, value_obj);
+        Py_DECREF(value_obj);
+        if (res < 0) {
+            goto fail;
+        }
+    }
+
+  Py_DECREF(buffer_array);
```
seberg:
```suggestion
    Py_DECREF(buffer_array);
```

numpy/_core/tests/test_getlimits.py
```diff
@@ -139,53 +138,53 @@ def test_instances():
...
+
+    # Test float64
+    f64 = finfo(np.float64)
+    assert_equal(f64.bits, 64)
+    assert_equal(f64.precision, 15)  # int(-log10(2**-52)) = 15
+    assert_equal(f64.nmant, 53)
+    assert_equal(f64.minexp, -1021)
+    assert_equal(f64.maxexp, 1024)
+
+    # Test longdouble (values vary by platform, just check they exist)
```
seberg: Happy to skip it.  It could be cool to add the old bit pattern as an if/else (i.e. we had 4 hard-coded ones I think, double, IEEE 80bit, IEEE 128bit, and double-double).  And if neither matches, I guess we would do nothing.

numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -4318,10 +4320,193 @@ PyArray_DescrFromType(int type)
...
+            val = @RADIX@;
+            break;
+        case NPY_CONSTANT_finfo_eps:
+            val = @ABB@_EPSILON;
+            break;
+        case NPY_CONSTANT_finfo_smallest_normal:
+            val = @ABB@_MIN;
+            break;
+        case NPY_CONSTANT_finfo_smallest_subnormal:
+            val = @ABB@_TRUE_MIN;
```
seberg, line 4471: This is one thing I am slightly worried about, although it may well be impossible/implausible for these types.

In principle a float type might not define subnormals at all, making smallest-subnormal == smallest-normal and the right thing would be to refuse return nothing (i.e. consider it as "undefined").

numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -4318,10 +4320,193 @@ PyArray_DescrFromType(int type)
...
+/*
+ * On PPC64 systems with IBM double-double format, LDBL_EPSILON from the C
+ * standard library is incorrect. IBM double-double is a pair of IEEE binary64
+ * values (not a true IEEE quad), giving ~106 bits of mantissa precision (53+53).
+ * The correct epsilon is 2^-105, but glibc incorrectly returns 2^-1074 (DBL_TRUE_MIN).
+ */
+#if defined(HAVE_LDOUBLE_IBM_DOUBLE_DOUBLE_LE) || \
+    defined(HAVE_LDOUBLE_IBM_DOUBLE_DOUBLE_BE)
+    #undef LDBL_EPSILON
+    #define LDBL_EPSILON 2.46519032881566189191e-32L  /* 2^-105 */
```
seberg:
```suggestion
    #define LDBL_EPSILON 0x1p-105L  /* 2^-105 */
```
I think you can also write it as a hex-float, which is maybe a bit less clumsy for a clear power of 2.

numpy/_core/getlimits.py
```diff
@@ -548,75 +225,105 @@ def __new__(cls, dtype):
...
-    def smallest_normal(self):
-        """Return the value for the smallest normal.
+        # Use precision+1 digits in exponential notation
+        fmt_str = _MACHAR_PARAMS.get(self.dtype.type, {}).get('fmt', '%s')
+        if fmt_str != '%s' and hasattr(self, 'max') and hasattr(self, 'min'):
+            max_str = (fmt_str % self.max).strip()
+            min_str = (fmt_str % self.min).strip()
+        else:
+            max_str = str(self.max)
+            min_str = str(self.min)
```
seberg, line 318: If we wouldn't have min/max, then this would fail... maybe we should just assume min/max always exists (how could it not)?

numpy/_core/getlimits.py
```diff
@@ -548,75 +225,105 @@ def __new__(cls, dtype):
...
-        d = self.__dict__.copy()
-        d['klass'] = c
-        return (("%(klass)s(resolution=%(resolution)s, min=-%(_str_max)s,"
-                 " max=%(_str_max)s, dtype=%(dtype)s)") % d)
 
-    @property
-    def smallest_normal(self):
-        """Return the value for the smallest normal.
+        # Use precision+1 digits in exponential notation
+        fmt_str = _MACHAR_PARAMS.get(self.dtype.type, {}).get('fmt', '%s')
```
seberg, line 312: Hmmm, would hvae hoped we can always use normal string formatting now (and maybe fix tests as needed)... but happy to consider that a follow-up that is OK if it never happens!

numpy/_core/tests/test_getlimits.py
```diff
@@ -139,53 +138,53 @@ def test_instances():
...
-            # Double double
-            ld_ma.smallest_subnormal
-            # This test may fail on some platforms
-            assert len(w) == 0
+        # Test for common float types
+        for dtype in [np.float16, np.float32, np.float64]:
+            f = finfo(dtype)
+            _ = f.smallest_subnormal
+        # Also test longdouble
+        with np.errstate(all='ignore'):
```
seberg, line 150: Curious if this can still happen, but let's not worry about it.

numpy/_core/getlimits.py
```diff
@@ -548,75 +225,105 @@ def __new__(cls, dtype):
...
+    def negep(self):
+        return frexp(self.epsneg)[1]
+
+    @cached_property
+    def nexp(self):
+        # considering all ones (inf/nan) and all zeros (subnormal/zero)
+        return math.ceil(math.log2(self.maxexp - self.minexp + 2))
+
+    @cached_property
+    def iexp(self):
```
seberg, line 256: I am a bit worried about not having a test for especially these calculated values (or do we have that?)

seberg (commented):
Thanks @SwayamInSync I am happy, and it's very nice to get rid of this code!

The main thing that truly is missing, which I am OK to make a follow up, though is that this is new public API that needs to be documented!

There is one other thing that I had not thought of.  We have integer constants mixed with "normal" constants.  I think that is OK, but I also think we should make it easy to distinguish, which means using a bit-pattern to mark the "integer" ones.

Otherwise, a few more tests would be nice (but I am really not sure what tests we already have), so I think I am OK as it is -- but I should probably check once more, although maybe even after merging.

SwayamInSync:
Thanks @seberg I'll be happy to take the iinfo and other generic dtypes related refactor in the respective follow ups as well. And within in this PR I was also thinking to document + release note the `finfo` related refactors we did. Along with refactoring the `machar` test file (which I deleted earlier) to ensure the correctness of currently derived properties.

numpy/_core/getlimits.py
SwayamInSync: I am not 100% sure but from what I've seen there are tests that used the values (in some ways) but it is not strict checking. Its possible to have cases where even the approximate value can work

numpy/_core/include/numpy/dtype_api.h
SwayamInSync: Got it and also radix I switched it back
```cpp
 #RADIX = 16384, 2, 2, 2# // (get casted properly)
 ```

**Comment by SwayamInSync** [comment]
@seberg I need an opinion here

From macros FLT_MANT_DIG=24 includes the implicit leading bit for precision (IEEE specifies 23 explicit fraction bits but effective 24-bit significand); 
FLT_MIN_EXP=-125 reflects C's radix-agnostic definition where the exponent is offset by +1 compared to IEEE's unbiased min of -126 for normalized numbers; similarly, FLT_MAX_EXP=128 offsets IEEE's max unbiased exponent of +127, marking the overflow threshold. 
With MAChar these values were hardcoded so somewhere the values are as per IEEE (nmant, minexp) and somewhere as per C macro (maxexp) 

So do you prefer the complete original behaviour or we should take the whatever value comes from the header as finfo and adjust the derived values correct with proper documentations around this?

**Comment by SwayamInSync** [comment]
Also orignally machar used values make more sense atleast for `nmant` and sticking to that won't break any user's code, in case if they were doing some `nmant` or `minexp` dependent calculations and suddenly after refactor upgrade they are seeing float32 showing 24 instead of 23 mantissa bits

**Comment by seberg** [comment]
> So do you prefer the complete original behaviour or we should take the whatever value comes from the header as finfo and adjust the derived values correct with proper documentations around this?

Ufff, that is annoying that these definitions are so subtly different!  For the Python side, we clearly can't change the values.

For `nmant`, I could see calling it `mant_dig` instead and using the C definition on the C-side.  However for `min_exp`/`max_exp` that sleigh of hands doesn't work that well, unless we name it `_c_max_exp`, but...
So think I lean towards using the identical `finfo_...` naming and documenting that these do *not* match the C definition unfortunately (at least for the new C docs).

(I would like to still tag the integer constants, but if you like I can push that also.)

**Review by SwayamInSync — COMMENTED** [review]
These changes ensure the values remains as original and we can document them in new finfo docs on what C macros returns and where we tweak it for  backwards compatibility.

This way if user will be aware of the behaviour and if needs the true C macro value then he can just add +1 to the respective fields to get them.

File: numpy/_core/src/multiarray/arraytypes.c.src [hunk 25]
```diff
@@ -4318,10 +4320,202 @@ PyArray_DescrFromType(int type)
...
+            val = @ABB@_TRUE_MIN;
+            break;
+        case NPY_CONSTANT_finfo_nmant:
+            *(npy_intp *)ptr = @ABB@_MANT_DIG - 1;
+            return 1;
+        case NPY_CONSTANT_finfo_min_exp:
+            /*
+            Definition: Minimum negative integer such that FLT_RADIX raised by power one less than that integer is a normalized float, double and long double respectively
+
+            refernce: https://en.cppreference.com/w/c/types/limits.html
```
SwayamInSync, line 4480: I left these comments for developers so that they can reference why we are subtracting 1

numpy/_core/tests/test_finfo.py
```diff
@@ -0,0 +1,82 @@
+import pytest
+import numpy as np
+from numpy._core import numerictypes as ntypes
+from numpy import exp2, log10
+
+
+class MachArLike:
```
SwayamInSync, line 8: This file mimics the old machar behaviour to test

SwayamInSync (commented):
These changes ensure the values remains as original and we can document them in new finfo docs on what C macros returns and where we tweak it for backwards compatibility.

This way if user will be aware of the behaviour and if needs the true C macro value then he can just add +1 to the respective fields to get them.

SwayamInSync:
@seberg if this look good enough then I can proceed with the documentation

numpy/_core/tests/test_finfo.py
SwayamInSync: kept in separate file since machar is not going to be around so not mixing this with other tests

SwayamInSync:
Hmm... test failures aren't looking to be coming from finfo refactor

numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -4318,10 +4320,202 @@ PyArray_DescrFromType(int type)
...
+#define HALF_MAX_EXP 16
+#define HALF_MIN_EXP -13
+#define HALF_MANT_DIG 11    /* 10 + 1 (implicit) */
+#define HALF_DIG 3
+
+/*
+ * On PPC64 systems with IBM double-double format, LDBL_EPSILON from the C
+ * standard library is incorrect. IBM double-double is a pair of IEEE binary64
+ * values (not a true IEEE quad), giving ~106 bits of mantissa precision (53+53).
+ * The correct epsilon is 2^-105, but glibc incorrectly returns 2^-1074 (DBL_TRUE_MIN).
```
WarrenWeckesser: Is there a semantic definition of `epsilon` that can used as a property to be tested that justifies this choice?  My understanding of `epsilon` is that it is the smallest positive value such that `1 + epsilon` is not `1`.  This is the "mainstream definition" of [*machine epsilon* according to wikipedia](https://en.wikipedia.org/wiki/Machine_epsilon).  If that is the property that should be satisfied by `epsilon`, then `2^-1074` (i.e. 5e-324) is the correct value.

If it is desirable to avoid subnormal numbers in a double-double representation, then `epsilon` would be `2^-1022` (i.e `np.finfo(np.float64).tiny = 2.2250738585072014e-308`).

I guess the justification for `2^-105` is by analogy to regular double precision, where `epsilon` is `2^-52` (i.e. use `2^(-nbits+1)`). But is that a useful definition?

seberg: The value is not modified in this PR, so I suggest moving the discussion to an issue, because I think we should put in the PR either way.

The point is what `eps` is useful for, I guess?  Yes, you could define it like that, but this isn't IEEE so it's tricky.
The point is, if this *was* an IEEE number and you ignored any number that is not representable by a comparable IEEE, then -105 is the correct value.

For IEEE floats eps is the spacing in the range of 1.0 to 2.0 (exclusive) but the guaranteed spacing is `2^-105` in this range and not any smaller value.

I may be willing to bet that in most numerical situations the latter definition is the meaningful one, but I am not sure how actually settle the bet :).

SwayamInSync: @WarrenWeckesser I read the same article when writing.
In the section **Mainstream definition (Interval machine epsilon)** provides a definition that is also being used by the ISO C Standard
>the difference between 1.0 and the next **representable** floating-point number larger than 1.0

and formula they provide is `b^{-(p-1)}` here b is 2 (radix) and `p` is the total number of significant digits (in base b) that can be stored in the significand.
So for double-double format that value is 106 (comes from combining the significands of two doubles) which leads to the epsilon as ~2.4651903e-32

SwayamInSync: Just found this, IBM AIX (also uses double-double format) here in page 60, **Table 8. Long double constants for Power systems** They are also seem to be using `LDBL_EPSILON = 0.24651903288156618919116517665087070E-31` ~ 2^-105
https://www.redbooks.ibm.com/redbooks/pdfs/sg246034.pdf

WarrenWeckesser: > The value is not modified in this PR, so I suggest moving the discussion to an issue, ...

Ah, sorry, my fault for making a drive-by comment without looking at the full scope of the change.

>> the difference between 1.0 and the next representable floating-point number larger than 1.0
>
> and formula they provide is b^{-(p-1)} 

That formula would apply to a standard IEEE float. But with double-double, the values are stored as two double precision floats, so 1.0 is (1.0, 0.0) internally.  The *next representable value* after (1.0, 0.0) is (1.0, 5e-324).

But as @seberg suggested, further discussion should be deferred to a separate issue.  Sorry for the noise.

seberg: I opened an issue, but maybe we can close it quickly again, too. The only value being changed in this PR was defining the smallest normal (previously undefined and before that just wrong).

What is true is that the comment is wrong in that we effectively opt for a different definition of epsilon rather than the header provided being wrong. Ours is a spacing, or the smallest _normal_ float above 1 one or so.

SwayamInSync: Right, I'll update the comment here

WarrenWeckesser: I left further comments over in https://github.com/numpy/numpy/issues/29871.  I understand better why `2**-105` is reasonable, so I'm resolving this comment.

seberg:
@SwayamInSync decided to just put it in.  I feel this should be reasonable both from a general API perspective and from the changes to the `finfo`.

(Although, I will expect that there may be some follow-ups with finfo, it's always trickier than you expect.)

I think it's just as well to include release notes that you think are helpful in a seperate PR to add docs (and at that point -- or earlier -- others will also get another chance to review the API choices here).

SwayamInSync:
Thanks @seberg , will do the doc PR asap

numpy/_core/include/numpy/dtype_api.h
```diff
@@ -389,7 +390,7 @@ typedef int (PyArrayMethod_PromoterFunction)(PyObject *ufunc,
 
 // used to separate dtype slots from arrfuncs slots
 // intended only for internal use but defined here for clarity
-#define _NPY_DT_ARRFUNCS_OFFSET (1 << 10)
+#define _NPY_DT_ARRFUNCS_OFFSET (1 << 11)
```
seberg, line 393: This was a pretty bad oops, I really don't know how we both missed it :(.  I.e. this is an ABI break.  We could try to work-around it in `npy_2_compat.h`, but for now let me post a weird work-around:
```
#if NPY_TARGET_VERSION >= 0x15  // NUMPY_2_4_API_VERSION
#define ARRFUNCS_OFFSET_FIX(v) (v)
#else
#define ARRFUNCS_OFFSET_FIX(v) \
  (v) - (NPY_DT_PyArray_ArrFuncs_getitem) + 1 + (((PyArray_RUNTIME_VERSION >= 0x15) ? (1 << 11) : (1 << 10)))
#endif
```
And then use: `ARRFUNCS_OFFSET_FIX(NPY_DT_PyArray_ArrFuncs_*)`.

The above "grows out" by itself after a few years when only NumPy 2.4+ is relevant anymore.

SwayamInSync: Ahh this messed up :( 
I am sorry for pushing this bad bad line.

This `npy_2_compat.h` fix will also require documenting this macro, I was thinking if somehow we can handle this inside NumPy at runtime? Like in `dtypemeta.c` we can check
```c
#define _NPY_DT_ARRFUNCS_OFFSET_OLD (1 << 10)
if (slot > _NPY_DT_ARRFUNCS_OFFSET_OLD &&
        slot <= NPY_NUM_DTYPE_PYARRAY_ARRFUNCS_SLOTS
                + _NPY_DT_ARRFUNCS_OFFSET_OLD) {
    slot = (slot - _NPY_DT_ARRFUNCS_OFFSET_OLD)
            + _NPY_DT_ARRFUNCS_OFFSET;
}
#undef _NPY_DT_ARRFUNCS_OFFSET_OLD
```

So this way maybe NumPy's slot processing code remaps old-offset slots `(1024+N)` to new-offset slots `(2048+N)`. I think this will also have zero burden on downstream extensions compiled against NumPy 2.0–2.3 just work on 2.4+ without recompiling or changing code.

numpy/_core/src/multiarray/dtypemeta.c
SwayamInSync, line 219: Precisely here

```suggestion
        }
        
#define _NPY_DT_ARRFUNCS_OFFSET_OLD (1 << 10)
if (slot > _NPY_DT_ARRFUNCS_OFFSET_OLD &&
        slot <= NPY_NUM_DTYPE_PYARRAY_ARRFUNCS_SLOTS
                + _NPY_DT_ARRFUNCS_OFFSET_OLD) {
    slot = (slot - _NPY_DT_ARRFUNCS_OFFSET_OLD)
            + _NPY_DT_ARRFUNCS_OFFSET;
}
#undef _NPY_DT_ARRFUNCS_OFFSET_OLD
```

numpy/_core/include/numpy/dtype_api.h
seberg: We can't do that in NumPy, because it needs to work both ways.  But we can of course explicitly re-define those macros to do the right thing at runtime (without user code changes) in `npy_2_compat.h`.
The above is just a quick downstream-side suggested fix.

I am trying to figure out a way that `ml_dtypes` can start using the new API without having a hard break (beyond maybe slightly unfortunate printing of the dtype).

The pattern here is to:
* Introduce a new `NPY_DT_legacy_proto` slot basically to pass in the _existing_ registration struct (which includes type code, kind code, elsize, etc.) -- this is *strictly* for non-parametric dtypes of course!
* And the trick: Backport this to old versions of NumPy with a slightly scary dance that works out (it isn't too bad, although there is one crazy little trick that luckily happens to work out).

I.e. "backport" means: If you compile with 2.4 (or copy paste our headers), the new pattern will work all the way back to NumPy 2.0, but only do crazy things on versions prior to NumPy 2.5.
Having something like NumPy 2.5+ isn't problematic for new DTypes I think, but for `ml_dtypes` it creates a major upgrading conundrum...

(If we can agree that forcing to copy the header or compile with new versions of NumPy in a year or so, we can e.g. start to force downstream to use the new casting implementation paths for real. -- Which is one of the biggest things push.
I would put a note on `NPY_DT_legacy_proto` that this is likely to morph over time or go away, but it doesn't do all that much besides assigning type number, kind, and char and providing defaults with the assumption that the only possible parameter is the byte-order.)

As part of that, I would just fix this as well.  Not a whole lot of harm done, just cryptic if anyone would try.

seberg: Oooops, but right, we shouldn't fix the macro definitions, as that would break static use (not hard to deal with, but you have to deal with it)...

However, if in the above I redefine `PyArrayDTypeMeta_InitFromSpec` (which we _can_ do), then this function can indeed apply the fix.  A bit brutal, as I guess we might actually mutate downstream static storage, but it will work (and the fixing is idempotent so mutation shouldn't be too scary in practice, I think).

numpy/_core/src/multiarray/dtypemeta.c
seberg: I added a compat shim in gh-31302. That said it does also need this code fix for maximum compatibility (although, since nobody noticed yet...).
In theory the best compatibility would be if we backport this sanitization to 2.4 and _also_ change it back to `1 << 10`  (I.e. sanitize `1 << 11` to `1 << 10`).

Then there is a 2.4.x release that is compatible with everything else and maybe I can just skip the gh-31302 dance.

(I think if we can try to revert to `1 << 10` for slightly more compat, but then nobody noticed yet...  Or we do the gh-31302 dance -- with additional in-code fix -- and just stick with `1 << 11` on grounds that there is a work-around and honestly, nobody even noticed even though 2.4 is out for a long time.)

PR #29841, opened 2025-09-30
DEP: Raise `TypeError` on attempt to convert array with `ndim > 0` to…

hpkfft:
This PR removes deprecated functionality, as requested in #29835

### DIFF
--- a/numpy/_core/tests/test_multiarray.py
+++ b/numpy/_core/tests/test_multiarray.py
@@ -3886,20 +3886,9 @@ class TestMethods:
                   '?', 'O']
         for dt in dtypes:
             a = np.array(7, dtype=dt)
-            b = np.array([7], dtype=dt)
-            c = np.array([[[[[7]]]]], dtype=dt)
-
             msg = f'dtype: {dt}'
             ap = complex(a)
             assert_equal(ap, a, msg)
-
-            with pytest.warns(DeprecationWarning):
-                bp = complex(b)
-            assert_equal(bp, b, msg)
-
-            with pytest.warns(DeprecationWarning):
-                cp = complex(c)
-            assert_equal(cp, c, msg)
 
     def test__complex__should_not_work(self):
         dtypes = ['i1', 'i2', 'i4', 'i8',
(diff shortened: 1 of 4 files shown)


numpy/_core/tests/test_multiarray.py
```diff
@@ -9260,9 +9250,24 @@ def __int__(self):
...
+            assert_raises(TypeError, float_func, np.array([2]))
+            assert_raises(TypeError, float_func, np.array([3.14]))
+            assert_raises(TypeError, float_func, np.array([[4.0]]))
+
+            assert_equal(5.0, float_func(np.array('5')))
+            assert_equal(5.1, float_func(np.array('5.1')))
+            assert_equal(6.0, float_func(np.bytes_(b'6')))
+            assert_equal(6.1, float_func(np.bytes_(b'6.1')))
+            assert_equal(7.0, float_func(np.str_('7')))
+            assert_equal(7.1, float_func(np.str_('7.1')))
```
seberg, line 9270: New tests?  I don't say no to new tests ever :).  (I think `float` and `__float__` is a bit double, since `float` just calls `__float__`, but OK.)

seberg (approved):
Nice, thanks a lot @hpkfft, I think Matti agreed it's about time, so I think we can give this a shot.

If someone finds this PR and has a larger fallout, please let us know, in theory we could e.g. escalate to a `VisibleDeprecationWarning` (but in practice, I think users also tend to ignore warnings either way).

NumPy 1.25 was 2.5 years ago (by the time this is released), which seems OK for this.
(We are not good about timely expiring, but this was always a noisy one, so I think it had to be well above 1 year of deprecation.)

seberg:
Windows FP16 seems to be very broken on some hardware or so, restarted CI, but whatever is going on, it is unrelated to this PR anyway.

numpy/_core/tests/test_multiarray.py
hpkfft: Thanks, I thought it best to be careful.

Since Python 3.8, [float()](https://docs.python.org/3/library/functions.html#float) will fall back to `__index__()`:
> For a general Python object `x`, `float(x)` delegates to `x.__float__()`. If `__float__()` is not defined then it falls back to `__index__()`.

eric-wieser:
In case anyone is curious of the history of the original deprecation, it can be found at https://github.com/numpy/numpy/pull/10615

PR #29846, opened 2025-09-30
TYP: Fix signatures of ``linalg.matmul`` and ``linalg.outer``

jorenham:
This fixes several issues in `np.linalg.outer` (positional-only params, a misplaced `out=...` param, and an orphanced `-> _ArrayT`), and also fixes `linalg.matmul` parameters to be positional-only. 

I also snuck in an an additional overload that improves the return type if both input array-likes have the same dtype.


numpy/linalg/_linalg.pyi
```diff
@@ -78,7 +78,7 @@ __all__ = [
     "vecdot",
 ]
 
-_ArrayT = TypeVar("_ArrayT", bound=NDArray[Any])
```
jorenham, line 81: This one was unused

charris:
Thanks Joren.

PR #29847, opened 2025-09-30
TYP: Fix incompatible defaults in ``polyfit``, ``histogram``, and ``histogramdd``

jorenham:
The default values weren't assignable to the types of the parameters.


charris:
Thanks Joren.

PR #29848, opened 2025-10-01
MAINT, TYP: bump ``mypy`` to ``1.18.2``

jorenham:
https://mypy.readthedocs.io/en/stable/changelog.html#mypy-1-18-2


charris:
Thanks Joren.

Issue #28095, opened 2025-01-03
BUG: F2py no longer properly handles nested parentheses in variable definitions


eia-akahan:
### Describe the issue:

Up to numpy 1.26.2, F2py successfully handled variables with nested defintions.  In 1.26.4, it does not.

This variable definition
```
INTEGER MX_INTL_CURVES
INTEGER MX_SUPPLY_CURVES
INTEGER MX_UNITS
PARAMETER(MX_INTL_CURVES = 12)
PARAMETER(MX_SUPPLY_CURVES = 14)
PARAMETER(MX_UNITS = 1800)
REAL*4       CMM_CL_BTUS((MX_SUPPLY_CURVES + MX_INTL_CURVES),MX_UNITS)  

```
produce this entry in the pyf for cmm_cl_btus in numpy 1.26.2

`real*4 dimension(26,1800) :: cmm_cl_btus`

starting in 1.26.4, we instead get the following entry in the pyf.

`real*4 dimension(mx_intl_curves + mx_supply_curves,1800) :: cmm_cl_btus`

I looked through, and it appears that in crackfortran,  the attrspec gets set improperly - During analyzecommmon(), the watch window shows that the block['vars'][n]['attrspec'] is showing this - ['dimension((mx_supply_curves+mx_intl_curves),mx_units)'].

I think it has trouble removing the second set of parentheses when it does the parameter lookup.

I'm not sure which change in crackfortran that would have caused the change in behavior to nested parentheses between 1.26.2 and 1.26.4.  

### Reproduce the code example:

```python
import numpy   
numpy.f2py.run_main(['-h',
                         'vartest.pyf',
                        'vars.f90',
                        '-m', 'vartest',
                        # 'only:', 'init_filer', 'get_data',
                        '--overwrite-signature'])

Contents of vars.f90

module util
INTEGER MX_INTL_CURVES
INTEGER MX_SUPPLY_CURVES
INTEGER MX_UNITS
PARAMETER(MX_INTL_CURVES = 12)
PARAMETER(MX_SUPPLY_CURVES = 14)
PARAMETER(MX_UNITS = 1800)
REAL*4 CMM_CL_BTUS((MX_SUPPLY_CURVES + MX_INTL_CURVES),MX_UNITS) 
contains
 subroutine test
 write(*,*) 'Hello'
 end subroutine test
 end module
```

### Error message:

```shell
!    -*- f90 -*-
! Note: the context of this file is case sensitive.

python module vartest ! in 
[...]

PR #29857, opened 2025-10-01
BUG: param_parse couldn't handle nested parentheses

eia-akahan:
BUG:  param_parse couldn't handle nested parentheses - see https://github.com/numpy/numpy/issues/28095.  

This adds a function that determines the appropriate order of nested parentheses and  properly assigns the dimensions based on the order of operations. It solves the case outlined in https://github.com/numpy/numpy/issues/28095, so that all parameters are mapped properly and assigned values.

BUG: param_eval expects v to be a string, but get_parameters sets v to integers in 2 cases.

```
            if real16pattern.search(v):
                v = 8
            elif real8pattern.search(v):
                v = 4
```

by enclosing those values in single quotes, now v is a string, which removes the error message: "param_eval: got eval() arg 1 must be a string, bytes or code object on 4." which is triggered when those two if statements are activated.

This is my first pull request for numpy - I'm not sure if I need to create test cases.

### DIFF
--- a/numpy/f2py/crackfortran.py
+++ b/numpy/f2py/crackfortran.py
@@ -3082,12 +3082,12 @@ def param_parse(d, params):
     >>> param_parse(d, params)
     3
     """
-    if "(" in d:
-        # this dimension expression is an array
-        dname = d[:d.find("(")]
+    dname = d[:d.find("(")] if "(" in d else ""
+    if dname:
+        # Text precedes the first "(", so this is array-parameter
+        # indexing such as pa(1) or myparamarray(nested(dim)).
         ddims = d[d.find("(") + 1:d.rfind(")")]
-        # this dimension expression is also a parameter;
-        # parse it recursively
+        # the index is itself an expression over params; parse it recursively
         index = int(param_parse(ddims, params))
         return str(params[dname][index])
     elif d in params:
--- a/numpy/f2py/tests/src/crackfortran/gh28095.f90
+++ b/numpy/f2py/tests/src/crackfortran/gh28095.f90
@@ -1,0 +1,13 @@
+module util
+integer mx_intl_curves
+integer mx_supply_curves
+integer mx_units
+parameter(mx_intl_curves = 12)
+parameter(mx_supply_curves = 14)
+parameter(mx_units = 1800)
+real*4 cmm_cl_btus((mx_supply_curves + mx_intl_curves), mx_units)
+contains
+ subroutine test
+ write(*,*) 'Hello'
+ end subroutine test
+end module util
--- a/numpy/f2py/tests/test_crackfortran.py
+++ b/numpy/f2py/tests/test_crackfortran.py
@@ -417,6 +417,42 @@ class TestParamEval:
         pytest.raises(ValueError, crackfortran.param_eval, v, g_params, params,
                       dimspec=dimspec)
 
+
+class TestParamParseNestedParens:
+    # issue gh-28095: grouping parens in a dimension expression such as
+    # (mx_supply_curves + mx_intl_curves) must fold to a concrete size,
+    # not be mistaken for pa(index) array-parameter indexing.
+    def test_grouping_parens_fold(self):
+        params = {"mx_supply_curves": 14, "mx_intl_curves": 12}
+        out = crackfortran.param_parse(
+            "(mx_supply_curves + mx_intl_curves)", params)
+        assert out.replace(" ", "") == "(14+12)"
+
+    def test_grouping_parens_with_trailing_factor(self):
+        # A grouping paren need not span the whole token; the factor
+        # outside the parentheses must survive substitution.
+        params = {"n": 3, "m": 1800}
+        out = crackfortran.param_parse("(n + 1)*m", params)
+        assert out.replace(" ", "") == "(3+1)*1800"
+
+    def test_array_index_still_works(self):
+        params = {"pa": {1: 3, 2: 5}}
+        assert crackfortran.param_parse("pa(1)", params) == "3"
+        assert crackfortran.param_parse("pa(2)", params) == "5"
+
+    def test_nested_array_index(self):
+        params = {"dim": 2, "nested": {1: 1, 2: 2, 3: 3},
+                  "myparamarray": {1: 10, 2: 20, 3: 30}}
+        assert crackfortran.param_parse(
+            "myparamarray(nested(dim))", params) == "20"
+
+    def test_dimension_substituted_in_module(self):
+        fpath = util.getpath("tests", "src", "crackfortran", "gh28095.f90")
+        mod = crackfortran.crackfortran([str(fpath)])
+        vs = mod[0]["vars"]["cmm_cl_btus"]
+        assert vs["dimension"] == ["26", "1800"]
+
+
 @pytest.mark.slow
 class TestLowerF2PYDirective(util.F2PyTest):
     sources = [util.getpath("tests", "src", "crackfortran", "gh27697.f90")]


numpy/f2py/crackfortran.py
```diff
@@ -3022,6 +3022,52 @@ def param_eval(v, g_params, params, dimspec=None):
...
+        variable_values: A dictionary where keys are variable names and values are their values.
+
+    Returns:
+        A new string with variables replaced by their values.
+        Raises a ValueError if a variable in the equation is not found in variable_values (which is the parameter dictionary).
+    """
+    string_equation=re.sub(r'\b[a-zA-Z_][a-zA-Z0-9_]*\b',
+                  lambda match: str(params[match.group(0)])
+                                if match.group(0) in params
+                                else _raise_value_error(match.group(0)),
```
seberg: I would just make inline `def ...` or move all into the helper. Now we have a helper but just to raise an error.

numpy/f2py/crackfortran.py
```diff
@@ -3082,12 +3128,9 @@ def param_parse(d, params):
     """
     if "(" in d:
         # this dimension expression is an array
-        dname = d[:d.find("(")]
-        ddims = d[d.find("(") + 1:d.rfind(")")]
```
seberg, line 3088: Do you know what you are getting here, i.e. what `d` is/can be?  Since the code here works, I would expect that this is pretty well formatted and the balancing might not be accounting for much here?

I think I am getting at the parentheses balancing seems a bit odd and I would like to know if there is even a reason for it.

(Also nit, but `replace_variables_in_equation` isn't recursive :).)

seberg (commented):
Some comments to get things started, I hope someone more famliar with f2py will chime in.
About adding a test: Yes please do unless it is incredibly painful, but I suspect it is easy (maybe even just modifying/extending an existing test a bit).

numpy/f2py/crackfortran.py
eia-akahan: Thanks for the questions - all good ones.

**Question 1.  Do you know what you are getting here, i.e. what d is/can be?**

That deleted code is relatively new - it was introduced in 1.26.4. https://github.com/numpy/numpy/commit/01a7265b03e52902b7daf56bd7a1056c2e5e0e43

It creates some problems - the dimensions of a fortran variable definition are fed in as d.  

Lets take an example where it works
```
INTEGER MX_UNITS
PARAMETER(MX_UNITS = 1800)
INTEGER MX_INTL_CURVES
PARAMETER(MX_SUPPLY_CURVES = 14)
INTEGER MX_INTL_CURVES
PARAMETER(MX_INTL_CURVES = 12)
REAL*4       CMM_CL_BTUS(MX_INTL_CURVES,MX_UNITS)`
```

For the final line, d will be   

`REAL*4       CMM_CL_BTUS(MX_INTL_CURVES,MX_UNITS)`

This code:
image

produces ddims to pass into the parameter lookup, and will sequentially pass in MX_INTL_CURVES, followed by MX_UNITS.

But that code can't handle when there are more than 1 set of parantheses.  Say the final line was 

```
REAL*4       CMM_CL_BTUS((MX_SUPPLY_CURVES + MX_INTL_CURVES),MX_UNITS)  
```

That'll pass in (MX_SUPPLY_CURVES + MX_INTL_CURVES) to the parameter lookup, which can't find a match in the parameter table because it still contains the parentheses. 

The new regular expression isolates variables in an equation string, looks up their values in a dictionary,
    and replaces them in the string.  We then use eval to use the math, preserving all the parentheses.

```    string_equation=re.sub(r'\b[a-zA-Z_][a-zA-Z0-9_]*\b',
                  lambda match: str(params[match.group(0)])
                                if match.group(0) in params
                                else _raise_value_error(match.group(0)),

```

**Question 2.  I think I am getting at the parentheses balancing seems a bit odd and I would like to know if there is even a reason for it.**

In the flow here, 

```
REAL*4       CMM_CL_BTUS((MX_SUPPLY_CURVES + MX_INTL_CURVES),MX_UNITS)  
```
when sent for processing, you get a list [((MX_SUPPLY_CURVES + MX_INTL_CURVES),MX_UNITS)], and each piece is sent to eval.   eval can't handle 'MX_UNITS)'.  Balance parentheses adds as many opening or closing parentheses as needed to allow eval to work.


**Question 3. (Also nit, but replace_variables_in_equation isn't recursive :).)**

good point on the recursion.  I switched to the regular expression/eval approach after trying something recursive.  I will remove the comment.

seberg: The reason I ask was because I am wondering for what kind of input you need the parentheses balancing.
But more importantly, I can't really say what's what here.  I can see that this might be off, or require an `_eval_length` call, but even if that seems likely, I am not sure that the eval length call would happen here rather than later.

But, on a more general level: The tests are all failing here, ideally you try to run the f2py tests locally first.

Ping @HaoZeke in case he has a quick insight, but we really need tests passing.

eia-akahan: ok.  I'll take a look at the tests.  Thanks for the guidance.

eia-akahan: @seberg - ran the tests locally, got them to pass.  Windows tests pass on CI, but linux test fails. Doesn't appear to be directly related to the new code, so I'm not sure.  [numpy ci test log](https://dev.azure.com/numpy/numpy/_build/results?buildId=45247&view=logs&j=651fc570-a910-517b-e9d3-812d284ce99f&t=ea240f47-a7ac-5f02-a5a1-32050ed91ebb&l=1081
)

eia-akahan: Made some changes and the tests are passing now.  And the code now handles a Fortran parameter with an equation in nested parentheses.

HaoZeke (changes requested):
Thanks for working on this @eia-akahan. Sorry for the late response. The core bug (`param_parse` treats `(a + b)` as array indexing) is correctly identified.

However the current approach replaces working logic with a more complex engine that has some issues:

1. `balance_parentheses` should not be needed. The dimension pieces from `markoutercomma` are already balanced. Silently "fixing" unbalanced parens masks upstream bugs.

2. The `except TypeError: int_dim ` int(balanced_dim.split('(')[0])` fallback is fragile. For expressions starting with `(` this would do `int("")` and crash. It works for `myparamarray(nested(dim))` by coincidence.

3. The `{ in str(params[...])` heuristic for detecting array parameters couples to Python dict repr formatting.

A simpler fix: check `dname` (text before the first `(`). 
- When `dname` is empty the parens are grouping, not array indexing, so strip outer parens and recurse.
- When `dname` is non-empty, keep the existing array-index logic.

The `v = '8'` / `v = '4'` fix for `real16pattern` / `real8pattern` is correct and could go as a separate commit.

Here is a concrete alternative (happy to co-author / cleanup and finalize this):

```python
if "(" in d:
    dname ` d[:d.find("(")]
    if dname:
        # Array parameter indexing, e.g. pa(1)
        ddims ` d[d.find("(") + 1:d.rfind(")")]
        index ` int(param_parse(ddims, params))
        return str(params[dname][index])
    else:
        # Grouping parentheses, e.g. (a + b)
        d ` d[d.find("(") + 1:d.rfind(")")]
        return param_parse(d, params)
```

This preserves all existing behavior for array parameters and adds the missing case. The `v = '8'` fix should be a separate commit.

eia-akahan:
>Here is a concrete alternative (happy to co-author / cleanup and finalize this):

I appreciate the review (from both @seberg and @HaoZeke ), and the suggested alternative solution. Using this pull request, will remove previously proposed approach, swap with that, test it out, and update.  

>The v = '8' fix should be a separate commit.

Will create a second pull request with with the proposed solution and a test.

eia-akahan:
Updated with @HaoZeke's implementation, and removed the solution to the v string int/string issue. All tests passing, requesting review.

HaoZeke:
@eia-akahan sorry for the delay here, do you mind if I make a few changes to finish this up?

eia-akahan:
> @eia-akahan sorry for the delay here, do you mind if I make a few changes to finish this up?

@HaoZeke , That would be great.  Thanks for the review.

HaoZeke (approved):

PR #29858, opened 2025-10-02
TST: Make temporary file usage thread safe

bwhitt7:
This PR is related to #29552 and also a sort of continuation of #29816. With the newest version of `pytest-run-parallel`, most usages of `tmp_path` should be thread safe. However there are a few instances of temporary files being used in the test suite that do not use the pytest `tmp_path` fixtures and thus are still thread-unsafe. This PR attempts to fix these thread safety issues.

In `numpy/_core/tests/test_multiarray.py::TestIO`:
- This was a bit tricky to fix and make thread safe. The way `pytest-run-parallel` patches `tmp_path` is by creating directories for each thread. However this only works if the `tmp_path` fixture is called directly by a test. In this case, `tmp_path` is called by another fixture, `tmp_filename`. I made it so it's an explicit method called by each test instead. This of course introduces more code changes since the fixture was parameterized, so I had to include these parameters wherever `tmp_filename` was originally used.
- Another change was to `test_fromfile_bad_dup`. This mutates `os.dup` in a way that's extremely thread unsafe and leads to following tests giving all sorts of errors. This test could be marked as thread unsafe, but I wanted to try using `monkeypatch`. This is a bit more of a pytest-like way of handling this mutation, and will automatically mark the test are thread-unsafe.

In `numpy/lib/tests/test__datasource.py`:
- A bit more of a simple change, makes it so the `datasource.Datasource` instance is created in each test using `tmp_path`. Originally the teardown had it so this datasource object was deleted. I'm unsure if this is needed, but if it is, it would require including `del ds` in each test that declares the datasource object. I can add this if folks think it's necessary.

### DIFF
--- a/numpy/_core/tests/test_multiarray.py
+++ b/numpy/_core/tests/test_multiarray.py
@@ -5579,20 +5579,18 @@ class TestIO:
 
     @pytest.fixture(params=["string", "path_obj"])
     def param_filename(self, tmp_path, request):
-        # This fixture covers two cases:
-        # one where the filename is a string and
-        # another where it is a pathlib object
-        filename = tmp_path / "file"
-        if request.param == "string":
-            filename = str(filename)
-        yield filename
+        # This fixtures returns string or path_obj
+        # so that every test doesn't need to have the
+        # paramterize marker.
+        return request.param
 
-    def _threaded_filename(self, path, thread_index):
-        # Makes sure the param_filename fixture is thread safe
-        if isinstance(path, str):
-            return path + str(thread_index)
-        else:
-            return pathlib.Path(path).with_name("file" + str(thread_index))
+    def _threaded_filename(self, tmp_path, param):
+        # Handles two cases, where filename should
+        # be a string, or a path object.
+        path = tmp_path / "file"
+        if param == "string":
+            return str(path)
+        return path
 
     def test_nofile(self):
         # this should probably be supported as a file
@@ -5623,22 +5621,22 @@ class TestIO:
         d = np.fromstring("1,2", sep=",", dtype=np.int64, count=0)
         assert d.shape == (0,)
 
-    def test_empty_files_text(self, param_filename, thread_index):
-        tmp_filename = self._threaded_filename(param_filename, thread_index)
+    def test_empty_files_text(self, tmp_path, param_filename):
+        tmp_filename = self._threaded_filename(tmp_path, param_filename)
         with open(tmp_filename, 'w') as f:
             pass
         y = np.fromfile(tmp_filename)
         assert_(y.size == 0, "Array not empty")
 
-    def test_empty_files_binary(self, param_filename, thread_index):
-        tmp_filename = self._threaded_filename(param_filename, thread_index)
+    def test_empty_files_binary(self, tmp_path, param_filename):
+        tmp_filename = self._threaded_filename(tmp_path, param_filename)
         with open(tmp_filename, 'wb') as f:
             pass
         y = np.fromfile(tmp_filename, sep=" ")
         assert_(y.size == 0, "Array not empty")
 
-    def test_roundtrip_file(self, param_filename, thread_index):
-        tmp_filename = self._threaded_filename(param_filename, thread_index)
+    def test_roundtrip_file(self, tmp_path, param_filename):
+        tmp_filename = self._threaded_filename(tmp_path, param_filename)
         x = self._create_data()
         with open(tmp_filename, 'wb') as f:
             x.tofile(f)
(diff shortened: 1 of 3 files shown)


numpy/_core/tests/test_multiarray.py
```diff
@@ -5577,15 +5577,14 @@ def _create_data(self):
         x[0, :, 1] = [np.nan, np.inf, -np.inf, np.nan]
         return x
 
-    @pytest.fixture(params=["string", "path_obj"])
-    def tmp_filename(self, tmp_path, request):
+    def _create_filename(self, path, param):
         # This fixture covers two cases:
         # one where the filename is a string and
         # another where it is a pathlib object
-        filename = tmp_path / "file"
```
lysnikolaou, line 5599: Since the tests only depend on the file created, this could be much simpler if we used the same directory from all threads and then created one file for each thread, by appending a suffix with the thread id. The way I see things, this would be enough isolation for everything to be okay with less code.

lysnikolaou (commented):
Thanks @bwhitt7 for opening the PR. `DataSource` changes look good to me. I'd prefer if we went with the less code approach I'm recommending below for the test in `TestIO`, but I'll let other decide on the best approach.

numpy/_core/tests/test_multiarray.py
bwhitt7: @lysnikolaou Unfortunately `thread_index` suffers from the same issue as `tmp_path` where it only gets updated when directly requested by a test. We could potentially try fixing this in `pytest-run-parallel`, but there's also the issue that fixtures generally get called only once per-test and don't have a "thread" scope. Hmmmm if we could get the thread id without calling into fixtures that would get rid of a lot of code here.

lysnikolaou: You could probably use `threading.current_thread().name` and that should be good enough.

bwhitt7:
@lysnikolaou Okay I think I figured out a way to minimize code changes for the tests in `TestIO`. The original fixture is still being used, making it so I don't need to add all those`pytest.mark.parametrize` on every test. I then use a method to modify the path returned by the fixture and make it thread safe. Fixtures are initialized only once per test, shared between the threads, so calling a method in the test is one of the only ways I've found to make tests like this thread safe.

bwhitt7:
Whoops forgot to check if the tests were failing. Looks like the usage of `thread_index` is causing tests to fail since they can't import the fixture without `pytest-run-parallel` in the CI job haha. Might wait on working on this PR some more after the thread marker PR gets merged.

bwhitt7:
Nevermind I'll add it now. This latest push adds in checks for `pytest-run-parallel` in the main `conftest.py` file, and creates a dummy `thread_index` fixture when `pytest-run-parallel` doesn't import. I'm thinking of moving this to the `test_multiarray.py` file for better readability, but in case anyone wants to use thread_index in the future, it might be good to keep it in the main confest file.

numpy/_core/tests/test_multiarray.py
ngoldbaum: @lysnikolaou the issue is that fixtures are evaluated on the main thread. If I use your suggestion, the temporary files will always have `MainThread` in their name. The only way to make things local is to call a function, which unfortunately leads to the repetitive calls here.

If pytest-run-parallel had a way to register fixtures to be made thread-safe then we can do the right thing in pytest-run-parallel. Maybe that's a better way forward? But I don't want to block adding pytest-run-parallel CI over here in NumPy on new pytest-run-parallel features. Since @bwhitt7 is running out of time on their internship.

numpy/_core/tests/test_multiarray.py
```diff
@@ -5592,14 +5578,19 @@ def _create_data(self):
...
     @pytest.fixture(params=["string", "path_obj"])
-    def tmp_filename(self, tmp_path, request):
-        # This fixture covers two cases:
-        # one where the filename is a string and
-        # another where it is a pathlib object
-        filename = tmp_path / "file"
-        if request.param == "string":
-            filename = str(filename)
-        yield filename
+    def param_filename(self, tmp_path, request):
```
ngoldbaum:
```suggestion
    def param_filename(self, request):
```

This isn't used so I think you can delete it.

numpy/_core/tests/test_multiarray.py
```diff
@@ -5592,14 +5578,19 @@ def _create_data(self):
...
+        # This fixtures returns string or path_obj
+        # so that every test doesn't need to have the
+        # paramterize marker.
+        return request.param
+
+    def _threaded_filename(self, tmp_path, param):
```
ngoldbaum: Maybe call this function `normalize_filename`? I think the underscore is confusing (who cares about public/private on a test?). It also doesn't need to be a method on the class or a staticmethod. you can move it to module scope and make it a regular function.

ngoldbaum:
Since the other PR has thread-unsafe markers that this fixes, I think the best thing is probably to merge this first and then update the thread-unsafe marker PR to unmark tests fixed here.

ngoldbaum:
I think the docs build is failing because the SciPy website is having intermittent issues: https://github.com/scipy/docs.scipy.org/issues/102. The SciPy intersphinx mapping isn't getting resolved.

I re-triggered the docs build.

@lysnikolaou are you ok with this approach? Given the lack of a way to register fixtures for pytest-run-parallel to inject on a per-thread basis into each worker thread, I think this is actually the easiest way to do this right now.

lysnikolaou:
> are you ok with this approach? Given the lack of a way to register fixtures for pytest-run-parallel to inject on a per-thread basis into each worker thread, I think this is actually the easiest way to do this right now.

Yes, LGTM!

ngoldbaum:
In it goes, thanks @bwhitt7!

PR #29092, opened 2025-05-30
TYP: fix `NDArray[integer]` inplace operator mypy issue

jorenham:
fixes python/mypy#19171

---

Note that this does not affect the return type; this just broadens the input (`self`) type.

The `unsignedinteger` overloads could be removed because `_ArrayLikeInt_co` already covers the array-likes of `integer` or `bool`.

PR #29862, opened 2025-10-02
TYP: Improved ``ndarray`` augmented assignment operators

jorenham:
By using type-restricted method-bound type parameters for `self` in the `ndarray.__i{op}__` "augmented assignment" operator methods, they'll now covariantly specialize, and side-step the need for higher-kinded types (sorry about the jargon). I.e., subtypes of `ndarray` no longer need to override the `__i{op}__` methods.
This allows us to get rid of a bunch of duplication in e.g. `np.ma.MaskedArray`, and users with their own `ndaray` subclasses are now also able to do the same.

This also addresses the issue described in https://github.com/python/mypy/issues/19171 more rigourously than #29092, since the `self`-bound type-parameters are now slightly wider. For example the `ndarray.__add__` overload that accepts "integer-like array-likes" (`_ArrayLikeInt_co`) now applies to all `np.number` arrays, instead of just the `np.integer` arrays. For floating-like array-likes the `np.floating` restriction is now similarly widened to `np.inexact`. I believe this resolves the remaining class of
issues described in https://github.com/python/mypy/issues/19171#issuecomment-3352204418 by @david-zwicker.


jorenham:
@MarcoGorelli do you think we could also apply this trick to get rid of the other arithmetic dunder overrides in `MaskedArray`, or do they behave differently?

numpy/__init__.pyi
```diff
@@ -3416,141 +3430,139 @@ class ndarray(_ArrayOrScalarCommon, Generic[_ShapeT_co, _DTypeT_co]):
...
+    def __iadd__(self: _BoolArrayT, other: _ArrayLikeBool_co, /) -> _BoolArrayT: ...
     @overload
-    def __iadd__(self: NDArray[integer], other: _ArrayLikeInt_co, /) -> ndarray[_ShapeT_co, _DTypeT_co]: ...
+    def __iadd__(self: _ComplexFloatingArrayT, other: _ArrayLikeComplex_co, /) -> _ComplexFloatingArrayT: ...
     @overload
-    def __iadd__(self: NDArray[floating], other: _ArrayLikeFloat_co, /) -> ndarray[_ShapeT_co, _DTypeT_co]: ...
+    def __iadd__(self: _InexactArrayT, other: _ArrayLikeFloat_co, /) -> _InexactArrayT: ...
     @overload
-    def __iadd__(self: NDArray[complexfloating], other: _ArrayLikeComplex_co, /) -> ndarray[_ShapeT_co, _DTypeT_co]: ...
+    def __iadd__(self: _NumberArrayT, other: _ArrayLikeInt_co, /) -> _NumberArrayT: ...
```
jorenham, line 3443: I had to flip the overload order here to appease mypy. It started complaining about non-existent overlap once I put the `_*ArrayT` typars in, demonstrating that the error is a false-positive.

charris:
@MarcoGorelli Ping.

charris:
Needs rebase.

charris:
Thanks Joren.

MarcoGorelli:
Sorry for the delay, was on holiday and then got back to a long backlog of issues to respond to

> @MarcoGorelli do you think we could also apply this trick to get rid of the other arithmetic dunder overrides in MaskedArray, or do they behave differently?

So, to `__add__` / `__radd__` / `__sub__` / `__rsub__` / multiplication / floordiv / truediv / power? Sure, I'm not aware of why not

PR #29867, opened 2025-10-02
TST: Fix misc thread unsafe data races

bwhitt7:
This PR is related to #29552 and also a sort of continuation of #29816. A few usages of `@pytest.mark.parametrize` resulted in data races when ran in parallel threads using [pytest-run-parallel](github.com/Quansight-Labs/pytest-run-parallel). By making it so the tests instead use a `.copy` of the parameterized values (as suggested by @lysnikolaou), this gets rid of the data race.

This also gives a small edit to `numpy/lib/tests/test_io.py` to fix up some data races there.

### DIFF
--- a/numpy/_core/tests/test_nditer.py
+++ b/numpy/_core/tests/test_nditer.py
@@ -2895,7 +2895,7 @@ def _is_buffered(iterator):
         return True
     return False
 
-@pytest.mark.parametrize("a",
+@pytest.mark.parametrize("arrs",
         [np.zeros((3,), dtype='f8'),
          np.zeros((9876, 3 * 5), dtype='f8')[::2, :],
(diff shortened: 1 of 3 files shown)


charris:
I'm curious as to how parametized tests have trouble with parallel processing.

charris:
The test failures are unrelated.

lysnikolaou:
> I'm curious as to how parametized tests have trouble with parallel processing.

It doesn't have to do with parametrized tests, but rather how `pytest-run-parallel` runs the tests in question. What it does, is it spawns x parallel threads that all run the test functions. It passes the same parameter objects to all of the parallel functions, which means that the parameters effectively become shared state. In the case of the tests in this PR, the parameters are being mutated, which leads to data races.

lysnikolaou (approved):
LGTM! Thanks @bwhitt7!

charris:
Thanks Britney.

PR #29868, opened 2025-10-03
BLD: Fix MSVC warnings and add CI check with allowlist

seiko2plus:
- Fixes the majority of MSVC compiler warnings in the codebase.
- Introduces an allowlist for known/acceptable warnings.
- Adds a CI step to detect new or unexpected warnings during MSVC builds.

### DIFF
--- a/.github/check-warnings/msvc-allowed-warnings.txt
+++ b/.github/check-warnings/msvc-allowed-warnings.txt
@@ -1,0 +1,31 @@
+../numpy/linalg/lapack_lite/f2c.c(194): warning C4244: 'return': conversion from 'double' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_c_lapack.c(230): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_c_lapack.c(250): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_c_lapack.c(1530): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_s_lapack.c(1625): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_s_lapack.c(1645): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_s_lapack.c(2865): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_s_lapack.c(2882): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_s_lapack.c(2894): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_config.c(1368): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c.c(194): warning C4244: 'return': conversion from 'double' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_c_lapack.c(230): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_c_lapack.c(250): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_c_lapack.c(1530): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_s_lapack.c(1625): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_s_lapack.c(1645): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_s_lapack.c(2865): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_s_lapack.c(2882): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_s_lapack.c(2894): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+numpy/random/_generator.cp311-win_arm64.pyd.p/numpy/random/_generator.pyx.c(26966): warning C4244: 'function': conversion from 'int64_t' to 'double'
+numpy/random/_generator.cp311-win_arm64.pyd.p/numpy/random/_generator.pyx.c(39070): warning C4244: 'function': conversion from 'int64_t' to 'double'
+numpy/random/_generator.cp311-win32.pyd.p/numpy/random/_generator.pyx.c(26966): warning C4244: 'function': conversion from 'int64_t' to 'double'
+numpy/random/_generator.cp311-win32.pyd.p/numpy/random/_generator.pyx.c(39070): warning C4244: 'function': conversion from 'int64_t' to 'double'
+cl : Command line warning D9025 : overriding '/arch:SSE2' with '/arch:AVX2'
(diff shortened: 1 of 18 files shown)


numpy/_core/src/multiarray/lowlevel_strided_loops.c.src
```diff
@@ -888,21 +891,20 @@ npy_uint64 _npy_halfbits_to_doublebits(npy_uint16 h){
...
 #  define _TO_RTYPE1(x) (@rtype1@)(x)
-
 #  if @is_float1@
 #    define _CONVERT_FN(x) npy_floatbits_to_halfbits(x)
-#    define _ROUND_TRIP(x) (@rtype1@)npy_halfbits_to_floatbits(_CONVERT_FN(x))
+#    define _ROUND_TRIP(x) npy_half_to_float(npy_float_to_half(x))
 #  elif @is_double1@
 #    define _CONVERT_FN(x) npy_doublebits_to_halfbits(x)
-#    define _ROUND_TRIP(x) (@rtype1@)_npy_halfbits_to_doublebits(_CONVERT_FN(x))
+#    define _ROUND_TRIP(x) npy_half_to_double(npy_double_to_half(x))
```
seiko2plus, line 907: @mattip, It looks a bug to me, do we have test case covering it?

numpy/_core/src/multiarray/stringdtype/static_string.c
```diff
@@ -177,14 +177,15 @@ arena_malloc(npy_string_arena *arena, npy_string_realloc_func r, size_t size)
...
         }
         else {
             newsize = arena->size + string_storage_size;
         }
-        if ((arena->cursor + size) >= newsize) {
+        // If there enough room for both the payload and its header
+        if ((arena->cursor + string_storage_size) > newsize) {
             // need extra room beyond the expansion factor, leave some padding
-            newsize = ARENA_EXPAND_FACTOR * (arena->cursor + size);
+            newsize = (size_t)(ARENA_EXPAND_FACTOR * (arena->cursor + string_storage_size));
```
seiko2plus, line 188: @ngoldbaum I'm not sure if I understand it correctly, but isn't it supposed to also check the header size?

.github/check-warnings/msvc-allowed-warnings.txt
```diff
@@ -0,0 +1,31 @@
...
+../numpy/linalg/lapack_lite/f2c_c_lapack.c(1530): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_s_lapack.c(1625): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_s_lapack.c(1645): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_s_lapack.c(2865): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_s_lapack.c(2882): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+../numpy/linalg/lapack_lite/f2c_s_lapack.c(2894): warning C4244: '=': conversion from 'real' to 'integer', possible loss of data
+numpy/random/_generator.cp311-win_arm64.pyd.p/numpy/random/_generator.pyx.c(26966): warning C4244: 'function': conversion from 'int64_t' to 'double'
+numpy/random/_generator.cp311-win_arm64.pyd.p/numpy/random/_generator.pyx.c(39070): warning C4244: 'function': conversion from 'int64_t' to 'double'
```
seiko2plus, line 21: Except for those two lines (auto-generated), the other warnings will be covered in follow-up PRs

numpy/_core/src/multiarray/lowlevel_strided_loops.c.src
```diff
@@ -888,21 +891,20 @@ npy_uint64 _npy_halfbits_to_doublebits(npy_uint16 h){
 #    define _ROUND_TRIP(x) (x)
 #  elif @is_bool2@
 #    define _CONVERT_FN(x) ((npy_bool)!npy_half_iszero(x))
-#    define _ROUND_TRIP(x) npy_float_to_half((float)(!npy_half_iszero(x)))
+#    define _ROUND_TRIP(x) npy_float_to_half((!npy_half_iszero(x))
```
charris: ~Needs a `)`?~

Has extra `(`.

numpy/_core/src/npysort/radixsort.cpp
```diff
@@ -26,8 +26,7 @@ KEY_OF(UT x)
     if (std::is_floating_point<T>::value) {
         // For floats, we invert the key if the sign bit is set, else we invert
         // the sign bit.
-        return ((x) ^ (-((x) >> (sizeof(T) * 8 - 1)) |
-                       ((UT)1 << (sizeof(T) * 8 - 1))));
+        return x ^ ((UT)(- (int)(x >> (sizeof(T) * 8 - 1))) | ((UT)1 << (sizeof(T) * 8 - 1)));
```
charris: Extra space after `-`. It would be nice to group things better.

numpy/_core/src/multiarray/stringdtype/static_string.c
ngoldbaum: @mhvk any chance you can look at this to see if it's a StringDType bug? I'm on vacation this week and will be busy next week so won't have time to look at this for a while. No worries if you don't have time either.

numpy/_core/src/multiarray/lowlevel_strided_loops.c.src
mattip: The x86_64 tests (with emulated float16) of float32/float64 -> float16 should hit this and AFAICT are passing

numpy/_core/src/multiarray/stringdtype/static_string.c
seberg: Changing `size` to `string_storage_size` seems right `size` should not be used for any storage computation ever.
That said, this whole branch looks like it should be deleted.  Go through the above parts, I don't see how it can ever be triggered.

For this PR, I think even just this is OK, TBH.

(Independently, the above code looks subtly wrong, though.  And I think it should be fixed in a separate PR.  There is a similar `grow_size_and_multiply` helper in textreading, unfortunately it is for `npy_intp` sized only and I think this is truly unsigned sizes on 32bit systems).

numpy/_core/src/multiarray/lowlevel_strided_loops.c.src
seberg: How come this doesn't fail CI!?  The definition must be unused for this branch?

seberg (approved):
I'll just say, I am generally happy with this, that unbalanced bracket is a bit surprising.
The `-usigned` warnings/changes are a bit tedious, but not too bad overall.

numpy/_core/src/multiarray/lowlevel_strided_loops.c.src
seiko2plus: My bad. This change was actually meant to break the build. Looking again, it makes sense: _ROUND_TRIP never gets called when the second type is bool. See:
https://github.com/numpy/numpy/blob/1b11ae0aef650345b8218bebf79af6fd0e1fe25e/numpy/_core/src/multiarray/lowlevel_strided_loops.c.src#L956-L962

numpy/_core/src/npysort/radixsort.cpp
seiko2plus: Fix it, also `int` should be replaced with `make_signed`

seiko2plus:
@seberg,
>  The -usigned warnings/changes are a bit tedious, but not too bad overall.

How about revert these changes and globally disables this warning via:
```C
#if defined(_MSC_VER)
#pragma warning(disable:4146)
#endif
```

charris:
> How about revert these changes and globally disables this warning via

That makes sense to me.

seiko2plus:
> That makes sense to me.

Done

charris (approved):

charris:
Let's get this in. Thanks Sayed.

@mhvk You might want to take a look at the string changes.

numpy/_core/src/multiarray/stringdtype/static_string.c
mhvk: Belatedly, does look correct to me!

seiko2plus:
As discussed, there is no need to perform a backport; the entire dstring branch check is unnecessary, as the float16 bug was newly introduced.

Issue #29677, opened 2025-09-05
DOC: Add memory sharing warnings and examples to sliding_window_view


ajoosen:
### Issue with current documentation:

[Current documentation](https://numpy.org/doc/stable/reference/generated/numpy.lib.stride_tricks.sliding_window_view.html) for `numpy.lib.stride_tricks.sliding_window_view` does not sufficiently warn about shared memory behavior when `writeable=True`. Current documentation under `writeable` mentions "writing to one location will cause others to change" but this brief mention at the end of the argument description does not mention that the original array is also modified and does not sufficiently convey the risk. This can cause silent unexpected behavior that is difficult to debug, affecting downstream data analysis that assumes window independence or that the original array was not changed. Current documentation contains no example demonstrating this behavior. 

### Idea or request for content:

I suggest adding a warning box similar to [this one](https://numpy.org/doc/stable/reference/generated/numpy.lib.stride_tricks.as_strided.html) in `as_strided` that says "This function creates views with overlapping memory. When `writeable=True`, writing to the view will modify the original array and may affect multiple view positions. See the examples below and this [guide](https://numpy.org/doc/stable/user/basics.copies.html) about the difference between copies and views." 

Additionally, I propose adding the examples below to demonstrate the behavior. 

Creating a view with `writeable=False` and then writing to it raises an error. 
```
>>> import numpy as np
>>> from numpy.lib.stride_tricks import sliding_window_view
>>> x = np.arange(6)
>>> x.shape
(6,)
>>> x
array([0, 1, 2, 3, 4, 5])
>>> v = sliding_window_view(x, 3)   # create view with default writeable=False
>>> v.shape
(4, 3)
>>> v
array([[0, 1, 2],
       [1, 2, 3],
       [2, 3, 4],
       [3, 4, 5]])
>>> v[0,1] = 10    # writing to view will raise error
Traceback (most recent call last):
  File "", line 1, in 
ValueError: assignment destination is read-only
```

Creating a view with `writeable=True` and then writing to it changes the value in the original array and multiple view positions. 
```
>>> v = sliding_window_view(x, 3, writeable=True)   # create view with writeable=True
>>> v
array([[0, 1, 2],
       [1, 2, 3],
       [2, 3, 4],
       [3, 4, 5]])
>>> x
array([0, 1, 2, 3, 4, 5])
>>> v[0,1] = 10   # write to position in view
>>> x
array([ 0, 10,  2,  3,  4,  5])   # value modified in original array
[...]

PR #29869, opened 2025-10-03
DOC: Add warning and examples for sliding_window_view

Alverok:
Closes #29677.

### Description of the problem solved

The existing documentation for `numpy.lib.stride_tricks.sliding_window_view` does not sufficiently warn users about the risks of using `writeable=True`. When the returned view is modified, it also modifies the original array and can affect multiple positions within the view itself due to overlapping memory. This can lead to silent, hard-to-debug errors in user code that assumes the view is an independent copy.

### Summary of changes

* A `.. warning::` directive has been added to the `Notes` section of the docstring. This warning explains the shared memory behavior and links to the user guide on copies and views.
* Two new code examples have been appended to the `Examples` section:
    1.  The first demonstrates the correct `ValueError` that is raised when trying to write to a default, read-only view.
    2.  The second demonstrates how writing to a `writeable=True` view modifies both the original array and other parts of the view.

These changes directly address the suggestions in the original issue and make the documentation safer and clearer for users.

---
### Checklist

- [x] I have read the [NumPy contributor guide](https://numpy.org/devdocs/dev/index.html).
- [x] This is a documentation-only change.
- [x] The new documentation is clear and readable.
- [x] The change is supported by the conversation in the linked issue.

PR #29870, opened 2025-10-03
DOC: Add memory sharing warnings and examples to sliding_window_view

ajoosen:
I opened the original issue @ajoosen (#29677) and am submitting this PR to address it.

Currently, the documentation for `numpy.lib.stride_tricks.sliding_window_view` does not sufficiently warn about the shared memory behavior when `writeable=True`.

This commit adds:

- A warning box explaining that writing to the view modifies the original array and can affect multiple positions.
- Examples showing the difference between `writeable=False` (raises an error) and `writeable=True` (modifies original array and overlapping view positions).

These additions clarify potential pitfalls and help prevent silent bugs when using sliding window views.


mattip:
Addressed in #29869

PR #29872, opened 2025-10-03
ENH: Set DLPack tensor `shape` and `strides` to NULL iff `ndim == 0`.

hpkfft:
DLPack allows, but does not require, `DLTensor.strides` to be `NULL`  if `PyArray_IS_C_CONTIGUOUS`.
Previously, in this case, NumPy set the `strides` pointer to `NULL`.
Now, NumPy will supply a valid pointer to correctly initialized strides when `ndim > 0`.

This change can improve performance for downstream users.  For example, nanobind can be used to import a NumPy array using the DLPack protocol to make the array available to C++ code.  If `DLTensor.strides` is `NULL`, nanobind allocates and initializes a `strides` array.  This redundant work is now avoided since NumPy provides a pointer to the `strides`, which it allocated and initialized anyway.

If `ndim == 0`, this PR sets both `shape` and `strides` to `NULL`.
Previously, `strides` was was set to `NULL` (since `PyArray_IS_C_CONTIGUOUS`), but `shape` pointed past the end of allocated memory.  For security and debugability, `NULL` is preferred.

### DIFF
--- a/numpy/_core/src/multiarray/dlpack.c
+++ b/numpy/_core/src/multiarray/dlpack.c
@@ -285,10 +285,6 @@ fill_dl_tensor_information(
     }
 
     dl_tensor->ndim = ndim;
-    if (PyArray_IS_C_CONTIGUOUS(self)) {
-        /* No need to pass strides, so just NULL it again */
-        dl_tensor->strides = NULL;
-    }
     dl_tensor->byte_offset = 0;
 
     return 0;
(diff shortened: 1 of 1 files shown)


seberg:
I don't care either way about NULLing, the important change here is: https://github.com/dmlc/dlpack/issues/177

This is useful for you and since it was brought up by Tianqi, I guess we might as well.  I'll note that `NULL` is actually convenient for consumers who expect a contiguous array, since they can skip any further checks.

Anyway, looks fine, a small test would be nice, though.  But I'll wait for the decision there, at least a bit, so ping again if we forget about it.

seberg:
OK, I'll just put it in, thanks @hpkfft.  Basically, I am doubting the discussion there changes.  And even if it does, this doesn't actually hurt, since I doubt other exporters add this logic to try and be helpful (it just doesn't help either).

PR #29877, opened 2025-10-05
BUG: Fix repr of polynomial for non-standard dtypes

Rishabh-git10:
Fixes #29668

**Problem:**

The `__repr__` method in `numpy.polynomial._polybase.ABCPolyBase` used fixed string slicing to format the `coef`, `domain`, and `window` arrays. This failed when a non-default `dtype` was present in the array's string representation, leading to an incorrect `repr`.

**Solution:**

This PR changes the implementation to robustly find the start `[` and end `]` brackets of the array's string representation. This correctly extracts the list of values regardless of whether a `dtype` is present, fixing the bug for non-standard dtypes while also preserving the correct output for default dtypes.

### DIFF
--- a/doc/release/upcoming_changes/29668.bug.rst
+++ b/doc/release/upcoming_changes/29668.bug.rst
@@ -1,0 +1,1 @@
+Fix ``__repr__`` for ``numpy.polynomial`` classes with non-standard dtypes.
--- a/numpy/polynomial/_polybase.py
+++ b/numpy/polynomial/_polybase.py
@@ -320,12 +320,17 @@ class ABCPolyBase(abc.ABC):
         self._symbol = symbol
 
     def __repr__(self):
-        coef = repr(self.coef)[6:-1]
-        domain = repr(self.domain)[6:-1]
-        window = repr(self.window)[6:-1]
-        name = self.__class__.__name__
-        return (f"{name}({coef}, domain={domain}, window={window}, "
-                f"symbol='{self.symbol}')")
+        s_coef = repr(self.coef)
+        coef = s_coef[s_coef.find('['):s_coef.rfind(']')+1]
+
+        s_domain = repr(self.domain)
+        domain = s_domain[s_domain.find('['):s_domain.rfind(']')+1]
+
+        s_window = repr(self.window)
+        window = s_window[s_window.find('['):s_window.rfind(']')+1]
+
+        return (f"{self.__class__.__name__}({coef}, "
+            f"domain={domain}, window={window}, symbol='{self.symbol}')")
 
     def __format__(self, fmt_str):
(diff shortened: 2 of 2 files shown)


jorenham (commented):
Could you revert the formatting changes? I'm not against it per se, but it makes reviewing difficult.

Rishabh-git10:
I've updated the pull request and reverted the automatic formatting changes, as you requested. The commits should be much cleaner to review now.
Let me know if anything else is needed!

jorenham:
> I've updated the pull request and reverted the automatic formatting changes, as you requested. The commits should be much cleaner to review now. Let me know if anything else is needed!

There's no need for AI-generated comments like these ;)

jorenham:
Could you show an example of what the `repr` looks like now vs before?

Rishabh-git10:
After reproducing the error using the provided code -

```py
import numpy as np
a = np.array([1,2,3], dtype=np.float32)
repr(np.polynomial.Polynomial(a))
```

```
"Polynomial([1., 2., 3.], dtype=float32, domain=[-1.,  1.], window=[-1.,  1.], symbol='x')"`
```

I cleaned up to produce a valid expression

```
Polynomial([1., 2., 3.], domain=[-1.,  1.], window=[-1.,  1.], symbol='x')
```

Rishabh-git10:
Do I need to do something about the 3 failed tests ?

jorenham:
> Do I need to do something about the 3 failed tests ?

Yea, they're linting failures, so you gotta do one of those `ruff check --fix`

jorenham:
> After reproducing the error using the provided code -
> 
> ```python
> import numpy as np
> a = np.array([1,2,3], dtype=np.float32)
> repr(np.polynomial.Polynomial(a))
> ```
> 
> ```
> "Polynomial([1., 2., 3.], dtype=float32, domain=[-1.,  1.], window=[-1.,  1.], symbol='x')"`
> ```
> 
> I cleaned up to produce a valid expression
> 
> ```
> Polynomial([1., 2., 3.], domain=[-1.,  1.], window=[-1.,  1.], symbol='x')
> ```

Hmm I don't see why you call this a bug, or why this changes makes it better. The `dtype=float32` seems relevant information, no?

Rishabh-git10:
The earlier implementation using slicing was not robust enough. Also the utility of `__repr__` as I understand is to return a valid executable string in python to paste back to Python without any modification to get back the object, but it was not working with non-standard datatypes as mentioned in the original issue using jax.numpy.
I implemented the ruff --fix. Let's see if it works this time.

jorenham:
Following https://github.com/numpy/numpy/issues/29668#issuecomment-3372286024, I'll be closing this

PR #29880, opened 2025-10-05
MAINT: Rewrite setitem to use the new API (mostly)

seberg:
This has a mild speed advantage since `PyArray_Pack` had this hack in place to ensure that we have an "array" to pass on.  This hack is now only used in extreme cases (i.e. user dtypes).

This gives a very slight speed advantage, but mostly, I like to change this.

It would be nice to change `getitem` as well, but `getitem` is trickier due to structured dtypes.

I wanted to first just remove `->setitem` but one (brutal!) test failed and I reconsidered things thinking that someone might be using `descr->setitem` without an array which would be broken without keeping an explicit definition around.

The 0-D unpacking path should be unimportant these days (it only applies to _nested_ arrays so the tests don't even notice it) but I kept it around for now anyway.

EDIT: It would be good to do the same thing for `_getitem` but it needs some thought and probably a second `_getitem` version that supports a `base` "owning" object...

### DIFF
--- a/numpy/_core/src/multiarray/array_coercion.c
+++ b/numpy/_core/src/multiarray/array_coercion.c
@@ -479,20 +479,13 @@ npy_cast_raw_scalar_item(
 NPY_NO_EXPORT int
 PyArray_Pack(PyArray_Descr *descr, void *item, PyObject *value)
 {
-    PyArrayObject_fields arr_fields = {
-            .flags = NPY_ARRAY_WRITEABLE,  /* assume array is not behaved. */
-        };
-    Py_SET_TYPE(&arr_fields, &PyArray_Type);
-    Py_SET_REFCNT(&arr_fields, 1);
-
     if (NPY_UNLIKELY(descr->type_num == NPY_OBJECT)) {
         /*
          * We always have store objects directly, casting will lose some
          * type information. Any other dtype discards the type information.
          * TODO: For a Categorical[object] this path may be necessary?
          */
-        arr_fields.descr = descr;
-        return PyDataType_GetArrFuncs(descr)->setitem(value, item, &arr_fields);
+        return NPY_DT_CALL_setitem(descr, value, item);
     }
 
     /* discover_dtype_from_pyobject includes a check for is_known_scalar_type */
@@ -527,8 +520,7 @@ PyArray_Pack(PyArray_Descr *descr, void *item, PyObject *value)
     if (DType == NPY_DTYPE(descr) || DType == (PyArray_DTypeMeta *)Py_None) {
         /* We can set the element directly (or at least will try to) */
         Py_XDECREF(DType);
-        arr_fields.descr = descr;
-        return PyDataType_GetArrFuncs(descr)->setitem(value, item, &arr_fields);
+        return NPY_DT_CALL_setitem(descr, value, item);
     }
     PyArray_Descr *tmp_descr;
     tmp_descr = NPY_DT_CALL_discover_descr_from_pyobject(DType, value);
@@ -546,8 +538,7 @@ PyArray_Pack(PyArray_Descr *descr, void *item, PyObject *value)
     if (PyDataType_FLAGCHK(tmp_descr, NPY_NEEDS_INIT)) {
         memset(data, 0, tmp_descr->elsize);
     }
-    arr_fields.descr = tmp_descr;
-    if (PyDataType_GetArrFuncs(tmp_descr)->setitem(value, data, &arr_fields) < 0) {
+    if (NPY_DT_CALL_setitem(tmp_descr, value, data) < 0) {
         PyObject_Free(data);
         Py_DECREF(tmp_descr);
         return -1;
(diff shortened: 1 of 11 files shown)


numpy/_core/tests/test_array_coercion.py
```diff
@@ -266,11 +266,6 @@ def test_scalar_coercion(self, scalar):
             # Ensure we have a full-precision number if available
             scalar = type(scalar)((scalar * 2)**0.5)
 
-        if type(scalar) is rational:
-            # Rational generally fails due to a missing cast. In the future
-            # object casts should automatically be defined based on `setitem`.
-            pytest.xfail("Rational to object cast is undefined currently.")
```
seberg, line 272: Must have been working for many years...

numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -366,28 +385,23 @@ NPY_NO_EXPORT int
...
-        PyObject *type, *value, *traceback;
-        PyErr_Fetch(&type, &value, &traceback);
-        if (PySequence_NoString_Check(op)) {
-            PyErr_SetString(PyExc_ValueError,
-                    "setting an array element with a sequence.");
-            npy_PyErr_ChainExceptionsCause(type, value, traceback);
-        }
-        else {
-            PyErr_Restore(type, value, traceback);
+        if (temp == (@type@)-1 && PyErr_Occurred()) {
```
seberg, line 388: This just moves it.  The first branch clearly can't fail, and the second one now has an `== -1` check (which required fixing the called helpers a bit).

seberg:
@mhvk maybe you can have a look?  In a sense this just an internal refactor ensuring that the new `setitem` now always exists and is the preferred entry-point.
Basically, realized that this is possible now, and also a minor speedup for `PyArray_Pack()` (but really very quite minor).

External code will still go via the old code paths mostly, but in most cases should use `PyArray_Pack` anyway.

numpy/_core/src/multiarray/array_coercion.c
```diff
@@ -479,20 +479,13 @@ npy_cast_raw_scalar_item(
 NPY_NO_EXPORT int
 PyArray_Pack(PyArray_Descr *descr, void *item, PyObject *value)
 {
-    PyArrayObject_fields arr_fields = {
```
mhvk, line 482: I looked at this earlier and had *no* idea what it was about, so nice to get rid of it...

numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -139,10 +153,16 @@ MyPyFloat_FromHalf(npy_half h)
     return PyFloat_FromDouble(npy_half_to_double(h));
 }
 
-/* Handle case of assigning from an array scalar in setitem */
+/*
+ * Handle case of assigning from an array scalar in setitem.
+ * NOTE/TODO(seberg): This was important, but is now only userd
```
mhvk: userd -> used

numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -808,13 +810,13 @@ STRING_setitem(PyObject *op, void *ov, void *vap)
         Py_DECREF(temp);
         return -1;
     }
-    memcpy(ov, ptr, PyArray_MIN(PyArray_ITEMSIZE(ap),len));
+    memcpy(ov, ptr, PyArray_MIN(descr->elsize,len));
```
mhvk: If you're editing again, might as well add a space after the ,

numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -1029,18 +1023,17 @@ _copy_and_return_void_setitem(_PyArray_LegacyDescr *dstdescr, char *dstdata,
 }
 
 static int
-VOID_setitem(PyObject *op, void *input, void *vap)
+VOID_setitem(PyArray_Descr *descr_, PyObject *op, char *ip)
 {
-    char *ip = input;
-    PyArrayObject *ap = vap;
-    int itemsize = PyArray_ITEMSIZE(ap);
+    _PyArray_LegacyDescr *descr = (_PyArray_LegacyDescr *)descr_;
```
mhvk, line 1028: More out of curiosity, but why is this needed? (It was there already, so I guess just that some aspects cannot be retrieved by our macro API yet?)

mhvk (approved):
Yes, that's quite a bit nicer -- the unclear stuff is now all in functions called `legacy`.

numpy/_core/src/multiarray/arraytypes.c.src
seberg: I prefer this and also the macro has to add an unnecessary if statement. `LegacyDescr` is a subclass (instance) of `PyArray_Descr` and to me it is actually the most reasonable thing to cast.

seberg:
Thanks for the review, I think this should be uncontroversial, the only thing might be small additional fixups, if anyone comes around and asks for it, I'll do that.

PR #29882, opened 2025-10-06
DOC: Remove unused arrays from the structured dtype ufunc example.

WarrenWeckesser:
The `funcs` and `types` arrays are not used in the example of the ufunc that handles a structured array dtype.


charris:
Thanks Warren.

PR #29885, opened 2025-10-06
MAINT: Simplify string arena growth strategy

seberg:
Not sure if there is any downsides to this, but I think this is much simpler and pretty much identical.
Plus, I don't quite like that the 1.25 factor could be large, although I admit it seems very unlikely (maybe impossible) to actually trigger a problem there.

EDIT: As a note, I have a similar function that caps overallocation and increments in "neat" values (which double as a minimum overallocation amount).  I somewhat assumed that doesn't matter much, but we could definitely do either or both here as well.

### DIFF
--- a/numpy/_core/src/multiarray/stringdtype/static_string.c
+++ b/numpy/_core/src/multiarray/stringdtype/static_string.c
@@ -155,8 +155,6 @@ vstring_buffer(npy_string_arena *arena, _npy_static_string_u *string)
     return (char *)((size_t)arena->buffer + string->vstring.offset);
 }
 
-#define ARENA_EXPAND_FACTOR 1.25
-
 static char *
 arena_malloc(npy_string_arena *arena, npy_string_realloc_func r, size_t size)
 {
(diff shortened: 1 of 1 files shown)


numpy/_core/src/multiarray/stringdtype/static_string.c
```diff
@@ -168,25 +166,14 @@ arena_malloc(npy_string_arena *arena, npy_string_realloc_func r, size_t size)
...
-            newsize = (size_t)(ARENA_EXPAND_FACTOR * arena->size);
-        }
-        else {
-            newsize = arena->size + string_storage_size;
-        }
-        // If there enough room for both the payload and its header
-        if ((arena->cursor + string_storage_size) > newsize) {
-            // need extra room beyond the expansion factor, leave some padding
-            newsize = (size_t)(ARENA_EXPAND_FACTOR * (arena->cursor + string_storage_size));
+    if ((arena->size - arena->cursor) < string_storage_size) {
```
seberg, line 169: I changed the `<=` to `<` here.  `==` seems fine, i.e. at the end `size == cursor` which should be valid, no?

ngoldbaum:
Thanks for looking at this! Just an FYI I am traveling in Portugal this week and may not have time to look at this until next week.

ngoldbaum:
Thanks for looking at this. I agree that the original code has issues and your fix is fine.

In your edit:

>  As a note, I have a similar function that caps overallocation and increments in "neat" values (which double as a minimum overallocation amount). I somewhat assumed that doesn't matter much, but we could definitely do either or both here as well.

This seems like a good idea too and I like the idea of having a utility function we can use for growing buffers. If you're up for it, it would be nice to re-use that code here. But also this is fine to merge as-is IMO.

ngoldbaum (approved):

seberg:
I'll just put this in as back-portable, although dunno if we will.   May look into consolidating, but that would make the change unnecessarily bloated for backporting (if it turns out nice).

PR #29886, opened 2025-10-06
CI: macos-13 --> macos-15-intel

andyfaff:
The macos-13 image is deprecated and will be removed soon. This PR removes `macos-13` entries from CI, moving them both to `macos-15-intel`. `macos-15-intel` is suggested because both of the entries were targeting x86_64.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -113,7 +113,7 @@ jobs:
       fail-fast: false
       matrix:
         build_runner:
-          - [ macos-13, "macos_x86_64" ]
+          - [ macos-15-intel, "macos_x86_64" ]
           - [ macos-14, "macos_arm64" ]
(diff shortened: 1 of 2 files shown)


.github/workflows/macos.yml
```diff
@@ -128,11 +128,6 @@ jobs:
       with:
         python-version: ${{ matrix.version }}
 
-    - uses: maxim-lobanov/setup-xcode@60606e260d2fc5762a71e64e74b2174e8ea3c8bd # v1.6.0
```
andyfaff, line 131: This was probably required for a specific reason, now lost to time.

rgommers: Indeed, I remember we needed a newer XCode at some point, but that's no longer a problem.

rgommers (approved):
Looked through the build logs, all LGTM. Thanks @andyfaff

PR #29889, opened 2025-10-07
DOC: Documentation related finfo refactors and new slot addition

SwayamInSync:
This is a follow-up PR of work happened in #29836 
Summary of prev work
- The deprecated `MachAr` runtime discovery mechanism has been removed.
- `np.finfo` fetches the constants provided by the compiler macros
- new slot to fetch the dtype related constants


SwayamInSync:
@seberg this is the first draft, you can have a look.
Wrote all the major things. `np.finfo`'s docstring might also needed to be modified will do it in this itself.

doc/release/upcoming_changes/29836.improvement.rst
```diff
@@ -0,0 +1,23 @@
...
+* Constants like ``eps``, ``min``, ``max``, ``smallest_normal``, and
+  ``smallest_subnormal`` now come directly from standard C macros (``FLT_EPSILON``,
+  ``DBL_MIN``, etc.), ensuring platform-correct values.
+
+* The deprecated ``MachAr`` runtime discovery mechanism has been removed.
+
+* Derived attributes have been corrected to match standard definitions:
+  ``machep`` and ``negep`` now use ``int(log2(eps))``; ``nexp`` accounts for
+  all exponent patterns; ``nmant`` excludes the implicit bit; and ``minexp``
+  follows the C standard definition.
```
seberg, line 17: Wait, did we change any values here?! Because that we must revert if we did.  (unless the old values were blatently wrong -- but even if wrong people may work around it...)

The only thing that we really should mention here I think is that some `longdouble`, specifically `smallest_normal` may now have changed to match those defined by the C-headers.
I.e. what changed for the end user, for the end user it doesn't matter that the constants come from C now, they are the same values after all :).

doc/source/reference/c-api/dtype.rst
```diff
@@ -502,3 +502,140 @@ format specifier in printf and related commands.
 .. c:macro:: NPY_UINTP_FMT
 
 .. c:macro:: NPY_LONGDOUBLE_FMT
+
+
+DType constant
+-------------------------
```
seberg: Probably good to make the line length match, some stuff didn't like mismatches.

doc/source/reference/c-api/dtype.rst
```diff
@@ -502,3 +502,140 @@ format specifier in printf and related commands.
...
+
+NumPy provides a mechanism for dtypes to expose constant values such as machine
+limits, special values (infinity, NaN), and floating-point characteristics.
+This is implemented through the ``NPY_DT_get_constant`` slot in the DType API.
+
```
seberg: I suppose right now this means setting the `kind` to `f` but maybe that is maybe OK.  Maybe it makes sense to actually say that implementing the float/finfo constants will make `np.finfo` work?  (Just one sentence)?

doc/release/upcoming_changes/29836.improvement.rst
```diff
@@ -0,0 +1,23 @@
...
+The ``numpy.finfo`` class has been completely refactored to obtain floating-point
+constants directly from C compiler macros rather than deriving them at runtime.
+This provides better accuracy, platform compatibility and corrected
+several attribute calculations:
+
+* Constants like ``eps``, ``min``, ``max``, ``smallest_normal``, and
+  ``smallest_subnormal`` now come directly from standard C macros (``FLT_EPSILON``,
+  ``DBL_MIN``, etc.), ensuring platform-correct values.
+
+* The deprecated ``MachAr`` runtime discovery mechanism has been removed.
```
seberg, line 12: This part is important.  You could even make it it's own `29836.expired.rst`  (It is OK if it is just this one bullet point as is!)

seberg (commented):
I would prefer much shorter release notes, but I like them so much I am also fine with keeping them :)

`array.rst` is way too large, but I suspect that `dtypes.rst` is only about the NumPy builtin DTypes and the whole new (and old) API for user DTypes is in `array.rst`.
That might be worth splitting up, but I still supect this is better to move there.

Anyway, love the docs otherwise, mostly wondering if it might be better to put in the other file.

doc/release/upcoming_changes/29836.improvement.rst
SwayamInSync: For the user side, values are exactly same as they were used to. I checked this against numpy 2.3.3 for float16, 32 and 64 and the values were exactly matching.
Here I wanted to convey that the instead of hard-coded values we use C side definitions and they were a bit different so we are explicitely handling it and in case someone wants the exact C definitions then they can do as the doc says.

Good point, so I added the longdouble related details in doc but surely give that a bullet point here as well

doc/source/reference/c-api/dtype.rst
SwayamInSync: Yes on that, I actually wanted to discuss it. 
Since the current constraints want dtype to be type of `inexact` (which we does by subclassing `np.floating`) and since there is no such explicit testing on `kind` so just a generation opinion that if we have a base class to `Inexact` then maybe legacy types like `ml_dtypes` can also support this?

SwayamInSync:
This is ready for the review

doc/release/upcoming_changes/29836.c_api.rst
```diff
@@ -0,0 +1,23 @@
+New ``NPY_DT_get_constant`` slot for DType constant retrieval
+----------------------------------------------------------------
```
seberg:
```suggestion
New ``NPY_DT_get_constant`` slot for DType constant retrieval
-------------------------------------------------------------
```

doc/release/upcoming_changes/29836.improvement.rst
```diff
@@ -0,0 +1,26 @@
+``numpy.finfo`` Refactor
+-----------------------------------------------------------------
```
seberg:
```suggestion
``numpy.finfo`` Refactor
------------------------
```

doc/source/reference/c-api/array.rst
```diff
@@ -3536,6 +3536,143 @@ member of ``PyArrayDTypeMeta_Spec`` struct.
...
+.. c:macro:: NPY_DT_get_constant
+
+.. c:type:: int (PyArrayDTypeMeta_GetConstant)( \
+                PyArray_Descr *descr, int constant_id, void *out)
+
+   If defined, allows the DType to expose constant values such as machine
+   limits, special values (infinity, NaN), and floating-point characteristics.
+   The *descr* is the descriptor instance, *constant_id* is one of the
+   ``NPY_CONSTANT_*`` macros, and *out* is a pointer to memory where the
+   constant value should be written. Returns 1 on success, 0 if the constant
```
seberg: Sorry hadn't noticed it's missing here (it's in the release notes where I don't think it's needed).

This should mention that `*out` is considered uninitialized and may be unaligned. (Although, honestly, I am not sure about the later and I don't mind if you write the opposite.)
I do feel uninitialized is reasonable/nice for the typical use-case (although there may be others).

doc/source/reference/c-api/array.rst
```diff
@@ -3536,6 +3536,143 @@ member of ``PyArrayDTypeMeta_Spec`` struct.
...
+       The zero value for the dtype.
+
+   .. c:macro:: NPY_CONSTANT_one
+
+       The one value for the dtype.
+
+   .. c:macro:: NPY_CONSTANT_minimum_finite
+
+       The minimum finite value representable by the dtype. For floating-point types, this is the most negative finite value
+       (e.g., -FLT_MAX).
```
seberg:
```suggestion
       The minimum finite value representable by the dtype. For floating-point types,
       this is the most negative finite value (e.g., ``-FLT_MAX``).
```
(mostly, break long line nit)

doc/source/reference/c-api/array.rst
```diff
@@ -3536,6 +3536,143 @@ member of ``PyArrayDTypeMeta_Spec`` struct.
...
+       floating-point types.
+
+   .. c:macro:: NPY_CONSTANT_finfo_eps
+
+       Machine epsilon: the difference between 1.0 and the next representable value
+       greater than 1.0. Corresponds to C macros like ``FLT_EPSILON``, ``DBL_EPSILON``.
+
+       .. note::
+           For long double in IBM double-double format (PowerPC), this is defined as
+           ``0x1p-105L`` (2^-105) based on the ~106 bits of mantissa precision.
```
seberg, line 3604: Nice note!

doc/source/reference/c-api/array.rst
```diff
@@ -3536,6 +3536,143 @@ member of ``PyArrayDTypeMeta_Spec`` struct.
...
+   special handling:
+
+   * ``LDBL_EPSILON`` is redefined to ``0x1p-105L`` (2^-105) based on the ~106 bits
+     of mantissa precision (53+53 bits from two doubles)
+   * The maximum representable value is computed using ``nextafter(inf, 0)`` to
+     match historical behavior
+
+   **Subnormal handling**: For platforms where ``FLT_TRUE_MIN``, ``DBL_TRUE_MIN``,
+   or ``LDBL_TRUE_MIN`` are not defined by the C compiler, NumPy computes them
+   using ``nextafter(0, 1)`` to obtain the smallest positive representable value.
```
seberg: This is nice, but to be honest... I don't think there is a reason to have this whole "platform specific handling" section.

We don't actually define those macros for users right now, so it doesn't need to be here!  We could mention the subnormal handling and the long-double crazyness, but maybe putting that into `finfo` makes more sense.

Even if we make this available as public C-API (we should probably), I think that function could just refer to `np.finfo` docs.

doc/release/upcoming_changes/29836.c_api.rst
```diff
@@ -0,0 +1,23 @@
...
+A new slot ``NPY_DT_get_constant`` has been added to the DType API, allowing
+dtype implementations to provide constant values such as machine limits and
+special values. The slot function has the signature::
+
+    int get_constant(PyArray_Descr *descr, int constant_id, void *ptr)
+
+It returns 1 on success, 0 if the constant is not available, or -1 on error.
```
seberg, line 9: I am still tempted to just shorten this to.  "Implementing this can be used by user DTypes to provide ``np.finfo`` values." or so.

seberg (approved):
Thanks for the update and working on this!

If you don't mind (and don't update), I think I might just apply the tiny changes and remove that one section and put it in very soon.
(I'll just leave shortening the release notes to Chuck if he thinks they are long.)

SwayamInSync:
Cool, made all the suggestions
- Removed the `platform specific` section, added related longdouble line inside finfo docs
- shorten the c_api release notes
- added part that `*out`  in `PyArrayDTypeMeta_GetConstant` point to uninitialized memory and can be unaligned

doc/source/reference/c-api/array.rst
```diff
@@ -3536,6 +3536,120 @@ member of ``PyArrayDTypeMeta_Spec`` struct.
...
+.. c:type:: int (PyArrayDTypeMeta_GetConstant)( \
+                PyArray_Descr *descr, int constant_id, void *out)
+
+   If defined, allows the DType to expose constant values such as machine
+   limits, special values (infinity, NaN), and floating-point characteristics.
+   The *descr* is the descriptor instance, *constant_id* is one of the
+   ``NPY_CONSTANT_*`` macros, and *out* is a pointer to uninitialized memory
+   where the constant value should be written. The memory pointed to by *out*
+   may be unaligned. Returns 1 on success, 0 if the constant is not available,
+   or -1 with an error set.
```
seberg:
```suggestion
   may be unaligned and is uninitialized.
   Returns 1 on success, 0 if the constant is not available,
   or -1 with an error set.
```

numpy/_core/getlimits.py
```diff
@@ -140,6 +143,12 @@ class finfo:
     fill the gap between 0 and ``smallest_normal``. However, subnormal numbers
     may have significantly reduced precision [2]_.
 
+    For ``longdouble``, the representation varies across platforms. On most
+    platforms it is IEEE 754 binary128 (quad precision) or binary64-extended
+    (80-bit extended precision). On PowerPC systems, it may use the IBM
+    double-double format (a pair of float64 values), which has special
+    characteristics for precision and range.
```
seberg, line 150: Thanks, the note is great here!

seberg (approved):

seberg:
Thanks @SwayamInSync!

PR #29892, opened 2025-10-08
DOC: Add Plausible analytics to the NumPy documentation

InessaPawson:


seberg:
Hmmm, this config isn't actually working/passing CI.  But I suspect @stefanv if you have a bit of time, maybe you just know what is right here and can have a look?

Alverok:
Hi @InessaPawson, I was looking into the failing checks on this PR and I noticed two things:

1.  The `Lint` check might be failing due to a small syntax error in `doc/source/conf.py` (it looks like a missing comma).
2.  The `ci/circleci: build` failure is likely due to a missing integrity hash for the new analytics script.

I would be happy to submit a PR that applies your changes and attempt to fix both of these issues, as it would be a great learning experience for me. Please let me know how you'd like to proceed!

doc/source/conf.py
```diff
@@ -290,6 +290,10 @@ def setup(app):
         "json_url": "https://numpy.org/doc/_static/versions.json",
     },
     "show_version_warning_banner": True,
+    "analytics": [
+        "plausible_analytics_domain": "numpy.org/doc/stable/",
+        "plausible_analytics_url": "https://views.scientific-python.org/js/script.file-downloads.hash.outbound-links.js",
```
stefanv: This is not the correct link. Where does script.file-downloads.hash.outbound come from?

InessaPawson: @stefanv Please see the screenshot from my Plausible dashboard below. I thought we might want to enable some optional measurements.

npdocs-plausible

InessaPawson:
> I would be happy to submit a PR that applies your changes and attempt to fix both of these issues, as it would be a great learning experience for me. Please let me know how you'd like to proceed!

Really appreciate the offer, @Alverok! But every once in a while, I like to wrestle with CI on my own.😄

doc/source/conf.py
InessaPawson: @stefanv I've updated the link to the standard one.

stefanv:
@InessaPawson Interesting, I don't think we've tried the optional analytics gathering yet! That script does exist, so I also learned something new today :) If you want to re-enable it in a follow-up PR, we can try again?

InessaPawson:
Sure, @stefanv! It would be great to explore additional data points about the NumPy documentation.

PR #29897, opened 2025-10-08
BUG: Fixup float16 conversion error path and add tests

seberg:
I had modified the code to not purely rely on PyErr_Occurred() for error propagation, but missed this conversion call. This also applied to float conversions actually.

Closes gh-29896


charris:
Thanks Sebastian.

PR #29900, opened 2025-10-08
ENH: Add registration for sorting loops using new ufunc convenience function

MaanasArora:
Follows #29737 to actually add the registration for the new-style sorts and argsorts. This is done using a `PyUFunc_AddLoopsFromSpecs` convenience function that can accept multiple ufunc names and arraymethod specs using slots, special-casing the `"sort"` and `"argsort"` names. The `sfloat` test is updated to use this registration.

Thanks for reviewing!

### DIFF
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -1886,6 +1886,34 @@ with the rest of the ArrayMethod API.
    the main ufunc registration function.  This adds a new implementation/loop
    to a ufunc.  It replaces `PyUFunc_RegisterLoopForType`.
 
+.. c:type:: PyUFunc_LoopSlot
+
+   Structure used to add multiple loops to ufuncs from ArrayMethod specs.
+   This is used in `PyUFunc_AddLoopsFromSpecs`.
+
+    .. c:struct:: PyUFunc_LoopSlot
+
+        .. c:member:: const char *name
+
+            The name of the ufunc to add the loop to.
+
+        .. c:member:: PyArrayMethod_Spec *spec
+
+            The ArrayMethod spec to use to create the loop.
+
+.. c:function:: int PyUFunc_AddLoopsFromSpecs( \
+                        PyUFunc_LoopSlot *slots)
+
+    .. versionadded:: 2.4
+
+    Add multiple loops to ufuncs from ArrayMethod specs. This also
+    handles the registration of methods for the ufunc-like functions
+    ``sort`` and ``argsort``. See :ref:`array-methods-sorting` for details.
+
+    The ``slots`` argument must be a  NULL-terminated array of
+    `PyUFunc_LoopSlot` (see above), which give the name of the
+    ufunc and spec needed to create the loop.
+
 .. c:function:: int PyUFunc_AddPromoter( \
                         PyObject *ufunc, PyObject *DType_tuple, PyObject *promoter)
 
@@ -2035,6 +2063,36 @@ code:
         loop_descrs[2] = PyArray_DescrFromType(NPY_FLOAT64);
         Py_INCREF(loop_descrs[2]);
     }
+
+.. _array-methods-sorting:
+
+Sorting and Argsorting
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Sorting and argsorting methods for dtypes can be registered using the
+ArrayMethod API. This is done by adding an ArrayMethod spec with the name
+``"sort"`` or ``"argsort"`` respectively.  The spec must have ``nin=1``
+and ``nout=1`` for both sort and argsort. Sorting is inplace, hence we
+enforce that ``data[0] == data[1]``. Argsorting returns a new array of
+indices, so the output must be of ``NPY_INTP`` type.
+
+The ``context`` passed to the loop contains the ``parameters`` field which
+for these operations is a ``PyArrayMethod_SortParameters *`` struct. This
+struct contains a ``flags`` field which is a bitwise OR of ``NPY_SORTKIND``
+values indicating the kind of sort to perform (that is, whether it is a
+stable and/or descending sort). If the strided loop depends on the flags,
+a good way to deal with this is to define :c:macro:`NPY_METH_get_loop`,
+and not set any of the other loop slots.
+
+.. c:struct:: PyArrayMethod_SortParameters
+
+    .. c:member:: NPY_SORTKIND flags
+
+        The flags passed to the sort operation. This is a bitwise OR of
+        ``NPY_SORTKIND`` values indicating the kind of sort to perform.
+
+These specs can be registered using :c:func:`PyUFunc_AddLoopsFromSpecs`
+along with other ufunc loops.
 
 API for calling array methods
 -----------------------------
--- a/numpy/_core/code_generators/cversions.txt
+++ b/numpy/_core/code_generators/cversions.txt
@@ -80,5 +80,7 @@
 0x00000013 = 2b8f1f4da822491ff030b2b37dff07e3
 # Version 20 (NumPy 2.3.0)
 0x00000014 = e56b74d32a934d085e7c3414cb9999b8,
-# Version 21 (NumPy 2.4.0) Add 'same_value' casting, header additions
-0x00000015 = e56b74d32a934d085e7c3414cb9999b8,
+# Version 21 (NumPy 2.4.0)
+# Add 'same_value' casting, header additions.
+# General loop registration for ufuncs, sort, and argsort
+0x00000015 = fbd24fc5b2ba4f7cd3606ec6128de7a5
--- a/numpy/_core/include/numpy/dtype_api.h
+++ b/numpy/_core/include/numpy/dtype_api.h
@@ -145,6 +145,13 @@ typedef struct {
 } PyArrayMethod_Spec;
 
 
+// This is used for the convenience function `PyUFunc_AddLoopsFromSpecs`
+typedef struct {
+    const char *name;
+    PyArrayMethod_Spec *spec;
+} PyUFunc_LoopSlot;
+
+
 /*
  * ArrayMethod slots
  * -----------------
--- a/numpy/_core/src/umath/_scaled_float_dtype.c
+++ b/numpy/_core/src/umath/_scaled_float_dtype.c
@@ -777,92 +777,6 @@ promote_to_sfloat(PyUFuncObject *NPY_UNUSED(ufunc),
 }
 
 
-/*
- * Add new ufunc loops (this is somewhat clumsy as of writing it, but should
- * get less so with the introduction of public API).
- */
-static int
-sfloat_init_ufuncs(void) {
-    PyArray_DTypeMeta *dtypes[3] = {
-            &PyArray_SFloatDType, &PyArray_SFloatDType, &PyArray_SFloatDType};
-    PyType_Slot slots[3] = {{0, NULL}};
-    PyArrayMethod_Spec spec = {
-        .nin = 2,
-        .nout =1,
-        .dtypes = dtypes,
-        .slots = slots,
-    };
-    spec.name = "sfloat_multiply";
-    spec.casting = NPY_NO_CASTING;
-
-    slots[0].slot = NPY_METH_resolve_descriptors;
-    slots[0].pfunc = &multiply_sfloats_resolve_descriptors;
-    slots[1].slot = NPY_METH_strided_loop;
-    slots[1].pfunc = &multiply_sfloats;
-    PyBoundArrayMethodObject *bmeth = PyArrayMethod_FromSpec_int(&spec, 0);
-    if (bmeth == NULL) {
-        return -1;
-    }
-    int res = sfloat_add_loop("multiply",
-            bmeth->dtypes, (PyObject *)bmeth->method);
-    Py_DECREF(bmeth);
-    if (res < 0) {
-        return -1;
-    }
-
-    spec.name = "sfloat_add";
-    spec.casting = NPY_SAME_KIND_CASTING;
-
-    slots[0].slot = NPY_METH_resolve_descriptors;
-    slots[0].pfunc = &add_sfloats_resolve_descriptors;
-    slots[1].slot = NPY_METH_strided_loop;
-    slots[1].pfunc = &add_sfloats;
-    bmeth = PyArrayMethod_FromSpec_int(&spec, 0);
-    if (bmeth == NULL) {
-        return -1;
-    }
-    res = sfloat_add_loop("add",
-            bmeth->dtypes, (PyObject *)bmeth->method);
-    Py_DECREF(bmeth);
-    if (res < 0) {
-        return -1;
-    }
-
-    /* N.B.: Wrapping isn't actually correct if scaling can be negative */
-    if (sfloat_add_wrapping_loop("hypot", dtypes) < 0) {
-        return -1;
-    }
-
-    /*
-     * Add a promoter for both directions of multiply with double.
-     */
-    PyArray_DTypeMeta *double_DType = &PyArray_DoubleDType;
-
-    PyArray_DTypeMeta *promoter_dtypes[3] = {
-            &PyArray_SFloatDType, double_DType, NULL};
-
-    PyObject *promoter = PyCapsule_New(
-            &promote_to_sfloat, "numpy._ufunc_promoter", NULL);
-    if (promoter == NULL) {
-        return -1;
-    }
-    res = sfloat_add_loop("multiply", promoter_dtypes, promoter);
-    if (res < 0) {
-        Py_DECREF(promoter);
-        return -1;
-    }
-    promoter_dtypes[0] = double_DType;
-    promoter_dtypes[1] = &PyArray_SFloatDType;
-    res = sfloat_add_loop("multiply", promoter_dtypes, promoter);
-    Py_DECREF(promoter);
-    if (res < 0) {
-        return -1;
-    }
-
-    return 0;
-}
-
-
 NPY_NO_EXPORT int
 sfloat_stable_sort_loop(
         PyArrayMethod_Context *context,
@@ -1044,53 +958,120 @@ sfloat_argsort_resolve_descriptors(
 }
 
 
+/*
+ * Add new ufunc loops (this is somewhat clumsy as of writing it, but should
+ * get less so with the introduction of public API).
+ */
 static int
-sfloat_init_sort(void)
-{
-    PyArray_DTypeMeta *dtypes[2] = {&PyArray_SFloatDType, &PyArray_SFloatDType};
-    PyType_Slot slots[3] = {
+sfloat_init_ufuncs(void) {
+    PyArray_DTypeMeta *all_sfloat_dtypes[3] = {
+            &PyArray_SFloatDType, &PyArray_SFloatDType, &PyArray_SFloatDType};
+    PyType_Slot multiply_slots[3] = {
+        {NPY_METH_resolve_descriptors, &multiply_sfloats_resolve_descriptors},
+        {NPY_METH_strided_loop, &multiply_sfloats},
+        {0, NULL}
+    };
+    PyArrayMethod_Spec multiply_spec = {
+        .nin = 2,
+        .nout = 1,
+        .dtypes = all_sfloat_dtypes,
+        .slots = multiply_slots,
+        .name = "sfloat_multiply",
+        .casting = NPY_NO_CASTING,
+    };
+
+    PyType_Slot add_slots[3] = {
+        {NPY_METH_resolve_descriptors, &add_sfloats_resolve_descriptors},
+        {NPY_METH_strided_loop, &add_sfloats},
+        {0, NULL}
+    };
+    PyArrayMethod_Spec add_spec = {
+        .nin = 2,
+        .nout = 1,
+        .dtypes = all_sfloat_dtypes,
+        .slots = add_slots,
+        .name = "sfloat_add",
+        .casting = NPY_SAME_KIND_CASTING,
+    };
+
+    PyArray_DTypeMeta *sort_dtypes[2] = {&PyArray_SFloatDType, &PyArray_SFloatDType};
+    PyType_Slot sort_slots[3] = {
         {NPY_METH_resolve_descriptors, &sfloat_sort_resolve_descriptors},
         {NPY_METH_get_loop, &sfloat_sort_get_loop},
         {0, NULL}
     };
-    PyArrayMethod_Spec spec = {
+    PyArrayMethod_Spec sort_spec = {
         .nin = 1,
         .nout = 1,
-        .dtypes = dtypes,
-        .slots = slots,
+        .dtypes = sort_dtypes,
+        .slots = sort_slots,
     };
-    spec.name = "sfloat_sort";
-    spec.casting = NPY_NO_CASTING;
-    spec.flags = NPY_METH_NO_FLOATINGPOINT_ERRORS;
+    sort_spec.name = "sfloat_sort";
+    sort_spec.casting = NPY_NO_CASTING;
+    sort_spec.flags = NPY_METH_NO_FLOATINGPOINT_ERRORS;
 
-    PyBoundArrayMethodObject *sort_meth = PyArrayMethod_FromSpec_int(&spec, 0);
-    if (sort_meth == NULL) {
+    PyArray_DTypeMeta *argsort_dtypes[2] = {&PyArray_SFloatDType, &PyArray_IntpDType};
+    PyType_Slot argsort_slots[3] = {
+        {NPY_METH_resolve_descriptors, &sfloat_argsort_resolve_descriptors},
+        {NPY_METH_get_loop, &sfloat_argsort_get_loop},
+        {0, NULL}
+    };
+    PyArrayMethod_Spec argsort_spec = {
+        .nin = 1,
+        .nout = 1,
+        .dtypes = argsort_dtypes,
+        .slots = argsort_slots,
+    };
+    argsort_spec.name = "sfloat_argsort";
+    argsort_spec.casting = NPY_NO_CASTING;
+    argsort_spec.flags = NPY_METH_NO_FLOATINGPOINT_ERRORS;
+
+    PyUFunc_LoopSlot loops[] = {
+        {"multiply", &multiply_spec},
+        {"add", &add_spec},
+        {"sort", &sort_spec},
+        {"argsort", &argsort_spec},
+        {NULL, NULL}
+    };
+    if (PyUFunc_AddLoopsFromSpecs(loops) < 0) {
         return -1;
     }
-    // TODO: once registration method is in place, use it instead of setting hidden slot
-    NPY_DT_SLOTS(&PyArray_SFloatDType)->sort_meth = sort_meth->method;
-    Py_INCREF(sort_meth->method);
-    Py_DECREF(sort_meth);
 
-    spec.name = "sfloat_argsort";
-    dtypes[1] = &PyArray_IntpDType;
-
-    slots[0].slot = NPY_METH_resolve_descriptors;
-    slots[0].pfunc = &sfloat_argsort_resolve_descriptors;
-    slots[1].slot = NPY_METH_get_loop;
-    slots[1].pfunc = &sfloat_argsort_get_loop;
-
-    // TODO: once registration method is in place, use it instead of setting hidden slot
-    PyBoundArrayMethodObject *argsort_meth = PyArrayMethod_FromSpec_int(&spec, 0);
-    if (argsort_meth == NULL) {
+    /* N.B.: Wrapping isn't actually correct if scaling can be negative */
+    if (sfloat_add_wrapping_loop("hypot", all_sfloat_dtypes) < 0) {
         return -1;
     }
-    NPY_DT_SLOTS(&PyArray_SFloatDType)->argsort_meth = argsort_meth->method;
-    Py_INCREF(argsort_meth->method);
-    Py_DECREF(argsort_meth);
+
+    /*
+     * Add a promoter for both directions of multiply with double.
+     */
+    int res = -1;
+    PyArray_DTypeMeta *double_DType = &PyArray_DoubleDType;
+
+    PyArray_DTypeMeta *promoter_dtypes[3] = {
+            &PyArray_SFloatDType, double_DType, NULL};
+
+    PyObject *promoter = PyCapsule_New(
+            &promote_to_sfloat, "numpy._ufunc_promoter", NULL);
+    if (promoter == NULL) {
+        return -1;
+    }
+    res = sfloat_add_loop("multiply", promoter_dtypes, promoter);
+    if (res < 0) {
+        Py_DECREF(promoter);
+        return -1;
+    }
+    promoter_dtypes[0] = double_DType;
+    promoter_dtypes[1] = &PyArray_SFloatDType;
+    res = sfloat_add_loop("multiply", promoter_dtypes, promoter);
+    Py_DECREF(promoter);
+    if (res < 0) {
+        return -1;
+    }
 
     return 0;
 }
+
 
 /*
  * Python entry point, exported via `umathmodule.h` and `multiarraymodule.c`.
@@ -1125,10 +1106,6 @@ get_sfloat_dtype(PyObject *NPY_UNUSED(mod), PyObject *NPY_UNUSED(args))
     }
 
     if (sfloat_init_ufuncs() < 0) {
-        return NULL;
-    }
-
-    if (sfloat_init_sort() < 0) {
         return NULL;
     }
 
--- a/numpy/_core/src/umath/dispatching.cpp
+++ b/numpy/_core/src/umath/dispatching.cpp
@@ -47,6 +47,7 @@
 
 #include "numpy/ndarraytypes.h"
 #include "numpy/npy_3kcompat.h"
+#include "npy_import.h"
 #include "common.h"
 #include "npy_pycompat.h"
 
@@ -190,6 +191,83 @@ PyUFunc_AddLoopFromSpec_int(PyObject *ufunc, PyArrayMethod_Spec *spec, int priv)
         return -1;
     }
     return PyUFunc_AddLoop((PyUFuncObject *)ufunc, info, 0);
+}
+
+
+/*UFUNC_API
+ * Add multiple loops to ufuncs from ArrayMethod specs. This also
+ * handles the registration of sort and argsort methods for dtypes
+ * from ArrayMethod specs.
+ */
+NPY_NO_EXPORT int
+PyUFunc_AddLoopsFromSpecs(PyUFunc_LoopSlot *slots)
+{
+    static PyObject *sort = NULL;
+    static PyObject *argsort = NULL;
+    if (sort == NULL) {
+        sort = npy_import("numpy", "sort");
+        if (sort == NULL) {
+            return -1;
+        }
+    }
+    if (argsort == NULL) {
+        argsort = npy_import("numpy", "argsort");
+        if (argsort == NULL) {
+            return -1;
+        }
+    }
+
+    PyUFunc_LoopSlot *slot;
+    for (slot = slots; slot->name != NULL; slot++) {
+        PyObject *ufunc = npy_import("numpy", slot->name);
+        if (ufunc == NULL) {
+            return -1;
+        }
+
+        if (ufunc == sort) {
+            Py_DECREF(ufunc);
+
+            PyArray_DTypeMeta *dtype = slot->spec->dtypes[0];
+            PyBoundArrayMethodObject *sort_meth = PyArrayMethod_FromSpec_int(slot->spec, 0);
+            if (sort_meth == NULL) {
+                PyErr_Format(PyExc_TypeError, "Failed to create sort method for %R", dtype);
+                return -1;
+            }
+
+            NPY_DT_SLOTS(dtype)->sort_meth = sort_meth->method;
+            Py_INCREF(sort_meth->method);
+            Py_DECREF(sort_meth);
+        }
+        else if (ufunc == argsort) {
+            Py_DECREF(ufunc);
+
+            PyArray_DTypeMeta *dtype = slot->spec->dtypes[0];
+            PyBoundArrayMethodObject *argsort_meth = PyArrayMethod_FromSpec_int(slot->spec, 0);
+            if (argsort_meth == NULL) {
+                PyErr_Format(PyExc_TypeError, "Failed to create argsort method for %R", dtype);
+                return -1;
+            }
+
+            NPY_DT_SLOTS(dtype)->argsort_meth = argsort_meth->method;
+            Py_INCREF(argsort_meth->method);
+            Py_DECREF(argsort_meth);
+        }
+        else {
+            if (!PyObject_TypeCheck(ufunc, &PyUFunc_Type)) {
+                PyErr_Format(PyExc_TypeError, "%s was not a ufunc!", slot->name);
+                Py_DECREF(ufunc);
+                return -1;
+            }
+
+            int ret = PyUFunc_AddLoopFromSpec_int(ufunc, slot->spec, 0);
+            Py_DECREF(ufunc);
+            if (ret < 0) {
+                return -1;
+            }
+        }
+    }
+
+    return 0;
 }
 
 
--- a/numpy/_core/src/umath/dispatching.cpp
+++ b/numpy/_core/src/umath/dispatching.cpp
@@ -224,7 +224,6 @@ PyUFunc_AddLoopsFromSpecs(PyUFunc_LoopSlot *slots)
             PyArray_DTypeMeta *dtype = slot->spec->dtypes[0];
             PyBoundArrayMethodObject *sort_meth = PyArrayMethod_FromSpec_int(slot->spec, 0);
             if (sort_meth == NULL) {
-                PyErr_Format(PyExc_TypeError, "Failed to create sort method for %R", dtype);
                 return -1;
             }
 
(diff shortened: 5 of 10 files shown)


doc/source/reference/c-api/array.rst
```diff
@@ -2036,6 +2066,42 @@ code:
         Py_INCREF(loop_descrs[2]);
     }
 
+.. _array-methods-sorting:
+
+Sorting and Argsorting
```
MaanasArora, line 2069: Not sure if this file is the best place for this doc?

doc/source/reference/c-api/array.rst
```diff
@@ -2036,6 +2066,42 @@ code:
...
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Sorting and argsorting methods for dtypes can be registered using the
+ArrayMethod API. This is done by adding an ArrayMethod spec with the name
+``"sort"`` or ``"argsort"`` respectively.  The spec must have ``nin=1``
+and ``nout=0`` for sort and ``nin=1`` and ``nout=1`` for argsort. Sorting is
```
mhvk: I think it has to have `nout=1` - there is one output, it is just inplace (as you write below).

doc/source/reference/c-api/array.rst
```diff
@@ -2036,6 +2066,42 @@ code:
...
+inplace, however we enforce that `data[0] == data[1]`. Argsorting returns a new
+array of indices, so the output must be of ``NPY_INTP`` type.
+
+The spec must define the `get_strided_loop` slot which returns a
+``PyArrayMethod_StridedLoop`` implementation. However, strides are currently
+ignored and the loops are always called with contiguous data. The spec can
+also define the `resolve_descriptors` slot if needed.
+
+The `context` passed to the loop contains the `parameters` field which is for
```
mhvk: delete "is"

doc/source/reference/c-api/array.rst
```diff
@@ -2036,6 +2066,42 @@ code:
...
+these operations is a `NpySortParameters` struct. This struct contains the
+boolean fields `stable` and `descending` to indicate whether the sort should be
```
mhvk: Why is this changed? Just pass the sort flags as we had defined! (which are discussed in the C api).

doc/source/reference/c-api/array.rst
mhvk: In a way, it is a bit illogical that this is in the array and not in the ufunc section, but then that's true for the preceding information as well -- `ufuncs.rst` seems to refer mostly to the old technique.

Anyway, somewhat out of scope here, maybe @seberg has a better suggestions.

numpy/_core/code_generators/cversions.txt
```diff
@@ -80,5 +80,7 @@
 0x00000013 = 2b8f1f4da822491ff030b2b37dff07e3
 # Version 20 (NumPy 2.3.0)
 0x00000014 = e56b74d32a934d085e7c3414cb9999b8,
-# Version 21 (NumPy 2.4.0) Add 'same_value' casting, header additions
-0x00000015 = e56b74d32a934d085e7c3414cb9999b8,
+# Version 21 (NumPy 2.4.0)
+# Add 'same_value' casting, header additions.
+# Sort and argsort array method registration.
```
mhvk: This would be `# General loop registration for ufuncs, sort, and argsort`

doc/source/reference/c-api/array.rst
```diff
@@ -1886,6 +1886,36 @@ with the rest of the ArrayMethod API.
...
+.. c:type:: PyUFunc_LoopSlot
+
+   Structure used to add multiple loops to ufuncs from ArrayMethod specs.
+   This is used in `PyUFunc_AddLoopsFromSpecs`.
+
+    .. c:struct:: PyUFunc_LoopSlot
+
+        .. c:member:: const char *ufunc_name
+
+            The name of the ufunc to add the loop to.
```
mhvk, line 1898: This is slightly vague. Where is this name found? Could one register for a ufunc defined outside numpy? I think by default it should be looked up on `np`, but one should be able to give a name including package. I guess one can test this using some of the private test ufuncs.

numpy/_core/src/umath/_scaled_float_dtype.c
```diff
@@ -1047,49 +1047,45 @@ sfloat_argsort_resolve_descriptors(
 static int
 sfloat_init_sort(void)
 {
-    PyArray_DTypeMeta *dtypes[2] = {&PyArray_SFloatDType, &PyArray_SFloatDType};
-    PyType_Slot slots[3] = {
+    PyArray_DTypeMeta *sort_dtypes[2] = {&PyArray_SFloatDType, &PyArray_SFloatDType};
```
mhvk, line 997: I think we should now combine this with `sfloat_init_ufuncs`, to show you can do all of them in one go!

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -193,6 +193,76 @@ PyUFunc_AddLoopFromSpec_int(PyObject *ufunc, PyArrayMethod_Spec *spec, int priv)
...
+                return -1;
+            }
+
+            NPY_DT_SLOTS(dtype)->argsort_meth = argsort_meth->method;
+            Py_INCREF(argsort_meth->method);
+            Py_DECREF(argsort_meth);
+        }
+        else {
+            PyObject *ufunc =
+                PyObject_GetAttrString(PyImport_ImportModule("numpy"), slot->ufunc_name);
```
mhvk: So, this is the part where I think if it fails, one perhaps should have a more general import - surely, there is something available that can help importing a `module.submodule.function` string...

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -193,6 +193,76 @@ PyUFunc_AddLoopFromSpec_int(PyObject *ufunc, PyArrayMethod_Spec *spec, int priv)
...
+            if (ufunc == NULL) {
+                PyErr_Format(PyExc_TypeError,
+                        "numpy.%s is not a ufunc!", slot->ufunc_name);
```
mhvk: I think here an `AttributeError` has already been set, so you can just return.

numpy/_core/src/umath/dispatching.cpp
mhvk: Aside: I think we have some neat shortcuts with static variables to avoid having to re-import numpy every time.

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -193,6 +193,76 @@ PyUFunc_AddLoopFromSpec_int(PyObject *ufunc, PyArrayMethod_Spec *spec, int priv)
...
+ * Add multiple loops to ufuncs from ArrayMethod specs. This also
+ * handles the registration of sort and argsort methods for dtypes
+ * from ArrayMethod specs.
+ */
+NPY_NO_EXPORT int
+PyUFunc_AddLoopsFromSpecs(PyUFunc_LoopSlot *slots)
+{
+    PyUFunc_LoopSlot *slot;
+    for (slot = slots; slot->ufunc_name != NULL; slot++) {
+        if (strcmp(slot->ufunc_name, "sort") == 0) {
```
mhvk: It may not be completely crazy to do the test after getting the attribute from `numpy` - i.e., check that it is `array_sort`. This means that if someone passes in `numpy.sort`, it would still get found. Note that I'm not sure you have access to `array_sort` here, so for now this is probably simply overthinking it!

mhvk (commented):
I like it! Mostly looks good  -- probably the biggest comment I have is that the sort context contains flags rather than bools... The rest is minor, with some just being extensions that could be added later too.

Curious to see what @seberg thinks - it seems a nice implementation of his idea!

doc/source/reference/c-api/array.rst
MaanasArora: Yes sorry, this is fixed, thanks!

doc/source/reference/c-api/array.rst
MaanasArora: Done, thanks!

doc/source/reference/c-api/array.rst
MaanasArora: In hindsight, I have no idea... sorry! It seems I somehow thought we were using the older PR. This is fixed, thanks!

numpy/_core/code_generators/cversions.txt
MaanasArora: Done, thanks!

numpy/_core/src/umath/dispatching.cpp
MaanasArora: Done.

numpy/_core/src/umath/_scaled_float_dtype.c
MaanasArora: Yes, makes sense, I'll do this now!

MaanasArora: This is done, thanks!

numpy/_core/src/umath/dispatching.cpp
MaanasArora: Yes, that could be nice. I'm looking at this, but for the static variables: I actually couldn't find a place where the ufuncs are looked up using a static module variable, instead, it seems the module `numpy` is passed as an argument into those functions?

doc/source/reference/c-api/array.rst
seberg: We could do this, I think.  The logic would be to allow for a `:`, if given, you import the part before the `:` and then use `GetAttr` for the part after (in principle should support `.` in the second part too).
It might be that the import mechanism can already do both in one go, but I am not sure.  If it is tedious in C, could create a little Python helper to do it since it is very easy there, e.g. something like [this](https://github.com/scientific-python/spatch/blob/551f99766d90db0e0e8d3ad83efa8e528c1af163/src/spatch/utils.py#L14-L19)

And if there is no `:` we would just go with `numpy` (and maybe the internal ufunc module as well to robustify it).

I am wondering if it makes sense to provision for when the object isn't found (or not a ufunc).  We could for example allow for `"?numpy.strings:..."` to make it easier to support across versions.

doc/source/reference/c-api/array.rst
seberg: Yeah, I guess all the newer stuff ended up here.  Similar issue with the finfo PR.
We should split up this file, if just because it is way too big, but I agree it is out of scope.

doc/source/reference/c-api/array.rst
```diff
@@ -2036,6 +2066,39 @@ code:
...
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Sorting and argsorting methods for dtypes can be registered using the
+ArrayMethod API. This is done by adding an ArrayMethod spec with the name
+``"sort"`` or ``"argsort"`` respectively.  The spec must have ``nin=1``
+and ``nout=1`` for both sort and argsort. Sorting is inplace, however we
+enforce that `data[0] == data[1]`. Argsorting returns a new array of
+indices, so the output must be of ``NPY_INTP`` type.
+
+The spec must define the `get_strided_loop` slot which returns a
```
seberg: Technically, it may not have to (unless we add an extra check), but I let's lie :).

doc/source/reference/c-api/array.rst
```diff
@@ -2036,6 +2066,39 @@ code:
...
+``PyArrayMethod_StridedLoop`` implementation. However, strides are currently
+ignored and the loops are always called with contiguous data. The spec can
+also define the `resolve_descriptors` slot if needed.
+
+The `context` passed to the loop contains the `parameters` field which for
+these operations is a `PyArrayMethod_SortParameters` struct. This struct
```
seberg:
```suggestion
these operations is a `PyArrayMethod_SortParameters *` struct. This struct
```
(Kinda clear if you read what `parmaters` is, but maybe the `*` is good as a reminder.)

numpy/_core/src/umath/_scaled_float_dtype.c
```diff
@@ -1044,54 +958,123 @@ sfloat_argsort_resolve_descriptors(
...
+        .nin = 1,
+        .nout = 1,
+        .dtypes = argsort_dtypes,
+        .slots = argsort_slots,
+    };
+    argsort_spec.name = "sfloat_argsort";
+    argsort_spec.casting = NPY_NO_CASTING;
+    argsort_spec.flags = NPY_METH_NO_FLOATINGPOINT_ERRORS;
+
+    PyUFunc_LoopSlot loops[5] = {
```
seberg:
```suggestion
    PyUFunc_LoopSlot loops[] = {
```

numpy/_core/src/umath/_scaled_float_dtype.c
```diff
@@ -1044,54 +958,123 @@ sfloat_argsort_resolve_descriptors(
...
-    NPY_DT_SLOTS(&PyArray_SFloatDType)->sort_meth = sort_meth->method;
-    Py_INCREF(sort_meth->method);
-    Py_DECREF(sort_meth);
 
-    spec.name = "sfloat_argsort";
-    dtypes[1] = &PyArray_IntpDType;
+    /*
+     * Add a promoter for both directions of multiply with double.
+     */
+    int res;
```
seberg: NIT: would move to first use, or init as `-1`.

numpy/_core/src/umath/dispatching.cpp
seberg: Either way works... that way we need to import and cache `numpy.sort` before this can be called (which is fine).  But overall, I don't think it matters.

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -193,6 +193,74 @@ PyUFunc_AddLoopFromSpec_int(PyObject *ufunc, PyArrayMethod_Spec *spec, int priv)
...
+        else {
+            PyObject *ufunc =
+                PyObject_GetAttrString(PyImport_ImportModule("numpy"), slot->ufunc_name);
+            if (ufunc == NULL) {
+                return -1;
+            }
+
+            if (!PyObject_TypeCheck(ufunc, &PyUFunc_Type)) {
+                PyErr_Format(PyExc_TypeError,
+                        "numpy.%s was not a ufunc!", slot->ufunc_name);
```
seberg:
```suggestion
                        "%s was not a ufunc!", slot->ufunc_name);
```
then if we support more, it just works out ;).

numpy/_core/src/umath/dispatching.cpp
seberg: We have `npy_cache_import_runtime` you can use for NumPy (if we support non-numpy can't cache those of course).
(For more general, see first comment, I think it should be `module.submodule:obj.attr` and if that tedious we can even do it in Python since I'll assume this doesn't matter speed wise -- there may be C helpers in the Python API, though)

seberg (commented):
Nice, thanks!  I would be happy to extend to outside NumPy as this should really be a convenience API for the biggest part.

numpy/_core/src/umath/dispatching.cpp
mhvk: I was indeed thinking about `npy_import` and `npy_cache_import_runtime`, both in `src/common/npy_import.h`
Since nothing here is performance critical, in this particular instance, you could just replace this with 
```
PyObject *ufunc = npy_import("numpy", slot->ufunc_name);
```

One might then use `npy_cache_import_nuntime("numpy", "sort")` (and `argsort`) somewhere on top and do a straight comparison of those here.

doc/source/reference/c-api/array.rst
mhvk: I keep thinking about nice ways to expand, but the more I do so, the more it is clear that we should not go there now! 

So, just a small comment: I would call this entry plain "name" -- it is already in the "UFunc_LoopSlot" namespace, so adding `ufunc_` seems superfluous.

doc/source/reference/c-api/array.rst
```diff
@@ -1886,6 +1886,36 @@ with the rest of the ArrayMethod API.
...
+
+        .. c:member:: PyArrayMethod_Spec *spec
+
+            The ArrayMethod spec to use to create the loop.
+
+.. c:function:: int PyUFunc_AddLoopsFromSpecs( \
+                        PyUFunc_LoopSlot *slots)
+
+   Add multiple loops to ufuncs from ArrayMethod specs. This also
+   handles the registration of sort and argsort methods for dtypes
```
mhvk: I think this gives a bit more detail than necessary. How about shortening this to
```
Add multiple loops to ufuncs from ArrayMethod specs. This also
handles the registration of methods for the ufunc-like functions
``sort`` and ``argsort``. See :ref:`array-methods-sorting` for details.
```

And then delete the sentence at l.1916.

doc/source/reference/c-api/array.rst
```diff
@@ -1886,6 +1886,36 @@ with the rest of the ArrayMethod API.
...
+   from ArrayMethod specs.
+
+   To register a loop for a ufunc, the ``slots`` argument must be a
```
mhvk: Again shorten a little, perhaps as follows:
```
The ``slots`` argument must be a  NULL-terminated array of
`PyUFunc_LoopSlot` (see above), which give the name of the
ufunc and spec needed to create the loop.
```

doc/source/reference/c-api/array.rst
```diff
@@ -2036,6 +2066,39 @@ code:
...
 
+.. _array-methods-sorting:
+
+Sorting and Argsorting
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Sorting and argsorting methods for dtypes can be registered using the
+ArrayMethod API. This is done by adding an ArrayMethod spec with the name
+``"sort"`` or ``"argsort"`` respectively.  The spec must have ``nin=1``
+and ``nout=1`` for both sort and argsort. Sorting is inplace, however we
```
mhvk: "... Sorting is inplace, hence we"

doc/source/reference/c-api/array.rst
mhvk: Do you mean that there is a default that will just look up the loop in the other slots? I do think it would be good to treat this as much as a regular ufunc array method as possible. In that respect, the only important bits may be that for `sort`, inplace is expected, and that one can expect aligned, contiguous data.  But if it is fine, therefore, to just define `contiguous_loop`, is there anything wrong with saying so?

doc/source/reference/c-api/array.rst
```diff
@@ -2036,6 +2066,39 @@ code:
...
+The spec must define the `get_strided_loop` slot which returns a
+``PyArrayMethod_StridedLoop`` implementation. However, strides are currently
+ignored and the loops are always called with contiguous data. The spec can
+also define the `resolve_descriptors` slot if needed.
+
+The `context` passed to the loop contains the `parameters` field which for
+these operations is a `PyArrayMethod_SortParameters` struct. This struct
+contains a `flags` field which is a bitwise OR of `NPY_SORTKIND` values
+indicating the kind of sort to perform (that is, whether it is a stable
+and/or descending sort).
```
mhvk: Since this is probably the reason to define `get_strided_loop`, maybe it would be good to keep the above shorter and here add "If the strided loop depends on these flags, a good way to deal with that is to define `get_strided_loop`, and not set any of the ``PyArrayMethod_StridedLoop``.

numpy/_core/include/numpy/dtype_api.h
```diff
@@ -145,6 +145,13 @@ typedef struct {
 } PyArrayMethod_Spec;
 
 
+// This is used for the convenience function `PyUFunc_AddLoopsFromSpecs`
+typedef struct {
+    const char *ufunc_name;
```
mhvk: So, I'd suggest simple `name` here.

mhvk (commented):
Review of just the documentation - will continue later, so perhaps wait with pushing changes...

doc/source/reference/c-api/array.rst
seberg: Yes, you could just a `strided_loop` and that will be picked and could contain the `parameter` logic there rather than inside `get_loop`.

mhvk: So, I think we probably should just delete this, then.

If we really do want to advertise that the array is going to be contiguous, one could do so at the end of l2079:
```
...type. For both
``sort`` and ``argsort``, the data provided is contiguous.
```
But do we want to advertize that?

numpy/_core/src/umath/_scaled_float_dtype.c
```diff
@@ -1044,54 +958,123 @@ sfloat_argsort_resolve_descriptors(
...
+    PyArrayMethod_Spec multiply_spec = {
+        .nin = 2,
+        .nout = 1,
+        .dtypes = multiply_dtypes,
+        .slots = multiply_slots,
+        .name = "sfloat_multiply",
+        .casting = NPY_NO_CASTING,
+    };
+
+    PyArray_DTypeMeta *add_dtypes[3] = {
```
mhvk: Not really your PR, but partially for my understanding; I think below one could just use `multiply_dtypes` again, since it is identical. Indeed, one could also use it for `sort` (but not `argsort`). Might it make sense to just define these as `all_sfloat_dtypes[3]` and use them where possible?

numpy/_core/src/umath/_scaled_float_dtype.c
mhvk: Indeed, might as well let the compiler do the work...

numpy/_core/src/umath/_scaled_float_dtype.c
```diff
@@ -1044,54 +958,123 @@ sfloat_argsort_resolve_descriptors(
...
+        {NULL, NULL}
+    };
+    if (PyUFunc_AddLoopsFromSpecs(loops) < 0) {
+        return -1;
+    }
 
-    PyBoundArrayMethodObject *sort_meth = PyArrayMethod_FromSpec_int(&spec, 0);
-    if (sort_meth == NULL) {
+    /* N.B.: Wrapping isn't actually correct if scaling can be negative */
+    if (sfloat_add_wrapping_loop("hypot", multiply_dtypes) < 0) {
```
mhvk: Hmm, would it be possible to do this with the new registration too? It would seem something that might be used a lot. Though maybe the easiest is to then define a new `PyUFunc_AddWrappingLoopsFromSpecs`. Anyway, would seem absolutely fine to leave this for later! (It is nice in that it would remove the need for `sfloat_get_ufunc`...)

p.s. Here `multiply_dtypes` is re-used, so I think it indeed can be done elsewhere too.

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -193,6 +193,74 @@ PyUFunc_AddLoopFromSpec_int(PyObject *ufunc, PyArrayMethod_Spec *spec, int priv)
...
+{
+    PyUFunc_LoopSlot *slot;
+    for (slot = slots; slot->ufunc_name != NULL; slot++) {
+        if (strcmp(slot->ufunc_name, "sort") == 0) {
+            PyArray_DTypeMeta *dtype = slot->spec->dtypes[0];
+
+            PyBoundArrayMethodObject *sort_meth = PyArrayMethod_FromSpec_int(slot->spec, 0);
+            if (sort_meth == NULL) {
+                PyErr_Format(PyExc_TypeError,
+                        "Failed to create sort method for %R",
```
mhvk: Absolute nitpick, but does this not fit on one line? (logic: error paths should take up the least space possible to minimize distraction)

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -193,6 +193,74 @@ PyUFunc_AddLoopFromSpec_int(PyObject *ufunc, PyArrayMethod_Spec *spec, int priv)
...
+            if (!PyObject_TypeCheck(ufunc, &PyUFunc_Type)) {
+                PyErr_Format(PyExc_TypeError,
+                        "numpy.%s was not a ufunc!", slot->ufunc_name);
+                Py_DECREF(ufunc);
+                return -1;
+            }
+
+            int ret = PyUFunc_AddLoopFromSpec_int(ufunc, slot->spec, 0);
+            if (ret < 0) {
+                Py_DECREF(ufunc);
```
mhvk: Move this before the `if` so that you can remove the one on l.256 as well. In fact, you could just fall through altogether, but perhaps that's too risky...

mhvk (commented):
OK, second half of second round: only small things, I think this is working out very neatly!

My only slightly more major comment/question is about registering wrapping loops, but I ended up concluding those are probably better done in a separate function. @seberg, does that make sense? It certainly does not seem to have to hold up this, unless we think it would affect the layout of the slots.

doc/source/reference/c-api/array.rst
MaanasArora: Makes sense - I agree adding more detail could be a bit broad at this time maybe! I've changed the `ufunc_name` to `name`, thanks!

numpy/_core/src/umath/dispatching.cpp
MaanasArora: Ah thanks! I've used `npy_import` on top, and then compared cached imports with the ufunc.

Edit: this means we will cache in the `npy_runtime_imports` struct as is done now?

doc/source/reference/c-api/array.rst
MaanasArora: Hmm, not sure. It could make it easier to support strides in the future if people actually write for them, so maybe not?

numpy/_core/src/umath/_scaled_float_dtype.c
MaanasArora: Done.

numpy/_core/src/umath/dispatching.cpp
MaanasArora: This is done, thanks!

doc/source/reference/c-api/array.rst
MaanasArora: Done, thanks!

doc/source/reference/c-api/array.rst
MaanasArora: Done as well.

doc/source/reference/c-api/array.rst
MaanasArora: Done.

doc/source/reference/c-api/array.rst
MaanasArora: Yes, I removed the previous paragraph and added this note.

numpy/_core/include/numpy/dtype_api.h
MaanasArora: Done as above.

numpy/_core/src/umath/dispatching.cpp
MaanasArora: This is fixed!

numpy/_core/src/umath/dispatching.cpp
MaanasArora: Did that, thanks!

numpy/_core/src/umath/_scaled_float_dtype.c
MaanasArora: It does seem to work - pushed! Actually, it seems to have been reused before when the spec was also being reused.

numpy/_core/src/umath/_scaled_float_dtype.c
MaanasArora: Hmm, I suspect that would indeed change the structure of the slots, so should consider. In particular for wrapping loops we need the dtype (and descriptor translate functions) and not the spec as far as I can tell. So, it seems there is not much common, maybe defining a new function would be best?

numpy/_core/src/umath/dispatching.cpp
MaanasArora: Hmm, the windows tests seem to be failing because we haven't imported numpy before we call this function - I've reverted to using `npy_import` for now for the comparisons. Figuring how to import (and cache) these before call - if needed?

numpy/_core/src/umath/_scaled_float_dtype.c
mhvk: Yeah, I was also going in that direction, but will defer to @seberg.
Anyway, I think it is basically out of scope here!

doc/source/reference/c-api/array.rst
```diff
@@ -2036,6 +2062,36 @@ code:
...
+and ``nout=1`` for both sort and argsort. Sorting is inplace, hence we
+enforce that `data[0] == data[1]`. Argsorting returns a new array of
+indices, so the output must be of ``NPY_INTP`` type.
+
+The `context` passed to the loop contains the `parameters` field which for
+these operations is a `PyArrayMethod_SortParameters *` struct. This struct
+contains a `flags` field which is a bitwise OR of `NPY_SORTKIND` values
+indicating the kind of sort to perform (that is, whether it is a stable
+and/or descending sort). If the strided loop depends on the flags, a good
+way to deal with this is to define `get_strided_loop`, and not set any of
```
mhvk: Maybe check that single quotes are OK, by doing a docs build and seeing if this properly refers to a location. I don't think so, it may have to be `NPY_METH_get_loop`?

numpy/_core/src/umath/_scaled_float_dtype.c
```diff
@@ -1044,54 +958,121 @@ sfloat_argsort_resolve_descriptors(
...
         {NPY_METH_get_loop, &sfloat_sort_get_loop},
         {0, NULL}
     };
-    PyArrayMethod_Spec spec = {
+    PyArrayMethod_Spec sort_spec = {
         .nin = 1,
         .nout = 1,
-        .dtypes = dtypes,
-        .slots = slots,
+        .dtypes = sort_dtypes,
```
mhvk, line 1006: Here you could also use `all_sfloat_dtypes` (it is overly long, but the code cannot care - it just expects `nin+nout`).

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -208,8 +209,11 @@ PyUFunc_AddLoopsFromSpecs(PyUFunc_LoopSlot *slots)
             return -1;
         }
 
-        npy_cache_import_runtime("numpy", "sort", &npy_runtime_imports._sort);
-        npy_cache_import_runtime("numpy", "argsort", &npy_runtime_imports._argsort);
+        if (npy_cache_import_runtime("numpy", "sort", &npy_runtime_imports._sort) < 0 ||
```
mhvk: Weird that this did not work. It looks good, though it should be outside the loop, and the object should not have an underscore (at least, that seems to be the habit, to just use the name).

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -193,6 +194,75 @@ PyUFunc_AddLoopFromSpec_int(PyObject *ufunc, PyArrayMethod_Spec *spec, int priv)
...
+PyUFunc_AddLoopsFromSpecs(PyUFunc_LoopSlot *slots)
+{
+    PyUFunc_LoopSlot *slot;
+    for (slot = slots; slot->name != NULL; slot++) {
+        PyObject *ufunc = npy_import("numpy", slot->name);
+        if (ufunc == NULL) {
+            return -1;
+        }
+
+        PyObject *sort = npy_import("numpy", "sort");
```
mhvk: So, we can still do this the inelegant way (but really, runtime import should work!). Outside of the loop,
```
static PyObject *sort = NULL;
static PyObject *argsort = NULL;
if (sort == NULL) {
    sort = npy_import("numpy", "sort");
    if (sort == NULL) {
        return -1;
    }
if (argsort == NULL) {
    argsort = npy_import("numpy", "argsort");
    if (argsort == NULL) {
        return -1;
    }
}
```

But I'm really confused why what you had failed (hard to believe it had to do with being inside the loop), i.e., 
```
if (npy_cache_import_runtime("numpy", "sort", &npy_runtime_imports.sort) == -1
    || npy_cache_import_runtime("numpy", "argsort", &npy_runtime_imports.argsort) == -1) {
    return NULL;
}
```

mhvk (commented):
Basically, looks great. Two confused comments, one to do only if there is another round. Maybe @seberg can advice about the interned objects; I don't understand why the windows build failed, but don't have a windows machine, so cannot test it.

p.s. Really not too important, but I usually try to avoid too many pushes, as it bugs me that every time the tests are run 70 times...

doc/source/reference/c-api/array.rst
MaanasArora: Yes, there were issues with references and code formatting - I did a local build and changed things so they look right (hopefully), thanks!

numpy/_core/src/umath/_scaled_float_dtype.c
MaanasArora: Hmm, would that be less clear, since we're creating dtypes for argsort anyway?

numpy/_core/src/umath/dispatching.cpp
MaanasArora: Yes, it's quite strange - though yes it should've been outside the loop! Could be the issue maybe, given it only failed in windows, maybe hit some corner case. But in the error, the `npy_runtime_imports` struct itself was not found...

numpy/_core/src/umath/dispatching.cpp
MaanasArora: I used that code block, thanks! I'll try experimenting with the cached imports more.

numpy/_core/src/umath/dispatching.cpp
mhvk: Yes, I noticed that. I just cannot understand how things can work everywhere else if it was a problem here...

numpy/_core/src/umath/_scaled_float_dtype.c
mhvk: It's fine to keep as is!

numpy/_core/src/umath/dispatching.cpp
mhvk: Should have commented here: Yes, I noticed that the windows build failed to find it. I just cannot understand how things can work everywhere else if it was a problem here...

numpy/_core/src/umath/_scaled_float_dtype.c
seberg: Yeah, let's not go there.  Seeing that C-API module init PEP, I wish I had thought of just not using a struct at all and putting _everything_ into slots, then this would be easy to do reasonably...

But I think a reasonable solution will be to just add the wrapping functions as slots.  Then if those slots are found you might even ignore some entries of the original struct.  (Or at least allow for example `nin = nout = 0.)

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -193,6 +194,81 @@ PyUFunc_AddLoopFromSpec_int(PyObject *ufunc, PyArrayMethod_Spec *spec, int priv)
...
+ * handles the registration of sort and argsort methods for dtypes
+ * from ArrayMethod specs.
+ */
+NPY_NO_EXPORT int
+PyUFunc_AddLoopsFromSpecs(PyUFunc_LoopSlot *slots)
+{
+    static PyObject *sort = NULL;
+    static PyObject *argsort = NULL;
+    if (sort == NULL) {
+        sort = npy_import("numpy", "sort");
```
seberg: If you do it here, please use the `cache_import_runtime`

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -193,6 +194,81 @@ PyUFunc_AddLoopFromSpec_int(PyObject *ufunc, PyArrayMethod_Spec *spec, int priv)
...
+            NPY_DT_SLOTS(dtype)->sort_meth = sort_meth->method;
+            Py_INCREF(sort_meth->method);
+            Py_DECREF(sort_meth);
+        }
+        else if (ufunc == argsort) {
+            PyArray_DTypeMeta *dtype = slot->spec->dtypes[0];
+
+            PyBoundArrayMethodObject *argsort_meth = PyArrayMethod_FromSpec_int(slot->spec, 0);
+            if (argsort_meth == NULL) {
+                PyErr_Format(PyExc_TypeError, "Failed to create argsort method for %R", dtype);
```
seberg: Doesn't matter, but FYI: this probably eats an in-flight exception.  But chaining is a bit tedious to repeat multiple times IMO.

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -193,6 +194,81 @@ PyUFunc_AddLoopFromSpec_int(PyObject *ufunc, PyArrayMethod_Spec *spec, int priv)
...
+            if (sort_meth == NULL) {
+                PyErr_Format(PyExc_TypeError, "Failed to create sort method for %R", dtype);
+                return -1;
+            }
+
+            NPY_DT_SLOTS(dtype)->sort_meth = sort_meth->method;
+            Py_INCREF(sort_meth->method);
+            Py_DECREF(sort_meth);
+        }
+        else if (ufunc == argsort) {
```
seberg: Should decref ufunc here and above also.

doc/source/reference/c-api/array.rst
seberg: I am happy to not do it here, but I am curious about your reasoning :).  I suppose it may be weird enough to support for non-numpy and honestly, writing this helper isn't hard (although this type of convenience seems to make more of a dent than I tend to expect).

Anyway, one thing we don't yet support, but happy to also defer is adding ufuncs that are _not_ available in the main namespace.

doc/source/reference/c-api/array.rst
```diff
@@ -1886,6 +1886,32 @@ with the rest of the ArrayMethod API.
...
+.. c:function:: int PyUFunc_AddLoopsFromSpecs( \
+                        PyUFunc_LoopSlot *slots)
+
+    Add multiple loops to ufuncs from ArrayMethod specs. This also
+    handles the registration of methods for the ufunc-like functions
+    ``sort`` and ``argsort``. See :ref:`array-methods-sorting` for details.
+
+    The ``slots`` argument must be a  NULL-terminated array of
+    `PyUFunc_LoopSlot` (see above), which give the name of the
+    ufunc and spec needed to create the loop.
```
seberg, line 1915: Would be good to add a `.. versionadded:: 2.4`

seberg (approved):
FWIW, I am happy with this thanks.  I think adding this convenience functino should be pretty uncontroversial API wise, and it is flexible enough to deprecate, etc.

Just a few nits.  I can see extending at least for `linalg.<...>` or `numpy.strings.<...>` but maybe no need to do it here and now.

numpy/_core/src/umath/dispatching.cpp
MaanasArora: Did this - FYI using the runtime struct was causing errors on Windows, not sure how the CI would end up, but hopefully what is there now is reasonable!

numpy/_core/src/umath/dispatching.cpp
MaanasArora: Done, thanks!

doc/source/reference/c-api/array.rst
MaanasArora: Done.

numpy/_core/src/umath/dispatching.cpp
MaanasArora: It seems it errors out even if we don't actually cache (in the struct)! See [this commit](https://github.com/numpy/numpy/pull/29900/files/cdcf12336f7c2fcfffbde602bff96dd1da4ad876..fdfc92f726b587efe881982256a1b7a0ca14dee5#diff-b0dfb4a9897d914cc202c0b4cadd6f7b79d1afa53d88ae7148d1ee9da1790986) also, @mhvk was confused about it too.

MaanasArora: Sorry for the many pushes, reverted this to fix the CI until we figure this out.

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -193,6 +194,83 @@ PyUFunc_AddLoopFromSpec_int(PyObject *ufunc, PyArrayMethod_Spec *spec, int priv)
...
+    static PyObject *sort = NULL;
+    static PyObject *argsort = NULL;
+    if (sort == NULL) {
+        sort = npy_import("numpy", "sort");
+        if (sort == NULL) {
+            return -1;
+        }
+    }
+    if (argsort == NULL) {
+        argsort = npy_import("numpy", "argsort");
```
seberg: Nit, but I can push it: Very sure the function includes the NULL check (no need to indent).

seberg:
I don't see this as controversial API since it adds a pretty straightforward convenience mostly. So planning to merge soon (if anyone has concers even after ee can follow up, of course!).

We do need to add a release note (or two, one for sorting), though.

numpy/_core/src/umath/dispatching.cpp
seberg: I saw later that you flipped back and forth here.  I fixed it, the problem was that this is a cpp file and the header was missing the `extern C`.

MaanasArora: Ah, that makes sense, thank you!

MaanasArora:
Thanks for reviewing! Yes, I suppose the main point of extension might be enabling the use of other namespaces.

I added two relatively short release notes, not sure if one should add a link to the API docs (probably not)?

mhvk:
@seberg - nice fix, this had me quite stymied, but I never even considered that there would be a difference with being in a c++ file...

Agreed that we should move forward! It may be good to have a new issue with leftovers. I think the main one is to broaden it to all ufuncs, i.e., allow the name to contain dots and perhaps a colon for the module (trying to understand the logic for it: is it to avoid someone having a package that is called like a numpy ufunc?). One could envision a call out to a little python helper either if the name contains : or ., or if the lookup in numpy fails.

seberg:
Colon is just the precise way to represent a full Python object reference: `module.submodule:obj.attribute`, see for example also entry-point definitions.

numpy/_core/src/umath/dispatching.cpp
mhvk: Hmm, not sure I agree here, why not just let the error pass through? I.e., `return -1`? This will be seen only by developers for whom it may be nicer to have the original error. Anyway, would change it here and above.

mhvk (approved):
One more thing, but I'll just push the fix up, so we can get this merged!

mhvk:
p.s. I took the opportunity to squash commits together, trying to leave the small contributes of @seberg and me in separate commits, so credit goes to where it belongs! Will merge once the tests have passed.

mhvk:
p.p.s. Generalization to non-numpy names forthcoming...

MaanasArora:
Thanks @mhvk and @seberg! Looking forward to the generalization and I'll start the follow-up for StringDType (and think more about the compare function idea).

PR #29901, opened 2025-10-09
TYP: add missing `__slots__`

jorenham:
as caught by https://github.com/JelleZijlstra/stubdefaulter, analogous to numpy/numtype#723


charris:
Thanks Joren.

PR #29902, opened 2025-10-09
TYP: wrong argument defaults in ``testing._private``

jorenham:
These are the backportable ones that were caught by https://github.com/JelleZijlstra/stubdefaulter.

PR #29903, opened 2025-10-09
TYP: fix incorrect ``ma.sort`` arg default for ``stable``

jorenham:
Same story as #29902, but this one shouldn't be backported


charris:
Thanks Joren.

PR #29914, opened 2025-10-09
TYP: minor fixes related to ``errstate``

jorenham:
- move the `np.errstate` stub from `__init__.pyi` to `_core/_ufunc_config.pyi` to better match the runtime situation.
- stub the `_unspecified` class and `_Unspecified` sentinel in `_core._ufunc_config`
- correct mismatching parameter type and parameter default of the `call` kwarg in `errstate.__init__`
- add missing `__all__` in `_core/_ufunc_config.pyi`

### DIFF
--- a/numpy/_core/_ufunc_config.pyi
+++ b/numpy/_core/_ufunc_config.pyi
@@ -1,12 +1,22 @@
 from _typeshed import SupportsWrite
 from collections.abc import Callable
-from typing import Any, Literal, TypeAlias, TypedDict, type_check_only
+from types import TracebackType
+from typing import Any, Final, Literal, TypeAlias, TypedDict, TypeVar, type_check_only
 
-from numpy import errstate as errstate
+__all__ = [
+    "seterr",
+    "geterr",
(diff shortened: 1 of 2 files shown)


numpy/_core/_ufunc_config.pyi
```diff
@@ -15,6 +25,36 @@ class _ErrDict(TypedDict):
...
+_Unspecified: Final[_unspecified]
+
+class errstate:
+    __slots__ = "_all", "_call", "_divide", "_invalid", "_over", "_token", "_under"
+
+    def __init__(
+        self,
+        /,
+        *,
+        call: _ErrCall | _unspecified = ...,  # = _Unspecified
```
seberg, line 41: Weird, I guess I didn't want to clean that up last time around, but we should just use `np._NoValue` it seems to me...
Anyway, doesn't matter for this.

(I don't really think this should be backported -- I can say that once every few months, right? :).)

seberg (approved):
Looks good, thanks, Joren.

PR #29915, opened 2025-10-10
TYP: move ``matrix`` from ``__init__.pyi`` to ``matrixlib/defmatrix.pyi``

jorenham:
This way it better matches the runtime definition. It also helps make `__init__.pyi` a bit less monolithic.

This also slightly narrows the matrix return types with unknown dtypes do have `matrix.dtype: np.dtype[Any]` instead of `matrix.dtype: Any`. It's less type-unsafe this way, and more in line with `ndarray`.

There's no need to backport this, btw.


jorenham:
circleci failure seems unrelated

mattip:
Thanks @jorenham

Issue #21532, opened 2022-05-18
BUG: Refcnt leak on failure (some static analysis reports)


Snape3058:
Please give every report a confirmation result. Refer to a report with its ID (the hex numbers in the parenthesis) for simplicity.
If there are multiple hex numbers in the parenthesis, they are the IDs for the reports in the group above or below.

For a group of reports, a summarized confirmation to the group is also OK.

If there are any questions about a report, please leave a comment with the report ID or link, and I will reply the detailed path provided by our static analysis report.

## Undecreased refcnt due to the `return NULL` statement in macro `import_array`.

This issue also affects other projects.

https://github.com/numpy/numpy/blob/59aad3cf436691d447e43038e3402500f13cd817/numpy/core/src/umath/_operand_flag_tests.c#L60
https://github.com/numpy/numpy/blob/59aad3cf436691d447e43038e3402500f13cd817/numpy/core/src/umath/_struct_ufunc_tests.c#L126
https://github.com/numpy/numpy/blob/59aad3cf436691d447e43038e3402500f13cd817/numpy/core/src/umath/_umath_tests.c.src#L705
https://github.com/numpy/numpy/blob/eecd16210ffb487d4e862c71e86cc6f88f2e9aa2/numpy/linalg/umath_linalg.cpp#L4535
https://github.com/numpy/numpy/blob/b222eb66c79b8eccba39f46f020ed8303614a87f/numpy/fft/_pocketfft.c#L2374
https://github.com/numpy/numpy/blob/b222eb66c79b8eccba39f46f020ed8303614a87f/numpy/linalg/lapack_litemodule.c#L397

(79bbae ca50d3 281b63 7ce5a3 ab547b 9bedd0)

## On failure, return NULL directly without decreasing the refcnt.

New reference returned here is inherited from its argument, but not decreased when returned on line 379. (Take the true branch on line 370) (a7dae8)
https://github.com/numpy/numpy/blob/b222eb66c79b8eccba39f46f020ed8303614a87f/numpy/core/src/multiarray/calculation.c#L362

New reference returned here is inherited from its argument, but not decreased when returned on line 410. (Take the true branch on line 390) (fbd6d9)
https://github.com/numpy/numpy/blob/b222eb66c79b8eccba39f46f020ed8303614a87f/numpy/core/src/multiarray/calculation.c#L381

New reference returned here is not decreased when returned on line 435. (347302)
[...]

PR #29917, opened 2025-10-10
Fix memory leak in import_array()

GermnGodoy:
## Summary
Fixes memory leak when import_array() or import_umath() fail by moving the import calls before PyModule_Create().

## Changes
- Moved import_array() and import_umath() before module creation in:
 - numpy/_core/src/umath/_operand_flag_tests.c
 - numpy/_core/src/umath/_struct_ufunc_tests.c

## Explanation
The import_array() and import_umath() macros contain `return NULL` statements. When called after PyModule_Create(), a failure causes the function to exit immediately without cleaning up the module object, resulting in a memory leak.

By calling these imports before creating any Python objects, if they fail, there's nothing to clean up.

### DIFF
--- a/numpy/_core/src/umath/_operand_flag_tests.c
+++ b/numpy/_core/src/umath/_operand_flag_tests.c
@@ -57,13 +57,13 @@ PyMODINIT_FUNC PyInit__operand_flag_tests(void)
     PyObject *m = NULL;
     PyObject *ufunc;
 
+    import_array();
+    import_umath();
+
     m = PyModule_Create(&moduledef);
     if (m == NULL) {
         goto fail;
     }
-
-    import_array();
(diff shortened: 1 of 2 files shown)


Snape3058:
Hi, I suggest removing the "close #21532" in the commit message, as this patch fixes a part of it, not all. Using this will close the original issue together with this PR.

GermnGodoy:
Thanks for the feedback! I've updated the commit message to remove the reference.

mhvk (approved):
While only used for testing, so unlikely to ever cause problems, it might indeed as well be right, as people (at least those like me), use test cases for inspiration. Thanks!

mhvk:
All tests pass, so merging. Thanks again!

PR #29922, opened 2025-10-10
TYP: minor fixes related to ``errstate`` (#29914)

charris:
Backport of #29914.

- move the `np.errstate` stub from `__init__.pyi` to `_core/_ufunc_config.pyi` to better match the runtime situation.
- stub the `_unspecified` class and `_Unspecified` sentinel in `_core._ufunc_config`
- correct mismatching parameter type and parameter default of the `call` kwarg in `errstate.__init__`
- add missing `__all__` in `_core/_ufunc_config.pyi`

### DIFF
--- a/numpy/_core/_ufunc_config.pyi
+++ b/numpy/_core/_ufunc_config.pyi
@@ -1,13 +1,21 @@
 from collections.abc import Callable
-from typing import Any, Literal, TypeAlias, TypedDict, type_check_only
+from types import TracebackType
+from typing import Any, Final, Literal, TypeAlias, TypedDict, TypeVar, type_check_only
 
-from _typeshed import SupportsWrite
-
-from numpy import errstate as errstate
+__all__ = [
+    "seterr",
+    "geterr",
+    "setbufsize",
+    "getbufsize",
+    "seterrcall",
+    "geterrcall",
+    "errstate",
+]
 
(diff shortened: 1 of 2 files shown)


numpy/_core/_ufunc_config.pyi
```diff
@@ -1,13 +1,21 @@
 from collections.abc import Callable
```
jorenham, line 1: There seems to be [a missing import](https://github.com/numpy/numpy/blob/6bb72644e16e5e7f590533bde670c42b199edb53/numpy/_core/_ufunc_config.pyi#L1):

```pyi
from _typeshed import SupportsWrite
```

jorenham: must be related to the changes in the ruff config that caused the imports to be reordered.

jorenham: I added the import myself; hope you don't mind

charris: I do mind :) All the backports now have that  problem, a PR would have been better.

charris: Rather, I rebased and force pushed and now your fix is gone.

jorenham: > I do mind :) All the backports now have that problem, a PR would have been better.

Ah ok sorry about that, I'll do that next time then

PR #29925, opened 2025-10-10
MAINT: Avoid assumptions about how memory is allocated

mhvk:
While trying to store data on the instance, I found that there were a few places where code needlessly relies on details of memory allocation. Since those cases are useful to fix even though I'm considering a better scheme that would not have problems, I thought I might as well put them in a separate PR.

For ease of reference, the three commits are:

MAINT: Avoid "fooling deallocator" in array_descr_set
Found this caused problems while trying to implement storing the
dimensions and strides on the instance.  Splitting it out since it is
generally bad form to assume too much about the inner workings of, e.g.,
the deallocator, and it has no real benefits beyond trying to be clever.

MAINT: Use PyArray_Resize to grow arrays in FromFile and FromIter
Found this while trying to implement storage on the instance, but
factored it out since it seems better form to just have one place
in numpy where arrays are resized inplace (which ideally would not
happen at all, but here is quite logical).

For PyArray_FromFile, there is little benefit beyond that, but in
PyArray_FromIter, it allows removing quite a bit of code that
duplicated what PyArray_Resize does.

MAINT: Ensure PyArray_Resize replaces empty allcation with 1 byte.
Minimum allocation should be 1, not elsize.  Also use the default
function to deallocate dimension & strides.

### DIFF
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -2390,7 +2390,7 @@ PyArray_FromInterface(PyObject *origin)
         goto fail;
     }
     if (use_scalar_assign) {
-        /* 
+        /*
          * NOTE(seberg): I honestly doubt anyone is using this scalar path and we
          * could probably just deprecate (or just remove it in a 3.0 version).
          */
@@ -3732,21 +3732,20 @@ PyArray_FromFile(FILE *fp, PyArray_Descr *dtype, npy_intp num, char *sep)
     }
     if (((npy_intp) nread) < num) {
         /*
-         * Realloc memory for smaller number of elements, use original dtype
-         * which may have include a subarray (and is used for `nread`).
+         * Resize array to smaller number of elements. Note that original
+         * dtype may have included a subarray, so we may not be 1-d.
          */
-        const size_t nsize = PyArray_MAX(nread,1) * dtype->elsize;
-        char *tmp;
-
-        /* The handler is always valid */
-        if((tmp = PyDataMem_UserRENEW(PyArray_DATA(ret), nsize,
-                                     PyArray_HANDLER(ret))) == NULL) {
+        npy_intp dims[NPY_MAXDIMS];
+        dims[0] = (npy_intp)nread;
+        for (int i = 1; i < PyArray_NDIM(ret); i++) {
+            dims[i] = PyArray_DIMS(ret)[i];
+        }
+        PyArray_Dims new_dims = {dims, PyArray_NDIM(ret)};
+        if (PyArray_Resize_int(ret, &new_dims, 0) < 0) {
             Py_DECREF(dtype);
             Py_DECREF(ret);
-            return PyErr_NoMemory();
+            return NULL;
         }
-        ((PyArrayObject_fields *)ret)->data = tmp;
-        PyArray_DIMS(ret)[0] = nread;
     }
     Py_DECREF(dtype);
     return (PyObject *)ret;
@@ -3998,6 +3997,7 @@ PyArray_FromIter(PyObject *obj, PyArray_Descr *dtype, npy_intp count)
     PyObject *iter = NULL;
     PyArrayObject *ret = NULL;
     npy_intp i, elsize, elcount;
+    npy_intp dims[NPY_MAXDIMS];
 
     if (dtype == NULL) {
         return NULL;
@@ -4037,6 +4037,9 @@ PyArray_FromIter(PyObject *obj, PyArray_Descr *dtype, npy_intp count)
     if (ret == NULL) {
         goto done;
     }
+    /* set up for possible resizing */
+    memcpy(dims, PyArray_DIMS(ret), PyArray_NDIM(ret)*sizeof(npy_intp));
+    PyArray_Dims new_dims = {dims, PyArray_NDIM(ret)};
 
     char *item = PyArray_BYTES(ret);
     for (i = 0; i < count || count == -1; i++, item += elsize) {
@@ -4044,14 +4047,12 @@ PyArray_FromIter(PyObject *obj, PyArray_Descr *dtype, npy_intp count)
         if (value == NULL) {
             if (PyErr_Occurred()) {
                 /* Fetching next item failed perhaps due to exhausting iterator */
-                goto done;
+                goto fail;
             }
             break;
         }
 
         if (NPY_UNLIKELY(i >= elcount) && elsize != 0) {
-            char *new_data = NULL;
-            npy_intp nbytes;
             /*
               Grow PyArray_DATA(ret):
               this is similar for the strategy for PyListObject, but we use
@@ -4060,31 +4061,18 @@ PyArray_FromIter(PyObject *obj, PyArray_Descr *dtype, npy_intp count)
                     be suitable to reuse here.
             */
             elcount = (i >> 1) + (i < 4 ? 4 : 2) + i;
-            if (!npy_mul_sizes_with_overflow(&nbytes, elcount, elsize)) {
-                /* The handler is always valid */
-                new_data = PyDataMem_UserRENEW(
-                        PyArray_BYTES(ret), nbytes, PyArray_HANDLER(ret));
-            }
-            if (new_data == NULL) {
-                PyErr_SetString(PyExc_MemoryError,
-                        "cannot allocate array memory");
+            dims[0] = elcount;
+            if (PyArray_Resize_int(ret, &new_dims, 0) < 0) {
                 Py_DECREF(value);
-                goto done;
+                goto fail;
             }
-            ((PyArrayObject_fields *)ret)->data = new_data;
-            /* resize array for cleanup: */
-            PyArray_DIMS(ret)[0] = elcount;
             /* Reset `item` pointer to point into realloc'd chunk */
-            item = new_data + i * elsize;
-            if (PyDataType_FLAGCHK(dtype, NPY_NEEDS_INIT)) {
-                /* Initialize new chunk: */
-                memset(item, 0, nbytes - i * elsize);
-            }
+            item = ((char *)PyArray_DATA(ret)) + i * elsize;
         }
 
         if (PyArray_Pack(dtype, item, value) < 0) {
             Py_DECREF(value);
-            goto done;
+            goto fail;
         }
         Py_DECREF(value);
     }
@@ -4093,46 +4081,22 @@ PyArray_FromIter(PyObject *obj, PyArray_Descr *dtype, npy_intp count)
         PyErr_Format(PyExc_ValueError,
                 "iterator too short: Expected %zd but iterator had only %zd "
                 "items.", (Py_ssize_t)count, (Py_ssize_t)i);
-        goto done;
+        goto fail;
     }
 
     /*
      * Realloc the data so that don't keep extra memory tied up and fix
      * the arrays first dimension (there could be more than one).
      */
-    if (i == 0 || elsize == 0) {
-        /* The size cannot be zero for realloc. */
+    dims[0] = i;
+    if (!PyArray_Resize_int(ret, &new_dims, 0)) {
+        goto done;
     }
-    else {
-        /* Resize array to actual final size (it may be too large) */
-        /* The handler is always valid */
-        char *new_data = PyDataMem_UserRENEW(
-                PyArray_DATA(ret), i * elsize, PyArray_HANDLER(ret));
 
-        if (new_data == NULL) {
-            PyErr_SetString(PyExc_MemoryError,
-                    "cannot allocate array memory");
-            goto done;
-        }
-        ((PyArrayObject_fields *)ret)->data = new_data;
+  fail:
+    Py_CLEAR(ret);
 
-        if (count < 0) {
-            /*
-             * If the count was smaller than zero, the strides may be all 0
-             * (even in the later dimensions for `count < 0`!
-             * Thus, fix all strides here again for C-contiguity.
-             */
-            int oflags;
-            _array_fill_strides(
-                    PyArray_STRIDES(ret), PyArray_DIMS(ret), PyArray_NDIM(ret),
-                    PyArray_ITEMSIZE(ret), NPY_ARRAY_C_CONTIGUOUS, &oflags);
-            PyArray_STRIDES(ret)[0] = elsize;
-            assert(oflags & NPY_ARRAY_C_CONTIGUOUS);
-        }
-    }
-    PyArray_DIMS(ret)[0] = i;
-
- done:
+  done:
     Py_XDECREF(iter);
     Py_XDECREF(dtype);
     if (PyErr_Occurred()) {
--- a/numpy/_core/src/multiarray/getset.c
+++ b/numpy/_core/src/multiarray/getset.c
@@ -500,15 +500,23 @@ array_descr_set(PyArrayObject *self, PyObject *arg, void *NPY_UNUSED(ignored))
         if (temp == NULL) {
             return -1;
         }
+        /* create new dimensions cache and fill it */
+        npy_intp new_nd = PyArray_NDIM(temp);
+        npy_intp *new_dims = npy_alloc_cache_dim(2 * new_nd);
+        if (new_dims == NULL) {
+            Py_DECREF(temp);
+            PyErr_NoMemory();
+            return -1;
+        }
+        memcpy(new_dims, PyArray_DIMS(temp), new_nd * sizeof(npy_intp));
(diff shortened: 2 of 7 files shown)


numpy/_core/src/multiarray/ctors.c
```diff
@@ -3731,21 +3731,22 @@ PyArray_FromFile(FILE *fp, PyArray_Descr *dtype, npy_intp num, char *sep)
...
+         * dtype may have included a subarray, so we may not be 1-d.
          */
-        const size_t nsize = PyArray_MAX(nread,1) * dtype->elsize;
-        char *tmp;
-
-        /* The handler is always valid */
-        if((tmp = PyDataMem_UserRENEW(PyArray_DATA(ret), nsize,
-                                     PyArray_HANDLER(ret))) == NULL) {
+        npy_intp *dims = npy_alloc_cache_dim(PyArray_NDIM(ret));
+        dims[0] = (npy_intp)nread;
```
seberg, line 3739: Hmmm, seems a bit inconvenient here with the size changes, might a more raw resize helper make it nicer?
(I.e. something internal that `PyArray_Resize()` itself just calls.)

Maybe that doesn't make sense... Although just the fact that `PyArray_Resize` returns None (and uses that `PyArray_Dims` struct?) might be worth a helper?

numpy/_core/src/multiarray/ctors.c
```diff
@@ -4036,6 +4038,10 @@ PyArray_FromIter(PyObject *obj, PyArray_Descr *dtype, npy_intp count)
     if (ret == NULL) {
         goto done;
     }
+    /* set up for resizing */
+    dims = npy_alloc_cache_dim(PyArray_NDIM(ret));
+    memcpy(dims, PyArray_DIMS(ret), PyArray_NDIM(ret)*sizeof(npy_intp));
+    PyArray_Dims new_dims = {dims, PyArray_NDIM(ret)};
```
seberg, line 4042: This is missing an OOM check, but really just use `npy_intp dims[NPY_MAXDIMS]` and stop worrying about cleanup.

seberg (commented):
Wondering slightly if a more raw thing might be helpful, but not sure, then you still need to update the shape correctly.  Although... all of these functions that reasonably resize clearly have to update shape all the time anyway?

(But otherwise the `None` return is a thorn, although I am not strictly sure it's worth removing...)

One thing: I think you should just use `dims[NPY_MAXDIMS]` at least for any of the temporary ones just to simplify the code...

numpy/_core/src/multiarray/ctors.c
mhvk: I think with your suggested `dims[NPY_MAXDIMS]` it is not so bad. There is also an advantage to using public API - good to stress test it and maybe find out what we would like to change!

numpy/_core/src/multiarray/ctors.c
mhvk: Done

numpy/_core/src/multiarray/textreading/rows.c
```diff
@@ -311,9 +310,6 @@ read_rows(stream *s,
                 if (data_array == NULL) {
                     goto error;
                 }
-                if (needs_init) {
-                    memset(PyArray_BYTES(data_array), 0, PyArray_NBYTES(data_array));
```
mhvk, line 315: There is no need for initializing the memory if the dtype needs it - already done by `PyArray_SimpleNewFromDescr`.

numpy/_core/src/multiarray/textreading/rows.c
```diff
@@ -354,22 +350,15 @@ read_rows(stream *s,
...
                 goto error;
             }
-            /* Replace the arrays data since it may have changed */
-            ((PyArrayObject_fields *)data_array)->data = new_data;
-            ((PyArrayObject_fields *)data_array)->dimensions[0] = new_rows;
-            data_ptr = new_data + row_count * row_size;
+            data_ptr = PyArray_DATA(data_array) + row_count * row_size;
             data_allocated_rows = new_rows;
-            if (needs_init) {
-                memset(data_ptr, '\0', (new_rows - row_count) * row_size);
```
mhvk, line 371: `PyArray_Resize` also takes care of initialization. Nice number of lines saved!

mhvk (commented):
@seberg - thanks for the comments! With your suggestion of just using a fixed `dims[NPY_MAXDIMS]` the code is indeed clearer (and more correct, in not forgetting to check for allocation errors...).

Note that I found another culprit, the textreader (which only uses `UserRENEW`, not `UserNEW`, hence an earlier grep failed to find it...). Here the use of `PyArray_Resize` is even nicer, since the equivalent of `dims` is already available in the form of `result_shape`.

numpy/_core/src/multiarray/ctors.c
```diff
@@ -3731,21 +3732,20 @@ PyArray_FromFile(FILE *fp, PyArray_Descr *dtype, npy_intp num, char *sep)
...
-        /* The handler is always valid */
-        if((tmp = PyDataMem_UserRENEW(PyArray_DATA(ret), nsize,
-                                     PyArray_HANDLER(ret))) == NULL) {
+        dims[0] = (npy_intp)nread;
+        for (int i = 1; i < PyArray_NDIM(ret); i++) {
+            dims[i] = PyArray_DIMS(ret)[i];
+        }
+        PyArray_Dims new_dims = {dims, PyArray_NDIM(ret)};
+        PyObject *tmp = PyArray_Resize(ret, &new_dims, 0, 0);
+        if(tmp == NULL) {
```
seberg:
```suggestion
        if (tmp == NULL) {
```

More importantly, `Py_DECREF(tmp)` is missing.

numpy/_core/src/multiarray/ctors.c
```diff
@@ -4059,26 +4061,19 @@ PyArray_FromIter(PyObject *obj, PyArray_Descr *dtype, npy_intp count)
...
             */
             elcount = (i >> 1) + (i < 4 ? 4 : 2) + i;
+            npy_intp nbytes = 0;
+            PyObject *res = NULL;
             if (!npy_mul_sizes_with_overflow(&nbytes, elcount, elsize)) {
-                /* The handler is always valid */
-                new_data = PyDataMem_UserRENEW(
-                        PyArray_BYTES(ret), nbytes, PyArray_HANDLER(ret));
+                dims[0] = elcount;
+                res = PyArray_Resize(ret, &new_dims, 0, 0);
```
seberg: Maybe we should put `NY_CORDER` as last argument for clarity.

numpy/_core/src/multiarray/ctors.c
```diff
@@ -3692,6 +3692,7 @@ PyArray_FromFile(FILE *fp, PyArray_Descr *dtype, npy_intp num, char *sep)
 {
     PyArrayObject *ret;
     size_t nread = 0;
+    npy_intp dims[NPY_MAXDIMS];
```
seberg: Slightly nicer to move down.

numpy/_core/src/multiarray/ctors.c
seberg: Anyway, looking at this now, it doesn't seem so bad anymore (resize is a bit clumsy, but...)

numpy/_core/src/multiarray/ctors.c
seberg: Yeah, I guess... just if that function returned 0 and -1, it would allow for much nicer `if (PyArray_Resize(...) < 0) {` calls and no silly cleanup for None.

numpy/_core/src/multiarray/ctors.c
```diff
@@ -4059,26 +4061,19 @@ PyArray_FromIter(PyObject *obj, PyArray_Descr *dtype, npy_intp count)
                     be suitable to reuse here.
             */
             elcount = (i >> 1) + (i < 4 ? 4 : 2) + i;
+            npy_intp nbytes = 0;
+            PyObject *res = NULL;
             if (!npy_mul_sizes_with_overflow(&nbytes, elcount, elsize)) {
```
seberg: This `mul_sizes_with_overflow` protects from overflows (mainly on 32bit systems).
`PyArray_Resize()` *should* already do that (I dunno if I trust it to, though).

So this branch should be unnecessary now (which also cleans up the slightly weird error logic)

numpy/_core/src/multiarray/ctors.c
```diff
@@ -4099,39 +4094,14 @@ PyArray_FromIter(PyObject *obj, PyArray_Descr *dtype, npy_intp count)
...
-            _array_fill_strides(
-                    PyArray_STRIDES(ret), PyArray_DIMS(ret), PyArray_NDIM(ret),
-                    PyArray_ITEMSIZE(ret), NPY_ARRAY_C_CONTIGUOUS, &oflags);
-            PyArray_STRIDES(ret)[0] = elsize;
-            assert(oflags & NPY_ARRAY_C_CONTIGUOUS);
-        }
+    dims[0] = i;
+    PyObject *res = PyArray_Resize(ret, &new_dims, 0, 0);
+    if (res == NULL) {
+        goto done;
```
seberg, line 4093: Deleting all of this is actually very nice :)!

numpy/_core/src/multiarray/getset.c
```diff
@@ -500,15 +500,23 @@ array_descr_set(PyArrayObject *self, PyObject *arg, void *NPY_UNUSED(ignored))
...
-        ((PyArrayObject_fields *)self)->strides = PyArray_STRIDES(temp);
+        ((PyArrayObject_fields *)self)->nd = new_nd;
+        ((PyArrayObject_fields *)self)->dimensions = new_dims;
+        ((PyArrayObject_fields *)self)->strides = new_dims + new_nd;
         newtype = PyArray_DESCR(temp);
-        Py_INCREF(PyArray_DESCR(temp));
-        /* Fool deallocator not to delete these*/
-        ((PyArrayObject_fields *)temp)->nd = 0;
-        ((PyArrayObject_fields *)temp)->dimensions = NULL;
+        Py_INCREF(newtype);
```
seberg, line 519: I guess we still mess with one error, but yeah I don't mind doing this slightly longer dance here.

numpy/_core/src/multiarray/shape.c
```diff
@@ -122,7 +122,7 @@ PyArray_Resize(PyArrayObject *self, PyArray_Dims *newshape, int refcheck,
             return NULL;
         }
         new_data = PyDataMem_UserRENEW(PyArray_DATA(self),
-                                       newnbytes == 0 ? elsize : newnbytes,
+                                       newnbytes == 0 ? 1 : newnbytes,
```
seberg, line 129: Nice catch, was changed everywhere else!

seberg (commented):
OK, a bit closer look, I am happy with this overall, a few comments (but I can make a pass tomorrow or so).

(Would be curious if this is noticable speed wise, but may try myself also.)

numpy/_core/src/multiarray/ctors.c
mhvk: Done

numpy/_core/src/multiarray/ctors.c
mhvk: Hmm, given that I did not remember to decref multiple times, I've now gone for it...

numpy/_core/src/multiarray/ctors.c
mhvk: But I think you were right after all...

numpy/_core/src/multiarray/ctors.c
mhvk: Good catch, indeed it does check. And I think we should trust it...

numpy/_core/src/multiarray/ctors.c
mhvk: Now that I've created the internal version, I just omitted that entry - it was unused.

numpy/_core/src/multiarray/ctors.c
mhvk: With your suggestion, it has gotten even better!

numpy/_core/src/multiarray/getset.c
mhvk: Agreed it is still not perfect, but it doesn't fit resize and here I just wanted not to mess with `temp`.

mhvk (commented):
Thanks for the detailed review! The multiple times I forgot to decref the `None` from `PyArray_Resize` convinced me that we should after all have an internal version...

seberg:
Looking at the freeing bug, there is one fun little difference here for objects.  Unlike `Resize` the old code always just NULLed everything for initialization and that means also shrinking doesn't need to clean up here.
But `Resize` actually fills as if you called `np.zeros`, which is a bit weird.  Dunno if that should worry, since if we have objects, we probably don't need to worry about the speed of that much.

numpy/_core/src/multiarray/getset.c
mhvk: I've got a PR halfway that changes `setstate` to basically deallocate all but itself (using that part of `array_dealloc`) and hopefully then allocate using the relevant part of ~`PyArray_FromDescrAndData_int`~ `PyArray_FromDescr_int`. 

For here, though, I'd suggest to leave it at this -- at least the deallocation of `temp` is now not messed with.

mhvk:
Hmm, I actually think it is better to be safe. But really, what should happen should be the same between contruction of an an array and resizing. I saw that `PyArray_FromDescr_int` has a whole dance between `calloc` and `fill_zero_info.func`, which seems very similar to what `PyArray_ZeroContiguousBuffer` does (and maybe could be factored out?). So, it seems reasonable to do the same for resizing of arrays in `FromIter` and `FromFile`.

seberg:
> Hmm, I actually think it is better to be safe.

I don't mind cutting this corner (since we size it to exactly what we wrote, everything else would be horribly wrong in more than one way). But... it is maybe not worthwhile avoiding this (at least for shrinking).  FWIW, for growing, I wouldn't mind being able to just init rather than zero fill.

`PyArray_FromDescr_int` could maybe use the `PyArray_ZeroContiguousBuffer` helper.  But it has to do parts of the same setup to be able to decide whether it can use `calloc` or not.

mhvk:
@seberg - was looking at this PR and trying to see what needs to be done. I think we were down to wondering whether or not objects should be initialized as NULL or `None` - my sense for now is that this only matters for speed in a corner case where speed is not very important, so maybe keep it simple here? If we ever want to, we can rewrite `resize` to initialize objects to `NULL` instead of `None`...

seberg:
Yes, agreed. I tried timing briefly, reading a csv file with:
```
In [3]: with open("/tmp/asdf.csv", "w") as f:
   ...:     f.write("1\n" * 1000000)
   ...: 

In [4]: %timeit np.loadtxt("/tmp/asdf.csv", dtype=object)
```
Which went from 26ms to 29ms on my computer (hot file cache).  46 to 37 if I use `dtype="O,i", delimiter=","` and a two column file.

That isn't nothing, but also not a crazy slowdown, I guess.  So yeah, I am fine with just doing this.  I suppose conceptually, I still think these internal uses are a bit different from Python side resizing (where a `np.zeros` style API is also weird, but full initialization makes more sense).

mhvk:
I think the reading slow downs are probably at the acceptable level, especially as it only affects objects, which are not very common. I also think we can solve them if we allow passing something in to tell `PyArray_ZeroContiguousBuffer` whether zeroing an object means writing `NULL`, or inserting `Py_None`. Indeed, that might well help code elsewhere...

mhvk:
p.s. Rebased to get rid of the merge conflict.

seberg:
> . I also think we can solve them if we allow passing something in to tell PyArray_ZeroContiguousBuffer whether zeroing an object means writing NULL, or inserting Py_None. Indeed, that might well help code elsewhere...

An additional helper `PyArray_InitContigouosBufferIfNeeded()` may be useful, but they seem two very different concepts. `ZeroContiguousBuffer` is explicitly for `np.zeros()` which isn't used in many places, so I don't think it would be helpful to mix up the two concepts.

But yeah, I am tempted to add a flag to the `_int` version (call it "unsafe" if you like) that causes only init-if-needed for growing and maybe also promises to only shrunk unused space.
Basically, because any sane code where this isn't OK is buggy anyway (i.e. file readers/fromiter).

Anway, we can just put it in when this passes, I may follow up.

mhvk:
Agreed with the "init_if_needed" idea. And also with doing it in follow-up :smiley_cat: (test passed so I went ahead and merged...)

seberg:
Thanks for merging, sorry for forgetting.

PR #29927, opened 2025-10-11
TST: Add unit test for RISC-V CPU features

ixgbe:
Test environment: Banana Pi BPI-F3 (SpacemiT K1 8-core RISC-V chip) gcc/g++ version 14.2.0
Test command: spin test -- -k "test_cpu_features" -v

Test_RISCV_Features

### DIFF
--- a/numpy/_core/tests/test_cpu_features.py
+++ b/numpy/_core/tests/test_cpu_features.py
@@ -439,10 +439,10 @@ class Test_RISCV_Features(AbstractTest):
     features = ["RVV"]
 
     def load_flags(self):
-            self.load_flags_auxv()
-            if not self.features_flags:
-                # Let the test fail and dump if we cannot read HWCAP.
-                return
-            hwcap = int(next(iter(self.features_flags)), 16)
-            if hwcap & (1 << 21):  # HWCAP_RISCV_V
-                self.features_flags.add("RVV")
+        self.load_flags_auxv()
+        if not self.features_flags:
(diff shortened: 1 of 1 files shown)


numpy/_core/tests/test_cpu_features.py
```diff
@@ -431,3 +431,21 @@ class Test_LOONGARCH_Features(AbstractTest):
...
+    def load_flags(self):
+        self.features_flags = set()
+        with open('/proc/cpuinfo') as fd:
+            for line in fd:
+                if line.startswith("isa"):
+                    isa_string = line.split(':', 1)[1].strip()
+                    base_isa_part = isa_string.split('_')[0]
+                    if 'v' in base_isa_part:
+                        self.features_flags.add("V")
+                    break
```
seiko2plus:
```suggestion

    def load_flags(self):
        self.load_flags_auxv()
        if not self.features_flags:
            # Let the test fail and dump if we cannot read HWCAP.
            return
        hwcap = int(next(iter(self.features_flags)), 16)
        if hwcap & (1 << 21):  # HWCAP_RISCV_V
            self.features_flags.add("RVV")
```

Have you tried to load the flags via `auxv` instead?

ixgbe: Yes, I've tested it with auxv. It works correctly and detects RVV properly. Thanks for the great suggestion!
image

seiko2plus (approved):

seiko2plus:
@ixgbe, Thank you

PR #29929, opened 2025-10-11
ENH: ensure intent(inplace) does not mangle input array

mhvk:
While looking at trying to store data on instances, I ran into problems with f2py assuming one could just swap data between array instances. This is pretty bad form, but was probably necessary at the time, since we did not yet have `WRITEBACKIFCOPY`. But now we do. So, this PR uses it.

The commits are
1. Make the required changes and test them with a new piece of fortran code;
2. Adjust the tests that check the wrapper directly. This needed a small addition to `wrap`;
3. Adjust the documentation;
4. Changelog entry.

I'm quite pleased with how this turned out!

### DIFF
--- a/doc/release/upcoming_changes/29929.improvement.rst
+++ b/doc/release/upcoming_changes/29929.improvement.rst
@@ -4,6 +4,16 @@ the input array was modified in-place, changing its dtype and
 replacing its data by a corrected copy.  Now, instead, the corrected
 copy is kept a separate array, which, after being passed and
 presumably modified by the fortran routine, is copied back to the
-input routine. This means one no longer has the risk that pre-existing
-views or slices of the input array start pointing to unallocated
-memory.
+input routine.  The above means one no longer has the risk that
+pre-existing views or slices of the input array start pointing to
+unallocated memory (at the price of increased overhead for the
+write-back copy at the end of the call).
+
+A potential problem would be that one might get very different results
+if one, e.g., previously passed in an integer array where a double
+array was expected: the writeback to integer would likely give wrong
+results.  To avoid such situations, ``intent(inplace)`` will now only
+allow arrays that have equivalent type to that used in the fortran
+routine, i.e., ``dtype.kind`` is the same. For instance, a routine
+expecting double would be able to receive float, but would raise on
+integer input.
--- a/doc/release/upcoming_changes/29929.improvement.rst
+++ b/doc/release/upcoming_changes/29929.improvement.rst
@@ -1,0 +1,9 @@
+For ``f2py``, the behaviour of ``intent(inplace)`` has improved.
+Previously, if an input array did not have the right dtype or order,
+the input array was modified in-place, changing its dtype and
+replacing its data by a corrected copy.  Now, instead, the corrected
+copy is kept a separate array, which, after being passed and
+presumably modified by the fortran routine, is copied back to the
+input routine. This means one no longer has the risk that pre-existing
+views or slices of the input array start pointing to unallocated
+memory.
--- a/doc/source/f2py/python-usage.rst
+++ b/doc/source/f2py/python-usage.rst
@@ -116,7 +116,7 @@ two notable exceptions:
   :term:`proper-contiguous <contiguous>` and have a compatible ``dtype``,
   otherwise an exception is raised.
 * ``intent(inplace)`` array arguments must be arrays. If these have
-  incompatible order or type, a converted copy is passed in, which is
+  incompatible order or size, a converted copy is passed in, which is
   copied back into the original array on exit (see the ``intent(inplace)``
   :ref:`attribute <f2py-attributes>` for more information).
 
--- a/doc/source/f2py/python-usage.rst
+++ b/doc/source/f2py/python-usage.rst
@@ -115,8 +115,9 @@ two notable exceptions:
 * ``intent(inout)`` array arguments must always be
   :term:`proper-contiguous <contiguous>` and have a compatible ``dtype``,
   otherwise an exception is raised.
-* ``intent(inplace)`` array arguments  will be changed *in situ* if the argument
-  has a different type than expected (see the ``intent(inplace)``
+* ``intent(inplace)`` array arguments must be arrays. If these have
+  incompatible order or type, a converted copy is passed in, which is
+  copied back into the original array on exit (see the ``intent(inplace)``
   :ref:`attribute <f2py-attributes>` for more information).
 
 In general, if a NumPy array is :term:`proper-contiguous <contiguous>` and has
--- a/doc/source/f2py/f2py.getting-started.rst
+++ b/doc/source/f2py/f2py.getting-started.rst
@@ -154,9 +154,8 @@ Fortran subroutine ``FIB`` is accessible via ``fib1.fib``::
     Clearly, this is unexpected, as Fortran typically passes by reference. That
     the above example worked with ``dtype=float`` is considered accidental.
 
-    F2PY provides an ``intent(inplace)`` attribute that modifies the attributes
-    of an input array so that any changes made by the Fortran routine will be
-    reflected in the input argument. For example, if one specifies the
+    F2PY provides an ``intent(inplace)`` attribute that ensures that changes
+    are copied back to the input argument. For example, if one specifies the
     ``intent(inplace) a`` directive (see :ref:`f2py-attributes` for details),
     then the example above would read::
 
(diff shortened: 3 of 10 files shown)


mhvk:
@charris, @HaoZeke - not sure you are the right ones, but most fixes for `f2py` recently seem to have been by you, so I was hoping you might be willing to have a look at this fairly involved PR, which makes `intent(inplace)` behave much more reasonably. Thanks!

doc/release/upcoming_changes/29929.improvement.rst
```diff
@@ -0,0 +1,9 @@
+For ``f2py``, the behaviour of ``intent(inplace)`` has improved.
+Previously, if an input array did not have the right dtype or order,
+the input array was modified in-place, changing its dtype and
+replacing its data by a corrected copy.  Now, instead, the corrected
+copy is kept a separate array, which, after being passed and
+presumably modified by the fortran routine, is copied back to the
+input routine. This means one no longer has the risk that pre-existing
+views or slices of the input array start pointing to unallocated
+memory.
```
seberg: Sorry, but will keep it at a bit of a fly-by comment for now.  This is scary (mostly the old code, it's scary enough that I suspect we should just change it... although I do wonder if we should maybe go towards an error)!
It seems to me that the swapping for outputs would lead to `out=float_arr` for integer array functions to just swap the `float_arr` to be an integer one!?

mhvk: I know, the old way meant that if you put in an integer array, your own *instance* would become a float array! With my new mechanism, you'd keep an integer array, with the copied-back float stuff truncated and possibly overflowing. So, definitely an aspect of "don't do that, then!".

I do think it is correct that if you put in a `float32` to a fortran routine taking only `float64`, you just get reduced precision output (instead of your array becoming `float64`!). It is like the `ufunc` would do it (with the same mechanism, I think...).

Anyway, much of this predates having machinery to deal with these cases, and I find it quite remarkable how well `f2py` has held up -- I've quite happily used it myself when I started with python, and wanted not to lose all the fortran code I had written.

charris: At first glance, I think this was a long standing problem that prevented full API compliance. Or maybe I am thinking of another one where data pointers were swapped ...

mhvk:
@HaoZeke - would you have time to look at this PR for `f2py`. I think it does solve a longstanding issue (and is needed for some of my other work... we really shouldn't be mangling arrays, swapping pointers to memory, etc.). I'm also fairly confident the code itself is right, but would want to be sure the changes I made to the documentation are useful/good.

seberg:
We just brought it up, and I think the main reason this might fail I think is that before there was never any casts, `out` was basically ignored (i.e. replaced) if it doesn't fit perfectly.
But with this, we cast the result to the given `out` with its dtype.

That is _clearly_ better, but will it break downstream code?  We could also just disable `inplace` or just raise an error when the dtype doesn't match and a copy would be needed?

To me that last bit makes sense anyway, the user wanted in-place and didn't get it?  That way we don't have to worry about _changed_ behavior only about new errors.

mattip:
We discussed this at a recent triage meeting and would prefer to just disallow this whole construct. Is that possible? There were some other opinions too.

HaoZeke:
Sorry for being AWOL I have penciled time in this weekend to review this properly and test some older projects.

mhvk:
Yes, seems certainly to make sense to error on bad cases, I should have considered that...

I'll wait with doing anything until @HaoZeke has had a chance to have a look, but FWIW, removing the option to have "inplace" would seem bad, in that it would somewhat gratuitously break code that worked perfectly well. However, refusing to run when the dtype changes seems reasonable, i.e., there would be a run-time failure if you put in, e.g., an integer array when a float was expected (compiled with "inplace").

The question then arises what we consider a "dtype change". Effectively, this is a casting question, and it seems one would definitely like to allow "equiv" (byte-order changes).  The "safe" option is not relevant, since casts will happen in both directions. I think one would also want to allow "same_kind", as it seems OK to put in, e.g., a `float32` array in `float64` code (or vice versa), and it would probably not materially affect the result. But I think we would then error on anything requiring "unsafe". (Ideally, I guess it would be an option, but I do not know how to implement that...)

p.s. I do hope we can make a decision that does not, effectively, prevent changes to how `ndarray` organizes its associated memory, though! Ideally, those become proper implementation details...

seberg:
Yeah, we should move on, it's just would be nicer to know how much and how badly it might break things.  It seemed like this isn't used much (thus why "do we need this" came up).

I tend towards it being OK to require exact (equivalent) dtype.  I do think we do it in a few places for `out=` parameters (but would have to try to remember where).
UFuncs are a bit different, in that they at least use buffered casts, so there is no full copy with `out=`.

To me an "in-place" use-case should almost always be truly in-place anyway and it may actually surprise some users if it's allowed to not be, so I would tend towards the easy way out.

We _could_ deprecate it even, but with f2py being transient and this apparently almost never used (and most use-cases probably unaffected), I would be happy to try to flip the switch for 2.5.  (We should add a release note).

mhvk:
@HaoZeke, @seberg - I've updated my fix to `intent(inplace)` to ensure the input array is the same kind as the array needed by the fortran routine. I decided against enforcing strictly the same dtype, as I worried that greatly increased the likelihood of breaking existing code (in particular, for 32 and 64 bit float and int). Obviously, happy to change it if you think it is better (e.g., we could explicitly allow only 32 and 64 bit versions, not 8 and 16, which are much less likely to have been used in any fortran code, and much more likely to give wrong results).

Note that I think there are other pieces that might break. E.g., it doesn't look like the current code checks byte order.

I think it would be reasonable to deprecate the use of `intent(inplace)` overall, but probably best as follow-up.

seberg:
I would prefer the strict version since that will shout at you in a useful way, while the other one could in principle just do the wrong thing silently.
(The main issue I see is that I am not sure if downstream builds newer versions with newer NumPy versions, allowing for updating a downstream library in a bug-fix release, which is undesireable either way though)

mhvk:
I can certainly do that, but as I mentioned, I worry about breaking reasonable code (float32 code being fed float64 in particular). Let's see what @HaoZeke thinks...

doc/source/f2py/python-usage.rst
```diff
@@ -115,8 +115,9 @@ two notable exceptions:
 * ``intent(inout)`` array arguments must always be
   :term:`proper-contiguous <contiguous>` and have a compatible ``dtype``,
   otherwise an exception is raised.
-* ``intent(inplace)`` array arguments  will be changed *in situ* if the argument
-  has a different type than expected (see the ``intent(inplace)``
+* ``intent(inplace)`` array arguments must be arrays. If these have
+  incompatible order or size, a converted copy is passed in, which is
+  copied back into the original array on exit (see the ``intent(inplace)``
```
HaoZeke, line 120: There is of course a performance cost now, but given that it is more correct strictly speaking, I think that's fine. Might be worth highlighting that for performance parity the arrays should be pre-aligned.

HaoZeke (approved):
One minor doc nit aside, I think this makes sense. We'll likely see some pushback in terms of the performance loss but I think it's fine. Thanks @mhvk and sorry for the very long delay!

doc/source/f2py/python-usage.rst
mhvk: Thanks! I made a small addition to the more expansive definition that this piece of text links to, and also added a note to the changelog entry. Overall, it should be a minimal effect - most code presumably does something sufficiently interesting that going from one to two copies does not really matter...

mhvk:
@seberg, @charris - OK to put this in now that @HaoZeke has approved?

seberg (approved):
Sure, I had a brief look, but all my comments were things that are weird all over F2Py, so no more comments...

FWIW, I am still very convinced that a clear error is the much better choice here rather than modifying results.

But if you two agree, and nobody else seems to have an opinion, we'll go this route (at least in a day or so).

(There is no performance gain for users in providing an output that needs a cast, just very mild convenience.
The main argument here is to not break existing code but we do so at the cost of potentially breaking some less code but in subtle ways.)

seberg:
If others tend to agree to me, I would also be happy with following up to putting a `VisibleDeprecationWarning` on the copying path.

mhvk:
> If others tend to agree to me, I would also be happy with following up to putting a `VisibleDeprecationWarning` on the copying path.

@HaoZeke - what do you think?  It seems reasonable given that the documentation already recommends `intent(in,out)` -- indeed, that could be suggested as part of the warning.

HaoZeke:
> > If others tend to agree to me, I would also be happy with following up to putting a `VisibleDeprecationWarning` on the copying path.
> 
> @HaoZeke - what do you think? It seems reasonable given that the documentation already recommends `intent(in,out)` -- indeed, that could be suggested as part of the warning.

Yup that'd be very useful. That way we can also hope to get rid of it in a few releases..

PR #29578, opened 2025-08-17
DOC: Add polynomial evaluation examples and improve polyval docs

olivvius:
- Add evaluation examples to polynomial quick reference table
- Enhance polyval docstring with clearer examples
- Add warning about domain/window scaling with fitted polynomials
- Show comparison between legacy and new polynomial APIs

Fixes #25587

PR #29930, opened 2025-10-12
DOC: Completed and fixed PR #29578

### LINKED ISSUE #25587: DOC: missing examples for np.polynomial.polynomial.polyval

I JUST ran into this issue #22104 and it took me way too long to find that github-issue to finally see an example of how to use the Polynomial correctly. 

Adding an example (like the one given in #22104) [here](https://numpy.org/doc/stable/reference/routines.polynomials.html#quick-reference) and [here](https://numpy.org/doc/stable/reference/generated/numpy.polynomial.polynomial.polyval.html), and perhaps a mention [here](https://numpy.org/doc/stable/reference/generated/numpy.polynomial.polynomial.Polynomial.html#numpy.polynomial.polynomial.Polynomial) (e.g. adding "evaluating the polynomial at positions" instead of, or in additon to, "Call self as a function") would have been very helpful.

Alverok:
Closes #25587 and #29578.

### DIFF
--- a/doc/source/reference/routines.polynomials.rst
+++ b/doc/source/reference/routines.polynomials.rst
@@ -47,22 +47,27 @@ The `~numpy.polynomial.polynomial.Polynomial` class is imported for brevity::
     from numpy.polynomial import Polynomial
 
 
-+------------------------+------------------------------+---------------------------------------+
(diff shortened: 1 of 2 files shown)


mattip:
Thanks @Alverok and @olivvius

PR #29931, opened 2025-10-12
ENH: In spec registration, allow looking up ufuncs in any module.

mhvk:
This allows the addition of loops from specs also from more complicated names such as `numpy:strings.str_len`. cc @MaanasArora 

p.s. I really shouldn't have gone for doing it in C, but it turned out fairly simple, so I guess why not...

### DIFF
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -1895,7 +1895,10 @@ with the rest of the ArrayMethod API.
 
         .. c:member:: const char *name
 
-            The name of the ufunc to add the loop to.
+            The name of the ufunc to add the loop to, in the form like that of
+            entry points, ``(module ':')? (object '.')* name``, with ``numpy``
+            the default module. Examples: ``sin``, ``strings.str_len``,
+            ``numpy.strings:str_len``.
 
         .. c:member:: PyArrayMethod_Spec *spec
 
--- a/numpy/_core/src/common/npy_import.c
+++ b/numpy/_core/src/common/npy_import.c
@@ -19,3 +19,44 @@ init_import_mutex(void) {
 #endif
     return 0;
 }
+
+
+/*! \brief Import a Python object from an entry point string.
+
+ * The name should be of the form "(module ':')? (object '.')* attr".
+ * If no module is present, it is assumed to be "numpy".
+ * On error, returns NULL.
+ */
+NPY_NO_EXPORT PyObject*
+npy_import_entry_point(const char *entry_point) {
+    PyObject *result;
+    const char *item;
+
+    const char *colon = strchr(entry_point, ':');
+    if (colon) { // there is a module.
+        result = PyUnicode_FromStringAndSize(entry_point, colon - entry_point);
+        if (result != NULL) {
+            Py_SETREF(result, PyImport_Import(result));
+        }
+        item = colon + 1;
+    }
+    else {
+        result = PyImport_ImportModule("numpy");
+        item = entry_point;
+    }
+
+    const char *dot = item - 1;
+    while (result != NULL && dot != NULL) {
+        item = dot + 1;
+        dot = strchr(item, '.');
+        PyObject *string = PyUnicode_FromStringAndSize(
+            item, dot ? dot - item : strlen(item));
+        if (string == NULL) {
+            Py_DECREF(result);
+            return NULL;
+        }
+        Py_SETREF(result, PyObject_GetAttr(result, string));
+        Py_DECREF(string);
+    }
+    return result;
+}
--- a/numpy/_core/src/umath/_scaled_float_dtype.c
+++ b/numpy/_core/src/umath/_scaled_float_dtype.c
@@ -947,7 +947,7 @@ sfloat_argsort_resolve_descriptors(
 {
     assert(given_descrs[1] == NULL || given_descrs[1]->type_num == NPY_INTP);
     assert(PyArray_IsNativeByteOrder(given_descrs[0]->byteorder));
-    
+
     loop_descrs[0] = given_descrs[0];
     Py_INCREF(loop_descrs[0]);
     loop_descrs[1] = PyArray_DescrFromType(NPY_INTP);
@@ -1026,11 +1026,12 @@ sfloat_init_ufuncs(void) {
     argsort_spec.casting = NPY_NO_CASTING;
     argsort_spec.flags = NPY_METH_NO_FLOATINGPOINT_ERRORS;
 
+    /* here we chose weirdish names to test the lookup mechanism */
     PyUFunc_LoopSlot loops[] = {
         {"multiply", &multiply_spec},
-        {"add", &add_spec},
-        {"sort", &sort_spec},
-        {"argsort", &argsort_spec},
+        {"_core._multiarray_umath.add", &add_spec},
+        {"numpy:sort", &sort_spec},
+        {"numpy._core.fromnumeric:argsort", &argsort_spec},
         {NULL, NULL}
     };
     if (PyUFunc_AddLoopsFromSpecs(loops) < 0) {
--- a/numpy/_core/meson.build
+++ b/numpy/_core/meson.build
@@ -759,6 +759,7 @@ py.extension_module('_multiarray_tests',
     'src/common/mem_overlap.c',
     'src/common/npy_argparse.c',
     'src/common/npy_hashtable.cpp',
+    'src/common/npy_import.c',
     src_file.process('src/common/templ_common.h.src')
   ],
   c_args: c_args_common,
--- a/numpy/_core/src/common/npy_import.h
+++ b/numpy/_core/src/common/npy_import.h
@@ -112,11 +112,20 @@ npy_cache_import_runtime(const char *module, const char *attr, PyObject **obj) {
 #endif
         Py_DECREF(value);
     }
-    return 0;    
+    return 0;
 }
 
 NPY_NO_EXPORT int
 init_import_mutex(void);
+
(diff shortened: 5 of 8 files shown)


doc/source/reference/c-api/array.rst
```diff
@@ -1895,7 +1895,9 @@ with the rest of the ArrayMethod API.
 
         .. c:member:: const char *name
 
-            The name of the ufunc to add the loop to.
+            The name of the ufunc to add the loop to, in the form
+            ``(module ':')? (submodule '.')* name``, with ``numpy`` the
+            default module. Examples: ``sin``, ``numpy:strings.str_len``.
```
MaanasArora: I noticed @seberg actually said:

> Colon is just the precise way to represent a full Python object reference: module.submodule:obj.attribute

So if I'm not misunderstanding, this doesn't seem conventional? Even if we don't need to change the code, it seems `submodule` can go before the colon at least (documentation-wise)?

``module ('.' submodule)* ':' name`` (edit: or something much cleaner :))

numpy/_core/src/common/npy_import.c
```diff
@@ -19,3 +19,46 @@ init_import_mutex(void) {
...
+    char *item = copy;
+    sep = item + (int)(sep - name);
+    if (*sep == ':') {  // there is a module.
+        *sep = 0;  // replace ":" with termination marker.
+        module = item;
+        item = sep + 1;
+        sep = strchr(item, '.');
+    }
+
+    result = PyImport_ImportModule(module);
```
MaanasArora: This can also import submodules, right? If so see above! Otherwise I guess we will still clarify the format and then decide if this is worth changing accordingly.

MaanasArora (commented):
Thanks, looks really nice! Actually it seems simpler to do it directly in C. I had just one question about the format, and was also wondering a bit about error handling, but we can probably rely on the Python C API to catch those.

doc/source/reference/c-api/array.rst
mhvk: hmm, I think you are right. I actually started that way, but then thought it was narrower. I know for sure that you should be able to have `.` after the colon, though... I can change it so it just doesn't matter where you put the colon...

numpy/_core/src/common/npy_import.c
mhvk: Not quite, like `__import__("numpy.fft")` it returns the top module (but checks that the others exist, I think).

numpy/_core/src/umath/_scaled_float_dtype.c
```diff
@@ -1026,11 +1026,12 @@ sfloat_init_ufuncs(void) {
...
 
+    /* here we chose weirdish names to test the lookup mechanism */
     PyUFunc_LoopSlot loops[] = {
         {"multiply", &multiply_spec},
-        {"add", &add_spec},
-        {"sort", &sort_spec},
-        {"argsort", &argsort_spec},
+        {"_core._multiarray_umath.add", &add_spec},
+        {"numpy:sort", &sort_spec},
+        {"numpy:_core.fromnumeric.argsort", &argsort_spec},
```
MaanasArora: Hmm, yes, it seems the format difference is real!

doc/source/reference/c-api/array.rst
MaanasArora: Hmm, I see! There's probably a clear difference between submodule and object, though? I suppose we can still have attributes of the object this way:

`module ('.' submodule)* ':' name ('.' attribute)*`

numpy/_core/src/common/npy_import.c
MaanasArora: Ah okay - then I guess the code could need a change. As above, I think @seberg's format maps to `module ('.' submodule)* ':' name ('.' attribute)*`

mhvk: OK, the code will now be generous on input. I'll wait with adjusting the text till I'm sure we've converged. In the meantime, I did add another example which should help clarify what is allowed.  Perhaps I should not try to give a formal explanation of it...

MaanasArora: Thanks a lot, yes I think this handles the use cases well! It is generous enough to work with anything in practice.

I wonder - are we worrying about the case someone says `module.object:attribute` or something to that effect (effectively accessing a submodule that doesn't exist)? Clearly very odd, so I don't think it matters, but I'll defer to @seberg here.

MaanasArora: p.s. To be clear, it doesn't matter to me at all, was just wondering! This looks good to me FWIW.

numpy/_core/src/common/npy_import.c
```diff
@@ -19,3 +19,54 @@ init_import_mutex(void) {
...
+    char *module = "numpy";  // default
+
+    char *dot = strchr(name, '.');
+    char *colon = strchr(name, ':');
+    if (dot == NULL && colon == NULL) {  // short-cut
+        return npy_import(module, name);
+    }
+
+    /* make a copy of name so we can insert 0 as we like. */
+    char *copy = malloc(strlen(name) + 1);
```
seberg: I am trying to generally stop using malloc directly, `PyMem_Malloc()` is just malloc, but allows Python the choice (could use the small workspace stuff, but not really worth wit.).

numpy/_core/src/common/npy_import.c
```diff
@@ -19,3 +19,54 @@ init_import_mutex(void) {
...
+    memcpy(copy, name, strlen(name) + 1);
+    char *item = copy;
+    dot = dot != NULL ? copy + (int)(dot - name) : NULL;
```
seberg: Are these `(int)` casts necessary, fine, just seems slightly weird?

numpy/_core/src/common/npy_import.c
seberg: The Python docs seem confused here, we have code like:
```
mod = PyImport_ImportModule("numpy._core._multiarray_umath");
obj = PyObject_GetAttrString(mod, "scalar");
```
I think it refers to `__init__` due to the machinery, but is actually less raw than `__init__` in this way and does return the submodule.

(It's too bad I am pedantic about supporting ufuncs attached somewhere, python is growing a `from . import x` support, but without that :).)

doc/source/reference/c-api/array.rst
```diff
@@ -1895,7 +1895,10 @@ with the rest of the ArrayMethod API.
 
         .. c:member:: const char *name
 
-            The name of the ufunc to add the loop to.
+            The name of the ufunc to add the loop to, in the form
+            ``(module ':')? (submodule '.')* name``, with ``numpy`` the
+            default module. Examples: ``sin``, ``numpy:strings.str_len``,
+            ``numpy.strings:str_len``.
```
seberg, line 1901: Yeah, so `numpy.strings:str_len` would be preferred, because the first only works due to numpy always importing all modules (or supporting lazy import).

I know I liked the idea of just supporting `obj.attribute`... but if you like, I am very happy to just go for simplicity here.
The point being... in the very rare event someone actually needs attribute walking it's OK if they can't use this convenience helper...

seberg (commented):
Thanks, you even wrote nice tests!  Anyway, very sure the import can be simplified.

And if you like, just drop the getattr walking support here.  It's not like we really are likely to need it.
(IIRC, numba creates such objects, but it seems unlikely someone will search for them here.)

numpy/_core/src/common/npy_import.c
mhvk: OK, changed to `PyMem_Malloc` - I actually tried the small workspace stuff, but then it seems I have to add not just `alloc.c` but even more stuff to `meson.build` - probably nicer if `npy_import.h` does not depend on too many things.

numpy/_core/src/common/npy_import.c
mhvk: I removed them - there was part of me that wanted to distinguish pointers from pointer differences, but, really, the compiler knows how to do that.

numpy/_core/src/common/npy_import.c
mhvk: I now import the whole module string - as you said, it does work, contrary to `__import__`.

doc/source/reference/c-api/array.rst
mhvk: I prefer keeping the attribute walker, since with we do what is described for [entry points]( 
https://packaging.python.org/en/latest/specifications/entry-points/):
```
import importlib
modname, qualname_separator, qualname = object_ref.partition(':')
obj = importlib.import_module(modname)
if qualname_separator:
    for attr in qualname.split('.'):
        obj = getattr(obj, attr)
```

For the ufuncs, this is indeed unnecessary, but I put it in `npy_import` because I thought it might be more generally useful...

Indeed, since it is so similar to entry points, I now called it `npy_import_entry_point`. I also realized one could avoid making a copy of the input string by using `PyUnicode_FromStrAndSize` (a unicode object is also created in [`PyImport_ImportModule`](https://github.com/python/cpython/blob/525dcfe5236ee21b234cad16d2a3d5769e77e0ec/Python/import.c#L3414-L3426)).

Anyway, spent more time on it than I should have, but hope this makes sense and looks OK.

seberg (approved):
Hah, impressive that you managed only a single error return and still get error handling right ;).
Yeah, this was not really necessary to think this much on it, but the result is pretty nice!

mhvk:
> Hah, impressive that you managed only a single error return and still get error handling right ;).

Thanks! I'm finding that `Py_SETREF` really helps...

seberg:
Thanks again Marten!

PR #29934, opened 2025-10-13
CI: Use POWER10 GHA runner for NumPy test jobs

sandeepgupta12:
This PR updates the CI configuration to run NumPy tests on a POWER10 self-hosted runner instead of the default environment (Power9). The change ensures SIMD-related tests (previously skipped on Power9) are now executed and validated on Power10 hardware where VSX4 is supported.

**Background / Motivation**

Earlier, the NumPy CI for ppc64le was running on POWER9 hardware.
Many tests (especially those in test_simd.py) were skipped due to lack of VSX4 SIMD support.
Follow-up investigation showed these tests pass correctly on POWER10 after fixes introduced in PR [https://github.com/numpy/numpy/pull/29893].

Moving to a POWER10 runner ensures better test coverage and validates modern hardware capabilities.

cc: @rgommers @ghatwala

### DIFF
--- a/.github/workflows/linux-ppc64le.yml
+++ b/.github/workflows/linux-ppc64le.yml
@@ -20,7 +20,7 @@ jobs:
     # It requires a native ppc64le GHA runner, which is not available on forks.
     # For more details, see: https://github.com/numpy/numpy/issues/29125
     if: github.repository == 'numpy/numpy'
-    runs-on: ubuntu-24.04-ppc64le
+    runs-on: ubuntu-24.04-ppc64le-p10
     name: "Native PPC64LE"
     steps:
     - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0


seiko2plus (commented):
Thank you. The default baseline should be updated to `VSX4` so we can also check the non-dynamically dispatched kernels. This can be done using the option --cpu-baseline=vsx4. Alternatively, native can be used, which deduces the host features:
```yml
    - name: Meson Build
      run: |
        spin build -- -Dallow-noblas=false -Dcpu-baseline=native
```

sandeepgupta12:
> Thank you. The default baseline should be updated to `VSX4` so we can also check the non-dynamically dispatched kernels. This can be done using the option --cpu-baseline=vsx4. Alternatively, native can be used, which deduces the host features:
> 
> ```yaml
>     - name: Meson Build
>       run: |
>         spin build -- -Dallow-noblas=false -Dcpu-baseline=native
> ```

Thanks for the suggestion. I tried building on Power10 with both --cpu-baseline=vsx4 and --cpu-baseline=native, but both failed.

With vsx4, the build errors out in loops_modulo.dispatch.c.src (no definition of key "t").

With native, Meson reports:

Test features "VSX VSX2 VSX3 VSX4" : Unsupported due to Arguments "-march=native"
ERROR: Option "native" doesn't support compiler gcc


So, it seems Power10 builds can’t currently use vsx4 or native baselines successfully.

seiko2plus:
@sandeepgupta12, thanks for the clarification. I will take a closer look and report back here. For now, let's merge it as-is to at least test the dynamically dispatched kernels for VSX3/VSX4.

charris:
@seiko2plus We don't test power in 2.3.x.

seiko2plus:
> We don't test power in 2.3.x.

@charris, Thanks. Should we drop the QEMU tests we have for ppc64le since we have native support via GitHub Actions?

> @sandeepgupta12, thanks for the clarification. I will take a closer look and report back here. For now, let's merge it as-is to at least test the dynamically dispatched kernels for VSX3/VSX4.

@sandeepgupta12, Building with `--cpu-baseline=vsx4` works for me even on GCC 11 after addressing a compiler bug due to conflicts with target attributes that are pushed via Highway #pragma on GCC < 13, so to fix this issue we will need to pass `-DHWY_COMPILE_ONLY_STATIC` to compile flags after merging & pulling Highway PR google/highway/pull/2782.

> With vsx4, the build errors out in loops_modulo.dispatch.c.src (no definition of key "t").

I didn't catch such an error; would you please provide the full build log or the compiler version you are using?

> Test features "VSX VSX2 VSX3 VSX4" : Unsupported due to Arguments "-march=native"
ERROR: Option "native" doesn't support compiler gcc

Weird, which compiler are you using? This means the flag test fails.

sandeepgupta12:
> I didn't catch such an error; would you please provide the full build log or the compiler version you are using?

Here is the complete logs with vsx4 flag- https://github.com/sandeepgupta12/numpy/actions/runs/18490699806/job/52683373860
I will try to build with -DHWY_COMPILE_ONLY_STATIC to compile flags after merging & pulling Highway PR https://github.com/google/highway/pull/2782.

> Weird, which compiler are you using? This means the flag test fails.

In logs I can see gcc 13.3 compiler

PR #29942, opened 2025-10-13
TST: Convert mixed_types_structured to method

bwhitt7:
I've been working on making the NumPy test suite more thread safe. When running tests in `numpy/lib/tests/test_loadtxt.py` that use the custom `mixed_types_structured` fixture, they fail when ran under [pytest-run-parallel](https://github.com/Quansight-Labs/pytest-run-parallel), giving AssertionErrors. Here's the [test run](https://gist.github.com/bwhitt7/f5bb34c3bdd1ed0a6fd8d3f4fd728b8e) with errors if you're interested in seeing that.

I'm somewhat unsure why the usage of a fixture is triggering this, but since fixtures are shared between threads, `np.loadtxt` may be modifying the fixture values in some way that result in subsequent tests dealing with different data. Regardless, changing this from a fixture to a method fixes the issue.

### DIFF
--- a/numpy/lib/tests/test_loadtxt.py
+++ b/numpy/lib/tests/test_loadtxt.py
@@ -40,10 +40,9 @@ def test_comment_multiple_chars(comment):
     assert_equal(a, [[1.5, 2.5], [3.0, 4.0], [5.5, 6.0]])
 
 
-@pytest.fixture
 def mixed_types_structured():
     """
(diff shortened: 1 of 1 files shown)


ngoldbaum (approved):
Makes sense, this doesn't need to be a fixture and these objects are all cheap to create so amortizing that cost doesn't help much.

PR #29944, opened 2025-10-13
BUG: Fix np.strings.slice if stop=None or start and stop >= len

aaronkollasch:
Python treats `slice(-1)` differently from `slice(-1, None)`: The first is interpreted as `slice(None, -1, None)`, while the second becomes `slice(-1, None, None)`, according to the logic in [`slice_new`](https://github.com/python/cpython/blob/be60e4b4f34a097d999d337446786ae14e35c964/Objects/sliceobject.c#L331-L336).

However, `np.strings.slice` treats these identically, as it cannot distinguish unset arguments from arguments set to None. This makes it impossible to get the last characters of each string, for example:
```python
>>> a = np.array(['hello', 'world'])
>>> np.strings.slice(a, -2, None)  # should return last two characters
array(['hel', 'wor'], dtype='<U5')
```

This PR fixes that behavior:
```python
>>> a = np.array(['hello', 'world'])
>>> np.strings.slice(a, -2, None)  # returns last characters as expected
array(['lo', 'ld'], dtype='<U5')
>>> np.strings.slice(a, -2)  # original behavior preserved if no stop given
array(['hel', 'wor'], dtype='<U5')
```

It does this by adding a `stop=np._NoValue` default argument to `np.strings.slice`, which can be overridden with `None`.

The PR adds test conditions to `numpy/_core/tests/test_strings.py::TestMethods::test_slice`, to verify that the slicing behavior matches Python's `slice` with these arguments.

It also fixes an error when start and stop are >= the string length and the dtype is `StringDType()`. To reproduce:
```python
>>> np.__version__
'2.3.3'
>>> a = np.array(['hello', 'world'], dtype="T")
>>> np.strings.slice(a, 5, 7)
  File "numpy-dev/lib/python3.12/site-packages/numpy/_core/strings.py", line 1823, in slice
    return _slice(a, start, stop, step)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
MemoryError: Failed to allocate string in slice
```
When running `spin ipython` based on main, these commands cause the python process to exit with code 251.
To fix this, I added bounds checks to `codepoint_offsets[start]` in `numpy/_core/src/umath/stringdtype_ufuncs.cpp`.


numpy/_core/src/umath/stringdtype_ufuncs.cpp
```diff
@@ -2267,7 +2267,9 @@ slice_strided_loop(PyArrayMethod_Context *context, char *const data[],
             npy_intp outsize = ((size_t)stop < num_codepoints
                                         ? codepoint_offsets[stop]
                                         : (unsigned char *)is.buf + is.size) -
-                               codepoint_offsets[start];
+                               ((size_t)start < num_codepoints
+                                        ? codepoint_offsets[start]
+                                        : (unsigned char *)is.buf + is.size);
```
seberg: Makes me think we should just use `start_ptr = ...`, `stop_ptr =` and then calculate the `outside` from that (and re-use the `start_ptr` below).

numpy/_core/strings.py
```diff
@@ -1726,8 +1726,12 @@ def translate(a, table, deletechars=None):
             [table] + _clean_args(deletechars)
         )
 
+
+NULL = "NULL"
+
+
 @set_module("numpy.strings")
-def slice(a, start=None, stop=None, step=None, /):
+def slice(a, start=None, stop=NULL, step=None, /):
```
seberg: We have that singleton, it is named `np._NoValue`.

numpy/_core/strings.py
```diff
@@ -1784,8 +1794,10 @@ def slice(a, start=None, stop=None, step=None, /):
 
     """
     # Just like in the construction of a regular slice object, if only start
-    # is specified then start will become stop, see logic in slice_new.
-    if stop is None:
+    # is specified then start will become stop.
+    # However if stop is specified as None, preserve its value.
+    # See logic in slice_new.
```
seberg: I would omit the last two comment lines.  Since this will be `stop is np._NoValue` I think it is clear from the code.  Feels more like an explanation of why we are changing it now.

seberg (commented):
I am slightly curious if the second branch is correct, but I think it should be since it iterates through one by one so that the iteration stopping criteria encodes the incorrect logic.

Thanks for looking into both issues!  I wonder if the `None` change should be mentioned, but overall it seems clear enough (and this function is still fairly new also anyway).

numpy/_core/src/umath/stringdtype_ufuncs.cpp
aaronkollasch: I made `start_bounded` and `stop_bounded` variables as `start_ptr` and `start` were already taken, unless I should try reusing one of those?

seberg (approved):
LGTM, thanks @aaronkollasch.  I think this is good to go in if tests pass (I assume they will).

aaronkollasch:
Sounds great, thanks for reviewing @seberg!

ngoldbaum:
Thanks so much for the fix and for expanding the tests!

Issue #29824, opened 2025-09-28
ENH: `np.vectorize` hides docstrings during jupyter introspection.


jagerber48:
### Proposed new feature or change:

If I write a function in a python module or in my juptyer notebook, when I am writing to call that function in a jupyter notebook, I can press `tab` to see the function signature and docstring. This helps me quickly write code, especially if the function has many arguments. However if I have wrapped the function in `np.vectorize` I instead see documentation for the `vectorize` class. This spoils the benefits of the tab complete tooling available in jupyter and other IDE-like environments.

I would like if I could still see the docstring for my function even when it is wrapped in `np.vectorize`. I know there are tools like `functools.wraps` that can help with this sort of thing, but I also know that it can still be tricky to get the desired behavior. I wonder if this has been considered before for `np.vectorize`?

seberg:
Hmmmm, interesting, `np.vectorize` does set `__doc__` similar to `np.mean` (implementation detail, this gets also wrapped) and using `functools.wraps` doesn't seem sufficient to me?
I must be missing something that I probably knew at some point. IPython does do this nicely for `np.mean` which also is a class (although defined in C?) but doens't show the class docs, but rather the `__doc__` that was set (or maybe via `__wrapped__`).

Curious, did you have a working invocation to fix it?

jagerber48:
@seberg thank you for your quick response and looking into it. No I don't currently have an invocation for this. I was curious to know the prior art within `numpy` before diving into figuring out what is going on. I see the same issue in Pycharm.

Ok, I did a little more testing. So the docstring set for the function DOES appropriately show up. Originally my function had no docstring.

However. what I really want to see is the function signature.

Here is an example without `np.vectorize` where I can helpfully see the function signature

<img width="298" height="176" alt="Image" src="https://github.com/user-attachments/assets/5df96c1c-2b2e-488d-aa6c-563ba12d4e36" />

Here is the same example with `np.vectorize` where I can't see the function signature. However, I can see the docstring.

<img width="766" height="469" alt="Image" src="https://github.com/user-attachments/assets/9ddfa7dd-3b7f-4bfd-996b-913e3dfb6560" />

Here is `np.mean` where I can see the signature.

<img width="616" height="383" alt="Image" src="https://github.com/user-attachments/assets/9a942f51-712e-4892-9d32-9d4e47b3529b" />

[...]

PR #29946, opened 2025-10-14
ENH: fix call signature display for np.vectorize during jupyter

Abhi210:
ENH: fix call signature display for np.vectorize during Jupyter introspection (gh-29824)

This enhancement makes two related improvements to `np.vectorize`:

- Adds `__signature__` and `__wrapped__` attributes to better expose the signature and docstring of the wrapped function.

- Moves the `vectorize` class docstring into `__init__` to avoid cluttering with "Class Docstring" in Jupyter notebook tooltips.

These changes were with user-defined functions, lambda functions, imported functions, and callable classes. The tooltip now correctly displays the function signature and docstring, though IPython still shows a redundant "Call Signature" field.


mattip:
Linting is failing. Can you reflow the help text to 88 or less columns?

Abhi210:
Hello! I have re flowed the help text to less than 88 columns (Checked with ruff tool). Thank you!

seberg:
CC @jagerber48 in case you want to take a look.

Can you check if the signature isn't also a bit of a stretch?  I.e. the original signature probably allows passing non-positionally, but `**args` doesn't really.  Although, if y'all think it's a big improvement even if stretching the truth, I don't mind it much.

jagerber48:
Can we have screenshots of the tooltip posted here like in the the issue thread? @seberg unfortunately I can't understand what you mean about "a bit of a stretch", `**args`, and "stretching the truth". I'll try to test on my own if I get the time.

seberg:
I mean that you can't pass arguments by `a=np.arange(10)` when the original function does, you have to pass positionally, I believe.

Abhi210:
@jagerber48  Here is the screenshot of tooltip before vs after changes :
PR1



@seberg I did not understand your comment (Sorry!). But, did you mean that if I test the following function:

```
@np.vectorize
def vec_foo(a, b, c, d=5):
    return a + b + c**2
```

Define:

```
A = np.arange(3)
B = np.arange(3) + 10
C = np.arange(3) + 100
```

And then try to get the function signature, I should get it properly? If so, here is the image showing these:


PR2

seberg:
Sorry, I meant that the signature looks like you can call the vectorized function with `func(a=1, b=2)` but I think you have to use `func(1, 2)` (positionally). So the signature itself is slightly wrong I think, but it is nicer to have it show up in the docs...

Abhi210:
Ah, I see! You mean that if the wrapped function only accepts positional arguments, then displaying a signature like `func(a, b, c, d=19) ` might make it look like you can call it as `func(a=1, b=2, ...)`, even though keyword arguments wouldn’t actually work. That makes sense. We could add a note to the documentation clarifying that the displayed signature reflects the wrapped Python function. For built-in functions or positional-only callables, the actual calling convention may differ. 

Now, what is the best place to add it? Should we put this warning in the Notes section of numpy.vectorize?

seberg:
Yeah nvm though, seems vectorize does a crazy dance to support kwargs... I did not expect that.

jagerber48:
Ok, this looks good to me. It has the improvements we discussed. 

The only thing I don't like is the pesky "Call Signature" line that is wasting space. But per the discussion at #29824, it doesn't seem like it will be easy to remove that. Perhaps that could be out-of-scope for this PR and can be opened as a separate issue after this one is resolved.

Abhi210:
Thanks for clarifying, @seberg. Good to know about the keyword argument behavior!
And yes, agreed that the “Call Signature” line seems like a separate concern and can be addressed in a follow-up issue after this PR.

Let me know if there’s anything else you’d like me to adjust before merging.

seberg:
Hmmmm, I forgot about this.  Should we just put this in?  I expect the try/except is safe enough.
Are we OK with annotations on the signature being a lie (because input/outputs are now arrays)?

Abhi210:
@seberg, can you do something about this PR? I want to submit another PR for a different issue, but I am getting an error saying I have reached the maximum number of open pull requests.

Abhi210:
I have updated this issue so as to resolve the merge conflict with the latest main.

PR #29947, opened 2025-10-14
BUG: support axis sequence in `np.trim_zeros`

jakevdp:
Fixes #29945

### DIFF
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -1986,6 +1986,14 @@ def trim_zeros(filt, trim='fb', axis=None):
     trim = trim.lower()
     if trim not in {"fb", "bf", "f", "b"}:
         raise ValueError(f"unexpected character(s) in `trim`: {trim!r}")
+    if axis is None:
+        axis_tuple = tuple(range(filt_.ndim))
+    else:
+        axis_tuple = _nx.normalize_axis_tuple(axis, filt_.ndim, argname="axis")
+
+    if not axis_tuple:
+        # No trimming requested -> return input unmodified.
+        return filt
 
     start, stop = _arg_trim_zeros(filt_)
     stop += 1  # Adjust for slicing
@@ -2000,20 +2008,13 @@ def trim_zeros(filt, trim='fb', axis=None):
         if 'b' not in trim:
             stop = (None,) * filt_.ndim
 
-    if len(start) == 1:
-        # filt is 1D -> don't use multi-dimensional slicing to preserve
+    sl = tuple(slice(start[ax], stop[ax]) if ax in axis_tuple else slice(None)
+               for ax in range(filt_.ndim))
+    if len(sl) == 1:
+        # filt is 1D -> avoid multi-dimensional slicing to preserve
         # non-array input types
-        sl = slice(start[0], stop[0])
-    elif axis is None:
-        # trim all axes
-        sl = tuple(slice(*x) for x in zip(start, stop))
-    else:
-        # only trim single axis
-        axis = normalize_axis_index(axis, filt_.ndim)
-        sl = (slice(None),) * axis + (slice(start[axis], stop[axis]),) + (...,)
-
-    trimmed = filt[sl]
-    return trimmed
+        return filt[sl[0]]
+    return filt[sl]
 
 
 def _extract_dispatcher(condition, arr):
--- a/numpy/lib/tests/test_function_base.py
+++ b/numpy/lib/tests/test_function_base.py
@@ -2,6 +2,7 @@ import decimal
 import math
 import operator
 import sys
+import typing
 import warnings
 from fractions import Fraction
 from functools import partial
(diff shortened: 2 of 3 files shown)


numpy/lib/tests/test_function_base.py
```diff
@@ -1469,6 +1469,39 @@ def test_unexpected_trim_value(self, trim):
...
+
+        a_trim0 = np.zeros((2, 4))
+        a_trim0[:, 1:3] = 1
+
+        a_trim1 = np.zeros((4, 2))
+        a_trim1[1:3, :] = 1
+
+        a_trim_both = np.ones((2, 2))
+
+        res = trim_zeros(a, axis=())
```
ngoldbaum: for other reviewers: the current behavior is to raise a TypeError

numpy/lib/tests/test_function_base.py
```diff
@@ -1469,6 +1469,39 @@ def test_unexpected_trim_value(self, trim):
...
+        assert_array_equal(res, a)
+
+        res = trim_zeros(a, axis=0)
+        assert_array_equal(res, a_trim0)
+
+        res = trim_zeros(a, axis=1)
+        assert_array_equal(res, a_trim1)
```
ngoldbaum: these two are testing the current behavior

numpy/lib/tests/test_function_base.py
```diff
@@ -1469,6 +1469,39 @@ def test_unexpected_trim_value(self, trim):
...
+
+        res = trim_zeros(a, axis=(-2,))
+        assert_array_equal(res, a_trim0)
+
+        res = trim_zeros(a, axis=(-1,))
+        assert_array_equal(res, a_trim1)
+
+        res = trim_zeros(a, axis=(0, -1))
+        assert_array_equal(res, a_trim_both)
```
ngoldbaum: All of these currently fail with a TypeError. IMO it makes sense to do something sensible here instead of raising an error.

numpy/lib/tests/test_function_base.py
```diff
@@ -1469,6 +1469,39 @@ def test_unexpected_trim_value(self, trim):
...
+
+        res = trim_zeros(a, axis=None)
+        assert_array_equal(res, a_trim_both)
```
ngoldbaum: This is also a behavior change. I'm not so sure about it. I fear that there might be code relying on this behavior:

```
>>> res = trim_zeros(a, axis=None)
>>> a
array([[0., 0., 0., 0.],
       [0., 1., 1., 0.],
       [0., 1., 1., 0.],
       [0., 0., 0., 0.]])
```

Unless I'm misinterpreting or made a mistake somewhere...

numpy/lib/tests/test_function_base.py
```diff
@@ -1469,6 +1469,39 @@ def test_unexpected_trim_value(self, trim):
...
+
```
ngoldbaum, line 1520: This is not a blocker for merging, just a general idea to improve testing:

I don't think you broke the ND case but IMO it might be a good idea to test with a 3D or higher-dimensional array instead of a 2D array to see if there are any corner cases. Not sure how good the existing test coverage is.

ngoldbaum (changes requested):
Thanks! Sorry for taking a little while to look at this. These sorts of changes to the Python parts of NumPy are always a little inscrutable.

It's borderline but it would probably help any possibly confused users to add a release note. ~~particularly for the `axis=None` change if that was intentional.~~

EDIT: removed an incorrect statement

numpy/lib/_function_base_impl.py
```diff
@@ -1987,6 +1987,14 @@ def trim_zeros(filt, trim='fb', axis=None):
...
     if trim not in {"fb", "bf", "f", "b"}:
         raise ValueError(f"unexpected character(s) in `trim`: {trim!r}")
+    if axis is None:
+        axis_tuple = tuple(range(filt_.ndim))
+    else:
+        axis_tuple = _nx.normalize_axis_tuple(axis, filt_.ndim, argname="axis")
+
+    if not axis_tuple:
+        # No trimming requested -> return input unmodified.
+        return filt
```
seberg: I wonder if we should return a view. But we have been making things more immutable, maybe it doesn't matter much anymore.

numpy/lib/tests/test_function_base.py
ngoldbaum: Oh wait, I'm sorry, total brainfart:

```
>>> res
array([[1., 1.],
       [1., 1.]])
```

ngoldbaum (approved):
NM, this is OK to merge. I confused myself. Added a 2.4.0 milestone to make sure this gets merged before the next release.

jakevdp:
Thanks! I'm happy to expand test coverage if you'd like that here – let me know!

ngoldbaum:
Go ahead if you'd like. It would make me feel a little more confident hitting the merge button without someone else looking at it. A release note about the fix/new feature would be nice too.

jakevdp:
The last commit adds some more comprehensive tests. I added a utility to construct (input, output) pairs in order to avoid having to manually define each test case; let me know if you think this approach is too complicated.

numpy/lib/tests/test_function_base.py
```diff
@@ -2,6 +2,7 @@
 import math
 import operator
 import sys
+import typing
```
ngoldbaum: I don't think we use `typing` at all in any of our tests. The only place it seems to show up is in a vendored test from meson:

```
goldbaum at Nathans-MBP in ~/Documents/numpy on main
± rg "import typing" **/test_*.py
vendored-meson/meson/mesonbuild/scripts/test_loaded_modules.py
5:import typing as T
```

So probably best to just delete the type hints since it makes this code hard to read for someone who is used to looking at other tests in NumPy.

numpy/lib/tests/test_function_base.py
```diff
@@ -1465,6 +1501,29 @@ def test_unexpected_trim_value(self, trim):
...
+        [(5, 6, 7, 8), None],
+        [(5, 6, 7, 8), ()],
+        [(5, 6, 7, 8), -2],
+        [(5, 6, 7, 8), (0, 1, 3)],
+    ])
+    @pytest.mark.parametrize("trim", ['fb', 'f', 'b'])
+    def test_multiple_axes(self, shape, axis, trim):
+        rng = np.random.default_rng(4321)
+        data, expected = self.construct_input_output(rng, shape, axis, trim)
+        assert_array_equal(trim_zeros(data, axis=axis, trim=trim), expected)
```
ngoldbaum, line 1519: this new test takes 0.35 s to run on my M3 macbook pro. I think that's acceptable.

ngoldbaum (commented):
The new test is complicated but easy enough to understand with the comments or by stepping through in a debugger. I'm fine with it. Just one comment about the type hints. It'd also still be really nice to get a release note for this.

numpy/lib/tests/test_function_base.py
jakevdp: Done

jakevdp:
> It'd also still be really nice to get a release note for this.

Done.

ngoldbaum:
Thanks @jakevdp!

PR #29948, opened 2025-10-14
STY: rename ``@classmethod`` arg to ``cls``

jorenham:
For some reason this was causing stubtest to crash :shrug:


charris:
Thanks Joren.

PR #29949, opened 2025-10-14
STY: rename @classmethod arg to cls

charris:
Backport of #29948.

For some reason this was causing stubtest to crash :shrug:


doc/changelog/2.3.4-changelog.rst
jorenham, line 1: is this meant to be in here?

charris: No. Forgot to base off the maintenance branch. Just noticed that.

PR #28576, opened 2025-03-24
API,ENH: Allow forcing an array result in ufuncs

seberg:
As of now a WIP, because it would be nice to see what `_methods.py`.  *Happy if someone just picks it up also, thought I would show how trivial it is now.*

The API change is to allow `out=...`, for multi-argument outputs it is currently `out=(..., ...)`, although I guess we should allow short-handing that (nobody needs to pick it per argument).

I.e. all of these return arrays (and unlike the last time around it works with subclasses, although right now it passes `...` as context so that may confuse them or `__array_ufunc__` implementations.):
```
np.add(1, 2, out=...)
a = np.add.reduce(np.arange(10), out=...)
```

There is some discussion in gh-14489 about this.  I still think `...` is the best, because `out=np.ndarray` doesn't generalize for `__array_ufunc__`, and yes, I like `...` because it also indicates an array return in indexing.  (`arr[i, ...]` always returns an array.)

---

The ufunc part works (not sure if 100% complete), but the `_methods.py` changes do not (and change behavior subtlely, if possibly correctly). The actual change is trivial (after I had refactored this a while ago).

I think we should do this, but the fact that the `_methods.py` changes are not trivial makes me think there may be some idea missing.

It may be that the solution is for these functions to use conversion helper once at the start and then convert to scalars based on that at the end though (rather than dragging around subclasses).

---

CC @mdhaber you may be interested in this.  We could probably just put this through (with a few new tests) without the `_methods.py` changes hashed out.  (There must be a few other low-hanging places where this would simplify code, that are not as maze-like as `_methods.py`.)

PR #29951, opened 2025-10-14
MAINT: replace use of ``asanyarray`` with ``out=...`` to keep arrays

mhvk:
With #28576, it has become possible to ensure that when calling a ufunc, the output is guaranteed to be an array.

This PR uses that to replace `np.asanyarray` calls in some functions. I'm not sure this is complete -- these were just parts of code for which test failed if `np.anyarray(scalar)` is made to return a read-only array (see #29876).

I do think there are all small improvements, though.

### DIFF
--- a/numpy/_core/_methods.py
+++ b/numpy/_core/_methods.py
@@ -185,15 +185,14 @@ def _var(a, axis=None, dtype=None, out=None, ddof=0, keepdims=False, *,
     # Compute sum of squared deviations from mean
     # Note that x may not be inexact and that we need it to be an array,
     # not a scalar.
-    x = asanyarray(arr - arrmean)
-
+    x = um.subtract(arr, arrmean, out=...)
     if issubclass(arr.dtype.type, (nt.floating, nt.integer)):
-        x = um.multiply(x, x, out=x)
+        x = um.square(x, out=x)
     # Fast-paths for built-in complex types
-    elif x.dtype in _complex_to_float:
-        xv = x.view(dtype=(_complex_to_float[x.dtype], (2,)))
-        um.multiply(xv, xv, out=xv)
-        x = um.add(xv[..., 0], xv[..., 1], out=x.real).real
+    elif (_float_dtype := _complex_to_float.get(x.dtype)) is not None:
+        xv = x.view(dtype=(_float_dtype, (2,)))
+        um.square(xv, out=xv)
+        x = um.add(xv[..., 0], xv[..., 1], out=x.real)
     # Most general case; includes handling object arrays containing imaginary
     # numbers and complex types with non-native byteorder
     else:
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -42,6 +42,7 @@ from numpy._core.umath import (
     arctan2,
     cos,
     exp,
+    floor,
     frompyfunc,
     less_equal,
     minimum,
@@ -4679,9 +4680,8 @@ def _lerp(a, b, t, out=None):
     out : array_like
         Output array.
     """
-    diff_b_a = subtract(b, a)
-    # asanyarray is a stop-gap until gh-13105
-    lerp_interpolation = asanyarray(add(a, diff_b_a * t, out=out))
+    diff_b_a = b - a
+    lerp_interpolation = add(a, diff_b_a * t, out=... if out is None else out)
     subtract(b, diff_b_a * (1 - t), out=lerp_interpolation, where=t >= 0.5,
              casting='unsafe', dtype=type(lerp_interpolation.dtype))
     if lerp_interpolation.ndim == 0 and out is None:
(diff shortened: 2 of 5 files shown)


numpy/_core/_methods.py
```diff
@@ -185,15 +185,14 @@ def _var(a, axis=None, dtype=None, out=None, ddof=0, keepdims=False, *,
     # Compute sum of squared deviations from mean
     # Note that x may not be inexact and that we need it to be an array,
     # not a scalar.
-    x = asanyarray(arr - arrmean)
-
+    x = um.subtract(arr, arrmean, out=...)
     if issubclass(arr.dtype.type, (nt.floating, nt.integer)):
-        x = um.multiply(x, x, out=x)
+        x = um.square(x, out=x)
```
mhvk, line 190: Sorry, couldn't resist making a few more small improvements as I was here...

numpy/lib/tests/test_function_base.py
```diff
@@ -3895,7 +3895,7 @@ def test_fraction(self):
 
         q = np.quantile(x, .5)
         assert_equal(q, 1.75)
-        assert_equal(type(q), np.float64)
+        assert isinstance(q, float)
```
mhvk, line 3898: This is a side effect, the function now returns regular `float`. Since here one is dealing with object arrays, that arguably is more appropriate. But it really should not matter at all -- it is not like one could count on an `np.float64`, in most cases one gets a `Fraction`.

numpy/linalg/_linalg.py
```diff
@@ -2016,18 +2016,14 @@ def cond(x, p=None):
...
     if nan_mask.any():
         nan_mask &= ~isnan(x).any(axis=(-2, -1))
         if r.ndim > 0:
             r[nan_mask] = inf
         elif nan_mask:
-            r[()] = inf
-
-    # Convention is to return scalars instead of 0d arrays
-    if r.ndim == 0:
-        r = r[()]
```
mhvk, line 2030: the change below takes care of this now.

mhvk (commented):
A few comments to help review.

numpy/lib/tests/test_function_base.py
seberg: Yeah, I think I had skipped some cases like this when first adding `out=...`, but should have followed up on it.  I think this is fine for object array input... It is arguably a bug-fix.
But it is also a very subtle small change that might break someone in theory!

numpy/lib/_function_base_impl.py
```diff
@@ -4679,9 +4680,8 @@ def _lerp(a, b, t, out=None):
     out : array_like
         Output array.
     """
-    diff_b_a = subtract(b, a)
-    # asanyarray is a stop-gap until gh-13105
-    lerp_interpolation = asanyarray(add(a, diff_b_a * t, out=out))
+    diff_b_a = b - a
```
seberg, line 4683: Is there a reason not to do `subtract(b, a, out=...)` actually and stick to arrays?
(but doesn't matter, TBH.)

numpy/linalg/_linalg.py
seberg: Hmmm, yeah with `r` potentially being scalar might be awkward, but I don't think this makes it less awkward and we use `dtype.type(val)` in enough places that I think we can rely on it.
(It seemed slightly awkward for `object` dtype, but it works even there.)

seberg (approved):
LGTM, thanks.  The change in quantile for object input is very subtle, but I think we should do it and it's technically fixing bugs for decimal or fractional inputs.

I'll wait a bit in case someone has thoughts, but I think we can put this in.

numpy/lib/_function_base_impl.py
mhvk: My logic was that if both are scalars, this is faster. I had hoped more generally to go to just using the operators, but in other places it turned out annoying, and then I decided to keep the changes minimal... Anyway, would suggest to leave it, but can change back if desired.

numpy/lib/tests/test_function_base.py
mhvk: Really only in theory, I think, since even before one couldn't count on getting `np.float64`...

seberg:
Let's get this in then, thanks Marten!

The biggest change (I am not sure how big), is that quantiles propagate Fractions better with this which might change things in some niche situation (arguably a fix, but possibly a subtle change).

neutrinoceros:
FYI it seems that, as a side effect, this fixed some odd behavior we had in unyt
xref: https://github.com/yt-project/unyt/issues/603

I don't understand it myself without a deeper dig, but I thought that'd be interesting to you guys !

mhvk:
@neutrinoceros - nice to hear. Agreed it is not that obvious what change actually made the code work better.

numpy/lib/_ufunclike_impl.py
```diff
@@ -57,7 +57,7 @@ def fix(x, out=None):
 
     """
     # promote back to an array if flattened
-    res = nx.asanyarray(nx.ceil(x, out=out))
+    res = nx.ceil(x, out=... if out is None else out)
```
rhshadrach, line 60: This change is causing a failure in pandas

https://github.com/pandas-dev/pandas/actions/runs/19195436099/job/54875728980?pr=63046#step:5:63

Not clear to me this was intentional. This is complicated by the fact that this was already xfailed on the main branch due to https://github.com/pandas-dev/pandas/pull/51082#issuecomment-1455882487, see also the comment a few below this one.

seberg: Not intentional, but we should allow `out=...` in pandas' `__array_ufunc__` implementation maybe (it is the same as `None` except for reductions).

This allows fixing weird corner cases with object dtype in NumPy, so it would be nice to not stall on it forever.

(I don't mind reverting this or similar cases for now if you prefer.)

seberg: Nvm.  We should just change `fix` to forward to `trunc`, that should be safe and good right now and not worry about this regression.

I'll have a look at the `__array_ufunc__` in pandas so that this type of change doesn't break anything in the future.

PR #29956, opened 2025-10-14
TYP: fix return type annotation for normalize_axis_tuple utility

jakevdp:
The tuple returned from this function can have any length.


jorenham (approved):

jorenham:
Thanks Jake

PR #29957, opened 2025-10-14
TST: Remove recwarn from tests

bwhitt7:
I'm working on making the NumPy test suite more thread-safe and able to be run under `pytest-run-parallel` (talk about it more here #29552). Looks like `recwarn` isn't thread-safe, so this PR introduces a change to make any tests that uses the `recwarn` fixture (so far just 1) use the `warnings.catch_warnings` context instead. The test in question already was using that, so it makes it so `warnings.catch_warnings` records the warnings now instead of `recwarn`.

### DIFF
--- a/numpy/lib/tests/test_nanfunctions.py
+++ b/numpy/lib/tests/test_nanfunctions.py
@@ -723,22 +723,22 @@ class TestNanFunctions_MeanVarStd(SharedNanFunctionsTestsMixin):
                 res = nf(_ndat, axis=1, ddof=ddof)
                 assert_almost_equal(res, tgt)
 
-    def test_ddof_too_big(self, recwarn):
+    def test_ddof_too_big(self):
         nanfuncs = [np.nanvar, np.nanstd]
         stdfuncs = [np.var, np.std]
         dsize = [len(d) for d in _rdat]
         for nf, rf in zip(nanfuncs, stdfuncs):
             for ddof in range(5):
-                with warnings.catch_warnings():
+                with warnings.catch_warnings(record=True) as w:
+                    warnings.simplefilter('always')
                     warnings.simplefilter('ignore', ComplexWarning)
                     tgt = [ddof >= d for d in dsize]
                     res = nf(_ndat, axis=1, ddof=ddof)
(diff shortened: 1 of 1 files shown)


ngoldbaum (approved):
LGTM. This also follows the pattern used elsewhere in this file (e.g. in `TestNanFunctions_MinMax.test_subclass`) to record and check for warnings. No other tests in all of numpy use `recwarn`. It's worth eliminating for that reason alone, even if we weren't trying to set up pytest-run-parallel.

numpy/lib/tests/test_nanfunctions.py
```diff
@@ -723,22 +723,22 @@ def test_ddof(self):
...
 
-    def test_ddof_too_big(self, recwarn):
+    def test_ddof_too_big(self):
         nanfuncs = [np.nanvar, np.nanstd]
         stdfuncs = [np.var, np.std]
         dsize = [len(d) for d in _rdat]
         for nf, rf in zip(nanfuncs, stdfuncs):
             for ddof in range(5):
-                with warnings.catch_warnings():
+                with warnings.catch_warnings(record=True) as w:
```
tylerjereddy, line 732: FWIW, it looks like the concurrency safety guarantee of `catch_warnings` is only strong at CPython >= 3.14 per https://docs.python.org/3/library/warnings.html#concurrent-safety-of-context-managers.

I suspect that isn't too surprising though, with 3.14 generally considered better for free-threading support.

ngoldbaum: Yes - maybe we should document somewhere that we're assuming people doing parallel testing are using 3.14t or 3.14 with context-aware warnings turned on? Otherwise you'll see a lot of failed tests due to warnings issues.

ngoldbaum: @bwhitt7 has a followup to set up a pytest-run-parallel CI job by repurposing the 3.14t MacOS CI job.

bwhitt7: Yeah generally we expect folks to run the tests in parallel using 3.14t, should definitely document that somewhere.

PR #29958, opened 2025-10-14
TYP: Fix inconsistent ``__all__`` stubs

jorenham:
They're not the most interesting of fixes, but stubtest was complaining about them, so this will bring us a bit closer to being able to start use stubtest in CI ("only" 527 errors left).


mattip:
Thanks @jorenham

PR #29960, opened 2025-10-14
DOC: fix formatting in ``np.percentile`` docstring

jakevdp:
The extra space here leads to `weights` appearing indented under `keepdims`: https://numpy.org/doc/2.3/reference/generated/numpy.percentile.html


mattip:
Thanks @jakevdp

PR #29961, opened 2025-10-14
TYP: update the ``finfo`` stubs

jorenham:
This syncs the stubs with the runtime changes made in #29836.


mattip:
Thanks @jorenham

PR #29962, opened 2025-10-14
MAINT: remove obsolete ``generic.tostring`` method descriptor and docstring

jorenham:
The method was removed in 2.3, but the method descriptor was still there on the `numpy.generic` type, which could be pretty confusing:

```pycon
>>> import numpy as np
>>> np.int8.tostring
<method 'tostring' of 'numpy.generic' objects>
>>> np.int8().tostring()
    np.int8().tostring()
    ~~~~~~~~~~~~~~~~~~^^
AttributeError: 'numpy.ndarray' object has no attribute 'tostring'
```


mattip:
Thanks @jorenham

PR #29963, opened 2025-10-15
MAINT: Remove removed array methods

jorenham:
(that's not a typo in the title)

The `ptp`, `itemset`, and `newbyteorder` methods of `ndarray` and `generic` were removed in numpy 2. However, they still existed, but attempting to call them resulted in an `AttributeError` with a paradoxical message

> `ptp` was removed from the ndarray class in NumPy 2.0. Use np.ptp(arr, ...) instead.

So that got me thinking: "how can something that is removed raise an error?". After a bit of digging, I found the relevant C code, and got rid of it at the root. That way, we avoid causality-breaking parodoxes, making it less likely that the universe will collapse in on itself.

Before:

```py
>>> np.ndarray.ptp
<attribute 'ptp' of 'numpy.ndarray' objects>

>>> np.array(1).ptp()
AttributeError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 np.array(1).ptp()

AttributeError: `ptp` was removed from the ndarray class in NumPy 2.0. Use np.ptp(arr, ...) instead.
```

After:

```py
>>> np.ndarray.ptp
AttributeError                            Traceback (most recent call last)
Cell In[1], line 1
----> 1 np.ndarray.ptp

AttributeError: type object 'numpy.ndarray' has no attribute 'ptp'

>>> np.array(1).ptp()
AttributeError                            Traceback (most recent call last)
Cell In[4], line 1
----> 1 np.array(1).ptp()

AttributeError: 'numpy.ndarray' object has no attribute 'ptp'
```

It's the same with `itemset` and  `newbyteorder`.

### DIFF
--- a/numpy/_core/src/multiarray/getset.c
+++ b/numpy/_core/src/multiarray/getset.c
@@ -865,34 +865,6 @@ array_matrix_transpose_get(PyArrayObject *self, void *NPY_UNUSED(ignored))
     return PyArray_MatrixTranspose(self);
 }
 
-static PyObject *
-array_ptp(PyArrayObject *self, void *NPY_UNUSED(ignored))
-{
-    PyErr_SetString(PyExc_AttributeError,
-                    "`ptp` was removed from the ndarray class in NumPy 2.0. "
-                    "Use np.ptp(arr, ...) instead.");
-    return NULL;
-}
-
-static PyObject *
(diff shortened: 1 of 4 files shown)


mattip:
A bit surprising that no test changes were needed, but CI is passing. Thanks @jorenham

mattip:
I don't think we need a release note, these were raising an error anyway.

PR #29964, opened 2025-10-15
TYP: add missing ``generic`` methods

jorenham:
Stubtest was complaining about missing methods, so I added them. But some of them always raise when used, so maybe we should just get rid of those as well (similar to #29962 and #29963)?


numpy/__init__.pyi
```diff
@@ -3599,10 +3638,33 @@ class generic(_ArrayOrScalarCommon, Generic[_ItemT_co]):
...
+    ) -> Never: ...
+    def diagonal(self: Never, /, offset: Never = ..., axis1: Never = ..., axis2: Never = ...) -> Never: ...  # type: ignore[misc]
+    def swapaxes(self: Never, /, axis1: Never, axis2: Never) -> Never: ...  # type: ignore[misc]
+    def sort(self: Never, /, axis: Never = ..., kind: Never = ..., order: Never = ...) -> Never: ...  # type: ignore[misc]
+    def nonzero(self: Never, /) -> Never: ...  # type: ignore[misc]
+    def setfield(self: Never, /, val: Never, dtype: Never, offset: Never = ...) -> None: ...  # type: ignore[misc]
+    def searchsorted(self: Never, /, v: Never, side: Never = ..., sorter: Never = ...) -> Never: ...  # type: ignore[misc]
+
+    # NOTE: this wont't raise, but won't do anything either
+    def resize(self, new_shape: L[0, -1] | tuple[L[0, -1]] | tuple[()], /, *, refcheck: builtins.bool = False) -> None: ...
```
jorenham, line 3662: > But some of them always raise when used, so maybe we should just get rid of those as well (similar to https://github.com/numpy/numpy/pull/29962 and https://github.com/numpy/numpy/pull/29963)?

So these ones is what I'm talking about here

mattip:
Should we revisit this after removing the deprecated methods or merge this now and fix up later?

jorenham:
> Should we revisit this after removing the deprecated methods or merge this now and fix up later?

I don't think these are deprecated; they just always raise a `ValueError` (or `AxisError`) when called (or do nothing in case of `resize`):

```python
In [1]: np.int8().diagonal()
Cell In[1], line 1
----> 1 np.int8().diagonal()

ValueError: diag requires an array of at least two dimensions

In [2]: np.int8().sort()
AxisError                                 Traceback (most recent call last)
Cell In[2], line 1
----> 1 np.int8().sort()

AxisError: axis -1 is out of bounds for array of dimension 0

In [3]: np.int8().nonzero()
Cell In[3], line 1
----> 1 np.int8().nonzero()

ValueError: Calling nonzero on 0d arrays is not allowed. Use np.atleast_1d(scalar).nonzero() instead.
```

So I think we should deprecate them first then.

mattip:
Thanks @jorenham

PR #29965, opened 2025-10-15
TYP: mark ``flexible`` as ``@final``

jorenham:
Because even though it has subclasses (`character`, `void`, `str_` and `bytes_`), it's impossible to subclass it at runtime.


mattip:
Thanks @jorenham

PR #29967, opened 2025-10-15
TYP: improved ``busdaycalendar`` annotations

jorenham:
The switch from `__new__` to `__init__` resolves a stubtest error.


mattip:
Thanks @jorenham

PR #29969, opened 2025-10-15
TYP: fix stubtest errors in ``lib._function_base_impl``

jorenham:
- `corrcoef`: wrong parameter name
- `diff`: default argument not assignable to parameter type
- ~`percentile` and `quantile`: missing (deprecated) `interpolation` kwarg~
- `piecewise`: the first three parameters shouldn't be positional-only
- `trim_zeros`: missing `axis` parameter (since 2.2.0)

### DIFF
--- a/numpy/lib/_function_base_impl.pyi
+++ b/numpy/lib/_function_base_impl.pyi
@@ -626,7 +626,6 @@ def percentile(
     keepdims: L[False] = False,
     *,
     weights: _ArrayLikeFloat_co | None = None,
-    interpolation: None = None,  # deprecated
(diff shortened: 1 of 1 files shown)


mattip:
`interpolation` as a kwarg was deprecated in 1.2. Perhaps it is time to expire the deprecation instead of adding it to typing

jorenham:
> `interpolation` as a kwarg was deprecated in 1.2. Perhaps it is time to expire the deprecation instead of adding it to typing

Yea seems like the way to go; I'll get on that then.

mattip:
Thanks @jorenham

PR #29970, opened 2025-10-15
BUG: Fix resize when it contains references

### LINKED ISSUE #29887: BUG: shrinking an object array with .resize does not decref the shrunk items
### Describe the issue:

While investigating `array.resize` in the context of #29878, I noticed that it seemed not to worry at all about the array containing references. This is fine if the array is expanded, but not if it is shrunk.

I think in principle, it is not difficult to solve, though definitely annoying, since an object can be inside a structured array too. Possibly, it makes more sense just to forbid resizing of arrays that hold references.

```python
import numpy as np
import sys
import gc

a = object()
b = np.array([a, a])
sys.getrefcount(a)
# 4
b.resize(1)
gc.collect()
sys.getrefcount(a)
# 4
del b
gc.collect()
sys.getrefcount(a)
# 3  (without the resize, this correctly states 2)
```

### Error message:

```shell

```

2.2.4
3.13.7 (main, Aug 20 2025, 22:17:40) [GCC 14.3.0]

### Runtime Environment:

Not relevant

### Context for the issue:

_No response_

seberg:
Small fix for resize with objects, which was always broken.

@mhvk one little thing is that we of course don't need to do this in all of the internal places in theory.

Closes gh-29887.

### DIFF
--- a/numpy/_core/tests/test_multiarray.py
+++ b/numpy/_core/tests/test_multiarray.py
@@ -6311,14 +6311,14 @@ class TestResize:
         assert_array_equal(a['k'][-5:], 0)
         assert_array_equal(a['k'][:-5], 1)
 
-    @_no_tracing
+    @pytest.mark.skipif(not HAS_REFCOUNT, reason="Python lacks refcounts")
     @pytest.mark.parametrize("dtype", ["O", "O,O"])
     def test_obj_obj_shrinking(self, dtype):
         # check that memory is freed when shrinking an array.
         test_obj = object()
         expected = sys.getrefcount(test_obj)
         a = np.array([test_obj, test_obj, test_obj], dtype=dtype)
-        a.size == 3
+        assert a.size == 3
         a.resize((2, 1))  # two elements, not three!
         assert a.size == 2
         del a
--- a/numpy/_core/tests/test_multiarray.py
+++ b/numpy/_core/tests/test_multiarray.py
@@ -6311,6 +6311,20 @@ class TestResize:
         assert_array_equal(a['k'][-5:], 0)
         assert_array_equal(a['k'][:-5], 1)
 
+    @_no_tracing
+    @pytest.mark.parametrize("dtype", ["O", "O,O"])
+    def test_obj_obj_shrinking(self, dtype):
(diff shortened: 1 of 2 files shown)


numpy/_core/tests/test_multiarray.py
```diff
@@ -6311,6 +6311,20 @@ def test_obj_obj(self):
...
         assert_array_equal(a['k'][:-5], 1)
 
+    @_no_tracing
+    @pytest.mark.parametrize("dtype", ["O", "O,O"])
+    def test_obj_obj_shrinking(self, dtype):
+        # check that memory is freed when shrinking an array.
+        test_obj = object()
+        expected = sys.getrefcount(test_obj)
+        a = np.array([test_obj, test_obj, test_obj], dtype=dtype)
+        a.size == 3
```
mhvk: Missing `assert` here!

mhvk (commented):
Looks good, just one missing `assert`.

mhvk:
> @mhvk one little thing is that we of course don't need to do this in all of the internal places in theory.

For object arrays, I think it pays to just be correct -- and everything is slow anyway,

numpy/_core/tests/test_multiarray.py
```suggestion
        assert a.size == 3
```

numpy/_core/tests/test_multiarray.py
```diff
@@ -6311,6 +6311,20 @@ def test_obj_obj(self):
...
+        # check that memory is freed when shrinking an array.
+        test_obj = object()
+        expected = sys.getrefcount(test_obj)
+        a = np.array([test_obj, test_obj, test_obj], dtype=dtype)
+        assert a.size == 3
+        a.resize((2, 1))  # two elements, not three!
+        assert a.size == 2
+        del a
+        # if all is well, then we reclaimed all references
+        assert sys.getrefcount(test_obj) == expected
```
ngoldbaum, line 6326: not available on pypy

mhvk (approved):
Looks good!

ngoldbaum:
Thanks @seberg thanks for the nudge.

PR #29971, opened 2025-10-15
TYP: update ``ScalarType`` type

jorenham:
This is needed because *someone* forgot to update the stubs after changing the runtime order of `ScalarType` in #29761.
Since #29761 was backported, this one probably should too.


mattip:
Docs CI failed, it could not download the scipy.org/scipy intersphinx file. This was happening earlier today as well. I restarted the circleci CI run.

mattip:
Thanks @jorenham

PR #29972, opened 2025-10-15
TYP: expand ``TypedDict`` kwargs in ``full`` to appease stubtest

jorenham:
It's a bit less DRY, but this way the default values are also known, which can help for documentation/introspection purposes.


mattip:
Oh well. So many overloads needed. Thanks @jorenham

jorenham:
> Oh well. So many overloads needed. Thanks @jorenham

This isn't even that bad actually; in [scipy-stubs](https://github.com/scipy/scipy-stubs/) there are a couple of methods with 60+ overloads 😅

Issue #11521, opened 2018-07-06
DEP: remove code with expired deprecations (tracking issue)


mattip:
A tracking issue of features with `DeprecationWarning`. You can grep around for deprecations in python code with `git grep "deprecated:: 1"` and `git grep "deprecated:: 2.0"`, and in C via `git grep "DEPRECATE("`

charris:
We have just left the ones in the API, there are no plans to remove them. There are a *lot* of later ones, they are still there because I am lazy :) Which I rationalize as not wanting too much churn before dropping 2.7. We might want to make a list and set up a schedule for removal.

YousefH:
@charris @mattip  I wanted to open an issue about the use of a deprecated kwarg, namely the `normed` kwarg, in matplotlib's `hist()` function, in an example in numpy's doc, but then saw this issue. To be honest after reading this I am still not sure, if it referenced the issue that I am referring to or not, so I was hesitant on whether to open a new issue or not. I will just link the page here and hope it will be helpful.

https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.uniform.html

rgommers:
> The deprecate decorator is used only for np.unique and np.testing.random, perhaps we should deprecate it too.

It's part of the public API and used regularly by other libraries (e.g. the majority of deprecations in `scipy`).

rgommers:
Thanks @mattip, this overview is quite helpful. After we've formulated the deprecation policy, probably the first step would be to go in and fix the deprecation messages. E.g. add when the function was deprecated, and when it will be removed.

rgommers:
Setting to the 1.16.0 milestone so we remember to update the deprecation message before the next release.

mattip:
@rgommers Did we fomulate a specific text /policy for deprecation messages? For instance, the warning in [arrayprint.FloatFormat]() is:

```
warnings.warn("FloatFormat has been replaced by FloatingFormat",
                      DeprecationWarning, stacklevel=2)
```
Should this become
```
warnings.warn("FloatFormat was replaced by FloatingFormat and will be removed in v1.17",
                      DeprecationWarning, stacklevel=2)
```
Should any of the old deprecations be turned into errors (now, for 1.16) for a few versions before removing?

shoyer:
@mattip I think you can remove `np.unique` from that list. The deprecated function is `np.lib.function_base.unique()`, but `np.unique` has pointed to `np.lib.arraysetops.unique` for a very long time.

[...]

PR #29973, opened 2025-10-15
DEP: Remove deprecated ``interpolation`` parameter from quantile/percentile

jorenham:
The `interpolation` kwargs of `[nan]percentile` and `[nan]quantile` were deprecated almost four years ago since 1.22.0.

ref: https://github.com/numpy/numpy/pull/29969#issuecomment-3404812838
towards #11521


mattip:
Thanks @jorenham

PR #29975, opened 2025-10-15
CI: Add 3.15 for CI on Linux, macOS, and Windows

clin1234:


.github/workflows/windows.yml
```diff
@@ -30,7 +30,12 @@ jobs:
     - name: Setup Python
       uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v6.0.0
       with:
-        python-version: "3.14t-dev"
+        python-version: |
+          3.14
+          3.14t
+          3.15-dev
+          3.15t-dev
```
tylerjereddy: Does this actually work? I'm pretty sure it doesn't--looking at the CI that runs, I think you're only getting the last entry here `3.15t-dev`. Normally I'd use a `matrix` structure to split a job over multiple Python versions. In this case it isn't clear that we want that though--what's the motivation for trying to add so many more Python versions vs. trying to do the same thing as was done before with a single free-threaded pre-release version?

.github/workflows/macos.yml
```diff
@@ -115,7 +115,7 @@ jobs:
         build_runner:
           - [ macos-15-intel, "macos_x86_64" ]
           - [ macos-14, "macos_arm64" ]
-        version: ["3.11", "3.14t-dev"]
+        version: ["3.11", "3.14", "3.14t", "3.15-dev", "3.15t-dev"]
```
tylerjereddy: This causes a fairly substantial combinatorial growth of runs in the build matrix--maybe we don't need all 5 entries?

ngoldbaum:
I don't think we should add explicit testing for the `main` branch of CPython. We usually start testing after the first beta release.

charris:
>explicit testing for the main branch of CPython.

Yep, we will probably start testing around next June, and we will drop 3.11 support first.

mattip:
Let's close this for now, it will bit-rot too much by June. Please try again then @clin1234

PR #29976, opened 2025-10-15
TYP: fix ``random.Generator.shuffle`` input type

### LINKED ISSUE #29974: TYP: Type hint for np.random.Generator.shuffle is wrong
### Describe the issue:

The type hint for np.random.Generator.shuffle is ArrayLike ([here](https://github.com/numpy/numpy/blob/a1467b5e35bc345fcb6204512e827d0d6a36d4f8/numpy/random/_generator.pyi#L853)), but the documentation says it takes a ndarray or MutableSequence. There are a few differences e.g.:
- A `list[object]` is a MutableSequence but not ArrayLike
- A `complex` is ArrayLike but not an ndarray nor MutableSequence (passing it crashes at runtime because it doesn't support `len`)

This looks similar to #19507, but that was claiming a type-check error when passing `list[int]` (which does actually work, because it is ArrayLike) and was found to be due to PyCharm getting type annotations from some external source.

```python
import numpy as np

rng = np.random.default_rng()
lst = [object(), object()]
rng.shuffle(lst)
rng.shuffle(1 + 2j)
```

### Error message:

```shell
shuffle_type.py:5: error: Argument 1 to "shuffle" of "Generator" has incompatible type "list[object]"; expected "Buffer | _SupportsArray[dtype[Any]] | _NestedSequence[_SupportsArray[dtype[Any]]] | complex | bytes | str | _NestedSequence[complex | bytes | str]"  [arg-type]


It should instead be complaining about like 6, where 1 + 2j is not valid for shuffle.
```

2.3.4

mypy 1.14.1

No command-line options beyond the filename, and no config file.

_No response_

jorenham:
Closes #29974

### DIFF
--- a/numpy/random/_generator.pyi
+++ b/numpy/random/_generator.pyi
@@ -1,4 +1,4 @@
-from collections.abc import Callable
+from collections.abc import Callable, MutableSequence
 from typing import Any, Literal, TypeAlias, TypeVar, overload
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


mattip:
Thanks @jorenham

PR #29981, opened 2025-10-16
TYP: change ``ndenumerate.__new__``  into ``__init__``

jorenham:
Because apparently it's slightly less incorrect this way. (But I don't think that backporting this is worthwhile.)


mattip:
Thanks @jorenham

PR #29982, opened 2025-10-16
TYP: change ``nditer.__new__``  into ``__init__`` and tighten its signature

jorenham:
stubtest was complaining about the `__new__` not being an `__init__`, which indeed seems to be less incorrect. This also overloads the `nditer.__init__` method in order to avoid certain false negatives.


mattip:
Thanks @jorenham

PR #29983, opened 2025-10-16
TYP: minor fixes and improvements in ``record`` and ``recarray``

jorenham:
No need to backport this though.


mattip:
Thanks @jorenham

PR #29984, opened 2025-10-16
DEP: remove the ``fix_imports`` parameter from ``save()``

jorenham:
It was deprecated 17 months ago in 2.1.

Towards #11521


mattip:
I think this one needs to wait for 2.5.

jorenham:
> I think this one needs to wait for 2.5.

True, 17 months is a bit shorter than usual, but I don't see why anyone would still using `fix_imports`. It was a Python 2 thing, no?

mattip:
Thanks @jorenham

PR #29986, opened 2025-10-16
DEP: Remove ``ndarray.ctypes.get_*`` methods (deprecated since 1.21)

jorenham:
This removes the following deprecated `numpy._core._internal._ctypes` methods:

- `get_data()`
- `get_shape()`
- `get_strides()`
- `get_as_parameter()`

Towards #11521


mattip:
Thanks @jorenham

PR #29909, opened 2025-10-09
DEP: finalize deprecation of numpy/fft.helpers and numpy.linalg.linalg (closes #28983)

anikchand461:
Post-2.0 cleanup: Remove deprecated shims and add release note.

Changes:
- Delete `numpy/linalg/linalg.py`
- Delete `numpy/fft/helpers.py`
- Add release note: `doc/release/upcoming/29000.expiring.rst` (placeholder; update to PR # post-creation)

No breaking changes—public APIs (e.g., `np.linalg.inv`, `np.fft.fft`) unchanged. CI will verify tests.

Bundle per @mattip's guidance in #28983. Closes #28983.

PR #29988, opened 2025-10-16
MAINT: remove remnants of ``linalg.linalg`` and ``fft.helper``

jorenham:
This should've been done in #29909


mattip:
Thanks @jorenham. Good catch, sorry i missed that

jorenham:
> sorry i missed that

right back at ya

PR #29989, opened 2025-10-16
BUG: Fix np.strings.slice if start > stop

aaronkollasch:
Found another bug in `np.strings.slice` while running the existing test parameters on longer strings.

If the slice start > stop, then `slice_strided_loop()` attempts to allocate a string of negative size, causing a MemoryError. To replicate:
```
>>> a = np.array(['test-strings'], dtype="T")
>>> np.strings.slice(a, 6, 5)
  File "lib/python3.12/site-packages/numpy/_core/strings.py", line 1813, in slice
    return _slice(a, start, stop, step)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
MemoryError: Failed to allocate string in slice
```


mattip:
Thanks @aaronkollasch. @charris - is this something you want to backport?

aaronkollasch:
Thanks @mattip!
@charris re backporting, the second commit in #29944 adds another missed bounds check to consider

PR #29990, opened 2025-10-16
TYP: missing parameter default in ``reshape`` for ``newshape``

jorenham:
Stubtest was complaining here, but I'm guessing that It's a bit too early to remove it, so II figured that we might as well annotate it correctly for the time being.

### DIFF
--- a/numpy/_core/fromnumeric.pyi
+++ b/numpy/_core/fromnumeric.pyi
@@ -193,6 +193,7 @@ def reshape(  # shape: index
     shape: SupportsIndex,
     order: _OrderACF = "C",
     *,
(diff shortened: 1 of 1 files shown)


seberg (approved):
LGTM, but we have a 1-year minimum policy normally.  And that would make this OK to finalize.  Considering that I think passing `newshape=` is very niche, we could also go with that.

jorenham:
> LGTM, but we have a 1-year minimum policy normally. And that would make this OK to finalize. Considering that I think passing `newshape=` is very niche, we could also go with that.

Ah nice, let's get rid of it then :)

PR #29991, opened 2025-10-16
TYP: some minor fixes for the constants in ``_core.multiarray``

jorenham:
`MAXDIMS` was still set at 32, and most others weren't marked as final (i.e. immutable(-ish)), or were missing a value stub.


mattip:
Thanks @jorenham

PR #29993, opened 2025-10-16
MAINT: avoid namespace pollution in ``_core._type_aliases``

jorenham:
This was causing stubtest to report these errors:

```
numpy._core._type_aliases.abstract_type is not present in stub
numpy._core._type_aliases.base_name is not present in stub
numpy._core._type_aliases.bits is not present in stub
numpy._core._type_aliases.concrete_type is not present in stub
numpy._core._type_aliases.extended_prec_name is not present in stub
numpy._core._type_aliases.full_name is not present in stub
numpy._core._type_aliases.is_complex is not present in stub
numpy._core._type_aliases.k is not present in stub
numpy._core._type_aliases.longdouble_type is not present in stub
numpy._core._type_aliases.sctype_key is not present in stub
numpy._core._type_aliases.sctype_list is not present in stub
numpy._core._type_aliases.type_group is not present in stub
numpy._core._type_aliases.type_info is not present in stub
numpy._core._type_aliases.v is not present in stub
```

And for those of us that use an IDE, these names would also show up in things like import autocompletion suggestions. 
For runtime introspection using `dir()` or something this could matter.

### DIFF
--- a/numpy/_core/_type_aliases.py
+++ b/numpy/_core/_type_aliases.py
@@ -36,6 +36,7 @@ _abstract_type_names = {
 
 for _abstract_type_name in _abstract_type_names:
     allTypes[_abstract_type_name] = getattr(ma, _abstract_type_name)
+    del _abstract_type_name
 
 for k, v in typeinfo.items():
     if k.startswith("NPY_") and v not in c_names_dict:
@@ -44,6 +45,8 @@ for k, v in typeinfo.items():
         concrete_type = v.type
         allTypes[k] = concrete_type
         sctypeDict[k] = concrete_type
+        del concrete_type
+    del k, v
 
 _aliases = {
     "double": "float64",
@@ -60,6 +63,7 @@ _aliases = {
 for k, v in _aliases.items():
     sctypeDict[k] = allTypes[v]
     allTypes[k] = allTypes[v]
+    del k, v
 
 # extra aliases are added only to `sctypeDict`
 # to support dtype name access, such as`np.dtype("float")`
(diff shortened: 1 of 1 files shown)


mattip:
@mtsokol could you give this a glance?

jorenham:
BTW, if you're not a fan of the self-executing function approach I used here, we could also go with a classic `del ...` approach.

mtsokol:
@jorenham Can we `del` these two new functions instead? I think people are more familiar with that approach.

mtsokol:
Apart from it LGTM

mtsokol (approved):

numpy/_core/_type_aliases.py
```diff
@@ -76,18 +80,21 @@
...
 for is_complex, full_name in [(False, "longdouble"), (True, "clongdouble")]:
-    longdouble_type: type = allTypes[full_name]
+    longdouble_type = allTypes[full_name]
 
-    bits: int = dtype(longdouble_type).itemsize * 8
-    base_name: str = "complex" if is_complex else "float"
-    extended_prec_name: str = f"{base_name}{bits}"
+    bits = dtype(longdouble_type).itemsize * 8
+    base_name = "complex" if is_complex else "float"
+    extended_prec_name = f"{base_name}{bits}"
```
jorenham, line 91: This probably looks very out of character, but on Python >=3.14 these were causing a module-level `__conditional_annotations__` name to be generated (no mention of this anywhere in the python docs, but it's a `set` of integer indices 🤷🏻). That was causing stubtest to report the following error:

```
numpy._core._type_aliases.__conditional_annotations__ is not present in stub
```

So it was indirectly also causing namespace pollution in that sense.

Usually I tend to ignore these kinds of stubtest errors, but in this case that'd be pretty annoying, seeing as it depends on the Python version. Ignoring errors in stubtest involves an "allowlist", and stubtest will also report an error for "unused allowlist entries". So to ignore this, we'd have to write a wrapper script that would conditionally select the allowlist, depending on the Python version used, and call stubtest with that.

And since these annotations were redundant anyways (type-checkers can infer those types themselves just fine), I thought we might as well get rid of them.

mtsokol: Thanks for an explanation! Yes, in that context it makes sense.

mattip:
~I think moving all this into a function would avoid the namespace pollution.~

Let's stick with this approach

mattip:
Thanks @jorenham

PR #29994, opened 2025-10-16
DEP: remove the ``newshape`` parameter from ``reshape()``

jorenham:
ref https://github.com/numpy/numpy/pull/29990#pullrequestreview-3343412273

cc @seberg

### DIFF
--- a/doc/release/upcoming_changes/29994.expired.rst
+++ b/doc/release/upcoming_changes/29994.expired.rst
@@ -1,0 +1,7 @@
+Remove ``newshape`` parameter from ``numpy.reshape``
+----------------------------------------------------
+
+The ``newshape`` parameter was deprecated in NumPy 2.1.0 and has been
+removed from ``numpy.reshape``.
+
+Use the ``shape`` parameter instead.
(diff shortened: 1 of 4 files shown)


doc/release/upcoming_changes/29994.expired.rst
```diff
@@ -0,0 +1,7 @@
+Remove ``newshape`` parameter from ``numpy.reshape``
+----------------------------------------------------
+
+The ``newshape`` parameter was deprecated in NumPy 2.1.0 and has been
+removed from ``numpy.reshape``.
+
+Use the ``shape`` parameter instead.
```
seberg:
```suggestion
Remove ``newshape`` parameter from ``numpy.reshape``
----------------------------------------------------

The ``newshape`` parameter was deprecated in NumPy 2.1.0 and has been
removed from ``numpy.reshape``. Pass it positionally or use ``shape=``
on newer NumPy versions.
```
Not sure, but might be nice to nudge towards positionally, since `shape=` would fail on old versions of NumPy?

(A bullet would also do, but that's just me trying to be short.)

seberg (approved):
LGTM, thanks Joren!  Might tweak the release notes to nudge towards passing positionally, since passing `shape=` would break compatible with old NumPy versions.

mattip:
Thanks @jorenham

PR #30000, opened 2025-10-16
ENH: Add ``stable`` kwarg to ``chararray.argsort``

jorenham:
For compatibility with `ndarray.argsort` (it was violating the Liskov substitution principle).

### DIFF
--- a/numpy/_core/defchararray.py
+++ b/numpy/_core/defchararray.py
@@ -718,7 +718,7 @@ class chararray(ndarray):
     def __rmod__(self, other):
         return NotImplemented
 
-    def argsort(self, axis=-1, kind=None, order=None):
+    def argsort(self, axis=-1, kind=None, order=None, *, stable=None):
         """
         Return the indices that sort the array lexicographically.
 
@@ -736,7 +736,7 @@ class chararray(ndarray):
               dtype='|S5')
 
         """
-        return self.__array__().argsort(axis, kind, order)
+        return self.__array__().argsort(axis, kind, order, stable=stable)
     argsort.__doc__ = ndarray.argsort.__doc__
 
     def capitalize(self):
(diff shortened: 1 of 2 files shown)


jorenham:
circleci failure seems unrelated

tylerjereddy (commented):
Obviously it would be straightforward to write a test here.

<details>
<summary> Sample regression test that fails on main, passes here </summary>

```diff
--- a/numpy/_core/tests/test_defchararray.py
+++ b/numpy/_core/tests/test_defchararray.py
@@ -694,6 +694,11 @@ def B(self):
         return np.array([['efg', '456'],
                          ['051', 'tuv']]).view(np.char.chararray)
 
+    def test_argsort(self):
+        arr = np.array(['abc'] * 4).view(np.char.chararray)
+        actual = arr.argsort(stable=True)
+        assert_array_equal(actual, [0, 1, 2, 3])
```

</details>

That said, I wonder how much appetite there is for making changes here, especially if marked as an enhancement, given that `chararray` is documented for backcompat only and not for new development.

jorenham:
> Obviously it would be straightforward to write a test here.
> 
> Sample regression test that fails on main, passes here

Yea sure; I guess it felt a bit obvious to explicitly test. But it's indeed pretty trivial and couldn't hurt.

> That said, I wonder how much appetite there is for making changes here, especially if marked as an enhancement, given that `chararray` is documented for backcompat only and not for new development.

Hm yea, you have a point. But it was either this, or override this method in the chararray stubs and add `# type: ignore[override]`. And I figured that since it's not deprecated (yet?), we might as well keep it compatible with `ndarray` (in terms of the Liskov substitution principle). Because otherwise we could get problematic situations like this one:

```py
def f(a: ndarray) -> ndarray:
    return a.argsort(stable=True)

a = np.array(["a", "b"]).view(np.char.chararray)
f(a)  # type-checker doesn't mind; raises TypeError at runtime
```

seberg:
Seems fine to add if it makes some linting/style stuff happy, thanks.

PR #30001, opened 2025-10-17
TYP: fix ``ndarray.sort(stable=True)``

jorenham:
The `stable` parameter of `ndarray.[arg]sort` and `generic.argsort` was annotated to only accept `np.bool` values, causing them to reject calls with `stable=True`.


mattip:
Thanks @jorenham

PR #30002, opened 2025-10-17
TYP: inconsistent ``strings.slice`` default argument for ``stop``

jorenham:
Mostly meant to appease stubtest, so I don't think it's worth backporting.


mattip:
Thanks @jorenham

PR #30003, opened 2025-10-17
TYP: remove implicit re-export in ``_core._exceptions``

jorenham:
Because it's not there at runtime either.

I believe that this takes care of the last (fixable) stubtest errors in `numpy._core` (assuming that #29991, #29993, #29994, #29999, and #30002 get merged).


mattip:
Thanks @jorenham

PR #30006, opened 2025-10-17
TYP: fix stubtest errors in ``numpy.lib.*``

jorenham:
There's still one error that remains. But that's on the runtime side of things, so I'll follow up on that in a separate PR.


mattip:
Thanks @jorenham

PR #28035, opened 2024-12-18
MAINT: Move `lib.format` and `ctypeslib` to submodules/private files

mtsokol:
Hi!

This PR is a backward compatible refactor of `np.lib.format` and `np.ctypeslib` to avoid exporting irrelevant objects.

These modules currently export some redundant items, such as:
- `np.lib.format.io`, `np.lib.format.os`, `np.lib.format.warnings`, etc.
- `np.ctypeslib.np` or `np.ctypeslib.os`.

These are mostly details but I think these were missing from Python API cleanup NEP. I moved both entities to dedicated directory/private files. 

AFAIC these changes don't touch the current state of the public API.

PR #30007, opened 2025-10-17
MAINT: Remove ``NDArrayOperatorsMixin.um`` class attribute ``umath`` reexport

jorenham:
Since #28035 the (internal) `numpy._core.umath` import was exposed as a *class attribute* `um` on `NDArrayOperatorsMixin`, and therefore also on all of its subclasses:

```python
In [1]: from numpy.lib.mixins import NDArrayOperatorsMixin

In [2]: class BadArray(NDArrayOperatorsMixin): ...

In [3]: BadArray.um
Out[3]: <module 'numpy._core.umath' from '/home/joren/Workspace/numpy/build-install/usr/lib/python3.14/site-packages/numpy/_core/umath.py'>
```

Judging from #28035 it doesn't look like this addition to the public API was intentional. I'm guessing that the intention was to reducing namespace pollution. 
But either way, it's probably best to get rid of this class class attribute before someone actually starts using it.

### DIFF
--- a/numpy/lib/mixins.py
+++ b/numpy/lib/mixins.py
@@ -1,6 +1,7 @@
 """
 Mixin classes for custom array types that don't inherit from ndarray.
 """
+from numpy._core import umath as um
 
 __all__ = ['NDArrayOperatorsMixin']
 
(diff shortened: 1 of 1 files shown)


seberg:
We discussed this, this seems clearly safe enough, weird thing but how would you use `um` usefully (and apparently it's not even be around forever, but added a year ago by accident).

Issue #28614, opened 2025-04-01
DOC: improve documentation on f2py and/or meson


wingkitlee0:
### Issue with current documentation:

Although `meson` will be required for python 3.12 or numpy 2.x, there are not much information on how to use meson or build a package with it.
Numpy (and f2py) are commonly used to build packages (unlike scipy that _uses_ meson). I think it will be beneficial to the community to have better documentation, esp examples, as meson itself is relatively new to the python build system.

### Idea or request for content:

I spent a weekend to get everything working in a Fortran wrapper https://github.com/wingkitlee0/pyigrf2. There are a lot of stuff requiring some effort to search in either meson doc or numpy doc. At the end, the most useful resource is scipy's meson.build files. There are a few tricks that deserved to be surfaced into the numpy documentation.

So I hope it will be useful to have examples on:
* f2py in a package
* numpy ufuncs extension (C API) in a package

Happy to contribute some docs.

ngoldbaum:
> Happy to contribute some docs.

Please!

ngoldbaum:
FWIW I agree the docs could definitely be better. If anyone wants to work on improving this I'm happy to help with code review and over low-latency chat.

wingkitlee0:
Thanks. Ya, I will find some time to think about what to add or change. Likely it's under the "Migrating to meson" page.

rgommers:
The numpy ufuncs part of this issue was completed in gh-30008.

mattip:
The fortran -> extension module via meson is now covered in the [docs](https://numpy.org/doc/stable/f2py/usage.html#building-with-meson-examples). I think that is sufficient to close the issue, since packaging that extention module into a package is not specific to f2py. Closing, please reopen or open a new issue if there is more needed.

Issue #20880, opened 2022-01-23
ENH: Future of npymath library


matthew-brett:
### Proposed new feature or change:

This is an issue to discuss the future of the `npymath` library in Numpy.

Numpy ships with a static library in `core/lib` named `npymath.lib` (Windows) or `libnpymath.a` (Unices).

It provides a suitable platform-specific implementation of various math routines.

Scipy links to `npymath` during its build process, in various places.

We (Scipy) have run into problems linking against Npymath, because of the combination of Numpy's use of the latest Visual Studio toolchain, and Scipy's use of the Mingw-w64 toolchain - discussed in https://github.com/MacPython/numpy-wheels/issues/145 .

The issue that arises is that it is in general more difficult to link static libraries across toolchains, than it is to link against dynamic libraries (e.g. DLLs) or to recompile the sources.   This issue is to discuss whether we should think of other solutions.   Possible options are:

* Leave things as they are.
* Ship the C header and sources for Npymath, so external libraries and packages can recompile them.
* Create a dynamic library from Npymath, and ship that, for linkage against external libraries.
* Your option here.

This issue is to house the discussion of different options.

rgommers:
Note that whether or not this is a priority is kinda coupled with the `vc141` vs. newer MS compiler toolchain discussion in https://github.com/MacPython/numpy-wheels/issues/145. For `vc141` it seems we can work with `libnpymath.a` in a Mingw-w64 based SciPy build, which with `vc142` things just seem broken and likely hard/fragile to repair.

We could do something proactively, but the urgency is a lot lower as long as there's no need to upgrade from `vc141`.

carlkl:
my suggestion is as follows:

- leave `npymath.lib` and `npyrandom.lib` as they are for numpy.
- Numpy could create two additional shared binaries  `npymath.dll` and `npyrandom.dll` from these static libraries.
- during the Numpy installation, place these two DLLs in a common location where Scipy or other third-party packages can locate them both during the build process and during import.


This approach has the following benefits:

- no need to change the build process of Numpy as it is only an extension of the existing build process.
- packages which depends on these libraries can still use the static libraries if **and only if** they use the same VS toolset as the Numpy uses. 
- packages which depends on these libraries using a different toolset or a different compiler can link against the shared libraries instead.

For a test I have successfully build `npymath.dll` with VS `link.exe`. It only needs a def file with all symbol names to be exported. A test to build scipy with this change will follow. 

There is a disadvantage as well:

[...]

PR #30008, opened 2025-10-17
DOC: Add concrete Meson build example for NumPy C ufunc extension

Aniketsy:
#28614 
This PR adds a new documentation page, ufunc_meson_example.rst

Thankyou !

### DIFF
--- a/doc/source/user/c-info.ufunc-tutorial.rst
+++ b/doc/source/user/c-info.ufunc-tutorial.rst
@@ -157,42 +157,54 @@ the module.
             return m;
         }
 
-To use the ``setup.py`` file, place ``setup.py`` and ``spammodule.c``
-in the same folder. Then ``python setup.py build`` will build the module to
-import, or ``python setup.py install`` will install the module to your
-site-packages directory.
+.. note::
 
-    .. code-block:: python
+   For modern Python packaging, it is recommended to use a ``pyproject.toml`` and select
+   either ``meson-python`` or ``scikit-build-core`` as your build backend. Both are well-designed,
+   well-maintained, and make your extension pip-installable and compatible with current Python packaging standards.
 
-        '''
-            setup.py file for spammodule.c
+   Example ``pyproject.toml``:
 
-            Calling
-            $python setup.py build_ext --inplace
-            will build the extension library in the current file.
+   .. code-block:: toml
 
-            Calling
-            $python setup.py build
-            will build a file that looks like ./build/lib*, where
-            lib* is a file that begins with lib. The library will
-            be in this file and end with a C library extension,
-            such as .so
+      [build-system]
+      requires = ["meson-python", "numpy>=2.0"]
+      build-backend = "mesonpy"
 
-            Calling
-            $python setup.py install
-            will install the module in your site-packages file.
+.. tab-set::
 
-            See the setuptools section 'Building Extension Modules'
-            at setuptools.pypa.io for more information.
-        '''
+   .. tab-item:: meson
 
-        from setuptools import setup, Extension
-        import numpy as np
+      .. code-block:: meson
 
-        module1 = Extension('spam', sources=['spammodule.c'])
+         project('spam', 'c')
 
-        setup(name='spam', version='1.0', ext_modules=[module1])
+         py = import('python').find_installation()
 
+         sources = files('spammodule.c')
+
+         extension_module = py.extension_module(
+           'spam',
+           sources,
+           install: true,
+           subdir: 'spam'
+         )
+
+      To build and install:
+
+      .. code-block:: bash
+
+         pip install .
+
+   .. tab-item:: setuptools
+
+      .. code-block:: python
+
+         from setuptools import setup, Extension
+
+         module1 = Extension('spam', sources=['spammodule.c'])
+
+         setup(name='spam', version='1.0', ext_modules=[module1])
 
 Once the spam module is imported into python, you can call logit
 via ``spam.logit``. Note that the function used above cannot be applied
@@ -339,49 +351,48 @@ the primary thing that must be changed to create your own ufunc.
             return m;
         }
 
-This is a ``setup.py`` file for the above code. As before, the module
-can be build via calling ``python setup.py build`` at the command prompt,
-or installed to site-packages via ``python setup.py install``. The module
-can also be placed into a local folder e.g. ``npufunc_directory`` below
-using ``python setup.py build_ext --inplace``.
+.. tab-set::
 
-    .. code-block:: python
+   .. tab-item:: meson
 
-        '''
-            setup.py file for single_type_logit.c
-            Note that since this is a numpy extension
-            we add an include_dirs=[get_include()] so that the
-            extension is built with numpy's C/C++ header files.
+      .. code-block:: meson
 
-            Calling
-            $python setup.py build_ext --inplace
-            will build the extension library in the npufunc_directory.
+         project('npufunc', 'c')
 
-            Calling
-            $python setup.py build
-            will build a file that looks like ./build/lib*, where
-            lib* is a file that begins with lib. The library will
-            be in this file and end with a C library extension,
-            such as .so
+         py = import('python').find_installation()
+         np_dep = dependency('numpy')
 
-            Calling
-            $python setup.py install
-            will install the module in your site-packages file.
+         sources = files('single_type_logit.c')
 
-            See the setuptools section 'Building Extension Modules'
-            at setuptools.pypa.io for more information.
-        '''
+         extension_module = py.extension_module(
+           'npufunc',
+           sources,
+           dependencies: [np_dep],
+           install: true,
+           subdir: 'npufunc'
+         )
 
-        from setuptools import setup, Extension
-        from numpy import get_include
+      To build and install:
 
-        npufunc = Extension('npufunc',
-                            sources=['single_type_logit.c'],
-                            include_dirs=[get_include()])
+      .. code-block:: bash
 
-        setup(name='npufunc', version='1.0', ext_modules=[npufunc])
+         pip install .
 
+   .. tab-item:: setuptools
 
+      .. code-block:: python
+
+         from setuptools import setup, Extension
+         from numpy import get_include
+         
+         # Note: since this is a numpy extension, we add include_dirs=[get_include()]
+         # so that the extension is built with numpy's C/C++ header files.
+         npufunc = Extension('npufunc',
+                             sources=['single_type_logit.c'],
+                             include_dirs=[get_include()])
+
+         setup(name='npufunc', version='1.0', ext_modules=[npufunc])
+    
 After the above has been installed, it can be imported and used as follows.
 
 >>> import numpy as np
@@ -403,7 +414,7 @@ Example NumPy ufunc with multiple dtypes
    pair: ufunc; adding new
 
 We finally give an example of a full ufunc, with inner loops for
-half-floats, floats, doubles, and long doubles. As in the previous
+floats, doubles, and long doubles. As in the previous
 sections we first give the ``.c`` file and then the corresponding
 ``setup.py`` file.
 
@@ -419,7 +430,6 @@ is the primary thing that must be changed to create your own ufunc.
         #include <Python.h>
         #include "numpy/ndarraytypes.h"
         #include "numpy/ufuncobject.h"
-        #include "numpy/halffloat.h"
         #include <math.h>
 
         /*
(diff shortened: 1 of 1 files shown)


Aniketsy:
@mhvk please review these changes when you get chance, and let me know if this needs improvement. I have created a new file and don't know if this is best approach .

doc/source/f2py/ufunc_meson_example.rst
```diff
@@ -0,0 +1,55 @@
+.. _ufunc-meson-example:
+
+===============================
```
mattip: Underline/Overline is too short

mattip (commented):
I think the added examples are missing. The other code examples are all in `f2py/code`, maybe put them under a `ufunc` subdirectory?

Aniketsy:
@mattip I've updated with changes.

mhvk (commented):
@Aniketsy - this looks fine but it is in the wrong place! I think this should be part of `https://numpy.org/doc/stable/user/c-info.ufunc-tutorial.html` -- and ideally where it introduces how to compile it using `setuptools`, you give an alternative with `meson`.  A big advantage is that then you can just adjust the sample for the ufunc c code that is already introduced there.

What would be brilliant is to use the same kind of switchable display as on https://numpy.org/doc/stable/building/index.html#building-from-source, where one has blocks different OS, and for different ways to create virtual environments, etc. Here, one could have one for `setuptools` and one for `meson`. Note, however, that I have not looked at the source code to see if it would be very tricky!

Aniketsy:
@mhvk i have updated, the changes as per suggestion, please let me know if it needs improvement.

ngoldbaum:
Awesome:

image

I didn't manually test that the meson example works but it looks correct at a quick glance.

My only suggestion is, IMO, we should be suggesting meson by default in the tab picker.

ngoldbaum:
Also probably worth suggesting writing a `pyproject.toml` and picking `meson-python` as the build backend to make your code pip-installable.

ngoldbaum:
The numpy-user-dtypes repo has a number of meson/meson-python examples that use the numpy C API you can take inspiration from: https://github.com/numpy/numpy-user-dtypes

Aniketsy:
@ngoldbaum Thanks! I've updated as per suggestions, please let me know if this need improvement.

doc/source/user/c-info.ufunc-tutorial.rst
```diff
@@ -193,6 +193,58 @@ site-packages directory.
 
         setup(name='spam', version='1.0', ext_modules=[module1])
 
+.. note::
+
+   For modern Python packaging, it is recommended to use a `pyproject.toml` and select
+   `meson-python` as your build backend. This makes your extension pip-installable and
```
rgommers: I'd add "or `scikit-build-core`" here. I am of course partial to `meson-python`, but I want to be fair and I usually explain that there is a choice of two modern and well-designed/maintained build backends. Both of those are a massive improvement over using `setuptools`.

doc/source/user/c-info.ufunc-tutorial.rst
```diff
@@ -193,6 +193,58 @@ site-packages directory.
...
+.. tab-set::
+
+   .. tab-item:: meson
+
+      .. code-block:: meson
+
+         project('spam', 'c')
+
+         py = import('python').find_installation()
+         numpy_include = run_command(py.full_path(), '-c', 'import numpy; print(numpy.get_include())').stdout().strip()
```
rgommers: This won't work reliably without more boilerplate. The much cleaner solution that works with numpy>=2.0 (which we can document now, since folks reading the docs today probably should not support 1.26 anymore:

```suggestion
np_dep = dependency('numpy')
```

doc/source/user/c-info.ufunc-tutorial.rst
```diff
@@ -193,6 +193,58 @@ site-packages directory.
...
+
+         sources = files('spammodule.c')
+
+         extension_module = py.extension_module(
+           'spam',
+           sources,
+           include_directories: [numpy_include],
```
rgommers: and then here:
```suggestion
           dependencies: [np_dep],
```

doc/source/user/c-info.ufunc-tutorial.rst
```diff
@@ -193,6 +193,58 @@ site-packages directory.
...
+           install: true
```
rgommers: Let's add `subdir: 'pkgname',` to this, which makes the extension module land in the right place.

doc/source/user/c-info.ufunc-tutorial.rst
```diff
@@ -193,6 +193,58 @@ site-packages directory.
...
+         )
+
+      To build and install:
+
+      .. code-block:: bash
+
+         meson setup builddir
+         meson compile -C builddir
+         meson install -C builddir
```
rgommers: This is rarely necessary. These three lines can be replaced with `pip install .`

doc/source/user/c-info.ufunc-tutorial.rst
rgommers: This comment applies to the other examples as well.

rgommers (changes requested):
Thanks for working on this @Aniketsy! I added some comments aiming to make the examples as idiomatic as possible.

Aniketsy:
@rgommers Thanks! I've updated with changes as per suggestion, please let me know if this need improvement.

doc/source/user/c-info.ufunc-tutorial.rst
rgommers: The `include_directories` line still has to be removed, here and in the other examples.

doc/source/user/c-info.ufunc-tutorial.rst
```diff
@@ -381,7 +433,61 @@ using ``python setup.py build_ext --inplace``.
 
         setup(name='npufunc', version='1.0', ext_modules=[npufunc])
 
+.. note::
+
+   For modern Python packaging, it is recommended to use a `pyproject.toml` and select
+   either `meson-python` or `scikit-build-core` as your build backend. Both are well-designed,
+   well-maintained, and make your extension pip-installable and compatible with current Python packaging standards.
```
rgommers: This note doesn't need to be repeated three times, can you just leave the first one and delete the other ones?

rgommers (changes requested):
This is getting closer. If you look at the rendered page (see [here](https://output.circle-artifacts.com/output/job/4f415ec0-3382-44e0-ba71-74166c6dcea3/artifacts/0/doc/build/html/user/c-info.ufunc-tutorial.html) for the latest version), the page doesn't flow very well though; it contains duplicate `setup.py` entries now above the `.. note::` blocks. Those can be deleted. The extra docstring content is mostly obsolete, except for the numpy-specific paragraphs like this one:
```
    setup.py file for single_type_logit.c
    Note that since this is a numpy extension
    we add an include_dirs=[get_include()] so that the
    extension is built with numpy's C/C++ header files.
```
Can you move those parts into the `setup.py` example in the tabs, and delete the rest (e.g., everything related to running `python setup.py xxx` is obsolete)?

There is only last issue with `libraries=["npymath"]`. I think we need to leave that as is but open a follow-up issue to change the example and remove those docs. We should stop encouraging anyone to use `libnpymath`, since we want to get rid of that. To do so, the example has to be changed to not use half-precision dtypes (it's a known issue that `float16` et al routines aren't available through the regular C API).

Aniketsy:
@rgommers I've update as per suggestion, please let me know if this needs further improvements. Thanks!

rgommers:
We're getting close! There's a rendering issue, see the double indent [here](https://output.circle-artifacts.com/output/job/815df588-bcf3-4b98-a5c5-26c0a04c6ce8/artifacts/0/doc/build/html/user/c-info.ufunc-tutorial.html#example-numpy-ufunc-with-multiple-dtypes):

image

doc/source/user/c-info.ufunc-tutorial.rst
```diff
@@ -157,42 +157,57 @@ the module.
...
+
+   Example `pyproject.toml`:
+
+   .. code-block:: toml
+
+      [build-system]
+      requires = ["meson-python", "numpy"]
```
rgommers: How about we use `numpy>=2.0` here for clarity? It won't work as is with 1.xx, and even if that's still supported at runtime, one still wants >=2.0 at build time.

doc/source/user/c-info.ufunc-tutorial.rst
```diff
@@ -339,49 +354,56 @@ the primary thing that must be changed to create your own ufunc.
...
+   .. code-block:: toml
 
-        '''
-            setup.py file for single_type_logit.c
-            Note that since this is a numpy extension
-            we add an include_dirs=[get_include()] so that the
-            extension is built with numpy's C/C++ header files.
+      [build-system]
+      requires = ["meson-python", "numpy"]
+      build-backend = "mesonpy"
```
rgommers: Can we also avoid repeating this 3 times? I think once is enough.

doc/source/user/c-info.ufunc-tutorial.rst
```diff
@@ -157,42 +157,57 @@ the module.
...
 
-To use the ``setup.py`` file, place ``setup.py`` and ``spammodule.c``
-in the same folder. Then ``python setup.py build`` will build the module to
-import, or ``python setup.py install`` will install the module to your
-site-packages directory.
+.. note::
 
-    .. code-block:: python
+   For modern Python packaging, it is recommended to use a `pyproject.toml` and select
+   either `meson-python` or `scikit-build-core` as your build backend. Both are well-designed,
```
rgommers: For the above 2 lines: single backticks are references in ReST, can you use double backticks instead?

rgommers (commented):
A few more small comments. Once those are resolved and the rendered version looks good, I'll do some local testing.

Aniketsy:
@rgommers  done with changes, please let me know if this needs further improvements. Thanks!

Aniketsy:
`There is only last issue with libraries=["npymath"]. I think we need to leave that as is but open a follow-up issue to change the example and remove those docs. We should stop encouraging anyone to use libnpymath, since we want to get rid of that. To do so, the example has to be changed to not use half-precision dtypes (it's a known issue that float16 et al routines aren't available through the regular C API).`

@rgommers I’d love to work on this issue if you allow. Please let me know if I can go ahead with this!

rgommers:
After 3 rounds of review there was still a lot of straightforward things to fix, like duplication of content and added examples that invalidated the existing text above it. So I pushed the needed changes for those directly.

The `meson.build` file for _Example NumPy ufunc with multiple dtypes_ isn't correct right now, since it does lot link `libnpymath`. I am thinking we need the following change there:

1. Remove `float16` support from the C example, to get the actual happy path multi-dtype support correct
2. Adjust the `setup.py` file to match
3. Then add a new example (or a diff) with `float16`, that then (a) does link `libnpymath` and (b) has a warning about that not being future-proof or super robust (since one sometimes needs to match the compilers used by `numpy` itself at wheel build time, link to https://github.com/numpy/numpy/issues/20880) but currently `NPY_HALF` & friends are not exposed in the regular NumPy API so this is your only choice currently. 
    - Add to that that `float16` != `bfloat16` and hence not super useful, please think twice about adding this support because you very likely do not need it.

@mhvk @ngoldbaum thoughts on that approach?

doc/source/user/c-info.ufunc-tutorial.rst
```diff
@@ -157,42 +157,54 @@ the module.
...
-        module1 = Extension('spam', sources=['spammodule.c'])
+   .. tab-item:: meson
 
-        setup(name='spam', version='1.0', ext_modules=[module1])
+      .. code-block:: meson
 
+         project('spam', 'c')
+
+         py = import('python').find_installation()
+         np_dep = dependency('numpy')
```
mhvk: This particular example actually doesn't use the numpy dependency, so this is not necessary.

doc/source/user/c-info.ufunc-tutorial.rst
```diff
@@ -157,42 +157,54 @@ the module.
...
+
+         extension_module = py.extension_module(
+           'spam',
+           ['spammodule.c'],
+           dependencies: [np_dep],
```
mhvk: This can go too.

doc/source/user/c-info.ufunc-tutorial.rst
```diff
@@ -157,42 +157,54 @@ the module.
...
+           install: true,
+           subdir: 'spam'
+         )
+
+   .. tab-item:: setuptools
+
+      .. code-block:: python
+
+         from setuptools import setup, Extension
+         import numpy as np
```
mhvk: And this one.

mhvk (commented):
Just had a quick look. Overall, looks very nice, just a few small comments on the existing changes.

On the later question of `float16`, I think for these examples it is also fine to not go on to an example that includes `float16`, but just skip to describing what that is tricky. In the end, these examples are meant to be for people writing their first ufunc, it does not have to be too elaborate.

rgommers:
> On the later question of `float16`, I think for these examples it is also fine to not go on to an example that includes `float16`, but just skip to describing what that is tricky. In the end, these examples are meant to be for people writing their first ufunc, it does not have to be too elaborate.

Agreed, that sounds like a good way to go. 

@Aniketsy if you're up for giving that a go, removing all `float16` related content from that example and ensuring it builds, then adding a sentence at the bottom as @mhvk suggests, please go for it.

Aniketsy:
I apologize for the late response, I was in the hospital.
@mhvk , @rgommers  thank you for the review. I’ve updated the PR with the suggested changes. Please let me know if this needs further improvements.

doc/source/user/c-info.ufunc-tutorial.rst
```diff
@@ -644,7 +627,11 @@ np.float64(0.0)
 >>> npufunc.logit(a)
 array([       -inf, -1.09861229,  0.        ,  1.09861229,         inf])
 
+.. note::
 
+   Supporting ``float16`` (half-precision) in custom ufuncs is more complex
+   due to its non-standard C representation and conversion requirements.
+   For most users, starting with ``float32`` and ``float64`` is recommended.
```
mhvk: How about replacing this with something a bit more positive: "If you really need it, have a look at the numpy source code to see how it is done there."

mhvk:
@Aniketsy - I had a look and tried everything. Some things still failed, in part because the `pyproject.toml` sample, as given, was not quite complete. So, I added a minimal complete one. While doing this, I saw other things to fix (including removing the last part of the file, the point of which completely escaped me), so I ended up thinking it was easier to just follow @rgommers's example and push my fixes directly to your branch. I also took the liberty to squash your commits together. Anyway, this ended up a larger refactoring than any of us thought, but I think the result is good! Thanks for starting it!!

@rgommers - I think this is now ready...

p.s. I am not sure the docs build can be viewed anywhere, but  I checked it locally, and it looks good.

rgommers:
> I am not sure the docs build can be viewed anywhere,

The link `ci/circleci: build artifact` in the list of CI jobs gets you to the preview of the html docs built in CI. You then still need to navigate to the actual page,

rgommers:
Thanks for another push on this @mhvk!

> I was in the hospital.

I hope you are feeling better @Aniketsy!

rgommers (approved):
This is in pretty good shape now, let's declare victory and get this in. Thanks @Aniketsy and @mhvk!

mhvk:
Great! And the result of a nice three-way collaboration, thanks both!

Aniketsy:
> I hope you are feeling better @Aniketsy!
@rgommers Thanks! I’m good for now :)

@mhvk  Thank you for the guidance and corrections.

PR #30010, opened 2025-10-17
TYP: add missing parameter default in ``reshape`` for ``shape``

jorenham:
Hmm, II thought I already took care of this one. But apparently not :shrug:

### DIFF
--- a/numpy/_core/fromnumeric.pyi
+++ b/numpy/_core/fromnumeric.pyi
@@ -225,7 +225,7 @@ def reshape(
 def reshape(  # shape: (int, ...) @ _AnyShapeT
     a: ArrayLike,
     /,
-    shape: _AnyShapeT,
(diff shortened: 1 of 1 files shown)


jakevdp:
I'm confused why `shape` needs to have a default value. It looks like it's a required argument at runtime: https://github.com/numpy/numpy/blob/989e9a6a83a0c1bf11dd5280086694324eac1f47/numpy/_core/fromnumeric.py#L298-L300
But wouldn't it be cleaner to communicate that via lack of a default rather than specifically checking for the default value and raising? Maybe the default value just part of the `new_shape` deprecation that we forgot to remove?

jorenham:
> Maybe the default value just part of the `new_shape` deprecation that we forgot to remove?

Ahh yea that makes sense. I'll just remove the default then.

PR #30011, opened 2025-10-17
TYP: ``linalg.svdvals``: fix inconsistent signature and add dtype overloads

jorenham:
This fixes a stubtest error (`x` should be positional-only) and improves the return type for 32- and 64-bits array-likes 
Oh yea and there weren't any type-tests for it yet, so I added some.


charris:
Thanks Joren.

Issue #27032, opened 2024-07-24
TYP: Overlapping overloads with incompatible return types


jorenham:
### Describe the issue:

The commonly used `@overload` pattern with `_ArrayLike{}_co` parameter types can lead to unexpected type inferences if the return types are incompatible with eachother.

An exampple of this is `numpy.dtype`, but there are many more at the moment (I'll compile a list of these soon).

### Reproduce the code example:

```python
from typing_extensions import assert_type
import numpy as np

assert_type(np.dtype(bool), np.dtype[np.bool])  # ok
assert_type(np.dtype(int), np.dtype[np.int_])  # ok
assert_type(np.dtype(bool | int), np.dtype[np.bool | np.int_])  # error
```

### Error message:

```shell
"assert_type" mismatch: expected "dtype[bool | signedinteger[Any]]" but received "dtype[bool] | dtype[signedinteger[Any]]"
```

### Python and NumPy Versions:

2.1.0.dev0+git20240724.3df4283
3.10.12 (main, Mar 22 2024, 16:50:05) [GCC 11.4.0]

### Type-checker version and settings:

basedpyright 1.15.0 (pyright 1.1.373)

### Additional typing packages.

typing-extensions 4.12.2

Jacob-Stevens-Haas:
Is the solution to use parametrized alias, not union aliases?  e.g.  
```
_ArrayInt_co: TypeAlias = NDArray[np.bool | integer[Any]]
```
would be replaced with:
```python
from typing import TypeVar
Int_co = TypeVar("Int_co", integer[Any], np.bool)
_ArrayInt_co: TypeAlias = NDArray[Int_co]
```

I think the dtype case (for overloads 2, 3, 4, 5) related but a different story though?  It looks like `z` matches [overload 2](https://github.com/numpy/numpy/blob/c118629dd5f08b23d090b4587ef1e899d33f5438/numpy/__init__.pyi#L776).  Letting `z` be a typevar seems to work, and could collapse some of the overloads:

```python
from typing import TypeVar
z = TypeVar("z", bound=type[complex])
def identity(typ: z) -> z: ...

reveal_type(np.dtype(identity(bool))) # OK: dtype[bool]
reveal_type(np.dtype(identity(int)))# OK: dtype[signedinteger[Any]]
reveal_type(np.dtype(identity(float)))  # OK: dtype[floating[_64Bit]]
[...]

PR #30013, opened 2025-10-17
TYP: ``linalg.tensordot``: fix inconsistent signature and simplify overloads

jorenham:
The duplication is a bit annoying, but is needed because in `_core.numeric` there the function parameters are all keyword or positional, while in `linalg._linalg` there are two pos-only parameters and one keyword-only.

The `| Any` in the `NDArray` scalar type-arguments might look redundant, but is needed to keep these overlapping overloads compatible. I've explained this in more detail in https://github.com/numpy/numpy/issues/27032#issuecomment-2278958251.


charris:
Needs rebase.

charris:
Thanks Joren.

PR #30014, opened 2025-10-17
TYP: stub ``linalg.lapack_lite.LapackError``

jorenham:
stubtest reported this as an error:

```
numpy.linalg.lapack_lite.LapackError is not present in stub
```


charris:
Thanks Joren.

PR #30016, opened 2025-10-17
DOC:Add note about debug build caveats in spin build --help

ChanchalSh2005:
Closes #29941

This PR adds a note in the `spin build --help` output explaining that debug builds (`-- -Dbuildtype=debug`) include additional checks and assertions, making them slower and meant for development or debugging use only.

### DIFF
--- a/.spin/cmds.py
+++ b/.spin/cmds.py
@@ -80,6 +80,15 @@ def changelog(token, revision_range):
 )
 @spin.util.extend_command(spin.cmds.meson.build)
 def build(*, parent_callback, with_scipy_openblas, **kwargs):
+    """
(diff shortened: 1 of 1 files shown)


mattip:
Isn;t that generally true of debug builds, and something anyone using debug would know? I don't think we need to state that here.

tylerjereddy:
Agreed, I think. There are other tradeoffs like sometimes-faster build times too of course, but seems more like general guidance on debugging compiled code than build tool-specific docs.

mattip:
Closing. The help presented by `spin` should show a little "what" and some "how", but not "why", that should be found elsewhere. The "how" request from #29941 is not solved by this PR.

PR #30018, opened 2025-10-17
TYP: fix stubtest errors in ``numpy.ma``

jorenham:
I wasn't able to figure out how to annotate the individual callable signatures of the `ma.core._frommethod`  "functions". Stubtest doesn't like the `@type_check_only` subclass trick, and there's currently no way to reliably pass an *overloaded* callable signature to a generic type, without it breaking along the way. I can't think of any other solutions. So I'm thinking of asking mypy to modify stubtest so that it'll consider `@type_check_only` subtypes as equivalent to its supertype at runtime (which makes sense if you think about it). 
But for the time being we're going to have to ignore these stubtest errors for `ma.count`, `ma.argmin`, and `ma.argmax`. That unfortunately means that we also won't know when there will be *actual* problems in those functions.

I also filled in a couple of missing annotations here and there, so this should also increases the "typing coverage".

There is actually still one error left in `ma.mrecords`, but  that's related to a deprecation from 1.22, so I'll follow-up on that by expiring it.


jorenham:
Hm. there's that scipy issue again in circleci

```
intersphinx inventory 'https://docs.scipy.org/doc/scipy/objects.inv' not fetchable due to <class 'requests.exceptions.ConnectTimeout'>: HTTPSConnectionPool(host='docs.scipy.org', port=443): Max retries exceeded with url: /doc/scipy/objects.inv (Caused by ConnectTimeoutError(<urllib3.connection.HTTPSConnection object at 0x7f1e66893e50>, 'Connection to docs.scipy.org timed out. (connect timeout=None)'))
```

For me the scipy docs have also been pretty slow lately 🤔

charris:
Thanks Joren.

PR #30019, opened 2025-10-18
MAINT, TST: Increase tolerance in fft test.

charris:
`test_identity_long_short_reversed` fails fairly often, so increase the test tolerance by a bit. Note that `random` is used throughout this test module without a seed, but should be OK as the tolerance is set by the type spacing at 1, which should be out of range for the generated random values.


mattip:
Thanks @charris

PR #30021, opened 2025-10-18
DEP: Remove ``delimitor`` kwarg from ``ma.mrecords.fromtextfile``

jorenham:
it has been deprecated since NumPy 1.22.0


mattip:
Thanks @jorenham

PR #28896, opened 2025-05-04
ENH: Modulate dispatched x86 CPU features

seiko2plus:
## Overview

This PR reorganizes NumPy's CPU build options by replacing individual x86 features with microarchitecture levels. This change aligns with the Google Highway project requirements and common Linux distribution practices.

This PR default setting for `cpu-baseline` on x86 has been raised to `x86-64-v2` microarchitecture as we're in 2025 and adding SIMD compatibility for antiquated  CPUs from before 2009 is no longer practical or efficient.
This can be changed to `cpu-baseline=none` during build time to support older CPUs, though manual SIMD optimizations for pre-2009 processors are no longer supported. This change improves performance and reduces binary 
size while only affecting hardware that is over 15 years old. 

## Key Changes

- **Consolidated into Microarchitecture Groups**: Replaced individual features with `X86_V2`, `x86_V3`, and `x86_V4` groups
- **Adjusted Baseline**: Set to micro-architecture level 2 (`x86_V2`), covering features from CPUs since 2009.
  This improves performance and reduces binary size
- **Improved `-` Operator**: Corrected to properly exclude successor features
- **Backward Compatibility**: Added redirection via meson for removed feature names
- **Stricter Compatibility**: Features like AVX512 without full mask operations now considered unsupported rather than using fallbacks

## Detailed CPU Feature Changes

- Removed individual features (`SSE`, `SSE2`, `SSE3`, `SSSE3`, `SSE4_1`, `SSE4_2`, `POPCNT`) → now in `X86_V2`
- Removed AMD legacy features (`XOP`, `FMA4`)
- Removed Xeon Phi support (`AVX512_KNL`, `AVX512_KNM`)
- Removed individual features (`AVX`, `AVX2`, `FMA3`, `F16C`) → now in `X86_V3`
- Removed `AVX512F`, `AVX512CD` (from dropping Xeon Phi support)
- Renamed `AVX512_SKX` to `X86_V4`
- Removed redundant groups `AVX512_CLX` and `AVX512_CNL`
- Updated `AVX512_ICL` to include `VAES`, `GFNI`, `VPCLMULQDQ`

## New Feature Group Hierarchy

| Name | Implies | Includes |
|------|---------|----------|
| X86_V2 | | SSE SSE2 SSE3 SSSE3 SSE4_1 SSE4_2 POPCNT CX16 LAHF |
| X86_V3 | X86_V2 | AVX AVX2 FMA3 BMI BMI2 LZCNT F16C MOVBE |
[...]

PR #30026, opened 2025-10-20
SIMD, BLD: Backport FPMATH mode on x86-32 and filter successor CPU features

seiko2plus:
Two major bugs need to be merged as part of gh-28896 and must be fixed:

- Force SSE-based floating-point on 32-bit x86 systems to avoid x87 FPU's 80-bit internal precision causing unpredictable rounding and overflow behavior when converting to smaller types.

- Fix filtering out successor features when an implied feature is removed through build options `--cpu-baseline` and `--cpu-dispatch`. Documentation `build-options` has been updated for this case.

- Updates `build-options` to add missing explanation about option `detect`.

Please, see [this comment](https://github.com/numpy/numpy/pull/28896) for more details.

### DIFF
--- a/doc/source/reference/simd/build-options.rst
+++ b/doc/source/reference/simd/build-options.rst
@@ -97,8 +97,23 @@ Having issues with AVX512 features?
 You may have some reservations about including of ``AVX512`` or
 any other CPU feature and you want to exclude from the dispatched features::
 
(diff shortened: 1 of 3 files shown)


seiko2plus:
It seems I need a rebase to get rid of linter errors, as PR #30029 merged.

charris:
Thanks Sayed.

PR #30031, opened 2025-10-20
TYP: fix stubtest errors in ``numpy.polynomial.*``

jorenham:
While working on this I noticed that there's a lot of room for improvement in the `numpy.polynomial` stubs. I tried my best to limit the changes to those that were needed to fix the stubtest errors, but in some cases I touched a some return annotations here and there. So  given that, and considering the side of this PR, I think that we probably shouldn't backport this. Splitting this up would be difficult, because most changes are affected by each other, i.e. via `ABCPolyBase` inheritance and the `_polytypes` callable types.

After this (and assuming that #30000 and #30007 get merged) there are only 5 (!) stubtest errors left. Two PR's should take care of those.


numpy/polynomial/_polybase.pyi
```diff
@@ -39,34 +37,36 @@ from ._polytypes import (
 
 __all__ = ["ABCPolyBase"]
 
-_NameCo = TypeVar(
-    "_NameCo",
-    bound=LiteralString | None,
-    covariant=True,
-    default=LiteralString | None
-)
```
jorenham, line 47: There's no real advantage to using `LiteralString` here, and needlessly complicates things, so that's why I got rid of it. This change should be backwards-compatible.

numpy/polynomial/_polybase.pyi
```diff
@@ -39,34 +37,36 @@ from ._polytypes import (
...
-_Other = TypeVar("_Other", bound=ABCPolyBase)
```
jorenham, line 48: `_Other` was renamed to `_PolyT`, following the `TypeVar` naming convention we use everywhere else.

numpy/polynomial/_polybase.pyi
```diff
@@ -39,34 +37,36 @@ from ._polytypes import (
...
-
+_NameT_co = TypeVar("_NameT_co", bound=str | None, default=str | None, covariant=True)
+_PolyT = TypeVar("_PolyT", bound=ABCPolyBase)
 _AnyOther: TypeAlias = ABCPolyBase | _CoefLike_co | _SeriesLikeCoef_co
-_Hundred: TypeAlias = Literal[100]
```
jorenham, line 51: This was only used once, so I inlined it

numpy/polynomial/_polybase.pyi
```diff
@@ -39,34 +37,36 @@ from ._polytypes import (
...
-    coef: _CoefSeries
-    domain: _Array2[np.inexact | np.object_]
-    window: _Array2[np.inexact | np.object_]
+    _symbol: str
+    @property
+    def symbol(self, /) -> str: ...
+
+    @property
+    @abc.abstractmethod
+    def domain(self) -> _Array2[np.float64 | Any]: ...
```
jorenham, line 59: You'll see some other changes like these, where I replaced a compicated abstract type annotation with e.g. `np.float64 | Any`. It's technically more type-unsafe (because of `Any`), but it's *a lot* less annoying to have to deal with from a user perspective. 
This is theoretically a backwards compatible change because of the [gradual (typing) guarantee](https://typing.python.org/en/latest/spec/concepts.html#the-gradual-guarantee):

> Removing type annotations (making the program more dynamic) should not result in additional static type errors. This is often referred to as the **gradual guarantee**.

numpy/polynomial/_polybase.pyi
```diff
@@ -78,39 +78,17 @@ class ABCPolyBase(Generic[_NameCo], abc.ABC):
...
     @overload
-    def __call__(self, /, arg: _Other) -> _Other: ...
-    # TODO: Once `_ShapeT@ndarray` is covariant and bounded (see #26081),
-    # additionally include 0-d arrays as input types with scalar return type.
+    def __call__(self, /, arg: _PolyT) -> _PolyT: ...
     @overload
-    def __call__(
-        self,
-        /,
-        arg: _FloatLike_co | decimal.Decimal | numbers.Real | np.object_,
```
jorenham, line 88: `numbers.Real` is problematic because the typeshed stubs for it are very vague (which is unavoidable because the `numbers` class hierarchy is, mildly put, awkward), and incomplete (i.e. most method parameter annotations are missing). The causes certain type-checkers (pyright notably) to label it, and everything that uses it, as "unknown". When e.g. pyright is configured in strict mode (which is pretty standard, and advisable), then every "unknown" type will be reported as an error. 
That means that everything that using `numbers.Real` as an annotation, like previously done here, will cause pyright to complain. 

Anyway, long story short; `numbers.Real` causes issues, so I got rid of it. It doubt it will cause backcompat issues for users though, as it was only meant to catch exotic number types that are incompatible with numpy's scalar types or the stdlib numeric types.

numpy/polynomial/_polybase.pyi
```diff
@@ -133,21 +111,18 @@ class ABCPolyBase(Generic[_NameCo], abc.ABC):
     def __rmod__(self, x: _AnyOther, /) -> Self: ...
     def __rdivmod__(self, x: _AnyOther, /) -> _Tuple2[Self]: ...
     def __len__(self, /) -> int: ...
-    def __iter__(self, /) -> Iterator[np.inexact | object]: ...
```
jorenham, line 136: `np.inexact | object` is equivalent to `object`, and is therefore only assignable to things of `object` type. That makes this maximally impractical to use (it can't get any worse than this), so I replaced it with a friendlier "gradual" `float64 | Any`.

numpy/polynomial/_polybase.pyi
```diff
@@ -133,21 +111,18 @@ class ABCPolyBase(Generic[_NameCo], abc.ABC):
...
     def __getstate__(self, /) -> dict[str, Any]: ...
     def __setstate__(self, dict: dict[str, Any], /) -> None: ...
 
     def has_samecoef(self, /, other: ABCPolyBase) -> bool: ...
     def has_samedomain(self, /, other: ABCPolyBase) -> bool: ...
     def has_samewindow(self, /, other: ABCPolyBase) -> bool: ...
-    @overload
-    def has_sametype(self, /, other: ABCPolyBase) -> TypeIs[Self]: ...
-    @overload
-    def has_sametype(self, /, other: object) -> Literal[False]: ...
```
jorenham, line 146: Overloading is redundant here, and can only confuse type-checkers

numpy/polynomial/_polytypes.pyi
```diff
@@ -113,22 +111,10 @@ _ArrayLikeCoef_co: TypeAlias = (
...
-    "_Name_co",
-    bound=LiteralString,
-    covariant=True,
-    default=LiteralString
-)
-
-@type_check_only
-class _Named(Protocol[_Name_co]):
-    @property
-    def __name__(self, /) -> _Name_co: ...
```
jorenham, line 126: This `__name__` property was causing stubtest (i.e. mypy) to report almost every module-level polynomial as incorrect. And if it was confusing stubtest, then it was probably also confusing mypy.

numpy/polynomial/_polytypes.pyi
```diff
@@ -113,22 +111,10 @@ _ArrayLikeCoef_co: TypeAlias = (
...
-
-_Line: TypeAlias = np.ndarray[tuple[Literal[1, 2]], np.dtype[_ScalarT]]
```
jorenham, line 128: A bit out of scope, but it turns out that using `Literal` in shape-types like this isn't a good idea (I won't bore you with the details).

numpy/polynomial/_polytypes.pyi
jorenham: Also, `Literal` strings tend to get upcast to `str` when used as generic type arguments (Pyright does this for performance reasons I believe).

numpy/polynomial/chebyshev.pyi
```diff
@@ -87,40 +95,39 @@ def _zseries_div(
 def _zseries_der(zs: npt.NDArray[_NumberOrObjectT]) -> _Series[_NumberOrObjectT]: ...
 def _zseries_int(zs: npt.NDArray[_NumberOrObjectT]) -> _Series[_NumberOrObjectT]: ...
 
-poly2cheb: _FuncPoly2Ortho[L["poly2cheb"]]
-cheb2poly: _FuncUnOp[L["cheb2poly"]]
+poly2cheb: Final[_FuncPoly2Ortho] = ...
```
jorenham, line 98: The `= ...` indicate the presence of a value at runtime, and the `Final[_]` markers prohibit reassignment.

jorenham (commented):
I left some comments for reviewers. Feel free to resolve them if it helps.

jorenham:
That scipy intersphinx issue is causing circleci to fail again

charris:
Thanks Joren.

PR #30032, opened 2025-10-20
TYP: `testing.check_support_sve`: fix inconsistent parameter default value

jorenham:
This fixes the following (duplicate) stubtest errors:

```
numpy.testing.check_support_sve is inconsistent, runtime parameter "__cache" has a default value of [False], which is different from stub parameter default []
numpy.testing._private.utils.check_support_sve is inconsistent, runtime parameter "__cache" has a default value of [False], which is different from stub parameter default []
```

So this one of the few cases where using `= ...` as default value stub is the way to go, since using `= [False]` here would be pretty confusing.

This isn't interesting enough to backport, as far as I'm concerned.


charris:
Thanks Joren.

PR #30033, opened 2025-10-20
TYP: fix stubtest error in ``numpy.typing``

jorenham:
Stubtest was reporting this error:

```
numpy.typing.PytestTester is not present at runtime
```

False positive or not, it suggests that mypy thinks that `numpy.typing.PytestTester` exists, and that it is public API.

This adds a minimal stub, which resolves this issue.


jorenham:
BTW, this was the last remaining stubtest error. So my plan now is to create a stubtest CI workflow, and perhaps a `spin stubtest` too.

charris:
Thanks Joren.

PR #30034, opened 2025-10-21
TYP: Add type annotations for ASIMD, NEON, and RVV targets

ixgbe:
Refer to https://github.com/numpy/numpy/pull/28517


jorenham (commented):
Hmm I'm not seeing these on my machine, so I take it that they're platform dependent? Because if so, then it might help to add a quick little comment about that there, similar to the comment for the `# NOTE` on line 4.

jorenham (approved):

jorenham:
Thanks @ixgbe

PR #30035, opened 2025-10-21
DEV: add a `spin stubtest` command

jorenham:
The 6 errors it currently reports (on Python 3.14) are addressed in  #29984, #30000, and #30007 are merged.


charris:
Thanks Joren.

PR #30036, opened 2025-10-21
TYP: restore abstract scalar type constructor parameters

### LINKED ISSUE #29977: TYP: np.dtype.type(value) no longer type-checking since numpy==2.3.4
### Describe the issue:

In `numpy==2.3.3`, the following code runs in Python and type-checks.

In `numpy==2.3.4`, the same code runs in Python but fails type-checks.

```python
from typing import Any

import numpy as np

def print_typed_one(my_type: np.dtype[np.integer[Any]]) -> None:
    print(my_type.type(1))

print_typed_one(np.dtype("=i1"))
```

### Error message:

```shell
bug.py:6: error: Too many arguments for "integer"  [call-arg]
bug.py:6: note: See https://mypy.rtfd.io/en/stable/_refs.html#code-call-arg for more info

and

bug.py
  bug.py:6:24 - error: Expected 0 positional arguments (reportCallIssue)
1 error, 0 warnings, 0 informations
```

Python 3.13.8
numpy 2.3.4

mypy 1.18.2
pyright 1.1.406

_No response_

jorenham:
This partially reverts #29629 and closes #29977


jorenham:
cc @bersbersbers

mattip:
Thanks @jorenham

PR #30039, opened 2025-10-21
DEV: Set correct ``PYTHONPATH`` in ``spin stubtest``

jorenham:
It turns out that I forgot to set the `PYTHONPATH` to the `build-install` dir in #30035. And because I had done a `uv pip install I didn't notice it was using `.venv`  instead.

The `meson._set_pythonpath` trick is also used in the `spin ipython`, so I assumed that this means it's ok to use then.


charris:
Thanks Joren.

PR #30040, opened 2025-10-21
DOC: Clarify signed vs unsigned ``intptr_t`` vs ``uintptr_t`` in basics.types

ParsaShemirani:
Clarified that uintptr_t is for pointers without sign as opposed to intptr_t.

Improves consistency across descriptions in regards to signed vs unsigned types.

Note: I am just a beginner with Numpy getting started with the docs so apologies if I misunderstood it.

### DIFF
--- a/doc/source/user/basics.types.rst
+++ b/doc/source/user/basics.types.rst
@@ -227,7 +227,7 @@ confusion with builtin python type names, such as `numpy.bool_`.
     * - N/A
       - ``'P'``
       - ``uintptr_t``
-      - Guaranteed to hold pointers. Character code only (Python and C).
+      - Guaranteed to hold pointers without sign. Character code only (Python and C).
 
     * - `numpy.int32` or `numpy.int64`
       - `numpy.long`


jorenham:
Seeing as the `u` prefix is commonly known to stand for "unsigned", I don't expect it to be very likely that someone in the target audience doesn't already know this. 
But I'll leave this open for a bit before closing in case another maintainer sees some benefit in this.

ParsaShemirani:
@jorenham Yes the u prefix is definitely well known. My suggestion was mainly for consistency across that section for the descriptions.

jorenham:
> My suggestion was mainly for consistency across that section for the descriptions.

Ah; consistency! I didn't know that **that's** what you are aiming to address here, seeing as it's not mentioned in the PR title or description. So, could you put that in there, so that the reviewers know what we should be looking at?

But anyway, I agree that this indeed improves consistency here :)

jorenham:
Thanks @ParsaShemirani

PR #30041, opened 2025-10-21
TYP: Various typing updates

charris:


jorenham:
I don't think we should backport [`600294e` (#30041)](https://github.com/numpy/numpy/pull/30041/commits/600294e0e938a280e362384630baf1603c3f637c)

numpy/_core/numerictypes.pyi
```diff
@@ -154,39 +154,40 @@ class _TypeCodes(TypedDict):
...
-    type[bytes_],
-    type[str_],
-    type[ubyte],
-    type[ushort],
-    type[uintc],
-    type[ulong],
-    type[ulonglong],
-    type[void],
-]
+typecodes: Final[_TypeCodes] = ...
```
jorenham, line 161:
```suggestion
def issubdtype(arg1: DTypeLike, arg2: DTypeLike) -> builtins.bool: ...

typecodes: Final[_TypeCodes] = ...
```

numpy/__init__.pyi
```diff
@@ -3509,8 +3511,11 @@ class ndarray(_ArrayOrScalarCommon, Generic[_ShapeT_co, _DTypeT_co]):
 # See https://github.com/numpy/numpy-stubs/pull/80 for more details.
 class generic(_ArrayOrScalarCommon, Generic[_ItemT_co]):
     @abstractmethod
-    def __new__(cls) -> Self: ...
-    def __hash__(self) -> int: ...
+    def __new__(cls, /, *args: Any, **kwargs: Any) -> Self: ...
+
+    if sys.version_info >= (3, 12):
+        def __buffer__(self, flags: int, /) -> memoryview: ...
+
```
jorenham:
```suggestion
    def __new__(cls, /, *args: Any, **kwargs: Any) -> Self: ...
    def __hash__(self) -> int: ...
```

numpy/__init__.pyi
```diff
@@ -405,6 +405,8 @@ from numpy._core.shape_base import (
 )
 
 from ._expired_attrs_2_0 import __expired_attributes__ as __expired_attributes__
+from ._globals import _CopyMode as _CopyMode
```
jorenham, line 408: On line 473 there already is a `from numpy._globals import _CopyMode`, so that one isn't needed anymore.
(the import order will require a `ruff lint --fix` though)

numpy/_core/numerictypes.pyi
jorenham, line 1: The `Final` import (from `typing`) might be missing here

charris:
I am afraid your pushes got lost (again).

jorenham:
The mypy failures in `numpy/typing/tests/data/reveal/numerictypes.pyi` can be resolved by including that entire `.pyi` here.

But the mypy failure in `numpy/typing/tests/data/fail/scalars.pyi` can be resolved by removing the `__new__` method from `class flexible` here: https://github.com/numpy/numpy/blob/49e4e89667730e86005aea81122f32eaf6577732/numpy/__init__.pyi#L5453-L5455

so that it's something like this:

```py
class flexible(_RealMixin, generic[_FlexibleItemT_co], Generic[_FlexibleItemT_co]): ...  # type: ignore[misc]
```

jorenham:
> I am afraid your pushes got lost (again).

I did not push anything though

jorenham:
The pyspark primer diff originates from [this line](https://github.com/apache/spark/blob/a8cfe0c682c57d2c7346fed20a0c2e76a2c7ff99/python/pyspark/pandas/series.py#L1210C27-L1210C30):

```py
tmp_val = arg[np._NoValue]  # type: ignore[attr-defined]
```

where arg is a `dict | pd.Series` instance.

So the previous `attr-defined` error now turned into a `call-overload` error; that's all.

numpy/__init__.pyi
```diff
@@ -1691,8 +1691,8 @@ class _ArrayOrScalarCommon:
         kind: _SortKind | None = ...,
         order: str | Sequence[str] | None = ...,
         *,
-        stable: bool | None = ...,
-    ) -> NDArray[Any]: ...
+        stable: builtins.bool | None = ...,
+    ) -> NDArray[intp]: ...
```
jorenham: This should fix the mypy failures I think

```suggestion
    ) -> NDArray[Any]: ...
```

jorenham:
> The mypy failures in `numpy/typing/tests/data/reveal/numerictypes.pyi` can be resolved by including that entire `.pyi` here.

With this the last two failures should be taken care of

jorenham (approved):
I went over it again, and I think it should be good to merge now.

jorenham:
Thanks @charris

PR #30042, opened 2025-10-21
BUG: np.fix rounds up when out == in

timkpaine:

### DIFF
--- a/numpy/lib/_ufunclike_impl.py
+++ b/numpy/lib/_ufunclike_impl.py
@@ -57,7 +57,7 @@ def fix(x, out=None):
 
     """
     # promote back to an array if flattened
-    res = nx.asanyarray(nx.ceil(x, out=out))
+    res = nx.asanyarray(nx.ceil(x, out=out, where=nx.less(x, 0)))
     res = nx.floor(x, out=res, where=nx.greater_equal(x, 0))
 
     # when no out argument is passed and no subclasses are involved, flatten
--- a/numpy/lib/tests/test_ufunclike.py
+++ b/numpy/lib/tests/test_ufunclike.py
@@ -47,6 +47,8 @@ class TestUfunclike:
         assert_equal(out, tgt)
         assert_equal(fix(3.14), 3)
 
+        assert_equal(fix(a), fix(a, out=a))
+
     def test_fix_with_subclass(self):
         class MyArray(np.ndarray):
             def __new__(cls, data, metadata=None):


numpy/lib/_ufunclike_impl.py
```diff
@@ -57,7 +57,7 @@ def fix(x, out=None):
 
     """
     # promote back to an array if flattened
-    res = nx.asanyarray(nx.ceil(x, out=out))
+    res = nx.asanyarray(nx.ceil(x, out=out if out is not x else None))
```
mattip: This will now allocate a new array, which is typically something to be avoided when using `out`. What about
```
res = nx.asanyarray(nx.ceil(x, out=out, where=np.less(x, 0))
```

numpy/lib/_ufunclike_impl.py
seberg, line 66: The ufunc is already doing this, this is duplicating it's logic! Just delte this whole if and comment.

numpy/lib/_ufunclike_impl.py
```diff
@@ -57,7 +57,7 @@ def fix(x, out=None):
 
     """
     # promote back to an array if flattened
-    res = nx.asanyarray(nx.ceil(x, out=out))
+    res = nx.asanyarray(nx.ceil(x, out=out, where=nx.less(x, 0)))
```
seberg, line 60:
```suggestion
    out = ... if out is None else out  # ensure `res` is not a scalar
    res = nx.ceil(x, out=out, where=nx.less(x, 0))
```
Although, I am not even sure that it even mattered in the old code really, but doesn't hurt now that we have it.

mattip:
@jorenham MyPy is unhappy, this seems like a false-positive?
```
res = nx.asanyarray(nx.ceil(x, out=out, where=nx.less(x, 0)))
E       TypeError: '<' not supported between instances of 'Object' and 'int'
```

jorenham:
> @jorenham MyPy is unhappy, this seems like a false-positive?
> 
> ```
> res = nx.asanyarray(nx.ceil(x, out=out, where=nx.less(x, 0)))
> E       TypeError: '<' not supported between instances of 'Object' and 'int'
> ```

This is a hard `TypeError`, so it's not a static typing thing. At first glance I can't really see where it's coming from though...

jorenham:
Ah I found the issue: https://github.com/numpy/numpy/blob/c318a2be91e387f17ec1ad773a81f5473974f925/numpy/typing/tests/data/pass/ufunclike.py#L36

This introduces a regression for user-defined duck-types such as `Object` in this test:

https://github.com/numpy/numpy/blob/c318a2be91e387f17ec1ad773a81f5473974f925/numpy/typing/tests/data/pass/ufunclike.py#L8-L22

seberg (commented):
With the test failure, Joren is right and that it needs fixing here.
But it is a problem with the test, the code change is OK the test is not.

Adding: `@functools.total_ordering` to that class should fix it or alternatively, implement `__lt__`.

jorenham:
> Adding: `@functools.total_ordering` to that class should fix it or alternatively, implement `__lt__`.

I suppose that it's kinda weird that the test was working before. But it indeed seems simple enough to fix like that

seberg:
Actually, I can't believe I missed that, but gh-30096 seems right.  We should just deprecate `np.fix` and tell users to use `trunc`.

We could possible just make it an alias, since trunc is implemented via `__trunc__` in python, it is probably already more reliable than the `ceil` used here likely.

seberg:
Closing, thereisa PR to deprecate. But thank you for opening and working on this!

PR #30043, opened 2025-10-21
CI, TYP: stubtest

jorenham:
This will run mypy's `stubtest` tool if there are any source code changes in `numpy/*`. Stubtest checks whether the `.pyi` stubs match their runtime equivalent implementation (e.g. `.py` or a C extension). For example, if a new function parameter is added to some function in a `.py` but not in the `.pyi`, then stubtest will complain about that.

TLDR; This'll effectively ensure that the stubs are always up-to-date. It'll also help prevent accidental public API changes, such as the one addressed in #30007.

### DIFF
--- a/.github/workflows/stubtest.yml
+++ b/.github/workflows/stubtest.yml
@@ -56,23 +56,9 @@ jobs:
           uv pip install
           -r requirements/build_requirements.txt
           -r requirements/test_requirements.txt
-          .
 
-      # workaround for https://github.com/python/mypy/issues/20064
-      # also addressed in https://github.com/numpy/numpy/pull/30044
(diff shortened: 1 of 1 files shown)


jorenham:
Ok nice; the stubtest errors are the same for both Python 3.11 and 3.14. Once #30000 and #30007 get merged, these errors should disappear, and then this should be ready to merge.

charris:
Thanks Joren. The test take about 3 minutes, so don't add much overhead.

jorenham:
> Thanks Joren. The test take about 3 minutes, so don't add much overhead.

Yea mypy has gotten a lot quicker recently. They're now even planning on rewriting mypy's parser in rust.

PR #30044, opened 2025-10-21
MAINT: bump ``hypothesis`` to ``6.142.2``

jorenham:
This is needed to avoid a crash in stubtest.

For context, see:
- https://github.com/HypothesisWorks/hypothesis/pull/4564
- https://github.com/python/mypy/issues/20064


charris:
Thanks Joren.

PR #30045, opened 2025-10-21
DEV: separate stubtest allowlist for py312+

jorenham:
This should take care of 3/6 stubtest errors (unused allowlist entries) on Python 3.11:
https://github.com/numpy/numpy/actions/runs/18691969608/job/53299696133?pr=30043


charris:
Thanks Joren. Note that we will drop Python 3.11 when 2.5 development starts after 2.4 is branched.

jorenham:
> Thanks Joren. Note that we will drop Python 3.11 when 2.5 development starts after 2.4 is branched.

Yea I can't wait 🎉

PR #30050, opened 2025-10-22
CI: Skip test runs if all changes are docs or stubs

jorenham:
This assumes that these runtime tests cannot be influenced by any `*.pyi`, `*.md`, and `*.rst` files.

### DIFF
--- a/.github/workflows/compiler_sanitizers.yml
+++ b/.github/workflows/compiler_sanitizers.yml
@@ -8,6 +8,10 @@ on:
     branches:
       - main
       - maintenance/**
+    paths-ignore:
+      - '**.pyi'
+      - '**.md'
+      - '**.rst'
 
 defaults:
   run:
--- a/.github/workflows/cygwin.yml
+++ b/.github/workflows/cygwin.yml
@@ -4,6 +4,10 @@ on:
     branches:
       - main
       - maintenance/**
+    paths-ignore:
+      - '**.pyi'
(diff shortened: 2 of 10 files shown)


ngoldbaum:
Here's the docs on this flag: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore

I like the idea of skipping unnecessary tests automatically. Can you clarify a little what the user experience for contributors will look like once we merge this?

jorenham:
> I like the idea of skipping unnecessary tests automatically. Can you clarify a little what the user experience for contributors will look like once we merge this?

Currently, PR's that only touch the typing stubs for example, will trigger 77 CI checks: https://github.com/numpy/numpy/pull/30033/checks
THere, most of the CI time is spent on testing code that has nothing to do with those changes, which clearly isn't ideal for development workflow, let alone the CI costs.
Also note that in this PR there are 76 checks, whereas in #30033 there are 77 checks. THe extra check is the mypy_primer workflow, which only runs on changes to `*.pyi` files.
So if this PR gets merged, future typing-only PR's will reduce the number of checks. I'm not sure how many that will be exactly, but it's probably somewhere below 40. That will significantly speed up the total CI runtime for typing PR's like that, as well as for docs-only PR's.

charris:
Let's give this a try. Thanks Joren.

PR #30052, opened 2025-10-22
TYP, STY: ``polynomial``: reformat the stubs

jorenham:
The code formatting in the `numpy.polynomial` stubs has been bothering me for a while now, so I ``ruff format``'ed them. Note that I used a `line-length` of `130`, as that's what the (official) Python typing spec says it should be in the [style guide](https://typing.python.org/en/latest/guides/writing_stubs.html#style-guide) for stubs.

I verified that this not introduce any (new) stubtest errors.


numpy/polynomial/_polybase.pyi
```diff
@@ -77,6 +74,7 @@ class ABCPolyBase(Generic[_NameT_co], abc.ABC):
         symbol: str = "x",
     ) -> None: ...
 
+    #
```
jorenham, line 77: These empty comments are needed to prevent `ruff format` from removing all blank lines between method declarations. So to improve readability, I tend to place `#` above `@overload`ed methods, and groups of methods that are similar in some way.

numpy/polynomial/chebyshev.pyi
```diff
@@ -82,16 +82,12 @@ __all__ = [
 ]
 
 _NumberOrObjectT = TypeVar("_NumberOrObjectT", bound=np.number | np.object_)
+_CoefScalarT = TypeVar("_CoefScalarT", bound=np.number | np.bool | np.object_)
```
jorenham, line 85: This was previously called `_RT`. Grouping the `TypeVar`s at the top of the file is prety much a convention in the numpy stubs (and also in scipy-stubs, for example).

charris:
Thanks Joren.

PR #30053, opened 2025-10-23
TYP: Type-checking the stubs

jorenham:
Previously, mypy was only used to type-check the "type-tests" living in `numpy/typing/tests/data/**`.

This updates the mypy config used by stubtest so that it will now also report (most) mypy errors it encounters in the stubs. Mypy was reporting several errors in the stubs. Some of them have been fixed here. But the errors for "incompatible overlapping overloads" (see #27032) have explicitly been disabled for the time being.

Now that mypy also type-checks the stubs, we'll be able to prevent a large class of potential bugs. Previously, even syntax errors could have gone by undetected if it wasn't covered by our explicit hand-writteen type-tests. Now, together with stubtest, the typing stubs are systematically checked for typing-errors, and compared with the runtime API to check for inconsistencies. Pretty cool, eh?

Mypy config docs: https://mypy.readthedocs.io/en/stable/config_file.html


jorenham:
We'll have to wait on #30043 for this to be enforced in CI. In the meantime, we can use the `spin stubtest` command for type-checking + stubtest the stubs.

charris:
Thanks Joren.

Issue #26404, opened 2024-05-08
BUG: Missing type annotations in `numpy.ma`


johnthagen:
### Describe the issue:

Running Mypy on some functions within `numpy.ma` lead to type checking errors due to missing annotations.

### Reproduce the code example:

```python
import numpy

a = numpy.ma.array([1, 2, 3, 4], mask=[0, 0, 1, 0])
b = numpy.ma.masked_array([1, 2, 3, 4], mask=[0, 0, 1, 0])
```

### Error message:

```shell
$ mypy test.py 
test.py:3: error: Call to untyped function "array" in typed context  [no-untyped-call]
test.py:4: error: Call to untyped function "MaskedArray" in typed context  [no-untyped-call]
```

### Python and NumPy Versions:

```
1.26.4
3.12.2 (main, Feb  6 2024, 20:19:44) [Clang 15.0.0 (clang-1500.1.0.2.5)]
```

### Runtime Environment:

```
[{'numpy_version': '1.26.4',
  'python': '3.12.2 (main, Feb  6 2024, 20:19:44) [Clang 15.0.0 '
            '(clang-1500.1.0.2.5)]',
  'uname': uname_result(system='Darwin', node='M2DF94775W-ML', release='23.4.0', version='Darwin Kernel Version 23.4.0: Fri Mar 15 00:12:37 PDT 2024; root:xnu-10063.101.17~1/RELEASE_ARM64_T6031', machine='arm64')},
 {'simd_extensions': {'baseline': ['NEON', 'NEON_FP16', 'NEON_VFPV4', 'ASIMD'],
                      'found': ['ASIMDHP'],
                      'not_found': ['ASIMDFHM']}},
 {'architecture': 'armv8',
  'filepath': '/Users/hagenjt1/Library/Caches/pypoetry/virtualenvs/non-package-mode-eSjUyYda-py3.12/lib/python3.12/site-packages/numpy/.dylibs/libopenblas64_.0.dylib',
  'internal_api': 'openblas',
  'num_threads': 16,
  'prefix': 'libopenblas',
  'threading_layer': 'pthreads',
  'user_api': 'blas',
  'version': '0.3.23.dev'}]
None
```

### Context for the issue:

We rely on type checking to find bugs in our software that uses Numpy before it is executed. This enables higher code quality and reliability. False positive type errors cost time to track down and can distract from real issues.

Type annotations also help IDEs such as PyCharm and VS Code perform better intellisense during development.

[...]

PR #30055, opened 2025-10-23
TYP: Annotate ``ma.array``, ``ma.asarray``, and ``ma.asanyarray``

jorenham:
towards #26404


jorenham:
The new `[assignment]` mypy error in xarray (https://github.com/pydata/xarray/blob/e49cfc4f2009eb660411128eaf05fa3e714a0c83/xarray/tests/test_backends.py#L2070) is a classic case of reassignment, so nothing to worry about.
The two `[unused-ignore]` are a good sign.

charris:
Thanks Joren.

PR #30057, opened 2025-10-23
DOC: Remove nonexistent ``order`` parameter docs of ``ma.asanyarray``

jorenham:
An alternative would be to actually add the `order` parameter (but then this would still be worth backporting).


charris:
I have mixed feelings about this. A backport would be good, but I would rather it be fixed in main.

jorenham:
> I have mixed feelings about this. A backport would be good, but I would rather it be fixed in main.

Let's discuss this Wednesday then

PR #30058, opened 2025-10-23
BUG: fix int left shift UB in CPU feature detection

ngoldbaum:
Fixes #29953

@seberg I *think* this is the right way to phrase this according to the C standard. I like this better than changing `reg` to be an unsigned int and adding a bunch of casts all over the place.


seberg (approved):
Yeah, looks good, thanks. If @seiko2plus or so prefers something different, I'll merge that too.

PR #30060, opened 2025-10-23
TYP: ``polynomial.polyutils``: fix callable type signatures

jorenham:
This only affects the private functions, so it's not the most interesting of fixes.


jorenham:
Well, at least we know now that stubtest is working

jorenham:
The stubtest failures are addressed in #30061

charris:
Might want to ignore distutils.

jorenham:
> Might want to ignore distutils.

see https://github.com/numpy/numpy/pull/30061 :)

jorenham:
I'll rebase, gimme a sec

charris:
Thanks Joren.

PR #30061, opened 2025-10-23
CI, TYP: Fix stubtest CI failures on py311

jorenham:
As can be seen in https://github.com/numpy/numpy/actions/runs/18756463021/job/53509587499?pr=30060, stubtest is failing on Python 3.11 because of missing `distutils` stubs. This adds a Python3.11 specific "allowlist" to ignore these errors.
We need a separate allowlist for this, because stubtest reports unused allowlist entries as errors. So we can't globally ignore the `numpy\.distutils\..*` pattern, as that would result in errors on all other Python versions.


charris:
Thank Joren.

PR #30064, opened 2025-10-23
TST: Remove unnecessary test__datasource thread-unsafe markers

bwhitt7:
In a previous PR (#29816) I marked up tests that were thread-unsafe when ran under [pytest-run-parallel](https://github.com/Quansight-Labs/pytest-run-parallel). A few markers I put there are no longer needed since I ended up fixing the issues with the tests in a separate PR (#29858). Forgot to actually get rid of the markers, so that's what this PR is for.


ngoldbaum:
Thanks @bwhitt7!

PR #30065, opened 2025-10-23
TYP: ``polynomial``: Simplify ``chebpts{1,2}`` function stubs

jorenham:
I'm not really a fan of those callable protocols in the type-check-only `_polytypes` module, and this is the only one that wasn't used in most submodules, so I got rid of it by "inlining" it.


jorenham:
stubtest will fail here too until rebased onto #30061

charris:
Thanks Joren.

Issue #16575, opened 2020-06-11
nanquantile is hundreds of times slower than quantile for certain cases


ashaffer:
nanquantile seems to be significantly slower than regular quantile, in certain cases. They seem to be identical in the case where you're processing a one dimensional vector, and also identical if you're processing a two dimensional vector as a whole. But if you apply it along one axis of a two dimensional vector, the performance is terrible. I guess it's possible that there's some intrinsic reason for this, but it's hard to imagine what that could be.

### Reproducing code example:

```python
import numpy as np

a = np.random.uniform(size=(27, 100))
%timeit np.quantile(a, 0.8, axis=0)
%timeit np.nanquantile(a, 0.8, axis=0)
```

For me the output of this is:

```
147 µs ± 11.4 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
10.3 ms ± 953 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

### Numpy/Python version information:

```
1.18.5 3.6.4 (default, Mar  9 2018, 23:15:12) 
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)]
```

seberg:
@ashaffer the reason is that these are functions written in Python.  They should be fairly efficient on 1-D slices, but will have quite significant overheads associated to each individual slice.

Your examples have many small 1-D slices on which the operation has to be performed, so those overheads will be large.  We do not have a good solution for this at the time.  For `np.quantile` speficially, it is itself implemented in Python to significantly reduce the overheads in this case, it would probably be required to move this to C.

There may be small performance improvements which are easier though (and e.g. general small-array performance improvements could have a large impact here).

carlosgmartin:
@seberg @ashaffer Someone wrote a workaround for this ([*np.nanpercentile() - there has to be a faster way!*](https://krstn.eu/np.nanpercentile()-there-has-to-be-a-faster-way/)), apparently resulting in a 160x speedup. It would be nice if this could be added to numpy.

max-sixty:
FYI numbagg now has a `nanquantile` function — it's slightly slower than numpy's on a 1D array, but 5x faster on 2D arrays where parallelization is possible

lbferreira:
I just released a library to speed up nanquantile calculation.
You can install using: 
```
pip install git+https://github.com/lbferreira/fastnanquantile
```
Example
```python

import numpy as np
import fastnanquantile as fnq

[...]

PR #30066, opened 2025-10-23
BUG: fix weighted-nanquantile broadcasting issue.

cakedev0:
#### Related issues/PRs:

Fixes: https://github.com/numpy/numpy/issues/29709
Also fixes partially (for the weighted case): https://github.com/numpy/numpy/issues/16575

Inspired by `sklearn.utils.stats._weighted_percentile`.

#### What does this implement/fix?

I refactored the logic of weighted `nanquantile`:
- **Before:** custom logic for weighted `nanquantile` (custom logic with a bug).
- **After:** zero the weights at NaNs values, and then run the logic from weighted quantile, which is what is done in scikit-learn and now scipy.stats.

I also extended the test surface for the weighted case.


numpy/lib/_nanfunctions_impl.py
```diff
@@ -1591,10 +1591,18 @@ def _nanquantile_unchecked(
     # so deal them upfront
     if a.size == 0:
         return np.nanmean(a, axis, out=out, keepdims=keepdims)
+    if weights is not None:
+        isnan = np.isnan(a)
+        if isnan.any():
+            # overwrite if `overwrite_input` is True?
+            weights = np.where(isnan, 0., weights)
```
cakedev0, line 1598: About this comment/line, the doc states:

> If True, then allow the input array `a` to be modified by intermediate calculations, to save memory. In this case, the contents of the input `a` after this function completes is undefined.

So I guess I should either:
- not modify `weights` (as I'm currently doing)
- modify `weights` in-place when `overwrite_input=True` and update the doc accordingly.

I think the second option is more aligned with the mindset of the `overwrite_input` parameter, but I'll wait for a reviewer's opinion before implementing it.

cakedev0:
cc @seberg if you want to take a look :pray:  it kinda is the follow-up from the one you've just merged

numpy/lib/_nanfunctions_impl.py
```diff
@@ -1618,54 +1625,35 @@ def _nanquantile_ureduce_func(
...
 
 def _nanquantile_1d(
-    arr1d, q, overwrite_input=False, method="linear", weights=None,
+    arr1d, q, overwrite_input=False, method="linear"
 ):
     """
     Private function for rank 1 arrays. Compute quantile ignoring NaNs.
     See nanpercentile for parameter usage
     """
-    # TODO: What to do when arr1d = [1, np.nan] and weights = [0, 1]?
```
cakedev0, line 1666: Before this PR: 
`RuntimeWarning: invalid value encountered in divide cdf /= cdf[-1, ...]  # normalization to 1` and returns 1

After: raises `ValueError: Weights included NaN, inf or were all zero.`

... I guess the expected behavior should be to return nan actually.

cakedev0: Hum, similar edge cases are interesting:

| x   | w   | current                | this PR   | desired   |
|:----|:----|:-----------------------|:-----|:----------|
| [1, 2] | [0, 0] | invalid divide warning | invalid weights error  | invalid weights error |
| [1, nan] | [0, 1] | invalid divide warning, returns 1 | invalid weights error | unsure? all nan slice warning, return nan? |
| [nan, nan] | [1, 1] | all-nan slice warning, return nan  |  invalid weights error | all-nan slice warning, return nan |

I need to at least fix the third case.

cakedev0: Back on this: the weights validation is not easy once intricated with NaNs. I pushed [a commit](https://github.com/numpy/numpy/pull/30066/commits/95975bf59edffa218be6647d2926f82653dd4e39) to match the desired column of the table above, but the code is bit convoluted, I'm not too happy about it.

The code could be quite simpler if we are ok with raising an invalid weights error for the case `x = [1, nan], w = [0, 1]`.

Wdyt @seberg?

seberg: Hmmmm, staring at it I am tempted to return NaN for the second case if it isn't terribly complicated (i.e. what you got as "desired").
People working with NaNs likely use it a bit like an NA value and I think propagating it makes sense (even the all-nan warning is maybe unnecessary, although it makes sense  here in that the normal path would give an error, I think).

This is a bit tedious, I admit. I can't think of a nicer way directly, turning around the logic isn't really helpful either, in the end "all NaN slice" here means "... wherever the weight wasn't zero".

The one thing I actually could see is to just return NaN+warning for the first case as well. But a problem with that is maybe that the path of "cannot have NaNs so just call the non-NaN version" doesn't work then.

cakedev0: > Hmmmm, staring at it I am tempted to return NaN for the second case if it isn't terribly complicated (i.e. what you got as "desired").

I agree. This preserve the equivalence between doing `np.nanquantile(x, q, weights=w, method="inverted_cdf")` and `np.nanquantile(x[w > 0], q, method="inverted_cdf")` if w is 1s and 0s.

> The one thing I actually could see is to just return NaN+warning for the first case as well

Hum... yes that makes sense! Using the same argument of equivalence than above and the fact that `np.nanquantile([], 0.5)` returns NaN.

And it's not so tedious to implement. I've just pushed a way to do it, it's a bit hacky, but not so hard to understand/maintain I think.

PR #26268, opened 2024-04-12
API: Disallow 0D input arrays in ``nonzero``

mtsokol:
Fixes #26238

This PR disallows 0D arrays input in `np.nonzero` and `np.ndarray.nonzero`.

PR #30068, opened 2025-10-24
MAINT: remove deprecated `style` argument and deprecations in `fromnumeric`

Abhi210:
This PR references gh-11521. It removes deprecated functionality and cleans up outdated code paths:

- Removed the deprecated `style` parameter from `array2string` in `numpy/_core/arrayprint.py`, along with its related documentation and tests.
- Replaced the deprecation warning for `np.sum` called on generators with a `TypeError` in `numpy/_core/fromnumeric.py`, since this behavior is now unsupported.
- Removed a leftover docstring that was intended to be deleted in PR #26268.


jorenham:
As you can see in the failing stubtest CI jobs, you'll also need to update the .pyi stubs to 
match these changes.

--- 

edit: actually I see now that it's a different issue, I'll have a closer look

numpy/_core/arrayprint.pyi
jorenham, line 1: Ah, stubtest is complaining that there are a couple of overloads that are now redundant:

```
numpy/_core/arrayprint.pyi:113: error: Overloaded function signature 2 will never be matched: signature 1's parameter type(s) are the same or broader  [overload-cannot-match]
numpy/_core/arrayprint.pyi:130: error: Overloaded function signature 3 will never be matched: signature 1's parameter type(s) are the same or broader  [overload-cannot-match]
numpy/_core/arrayprint.pyi:147: error: Overloaded function signature 4 will never be matched: signature 1's parameter type(s) are the same or broader  [overload-cannot-match]
numpy/_core/arrayprint.pyi:164: error: Overloaded function signature 5 will never be matched: signature 1's parameter type(s) are the same or broader  [overload-cannot-match]
```

If you remove all other overloads except for the first one, and also remove the remaining `@overload` decorator, then these errors should disappear.

numpy/_core/fromnumeric.py
```diff
@@ -2430,19 +2416,12 @@ def sum(a, axis=None, dtype=None, out=None, keepdims=np._NoValue,
     """
     if isinstance(a, _gentype):
         # 2018-02-25, 1.15.0
-        warnings.warn(
+        raise TypeError(
             "Calling np.sum(generator) is deprecated, and in the future will "
             "give a different result. Use np.sum(np.fromiter(generator)) or "
             "the python sum builtin instead.",
```
jorenham, line 2422: This message could do with an update

numpy/_core/arrayprint.pyi
Abhi210: I removed the extraneous overloaded signature functions. Furthermore, I removed the `@overload` since we now have only one call signature. However, the test is failing due to the following file:

```
AR: npt.NDArray[np.float64]
func1: Callable[[Any], str]
func2: Callable[[np.integer], str]

np.array2string(AR, legacy="1.14")  # type: ignore[call-overload]
np.array2string(AR, sign="*")  # type: ignore[call-overload]
np.array2string(AR, floatmode="default")  # type: ignore[call-overload]
np.array2string(AR, formatter={"A": func1})  # type: ignore[call-overload]
np.array2string(AR, formatter={"float": func2})  # type: ignore[call-overload]
```

It seems that these tests were supposed to fail. However, removing the overloaded call signatures now triggers a MYPY warning. Should we remove the `# type: ignore[call-overload]` in all the lines above?

jorenham: I'm guessing that the error codes changed from `call-overload` to `arg-type`, so replacing those here should fix that

numpy/_core/arrayprint.pyi
```diff
@@ -93,16 +90,13 @@ def set_printoptions(
 ) -> None: ...
 def get_printoptions() -> _FormatOptions: ...
 
-# public numpy export
```
jorenham, line 96: It's still a public numpy export

doc/release/upcoming_changes/30068.expired.rst
```diff
@@ -0,0 +1,8 @@
+Removal of deprecated functions and arguments
```
jorenham: Could you make the title more specific?

doc/release/upcoming_changes/30068.expired.rst
```diff
@@ -0,0 +1,8 @@
+``numpy.array2string`` and ``numpy.sum`` deprecations finalized
+---------------------------------------------
```
jorenham: This is causing CI to fail

```suggestion
``numpy.array2string`` and ``numpy.sum`` deprecations finalized
---------------------------------------------------------------
```

doc/release/upcoming_changes/30068.expired.rst
```diff
@@ -0,0 +1,8 @@
+``numpy.array2string`` and ``numpy.sum`` deprecations finalized
+---------------------------------------------------------------
+
+The following long-deprecated APIs have been removed or converted to errors:
+
+* The ``style`` parameter has been removed from ``numpy.array2string``. This argument had no effect since Numpy 1.14.0.
+
+* Calling ``np.sum(generator)`` directly on a generator object now raises a `TypeError`. This behavior was deprecated in NumPy 1.15.0. Use ``np.sum(np.fromiter(generator))`` or the python ``sum`` builtin instead.
```
mattip: Why doesn't the linter complain about this long line? Please reflow the note to 88 columns.

numpy/_core/tests/test_arrayprint.py
```diff
@@ -735,11 +735,6 @@ def test_0d_arrays(self):
         # str is unaffected
         assert_equal(str(x), "1")
 
-        # check `style` arg raises
-        pytest.warns(DeprecationWarning, np.array2string,
-                                         np.array(1.), style=repr)
-        # but not in legacy mode
-        np.array2string(np.array(1.), style=repr, legacy='1.13')
         # gh-10934 style was broken in legacy mode, check it works
```
mattip:
```suggestion
        # check it works
```

mattip (approved):
Looks good, a few formatting/comment nits.

doc/release/upcoming_changes/30068.expired.rst
Abhi210: I used to check the linting using the `ruff` tool. However, I found that it can only check the Python files. In `requirements/linter_requirements`, only `ruff`, `GitPython`, and `cython-lint` are present. And, I checked that none of these linters check the `.rst` files in the `doc/release/upcoming_changes` directory. So, `python tools/linter.py` passes successfully. Is there any linting tool to check those `.rst` files?

mattip:
Thanks @Abhi210

mhvk:
Hmm, this cleanup is causing problems in astropy because the order of positional arguments changed for `array2string`. Clearly, we can fix this, but would it be an idea to make everything after `prefix` keyword only? Almost certainly, there is not much code out there that uses positional up to >=6 arguments, but if there is, one gets weird errors.

jorenham:
> would it be an idea to make everything after `prefix` keyword only?

Yea seems like a good idea to me

mhvk:
OK, thanks! See #30139.

PR #30070, opened 2025-10-24
FIX: fix reduction issue in weighted quantile

cakedev0:
#### Related issues/PRs:

Fixes: https://github.com/numpy/numpy/issues/30069

#### How does this PR fixes the issue?

Handle `weights` in `_ureduce` the same way the main array `a` is handled.

Add tests that would have failed before the fix and now succeed (see for this [failed CI job](https://github.com/numpy/numpy/actions/runs/18786246946/job/53605130133) for instance)

### DIFF
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -3857,13 +3857,21 @@ def _ureduce(a, func, keepdims=False, **kwargs):
         if len(axis) == 1:
             kwargs['axis'] = axis[0]
         else:
-            keep = set(range(nd)) - set(axis)
+            keep = sorted(set(range(nd)) - set(axis))
             nkeep = len(keep)
-            # swap axis that should not be reduced to front
-            for i, s in enumerate(sorted(keep)):
-                a = a.swapaxes(i, s)
-            # merge reduced axis
-            a = a.reshape(a.shape[:nkeep] + (-1,))
+
+            def reshape_arr(a):
+                # move axis that should not be reduced to front
+                a = np.moveaxis(a, keep, range(nkeep))
+                # merge reduced axis
+                return a.reshape(a.shape[:nkeep] + (-1,))
+
+            a = reshape_arr(a)
+
+            weights = kwargs.get("weights")
+            if weights is not None:
+                kwargs["weights"] = reshape_arr(weights)
+
             kwargs['axis'] = -1
     elif keepdims and out is not None:
         index_out = (0, ) * nd
--- a/numpy/lib/tests/test_function_base.py
+++ b/numpy/lib/tests/test_function_base.py
@@ -4228,6 +4228,17 @@ class TestQuantile:
                 )
         assert_allclose(q, q_res)
 
+        # axis is a tuple of all axes
+        q = np.quantile(y, alpha, weights=w, method=method, axis=(0, 1, 2))
+        q_res = np.quantile(y, alpha, weights=w, method=method, axis=None)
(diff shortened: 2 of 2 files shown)


cakedev0:
ping @seberg as you've reviewed my preceding PR on weighted quantile.

numpy/lib/_function_base_impl.py
```diff
@@ -3840,11 +3840,21 @@ def _ureduce(a, func, keepdims=False, **kwargs):
...
-                a = a.swapaxes(i, s)
-            # merge reduced axis
-            a = a.reshape(a.shape[:nkeep] + (-1,))
+
+            def reshape_arr(a):
+                # swap axis that should not be reduced to front
+                for i, s in enumerate(sorted(keep)):
+                    a = a.swapaxes(i, s)
+                # merge reduced axis
+                return a.reshape(a.shape[:nkeep] + (-1,))
```
seberg, line 3867: Hehe, for once einops-like syntax would be nice for us also.

I do wonder if it isn't nicer to just duplicate things by using `moveaxis`.
```
a = np.moveaxis(a, keep, range(nkeep))
a = a.reshape(a.shape[:nkeep] + (-1,)) 
```
And then just repeat, because it's just two lines?  The `asanyarray(weights)` below should not be necessary here, I think.  We need additional preparation for the `weights` already (because they have funky logic for when the weights match the number of reduction axes)

seberg (approved):
Looks good, thanks!  I think a small iteration may be nice to make use of `moveaxis` rather than that loop here.
(The test is a bit large, but seems reasonable to just grow it here.)

numpy/lib/_function_base_impl.py
cakedev0: Thanks for the review and the nice suggestions:
- no need for `asanyarray(weights)` indeed
- using `np.moveaxis` is cleaner indeed
- though I think I prefer keeping the logic in a function. Not because I don't want to repeat 2 lines, but just because I think it's easier to read this way: you immediately see it's the same transformation applied to `a` and `weights`.

cakedev0:
This PR was approved, and I took into account some suggestions, maybe we can merged it?

cc @seberg

mattip:
Please avoid force pushing after approval, it complicates the review process.

cakedev0:
Sorry!

I've always worked in a squashed-on-merge environment, so I'm not familiar with the do-not-squash / rebase main kind of workflow.

For instance, I thought I was supposed to keep a clean commit history, so I merged the small changes I made after the review into the previous commit (as it was still about doing the same thing than in the previous commit). But yeah, now I do realize that it's unpractical for reviewers. What was I supposed to do? Commit those small changes with a message like:

> MNT: applied some review suggestions
> 
> - use `np.moveaxis` instead of iterating on `np.swapaxes`
> - remove useless `asanyarray`

I'm not sure such a commit has his place in the main history?

mattip:
We can squash-merge. Linear history is iverrated.

seberg:
Yeah, let's get this in, thanks again, sorry not merging it earlier. Should be backportable, I think.

seberg:
> For instance, I thought I was supposed to keep a clean commit history, so I merged the small changes

Yeah, we operated a bit like this for many years and some still like to do this. But overall, unless a PR is a bit larger and someone took a lot of care, I will squash-merge typically.

> Commit those small changes with a message like:

Honestly, any message, if the merger is lazy, the squash-merge will have a bullet with that, which should be OK.

Issue #28829, opened 2025-04-26
BUG: crashes found by fuzzing with fusil


devdanzin:
These are some crashes found by fuzzing Numpy with [fusil](https://github.com/devdanzin/fusil), condensed into a single issue as requested.

### 1- `strings._expandtabs_length.reduce(numpy.zeros(200))`

```python
import numpy._core._strings

numpy._core.strings._expandtabs_length.reduce(numpy.zeros(200))
```

Backtrace:
```
Thread 1 "python" received signal SIGSEGV, Segmentation fault.
0x00007fffb3907e4c in string_expandtabs_length_promoter(_object*, PyArray_DTypeMeta* const*, PyArray_DTypeMeta* const*, PyArray_DTypeMeta**) () from /home/danzin/venvs/3.13_upstream_venv/lib/python3.13t/site-packages/numpy/_core/_multiarray_umath.cpython-313t-x86_64-linux-gnu.so

#0  0x00007fffb3907e4c in string_expandtabs_length_promoter(_object*, PyArray_DTypeMeta* const*, PyArray_DTypeMeta* const*, PyArray_DTypeMeta**) ()
   from /home/danzin/venvs/3.13_upstream_venv/lib/python3.13t/site-packages/numpy/_core/_multiarray_umath.cpython-313t-x86_64-linux-gnu.so
#1  0x00007fffb38b2137 in call_promoter_and_recurse(_tagPyUFuncObject*, _object*, PyArray_DTypeMeta**, PyArray_DTypeMeta**, tagPyArrayObject* const*) ()
   from /home/danzin/venvs/3.13_upstream_venv/lib/python3.13t/site-packages/numpy/_core/_multiarray_umath.cpython-313t-x86_64-linux-gnu.so
#2  0x00007fffb38b1d0f in promote_and_get_info_and_ufuncimpl(_tagPyUFuncObject*, tagPyArrayObject* const*, PyArray_DTypeMeta**, PyArray_DTypeMeta**, unsigned char) ()
   from /home/danzin/venvs/3.13_upstream_venv/lib/python3.13t/site-packages/numpy/_core/_multiarray_umath.cpython-313t-x86_64-linux-gnu.so
#3  0x00007fffb38b26e1 in promote_and_get_ufuncimpl ()
   from /home/danzin/venvs/3.13_upstream_venv/lib/python3.13t/site-packages/numpy/_core/_multiarray_umath.cpython-313t-x86_64-linux-gnu.so
[...]

PR #30071, opened 2025-10-24
BUG: avoid negating INT_MIN in PyArray_Round implementation

ngoldbaum:
c.f. https://huntr.com/bounties/49928a2c-c6bb-4c1c-80ec-5d7bf708bf28 where this almost led to a CVE getting reported against NumPy.

Addresses one of the issues reported in https://github.com/numpy/numpy/issues/28829.

For those who are unaware: the value of `INT_MIN` is `-INT_MAX - 1` (negating using two's complement arithmetic), so that means that if C used two's complement `-INT_MIN == INT_MIN`. For that reason, `-INT_MIN` is UB according to the C standard. You are always supposed to do a check like this when negating a signed integer, but it is often skipped.

I learned while working on this that passing `round` a negative `ndigits` is supported. Round with `ndigits!=0` is the same as round with `ndigits==0`, but transformed in the following way: `round(x * 10**ndigits, 0) / 10**ndigits`. The same formula holds for positive and negative ndigits. 

No tests because it's annoying to write a test for this case (see https://github.com/python/cpython/issues/132474 -- it turns out `round(2**31, -2**31)` will hang CPython...).

Open to suggestions if people want to do a more thorough job of this but at least this prevents the segfault.

Ping @devdanzin, this seems relevant to your interests.

### DIFF
--- a/numpy/_core/src/multiarray/calculation.c
+++ b/numpy/_core/src/multiarray/calculation.c
@@ -652,7 +652,15 @@ PyArray_Round(PyArrayObject *a, int decimals, PyArrayObject *out)
     else {
         op1 = n_ops.true_divide;
         op2 = n_ops.multiply;
-        decimals = -decimals;
+        if (decimals == INT_MIN) {
+            // not technically correct but it doesn't matter because no one in
+            // this millenium is using floating point numbers with enough
+            // accuracy for this to matter
+            decimals = INT_MAX;
+        }
+        else {
+            decimals = -decimals;
+        }
     }
     if (!out) {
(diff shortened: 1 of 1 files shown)


devdanzin:
Thanks for the ping and for the reference! I guess having a report avoid a CVE isn't as sexy as having a report result in a CVE, but it's still cool :)

seberg (approved):
Thanks, could test the special case, but I am OK with just putting it in as is.  (The result seems 0 or `inf`, although I guess the inf is just overflowing intermediates.)

Please go ahead and merge if you think a special case test isn't worth it, happy with that..

Issue #30022, opened 2025-10-18
DOC: Confusing advanced indexing when mixing tuple, integer, and range


rocketcollider:
### Issue with current documentation:

When addressing a 3+ dim array with a tuple in its 3rd or later dims, the returned shape depends on the order of previous ranges/integers:

    >>> k = np.ones((4,5,6),)
    >>> k[:,0,(1,2)].shape
    (4, 2) #expected
    >>> k[0,:,(1,2)].shape
    (2, 5) #unexpected! actually transpose of right answer
    >>> k[0,:,:][:,(1,2)].shape
    (5, 2) #workaround
    >>> k[:1,:,(1,2)].shape
    (1, 5, 2) #workaround 

This *might* be an actual bug. It seems the actually unexpected case should be canonically correct: zero getting broadcast to (0,0). The documentation is unclear as to how this situation should behave, the example in that case uses unclear variable names:

    result[i_1, ..., i_M] == x[ind_1[i_1, ..., i_M], ind_2[i_1, ..., i_M],
                           ..., ind_N[i_1, ..., i_M]]

from here:
https://numpy.org/doc/stable/user/basics.indexing.html#integer-array-indexing

### Idea or request for content:

The priority between ranges and tuples seems to be undefined and I don't know what it *should* be. If the described behavior is expected, the documentation should describe behavior when mixing with ranges better.

jakevdp:
Hi – agreed that this can be confusing, but it is working as expected. This situation is covered by the docs in this section:  https://numpy.org/devdocs/user/basics.indexing.html#combining-advanced-and-basic-indexing

> Two cases of index combination need to be distinguished:
>
> 1. The advanced indices are separated by a slice, [Ellipsis](https://docs.python.org/3/library/constants.html#Ellipsis) or [newaxis](https://numpy.org/devdocs/reference/constants.html#numpy.newaxis). For example `x[arr1, :, arr2]`.
>
> 2. The advanced indices are all next to each other. For example `x[..., arr1, arr2, :]` but not `x[arr1, :, 1]` since 1 is an advanced index in this regard.
>
> In the first case, the dimensions resulting from the advanced indexing operation come first in the result array, and the subspace dimensions after that. In the second case, the dimensions from the advanced indexing operations are inserted into the result array at the same spot as they were in the initial array

It goes on to show some specific examples of this.

[...]

PR #30078, opened 2025-10-26
DOC: Clarify advanced vs basic indexing tuple definition

muhammed-soysl:
The existing definition of advanced indexing was slightly confusing about what triggers it versus basic indexing when using a tuple.

This commit adds a note to explicitly state that a tuple containing *only* slices, integers, ellipsis, or newaxis triggers basic indexing, while a tuple containing *at least one* list or ndarray triggers advanced indexing.

Relates to #30022.

### DIFF
--- a/doc/source/user/basics.indexing.rst
+++ b/doc/source/user/basics.indexing.rst
@@ -277,11 +277,14 @@ operations. For example::
 Advanced indexing
 -----------------
 
-Advanced indexing is triggered when the selection object, *obj*, is a
-non-tuple sequence object, an :class:`ndarray` (of data type integer or bool),
+Advanced indexing is triggered when the selection object *obj* is a non-tuple
+sequence object, an :class:`ndarray` (of data type integer or bool),
 or a tuple with at least one sequence object or ndarray (of data type
-integer or bool). There are two types of advanced indexing: integer
-and Boolean.
+integer or bool). In short, advanced indexing occurs when one of the indices
+is an array or sequence, and :ref:`basic-indexing` occurs for all
+other cases (e.g., integer scalars, :class:`slice` objects, or tuples of
+only these). There are two types of advanced indexing: integer and Boolean.
+
 
 Advanced indexing always returns a *copy* of the data (contrast with
 basic slicing that returns a :term:`view`).


doc/source/user/basics.indexing.rst
```diff
@@ -283,6 +283,17 @@ or a tuple with at least one sequence object or ndarray (of data type
 integer or bool). There are two types of advanced indexing: integer
 and Boolean.
 
+.. note::
+
+   The key difference from *basic* indexing lies in what the :class:`tuple`
+   contains. A tuple containing *only* integers, :class:`slice` objects,
```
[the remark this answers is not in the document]
jorenham: It's not a bad suggestion actually, and also I agree that it's a nitpick, so feel free to ignore it if you disagree.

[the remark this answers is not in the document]
muhammed-soysl: @jorenham Thanks for the review! That's a good suggestion for clarity. I've applied the change.

doc/source/user/basics.indexing.rst
```diff
@@ -283,6 +283,17 @@ or a tuple with at least one sequence object or ndarray (of data type
...
+.. note::
+
+   The key difference from *basic* indexing lies in what the :class:`tuple`
+   contains. A tuple containing *only* integer scalars, :class:`slice` objects,
+   :py:data:`Ellipsis`, or :const:`newaxis` objects triggers
+   :ref:`basic-indexing`.
+
+   In contrast, a tuple containing at least one :class:`list`,
+   :class:`ndarray`, or other sequence object triggers
+   :ref:`advanced-indexing`.
```
jakevdp: This note seems kind of redundant with the paragraph that precedes it – I wonder if you could add information to that paragraph that would make it more clear? I'm not sure what to suggest – what are the particular aspects that you think are unclear?

muhammed-soysl: Hi @jakevdp, thanks for taking the time to review this

My goal was to address the confusion (from Issue #30022) about what differentiates a "basic" tuple (which returns a view) from an "advanced" tuple (which returns a copy).

The original paragraph explained what *triggers* advanced indexing (a list/array in the tuple), but it didn't explicitly contrast it with what *triggers* basic indexing (a tuple with *only* integer scalars/slices). My note was an attempt to put those two rules side-by-side.

I agree that editing the original paragraph directly is a much better solution.

How about this: I can remove the `.. note::` block and instead modify the original paragraph to include this contrast? Something like:
"...This contrasts with *basic indexing* tuples, which contain *only* integer scalars, :class:`slice` objects, etc. An index tuple that contains *at least one* sequence object, :class:`ndarray`, or :class:`list` triggers :ref:`advanced-indexing`."

Let me know if an approach like that sounds better, and I'll update the PR.

jakevdp: I think this makes it far more complicated than it needs to be. In a sentence: advanced indexing is when one of your indices is an array or sequence, and basic indexing is everything else.

Arguably even the current paragraph is too long: I don't think we need two additional paragraphs to fix that.

muhammed-soysl: @jakevdp You are 100% right, that is a much simpler and clearer way to define it.

I was focused on the tuple confusion and over-complicated the explanation. I will remove my previous changes and update the paragraph using your "advanced is array/sequence, basic is everything else" logic.

Thank you for the guidance! I'll push the new update shortly.

jorenham:
I'm sorry but I don't consider this to be an improvement, and I don't see this becoming one either, so I'm going to close this.

PR #30083, opened 2025-10-27
DOC: Fix the first small 'process_core_dims()' example.

WarrenWeckesser:
Argh, even after https://github.com/numpy/numpy/pull/30074, I was oblivious to another goof in that code snippet 🤦‍♂️ .


mhvk (approved):
Ouch, but, yes, so easy to overlook...

PR #30067, opened 2025-10-23
TYP: ``numpy.ma``: Annotate 27 functions related to masks and fill-values

jorenham:
This adds type annotations for the following (currently untyped) public `numpy.ma` functions:

- `default_fill_value`
- `minimum_fill_value`
- `maximum_fill_value`
- `common_fill_value`
- `set_fill_value`
- `set_fill_value`
- `fix_invalid`
- `getdata`
- `getmaskarray`
- `make_mask`
- `make_mask_descr`
- `make_mask_none`
- `mask_or`
- `flatten_mask`
- `flatten_structured_array`
- `masked_invalid`
- `masked_where`
- `masked_greater`
- `masked_greater_equal`
- `masked_less`
- `masked_less_equal`
- `masked_not_equal`
- `masked_equal`
- `masked_inside`
- `masked_outside`
- `masked_object`
- `masked_values`

I didn't check how much this improves the typing coverage %, but this should help quite a bit.

Towards #26404

PR #30084, opened 2025-10-28
TYP: ``numpy.ma``: Annotate the callable wrapper classes

jorenham:
This is the next batch of missing `numpy.ma` annotations, following #30067.


charris:
Thanks Joren.

PR #30091, opened 2025-10-28
BUG, TYP: Fix ``ma.core._frommethod`` function signatures

jorenham:
This converts the `numpy.ma.core._frommethod` method wrapper utility class to a function, and modifies the wrapped `inspect.signature` to match that of the wrapper. 

Before:

```py
In [2]: np.ma.compress
Out[2]: <numpy.ma.core._frommethod at 0x7559fd926ea0>

In [3]: inspect.signature(np.ma.compress)
Out[3]: <Signature (a, *args, **params)>
```

After:

```py
In [2]: np.ma.compress
Out[2]: <function numpy.ma.core.compress(condition, a, axis=None, out=None)>

In [3]: inspect.signature(np.ma.compress)
Out[3]: <Signature (condition, a, axis=None, out=None)>
```

As you can see in the "before" output, the `a` parameter was incorrectly reported as being the first one, which isn't the case. That's why I've labeled this as a bug.

Before this, it wasn't possible to meaningfully annotate these `_frommethod` functions, as stubtest did not recognize them as functions, but rather as instances of the `_frommethod` type. But now that they're actual functions (i.e. instances of `FunctionType`), we can annotate them like any other function, allowing us to use overloads and such.

Note that there's also a similar situation for `np.ma.core._convert2ma`. But I figured I'd first start with this one in case there are any objections, and otherwise address `_convert2ma` in a follow-up.

---

~Note that stubtest will not like this, so I'll now be working on updating the stubs. I'm not planning on  touching `core.py` anymore, so that part can already be reviewed. (I'll have to do a quick scipy-stubs release in the meantime though.)~

### DIFF
--- a/doc/source/user/basics.copies.rst
+++ b/doc/source/user/basics.copies.rst
@@ -123,8 +123,8 @@ shape attribute of the array. For example::
     AttributeError: Incompatible shape for in-place modification. Use
     `.reshape()` to make a copy with the desired shape.
 
-Taking the example of another operation, :func:`.ravel` returns a contiguous
-flattened view of the array wherever possible. On the other hand,
+Taking the example of another operation, :func:`numpy.ravel` returns a
+contiguous flattened view of the array wherever possible. On the other hand,
 :meth:`.ndarray.flatten` always returns a flattened copy of the array.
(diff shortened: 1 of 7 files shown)


charris:
Let's give this a shot. Thanks Joren.

PR #30093, opened 2025-10-29
DOC: Correct grammatical usage like a/an

star1327p:
Correct grammatical usage like a/an

### DIFF
--- a/numpy/lib/_index_tricks_impl.py
+++ b/numpy/lib/_index_tricks_impl.py
@@ -473,8 +473,8 @@ class RClass(AxisConcatenator):
     expression can be used to change the output. The strings 'r' or 'c' result
     in matrix output. If the result is 1-D and 'r' is specified a 1 x N (row)
     matrix is produced. If the result is 1-D and 'c' is specified, then
-	an N x 1 (column) matrix is produced.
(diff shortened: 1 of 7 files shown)


numpy/lib/_index_tricks_impl.py
```diff
@@ -472,9 +472,9 @@ class RClass(AxisConcatenator):
     Optional character strings placed as the first element of the index
     expression can be used to change the output. The strings 'r' or 'c' result
     in matrix output. If the result is 1-D and 'r' is specified a 1 x N (row)
-    matrix is produced. If the result is 1-D and 'c' is specified, then a N x 1
-    (column) matrix is produced. If the result is 2-D then both provide the
-    same matrix result.
+    matrix is produced. If the result is 1-D and 'c' is specified, then
+	an N x 1 (column) matrix is produced.
+	If the result is 2-D then both provide the same matrix result.
```
jorenham: `\t` vs `    `
```suggestion
    an N x 1 (column) matrix is produced.
    If the result is 2-D then both provide the same matrix result.
```

this should fix the circle-ci failure:

```
docstring of numpy.r_:24: ERROR: Unexpected indentation.
```

jorenham:
thanks @star1327p

PR #30097, opened 2025-10-29
CI: Update ARM job (armhf_test) to use Ubuntu 24.04

### LINKED ISSUE #30086: CI: Update ARM job ( armhf_test )  to use Ubuntu 24.04 image
Following the discussion in gh-30075, the image based on Ubuntu 24.04 seems more appropriate for the `armhf_test` CI job.

Notes:
* The GCC version to compile Numpy in the Ubuntu 22.04 image for ARM seems GCC 11, which is a bit old.
* According to the information below, the Ubuntu 24.04 image for ARM is available:
  * https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-24-image.md

(@mattip  Thank you for the kind suggestion!)

riku-sakamoto:
Closes #30086 

* Update the image used for the armhf_test CI job to one based on Ubuntu 24.04. (`ubuntu-24.04-arm`)
* Use a virtual environment to comply with [PEP 668](https://peps.python.org/pep-0668/), which is automatically enforced on Ubuntu 24.04.

Notes:

One job running on the `ubuntu:24.04-arm` runner was abruptly canceled. Here is the log for reference: 
* https://github.com/numpy/numpy/actions/runs/18915094882/job/53996862849

Similar issues were reported about nine months ago.
* https://github.com/orgs/community/discussions/148648#discussioncomment-12080023

I have encountered this issue only once after making this change.
If it continues to occur, we may need to consider reverting to `ubuntu-22.04-arm`.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -166,7 +166,7 @@ jobs:
     # running on aarch64 (ARM 64-bit) GitHub runners.
     needs: [smoke_test]
     if: github.repository == 'numpy/numpy'
-    runs-on: ubuntu-22.04-arm
+    runs-on: ubuntu-24.04-arm
     steps:
     - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
       with:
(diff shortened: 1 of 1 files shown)


mattip:
I think the [`pip install` lines](https://github.com/numpy/numpy/actions/runs/18913428423/workflow?pr=30097#L183) should be moved down to the meson build step and that whole step be done after setting up a virtualenv.

riku-sakamoto:
@mattip 
Thank you for the comment!
As you suggested, I have updated CI settings and confirmed that it works successfully.

Also,  I have encountered a case where the job is abruptly cancelled, so I added a note about it in the PR description.

mattip:
> If it continues to occur, we may need to consider reverting to ubuntu-22.04-arm.

Are you sure the cancellation is due to the change in image? We can compare the frequency of cancellation to the remaining `ubuntu-22.04-arm` image in the linux-simd CI runs (which should also be moved to `ubuntu-24.04-arm` at some point).

mattip:
Thanks @riku-sakamoto

riku-sakamoto:
> Are you sure the cancellation is due to the change in image? We can compare the frequency of cancellation to the remaining ubuntu-22.04-arm image in the linux-simd CI runs (which should also be moved to ubuntu-24.04-arm at some point).

Thanks. As you pointed out, it seems early to tell whether the image update is responsible.
I’ll keep an eye on the CI runs for a while and gather more data.

PR #30098, opened 2025-10-30
DEP: Deprecate numpy.fix

hellerve:
This PR fixes #30096 (especially its follow-up comments) by deprecating `numpy.fix`.

Cheers


numpy/lib/_ufunclike_impl.py
```diff
@@ -56,6 +61,10 @@ def fix(x, out=None):
     array([ 2.,  2., -2., -2.])
 
     """
+    # NumPy 2.4.0
+    warnings.warn('The function numpy.fix is deprecated in favor of numpy.trunc since NumPy 2.4.0.',
+                  DeprecationWarning,
+                  stacklevel=2)
     # promote back to an array if flattened
```
ngoldbaum: There's a small chance it will break code somehow, but also maybe we can make this function an alias for `np.trunc` while we're at it. I'm reasonably sure that this will improve performance for all inputs - it was a factor of 2 speedup for a million-element array on my laptop. I don't think we need to do anything more fancy than deleting everything below and replacing it with `return np.trunc(x, out=out)`. This was originally @seberg's idea: https://github.com/numpy/numpy/issues/30096#issuecomment-3463314780

I realize that deprecating a thing normally means we don't touch it, but in this case what `fix` is doing is so silly I'm tempted to also do the performance optimization.

ngoldbaum (commented):
Thanks! Probably best to wait for someone else to weigh in before implementing my suggestion. Otherwise this all looks great.

hellerve:
It seems that my last error is due to this section: https://github.com/numpy/numpy/pull/30098/files#diff-6cdc3b490da76e0f287028f0c0fa780b24100106713a4216594f8d33d162fec6R36-R47

which is test data, but not picked up as tests, and a check complains about `warnings.filterwarnings` being used outside of tests (full message: `AssertionError: ignore filters should only be used in tests; found in C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\site-packages\numpy\typing\tests\data\pass\ufunclike.py on line 37`). Not sure how to resolve that best.

doc/release/upcoming_changes/30098.deprecation.rst
jorenham, line 1: the PR number in the filename is wrong

hellerve: Ah, I used the issue number. Sorry about that.

numpy/lib/_ufunclike_impl.py
```diff
@@ -56,15 +61,14 @@ def fix(x, out=None):
...
-    # promote back to an array if flattened
-    res = nx.ceil(x, out=... if out is None else out)
-    res = nx.floor(x, out=res, where=nx.greater_equal(x, 0))
-
-    # when no out argument is passed and no subclasses are involved, flatten
-    # scalars
-    if out is None and type(res) is nx.ndarray:
-        res = res[()]
-    return res
+    # NumPy 2.4.0
```
seberg: Let's put the full `# Deprecated NumPy 2.4, 2025-10-30` or so.

numpy/lib/tests/test_ufunclike.py
```diff
@@ -87,11 +97,15 @@ def test_scalar(self):
...
-        assert_equal(actual, expected)
-        assert_equal(type(actual), type(expected))
-
-        out = np.array(0.0)
-        actual = np.fix(x, out=out)
-        assert_(actual is out)
+        with warnings.catch_warnings():
+            warnings.filterwarnings(
+                "ignore", "The function numpy.fix is deprecated", DeprecationWarning
+            )
```
seberg, line 103: If we add a tiny test in `test_deprecations.py` we could use:
```
@pytest.mark.filterwarnings('ignore::DeprecationWarning')
```
and leave the test content unchanged (although this one should have it's own test function then).

seberg (approved):
Nice, thanks.  LGTM, two nits to change docs and maybe add to `test_deprecations.py` which would be nice (just to discover it, since this isn't C, I don't care too much about it).

numpy/typing/tests/data/pass/ufunclike.py
```diff
@@ -29,12 +33,18 @@ def __array__(self, dtype: np.typing.DTypeLike | None = None,
...
-np.fix(AR_LIKE_i)
-np.fix(AR_LIKE_f)
-np.fix(AR_LIKE_O)
-np.fix(AR_LIKE_f, out=AR_U)
+with warnings.catch_warnings():
+    warnings.filterwarnings(
+        "ignore",
+        message="The function numpy.fix is deprecated in favor of numpy.trunc",
+        category=DeprecationWarning,
+    )
```
ngoldbaum: @jorenham is there a proper way to call things that trigger deprecation warnings in the typing test data? this seems to run afoul of the test that checks the codebase for warning filters.

We could also patch that test to skip the typing test data. Maybe it doesn't matter if the typing tests trigger deprecation warnings? I don't actually know.

numpy/lib/_ufunclike_impl.py
```diff
@@ -56,15 +61,14 @@ def fix(x, out=None):
...
-    res = nx.floor(x, out=res, where=nx.greater_equal(x, 0))
-
-    # when no out argument is passed and no subclasses are involved, flatten
-    # scalars
-    if out is None and type(res) is nx.ndarray:
-        res = res[()]
-    return res
+    # NumPy 2.4.0, 2025-10-30
+    warnings.warn(
+        'The function numpy.fix is deprecated in favor of numpy.trunc '
```
ngoldbaum: Tiny grammar tweak. This style is a little bit closer to the other deprecation warnings we've written in the past.

You'll need to update the test for this message too.

```suggestion
        'Calling numpy.fix is deprecated. Use numpy.trunc instead.'
```

numpy/typing/tests/data/pass/ufunclike.py
seberg: Completely fine to change ignore it in those tests in principle, I think. The only reason for the filter in other tests was that it is nicer to have another solution (e.g. `pytest.warns` or `@pytest.filterwarnings`).

(The test comes from a different time and age and the original purpose is moot, now we just have nicer spellings :))

hellerve: I am unsure what this means. Do I get rid of these calls completely or should I look at how to change the test script? I can do the latter, but that seems like a separate PR to me, right?

jorenham: > @jorenham is there a proper way to call things that trigger deprecation warnings in the typing test data? this seems to run afoul of the test that checks the codebase for warning filters.
> 
> We could also patch that test to skip the typing test data. Maybe it doesn't matter if the typing tests trigger deprecation warnings? I don't actually know.

Yea there is the `typing_extensions.deprecated` decorator that can be used in the stubs. It can be applied to classes, functions, or individual overloads. See for example https://github.com/numpy/numpy/blob/d8eb2256ee4f21c90e8498c6008e08c960b3d174/numpy/_core/arrayprint.pyi#L96-L188

ngoldbaum: Ah, I see, that's what's missing. This PR also needs to update the type stubs for `np.fix`: https://github.com/numpy/numpy/blob/d8eb2256ee4f21c90e8498c6008e08c960b3d174/numpy/lib/_ufunclike_impl.pyi#L16-L35

numpy/lib/_ufunclike_impl.pyi
jorenham, line 1: You may want to `from typing_extensions import deprecated` as well

numpy/typing/tests/data/pass/ufunclike.py
hellerve: If I use that decorator locally (after importing of course 🤦 ), I still seem to be getting the errors on the deprecation message. So I suppose the question remains whether the checks should be dropped from `numpy/typing/tests/data/pass/ufunclike.py` and `numpy/typing/tests/data/reveal/ufunclike.pyi`?

hellerve:
So, it seems like the easiest way to deal with this deprecation is to

1. remove the examples from the doc string (those get checked and currently fail), and
2. remove the typing tests.

Both of these mean losing fidelity, but I’m not well-versed enough in the codebase to propose alternatives, unfortunately.

jorenham:
> 2\. remove the typing tests.

You could also smack a `# type: ignore[deprecated]` onto them, AFAIK.

hellerve:
> You could also smack a `# type: ignore[deprecated]` onto them, AFAIK.

I might be misunderstanding, but doesn’t that then defeat the prupose of the tests?

jorenham:
> > You could also smack a `# type: ignore[deprecated]` onto them, AFAIK.
> 
> I might be misunderstanding, but doesn’t that then defeat the prupose of the tests?

it would only suppress the `deprecated` errors; so any other mypy errors would still be reported

rgommers (dismissed):
This deprecation looks justified (despite us not removing this for 2.0 because of concerns of widespread usage), but a blocking comment for merge with a few comments:
-  This should be proposed on the NumPy mailing list before it can proceed, like all deprecations.
- This should not be merged this late in a release cycle, so please make it 2.5.0 not 2.4.0
- This cannot be merged before downstream users of our nightlies have been updated. E.g., I see usages in scikit-image and in PyWavelets. PRs to replace those usages would be nice to help this move along.

hellerve:
> > > You could also smack a `# type: ignore[deprecated]` onto them, AFAIK.
> > 
> > 
> > I might be misunderstanding, but doesn’t that then defeat the prupose of the tests?
> 
> it would only suppress the `deprecated` errors; so any other mypy errors would still be reported

Looks like that has the same issue, since now the assertion error is: `AssertionError: ignore filters should only be used in tests; found in /numpy/venv/lib/python3.11/site-packages/numpy/typing/tests/data/pass/ufunclike.py on line 37`. So neither using warning filters nor ignore filters seems to work here.

ngoldbaum:
> This should not be merged this late in a release cycle, so please make it 2.5.0 not 2.4.0

Fair enough - I cleared the 2.4.0 milestone.

@hellerve - if you're interested, it might be a fun exercise to contribute to a bunch of other popular projects as well if you could take a shot at doing the migration in a few popular packages. In addition to PyWavelets and scikit-image, which both looks like they use `fix` in a way that can be straightforwardly replaced by `trunc`, It looks like `pandas` has two uses that can be replaced by `trunc` and explicitly tests support for `fix` in one spot. That test will likely need to handle the new deprecation warning. 

It also looks like `astropy` has a few uses, but only to test or support wrapping of NumPy. That code might also need to handle the deprecation warning. Probably worth raising an issue over in astropy.

It looks like `jax` implements `np.fix` - maybe they care about this change. Pythran also has some tests that will need updating.

SciPy and matplotlib don't use it, as far as I can see.

ngoldbaum:
> Looks like that has the same issue, since now the assertion error is: AssertionError: ignore filters should only be used in tests; found in /numpy/venv/lib/python3.11/site-packages/numpy/typing/tests/data/pass/ufunclike.py on line 37. So neither using warning filters nor ignore filters seems to work here.

IMO it's fine to add an exception to the test that checks for warning filters to ignore code in the `typing/tests/data` directory. You'd have to add a new `if` block here:

https://github.com/numpy/numpy/blob/1e424dae42a2d560520b6e053e8e60ac4205bfc7/numpy/tests/test_warnings.py#L67-L74

hellerve:
> > This should not be merged this late in a release cycle, so please make it 2.5.0 not 2.4.0
> 
> Fair enough - I cleared the 2.4.0 milestone.
> 
> @hellerve - if you're interested, it might be a fun exercise to contribute to a bunch of other popular projects as well if you could take a shot at doing the migration in a few popular packages. In addition to PyWavelets and scikit-image, which both looks like they use `fix` in a way that can be straightforwardly replaced by `trunc`, It looks like `pandas` has two uses that can be replaced by `trunc` and explicitly tests support for `fix` in one spot. That test will likely need to handle the new deprecation warning.
> 
> It also looks like `astropy` has a few uses, but only to test or support wrapping of NumPy. That code might also need to handle the deprecation warning. Probably worth raising an issue over in astropy.
> 
> It looks like `jax` implements `np.fix` - maybe they care about this change. Pythran also has some tests that will need updating.
> 
> SciPy and matplotlib don't use it, as far as I can see.

I feel totally up to it (more than for the mailing list proposal, anyway 😅). I’ll take a look and start contributing PRs. Is there a script you use to check for usages or just go through a set of known repositories and check?

hellerve:
> > Looks like that has the same issue, since now the assertion error is: AssertionError: ignore filters should only be used in tests; found in /numpy/venv/lib/python3.11/site-packages/numpy/typing/tests/data/pass/ufunclike.py on line 37. So neither using warning filters nor ignore filters seems to work here.
> 
> IMO it's fine to add an exception to the test that checks for warning filters to ignore code in the `typing/tests/data` directory. You'd have to add a new `if` block here:
> 
> https://github.com/numpy/numpy/blob/1e424dae42a2d560520b6e053e8e60ac4205bfc7/numpy/tests/test_warnings.py#L67-L74

As part of this PR or a different one (since it’s more or less unconcerned with this particular deprecation)?

rgommers:
> Is there a script you use to check for usages or just go through a set of known repositories and check?

I just search known downstream projects first to get a sense, and then I may check what https://github.com/search?type=code says.

ngoldbaum:
> As part of this PR or a different one (since it’s more or less unconcerned with this particular deprecation)?

Doing it in this one is fine. I doubt it will come up again before this is merged.

ngoldbaum:
I just pinged the mailing list: https://mail.python.org/archives/list/numpy-discussion@python.org/thread/LYRHETXPN32S7FNHKNLJFDIYWGQHLODJ/

hellerve:
Alright, looks like we got the tests green, and I also adjusted the code comments to make the target `2.5.0`. I will go through the important downstream projects I can find in the coming days and submit PRs.

hellerve:
I did an (admittedly small) amount of digging for references. I unfortunately don’t have good insight into what repositories are considered "important", and Github search limits me to 5 largely unordered pages of results.

I did submit 4 PRs just now, among others to [pywt](https://github.com/PyWavelets/pywt/pull/830) and [scikit-image](https://github.com/scikit-image/scikit-image/pull/7933). I also checked a few others but couldn’t find much. (No guarantees, though)

ngoldbaum:
It looks like there are also a couple other uses in pandas that can be straightforwardly updated, if you want another easy contribution:

```
pandas/_libs/tslibs/timestamps.pyx
3495:                np.fix((153 * month - 457) / 5) +

pandas/core/arrays/datetimes.py
2290:            + np.fix((153 * month - 457) / 5)
```

There's another use in the tests that's testing explicitly for support of `np.fix` and probably shouldn't be updated.

hellerve:
> It looks like there are also a couple other uses in pandas that can be straightforwardly updated, if you want another easy contribution [...]

Contributed a PR there too.

seberg:
Thanks @hellerve, it's cool to see that you updated things there!

There was a small unintended regression in `np.fix` but I guess we'll wait for one release to actually add the deprecation.
Because of that, I think we should split out the change to `np.trunc` into a separate PR to be merged quickly.  I can just do that, but if you are interested in that, that would also be great!

hellerve:
I’ll solve the merge conflicts once #30168 has also been merged (I’m assuming it will be merged before this PR is even considered).

charris:
Needs rebase. #30168 is in.

hellerve:
@charris Thanks for the heads-up! Done!

jorenham:
this needs a rebase

hellerve:
Um... I tried for a second here, but then I realized #30644 landed in the meantime that fixes this?

PR #30100, opened 2025-10-30
BLD: use blobless checkout on CircleCI

mattip:
CirrusCI added an [option to use blobless checkout](https://circleci.com/changelog/introducing-a-faster-checkout-option). This will become the default on Nov 3. Let's proactively make sure we can use it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,8 @@ jobs:
   build:
     <<: *defaults
     steps:
-      - checkout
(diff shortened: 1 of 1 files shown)


mattip:
All good, the new default seems to work. We could merge this, or ignore it and `blobless` will become the default next week.

charris:
Thanks Matti. I merged to document the option for those who may not be familiar with the history.

PR #30101, opened 2025-10-30
TST: Add thread-safe testing guidelines

bwhitt7:
With the addition of [pytest-run-parallel](https://github.com/Quansight-Labs/pytest-run-parallel) to the NumPy CI, any new tests that get added will need to be thread-safe to ensure they pass. To help out, I tried adding some tips on writing thread-safe tests to the NumPy docs' testing guidelines. I also modified the sections that talk about setup/teardown methods and random data to reflect thread-safe practices.

### DIFF
--- a/doc/TESTS.rst
+++ b/doc/TESTS.rst
@@ -235,7 +235,7 @@ Similarly for methods::
           assert_(zzz() == 'Hello from zzz')
 
 Setup and teardown methods
----------------------------------------------
+--------------------------
 
 NumPy originally used xunit setup and teardown, a feature of `pytest`. We now encourage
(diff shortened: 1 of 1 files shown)


doc/TESTS.rst
```diff
@@ -225,36 +234,34 @@ Similarly for methods::
       def test_simple(self):
           assert_(zzz() == 'Hello from zzz')
 
-Easier setup and teardown functions / methods
+Setup and teardown methods
 ---------------------------------------------
```
charris: Be good to match the length of header here.

bwhitt7: Thank you for the catch!

charris:
Thanks @bwhitt7 .

PR #30102, opened 2025-10-30
ENH: Make FPE blas check a runtime check for all apple arm systems

### LINKED ISSUE #29820: BUG: "divide by zero encountered in matmul" on MacOS M4 with numpy v2.3.3
### Describe the issue:

We are seeing an error that appears identical to https://github.com/numpy/numpy/issues/28687 in our environment for the project https://github.com/brendanjmeade/celeri/issues/259#issuecomment-3339244337.

```python
import numpy as np
print(f"numpy version: {np.__version__}")
print(np.identity(n=15) @ np.identity(n=15)) # produces warning
print(np.identity(n=14) @ np.identity(n=14)) # no warning
```

### Error message:

```shell
numpy version: 2.3.3
/Users/meade/Desktop/celeri-org/celeri/celeri/scripts/celeri_solve.py:15: RuntimeWarning: divide by zero encountered in matmul
  print(f"{np.identity(n=15) @ np.identity(n=15)=}")
/Users/meade/Desktop/celeri-org/celeri/celeri/scripts/celeri_solve.py:15: RuntimeWarning: overflow encountered in matmul
  print(f"{np.identity(n=15) @ np.identity(n=15)=}")
/Users/meade/Desktop/celeri-org/celeri/celeri/scripts/celeri_solve.py:15: RuntimeWarning: invalid value encountered in matmul
  print(f"{np.identity(n=15) @ np.identity(n=15)=}")
np.identity(n=15) @ np.identity(n=15)=array([[1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
       [0., 1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
       [0., 0., 1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0., 1., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0., 0., 1., 0., 0., 0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0., 0., 0., 1., 0., 0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0., 0., 0., 0., 1., 0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 1., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1., 0., 0., 0.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1., 0., 0.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1., 0.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.]])
np.identity(n=14) @ np.identity(n=14)=array([[1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
       [0., 1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
       [0., 0., 1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 1., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 1., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0., 1., 0., 0., 0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0., 0., 1., 0., 0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0., 0., 0., 1., 0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0., 0., 0., 0., 1., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 1., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1., 0., 0., 0.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1., 0., 0.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1., 0.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 1.]])
```

Python 3.13.7
NumPy 2.3.3
MacOS: M4 with Sequoia 15.6.1 (24G90)

I will provide the exact command output on Monday.

### Runtime Environment:

Installed from conda-forge via `pixi install`. [lockfile](https://github.com/brendanjmeade/celeri/blob/bd013bf124cedd6eb4e5ecce4d87fbaee00ab95d/pixi.lock)

I will provide the exact command output on Monday

### Context for the issue:

This bug results in a lot of spurious, confusing, and scary warnings that distract from real problems.

seberg:
If an arm system uses SME some BLAS versions just set FPEs spuriously. The culprit is really Accelerate so we might limit this to Mac OS as well (and then with an SME check also -- see changes prior to this commit).

However, some OpenBLAS versions also caused this, although OpenBLAS is likely to clear the FPEs on their side.

Closes gh-29820

### DIFF
--- a/numpy/__init__.py
+++ b/numpy/__init__.py
@@ -870,6 +870,23 @@ else:
             del w
     del _mac_os_check
 
+    def blas_fpe_check():
+        # Check if BLAS adds spurious FPEs, mostly seen on M4 arms with Accelerate.
+        with errstate(all='raise'):
+            x = ones((20, 20))
+            try:
+                x @ x
+            except FloatingPointError:
+                res = _core._multiarray_umath._blas_supports_fpe(False)
+                if res:  # res was not modified (hardcoded to True for now)
+                    warnings.warn(
+                        "Spurious warnings given by blas but suppression not "
+                        "set up on this platform. Please open a NumPy issue.",
+                        UserWarning, stacklevel=2)
+
+    blas_fpe_check()
+    del blas_fpe_check
+
     def hugepage_setup():
         """
         We usually use madvise hugepages support, but on some old kernels it
--- a/numpy/_core/src/common/blas_utils.c
+++ b/numpy/_core/src/common/blas_utils.c
@@ -37,13 +37,13 @@ set_BLAS_causes_spurious_FPEs(void)
 {
     // These are all small, so just work on stack to not worry about error
     // handling.
-    double *x = PyMem_Malloc(20*20*3*sizeof(double));
+    double *x = PyMem_Calloc(20 * 20 * 3, sizeof(double));
     if (x == NULL) {
         PyErr_NoMemory();
         return -1;
     }
-    double *y = x + 20*20;
-    double *res = y + 20*20;
+    double *y = x + 20 * 20;
+    double *res = y + 20 * 20;
 
     npy_clear_floatstatus_barrier((char *)x);
 
--- a/numpy/_core/src/common/blas_utils.c
+++ b/numpy/_core/src/common/blas_utils.c
@@ -1,5 +1,3 @@
-#include <Python.h>
-
 #include "numpy/npy_math.h"     // npy_get_floatstatus_barrier
 #include "numpy/numpyconfig.h"  // NPY_VISIBILITY_HIDDEN
 #include "blas_utils.h"
@@ -17,46 +15,7 @@
 /*
  * Static variable to cache runtime check of BLAS FPE support.
  */
- static bool blas_supports_fpe = true;
-
-/*
- * ARM Scalable Matrix Extension (SME) raises all floating-point error flags
- * when it's used regardless of values or operations.  As a consequence,
- * when SME is used, all FPE state is lost and special handling is needed.
- *
- * For NumPy, SME is not currently used directly, but can be used via
- * BLAS / LAPACK libraries.  This function does a runtime check for whether
- * BLAS / LAPACK can use SME and special handling around FPE is required.
- *
- * This may be an Accelerate bug (at least OpenBLAS consider it that way)
- * but when we find an ARM system with SVE we do a runtime check for whether
- * FPEs are spuriously given.
- */
-static inline int
-set_BLAS_causes_spurious_FPEs(void)
-{
-    // These are all small, so just work on stack to not worry about error
-    // handling.
-    double *x = PyMem_Calloc(20 * 20 * 3, sizeof(double));
-    if (x == NULL) {
-        PyErr_NoMemory();
-        return -1;
-    }
-    double *y = x + 20 * 20;
-    double *res = y + 20 * 20;
-
-    npy_clear_floatstatus_barrier((char *)x);
-
-    CBLAS_FUNC(cblas_dgemm)(
-        CblasRowMajor, CblasNoTrans, CblasNoTrans, 20, 20, 20, 1.,
-        x, 20, y, 20, 0., res, 20);
-    PyMem_Free(x);
-
-    int fpe_status = npy_get_floatstatus_barrier((char *)x);
-    // Entries were all zero, so we shouldn't see any FPEs
-    blas_supports_fpe = fpe_status != 0;
-    return 0;
-}
+static bool blas_supports_fpe = true;
 
 #endif // NPY_BLAS_CHECK_FPE_SUPPORT
 
@@ -71,13 +30,14 @@ npy_blas_supports_fpe(void)
 #endif
 }
 
-NPY_VISIBILITY_HIDDEN int
-npy_blas_init(void)
+NPY_VISIBILITY_HIDDEN bool
+npy_set_blas_supports_fpe(bool value)
 {
 #if NPY_BLAS_CHECK_FPE_SUPPORT
-    return set_BLAS_causes_spurious_FPEs();
+    blas_supports_fpe = (bool)value;
+    return blas_supports_fpe;
 #endif
-    return 0;
+    return true;  // ignore input not set up on this platform
 }
 
 NPY_VISIBILITY_HIDDEN int
--- a/numpy/_core/src/common/blas_utils.c
+++ b/numpy/_core/src/common/blas_utils.c
@@ -1,6 +1,9 @@
+#include <Python.h>
+
 #include "numpy/npy_math.h"     // npy_get_floatstatus_barrier
 #include "numpy/numpyconfig.h"  // NPY_VISIBILITY_HIDDEN
 #include "blas_utils.h"
+#include "npy_cblas.h"
 
 #include <stdbool.h>
 #include <stdio.h>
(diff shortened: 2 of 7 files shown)


numpy/_core/src/common/blas_utils.h
```diff
@@ -2,18 +2,22 @@
...
-/* NPY_BLAS_CHECK_FPE_SUPPORT controls whether we need a runtime check
+/* 
+ * NPY_BLAS_CHECK_FPE_SUPPORT controls whether we need a runtime check
  * for floating-point error (FPE) support in BLAS.
+ * The known culprit right now is SVM likely only on mac, but that is not
+ * quite clear.
+ * This checks always on all ARM (it is a small check overall).
  */
-#if defined(__APPLE__) && defined(__aarch64__) && defined(ACCELERATE_NEW_LAPACK)
+#if defined(__aarch64__)
```
seberg: I wasn't sure what to put here, so went with the broad `__aarch64__`.  We could add `__APPLE__` if we think this is largely Accelerate only (i.e. OpenBLAS considered this a bug on their side, but for now I am not sure...).

It's not like it is utterly terrible if we have the runtime check for the global...

@mattip can you test it on an M4, my M1 hits the checks, but wouldn't actually trigger the spurious FPEs.

numpy/_core/src/common/blas_utils.c
```diff
@@ -11,92 +14,50 @@
...
-    }
+    npy_clear_floatstatus_barrier((char *)x);
 
-    if(has_SME){
-        return true;
-    }
-#endif
+    CBLAS_FUNC(cblas_dgemm)(
+        CblasRowMajor, CblasNoTrans, CblasNoTrans, 20, 20, 20, 1.,
+        x, 20, y, 20, 0., res, 20);
```
seberg: One of the jobs segfaulted (previous run), did I get these parameters wrong?

mattip:
Why do the check in C when we can do it in python in `__init__.py::_sanity_check`?

seberg:
No reason, I think I just ended there because it seemed like the smaller change to keep some of the logic in-place to skip it on unaffected machines (and a single call is simple enough, although something is off on windows!?).  Happy to just move it to Python fully as well.

mattip:
I think it would be better in python, no need to worry about calling conventions and allocations. You could even limit the check to `np._core._multiarray_umath.__cpu_features__['SVE'] is True`. We don't currently have a `SME` check.

seberg:
OK, let me just move it to Python.
> You could even limit the check to  `np._core._multiarray_umath.__cpu_features__['SVE'] is True`

I first had that, but then read [this](https://en.wikipedia.org/wiki/Apple_M4#Design):
> It supports the Scalable Matrix Extension (SME) but not the [Scalable Vector Extension](https://en.wikipedia.org/wiki/Scalable_Vector_Extension) (SVE). Because of the lack of SVE support, the [LLVM](https://en.wikipedia.org/wiki/LLVM) compiler officially flags the M4 as supporting ARMv8.7a.[[7]](https://en.wikipedia.org/wiki/Apple_M4#cite_note-7)

mattip:
Ahh, oops, right. Here is what the M4 reports:
```
>>> [k for k,v in np._core._multiarray_umath.__cpu_features__.items() if v is True]
['NEON', 'NEON_FP16', 'NEON_VFPV4', 'ASIMD', 'FPHP', 'ASIMDHP', 'ASIMDDP']
```

seberg:
> Ahh, oops, right. Here is what the M4 reports:

Which of these do you actually want to add? Right now it's hardcoded to just ARM on the C side, although because of that I am also not sure I really prefer moving this to Python.  Of course now it would be easier to add this, but I am not sure what (or if it even matters if we limit to ARM already).

charris:
The circleci failure is unrelated.

numpy/__init__.py
```diff
@@ -870,6 +870,23 @@ def _mac_os_check():
...
 
+    def blas_fpe_check():
+        # Check if BLAS adds spurious FPEs, mostly seen on M4 arms with Accelerate.
+        with errstate(all='raise'):
+            x = ones((20, 20))
+            try:
+                x @ x
+            except FloatingPointError:
+                res = _core._multiarray_umath._blas_supports_fpe(False)
+                if res:  # res was not modified (hardcoded to True for now)
```
mattip, line 881: `res` is not modified?

mattip (commented):
Could `_blas_supports_fpe()` (called with no argument) return the current state? Then
- I would prefer we not blanket-disable FPEs on ARM64, and use the import test to disable it if needed. 
- add the current state of `_blas_supports_fpe()` to `show_runtime()`.

numpy/__init__.py
seberg: > _blas_supports_fpe 

I made it `_blas_supports_fpe(None)` just to keep the C code a very bit simpler, didn't seem useful enough to care?

`res` is the state _after_ changing it.  If we are not on an ARM machine, it will still be True.

> add the current state of _blas_supports_fpe() to show_runtime().

Hmmm, yeah that makes sense.

seberg: Added the show runtime

> I would prefer we not blanket-disable FPEs on ARM64, and use the import test to disable it if needed.

Sorry, but what do you mean?  On non ARM64, it's blanket enabled.  On ARM64 it's disabled based on this test.  I have not bothered to skip this test on non-ARM64, instead it might give that warning when things seem off.

charris:
Is this ready? I need to make a decision about backporting it.

seberg:
I suspect so, unless you/Matti have thoughts on it. Unfortunately, I can't confirm for sure that it works :(.

charris:
I'll probably put this in, but wait for @mattip tomorrow. I do wonder why arm did things that way for sme, too hard?

mattip:
Did we have reports of aarch64 (linux) emitting FPEs? I thought the problem was only on macOS.

numpy/_core/src/common/blas_utils.h
```diff
@@ -2,27 +2,30 @@
...
-/* NPY_BLAS_CHECK_FPE_SUPPORT controls whether we need a runtime check
+/* 
+ * NPY_BLAS_CHECK_FPE_SUPPORT controls whether we need a runtime check
  * for floating-point error (FPE) support in BLAS.
+ * The known culprit right now is SVM likely only on mac, but that is not
+ * quite clear.
+ * This checks always on all ARM (it is a small check overall).
  */
-#if defined(__APPLE__) && defined(__aarch64__) && defined(ACCELERATE_NEW_LAPACK)
+#if defined(__aarch64__) && defined(HAVE_CBLAS)
```
mattip: Let's leave this restricted to macos, we have not seen reports of problems with linux.
```suggestion
#if defined(__APPLE__) && defined(__aarch64__) && defined(HAVE_CBLAS)
```

seberg:
> Did we have reports of aarch64 (linux) emitting FPEs? I thought the problem was only on macOS.

Not yet at least, I thought openblas may well be broken on others also (But I guess openblas may work around it).

mattip:
Thanks @seberg

Issue #30103, opened 2025-10-30
TYP: typing issue with `np.object_` dtype iteration


acabal33uab:
### Describe the issue:

`np.ndarray.__iter__` type does not correctly type the return value when `dtype` is a union containing `np.object_`

### Reproduce the code example:

```python
## The iterated value is correctly typed with unions.

import numpy as np
from typing import Any, reveal_type


def str_union(a: np.ndarray[tuple[Any], np.dtype[np.int8 | np.str_]]) -> str:
    for i in a:
        reveal_type(i)
        if isinstance(i, str):
            return i
    return "a"


kk.py:7: note: Revealed type is "numpy.signedinteger[numpy._typing._nbit_base._8Bit] | numpy.str_"



## The iterated value is correctly typed to `Any` with np.object_.

import numpy as np
from typing import Any, reveal_type

def object(a: np.ndarray[tuple[Any], np.dtype[np.object_]]) -> np.ndarray:
    for i in a:
        reveal_type(i)
        if isinstance(i, np.ndarray):
            return i
    return a


kk.py:15: note: Revealed type is "Any"


## The iterated value is incorrectly typed to `int8 | np.object_` with `np.object_` union.
It should be typed to `Any` or `int8 | Any`.

def object_union(a: np.ndarray[tuple[Any], np.dtype[np.int8 | np.object_]]) -> np.ndarray:
    for i in a:
        reveal_type(i)
        if isinstance(i, np.ndarray):
            return i
    return a


kk.py:23: note: Revealed type is "numpy.signedinteger[numpy._typing._nbit_base._8Bit] | numpy.object_"
```

### Error message:

```shell
kk.py:24: error: Subclass of "object_" and "ndarray[Any, Any]" cannot exist: "object_" is final  [unreachable]
            if isinstance(i, np.ndarray):
                          ^
kk.py:24: error: Subclass of "signedinteger[_8Bit]" and "ndarray[Any, Any]" cannot exist: would have incompatible method signatures 
[unreachable]
            if isinstance(i, np.ndarray):
                          ^
kk.py:25: error: Statement is unreachable  [unreachable]
                return i
                ^~~~~~~~
```

### Python and NumPy Versions:

2.3.4
3.14.0 (main, Oct 27 2025, 12:54:43) [GCC 14.2.0]

### Runtime Environment:

[{'numpy_version': '2.3.4',
  'python': '3.14.0 (main, Oct 27 2025, 12:54:43) [GCC 14.2.0]',
[...]

PR #30104, opened 2025-10-30
BUG, TYP: ufunc method signatures

### LINKED ISSUE #30095: BUG: Missing signatures for ufuncs
### Describe the issue:

Numpy ufuncs do not have valid signatures.

```python
import numpy as np
import inspect

inspect.signature(np.ufunc.accumulate)
inspect.signature(np.ufunc.at)
inspect.signature(np.ufunc.outer)
inspect.signature(np.ufunc.reduce)
inspect.signature(np.ufunc.reduceat)
inspect.signature(np.ufunc.resolve_dtypes)
```

### Error message:

```shell
Cell In[28], line 1
----> 1 inspect.signature(np.ufunc.accumulate)

File ~/miniconda3/lib/python3.10/inspect.py:3254, in signature(obj, follow_wrapped, globals, locals, eval_str)
   3252 def signature(obj, *, follow_wrapped=True, globals=None, locals=None, eval_str=False):
   3253     """Get a signature object for the passed callable."""
-> 3254     return Signature.from_callable(obj, follow_wrapped=follow_wrapped,
   3255                                    globals=globals, locals=locals, eval_str=eval_str)

File ~/miniconda3/lib/python3.10/inspect.py:3002, in Signature.from_callable(cls, obj, follow_wrapped, globals, locals, eval_str)
   2998 @classmethod
   2999 def from_callable(cls, obj, *,
   3000                   follow_wrapped=True, globals=None, locals=None, eval_str=False):
   3001     """Constructs Signature for the given callable object."""
-> 3002     return _signature_from_callable(obj, sigcls=cls,
   3003                                     follow_wrapper_chains=follow_wrapped,
   3004                                     globals=globals, locals=locals, eval_str=eval_str)

File ~/miniconda3/lib/python3.10/inspect.py:2468, in _signature_from_callable(obj, follow_wrapper_chains, skip_bound_arg, globals, locals, eval_str, sigcls)
   2463     return _signature_from_function(sigcls, obj,
   2464                                     skip_bound_arg=skip_bound_arg,
   2465                                     globals=globals, locals=locals, eval_str=eval_str)
   2467 if _signature_is_builtin(obj):
-> 2468     return _signature_from_builtin(sigcls, obj,
   2469                                    skip_bound_arg=skip_bound_arg)
   2471 if isinstance(obj, functools.partial):
   2472     wrapped_sig = _get_signature_of(obj.func)

File ~/miniconda3/lib/python3.10/inspect.py:2275, in _signature_from_builtin(cls, func, skip_bound_arg)
   2273 s = getattr(func, "__text_signature__", None)
   2274 if not s:
-> 2275     raise ValueError("no signature found for builtin {!r}".format(func))
   2277 return _signature_fromstr(cls, func, s, skip_bound_arg)

ValueError: no signature found for builtin <method 'accumulate' of 'numpy.ufunc' objects>
```

2.2.6
3.10.12 | packaged by conda-forge | (main, Jun 23 2023, 22:41:52) [Clang 15.0.7 ]

### Runtime Environment:

_No response_

### Context for the issue:

In Dask we fixed a bug where `dask.array.cumprod` wasn't passing the dtype down to the underlying function calls. To do this we inspect the funcs passed in for a dtype kwarg, and if set to use a `partial` to pass the dtype on.

When used with numpy ufuncs we are now seeing exceptions because the methods do not have valid signatures so we can't check if they have a dtype kwarg. See https://github.com/dask/dask/issues/12117

We could special case this in Dask as all ufuncs have a dtype kwarg, but it would be better if we could inspect them to discover this at runtime.

#23993 is related

jorenham:
closes #30095

Because stubtest is now able to pick up on the `ufunc` method signatures, this required updating the stubs accordingly. The ufunc stubs are pretty messy as you can see, so it took a bit of effort to get it all aligned.

### DIFF
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -4799,7 +4799,7 @@ add_newdoc('numpy._core.umath', '_add_newdoc_ufunc',
 
 add_newdoc('numpy._core.multiarray', 'get_handler_name',
     """
-    get_handler_name(a: ndarray) -> str,None
+    get_handler_name(a: ndarray) -> str | None
 
     Return the name of the memory handler used by `a`. If not provided, return
     the name of the memory handler that will be used to allocate data for the
@@ -5171,6 +5171,9 @@ add_newdoc('numpy._core', 'ufunc', ('signature',
 
 add_newdoc('numpy._core', 'ufunc', ('reduce',
     """
+    reduce($self, array, /, axis=0, dtype=None, out=None, **kwargs)
+    --
+
     reduce(array, axis=0, dtype=None, out=None, keepdims=False, initial=<no value>, where=True)
 
     Reduces `array`'s dimension by one, by applying ufunc along one axis.
@@ -5297,6 +5300,9 @@ add_newdoc('numpy._core', 'ufunc', ('reduce',
 
 add_newdoc('numpy._core', 'ufunc', ('accumulate',
     """
+    accumulate($self, array, /, axis=0, dtype=None, out=None)
+    --
+
     accumulate(array, axis=0, dtype=None, out=None)
 
     Accumulate the result of applying the operator to all elements.
@@ -5375,6 +5381,9 @@ add_newdoc('numpy._core', 'ufunc', ('accumulate',
 
 add_newdoc('numpy._core', 'ufunc', ('reduceat',
     """
+    reduceat($self, array, /, indices, axis=0, dtype=None, out=None)
+    --
+
     reduceat(array, indices, axis=0, dtype=None, out=None)
 
     Performs a (local) reduce with specified slices over a single axis.
@@ -5483,6 +5492,9 @@ add_newdoc('numpy._core', 'ufunc', ('reduceat',
 
 add_newdoc('numpy._core', 'ufunc', ('outer',
     r"""
+    outer($self, A, B, /, **kwargs)
+    --
+
     outer(A, B, /, **kwargs)
 
     Apply the ufunc `op` to all pairs (a, b) with a in `A` and b in `B`.
@@ -5554,6 +5566,9 @@ add_newdoc('numpy._core', 'ufunc', ('outer',
 
 add_newdoc('numpy._core', 'ufunc', ('at',
     """
+    at($self, a, indices, b=None, /)
+    --
+
     at(a, indices, b=None, /)
 
     Performs unbuffered in place operation on operand 'a' for elements
@@ -5605,6 +5620,9 @@ add_newdoc('numpy._core', 'ufunc', ('at',
 
 add_newdoc('numpy._core', 'ufunc', ('resolve_dtypes',
     """
+    resolve_dtypes($self, dtypes, *, signature=None, casting=None, reduction=False)
+    --
+
     resolve_dtypes(dtypes, *, signature=None, casting=None, reduction=False)
 
     Find the dtypes NumPy will use for the operation.  Both input and
@@ -5677,6 +5695,9 @@ add_newdoc('numpy._core', 'ufunc', ('resolve_dtypes',
 
 add_newdoc('numpy._core', 'ufunc', ('_resolve_dtypes_and_context',
     """
+    _resolve_dtypes_and_context($self, dtypes, *, signature=None, casting=None, reduction=False)
+    --
+
     _resolve_dtypes_and_context(dtypes, *, signature=None, casting=None, reduction=False)
 
     See `numpy.ufunc.resolve_dtypes` for parameter information.  This
@@ -5700,6 +5721,9 @@ add_newdoc('numpy._core', 'ufunc', ('_resolve_dtypes_and_context',
 
 add_newdoc('numpy._core', 'ufunc', ('_get_strided_loop',
     """
+    _get_strided_loop($self, call_info, /, *, fixed_strides=None)
+    --
+
     _get_strided_loop(call_info, /, *, fixed_strides=None)
 
     This function fills in the ``call_info`` capsule to include all
@@ -6453,7 +6477,7 @@ add_newdoc('numpy._core.multiarray', 'dtype', ('__le__',
 
 add_newdoc('numpy._core.multiarray', 'dtype', ('__gt__',
     """
-    __ge__(value, /)
+    __gt__(value, /)
 
     Return ``self > value``.
 
--- a/numpy/_core/function_base.py
+++ b/numpy/_core/function_base.py
@@ -1,4 +1,5 @@
 import functools
+import inspect
 import operator
 import types
 import warnings
@@ -477,6 +478,9 @@ def _add_docstring(obj, doc, warn_on_python):
             "Prefer to attach it directly to the source.",
             UserWarning,
             stacklevel=3)
+
+    doc = inspect.cleandoc(doc)
+
     try:
         add_docstring(obj, doc)
     except Exception:
@@ -494,10 +498,10 @@ def add_newdoc(place, obj, doc, warn_on_python=True):
     ----------
     place : str
         The absolute name of the module to import from
-    obj : str or None
+    obj : str | None
         The name of the object to add documentation to, typically a class or
         function name.
-    doc : {str, Tuple[str, str], List[Tuple[str, str]]}
+    doc : str | tuple[str, str] | list[tuple[str, str]]
         If a string, the documentation to apply to `obj`
 
         If a tuple, then the first element is interpreted as an attribute
@@ -534,12 +538,10 @@ def add_newdoc(place, obj, doc, warn_on_python=True):
     if isinstance(doc, str):
         if "${ARRAY_FUNCTION_LIKE}" in doc:
             doc = overrides.get_array_function_like_doc(new, doc)
-        _add_docstring(new, doc.strip(), warn_on_python)
+        _add_docstring(new, doc, warn_on_python)
     elif isinstance(doc, tuple):
         attr, docstring = doc
-        _add_docstring(getattr(new, attr), docstring.strip(), warn_on_python)
+        _add_docstring(getattr(new, attr), docstring, warn_on_python)
     elif isinstance(doc, list):
         for attr, docstring in doc:
-            _add_docstring(
-                getattr(new, attr), docstring.strip(), warn_on_python
-            )
+            _add_docstring(getattr(new, attr), docstring, warn_on_python)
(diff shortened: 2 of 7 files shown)


numpy/_core/_add_newdocs.py
```diff
@@ -4799,7 +4799,7 @@
 
 add_newdoc('numpy._core.multiarray', 'get_handler_name',
     """
-    get_handler_name(a: ndarray) -> str,None
+    get_handler_name(a: ndarray) -> str | None
```
jorenham, line 4802: sorry, I couldn't help myself

numpy/_core/_add_newdocs.py
```diff
@@ -5171,7 +5171,8 @@
 
 add_newdoc('numpy._core', 'ufunc', ('reduce',
     """
-    reduce(array, axis=0, dtype=None, out=None, keepdims=False, initial=<no value>, where=True)
+    reduce(array, axis=0, dtype=None, out=None, *, keepdims=False, where=True, **kwargs)
```
jorenham: I wasn't able to get the `initial=np._NoValue` default working. So it was either this, or lie about the default being `None` or `...` or something.

jorenham:
Let's see how much stubtest will complain

---

edit: ~127~ 112 errors; nice 🎉

numpy/_core/_add_newdocs.py
seberg: Hehe, but if we add the `--` to make it a signature, I guess this'll break anyway :P

jorenham:
Stubtest seems to be pretty confused about the `self` parameter now:

```
error: numpy.ufunc.resolve_dtypes is inconsistent, stub parameter "self" differs from runtime parameter "dtypes"
Stub: in file /home/joren/Workspace/numpy/numpy/__init__.pyi:5717
def (numpy.ufunc, dtypes: builtins.tuple[numpy.dtype[Any] | builtins.type | None, ...], *, signature: builtins.tuple[numpy.dtype[Any] | None, ...] | None =, casting: Literal['no'] | Literal['equiv'] | Literal['safe'] | Literal['same_kind'] | Literal['same_value'] | Literal['unsafe'] | None =, reduction: builtins.bool =) -> builtins.tuple[numpy.dtype[Any], ...]
Runtime:
def (dtypes, *, signature=None, casting=None, reduction=False)
```

So I think that we might need to add a `$self`, after all 🤔

numpy/_core/_add_newdocs.py
seberg: Hmmmmpf, that is annoying since it'll actually print with `**kwargs` in the docs as well now for (practically) no reason.

I suppose duplicating the first row will probably also just duplicate it on the sphinx output?  Because if it doesn't, then I think I would just opt for that.

numpy/_core/_add_newdocs.py
```diff
@@ -6453,7 +6462,7 @@
 
 add_newdoc('numpy._core.multiarray', 'dtype', ('__gt__',
     """
-    __ge__(value, /)
+    __gt__(value, /)
```
seberg, line 6480: Haha, nice one!

numpy/_core/function_base.py
```diff
@@ -477,6 +478,9 @@ def _add_docstring(obj, doc, warn_on_python):
             "Prefer to attach it directly to the source.",
             UserWarning,
             stacklevel=3)
+
+    doc = inspect.cleandoc(doc)
```
seberg, line 482: Ah, I guess that is a bit nicer. I do have a slight feeling we may have avoided `inspect` due to not super-fast import at some point, but maybe that was just calling it a lot?

seberg (commented):
Thanks for looking into it!  A bit too bad about the `__text_signature__` limitations, makes me wonder if Python has hacks around it, but I guess probably not...

numpy/_core/function_base.py
jorenham:
```
In [60]: %timeit inspect.cleandoc(np.ufunc.reduce.__doc__)
16.8 μs ± 12.6 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [61]: import textwrap

In [62]: %timeit textwrap.dedent(np.ufunc.reduce.__doc__)
10.8 μs ± 5.41 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

🤷🏻

numpy/_core/_add_newdocs.py
jorenham: I'm not sure if that'll work though, but I'll try

jorenham: Yea... doesn't work.

So, should we just lie about it and say it's `None` by default? Or perhaps `=...`?


---

hmm strange, I tried using `NotADirectoryError` as a default, but that doesn't work for some reason 🤔

numpy/_core/function_base.py
seberg: Yeah, should be fine, and at least on first glance, `import inspect` seems to be vastly faster than `import numpy`, so nothing to see!

numpy/_core/_add_newdocs.py
jorenham: The closest I was able to get it to `np._NoValue` is with `"($self, array, axis=0, dtype=None, out=None, keepdims=False, initial='<no default>', where=True)"`. But I think I prefer `None`

seberg: Yeah, it only supports `ast.literal_eval()`.

Did you actually try to include the signature twice?  Because on my quick try it seemed to work OK for the html docs (but I don't think my docs build fully, so...):
```
at(arg, *kwargs)
--

at(arg, initial=hahahahahahah)
```

Now, with `ipython` it's not too bad, with just `help()` in python it looks like:
```
at(a, indices, b, /) unbound numpy.ufunc method
    at(a, indices=hahahah, b=None, /)

    Performs unbuffered...
```
but maybe that is still OK? I don't like it, but I also don't really like having something wrong in the signature.

jorenham: > Yeah, it only supports `ast.literal_eval()`.
> 
> Did you actually try to include the signature twice? Because on my quick try it seemed to work OK for the html docs (but I don't think my docs build fully, so...):
> 
> ```
> at(arg, *kwargs)
> --
> 
> at(arg, initial=hahahahahahah)
> ```
> 
> Now, with `ipython` it's not too bad, with just `help()` in python it looks like:
> 
> ```
> at(a, indices, b, /) unbound numpy.ufunc method
>     at(a, indices=hahahah, b=None, /)
> 
>     Performs unbuffered...
> ```
> 
> but maybe that is still OK? I don't like it, but I also don't really like having something wrong in the signature.

Ahh I tried it before the `--`, lemme try this one then as well.

jorenham:
```
Signature: np.ufunc.reduce(self, /, array, axis=0, dtype=None, out=None, **kwargs)
Docstring:
reduce(array, axis=0, dtype=None, out=None, *, keepdims=False, initial=<no value>, where=True)

Reduces `array`'s dimension by one, by applying ufunc along one axis.
```

```py
In [2]: np.ufunc.reduce.__text_signature__
Out[2]: '($self, array, axis=0, dtype=None, out=None, **kwargs)'

In [3]: inspect.signature(np.ufunc.reduce)
Out[3]: <Signature (self, /, array, axis=0, dtype=None, out=None, **kwargs)>
```

I can live with that. Can you?

seberg: If it looks right in the html, I think that is the best of all the bad options (unless we just don't care about `__signature__` but I guess we do).

jorenham: I'll do the same for the other methods to keep things a bit consistent then

jorenham:
> Thanks for looking into it! A bit too bad about the `__text_signature__` limitations, makes me wonder if Python has hacks around it, but I guess probably not...

I'm not sure about C, but with rust + pyo3 you can do things like

```rust
use pyo3::prelude::*;

#[pymodule]
pub mod example {
   use pyo3::prelude::*;

   #[pyfunction]
   #[pyo3(signature = (arg: "list[int]") -> "list[int]")]
   fn list_of_int_identity(arg: Bound<'_, PyAny>) -> Bound<'_, PyAny> {
      arg
   }
}
```

and 

```rust
use pyo3::prelude::*;

/// This function adds two unsigned 64-bit integers.
#[pyfunction]
#[pyo3(signature = (a, b=0, /), text_signature = "(a, b=0, /)")]
fn add(a: u64, b: u64) -> u64 {
    a + b
}
```

[docs](https://pyo3.rs/v0.26.0/function/signature.html)

I think that's pretty slick (and completely irrelevant here)

jorenham:
I took a look at the new mypy_primer error for colour ([src](https://github.com/colour-science/colour/blob/638f7effb713582a8bdb9194879197b062740647/colour/phenomena/tmm.py#L176-L181)), and I'm pretty sure that it's a false positive, related to mypy's incomplete support for the overload typing spec, which could have the same underlying problem as #30103. 
It's not directly related to the changes here, and rather a consequence of mypy now being able to better understand ufuncs.

for reference, here's the static_frame primer diff source: https://github.com/static-frame/static-frame/blob/3807481e76b8305b5fb83b834cf5accdb02a5a12/static_frame/core/util.py#L2585

numpy/_core/_add_newdocs.py
```diff
@@ -5483,6 +5492,9 @@
 
 add_newdoc('numpy._core', 'ufunc', ('outer',
     r"""
+    outer($self, A, B, /, **kwargs)
+    --
+
```
seberg, line 5497: The duplication is needed to avoid the self in the html docs?

seberg (approved):
Thanks I am happy, just curios if duplication is needed for the valis cases.

numpy/_core/_add_newdocs.py
jorenham: Yea pretty much, but for consistency's sake too

charris:
Thanks Joren.

numpy/_core/_add_newdocs.py
```diff
@@ -5297,6 +5300,9 @@
 
 add_newdoc('numpy._core', 'ufunc', ('accumulate',
     """
+    accumulate($self, array, /, axis=0, dtype=None, out=None)
+    --
+
     accumulate(array, axis=0, dtype=None, out=None)
```
j-bowhay, line 5306: For my understanding, why is the signature needed twice?

jorenham: The first one is removed from the docstring and only used by e.g. `inspect.signature`. The second one is there for documentation purposes only.

PR #30108, opened 2025-10-31
TYP: shape-type-aware ``swapaxes``

jorenham:
Where "shape-type" means the type of `ndarray.shape` attribute, i.e. a tuple of integers types (not literals).

This also makes the `axis1` and `axis2` params positional only, because unlike what the docstring suggests, that's what they are:

```pycon
>>> import numpy as np
>>> np.eye(3).swapaxes(axis1=1, axis2=0)
    np.eye(3).swapaxes(axis1=1, axis2=0)
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
TypeError: ndarray.swapaxes() takes no keyword arguments
```


charris:
Thanks Joren.

PR #30110, opened 2025-10-31
ENH: Hash-based np.unique with support for return_index/return_inverse/return_counts

### LINKED ISSUE #28374: Hash based np.unique to support return_index, return_inverse, and return_counts
https://github.com/numpy/numpy/pull/26018 calculates unique values using a hash set, which means `return_index`, `return_inverse`, and `return_counts` are not supported.

We can use a hash map instead and add support for those functionalities as well.

math-hiyoko:
close #28374


math-hiyoko:
This PR now have a lot of conflicts, so I split the work into a separate branch and opened a new PR.
https://github.com/numpy/numpy/pull/30496

PR #30111, opened 2025-10-31
DOC: Add a plot to the 'unwrap' docstring.

WarrenWeckesser:
I recently implemented [unwrap](https://github.com/WarrenWeckesser/ufunclab?tab=readme-ov-file#unwrap) as a gufunc in [ufunclab](https://github.com/WarrenWeckesser/ufunclab), and I thought the visualization that I used in an example there would be a good addition to the docstring of `numpy.unwrap`.


mattip:
Thanks @WarrenWeckesser

PR #30112, opened 2025-10-31
ENH: New-style sorting for StringDType

### LINKED ISSUE #26510: ENH: np.unique and sorting is slow for StringDType

MaanasArora:
Implements merge sorts for StringDType using the feature introduced in #29737. The sorts allocate the mutex lock only once, thus fixes #26510.

Instead of utilizing the generic sorts as done in #29987, this PR completely rewrites the sorts. While that is more verbose, just duplicating the code can help move to more specialized string sorting algorithms and make optimizations independently from the `npysort` code. Also, we don't need to rely on the 'hack' in the other PR i.e. indirect use of sort compare function.

Posting a benchmark in a comment below; there is ~3x gain in speed on stable sort. ping @seberg - thanks!

Edit: also supports descending sorts now!


MaanasArora:
Running this script:

```python
import numpy
import random
import timeit

print(numpy.__version__)

options = ["left", "right", "leftovers", "righty", "up", "down", numpy.nan]
arr_s = numpy.array(lst, dtype="T")

time = timeit.timeit(
    "numpy.sort(arr_s, kind='stable')",
    globals=globals(),
    number=1000,
    setup="import numpy",
)
print(f"Time taken for stable sort: {time:.6f} seconds")
```

produces on main:

```
2.4.0.dev0+git20251030.d8eb225
Time taken for stable sort: 0.397690 seconds
```

and with this PR:

```
2.4.0.dev0+git20251030.dab50e6
Time taken for stable sort: 0.120393 seconds
```

Edit: just noting that this doesn't implement quick (non stable) sort so far - I didn't find a significant performance difference when working with this script and a few other tests, but I didn't really try out a lot of variety. It would also be more verbose if we added it. I'll try running experiments sometime.

numpy/_core/src/multiarray/array_method.c
```diff
@@ -342,13 +342,13 @@ fill_arraymethod_from_slots(
...
-                        spec->name);
-                return -1;
-            }
+            // if (i >= meth->nin && NPY_DT_is_parametric(res->dtypes[i])) {
+            //     PyErr_Format(PyExc_TypeError,
+            //             "must provide a `resolve_descriptors` function if any "
+            //             "output DType is parametric. (method: %s)",
+            //             spec->name);
+            //     return -1;
+            // }
```
MaanasArora, line 351: This should not be merged!!! I noticed that we actually cannot skip `resolve_descriptors` for any parametric dtype, at least for sorts, because the output is parametric. Should we special-case somehow?

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -3183,6 +3183,9 @@ PyArray_Sort(PyArrayObject *op, int axis, NPY_SORTKIND flags)
         // Arrays are always contiguous for sorting
         npy_intp strides[2] = {loop_descrs[0]->elsize, loop_descrs[1]->elsize};
 
+        NPY_ARRAYMETHOD_FLAGS flags = 0;
+        method_flags = &flags;
+
```
MaanasArora: Fixed a bug here and below!

numpy/_core/src/multiarray/stringdtype/sorts.c
```diff
@@ -0,0 +1,321 @@
...
+    PyBoundArrayMethodObject *argsort_method = PyArrayMethod_FromSpec_int(&argsort_spec, 1);
+    if (argsort_method == NULL) {
+        return -1;
+    }
+    NPY_DT_SLOTS(&PyArray_StringDType)->argsort_meth =
+            (PyArrayMethodObject *)argsort_method->method;
+    Py_INCREF(argsort_method->method);
+    Py_DECREF(argsort_method);
+
+    return 0;
```
MaanasArora, line 302: Unfortunately we cannot use `PyUFunc_AddLoopsWithSpecs` if we group this with the ufuncs because NumPy is not fully initialized when multiarray is initialized.

ngoldbaum:
Added a 2.4.0 milestone - it'd be really nice to get the performance boost shipped this month. I'll try to give this a once-over this week sometime.

numpy/_core/src/multiarray/stringdtype/sorts.h
```diff
@@ -0,0 +1,44 @@
+#ifndef _NPY_CORE_SRC_MULTIARRAY_STRINGDTYPE_SORTS_H_
+#define _NPY_CORE_SRC_MULTIARRAY_STRINGDTYPE_SORTS_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define SMALL_STRING_MERGESORT 20
```
ngoldbaum, line 8: what motivates this choice? maybe add a comment

numpy/_core/src/multiarray/array_method.c
ngoldbaum: I didn't follow all the discussions that led to using an ArrayMethod here. I think you explained elsewhere but I'm not finding it: why don't the sort ArrayMethods care about `resolve_descriptors`? Because sorts have nin = nout = 1 and the input *is* the output, or something along those lines? If that's what it is, maybe this check needs to be conditional on the input and output descriptors so that trivial ArrayMethods can skip it.

numpy/_core/src/multiarray/stringdtype/sorts.c
```diff
@@ -0,0 +1,303 @@
...
+            pk = pi - elsize;
+            while (pj > pl && _cmp(vp, pk, sdescr, descending) < 0) {
+                memcpy(pj, pk, elsize);
+                pj -= elsize;
+                pk -= elsize;
+            }
+            memcpy(pj, vp, elsize);
+        }
+    }
+}
```
ngoldbaum, line 111: This is pretty tricky C code - I would be very careful to make sure you're implementing it correctly (e.g. by adding a lot of tests). Or maybe use the C++ standard library instead?

numpy/_core/src/multiarray/stringdtype/sorts.c
```diff
@@ -0,0 +1,303 @@
...
+            vp = v + vi * elsize;
+            pj = pi;
+            pk = pi - 1;
+            while (pj > pl && _cmp(vp, v + (*pk) * elsize, sdescr, descending) < 0) {
+                *pj-- = *pk--;
+            }
+            *pj = vi;
+        }
+    }
+}
```
ngoldbaum, line 207: this too - can we maybe rely on the C++ standard library instead of having an algorithm implementation here in the stringdtype internals?

numpy/_core/src/umath/stringdtype_ufuncs.cpp
```diff
@@ -3128,5 +3129,9 @@ init_stringdtype_ufuncs(PyObject *umath)
         return -1;
     }
 
+    if (init_stringdtype_sorts() < 0) {
+        return -1;
+    }
```
ngoldbaum, line 3134: maybe you can avoid the numpy initialization problem you're having if you move this into `init_string_dtype` over in the `dtype.c` file in the `multiarray/stringdtype` folder.

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -1441,7 +1441,7 @@ _new_argsortlike(PyArrayObject *op, int axis, PyArray_ArgSortFunc *argsort,
     rstride = PyArray_STRIDE(rop, axis);
     needidxbuffer = rstride != sizeof(npy_intp);
 
-    if (method_flags != NULL) {
+    if (strided_loop != NULL) {
```
ngoldbaum, line 1444: are the two changes above bugfixes?

numpy/_core/src/multiarray/stringdtype/sorts.c
ngoldbaum: see https://github.com/numpy/numpy/pull/30112/files#r2501032204

ngoldbaum (commented):
Overall seems nice but incomplete. Left some comments inline.

Did you think about using C++ instead of C?

IMO this needs explicit string sorting tests, especially given the custom algorithm code in this PR. I'm not sure offhand if there's a good test corpus - maybe look in the CPython test suite?

Even if you change to use a C++ standard library sorting algorithm, I'd still like to see some more tests for sorting in `test_stringdtype.py` before this is merged.

numpy/_core/src/multiarray/item_selection.c
MaanasArora: Yes, bug fixes. This adjusts for the fact (below) that `method_flags` should actually never point to NULL, so it can be set in `get_strided_loop`.

numpy/_core/src/multiarray/stringdtype/sorts.c
MaanasArora: Sure, one could move to C++ standard library! I didn't exactly think about it because this code is directly copied and adjusted from the `npysort` merge sorts: https://github.com/numpy/numpy/blob/2b216b1a6072d28e4aa7b02a7e13e69713a20c5d/numpy/_core/src/npysort/mergesort.cpp#L338-L412

I think we need an extensive test suite either way, so I'll start with that (great point, thanks)! I don't think it matters, but it also might be slightly harder to make future optimizations or extensions if we use C++ sorts.

numpy/_core/src/multiarray/stringdtype/sorts.c
MaanasArora: Same as above - I'd add an extensive test suite, and if it still seems to be an issue (due to complexity for example) we can see!

numpy/_core/src/multiarray/stringdtype/sorts.c
MaanasArora: Unfortunately I tried `init_string_dtype` at first, and it still doesn't work because that is also initialized in `multiarray` as far as I can tell.

numpy/_core/src/multiarray/stringdtype/sorts.h
MaanasArora: Same as the algorithm, this is directly taken from `npysort` :

https://github.com/numpy/numpy/blob/2b216b1a6072d28e4aa7b02a7e13e69713a20c5d/numpy/_core/src/npysort/mergesort.cpp#L40

numpy/_core/src/umath/stringdtype_ufuncs.cpp
MaanasArora: See https://github.com/numpy/numpy/pull/30112/files#r2480095384

numpy/_core/src/multiarray/array_method.c
MaanasArora: Yes, it's because the output is essentially fake for sorts (we enforce `data[0] == data[1]`. Do you mean we can check if that is the case here and only error if it's not? If so makes sense -- I'll do that, thank you!

MaanasArora (commented):
Thanks for reviewing! I didn't look much at C++ - it makes sense to do that I think, though as I've said in a comment I'm unsure if using their standard sorts is the way to go:

- The sorting code is directly copied from the `npy_mergesort` code in `npysort`, so this implementation provides some consistency.
- In the future, if someone wants to do this more efficiently, add features, or even use SIMD (!), having custom code makes that easier.

That being said I totally see it can get very complicated to have lots of tricky algorithm code lying around. I actually see it as a very good reason to actually make our sorting infrastructure reusable, a public library, and perhaps use C++ templates to expose the generic sorts. Maybe this would be preferable to using C++ sorts simply because we can make extensions, even if we actually use C++ sorts internally! (See #29987 for a very initial step).

For the short term though, if it is very tricky to review this even given the reuse, I am happy to move to the C++ standard library - just making a general point. I'll add expanded test coverage soon. Thanks again!

numpy/_core/src/multiarray/array_method.c
MaanasArora: Sorry, I just realized this is initialization, so we don't actually have the descriptors. Not sure what the condition could be...

ngoldbaum:
Ah, that makes more sense, I didn't realize the sorting algorithm code was copied from elsewhere. Maybe we can refactor things so this calls into `npy_mergesort`? That would avoid duplication. Or maybe that's what you were saying about making the sorting code reusable.

I'll try to poke around in a debugger to see if I can come up with a suggestion for the `resolve_descriptors` stuff.

MaanasArora:
Thanks! Yeah unfortunately the main issue with calling into `npy_mergesort` is that the `sort_compare` is not implemented yet, nor descending. (Actually, #29987 has a workaround for exactly this.)

ngoldbaum:
I guess that argues for merging that PR before this one. I'll try to give that one a look soon. Sorry for letting it drop...

MaanasArora:
No worries at all! It actually does `StringDType` too, but it's a larger change and there are some confusing points about the API there. Completely fine with doing this just for now!

MaanasArora:
I've added tests using an ad-hoc string corpus; I couldn't find anything standard, sorry. It is two lorem ipsum passsages with some random unicode strings in between.

I'm getting some very bizzare failures with running the new test, only for argsort, and only when `na_object` is `unset` and the corpus has duplicates... I suspect it has to do something with the allocator. Debugging now.

MaanasArora:
Closing in favor of #30266.

PR #30114, opened 2025-10-31
BUG, TYP: ``ndarray`` method runtime signatures

jorenham:
Same story as #30104, but for ``ndarray`` this time.

This fixes a `ValueError` from being raised when using `inspect.signature` on the `ndarray` methods. 
The typing stubs changes are needed because stubtest is now able to "see" the actual signatures, which was causing several errors (related to pos-only parameters mostly).

This also adds a (very minimal) docstring for `ndarray.dot`, which was apparently missing (see the rendered docs from [before](https://numpy.org/devdocs/reference/generated/numpy.ndarray.dot.html) and [after](https://output.circle-artifacts.com/output/job/742a3ba6-4e81-4f2d-9221-081994e6f1ba/artifacts/0/doc/build/html/reference/generated/numpy.ndarray.dot.html))

### DIFF
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -2982,6 +2982,7 @@ add_newdoc('numpy._core.multiarray', 'ndarray', ('mT',
             [5, 7]]])
 
     """))
+
 ##############################################################################
 #
 # ndarray methods
@@ -2991,6 +2992,9 @@ add_newdoc('numpy._core.multiarray', 'ndarray', ('mT',
 
 add_newdoc('numpy._core.multiarray', 'ndarray', ('__array__',
     """
+    __array__($self, dtype=None, /, *, copy=None)
+    --
+
     a.__array__([dtype], *, copy=None)
 
     For ``dtype`` parameter it returns a new reference to self if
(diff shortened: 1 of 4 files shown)


numpy/_core/_add_newdocs.py
```diff
@@ -3028,6 +3038,9 @@
 
 add_newdoc('numpy._core.multiarray', 'ndarray', ('__copy__',
     """
+    __copy__($self, /)
+    --
+
     a.__copy__()
 
     Used if :func:`copy.copy` is called on an array. Returns a copy of the array.
```
charris, line 3046: Is it a deep copy of object  arrays?

numpy/_core/_add_newdocs.py
```diff
@@ -3154,6 +3188,9 @@
 
 add_newdoc('numpy._core.multiarray', 'ndarray', ('argmin',
     """
+    argmin($self, /, axis=None, out=None, *, keepdims=False)
+    --
+
     a.argmin(axis=None, out=None, *, keepdims=False)
 
     Return indices of the minimum values along the given axis.
```
charris, line 3196: Just a single index I believe.

charris:
The docs could use a bit of work, but I'm going to put this in as a good start. Thanks Joren.

PR #30117, opened 2025-10-31
ENH: Speedup `np.vectorize` wrapped functions when keyword args are present

Yeshwanth-G:
Fixes https://github.com/numpy/numpy/issues/30116

---

`np.vectorize` wrapped functions, when keyword arguments are involved were doing a some sanitization steps in every call to the function.

This commit enhances `vectorize._call_as_normal` to avoid these sanization steps when function only has pure positional / positional_or_keyword type params. We filter them by looking into `inspect.signature(foo).parameters`

Taking example from https://github.com/numpy/numpy/issues/30116 with this changes:
```
>>> def func(a, b):
...     return a + b

>>> %time _ = np.vectorize(func)(np.arange(10**7), 1)
CPU times: user 534 ms, sys: 76.8 ms, total: 611 ms
Wall time: 610 ms

>>> %time _ = np.vectorize(func)(np.arange(10**7), b=1) # equivalent performance as above.
CPU times: user 543 ms, sys: 67.3 ms, total: 610 ms
Wall time: 614 ms
```

Thanks!

### DIFF
--- a/benchmarks/benchmarks/bench_function_base.py
+++ b/benchmarks/benchmarks/bench_function_base.py
@@ -377,3 +377,19 @@ class Where(Benchmark):
 
     def time_interleaved_ones_x8(self):
         np.where(self.rep_ones_8)
+
+
+class Vectorize(Benchmark):
+    """
+    Performance tests for numpy.vectorize
+    """
+    def setup(self):
+        self.a = np.arange(10)
+        self.b = 20
+        self.c = 30
+    
+    def time_vectorize_positional_or_kw_args(self):
+        # Test performance of np.vectorize when there are no keyword only arguments.
+        def func(a, b, /, c):
+            return a + b + c
+        res = np.vectorize(func)(self.a, self.b, c=self.c)
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -2,6 +2,8 @@ import builtins
 import collections.abc
 import functools
 import re
+import inspect
+import sys
 import warnings
 
 import numpy as np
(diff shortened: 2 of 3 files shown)


benchmarks/benchmarks/bench_function_base.py
```diff
@@ -377,3 +377,19 @@ def time_interleaved_ones_x4(self):
...
+    def setup(self):
+        self.a = np.arange(10)
+        self.b = 20
+        self.c = 30
+    
+    def time_vectorize_positional_or_kw_args(self):
+        # Test performance of np.vectorize when there are no keyword only arguments.
+        def func(a, b, /, c):
+            return a + b + c
+        res = np.vectorize(func)(self.a, self.b, c=self.c)
```
jorenham, line 395: it's missing one of those trailing newline

numpy/lib/_function_base_impl.py
```diff
@@ -2486,9 +2488,36 @@ def _call_as_normal(self, *args, **kwargs):
         `args` and `kwargs` not in `excluded`.
         """
         excluded = self.excluded
+        
+        try:
+            pyfunc_params = dict(inspect.signature(self.pyfunc).parameters.items())
```
jorenham, line 2493: `inpsect.Signature.params` already is a dict :)

```suggestion
            pyfunc_params = inspect.signature(self.pyfunc).parameters
```

numpy/lib/_function_base_impl.py
```diff
@@ -2486,9 +2488,36 @@ def _call_as_normal(self, *args, **kwargs):
...
+        except:
```
jorenham, line 2494: base `except` = evil, as it will also catch `SystemExit` and stuff

```suggestion
        except ValueError:
```

numpy/lib/_function_base_impl.py
```diff
@@ -2486,9 +2488,36 @@ def _call_as_normal(self, *args, **kwargs):
...
+            pyfunc_params = None
+
         if not kwargs and not excluded:
             func = self.pyfunc
             vargs = args
+        elif (not excluded and pyfunc_params is not None and
+              all([param.kind is param.POSITIONAL_ONLY or
+                   param.kind is param.POSITIONAL_OR_KEYWORD
+                   for param in pyfunc_params.values()])):
```
jorenham, line 2503:
```suggestion
              all(param.kind in (param.POSITIONAL_ONLY, param.POSITIONAL_OR_KEYWORD)
                  for param in pyfunc_params.values())):
```

this should still fit within the 88 character limit I think

numpy/lib/_function_base_impl.py
```diff
@@ -2486,9 +2488,36 @@ def _call_as_normal(self, *args, **kwargs):
...
+                    effective_args.append(args[ind])
+                    ind += 1
+                elif name in kwargs:
+                    if param.kind is inspect.Parameter.POSITIONAL_ONLY:
+                        raise TypeError(f"Param {name} is positional only")
+                    effective_args.append(kwargs[name])
+                elif param.default is not inspect.Parameter.empty:
+                    effective_args.append(param.default)
+                else:
+                    raise TypeError(f"Missing param: {name}")
```
jorenham, line 2518: Could you add some concise comments to help explain what's going on here? Because it looks pretty complex.

jorenham (commented):
Thanks for this. I've coincidentally been using `inspect.signature`  quite a bit, so I left some comments I think could help improve this.

Yeshwanth-G:
Thanks for quick review @jorenham! I also see some lint failures. I'll address comments, failures and get back on this

PR #30118, opened 2025-10-31
CI: disable flaky ubuntu UBsan CI job

ngoldbaum:
To help with #30115

I don't think this will completely fix the flakyness but it should help by disabling an expensive CI job and by tweaking the remaining job.


ngoldbaum:
Ping @m-clare (just FYI)

.github/workflows/compiler_sanitizers.yml
```diff
@@ -49,16 +49,16 @@ jobs:
         pyenv --version
     - name: Set up LLVM
       run: |
-        brew install llvm@19
-        LLVM_PREFIX=$(brew --prefix llvm@19)
+        brew install llvm
+        LLVM_PREFIX=$(brew --prefix llvm)
```
jorenham: no pin?

ngoldbaum: Yeah, I wanted to update LLVM as well, I don't think pinning here makes any sense.

ngoldbaum: It looks like there are issues with linking against the C++ standard library with LLVM 21, so I guess pinning to LLVM 20 is necessary for now 😢

rgommers (approved):
LGTM, thanks @ngoldbaum

rgommers:
There are two tests with a complaint by the way, not sure if those are on your radar?

```
numpy/_core/tests/test_array_coercion.py::TestArrayLikes::test_arraylike_classes PASSED
==24803==WARNING: AddressSanitizer failed to allocate 0x4000000000000000 bytes

 numpy/_core/tests/test_multiarray.py::TestCreation::test_from_attribute PASSED
==24803==WARNING: AddressSanitizer failed to allocate 0x7fffffffffffffff bytes
```

PR #30119, opened 2025-10-31
Fix link text in contribution guide section

Naseer-fez:


jorenham:
This doesn't look like an improvement to me.

Naseer-fez:
> This doesn't look like an improvement to me.

Yeah it is not, i just saw some video on this , so i tried , i will not repeat this kind of behavior again

jorenham:
> i just saw some video on this

Ah yea you're not the first one. Well, I appreciate you're honesty.

PR #30121, opened 2025-10-31
BUG, TYP: scalar-type constructor runtime signatures

jorenham:
Following #30104 and #30114, this fixes `inspect.signature` from raising an error when called on the scalar-type constructors, e.g. `inspect.signature(np.int8)`.

To that end, I performed some surgery on `add_newdoc_for_scalar_type`, and replaced an used parameter with one for the `__text_signature__` string, making things a bit more dry and readable. I'm assuming here that `add_newdoc_for_scalar_type` is private API.

Oh and as it turns out, the docstrings for `long` and `ulong` were completely missing, and therefore also for `int32`/`uint32` on some windows platforms.


charris:
Thanks Joren.

PR #30124, opened 2025-11-01
BUG, TYP: ``flatiter`` method runtime signatures, and better stubs

jorenham:
After #30104, #30114, and #30121, this fixes the `inspect.signature` for the `copy` and `__array__` methods of `numpy.flatiter`. 
While updating the relevant stubs I improved some of the other `flatiter` method stubs. It is now better able to deal with shape-types and dtypes. I also brushed up the type-tests for it, which are now pretty thorough.


charris:
Thanks Joren.

PR #30126, opened 2025-11-01
BUG: ``nditer`` runtime signatures

jorenham:
This is the next batch of `inspect.signature` fixes after  #30104, #30114, and #30121, #30124, this time addressing the `signature(np.nditer)` constructor and that of its methods.

But wait, where are the typing changes?
Well, there's not really needed this time, because the signatures in the stubs where already spot-on (according to stubtest). And even though I noticed a bunch of ways to improve `nditer`, I miraculously managed to keep myself from going out-of-scope, i.e. my signature move (pun intended), this time.

### DIFF
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -154,6 +154,19 @@ add_newdoc('numpy._core', 'flatiter', ('copy',
 
 add_newdoc('numpy._core', 'nditer',
     """
+    nditer(
+        op,
+        flags=None,
+        op_flags=None,
+        op_dtypes=None,
+        order='K',
+        casting='safe',
+        op_axes=None,
+        itershape=None,
+        buffersize=0,
+    )
+    --
+
     nditer(op, flags=None, op_flags=None, op_dtypes=None, order='K',
         casting='safe', op_axes=None, itershape=None, buffersize=0)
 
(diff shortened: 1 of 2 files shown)


numpy/_core/_add_newdocs.py
```diff
@@ -504,32 +536,50 @@
...
+    """))
+
+# nested_iters
+
 add_newdoc('numpy._core', 'nested_iters',
     """
-    nested_iters(op, axes, flags=None, op_flags=None, op_dtypes=None, \
-    order="K", casting="safe", buffersize=0)
+    nested_iters(op, axes, flags=None, op_flags=None, op_dtypes=None,
+    order='K', casting='safe', buffersize=0)
```
charris, line 568: Hmm, will the space go missing?

charris: Seems the signature keeps the line break and this looks the same as previously.

charris:
Thanks Joren.

numpy/_core/_add_newdocs.py
jorenham: When I run `np.nested_iters?` in an `ipython` console on numpy 2.3, I see

```
nested_iters(op, axes, flags=None, op_flags=None, op_dtypes=None,     order="K", casting="safe", buffersize=0)
```

I thought that an explicit linebreak looked better than these additional spaces

PR #30132, opened 2025-11-03
BLD: use scipy-openblas 0.3.30.7

mattip:
Fixes #30092.

Rather than update the OpenBLAS version, this PR uses a patched version of OpenBLAS that incorporate the fix from https://github.com/OpenMathLib/OpenBLAS/issues/5520. That should minimize the risk in updating the dependency so close to the new release cut-off date.

### DIFF
--- a/requirements/ci32_requirements.txt
+++ b/requirements/ci32_requirements.txt
@@ -1,3 +1,3 @@
 spin==0.15
 # Keep this in sync with ci_requirements.txt
-scipy-openblas32==0.3.30.0.6
+scipy-openblas32==0.3.30.0.7
--- a/requirements/ci_requirements.txt
+++ b/requirements/ci_requirements.txt
@@ -1,4 +1,4 @@
 spin==0.15
 # Keep this in sync with ci32_requirements.txt
-scipy-openblas32==0.3.30.0.6
-scipy-openblas64==0.3.30.0.6
+scipy-openblas32==0.3.30.0.7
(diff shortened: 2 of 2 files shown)


mattip:
@FreddieWitherden could you try this out on your system?

Edit: it should be enough to update scipy-openblas64 on the reproducer from #30092, since that seems to be a local build and not a PyPI install.

FreddieWitherden:
> @FreddieWitherden could you try this out on your system?
> 
> Edit: it should be enough to update scipy-openblas64 on the reproducer from #30092, since that seems to be a local build and not a PyPI install.

Will action this evening and get back to you.

charris:
Note that this will also be needed in the `numpy-release` repo which currently uses 0.3.30.0.2. The maintenance branch also builds onn macos-13, which I think should be good through Dec.

andyfaff:
I think the macos-13 limit is fully retired by Dec 4th.

FreddieWitherden:
> @FreddieWitherden could you try this out on your system?
> 
> Edit: it should be enough to update scipy-openblas64 on the reproducer from #30092, since that seems to be a local build and not a PyPI install.

I can confirm this resolves the deadlock issue.

Regards, Freddie.

mattip:
> Note that this will also be needed in the `numpy-release` repo

I opened numpy/numpy-release#21 to simplify that.

rgommers (approved):
Thanks @mattip and @FreddieWitherden!

PR #30137, opened 2025-11-03
BUG: ``broadcast`` runtime signatures

jorenham:
Following #30104, #30114, #30121, #30124, and #30126, this adds the missing `__text_signature__`s to the ``broadcast`` constructor and its `reset` method.

before:

```pycon
>>> signature(np.broadcast)
[...]
ValueError: no signature found for builtin type <class 'numpy.broadcast'>
>>> signature(np.broadcast.reset)
[...]
ValueError: no signature found for builtin <method 'reset' of 'numpy.broadcast' objects>
```

after:

```pycon
>>> signature(np.broadcast)
<Signature (*arrays)>
>>> signature(np.broadcast.reset)
<Signature (self, /)>
```

### DIFF
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -638,6 +638,9 @@ add_newdoc('numpy._core', 'nested_iters',
 
 add_newdoc('numpy._core', 'broadcast',
     """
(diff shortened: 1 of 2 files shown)


charris:
Thanks Joren.

PR #30138, opened 2025-11-03
BUG: array construction function runtime signatures

jorenham:
Following #30104, #30114, #30121, #30124, #30126, and #30137; this adds the missing `__text_signature__`s to `array`, `asarray`, `asanyarray`, `ascontiguousarray`, and `asfortranarray`.

---

Before:

```pycon
>>> import numpy as np
>>> signature(np.array)
    signature(np.array)
    ~~~~~~~~~^^^^^^^^^^
  File "/home/joren/.local/share/uv/python/cpython-3.14-linux-x86_64-gnu/lib/python3.14/inspect.py", line 3312, in signature
    return Signature.from_callable(obj, follow_wrapped=follow_wrapped,
           ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                   globals=globals, locals=locals, eval_str=eval_str,
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                   annotation_format=annotation_format)
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/joren/.local/share/uv/python/cpython-3.14-linux-x86_64-gnu/lib/python3.14/inspect.py", line 3027, in from_callable
    return _signature_from_callable(obj, sigcls=cls,
                                    follow_wrapper_chains=follow_wrapped,
                                    globals=globals, locals=locals, eval_str=eval_str,
                                    annotation_format=annotation_format)
  File "/home/joren/.local/share/uv/python/cpython-3.14-linux-x86_64-gnu/lib/python3.14/inspect.py", line 2508, in _signature_from_callable
    return _signature_from_builtin(sigcls, obj,
                                   skip_bound_arg=skip_bound_arg)
  File "/home/joren/.local/share/uv/python/cpython-3.14-linux-x86_64-gnu/lib/python3.14/inspect.py", line 2292, in _signature_from_builtin
    raise ValueError("no signature found for builtin {!r}".format(func))
ValueError: no signature found for builtin <built-in function array>
```

After:

```pycon
>>> import numpy as np
>>> signature(np.array)
<Signature (object, dtype=None, *, copy=True, order='K', subok=False, ndmin=0, ndmax=None, like=None)>
```

### DIFF
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -847,6 +847,19 @@ add_newdoc('numpy._core', 'broadcast', ('reset',
 
 add_newdoc('numpy._core.multiarray', 'array',
     """
+    array(
+        object,
+        dtype=None,
+        *,
+        copy=True,
+        order='K',
+        subok=False,
+        ndmin=0,
+        ndmax=None,
+        like=None,
+    )
+    --
+
     array(object, dtype=None, *, copy=True, order='K', subok=False, ndmin=0,
           ndmax=None, like=None)
 
@@ -925,7 +938,7 @@ add_newdoc('numpy._core.multiarray', 'array',
     ones : Return a new array setting values to one.
     zeros : Return a new array setting values to zero.
     full : Return a new array of given shape filled with value.
-    copy: Return an array copy of the given object.
+    copy : Return an array copy of the given object.
 
 
     Notes
@@ -997,6 +1010,9 @@ add_newdoc('numpy._core.multiarray', 'array',
 add_newdoc('numpy._core.multiarray', 'asarray',
     """
     asarray(a, dtype=None, order=None, *, device=None, copy=None, like=None)
+    --
+
+    asarray(a, dtype=None, order=None, *, device=None, copy=None, like=None)
 
     Convert the input to an array.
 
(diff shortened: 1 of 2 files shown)


jorenham:
I'm kinda tempted to make the first argument (called `object` or `a`) positional-only in these signatures (and only there). That will cause stubtest to require them to also be annotated that way in the stubs. It'll be a bit of a lie though.
Thoughts?

numpy/_core/_add_newdocs.py
```diff
@@ -996,6 +1009,9 @@
 
 add_newdoc('numpy._core.multiarray', 'asarray',
     """
+    asarray(a, dtype=None, order=None, *, device=None, copy=None, like=None)
+    --
+
```
charris, line 1014: I'm curious about these repeated signatures and the `--`. Why are they needed?

jorenham: It's an [Argument Clinic](https://devguide.python.org/development-tools/clinic/) thing:  The signature above the `"\n--\n\n"` will be converted by cpython into a `__text_signature__`, which is understood by `inspect.signature`. It'll be stripped from the `__doc__`, so that's why I left the other one.

charris:
Thanks Joren.

PR #30139, opened 2025-11-04
MAINT,BUG: make later arguments in array2string keyword only.

mhvk:
In #30068, the deprecated `style` argument from `array2string` was removed. With that, however, all following arguments shift one up if used as positional arguments. This seems highly undesirable, as any code that used those positional arguments will now fail.

The only solution would seem to be to make those later arguments keyword-only -- which this PR does.

p.s. It seems highly unlikely that all positional arguments were used in the wild, but our code in astropy that implements `__array_function__` did, and broke by #30068. We don't mind adjusting, but it seemed better to just fix the problem here.

### DIFF
--- a/doc/release/upcoming_changes/30068.expired.rst
+++ b/doc/release/upcoming_changes/30068.expired.rst
@@ -4,7 +4,8 @@
 The following long-deprecated APIs have been removed or converted to errors:
 
 * The ``style`` parameter has been removed from ``numpy.array2string``.
-  This argument had no effect since Numpy 1.14.0.
+  This argument had no effect since Numpy 1.14.0.  Any arguments following
+  it, such as ``formatter`` have now been made keyword-only.
 
 * Calling ``np.sum(generator)`` directly on a generator object now raises a `TypeError`.
   This behavior was deprecated in NumPy 1.15.0. Use ``np.sum(np.fromiter(generator))``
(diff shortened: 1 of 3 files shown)


doc/release/upcoming_changes/30068.expired.rst
```diff
@@ -4,7 +4,8 @@
 The following long-deprecated APIs have been removed or converted to errors:
 
 * The ``style`` parameter has been removed from ``numpy.array2string``.
-  This argument had no effect since Numpy 1.14.0.
+  This argument had no effect since Numpy 1.14.0.  Any arguments following
+  it, such as ``formatter`` have now been made keyword-only.
```
jorenham, line 8:
```suggestion
  This argument had no effect since Numpy 1.14.0.  The parameters following
  it, such as ``formatter``, have now been made keyword-only.
```

jorenham (approved):
Apart from one nit this is looking good AFAIK

doc/release/upcoming_changes/30068.expired.rst
mhvk: In principle happy to change, but it is called `*args, **kwargs`, so I'm confused why one would write parameter instead of argument (even if "argument" maybe applies to what we are having -- a friendly one -- rather than computer code...).

jorenham: Well, basically, an argument is what you pass a parameter. So the parameter is how you define it, and the argument is how you use it.

charris:
Thanks Marten.

PR #30140, opened 2025-11-04
BUG, DOC, TYP: ``empty`` and ``zeros`` runtime signatures, and missing ``device`` parameter docs

jorenham:
Following #30104, #30114, #30121, #30124, #30126, #30137, and #30138; this adds the missing `__text_signature__`s to `empty`, `zeros`, and updates the relevant stubs.
Note that because `ones` and `full` are python functions, they did not have this issue.

This also adds docs for the `device` parameter of `zeros`, which apparently was missing.

---

Before:

```pycon
>>> import numpy as np
>>> signature(np.empty)
    signature(np.empty)
    ~~~~~~~~~^^^^^^^^^^
  File "/home/joren/.local/share/uv/python/cpython-3.14-linux-x86_64-gnu/lib/python3.14/inspect.py", line 3312, in signature
    return Signature.from_callable(obj, follow_wrapped=follow_wrapped,
           ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                   globals=globals, locals=locals, eval_str=eval_str,
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                   annotation_format=annotation_format)
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/joren/.local/share/uv/python/cpython-3.14-linux-x86_64-gnu/lib/python3.14/inspect.py", line 3027, in from_callable
    return _signature_from_callable(obj, sigcls=cls,
                                    follow_wrapper_chains=follow_wrapped,
                                    globals=globals, locals=locals, eval_str=eval_str,
                                    annotation_format=annotation_format)
  File "/home/joren/.local/share/uv/python/cpython-3.14-linux-x86_64-gnu/lib/python3.14/inspect.py", line 2508, in _signature_from_callable
    return _signature_from_builtin(sigcls, obj,
                                   skip_bound_arg=skip_bound_arg)
  File "/home/joren/.local/share/uv/python/cpython-3.14-linux-x86_64-gnu/lib/python3.14/inspect.py", line 2292, in _signature_from_builtin
    raise ValueError("no signature found for builtin {!r}".format(func))
ValueError: no signature found for builtin <built-in function empty>
```

After:

```pycon
>>> import numpy as np
>>> signature(np.empty)
<Signature (shape, dtype=None, order='C', *, device=None, like=None)>
>>> signature(np.zeros)
<Signature (shape, dtype=None, order='C', *, device=None, like=None)>
```

### DIFF
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -1293,7 +1293,10 @@ add_newdoc('numpy._core.multiarray', 'asfortranarray',
 
 add_newdoc('numpy._core.multiarray', 'empty',
     """
-    empty(shape, dtype=float, order='C', *, device=None, like=None)
+    empty(shape, dtype=None, order='C', *, device=None, like=None)
+    --
+
+    empty(shape, dtype=None, order='C', *, device=None, like=None)
 
     Return a new array of given shape and type, without initializing entries.
 
@@ -1306,8 +1309,7 @@ add_newdoc('numpy._core.multiarray', 'empty',
         `numpy.float64`.
     order : {'C', 'F'}, optional, default: 'C'
         Whether to store multi-dimensional data in row-major
-        (C-style) or column-major (Fortran-style) order in
-        memory.
+        (C-style) or column-major (Fortran-style) order in memory.
     device : str, optional
         The device on which to place the created array. Default: ``None``.
         For Array-API interoperability only, so must be ``"cpu"`` if passed.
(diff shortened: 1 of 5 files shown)


jorenham:
the primer error is a bit weird, but it's not an indiciation of a regression or something. Basically "need type annotations" means that mypy isn't able to infer the type itself, but in many cases where it also can't infer the type, mypy doesn't report this error.
So this diff doesn't really tell us much.

https://github.com/pydata/xarray/blob/ea35e64cec1a72e4c3270f60c7a937e59a36f0ce/xarray/tests/test_namedarray.py#L246-L250

numpy/_core/_add_newdocs.py
```diff
@@ -1293,7 +1293,10 @@
 
 add_newdoc('numpy._core.multiarray', 'empty',
     """
-    empty(shape, dtype=float, order='C', *, device=None, like=None)
+    empty(shape, dtype=None, order='C', *, device=None, like=None)
```
charris, line 1296: Hmm, python does pass `None` here, but that defaults to float64.  I guess this is more accurate.

charris:
Thanks Joren.

PR #30141, opened 2025-11-04
MAINT: fix math markup (\times -> \\times) in numpy.linalg.multidot docstring Notes section

treycole:
On the NumPy documentation page for [numpy.linalg.mutli_dot](https://numpy.org/doc/stable/reference/generated/numpy.linalg.multi_dot.html#numpy.linalg.multi_dot), the math markup in Notes was previously

```python
:math:`A_{10 \times 100}, B_{100 \times 5}, C_{5 \times 50}`.
```

which did not render correctly as it was detected as an escape sequence. This has been updated to

```python
:math:`A_{10 \\times 100}, B_{100 \\times 5}, C_{5 \\times 50}`.
```

### DIFF
--- a/numpy/linalg/_linalg.py
+++ b/numpy/linalg/_linalg.py
@@ -2938,7 +2938,7 @@ def multi_dot(arrays, *, out=None):
             return A.shape[0] * A.shape[1] * B.shape[1]
 
     Assume we have three matrices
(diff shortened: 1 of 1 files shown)


mattip:
Thanks @treycole

PR #30142, opened 2025-11-04
MAINT: Migrate einsum.c.src to C++ (einsum.cpp)

athurdekoos:
**Motivation:**
Part of #29528, migrates einsum.c.src to C++.

No C++ templates were needed as einsum.c.src was not templated but was still a c.src file for historical reason. This is a good starting point for future c++ migration work.

**What this PR does:** 
- Renamed `einsum.c.src` to `einsum.cpp`
- Updated refrences in genapi.py and meson.build
- In `einsum.cpp`:
  - Added `extern "C"` C linkage where necessary
  - Added 4 explicit casts to maintain functionality


mattip:
> No C++ templates were needed as einsum.c.src was not templated

Wouldn't it be easier to rename it to `einsum.c`, with no further changes?

mattip:
There is a [failure in compilation](https://github.com/numpy/numpy/actions/runs/19056916192/job/54429123184?pr=30142#step:5:901), the `requires` identifier is reserved in C++-20. 

https://github.com/numpy/numpy/blob/17d15c74c0f7e2b56c86f301924d83abd1e98fe8/numpy/_core/src/multiarray/ctors.h#L69-L75

1. I am not sure why we have identifiers in the function declarations [0], but it should be enough to change the identifiers in the header file (and maybe elsewhere in the headers?) to match the [documentation](https://numpy.org/devdocs/reference/c-api/array.html#c.PyArray_CheckFromAny)
2. I am not sure where the `--Werror=c++20-compat` is coming from, it is specific to this OpenBLAS nightly build. I see `-Werror -std=c++17` in the compilation command, but no `-std-c++20`. Looking at the [`scipy-openblas wheel used](https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/scipy-openblas32/0.3.30.313/scipy_openblas32-0.3.30.313-py3-none-manylinux2014_x86_64.manylinux_2_17_x86_64.whl) I don't see any such directive in the pkg_config file. Confusing.

[0] The [C style guide in NEP 45](https://numpy.org/neps/nep-0045-c_style_guide.html#c-dialect) says "All function declarations and definitions must use full prototypes (that is, specify the types of all arguments)", which I don't understand. Should that be "that is, specify the types and names of all arguments"? Certainly the point of a C declaration is to "specify the types of all arguments". It also specifies C99 without c++-17

rgommers:
> 2\. I am not sure where the `--Werror=c++20-compat` is coming from

That isn't present as far as I can tell. All there is is `-Werror -std=c++17`, which is only present in this job because the CI config for it passes `--werror` to `meson setup`. It's a bit of a mystery, but not worth digging into it looks like, since it's trivial to fix this up in a separate PR.

rgommers:
> Should that be "that is, specify the types and names of all arguments"?

Yes, I think so - want to open a separate PR for that?

athurdekoos:
> Wouldn't it be easier to rename it to `einsum.c`, with no further changes?

Yes.

I was initially just mechanically going off of the list in #29528. When I noticed no templates, the difference was moot. I also wanted to prove the workflow and see if the community wanted a `einsum.c` or `einsum.cpp` file in this case.

rgommers (approved):
The casts all look correct, so this seems mergeable as is (EDIT: modulo the `requires` fix of course, which is good to do anyway). Re `einsum.c` vs. `einsum.cpp`, I don't have a strong opinion. Using `einsum.c` would be more natural normally; the one reason to use `einsum.cpp` would be that `einsum_sumprod.c.src` does have templates and is planned to be converted to C++, so there may be some usefulness to having it all in C++.

@mattip any preference?

athurdekoos:
#30151 issue created to track `requires` fix.

mattip:
As long as CI is passing, I don't really care. The problem right now is there is an `-Werror=c++20-compat` error from `einsum.cpp` including `ctors.h`: 
```
In file included from ../numpy/_core/src/multiarray/einsum.cpp:25:
../numpy/_core/src/multiarray/ctors.h:71:45: error: identifier ‘requires’ is a keyword in C++20 [-Werror=c++20-compat]
   71 |                          int max_depth, int requires, PyObject *context);
      |                                             ^~~~~~~~
cc1plus: all warnings being treated as errors
```

So either 
- the file name should be `einsum.c` or
- that identifier needs changing or
- the place where `-Werror=c++20-compat` is added needs changing

rgommers:
@athurdekoos now that the `requires` fix is in, can you rebase this to get CI green? Also, it may be useful to squash the commits and make the commit message adhere to the numpy standard form.

numpy/_core/src/multiarray/einsum.c.src
```diff
@@ -7,7 +7,6 @@
  *
  * See LICENSE.txt for the license.
  */
-
```
ngoldbaum, line 10: super minor nit: try to avoid whitespace changes like this

ngoldbaum (approved):
Awesome, it seems like everyone else has already given this a bunch more review and now things are looking good. I prefer having this be C++ if the next step is to convert `einsum_sumprod`.

numpy/_core/src/multiarray/einsum.c.src
athurdekoos: Good Catch! I've updated the file to re-include the whitespace.

rgommers:
Looks like we're all set here, CI & humans are happy - in it went.

PR #30145, opened 2025-11-04
DOC: add example for out in concatenate

antareepsarkar:
## Description
This adds an example of the 'out' parameter with an argument in the 'concatenate' function present in 'numpy/_core/multiarray.py'.

## Motivation
It may be a bit tricky to understand what the 'out' parameter does.

### DIFF
--- a/numpy/_core/multiarray.py
+++ b/numpy/_core/multiarray.py
@@ -256,7 +256,12 @@ def concatenate(arrays, axis=None, out=None, *, dtype=None, casting=None):
     >>> import numpy as np
     >>> a = np.array([[1, 2], [3, 4]])
     >>> b = np.array([[5, 6]])
-    >>> np.concatenate((a, b), axis=0)
+    >>> c = np.array([[7, 8], [9, 10], [11, 12]])
+    >>> np.concatenate((a, b), axis=0, out=c)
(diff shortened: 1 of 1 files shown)


antareepsarkar:
Why did the test fail? I only edited a comment.

mattip:
Thanks for the added example, but I am not sure I understand the problem. What is not clear with the description of `out` 
> If provided, the destination to place the result. The shape must be correct, matching that of what concatenate would have returned if no out argument were specified.

CI sometimes is flaky, in this case a temporary network failure prevented downloading a package.

antareepsarkar:
> Thanks for the added example, but I am not sure I understand the problem. What is not clear with the description of `out`
> 
> > If provided, the destination to place the result. The shape must be correct, matching that of what concatenate would have returned if no out argument were specified.
> 
> CI sometimes is flaky, in this case a temporary network failure prevented downloading a package.

Thanks Sir, for your insight. I just thought that it would be beneficial if an example with an argument for `out` is also given. At first, I thought that the `out` array's value becomes the concatenated value added with the `out` array's previous value. But, then I understood that it reassigns the value.
Actually, I am quite new to open source, so please proceed as you think.

PR #30143, opened 2025-11-04
BUG, TYP: ``_core.multiarray.*`` function runtime signatures

jorenham:
This is yet another batch of `inspect.signature` fixes, similar to #30104, #30114, #30121, #30124, #30126, #30137, #30138, and #30140. The following functions are affected:

- `numpy.empty_like`
- `numpy.concatenate`
- `numpy.inner`
- `numpy.where`
- `numpy.lexsort`
- `numpy.can_cast`
- `numpy.min_scalar_type`
- `numpy.result_type`
- `numpy.dot`
- `numpy.vdot`
- `numpy.bincount`
- `numpy.ravel_multi_index`
- `numpy.unravel_index`
- `numpy.copyto,`
- `numpy.putmask`
- `numpy.packbits`
- `numpy.unpackbits`
- `numpy.shares_memory`
- `numpy.may_share_memory`
- `numpy.is_busday`
- `numpy.busday_offset`
- `numpy.busday_count`
- `numpy.datetime_as_string`

Issue #10167, opened 2017-12-06
add_newdocs should happen before PyType_Ready


mattip:
Change ``add_newdocs`` to e.g. generate a header file at compile time with the docs (as is done in [scipy.special](https://github.com/scipy/scipy/blob/master/scipy/special/add_newdocs.py)), rather than patching them at runtime after ``PyType_Ready``. Requires some hunting for the correct places to specify the docstrings. The current way ``tp_doc`` is hacked (after ``PyTypeReady``) not only leaks the ``char*`` string yanked out of the ``PyStringObject``, it does not work on PyPy.

gleaned from #10157, plagarized @pv 's hints

eric-wieser:
The problem with generating the header file at compile time is that fixing a typo in one docstring now causes that header file to change, and now every single C file which includes that header has to be recompiled.

For testing whether a sphinx construct renders correctly, this is a huge burden.

Perhaps we need to generate a `.h` file for each function? So `/doc_header/concatenate.h` would contain a single `#define` of `PyArray_docstring_concatenate`, or similar?

njsmith:
Would it be terrible to move the actual docstrings into `.c` files in the first place, like most C API functions work, and then we wouldn't need to generate anything at all?

IIUC the only reason they're in a `.py` file in the first place is that building numpy used to be much harder, and they were split out so contributors could edit the docstrings without having to go through a build cycle. This is *much* less of an issue these days.

pv:
Or, you could autogen the C files. Writing it in python files avoids dealing with C quoting and the "\n" everywhere...

njsmith:
I guess I don't care that much -- it's not like we have any shortage of autogen code already :-)

If we do this, then instead of generating `.h` files, I think we can generate a `.c` file, and link to it. So `docstring.c` would export a `char*` called `Pyarray_concatenate_docstring` or whatever. This would avoid the rebuild-the-world issue.

mattip:
For one idea how to use a function to generate docstrings in header files, see the code [in pnumpy](https://github.com/Quansight/numpy-threading-extensions/blob/main/_add_newdocs.py)

ngoldbaum:
We added `__dict__` to the ufunc object in #26233, so now we don't really need to this C-level dance at all anymore. We can just set `__doc__` at runtime when we set up numpy.

PR #30146, opened 2025-11-04
BUG: ``np.dtype`` and ``np.dtypes.*`` runtime signatures (C)

jorenham:
For this one I had to go out of my confort zone, into the scary lands of C. So be sure to check if I didn't mess up anywhere. I've checked a whole nbunch of related signatures in IPython to ensure that everything still works as it should, and the tests are pretty thorough, but C tends to strictly follow Murphy's law for some reason, so I guess you never  know :shrug:

Anyway, this fixes the `inspect.signature` errors when called on `np.dtype`, `np.dtype.newbyteorder`, and any of the `numpy.dtypes` subclasses of `dtype` (including `StringDType`)

Related to #30104, #30114, #30121, #30124, #30126, #30137, #30138, #30140, and #30143


charris:
LGTM after a quick scan, but let's ping @seberg .

numpy/_core/src/multiarray/dtypemeta.c
```diff
@@ -1175,6 +1175,41 @@ dtypemeta_wrap_legacy_descriptor(
...
+    /* Build text signature and prepend to docstring */
+    switch (descr->type_num) {
+        case NPY_STRING:
+        case NPY_UNICODE:
+        case NPY_VOID:
+            signature = "(size, /)";  // bytes_, str_, void_
+            break;
+        case NPY_DATETIME:
+        case NPY_TIMEDELTA:
+            signature = "(unit, /)";  // datetime64, timedelta64
```
seberg, line 1193: We _should_ support it, but it's actually not I think?!

numpy/_core/src/multiarray/dtypemeta.c
```diff
@@ -1175,6 +1175,41 @@ dtypemeta_wrap_legacy_descriptor(
...
+            + 6;  // == len("\n--\n\n") + null terminator
+    char *new_doc = PyMem_Malloc(new_doc_size);
+    if (new_doc == NULL) {
+        Py_DECREF(dtype_class);
+        PyErr_NoMemory();
+        return NULL;
+    }
+    snprintf(new_doc, new_doc_size, "%s%s\n--\n\n%s",
+             short_name, signature, base_doc);
+    ((PyTypeObject *)dtype_class)->tp_doc = new_doc;
```
seberg, line 1211: This is a bit awkward for user dtypes like `rational`, since they have no good name, they still just use `numpy.dtype[name]`.  Although, I suppose it's technically memory safe at least :).

I am wondering if we should just not set any `tp_doc` here and then just use `add_newdoc` from Python (it doesn't like when docs are set).
At least for the builtin ones (user dtypes are a bit too bad, but also probably OK to just not have docs, heh.).

But I don't mind it too much, since we leak the reference everywhere you could just use `PyUnicode_Format("%s%s\n--\n\n", ...)` and then leak it to avoid the annoying size claculation.

numpy/_core/src/multiarray/dtypemeta.c
```diff
@@ -1175,6 +1175,41 @@ dtypemeta_wrap_legacy_descriptor(
         Py_DECREF(dtype_class);
         return NULL;
     }
+
+    /* Set the text signature for inspect.signature() */
+    const char *base_doc = ((PyTypeObject *)dtype_class)->tp_doc;
```
seberg, line 1180: Might as well move the base doc from above here, since we set `tp_doc` later anyway I think.

seberg (commented):
Yeah, should be fine if a bit awkward, but then this whole construct is awkward, so...
We could move it to Python, but unless we want to add type specfic docs it doesn't seem super worthwhile, so I don't have a strong opinion.

numpy/_core/src/multiarray/dtypemeta.c
jorenham: Yea that's true. In the stubs it's annotated in such a way that it'll be rejected. But I figured that since it'll always error right now, we might as well prepare for when it'll actually be supported while we're at it.
But I don't mind removing this if you think that's better.

numpy/_core/src/multiarray/dtypemeta.c
jorenham: > This is a bit awkward for user dtypes like `rational`, since they have no good name, they still just use `numpy.dtype[name]`.

Hmm I'm not sure I follow

> I am wondering if we should just not set any `tp_doc` here and then just use `add_newdoc` from Python (it doesn't like when docs are set).

The `add_newdoc` seems a bit hacky to me, and I can imagine that there's also a bit of overhead involved at import time. Tbh I put in in C instead of Python because that's where the docs are currently set at. But I don't mind moving it to Python

> But I don't mind it too much, since we leak the reference everywhere you could just use `PyUnicode_Format("%s%s\n--\n\n", ...)` and then leak it to avoid the annoying size claculation.

Eh not sure what you mean with that :sweat_smile:

jorenham:
I think I'll create an alternative PR for this in pure Python, so we can see which one we like better.

numpy/_core/src/multiarray/dtypemeta.c
seberg: Check what `rational` is, it's called `numpy.dtype[rational]` not `numpy.dtypes.<SomeName>`.  It's kinda fine because there are exactly those two schemes, but the naming scheme is not guaranteed here.

> The add_newdoc seems a bit hacky to me,

Oh, it is hacky, but it is a hack we use a lot and it's there for a reason (it's a pain to do large docstrings in C).  So I feel we use it for everything anyway, so there is no point in avoiding it.

But... this matters only if we have real docs, if we add just this, then it's not too useful.

> Eh not sure what you mean with that

Was thinking to just replace the `snprintf` and custom malloc by creating a Python string and then fetching its data.  But OK.

numpy/_core/src/multiarray/dtypemeta.c
seberg: Hmmm, yeah, mostly noticed it, I think I might lean slightly to keep it correct and not include it for now, but if you prefer this, then also good.

mattip:
There is #10167 to refactor the hacky `add_newdocs` across all of NumPy.

jorenham:
closing in favor of #30169

PR #30147, opened 2025-11-04
BUG, TYP: add the remaining ``_core.multiarray`` function runtime signatures

jorenham:
This fixes `inspect.signature` from raising a `ValueError` for the following functions:

- `numpy.arange`
- `numpy.char.compare_chararrays`
- `numpy.busdaycalendar`
- `numpy.datetime_data`
- `numpy.from_dlpack`
- `numpy.frombuffer`
- `numpy.fromfile`
- `numpy.fromiter`
- `numpy.frompyfunc`
- `numpy.fromstring`
- `numpy.nested_iters`
- `numpy.promote_types`

The tests were getting a bit messy, so I cleaned them up a bit, and grouped the `signature`-related onces into a class-bsaed test.

~I'll try and improve the `arange` signature in a follow-up, as that'll require quite a few changes in the stubs.~ 

Related to #30104, #30114, #30121, #30124, #30126, #30137, #30138, #30140, #30143, and #30146.

FYI, still on my `inspect.signature` to-do list are the classes in `numpy.random`, `ndarray.to_device` (which is missing docs), and 53 of the `np.generic` methods.


jorenham:
Hmm apparently the array api testsuite doesn't understand that `*args` and `**kwargs` are (always) compatible with Liskov's substitution principle. I guess I'll have to put the planned `arange` changes in this PR then.

jorenham:
Ah woops, `step` can also be `None` apparently.

jorenham:
In order to be compatible with the Array API spec for `arange`, the first `start` parameter had to be made positional-only. I've tried making it positional-or-keyword-only, but I now think that's not possible to do in a type-safe manner. 

This won't cause any backwards-incompatibily issues at runtime, as you can still call `arange(start=..., stop=...)` just fine. However, static type-checkers will no longer accept this. But then again, I don't see any way around it.

But as mypy_primer shows, calling `arange` this way isn't as uncommon as I'd have hoped. ~So I think I'll include a release note explaining why this change was needed. That'll have to wait until tomorrow though.~

jorenham:
re: the primer diff (https://github.com/numpy/numpy/pull/30147#issuecomment-3488507358):

The errors for Jax and Colour look correct to me; `arange` does not support `ndarray` input, so mypy rightly rejects it.

The xarray errors are a consequence of the `start` parameter no longer being usable as keyword-argument, as I've explained in https://github.com/numpy/numpy/pull/30147#issuecomment-3488515449. Luckily, it's very easy to fix these errors; i.e. pass `start` positionally instead of by keyword.

charris:
Thanks Joren.

rgommers:
The blast radius for this PR, and the whole "add more signatures" effort in general, seems quite large (the `mypy_primer` output and incoming links show JAX, AstroPy, SciPy, scikit-image and Colour from just this PR) so this seems to be in need of a prominent entry in the release notes at least.

jorenham:
> The blast radius for this PR, and the whole "add more signatures" effort in general, seems quite large 

So far, the reactions have been mostly positive. But the impact that this seems to have is  pretty surprising to me.

> (the `mypy_primer` output and incoming links show JAX, AstroPy, SciPy, scikit-image and Colour from just this PR) 

The mypy_primer output looks pretty scary at first glance, but I'm pretty sure that the new mypy_errors that this will introduce for jax and colour, are justified; i.e. that they're either true positives. And in case of xarray, related to the `arange(start=)` change. See https://github.com/numpy/numpy/pull/30147#issuecomment-3488529512 for some details. 

If you consider the reported line numbers  in the mypy_primer diff, you'll see that we're talking about 5 lines of code here; so most what you're seeing is just irrelevant noise :)

> so this seems to be in need of a prominent entry in the release notes at least.

That's also what I was thinking, yea. I wasn't sure yet what the best way to go about that was. Open a separate PR for the release note (highlight?) and link that PR, or use the PR number of one these actual PR's?

jorenham:
> so this seems to be in need of a prominent entry in the release notes at least.

#30208 :)

PR #30149, opened 2025-11-05
MAINT: Not show signature in git_version

cielavenir:
When $HOME/.gitconfig has `log.showSignature=true`, git log is like this

```
(main) $ git log -1 --format="%H %aI"
gpg: Signature made Wed 05 Nov 2025 07:57:32 AM JST
gpg:                using RSA key B5690EEEBB952194
gpg: Can't check signature: No public key
7aa023f008969a43648aaf4c3e707349abc7a2f7 2025-11-04T23:57:32+01:00
```

In this way, git_version() is prevented. We should pass -c log.showSignature=false explicitly to hide the signature

### DIFF
--- a/numpy/_build_utils/gitversion.py
+++ b/numpy/_build_utils/gitversion.py
@@ -28,7 +28,7 @@ def git_version(version):
     git_hash = ''
     try:
         p = subprocess.Popen(
-            ['git', 'log', '-1', '--format="%H %aI"'],
+            ['git', '-c', 'log.showSignature=false', 'log', '-1', '--format="%H %aI"'],
             stdout=subprocess.PIPE,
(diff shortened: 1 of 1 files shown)


mattip (approved):
Makes sense, even though 99.9% of the commiters here do not use `commit -s` to add a signature. This may show up in local builds of NumPy forks (i.e. behind corporte firewalls). I wonder if there are other flags that can mess with the output.

seberg:
Yeah, a bit weird, but agree that there is no reason to just put this in, thanks!

cielavenir:
@mattip Thank you for checking. The background is that we build numpy from local git to use intel-mkl backend, and one of our collegues is using log.showSignature.

PR #30150, opened 2025-11-05
ENH: add binary exponentiation in ``polynomial.polyutils._pow``

antareepsarkar:
## Description 
This adds binary exponentiation to the `_pow` function in 'numpy/polynomial/polyutils.py'

## Motivation
To optimize exponentiation

### DIFF
--- a/numpy/polynomial/polyutils.py
+++ b/numpy/polynomial/polyutils.py
@@ -694,11 +694,12 @@ def _pow(mul_f, c, pow, maxpower):
     elif power == 1:
         return c
     else:
-        # This can be made more efficient by using powers of two
-        # in the usual way.
-        prd = c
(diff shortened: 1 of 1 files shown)


antareepsarkar:
I think the checks failed due to decimal value related issues i.e. rounding off, etc.

mattip:
Once tests are passing, this needs benchmarks to prove it is actually an improvement

antareepsarkar:
> Once tests are passing, this needs benchmarks to prove it is actually an improvement

Probably, the tests are failing due to how decimal values are handled by Python(e.g.: 0.33333333333333333*3 is 1.0 in Python).
The expected values and the actual values in the errors are different at the decimal places because the order in which the terms are being multiplied can matter in the case of decimal values.

Please look into this.

PR #30153, opened 2025-11-05
BUG: decref on error in PyArray_NewFromDescr (#30152)

mdavidsaver:
Attempt to address #30152 by reordering the error tests.  Check `descr == NULL` first, then add a `Py_DECREF(descr);` on error to the following test.


mattip:
Hmm. I don't remember seeing the macOS-arm64-Accelerate `RuntimeWarning: invalid value encountered in matmul` in `test_linalg.py::TestQR::test_stacked_inputs[float32-outer_size2-size1]` before. It is not connected to this PR.

Edit: rerunning cleared it.

mattip:
Thanks @mdavidsaver

PR #30160, opened 2025-11-06
TYP: fix an invalid default value for `array`'s `ndmax` parameter

### LINKED ISSUE #30159: TYP: <Please write a comprehensive title after the 'TYP: ' prefix>
### Describe the issue:

Hello from downstream, @jorenham !

first of: we also have a couple inspect-based tests in astropy intended to catch discrepancies between numpy signatures and our NEP 18 dispatch functions. Functions which didn't have signatures available at runtime were simply ignored, but your recent work has allowed finding many more issues with our NEP 18 implementations, so thanks a lot for that !

I'm seeing a small issue with gh-29569 though; `None` doesn't appear to be a valid value for `array`'s `ndmax` argument at runtime (an integer is expected). Instead, and based on my understanding of https://github.com/numpy/numpy/pull/29569, it seems the default value should actually be `0`.
It can be worked around in astropy alright, but it feels slightly off that I need to both re-expose the default value as `None` (granted, this part is imposed by *our* tests) *and* include logic in the implementation to handle this special case. It's also a bit awkward that the parameter is typed as `int` but uses `None` as a default.

```python
# a script will not cut it
```

### Error message:

```shell

```

Python 3.11.18 (irrelevant)
numpy 2.4.0dev (d8e1c8ce6f0bd7c2dda825f88eb61e02525829ed)

not actually detected at typecheck-time

_No response_

neutrinoceros:
fix gh-30159

### DIFF
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -863,13 +863,13 @@ add_newdoc('numpy._core.multiarray', 'array',
         order='K',
         subok=False,
         ndmin=0,
-        ndmax=None,
+        ndmax=0,
         like=None,
     )
     --
 
     array(object, dtype=None, *, copy=True, order='K', subok=False, ndmin=0,
-          ndmax=None, like=None)
+          ndmax=0, like=None)
 
     Create an array.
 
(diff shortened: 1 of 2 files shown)


neutrinoceros:
The failure I got on macOS seems unrelated and flaky; I was able to reproduce it locally but only after a handful of attempts.

neutrinoceros:
Although a different warning, I wonder if it might stem out of the same bug than gh-29820, which already as an in-PR fix.

```
FAILED numpy/linalg/tests/test_linalg.py::TestQR::test_stacked_inputs[float32-outer_size2-size1] - RuntimeWarning: invalid value encountered in matmul
```

Happy to report this independently if needed.

mattip:
Ahh, I saw the same `TestQR::test_stacked_inputs[float32-outer_size2-size1]` failure in another CI job, then rerunning it cleared the error. It seems we have not yet ironed out all the Accelerate problems. I wonder what changed lately that could affect this. We updated OpenBLAS but this is using Accelerate.

neutrinoceros:
Since I can reproduce locally I should be able to bisect this. Does it deserve its own issue ?

mattip:
Sure that would be great

neutrinoceros:
gh-30162 👀

jorenham (commented):
Thanks! It seems a bit weird at first, but the added explanation in the docstring makes sense.

Could you also update the default in `_core/multiarray.pyi?`

neutrinoceros:
you got it

jorenham (approved):

mattip:
Thanks @neutrinoceros

PR #30163, opened 2025-11-06
ENH: Add ``order`` parameter to ``np.ma.asanyarray``

jorenham:
As discussed at yesterday's community meeting, here's the follow-up of #30057.

I shamelessly plagiarised the docstring description from the `MaskedArray` constructor docs.

Note that unlike `numpy.asanyarray`, this one uses `A` instead of `C` as default for backwards compatibility.


charris:
Thanks Joren.

seberg:
> Note that unlike numpy.asanyarray, this one uses A instead of C as default for backwards compatibility.

Sorry, a bit random and more of a nit, but I think the default should be `K` or `None`?  Admittedly, that has practically the same meaning (the difference is when a copy is made `A` will make it contiguous one way or another, `K` will not).

jorenham:
> the default should be `K` or `None`?

Hmm, I guess I don't really understand the difference between 'C', 'K', and `None` then :thinking:

seberg:
`None` has no meaning, but should alias to whatever the default is (`K` in this case).  `K` is similar to `A`, but *if* a copy is made `A` will make the result C (or F) contiguous, while `K` will keep the order as it was (e.g. if you have a `.transpose(1, 0, 2)` doing that same transpose on the result of a `K` copy will give you a contiguous array, while an `A` copy would be contiguous without that transpose)

jorenham:
> `None` has no meaning, but should alias to whatever the default is (`K` in this case). `K` is similar to `A`, but _if_ a copy is made `A` will make the result C (or F) contiguous, while `K` will keep the order as it was (e.g. if you have a `.transpose(1, 0, 2)` doing that same transpose on the result of a `K` copy will give you a contiguous array, while an `A` copy would be contiguous without that transpose)

So if we change this to `K` as default, then we wouldn't need those conditional checks I added here anymore?

seberg:
No? For those checks (no-copy) `A` and `K` is the same.  I guess masked array creation always returns a view, but also, it probably makes sense to keep this fast-path...

jorenham:
> it probably makes sense to keep this fast-path

what's that?

seberg:
> those conditional checks

You were talking about the refined check for when to just return the input unmodified, no?  That is the fast-path.

jorenham:
> That is the fast-path.

Ah ok; I just wasn't familiar with the term

PR #30166, opened 2025-11-06
CI: add check for numpy-release version of scipy-openblas

rgommers:
This ensures that the `scipy-openblas32|64` version pinned in the `numpy-release` repo isn't falling behind the one in this repo. Implemented as discussed in https://github.com/numpy/numpy-release/pull/21 and https://github.com/numpy/numpy-release/issues/6.

### DIFF
--- a/tools/check_openblas_version.py
+++ b/tools/check_openblas_version.py
@@ -1,20 +1,107 @@
 """
-usage: check_openblas_version.py <min_version>
+Checks related to the OpenBLAS version used in CI.
 
-Check the blas version is blas from scipy-openblas and is higher than
-min_version
-example: check_openblas_version.py 0.3.26
+Options:
+1. Check that the BLAS used at build time is (a) scipy-openblas, and (b) its version is
+   higher than a given minimum version. Note: this method only seems to give
+   the first 3 version components, so 0.3.30.0.7 gets translated to 0.3.30 when reading
+   it back out from `scipy.show_config()`.
+2. Check requirements files in the main numpy repo and compare with the numpy-release
+   repo. Goal is to ensure that `numpy-release` is not behind.
+
+Both of these checks are primarily useful in a CI job.
+
+Examples:
+
+    # Requires install numpy
+    $ python check_openblas_version.py --min-version 0.3.30
+
+    # Only needs the requirements files
+    $ python check_openblas_version.py --req-files \
+            ../numpy-release/requirements/openblas_requirements.txt
 """
 
+import argparse
+import os.path
 import pprint
-import sys
 
-import numpy
 
-version = sys.argv[1]
-deps = numpy.show_config('dicts')['Build Dependencies']
-assert "blas" in deps
-print("Build Dependencies: blas")
-pprint.pprint(deps["blas"])
-assert deps["blas"]["version"].split(".") >= version.split(".")
-assert deps["blas"]["name"] == "scipy-openblas"
+def check_built_version(min_version):
+    import numpy
+    deps = numpy.show_config('dicts')['Build Dependencies']
+    assert "blas" in deps
+    print("Build Dependencies: blas")
+    pprint.pprint(deps["blas"])
+    assert deps["blas"]["version"].split(".") >= min_version.split(".")
+    assert deps["blas"]["name"] == "scipy-openblas"
+
+
+def check_requirements_files(reqfile):
+    if not os.path.exists(reqfile):
+        print(f"Path does not exist: {reqfile}")
+
+    def get_version(line):
+        req = line.split(";")[0].split("==")[1].split(".")[:5]
+        return tuple(int(s) for s in req)
+
+    def parse_reqs(reqfile):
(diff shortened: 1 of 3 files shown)


tools/check_openblas_version.py
```diff
@@ -1,20 +1,103 @@
...
+
+    this_dir = os.path.abspath(os.path.dirname(__file__))
+    reqfile_thisrepo = os.path.join(this_dir, '..', 'requirements',
+                                    'ci_requirements.txt')
+
+    v32_thisrepo, v64_thisrepo = parse_reqs(reqfile_thisrepo)
+    v32_rel, v64_rel = parse_reqs(reqfile)
+
+    def compare_versions(v_rel, v_thisrepo, bits):
+        if not v64_rel >= v64_thisrepo:
```
mattip:
```suggestion
        if not v_rel >= v_thisrepo:
```

tools/check_openblas_version.py
```diff
@@ -1,20 +1,103 @@
...
+            lines = f.readlines()
+
+        v32 = None
+        v64 = None
+        for line in lines:
+            if "scipy-openblas32" in line:
+                v32 = get_version(line)
+            if "scipy-openblas64" in line:
+                v64 = get_version(line)
+        return v32, v64
```
mattip, line 62: Might be nice to check and raise a nicer error if `v32` or `v64` are None. On the other hand, the NumPy repo has both (the unfortunately named) `ci_requirements.txt` and `ci32_requirements.txt`. Shouldn't check_requirements check both?

mattip (commented):
Nice. A few comments on the new function. At some point we should rename `ci_requirements.txt` and `ci32_requirements.txt` and remove non-openblas content, could be in a separate PR.

tools/check_openblas_version.py
rgommers: Oops, good catch

tools/check_openblas_version.py
rgommers: I considered handling / special-casing the 32-bit only file, but it didn't seem necessary - it already has a very explicit comment "keep in sync with `ci_requirements.txt`" and I didn't see that ever going wrong, so that's not the type of thing we typically also check in CI.



> Might be nice to check and raise a nicer error if `v32` or `v64` are None.

Agreed, that could, will update.

tools/check_openblas_version.py
```diff
@@ -1,20 +1,103 @@
...
-Check the blas version is blas from scipy-openblas and is higher than
-min_version
-example: check_openblas_version.py 0.3.26
+Options:
+1. Check that the BLAS used at build time is (a) scipy-openblas, and (b) its version is
+   higher than a given minimum version. Note: this method only seems to give
+   the first 3 version components, so 0.3.30.0.7 gets translated to 0.3.30 when reading
+   it back out from `scipy.show_config()`.
+2. Check requirements files in the main numpy repo and compare with the numpy-release
+   repo. Goal is to ensure that `numpy-release` is not behind.
```
andyfaff, line 10: Would it be better for the check to be done in `numpy/numpy-release` instead. That triggers less then the jobs here. It could be done as a smoke test off the bat, on which all other jobs depend.

rgommers: I don't think so, for two reasons:
1. It doesn't give any signal before CI actually fails on unrelated PRs, because updates go to this repo first. 
2. In this repo the check is quite cheap to run once; in the other repo we don't have any suitable jobs that run just once, we only have sdist and wheel build jobs.

tools/check_openblas_version.py
rgommers: New error is:
```
AssertionError: Expected `scipy-openblas32` and `scipy-openblas64` in `ci_requirements.txt`, got:
  spin==0.15
  # Keep this in sync with ci_requirements.txt
  scipy-openblas32==0.3.30.0.7
```

mattip: Hmm. I added the `ci32_requirements.txt` file specifically to ~reject~ not install `scipy-openblas64` on 32-bit systems, since they cannot use the ILP64 interfaces in `scipy-openblas64`. Not sure how to handle that here. Maybe a flag for 32/64? Maybe allow None and check later on that None is expected?

mattip: Wait, when I run this with the `ci_requirements.txt` file, everything is OK. Are you running it with `ci32_requirements.txt`? I thought we were not going to do that (comment above)?

rgommers:
CircleCI docs failure is unrelated.

tools/check_openblas_version.py
rgommers: Yes indeed, we weren't going to do that - I ran it once here just to demonstrate the error message.

Everything is okay here; if you'd delete one of the two entries in `ci_requirements.txt` then you will see this error as well.

mattip: Ahh i was confused

mattip:
Thanks @rgommers

PR #30168, opened 2025-11-07
TYP, DEP: ``numpy.fix`` pending deprecation

jorenham:
ref: https://mail.python.org/archives/list/numpy-discussion@python.org/thread/LYRHETXPN32S7FNHKNLJFDIYWGQHLODJ/

related: #30098


jorenham:
I'm not sure if I should add a release note for this though, since it's only a *pending* deprecation :thinking:

ngoldbaum:
> I'm not sure if I should add a release note for this though, since it's only a pending deprecation 🤔

Worth doing IMO

charris:
Thanks Joren.

PR #30164, opened 2025-11-06
BUG: ``numpy.random.*`` class runtime signatures

jorenham:
This fixes `inspect.signature` for

- `np.random.BitGenerator`
- `np.random.Generator`
- `np.random.MT19937`
- `np.random.PCG64`
- `np.random.PCG64DXSM`
- `np.random.Philox`
- `np.random.RandomState`
- `np.random.SFC64`
- `np.random.SeedSequence`
- `np.random.bit_generator.SeedlessSeedSequence`

This also fixes a typo in `bit_generator.pxd` that accidentally defined an empty unused class `np.random.bit_generator.SeedlessSequence`.

Related to #30104, #30114, #30121, #30124, #30126, #30137, #30138, #30140, #30143, #30146, #30147, and #30155

PR #30155, opened 2025-11-05
BUG, DOC: ``ndarray`` dunder method runtime signatures and missing docs

jorenham:
This fixes `inspect.signature` for 

- `ndarray`
- `ndarray.__array_function__`
- `ndarray.__array_ufunc__`
- `ndarray.__complex__`
- `ndarray.__format__`
- `ndarray.__sizeof__`
- `ndarray.__reduce_ex__`
- `ndarray.__array_namespace__`
- `ndarray.__dlpack__`
- `ndarray.__dlpack_device__`
- `ndarray.to_device`

Docstrings *were* also missing for these methods.

Related to #30104, #30114, #30121, #30124, #30126, #30137, #30138, #30140, #30143, #30146, and #30147.

PR #30169, opened 2025-11-07
BUG: ``np.dtype`` and ``np.dtypes.*`` runtime signatures

jorenham:
Alternative to #30146 

Ref: https://github.com/numpy/numpy/pull/30146#issuecomment-3499948505

Related to #30104, #30114, #30121, #30124, #30126, #30137, #30138, #30140, #30143, #30146, #30147, #30155, and #30164

---

Note that for reason the signature docstring header bits aren't stripped from `__doc__` as it should. The result is that the [`dtype` reference docs](https://output.circle-artifacts.com/output/job/affede45-6a2c-495d-995e-b8374e7089f9/artifacts/0/doc/build/html/reference/generated/numpy.dtype.html) now contains a weird `--` in its description.
In the recent triage meeting we decided to take this for granted for now, as the benefits of this PR outweigh the costs.

### DIFF
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -7425,17 +7425,26 @@ for _dtype_name, _signature, _sctype_name in (
     ("DateTime64DType", "(unit, /)", "datetime64"),
     ("TimeDelta64DType", "(unit, /)", "timedelta64"),
 ):
+    _extra_docs = ""
+    if _dtype_name in {"VoidDType", "DateTime64DType", "TimeDelta64DType"}:
+        _extra_docs = f"""
+        .. warning::
+            ``np.dtypes.{_dtype_name}`` cannot be instantiated directly.
+            Use ``np.dtype("{_sctype_name}[{{unit}}]")`` instead.
+        """
+
     add_newdoc('numpy.dtypes', _dtype_name,
         f"""
         {_dtype_name}{_signature}
         --
 
         DType class corresponding to the `numpy.{_sctype_name}` scalar type.
-
-        Please see `numpy.dtype` for the typical way to create dtype instances and
-        :ref:`arrays.dtypes` for additional information.
+        {_extra_docs}
+        See `numpy.dtype` for the typical way to create dtype instances
+        and :ref:`arrays.dtypes` for additional information.
         """)
-    del _dtype_name, _signature, _sctype_name  # avoid namespace pollution
+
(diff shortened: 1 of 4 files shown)


numpy/_core/_add_newdocs.py
```diff
@@ -6192,6 +6192,9 @@
 
 add_newdoc('numpy._core.multiarray', 'dtype',
     """
+    dtype(dtype, align=False, copy=False, **kwargs)
```
jorenham, line 6195: I tried using `metadata={}` before, but passing `metadata={}` isn't quite the same as omitting it.

It also looked a bit off in the docs; not sure why though.
image

jorenham: But I don't really like how sphinx now uses the `__text_signature__` instead of the the one below it with `, [metadata])`.

seberg: > I tried using metadata={} before, but passing metadata={} isn't quite the same as omitting it.

Hmmm, that is silly, we could (and should) allow `metadata=None` and translate that to NULL.
But yeah, that is slightly independent from this and fine.

jorenham: The issue seems to be that the `{signature}\n--\n\n` bit isn't stripped off for the dtype classes, like it happens for all other classes like `broadcast` and `flatiter`. My guess is that it might have something to do with the metaclass, but I have no idea how to get around that.

seberg: Ah, fun... The difference may also be that this is a heaptype (i.e. more like a Python defined class).  That means we could in principle add a `__signature__` probably as well.
That is different in that all other classes are static types.

jorenham: Hmm I see. Yesterday I tried adding a `__text_signature__` getset descriptor to the dtype meta in C, but sseemed to also overrule the text signatures from the `np.dtypes.*` classes, and I wasn't able to figure out how to get around that. Any ideas?

jorenham: As discussed in the recent triage meeting; let's just ignore this documentation glitch for now. Those stray `--` aren't the end of the world

mattip:
Thanks @jorenham. I assume there is a release note for all the runtime signature improvements already.

jorenham:
> I assume there is a release note for all the runtime signature improvements already.

not yet, but there will be, yea

jorenham:
> Thanks @jorenham. I assume there is a release note for all the runtime signature improvements already.

#30208 :)

PR #30174, opened 2025-11-07
TYP: Reject ``NDArray[datetime64] / NDArray[Any]``

### LINKED ISSUE #30173: TYP: numpy allows a datetime array to be divided by another array
### Describe the issue:

If a numpy array has a datetime type, and is divided by another array, typecheckers are not picking that up as wrong.

```python
from typing import reveal_type
import numpy as np

s = np.array([np.datetime64(f"2025-10-{d:02d}") for d in (23, 24, 25)], np.datetime64)
t = np.array([1, 2, 3])
reveal_type(s)
reveal_type(t)
reveal_type(s.__truediv__(t))

s / t
```

### Error message:

```shell
The type revealed of the division operation is `Any`
```

python 3.12
numpy 2.3.4

pyright 1.1.407
mypy 1.18.2

Note:  This came up with `pandas-stubs`.  

We think the solution is to add an overload that returns `Never` .

jorenham:
Closes #30173


jorenham:
cc @Dr-Irv

Dr-Irv:
Regarding the diffs on `pandas-stubs`, we have a PR that I'm about to merge that will probably affect the results here, and we'll have to merge that in until a new numpy comes out, and when that happens, we can fix our stubs to handle this right.

jorenham:
The new colour error originates from https://github.com/colour-science/colour/blob/ad9a3f0ac70c4445e839685816c0511256a564e0/colour/models/igpgtg.py#L116-L122, where an `ArrayLike` is divided by a float array. Because of this change, and because `ArrayLike` *could* be a `datetime64` array, type-checkers will now reject this.

I can imagine that that this is a pretty common usage pattern, i.e. dividing an `ArrayLike` by an array. So although is technically more type-safe, I'm not sure whether this is actually worth it.

Dr-Irv:
> I can imagine that that this is a pretty common usage pattern, i.e. dividing an `ArrayLike` by an array. So although is technically more type-safe, I'm not sure whether this is actually worth it.

We've been dealing with this in pandas-stubs.  We've created overloads that say that when we don't know the dtype, we allow the typechecker to say the computation is OK.  But if we DO know the dtype and we want to have the type checker pick it up, then we have the result be `Never` .  

So maybe an overload such as:
```python
 def __truediv__(self: NDArray[datetime64], other: Any) -> Never: ...
```
would be a way of catching this.

jorenham:
> > I can imagine that that this is a pretty common usage pattern, i.e. dividing an `ArrayLike` by an array. So although is technically more type-safe, I'm not sure whether this is actually worth it.
> 
> We've been dealing with this in pandas-stubs. We've created overloads that say that when we don't know the dtype, we allow the typechecker to say the computation is OK. But if we DO know the dtype and we want to have the type checker pick it up, then we have the result be `Never` .
> 
> So maybe an overload such as:
> 
> ```python
>  def __truediv__(self: NDArray[datetime64], other: Any) -> Never: ...
> ```
> 
> would be a way of catching this.

There are one or two places where it's still being used in the NumPy codebase, but the plan is to get rid of those. Because it turns out that it's not technically valid, as it leads to incompatible overlap, a.k.a. undefined type-checker behavior. So even though it might currently work, there's the risk that it'll break after a mypy or pyright update.

And because this (mis)use of `Never`/`NoReturn` is apparently a common thing, there recently was a proposal for adding a decorator that could be used to reject specific overloads: https://discuss.python.org/t/specifying-invalid-signatures-in-overloads/75405, but I'm guessing that you already knew about that.

Dr-Irv:
> There are one or two places where it's still being used in the NumPy codebase, but the plan is to get rid of those. Because it turns out that it's not technically valid, as it leads to incompatible overlap, a.k.a. undefined type-checker behavior. So even though it might currently work, there's the risk that it'll break after a mypy or pyright update.

Sure, but our philosophy in `pandas-stubs` is that "it works now, and we'll fix it later  if/when it breaks".  

Both `mypy` and `pyright` are doing their CI tests with pandas-stubs, so if a change will break what we're doing, we'll hear about it  before they release it and discuss with them how they suggest we should resolve it.

> 
> And because this (mis)use of `Never`/`NoReturn` is apparently a common thing, there recently was a proposal for adding a decorator that could be used to reject specific overloads: https://discuss.python.org/t/specifying-invalid-signatures-in-overloads/75405, but I'm guessing that you already knew about that.

Yes, you can see I was involved in that discussion, which seems to have ended up with no resolution.

jorenham:
> Sure, but our philosophy in `pandas-stubs` is that "it works now, and we'll fix it later if/when it breaks".

I apply that same philosophy to scipy-stubs. But for NumPy, we have to be more cautious.

On several occasions type-checkers held off on implementing a fix because it would have caused problems for users of numpy (and therefore downstream libraries, including pandas). Then there's also NumPy's release schedule to consider, and the multi-year support window ([SPEC 0](https://scientific-python.org/specs/spec-0000/)). So that's why I'm very cautious when it comes to changes like these.

> Yes, you can see I was involved in that discussion, which seems to have ended up with no resolution.

Hm yea, unfortunately that's what usually happens there...

Dr-Irv:
> On several occasions type-checkers held off on implementing a fix because it would have caused problems for users of numpy (and therefore downstream libraries, including pandas). Then there's also NumPy's release schedule to consider, and the multi-year support window ([SPEC 0](https://scientific-python.org/specs/spec-0000/)). So that's why I'm very cautious when it comes to changes like these.
> 
> >

Understood.  Avoiding a support policy with the stubs is another reason why having the separation of the project code from the stubs code helps.  With `pandas-stubs`, we don't support old versions. The latest version of the stubs supports the latest release of pandas, and that's it.  We don't have the resources to support multiple versions of pandas with the stubs.  We also can do releases of the stubs more frequently.

Has `numpy` considered having a separate `numpy-stubs` project?

jorenham:
> > On several occasions type-checkers held off on implementing a fix because it would have caused problems for users of numpy (and therefore downstream libraries, including pandas). Then there's also NumPy's release schedule to consider, and the multi-year support window ([SPEC 0](https://scientific-python.org/specs/spec-0000/)). So that's why I'm very cautious when it comes to changes like these.
> > >
> 
> Understood. Avoiding a support policy with the stubs is another reason why having the separation of the project code from the stubs code helps. With `pandas-stubs`, we don't support old versions. The latest version of the stubs supports the latest release of pandas, and that's it. We don't have the resources to support multiple versions of pandas with the stubs. We also can do releases of the stubs more frequently.

I fully agree. 

> Has `numpy` considered having a separate `numpy-stubs` project?

Yes: https://github.com/numpy/numtype

Dr-Irv:
> > Has `numpy` considered having a separate `numpy-stubs` project?
> 
> Yes: https://github.com/numpy/numtype

If you started "supporting" that via a pip install, and it resolved problems like this one, I'd have no issues in saying that `pandas-stubs` requires `numpy-stubs` , presuming we could get it all to work.

jorenham:
> > > Has `numpy` considered having a separate `numpy-stubs` project?
> > 
> > 
> > Yes: [numpy/numtype](https://github.com/numpy/numtype?rgh-link-date=2025-11-08T16%3A18%3A16Z)
> 
> If you started "supporting" that via a pip install, and it resolved problems like this one, I'd have no issues in saying that `pandas-stubs` requires `numpy-stubs` , presuming we could get it all to work.

Good to hear :)

Yea that's the plan, and I hope to be able to do that before the end of the year. But at the moment there's just not enough time in my schedule, and there are still a couple of things that need to happen before releasing it.

jorenham:
Anyway, I'll be closing this then, and I hope that it doesn't inconvenience pandas-stubs too much. Thanks for the report and your time :)

PR #30175, opened 2025-11-07
ENH: Updates for the `spin bench` command.

WarrenWeckesser:
* Use `sys.executable` to run a Python script in the bench function.
* Expose `--factor` as a command line option for `spin bench`.

I found it useful to tweak the `--factor` option of `asv` in a recent pull request.

### DIFF
--- a/.spin/cmds.py
+++ b/.spin/cmds.py
@@ -454,7 +454,7 @@ def bench(ctx, tests, compare, verbose, quick, commits, build_dir):
         meson._set_pythonpath(build_dir)
 
         p = spin.util.run(
-            ['python', '-c', 'import numpy as np; print(np.__version__)'],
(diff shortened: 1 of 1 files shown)


charris:
Makes sense. Thanks @WarrenWeckesser .

mattip:
>  I found it useful to tweak the --factor option of asv in a recent pull request.

I wonder if we could use `--` to pass additional options on to asv?

WarrenWeckesser:
Possibly, but because our spin command uses a default factor of 1.05, and the default factor for the command `asv continuous` is 1.1, we would have to check whether or not the user provided the factor after the `--` separator, and if not, explicitly add the option `--factor 1.05`.  (Unless `asv` is happy to accept multiple `--factor` options and only use the last one.)

PR #30176, opened 2025-11-07
BUG: Fix check of PyMem_Calloc return value.

WarrenWeckesser:
Stumbled across this while reviewing our handling of `PyMem_` allocation functions.


charris:
Thanks Warren.

PR #30179, opened 2025-11-08
MAINT,API: Introduce __numpy_dtype__ and fix dtype attribute recursion

seberg:
This is a lot more complex then I expected. We, correctly, deprecated `.dtype` attribute lookup recursion, however...
The code still had a `try/except` that still recursed and that try except carried meaning at least in a weird path of:
```
class mydtype(np.void):
    dtype = ...
```

Now does that make sense? Maybe not... we also fall back to object in some paths which should have been broken when a dtype attribute existed on a type but it is a property/descriptor.

So, this removes the recursion, but adds a check for `__get__` to filter out those cases, this is something we successfully did for other protocols `__array__`, `__array_interface__`, etc.

### DIFF
--- a/doc/release/upcoming_changes/30179.new_feature.rst
+++ b/doc/release/upcoming_changes/30179.new_feature.rst
@@ -1,0 +1,13 @@
+New ``__numpy_dtype__`` protocol
+--------------------------------
+NumPy now has a new ``__numpy_dtype__`` protocol. NumPy will check
+for this attribute when converting to a NumPy dtype via ``np.dtype(obj)``
+or any ``dtype=`` argument.
+
+Downstream projects are encouraged to implement this for all dtype like
+objects which may previously have used a ``.dtype`` attribute that returned
+a NumPy dtype.
+For array-like objects we encourage you to implement ``__numpy_dtype__``
+with a warning or error to _prevent_ using e.g. ``dtype=dataframe`` in
+NumPy functions (it may be good to go via a Deprecation).
+We expect deprecating coercion via ``.dtype`` in the future.
--- a/doc/release/upcoming_changes/30179.new_feature.rst
+++ b/doc/release/upcoming_changes/30179.new_feature.rst
@@ -7,7 +7,7 @@ or any ``dtype=`` argument.
 Downstream projects are encouraged to implement this for all dtype like
 objects which may previously have used a ``.dtype`` attribute that returned
 a NumPy dtype.
-For array-like objects we encourage you to implement ``__numpy_dtype__``
-with a warning or error to _prevent_ using e.g. ``dtype=dataframe`` in
-NumPy functions (it may be good to go via a Deprecation).
-We expect deprecating coercion via ``.dtype`` in the future.
+We expect to deprecate ``.dtype`` in the future to prevent interpreting
+array-like objects with a ``.dtype`` attribute as a dtype.
+If you wish you can implement ``__numpy_dtype__`` to ensure an earlier
+warning or error (``.dtype`` is ignored if this is found).
--- a/numpy/_core/tests/test_dtype.py
+++ b/numpy/_core/tests/test_dtype.py
@@ -1580,19 +1580,19 @@ class TestFromDTypeAttribute:
         assert np.dtype(dt) == np.float64
         assert np.dtype(dt()) == np.float64
 
-    @pytest.mark.skipif(IS_PYSTON, reason="Pyston disables recursion checking")
-    @pytest.mark.skipif(IS_WASM, reason="Pyodide/WASM has limited stack size")
-    def test_recursion(self):
+    def test_recursive(self):
+        # This used to recurse. It now doesn't, we enforce the
+        # dtype attribute to be a dtype (and will not recurse).
         class dt:
             pass
 
         dt.dtype = dt
-        with pytest.raises(RecursionError):
+        with pytest.raises(ValueError):
             np.dtype(dt)
 
         dt_instance = dt()
         dt_instance.dtype = dt
-        with pytest.raises(RecursionError):
+        with pytest.raises(ValueError):
             np.dtype(dt_instance)
 
     def test_void_subtype(self):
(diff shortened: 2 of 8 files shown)


mattip:
Would adding a `__numpy_dtype__` attribute to `np.float64` make the detection logic simpler? The documentation mentions one must use `np.dtype(np.float64)`, maybe we could have `np.float64.__numpy_dytpe__` instead

seberg:
Yeah, that is a nice thought, we should maybe add it to deprecate the use of the dtype on the instance.
I think there are two things that make me not jump at it. First, we need type lookup for some things anyway (i.e. I would prefer not to require this for user DTypes; that said such user dtypes don'tsupport the conversion right nowI suspect, that should be fixed).
Second, it does need a careful property dance, since the class attribute needs to give a dtype, while the instance attribute should give an error. That said... while I sm not sure that the Python API makes that simple, it alsodoesn't make it hard.

I guess that makesme lean to: worth an issue/followup, but maybe not needed here?

doc/release/upcoming_changes/30179.new_feature.rst
```diff
@@ -0,0 +1,13 @@
...
+NumPy now has a new ``__numpy_dtype__`` protocol. NumPy will check
+for this attribute when converting to a NumPy dtype via ``np.dtype(obj)``
+or any ``dtype=`` argument.
+
+Downstream projects are encouraged to implement this for all dtype like
+objects which may previously have used a ``.dtype`` attribute that returned
+a NumPy dtype.
+For array-like objects we encourage you to implement ``__numpy_dtype__``
+with a warning or error to _prevent_ using e.g. ``dtype=dataframe`` in
+NumPy functions (it may be good to go via a Deprecation).
```
timhoffm: Do we really need other array libraries to take action? If they don't do anything, the status-quo for these types of array-likes is maintained (works or doesn't work depending on the case).

A warning / deprecation / error  can be handled centrally in numpy. You could even make this behavior dependent on whether the passed object is array-like and/or whether they return a numpy dtype.

seberg: Let me rephrase it to say that you _could_ do this. NumPy would indeed deprecate it for you as soon as we actually deprecate `.dtype`.
I agree with that deprecation, but think we may want to wait a little bit, implementing it in the downstream library would be an early opt-in effectively.

inakleinbottle (commented):
I had a look through the changes to the C code. I can't see any obvious issues in the C code.

numpy/_core/tests/test_dtype.py
```diff
@@ -1580,19 +1580,19 @@ class dt:
...
 
         dt.dtype = dt
-        with pytest.raises(RecursionError):
+        with pytest.raises(ValueError):
             np.dtype(dt)
 
         dt_instance = dt()
         dt_instance.dtype = dt
-        with pytest.raises(RecursionError):
+        with pytest.raises(ValueError):
```
ngoldbaum, line 1595: Does this behavior change deserve a release note?

ngoldbaum (approved):
LGTM, just one small comment about whether the changing error type deserves a release note

numpy/_core/tests/test_dtype.py
seberg: :shrug: I don't think recursion errors are a type of error that anyonen should ever run into really.

charris:
Thanks Sebastian.

jorenham:
I guess I'll do the typing stuff for this tomorrow then (in 12 hours or so). I assume that branching 2.4 is still planned for his weekend?

jorenham:
Maybe we could add a `__numpy_dtype__` to `dtype` itself, i.e.e.g. `dtype(int8).__numpy_dtype__ == dtype(int8)`?

And how about also adding it to the scalar-type *classes*, e.g.i.e. `int8.__numpy_dtype__ == dtype(int8)`?

That way we'd be able to simplify `numpy._typing._SupportsDType` and `numpy.typing.DTypeLike`, from

```py
# numpy._typing
class _SupportsDType[DTypeT: np.dtype = np.dtype](Protocol):
    @property
    def dtype(self) -> DTypeT: ...

type _VoidDTypeLike = # <omitted spaghetti>
type _DTypeLike[ScalarT: np.generic] = type[ScalarT] | np.dtype[ScalarT] | _SupportsDType[np.dtype[ScalarT]]

# numpy.typing
type DTypeLike =  np.dtype | type | str | _SupportsDType | _VoidDTypeLike
```

to just

```py
# numpy._typing
class _SupportsDType[DTypeT: np.dtype = np.dtype](Protocol):
    @property
    def __numpy_dtype__(self) -> DTypeT: ...

type _VoidDTypeLike = # <omitted spaghetti>
type _DTypeLike[ScalarT: np.generic] = _SupportsDType[np.dtype[ScalarT]]

# numpy.typing
type DTypeLike = type | str | _DTypeLike | _VoidDTypeLike
```

(this uses Python 3.13+ PEP 696 syntax for aesthetics)

Typing aside, I suppose it'd also be nice if we'd "practice what we preach", analogous to e.g. `ndarray.__array__`, `int.__int__`, `str.__str__`, etc.

seberg:
> And how about also adding it to the scalar-type classes

Possible, but is it worth it? You need to make a funky dance with a custom descriptor stuffed into the `__dict__` again, because you must error when accessing it from the class...

I do also do like to (in principle!) allow tying scalar types to dtypes that do not allow adding a custom attribute like this (becuase they are defined by a 3rd party).  I.e. that still needs something like the `sctypeDict` "registration".

EDIT: Although, I guess typing wise the last thing might not matter too much, in the sense that I am not sure user dtypes are supported typing wise to begin with...

charris:
> I assume that branching 2.4 is still planned for his weekend?

Only if we are ready.

jorenham:
> I.e. that still needs something like the `sctypeDict` "registration".
> 
> EDIT: Although, I guess typing wise the last thing might not matter too much, in the sense that I am not sure user dtypes are supported typing wise to begin with...

Yea, that's right. It'd be similar to that `ABCMeta.register` stuff, that type-checkers also aren't able to deal with. I'm not even sure if there exists a type theory (yes; there's not just one "type-theory") that would allow for such a thing, considering the (arguably) most important static typing axiom, "types don't change" (at runtime), that all type theories share (as far as I'm aware of).

The approach that we currently use to statically determine a scalar type's associated dtype, via its `.dtype` attribute (with a `typing.Protocol`). But since it's an instance attribute, and the `dtype` constructor only accepts scalar classes, that can be pretty awkward to deal with.

jorenham:
> Possible, but is it worth it? You need to make a funky dance with a custom descriptor stuffed into the `__dict__` again, because you must error when accessing it from the class...

Can't we use `_DTypeMeta` for that?

seberg:
You are confusing scalar types and dtypes.

jorenham:
> You are confusing scalar types and dtypes.

Ah yea, you're right; classic mistake :P

Issue #30177, opened 2025-11-08
Unchecked memory allocations


WarrenWeckesser:
In almost all cases, if a memory allocation in our C & C++ code fails, the failure is handled so that eventually a `MemoryError` exception is raised by Python. I think that is a general rule that we apply to all low-level code: failure to allocate memory should ultimately result in a Python exception and not crash the interpreter.

Here are a few places where we don't check for a memory allocation failure:

- [ ] https://github.com/numpy/numpy/blob/f39abd4ffbca44f01eec353115ffd2d7ad296fb2/numpy/_core/src/multiarray/dtype_transfer.c#L171-L173
- [ ] https://github.com/numpy/numpy/blob/f39abd4ffbca44f01eec353115ffd2d7ad296fb2/numpy/_core/src/multiarray/stringdtype/casts.cpp#L856-L861
- [ ] https://github.com/numpy/numpy/blob/f39abd4ffbca44f01eec353115ffd2d7ad296fb2/numpy/_core/src/multiarray/stringdtype/casts.cpp#L1117-L1118
- [ ] https://github.com/numpy/numpy/blob/f39abd4ffbca44f01eec353115ffd2d7ad296fb2/numpy/_core/src/umath/stringdtype_ufuncs.cpp#L1121
- [ ] https://github.com/numpy/numpy/blob/f39abd4ffbca44f01eec353115ffd2d7ad296fb2/numpy/_core/src/umath/stringdtype_ufuncs.cpp#L1243
- [ ] https://github.com/numpy/numpy/blob/f39abd4ffbca44f01eec353115ffd2d7ad296fb2/numpy/_core/src/umath/stringdtype_ufuncs.cpp#L1458
- [ ] https://github.com/numpy/numpy/blob/f39abd4ffbca44f01eec353115ffd2d7ad296fb2/numpy/_core/src/umath/stringdtype_ufuncs.cpp#L1577

The following uses of `std::unordered_set` in `unique.cpp` can throw an exception if an allocation fails, which will crash Python:

- [x] https://github.com/numpy/numpy/blob/f39abd4ffbca44f01eec353115ffd2d7ad296fb2/numpy/_core/src/multiarray/unique.cpp#L212-L214
- [x] https://github.com/numpy/numpy/blob/f39abd4ffbca44f01eec353115ffd2d7ad296fb2/numpy/_core/src/multiarray/unique.cpp#L220

(This was not an exhaustive search.)

seberg:
[...]

PR #30180, opened 2025-11-08
MAINT: Take advantage of C++ for more robust use of the PyGILState_* API.

WarrenWeckesser:
In C++ code, instead of writing

    NPY_ALLOW_C_API_DEF
    NPY_ALLOW_C_API
    [code that uses the Python C API here]
    NPY_DISABLE_C_API

we can write

    {
        EnsureGIL ensure_gil{};
        [code that uses the Python C API here]
    }

This ensures that `PyGILState_Release(__save__)` is called,  even if the wrapped code throws an exception or executes a return or a goto. The latter currently occurs in a few places in our code.

### DIFF
--- a/numpy/_core/include/numpy/npy_common.h
+++ b/numpy/_core/include/numpy/npy_common.h
@@ -387,6 +387,39 @@ typedef struct
     long double _Val[2];
 } npy_clongdouble;
 
+//
+// In C++ code, use this at the beginning of a scope, e.g.
+//
+//     {
+//         EnsureGIL ensure_gil{};
+//         [code that uses the Python C API here]
+//     }
+//
+// instead of
+//
+//     NPY_ALLOW_C_API_DEF
+//     NPY_ALLOW_C_API
+//     [code that uses the Python C API here]
+//     NPY_DISABLE_C_API
+//
+// This ensures that PyGILState_Release(__save__) is called,  even if the
+// wrapped code throws an exception or executes a return or a goto.
+//
+class EnsureGIL
+{
+    PyGILState_STATE __save__;
+
+public:
+
+    EnsureGIL() {
+        __save__ = PyGILState_Ensure();
+    }
+
+    ~EnsureGIL() {
+        PyGILState_Release(__save__);
+    }
+};
+
 #else
 
 #include <complex.h>
--- a/numpy/_core/src/multiarray/stringdtype/casts.cpp
+++ b/numpy/_core/src/multiarray/stringdtype/casts.cpp
@@ -1933,8 +1933,7 @@ string_to_bytes(PyArrayMethod_Context *context, char *const data[],
 
         for (size_t i=0; i<s.size; i++) {
             if (((unsigned char *)s.buf)[i] > 127) {
-                NPY_ALLOW_C_API_DEF;
-                NPY_ALLOW_C_API;
+                EnsureGIL ensure_gil{};
                 PyObject *str = PyUnicode_FromStringAndSize(s.buf, s.size);
 
                 if (str == NULL) {
@@ -1962,7 +1961,6 @@ string_to_bytes(PyArrayMethod_Context *context, char *const data[],
                 PyErr_SetObject(PyExceptionInstance_Class(exc), exc);
                 Py_DECREF(exc);
                 Py_DECREF(str);
-                NPY_DISABLE_C_API;
                 goto fail;
             }
         }
--- a/numpy/_core/src/multiarray/unique.cpp
+++ b/numpy/_core/src/multiarray/unique.cpp
@@ -186,11 +186,12 @@ unique_numeric(PyArrayObject *self, npy_bool equal_nan)
     * Returns a new NumPy array containing the unique values of the input array of numeric (integer or complex).
     * This function uses hashing to identify uniqueness efficiently.
     */
-    NPY_ALLOW_C_API_DEF;
-    NPY_ALLOW_C_API;
-    PyArray_Descr *descr = PyArray_DESCR(self);
-    Py_INCREF(descr);
-    NPY_DISABLE_C_API;
+    PyArray_Descr *descr;
+    {
+        EnsureGIL ensure_gil{};
+        descr = PyArray_DESCR(self);
+        Py_INCREF(descr);
+    }
 
     PyThreadState *_save1 = PyEval_SaveThread();
 
@@ -223,23 +224,25 @@ unique_numeric(PyArrayObject *self, npy_bool equal_nan)
     npy_intp length = hashset.size();
 
     PyEval_RestoreThread(_save1);
-    NPY_ALLOW_C_API;
-    PyObject *res_obj = PyArray_NewFromDescr(
-        &PyArray_Type,
-        descr,
-        1, // ndim
-        &length, // shape
-        NULL, // strides
-        NULL, // data
-        // This flag is needed to be able to call .sort on it.
-        NPY_ARRAY_WRITEABLE, // flags
-        NULL // obj
-    );
+    PyObject *res_obj;
+    {
+        EnsureGIL ensure_gil{};
+        res_obj = PyArray_NewFromDescr(
+            &PyArray_Type,
+            descr,
+            1, // ndim
+            &length, // shape
+            NULL, // strides
+            NULL, // data
+            // This flag is needed to be able to call .sort on it.
+            NPY_ARRAY_WRITEABLE, // flags
+            NULL // obj
+        );
 
-    if (res_obj == NULL) {
-        return NULL;
+        if (res_obj == NULL) {
+            return NULL;
+        }
     }
-    NPY_DISABLE_C_API;
     PyThreadState *_save2 = PyEval_SaveThread();
     auto save2_dealloc = finally([&]() {
         PyEval_RestoreThread(_save2);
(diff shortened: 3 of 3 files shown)


WarrenWeckesser:
If we decide to add this (or something like it), we'll also need to decide if we will make it part of <strike>the</strike> a public C++ API, and document like we do the [C macros](https://numpy.org/doc/stable/reference/c-api/array.html#group-2).

numpy/_core/src/multiarray/stringdtype/casts.cpp
```diff
@@ -1933,8 +1933,7 @@ string_to_bytes(PyArrayMethod_Context *context, char *const data[],
 
         for (size_t i=0; i<s.size; i++) {
             if (((unsigned char *)s.buf)[i] > 127) {
-                NPY_ALLOW_C_API_DEF;
-                NPY_ALLOW_C_API;
+                EnsureGIL ensure_gil{};
```
jorenham, line 1936: Won't the `goto`'s be a problem here?

WarrenWeckesser: No, the destructor is called when the scope is exited, even with a goto.  Here's a self-contained demo:

`main.cpp`

```
#include <cstdio>

class EnsureGIL
{
    long __save__;

public:

    EnsureGIL() {
        printf("EnsureGIL()\n");
        __save__ = 1;
    }

    ~EnsureGIL() {
        printf("~EnsureGIL()\n");
        __save__ = 0;
    }
};


int main()
{
    {
        EnsureGIL ensure_gil{};
        printf("AAAAA\n");
        goto skip_ahead;
        printf("BBBBB\n");
    }

skip_ahead:
    printf("Done.\n");

    return 0;
}
```

In the output, you can see that the destructor is called:

```
% g++ -std=c++17 main.cpp -o main
% ./main
EnsureGIL()
AAAAA
~EnsureGIL()
Done.
```

jorenham: Thanks; TIL :)

numpy/_core/include/numpy/npy_common.h
```diff
@@ -387,6 +387,39 @@ typedef struct
...
+
+    EnsureGIL() {
+        __save__ = PyGILState_Ensure();
+    }
+
+    ~EnsureGIL() {
+        PyGILState_Release(__save__);
+    }
+};
+
```
mattip, line 422: > we'll also need to decide if we will make it part of the public C++ API

I see you decided to do so. Is any of this documented in the RST documentation?

mattip (commented):
Nice. Do we actually have any code that will throw a C++ exception? If so that should definitely use this construct and not the `NPY_ALLOW_C_API_DEF` macros

rgommers:
I don't think we have a C++ API? We have a C API that can be used from C++, I don't think we have any public API elements that are C++ only.

The insertion of this `EnsureGIL` code is in the wrong place even if we do decide to make it public - you inserted it in the middle of the code defining `npy_cdouble` & co which happened to need an `#if defined(__cplusplus)` for struct types.

Creating a new header-only C++ API with functionality that's C++-only seems like a large decision. I'd lean to no, but if we'd go that way than it would at a minimum need a separate header file.

WarrenWeckesser:
> The insertion of this EnsureGIL code is in the wrong place ...

🤦 Yeah, I'll move it. To address the comments about "public C++ API", I'll put it someplace private.  Whether or not to establish a public C++ API, and where to put it if we do (and how to name things, including a possible namespace) can be decided later.

seberg:
Yeah, I agree that this should just not be public, that GIL stuff feels like not our job and easily stolen. I would not mind C++ header only API, but I don't really feel like doing it for Python things.
(It may make sense to start such a thing outside NumPy. I have thought that it could be nice to allow re-use of NumPy patterns e.g. to define sorts/ufuncs for user DTypes.)

I am also a bit confused, while I think this makes sense, I think we need a RAII pattern for a `nogil` block rather than one for ensure GIL (which is usually more confined to error handling anyway and we have a helper for that)?

EDIT: OK, I see the current code seems a bit messy/unclear, which is why ensure GIL is the more obvious stary.

numpy/_core/src/multiarray/unique.cpp
```diff
@@ -263,11 +266,12 @@ unique_string(PyArrayObject *self, npy_bool equal_nan)
...
-    NPY_ALLOW_C_API;
-    PyArray_Descr *descr = PyArray_DESCR(self);
-    Py_INCREF(descr);
-    NPY_DISABLE_C_API;
+    PyArray_Descr *descr;
+    {
+        EnsureGIL ensure_gil{};
+        descr = PyArray_DESCR(self);
+        Py_INCREF(descr);
+    }
```
seberg, line 274: You probably already on it, but the thing that rubs me wrong about this, is that all of these places shouldn't exist to begin with, I think.

Replacing the finally action with a full-blown RAII pattern will be an improvement! But the thing is that for example this block is useless and because of this obfuscating what is simple into something that seems difficult, IMO.

`unique_*` are called with the GIL held and defensively ensuring we hold it just obfuscates the code, one could assert that if we must, but that would be the most
.
Now, there is the annoyance that we release the GIL twice within the function, but that is still far better done via:
```
// code with gil
{
    no_gil release via RAII
    // code without GIL
}
// code with gil
{
    no_gil release via RAII
}
// code with gil
```
Somewhere along the line, what was simple enough restoring either got messed up and then apparently these defensive GIL ensuring things were added.
As a rule of thumb though, defensively grabbing the GIL only makes sense for setting an error, and we even have `npy_gil_error()` for that.

numpy/_core/include/numpy/npy_common.h
WarrenWeckesser: This will be changed so the new code is not public.

numpy/_core/src/multiarray/unique.cpp
WarrenWeckesser: In my next update, I will have a (very conventional) RAII class that wraps `PyEval_SaveThread()` and `PyEval_RestoreThread(thread_state)` using the same constructor/destructor pattern as `EnsureGIL`, which I think is what you are suggesting above.

I agree that if we are sure the various `unique` functions are always called with the GIL (which is reasonable--`unique` is not a ufunc, after all), then the current use of `NPY_ALLOW_C_API`, etc, in there is unnecessary.  I'll remove that in the next update.

WarrenWeckesser: My updates in `unique.cpp` are becoming nontrivial: in addition to removing the use of the macros related to ensuring the GIL (`NPY_ALLOW_C_API`, etc) and adding the use of RAII objects for `PyEval_SaveThread/_RestoreThread` management, there is some incidental DRY clean up and encapsulation that works out nicely.  That's not a problem, but these changes will cause big conflicts with @math-hiyoko's recent PR to use a hash map to implement the various `return_` options of `unique` (https://github.com/numpy/numpy/pull/30110).  I haven't looked at that PR closely yet, but if it looks good enough to eventually merge, we should get that in before I pursue my PR here.  I suspect that at least some of what I am thinking of changing could be incorporated into #30110.

math-hiyoko: Thank you for the consideration. 
#30110 is still under active development and there are several parts that I plan to revise, so waiting for it to be completed might delay this PR.

I would prefer to have this PR merged first, and then update #30110 to incorporate the changes from it.

WarrenWeckesser: OK, thanks @math-hiyoko. I'll continue working on this, and keep an eye on #30110.

WarrenWeckesser:
I have a couple notes/questions about the hash-based `unique`:

* Currently there is code that implements the hash-based `unique` for complex types, but not floats.  Unless there is some objection (or technical obstacle that I don't see yet), I'll add the handling of floats. Like complex, they have to account for the equality of -0.0 and 0.0 when hashing.
* Similarly, `datetime64` is handled, but not `timedelta64`. I'll also add `timedelta64`.
* The handling of `datetime64` currently ignores the `equal_nan` option, even though `NaT` exists. Should that be added?  For example, should there be *two* occurrences of `NaT` in the output shown in `Out [204]` here?

  ```
  In [203]: dt
  Out[203]: array([       'NaT',        'NaT', '1984-01-01'], dtype='datetime64[D]')

  In [204]: np.unique(dt, equal_nan=False)
  Out[204]: array(['1984-01-01',        'NaT'], dtype='datetime64[D]')
  ```

seberg:
I thought we had float before complex, I am a bit surprised about that?! Only having one of datetime/timedelta doesn't make sense of course.

If `equal_nan=True` things are great, since it's just an int64. I guess `equal_nan=False` is funky, because on the one hand, I doubt there is much of a use-case for it beyond supporting `equal_nan=True` is sometimes tricky! But, I guess it makes sense to support it even if it is extra work here (rather than the other way around).

math-hiyoko:
When implementing the hash-based unique for complex types, I also evaluated the float case.
However, the existing sort-based implementation for floats turned out to be faster, so We decided not to add a hash-based version for floats.
https://github.com/numpy/numpy/pull/29537#issuecomment-3263829324

WarrenWeckesser:
I pulled the RAII-related changes that I've been using locally for this PR and put them in a self-contained PR here: https://github.com/numpy/numpy/pull/30312.  I'll use those in the updates to the hash-based `unique` that I am working on here to fix the issue reported in https://github.com/numpy/numpy/issues/30177.

WarrenWeckesser:
This PR is superseded by https://github.com/numpy/numpy/pull/30329.  It was simpler to start fresh than to undo the obsolete changes in this PR.

PR #30181, opened 2025-11-09
DEP: Deprecate ndarray.resize

eendebakpt:
Addresses part of https://github.com/numpy/numpy/issues/28800

We deprecate `ndarray.resize` in the type hints. In a followup PR (probably for a next numpy release) we will issue a deprecating working when `ndarray.resize` is called.


jorenham (approved):
Got nothing to complain :ok_hand:

mhvk (approved):
Yes, makes sense!

doc/release/upcoming_changes/30181.deprecation.rst
```diff
@@ -0,0 +1,6 @@
+Resizing a Numpy array inplace is deprecated
+-----------------------------------------------
```
jorenham:
```suggestion
Resizing a Numpy array inplace is deprecated
--------------------------------------------
```

doc/release/upcoming_changes/30181.deprecation.rst
```diff
@@ -0,0 +1,6 @@
...
+Resizing a Numpy array inplace is deprecated since mutating
+an array is unsafe if an array is shared, especially by multiple
+threads.  As an alternative, you can create a resized array via `np.resize`.
```
jorenham:
```suggestion
threads.  As an alternative, you can create a resized array via ``np.resize``.
```

seberg:
Will we make type-hint only deprecations a pattern now? FWIW, it seems nice it seems like a nice way to potentially get some feedback about impact (and I guess some libraries may clean it up before there even is a warning, which is a small bonus too)!

I do have a slight worry with this one that while it won't be much code that is affected, it may not always be obvious how to replace `resize`. Typically, one can `concate`, sometimes use `fromiter` occasionally it may just be a `reshape` anyway.

The least obvious extreme case might be that the resize actually does make sense and one might have to use `data = bytearray(...); tmp_view = np.frombuffer(data)` (to actually keep reallocating).

mhvk:
Internally, we seem to use it mostly in places where we're reading from a file, and the array should be extended. I think this would often involve a copy (not sure how likely it is memory right after an array allocated earlier is still available). In any python code at least, this is fairly trivially done with `concatenate`.

mhvk (approved):
Yes, this seems a good first step!

numpy/__init__.pyi
```diff
@@ -2248,8 +2248,10 @@ class ndarray(_ArrayOrScalarCommon, Generic[_ShapeT_co, _DTypeT_co]):
     def tolist(self, /) -> Any: ...
 
     @overload
+    @deprecated("Resizing a NumPy array inplace has been deprecated in NumPy 2.4")
     def resize(self, new_shape: _ShapeLike, /, *, refcheck: py_bool = True) -> None: ...
     @overload
+    @deprecated("Resizing a NumPy array inplace has been deprecated in NumPy 2.4")
```
tylerjereddy, line 2254: Should these have been `2.5`?

jorenham: Yup...

eendebakpt: I will make a prnto correct this tonight

mhvk: Oops, sorry I missed that in review...

eendebakpt: PR is up at https://github.com/numpy/numpy/pull/30569

PR #30182, opened 2025-11-09
TST: Add tests for savez and savez_compressed reproducibility

### LINKED ISSUE #9439: `savez()` is not idempotent
The results of `savez()` depend on the current time, because the underlying implementation uses `tempdir` to cache result about to be written.  From there the internal instance of `ZipFile()` auto-detects the time/date and encodes them into the zipfile headers somewhere.  

Here is a simple demo:
```python
In [1]: import numpy as np

In [2]: import io

In [3]: bio1 = io.BytesIO()

In [4]: bio2 = io.BytesIO()

In [5]: np.savez_compressed(bio1, foo="bar")

In [6]: import time

In [7]: time.sleep(1)

In [8]: np.savez_compressed(bio2, foo="bar")

In [9]: bio1.getvalue()
Out[9]: b'PK\x03\x04\x14\x00\x00\x00\x08\x00\x84s\xf2J\xd6i\x8cTK\x00\x00\x00\\\x00\x00\x00\x07\x00\x00\x00foo.npy\x9b\xec\x17\xea\x1b\x10\xc9\xc8\xe0\xc6P\xad\x9e\x92Z\x9c\\\xa4n\xa5\xa0n\x13j\xac\xae\xa3\xa0\x9e\x96_TR\x94\x98\x17\x9f_\x94\x92\n\x12wK\xcc)N\x05\x8a\x17g$\x16\xa4\x02\xf9\x1a\x9a:\n\xb5\n(\x80+\x89\x81\x81!\x11\x88\x8b\x80\x18\x00PK\x01\x02\x14\x03\x14\x00\x00\x00\x08\x00\x84s\xf2J\xd6i\x8cTK\x00\x00\x00\\\x00\x00\x00\x07\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x80\x81\x00\x00\x00\x00foo.npyPK\x05\x06\x00\x00\x00\x00\x01\x00\x01\x005\x00\x00\x00p\x00\x00\x00\x00\x00'
                                               ^ HERE
In [10]: bio2.getvalue()
Out[10]: b'PK\x03\x04\x14\x00\x00\x00\x08\x00\x8cs\xf2J\xd6i\x8cTK\x00\x00\x00\\\x00\x00\x00\x07\x00\x00\x00foo.npy\x9b\xec\x17\xea\x1b\x10\xc9\xc8\xe0\xc6P\xad\x9e\x92Z\x9c\\\xa4n\xa5\xa0n\x13j\xac\xae\xa3\xa0\x9e\x96_TR\x94\x98\x17\x9f_\x94\x92\n\x12wK\xcc)N\x05\x8a\x17g$\x16\xa4\x02\xf9\x1a\x9a:\n\xb5\n(\x80+\x89\x81\x81!\x11\x88\x8b\x80\x18\x00PK\x01\x02\x14\x03\x14\x00\x00\x00\x08\x00\x8cs\xf2J\xd6i\x8cTK\x00\x00\x00\\\x00\x00\x00\x07\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x80\x81\x00\x00\x00\x00foo.npyPK\x05\x06\x00\x00\x00\x00\x01\x00\x01\x005\x00\x00\x00p\x00\x00\x00\x00\x00'
                                               ^ HERE
In [11]: bio1.getvalue() == bio2.getvalue()
Out[11]: False
```

A minimum of 1 second difference is necessary for byte-wise similarity to fail.

tovrstra:
Closes #9439

@meawoppl Can you check if this is ok for you?

### DIFF
--- a/numpy/lib/tests/test_io.py
+++ b/numpy/lib/tests/test_io.py
@@ -2855,3 +2855,19 @@ def test_savez_nopickle():
     with temppath(suffix='.npz') as tmp:
         with pytest.raises(ValueError, match="Object arrays cannot be saved when.*"):
             np.savez_compressed(tmp, obj_array, allow_pickle=False)
+
+
+@pytest.mark.parametrize("savez_func", [np.savez, np.savez_compressed])
+def test_savez_reproducible(savez_func):
+    # Verify that saving the same data twice results in the same file.
+    # See gh-9439
+    arr = np.array([3.14])
+    with BytesIO() as stream:
+        savez_func(stream, arr=arr)
+        blob1 = stream.getvalue()
+    # Note that time resolution in ZIP files is 1 second.
+    time.sleep(1.1)
+    with BytesIO() as stream:
+        savez_func(stream, arr=arr)
+        blob2 = stream.getvalue()
+    assert blob1 == blob2


tovrstra:
All failing tests seem to be unrelated to the proposed test. This is a representative error message:

```text
Attempt to get headers for http://downloads.sourceforge.net/project/pkgconfiglite/0.28-1/pkg-config-lite-0.28-1_bin-win32.zip failed.
  The remote file either doesn't exist, is unauthorized, or is forbidden for url 'http://downloads.sourceforge.net/project/pkgconfiglite/0.28-1/pkg-config-lite-0.28-1_bin-win32.zip'. Exception calling "GetResponse" with "0" argument(s): "The remote server returned an error: (403) Forbidden."
Downloading pkgconfiglite 64 bit
  from 'http://downloads.sourceforge.net/project/pkgconfiglite/0.28-1/pkg-config-lite-0.28-1_bin-win32.zip'
ERROR: The remote file either doesn't exist, is unauthorized, or is forbidden for url 'http://downloads.sourceforge.net/project/pkgconfiglite/0.28-1/pkg-config-lite-0.28-1_bin-win32.zip'. Exception calling "GetResponse" with "0" argument(s): "The remote server returned an error: (403) Forbidden."
The install of pkgconfiglite was NOT successful.
```

The download from sourceforge seems to cause problems.

charris:
Close/reopen

meawoppl:
This fixes the problems I reported...8 years ago?

NGL, I don't even remember the original context in which it percolated, so ... thank you!

tovrstra:
You're welcome. Better late than never, but note that it is just a test to verify deterministic behavior. The fix itself is way older.

I was looking into this because I'd be interested in making the NPZ files also platform-independent, which goes just a bit further. (One step at a time.)

meawoppl:
Great to hear!

tovrstra:
@charris This should be ready for review.

One point I'm not 100% sure about is the delay I've put into the test to ensure that two NPZ streams would have different time stamps (with the second resolution used in the ZIP format) if this information would not be stripped out by `savez()`. This makes the test inherently slow.

Another option is to reopen the ZIP file with Python's built-in `zipfile` module and to check the metadata for anything non-deterministic.

ADSilberman:
The test also waits only 1.1 seconds. Due to zip using MS-DOS formatting, the timestamp only changes every two seconds (speaking of poor documentation, it's barely given a line, and doesn't specify the implication--that zip only saves in 2s increments) https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT.

PR #26574, opened 2024-05-30
CI: replace pkgconfig-lite from Chocolatey with `pip install pkgconf`

rgommers:
This makes the build env setup simpler and much faster. In GitHub Actions jobs, the pkg-config step for Windows goes from 50 seconds to 3 seconds, with cleaner logging. It should also make things more robust, because Chocolatey can be flaky (e.g., see gh-25980).

Also ensure that BLAS is found in the Windows CI jobs, don't allow a fallback to lapack-lite when we do install `scipy-openblas`.

[skip cirrus] [skip circle]

PR #30189, opened 2025-11-10
BLD: use pykg-config on windows instead of pkgconfiglite

mattip:
Fixes #30187 by using the pure-python pykg-config. A possible improvement would be to add pykg-config as a build requirement.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -62,7 +62,7 @@ jobs:
       - name: pkg-config-for-win
         if: runner.os == 'windows'
         run: |
-		  python3 -m pip install pykg-config
+          python3 -m pip install pykg-config
           $CIBW = "${{ github.workspace }}/.openblas"
           # pkgconfig needs a complete path, and not just "./openblas since the
           # build is run in a tmp dir (?)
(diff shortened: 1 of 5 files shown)


mattip:
CI is passing. I think it is fine for now to leave installing it as part of the CI environment, since we also assume there is a `pkg-config` available in posix builds. We don't explicitly mention how to install `pkg-config` in the [building](https://numpy.org/devdocs/building/index.html) page. We do mention it in passing as part of the ""conda env" tab.

Edit: actually, the wheel workflow failed but gave no indication of a problem :frowning_face:

charris:
The numpy-release builds use also use choco to install pkgconfiglite, I'm guessing cibuildwheel requires it.

mattip:
Bleh. The package does not work, it seems I was misled by a package installed with perl on the windows images. Closing

rgommers:
> A possible improvement would be to add pykg-config as a build requirement.

This is now moot, but: we shouldn't do that so easily, it's not a Python package and it's not needed for almost all build scenarios.

This PR reminded me of gh-26574, which we really should find some time to finish. Might need a Christmas holiday ....

PR #30190, opened 2025-11-10
MAINT: ``ma.asanyarray``: use ``order=None`` as default

jorenham:
ref: https://github.com/numpy/numpy/pull/30163#issuecomment-3510826797


seberg (approved):
Thanks for the follow-up!  This is indeed now matching main NumPy in behavior.

charris:
Thanks Joren.

Issue #29762, opened 2025-09-16
BUG: compilation fails when building in custom directory identical to source directory of pyf file


2sn:
### Describe the issue:

`numpy/f2py/_backends/_meson.py` has an uncaught exception when source and target are identical.

Please find my proposed fix (patch) below.

### Reproduce the code example:

```python
listing essential flags.

after a script creates the fortran source `_interface.f90` in `/home/alex/python/source/multistar/_library`, `f2py -h` creates `_interface.pyf` in same directory, after which I want to build in this directory s well.


python3.13 "-m" "numpy.f2py" "--verbose" "--build-dir" "/home/alex/python/source/multistar/_library" "-c" "-m" "_interface" "/home/alex/python/source/multistar/_library/library.a" "interface.f90" "/home/alex/python/source/multistar/_library/_interface.pyf"
```

### Error message:

```shell
File "/home/alex/Python/lib/python3.13/site-packages/numpy/f2py/_backends/_meson.py", line 188, in compile
    self.sources = _prepare_sources(self.modulename, self.sources, self.build_dir)
                   ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alex/Python/lib/python3.13/site-packages/numpy/f2py/_backends/_meson.py", line 200, in _prepare_sources
    shutil.copy(source, bdir)
    ~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/home/alex/Python/lib/python3.13/shutil.py", line 428, in copy
    copyfile(src, dst, follow_symlinks=follow_symlinks)
```

### Python and NumPy Versions:

2.3.2
3.13.7 (main, Sep  5 2025, 00:35:26) [GCC 15.2.1 20250808 (Red Hat 15.2.1-1)]

### Runtime Environment:

<details>

```
[{'numpy_version': '2.3.2',
  'python': '3.13.7 (main, Sep  5 2025, 00:35:26) [GCC 15.2.1 20250808 (Red '
            'Hat 15.2.1-1)]',
  'uname': uname_result(system='Linux', release='6.16.4-200.fc42.x86_64', version='#1 SMP PREEMPT_DYNAMIC Thu Aug 28 19:47:10 UTC 2025', machine='x86_64')},
 {'simd_extensions': {'baseline': ['SSE', 'SSE2', 'SSE3'],
                      'found': ['SSSE3',
                                'SSE41',
[...]

Issue #24874, opened 2023-10-07
BUG: f2py - Invalid command line options do not warn or error on Python 3.12


2sn:
### Describe the issue:

In python 3.12 there seems to be an issue with directory handling that is not present in 3.11.  In the very least it does behave differently.   It seems to be ignoring some path specifications as well and created files in directories where there were none before, ignoring path specifications.  Same numpy version, only difference is Python 3.12 instead of 3.11.6.

Likely related to `meson` backend, and that I use a runtime parameter combination that may not have been used/tested before/otherwise.

Maybe this can be fixed easily by someone familiar with the meson backend.
(or advise what I am doing wrong - that said, it does work with Python 3.11 and compatibility may be preferable)

### Reproduce the code example:

```python
my calls are

f2py3 --verbose -m _interface -h /home/alex/python/source/multistar/_build/_interface.pyf --f2cmap /home/alex/python/source/multistar/.f2py_f2cmap interface.f90 --overwrite-signature

then
```shell
f2py3 --verbose --build-dir /home/alex/python/source/multistar/_build --f90flags="-fPIC -fno-second-underscore -O3 -funroll-loops -fconvert=big-endian -I /home/alex/python/source/multistar/_library" --f77flags="-fPIC -fno-second-underscore -O3 -funroll-loops -fconvert=big-endian -I /home/alex/python/source/multistar/_library" --f2cmap /home/alex/python/source/multistar/.f2py_f2cmap -c -m _interface /home/alex/python/source/multistar/_library/library.a interface.f90 /home/alex/python/source/multistar/_build/_interface.pyf
```
```


### Error message:

```shell
f2py3 --verbose -m _interface -h /home/alex/python/source/multistar/_build/_interface.pyf --f2cmap /home/alex/python/source/multistar/.f2py_f2cmap interface.f90 --overwrite-signature
Reading f2cmap from '/home/alex/python/source/multistar/.f2py_f2cmap' ...
        Mapping "real(kind=real32)" to "float"
        Mapping "real(kind=real64)" to "double"
        Mapping "real(kind=real128)" to "long_double"
        Mapping "integer(kind=int8)" to "signed_char"
        Mapping "integer(kind=int16)" to "short"
        Mapping "integer(kind=int32)" to "int"
        Mapping "integer(kind=int64)" to "long"
        Mapping "integer(kind=int128)" to "long_long"
        Mapping "complex(kind=comp32)" to "complex_float"
[...]

PR #30192, opened 2025-11-10
TST: Add F2Py test of building in source directory.

DWesl:
Test for #29762 with fix suggested by #29762 and #24874

### DIFF
--- a/numpy/f2py/tests/test_modules.py
+++ b/numpy/f2py/tests/test_modules.py
@@ -24,6 +24,23 @@ class TestModuleFilterPublicEntities(util.F2PyTest):
 
 
 @pytest.mark.slow
+class TestBuildInSourceDirectory(util.F2PyTest):
+    sources = [
+        util.getpath(
+            "tests", "src", "modules", "gh26920",
+            "two_mods_with_one_public_routine.f90"
+        )
+    ]
+    # we filter the only public function mod2
+    only = ["mod1_func1", ]
+    options = ["--build-dir", util.get_module_dir()]
+
+    def test_gh26920(self):
+        # if it compiles and can be loaded, things are fine
+        pass
+
+
+@pytest.mark.slow
 class TestModuleWithoutPublicEntities(util.F2PyTest):
     sources = [
         util.getpath(
--- a/numpy/f2py/_backends/_meson.py
+++ b/numpy/f2py/_backends/_meson.py
@@ -204,7 +204,10 @@ def _prepare_sources(mname, sources, bdir):
     # Copy sources
     for source in sources:
         if Path(source).exists() and Path(source).is_file():
-            shutil.copy(source, bdir)
+            try:
+                shutil.copy(source, bdir)
+            except shutil.SameFileError:
+                pass
     generated_sources = [
         Path(f"{mname}module.c"),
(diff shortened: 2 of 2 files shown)


DWesl:
@2sn, applying your suggested fix in the obvious spots did not make the test pass.  Was there anything else you did to make this scenario work?

numpy/f2py/tests/test_modules.py
```diff
@@ -23,6 +23,23 @@ def test_gh26920(self):
...
+class TestBuildInSourceDirectory(util.F2PyTest):
+    sources = [
+        util.getpath(
+            "tests", "src", "modules", "gh26920",
+            "two_mods_with_one_public_routine.f90"
+        )
+    ]
+    # we filter the only public function mod2
+    only = ["mod1_func1", ]
+    options = ["--build-dir", util.get_module_dir()]
```
rgommers, line 36: `util.get_module_dir()` returns a fresh temporary directory:

https://github.com/numpy/numpy/blob/58b1ad7b608d8ccf3410cdb2694a6c2eb510c737/numpy/f2py/tests/util.py#L150-L157

How is that a test for building inside the source directory?

DWesl: `util.get_module_dir()` also caches that module directory and returns the same one until it's cleared.

[`util.build_module()`](https://github.com/numpy/numpy/blob/128d684bdc8d756c63341baab0fd7e4df3fbc09c/numpy/f2py/tests/util.py#L198) uses that function to [get a directory](https://github.com/numpy/numpy/blob/128d684bdc8d756c63341baab0fd7e4df3fbc09c/numpy/f2py/tests/util.py#L206) to [put the source files for a given test in](https://github.com/numpy/numpy/blob/128d684bdc8d756c63341baab0fd7e4df3fbc09c/numpy/f2py/tests/util.py#L217-L219): since `util.get_module_dir()` caches its result, the option will have f2py use as a build directory the same temporary directory it put the sources in.

I think T2PyTest cleanup will clear the temporary directory.

charris:
@DWesl Do you want to continue working on this?

DWesl:
There's not been any feedback from the author of the original issue, so I don't have any direction to work on this, leaving me inclined to leave this until a second person runs into this issue

2sn:
Seemed an obvious fix, I now worked my way around it.  As long as Numpy quality control is happy with broken code ...

DWesl:
> Seemed an obvious fix, I now worked my way around it. As long as Numpy quality control is happy with broken code ...

You haven't yet confirmed that I managed to reproduce your failure, let alone that I properly applied your suggested fix.

This, of course, leaves aside that the applied change doesn't actually fix the problem I found, which suggests you tried other things when trying to find a fix and one of those helped (this is part of why you were encouraged to create the PR yourself).

2sn:
@DWesl At the time, the changed seemed to have fixed it in my specific setup, but I have change my setup and the NumPy code be may have changed as well. 

I have to admit I no longer recall all the details, things work for now with my current usage of `f2py`.   Better to revisit if another use case with demonstration surfaces.

Nowadays AI can help a lot to work around things, and the next set of changes is on the horizon.

PR #30194, opened 2025-11-11
BUG: Change return type of sign(timedelta64) to float64.

### LINKED ISSUE #8463: sign for any positive timedelta should be/compare equal
The current behaviour of `sign` on `timedelta64` objects is somewhat surprising:    

```
In [64]: sign(np.timedelta64(3, 'm')) == sign(np.timedelta64(3, 's'))
Out[64]: False
```

I'm not sure if the return type for the sign of a `timedelta64` should even be a `timedelta64` or or simply numerical.  Of course, normally `sign(x)` returns a result of the same `dtype` as `x`, but for any numerical data we have `uint8(1) == int8(1) == float32(1) == ....` etc.  I'm not sure what the best return value would be for the sign of a `timedelta64`, but I think it would be desirable if their signs compared equal.

### LINKED ISSUE #29496: BUG: `sign(timedelta64('nat'))` returns -1; expected `timedelta64('nat')`
### Describe the issue:

`np.sign(np.timedelta64('nat'))` returns `timedelta64(-1)`.  For consistency with the behavior of the inexact types (e.g. `sign(nan)` returns `nan`), it should return `timedelta64('nat')`.

```python
In [10]: td = np.timedelta64('nat')

In [11]: np.sign(td)
Out[11]: np.timedelta64(-1)
```

### Error message:

```shell
n/a
```

2.4.0.dev0+git20250801.9da497f
3.13.5 (main, Jul 15 2025, 14:50:34) [GCC 11.4.0]

### Runtime Environment:

n/a

### Context for the issue:

_No response_

WarrenWeckesser:
With this PR, `sign(timedelta64)` returns a `float64`.  The ensures that the sign function has the basic property `x = np.sign(x)*np.abs(x)` when `x` has dtype `timedelta64`, including the case where `x` is `NaT`.

*The PR makes the change with no warnings and no deprecation.*

The issue was brought up on the mailing list: https://mail.python.org/archives/list/numpy-discussion@python.org/thread/7ELIJOAAZAMFIGNDAUMOCLXYBRYYKE2Z/

Closes gh-8463.
Closes gh-29496.

### DIFF
--- a/numpy/_core/src/umath/ufunc_type_resolution.c
+++ b/numpy/_core/src/umath/ufunc_type_resolution.c
@@ -445,6 +445,9 @@ PyUFunc_SignTypeResolver(PyUFuncObject *ufunc,
 {
     if (PyArray_DESCR(operands[0])->type_num == NPY_TIMEDELTA) {
         out_dtypes[0] = NPY_DT_CALL_ensure_canonical(PyArray_DESCR(operands[0]));
+        if (out_dtypes[0] == NULL) {
+            return -1;
+        }
         out_dtypes[1] = PyArray_DescrFromType(NPY_DOUBLE);
         return 0;
     }
@@ -609,6 +612,9 @@ PyUFunc_SimpleUniformOperationTypeResolver(
             descr = PyArray_DESCR(operands[0]);
         }
         out_dtypes[0] = NPY_DT_CALL_ensure_canonical(descr);
+        if (out_dtypes[0] == NULL) {
+            return -1;
+        }
     }
 
     /* All types are the same - copy the first one to the rest */
@@ -675,6 +681,9 @@ PyUFunc_IsNaTTypeResolver(PyUFuncObject *ufunc,
     }
 
     out_dtypes[0] = NPY_DT_CALL_ensure_canonical(PyArray_DESCR(operands[0]));
+    if (out_dtypes[0] == NULL) {
+        return -1;
+    }
     out_dtypes[1] = PyArray_DescrFromType(NPY_BOOL);
 
     return 0;
@@ -694,6 +703,9 @@ PyUFunc_IsFiniteTypeResolver(PyUFuncObject *ufunc,
     }
 
     out_dtypes[0] = NPY_DT_CALL_ensure_canonical(PyArray_DESCR(operands[0]));
+    if (out_dtypes[0] == NULL) {
+        return -1;
+    }
     out_dtypes[1] = PyArray_DescrFromType(NPY_BOOL);
 
     return 0;
--- a/numpy/_core/src/umath/ufunc_type_resolution.c
+++ b/numpy/_core/src/umath/ufunc_type_resolution.c
@@ -429,6 +429,31 @@ PyUFunc_NegativeTypeResolver(PyUFuncObject *ufunc,
     return ret;
 }
 
+/*
+ * This function applies special type resolution rules for the 'sign' ufunc.
+ * 'sign' converts timedelta64 to float64, so isn't covered by the simple
+ * unary type resolution.
+ *
+ * Returns 0 on success, -1 on error.
+ */
+NPY_NO_EXPORT int
+PyUFunc_SignTypeResolver(PyUFuncObject *ufunc,
+                         NPY_CASTING casting,
+                         PyArrayObject **operands,
+                         PyObject *type_tup,
+                         PyArray_Descr **out_dtypes)
+{
+    if (PyArray_DESCR(operands[0])->type_num == NPY_TIMEDELTA) {
+        out_dtypes[0] = NPY_DT_CALL_ensure_canonical(PyArray_DESCR(operands[0]));
+        out_dtypes[1] = PyArray_DescrFromType(NPY_DOUBLE);
+        return 0;
+    }
+    else {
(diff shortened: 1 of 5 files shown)


WarrenWeckesser:
*If* this change is accepted, it should not be merged until after 2.4.0 is branched, so it would be in 2.5.0.

WarrenWeckesser:
No response on the mailing list to https://mail.python.org/archives/list/numpy-discussion@python.org/thread/7ELIJOAAZAMFIGNDAUMOCLXYBRYYKE2Z/; not surprising, it is a bit of a niche topic.

I'll take this out of "draft" status after 2.4.0 is branched and 2.5.0 development starts on main.

mhvk (approved):
Nice that it actually turns out to be quite simple to code!

Approving now; doesn't look like there is going to be much discussion...

WarrenWeckesser:
Moved to "ready to review" status.

It occurred to me that if we had an *integer with nan* type, it could be used for the sign of the floating point and timedelta64 types.  But I don't think we'll add that just to make the output of `sign()` a bit nicer.

mhvk (approved):
Reapproving. Maybe @seberg can have a quick look too?

numpy/_core/src/umath/ufunc_type_resolution.c
```diff
@@ -429,6 +429,31 @@ PyUFunc_NegativeTypeResolver(PyUFuncObject *ufunc,
...
+ */
+NPY_NO_EXPORT int
+PyUFunc_SignTypeResolver(PyUFuncObject *ufunc,
+                         NPY_CASTING casting,
+                         PyArrayObject **operands,
+                         PyObject *type_tup,
+                         PyArray_Descr **out_dtypes)
+{
+    if (PyArray_DESCR(operands[0])->type_num == NPY_TIMEDELTA) {
+        out_dtypes[0] = NPY_DT_CALL_ensure_canonical(PyArray_DESCR(operands[0]));
```
seberg, line 447: I guess it's practically impossible, but I think this call can actually fail.
Maybe I should actually fix this, so that it can't fail for the simple dtypes (i.e. the simple ones can actually just return the `singleton` here, but they create a new dtype right now).

Other than this nit (and I think it's probably wrong in at least one other place, so I am happy to change `_ensure_canonical` here for the simple dtypes), it looks good to me.

@mhvk I am fine with putting this in, don't see much chance of downstream annoyance. I'll note that astropy asked before to not merge weirder things while the rc is in progress because it can be confusing in CI (i.e. you expect the change to be part of the rc release, but it's not).
But I am fine with just merging either way.

WarrenWeckesser: Thanks @seberg.  I took inspiration from the simpler type resolvers, and there are two of those where the return value of `NPY_DT_CALL_ensure_canoncial()` is not checked for NULL: `PyUFunc_IsNaTTypeResolver` and `PyUFunc_IsFiniteTypeResolver`.  It looks like there is a path in `PyUFunc_SimpleUniformOperationTypeResolver` where it does not check for NULL.  All the other occurrences in this file have a check for a NULL return value.

I'll fix this one, and I can fix the others while I'm at it, unless your idea to change _ensure_canonical is something that you'll do soon.

> I'll note that astropy asked before to not merge weirder things while the rc is in progress...

No hurry, whatever waiting time seems reasonable is fine with me.

seberg: Ahh, sorry please just fix it here. I got it wrong because times are parametric dtypes so this will always be able to fail (unless we add a singleton per unit, which we currently don't).

Code that uses the trivial builtins where it is obvious that it can't fails, should just avoid this function (e.g. use `PyArray_DescrFromType(...)` or `NPY_DTYPE(descr)->singleton).

WarrenWeckesser: Done.

WarrenWeckesser:
Test failures after the last commit are not related:
* [freebsd_test](https://github.com/numpy/numpy/pull/30194/checks?check_run_id=57308720049): infrastructure problem (404 error).
* *Edit: fixed by librt 0.7.2.* <strike>[Linux tests / armhf_test (pull_request)](https://github.com/numpy/numpy/actions/runs/19981546122/job/57309005300?pr=30194) and [Native ppc64le Linux Test / Native PPC64LE (pull_request)](https://github.com/numpy/numpy/actions/runs/19981546102/job/57308720317?pr=30194): a transitive dependency, [`librt`](https://pypi.org/project/librt/), that is not available as a wheel on these platforms fails to build from the tarball. Version 0.7.0 of `librt` was released today. I reported the problem here: https://github.com/mypyc/librt/issues/20.  See https://github.com/numpy/numpy/pull/30385 for @jorenham's proposed fix for NumPy CI.</strike>

jorenham:
> * [freebsd_test](https://github.com/numpy/numpy/pull/30194/checks?check_run_id=57308720049): infrastructure problem (404 error).

This is already resolved on on main, so rebasing will fixed this one.

WarrenWeckesser:
I added another small test that `sign(nat)` is `nan` for a scalar `nat`, and rebased on the main branch to get CI all green.

mattip:
LGTM, and others have approved. 2.5 was branched.

mattip:
Thanks @WarrenWeckesser

PR #30196, opened 2025-11-11
DOC: Fix some broken refs and Typos.

Carreau:
The default role is set to autolink (which fallbacks to emphasis if not resolved) And I believe there is typo in ALIGNMENT vs ALIGNENT.

I would suggest moving away from autolink for default role, and have it raise an error if not resolved.
This would also make sure users use double backticks for code.


rossbar (approved):
Thanks @Carreau !

PR #30197, opened 2025-11-11
ENH,MAINT: rewrite np.fix to use np.trunc internally

hellerve:
As discussed in #30098, this PR is a break-out PR to rewrite `np.fix` to use `np.trunc` internally. @seberg I hope I understood your comment in https://github.com/numpy/numpy/pull/30098#issuecomment-3516011752 correctly.

This PR should have no observable effects except maybe a speed-up of `np.fix`.

Cheers

### DIFF
--- a/numpy/lib/_ufunclike_impl.py
+++ b/numpy/lib/_ufunclike_impl.py
@@ -56,15 +56,7 @@ def fix(x, out=None):
     array([ 2.,  2., -2., -2.])
 
     """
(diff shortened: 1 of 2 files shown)


seberg:
Yeah, exactly, thanks!  I suspect you'll need this one small typing fixup also to make CI pass.

seberg:
Thanks @hellerve, this fixes the regression and hopefully nobody actually notices the change :).

Issue #29821, opened 2025-09-27
numpy.result_type acts inconsistently on Python/NumPy scalars


MustafaYounes1:
### Describe the issue:

According to the API Reference, the algorithm of [numpy.result_type](https://numpy.org/doc/stable/reference/generated/numpy.result_type.html#numpy.result_type) determines the output type following these steps:

1. Categories are determined by first checking which of boolean, integer (int/uint), or floating point (float/complex) the maximum kind of all the arrays and the scalars are.

2.  If there are only scalars or the maximum category of the scalars is higher than the maximum category of the arrays, the data types are combined with [promote_types](https://numpy.org/doc/stable/reference/generated/numpy.promote_types.html#numpy.promote_types) to produce the return value.

3.  Otherwise, [min_scalar_type](https://numpy.org/doc/stable/reference/generated/numpy.min_scalar_type.html#numpy.min_scalar_type) is called on each scalar, and the resulting data types are all combined with [promote_types](https://numpy.org/doc/stable/reference/generated/numpy.promote_types.html#numpy.promote_types) to produce the return value.

However, when it's provided with a Python scalar and an ndarray of the same category (e.g., int), it seems it completely ignores the scalar and outputs whatever datatype the ndarray has (and that contradicts the third point above; according to the doc, it should apply `min_scalar_type` on the scalar, and follow that with a call to `promote_types`).

However, when the provided scalar is a NumPy scalar, the behavior is different: what it seemingly does is only a call to `promote_types`

So, I'm wondering what exactly the role of `min_scalar_type` is in this context?

Please take a look at the provided code that shows this unexpected behaviour with a couple of examples.  

### Reproduce the code example:

```python
import numpy as np

#############
# Example 1 #
#############

# Both the ndarray and the scalar are of the same category (int)

a = np.empty(1, dtype='i1')

# Python Scalar & ndarray

# The following outputs 'int8' (completely ignoring the scalar)
print(np.result_type(-129, a))

# according to the doc; it should have called:
print(np.promote_types(np.min_scalar_type(-129), a.dtype))  # gives int16

# NumPy Scalar & ndarray

# gives 'int16' (seemingly only calls np.promote_types)
print(np.result_type(np.int16(-129), a))  
 
#############
# Example 2 #
#############

# The ndarray has a higher category (float vs int)

b = np.empty(1, np.float32)

# Python Scalar & ndarray

# The following outputs 'float32' (completely ignoring the scalar)
[...]

PR #30199, opened 2025-11-11
DOC: update result_type docs to link to promotion rules

hellerve:
This PR addresses #29821 by linking to [the data type promotion rules](https://numpy.org/doc/stable/reference/arrays.promotion.html) documentation instead of trying re-iterate it inline. While it would be nice to provide an overview in the function’s docstring, I think it’s better to keep it all in one place and make that place the canonical document.

Let me know what you think!

Cheers

### DIFF
--- a/numpy/_core/multiarray.py
+++ b/numpy/_core/multiarray.py
@@ -701,8 +701,7 @@ def result_type(*arrays_and_dtypes):
     result_type(*arrays_and_dtypes)
 
     Returns the type that results from applying the NumPy
-    type promotion rules to the arguments
-    (`see also <https://numpy.org/doc/stable/reference/arrays.promotion.html>`_).
+    :ref:`type promotion <arrays.promotion>` rules to the arguments.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


numpy/_core/multiarray.py
```diff
@@ -701,19 +701,8 @@ def result_type(*arrays_and_dtypes):
...
-    of the scalar is taken into account.
-
-    For example, calculating 3*a, where a is an array of 32-bit floats,
-    intuitively should result in a 32-bit float output.  If the 3 is a
-    32-bit integer, the NumPy rules indicate it can't convert losslessly
-    into a 32-bit float, so a 64-bit float should be the result type.
-    By examining the value of the constant, '3', we see that it fits in
-    an 8-bit integer, which can be cast losslessly into the 32-bit float.
+    type promotion rules to the arguments
+    (`see also <https://numpy.org/doc/stable/reference/arrays.promotion.html>`_).
```
mattip:
```suggestion
    :ref:`type promotion rules <arrays.promotion>` to the arguments
```

mattip (commented):
Thanks, makes sense. I suggest you simplify the link a bit. BTW, if you add `[skip azp][skip actions][skip cirrus]` to the commit message, it will only build the documentation rather than run all the CI

numpy/_core/multiarray.py
hellerve: Thank you! I wasn’t aware of that syntax!

mattip:
Thanks @hellerve

PR #30201, opened 2025-11-11
ENH: Detect Fortran vs C order in array_assign_boolean_subscript

hellerve:
This PR fixes #30156 by detecting when to set `KEEPORDER` order in `array_assign_boolean_subscript`.

On my machine, before this fix, using the benchmark I added:

```
[ 0.00%] ·· Benchmarking existing-py_Users_veitheller_.pyenv_versions_3.12.11_bin_python3.12
[50.00%] ··· Running (bench_indexing.BooleanAssignmentOrder.time_boolean_assign_scalar--).
[100.00%] ··· bench_indexing.BooleanAssignmentOrder.time_boolean_assign_scalar                                          ok
[100.00%] ··· ======= =============
               order
              ------- -------------
                 C       979±10μs
                 F     1.20±0.01ms
              ======= =============

[100.00%] ···· For parameters: 'C'
               NumPy CPU features: nothing enabled

               For parameters: 'F'
               NumPy CPU features: nothing enabled
```

And after:

```
[ 0.00%] ·· Benchmarking existing-py_Users_veitheller_.pyenv_versions_3.12.11_bin_python3.12
[50.00%] ··· Running (bench_indexing.BooleanAssignmentOrder.time_boolean_assign_scalar--).
[100.00%] ··· bench_indexing.BooleanAssignmentOrder.time_boolean_assign_scalar                                          ok
[100.00%] ··· ======= ==========
               order
              ------- ----------
                 C     974±10μs
                 F     984±10μs
              ======= ==========

[100.00%] ···· For parameters: 'C'
               NumPy CPU features: nothing enabled

               For parameters: 'F'
               NumPy CPU features: nothing enabled
```

I hope this is the fix we had in mind, I operated on the simple assumption that number_go_down==good.

Do we need a release note for this?

Cheers

### DIFF
--- a/benchmarks/benchmarks/bench_indexing.py
+++ b/benchmarks/benchmarks/bench_indexing.py
@@ -94,11 +94,7 @@ class BooleanAssignmentOrder(Benchmark):
         # emulate gh-30156: boolean assignment into a sliced Fortran/C view
         self.view = self.base[1:-1, 1:-1, 1:-1]
         mask = np.random.RandomState(0).rand(*self.view.shape) > 0.5
-        if order == 'F':
-            mask = np.asfortranarray(mask)
-        else:
-            mask = np.ascontiguousarray(mask)
-        self.mask = mask
+        self.mask = mask.copy(order)
         self.value = np.uint32(7)
 
     def time_boolean_assign_scalar(self, order):
--- a/benchmarks/benchmarks/bench_indexing.py
+++ b/benchmarks/benchmarks/bench_indexing.py
@@ -84,6 +84,27 @@ class ScalarIndexing(Benchmark):
             arr[indx] = val
 
 
+class BooleanAssignmentOrder(Benchmark):
+    params = ['C', 'F']
+    param_names = ['order']
+
+    def setup(self, order):
+        shape = (64, 64, 64)
+        self.base = np.zeros(shape, dtype=np.uint32, order=order)
+        # emulate gh-30156: boolean assignment into a sliced Fortran/C view
+        self.view = self.base[1:-1, 1:-1, 1:-1]
+        mask = np.random.RandomState(0).rand(*self.view.shape) > 0.5
+        if order == 'F':
+            mask = np.asfortranarray(mask)
+        else:
+            mask = np.ascontiguousarray(mask)
+        self.mask = mask
+        self.value = np.uint32(7)
+
+    def time_boolean_assign_scalar(self, order):
+        self.view[self.mask] = self.value
+
+
 class IndexingSeparate(Benchmark):
     def setup(self):
         self.tmp_dir = mkdtemp()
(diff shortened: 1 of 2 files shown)


benchmarks/benchmarks/bench_indexing.py
```diff
@@ -84,6 +84,27 @@ def time_assign_cast(self, ndim):
...
 
+class BooleanAssignmentOrder(Benchmark):
+    params = ['C', 'F']
+    param_names = ['order']
+
+    def setup(self, order):
+        shape = (64, 64, 64)
+        self.base = np.zeros(shape, dtype=np.uint32, order=order)
+        # emulate gh-30156: boolean assignment into a sliced Fortran/C view
+        self.view = self.base[1:-1, 1:-1, 1:-1]
```
seberg: Slicing shouldn't matter, I think, although, maybe we have a C-contig short-cut somewhere and then it would.

numpy/_core/src/multiarray/mapping.c
```diff
@@ -1176,6 +1176,10 @@ array_assign_boolean_subscript(PyArrayObject *self,
     }
 
     v_data = PyArray_DATA(v);
+    if (v_stride == 0) {
+        /* Scalar/broadcast values benefit from KEEPORDER to match self layout */
+        order = NPY_KEEPORDER;
```
seberg: Agreed that this is good enough, but might as well move it into the `v_stride` branch above?
(Honestly, passing in the order seems silly to begin with, we could just set it to KEEPORDER in the second branch and C_ORDER in the first...)

Since I'm nitpicky, I might say that the comment could say: If the same value is repeated, iteration order does not matter.

benchmarks/benchmarks/bench_indexing.py
```diff
@@ -84,6 +84,27 @@ def time_assign_cast(self, ndim):
...
+        mask = np.random.RandomState(0).rand(*self.view.shape) > 0.5
+        if order == 'F':
+            mask = np.asfortranarray(mask)
+        else:
+            mask = np.ascontiguousarray(mask)
+        self.mask = mask
```
seberg:
```suggestion
        self.mask = mask.copy(order)
```
or similar also works, not that it matters.

seberg (approved):
Nice, thanks, the benchmark also looks good, a few small nitpick, none really blocking.

I don't think we need a release note, but if you like you can add a short one.

hellerve:
I won’t add a release note then, I’m not sure whether this change is impactful enough in real terms for a release note.

seberg:
Thanks a lot @hellerve ! Got to like a 1-line fix :), and great to have the new benchmark!
(I am almost surprised at how small the speedup is, but I guess for large, out of cache, arrays it is probably a lot bigger.)

hellerve:
Yeah, and 20% is not nothing, especially if you do it a lot.

william-silversmith:
Fantastic! The volumes I process are typically 512^3 voxels. The speedup is substantially more once you are larger than on-die caches. Thank you!

PR #30210, opened 2025-11-14
ENH:  einsum_sumprod.c.src to cpp in numpy.multiarray see #29528

athurdekoos:
This PR addresses a part of #29528.

### Description
This change ports einsum_sumprod from the generated einsum_sumprod.c.src file to a C++ template source file (einsum_sumprod.cpp).  The goal is to improve readability and maintainability while preserving the existing behavior and performance characteristics.

There are no intended changes to the public API. 

### Summary of changes:
- Replacement of `einsum_sumprod.c.src` with  `einsum_sumprod.cpp`.
- Original function structure was preserved as closely as possible.
  - Function patterns remain the same.
  - Function names have been preserved where possible.
  - Variable names have been preserved where possible.
- Logic was maintained as close to the original as possible. 
- Some larger functions have been refactored into smaller discrete implementations.
- Helper functions added. 
- If possible, macro based optimizations have been preserved. 
- Introduced new macro implementations primarily to handle NPY_SIMD_F32 and NPY_SIMD_F64.
- Optimizations that were only present in some code paths in the original implementation have been ported to the corresponding functions in the new structure when possible. 
- Expanded and clarified some comments.
- Some templated parameters from the original code were removed where they were not implement or did not materially improve performance or clarity.

### Notes: 
- Prioritization of this port was readability over aggressive abstraction.
- When possible templating is resolved at compile time. 

### Considerations not implemented: 
- Considered introducing C++ namespaces but did not do so to keep in line with the overall C-API style. 
  - Some function names as a result are a long but descriptive. 
- Considered using templates as a replacement for some macros, but opted against this where it hurt clarity

### Benchmarks 
- Across multiple machines were no major performance changes were observed.
- However, on my personal machine, benchmarks seem to improve dramatically in several cases with the exception of a small inconsistent regression 

 
| Change | Before [fabf1844] <main> | After [fad21058] <einsum_sumprod_to_cpp> | Ratio | Benchmark (Parameter)                      |
|--------|---------------------------|-------------------------------------------|--------|--------------------------------------------|
| +      | 194±7μs                  | 220±10μs                                  | 1.13   | bench_linalg.Eindot.time_inner_trans_a_a   |
| -      | 25.5±1μs                 | 23.5±0.3μs                                | 0.92   | bench_linalg.Eindot.time_dot_d_dot_b_c     |
| -      | 1.10±0.1ms               | 987±30μs                                  | 0.89   | bench_linalg.Eindot.time_dot_trans_at_a    |
| -      | 106±10μs                 | 70.8±7μs                                  | 0.66   | bench_linalg.Eindot.time_dot_trans_a_atc   |
| -      | 1.44±0.02ms              | 899±30μs                                  | 0.62   | bench_linalg.Eindot.time_einsum_i_ij_j     |
| -      | 110±5ms                  | 33.1±3ms                                  | 0.30   | bench_linalg.Eindot.time_einsum_ijk_jil_kl |
| -      | 114±6ms                  | 6.11±0.08ms                               | 0.05   | bench_linalg.Eindot.time_einsum_ij_jk_a_b  |

Please let me know if you'd like me to adjust naming or structure.


athurdekoos:
@inakleinbottle quick fyi ping to keep you apprised of my current status

numpy/_core/src/multiarray/einsum_sumprod.cpp
```diff
@@ -0,0 +1,2281 @@
...
+/* Helper traits mapping a (T, Temptype) pair to the SIMD wrapper */
+template <typename T, typename TempType>
+struct SumSIMD {};
+
+#if NPY_SIMD_F32
+/* Wrapper around the npyv float32 SIMD, provides a uniform interface between SIMDF32
+ * and SIMDF64 */
+template <>
+struct SumSIMD<npy_float, npy_float> {
+    using SimdType = NpySIMDF32;
```
inakleinbottle: I think this, and a great deal of the other supporting structure will be useful elsewhere and should probably be moved to a header file that can just be included in all these replacement files as they are added. It might be worth having a discussion with the optimization team too for the SIMD stuff to avoid replicating work.

numpy/_core/src/multiarray/einsum_sumprod.cpp
```diff
@@ -0,0 +1,2281 @@
...
+                                                                              count);
+#else   // !NPY_SIMD_F32
+    return scaller_sum_of_arr<T, Temptype>(data, count);
+#endif  // NPY_SIMD_F32
+}
+
+/* Template where (npy_double, npy_double) will allow the SIMD
+ * capable version.*/
+template <typename T, typename Temptype,
+          typename std::enable_if<std::is_same<T, Temptype>::value &&
```
inakleinbottle: The enable_if construct is a little odd here. First, since we have access to C++17, you might want to use the _t and _v variants which make this a bit more readable. However, here I think simple overloading would serve the same purpose. C++ always prefers concrete instantiations over templates if it is the best match, and enable_if constructions have a large cost for instantiation whereas overloading has a much smaller cost. I think though, that what is needed here is actually a partially specializable struct template that can be used as a customization point inside the `sum_of_arr` function. For instance:

```c++
template <typename T, typename TempType, typename SFINAE=void>
struct SumOfArr {
    static TempType eval(T* daa, npy_intp size) noexcept(?) {}
};
template <typename T>
static inline TempTypeOf<T> sum_of_arr(T* data, npy_intp count) noexcept (?) {
    using Helper = SumOfArray<T, TempTypeOf<T>>;
    return Helper::eval(data, count);
}
```
This is just an idea, but it would give you relatively low-overhead control the implementation based on the type and TempType (which I've shorthanded here as a trait, but could be left as a template argument if multiple choices are necessary).

inakleinbottle:
Thanks @athurdekoos for tagging me in. I had a quick look over the code and made a couple of what I hope are helpful comments, or at least things to think about. To be clear, I don't think these are necessary changes, so this isn't a review of any kind. Both comments are "looking ahead" to the other c.src modules too, so establishing some standard patterns and reusable components might be a good idea. Very happy discuss or explain further.

numpy/_core/src/multiarray/einsum_sumprod.cpp
ngoldbaum: ping @seiko2plus, @Mousius for SIMD/C++ comments.

athurdekoos: @inakleinbottle I agree whole heartedly.

numpy/_core/src/multiarray/einsum_sumprod.cpp
```diff
@@ -0,0 +1,2281 @@
...
+        v_accum = SimdType::npyv_add(a, v_accum);
+    }
+    T accum = SimdType::npyv_sum(v_accum);
+    SimdType::cleanup();
+    return accum;
+}
+
+template <typename T, typename Temptype>
+static inline NPY_GCC_OPT_3 Temptype
+scaller_sum_of_arr(const T *data, npy_intp count)
```
ngoldbaum: what does scaller mean? It's not clear to me and I don't think it was used in the old .c.src implementation

numpy/_core/src/multiarray/einsum_sumprod.cpp
```diff
@@ -0,0 +1,2281 @@
...
+    for (; count > 0; count -= vstep, data += vstep) {
+        SimdReg a = SimdType::load_tillz(data, count);
+        v_accum = SimdType::npyv_add(a, v_accum);
+    }
+    T accum = SimdType::npyv_sum(v_accum);
+    SimdType::cleanup();
+    return accum;
+}
+
+template <typename T, typename Temptype>
```
ngoldbaum: can we give these template variables more meaningful names? here and throughout this code

ngoldbaum (commented):
I skimmed over the entire thing but didn't go through in detail. I left a couple comments about naming.

Overall I think I prefer this even if it does end up substantially longer. The templates you started with are very hard to read and understand given the multiple levels of string substitution happening.

It might also help for future readability to add a top-level comment describing the structure of the templating and how various kinds of einsum operations are dispatched. An ASCII table/flowchart wouldn't hurt IMO.

numpy/_core/src/multiarray/einsum_sumprod.cpp
athurdekoos: I agree this is a good idea.
 
I'm not sure what a better name would be though.  `@temptype@` is what the c.src file called them. 

My best idea would be something like `Accum_Type` since it looks like it's mostly used in conjunction with the `accum` or other accumulator linked intermediates. 

I assume this was done for greater precision but am not entirely sure

For reference here are the differences

| Type             | temptype       |
|------------------|----------------|
| npy_half         | npy_float      |
| npy_cfloat       | npy_float      |
| npy_cdouble      | npy_double     |
| npy_clongdouble  | npy_longdouble |

athurdekoos:
Changes made:
- Created new file `multiarray/simd_accum_traits.h` to host SIMD wrappers.  
  - Relocated SIMD related supporting structures to `multiarray/simd_accum_traits.h`.
- Removed all `enable_if`, replaced with templated structs. 
- Removed all `typename Temptype` replaced with `typename AccumType` for better identification. 
- Removed all mention of `scaller*` replaced with `elementwise*`

Apologies for large diff!

numpy/_core/src/multiarray/simd_accum_traits.h
```diff
@@ -0,0 +1,116 @@
...
+
+/* Helper traits mapping a (T, AccumType) pair to the SIMD wrapper */
+template <typename T, typename AccumType>
+struct SumSIMD {};
+
+#if NPY_SIMD_F32
+/* Wrapper around the npyv float32 SIMD, provides a uniform interface between SIMDF32
+ * and SIMDF64 */
+template <>
+struct SumSIMD<npy_float, npy_float> {
```
inakleinbottle: Personally I would not define a new trait unless there is a need to prserve the NpySIMDF64/NpySIMDF32 structs independently and just declare these as template specialisations. This means one list `::SimdType` thing later on and unifies the traits for both SIMD supporting types vs non-supporting types. However, I would wait to see what the optimisation team say before committing to a path.

inakleinbottle: This would require having a default implementation that doesn't do any SIMD ops too.

numpy/_core/src/multiarray/einsum_sumprod.cpp
```diff
@@ -0,0 +1,2181 @@
...
+
+#define NPY_NO_DEPRECATED_API NPY_API_VERSION
+#define _MULTIARRAYMODULE
+
+#include "einsum_sumprod.h"
+
+#include "common.h"
+#include "einsum_debug.h"
+
+#include "simd/simd.h"
```
seiko2plus:
```suggestion
#include "simd/simd.hpp"
```

We have decided (NEP-54) to move Highway rather than implementing a C++ interface for C universal intrinsics.

numpy/_core/src/multiarray/einsum_sumprod.cpp
```diff
@@ -0,0 +1,2181 @@
...
+#include <array>
+#include <iostream>
+#include <numpy/arrayobject.h>
+#include <numpy/halffloat.h>
+#include <numpy/ndarraytypes.h> /* for NPY_NTYPES_LEGACY */
+#include <numpy/npy_common.h>
+
+#include "simd_accum_traits.h"
```
seiko2plus:
```suggestion
```
Therefore, this header should be removed.

numpy/_core/src/multiarray/einsum_sumprod.cpp
```diff
@@ -0,0 +1,2181 @@
...
+#include "einsum_sumprod.h"
+
+#include "common.h"
+#include "einsum_debug.h"
+
+#include "simd/simd.h"
+#include <array>
+#include <iostream>
+#include <numpy/arrayobject.h>
+#include <numpy/halffloat.h>
```
seiko2plus, line 24:
```suggestion
#include "common.hpp"
```
Use the C++ class `Half` instead, which is implemented by `common/half.hpp`, rather than the C wrapper.

numpy/_core/src/multiarray/einsum_sumprod.cpp
```diff
@@ -0,0 +1,2181 @@
...
+{
+    return SumOfArr<T, AccumType>::eval(data, count);
+}
+
+template <typename SimdType>
+static inline NPY_GCC_OPT_3 void
+floating_point_sum_of_products_muladd(const typename SimdType::T *data,
+                                      typename SimdType::T *data_out,
+                                      typename SimdType::AccumType scalar,
+                                      npy_intp count)
```
seiko2plus:
```suggestion
template <typename Lane, typename LaneAcc>
static inline NPY_GCC_OPT_3 void
floating_point_sum_of_products_muladd(const Lane *data,
                                      Lane *data_out,
                                      LaneAcc scalar,
                                      SSize count)
```
Just deduce the scalar type directly; regarding `npy_intp`, you can use `SSize`, which maps to `Py_ssize_t` or `ptrdiff_t`.

numpy/_core/src/multiarray/einsum_sumprod.cpp
```diff
@@ -0,0 +1,2181 @@
...
+template <>
+inline npy_half
+to<npy_float, npy_half>(npy_float v)
+{
+    return npy_float_to_half(v);
+}
+
+template <typename SimdType>
+static inline NPY_GCC_OPT_3 typename SimdType::T
+floating_point_sum_of_arr(const typename SimdType::T *data, npy_intp count)
```
seiko2plus:
```suggestion
template <typename TLane>
static inline NPY_GCC_OPT_3 TLane
floating_point_sum_of_arr(const TLane *data, ptrdiff_t count)
```
Just deduce the scalar type directly; regarding `npy_intp`, you can use `SSize`, which maps to `Py_ssize_t` or `ptrdiff_t`.

numpy/_core/src/multiarray/einsum_sumprod.cpp
```diff
@@ -0,0 +1,2181 @@
...
+#include <numpy/npy_common.h>
+
+#include "simd_accum_traits.h"
+
+// ARM/Neon don't have instructions for aligned memory access
+#ifdef NPY_HAVE_NEON
+#    define EINSUM_IS_ALIGNED(x) 0
+#else  // NPY_HAVE_NEON
+#    define EINSUM_IS_ALIGNED(x) npy_is_aligned(x, NPY_SIMD_WIDTH)
+#endif  // NPY_HAVE_NEON
```
seiko2plus:
```suggestion
```
Specializing loops for aligned SIMD load and store operations is generally unnecessary, as modern CPUs handle unaligned memory accesses effectively.

numpy/_core/src/multiarray/einsum_sumprod.cpp
```diff
@@ -0,0 +1,2181 @@
...
+        v_accum = SimdType::npyv_add(a0123, v_accum);
+    }
+
+    for (; count > 0; count -= vstep, data += vstep) {
+        SimdReg a = SimdType::load_tillz(data, count);
+        v_accum = SimdType::npyv_add(a, v_accum);
+    }
+    T accum = SimdType::npyv_sum(v_accum);
+    SimdType::cleanup();
+    return accum;
```
seiko2plus:
```suggestion
    using namespace np::simd;
    using V = Vec<TLane>;
    
    const size_t vstep = Lanes<TLane>();
    const size_t wstep = vstep * 4;
    V zero = Zero<Lane>();
    V v_accum = zero;

    for (; count >= wstep; count -= wstep, data += wstep) {
        const V a0 = LoadU(data + vstep * 0);
        const V a1 = LoadU(data + vstep * 1);
        const V a2 = LoadU(data + vstep * 2);
        const V a3 = LoadU(data + vstep * 3);
        const V a01 = Add(a0, a1);
        const V a23 = Add(a2, a3);
        const V a0123 = Add(a01, a23);
        v_accum = Add(a0123, v_accum);
    }
    for (; count > 0; count -= vstep, data += vstep) {
        V a = LoadNOr(zero, data, count);
        v_accum = Add(a, v_accum);
    }
    return ReduceSum(v_accum);
```

Rather than continue with our own SIMD solution, we decided to move to Google’s Highway as part of the C++ transition, as demonstrated by `nep-54`. However, NumPy uses a lightweight wrapper to eliminate Highway tags required by some intrinsics, and we also use modern C++ features. During the migration of new kernels, we may need to expand this wrapper. For this kernel, we need to wrap two intrinsics: `LoadNOr` and `ReduceSum`.

e.g. you can add `LoadNOr` under:
https://github.com/numpy/numpy/blob/489c0eae45f42587483eb5ee6db9e40c9b960297/numpy/_core/src/common/simd/simd.inc.hpp#L46-L52

```C++
template <typename TLane>
HWY_API Vec<TLane>
LoadNOr(const Vec<TLane> &no, const TLane *p, size_t max_lanes_to_load)
{
    return hn::LoadNOr(no, _Tag<TLane>(), p, max_lanes_to_load);
}
```

numpy/_core/src/multiarray/einsum_sumprod.cpp
```diff
@@ -0,0 +1,2181 @@
...
+{
+    return (AccumType)v;
+}
+
+template <>
+inline npy_float
+from<npy_half, npy_float>(npy_half v)
+{
+    return npy_half_to_float(v);
+}
```
seiko2plus, line 44:
```suggestion
```
Use `static_cast` instead. The C++ class `np::Half` should be used directly rather than using the C wrapper. This is also important for inlining optimization.

numpy/_core/src/multiarray/einsum_sumprod.cpp
```diff
@@ -0,0 +1,2181 @@
...
+#endif
+    }
+};
+
+template <typename T, typename AccumType>
+static inline NPY_GCC_OPT_3 AccumType
+sum_of_arr(T *data, npy_intp count) noexcept
+{
+    return SumOfArr<T, AccumType>::eval(data, count);
+}
```
seiko2plus, line 155:
```suggestion
template <typename T>
static inline NPY_GCC_OPT_3 void
sum_of_products_contig_outstride0_one(int nop, char **dataptr, npy_intp const *strides,
                                      npy_intp count)
{
    NPY_EINSUM_DBG_PRINT1("sum_of_products_contig_outstride0_one (%d)\n", (int)count);
    if constexpr (std::is_same_v<T, np::Half>) {
      // speclized loop for wide cast
    }
#if NPY_HWY
    else if constexpr (np::simd::kSupportLane<T>) {
      // simd code
    }
#endif
    else {
      // scalar loop 
    }
}

template <typename T>
static inline NPY_GCC_OPT_3 void
sum_of_products_contig_outstride0_one<std::complex<T>>(int nop, char **dataptr, npy_intp const *strides,
                                      npy_intp count)
{
}
```
You can reduce the complexity of the code by using `np::simd::kSupportLane<T>` to check whether the lane type is supported by the SIMD interface. Also I see here unsecery depth that can be avoided.

seiko2plus (changes requested):
Thank you for working on it. A wrapper for C universal intrinsics cannot be accepted, as we agreed to use Google Highway as part of the C++ transition within NEP-54. I’ve left some suggestions below. I also see an opportunity to reduce the complexity of the C++ code by leveraging more C++17 features—for example, variadic templates.

numpy/_core/src/multiarray/einsum_sumprod.cpp
athurdekoos: Updated to `#include "simd/simd.hpp"`

numpy/_core/src/multiarray/einsum_sumprod.cpp
athurdekoos: Removed and deleted

numpy/_core/src/multiarray/einsum_sumprod.cpp
athurdekoos: Removed and deleted

numpy/_core/src/multiarray/einsum_sumprod.cpp
athurdekoos: created LoadNOr as suggested, thank you!

numpy/_core/src/multiarray/einsum_sumprod.cpp
athurdekoos: I did not implement this. As this diff is already huge I wanted to keep as much original logic as possible

numpy/_core/src/multiarray/einsum_sumprod.cpp
athurdekoos: I did not implement this. As this diff is already huge I wanted to keep as much original logic as possible

numpy/_core/src/multiarray/einsum_sumprod.cpp
athurdekoos: I did not implement this. As this diff is already huge I wanted to keep as much original logic as possible

numpy/_core/src/multiarray/einsum_sumprod.cpp
athurdekoos: I did not implement this. As this diff is already huge I wanted to keep as much original logic as possible

mattip:
It would be nice to rerun the einsum benchmarks on the latest changes.

athurdekoos:
@mattip good idea!
I'm showing this result on my machine: `BENCHMARKS NOT SIGNIFICANTLY CHANGED`

BastienGauthier:
Hi, and big thanks for your work ! Do you think this MR will be merged for 2.5 ? Given your benchmark, it seems game changer for my use cases ! (5-10sec einsum, even after going to f32 and order F and optimized ).

mattip:
Hmm. How do we reconcile these two conflicting statements?
> I'm showing this result on my machine: BENCHMARKS NOT SIGNIFICANTLY CHANGED

> Given your benchmark, it seems game changer for my use cases

BastienGauthier:
I understood there were no change compared to the initial benchmark of the PR. Have I mixed the informations ?

PR #30211, opened 2025-11-14
ENH: ``ufunc.__signature__``

jorenham:
This adds a dynamic `__signature__` attribute to the individual ufuncs, which gets used by introspection tools like `inspect.signature`.

`tp_getattro` is used in order to avoid `ufunc.__signature__` from becoming a class attribute, which sphinx did not seem to like.

Before, IPython showed something like this for `np.exp?`:

```py
Signature:       np.exp(*args, **kwargs)
Type:            ufunc
String form:     <ufunc 'exp'>   
[...]
```

After this change, this is what it looks like:

```py
Signature:      
np.exp(
    x,
    /,
    out=None,
    *,
    where=True,
    casting='same_kind',
    order='K',
    dtype=None,
    subok=True,
    signature=None,
)
Call signature:  np.exp(*args, **kwargs)
Type:            ufunc
String form:     <ufunc 'exp'>     
[...]
```

The "call signature" refers to `np.ufunc.__call__.__text_signature__`, which remains unchanged.

In case of `divmod`, the signature is

```py
np.divmod(
    x1,
    x2,
    /,
    out=(None, None),
    *,
    where=True,
    casting='same_kind',
    order='K',
    dtype=None,
    subok=True,
    signature=None,
)
```

GUFuncs include `**kwargs`, because it's not known whether they have an `axis` parameter:

```py
np.vecdot(
    x1,
    x2,
    /,
    out=None,
    *,
    casting='same_kind',
    order='K',
    dtype=None,
    subok=True,
    signature=None,
    **kwargs,
)
```


numpy/_core/_internal.py
```diff
@@ -893,6 +893,53 @@ def _ufunc_doc_signature_formatter(ufunc):
...
+    params.extend((
+        Parameter("casting", Parameter.KEYWORD_ONLY, default='same_kind'),
+        Parameter("order", Parameter.KEYWORD_ONLY, default='K'),
+        Parameter("dtype", Parameter.KEYWORD_ONLY, default=None),
+        Parameter("subok", Parameter.KEYWORD_ONLY, default=True),
+        Parameter("signature", Parameter.KEYWORD_ONLY, default=None),
+    ))
+
+    if ufunc.signature is not None:
+        # some gufuncs additionally support an `axis` parameters
```
mhvk: Note that all `gufunc` support `axes` and `keepdims`, while `axis` can be used only if there is only a single core dimension -- see https://numpy.org/doc/stable/reference/ufuncs.html

Maybe overall it makes most sense to just add those explicitly above, as an `else` clause to `ufunc.signature is None` -- they are probably more useful to see than `casting`, etc. (and it is also the order in the docs). The one annoyance is that while `keepdims=False` by default, `axis` and `axes` don't really have a default - it is their presence that triggers examination; maybe `np._NoValue`?  (At least for `axis`, that leaves the option to at some point in the future make `axis=None` mean `ravel` like for reductions...) Note that probably the same is true for things like `signature`, so maybe I'm overthinking.

numpy/_core/_internal.py
```diff
@@ -893,6 +893,53 @@ def _ufunc_doc_signature_formatter(ufunc):
     return f'{ufunc.__name__}({in_args}{out_args}, *{kwargs})'
 
 
+def _ufunc_inspect_signature_builder(ufunc):
+    """
+    Builds a ``__text_signature__`` string.
+
+    Should be kept in sync with `_ufunc_doc_signature_formatter`.
```
mhvk, line 902: I wondered whether there was any way to combine the two, but probably difficult and overkill. 

However, do also add a note to `_ufunc_doc_signature_formatter` that the code here should be kept in sync with this one.

numpy/_core/tests/test_ufunc.py
```diff
@@ -3336,3 +3336,35 @@ def test_long_arrays(self):
...
+    assert "x" in sig12.parameters
+    assert "x1" not in sig12.parameters
+    assert sig12.parameters["out"].default == (None, None)
+
+    sig22 = inspect.signature(np.divmod)
+    assert "x" not in sig22.parameters
+    assert "x1" in sig22.parameters
+    assert "x2" in sig22.parameters
+    assert "x3" not in sig22.parameters
+    assert sig22.parameters["out"].default == (None, None)
```
mhvk: I'd check maybe one or two `gufuncs` as well, like `vecdot` or `matmul`.

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -6721,6 +6747,9 @@ static PyGetSetDef ufunc_getset[] = {
     {"__doc__",
         (getter)ufunc_get_doc, (setter)ufunc_set_doc,
         NULL, NULL},
+    {"__signature__",
```
mhvk: Nit and arguably out of scope, but maybe take the opportunity to move `__name__` up here too, so that the order matches https://numpy.org/doc/stable/reference/ufuncs.html

mhvk (commented):
Very nice! Main comment is about being explicit for gufuncs as well.

Also, as I was looking at https://numpy.org/doc/stable/reference/ufuncs.html to remind myself of the actual signatures, I saw that the text mentions `__doc__` - do you want to mention `__signature__` there as well? It does look like that list is complete, so might as well keep it like that!

Separately, currently `np.add.reduce?` just shows the docstring, so all fine from a user perspective. But does it work for introspection? If  so, definitely for another PR, but I thought I'd mention it in part in case it has implications here, that really the `__signature__` should somehow be attached to `__call__` rather than the ufunc itself.

charris:
Circleci failure looks relevant.

```
WARNING: Failed to get a constructor signature for numpy.ufunc: unexpected object <attribute '__signature__' of 'numpy.ufunc' objects> in __signature__ attribute
WARNING: Failed to get a constructor signature for numpy.ufunc: unexpected object <attribute '__signature__' of 'numpy.ufunc' objects> in __signature__ attribute
```

numpy/_core/_internal.py
jorenham: > Note that all `gufunc` support `axes` and `keepdims`, while `axis` can be used only if there is only a single core dimension -- see [numpy.org/doc/stable/reference/ufuncs.html](https://numpy.org/doc/stable/reference/ufuncs.html)

Ah nice, good to know; I'll add them.

numpy/_core/_internal.py
jorenham: Oh woops, I forgot to update that. It's what I initially tried, but unlike `__signature__`, `inspect.signature` didn't seem to pick up on it.

numpy/_core/src/umath/ufunc_object.c
jorenham: Nit away; I'm a sucker for code style and things like that, so I'm all ears.

jorenham:
> Separately, currently `np.add.reduce?` just shows the docstring, so all fine from a user perspective.

That even seems necessary, judging by the circleCI errors; I'll take of care of it.

> I'd mention it in part in case it has implications here, that really the `__signature__` should somehow be attached to `__call__` rather than the ufunc itself.

That's initially what I had planned on doing. But that'd require a separate signature per ufunc instance, whereas for the other methods a one-size-fits-all signature is used. The C machinery surrounding `__call__` was also quite different from the other methods. So I wasn't able to figure out how I'd be able to attach a specializing `__text_signature__` or `__signature__` to `ufunc.__call__`. That's why I ended up settling for `ufunc.__signature__`, because that's what `inspect.signature` will also support.

numpy/_core/_internal.py
jorenham: And yea I agree that an explicit `axis` would be better than a `**kwargs`, as it'll be a less type-unsafe that way, and is structurally closer to the "actual" signature.
And `_NoValue` as default seems like a good choice. For `__text_signature__` that wouldn't have been possible, but with `__signature__` we have a lot more flexibility in that sense.

numpy/_core/tests/test_ufunc.py
jorenham: Note to self; these are the gufuncs that currently exist in numpy:

```pycon
>>> import numpy as np
>>> {k for k, v in vars(np).items() if isinstance(v, np.ufunc) and v.signature is not None}
{'matmul', 'matvec', 'vecmat', 'vecdot'}
```

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -6635,6 +6635,57 @@ ufunc_set_doc(PyUFuncObject *ufunc, PyObject *doc, void *NPY_UNUSED(ignored))
...
+
+    return signature;
+}
+
+static PyObject *
+ufunc_getattro(PyObject *obj, PyObject *name)
+{
+    // __signature__ special-casing to prevent class attribute access
+    if (PyUnicode_Check(name) &&
+        PyUnicode_CompareWithASCIIString(name, "__signature__") == 0) {
```
mhvk: Maybe compare with `npy_interned_str.__signature__`?

Aside: interesting that it needs the attribute getter, I had not thought of this... is it needed just for the caching in `__dict__`? since signature access is unlikely to be frequent or performance critical, I think it is actually fine not to cache. But this is OK too!

mhvk (approved):
Looks good! I have just one in-line comment, but since that is just one small suggested improvement plus an optional comment, I'll approve now.

numpy/_core/src/umath/ufunc_object.c
jorenham: > Maybe compare with `npy_interned_str.__signature__`?

Good idea :ok_hand: 

> `__dict__`? since signature access is unlikely to be frequent or performance critical, I think it is actually fine not to cache. But this is OK too!

It's needed so that there won't be a class-level attribute exposed as `ufunc.__signature__`, because sphinx didn't know what to do with that :shrug:.

mhvk:
Thanks, let's get it in!

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -6635,6 +6635,57 @@ ufunc_set_doc(PyUFuncObject *ufunc, PyObject *doc, void *NPY_UNUSED(ignored))
...
+}
+
+static PyObject *
+ufunc_getattro(PyObject *obj, PyObject *name)
+{
+    // __signature__ special-casing to prevent class attribute access
+    if (PyUnicode_Check(name) &&
+        PyUnicode_Compare(name, npy_interned_str.__signature__) == 0) {
+        return ufunc_get_inspect_signature((PyUFuncObject *)obj, NULL);
+    }
```
seberg, line 6683: (Sorry, I had noticed this yesterday, and should have commented right away.)

How bad is it if we don't do this dance?  A custom getattr feels like a bit of a code smell, and disables some optimizations I am pretty sure (e.g. `np.add.reduce(...)` will now have to actually _create_ `np.add.reduce` (the bound method) and call that rather than calling `np.add.reduce` directly.

In Python, it is simple to hack around it. Unfortunately, in C it seems harder unfortunately.
(I think we might have to make `ufunc` a heaptype so that it has a `__dict__` itself and we can use a custom descriptor just like we could in Python.)

So I guess a comment for now, although since converting to the heaptype is probably the trickiest part of this, maybe it is actually not that bad.

mhvk: I think we could still go back to having it as a `getter` and just not caching the result in `__dict__`. But I'm not 100% that was the reason for putting it here. @jorenham?

jorenham: I've tried a couple of other ways to avoid the `ufunc.__signature__` class-attribute from being created, but this was the only reasonable one I was able to find. I also tried supressing the warning in sphinx, but that required adding a bunch of extra special-casing to `ocs/source/conf.py`. 

But to be clear, I have no problem with changing this back to a descriptor. And a heap-type also sounds fine to me. It's just that I don't know how to do that without also breaking Sphinx :shrug:

jorenham: For reference: https://app.circleci.com/pipelines/github/numpy/numpy/36234/workflows/a5cbaba2-0fee-4e61-b4f7-329e8f0e165c/jobs/51422

```
WARNING: Failed to get a constructor signature for numpy.ufunc: unexpected object <attribute '__signature__' of 'numpy.ufunc' objects> in __signature__ attribute
WARNING: Failed to get a constructor signature for numpy.ufunc: unexpected object <attribute '__signature__' of 'numpy.ufunc' objects> in __signature__ attribute
```

seberg: Yeah, one silly work-around might be to use `__text_signature__` but I guess that also has some limitations, which isn't nice.

Unless we know this warning originates in sphinx or so (which I doubt) there is no easy work-around, I suppose.  (One can do `isinstance(obj, type) and hasattr(attribute, "__get__")` to guess at this.)

But the best solution is probably a HeapType there might even be some draft PR around for that (this would be nice anyway, because it unlocks a few other things). But I am not quite sure how trivial this conversion is.

jorenham: `__text_signature__` wouldn't work because that would apply to the entire `ufunc` class,

seberg: Ah, yes, but if you _also_ attached a `__text_signature__` explicitly the way you did it here, then you should find that one for `np.add.__text_signature__` and you might just still find the other one for `np.ufunc.__text_signature__`.

(I am not sure about this, it may depend a bit on how all of this is implemented within CPython.)

mhvk: Hmm, the documentation at https://docs.python.org/3/library/inspect.html#inspect.signature is decidedly unhelpful:
> CPython implementation detail: If the passed object has a __signature__ attribute, we may use it to create the signature. The exact semantics are an implementation detail and are subject to unannounced changes. Consult the source code for current semantics.

mhvk: If it is a matter of class vs instance, is it possible to check that in a getter with the `self` that is passed in? (in the code above, typically called `PyUFuncObject *ufunc`?).  [Sorry if this is completely off base!]

jorenham: > is it possible to check that in a getter with the `self` that is passed in? (in the code above, typically called `PyUFuncObject *ufunc`?)

I also tried that before, but alas

seberg: @mhvk yes that is possible in Python but not on a static type. For that to work, it has to be defined as an attribute on the class/type itself.
But, static types don't have a place to add explicit class attributes (as far as I can tell).  HeapTypes do have space for it in their `__dict__` (you can add properties, but not custom ones -- a metaclass would work too, but that is way too heavy machinery).

seberg: :facepalm: type objects do have a `tp_dict`, it's just "disabled" by it being immutable by default, let's see if that works.

seberg: That indeed works just fine: https://github.com/numpy/numpy/pull/30270

PR #30214, opened 2025-11-14
BUG: Fix build on s390x with clang

yasiribmcon:
For clang 20+ (_VEC_ >= 10305), fallback check relied on macro definitions for vector operations, but clang provides regular functions instead. This caused the fallback to redefine them as macros, overriding the actual vector operation logic.


charris:
Thanks @yasiribmcon .

Issue #24410, opened 2023-08-13
Restructuring GitHub Actions CI jobs


rgommers:
These CI jobs are hard to navigate with several levels of indirection (a workflow file, then an action file, then two `tools/travis-*.sh` scripts), and now that we have to move a lot of jobs over to Meson it's a good time to clean up. The `build_test.yml` is particularly messy, and contains 22 jobs which is too much (quite a few are unnecessary too).

A few thoughts:
- The `windows_meson.yml` and `windows_clangcl.yml` can be merged, and named `windows.yml`
- It's useful to keep unusual / niche platform jobs in separate files. So let's keep `emscripten.yml`, `cygwin.yml`, `linux_musl.yml` and `linux_compiler_sanitizers.yml`.
- Let's break up `build_test.yml`:
    - `linux_simd.yml`: all the SIMD related jobs (~8 jobs) and the `old_gcc` one
    -  `linux_blas.yml`: all the BLAS/LAPACK related variations we're testing (should end up with ~6 jobs)
    - `linux.yml`: testing across Python versions (PyPy, python-dbg, a pre-release Python in summer time), build-via-sdist, run benchmarks, measure code coverage, and other build options like relaxed-strides. This will also fold in `linux_meson.yml`.
- The SIMD tests are the only ones that may use a `.sh` script, for the checks that are now in `travis-test.sh`. For the rest, those shell script are just hard to read cruft. 
- I think we'd like a summary of what the larger files to at the top. In particular for SIMD, BLAS and the main `linux.yml` ones, that would be really helpful. 
- No more matrix'ed tests on a single variable, that's too much repetition of the exact same thing. We must be testing Python 3.9 + OpenBLAS + ubuntu-latest with its default GCC version more than 5 times by now, and it doesn't do much except heat the planet.

rgommers:
> One downside of breaking up the jobs is how to do a smoke test run before starting many other runs

That's not too difficult, I think smoke tests can be run for the 3 main files (simd, blas, linux). I had thought of that already, will keep it as a required feature for now.

> I don't know how we could pause all other CI runs until it passes, extra points for pausing cirrus travis, and azure

Controls between different CI providers are a pain, and I think an orthogonal topic to GHA changes. I'd much prefer not to think about that here.
 

> There is a persistent failure in travis, we should mark that as "not fail".

We should turn TravisCI off for anything but SIMD PRs.

rgommers:
We discussed this in the community meeting today. Overall sentiment was that:
[...]

PR #30222, opened 2025-11-15
CI: remove (mainly windows) jobs from Azure pipelines

andyfaff:
Removes two windows jobs from Azure Pipelines, transferring them to GHA.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -19,7 +19,7 @@ permissions:
 
 jobs:
   clangcl_python64bit_openblas32:
-    name: x86-64, LP64 OpenBLAS
+    name: Clang-cl, x86-64, fast, openblas32
     runs-on: windows-2022
     # To enable this job on a fork, comment out:
     if: github.repository == 'numpy/numpy'
@@ -69,7 +69,7 @@ jobs:
 
   #=======================================================================================
   msvc_python32bit_no_openblas:
-    name: MSVC, ${{ matrix.architecture }} Python , no BLAS
+    name: MSVC, ${{ matrix.architecture }}, fast, no BLAS
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
@@ -122,18 +122,16 @@ jobs:
 
   #=======================================================================================
   msvc_python64bit_openblas:
-    name: MSVC, ${{ matrix.TEST_MODE }}, BLAS${{ matrix.BLAS }}
+    name: MSVC, x86_64, ${{ matrix.TEST_MODE }}, openblas${{ matrix.BLAS }}
     runs-on: windows-2022
     strategy:
       fail-fast: false
       matrix:
         include:
-          - architecture: x64
-            BLAS: 64
+          - BLAS: 64
             TEST_MODE: full
-            pyver: '3.12'
-          - architecture: x64
-            BLAS: 32
+            pyver: '3.14'
+          - BLAS: 32
             TEST_MODE: fast
             pyver: '3.11'
 
@@ -151,20 +149,13 @@ jobs:
         uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v6.0.0
         with:
           python-version: ${{ matrix.pyver }}
-          architecture: ${{ matrix.architecture }}
 
-      - name: Setup MSVC
-        uses: bus1/cabuild/action/msdevshell@e22aba57d6e74891d059d66501b6b5aed8123c4d  # v1
-        with:
-          architecture: ${{ matrix.architecture }}
-
-      - name: pkg-config, rtools
+      - name: pkg-config
         run: |
           choco install -y --stoponfirstfailure --checksum 6004DF17818F5A6DBF19CB335CC92702 pkgconfiglite
 
       - name: Dependencies
         run: |
-          python -m pip install --upgrade pip wheel
           python -m pip install -r requirements/test_requirements.txt
 
       - name: Build and install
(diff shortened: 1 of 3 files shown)


rgommers:
Sounds great, finishes off the migration we discussed in gh-24410. Thanks for working on this @andyfaff! Having a closer look now.

azure-pipelines.yml
```diff
@@ -37,29 +37,6 @@ stages:
   dependsOn: Check
   jobs:
 
-  - job: Lint
```
rgommers, line 40: This was indeed fairly pointless duplication of a lint job, and the `Python.h` check fits better on Linux, so this can be removed.

.github/workflows/windows.yml
```diff
@@ -117,3 +119,68 @@ jobs:
...
+        with:
+          architecture: ${{ matrix.architecture }}
+
+      - name: pkg-config, rtools
+        run: |
+          choco install -y --stoponfirstfailure --checksum 6004DF17818F5A6DBF19CB335CC92702 pkgconfiglite
+
+      - name: Dependencies
+        run: |
+          python -m pip install --upgrade pip wheel
```
rgommers: We shouldn't need `wheel` for anything anymore. And `pip` is already at 25.3; any recent version works for us. So this line can be dropped.

.github/workflows/windows.yml
```diff
@@ -117,3 +119,68 @@ jobs:
...
+
+  #=======================================================================================
+  msvc_python64bit_openblas:
+    name: MSVC, ${{ matrix.TEST_MODE }}, BLAS${{ matrix.BLAS }}
+    runs-on: windows-2022
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - architecture: x64
```
rgommers: Architecture can be dropped here, since it's twice the same. It doesn't need to be specified to `setup-python` either, it's the default. If you want to specify it anyway, you can use the `'x64'` string directly.

.github/workflows/windows.yml
```diff
@@ -117,3 +119,68 @@ jobs:
...
+      - name: Setup Python
+        uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v6.0.0
+        with:
+          python-version: ${{ matrix.pyver }}
+          architecture: ${{ matrix.architecture }}
+
+      - name: Setup MSVC
+        uses: bus1/cabuild/action/msdevshell@e22aba57d6e74891d059d66501b6b5aed8123c4d  # v1
+        with:
+          architecture: ${{ matrix.architecture }}
```
rgommers: This whole action isn't needed since x64 is the default architecture and `--vsenv` is already passed further down.

.github/workflows/windows.yml
```diff
@@ -117,3 +119,68 @@ jobs:
...
+    name: MSVC, ${{ matrix.TEST_MODE }}, BLAS${{ matrix.BLAS }}
+    runs-on: windows-2022
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - architecture: x64
+            BLAS: 64
+            TEST_MODE: full
+            pyver: '3.12'
```
rgommers: Can we make this `3.14`? If we test two Python versions, it should be minimum and maximum supported ones for better coverage.

.github/workflows/windows.yml
```diff
@@ -117,3 +119,68 @@ jobs:
...
+          python -m pip install -r requirements/test_requirements.txt
+
+      - name: Build and install
+        run: |
+          pip install -r requirements/ci_requirements.txt
+          spin config-openblas --with-scipy-openblas=${{ matrix.BLAS }}
+          $env:PKG_CONFIG_PATH="$pwd/.openblas"
+          python -m pip install . -v -Csetup-args="--vsenv"
+
+      - name: Run test suite (fast)
```
rgommers: This should be `${{ matrix.TEST_MODE }}` instead of `fast`

.github/workflows/windows.yml
```diff
@@ -117,3 +119,68 @@ jobs:
...
+        run: |
+          cd tools
+          # Get a gfortran onto the path for f2py tests
+          $env:PATH = "c:\\rtools45\\x86_64-w64-mingw32.static.posix\\bin;$env:PATH"
+          If ( $env:TEST_MODE -eq "full" ) {
```
rgommers: This isn't working in the BLAS64 job it looks like, the test suite is still fast rather than full. There are 119 tests marked as slow (says a quick `rg "mark.slow" | wc -l`) and the difference in skipped tests is only a handful in the CI logs, and runtime is almost identical.

I think this should be `${{ matrix.TEST_MODE }}`. Did a quick docs search to confirm: Azure puts matrix values in `$env` by default while GHA does not.

.github/workflows/windows.yml
```diff
@@ -117,3 +119,68 @@ jobs:
...
+          architecture: ${{ matrix.architecture }}
+
+      - name: pkg-config, rtools
+        run: |
+          choco install -y --stoponfirstfailure --checksum 6004DF17818F5A6DBF19CB335CC92702 pkgconfiglite
+
+      - name: Dependencies
+        run: |
+          python -m pip install --upgrade pip wheel
+          python -m pip install -r requirements/test_requirements.txt
```
rgommers, line 159: It may be nice to pass `-Csetup-args=-Dallow-noblas=false` here to prevent regressions. `scipy-openblas` has been known to go silently missing on occasion.

rgommers (changes requested):
This looks pretty good, thanks @andyfaff! And the jobs are way faster on GHA than they were on Azure.

I made some inline comments that would be good to address. In addition, it'd be nice to make the job names more consistent, because it's now hard to spot the difference between the two platforms. I'd propose a fixed pattern: compiler, platform, fast/full, BLAS. Which would yield:

- Clang-cl, x86-64, fast, openblas32
- MSVC, x86, fast, no BLAS
- MSVC, arm64, fast, no BLAS
- MSVC, x86-64, full, openblas64
- MSVC, x86-64, fast, openblas32

andyfaff:
I'm going to close this PR now to do further work on my fork. I'll reopen when it's ready to be evaluated again.

azure-pipelines.yml
andyfaff: This is being done in #30225

.github/workflows/windows.yml
andyfaff: Don't you mean on L175, when we build and install?

rgommers: Oh yes that is what I meant.

.github/workflows/windows.yml
```diff
@@ -117,3 +119,59 @@ jobs:
...
+      - name: Dependencies
+        run: |
+          python -m pip install -r requirements/test_requirements.txt
+
+      - name: Build and install
+        run: |
+          pip install -r requirements/ci_requirements.txt
+          spin config-openblas --with-scipy-openblas=${{ matrix.BLAS }}
+          $env:PKG_CONFIG_PATH="$pwd/.openblas"
+          python -m pip install . -v -Csetup-args="--vsenv" -Csetup-args="-Dallow-noblas=true"
```
rgommers: This should be `allow-noblas=false` not `true`. `true` means "it's allowed to not link any BLAS library", `false` will make the build fail if BLAS goes missing.

```suggestion
          python -m pip install . -v -Csetup-args="--vsenv" -Csetup-args="-Dallow-noblas=false"
```

rgommers (approved):
Thanks for the update Andrew, LGTM now modulo one very minor last thing.

rgommers:
In it goes, thanks Andrew!

PR #30224, opened 2025-11-15
TYP: ``ravel``: less awkward return types

jorenham:
this is a backwards-compatible change that makes the return types of `np.ravel` for nested sequences of python scalars less annoying, while still avoiding overlapping overloads (see https://github.com/numpy/numpy/issues/27032#issuecomment-2278958251 for details).


rgommers (approved):
Thanks @jorenham. Seems fine to me. Given the discussion in gh-27032, pinging @Jacob-Stevens-Haas in case he has an opinion.

charris:
Thanks Joren.

Jacob-Stevens-Haas:
Thanks for suggesting me - it's nice to refocus back on typing problems.  While I didn't quite understand that discussion before, I think I understand now.  Why is `| Any` is more user friendly than `| np.bool_ | np.int_`, etc.?  Is it just because the unions are verbose?

If brevity is the concern, could we return standard unions, e.g. `PromotesToInt_`, `PromotesToFloat64` or `PromotesToCDouble`, which seems preferable to returning `Any`. Or does this problem occur beyond just numeric scalars, e.g. nested dtypes?

jorenham:
> Why is `| Any` is more user friendly than `| np.bool_ | np.int_`, etc.? Is it just because the unions are verbose?

If you have some `x: np.float64 | np.int_ | np.bool_`, while you know for certain that at runtime it's always a `np.float64`, and you pass it to e.g. some `f(y: np.float64)`, type-checkers won't let you, which is mildly infuriating .
But if `x: np.float64 | Any` instead, `f(x)` is perfectly fine. 

In typeshed this is used quite frequently (a bit *too* frequently if you ask me), so much so that they frivolously call it ["the Any trick"](https://typing.python.org/en/latest/guides/writing_stubs.html#the-any-trick).

Jacob-Stevens-Haas:
Ok this makes sense.  It seems like that link suggests, where the type system can't return something specific enough, we still want to allow the user some benefits of type checking without requiring them to explicitly check types.  Is that an accurate reading of "In cases where a function or method can return None, but where forcing the user to explicitly check for None can be detrimental..."?

jorenham:
> Is that an accurate reading of "In cases where a function or method can return None, but where forcing the user to explicitly check for None can be detrimental..."?

Yea something like that. From what I can tell, the philisophy of the typeshed maintainers (of which many have contributed to the typing spec) is that users should not have to change their code when adopting Python typing, as that wouldn't be user-friendly, and would increase the amount of reported issues [^1]. But personally, I think that higher priority should be given to preventing bugs, i.e. type-safety. If that requires users to have to change their code, then that's a *good* thing, as it makes it less prone to bugs.

A slightly less type-unsafe solution that has been discussed before, would be to introduce a new `AnyOf[...]` special form. It can be thought of as an `Any`-ish union, where `AnyOf[A, B]` accepts both `A` and `B`, and is itself assignable to only `A` or `B`. So if `f() -> AnyOf[int, None]`, it'd be valid to write `x: int = f()` or `x: None = f()`, but `x: bool = f()` and `x: str | None = f()` are not. Note that `AnyOf[<all possible types>]` is then equivalent to `Any`, so theoretically it generalizes `Any`.

[^1]: I base this on my experience in interacting with them, and some other "typing experts" I've talked to about this agree. As far as I'm aware, this hasn't been explicitly confirmed by  by them.

Jacob-Stevens-Haas:
Thanks @jorenham that context and the explanation of `AnyOf` really helps

PR #30225, opened 2025-11-15
CI: remove azure pipelines runs

andyfaff:
Removes azure pipelines runs by removing last job (i686 linux).

Follows (and is branched from) #30222. That PR should be merged first. 

Note, in this PR I'm checkout out code manually, rebasing PRs onto HEAD. This is because `actions/checkout` isn't available in a container.


andyfaff:
Alternatively, just merge this PR instead of 30222. Let me know if commits need to be squashed.

.github/workflows/linux.yml
```diff
@@ -362,3 +366,51 @@ jobs:
...
+      run: |
+        python3.11 -m venv venv
+        source venv/bin/activate
+        pip install --upgrade pip
+        pip install -r requirements/ci32_requirements.txt
+        pip install -r requirements/test_requirements.txt
+        
+        export CFLAGS="-msse2 -std=c99 -UNDEBUG"
+        export F77=gfortran-5
+        export F90=gfortran-5
```
rgommers, line 407: `F77` and `F90` do nothing, that's a leftover from the `distutils` builds - lines can be removed. We don't even need a Fortran compiler to build.

.github/workflows/linux.yml
```diff
@@ -362,3 +366,51 @@ jobs:
...
+
+    - name: build
+      run: |
+        python3.11 -m venv venv
+        source venv/bin/activate
+        pip install --upgrade pip
+        pip install -r requirements/ci32_requirements.txt
+        pip install -r requirements/test_requirements.txt
+        
+        export CFLAGS="-msse2 -std=c99 -UNDEBUG"
```
rgommers, line 405: None of these flags are necessary anymore, all already included in the build.

.github/workflows/linux.yml
```diff
@@ -362,3 +366,51 @@ jobs:
...
+        export F77=gfortran-5
+        export F90=gfortran-5
+        spin config-openblas --with-scipy-openblas=32
+        export PKG_CONFIG_PATH=/numpy/.openblas
+        python -m pip install . -v
```
rgommers, line 410: Same as in my other review, I suggest using `-Dallow-noblas=false` here.

.github/workflows/linux.yml
```diff
@@ -362,3 +366,51 @@ jobs:
         rm -rf build-install
         ./vendored-meson/meson/meson.py install -C build --destdir ../build-install --tags=runtime,python-runtime,devel
         python tools/check_installed_files.py $(find ./build-install -path '*/site-packages/numpy') --no-tests
+
+
+
+  Linux_Python_311_32bit_full_with_asserts:
+    name: i686, cp311, full, with asserts
```
rgommers, line 373: The "with asserts" can be removed, that's a leftover from long ago and doesn't mean anything anymore.

.github/workflows/linux.yml
```diff
@@ -362,3 +366,51 @@ jobs:
...
+    needs: [smoke_test]
+    runs-on: ubuntu-latest
+    container:
+       # There are few options for i686 images at https://quay.io/organization/pypa,
+       # use the glibc2.17 one (manylinux2014)
```
rgommers, line 378: It doesn't matter much, but we did drop `manylinux2014` support already, so if there's a `2_28` image, that may be slightly better.

rgommers (commented):
Thanks @andyfaff! Looks much better than on Azure. It'd be good to merge gh-30222 first and keep this one separate I think; they're Windows vs. Linux and easier to review if they're a bit smaller.

.github/workflows/linux.yml
mattip: Maybe `spin build --with-scipy-openblas=32 -- -Dallow-noblas=false -Dpkg_config_path=/numpy/.openblas`
and then 
`spin test` below

.github/workflows/windows.yml
```diff
@@ -117,3 +119,68 @@ jobs:
...
+        run: |
+          python -m pip install --upgrade pip wheel
+          python -m pip install -r requirements/test_requirements.txt
+
+      - name: Build and install
+        run: |
+          pip install -r requirements/ci_requirements.txt
+          spin config-openblas --with-scipy-openblas=${{ matrix.BLAS }}
+          $env:PKG_CONFIG_PATH="$pwd/.openblas"
+          python -m pip install . -v -Csetup-args="--vsenv"
```
mattip, line 175:
```suggestion
          spin build --with-scipy-openblas=32 -- -Dallow-noblas=false -Dpkg_config_path=/numpy/.openblas
```

and then `spin test --` instead of `python -m pytest` below?

.github/workflows/linux.yml
rgommers: I like using `pip install` in a couple of jobs, since (a) it uses fewer dependencies and (b) tests that using a scipy-openblas wheel without vendoring it in works as it should.

andyfaff:
I just squashed + force pushed to make commits a bit clearer from my end. I'm going to close this PR now to do further work on my fork. I'll reopen when it's ready to be evaluated again.

.github/workflows/windows.yml
andyfaff: Leaving as `pip install .`

PR #30231, opened 2025-11-15
TYP: Shape-typing in ``lib._twodim_base_impl``

jorenham:
This adds shape-typing support to the following functions:

- `np.diag`
- `np.diagflat`
- `np.eye`
- `np.fliplr`
- `np.flipud`
- `np.tri`
- `np.triu`
- `np.tril`
- `np.vander`
- `np.histogram2d`
- `np.mask_indices`
- `np.tril_indices`
- `np.tril_indices_from`
- `np.triu_indices`
- `np.triu_indices_from`

If the output shape depends on that of the input, and the shape-type of the input type is unknown, then the output shape-type will also be unknown (i.e. the "graudual" shape-type  `tuple[Any, ...]` is used).


charris:
Needs rebase.

charris:
Thanks Joren.

PR #30232, opened 2025-11-16
CI: removes azure pipelines

andyfaff:
I force pushed to #30225 whilst closed, which meant that I wasn't able to reopen it. Sorry for the noise.


.github/workflows/linux.yml
```diff
@@ -362,3 +366,47 @@ jobs:
...
         python tools/check_installed_files.py $(find ./build-install -path '*/site-packages/numpy') --no-tests
+
+
+  Linux_Python_311_32bit_full:
+    name: i686, cp311, full, with asserts
+    needs: [smoke_test]
+    runs-on: ubuntu-latest
+    container:
+       # There are few options for i686 images at https://quay.io/organization/pypa,
+       # use the glibc2.17 one (manylinux2014)
```
mattip:
```suggestion
       # use the glibc2.28 one
```

.github/workflows/linux.yml
```diff
@@ -362,3 +366,47 @@ jobs:
...
+      run: |
+        python3.11 -m venv venv
+        source venv/bin/activate
+        pip install --upgrade pip
+        pip install -r requirements/ci32_requirements.txt
+        pip install -r requirements/test_requirements.txt
+
+        spin config-openblas --with-scipy-openblas=32
+        export PKG_CONFIG_PATH=/numpy/.openblas
+        python -m pip install . -v -Dallow-noblas=false
```
mattip:
```suggestion
        python -m pip install . -v -Csetup-args="-Dallow-noblas=true"
```

mattip:
The linux pypy run failed, something about underflow in a check of the `conj` ufunc. I reran it, let's see if it is transient.

rgommers:
With gh-30222 merged, this one can be updated now.

.github/workflows/linux.yml
```diff
@@ -366,3 +366,47 @@ jobs:
...
+      run: |
+        python3.11 -m venv venv
+        source venv/bin/activate
+        pip install --upgrade pip
+        pip install -r requirements/ci32_requirements.txt
+        pip install -r requirements/test_requirements.txt
+
+        spin config-openblas --with-scipy-openblas=32
+        export PKG_CONFIG_PATH=/numpy/.openblas
+        python -m pip install . -v -Csetup-args="-Dallow-noblas=true"
```
rgommers: This should say `false`. And it would have caught the problem that I was suggested it could, since the build log shows:
```
Run-time dependency scipy-openblas found: NO (tried pkgconfig)
```
It's not clear from the Azure logs whether `scipy-openblas32` was successfully found in the Azure version of this job unfortunately, because the build log is hidden there. I suspect also not; the path `/numpy/.openblas` looks wrong, it shouldn't start with a slash.

rgommers (changes requested):
Thanks @andyfaff. This is almost ready, just one issue with BLAS detection. 

The job is also quite a bit faster; runtime goes from 13 to 8 minutes, which is nice.

.github/workflows/linux.yml
```diff
@@ -366,3 +366,47 @@ jobs:
         rm -rf build-install
         ./vendored-meson/meson/meson.py install -C build --destdir ../build-install --tags=runtime,python-runtime,devel
         python tools/check_installed_files.py $(find ./build-install -path '*/site-packages/numpy') --no-tests
+
+
+  Linux_Python_311_32bit_full:
+    name: i686, cp311, full, with asserts
```
rgommers:
```suggestion
    name: i686, cp311, full
```

rgommers (approved):
LGTM now, thanks @andyfaff. I'll apply one last tiny tweak to the job name, then merge.

PR #30239, opened 2025-11-16
DOC: Correct an equation error in `numpy.random.Generator.pareto`

star1327p:
Correct an equation error in `numpy.random.Generator.pareto`
https://numpy.org/devdocs/reference/random/generated/numpy.random.Generator.pareto.html

image

Replace "x + 1" with "( x + 1 )".

### DIFF
--- a/numpy/random/_generator.pyx
+++ b/numpy/random/_generator.pyx
@@ -2083,7 +2083,7 @@ cdef class Generator:
         -----
         The probability density for the Pareto II distribution is
 
-        .. math:: p(x) = \\frac{a}{{x+1}^{a+1}} , x \ge 0
+        .. math:: p(x) = \\frac{a}{{(x+1)}^{a+1}} , x \ge 0
 
         where :math:`a > 0` is the shape.
 
--- a/numpy/random/_generator.pyx
+++ b/numpy/random/_generator.pyx
@@ -2083,7 +2083,7 @@ cdef class Generator:
         -----
         The probability density for the Pareto II distribution is
 
-        .. math:: p(x) = \\frac{a}{{(x+1)}^{a+1}} , x \ge 0
+        .. math:: p(x) = \\frac{a}{(x+1)^{a+1}} , x \ge 0
 
         where :math:`a > 0` is the shape.
 


numpy/random/_generator.pyx
```diff
@@ -2083,7 +2083,7 @@ cdef class Generator:
         -----
         The probability density for the Pareto II distribution is
 
-        .. math:: p(x) = \\frac{a}{{x+1}^{a+1}} , x \ge 0
+        .. math:: p(x) = \\frac{a}{{(x+1)}^{a+1}} , x \ge 0
```
jorenham: It looks like the wrong braces were used before:

```suggestion
        .. math:: p(x) = \\frac{a}{(x+1)^{a+1}} , x \ge 0
```

jorenham: And while we're at it; the $x \ge 0$ could use a bit of breathing room;

```suggestion
        .. math:: p(x) = \\frac{a}{{(x+1)}^{a+1}} ,\\ x \ge 0
```

before

$$
\frac{a}{{(x+1)}^{a+1}} , x \ge 0
$$

after


$$
\frac{a}{{(x+1)}^{a+1}} ,\ x \ge 0
$$

I'm not sure if the escaped `\\` is needed, but it probably wouldn't hurt

jorenham (approved):
Nitpicks aside; this was indeed incorrect before, and this fixes it.

Fun fact; this isn't the Pareto distribution, because the "Pareto distribution" usually [^0] refers to the type I Pareto distribution, while this is the [type II Pareto distribution](https://en.wikipedia.org/wiki/Pareto_distribution#Pareto_types_I%E2%80%93IV), often referred to as the [Lomax distribution](https://en.wikipedia.org/wiki/Lomax_distribution) [^1].

But IMO Pareto II is better than Pareto I, so I have no problem this redefinition :stuck_out_tongue: 

[^0]: with "usually" I mean the particular branch of scientific literature that I have read (related to [L-moments](https://en.wikipedia.org/wiki/L-moment), mostly), so it could be that it's different in other fields I don't know about.
[^1]: Technically speaking type II isn't Lomax because Lomax doesn't have a $\mu$ (location) parameter (and that extra parameter is why it's the "type II" Pareto distribution, as in "type I plus one parameter"). But here that doesn't really matter because we standardize and consider it as a [location-scale distribution](https://en.wikipedia.org/wiki/Location%E2%80%93scale_family) (i.e. by allowing linear transformations by shifting and scaling). And if we set $\mu=0$ for a type II Pareto distribution, then that's exactly the Lomax distribution.

mattip:
Note there are two different `pareto` functions: the [legacy one](https://numpy.org/doc/stable/reference/random/generated/numpy.random.pareto.html#numpy.random.pareto), which is the traditional Pareto I, and the [new generator one](https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.pareto.html), Pareto II which is discussed here. Confusing.

In any case this is a good fix, thanks @star1327p

rkern:
No, the legacy is also Pareto II (as documented); the implementations are the same (only changing `exp(.) - 1` to `expm1(.)` for numerical purposes). The explanatory text for the legacy one might be wrong, though. It was written by volunteers in the Subversion Days before PR reviews and might be describing Pareto I instead.

mattip:
Hah! I opened #30250

PR #23513, opened 2023-03-31
ENH: speed up einsum with optimize using batched matmul

jcmgray:
This PR enables dispatching *all* pairwise calls in `einsum` when `optimize=` is turned on to batched matrix multiply (or non-batched `matmul` and `multiply` as special cases). For a short explanation of the implementation see here - https://github.com/jcmgray/einsum_bmm.

Currently falling back to `c_einsum` for these, when there exist batch indices, induces potentially many orders of magnitude slow downs (#22604, e.g. 7000x), particularly in the case of large or high dimensional tensors.

After fairly extensive benchmarking (below), I *believe* that this should be a uniform improvement across essentially all the cases when one would use `optimize=True`. This is partly enabled by the modern performance of matmul which in the past was not always faster than `einsum`. However, I know `einsum` has an enormous number of uses and I'd gladly accept more suggestions for benchmarks, and would understand people's hesitation! 

(@dgasmith, @seberg)

## Other notes:
- This replaces all the `can_blas` logic and `tensordot` calls, and simplifies everything into the `bmm_einsum`, which encapsulates how `tensordot` is implemented
- When there are no contracted indices, `numpy.multiply` is used for slightly better performance
- The current implementation can simply pass `out` and other kwargs on to `mutiply` and `matmul`, unlike tensordot, meaning `einsum('ij,jk->ik', x, y, optimize=True, out=z)` should write directly to `z`
- The fusing allows `einsum(eq, ..., optimize=True)` to be used when there are more than 32 indices involved (#5744).
- The low overhead of the actuall `bmm_einsum` implementation comes from caching what operations to use based on `(eq, x.shape, y.shape)`, using `@functools.lru_cache(2**12)`.
- the performance broadly speaking is brought in line with `torch.einsum`

## TODO: 
- [ ] update docs?
- [ ] decide cache size?
- [x] should I lint the existing `einsumfunc.py` code?

## Benchmarks

In the following the kernels are the following:

- `no optimize`: the base `np.einsum(eq, x, y)` i.e. `c_einsum`
- `optimize + dot`: the current `np.einsum(eq, x, y, optimize=True)` which calls `tensordot` where possible, and also induces some overhead from the potential path optimization (even though there are only two terms in the following, the equation must be checked and parsed etc.)
- `optimize + bmm`: the proposed `np.einsum(eq, x, y, optimize=True)` which calls batched matmul, and still induces some overhead from the potential path optimization
[...]

PR #30242, opened 2025-11-17
BUG: fix einsum ``optimize=True`` parsing error

jcmgray:
This fixes some edge case parsing errors (https://github.com/scipy/scipy/issues/23997) introduced in #23513, and adds new tests that catch such cases.

### DIFF
--- a/numpy/_core/einsumfunc.py
+++ b/numpy/_core/einsumfunc.py
@@ -988,15 +988,11 @@ def _parse_eq_to_batch_matmul(eq, shape_a, shape_b):
     if len(b_term) != len(shape_b):
         raise ValueError(f"Term '{b_term}' does not match shape {shape_b}.")
 
-    bat_inds = []  # appears on A, B, O
-    con_inds = []  # appears on A, B, .
-    a_keep = []  # appears on A, ., O
-    b_keep = []  # appears on ., B, O
     sizes = {}
     singletons = set()
 
-    # parse left term
-    seen = set()
+    # parse left term to unique indices with size > 1
+    left = dict()
     for ix, d in zip(a_term, shape_a):
         if d == 1:
             # everything (including broadcasting) works nicely if simply ignore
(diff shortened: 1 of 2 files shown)


numpy/_core/einsumfunc.py
```diff
@@ -988,63 +988,62 @@ def _parse_eq_to_batch_matmul(eq, shape_a, shape_b):
...
-    con_inds = []  # appears on A, B, .
-    a_keep = []  # appears on A, ., O
-    b_keep = []  # appears on ., B, O
     sizes = {}
     singletons = set()
 
-    # parse left term
-    seen = set()
+    # parse left term to unique indices with size > 1
+    left = dict()
```
jorenham:
```suggestion
    left = {}
```

there's a ruff rule for this I believe, although it might not be enabled (yet)

numpy/_core/einsumfunc.py
```diff
@@ -988,63 +988,62 @@ def _parse_eq_to_batch_matmul(eq, shape_a, shape_b):
...
-                bat_inds.append(ix)
-            else:
-                con_inds.append(ix)
-        elif ix in out:
-            a_keep.append(ix)
-
-    # parse right term
-    seen.clear()
+    # parse right term to unique indices with size > 1
+    right = dict()
```
jorenham:
```suggestion
    right = {}
```

numpy/_core/einsumfunc.py
```diff
@@ -988,63 +988,62 @@ def _parse_eq_to_batch_matmul(eq, shape_a, shape_b):
...
         if d == 1:
             # everything (including broadcasting) works nicely if simply ignore
             # such dimensions, but we do need to track if they appear in output
             # and thus should be reintroduced later
             singletons.add(ix)
             continue
-
-        # set or check size
         if sizes.setdefault(ix, d) != d:
+            # set or check size
```
jorenham:
```suggestion
            # set and check size
```

numpy/_core/einsumfunc.py
```diff
@@ -988,63 +988,62 @@ def _parse_eq_to_batch_matmul(eq, shape_a, shape_b):
...
-                b_keep.append(ix)
+                bat_inds.append(ix)
+            else:
+                con_inds.append(ix)
+        elif ix in out:
+            a_keep.append(ix)
+    # now only indices unique to right remain
+    for ix in right:
+        if ix in out:
+            b_keep.append(ix)
```
jorenham, line 1046: this could be simplified a bit now:

```suggestion
    b_keep.extend(set(out) & right.keys())
```

Or maybe it's an idea to define something like an `outset = set(out)` earlier on, so that all other `_ in out` O(n) calls can be replaced with `_ in outset` O(1) calls? Although I wonder if it'll actually be faster in practice :thinking: 

(feel free to ignore if you don't like this btw)

numpy/_core/tests/test_einsum.py
```diff
@@ -1119,6 +1119,41 @@ def test_output_order(self):
...
+        eq = "jhcabhijaci,dfijejgh->fgje"
+        shapes = (
+            (1, 1, 1, 1, 3, 1, 1, 1, 1, 1, 1),
+            (3, 1, 3, 1, 1, 1, 1, 2),
+        )
+        arrays = [np.random.rand(*shape) for shape in shapes]
+        self.optimize_compare(eq, operands=arrays)
+
+        eq = "baegffahgc,hdggeff->dhg"
+        shapes = ((2, 1, 4, 1, 1, 1, 1, 2, 1, 1), (1, 1, 1, 1, 4, 1, 1))
```
jorenham, line 1137: for consistency with line 1124

```suggestion
        shapes = (2, 1, 4, 1, 1, 1, 1, 2, 1, 1), (1, 1, 1, 1, 4, 1, 1)
```

and same story for the 4 similar cases below

or alternatively you could add some parentheses to 1124 (I have no preference)

jorenham (commented):
Thanks! 

I left some superficial nitpicks; hope you don't mind.

numpy/_core/einsumfunc.py
jcmgray: Yes it feels like sets should be the right approach but a few things make this faster in practice: 1) the order matters (so set operations aren't good) 2) all of these collections are typically only a few elements and 3) checking membership in a string is weirdly fast.

jcmgray:
One other thing that occurs to me is that the docstring for the `optimize` kwarg could possibly be updated to state that it turns on *both* intermediate optimization and *also* turns on performing these contractions via BLAS (itself requiring the path optimization, #30171). Maybe with special mention that if there are only two terms it is essentially a flag for using BLAS ('essentially' since there is also some single term simplifications).

mattip:
> One other thing that occurs to me is that the docstring for the optimize kwarg could possibly be updated...

I think this could go into a rewrite of the "Note" section about `optimize`. We should also explicitly mention when `optimize=False` is preferable. But this can go into a separate "DOC" PR where this one is a bugfix.

mattip:
Thanks @jcmgray.

rgommers:
Thanks @jcmgray for the quick fix! Confirmed that this resolved the issues that were visible in SciPy CI.

PR #30243, opened 2025-11-17
BUG: Add missing `PyErr_Occurred()` check to fast-path

seberg:
We could side-step this check unless it's a legacy ufunc, but there is little point as this is barely measureable.

Since someone would otherwise ask: No you can't move it into the legacy wrapping stuff because this has to happen at finalization when the GIL is held again.

### DIFF
--- a/numpy/_core/tests/test_umath.py
+++ b/numpy/_core/tests/test_umath.py
@@ -4893,6 +4893,15 @@ def test_bad_legacy_ufunc_silent_errors():
         ncu_tests.always_error.at(arr, [0, 1, 2], arr)
 
 
+def test_bad_legacy_unary_ufunc_silent_errors():
+    # Unary has a special scalar path right now, so test it explicitly.
+    with pytest.raises(RuntimeError, match=r"How unexpected :\)!"):
+        ncu_tests.always_error_unary(np.arange(3).astype(np.float64))
+
+    with pytest.raises(RuntimeError, match=r"How unexpected :\)!"):
+        ncu_tests.always_error_unary(1.5)
+
+
 @pytest.mark.parametrize('x1', [np.arange(3.0), [0.0, 1.0, 2.0]])
 def test_bad_legacy_gufunc_silent_errors(x1):
     # Verify that an exception raised in a gufunc loop propagates correctly.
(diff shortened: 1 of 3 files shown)


numpy/_core/tests/test_umath.py
```diff
@@ -1249,6 +1249,17 @@ def test_integer_to_negative_power(self):
             assert_raises(ValueError, np.power, one, b)
             assert_raises(ValueError, np.power, one, minusone)
 
+    def test_integer_to_negative_power_scalar(self):
```
WarrenWeckesser: This could use a brief comment to explain why `np.power(two, minusone)` is being called twice.  Seems a bit mysterious without it.

WarrenWeckesser (approved):
I confirmed that this fixes the SciPy issue.

Other than one suggestion for an added comment, looks good to me.

I'd double-check with @mhvk before merging.

numpy/_core/tests/test_umath.py
```diff
@@ -4893,6 +4904,15 @@ def test_bad_legacy_ufunc_silent_errors():
         ncu_tests.always_error.at(arr, [0, 1, 2], arr)
 
 
+def test_bad_legacy_unary_ufunc_silent_errors():
+    # Unary has a special scalar path right now, so test it explicitly.
+    with pytest.raises(RuntimeError, match=r"How unexpected :\)!"):
+        ncu_tests.always_error_unary(np.arange(3).astype(np.float64))
```
mhvk, line 4899: if you're editing anyway, just `np.arange(3.0)`? The `.astype()` is distracting here.

numpy/_core/tests/test_umath.py
seberg: Ah, I'll just delete the test... wanted to use it first, but it's nota unary function so doesn't take the fast path anyway.

mhvk (approved):
Nice that the test machinery was basically there!

Just a small comment, but actually not worth pushing for again...

numpy/_core/tests/test_umath.py
seberg: Ah true, could just be ones (copy pasted it).  It needs to be float64 for the ufunc to work, though.

seberg:
I'll just put this in with two approvals, it's a simple fix really.

PR #30247, opened 2025-11-18
DOC: Update wording in numpy.coremath

athurdekoos:
Simple documentation update. 

The following sentence was removed as it seems to have been left in by mistake during a previous update. 

> "The numpy core math library (``npymath``) is a first step in this direction."

### DIFF
--- a/doc/source/reference/c-api/coremath.rst
+++ b/doc/source/reference/c-api/coremath.rst
@@ -1,8 +1,7 @@
 NumPy core math library
 =======================
 
-The numpy core math library (``npymath``) is a first step in this direction. This
-library contains most math-related C99 functionality, which can be used on
+This library contains most math-related C99 functionality, which can be used on
 platforms where C99 is not well supported. The core math functions have the
 same API as the C99 ones, except for the ``npy_*`` prefix.
 


mattip:
Thanks @athurdekoos. For future doc PRs, if you add `[skip actions][skip cirrus]` to the commit message, it will only run the doc build CI and save some CI cycles. It was not needed for this one since #30050 skips most CI if the change is only to `*.rst` files.

@jorenham it seems the [wheel builder CI run](https://github.com/numpy/numpy/actions/runs/19449114237) and the [mypy run](https://github.com/numpy/numpy/actions/runs/19449114259) do not have the stanza used in #30050 for the other CI runs. Is that on purpose?

jorenham:
> Is that on purpose?

Nope... I guessing that some stars are needed here then:
https://github.com/numpy/numpy/blob/7440ac96276f552de051eeebd92a7549dcce33f1/.github/workflows/mypy.yml#L20-L28

And I probably just forgot about the wheel builder workflow.

mattip:
@jorenham well, it is nice that the exceptions did mostly work and saved many CI runs

PR #30248, opened 2025-11-18
DOC: remove mention of 'skip azp' since we no longer use azure

mattip:
Small cleanup PR after removal of azure CI runs.


jorenham:
Thanks Matti

PR #30254, opened 2025-11-18
BUG: fix data race in ufunc scalar fast path

### LINKED ISSUE #30251: BUG: TSan race in ufunc scalar fast path
I think this PR introduced a race condition that we see in JAX's tsan CI (https://github.com/jax-ml/jax/actions/runs/19400717531):

```
WARNING: ThreadSanitizer: data race (pid=95179)
  Write of size 8 at 0x7fffb4060f78 by thread T3 (mutexes: read M0, write M1):
    #0 PyArrayIdentityHash_SetItem /__w/jax/jax/numpy/.mesonpy-tk47p_zd/../numpy/_core/src/common/npy_hashtable.cpp:224:20 (_multiarray_umath.cpython-314t-x86_64-linux-gnu.so+0x41dbbb) (BuildId: fe065841b00019c85589d862249e0c30081a9a3f)
    #1 promote_and_get_info_and_ufuncimpl(_tagPyUFuncObject*, tagPyArrayObject* const*, PyArray_DTypeMeta**, PyArray_DTypeMeta**, unsigned char) /__w/jax/jax/numpy/.mesonpy-tk47p_zd/../numpy/_core/src/umath/dispatching.cpp:871:17 (_multiarray_umath.cpython-314t-x86_64-linux-gnu.so+0x435463) (BuildId: fe065841b00019c85589d862249e0c30081a9a3f)
    #2 promote_and_get_info_and_ufuncimpl_with_locking(_tagPyUFuncObject*, tagPyArrayObject* const*, PyArray_DTypeMeta**, PyArray_DTypeMeta**, unsigned char) /__w/jax/jax/numpy/.mesonpy-tk47p_zd/../numpy/_core/src/umath/dispatching.cpp:1004:12 (_multiarray_umath.cpython-314t-x86_64-linux-gnu.so+0x432fd9) (BuildId: fe065841b00019c85589d862249e0c30081a9a3f)
    #3 promote_and_get_ufuncimpl /__w/jax/jax/numpy/.mesonpy-tk47p_zd/../numpy/_core/src/umath/dispatching.cpp:1101:22 (_multiarray_umath.cpython-314t-x86_64-linux-gnu.so+0x432fd9)
    #4 ufunc_generic_fastcall /__w/jax/jax/numpy/.mesonpy-tk47p_zd/../numpy/_core/src/umath/ufunc_object.c:4683:38 (_multiarray_umath.cpython-314t-x86_64-linux-gnu.so+0x4404df) (BuildId: fe065841b00019c85589d862249e0c30081a9a3f)
...

  Previous read of size 8 at 0x7fffb4060f78 by thread T2 (mutexes: read M0):
    #0 find_item(PyArrayIdentityHash const*, _object* const*) /__w/jax/jax/numpy/.mesonpy-tk47p_zd/../numpy/_core/src/common/npy_hashtable.cpp:79:13 (_multiarray_umath.cpython-314t-x86_64-linux-gnu.so+0x41ddd0) (BuildId: fe065841b00019c85589d862249e0c30081a9a3f)
    #1 PyArrayIdentityHash_GetItem /__w/jax/jax/numpy/.mesonpy-tk47p_zd/../numpy/_core/src/common/npy_hashtable.cpp:240:21 (_multiarray_umath.cpython-314t-x86_64-linux-gnu.so+0x41ddd0)
    #2 try_trivial_scalar_call /__w/jax/jax/numpy/.mesonpy-tk47p_zd/../numpy/_core/src/umath/ufunc_object.c:4340:22 (_multiarray_umath.cpython-314t-x86_64-linux-gnu.so+0x43f606) (BuildId: fe065841b00019c85589d862249e0c30081a9a3f)
    #3 ufunc_generic_fastcall /__w/jax/jax/numpy/.mesonpy-tk47p_zd/../numpy/_core/src/umath/ufunc_object.c:4425:13 (_multiarray_umath.cpython-314t-x86_64-linux-gnu.so+0x43f606)
    #4 ufunc_generic_vectorcall /__w/jax/jax/numpy/.mesonpy-tk47p_zd/../numpy/_core/src/umath/ufunc_object.c:4766:12 (_multiarray_umath.cpython-314t-x86_64-linux-gnu.so+0x43d60e) (BuildId: fe065841b00019c85589d862249e0c30081a9a3f)
    #5 _PyObject_VectorcallTstate /__w/jax/jax/cpython/./Include/internal/pycore_call.h:169:11 (python3.14+0x1fa17a) (BuildId: 010258c9a6c23915fd3f27ed01befca5ea57f252)
```

I suspect that's enough for you to figure it out but if you like I can make a better repro.

_Originally posted by @hawkinsp in https://github.com/numpy/numpy/issues/29819#issuecomment-3538920384_
            
See further replies in that thread for more details.

kumaraditya303:
Fixes https://github.com/numpy/numpy/issues/30251

This PR adds read lock while accessing the dispatch cache in `try_trivial_scalar_call` to fix the data race.


numpy/_core/src/common/npy_hashtable.cpp
```diff
@@ -240,3 +240,20 @@ PyArrayIdentityHash_GetItem(PyArrayIdentityHash *tb, PyObject *const *key)
     PyObject *res = find_item(tb, key)[0];
     return res;
 }
+
+NPY_NO_EXPORT PyObject *
+PyArrayIdentityHash_GetItemWithLock(PyArrayIdentityHash *tb, PyObject *const *key)
+{
+    PyObject *res;
+#ifdef Py_GIL_DISABLED
```
ngoldbaum: Can you move this ifdef over to the call site in `ufunc_object.c` and make the `else` case call `PyArrayIdentityHash_GetItem`? That makes it a little clearer there's only locking happening on the free-threaded build. It also follows the pattern in dispatching.cpp where we only call `promote_and_get_info_and_ufuncimpl_with_locking` on the free-threaded build.

I also think the logic in `PyArrayIdentityHash_GetItemWithLock` is shared with the first half of the implementation of `promote_and_get_info_and_ufuncimpl_with_locking`, so you can call `PyArrayIdentityHash_GetItemWithLock` and delete some duplicated code over in that function.

ngoldbaum (commented):
Nice, just a couple suggestions to make this a little better organized and easier for future readers to understand.

numpy/_core/src/common/npy_hashtable.cpp
kumaraditya303: Done, also since `PyArrayIdentityHash_GetItemWithLock` is only used on free-threading builds, it is now conditionally defined.

ngoldbaum (approved):
@mhvk how does this look to you?

ngoldbaum:
Can we also get an explicit test for this? I think `test_parallel_ufunc_execution` just needs a case that calls e.g. `np.add()` on a scalar with itself or a pair of scalars to trigger a TSan warning on an unpatched version of NumPy.

mhvk:
> Can we also get an explicit test for this? I think `test_parallel_ufunc_execution` just needs a case that calls e.g. `np.add()` on a scalar with itself or a pair of scalars to trigger a TSan warning on an unpatched version of NumPy.

Yes, but you need to be sure to use a unary ufunc to test it.

numpy/_core/src/common/npy_hashtable.cpp
```diff
@@ -240,3 +240,20 @@ PyArrayIdentityHash_GetItem(PyArrayIdentityHash *tb, PyObject *const *key)
     PyObject *res = find_item(tb, key)[0];
     return res;
 }
+
+#ifdef Py_GIL_DISABLED
```
mhvk, line 244: I wonder whether a separate function is needed at all. I.e., could one instead just define `PyArrayIdentityHash_GetItem` differently for the with and without GIL case? (have an `#ifdef !Py_GIL_DISABLED` above, and an `#else` here). That way, the rest of the code does not have to worry about locking at all. (I presume CPython does something similarly, since I don't think there is a separate `PyDict_GetItem` for the no-GIL case).

mhvk (commented):
Looks good, but as mentioned in-line, I wonder if there is anything wrong with my initial suggestion, to just defining `PyArrayIdentifyHash_GetItem` differently for the GIL and no-GIL cases. If that is in principle OK, one can (and probably should) do the same for `*SetItem`, so that all the code elsewhere does not have to worry about the locking any more. Would that work?

numpy/_core/src/common/npy_hashtable.cpp
ngoldbaum: Sure, that would work too. Then callers would only have to care about the mutex if they need to write to the cache, which should only ever happen inside the dispatching implementation where it happens right now.

ngoldbaum: > do the same for *SetItem, so that all the code elsewhere does not have to worry about the locking any more. Would that work?

The `SetItem` call is only ever used in the dispatching implementation. IMO I'd rather not do this refactor for `SetItem` unless there's a customer for mutating the dispatch cache outside the dispatching implementation. (There's one use in `_multiarray.tests.c.src`, but IMO that's not worth worrying about)

ngoldbaum: > Sure, that would work too

Looking at the code again, an issue with this is that `promote_and_get_info_and_ufuncimpl` is recursive and we'd need to add a private version of `PyArrayIdentityHash_GetItem` that doesn't do locking. This ends up being a bigger refactor than just doing the `WithLocking` variant and making it clear in the source code that only the free-threaded build needs locking.

I don't think `shared_mutex` is guaranteed to be recursive, so you need to make sure you only acquire the mutex on the outermost layer of recursion otherwise you might cause a crash or deadlock.

kumaraditya303: > Sure, that would work too. Then callers would only have to care about the mutex if they need to write to the cache, which should only ever happen inside the dispatching implementation where it happens right now.

If we do the locking inside `PyArrayIdentityHash_GetItem` then the callers must not hold the mutex, as it is undefined behavior to read lock a mutex which is already locked[^1]. It would require further refactoring of `promote_and_get_info_and_ufuncimpl_with_locking` which currently calls `PyArrayIdentityHash_GetItem` with exclusive locked mutex.

I don't think  moving locking inside `PyArrayIdentityHash_GetItem` makes things simpler. 


[^1]: https://en.cppreference.com/w/cpp/thread/shared_mutex/lock_shared.html

mhvk: I'm not sure I understand the problem with recursive code: if `GetItem` acquires and releases the lock, then the recursion just means you get the lock a couple of times (something which would happen only the first time a function is tried for the given descriptors). So, is there then still a need for the dispatching code to worry about the GIL at all? I think not, i.e., with local locking, we can get rid of `promote_and_get_info_and_ufuncimpl_with_locking` altogether.

Indeed, this is my overall point: we do not have to worry about special handling anywhere else any more if you change both the `GetItem` and `SetItem` definitions depending on GIL availability (again, just like [I think] you don't have to care for `PyDict_GetItem`.)

p.s. I realize some of this may get annoying to really reason through (what if one thread sets the item while another is in the middle of its recursion, etc.), so if there is a real worry with doing it locally, I'm OK with getting this in too. I just want to be sure we really consider what I think is the better option, as it is unlikely we'll get back to it any time soon...

ngoldbaum: The problem is that we hold the write lock in the slow path when there's a cache miss:

https://github.com/numpy/numpy/blob/eab17dc13f3313b6598e214e340f4e28d14accad/numpy/_core/src/umath/dispatching.cpp#L999-L1006

`promote_and_get_info_and_ufuncimpl` needs to call into a variant of `PyArrayIdentityHash_GetItem` that doesn't do locking. We could refactor things so the API looks like that (e.g. call `_PyArrayIdentityHash_GetItem`), but then IMO we've left things less clear than they are already.

I like that the way things are currently written it's very clear there's a lock but it's only being used on the free-threaded build.

mhvk: I keep coming back to thinking `PyArrayIdentifyHash` should behave just be like `PyDict` (indeed, I argued with @seberg that we should *use* `PyDict` instead of his custom version, but the custom version had sufficiently better performance to keep it -- an unfortunate side effect is that now you have to ensure this custom version is thread-safe...).

But I don't want to hold up fixing a problem any longer -- just go ahead and merge if you think the current code is better (or indeed good enough...)!

seberg: Right, when locking had to be introduced, I was wondering if that small perf bonus is still worth it, it may not be (for free-threading). There may also be ways to get a bit more speed out of it in other ways.

But either way, I wouldn't mind using a dict, but I think for now this is a single special case and the recursion is a good reason of why it doesn't help much to put it into a utility...
So for now I think we might as well just do this.

(We also probably shouldn't use a borrowed reference anymore in free-threading...)

ngoldbaum: Just using a regular dict wouldn't help either. The thread safety problem in this cache isn't that the data structure isn't atomic, it's that the whole recursive ufunc promotion algorithm isn't atomic. As soon as there's a cache miss there needs to be some kind of lock preventing multiple threads from computing the value that fills the cache simultaneously.

It's definitely a subtle point: this code went through several rounds of iteration. I even highlighted it in the PyCon talk I co-presented this year as a spot that caused a multithreaded scaling issue. The original implementation we shipped used a critical section  instead of the shared_mutex and that scaled poorly.

mhvk: Thanks - that does help! If any changes are still to be made, I think it should be comments in the code clarifying this!

seberg:
Thanks @kumaraditya303 and Nathan!

PR #30255, opened 2025-11-18
BUG: always ignore FPE when Accelerate is the BLAS backend

### LINKED ISSUE #30162: TST: a flaky test on macOS arm64 (RuntimeWarning: invalid value encountered in matmul)
Seen in gh-30160
@mattip reported seeing this independently too

Basically, the test (sometimes) fails as
```
FAILED numpy/linalg/tests/test_linalg.py::TestQR::test_stacked_inputs[float32-outer_size2-size1] - RuntimeWarning: invalid value encountered in matmul
```

I'm able to reproduce this locally (although unreliably), so I'll attempt a bisection.
possibly related to gh-29820

mattip:
Fixes #30162 which seems to be a regression in Accelerate. It started  happening when we moved to a newer macOS image in CI (not an M4 machine) and was independently reported by @neutrinoceros on macOS 15.7.1 with an M2 machine. 

Continuation of #30102. Set the default `blas_supports_fpe=false` when compiling with Accelerate to "ignore all FPE in calls to BLAS". #30102 added a check when importing NumPy to see if a simple matmul triggers an FPE. If so, then `blas_supports_fpe` is set to `false`. But since the check is flaky it does not reliably set the value to `false` and we see FPE warnings.

Downstream packagers like conda that allow switching out the blas support from Accelerate to OpenBLAS to MKL and others will have to develop a patch to deal with the flaky Accelerate library and FPEs. The available handles are:
- at compile time they could choose to always set `blas_supports_fpe=false` and never see the warnings, at the cost of not knowing when overflow/underflow/divide-by-zero occurs
- at runtime, they could call `numpy._core._multiarray_umath._blas_supports_fpe(False)` when Accelerate is swapped in, to work around the flaky FPE false-positives.

### DIFF
--- a/numpy/_core/src/common/blas_utils.c
+++ b/numpy/_core/src/common/blas_utils.c
@@ -14,8 +14,13 @@
 #if NPY_BLAS_CHECK_FPE_SUPPORT
 /*
  * Static variable to cache runtime check of BLAS FPE support.
+ * Will always be false (ignore all FPE) when accelerate is the compiled backend
  */
+  #if defined(ACCELERATE_NEW_LAPACK)
+static bool blas_supports_fpe = false;
+  #else
 static bool blas_supports_fpe = true;
+  #endif // ACCELERATE_NEW_LAPACK
 
 #endif // NPY_BLAS_CHECK_FPE_SUPPORT
 


neutrinoceros:
Just want to report I've been running my reproducer from #30162 in a loop for 40min and never hit the exception. I think I can safely conclude that the fix is confirmed. Thanks Matti !

charris:
Thanks Matti. I would have stuck a `const` on variable, but I don't think it worth worrying about.

matthew-brett:
Do the Apple developer team know about this bug?  (Sorry if I missed the report).

mattip:
I did ping @Developer-Ecosystem-Engineering [in the issue](https://github.com/numpy/numpy/issues/30162#issuecomment-3507573201) but will do so again here

seberg:
@mattip I think we need to remove that silly warning that I added if this is flaky?  I mean, just deleting is fine, it was maybe a bad idea to begin with.

mattip:
I think the warning is fine. Let's see if we get reports of it triggering.

Developer-Ecosystem-Engineering:
> I did ping @Developer-Ecosystem-Engineering [in the issue](https://github.com/numpy/numpy/issues/30162#issuecomment-3507573201) but will do so again here

Hi Matt,

We saw it =) No update at the moment but it's in the queue.

PR #30256, opened 2025-11-18
CI: update ``paths-ignore`` for mypy and wheels workflows

jorenham:
ref https://github.com/numpy/numpy/pull/30247#issuecomment-3545738248


mattip:
Thanks @jorenham

Issue #30085, opened 2025-10-28
BUG: Running test suite in parallel causes ThreadSanitizer data races


bwhitt7:
### Describe the issue:

When running the test suite under [pytest-run-parallel](github.com/Quansight-Labs/pytest-run-parallel) and [ThreadSanitizer](https://clang.llvm.org/docs/ThreadSanitizer.html), many tests cause data races. The affected tests (~80 test) are listed in [this spreadsheet](https://docs.google.com/spreadsheets/d/1iJHNqrcPAP3GJM0oooZuv7-YI9SAE7Db-U0ZIzFzMw8/edit?usp=sharing). Below are some tests I thought were notable.

#### Tests cause a crash, with no output (some of these overlap with [ASan errors I found previously](https://github.com/numpy/numpy/issues/29890), and likely have to do with the tests crashing due to memory issues):
- [ ] numpy/_core/tests/test_arrayprint.py::test_multithreaded_array_printing
- [ ] numpy/_core/tests/test_multiarray.py::TestCreation::test_zeros_big
- [ ] numpy/_core/tests/test_regression.py::TestRegression::test_to_ctypes
- [ ] numpy/_core/tests/test_shape_base.py::TestConcatenate::test_huge_list_error

#### Segmentation fault after data race
- [ ] numpy/_core/tests/test_custom_dtypes.py::TestSFloat::test_wrapped_and_wrapped_reductions

#### RandomState data race (may be a linux-only bug)
- [ ] numpy/linalg/tests/test_linalg.py::TestCond::test_stacked_singular
- [ ] numpy/_core/tests/test_umath.py::TestLog::test_log_strides
- [ ] numpy/linalg/tests/test_linalg.py::TestCond::test_stacked_singular

#### dec_addstatus data races
- [ ] numpy/_core/tests/test_umath.py::TestRationalFunctions::test_decimal
- [ ] numpy/lib/tests/test_function_base.py::TestAverage::test_object_dtype
- [ ] numpy/lib/tests/test_function_base.py::TestAverage::test_object_no_weights

#### Caused test failure
- [ ] numpy/lib/tests/test_io.py::TestSaveTxt::test_large_zip
- [ ] numpy/testing/tests/test_utils.py::TestAssertNoGcCycles::test_asserts

### Steps to reproduce
- Use Python `3.14t` with ThreadSanitizer enabled.
   - Steps on how to do this can be found [here](https://py-free-threading.github.io/thread_sanitizer/). 
[...]

PR #30259, opened 2025-11-19
TST: mark tests which call `gc.collect()` as thread unsafe

kumaraditya303:
Relates #30085

This PR marks tests as thread unsafe if it relies on `gc.collect()` to break reference cycles.


ngoldbaum:
Thanks @kumaraditya303!

FWIW, I opened https://github.com/Quansight-Labs/pytest-run-parallel/issues/147. We can probably detect this pattern and automatically mark tests more often than not.

PR #30263, opened 2025-11-19
TST: scalar fast path multithreaded test

ngoldbaum:
Followup for #30254 and the discussion there.

I checked that the test I modified generates TSan race reports on the commit before Kumar's PR was merged.

Ping @mhvk: are the comments I added here what you were looking for?

### DIFF
--- a/numpy/_core/tests/test_multithreading.py
+++ b/numpy/_core/tests/test_multithreading.py
@@ -29,11 +29,12 @@ def test_parallel_ufunc_execution():
     # if the loop data cache or dispatch cache are not thread-safe
     # computing ufuncs simultaneously in multiple threads leads
     # to a data race that causes crashes or spurious exceptions
-    def func():
-        arr = np.random.random((25,))
-        np.isnan(arr)
+    for ufunc in [np.isnan, np.sin]:
+        for op in [np.random.random((25,)), np.int32(25)]:
+            def func():
+                ufunc(op)
 
-    run_threaded(func, 500)
+            run_threaded(func, 500)
 
     # see gh-26690
     NUM_THREADS = 50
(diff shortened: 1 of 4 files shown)


numpy/_core/tests/test_multithreading.py
```diff
@@ -29,11 +29,12 @@ def test_parallel_ufunc_execution():
     # if the loop data cache or dispatch cache are not thread-safe
     # computing ufuncs simultaneously in multiple threads leads
     # to a data race that causes crashes or spurious exceptions
-    def func():
-        arr = np.random.random((25,))
-        np.isnan(arr)
+    for ufunc in [np.isnan, np.sin]:
+        for op in [np.random.random((25,)), np.int32(25)]:
```
mhvk: I'd add a `np.float32(1.5)` case to ensure we also cover the case of a scalar for which a loop is present (not sure it would have failed, though I guess these tests are on a "clean" numpy?).

mhvk (commented):
Thanks for adding the comments! Just a comment on the test, since `np.sin` does not have a loop for integer.

numpy/_core/tests/test_multithreading.py
ngoldbaum: I didn't try float32 but I bet int32 promotes to float32 so it works out to the same thing. I'll check if float32 triggers the race on an unpatched numpy.

I first tried float64 and that didn't trigger the race then switched to int32, saw it raced, and didn't think about it much more.

ngoldbaum: If a loop is present there's no possibility of a race. Because it's filled in at import time. I'll go ahead and add float64 and float32 as well, because why not.

mhvk: Promotion means that the fast-track route gets abandoned immediately.

And, yes, I worried that the `float` examples would just test access, not creation.

mhvk (approved):
Thanks!

seberg:
Thanks a lot! Let's get it in.

PR #30270, opened 2025-11-20
ENH: Use descriptor rather than custom `tp_getattro`

seberg:
To me a custom `tp_getattro` feels a bit wrong, so instead use a descriptor protocol to solve this.
That is unfortunately slightly annoying/complex.  Static types do however have a `tp_dict` (a fact that I had missed...).

So while normal properties don't behave well enough for signatures putting a `__signature__` into the `__dict__` does work perfectly fine, although admittedly it is bit awkward to set up.

This should ensure that fast-paths that rely on the default getattr not being replaced should remain active.

### DIFF
--- a/numpy/_core/src/multiarray/multiarraymodule.c
+++ b/numpy/_core/src/multiarray/multiarraymodule.c
@@ -4966,7 +4966,18 @@ _multiarray_umath_exec(PyObject *m) {
         return -1;
     }
 
+    /* Set __signature__ to None on the type (the instance has a property) */
+    s = npy_import("numpy._globals", "_SignatureDescriptor");
+    if (s == NULL) {
+        return -1;
+    }
+    PyUFunc_Type.tp_dict = Py_BuildValue(
+        "{ON}", npy_interned_str.__signature__, s);
+    if (PyUFunc_Type.tp_dict == NULL) {
+        return -1;
+    }
     if (PyType_Ready(&PyUFunc_Type) < 0) {
+        Py_CLEAR(PyUFunc_Type.tp_dict);
         return -1;
     }
 
--- a/numpy/_core/src/multiarray/multiarraymodule.c
+++ b/numpy/_core/src/multiarray/multiarraymodule.c
@@ -4967,7 +4967,7 @@ _multiarray_umath_exec(PyObject *m) {
     }
 
     /* Set __signature__ to None on the type (the instance has a property) */
-    s = npy_import("numpy._globals", "_SignatureDescriptor");
+    s = npy_import("numpy._globals", "_signature_descriptor");
     if (s == NULL) {
         return -1;
     }
--- a/numpy/_core/src/umath/ufunc_object.c
+++ b/numpy/_core/src/umath/ufunc_object.c
@@ -6755,6 +6755,7 @@ static PyGetSetDef ufunc_getset[] = {
     {"signature",
         (getter)ufunc_get_signature,
         NULL, NULL, NULL},
+    // __signature__ stored in `__dict__`, see `_globals._SignatureDescriptor`
     {NULL, NULL, NULL, NULL, NULL},  /* Sentinel */
 };
 
--- a/numpy/_core/src/umath/ufunc_object.c
+++ b/numpy/_core/src/umath/ufunc_object.c
@@ -6646,57 +6646,6 @@ ufunc_set_doc(PyUFuncObject *ufunc, PyObject *doc, void *NPY_UNUSED(ignored))
 }
 
 static PyObject *
-ufunc_get_inspect_signature(PyUFuncObject *ufunc, void *NPY_UNUSED(ignored))
-{
-    PyObject *signature;
-
-    // If there is a __signature__ in the instance __dict__, use it.
-    int result = PyDict_GetItemRef(ufunc->dict, npy_interned_str.__signature__,
-                                   &signature);
-    if (result == -1) {
-        return NULL;
-    }
-    else if (result == 1) {
-        return signature;
-    }
-
-    if (npy_cache_import_runtime(
-            "numpy._core._internal", "_ufunc_inspect_signature_builder",
-            &npy_runtime_imports._ufunc_inspect_signature_builder) == -1) {
-        return NULL;
-    }
-
-    signature = PyObject_CallFunctionObjArgs(
-            npy_runtime_imports._ufunc_inspect_signature_builder,
-            (PyObject *)ufunc, NULL);
-    if (signature == NULL) {
-        return NULL;
-    }
-
-    // Cache the result in the instance dict for next time
-    if (PyDict_SetItem(ufunc->dict, npy_interned_str.__signature__,
-                       signature) < 0) {
-        Py_DECREF(signature);
-        return NULL;
-    }
-
-    return signature;
-}
-
-static PyObject *
-ufunc_getattro(PyObject *obj, PyObject *name)
-{
-    // __signature__ special-casing to prevent class attribute access
-    if (PyUnicode_Check(name) &&
-        PyUnicode_Compare(name, npy_interned_str.__signature__) == 0) {
-        return ufunc_get_inspect_signature((PyUFuncObject *)obj, NULL);
-    }
-    
-    // For all other attributes, use default behavior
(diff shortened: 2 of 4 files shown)


numpy/_core/src/umath/ufunc_object.c
```diff
@@ -6645,57 +6645,6 @@ ufunc_set_doc(PyUFuncObject *ufunc, PyObject *doc, void *NPY_UNUSED(ignored))
     }
 }
 
-static PyObject *
```
jorenham, line 6651: Maybe leave a little comment here, so people searching for it know where to go? (Cus I know that I'm probably gonna forget this)

numpy/_globals.py
```diff
@@ -94,3 +94,28 @@ def __bool__(self):
             return False
 
         raise ValueError(f"{self} is neither True nor False.")
+
+
+class _SignatureDescriptorClass:
```
mhvk: Total nitpick, but my PEP 8 mind would like to call this `_SignatureDescriptor` and below define `_signature_descriptor = _SignatureDescriptor()`

numpy/_core/src/umath/ufunc_object.c
mhvk: I agree a comment is useful. Maybe best near where `__doc__` and `name` are defined? 
...
... Ah, you already did that!
... 
I'd also tend to cross-reference the descriptor inside `_ufunc_inspect_signature_builder`?

mhvk (commented):
Very nice! And actually elegant, too. Just a few very small comments.

jorenham:
In case you missed it, there's a pypy thingy:

```
  FAILED numpy/_core/tests/test_umath.py::TestSpecialMethods::test_ufunc_docstring - AssertionError: assert {} == {'__signature...nature=None)>}
    Right contains 1 more item:
    {'__signature__': <Signature (x1, x2, /, out=None, *, where=True, casting='same_kind', order='K', dtype=None, subok=True, signature=None)>}
    Full diff:
    - {'__signature__': <Signature (x1, x2, /, out=None, *, where=True, casting='same_kind', order='K', dtype=None, subok=True, signature=None)>}
    + {}
  = 1 failed, 47103 passed, 1401 skipped, 2820 deselected, 58 xfailed, 38 xpassed in 563.11s (0:09:23) =
```

mattip:
> there's a pypy thingy

Probably best to skip the test on PyPy rather than spend time debugging it.

jorenham:
> > there's a pypy thingy
> 
> Probably best to skip the test on PyPy rather than spend time debugging it.

weren't we getting rid of it anyway :thinking: ?

seberg:
I think we are getting dropping it naturally when dropping Python 3.11, which IIRC was for the 2.5 release.

I guess it shouldn't be too surprising that PyPy doesn't like this dance, it's rather weird, anyway it isn't like a bad `__signature__` on PyPy is something that we need to worry over.

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -4966,7 +4966,18 @@ _multiarray_umath_exec(PyObject *m) {
...
+    if (s == NULL) {
+        return -1;
+    }
+    PyUFunc_Type.tp_dict = Py_BuildValue(
+        "{ON}", npy_interned_str.__signature__, s);
+    if (PyUFunc_Type.tp_dict == NULL) {
+        return -1;
+    }
     if (PyType_Ready(&PyUFunc_Type) < 0) {
+        Py_CLEAR(PyUFunc_Type.tp_dict);
```
mhvk, line 4980: Just to be sure, do we have to deallocate this ourselves in `dealloc`?

mhvk (approved):
Modulo the question in-line, this looks all good. I do actually quite like how this works out, allowing a pure Python solution. It may well help for other descriptors on other classes too!

jorenham:
> Modulo the question in-line, this looks all good. I do actually quite like how this works out, allowing a pure Python solution. It may well help for other descriptors on other classes too!

It might be nice to clean these ones up as well:

```py
>>> np.ufunc.nin
<attribute 'nin' of 'numpy.ufunc' objects>
>>> np.ufunc.nout
<attribute 'nout' of 'numpy.ufunc' objects>
>>> np.ufunc.nargs
<attribute 'nargs' of 'numpy.ufunc' objects>
>>> np.ufunc.ntypes
<attribute 'ntypes' of 'numpy.ufunc' objects>
>>> np.ufunc.types
<attribute 'types' of 'numpy.ufunc' objects>
>>> np.ufunc.identity
<attribute 'identity' of 'numpy.ufunc' objects>
>>> np.ufunc.signature
<attribute 'signature' of 'numpy.ufunc' objects>
```

numpy/_core/src/multiarray/multiarraymodule.c
seberg: There is no such thing as a type dealloc, the type is immortal.

seberg:
> It might be nice to clean these ones up as well:

I am -1 on that unless you can come up with some *very* good reason. For signature we needed this dance to paper over an issue that otherwise might be fixed ~downstream~ upstream (by ignoring descriptors there).  I am not willing to make a custom dance in NumPy for some purity when literally nothing else in the Python world does this.

If we care about this, we should open an issue at CPython to allow this and make it simple (e.g. including an `@property(invisible_on_class=True)` style thing or so, although dunno how one would piggy-back it into CPython).

EDIT: and we may yet want to open an issue that `inspect.signature()` should ignore unbound descriptors.

jorenham:
> I am -1 on that unless you can come up with some _very_ good reason.

The only reason I have is that, from a user-perspective, they're completely useless. But for things like static typing it's absolutely no problem, and type-checkers consider them as if they're `@property`s.

numpy/_core/src/multiarray/multiarraymodule.c
mhvk: Ah, yes, of course, I was confusing myself with creation of an actual ufunc.

seberg:
> The only reason I have is that, from a user-perspective, they're completely useless.

Yeah, but it is also weird that this pattern makes `ufunc.__signature__` effectively inaccessible even though it exists, so I am not sure there is anything to do.

I think the bug may actually more be that it would be nice to have a `getinstanceattr()` that omits descriptors on classes or so.  And also if you feel like doing a small PR on Python, I do think that `inspect.signature()` could ignore descriptors.
(Not sure how to best detect it, we always check for hasattr `__get__`, but there might actually be a better way via type flags, although I guess that may not existing in Python.)

numpy/_core/src/multiarray/multiarraymodule.c
jorenham: > There is no such thing as a type dealloc, the type is immortal.

This brings me great joy

jorenham:
> getinstanceattr

So something like this then?

```py
descriptor_types = (
    types.GetSetDescriptorType,
    types.MemberDescriptorType,
    types.MethodDescriptorType,
    types.ClassMethodDescriptorType,
)

def getinstanceattr(obj, name):
    attr = getattr(obj, name)
    if isinstance(attr, descriptor_types):
        raise AttributeError(obj=obj, name=name)
    return attr
```

Kinda like the opposite of [`inspect.getattr_static`](https://docs.python.org/3/library/inspect.html#fetching-attributes-statically)

seberg:
I think you would more do something like a `hasattr(..., "__get__")` but yeah.  I didn't realize `inspect.getattr_static` I suppose it's probably a bit like `if attr is inspect.getattr_static(obj, name): pass` (glancing at `getattr_static` it isn't quite clear to me why it seems to want a `__delete__` or `__set__` in some branch, but not sure what the branch exactly does, so...).

mhvk:
I think the discussion is wandering off to very interesting things, but not a reason to just merge this. I think it is a really nice solution!

seberg:
Quite seriously though, in case you stumble on a nicer solution, we actually have this problems in a few places (`__array_interface__`/struct, `__numpy_dtype__`/`dtype`, and maybe more).
And there we use something like `isinstance(obj, type) and hasattr(attr, "__get__")`.

PR #30275, opened 2025-11-21
ENH: use `PyMutex` based rwlock for better ufunc scaling

kumaraditya303:

### DIFF
--- a/numpy/_core/src/common/npy_rwlock.c
+++ b/numpy/_core/src/common/npy_rwlock.c
@@ -56,7 +56,7 @@ PyRWMutex_RLock(PyRWMutex *rwmutex)
         PyMutex_Lock(&rwmutex->writer_lock);
     }
     assert(PyMutex_IsLocked(&rwmutex->writer_lock));
-    assert(rwmutex->writer_id == 0 || rwmutex->writer_id == PyThread_get_thread_ident());
+    assert(rwmutex->writer_id == 0);
     PyMutex_Unlock(&rwmutex->reader_lock);
 }
 
--- a/numpy/_core/src/common/npy_rwlock.c
+++ b/numpy/_core/src/common/npy_rwlock.c
@@ -1,0 +1,72 @@
+#include "npy_rwlock.h"
+
+#include <stdatomic.h>
+#include <stdbool.h>
+
+NPY_NO_EXPORT void
+PyRWMutex_Lock(PyRWMutex *rwmutex)
+{
+    unsigned long thread_id = PyThread_get_thread_ident();
+    unsigned long writer_id = atomic_load_explicit((_Atomic unsigned long *)&rwmutex->writer_id, memory_order_acquire);
+    if (writer_id == thread_id) {
+        // If the current thread already holds the write lock, allow recursive write lock
+        rwmutex->level++;
+        return;
+    }
+    PyMutex_Lock(&rwmutex->writer_lock);
+    atomic_store_explicit((_Atomic unsigned long *)&rwmutex->writer_id, thread_id, memory_order_release);
+}
+
+NPY_NO_EXPORT void
+PyRWMutex_Unlock(PyRWMutex *rwmutex)
+{
+    unsigned long thread_id = PyThread_get_thread_ident();
+    if (rwmutex->writer_id != thread_id) {
+        Py_FatalError("Attempt to unlock a RWMutex not owned by the thread");
+    }
+    if (rwmutex->level > 0) {
+        rwmutex->level--;
+        return;
+    }
+    atomic_store_explicit((_Atomic unsigned long *)&rwmutex->writer_id, 0, memory_order_release);
+    PyMutex_Unlock(&rwmutex->writer_lock);
+}
+
+NPY_NO_EXPORT void
+PyRWMutex_RLock(PyRWMutex *rwmutex)
+{
+    unsigned long thread_id = PyThread_get_thread_ident();
+    unsigned long writer_id = atomic_load_explicit((_Atomic unsigned long *)&rwmutex->writer_id, memory_order_acquire);
+    // If current thread holds the write lock, allow recursive read lock
+    if (writer_id == thread_id) {
+        rwmutex->level++;
+        return;
+    }
+
+    PyMutex_Lock(&rwmutex->reader_lock);
+    rwmutex->reader_count++;
+    if (rwmutex->reader_count == 1) {
+        // First reader acquires the writer lock to block writers
+        PyMutex_Lock(&rwmutex->writer_lock);
+        // zero means locked by reader
+        atomic_store_explicit((_Atomic unsigned long *)&rwmutex->writer_id, 0, memory_order_release);
+    }
+    PyMutex_Unlock(&rwmutex->reader_lock);
+}
+
+NPY_NO_EXPORT void
+PyRWMutex_RUnlock(PyRWMutex *rwmutex)
+{
+    if (rwmutex->level > 0) {
+        rwmutex->level--;
+        return;
+    }
+    PyMutex_Lock(&rwmutex->reader_lock);
+    rwmutex->reader_count--;
+    if (rwmutex->reader_count == 0) {
+        // Last reader releases the writer lock
+        atomic_store_explicit((_Atomic unsigned long *)&rwmutex->writer_id, 0, memory_order_release);
+        PyMutex_Unlock(&rwmutex->writer_lock);
+    }
+    PyMutex_Unlock(&rwmutex->reader_lock);
+}
--- a/numpy/_core/src/common/npy_rwlock.c
+++ b/numpy/_core/src/common/npy_rwlock.c
@@ -3,6 +3,11 @@
 
 #ifdef Py_GIL_DISABLED
 
+// PyMutex_IsLocked is available only in Python 3.14+
+#ifndef PyMutex_IsLocked
+#define PyMutex_IsLocked(m) (1)
+#endif
+
 NPY_NO_EXPORT void
 PyRWMutex_Lock(PyRWMutex *rwmutex)
 {
@@ -11,6 +16,7 @@ PyRWMutex_Lock(PyRWMutex *rwmutex)
     if (writer_id == thread_id) {
         // If the current thread already holds the write lock, allow recursive write lock
         rwmutex->level++;
+        assert(PyMutex_IsLocked(&rwmutex->writer_lock));
         return;
     }
     PyMutex_Lock(&rwmutex->writer_lock);
@@ -21,8 +27,9 @@ PyRWMutex_Lock(PyRWMutex *rwmutex)
 NPY_NO_EXPORT void
 PyRWMutex_Unlock(PyRWMutex *rwmutex)
 {
+    assert(PyMutex_IsLocked(&rwmutex->writer_lock));
+    assert(rwmutex->writer_id == PyThread_get_thread_ident());
     if (rwmutex->level > 0) {
-        assert(rwmutex->writer_id == PyThread_get_thread_ident());
         rwmutex->level--;
         return;
     }
@@ -33,14 +40,22 @@ PyRWMutex_Unlock(PyRWMutex *rwmutex)
 NPY_NO_EXPORT void
 PyRWMutex_RLock(PyRWMutex *rwmutex)
 {
+    unsigned long thread_id = PyThread_get_thread_ident();
+    unsigned long writer_id = npy_atomic_load_ulong(&rwmutex->writer_id);
+    // If current thread holds the write lock, allow recursive read lock
+    if (writer_id == thread_id) {
+        rwmutex->level++;
+        assert(PyMutex_IsLocked(&rwmutex->writer_lock));
+        return;
+    }
+
     PyMutex_Lock(&rwmutex->reader_lock);
     rwmutex->reader_count++;
     if (rwmutex->reader_count == 1) {
         // First reader acquires the write lock to block writers
         PyMutex_Lock(&rwmutex->writer_lock);
-        // zero means locked by reader
-        npy_atomic_store_ulong(&rwmutex->writer_id, 0);
     }
+    assert(PyMutex_IsLocked(&rwmutex->writer_lock));
     assert(rwmutex->writer_id == 0 || rwmutex->writer_id == PyThread_get_thread_ident());
     PyMutex_Unlock(&rwmutex->reader_lock);
 }
@@ -48,8 +63,14 @@ PyRWMutex_RLock(PyRWMutex *rwmutex)
 NPY_NO_EXPORT void
 PyRWMutex_RUnlock(PyRWMutex *rwmutex)
 {
-    PyMutex_Lock(&rwmutex->reader_lock);
+    assert(PyMutex_IsLocked(&rwmutex->writer_lock));
     assert(rwmutex->writer_id == 0 || rwmutex->writer_id == PyThread_get_thread_ident());
+    if (rwmutex->level > 0) {
+        assert(rwmutex->writer_id == PyThread_get_thread_ident());
+        rwmutex->level--;
+        return;
+    }
+    PyMutex_Lock(&rwmutex->reader_lock);
     rwmutex->reader_count--;
     if (rwmutex->reader_count == 0) {
         // Last reader releases the writer lock
--- a/numpy/_core/src/common/npy_rwlock.c
+++ b/numpy/_core/src/common/npy_rwlock.c
@@ -9,7 +9,7 @@
 #endif
 
 NPY_NO_EXPORT void
-PyRWMutex_Lock(PyRWMutex *rwmutex)
+NPyRWMutex_Lock(NPyRWMutex *rwmutex)
 {
     unsigned long thread_id = PyThread_get_thread_ident();
     unsigned long writer_id = npy_atomic_load_ulong(&rwmutex->writer_id);
@@ -25,7 +25,7 @@ PyRWMutex_Lock(PyRWMutex *rwmutex)
 }
 
 NPY_NO_EXPORT void
-PyRWMutex_Unlock(PyRWMutex *rwmutex)
+NPyRWMutex_Unlock(NPyRWMutex *rwmutex)
 {
     assert(PyMutex_IsLocked(&rwmutex->writer_lock));
     assert(rwmutex->writer_id == PyThread_get_thread_ident());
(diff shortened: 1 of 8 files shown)


numpy/_core/src/common/npy_rwlock.h
```diff
@@ -0,0 +1,55 @@
...
+
+    Structure members:
+    - reader_lock: Mutex to protect reader_count.
+    - writer_lock: Mutex to protect write access.
+    - reader_count: Number of active readers.
+    - writer_id: Thread ID of the writer holding the lock (0 if no writer).
+    - level: Recursion level for the current thread holding the write lock.
+*/
+
+typedef struct PyRWMutex {
```
ngoldbaum: probably needs to be Npy instead of Py. Here and everywhere else.

numpy/_core/src/common/npy_rwlock.c
```diff
@@ -0,0 +1,62 @@
...
+{
+    PyMutex_Lock(&rwmutex->reader_lock);
+    rwmutex->reader_count++;
+    if (rwmutex->reader_count == 1) {
+        // First reader acquires the writer lock to block writers
+        PyMutex_Lock(&rwmutex->writer_lock);
+        // zero means locked by reader
+        npy_atomic_store_ulong(&rwmutex->writer_id, 0);
+    }
+    assert(rwmutex->writer_id == 0 || rwmutex->writer_id == PyThread_get_thread_ident());
```
mhvk: This part confuses me. How can it be equal to one's own thread id? If I'm the first reader, I've acquired the write lock and set the id to 0, while if I'm a later one, I cannot have the write lock.

If this is meant for the case where the thread itself acquires the write lock before the read lock, then is it correct to write 0 above?

numpy/_core/src/common/npy_rwlock.c
```diff
@@ -0,0 +1,62 @@
...
+
+NPY_NO_EXPORT void
+PyRWMutex_RUnlock(PyRWMutex *rwmutex)
+{
+    PyMutex_Lock(&rwmutex->reader_lock);
+    assert(rwmutex->writer_id == 0 || rwmutex->writer_id == PyThread_get_thread_ident());
+    rwmutex->reader_count--;
+    if (rwmutex->reader_count == 0) {
+        // Last reader releases the writer lock
+        npy_atomic_store_ulong(&rwmutex->writer_id, 0);
```
mhvk: Relatedly, I guess I'm confused why there is a need to store `0` -- it seems to me this is guaranteed to be zero.

numpy/_core/src/common/npy_rwlock.h
```diff
@@ -0,0 +1,55 @@
...
+    PyMutex reader_lock;
+    PyMutex writer_lock;
+    uint32_t reader_count;
+    unsigned long writer_id;
+    unsigned long level;
```
mhvk: Also `uint32_t`?

mhvk (commented):
This seems quite nice and clear -- I like the `assert`! Though some comments because of those!

numpy/_core/src/common/npy_rwlock.h
kumaraditya303: I  have changed it to use uint32_t now

numpy/_core/src/common/npy_rwlock.c
kumaraditya303: No, you are correct. This was leftover from older version of PR as I was debugging the arm64 segfault hence the draft status.

I have pushed the fixes now and improved assertions so things should be clearer now. Thanks for the review.

numpy/_core/src/common/npy_rwlock.h
kumaraditya303: I wasn't sure about that because the hash table uses `Py` prefix as well, anyways I'll change it to `NPy`.

numpy/_core/src/common/npy_rwlock.c
```diff
@@ -0,0 +1,82 @@
...
+    assert(PyMutex_IsLocked(&rwmutex->writer_lock));
+    assert(rwmutex->writer_id == 0);
+    PyMutex_Unlock(&rwmutex->reader_lock);
+}
+
+NPY_NO_EXPORT void
+NPyRWMutex_RUnlock(NPyRWMutex *rwmutex)
+{
+    assert(PyMutex_IsLocked(&rwmutex->writer_lock));
+    if (rwmutex->level > 0) {
```
mhvk, line 65: I think this can fail. Suppose first `_Lock` was called, i.e., the write lock was taken, and then `_RLock`. As is, this means one ends up with `level=1`. Now suppose the code was buggy and `_Unlock` was called before  `_RUnLock`. Then, `_UnLock` would have found `level=1`, decreased it to 0, and here one would just pass through. In principle, this would fail the `assert` just below, but in non-debug mode one would release the write lock without setting thread id to 0. 

I think this can be solved by changing the logic slightly: when one takes the write lock the first time, already set `level = 1` (and inside `_UnLock`, unconditionally decrease the level and return if still non-zero).
That way, here one can simply call `_UnLock` -- it would generally just decrease the level and in the buggy order example, it would also release the lock, setting writer id to zero. 

Possibly that also allows to simplify `_RLock` as in principle one can just check whether `level > 0` to know whether a write lock is in place, i.e., one could do something like
```
if `rwmutex->level` > 0) { // could also just call _Lock, though that does needless work.
    assert(rwmutex->writer_id == PyThread_get_thread_ident());
    rwmutix->level++;
    return;
}
```
I'm not sure, though, whether in that case `level` would need protection. If so, obviously one can just  stick with that you have (the `IsLocked` function would actually be quite handy... backport it?).

mhvk (commented):
@kumaraditya303 - no worries, I had seen it was a draft, but was simply curious!

Your asserts are really great, as they help think through the logic. I *think* I found a possible small problem still -- see inside, also for a suggested solution.

numpy/_core/src/common/npy_rwlock.c
kumaraditya303: I thought about this when initially designing the lock for `lxml` but ultimately I think that allowing `NPyRWMutex_Unlock` to unlock a read lock is an anti-pattern and makes it harder to reason. Currently, it is simple that for every `Rlock` you need `RUnlock` and for every `Lock` you need `Unlock` in order of acquisition. This is same as if you would take say two locks, you need to release them in order to avoid lock ordering deadlock so this follows same pattern.

I looked that other implementations of RWLock like C++ `shared_mutex` and they  don't allow recursion at all in the exclusive case and mark this case as "undefined". 
However this implementation allows that and if the user code is incorrect, it will cause assertion failure which seems like a good compromise than just "undefined" behavior. We can turn the assert into a fatal error but I don't think it is worth it, the CI runs tests in debug mode so it would catch it if the code is buggy.

kumaraditya303: > I'm not sure, though, whether in that case level would need protection.

yes, in this case level would need to be atomic as well

mhvk: Makes sense. I would still start with `level=1`, though, so that one does not have to rely on an `assert` in debug mode for inverted unlock calls.

mhvk: I guess my larger point would be that effectively you have two cases you are dealing with here, a read-write lock and a read-shared, no-write lock. Effectively, you've separate these mostly already; my suggestion would just finish that separation.

kumaraditya303: > I think this can be solved by changing the logic slightly: when one takes the write lock the first time, already set level = 1 (and inside _UnLock, unconditionally decrease the level and return if still non-zero).
That way, here one can simply call _UnLock -- it would generally just decrease the level and in the buggy order example, it would also release the lock, setting writer id to zero.

I am not sure how would this work. Let the first write lock set `level=1`, then read lock would increment it to `level=2`, then in the buggy code write unlock is called which would decrease the level first to `level=1` and since it is non-zero just return. Finally read unlock would decrease the `level=0` and since writer id is not zero there will be assertion failure. This is the same as what we have now. Am I missing something?

numpy/_core/src/common/npy_rwlock.c
```diff
@@ -0,0 +1,82 @@
...
+#ifndef PyMutex_IsLocked
+#define PyMutex_IsLocked(m) (1)
+#endif
+
+NPY_NO_EXPORT void
+NPyRWMutex_Lock(NPyRWMutex *rwmutex)
+{
+    unsigned long thread_id = PyThread_get_thread_ident();
+    unsigned long writer_id = npy_atomic_load_ulong(&rwmutex->writer_id);
+    if (writer_id == thread_id) {
```
mhvk: Small: If here you do `writer_id != 0` than you don't have to get the thread id (replacing it with the assert like below.

numpy/_core/src/common/npy_rwlock.c
mhvk: Sorry, yes, one still has to call `_Unlock`, but without starting at 1 even that would not be good enough. As I mentioned, I think it is best to see (and effectively code) it as two options, a RW lock and a shared R lock. With my suggestion, there is no need for ordering at, as far as I can see, no extra cost.

Anyway, I don't claim to be an expert in any of this, so I definitely won't insist - maybe better to get a review from someone who is.

numpy/_core/src/common/npy_rwlock.c
kumaraditya303: That wouldn't work because It is possible that another thread is holding the write lock so `writer_id != 0` would be true even though it is not locked by current thread.

mhvk: Ah, yes, of course. So, if getting the thread id is at all slow, one could get a bit of a speed advantage by doing,
```
if (writer_id != 0 && writer_id == npy_atomic_load_ulong(&rwmutex->writer_id) {
   ...
}
```

kumaraditya303: I think you mean `writer_id !=0 && writer_id == PyThread_get_thread_ident()`, that would work. 
I think rather than doing that, we can just get thread id faster by taking address of tls variable. On linux that would be just two machine instructions so be much faster than function call.

kumaraditya303: I pushed a commit to use your non-zero optimization, leaving the tls variable as an future experiment.

ngoldbaum:
Ping @colesbury - what do you think about using an RWLock based on `PyMutex` instead of `std::shared_lock`?

numpy/_core/src/common/npy_rwlock.h
ngoldbaum: NumPy also owns the PyArray namespace. We're trying to use Npy for newer stuff.

numpy/_core/src/common/npy_rwlock.c
mhvk: I'll unresolve this for a bit, as I still think my suggestion is worth being seen by another reviewer.

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -981,10 +981,12 @@ promote_and_get_info_and_ufuncimpl_with_locking(
         PyArray_DTypeMeta *op_dtypes[],
         npy_bool legacy_promotion_is_possible)
 {
-    std::shared_mutex *mutex = ((std::shared_mutex *)((PyArrayIdentityHash *)ufunc->_dispatch_cache)->mutex);
-    PyObject *info = PyArrayIdentityHash_GetItemWithLock(
+    NPyRWMutex *mutex = &((PyArrayIdentityHash *)ufunc->_dispatch_cache)->mutex;
```
mhvk, line 984: One nice aspect of the separate read lock that you created is that now it is possible to move it inside `PyArrayIdentityHash_GetItem` - that way you do not need to use it here and in `ufunc_object.c`. Furthermore, it guarantees that ordering is always OK at least for this use of your new lock!

(But can do this in follow-up too.)

kumaraditya303:
I benchmarked `NPyRWMutex` against C++ `shared_mutex` which is currently used using https://github.com/kumaraditya303/lockbench.

Results using 10 threads and 10000000 read acquisitions on each thread with no writer:
- `NPyRWMutex`: 0.695488s
- `shared_mutex`: 4.04059s


So in the un-contended general case of multiple readers, this is 5.8x faster than the current C++ `shared_mutex`.

seberg:
I am a bit curious what the chance is of having this in CPython eventually (and the python capi compat headers).

Just scrolling through, this seems simple enough (and I guess not just used here?) to do.  Not sure if I think it is strictly worth it for the ufunc use-case, but I suspect we have more places (e.g. the string array locking) and more may come in the future.

mhvk:
That's a very nice improvement! Related to @seberg's query, if the current CPython implementations of, e.g., `PyDict_{Get,Set}Item` use the slow version, then indeed it would seem worthwhile to upstream (and if not, would be nice to compare).

seberg:
Oh, I don't mean to upstream for GetItem/SetItem, I would be surprised if that needs it. I was wondering whther CPython (or so) should have this as public API for us to use.  Because NumPy isn't going to be the only project for which this will be useful.

ngoldbaum:
I think originally the plan was to make a full set of multithreaded programming utilities available in the C API. However, if you look at the discussion around adding PyMutex and the critical section APIs, you'll see a decent amount of controversy around this. There isn't a ton of appetite to make the C API maintainers also maintain a set of general-purpose programming utilities.

That's not to say that adding more stuff to the C API isn't a good idea. Just that someone would need to champion that and convince the C API workgroup.

That said, a header-only or just easily vendorable library gets you most of the way there.

I'm hoping Sam weighs in here with more context and/or opinions :)

mhvk:
> Oh, I don't mean to upstream for GetItem/SetItem, I would be surprised if that needs it. I was wondering whther CPython (or so) should have this as public API for us to use. Because NumPy isn't going to be the only project for which this will be useful.

Not sure:  I think multiple readers would be a real use case for `PyDict_GetItem`.  But it may well be that they have a similarly smart locking method, so one can check with their implementation.

Anyway, none of this should stop us using the present PR, it seems nice and clean!

colesbury:
@ngoldbaum How is the ufunc cache used? Are entries ever removed or modified or are they only added?

I'm seeing slightly worse performance with the `PyMutex` based lock using https://github.com/kumaraditya303/lockbench, not better performance. None of these reader-writer locks scale well, so best performance occurs when run on 1 CPU.

```
> taskset -c 0-9 python -c "import lockbench; lockbench.rwlock_lock()"
Elapsed time (RWLock): 18.9537s
> taskset -c 0-9 python -c "import lockbench; lockbench.shared_lock()"
Elapsed time (shared_lock): 18.0525s
```

```
> taskset -c 0 python -c "import lockbench; lockbench.rwlock_lock()"
Elapsed time (RWLock): 3.35476s
> taskset -c 0 python -c "import lockbench; lockbench.shared_lock()"
Elapsed time (shared_lock): 1.95423s
```

ngoldbaum:
> How is the ufunc cache used? Are entries ever removed or modified or are they only added?

They're only ever added to the cache. It's in the hot path of ufunc execution though so all ufunc calls are sensitive to this.

kumaraditya303:
> I'm seeing slightly worse performance with the PyMutex based lock using https://github.com/kumaraditya303/lockbench, not better performance. 

I tried running benchmarks on both mac and linux and I can't reproduce the slowdown. Here are the numbers on linux machine:
```console
lockbench on  main via C v13.3.0-gcc via 🐍 v3.14.0 (env) 
❯ taskset -c 0-9 python -c "import lockbench; lockbench.rwlock_lock()"
Elapsed time (RWLock): 2.33842s

lockbench on  main via C v13.3.0-gcc via 🐍 v3.14.0 (env) took 3s 
❯ taskset -c 0-9 python -c "import lockbench; lockbench.shared_lock()"
Elapsed time (shared_lock): 11.7777s
```
I am confused at how C++ shared lock can outperform `PyMutex` based lock, are you able to reproduce slowdown on a different machine as well?

colesbury:
> are you able to reproduce slowdown on a different machine as well?

Yes, I see similar behavior on two Linux machines (Intel Broadwell 36 cores glibc 2.34; Intel Core Ultra 155H, 6+8+2 heterogeneous cores glibc 2.39). I see better performance with the `PyMutex` based lock on macOS.

> They're only ever added to the cache. It's in the hot path of ufunc execution though so all ufunc calls are sensitive to this.

I think that you'd be better off trying to get rid of the rwlock than trying to improve its performance. Reader-writer locks are rarely a good solution. They ostensibly offer concurrency for multiple readers, but don't scale well because of the modifications to the counter that tracks the number of readers. They are a good fit only if the read-side critical section is large/slow.

The ufunc use case is pretty restricted, which makes it easier to implement:

* Entries are only added, never removed or modified
* It's okay for lookups to spuriously fail and fall back to a slower locking code path.

Here's a sketch of an implementation:

* Modifications are protected by a `PyMutex`. Readers don't acquire any locks.
* Readers load the buckets atomically and for each entry load [`item[0]`](https://github.com/numpy/numpy/blob/cd352ddc8a00a869df136d8def51c3cb630b5d09/numpy/_core/src/common/npy_hashtable.cpp#L79) atomically with acquire memory ordering
* The writer fills in `items[1...(tb->key_len-1)]` first (non atomically is fine). `items[0]` is stored with release memory ordering. In other words, if a reader sees `items[0] != NULL`, that means that `items[1..(tb->key_len-1)] are already filled in and valid.
* Updating the cache should fill in the entry only if it's not already filled. In other words, it should behave like `PyDict_SetDefaultRef` not like `PyDict_SetItem`.
* When resizing the table, copy all the entries to the new buckets, then store the buckets atomically with release memory ordering.
* In the free threaded build, don't free the old buckets when resizing the table because some reader may still be using them. Since the buckets double in size every time you resize the table, this uses at most 2x as much memory for the cache.
* When the cache is deallocated, free all the buckets

Your buckets struct can look something like:

```c
struct buckets {
    struct buckets *prev; // pointer to smaller buckets or NULL
    PyObject *array[]; // flexible array member for actual table.
};
```

If we exposed [`_PyMem_FreeDelayed`](https://github.com/python/cpython/blob/e1f1df4082e2bc6cdb46be9984124c88a7b2ff0f/Include/internal/pycore_pymem.h#L92-L93) you could use that for safely freeing the buckets. That's what we use internally in CPython for this kind of thing, but it's not currently exposed for extensions.

ngoldbaum:
We ended up merging https://github.com/numpy/numpy/pull/30593, so I'm closing this.

PR #30277, opened 2025-11-21
BUG: Fix misleading ValueError in convolve on empty inputs due to variable swapping

Jongwan93:
#### Reference issue
Fixes #30272

#### What does this implement/fix?
This PR fixes a misleading error message in `numpy.convolve` when the first argument is empty and the second is not.

Previously, an internal optimization swapped the arguments (`a` and `v`) based on their length before performing the empty check. This caused the `ValueError` to incorrectly report that the **second** argument was empty when it was actually the **first** argument that was empty.

This change moves the validation logic before the optimization swap.

#### Additional information
I have added a regression test case in `numpy/_core/tests/test_numeric.py` to ensure the correct error message is raised.

image


numpy/_core/tests/test_numeric.py
```diff
@@ -4282,4 +4296,4 @@ def test_astype(self):
         assert_equal(actual.dtype, expected.dtype)
 
         with pytest.raises(TypeError, match="Input should be a NumPy array"):
-            np.astype(data, np.float64)
+            np.astype(data, np.float64)
```
seberg: Would be nice to not change this.

seberg (commented):
I like that the fix is to just move things, but the test is failing, mixing up the parameter names.
(You wrote the test as if the variable name matters, but the argument name matters and we are passing things positionally.)

numpy/_core/numeric.py
```diff
@@ -893,12 +893,12 @@ def convolve(a, v, mode='full'):
 
     """
     a, v = array(a, copy=None, ndmin=1), array(v, copy=None, ndmin=1)
-    if (len(v) > len(a)):
-        a, v = v, a
     if len(a) == 0:
         raise ValueError('a cannot be empty')
     if len(v) == 0:
         raise ValueError('v cannot be empty')
+    if (len(v) > len(a)):
```
jorenham:
```suggestion
    if len(v) > len(a):
```

might as well :shrug:

Jongwan93: Thank you for pointing this out. made the change in next commit

Jongwan93:
@seberg 
Thanks for the review!

I've refactored the test to pass arrays directly  to avoid confusion between variable names and argument positions.
I also reverted the unrelated formatting changes.

Let me know if it looks okay now!

charris:
Lint error needs fixing, just needs the file to end with a newline.

Jongwan93:
> Lint error needs fixing, just needs the file to end with a newline.

Just added missing new line at the end of `test_numeric.py` file. Thank you

numpy/_core/tests/test_numeric.py
```diff
@@ -3686,6 +3686,16 @@ def test_mode(self):
         with assert_raises(TypeError):
             np.convolve(d, k, mode=None)
 
+    def test_convolve_empty_input_error_message(self):
+        """
+        Test that convolve raises the correct error message when inputs are empty.
+        Regression test for gh-issue-number (variable swapping bug).
```
WarrenWeckesser: `gh-issue-number` should be `gh-30272`.

WarrenWeckesser:
```suggestion
        Regression test for gh-30272 (variable swapping bug).
```

WarrenWeckesser:
Thanks @Jongwan93.  I pushed a small fix to a comment.  I'll merge once the latest CI run completes.

PR #30278, opened 2025-11-21
BUG: fix np.resize refchk on python 3.14

### LINKED ISSUE #30265: BUG: ndarray.resize doesn't detect aliasing in python 3.14
### Describe the issue:

In-place resizing of an array with multiple references fails inconsistently with python 3.14.  [This check](https://github.com/numpy/numpy/blob/eab17dc13f3313b6598e214e340f4e28d14accad/numpy/_core/src/multiarray/shape.c#L102) seems wrong.  The code below doesn't raise ValueError, but does if the commented line is uncommented

```python
import numpy as np


def test_resize():
    x = np.zeros((2, 2))
    y = x
    # z = x # raises ValueError if uncommented
    x.resize((5, 5))


if __name__ == "__main__":
    test_resize()
```

### Error message:

```shell

```

2.3.5
3.14.0 | packaged by conda-forge | (main, Oct 12 2025, 19:41:36) [GCC 14.3.0]

### Runtime Environment:

[{'numpy_version': '2.3.5',
  'python': '3.14.0 | packaged by conda-forge | (main, Oct 12 2025, 19:41:36) '
            '[GCC 14.3.0]',
  'uname': uname_result(system='Linux', node='qgpu1', release='6.8.0-51-generic', version='#52-Ubuntu SMP PREEMPT_DYNAMIC Thu Dec  5 13:09:44 UTC 2024', machine='x86_64')},
                      'found': ['SSSE3',
                                'SSE41',
                                'POPCNT',
                                'SSE42',
                                'AVX',
                                'F16C',
                                'FMA3',
                                'AVX2'],
                      'not_found': ['AVX512F',
                                    'AVX512CD',
                                    'AVX512_KNL',
                                    'AVX512_KNM',
                                    'AVX512_SKX',
                                    'AVX512_CLX',
                                    'AVX512_CNL',
                                    'AVX512_ICL',
                                    'AVX512_SPR']}},
 {'architecture': 'Zen',
  'filepath': '/home/rtimpe/git/torch-build/.pixi/envs/repro/lib/libopenblasp-r0.3.30.so',
  'internal_api': 'openblas',
  'num_threads': 48,
  'prefix': 'libopenblas',
  'user_api': 'blas',
  'version': '0.3.30'}]

### Context for the issue:

This came up during testing for pytorch and also affects tensors created with `torch.from_numpy`

ngoldbaum:
Fixes #30265

### DIFF
--- a/numpy/_core/src/multiarray/shape.c
+++ b/numpy/_core/src/multiarray/shape.c
@@ -10,6 +10,7 @@
 #include "numpy/npy_math.h"
 
 #include "npy_config.h"
+#include "npy_pycompat.h"
 #include "arraywrap.h"
 #include "ctors.h"
 #include "shape.h"
@@ -32,7 +33,6 @@ PyArray_Resize_int(PyArrayObject *self, PyArray_Dims *newshape, int refcheck)
     npy_intp oldnbytes, newnbytes;
     npy_intp oldsize, newsize;
     int new_nd=newshape->len, k, elsize;
-    int refcnt;
     npy_intp* new_dimensions=newshape->ptr;
     npy_intp new_strides[NPY_MAXDIMS];
     npy_intp *dimptr;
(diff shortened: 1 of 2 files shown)


seberg:
Hmmmpf, this doesn't seem to work on older versions, I guess because the method call is passed a new reference (i.e. it was `refcount > 2`).
Maybe on 3.14 it actually works out in the end (not sure from the tests), because this shouldn't need a new reference in theory as we don't need a bound method.

---

Oh well, if there is a simple solution great, if not... then too bad but OK, it's not the end of the world if this is just a best effort and I am not worried about the pattern in the issue, just thought there might be more.

ngoldbaum:
The fix ended up being to make the code python-version-dependent and avoid the definition in the compat header. The 2 references used to include a reference from the Python stack, but that's elided in 3.14, so there's only one reference.

seberg:
Thanks Nathan, makes sense that it changed in 3.14, seems to work so let's put it in!

PR #30279, opened 2025-11-21
MAINT: refactor unary temporary elision check

ngoldbaum:
I noticed this while working on https://github.com/numpy/numpy/pull/30278.

If `check_unique_temporary` returns true, that implies that `Py_REFCNT` will return 1, so doing both is redundant. Also `check_unique_temporary` just reduces down to checking `Py_REFCNT == 1` on the GIL-enabled build.


seberg:
Thanks, Nathan. Nice little cleanup!

Issue #2111, opened 2012-10-19
unique and NaN entries (Trac #1514)


thouis:
_Original ticket http://projects.scipy.org/numpy/ticket/1514 on 2010-06-18 by trac user rspringuel, assigned to unknown._

When unique operates on an array with multiple NaN entries its return includes a NaN for each entry that was NaN in the original array.

Examples:
a = random.randint(5,size=100).astype(float)

> > > a[12] = nan #add a single nan entry
> > > unique(a)
> > > array([  0.,   1.,   2.,   3.,   4.,  NaN])
> > > a[20] = nan #add a second
> > > unique(a)
> > > array([  0.,   1.,   2.,   3.,   4.,  NaN,  NaN])
> > > a[13] = nan
> > > unique(a) #and a third
> > > array([  0.,   1.,   2.,   3.,   4.,  NaN,  NaN,  NaN])

This is probably due to the fact that x == y evaluates to False if both x and y are NaN.  Unique needs to have "or (isnan(x) and isnan(y))" added to the conditional that checks for the presence of a value in the already identified values.  I don't know were unique lives in numpy and couldn't find it when I went looking, so I can't make the change myself (or even be sure what the exact syntax of the conditional should be).

Also, the following function can be used to patch over the behavior.

def nanunique(x):
    a = numpy.unique(x)
    r = []
    for i in a:
        if i in r or (numpy.isnan(i) and numpy.any(numpy.isnan(r))):
            continue
        else:
            r.append(i)
    return numpy.array(r)

maxalbert:
I'm still seeing this issue with latest master. Which commit should have fixed it? Unless I'm missing something I'd suggest re-opening this issue.

jaimefrio:
This is easy to fix for floats, but I don't see an easy way out for complex or structured dtypes. Will put a quick PR together and we can discuss the options there.

charris:
@jaimefrio I have it fixed for unique using
```
    if issubclass(aux.dtype.type, np.inexact):
        # nans always compare unequal, so encode as integers
        tmp = aux.searchsorted(aux)
    else:
        tmp = aux
    flag = np.concatenate(([True], tmp[1:] != tmp[:-1]))
```
but it looks like all the other operations also have problems. Maybe we need  `nan_equal, nan_not_equal` ufuncs, or maybe something in nanfuntions.

jaimefrio:
[...]

PR #30281, opened 2025-11-22
BUG: Return first NaN occurence in np.unique for complex arrays

MaanasArora:
Fixes #30113.

This is somewhat an experiment, because while it works, there is a clear performance deficit if there are many NaNs or when indices are not needed. That said, since sorting doesn't preserve nan order, I couldn't find a way yet to avoid this much... perhaps we can just wait for #30110 to finish if this is not critical.

There is some history in #2111 it seems, too, and a test had to be changed.

I'll add a release note if we decide to do this. Thanks for reviewing!

### DIFF
--- a/numpy/lib/_arraysetops_impl.py
+++ b/numpy/lib/_arraysetops_impl.py
@@ -378,7 +378,7 @@ def _unique1d(ar, return_index=False, return_inverse=False,
             return (conv.wrap(hash_unique),)
 
     # If we don't use the hash map, we use the slower sorting method.
-    if optional_indices:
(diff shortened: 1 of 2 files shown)


ngoldbaum:
ping @math-hiyoko

math-hiyoko:
I believe this bug will be resolved by the implementation in #30110, but I think this fix is good as it stands.

MaanasArora:
Thanks @math-hiyoko , yes, this is a temporary fix until that PR. I'm not sure this can/should be backported, which I suppose is the only case in which we should apply this, now that 2.4 is out?

PR #29536, opened 2025-08-10
DEP: Deprecate setting the shape attribute of a numpy array

eendebakpt:
Addresses part of https://github.com/numpy/numpy/issues/28800. Also see the discussion in https://github.com/numpy/numpy/pull/29523.

PR #30282, opened 2025-11-22
DEP, TYP: ``ndarray.shape`` setter pending deprecation

jorenham:
Similar to #30168, but for #29536, towards #28800.


jorenham:
Note that there's no need to also do this for `ndarray.dtype`, because it doesn't have a setter in the stubs (apparently stubtest didn't notice)

jorenham:
Are there any other pending deprecations that I (also (almost)) forgot about?

charris:
If more cases turn up, we can add them. Thanks Joren.

charris:
I added a triage label just so we can review the deprecations next triage meeting.

jorenham:
> Are there any other pending deprecations that I (also (almost)) forgot about?

@eendebakpt, can you perhaps think of any?

PR #30284, opened 2025-11-23
DEP: deprecate ``numpy.lib.user_array.container``

### LINKED ISSUE #30283: DEP: Deprecating `lib.user_array.container`
> This is a really old module: "Container class for backward compatibility with NumArray" (!)
> 
> We should consider deleting it entirely :)
> https://github.com/numpy/numpy/blob/06f987b7b77453aae7fa11de15f728567fccd407/numpy/tests/test_public_api.py#L196 

 _Originally posted by @shoyer in [#28313](https://github.com/numpy/numpy/issues/28313#issuecomment-2646730835)_

---

And besides it being very old, a quick [code search](https://github.com/search?q=%2F+numpy%5C.lib%5C.user_array%2F+language%3APython++NOT+is%3Aarchived+NOT+path%3A%2F%28venv%7Csite-packages%29%5C%2F%2F+NOT+path%3A%2F%5Enumpy%5C%2F%2F+NOT+repo%3Anumpy%2Fnumpy++NOT+is%3Afork&type=code) shows that it's barely used anymore.

jorenham:
Closes #30283

### DIFF
--- a/numpy/lib/_user_array_impl.py
+++ b/numpy/lib/_user_array_impl.py
@@ -54,6 +54,7 @@ class container:
 
     """
     def __init_subclass__(cls) -> None:
+        # Deprecated in NumPy 2.4, 2025-11-24
         import warnings
 
         warnings.warn(
--- a/numpy/lib/_user_array_impl.py
+++ b/numpy/lib/_user_array_impl.py
@@ -53,6 +53,16 @@ class container:
     astype
 
     """
+    def __init_subclass__(cls) -> None:
+        import warnings
+
+        warnings.warn(
+            "The numpy.lib.user_array.container class is deprecated and will be "
+            "removed in a future version.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+
     def __init__(self, data, dtype=None, copy=True):
         self.array = array(data, dtype, copy=copy)
 
--- a/doc/release/upcoming_changes/30284.deprecation.rst
+++ b/doc/release/upcoming_changes/30284.deprecation.rst
@@ -1,0 +1,3 @@
+Deprecation of ``numpy.lib.user_array.container``
+-------------------------------------------------
+The ``numpy.lib.user_array.container`` class is deprecated and will be removed in a future version.
--- a/numpy/lib/_user_array_impl.pyi
+++ b/numpy/lib/_user_array_impl.pyi
@@ -1,7 +1,7 @@
 from _typeshed import Incomplete
 from types import EllipsisType
 from typing import Any, Generic, Self, SupportsIndex, TypeAlias, overload
-from typing_extensions import TypeVar, override
+from typing_extensions import TypeVar, deprecated, override
 
 import numpy as np
 import numpy.typing as npt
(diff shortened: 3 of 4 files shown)


numpy/lib/_user_array_impl.py
```diff
@@ -53,6 +53,16 @@ class container:
     astype
 
     """
+    def __init_subclass__(cls) -> None:
```
charris, line 56: Does this work? AFAICT, it will only be called when a subclass is instantiated.

jorenham: First thing I tried was warning on import, but distutils didn't like that for some reason, so I went with this approach. 
If you prefer, I could try the module-level `__getattr__` trick?

charris: I think you can test with:
```
python numpy/lib/_user_array_impl.py
```
There is also an existing error in that test :)

Why not add the deprecation to the `__init__` ?

jorenham: > I think you can test with:
> 
> ```
> python numpy/lib/_user_array_impl.py
> ```

I kinda assumed that the only use-case of `contasiner` was as a base class. I doubt that anyone would use it directly like done in this test. But maybe I'm wrong :shrug: 

> There is also an existing error in that test :)

That's pretty funny haha

> Why not add the deprecation to the `__init__` ?

There's a higher chance of someone overriding that I guess. But I maybe I could do both; what do you think?

charris: Let's ping @stefanv, he is one of the early editors of the file.

charris:
This class is actually documented in `arrays.classes.rst`, and it comes from numeric UserArray. It seems the idea was to allow subclassing of the numeric arrays implemented in C, the original masked arrays would be an example of such use. Grok offers an example:
```
from Numeric import array, Float
from UserArray import UserArray

class MyLabeledArray(UserArray):
    def __init__(self, data, label=""):
        UserArray.__init__(self, array(data, Float))
        self.label = label

    def __repr__(self):
        return f"<MyLabeledArray {self.label}: {UserArray.__repr__(self)}>"'
```
and offers units as another application. I suspect the motivation was difficulty inheriting from C classes.

In any case, you are correct that the resulting Python class was intended as a base class. I'm not sure it wasn't a better idea than the current masked array implementation :) Have you tested `__init_subclass__`?

jorenham:
> This class is actually documented in `arrays.classes.rst`, and it comes from numeric UserArray. It seems the idea was to allow subclassing of the numeric arrays implemented in C, the original masked arrays would be an example of such use. Grok offers an example:
> 
> ```
> from Numeric import array, Float
> from UserArray import UserArray
> 
> class MyLabeledArray(UserArray):
>     def __init__(self, data, label=""):
>         UserArray.__init__(self, array(data, Float))
>         self.label = label
> 
>     def __repr__(self):
>         return f"<MyLabeledArray {self.label}: {UserArray.__repr__(self)}>"'
> ```
> 
> and offers units as another application. I suspect the motivation was difficulty inheriting from C classes.

Ohh, that goes pretty far back then!

> Have you tested `__init_subclass__`?

Yup: 

```py
from numpy.lib.user_array import container

class MyArray(container):
    pass
```

Running this `demo.py` with `spin python demo.py` (python 3.11) shows:

```
/home/joren/Workspace/numpy/demo.py:3: DeprecationWarning: The numpy.lib.user_array.container class is deprecated and will be removed in a future version.
  class MyArray(container):
```

charris:
Sounds good them. Should probably leave a note in the documentation as well, search for UserArray.

charris (approved):

charris:
Oh, and please add a comment with the date and release, something like:
```
# Deprecated in NumPy 2.4, 2025-11-24
```

jorenham:
> Oh, and please add a comment with the date and release, something like:
> 
> ```
> # Deprecated in NumPy 2.4, 2025-11-24
> ```

Sharp; I completely forgot about that

charris:
Thanks Joren.

PR #30285, opened 2025-11-23
WIP: ENH: Avoid nan check in testing routines for integer types

hmaarrfk:
For large integer arrays, this makes a big difference in the runtime.


jorenham:
This seems to have become stale, so I'm gonna close this. If you still want to work on this, let me know, and I'll re-open it.

PR #30286, opened 2025-11-24
MAINT: add ``matmul`` to ``_core.umath.__all__``

jorenham:
All other ufuncs are in there, so `matmul` probably should too.


charris:
Thanks Joren.

PR #30289, opened 2025-11-24
TYP: ``_core.overrides.set_module`` implicit re-export

jorenham:
This fixes mypy errors in e.g. `numpy/fft/_helper.py`, which imports `set_module` from `numpy._core.overrides`.


charris:
Thanks Joren.

PR #30290, opened 2025-11-24
TYP: move the ``normalize_axis_*`` function definitions from ``lib`` to ``_core``

jorenham:
It more closely matches the runtime definitions this way, and helps with IDE code navigation.


charris:
Thanks Joren.

PR #29175, opened 2025-06-11
TYP: Fix invalid inline annotations in ``lib._function_base_impl``

jorenham:
``_ : np.array`` is not a valid type annotations, because ``np.array`` is a function, and not a type expression.

PR #30291, opened 2025-11-24
TYP: ``lib._function_base_impl._quantile_ureduce_func`` inline annotation fix

jorenham:
I seem to have missed this one in #29175.


charris:
Thanks  Joren.

PR #30292, opened 2025-11-24
BUG: Blackman window has negative ends #30249

G-Guillard:
Fixes #30249 Blackman window has negative ends

### DIFF
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -3130,7 +3130,10 @@ def blackman(M):
     if M == 1:
         return ones(1, dtype=values.dtype)
     n = arange(1 - M, M, 2)
-    return 0.42 + 0.5 * cos(pi * n / (M - 1)) + 0.08 * cos(2.0 * pi * n / (M - 1))
+    res = 0.42 + 0.5 * cos(pi * n / (M - 1)) + 0.08 * cos(2.0 * pi * n / (M - 1))
+    res[0] = 0
+    res[-1] = 0
+    return res
 
 
 @set_module('numpy')
--- a/numpy/lib/tests/test_function_base.py
+++ b/numpy/lib/tests/test_function_base.py
@@ -2327,6 +2327,9 @@ class TestFilterwindows:
         # check symmetry
         assert_equal(w, flipud(w))
 
+        # check positivity
+        assert (w >= 0).all()
+
         # check known value
         if scalar < 1:
             assert_array_equal(w, np.array([]))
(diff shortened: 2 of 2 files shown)


numpy/lib/_function_base_impl.py
```diff
@@ -3130,7 +3130,10 @@ def blackman(M):
     if M == 1:
         return ones(1, dtype=values.dtype)
     n = arange(1 - M, M, 2)
-    return 0.42 + 0.5 * cos(pi * n / (M - 1)) + 0.08 * cos(2.0 * pi * n / (M - 1))
+    res = 0.42 + 0.5 * cos(pi * n / (M - 1)) + 0.08 * cos(2.0 * pi * n / (M - 1))
+    res[0] = 0
+    res[-1] = 0
```
jorenham, line 3135: That's not quite right:

```pycon
>>> import numpy as np
>>> np.blackman(1)
array([1.])
```

And in case of `np.blackman(0)` this will raise an error

G-Guillard: Those cases are handled by the tests on the M value at lines 3128:3131.

jorenham, line 3135: I'm not sure what's going on here, but this would change it :thinking: 

```pycon
>>> import numpy as np
>>> np.blackman(2+2j)
array([-1.38777878e-17+0.j        , -5.47215867e+00+0.66712036j])
```

G-Guillard: I guess that'll be a bonus from the fix. M should be an integer, not a complex, and the Blackman window should be symmetric anyway (your example would break the test on line 2328 below).

jorenham: Maybe some users are relying on this "accidental analytic continuation"? Maybe a code search could help shed some more light on this.

G-Guillard: That'd be quite weird, but I took a look anyway, and the only occurrences I found on github of a complex argument used in this function are copies of this line from `typing/tests/data/fail/lib_function_base.pyi` :
```np.blackman(1j)  # type: ignore[arg-type]```

Comment (jorenham) [reply numpy/lib/_function_base_impl.py:3135]: Nice; it should be safe enough to ignore that then. And it helps that the stubs already prohibit it.

**Review by jorenham — APPROVED** [review]

File: numpy/lib/_function_base_impl.py (hunk 1 above)
Comment (eendebakpt) [inline numpy/lib/_function_base_impl.py:3135]:
```suggestion
    x = pi * n / (M - 1)
    res = (0.42 + 0.08 * cos(2.0 * x)) + 0.5 * cos(x)
```
Works as well.

Side note: setting `res[0] = 0` changes behaviour for `blackman(3.5)`, altough I doubt it is used that way.

numpy/lib/_function_base_impl.py
eendebakpt, line 3089: This needs to be updated

eendebakpt:
Two minor notes, but lets try to get this merged.

PR #30294, opened 2025-11-24
TYP: `_core.*`: stubs for some private functions and constants

jorenham:
These were used in `_core/_ufunc_confg.py` and `lib/_shape_base_impl.py`, but because they had no stubs, they were causing errors to be reported by type-checkers. 
So this effectively gets rid of (almost) all of the annoying squiggly lines in those two modules.


charris:
Thanks Joren.

PR #30295, opened 2025-11-25
MAINT: remove ``lib._shape_base_impl._replace_zero_by_x_arrays``

jorenham:
It's not used or mentioned anywhere, and calls the removed `sometrue` function, so it wouldn't even work.


mattip:
Thanks @jorenham

PR #30296, opened 2025-11-25
TYP: ``lib.*``: stubs for some private functions used by ``_function_base_impl``

jorenham:
This gets rid of two squigglies in `lib._function_base_impl`, and is related to #30294.

Closes numpy/numtype#582


mattip:
Thanks @jorenham

PR #30297, opened 2025-11-25
MAINT: ``broadcast_shapes``: update presumed ``NPY_MAXARGS`` from 32 to 64

jorenham:
This should make `broadcast_shapes` a wee bit faster in case of >32 shapes


mattip:
Thanks @jorenham

PR #30298, opened 2025-11-25
BUG: C code generated by f2py is not C23 compliant

### LINKED ISSUE #30167: BUG: C code generated by f2py is not C23 compliant
### Describe the issue:

Looks like the C code generated by f2py is not C23 compliant and fails to compile with gcc 15.

For instance, this can be seen trying to build scikits.odes-daepack on rolling linux distros that are already at gcc 15. Editing the `meson.build` file to include a 'c_std=c17' seems to fix the issue.

The problem seems to be triggered when generating functions that take a function as a parameter. The latter gets in the signature as `f2py_func()` that is a function taking an arbitrary number of parameters in C<23 but is not the same in C23 that wants a full signature.

```python
`uv add scikits.odes` on a system with gcc 15.
```

### Error message:

```shell
N/A
```

python 3.13.9
numpy 2.3.4

### Runtime Environment:

N/A

### Context for the issue:

Not necessarily a problem, if properly documented.

Aniketsy:
Fix #30167 

This PR  updates the condition to `if r and r.strip():`, the code now properly ignores empty or whitespace only blocks. This ensures that when no arguments are provided, F2PY correctly falls back to generating `void` as the argument list, producing a valid and C23-compliant function signature.

Thankyou !

### DIFF
--- a/numpy/f2py/auxfuncs.py
+++ b/numpy/f2py/auxfuncs.py
@@ -699,7 +699,7 @@ def getcallstatement(rout):
 
 def getcallprotoargument(rout, cb_map={}):
     r = getmultilineblock(rout, 'callprotoargument', comment=0)
-    if r:
+    if r and r.strip():
         return r
     if hascallstatement(rout):
         outmess(


Aniketsy:
@rgommers  When you get a chance, could you please take a look at this? I’ve made the small changes that should address the issue, please let me know if I’m missing anything.

rgommers:
@Aniketsy could you please explain in the PR description what the problem was and why this one line changes fixes it? A diff of the generated code with/without this change would also be helpful. That way, the reviewer as well as people who may look at the PR in the future can easily tell the intent and effect here.

Aniketsy:
@rgommers Thanks for pointing this out, apologies for not including the detailed description earlier. I’ve updated the description.
I’ll make sure to include such details in all future PRs. please let me know if anything else needs to be improved.

rgommers:
@Aniketsy I'm still struggling with that description - it leaves a lot of work to the reviewer. What would be helpful is to state how you tested this. E.g., did you reproduce the `scikits.odes` failures in an environment with GCC 15, and is it fixed with this PR? If so, showing the diff of the generated C code by f2py before/after would be helpful.

Aniketsy:
@rgommers  I apologize for creating extra work for the reviewers. I tried to test this locally, but I wasn’t able to get it working in my setup. I also explored adding unit tests but couldn’t determine the correct approach.

If you could share the documentation or workflow for testing this locally, that would be very helpful. I’ll also make sure not to create extra work for you going forward.

mattip:
How did you come up with the solution if you cannot reproduce the problem?

Aniketsy:
I looked into the underlying reason for the issue and what could be causing the failure. After that, I tried to test the fix locally, but I wasn’t able to get it working in my setup. If my understanding is incorrect please let me know and I’ll look into it again.

rgommers:
I'm not sure if your understanding is incorrect, hard to tell without actually doing the reproducing/testing that you didn't manage here - which is most of the work. Since I don't have time to do that, I'll move this to draft. 

Note that you don't actually need GCC 15 to test this yourself, if that was your problem. Any compiler is fine, since `f2py` gives the same output. You could try to build `scikits.odes` with and without this patch, and just diff the files generated by `f2py`.

PR #30301, opened 2025-11-25
BUG: recarray, masked_array and datetime break fill value calculation

### LINKED ISSUE #29818: BUG: recarray, masked_array and datetime break fill value calculation
### Describe the issue:

I am unable to sort a masked_array of records, as the records contain timestamps, and this breaks the minimum_fill_value function.

This might either be due to the unusual wrapping of the value into an extra array that was introduced by afcc560 (see Example 1) or it might be that there is an inconsistency on array construction in when unpacking values in the np.array function (see Example 2) which is a property that afcc560 relies on. 

```python
import numpy as np
import datetime as dt
# Example 1
rec = np.array([(dt.date(2025,4,1),)], dtype=[('foo','<M8[D]')])
ma = np.ma.masked_array(rec)
np.sort(ma)
# Example 2
np.array((np.array("Hello"),), dtype=[('foo', '<U8')]) # This works 
np.array((np.array(dt.date(2025,4,1)),), dtype=[('foo', '<M8[D]')]) # This does not
```

### Error message:

```shell
  File "..../numpy_bug.py", line 20, in <module>
    np.ma.minimum_fill_value(ma)
  File "/usr/lib/python3/dist-packages/numpy/ma/core.py", line 347, in minimum_fill_value
    return _extremum_fill_value(obj, min_filler, "minimum")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/numpy/ma/core.py", line 296, in _extremum_fill_value
    return _recursive_fill_value(dtype, _scalar_fill_value)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/numpy/ma/core.py", line 211, in _recursive_fill_value
    return np.array(vals, dtype=dtype)[()]  # decay to void scalar from 0d
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: Could not convert object to NumPy datetime
```

1.26.4

### Runtime Environment:

_No response_

### Context for the issue:

_No response_

Aniketsy:
fix #29818 

- Handles both datetime types correctly: adds logic to detect datetime64 fields and properly convert both Python datetime objects and integer extremum values to the appropriate NumPy type.

- Ensures proper structured scalar creation: converts the field value list into a tuple before array creation, ensuring a correct 0-d structured scalar rather than an unintended 1-d array.

Thankyou !


numpy/ma/tests/test_core.py
```diff
@@ -4975,9 +4984,10 @@ def test_make_mask(self):
         bools = [True, False]
         dtypes = [MaskType, float]
         msgformat = 'copy=%s, shrink=%s, dtype=%s'
-        for cpy, shr, dt in itertools.product(bools, bools, dtypes):
-            res = make_mask(nomask, copy=cpy, shrink=shr, dtype=dt)
-            assert_(res is nomask, msgformat % (cpy, shr, dt))
+        # renamed 'dt' to 'dtype'
```
Aniketsy: Not sure if this is the correct practice, I applied this change after the ci- checks fails.

numpy/ma/core.py
```diff
@@ -223,10 +224,17 @@ def _recursive_fill_value(dtype, f):
...
-        vals = tuple(
-                np.array(_recursive_fill_value(dtype[name], f))
-                for name in dtype.names)
-        return np.array(vals, dtype=dtype)[()]  # decay to void scalar from 0d
+        vals = []
+        for name in dtype.names:
+            field_dtype = dtype[name]
+            val = _recursive_fill_value(field_dtype, f)
+            if np.issubdtype(field_dtype, np.datetime64):
+                if isinstance(val, (datetime.date, datetime.datetime)):
```
jorenham: `datetime.datetime` is a subclass of `datetime.date`:

```suggestion
                if isinstance(val, datetime.date):
```

numpy/ma/core.py
```diff
@@ -20,6 +20,7 @@
 
 """
 import builtins
+import datetime
```
jorenham: the convention is to `import datetime as dt`

numpy/ma/core.py
```diff
@@ -223,10 +224,17 @@ def _recursive_fill_value(dtype, f):
...
+        vals = []
+        for name in dtype.names:
+            field_dtype = dtype[name]
+            val = _recursive_fill_value(field_dtype, f)
+            if np.issubdtype(field_dtype, np.datetime64):
+                if isinstance(val, (datetime.date, datetime.datetime)):
+                    val = np.datetime64(val)
+                elif isinstance(val, (int, np.integer)):
+                    val = np.array(val).astype(field_dtype)
+            vals.append(np.array(val))
```
jorenham: it would be better to `val = np.array(val)` in an `else:`, to avoid doing it twice.

numpy/ma/tests/test_core.py
```diff
@@ -5308,9 +5318,10 @@ class TestMaskedObjectArray:
 
     def test_getitem(self):
         arr = np.ma.array([None, None])
-        for dt in [float, object]:
-            a0 = np.eye(2).astype(dt)
-            a1 = np.eye(3).astype(dt)
+        # renamed 'dt' to 'dtype'
```
jorenham:
```suggestion
```

numpy/ma/tests/test_core.py
jorenham: > Not sure if this is the correct practice, I applied this change after the ci- checks fails.

there's no need :)

```suggestion
```

numpy/ma/tests/test_core.py
```diff
@@ -2272,6 +2273,14 @@ def test_check_on_scalar(self):
         assert_raises(TypeError, _check_fill_value, 1e+20, int)
         assert_raises(TypeError, _check_fill_value, 'stuff', int)
 
+    def test_fill_value_datetime_structured(self):
+        # gh-29818
+        rec = np.array([(dt.date(2025, 4, 1),)], dtype=[('foo', '<M8[D]')])
+        ma = np.ma.masked_array(rec)
+        np.sort(ma)
+        res = np.ma.minimum_fill_value(ma)
+        assert isinstance(res['foo'], np.datetime64)
```
jorenham, line 2284: The added special-casing for `int | np.integer` could also use a test or two, and the same goes for `dt.datetime`.

jorenham (changes requested):

Aniketsy:
@jorenham  Thanks for the review. I’ve applied changes as per suggestions, please let me know if there’s anything else that needs improvement.

jorenham:
there are some ruff errors

Aniketsy:
Ci checks are green.

numpy/ma/core.py
```diff
@@ -223,10 +224,22 @@ def _recursive_fill_value(dtype, f):
         # for integer casts, this allows the use of 99999 as a fill value
         # for int8.
         # TODO: This is probably a mess, but should best preserve behavior?
```
jorenham: I suppose this comment can be (re)moved now

jorenham (approved):
this looks good now :)

numpy/ma/core.py
Aniketsy: Sure, I’ll remove this comment.

jorenham:
In it goes; thanks @Aniketsy !

PR #30304, opened 2025-11-25
ENH: is monotonic function to simplify checking ascending and descending values in arrays.

Vybh:
is monotonic function to simplify checking ascending and descending values in arrays.

### DIFF
--- a/doc/release/upcoming_changes/00000.new_function.rst
+++ b/doc/release/upcoming_changes/00000.new_function.rst
@@ -1,0 +1,7 @@
+``numpy.ismonotonic`` helper
+----------------------------
+Added the `numpy.ismonotonic` routine, which reports whether the flattened
+input follows an increasing or decreasing trend and optionally enforces
+strictly monotonic behaviour.
+
+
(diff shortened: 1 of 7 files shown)


WarrenWeckesser:
Such a function certainly has uses (e.g. some interpolation code in SciPy expects the input to be monotone), but I don't know if it will be considered sufficiently widely useful to be included in NumPy.

If it was included in NumPy, it should act along 1-d slices of the input array, and accept an `axis` parameter.  It could be implemented as a [gufunc](https://numpy.org/doc/stable/reference/c-api/generalized-ufuncs.html).

An argument against including in NumPy might be that, for any given use-case, "it is practically a one-liner".  But I would argue that if it were included in NumPy, these days the underlying code should be SIMDeified where possible (e.g. integers, float, double).  I've experiemented with a simple C++ function `is_increasing()` using XSIMD in https://github.com/WarrenWeckesser/xsimd_funcs/blob/main/include/xsimd_funcs/ordered.hpp, so it is certainly possible.

WarrenWeckesser:
@Vybh, I forgot to say in my earlier comment: proposals for new functions should be brought up on the [mailing list](https://mail.python.org/mailman3/lists/numpy-discussion.python.org/).

PR #30308, opened 2025-11-26
BUG: Fix view semantics in user_array.container (slices returned copies)

### LINKED ISSUE #30307: user_array.container breaks view semantics: in-place slice modifications fail silently
### Describe the issue:

I discovered a critical logic bug in the numpy.lib.user_array.container class (specifically in the internal implementation _user_array_impl.py).

The container breaks standard NumPy array semantics: slicing the container returns a deep copy of the data rather than a view.

This causes in-place operations on slices (e.g., container[0:5] = 1) to fail silently. The operation modifies the returned copy, which is immediately discarded, leaving the original data unchanged. This effectively makes it impossible to use this class for standard tasks like image masking or windowed operations without data corruption.

The issue stems from the _rc helper method initializing new instances with the default copy=True.

```python
import numpy as np
from numpy.lib.user_array import container

# 1. Create a simple container wrapping a zeros array
data = np.zeros(5)
wrapped = container(data)

# 2. Create a slice (First 3 elements)
# EXPECTATION: This should be a view. Modifying it should modify 'wrapped'.
my_slice = wrapped[0:3]

# 3. Modify the slice
my_slice[:] = 999

# 4. Check the original array
print(f"Slice values:    {my_slice.array}")
print(f"Original values: {wrapped.array}")

if wrapped.array[0] == 0:
    print("\n[FAIL] Bug reproduced: Original array was NOT updated (Slice was a copy).")
else:
    print("\n[PASS] Original array updated.")
```

### Error message:

```shell

```

NumPy Version: 2.0.2
Python Version: 3.12.12 (main, Oct 10 2025, 08:52:57) [GCC 11.4.0]

### Runtime Environment:

_No response_

### Context for the issue:

I am using this class for image preprocessing tasks where I need to wrap image data and mask out specific sensitive regions using slicing (e.g., img[x:y] = 0). Because of this bug, the masking operations fail silently, and the raw data is passed through unredacted. This poses a significant data safety risk.

pareshjoshij:
Closes #30307

This PR fixes a critical logic bug in numpy.lib.user_array.container where slicing operations returned a deep copy of the data instead of a view. This behavior violated standard NumPy array semantics and caused in-place operations on slices (e.g., masking) to fail silently, as detailed in the issue user_array.container breaks view semantics: in-place slice modifications fail silently (#30307).

The Fix

Modified the internal _rc (return container) helper method in numpy/lib/_user_array_impl.py.

Before: return self.__class__(a) (Defaults to copy=True)

After: return self.__class__(a, copy=False)

Passing copy=False ensures that operations returning new container instances (like slicing) maintain a view of the original data rather than allocating new memory.

### DIFF
--- a/numpy/lib/_user_array_impl.py
+++ b/numpy/lib/_user_array_impl.py
@@ -255,7 +255,7 @@ class container:
         if len(shape(a)) == 0:
             return a
         else:
-            return self.__class__(a)
+            return self.__class__(a, copy=False)
 
     def __array_wrap__(self, *args):
         return self.__class__(args[0])


jorenham:
Like I said in https://github.com/numpy/numpy/issues/30307#issuecomment-3579644294, `container` is deprecated (see #30284).

Also, I wouldn't call this a "critical bug", but rather a minor inconsistency.

So those two things considered, I don't see the need for this change. But I'll leave this open for a bit in case there any other maintainers who think otherwise.

PR #30312, opened 2025-11-26
MAINT: Implement some RAII classes and use them in stringdtype/casts.cpp

WarrenWeckesser:
This PR introduces a few simple RAII classes for managing the acquisition and release of some resources in C++ code. These are private classes, for internal use.

The classes and the acquire/release functions managed by the class are:

| Class            | Acquire                             | Release                             |
| ---------------- | ----------------------------------- | ----------------------------------- |
| `EnsureGIL`      | `gil_state = PyGILState_Ensure()`   | `PyGILState_Release(gil_state)`     |
|`SaveThreadState` | `thread_state = PyEval_SaveThread()` | `PyEval_RestoreThread(thread_state)` |
| `NpyStringAcquireAllocator` | `allocator = NpyString_acquire_allocator(descr)` | `NpyString_release_allocator(allocator)` |

The method `NpyStringAcquireAllocator.allocator()` returns the allocator instance for use in the scope where the instance is created.

In this PR, `EnsureGIL` and `NpyStringAcquireAllocator` are put to use in `numpy/_core/src/multiarray/stringdtype/casts.cpp`. 

This change evolved out of changes that I have been working on in `unique.cpp` in https://github.com/numpy/numpy/pull/30180.  This PR is a self-contained step towards the use of RAII, with an application that is independent of the work on the hash-based `unique` code.

 `SaveThreadState` is not used here, but it will be in https://github.com/numpy/numpy/pull/30180.


numpy/_core/src/common/raii_utils.hpp
```diff
@@ -0,0 +1,151 @@
...
+class EnsureGIL
+{
+    PyGILState_STATE gil_state;
+
+public:
+
+    EnsureGIL() {
+        gil_state = PyGILState_Ensure();
+    }
+
```
eric-wieser, line 61: You probably want `EnsureGIL(EnsureGIL const&) = delete;` too (and the assignment version, and perhaps also the `&&` version), to avoid doubly-releasing.

WarrenWeckesser: Thanks, done.

seberg (approved):
LGTM, the only bikeshed I can think of is whether `SafeThread` really reads clearly for releasing the GIL, and that seems not worth worrying about (and easy to change).

charris:
I put this in for 2.4.x so that potential backports will be easier. Thanks Warren.

numpy/_core/src/common/raii_utils.hpp
```diff
@@ -0,0 +1,166 @@
...
+// Forward declarations not currently in a header.
+// XXX Where should these be moved?
+NPY_NO_EXPORT npy_string_allocator *
+NpyString_acquire_allocator(const PyArray_StringDTypeObject *descr);
+NPY_NO_EXPORT void
+NpyString_release_allocator(npy_string_allocator *allocator);
+
+
+namespace np { namespace raii {
```
WarrenWeckesser, line 25: Note to self, and to anyone who edits this file in the future: this could have been written more succinctly as
```
namespace np::raii {
```
It isn't worth a PR just for this change, but it would be a reasonable bit of code maintenance the next time this file is updated.

Issue #30311, opened 2025-11-26
TSK: Dropping Python 3.11 followup


ngoldbaum:
When NumPy 2.5 is branched (hopefully this weekend) we will be able to drop Python 3.11 support. That will allow us to do some long-term cleanup:

* [ ] Drop `numpy.distuils`
* [ ] Clean up CI and drop 3.11 builds or any builds checking numpy.distutils.
* [ ] Update type stubs to account for [PEP 695](https://peps.python.org/pep-0695/).
* [ ] Replace `warnings.warn` `stacklevel` keyword argument with [`skip_file_prefixes`](https://docs.python.org/3/library/warnings.html#warnings.warn).

Aniketsy:
Hi, I’d be happy to take care of the remaining three items, dropping `numpy.distutils`, `cleaning up the CI` (removing 3.11 builds), and updating the warnings.
Please let me know if you're okay with that, and I’ll proceed accordingly.

rgommers:
>  Unpin setuptools, gets rid of four security warnings. Do we still need it?

We will no longer need that pin. We do still need `setuptools` as a test dependency for some compile tests. EDIT: it turns out those `setup.py` files are also no longer used, all compile tests use `meson` even if a `setup.py` file is still present for example purposes.

> Replace `warnings.warn` `stacklevel` keyword argument with [`skip_file_prefixes`]

This looks like a separate thing from the topic in this issue, and not a good idea to do wholesale. `skip_file_prefixes` may sometimes be useful, and sometimes just a more cumbersome and verbose way of achieving the same thing - it depends on code structure.

ngoldbaum:
> skip_file_prefixes may sometimes be useful, and sometimes just a more cumbersome and verbose way of achieving the same thing - it depends on code structure.

From my point-of-view it's a better, less error-prone construction. Can you point to a specific example where you'd prefer to use `stacklevel`?

IMO the goal with `stacklevel` is to make sure warnings are generated with context in user code. To get that right, you need to either analyze the code or, more likely, trigger the warning and manually adjust the stacklevel so the warning context shows up in user code. If someone ever refactors the code so the stacklevel changes, they also need to adjust the warning.

Here's what I mean:

```python
# my_package/library.py
import warnings

def code_that_generates_warning():
    warnings.warn("warning!", UserWarning, stacklevel=4)

def library_function():
    code_that_generates_warning()

def main():
    library_function()
```

[...]

PR #30319, opened 2025-11-28
TYP, STY: PEP 695 and other Ruff py312+ fixes

jorenham:
Follow-up of #30339 (and #30357)

After branching 2.4 we'll be able to drop support Python 3.11, opening the doors to the Python 3.12+ type parameter syntax from [PEP 695](https://peps.python.org/pep-0695/).

Ruff actually requires that we use this new syntax now. But it waa a bit disappointing to see that Ruff's PEP695 auto-fixes (pyupgrade I believe) ignored most cases, and the ones it did "fix" it actually broke the code. So in the end I had to do almost everything by hand.

When I locally bump `requires-python`  to `>=3.12` in `pyproject.toml`, ruff, mypy, and stubtest all pass. So CI is expected to fail until that's also the case on main, hence the draft status.

There were also some other ruff errors, most of them related to now-irrelevant `if sys.version_info` code paths, that this addresses. I put those in here because I ran `ruff lint --fix`, which has to happen either way, and given the size of this PR it's basically pocket chance anyway.

Also, we no longer need `typing_extensions` (I think), so it's no longer a test dependency.

Towards #30311


jorenham:
> Copilot reviewed 126 out of 128 changed files in this pull request and generated no comments.

sheesh; boring! :yum:

jorenham:
luckily the mypy_primer output is less boring :sweat_smile: 
(I'll look at it tomorrow)

jorenham:
@Dr-Irv this change from `numpy.typing.NDArray` as `_: TypeAlias` to a `type _` (i.e. a `typing.TypeAliasType` instance), the [`check` function](https://github.com/pandas-dev/pandas-stubs/blob/b983c83b138f20222fee3d80d324ef1bf10e61a2/tests/__init__.py#L94) no longer seems to work (I'm guessing also at runtime).

A fancy solution would be to use `klass: typing_extensions.TypeForm` ([docs](https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.TypeForm)) and then do something like `klass = getattr(klass, "__value__", klass)`. Replacing the `npt.NDArray` with `np.ndarray` in the `check` calls could also work. 

Is that ok for you? Reverting `NDArray` to a `: TypeAlias` is an option, but would only be a temporary one, since I'm planning on adding a second optional type-parameter for the shape-type to `NDArray`, which given their ordering will require us to use the new `type _` syntax either way.

Dr-Irv:
> @Dr-Irv this change from `numpy.typing.NDArray` as `_: TypeAlias` to a `type _` (i.e. a `typing.TypeAliasType` instance), the [`check` function](https://github.com/pandas-dev/pandas-stubs/blob/b983c83b138f20222fee3d80d324ef1bf10e61a2/tests/__init__.py#L94) no longer seems to work (I'm guessing also at runtime).

Can you indicate where it no longer works?  I think you're looking at where we pass `npt.NDArray` as the `klass` argument, and I think we are only doing that in one particular test file `test_plotting.py`, so if that is what you are referring to, we can change our test code to change the class we are checking to be deriving from `np.ndarray`, rather than `np.typing.NDArray` .  We may also have to change some of our type definitions in `pandas-stubs/_typing.pyi` .   If we make that change now, will the tests still work with the current released version of `numpy` ?

jorenham:
> Can you indicate where it no longer works? I think you're looking at where we pass `npt.NDArray` as the `klass` argument, and I think we are only doing that in one particular test file `test_plotting.py`, so if that is what you are referring to, we can change our test code to change the class we are checking to be deriving from `np.ndarray`, rather than `np.typing.NDArray` .

Yea it's indeed `test_plotting.py` I'm talking about. See https://github.com/numpy/numpy/pull/30319#issuecomment-3603316967 for the specific line numbers.

>  We may also have to change some of our type definitions in `pandas-stubs/_typing.pyi` . If we make that change now, will the tests still work with the current released version of `numpy` ?

There should be no need to change anything else; it's only the `type[npt.NDArray]` that this changes. So this only involves only applications of "meta-typing" such as the `check` function, where a *type expression* itself is used as a function argument. I don't expect this to affect any `.pyi` stub files.

---

Note that this change is planned for NumPy 2.5, so that'll be in roughly 7 months, so there's no hurry :)

Dr-Irv:
> There should be no need to change anything else; it's only the `type[npt.NDArray]` that this changes. So this only involves only applications of "meta-typing" such as the `check` function, where a _type expression_ itself is used as a function argument. I don't expect this to affect any `.pyi` stub files.

So a `TypeAlias` of a `_type` is OK to pass to the `check` function?

jorenham:
> > There should be no need to change anything else; it's only the `type[npt.NDArray]` that this changes. So this only involves only applications of "meta-typing" such as the `check` function, where a _type expression_ itself is used as a function argument. I don't expect this to affect any `.pyi` stub files.
> 
> So a `TypeAlias` of a `_type` is OK to pass to the `check` function?

Yea exactly, an "old" `_: TypeAlias` of a type (subscripted or not) is considered as a `type` by (at least) mypy.

jorenham:
I expect that #30364 will resolve the new mypy_primer errors

charris:
Needs rebase.

jorenham:
> I expect that #30364 will resolve the new mypy_primer errors

In case it wasn't clear yet; I think we should merge #30364 first

charris:
Needs another rebase.

jorenham:
> Needs another rebase.

... and a pretty tricky on at that, which will have to wait until later today (+6 hours or something)

jorenham:
Since mypy 1.19 it now requires `librt`, which is what's causing the PPC64LE job to fail. I'm not sure why this wasn't an issue before, though :thinking:. Anyway, I guess I'll address that in a separate PR.

EDIT: here it is: https://github.com/numpy/numpy/pull/30385

jorenham:
It looks like that first mypy_primer pandas error is the result of pandas having pinned the python version to 3.11 in their mypy config:
https://github.com/pandas-dev/pandas/blob/e450f0c52a564d1e813d09681311c87d57c81ee6/pyproject.toml#L661
So there's nothing we can do about that from our side of things.

jorenham:
All ready now!

environment.yml
```diff
@@ -7,7 +7,7 @@ name: numpy-dev
 channels:
   - conda-forge
 dependencies:
-  - python=3.12  # need to pin to avoid issues with builds
+  - python=3.12 # need to pin to avoid issues with builds
```
charris, line 10: double spaces, IIRC, is PEP 8. But not worth arguing about.

charris:
Thanks Joren.

environment.yml
jorenham: hmm yea not sure what happened there

charris: My guess is yml vs py.

PR #30320, opened 2025-11-28
Update README.md

Mohataseem89:
just formatted text a little.
Improve readability.


mattip:
I understand the urge to contribute, but it would be better to triage some old issues. This kind of contribution is not really helpful. The phrases are part of a single sentence and do not need capitalization.

PR #30324, opened 2025-11-28
TYP: ``ndenumerate`` generic type parameter default

jorenham:
This way, type-checkers configured in strict mode will allow you to write `_: np.ndenumerate` instead of `_: np.ndenumerate[Any]`. This matters because the latter isn't valid at runtime, raising a `TypeError`.


charris:
Thanks Joren.

PR #30326, opened 2025-11-28
TYP: ``ma.mrecords.MaskedRecords`` generic type parameter defaults

jorenham:
Same story as #30324, but more esoteric.


charris:
Thanks Joren.

PR #30327, opened 2025-11-28
TYP: ``_core._umath_tests`` module stubs

jorenham:
This resolves a bunch of squiggly lines in `numpy/_core/tests/test_*.py`, and makes the process of working on these umath tests (in an IDE / with an LSP) a bit nicer in general :). It could also be viewed as a form of internal documentation.


charris:
Thanks Joren.

PR #30329, opened 2025-11-29
MAINT: Use RAII objects in unique.cpp to ensure safe resource management.

WarrenWeckesser:
The functions `unique_numeric()`, `unique_string()` and `unique_vstring()` in `unique.cpp` are updated to ensure that if an exception is thrown during processing of the `std::unordered_set()` (e.g. throwing `std::bad_alloc`), the GIL, refcounts, and NpyString allocators are properly restored.

<strike>Another RAII object is added, ScopedINCREF, that increments the refcount of an object that must be used for the duration of a scope. The refcount is decreased when the scope is exited.</strike>

This addresses the issue about `unique()` reported in https://github.com/numpy/numpy/issues/30177.

This PR replaces https://github.com/numpy/numpy/pull/30180.  I mentioned over there that I would add handling of `datetime64` and `timedelta64`, but those changes didn't make it into this PR.  I scaled back the scope so that this PR should preserve the existing behavior.  `datetime64` and `timedelta64` can be handled in a follow-up PR.

### DIFF
--- a/numpy/_core/src/common/raii_utils.hpp
+++ b/numpy/_core/src/common/raii_utils.hpp
@@ -25,6 +25,46 @@ NpyString_release_allocator(npy_string_allocator *allocator);
 namespace np { namespace raii {
 
 //
+// RAII for holding a Python reference.
+// This is a wrapper for Py_INCREF/Py_DECREF.
+//
+// In C++ code, use this at the beginning of a scope, e.g.
+//
+//     {
+//         np::raii::ScopedINCREF(obj);
+//         [code that requires holding a reference to obj]
+//     }
+//
+// instead of
+//
+//     Py_INCREF(obj);
+//     [code that requires holding a reference to obj]
+//     Py_DECREF(obj);
+//
+// This ensures that Py_DECREF(obj) is called, even if the wrapped
+// code throws an exception or executes a return or a goto.
+//
+class ScopedINCREF
+{
+    PyObject* _obj;
+
+public:
+
+    ScopedINCREF(PyObject *obj) : _obj(obj) {
+        Py_INCREF(_obj);
+    }
+
+    ~ScopedINCREF() {
+        Py_DECREF(_obj);
+    }
+
+    ScopedINCREF(const ScopedINCREF&) = delete;
+    ScopedINCREF(ScopedINCREF&& other) = delete;
+    ScopedINCREF& operator=(const ScopedINCREF&) = delete;
+    ScopedINCREF& operator=(ScopedINCREF&&) = delete;
+};
+
+//
 // RAII for PyGILState_* API.
 //
 // In C++ code, use this at the beginning of a scope, e.g.
--- a/numpy/_core/src/common/raii_utils.hpp
+++ b/numpy/_core/src/common/raii_utils.hpp
@@ -25,46 +25,6 @@ NpyString_release_allocator(npy_string_allocator *allocator);
 namespace np { namespace raii {
 
 //
-// RAII for holding a Python reference.
-// This is a wrapper for Py_INCREF/Py_DECREF.
-//
-// In C++ code, use this at the beginning of a scope, e.g.
-//
-//     {
-//         np::raii::ScopedINCREF(obj);
-//         [code that requires holding a reference to obj]
-//     }
-//
-// instead of
-//
-//     Py_INCREF(obj);
-//     [code that requires holding a reference to obj]
-//     Py_DECREF(obj);
-//
-// This ensures that Py_DECREF(obj) is called, even if the wrapped
-// code throws an exception or executes a return or a goto.
-//
-class ScopedINCREF
-{
-    PyObject* _obj;
-
-public:
-
-    ScopedINCREF(PyObject *obj) : _obj(obj) {
-        Py_INCREF(_obj);
-    }
-
-    ~ScopedINCREF() {
-        Py_DECREF(_obj);
-    }
-
-    ScopedINCREF(const ScopedINCREF&) = delete;
-    ScopedINCREF(ScopedINCREF&& other) = delete;
-    ScopedINCREF& operator=(const ScopedINCREF&) = delete;
-    ScopedINCREF& operator=(ScopedINCREF&&) = delete;
-};
-
-//
 // RAII for PyGILState_* API.
 //
 // In C++ code, use this at the beginning of a scope, e.g.
@@ -166,9 +126,11 @@ public:
 //
 // Instead of
 //
+//   Py_INCREF(descr);
 //   npy_string_allocator *allocator = NpyString_acquire_allocator(descr);
 //   [code that uses allocator]
 //   NpyString_release_allocator(allocator);
+//   Py_DECREF(descr);
 //
 // use
 //
@@ -179,16 +141,19 @@ public:
 //
 class NpyStringAcquireAllocator
 {
+    PyArray_StringDTypeObject *_descr;
     npy_string_allocator *_allocator;
 
 public:
 
-    NpyStringAcquireAllocator(PyArray_StringDTypeObject *descr) {
-        _allocator = NpyString_acquire_allocator(descr);
+    NpyStringAcquireAllocator(PyArray_StringDTypeObject *descr) : _descr(descr) {
+        Py_INCREF(_descr);
+        _allocator = NpyString_acquire_allocator(_descr);
     }
 
     ~NpyStringAcquireAllocator() {
         NpyString_release_allocator(_allocator);
+        Py_DECREF(_descr);
     }
 
     NpyStringAcquireAllocator(const NpyStringAcquireAllocator&) = delete;
--- a/numpy/_core/src/multiarray/unique.cpp
+++ b/numpy/_core/src/multiarray/unique.cpp
@@ -370,9 +370,8 @@ unique_vstring(PyArrayObject *self, npy_bool equal_nan)
         std::invoke([&]() -> const set_type {
             PyArray_StringDTypeObject *descr =
                 reinterpret_cast<PyArray_StringDTypeObject *>(PyArray_DESCR(self));
-            np::raii::ScopedINCREF incref_descr(reinterpret_cast<PyObject *>(descr));
-            np::raii::SaveThreadState save_thread_state{};
             np::raii::NpyStringAcquireAllocator alloc(descr);
+            np::raii::SaveThreadState save_thread_state{};
 
             set_type hashset(std::min(isize, HASH_TABLE_INITIAL_BUCKETS), hash, equal);
 
@@ -407,9 +406,8 @@ unique_vstring(PyArrayObject *self, npy_bool equal_nan)
         reinterpret_cast<PyArray_StringDTypeObject *>(PyArray_DESCR(res_obj));
 
     {
-        np::raii::ScopedINCREF incref_res_descr(reinterpret_cast<PyObject *>(res_descr));
-        np::raii::SaveThreadState save_thread_state{};
         np::raii::NpyStringAcquireAllocator alloc(res_descr);
+        np::raii::SaveThreadState save_thread_state{};
 
         char *odata = PyArray_BYTES(res_obj);
         npy_intp ostride = PyArray_STRIDES(res_obj)[0];
--- a/numpy/_core/src/multiarray/unique.cpp
+++ b/numpy/_core/src/multiarray/unique.cpp
@@ -58,10 +58,6 @@ empty_array_like(PyArrayObject *arr, npy_intp length)
                 NULL                    // obj
             )
         );
-    if (res_obj == NULL) {
-        Py_DECREF(descr);
-        return NULL;
-    }
     return res_obj;
 }
 
--- a/numpy/_core/src/multiarray/unique.cpp
+++ b/numpy/_core/src/multiarray/unique.cpp
@@ -502,21 +502,23 @@ array__unique_hash(PyObject *NPY_UNUSED(module),
         return NULL;
     }
 
+    PyObject *result = NULL;
     try {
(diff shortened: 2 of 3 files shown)


numpy/_core/src/common/raii_utils.hpp
```diff
@@ -24,6 +24,46 @@ NpyString_release_allocator(npy_string_allocator *allocator);
...
+// instead of
+//
+//     Py_INCREF(obj);
+//     [code that requires holding a reference to obj]
+//     Py_DECREF(obj);
+//
+// This ensures that Py_DECREF(obj) is called, even if the wrapped
+// code throws an exception or executes a return or a goto.
+//
+class ScopedINCREF
```
seberg: For this one I would prefer the idea of an `OwnedRef` or so object that allows a `.get()` and a `.release()` (one giving the `PyObject *`, the other returning it but invalidating itself).
That is the pattern I have seen around, I think.

The assumption would then be that creation/assignment is given a reference, I think, but with `Py_NEWREF()` that is OK?
(Or maybe we can make it obvious with methods `.takeref()` and `newref()`, or so, but dunno?)

WarrenWeckesser: Something like that should be fine.  But looking at this again, I see that currently I only use `ScopedINCREF(descr)` in two places. Those are in `unique_vstring()`, and they hold a reference to `descr` while it is used in `NpyStringAcquireAllocator` to ensure that `descr` stays live while the allocator created by `NpyString_acquire_allocator()` is in use.

So now I think incrementing and decrementing the refcount of `descr` should be built in to `NpyStringAcquireAllocator`.  With that change, there would be no need for a separate `ScopedINCREF` or `OwnedRef` class--at least not in this PR.  Do you see anything wrong with that?

WarrenWeckesser: I made that change, so `ScopedINCREF` is no longer part of this PR.

WarrenWeckesser:
Currently the PR attempts to make just enough changes to ensure that resources are cleaned up correctly if a C++ exception occurs while using the `std::unordered_set`.

There is more that could be done here:
* `unique_numeric()` and `unique_string()` are very similar, and with a bit of refactoring and additional templating could probably be made one function. (`unique_vstring()` is more involved.)
* There is a pretty significant inefficiency in the handling of variable strings.  A full copy of the input array containing the unpacked string variables ends up getting stored in a C++ vector. This happens because the set of unique values is implemented with `std::unordered_set`, with the keys being *pointers* to the strings.  The vector of unpacked strings holds the strings that the pointers in the set point to.
  An alternative to this would be to use the unpacked strings themselves as the keys of the `std::unordered_set`.  This avoids the full copy of the input array, because only the unique strings are maintained in memory while creating the set of unique values.  Memory-wise, this would be no worse than the current implementation, and often would be much better.
* As mentioned in the opening comment, `datetime64` and `timedelta64` could be added to the implementation, and the handling of `NaT` made consistent with how `NaN` is handled for the floating point and complex types.

Any of these could be addressed with more work in this PR, or could be handled in follow-up PRs.

WarrenWeckesser:
*Edited...*

<strike>In `unique_vstring()`, the constructor of the `std::vector` needs to be protected too, but currently it is not.</strike>

Correction: that is not a problem.  No additional resources are held at that point, so the `std::vector` declaration is OK where it is.

numpy/_core/src/multiarray/unique.cpp
```diff
@@ -204,157 +220,112 @@ unique_numeric(PyArrayObject *self, npy_bool equal_nan)
...
-    char *idata = PyArray_BYTES(self);
-    npy_intp istride = PyArray_STRIDES(self)[0];
-    for (npy_intp i = 0; i < isize; i++, idata += istride) {
-        hashset.insert((T *)idata);
-    }
+    // Create hashset within the scope of a lambda to allow RAII objects
+    // to manage resources.
+    using set_type = std::unordered_set<T *, decltype(hash), decltype(equal)>;
+    const set_type hashset =
+        std::invoke([&]() -> const set_type {
```
eric-wieser: This seems like a slightly strange use of `std::invoke` to me; [CppCoreGuidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es28-use-lambdas-for-complex-initialization-especially-of-const-variables) suggest that `()` alone is fine (though do not mention `std::invoke` here).

WarrenWeckesser: Yeah, that is simpler, so probably better. I guess I got led astray after reading some suggestions that `std::invoke` was somehow more "readable".

I pushed a fix to call the lambda the usual way.

numpy/_core/src/multiarray/unique.cpp
```diff
@@ -204,157 +220,112 @@ unique_numeric(PyArrayObject *self, npy_bool equal_nan)
...
+            // number of elements in the input array
+            npy_intp isize = PyArray_SIZE(self);
+            set_type hashset(std::min(isize, HASH_TABLE_INITIAL_BUCKETS), hash, equal);
+            char *idata = PyArray_BYTES(self);
+            npy_intp istride = PyArray_STRIDES(self)[0];
+            for (npy_intp i = 0; i < isize; i++, idata += istride) {
+                hashset.insert(reinterpret_cast<T *>(idata));
+            }
+            return hashset;
+        }();
```
seberg: Hmmm, is the lambda worth it?  As far as I understand it is just to make it easier to create the hash within the nogil block?
Seems to me that it is probably OK to just not do that?

numpy/_core/src/common/raii_utils.hpp
```diff
@@ -126,9 +126,11 @@ class SaveThreadState
 //
 // Instead of
 //
+//   Py_INCREF(descr);
 //   npy_string_allocator *allocator = NpyString_acquire_allocator(descr);
 //   [code that uses allocator]
 //   NpyString_release_allocator(allocator);
+//   Py_DECREF(descr);
```
seberg, line 133: Hmmm, yeah, maybe this is better. OTOH it seems to me like if the descriptor gets cleaned up before the lock that seems very awkward and likely very wrong.

OTOH... I guess it is plausible to write a function that uses immediate C-cleanup on error of the array (and thus descriptor), before the scope is finished.

seberg (approved):
Thanks LGTM and a nice cleanup.  I am wondering if the lambda dance is really worth it.
Yeah, it is completely fine, but feels a bit like making the code harder to read for releasing the GIL a minimal amount earlier?

Mohataseem89:
discussion around the lambda used in building unordered_set: agree with the above comments, this lambda adds a level of indirection without much benefit; given nothing in the construction phase holds additional resources creating the set directly outside of the lambda both seems cleaner and consistent with the surrounding style. keeping the logic flat would probably improve readability without altering the error-safety guarantees introduced in this PR

and the resource management changes (including moving INCREF/DECREF into NpyStringAcquireAllocator) make the code more robust

numpy/_core/src/multiarray/unique.cpp
WarrenWeckesser: For large arrays when using multiple threads, there can be a benefit to releasing the GIL.  I wrote a little script to test this: https://github.com/WarrenWeckesser/experiments/blob/main/python/numpy/time-unique/time_unique.py.

I ran the script with the current pull request, and again with the lines `np::raii::SaveThreadState save_thread_state{};` commented out.  The following summarizes the results for some tests with `uint8`:

```
data type: uint8
array size: 25000000

              Execution Time
num_threads:    1       2       3      4
hold GIL:     0.4791  0.4628  0.4561 0.4553
release GIL:  0.4711  0.2436  0.1695 0.1289
```

and for variable strings:

```
data type: StringDType()
array size: 5000000

              Execution Time
num_threads:    1       2       3      4
hold GIL:     2.104   2.092   2.066  1.988
release GIL:  2.091   1.178   0.9379 0.7888
```

WarrenWeckesser: ... but I guess you're suggesting to move the declaration/initialization of the set out of the lambda, and only release the GIL for the loop that inserts values into the set.  That should not require a lambda.  I'll experiment with that.

WarrenWeckesser:
The `freebsd_test` is fixed in `main`.  Git workflow question:  What is the current recommended procedure when working on a PR to get `main` into the PR?  Merge `main`, or rebase on `main` and force-push?  In the past, Chuck has suggested to not merge main, because it can complicate a potential backport (but maybe that is not relevant if the PR will be squash-merged). Rebasing messes with ongoing reviews.

numpy/_core/src/multiarray/unique.cpp
WarrenWeckesser: OK, fixed.  The lambda's were not needed.

Mohataseem89 (commented):
this looks like an important fix—adding the missing Py_XDECREF(arr) ensures we don't leak the allocated array object on this error path, the logic stays the same, but the cleanup is now consistent with the rest of the function


(but this adds an extra Py_XDECREF call on an early-return path, so if arr is ever expected to be borrowed rather than owned at this point, it could introduce a double-free — though from the surrounding code, that looks unlikely )

seberg:
@WarrenWeckesser whatever you like, unless it's very clean, I would just squash merge.

numpy/_core/src/multiarray/unique.cpp
```diff
@@ -549,24 +499,27 @@ array__unique_hash(PyObject *NPY_UNUSED(module),
...
         PyErr_NoMemory();
-        return NULL;
+        result = NULL;
     }
     catch (const std::exception &e) {
         PyErr_SetString(PyExc_RuntimeError, e.what());
-        return NULL;
+        result = NULL;
     }
+    Py_DECREF(arr);
```
seberg, line 523: Wow, how did this sneak in!

seberg (approved):
Thanks for the follow-ups and misc cleanups! Looks good to me, let's get it in.
A bit large, but should backport this or at least those xdecref fixes, but probably can just do the whole thing.

seberg:
CircleCI here failed unrelated, probably needs a look at soon:
```
  File "/home/circleci/repo/venv/lib/python3.11/site-packages/sphinx/config.py", line 358, in eval_config_file
    exec(code, namespace)  # NoQA: S102
    ^^^^^^^^^^^^^^^^^^^^^
  File "/home/circleci/repo/doc/source/conf.py", line 65, in <module>
    replace_scalar_type_names()
  File "/home/circleci/repo/doc/source/conf.py", line 46, in replace_scalar_type_names
    import numpy
  File "/home/circleci/repo/build-install/usr/lib/python3.11/site-packages/numpy/__init__.py", line 454, in <module>
    from . import lib, matrixlib as _mat
  File "/home/circleci/repo/build-install/usr/lib/python3.11/site-packages/numpy/lib/__init__.py", line 18, in <module>
    from . import (
  File "/home/circleci/repo/build-install/usr/lib/python3.11/site-packages/numpy/lib/_arraypad_impl.py", line 10, in <module>
    from numpy.lib._index_tricks_impl import ndindex
  File "/home/circleci/repo/build-install/usr/lib/python3.11/site-packages/numpy/lib/_index_tricks_impl.py", line 8, in <module>
    import numpy.matrixlib as matrixlib
  File "/home/circleci/repo/build-install/usr/lib/python3.11/site-packages/numpy/matrixlib/__init__.py", line 4, in <module>
    from . import defmatrix
  File "/home/circleci/repo/build-install/usr/lib/python3.11/site-packages/numpy/matrixlib/defmatrix.py", line 13, in <module>
    from numpy.linalg import matrix_power
  File "/home/circleci/repo/build-install/usr/lib/python3.11/site-packages/numpy/linalg/__init__.py", line 87, in <module>
    from . import _linalg
  File "/home/circleci/repo/build-install/usr/lib/python3.11/site-packages/numpy/linalg/_linalg.py", line 77, in <module>
    from numpy._typing import NDArray
  File "/home/circleci/repo/build-install/usr/lib/python3.11/site-packages/numpy/_typing/__init__.py", line 3, in <module>
    from ._array_like import (
  File "/home/circleci/repo/build-install/usr/lib/python3.11/site-packages/numpy/_typing/_array_like.py", line 15
    type NDArray[ScalarT: np.generic] = np.ndarray[_AnyShape, np.dtype[ScalarT]]
         ^^^^^^^
SyntaxError: invalid syntax

  File "/home/circleci/repo/venv/lib/python3.11/site-packages/sphinx/cmd/build.py", line 293, in build_main
    app = Sphinx(args.sourcedir, args.confdir, args.outputdir,
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/circleci/repo/venv/lib/python3.11/site-packages/sphinx/application.py", line 211, in __init__
    self.config = Config.read(self.confdir, confoverrides or {}, self.tags)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/circleci/repo/venv/lib/python3.11/site-packages/sphinx/config.py", line 181, in read
    namespace = eval_config_file(filename, tags)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/circleci/repo/venv/lib/python3.11/site-packages/sphinx/config.py", line 361, in eval_config_file
    raise ConfigError(msg % err) from err
sphinx.errors.ConfigError: There is a syntax error in your configuration file: invalid syntax (_array_like.py, line 15)


Configuration error:
There is a syntax error in your configuration file: invalid syntax (_array_like.py, line 15)
```

mattip:
CircleCI does not properly merge the config from main, and so this PR is running the build with python3.11 (from the branch) not 3.12 (from main)

PR #30330, opened 2025-11-29
ENH: Show warning when polynomial.polynomial.fit can behave inconsistently

### LINKED ISSUE #28378: BUG: Polynomial.fit with a list of degrees behaves inconsistently
### Describe the issue:

Polynomial.fit of asymmetric x-range with a list of selected degrees using the default domain produces a wrong answer. It gives the correct answer when (1) the degree is given by an integer, (2) a list of the full degrees is given, or (3) domain is forced to be [-1,1] instead of automatically generated [xmin, xmax] .
If the x range is symmetric, then this issue doesn't occur.

```python
from numpy import linspace, arange
from numpy import poly1d
import numpy.polynomial.polynomial as nppp
import matplotlib.pyplot as plt

#x = linspace(-5,5,11)
x = linspace(-3,7,11)
y = -x + 2.1*x**2 - 0.34e-1*x**4

ppn = nppp.Polynomial.fit(x,y,5).convert().coef
ppf = nppp.Polynomial.fit(x,y,deg=[0,1,2,3,4,5]).convert().coef
ppl = nppp.Polynomial.fit(x,y,deg=[0,1,2,4]).convert().coef ### ppl gives wrong answers
ppc = nppp.Polynomial.fit(x,y,deg=[0,1,2,4],domain=[-1,1]).convert().coef

xx=linspace(x[0],x[-1],101)
yn=sum([p*xx**k for k,p in enumerate(ppn)])
yf=sum([p*xx**k for k,p in enumerate(ppf)])
yl=sum([p*xx**k for k,p in enumerate(ppl)])
yc=sum([p*xx**k for k,p in enumerate(ppc)])
plt.plot(x,y,'ko',xx,yn,'b-',xx,yf,'g-',xx,yl,'r-',xx,yc,'g.'); plt.pause(0.1)
```

### Error message:

```shell

```

2.2.3
3.10.9 (tags/v3.10.9:1dd9be6, Dec  6 2022, 20:01:21) [MSC v.1934 64 bit (AMD64)]

### Runtime Environment:

_No response_

### Context for the issue:

_No response_

antareepsarkar:
resolves #28378

`polynomial.polynomial.fit` does not always return a polynomial with only the degrees it is asked for. This PR is to show a warning in cases it behaves inconsistently.


antareepsarkar:
Please review this.

antareepsarkar:
Please review this.

mattip:
I am not really a polynomial expert, but this should have a test that emits the warning.

mattip:
You need to rebase off `main` since the circleCI job is still using 3.11 instead of 3.12

antareepsarkar:
> You need to rebase off `main` since the circleCI job is still using 3.11 instead of 3.12

Okay, Thanks.
I did it.

mattip:
Note this fix is different from the one suggested by @eendebakpt in the issue. Are they equivalent?

antareepsarkar:
> Note this fix is different from the one suggested by @eendebakpt in the issue. Are they equivalent?

They are similar. The only thing is, this one is more specific about when the inconsistency may occur. They considered only the `domain`, but, it may also occur due to the `window`.

antareepsarkar:
Should I make any other changes? Or, should I make them like as suggested in the issue?

numpy/polynomial/_polybase.py
```diff
@@ -1024,6 +1025,12 @@ class domain in NumPy 1.4 and ``None`` in later versions.
 
         xnew = pu.mapdomain(x, domain, window)
         res = cls._fit(xnew, y, deg, w=w, rcond=rcond, full=full)
+
+        if domain[0] + domain[1] != 0 or window[0] + window[1] != 0:
```
eendebakpt: The condition `domain[0] + domain[1] != 0` is hard to understand for me. What if a users explicitly passes `domain=[1, 2]`? We then get a warning (if the `deg` parameter is missing terms).

antareepsarkar: Okay, so the previous change had some problems.
Looks like that it only behaves inconsistently when the `off` variable of  the `mapparms` function of `numpy.polynomial.polyutils.py` is not equal to 0.

numpy/polynomial/_polybase.py
```diff
@@ -1024,6 +1025,14 @@ class domain in NumPy 1.4 and ``None`` in later versions.
 
         xnew = pu.mapdomain(x, domain, window)
         res = cls._fit(xnew, y, deg, w=w, rcond=rcond, full=full)
+
+        off = (domain[1] * window[0] - domain[0] * window[1])
```
eendebakpt, line 1031: `off` is the offset parameter (upto a scale factor) from 

https://github.com/numpy/numpy/blob/1883d84250008daf7aaf770fc62a2722301cb05c/numpy/polynomial/polyutils.py#L284

The reasoning is that off is zero, then scaling a polynomial `sum_i a_i x^i` by `x -> scale * x`  will not change which of the coefficients `a_i` are non-zero.

numpy/polynomial/_polybase.py
```diff
@@ -1024,6 +1025,14 @@ class domain in NumPy 1.4 and ``None`` in later versions.
...
+
+        if off != 0:
+            if hasattr(deg, "__iter__") and len(deg) - 1 != max(deg):
+                msg = "The polynomial may contain degrees other than those in deg"
```
eendebakpt: The warning message should be improved:

1. It is does not give a hint to the user how to resolve the issue (e.g. specify domain and window in such a way that the fitted polynomial will only have the degrees specified)
2. The message is not strong enough: the resulting fit will be incorrect
3. In the message `The polynomial may contain degrees other than those in deg`: which representation of the fitted polynomial are we talking about? (the coeffients which are non-zero will depend on the choice of domain/window)

numpy/polynomial/_polybase.py
eendebakpt: The `off` condition is correct, but maybe it is safer to just generate a warning when the input domain/window is not set explicitly. I think using the `deg` argument with omitted degrees is tricky, and users that really want this should be aware of the considerations and set the domain/window explicitly.

eendebakpt (commented):
Could you also update the documentation of the `deg` parameter of `ABCPolyBase.fit`?

numpy/polynomial/_polybase.py
antareepsarkar: Is the last commit okay?

antareepsarkar: Please review this.

numpy/polynomial/_polybase.py
```diff
@@ -1024,6 +1027,14 @@ class domain in NumPy 1.4 and ``None`` in later versions.
 
         xnew = pu.mapdomain(x, domain, window)
         res = cls._fit(xnew, y, deg, w=w, rcond=rcond, full=full)
+
+        off = (domain[1] * window[0] - domain[0] * window[1])
+
+        if off != 0:
+            if hasattr(deg, "__iter__") and len(deg) - 1 != max(deg):
```
eendebakpt, line 1034: Instead of checking `__iter__` I would use what `cls._fit` uses internally (e.g. `deg = np.asarray(deg)`)

numpy/polynomial/_polybase.py
```diff
@@ -1024,6 +1027,14 @@ class domain in NumPy 1.4 and ``None`` in later versions.
...
+                msg = "Change the domain/window to get only the specified degrees"
```
eendebakpt, line 1035: This will now issue a warning even when the user species the `domain`, `window` and `deg` (which could be a valid use case). My original suggestion was to only issue a warning when the `domain` and `window` have not been set explicitly.

eendebakpt (commented):
The docstring change is fine with me. But the changes the to code am not sure about yet. I will put the PR on the triage list to get a few more opinions.

Having the `deg` argument with `domain` / `window` is something tricky anyway. Maybe we should disallow it altogether?

numpy/polynomial/_polybase.py
antareepsarkar: So, should I make another commit based on the changes you are talking about?

eendebakpt: Lets wait a bit untill the issue has been discussed. I added it to the triage meeting https://hackmd.io/76o-IxCjQX2mOXO_wwkcpg?view#New-topics.

My preference would be to add the docs, and add a warning if the `domain` and `window` have not been set (e.g. None). If the users explicitly sets the `domain` and `window`, then it is is his/her responsibility to take the degrees into account.

antareepsarkar:
@eendebakpt 
Any updates on this?

eendebakpt:
> @eendebakpt Any updates on this?

Seems it was discussed (see the notes in https://hackmd.io/76o-IxCjQX2mOXO_wwkcpg?view#New-topics), so lets be patient.

PR #30332, opened 2025-11-29
MAINT: Add some pedantic Py_DECREFs

WarrenWeckesser:
A common pattern throughout the code is

    PyArray_Descr *descr = PyArray_DESCR(obj);
    Py_INCREF(descr);
    new_arr = PyArray_From[...](..., descr, ...)
    if (new_arr == NULL) {
        // Handle the failure.
    }

The array creation functions PyArray_From[...] steal a reference to descr, but if the function fails, the refcount of descr should be decreased. The Py_DECREF to do this was missing in many places.

This is not a high priority fix.  The array creation failure will end up as an exception in the calling code, and I doubt there are very many applications that will catch the exception *and keep going* long enough to notice a memory leak associated with the dtype.  So, as it says on the tin, *pedantic*...

### DIFF
--- a/numpy/_core/src/multiarray/_multiarray_tests.c.src
+++ b/numpy/_core/src/multiarray/_multiarray_tests.c.src
@@ -1077,6 +1077,7 @@ test_as_c_array(PyObject *NPY_UNUSED(self), PyObject *args)
                     dims,
                     1,
                     descr) < 0) {
+                Py_DECREF(descr);
                 PyErr_SetString(PyExc_RuntimeError, "error converting 1D array");
                 return NULL;
             }
@@ -1090,6 +1091,7 @@ test_as_c_array(PyObject *NPY_UNUSED(self), PyObject *args)
                     dims,
                     2,
                     descr) < 0) {
+                Py_DECREF(descr);
                 PyErr_SetString(PyExc_RuntimeError, "error converting 2D array");
                 return NULL;
             }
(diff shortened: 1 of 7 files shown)


mattip:
Drive by comment: maybe it is time to stop stealing refs in PyArray_From* functions and create PyArray_From2* that do the right thing

mhvk (commented):
Are you sure this is correct? I thought all `PyArray_From*` unconditionally steal the reference, as done in what I think is the main routine they all call: https://github.com/numpy/numpy/blob/250285f3cefd950e3a41e049a8733a3f6a9b920c/numpy/_core/src/multiarray/ctors.c#L1491-L1504

WarrenWeckesser:
> Are you sure this is correct? 

Well, not now, no ☹️.   I was working under the assumption that the creation functions would *not* change the state of the caller's input when an error occurred, but that's not the case.  [When you *assume*](https://xkcd.com/1339/)...

seberg:
Yeah, the caller also steals on errors, it's a bug in the caller if that isn't the case.

I would be fine with adding new non-stealing aliases for what it's worth (heck, they can be macros/static-inline functions, so that you can use them on old versions of NumPy also downstream).

mhvk:
OK, that at least solves it! Aside: agreed with @mattip that it is surprising and rather annoying that the reference gets stolen... But not quite sure it is worth new versions...

seberg:
I could imagine it now, stealing was never great, but Python pushed a lot to get rid of returning borrowed refs and that makes me think there is a move to never steal or return borrowed.
OTOH, need at least a nice name and two versions is arguably also confusing unfortunately...

PR #30334, opened 2025-11-29
BUG: Allow np.interp([], [], []) to return an empty array instead of error (GH-30316)

### LINKED ISSUE #30316: BUG: np.interp([], [], []) fails, when it could return an empty array
### Describe the issue:

```python
```

### Error message:

```shell
```

2.3.5

### Runtime Environment:

[{'numpy_version': '2.3.5',
            '[Clang 20.1.8 ]',
                      'found': ['ASIMDHP'],
                      'not_found': ['ASIMDFHM']}},

### Context for the issue:

Sujit-Hiwale:
BUG: Fix np.interp([], [], []) returning ValueError (GH-30316)

When both `x` and `xp` are empty, np.interp previously raised
ValueError("array of sample points is empty"). This behavior is overly
strict, as empty input should naturally result in an empty output array
with the same shape as `x`.

This commit implements special handling for the case xp == [] and
x == [], returning an empty output array, while preserving all existing
error conditions:

* xp == [] and x non-empty → still raises ValueError
* len(fp) != len(xp) → still raises ValueError
* xp non-empty → normal interpolation

Tests are added to verify the corrected behavior.
Fixes GH-30316.

### DIFF
--- a/numpy/_core/meson.build
+++ b/numpy/_core/meson.build
@@ -927,7 +927,7 @@ foreach gen_mtargets : [
     'loops_arithm_fp.dispatch.h',
     src_file.process('src/umath/loops_arithm_fp.dispatch.c.src'),
     [
-      X86_V3, X86_V2, 
+      X86_V3, X86_V2,
       ASIMD, NEON,
       VSX3, VSX2,
       VXE, VX,
@@ -960,7 +960,7 @@ foreach gen_mtargets : [
     'loops_exponent_log.dispatch.h',
     src_file.process('src/umath/loops_exponent_log.dispatch.c.src'),
     [
-      X86_V4, X86_V3, 
+      X86_V4, X86_V3,
     ]
   ],
   [
@@ -1403,6 +1403,7 @@ python_sources = [
   '_type_aliases.pyi',
   '_ufunc_config.py',
   '_ufunc_config.pyi',
+  '_umath_tests.pyi',
   'arrayprint.py',
   'arrayprint.pyi',
   'cversions.py',
--- a/numpy/lib/tests/test_function_base.py
+++ b/numpy/lib/tests/test_function_base.py
@@ -3087,6 +3087,12 @@ class TestInterp:
         assert_raises(ValueError, interp, 0, [0, 1], [1, 2], period=0)
         assert_raises(ValueError, interp, 0, [], [], period=360)
         assert_raises(ValueError, interp, 0, [0], [1, 2], period=360)
+        
+    def test_empty_triple_case(self):
+        # New expected behavior: np.interp([], [], []) returns []
+        res = np.interp([], [], [])
+        assert_equal(res, np.array([]))
+        assert res.shape == (0,)
 
     def test_basic(self):
         x = np.linspace(0, 1, 5)
--- a/numpy/_core/_umath_tests.pyi
+++ b/numpy/_core/_umath_tests.pyi
@@ -1,0 +1,47 @@
+# undocumented internal testing module for ufunc features, defined in
+# numpy/_core/src/umath/_umath_tests.c.src
+
+from typing import Final, Literal as L, TypedDict, type_check_only
+
+import numpy as np
+from numpy._typing import _GUFunc_Nin2_Nout1, _UFunc_Nin1_Nout1, _UFunc_Nin2_Nout1
+
+@type_check_only
+class _TestDispatchResult(TypedDict):
+    func: str  # e.g. 'func_AVX2'
+    var: str  # e.g. 'var_AVX2'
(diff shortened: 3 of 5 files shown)


numpy/lib/tests/test_function_base.py
```diff
@@ -3087,6 +3087,29 @@ def test_exceptions(self):
...
+
+    def test_empty_xp_nonempty_x_raises(self):
+        # xp empty but x non-empty → still an error
+        with pytest.raises(ValueError, match="sample points"):
+            np.interp([1, 2], [], [])
+
+    def test_fp_xp_length_mismatch_raises(self):
+        # mismatched fp/xp lengths must still error
+        with pytest.raises(ValueError, match="fp and xp"):
+            np.interp([1], [1, 2], [3])
```
mattip: These tests are covered in test_exceptions, no?

numpy/lib/tests/test_function_base.py
```diff
@@ -3087,6 +3087,29 @@ def test_exceptions(self):
...
         assert_raises(ValueError, interp, 0, [0], [1, 2], period=360)
+        
+    def test_empty_xp_empty_x(self):
+        # GH-30316: interp([], [], []) should return empty array
+        out = np.interp([], [], [])
+        assert_array_equal(out, [])
+
+        # fp non-empty should still return empty output
+        out = np.interp([], [], [1, 2])
+        assert_array_equal(out, [])
```
mattip: I think this is wrong. The [documentation](https://numpy.org/doc/stable/reference/generated/numpy.interp.html#numpy-interp) states "Raises ValueError if `xp` and `fp` have different length

mattip (commented):
I think only the case `np.interp([], [], [])` should return empty. All others should be errors.

numpy/lib/tests/test_function_base.py
Sujit-Hiwale: Yes they are

Sujit-Hiwale:
> I think only the case `np.interp([], [], [])` should return empty. All others should be errors.

Yes, agreed — only the specific case `np.interp([], [], [])` should
return an empty array. For all other cases, including any xp/fp length
mismatch or xp being empty while x is non-empty, the function should
continue to raise the appropriate ValueError.

Let me update the patch accordingly.

Sujit-Hiwale:
Thanks @mattip for the feedback.

Per your comments:

- Removed all additional tests that duplicate test_exceptions

- Removed the incorrect test allowing np.interp([], [], [1,2])
 
- Kept only the intended new behavior:
  np.interp([], [], []) returns an empty array
 
- Updated logic in compiled_base.c accordingly

Let me know if any further adjustments are needed!

mattip:
Let's try to minimize the feature creep. Is the testing module stubs critical for this PR?

Also: once we start reviewing, if you force push that means we have to review from the beginning. I prefer just adding more commits, as long as there are no merge conflicts.

numpy/_core/meson.build
```diff
@@ -927,7 +927,7 @@ foreach gen_mtargets : [
     'loops_arithm_fp.dispatch.h',
     src_file.process('src/umath/loops_arithm_fp.dispatch.c.src'),
     [
-      X86_V3, X86_V2, 
+      X86_V3, X86_V2,
```
mattip, line 930: Please avoid changes not connected to this PR.

numpy/lib/tests/test_function_base.py
```diff
@@ -3087,6 +3087,12 @@ def test_exceptions(self):
         assert_raises(ValueError, interp, 0, [0, 1], [1, 2], period=0)
         assert_raises(ValueError, interp, 0, [], [], period=360)
         assert_raises(ValueError, interp, 0, [0], [1, 2], period=360)
+        
+    def test_empty_triple_case(self):
+        # New expected behavior: np.interp([], [], []) returns []
+        res = np.interp([], [], [])
+        assert_equal(res, np.array([]))
+        assert res.shape == (0,)
```
mattip, line 3095: Why both checks?

numpy/lib/tests/test_function_base.py
```diff
@@ -3087,6 +3087,12 @@ def test_exceptions(self):
         assert_raises(ValueError, interp, 0, [0, 1], [1, 2], period=0)
         assert_raises(ValueError, interp, 0, [], [], period=360)
         assert_raises(ValueError, interp, 0, [0], [1, 2], period=360)
+        
+    def test_empty_triple_case(self):
+        # New expected behavior: np.interp([], [], []) returns []
```
mattip, line 3092:
```suggestion
```

PR #30335, opened 2025-11-30
TYP: add overloads for `generic.__getitem__`

nstarman:
"NumPy scalars also have many of the same methods arrays do." This includes `__getitem__`. I recently found that my type checker was complaining about `__getitem__` with key `None` on a `generic(...)`. This PR should fix this specific issue. There are a large number of remaining overloads required to describe the rest of the API surface for `__getitem__`, which interested parties can fill in or which I might contribute as need arises.

Edit: Let's just do it all.


numpy/__init__.pyi
```diff
@@ -3678,6 +3678,11 @@ class generic(_ArrayOrScalarCommon, Generic[_ItemT_co]):
     @overload
     def __array__(self, dtype: _DTypeT, /) -> ndarray[tuple[()], _DTypeT]: ...
 
+    #
+    @overload
```
jorenham, line 3682: It's the only one, so there's nothing to overload here

```suggestion
```

numpy/__init__.pyi
```diff
@@ -3678,6 +3678,11 @@ class generic(_ArrayOrScalarCommon, Generic[_ItemT_co]):
...
+    def __getitem__(self, key: None, /) -> ndarray[tuple[L[1]], dtype[Self]]: ...
```
jorenham: Literal shape axes can (unfortunately) be problematic in practice (there are still some places in the stubs where it's used, but I plan to remove those soon), so it's better to try and stick to boring `int`'s.

```suggestion
    def __getitem__(self, key: None, /) -> ndarray[tuple[int], dtype[Self]]: ...
```

jorenham (commented):
Thanks; I had no idea this was a thing tbh. 
I would have expected stubtest to detect these missing `__getitem__` methods though; odd...

jorenham:
> There are a large number of remaining overloads required to describe the rest of the API surface for `__getitem__`, which interested parties can fill in or which I might contribute as need arises.

After playing around with it a bit, I found that `_[()]` and `_[...]` return the scalars themselves, and that e.g. `_[None, None]` returns a `(1, 1)` array. Are there any others that you know of?

nstarman:
> Are there any others that you know of?

No, I just couldn't find a list of what was supported. And streaming numpy through an LLM likewise didn't get me a comprehensive list.

numpy/__init__.pyi
nstarman: True.

numpy/__init__.pyi
nstarman: That is unfortunate.

nstarman:
Interesting failures. I didn't get these on my GH codespace.

jorenham:
> > Are there any others that you know of?
> 
> No, I just couldn't find a list of what was supported. And streaming numpy through an LLM likewise didn't get me a comprehensive list.

Seems like its indeed limited to `...`, `None`, and tuples thereof:
https://github.com/numpy/numpy/blob/2de293aed2d8307db261faf4f5071f79376ccc85/numpy/_core/src/multiarray/scalartypes.c.src#L4155-L4176

jorenham:
> Interesting failures. I didn't get these on my GH codespace.

Hmm, did you run `spin stubtest` and `spin mypy` there (assuming that's even a thing in codespaces)?

nstarman:
> Seems like its indeed limited to ..., None, and tuples thereof:

That's a short enough list I can do it here.

jorenham:
mypy_primer reports two new errors for freqtrade:
https://github.com/freqtrade/freqtrade/blob/52c73eabeb36e0e44e7abb56e9a66626f07fff69/freqtrade/rpc/telegram.py#L408-L409

I'm not sure why they're only showing up *now*, but I don't see any way to avoid these, so it's ok to ignore them afaik.

And if you look at the line numbers in the diff, you'll see that the remaining green lines are nothing new.

nstarman:
All overloads are done. The tests are maybe too comprehensive. Also I squashed to 1 commit 🤷.

numpy/__init__.pyi
```diff
@@ -3678,6 +3678,31 @@ class generic(_ArrayOrScalarCommon, Generic[_ItemT_co]):
...
+        self, key: tuple[None, None], /
+    ) -> ndarray[tuple[int, int], dtype[Self]]: ...
+    @overload
+    def __getitem__(
+        self, key: tuple[None, None, None], /
+    ) -> ndarray[tuple[int, int, int], dtype[Self]]: ...
+    @overload  # Limited support for (None,) * N > 3
+    def __getitem__(
+        self, key: tuple[None, ...], /
+    ) -> ndarray[tuple[int, ...], dtype[Self]]: ...
```
jorenham: `NDArray` uses the "gradual" tuple type, `tuple[Any, ...]`. That way it'll also be assignable to e.g. arrays with 1d shape type:

```suggestion
    def __getitem__(self, key: tuple[None, ...], /) -> NDArray[Self]: ...
```

numpy/typing/tests/data/pass/scalars.py
```diff
@@ -247,3 +247,16 @@ def __float__(self) -> float:
...
+b[None]
+i8[None]
+u8[None]
+f8[None]
+c16[None]
+c16[...]
+c16[()]
+c16[(None,)]
+c16[(...,)]
+c16[(None, None)]
```
jorenham: nitpick

```suggestion
c16[None, None]
```

numpy/typing/tests/data/reveal/scalars.pyi
```diff
@@ -44,6 +44,86 @@ assert_type(c16.imag, np.float64)
...
+assert_type(c16[(...,)], np.ndarray[tuple[()], np.dtype[np.complex128]])
+assert_type(U[...], np.ndarray[tuple[()], np.dtype[np.str_]])
+assert_type(U[(...,)], np.ndarray[tuple[()], np.dtype[np.str_]])
+assert_type(S[...], np.ndarray[tuple[()], np.dtype[np.bytes_]])
+assert_type(S[(...,)], np.ndarray[tuple[()], np.dtype[np.bytes_]])
+
+None1 = (None,)
+None2 = (None, None)
+None3 = (None, None, None)
+None4 = (None, None, None, None)
```
jorenham, line 80: pretty slick :ok_hand:

jorenham:
> The tests are maybe too comprehensive.

I don't mind. AFAIK, the best kill is overkill (when testing)

numpy/__init__.pyi
nstarman: SGTM.

numpy/typing/tests/data/reveal/scalars.pyi
nstarman: 😎

numpy/typing/tests/data/pass/scalars.py
nstarman: ✅

numpy/__init__.pyi
nstarman: ✅

numpy/__init__.pyi
```diff
@@ -5708,7 +5731,8 @@ class void(flexible[bytes | tuple[Any, ...]]):  # type: ignore[misc]
     @overload
     def __new__(cls, length_or_data: object, /, dtype: _DTypeLikeVoid) -> Self: ...
 
-    @overload
+    #
+    @overload  # type: ignore[override]
```
jorenham: things like `void(b"")[None]` and such also work, no?

nstarman: Yeah. Let's do it.

nstarman:
Ok. Fully passed (locally) `spin build && spin mypy && spin stubtests`. Let's see how commensurate with CI that is.

numpy/__init__.pyi
nstarman: Ok. Required the full set of `__getitem__`.

nstarman:
@jorenham by ignore mypy_primer, do you mean here in the PR... I'm not seeing anything except the GH workflow.

jorenham:
> @jorenham by ignore mypy_primer, do you mean here in the PR... I'm not seeing anything except the GH workflow.

https://github.com/numpy/numpy/pull/30335#issuecomment-3593446850

nstarman:
> I'm not sure why they're only showing up now, but I don't see any way to avoid these, so it's ok to ignore them afaik.

> - colour/algebra/common.py:614: error: Value of type "Any | floating[_16Bit] | floating[_32Bit] | float64" is not indexable  [index]

Yeah, I was looking for a place to add a `type: ignore` or tell the CI that this is a known thing.

jorenham:
> > I'm not sure why they're only showing up now, but I don't see any way to avoid these, so it's ok to ignore them afaik.
> 
> > * colour/algebra/common.py:614: error: Value of type "Any | floating[_16Bit] | floating[_32Bit] | float64" is not indexable  [index]
> 
> Yeah, I was looking for a place to add a `type: ignore` or tell the CI that this is a known thing.

You'd have to do that in [colour](https://github.com/colour-science/colour) itself then.

nstarman:
Another day 😆

jorenham (approved):

jorenham:
Thanks @nstarman

PR #30338, opened 2025-12-01
BUG: Avoid hwcap for RVV version detection due to faulty v0.7/v1.0 re…

ixgbe:
Requires a distribution with a Linux 6.5 kernel or greater.
Refer to https://github.com/ggml-org/llama.cpp/pull/17461#discussion_r2568726154

Refer to https://riscv-optimization-guide.riseproject.dev/#_detecting_risc_v_extensions_on_linux as well. It will be helpful to extend to extract the vendor and micro-architecture ID if/when we want to have per-micro-architecture optimizations.


rgommers:
Thanks for this PR @ixgbe, and sorry it sat here unreviewed for so long.

This fix seems to overlap with gh-30988. I'm not sure if they're conflicting - needs a closer look.

We're adding RISC-V CI in gh-30995, which will help validate this PR.

luhenry (commented):
hwprobe is definitely the better way to go. It will also allow to detect for more extensions like Zvfh (vector half-float operations) and many more

gounthar:
Looks right to me. `riscv_hwprobe` with `RISCV_HWPROBE_KEY_IMA_EXT_0` + `RISCV_HWPROBE_IMA_V` targets RVV 1.0 specifically, not the pre-standard 0.7. `hwcap` can't make that distinction, so this is a real fix, not just a cleanup.

Fallback to `hwcap` when hwprobe isn't available is the right call. hwprobe landed in Linux 6.4, so older kernels are still covered.

On the overlap with #30988: that test still uses `hwcap` for RISC-V detection. On real RVV 1.0 hardware (SpaceMiT X60, VisionFive 2), both paths agree anyway, so no test breakage there. The mismatch would only show on pre-standard v0.7 boards, which is the exact case this PR guards against. I don't have 0.7 hardware to test that path directly, but the logic looks sound. A short comment in the test explaining why hwcap is still acceptable there would help future readers.

Once #30995 lands and CI runs natively on a SpaceMiT X60, we'll have end-to-end confirmation on real RVV 1.0 hardware.

ixgbe:
Thanks all. My understanding of the overlap with gh-30988 is:

- gh-30338 changes runtime feature detection in `npy_cpu_features.c`.
- gh-30988 changes the test-side helper for reading `AT_HWCAP`.
- On normal RVV 1.0 hardware, both `riscv_hwprobe(RISCV_HWPROBE_KEY_IMA_EXT_0)` and `HWCAP_RISCV_V` should report V, so the runtime and test paths agree.
- The only expected mismatch is pre-standard v0.7/0.7.1 hardware or older/backported kernels where HWCAP may advertise `V` without proving RVV 1.0. That is exactly why this PR prefers `riscv_hwprobe` for the runtime path.

I can add a short comment in `npy_cpu_features.c` clarifying that `RISCV_HWPROBE_IMA_V` is the RVV 1.0 signal, while `HWCAP` is kept only as a compatibility fallback when hwprobe is unavailable. If NumPy would prefer strict avoidance of HWCAP false positives on older/backported kernels, I can also change the fallback policy, but that would intentionally stop advertising RVV on kernels without hwprobe.

rgommers:
Thanks for the quick reviews @luhenry and @gounthar.  


> If NumPy would prefer strict avoidance of HWCAP false positives on older/backported kernels, I can also change the fallback policy, but that would intentionally stop advertising RVV on kernels without hwprobe.

Not necessary I think, the way you have it seems fine.

> I can add a short comment in `npy_cpu_features.c` clarifying that `RISCV_HWPROBE_IMA_V` is the RVV 1.0 signal, while `HWCAP` is kept only as a compatibility fallback when hwprobe is unavailable.

That seems useful to me.

ixgbe:
Hi @rgommers, thanks again for the review.

I pushed a follow-up commit adding the comment we discussed, clarifying that `RISCV_HWPROBE_IMA_V` is the RVV 1.0 signal and that `HWCAP` is only used as a compatibility fallback when `hwprobe` is unavailable. The PR checks are green now.

Could you let me know if there is anything else I should change to make this ready for merge? In particular, should this PR wait for the native RISC-V CI work in gh-30995, or is the current coverage enough for this runtime detection fix?

Thanks!

rgommers (approved):
LGTM and we have CI coverage now, so in it goes. Thanks @ixgbe and reviewers!

> In particular, should this PR wait for the native RISC-V CI work in https://github.com/numpy/numpy/pull/30995,

Sorry I forgot to answer this - yes, I thought CI coverage was close, so good to wait.

PR #30344, opened 2025-12-01
ENH: Remove unnecessary ValueError exception handler in trapezoid

sabasiddique1:
~Fix:~ Towards #30337
(edit: will not fix all the points in the issue, @mattip)
## What Changed

* Removed the unused `try/except ValueError` block in `numpy/lib/_function_base_impl.py`.

## Why Changed

The `ValueError` handler was never triggered by any valid input type, making it dead code.

## Comprehensive analysis shows:

1. **.sum(axis)** on NumPy arrays never raises ValueError 
2. **Broadcasting** is guaranteed by explicit reshape of d to match y's shape
(lines 5012-5014), so ValueError from broadcasting cannot occur
3. All **existing comprehensive tests** (including masked arrays, which are
the most likely to cause issues) pass without triggering the handler
4. The **fallback path does essentially the same computation as the normal
path** (converts to ndarray and uses add.reduce, which is what .sum()
uses internally anyway)

The handler appears to be defensive code added 'just in case' but handles
a case that doesn't occur with standard NumPy array types.

### DIFF
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -5019,13 +5019,7 @@ def trapezoid(y, x=None, dx=1.0, axis=-1):
     slice2 = [slice(None)] * nd
     slice1[axis] = slice(1, None)
     slice2[axis] = slice(None, -1)
-    try:
-        ret = (d * (y[tuple(slice1)] + y[tuple(slice2)]) / 2.0).sum(axis)
-    except ValueError:
-        # Operations didn't work, cast to ndarray
-        d = np.asarray(d)
-        y = np.asarray(y)
-        ret = add.reduce(d * (y[tuple(slice1)] + y[tuple(slice2)]) / 2.0, axis)
+    ret = (d * (y[tuple(slice1)] + y[tuple(slice2)]) / 2.0).sum(axis)
     return ret
 
 
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -5019,7 +5019,14 @@ def trapezoid(y, x=None, dx=1.0, axis=-1):
     slice2 = [slice(None)] * nd
     slice1[axis] = slice(1, None)
     slice2[axis] = slice(None, -1)
-    ret = (d * (y[tuple(slice1)] + y[tuple(slice2)]) / 2.0).sum(axis)
+    try:
+        ret = (d * (y[tuple(slice1)] + y[tuple(slice2)]) / 2.0).sum(axis)
+    except ValueError:
+        # Operations didn't work, cast to ndarray
+        # maybe a subclass like matrix?
+        d = np.asarray(d)
+        y = np.asarray(y)
+        ret = add.reduce(d * (y[tuple(slice1)] + y[tuple(slice2)]) / 2.0, axis)
     return ret
 
 


mattip:
Please check in the git history when this code was added, and look for hints there why it was done. This seems like it is AI generated, if so please clearly state it.

sabasiddique1:
> Please check in the git history when this code was added, and look for hints there why it was done. This seems like it is AI generated, if so please clearly state it.

Thank you for pointing this out and for the suggestion to check the git history.

**Commit:** `d1a661df48625ce5544d83dc022c96cf8c5d41c7`
I reviewed the history, and this `try/except ValueError` block was added in a 2010 commit by Ralf Gommers, along with switching `asarray` → `asanyarray` to support `ndarray` subclasses. In current NumPy, `array.sum(axis=…)` doesn’t raise `ValueError` for supported array types (invalid axes raise `AxisError`), and the tests from that commit pass without ever taking the `ValueError` path.

I did use an AI assistant to sketch the initial explanation, but I edited and refined everything myself. I also ran the full test suite locally (including the new tests I added), and none of them hit the `ValueError` branch, so removing it is completely safe.

Since this branch isn’t documented, tested, or triggered, I’d lean toward removing it, unless you’d prefer that I keep it and add targeted tests. Happy to adjust based on your feedback.

mattip:
Indeed, the test case of matrixes in d1a661df486 hits the path this PR wishes to remove. That test seems to have disappeared, and as @seberg says (edit: in the issue) the whole matrix class is deprecated, but some other subclass could hit the path. I don't think we should accept this PR to remove code that could be used by a subclass.
```
>>> import numpy as np
>>> x = np.linspace(0, 5)
>>> y = x * x
>>> mx = np.matrix(x)
>>> my = np.matrix(y)
>>> mr = np.trapezoid(my, mx)    # Hits the path in question
```

sabasiddique1:
> Indeed, the test case of matrixes in [d1a661d](https://github.com/numpy/numpy/commit/d1a661df48625ce5544d83dc022c96cf8c5d41c7) hits the path this PR wishes to remove. That test seems to have disappeared, and as @seberg says (edit: in the issue) the whole matrix class is deprecated, but some other subclass could hit the path. I don't think we should accept this PR to remove code that could be used by a subclass.
> 
> ```
> >>> import numpy as np
> >>> x = np.linspace(0, 5)
> >>> y = x * x
> >>> mx = np.matrix(x)
> >>> my = np.matrix(y)
> >>> mr = np.trapezoid(my, mx)    # Hits the path in question
> ```

Given that the fallback path is still reachable for subclasses (like np.matrix), should I go ahead and add tests and documentation for this behavior, or would it be better to close this PR instead?

mattip:
I think it is sufficient to make a 1-line change to extend the comment so others won't try to delete this until we get rid of matrix
```diff
        # Operations didn't work, cast to ndarray
+       # maybe a subclass like matrix?
```

MikhailRyazanov:
> I think it is sufficient to make a 1-line change to extend the comment so others won't try to delete this until we get rid of matrix

Let me jump in, since this is related to my #30337 and directly affects _what_ I would need to implement.
Are there any other subclasses that make sense as function arguments and support arithmetic operations but not, say ``numpy.sum``? If yes, please add a test.

seberg:
Maybe we can deprecate this? The `try/except` seems rather broader here than is nice for the sake of matrices that also seem like a rather niche case for being passed into `trapezoid` to begin with.

sabasiddique1:
I’ve added the comment about the fallback path.
Should any further changes (tests or deprecation) be done in this PR or in a separate one? @mattip

mattip:
> Are there any other subclasses

Who knows? I think the point is less `obj.sum` vs `np.add.reduce` and more the `np.asarray` before calling `*`. So the problem is subclasses that override `*` (like matrix)

seberg:
Sorry, but I think I am going to close this.  It would be nice to try and improve `trapezoid`, but the single line comment doesn't matter much, let's open a new PR for improvements!

PR #30348, opened 2025-12-01
MAINT: Update circleci python to 3.12.

charris:
We could go to 3.14, but I would prefer to let it settle a bit longer.

[skip actions] [skip cirrus]

NOTE: the build fails with Python 3.13, cannot parse `try_examples.json`. I suspect that maybe the environment needs an update.


charris:
Self merging so other Python 3.11 removals can proceed.

PR #30351, opened 2025-12-02
BUG: raise BufferError when creating dlpack with wrong device tuple

mattip:
Fixes #30341

Raise compliant BufferError exception


rgommers (approved):
Thanks Matti. That looks correct to me according to the spec.

I'll leave it open for a bit, since @seberg may want to have a look too.

seberg (approved):
Yeah, looks fine to me.  (On second thought, I guess we thought of of it as a bad user input here, but I think there is no real reaason to not use a `BufferError` anyway, and it seems like sklearn has a reason why they need it to be "can't export it via this protocol".)

seberg:
Thanks Matti, can just put this in I think.

PR #30359, opened 2025-12-02
MAINT: bump ``ruff`` to ``0.14.7``

jorenham:


mattip:
Did something in the changes between 0.14.0 and 0.14.7 catch your eye? I wonder why dependabot did not update this.

mattip (approved):

jorenham:
> Did something in the changes between 0.14.0 and 0.14.7 catch your eye? 

None that are relevant to numpy, but to be honest, I didn't look at the changelist that thoroughly.

> I wonder why dependabot did not update this.

I guess that these requirements.txt and environment.yml aren't included in the config then. But I wonder if that's even possible :thinking:

PR #30360, opened 2025-12-02
BUG: fix free-threaded races in RandomState

ngoldbaum:
Fixes one of the races in #30085.

I'd appreciate it if one of @lysnikolaou, @kumaraditya303, or @crusaderky could look this over, since you're all doing work related to this change.

I verified that the test I added generates TSan races without the cython change to add critical sections.

As far as I can tell, all other reads and writes to internal state are either atomic (e.g. in `get_state`) or are done with a `threading.lock` held and don't need any additional synchronization.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -20,7 +20,7 @@ Building NumPy requires the following installed software:
    e.g., on Debian/Ubuntu one needs to install both `python3` and
    `python3-dev`. On Windows and macOS this is normally not an issue.
 
-2) Cython >= 3.1.0
+2) Cython >= 3.0.6
 
 3) pytest__ (optional)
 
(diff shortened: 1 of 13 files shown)


ngoldbaum:
Looks like this also requires updating the minimum Cython version, so I'm going to remove the backport label...

ngoldbaum:
Well, looks like that didn't work on the TSan CI job. Likely will need a different approach. Converting to a draft PR for now.

ngoldbaum:
Nice! Just using the generator lock seems to work. Re-added the backport label since this doesn't require a Cython version bump using the lock.

Ping @rkern too.

rkern:
Seems sound to me.

mattip:
Thanks @ngoldbaum

PR #30361, opened 2025-12-03
MAINT: refactor `scalartypes.c.src` to pure C++

scratchmex:
remove begin repeat and use C macros. work towards https://github.com/numpy/numpy/issues/29528

for now I'm just refactoring to use macros and whenever the c++ templates are needed I will change to it. moving to C macros still has it's advantages cause I'm aiming for grepability and thus LSP advantages: `Py@NAME@ArrType_Type` vs `PyNumberArrType_Type`

the objective is to have a `scalartypes.c` or `scalartypes.cpp`

edit: the objective now is `scalartypes.cpp`


seberg:
I think the helpful first step may to be convert the file to C++ to continue in steps from there.  These custom macros seem a step in the wrong direction, since if we are to replace custom templating with C++ templating, we should use that.

scratchmex:
I'm going to do that in another PR because apparently is not as trivial as I thought. converting from `scalartypes.c.src` to `scalartypes.cpp.src`

edit: I pushed the changes to review, and because this will have a dependency on it. Just confirm me if you want this in separate PR for easy review and to avoid noise on this

edit 2: well apparently for the migration to C++ in windows it requires C++ 20 :/. do you have a suggestion? I think it is not suitable to do it without designated initializers `{ .tp_name = val }`

> error C7555: use of designated initializers requires at least '/std:c++20'

edit 3: learning stuff: struct factories at top level and compile time. introduce `make_pyarr_type`

edit 4: removed all designated initializers just trying to make it compile in MSVC so that I can mark as "completed" the migration to C++.

scratchmex:
nicee. finally C++ compiles jaja!

Just confirm me if you want the migration to C++ in a separate PR for easy review and to avoid noise on the removal of `.src` x2

edit 5: cannot make factories for array of methods because a) trying to use static variables inside a constexpr is not allowed and b) returning a stack address associated with a local variable is not something we want. 

a) `definition of a static variable in a constexpr function is a C++23 extension`
b) `address of stack memory associated with local variable 'methods' returned`

because of that, I will just repeat the definition of those

edit 6: cannot use `constexpr std::map` until C++26 to avoid using a chain of `if-else constexpr`

edit 7: cannot use `template <const char* mystr>` to create function factories. nor I can make a function return another function (lambda). will need to make definition of each function that wraps the generic one

seberg:
> I think it is not suitable to do it without designated initializers { .tp_name = val }

I think they are allowed to be written as long as they are in the exactly correct order? I am not sure how many holes they have, this is one annoyance with C++. If there are few we might also just fill them in explicitly, otherwise, I guess this tricky is fine.

Maybe it makes sense to keep going here for now and then see if we want to split out just the initial conversion for example.  (A bit conflicting, I would prefer to split it out, but if we don't follow through with the full thing there is no point.)

For what it's worth, I think the changes you did beyond the main change are not far reaching enough to be nice, we need to go all-in on templates for things to make sense. (Almost) anything repeated should be a template instead, the bufferprocs are a good example, that should just be something like `bufferprocs<typename T>`.
(The gentype version is then the overload that kicks in if a more specialized version is not defined.)

EDIT: CC @athurdekoos for awareness (and maybe if you want to even review).

scratchmex:
> the bufferprocs are a good example, that should just be something like `bufferprocs<typename T>`

you want
```
static auto unicode_arrtype_as_buffer = generic_pybufferprocs<unicode_getbuffer>;
static auto bool_arrtype_as_buffer = generic_pybufferprocs<dunder_buffer_impl1<npy_bool, PyBoolScalarObject>>;
```
instead of
```
static auto unicode_arrtype_as_buffer = make_pybufferprocs(unicode_getbuffer);
static auto bool_arrtype_as_buffer = make_pybufferprocs(dunder_buffer_impl1<npy_bool, PyBoolScalarObject>);
```
?

seberg:
No, `unicode_arrtype_as_buffer` need not exisst at all.

scratchmex:
`unicode_arrtype_as_buffer` needs to live in the stack at the top level scope because later its pointer is passed to their type definition like this
```cpp
// that's why I do
static auto unicode_arrtype_as_buffer = generic_pybufferprocs<unicode_getbuffer>;  // << top level
PyUnicodeArrType_Type.tp_as_buffer = &unicode_arrtype_as_buffer;
// instead of
PyUnicodeArrType_Type.tp_as_buffer = &make_pybufferprocs(unicode_getbuffer);
```

and doing something like 
```cpp
PyUnicodeArrType_Type.tp_as_buffer = &arrtype_as_buffer<unicode_getbuffer>
```

I wasn't able to do it yet. Can you give an example for that?

scratchmex:
I think the PR can be taken a first look. Remarking the following points

1. The most difficult part of the migration was the `__new__` impls block, I would like a review on that
2. I want to get rid of the macros but I don't find a better way to do it right now
3. seberg comment about assigning a pointer to a templated struct. I don't know how to do it

ccing the top 3 commiters of `scalartypes.c.src` @seberg @ngoldbaum @mtsokol

numpy/_core/src/multiarray/scalartypes.cpp
```diff
@@ -685,113 +715,152 @@ gentype_format(PyObject *self, PyObject *args)
...
-    for(dptr = ip + len - 1; len > 0 && *dptr == 0; len--, dptr--);
-    new = PyUnicode_FromKindAndData(PyUnicode_4BYTE_KIND, ip, len);
-    if (new == NULL) {
-        PyMem_Free(ip);
+    for (dptr = ip + len - 1; len > 0 && *dptr == 0; len--, dptr--);
+    PyObject *new_ = PyUnicode_FromKindAndData(PyUnicode_4BYTE_KIND, ip, len);
+    PyMem_Free(ip);
+    if (new_ == NULL) {
         return NULL;
     }
```
scratchmex, line 843:
```suggestion
    if (new_ == NULL) {
        PyMem_Free(ip);
        return NULL;
    }
```

ngoldbaum:
ping @athurdekoos as well

numpy/_core/src/multiarray/scalartypes.cpp
```diff
@@ -460,46 +475,56 @@ gentype_power(PyObject *m1, PyObject *m2, PyObject *modulo)
...
 }
 
+#define DEF_BIT_COUNT(fname, npy_type, ScalarName, npy_popcount_func, ConvertFrom) \
+    static PyObject *                                                              \
+    fname(PyObject *self, PyObject *NPY_UNUSED(args))                              \
+    {                                                                              \
+        npy_type scalar = PyArrayScalar_VAL(self, ScalarName);                     \
+        uint8_t count = npy_popcount_func(scalar);                                 \
+        return ConvertFrom(count);                                                 \
+    }
```
ngoldbaum, line 485: can't this be a template function?

numpy/_core/src/multiarray/scalartypes.cpp
```diff
@@ -2103,72 +2328,321 @@ static PyObject *
...
+gentype_ravel(PyObject *self, PyObject *args, PyObject *kwds)
+{
+    return gentype_generic_method(self, args, kwds, "ravel");
+}
+
+static PyObject *
+gentype_squeeze(PyObject *self, PyObject *args, PyObject *kwds)
+{
+    return gentype_generic_method(self, args, kwds, "squeeze");
+}
```
ngoldbaum, line 2638: given that it's trivial it's probably better to make this a macro though, it'll be a lot more concise.

numpy/_core/src/multiarray/scalartypes.cpp
```diff
@@ -2841,132 +3345,281 @@ static PyMethodDef numbertype_methods[] = {
...
+static PyMethodDef inttype_methods[] = {
+    /* for typing */
+    {"__class_getitem__",
+        (PyCFunction)numbertype_class_getitem,
+        METH_CLASS | METH_O, NULL},
+    {"bit_count",
+        (PyCFunction)npy_int_bit_count,
+        METH_NOARGS, NULL},
+    {NULL, NULL, 0, NULL} /* sentinel */
+};
```
ngoldbaum, line 3526: I like that these are all explicitly defined in the source now, it's OK that it's more verbose.

numpy/_core/src/multiarray/scalartypes.cpp
```diff
@@ -3170,35 +3835,84 @@ gentype_arrtype_getbuffer(PyObject *self, Py_buffer *view, int flags)
...
+    else if constexpr (std::is_same_v<PT, PyLongDoubleScalarObject>) {
+        return "g";
+    }
+    else if constexpr (std::is_same_v<PT, PyCFloatScalarObject>) {
+        return "Zf";
+    }
+    else if constexpr (std::is_same_v<PT, PyCDoubleScalarObject>) {
+        return "Zd";
+    }
+    else if constexpr (std::is_same_v<PT, PyCLongDoubleScalarObject>) {
```
ngoldbaum, line 3901: general question about code like this from a non-expert in C++, code like this using `constexpr` is evaluated at compile-time, right?

ngoldbaum (commented):
I skimmed over most of the changes in `scalartypes.cpp` and left a few comments. No careful review here.

seberg:
> seberg comment about assigning a pointer to a templated struct. I don't know how to do it

Maybe check out what `jax/ml_dtypes` is doing and see if you get inspired from a pattern. I don't really want to try around right now, but I think this has to happen in some form.
I would hope all explicit instantiation with a name is unnecessary with the exception a single global type probably.
(In part, focusing on this type of discussion is why it might be helpful to do just the conversion first.)

PR #30364, opened 2025-12-03
TYP: default to ``datetime64[Any]`` and ``timedelta64[Any]``

jorenham:
This changes the default of the `datetime64` type parameter from `date | datetime | int | None` to `Any`, and that of `timedelta64` from `timedelta | int | None` to `Any`. 

Technically it's more type-unsafe this way. But in practice, these broad defaults were causing quite a lot of annoying issues. See https://github.com/microsoft/pyright/issues/11135 for an example. 

As you can see in the type-test changes, using `Any` as generic default seems to trigger some strange behavior in mypy, causing it to arbitrarily change return types to e.g. `Any` and `tuple[Any, ...]`. However, according to the [gradual guarantee](https://typing.python.org/en/latest/spec/concepts.html#the-gradual-guarantee), this shouldn't be a breaking change.

However, both mypy and pyright currently deviate from the typing spec when it comes to overload behavior when `Any` is involved. This causes them to violate the gradual guarantee. The workaround is to (ab)use the `Never` *bottom type* to "capture" `Any`, and only `Any`. This `Never`-trick is already used in several places in the numpy stubs (before I came into the picture even), so it's nothing new.


jorenham:
The net mypy_primer diff is +2 "unused-ignore" errors, i.e. 2 errors resolved. All other diffs seem to be changed error messages.

jorenham:
The new primer errors for pandas-stubs in `tests/scalars/test_scalars.py` are likely caused by the same mypy issue that I mentioned in the PR description. Because even though I did not touch any of the `__eq__` or `__ne__` methods, mypy is now inferring those to return `Any` for some reason: https://github.com/pandas-dev/pandas-stubs/blob/7bd80651ef1d18831cd7765fb9c5f6bae4cbf315/tests/scalars/test_scalars.py#L893-L913. So the only conclusion I can draw here is that it's a bug in mypy.

The two new pandas-stubs `test_sub.py` errors are probably my fault; I'll see if I can find the issue.

EDIT: Nope; those `test_sub.py` errors also appear to be caused by a (different) mypy bug ([repro](https://mypy-play.net/?mypy=latest&python=3.14&gist=3219febc559a863ae1a799b9307c95fe)). I should be able to work around it though.

jorenham:
Ah nice; all pandas-stubs errors are gone now. That makes this a net gain of 2 resolved errors (in static-frame).

charris:
Needs rebase.

charris:
Thanks Joren.

PR #30366, opened 2025-12-03
DOC: Add more examples for numpy.sum

karansingh012:
This PR adds two additional examples to the numpy.sum documentation:
	•	An example demonstrating the use of the where parameter
	•	An example demonstrating the use of the initial parameter

These examples make the optional arguments easier to understand for beginners and improve the clarity of the documentation.

### DIFF
--- a/numpy/_core/fromnumeric.py
+++ b/numpy/_core/fromnumeric.py
@@ -2413,6 +2413,10 @@ def sum(a, axis=None, dtype=None, out=None, keepdims=np._NoValue,
 
     >>> np.sum([10], initial=5)
     15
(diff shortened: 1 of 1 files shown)


karansingh012:
This change only affects documentation.  
The failing CI checks seem unrelated to the modification.  
Please let me know if anything needs to be updated. Thanks!

jorenham:
There are already examples of both `where` and `initial`?

karansingh012:
Thanks for pointing that out!

I checked again and you're right — the documentation already includes examples for both `where` and `initial`. I’ll close this PR to avoid duplication.

Appreciate the feedback!

PR #30373, opened 2025-12-04
DOC: Improve cross-links in thread safety documentation

Carreau:
Those are mostly cosmetic changes and extra cross-links to improve the documentation. In particular I'd like to use this page as an example of good top-level documentation for threading/freethreading in the py-threading-guide https://py-free-threading.github.io/

This also does a number of updates to the misc.rst file which was basically too indented, and missing a bunch of links. I think many of the info misc.rst on how to interop with C/C++ could be removed, and likely all the information on seterr/geterr moved to routines.err.rst but I don't want to include this moves as part of this PR as it's likely more subject to discussion.

There are minor updates in conf.py to avoid 1 sphinx warnings, and I know the thread safety backlink in geterr/seterr/...etc, are a bit overkill, but as numpy will be used as an example and in general is referred to as one of the core scientific Python project I thinks it's OK go go a bit further than usual.

[skip actions]
[skip cirrus]

### DIFF
--- a/doc/source/reference/routines.err.rst
+++ b/doc/source/reference/routines.err.rst
@@ -75,8 +75,6 @@ Examples
     saw stupid error!
     array([nan, nan, nan, nan, nan])
 
-::
-
 Setting and getting error handling
 ----------------------------------
 
--- a/doc/source/reference/routines.err.rst
+++ b/doc/source/reference/routines.err.rst
@@ -1,7 +1,81 @@
+.. _fp_error_handling:
+
 Floating point error handling
 =============================
 
 .. currentmodule:: numpy
+
+Error handling settings are stored in :py:mod:`python:contextvars`
+allowing different threads or async tasks to have independent configurations.
+For more information, see :ref:`thread_safety`.
+
+.. _misc-error-handling:
+
+How numpy handles numerical exceptions
+--------------------------------------
+
+The default is to ``'warn'`` for ``invalid``, ``divide``, and ``overflow``
+and ``'ignore'`` for ``underflow``.  But this can be changed, and it can be
+set individually for different kinds of exceptions. The different behaviors
+are:
+
+- ``'ignore'`` : Take no action when the exception occurs.
+- ``'warn'``   : Print a :py:exc:`RuntimeWarning` (via the Python :py:mod:`warnings` module).
+- ``'raise'``  : Raise a :py:exc:`FloatingPointError`.
+- ``'call'``   : Call a specified function.
+- ``'print'``  : Print a warning directly to ``stdout``.
+- ``'log'``    : Record error in a Log object.
+
+These behaviors can be set for all kinds of errors or specific ones:
+
+- ``all``       : apply to all numeric exceptions
+- ``invalid``   : when NaNs are generated
+- ``divide``    : divide by zero (for integers as well!)
+- ``overflow``  : floating point overflows
+- ``underflow`` : floating point underflows
+
+Note that integer divide-by-zero is handled by the same machinery.
+
+The error handling mode can be configured :func:`numpy.errstate`
+context manager.
+
+Examples
+--------
+
+::
+
+    >>> with np.errstate(all='warn'):
+    ...     np.zeros(5, dtype=np.float32) / 0.0
+    <python-input-1>:2: RuntimeWarning: invalid value encountered in divide
+    array([nan, nan, nan, nan, nan], dtype=float32)
+
+::
+
+    >>> with np.errstate(under='ignore'):
+    ...     np.array([1.e-100])**10
+    array([0.])
+
+::
+
+    >>> with np.errstate(invalid='raise'):
+    ...     np.sqrt(np.array([-1.]))
+    ...
+    Traceback (most recent call last):
+      File "<python-input-1>", line 2, in <module>
+        np.sqrt(np.array([-1.]))
+        ~~~~~~~^^^^^^^^^^^^^^^^^
+    FloatingPointError: invalid value encountered in sqrt
+
+::
+
+    >>> def errorhandler(errstr, errflag):
+    ...     print("saw stupid error!")
+    >>> with np.errstate(call=errorhandler, all='call'):
+    ...     np.zeros(5, dtype=np.int32) / 0
+    saw stupid error!
+    array([nan, nan, nan, nan, nan])
+
+::
 
 Setting and getting error handling
 ----------------------------------
--- a/doc/source/reference/routines.io.rst
+++ b/doc/source/reference/routines.io.rst
@@ -59,8 +59,15 @@ Memory mapping files
    memmap
    lib.format.open_memmap
 
+.. _text_formatting_options:
+
 Text formatting options
 -----------------------
+
+Text formatting settings are maintained in a :py:mod:`context variable <python:contextvars>`,
+allowing different threads or async tasks to have independent configurations.
+For more information, see :ref:`thread_safety`.
+
 .. autosummary::
    :toctree: generated/
 
--- a/doc/source/reference/thread_safety.rst
+++ b/doc/source/reference/thread_safety.rst
@@ -5,7 +5,7 @@ Thread Safety
 *************
 
 NumPy supports use in a multithreaded context via the `threading` module in the
-standard library. Many NumPy operations release the GIL, so unlike many
+standard library. Many NumPy operations release the :term:`python:GIL`, so unlike many
 situations in Python, it is possible to improve parallel performance by
 exploiting multithreaded parallelism in Python.
 
@@ -22,7 +22,7 @@ are not reproducible, let alone correct. It is also possible to crash the Python
 interpreter by, for example, resizing an array while another thread is reading
 from it to compute a ufunc operation.
 
-In the future, we may add locking to ndarray to make writing multithreaded
+In the future, we may add locking to :class:`~numpy.ndarray` to make writing multithreaded
 algorithms using NumPy arrays safer, but for now we suggest focusing on
 read-only access of arrays that are shared between threads, or adding your own
 locking if you need to mutation and multithreading.
@@ -32,6 +32,18 @@ from use of the `threading` module, and instead might be better served with
 `multiprocessing`. In particular, operations on arrays with ``dtype=np.object_``
 do not release the GIL.
 
+Context-local state
+-------------------
+
+NumPy maintains some state for ufuncs context-local basis, which means each
+thread in a multithreaded program or task in an asyncio program has its own
+independent configuration of the `numpy.errstate` (see
+:doc:`/reference/routines.err`), and of :ref:`text_formatting_options`.
+
+You can update state stored in a context variable by entering a context manager.
+As soon as the context manager exits, the state will be reset to its value
+before entering the context manager.
+
 Free-threaded Python
 --------------------
 
@@ -40,12 +52,27 @@ Free-threaded Python
 Starting with NumPy 2.1 and CPython 3.13, NumPy also has experimental support
 for python runtimes with the GIL disabled. See
 https://py-free-threading.github.io for more information about installing and
-using free-threaded Python, as well as information about supporting it in
-libraries that depend on NumPy.
+using :py:term:`free-threaded <python:free threading>` Python, as well as
+information about supporting it in libraries that depend on NumPy.
 
-Because free-threaded Python does not have a global interpreter lock to
-serialize access to Python objects, there are more opportunities for threads to
-mutate shared state and create thread safety issues. In addition to the
-limitations about locking of the ndarray object noted above, this also means
-that arrays with ``dtype=np.object_`` are not protected by the GIL, creating data
-races for python objects that are not possible outside free-threaded python.
+Because free-threaded Python does not have a
+global interpreter lock to serialize access to Python objects, there are more
+opportunities for threads to mutate shared state and create thread safety
+issues. In addition to the limitations about locking of the
+:class:`~numpy.ndarray` object noted above, this also means that arrays with
+``dtype=np.object_`` are not protected by the GIL, creating data races for python
+objects that are not possible outside free-threaded python.
+
+C-API Threading Support
+-----------------------
+
+For developers writing C extensions that interact with NumPy, several parts of
+the :doc:`C-API array documentation </reference/c-api/array>` provide detailed
+information about multithreading considerations.
+
+See Also
+--------
+
+* :doc:`/reference/random/multithreading` - Practical example of using NumPy's
+  random number generators in a multithreaded context with
+  :mod:`concurrent.futures`.
--- a/doc/source/user/misc.rst
+++ b/doc/source/user/misc.rst
@@ -7,7 +7,7 @@ Miscellaneous
 IEEE 754 floating point special values
 --------------------------------------
 
-Special values defined in numpy: nan, inf,
+Special values defined in numpy: :data:`~numpy.nan`, :data:`~numpy.inf`
 
 NaNs can be used as a poor-man's mask (if you don't care what the
 original value was)
@@ -17,29 +17,39 @@ Note: cannot use equality to test NaNs. E.g.: ::
  >>> myarr = np.array([1., 0., np.nan, 3.])
  >>> np.nonzero(myarr == np.nan)
  (array([], dtype=int64),)
+
+::
+
  >>> np.nan == np.nan  # is always False! Use special numpy functions instead.
  False
+
+::
+
  >>> myarr[myarr == np.nan] = 0. # doesn't work
  >>> myarr
  array([  1.,   0.,  nan,   3.])
+
+::
+
  >>> myarr[np.isnan(myarr)] = 0. # use this instead find
  >>> myarr
  array([1.,  0.,  0.,  3.])
 
-Other related special value functions: ::
+Other related special value functions:
 
- isinf():    True if value is inf
- isfinite(): True if not nan or inf
- nan_to_num(): Map nan to 0, inf to max float, -inf to min float
+- :func:`~numpy.isnan` - True if value is nan
+- :func:`~numpy.isinf` - True if value is inf
+- :func:`~numpy.isfinite` - True if not nan or inf
+- :func:`~numpy.nan_to_num` - Map nan to 0, inf to max float, -inf to min float
 
 The following corresponds to the usual functions except that nans are excluded
-from the results: ::
+from the results:
 
- nansum()
- nanmax()
- nanmin()
- nanargmax()
- nanargmin()
+- :func:`~numpy.nansum`
+- :func:`~numpy.nanmax`
+- :func:`~numpy.nanmin`
+- :func:`~numpy.nanargmax`
+- :func:`~numpy.nanargmin`
 
  >>> x = np.arange(10.)
  >>> x[3] = np.nan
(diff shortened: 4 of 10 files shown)


doc/source/user/misc.rst
```diff
@@ -56,159 +68,178 @@ and ``'ignore'`` for ``underflow``.  But this can be changed, and it can be
...
+   array([nan, nan, nan, nan, nan])
+
+::
+
+   >>> j = np.seterr(**oldsettings) # restore previous
+   ...                              # error-handling settings
 
 Interfacing to C
 ----------------
 Only a survey of the choices. Little detail on how each works.
```
Carreau: I think most of what is below here could/should use some pruning. But that might be another PR.

Carreau:
Maybe tag as free-threading ?

rgommers:
sure, done

doc/source/reference/thread_safety.rst
```diff
@@ -22,15 +22,30 @@ are not reproducible, let alone correct. It is also possible to crash the Python
...
+its own independent configuration:
+
+* **Error handling** - The floating-point error handling mode (configured via
+  :func:`numpy.seterr`, :func:`numpy.geterr`, and :func:`numpy.errstate`) is
+  thread-local. Each thread can have different settings for how to handle
+  invalid operations, division by zero, overflow, and underflow. See
+  :ref:`misc-error-handling` for more details.
+
+* **Buffer size** - The internal buffer size used by ufuncs for processing
+  misaligned or type-converted data is configurable per-thread.
```
mattip: Since both these options are handled by the errstate, I think it would be better to describe the functionality there.

```suggestion
NumPy maintains some state for ufuncs on a per-thread basis, which means each thread has its 
own independent configuration of the `numpy.errstate`.
```

doc/source/reference/thread_safety.rst
```diff
@@ -5,14 +5,14 @@ Thread Safety
 *************
 
 NumPy supports use in a multithreaded context via the `threading` module in the
-standard library. Many NumPy operations release the GIL, so unlike many
+standard library. Many NumPy operations release the :term:`python:GIL`, so unlike many
```
mattip, line 8: Typically only the first reference to GIL should be a link, especially when the occurrences are so close together.

doc/source/reference/thread_safety.rst
```diff
@@ -40,12 +55,49 @@ Free-threaded Python
...
-Because free-threaded Python does not have a global interpreter lock to
-serialize access to Python objects, there are more opportunities for threads to
-mutate shared state and create thread safety issues. In addition to the
-limitations about locking of the ndarray object noted above, this also means
-that arrays with ``dtype=object`` are not protected by the GIL, creating data
-races for python objects that are not possible outside free-threaded python.
+using :py:term:`free-threaded <python:free threading>` Python, as well as
+information about supporting it in libraries that depend on NumPy.
+
+Because :py:term:`free-threaded <python:free threading>` Python does not have a
```
mattip: Here too only the first reference should be a link, especially when they are so close together.

doc/source/reference/thread_safety.rst
```diff
@@ -40,12 +55,49 @@ Free-threaded Python
...
+``dtype=object`` are not protected by the GIL, creating data races for python
+objects that are not possible outside :py:term:`free-threaded <python:free
+threading>` python.
+
+C-API Threading Support
+-----------------------
+
+For developers writing C extensions that interact with NumPy, several parts of
+the :doc:`C-API array documentation </reference/c-api/array>` provide detailed
+information about multithreading considerations:
```
mattip: Hmm. Much of this repeats the information already in the referenced page. Perhaps there is a different way to structure this so as to not repeat that information here?

doc/source/user/misc.rst
mattip, line 1: This page is strange. It has the :orphan: tag because it is not included in a TOC, so it is not really part of the user documentation. It is not clear how users would stumble across it. The information here would probably be best distributed into the different areas of reference, rather than extending it.

The IEE 754 floation point special values section could be folded into the [constants](https://numpy.org/devdocs/reference/constants.html#constants) page.

The `How numpy handles numerical exceptions` section duplicates information in the `errstate` documentation, and would be best reduced to a short "see `errstate` instead. Maybe some of this content could be used to extend the `errstate` documentation.

The "Interfacing to C" section is covered in the [Using NumPy C-API](https://numpy.org/devdocs/user/c-info.html) and [interoperability](https://numpy.org/devdocs/user/basics.interoperability.html) sections.

numpy/_core/_ufunc_config.py
```diff
@@ -68,6 +68,11 @@ def seterr(all=None, divide=None, over=None, under=None, invalid=None):
     - Invalid operation: result is not an expressible number, typically
       indicates that a NaN was produced.
 
+    **Thread safety:** Error handling settings are maintained on a per-thread
+    basis. Each thread can have its own independent error handling configuration.
+    See :ref:`thread_safety` for more information about using NumPy in
```
mattip: Probably better to directly reference `errstate`. and find a way to not repeat this information in slightly different forms each time

mattip (commented):
Thanks for thinking about the documentation. In general cross-linking to the same term should be limited to a single reference per page, since it is visually disturbing.

> In particular I'd like to use this page as an example of good top-level documentation

Cool!

> This also does a number of updates to the misc.rst file

Hah, surprised you found it at all. That page is really an orphan that probably should have been removed at some stage. I would prefer to not add to it, rather to remove from it and move the content elsewhere.

doc/source/reference/thread_safety.rst
```diff
@@ -22,15 +22,30 @@ are not reproducible, let alone correct. It is also possible to crash the Python
...
 locking if you need to mutation and multithreading.
 
-Note that operations that *do not* release the GIL will see no performance gains
+Note that operations that *do not* release the :term:`python:GIL` will see no performance gains
 from use of the `threading` module, and instead might be better served with
 `multiprocessing`. In particular, operations on arrays with ``dtype=object`` do
-not release the GIL.
+not release the :term:`python:GIL`.
+
+Thread-local state
```
ngoldbaum: errstate is stored in a contextvar, so it's actually context-local state. It's async-safe and thread-safe.

Maybe you could call it `Context-local state`? I'm not sure if there's a term for it that asyncio or CPython more generally uses.

Also the printoptions are a contextvar as well.

ngoldbaum: The custom allocator object is also context-local: https://numpy.org/devdocs/reference/c-api/data_memory.html#configurable-memory-routines-in-numpy-nep-49

doc/source/user/misc.rst
ngoldbaum: I'd just skip editing this file in this PR. Move the new content you added where it makes sense elsewhere. Open an issue for a followup task to clean this file and other orphan files in the docs up.

numpy/_core/_ufunc_config.py
ngoldbaum: I agree, it might be nice to have a custom adominition that inserts this text, for example.

numpy/_core/_ufunc_config.py
```diff
@@ -68,6 +68,11 @@ def seterr(all=None, divide=None, over=None, under=None, invalid=None):
     - Invalid operation: result is not an expressible number, typically
       indicates that a NaN was produced.
 
+    **Thread safety:** Error handling settings are maintained on a per-thread
```
ngoldbaum: It's not really a safety note, it's more of a usage note. Maybe "Concurrency notes"?

Also concurrency is better because this all applies to asyncio as well.

doc/source/reference/routines.err.rst
```diff
@@ -1,8 +1,86 @@
+.. _fp_error_handling:
```
Carreau, line 1: adding target as `See Also` links parsing is more limited than sphinx/pure rst, improvment to add to docscrape would be support multi-colon in roles (like `:std:doc:`, and relax allowed character in between backticks: 
``` 
:ref:`/reference/routine.err`
:ref:`things-with-dashed`
```

Appear to also not please docscrape parsing.

doc/source/reference/routines.err.rst
```diff
@@ -1,8 +1,86 @@
...
+   >>> np.zeros(5, dtype=np.int32)/0
+   saw stupid error!
+   array([nan, nan, nan, nan, nan])
+
+::
+
+   >>> j = np.seterr(**oldsettings) # restore previous
+   ...                              # error-handling settings
+
+
```
Carreau: All this is moved from the misc page; I could also as pointed, just not touch the misc page, and do all this later.

doc/source/reference/routines.io.rst
```diff
@@ -59,8 +59,15 @@ Memory mapping files
    memmap
    lib.format.open_memmap
 
+.. _text_formatting_options:
```
Carreau, line 62: Same thing here, adding target to be able to crosslink

doc/source/reference/thread_safety.rst
```diff
@@ -5,14 +5,14 @@ Thread Safety
...
+standard library. Many NumPy operations release the :term:`python:GIL`, so unlike many
 situations in Python, it is possible to improve parallel performance by
 exploiting multithreaded parallelism in Python.
 
 The easiest performance gains happen when each worker thread owns its own array
 or set of array objects, with no data directly shared between threads. Because
-NumPy releases the GIL for many low-level operations, threads that spend most of
-the time in low-level code will run in parallel.
+NumPy releases the GIL for many low-level operations, threads
+that spend most of the time in low-level code will run in parallel.
```
Carreau: Todo: revert to original.

numpy/_core/_ufunc_config.py
```diff
@@ -55,7 +55,8 @@ def seterr(all=None, divide=None, over=None, under=None, invalid=None):
     See also
     --------
     seterrcall : Set a callback function for the 'call' mode.
-    geterr, geterrcall, errstate
+    geterr, geterrcall, errstate, :ref:`fp_error_handling`
```
Carreau: Option 1: 

I can do everything like this in See Also:
 - advantage, it's juste a single link.
 - disadvantage, I can't say it's about concurrency, and can't choose the text (unless we fix docscrape to accept the `text <ref>` syntax.

numpy/_core/_ufunc_config.py
```diff
@@ -127,6 +128,8 @@ def geterr():
     For complete documentation of the types of floating-point exceptions and
     treatment options, see `seterr`.
 
+    **Concurrency note:** see :doc:`/reference/routines.err`
```
Carreau, line 133: Or Options 2:

Add in Notes with a text. We are free to use arbitrary RST.
 - advantage, it can say "concurrency" or threading.
 - disadvantage; it's verbose.

Let me know which option you like.

numpy/_core/_ufunc_config.py
```diff
@@ -392,13 +412,19 @@ class errstate:
...
     -----
     For complete documentation of the types of floating-point exceptions and
     treatment options, see `seterr`.
 
+    **Multithreading:** As of version 2.0, `errstate` works correctly with
+    threading and asyncio. Error handling settings are maintained on a per-thread
+    basis, so using `errstate` in one thread does not affect other threads.
+    See :ref:`thread_safety` for more information about using NumPy in
+    multithreaded contexts.
+
```
Carreau: Todo, remove.

numpy/_core/arrayprint.py
```diff
@@ -255,6 +256,7 @@ def set_printoptions(precision=None, threshold=None, edgeitems=None,
     * Use `printoptions` as a context manager to set the values temporarily.
     * These print options apply only to NumPy ndarrays, not to scalars.
 
+
```
Carreau, line 260:
```suggestion
```

Carreau (commented):
Some suggestions, in particular I'd like  your input as to wether put in `See Also`, or in `Note` the ref from the functions docstrings, or just not put it at all. 

I'm not a huge fan of See Also, as there is no way reading it to know it's about concurency; and same with saying "Concurency Note", if you search for "Thread" it does not match, so one may miss the text.

doc/source/reference/routines.err.rst
```diff
@@ -1,8 +1,86 @@
+.. _fp_error_handling:
+
 Floating point error handling
 =============================
 
 .. currentmodule:: numpy
 
+Error handling settings are maintained on in a :py:mod:`python:contextvars`
```
ngoldbaum:
```suggestion
Error handling settings are stored in :py:mod:`python:contextvars`
```

doc/source/reference/routines.err.rst
```diff
@@ -1,8 +1,86 @@
...
+
+The default is to ``'warn'`` for ``invalid``, ``divide``, and ``overflow``
+and ``'ignore'`` for ``underflow``.  But this can be changed, and it can be
+set individually for different kinds of exceptions. The different behaviors
+are:
+
+- ``'ignore'`` : Take no action when the exception occurs.
+- ``'warn'``   : Print a :py:exc:`RuntimeWarning` (via the Python :py:mod:`warnings` module).
+- ``'raise'``  : Raise a :py:exc:`FloatingPointError`.
+- ``'call'``   : Call a function specified using the :func:`~numpy.seterrcall` function.
```
ngoldbaum:
```suggestion
- ``'call'``   : Call a specified function.
```

I wouldn't mention `seterrcall` here, that's a legacy interface. All it does it set the "call" key of the default error state.

doc/source/reference/routines.err.rst
```diff
@@ -1,8 +1,86 @@
...
+- ``'print'``  : Print a warning directly to ``stdout``.
+- ``'log'``    : Record error in a Log object specified by :func:`~numpy.seterrcall`.
```
ngoldbaum: ditto here

doc/source/reference/routines.err.rst
```diff
@@ -1,8 +1,86 @@
...
+- ``all``       : apply to all numeric exceptions
+- ``invalid``   : when NaNs are generated
+- ``divide``    : divide by zero (for integers as well!)
+- ``overflow``  : floating point overflows
+- ``underflow`` : floating point underflows
+
+Note that integer divide-by-zero is handled by the same machinery.
+
+The error handling mode can be configured using :func:`numpy.seterr`,
+:func:`numpy.geterr`, :func:`numpy.seterrcall`, and the :func:`numpy.errstate`
```
ngoldbaum: I wouldn't mention `geterr`, `seterr`, and `seterrcall` here. Or if you feel like they *have* to be mentioned, I'd emphasize that `errstate` is the modern interface and the others should be avoided because context managers are easier to understand and there's no reason for users to implement their own.

doc/source/reference/routines.err.rst
```diff
@@ -1,8 +1,86 @@
...
+context manager.
+
+Examples
+--------
+
+::
+
+   >>> oldsettings = np.seterr(all='warn')
```
ngoldbaum: Should use `np.errstate`, ditto all the examples below.

doc/source/reference/routines.err.rst
```diff
@@ -1,8 +1,86 @@
...
+   >>> np.zeros(5,dtype=np.float32)/0.
+   Traceback (most recent call last):
+   ...
+   RuntimeWarning: invalid value encountered in divide
```
ngoldbaum: Why is there a traceback here? It's a warning. Here's what I see in my prompt:

```
>>> import numpy as np
>>> with np.errstate(all='warn'):
...     np.zeros(5, dtype=np.float32)/0
...
<python-input-1>:2: RuntimeWarning: invalid value encountered in divide
array([nan, nan, nan, nan, nan], dtype=float32)
```

doc/source/reference/routines.err.rst
```diff
@@ -1,8 +1,86 @@
...
+   Traceback (most recent call last):
+   ...
+   FloatingPointError: invalid value encountered in sqrt
+
+::
+
+   >>> def errorhandler(errstr, errflag):
+   ...      print("saw stupid error!")
+   >>> np.seterrcall(errorhandler)
+   >>> j = np.seterr(all='call')
```
ngoldbaum: You can collapse this into a single line using `errstate`.

doc/source/reference/routines.err.rst
ngoldbaum: Well, I just gave a bunch of review to modernize it, so might as well with my suggestions applied I guess 😅

doc/source/reference/routines.io.rst
```diff
@@ -59,8 +59,15 @@ Memory mapping files
...
+
 Text formatting options
 -----------------------
+
+Text formatting settings are maintained in a :py:mod:`python:contextvars`,
```
ngoldbaum: I *think* this is the correct syntax. I want this to read grammatically correct no matter what the generated hyperlink text is.

```suggestion
Text formatting settings are maintained in a :py:mod:`context variable <python:contextvars>`,
```

doc/source/reference/thread_safety.rst
```diff
@@ -32,6 +32,14 @@ from use of the `threading` module, and instead might be better served with
 `multiprocessing`. In particular, operations on arrays with ``dtype=object`` do
 not release the GIL.
 
+Context-local state
+-------------------
+
+NumPy maintains some state for ufuncs on a per-thread basis, which means each
```
ngoldbaum:
```suggestion
NumPy maintains some state for ufuncs context-local basis, which means each
```

doc/source/reference/thread_safety.rst
```diff
@@ -32,6 +32,14 @@ from use of the `threading` module, and instead might be better served with
...
+thread/task has its own independent configuration of the `numpy.errstate` (see
```
ngoldbaum:
```suggestion
thread in a multithreaded program or task in an asyncio program has its own independent configuration of the `numpy.errstate` (see
```

doc/source/reference/thread_safety.rst
```diff
@@ -32,6 +32,14 @@ from use of the `threading` module, and instead might be better served with
...
+:doc:`/reference/routines.err`), and of
+:ref:`text_formatting_options`.
```
ngoldbaum: I'd add the following sentence to the end of this paragraph:

You can update state stored in a context variable by entering a context manager. As soon as the context manager exits, the state will be reset to its value before entering the context manager.

numpy/_core/_ufunc_config.py
ngoldbaum: I think putting it under notes makes sense. Maybe in a special box?

Carreau: I want to avoid special boxes until we settle on a shared directive.

Carreau:
Squashed & rebased, sorry for the loss of history bur fixing each commit conflict with #30394 was annoying.

doc/source/reference/routines.err.rst
Carreau: I don't think the _have_ to be mentioned, there are 2 tensions here:
 - do we address users of code that already contain those, in which case it maybe we should.
 - Or do we target new uses that should not even be aware of seterr and co. so that it can be removed. I think we are more in the second case.

doc/source/reference/routines.err.rst
Carreau: All exmamples re-written using errsate.

doc/source/reference/thread_safety.rst
Carreau: I don't completely agree we need that; but I don't think it harms to add it. 

I read this as "if you set it in one thread/task, it will only affects this thread/task" (which is either good or bad depending on what you want to achieve), the fact that you can use a context manager is independent of that. 

Said in a different way, we state things about spacial-locality[^1] (thread/task), vs time-locality (context-manager)


[^1]: Yes, I know spacial locality is more about where the data is, 'execution context' if you prefer, but I want to avoid 1 more use of "context", and yes threads may interleave in time as well, but I don't think users think about threads this way.

Carreau:
I think I got everything.

mattip (approved):

ngoldbaum (approved):
Awesome, huge improvement!

@charris, I'm adding the backport-candidate label because I'd like to see these docs included in the 2.4.0 release. Feel free to disregard if you'd prefer not to.

PR #30375, opened 2025-12-04
MAINT: don't assert RecursionError in monster dtype test

### LINKED ISSUE #30370: Python 3.14.1t seems to be failing `test_tuple_recursion` on macos x86_64
See #30369 that pins to 3.14.0t as a fix.

ngoldbaum:
Fixes #30370.

Newer CPython versions can recurse more deeply without raising a RecursionError and blowing out the stack. It happens that on x86_64 MacOS the `np.dtype` call succeeds and no error is raised. IMO it's fine if there's no error, the test was simply checking for the observed behavior of CPython and verifying that the case doesn't lead to a seg fault. See https://github.com/numpy/numpy/commit/bb76bff17f86e468ca9fbc5071d95782dc932c10 for context.

### DIFF
--- a/numpy/_core/tests/test_dtype.py
+++ b/numpy/_core/tests/test_dtype.py
@@ -991,8 +991,12 @@ class TestMonsterType:
         d = np.int32
         for i in range(100000):
             d = (d, (1,))
-        with pytest.raises(RecursionError):
+        # depending on OS and Python version, this might succeed
+        # see gh-30370 and cpython issue #142253
+        try:
             np.dtype(d)
+        except RecursionError:
+            pass
 
     @pytest.mark.skipif(IS_PYSTON, reason="Pyston disables recursion checking")
     @pytest.mark.skipif(IS_WASM, reason="Pyodide/WASM has limited stack size")
--- a/numpy/_core/tests/test_dtype.py
+++ b/numpy/_core/tests/test_dtype.py
@@ -1,3 +1,4 @@
+import contextlib
 import ctypes
 import gc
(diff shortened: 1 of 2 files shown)


numpy/_core/tests/test_dtype.py
```diff
@@ -991,8 +991,12 @@ def test_tuple_recursion(self):
         d = np.int32
         for i in range(100000):
             d = (d, (1,))
-        with pytest.raises(RecursionError):
+        # depending on OS and Python version, this might succeed
+        # see gh-30370 and cpython issue #142253
+        try:
             np.dtype(d)
+        except RecursionError:
+            pass
```
jorenham: nit: [`contextlib.suppress`](https://docs.python.org/3/library/contextlib.html#contextlib.suppress) could also be used here

colesbury:
This makes sense to me. For what it's worth, I was able to repro this with `setup/python` on GHA but not `uv`'s version of Python. I think it's just that we don't hit the recursion limit at 100,000 iterations. Increasing the iterations to 1,000,000 seemed to trigger the RecursionError on macOS x86-64:

Here are runs from a test repo:

https://github.com/colesbury/issue142253/actions/runs/19940622355 (100,000)
https://github.com/colesbury/issue142253/actions/runs/19940738167 (1,000,000)

charris:
Thanks Nathan. Should we do the same for the list recursion test?

ngoldbaum:
Nope, that one is self-referential so it has to error eventually.

Issue #29315, opened 2025-07-03
BUG: quantile inconsitent with median for size=0


DerWeh:
### Describe the issue:

For empty inputs (size=0), quantile yields an Index error:

```python
>> np.quantile([], 0.5)
IndexError: index -1 is out of bounds for axis 0 with size 0
```

While median yields `np.float64(nan)` (with warnings):

```python
>> np.median([])
np.float64(nan)
```

The `See also` section, suggests ``np.quantile(..., 0.5) == np.median(...)``, else I find no information on the case size=0, or any mention of raised errors.
The behavior of median is consistent with mean, which follows the specification of https://data-apis.org/array-api/latest/API_specification/generated/array_api.mean.html.

I would expect quantile to return NaN for empty arrays.

### Reproduce the code example:

```python
import numpy as np
np.quantil([], 0.5)
```

### Error message:

```shell

```

### Python and NumPy Versions:

2.2.6
3.12.6 | packaged by conda-forge | (main, Sep 30 2024, 18:08:52) [GCC 13.3.0]

DerWeh:
Probably, we need more testing... [hypothesis](https://hypothesis.readthedocs.io/en/latest/) reveals more incorrect edge cases...

```python
>> import numpy as np
>> np.quantile(np.array([32767, -1], dtype=np.int16), 0.5)
np.float64(49151.0)
>> np.median(np.array([32767, -1], dtype=np.int16))
np.float64(16383.0)
```

Apparently, the cast to `np.float64` was done after calculating the difference... 
(Tested also with NumPy 2.3.1)

I would suggest adding hypothesis tests `np.quantile(..., 0) == np.median(...)` and `np.min(...) <= np.quantile(..., q) <= np.max(...)` (of course, paying attention to NaNs).

imran4444shaik:
Thanks for raising this! 

Right now, **np.quantile([], x)** raises an IndexError on empty inputs even though **np.median([])** and **np.mean([])** both return nan, so the API is inconsistent. We can resolve this by inserting an early guard clause that checks for zero‑length axes and directly returns a float64 array of nan.

I’d be happy to draft a PR implementing these changes and adding unit tests to cover these edge cases!

eendebakpt:
The float16 edge cases might be addressed by https://github.com/numpy/numpy/pull/29105/

DerWeh:
[...]

PR #29105, opened 2025-05-31
BUG: Allow np.percentile to operate on float16 data

eendebakpt:
Fixes #29003

The approach taken to address the issue is to perform index calculations with increased precision, but keep the behaviour that for `arr` of dtype `float16` the output `np.quantile(arr, q)` is of the same type.

The behaviour to have the same output type makes sense for float input, but cannot hold for integer input. E.g. we want `np.quantile([1, 2], 0.5, method='linear')` to output 1.5. For that reason it might also be ok to just upcase the type to float64.

There are also some corner cases. For example input (of either the array or the quantile) of type `Fraction`. Possible output types could be either `Fraction` or `float64`.

PR #30376, opened 2025-12-04
BUG: fix numpy.quantile for integer inputs

DerWeh:
As mentioned in #29315 (see https://github.com/numpy/numpy/issues/29315#issuecomment-3033546960), `numpy.quantile` currently can overflow for integer inputs in spite of returning a float:

```python
>>> np.quantile(np.array([32767, -1], dtype=np.int16), 0.5)
np.float64(49151.0)
```

This can be fixed by casting integers to floats, before interpolating.
This yields the correct result:
```python
>>> np.quantile(np.array([32767, -1], dtype=np.int16), 0.5)
np.float64(16383.0)
```
We ensure a consistent behavior by adding a property test:

```python
quantile = np.quantile(a, q)
assert np.min(a) <= quantile
assert quantile <= np.max(a)
```

So far, this is just a quick fix, I haven't check if all paths are correct.

We probably need to parameterize by the additional parameters of `np.quantile`.
I just wanted to first cross-check if this is the right approach to fix the issue.

At this point, we still incorrectly get:

```python
>>> np.quantile([1], 0)
np.int64(1)
```
While the documentation ensures us that we should get `np.float64(0.0)`.
Currently, the return type depends on the value of `q`:

```python
>>> np.quantile([1], 0.5)
np.float64(1.0)
```

which is undesired behavior.
I can add the return type to the property test.
We also haven't looked into the second part of the documentation: "If the input contains ... floats smaller than float64, the output data-type is float64."

-----

Another point worth discussing is using `np.int64` with values bigger than can be exactly represented by `np.float64` (i.e. integers larger than 2⁵³).
I am not quite sure, if we might lose accuracy in this case.
On the other hand, it is clearly documented that we cat `np.float64`, so we should expect a loss of accuracy for integers outside the representable range.

### DIFF
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -4794,6 +4794,11 @@ def _quantile(
                                method_props, gtype)
             result_shape = virtual_indexes.shape + (1,) * (arr.ndim - 1)
             gamma = gamma.reshape(result_shape)
+            if arr.dtype.kind in ("i", "u"):
+                # as documented, if the inputs are integers, return float64
+                # cast before _lerp to avoid overflows
+                previous = np.astype(previous, np.float64)
+                next = np.astype(next, np.float64)
             result = _lerp(previous,
                         next,
                         gamma,
--- a/numpy/lib/tests/test_function_base.py
+++ b/numpy/lib/tests/test_function_base.py
@@ -9,7 +9,8 @@ from functools import partial
 import hypothesis
 import hypothesis.strategies as st
 import pytest
-from hypothesis.extra.numpy import arrays
+from hypothesis.extra.array_api import make_strategies_namespace
+from hypothesis.extra.numpy import array_shapes, arrays
 
 import numpy as np
 import numpy.lib._function_base_impl as nfb
@@ -4323,6 +4324,22 @@ class TestQuantile:
         value = np.quantile(a, q)
         assert value == q * 50_000
         assert value.dtype == np.float16
+
+
+@hypothesis.given(
+    a=arrays(
+        # real_dtype online in array_api, not in numpy
+        dtype=make_strategies_namespace(np).real_dtypes(),
+        shape=array_shapes(max_dims=1, min_side=1),
+    ).filter(lambda a: np.all(np.isfinite(a))),
+    q=st.floats(min_value=0, max_value=1),
+)
+@pytest.mark.filterwarnings("ignore::RuntimeWarning")
+def test_quantile_properties(a, q):
+    """Check that quantile produces a value in [min(a), max(a)]."""
+    quantile = np.quantile(a, q)
+    assert np.min(a) <= quantile
+    assert quantile <= np.max(a)
 
 
 class TestLerp:


DerWeh:
I fail the test
```python
    def test_quantile_gh_29003_Fraction(self):
        r = np.quantile([1, 2], q=Fraction(1))
        assert r == Fraction(2)
        assert isinstance(r, Fraction)

        r = np.quantile([1, 2], q=Fraction(.5))
        assert r == Fraction(3, 2)
        assert isinstance(r, Fraction)
```
Introduced in #29105. 
@eendebakpt says: 
> There are also some corner cases. For example input (of either the array or the quantile) of type `Fraction`. Possible output types could be either `Fraction` or `float64`.

This is undocumented behavior, and I am not quite sure in which case you would want that...

Documentation only says:
> If the input contains integers or floats smaller than float64, the output data-type is float64. Otherwise, the output data-type is the same as that of the input. If out is specified, that array is returned instead.

So my claim is, that if you for whatever reason want `Fraction` as output, you should either explicitly give `out` or convert `[1, 2]` to an array of `Fraction`s.
To my understanding, the test contradicts the documentation.

numpy/lib/tests/test_function_base.py
```diff
@@ -4325,6 +4326,22 @@ def test_float16_gh_29003(self):
...
+@hypothesis.given(
+    a=arrays(
+        # real_dtype online in array_api, not in numpy
+        dtype=make_strategies_namespace(np).real_dtypes(),
+        shape=array_shapes(max_dims=1, min_side=1),
+    ).filter(lambda a: np.all(np.isfinite(a))),
+    q=st.floats(min_value=0, max_value=1),
+)
+@pytest.mark.filterwarnings("ignore::RuntimeWarning")
+def test_quantile_properties(a, q):
```
MaanasArora, line 4338: Should this not be under the `TestQuantile` class? Also, perhaps a more descriptive name would be better (maybe `test_quantile_within_bounds`)?

numpy/lib/_function_base_impl.py
```diff
@@ -4794,6 +4794,11 @@ def _quantile(
                                method_props, gtype)
             result_shape = virtual_indexes.shape + (1,) * (arr.ndim - 1)
             gamma = gamma.reshape(result_shape)
+            if arr.dtype.kind in ("i", "u"):
```
MaanasArora, line 4797: Can this be merged into the if statement above?

https://github.com/numpy/numpy/blob/ad80c0a3dde15d331487f337ed4b52b9eb968078/numpy/lib/_function_base_impl.py#L4785-L4786

MaanasArora (commented):
Seems to patch much of the issue! Agreed that the docs are a bit inconsistent for floats still, so let's wait for maintainer review.

Not sure about the fraction edge case, not familiar with this code. But who knows if this will fail with other object dtypes - perhaps one could special-case for `q` or even `gamma`.

numpy/lib/_function_base_impl.py
DerWeh: Yes, you are probably right. But first, we (or I) need to cover all branches. As pointed out above, in the current draft version, the return type still depends on the quantile `q`.

I hoped to get feedback from the maintainer if I'm on the right track, before putting in all the effort. As soon as we have the correct tests in place, we should indeed try to simplify the code. Thanks for the input.

numpy/lib/tests/test_function_base.py
DerWeh: > Should this not be under the `TestQuantile` class?

I don't know, but I can put it there. The function doesn't depend on any attributes and I see no application for inheritance, so I thought a free/plain function is more idiomatic than a method.


> Also, perhaps a more descriptive name would be better (maybe `test_quantile_within_bounds`)?

If it stays like this, I agree. I planned on adding the type checks here (and any other property we agree on) and extending the test to cover all `method`s.

Alternatively, we could write multiple short functions, but it seemed more efficient to me, to scan all inputs only once.

numpy/lib/_function_base_impl.py
seberg: FWIW, I think this should be moved into `_lerp` if anything. (this brings up my long term thought that a `lerp` ufunc would maybe the cleanest thing to solve all of these issues.)
If you have it there, you can also make it quicker than just doing a full cast here I think (at least for very large arrays).  Although it may be nice to at least not do this when the input is clearly OK.

EDIT: If you are actually interested in that, at least I think that adding a `lerp` ufunc would be worthwhile.

PR #30380, opened 2025-12-05
ENH: float16.h in NumPy

athurdekoos:
This issue will close #30280

## Summary of Changes:
- New `float16.h` 
  - This is a header only implementation of `halffloat.h`. Effectively eliminating the need to include `npy_math` when using `halffloat.h` functions. 
  - A majority of the logic has been `Tactically Acquired` from the old [`halffloat.c`](https://github.com/numpy/numpy/blob/bba99dbd606d173a12c7de1867e2337de17c60e9/numpy/core/src/npymath/halffloat.c) implementation to reduce the need to rewrite everything form scratch. 
  - I have placed header guards in `float16.h` where needed to allow for both `halffloat.h` and `float16.h` in case of simultaneous includes. Functions have also been renamed to avoid conflict.
  - `#define NPY_HALF_GENERATE_OVERFLOW ` and `#define NPY_HALF_GENERATE_INVALID` has been removed and logic has been updated to remove the code path.
 
- New `test_float16.py` and `_float16_tests.c` to support tests for float16

Update 12/8/25:
- New `tests/examples/float16/` directory has been created.
  - `_float16_tests.c` has been relocated to this directory.
  - New `float16/setup.py` & `float16/meson.build` files have been created to assist with testing. 
- Update in`test_float16.py` with logic similar to [tests/test_limited_api.py](https://github.com/numpy/numpy/blob/60bf05c38a1059cef0d897a83c626bad702ae1ba/numpy/_core/tests/test_limited_api.py) to better test `float16.h` per [Ralf's recommendation ](https://github.com/numpy/numpy/pull/30380#pullrequestreview-3549564449) (thanks Ralf!)

Update 12/9/25:
- New `float16.rst` document has been added for documentation on the new `float16.h` file. 
- Update in `coremath.rst`, a user facing note has been added to steer users to `float16.h`.
- Update in`c-api/index.rst` to add the new `float16.rst` document to the file tree.

Update 12/10/25:
- Update in `coremath.rst` did not have a a reference to `NPY_MAX_HALF`, a new one has been created and added. 
- Update in `float16.h`,  `npy_float16_spacing` was removed.
  -  `_npy_float16_header_invalid` was removed as it is now unneeded.


athurdekoos:
I would love some feedback on[ _float16_tests.c ](https://github.com/numpy/numpy/blob/ff6ada95a9dac0457d9b86dcfa4510a60c2e9c23/numpy/_core/src/npymath/_float16_tests.c)

I have the same function logic essentially repeated multiple times right now to help with debug but I am a bit back and forth on if I should just factor this out.

numpy/_core/src/npymath/_float16_tests.c
```diff
@@ -0,0 +1,364 @@
+#define PY_SSIZE_T_CLEAN
+#include <Python.h>
+
+#define NPY_NO_DEPRECATED_API NPY_API_VERSION
+#include "numpy/arrayobject.h"
+#include "numpy/npy_common.h"
+#include "common.h"  /* for error_converting */
```
rgommers: This isn't a public header, so should not be used in the test (an actual user of `float16.h` would not have access to this).

numpy/_core/src/npymath/_float16_tests.c
```diff
@@ -0,0 +1,364 @@
...
+ */
+static PyObject *
+float16tests_float16_to_float(PyObject *NPY_UNUSED(self), PyObject *args)
+{
+    unsigned int hbits;
+    if (!PyArg_ParseTuple(args, "I", &hbits)) {
+        return NULL;
+    }
+
+    npy_half h = (npy_half)(npy_uint16)hbits;
```
rgommers: Honest question: are the double casts necessary? What happens when casting straight to `npy_half`?

rgommers (commented):
Thanks @athurdekoos. The implementation looks pretty solid. I didn't verify line by line that it matches the old `halffloat.c` content yet, but I assume it does - and everything else LGTM. The one thing there is to decide whether to include all functions or not, which can be left to the very end.

The tests look less ideal to me. The compilation uses `common.h` which wouldn't work for an external user. And shipping a pre-compiled extension module increases the binary size and is less good of a test than compiling something at test time. That's what is done for some other tests, like in `test_cython.py` and `test_limited_api.py`. I think the limited API one is the example to follow:
- The C code that uses `float16.h` goes under `_core/tests/examples/` (like https://github.com/numpy/numpy/tree/60bf05c38a1059cef0d897a83c626bad702ae1ba/numpy/_core/tests/examples/limited_api)
- The `test_float16.py` can use this pattern to compile an extension module at test time: https://github.com/numpy/numpy/blob/60bf05c38a1059cef0d897a83c626bad702ae1ba/numpy/_core/tests/test_limited_api.py
    - Note that the test skips will be simpler than for the limited API test - you should only need the `IS_EDITABLE` and `IS_WASM` ones

Could you try to change to that testing method?

numpy/_core/src/npymath/_float16_tests.c
athurdekoos: Not at all, that was relic from the hacking phase, removed.

numpy/_core/src/npymath/_float16_tests.c
athurdekoos: Removed

numpy/_core/include/numpy/float16.h
```diff
@@ -0,0 +1,662 @@
...
+#ifndef NPY_HALF_NAN
+#define NPY_HALF_NAN    (0x7e00u)
+#endif
+
+#ifndef NPY_MAX_HALF
+#define NPY_MAX_HALF    (0x7bffu)
+#endif
+
+#include "numpy/ndarraytypes.h"
+#include "numpy/npy_common.h"
```
rgommers: These should go at the top with the other includes

rgommers:
Looks good now, I marked it as ready for review.

rgommers:
Actually modulo docs that it. The docs should explain the `float16.h` usage. And I think https://numpy.org/devdocs/reference/c-api/coremath.html#half-precision-functions should get a `.. note::` recommending to not use `halffloat.h` anymore but `float16.h` instead because it doesn't require linking `libnpymath`.

numpy/_core/tests/examples/float16/meson.build
```diff
@@ -0,0 +1,20 @@
...
+    ], check: true).stdout().strip()
+
+py.extension_module(
+  '_float16_tests',
+  '_float16_tests.c',
+  install: false,
+  include_directories: [npy_include_path],
+    c_args: [
+      '-DNPY_TARGET_VERSION=NPY_2_0_API_VERSION',
+    ],
```
rgommers: These 3 lines can be deleted, they're not necessary (unless I am missing something). Indentation is also off.

ngoldbaum:
I wonder if it makes sense to also move the floating point status error handling code into its own header-only library as well. That would avoid duplicating it in `float16.h`.

Do we eventually want to make numpy's half float use this new library to implement the dtype? This PR would leave numpy's half float implementation using the npymath implementation still.

athurdekoos:
`c-api/float16.rst` was added as a new stand alone file and added to the file tree under c-api. I recognize that this is likely not the most ideal location for this document, however I also did not believe it should be included in `c-api/coremath.rst`. 

Feedback on appropriate location for this document would be appreciated.

numpy/_core/tests/examples/float16/meson.build
athurdekoos: Removed

rgommers:
> I wonder if it makes sense to also move the floating point status error handling code into its own header-only library as well. That would avoid duplicating it in `float16.h`.

We can't change the public `void npy_set_floatstatus_overflow` version, so a separate public header-only version would need to be duplicate anyway. I can't find usages in OSS code that aren't just copies of `numpy` code, so I think we'd better not do this until a need shows up?

> Do we eventually want to make numpy's half float use this new library to implement the dtype? This PR would leave numpy's half float implementation using the npymath implementation still.

I doubt it. The internal implementation was moved to C++ a while back, and seems to do the job. Using the version in this PR would probably also increase the binary size a bit due to multiple copies from `static inline`?

doc/source/reference/c-api/float16.rst
```diff
@@ -0,0 +1,263 @@
+.. _float16:
+
+Header only half-precision functions
```
rgommers: I'd drop the "Header only" here. It's useful to mention in the second paragraph as you did, but in the end it's an implementation detail - the section header should just say what the topic of the page is. It shows up in the sidebar as well. So I'd say _Half-precision API_ or _float16 API_ (API rather than functions for consistency, see [the rendered version](https://output.circle-artifacts.com/output/job/36f168fc-9715-4153-932e-b64f0dbe473e/artifacts/0/doc/build/html/reference/c-api/float16.html))

doc/source/reference/c-api/float16.rst
```diff
@@ -0,0 +1,263 @@
...
+
+__ https://ieeexplore.ieee.org/document/4610935/
+__ https://en.wikipedia.org/wiki/Half-precision_floating-point_format
+__ https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_half_float_pixel.txt
+__ https://www.openexr.com/about.html
+
+Constants
+---------
+
+:c:macro:: NPY_HALF_ZERO
```
rgommers: These aren't rendered correctly, see the rendered version in CI

numpy/_core/tests/examples/float16/meson.build
rgommers: note that indentation is still off in this file, it should be consistently 2 or 4 spaces. You can run `meson fmt` to autofix

numpy/_core/tests/examples/float16/setup.py
```diff
@@ -0,0 +1,25 @@
...
+Build an example package using the limited Python C API.
+"""
+
+import os
+
+from setuptools import Extension, setup
+
+import numpy as np
+
+macros = [("NPY_TARGET_VERSION", "NPY_2_0_API_VERSION")]
```
rgommers: Can you delete this? And the `define_macros=macros,` line. It's not needed and was already removed from the `meson.build` file.

numpy/_core/tests/examples/float16/setup.py
```diff
@@ -0,0 +1,25 @@
...
+    "_float16_tests",
+    sources=[os.path.join('.', "_float16_tests.c")],
+    include_dirs=[np.get_include()],
+    define_macros=macros,
+)
+
+extensions = [float16_tests]
+
+setup(
+    ext_modules=extensions
```
rgommers: `ext_modules=[float16_tests]` and remove line 21 above?

with both those changes, this file becomes pretty concise.

numpy/_core/tests/examples/float16/setup.py
```diff
@@ -0,0 +1,25 @@
+"""
+Build an example package using the limited Python C API.
```
rgommers: copy-paste issue, it's not the limited API but about float16 C API

doc/source/reference/c-api/float16.rst
```diff
@@ -0,0 +1,248 @@
+.. _float16:
+
+Half-precision API
+==================
+
+The header file ``<numpy/float16.h>`` provides helper routines for working
+with IEEE 754-2008 16-bit floating-point values
```
seberg, line 7: Not sure where, but the `float16.h` header is special in the sense that none of the functions here require "importing" NumPy (since they don't call any functions).
I think it would be nice to mention that somewhere.

I guess we need `ndarraytypes.h` to have `npy_float16` and `npy_half` defined? Would be cool if we didn't need the full thing, or need it at all (it's rather large).
But I am not really worried about that.

doc/source/reference/c-api/float16.rst
```diff
@@ -0,0 +1,248 @@
...
+
+Unlike ``<numpy/halffloat.h>``, this header is a header only implmentation
+and does not require linking against the ``npymath`` static library. All
+functionality is implemented directly in the header using ``<math.h>`` and
+``<fenv.h>``.
```
seberg: I know you'll see the warning when compiling with older versions. But how about adding a deprecation warning to `numpy/halffloat.h` so we can forget about it soonish hopefully.

doc/source/reference/c-api/float16.rst
```diff
@@ -0,0 +1,248 @@
...
+
+__ https://ieeexplore.ieee.org/document/4610935/
+__ https://en.wikipedia.org/wiki/Half-precision_floating-point_format
+__ https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_half_float_pixel.txt
+__ https://www.openexr.com/about.html
+
+Constants
+---------
+
+:c:macro:`NPY_HALF_ZERO`
```
seberg: Should we just name these `NPY_FLOAT16_ZERO` also?  `npy_float16` is a valid alias anyway so if users have to update the function names already, might as well update these?

numpy/_core/tests/examples/float16/_float16_tests.c
```diff
@@ -0,0 +1,343 @@
...
+};
+
+static int
+float16tests_exec(PyObject *m)
+{
+    Py_INCREF(m);
+    if (PyModule_AddObject(m, "float16", m) < 0) {
+        Py_DECREF(m);
+        return -1;
+    }
```
seberg: Why would the module add itself to the namespace?

numpy/_core/tests/examples/float16/_float16_tests.c
```diff
@@ -0,0 +1,343 @@
...
+        return NULL;
+    }
+
+    npy_half h1 = h1bits;
+    npy_half h2 = h2bits;
+
+    npy_half mod;
+    npy_half quo = npy_float16_divmod(h1, h2, &mod);
+
+    return Py_BuildValue("II", quo, mod);
```
seberg: This seems incorrect on some platforms. You need to make sure the types match. I assume `I` is `unsigned int`, so you have to type or cast `mod` and `quo` accordingly.  (`PyLong_FromUnsignedLong` is OK, since it is a typed function.)

(But from a quick glance, I guess that is the only occurrence.)

The same is true for `ParseTuple`, but I think there it is correct everywhere.

seberg (commented):
A few comments, but overall looks good to me, thanks.  I am not quite sure we have settled if we want to remove anything again here.

numpy/_core/tests/examples/float16/_float16_tests.c
seberg: Sorry, nevermind. I think integers shorter than integer are passed as integers in the C standard.
(Recently ran into this issue with integers _longer_ than int.)

doc/source/reference/c-api/float16.rst
athurdekoos: That works for me, it'll be a cleaner implementation overall to use `NPY_FLOAT16*` and allow me to drop some header guards

athurdekoos:
> A few comments, but overall looks good to me, thanks. I am not quite sure we have settled if we want to remove anything again here.

I think it's worth removing the *nonane variants. Is there anything else that should go?

seberg:
To me, `divmod` seems like an odd one out that is probably not worth having.

numpy/_core/tests/examples/float16/_float16_tests.c
athurdekoos: This was a silly idea, removed and updated

doc/source/reference/c-api/float16.rst
athurdekoos: I've worded to this effect

doc/source/reference/c-api/float16.rst
athurdekoos: Full disclosure, I am not sure how to properly do this just yet. I've tried a few different methods, but I need to dig deeper into the docs and PR history to see if I can find an elegant way to do this since any compiler warning seem to cause the CI to fail. I've added a comment to halffloat.h in the meantime.

seberg:
Thanks for the update. From my side, I happy if this gets in as is (and if there is feedback we can incorporate it).
If we end up with the deprecation warning might as well wait till after 2.4 is released (around the corner anyway).

This should get a release note fragment as well. It can be very brief though and mostly link to the other (very nice!) docs.  (If you have not done that yet, see `doc/release/upcoming_changes/README.md`)

athurdekoos:
Looks like my CI jobs are failing due to a unrelated docker issue.

PR #30381, opened 2025-12-05
ENH: Pixi package definitions for downstream development

lucascolley:
For https://github.com/scipy/scipy/pull/24066. cc @ngoldbaum @FFY00 @rgommers 

Depends on https://github.com/python/cpython/pull/142469.

### DIFF
--- a/.github/workflows/pixi-packages.yml
+++ b/.github/workflows/pixi-packages.yml
@@ -1,0 +1,39 @@
+name: Pixi packages tests
+
+on:
+  pull_request:
+    branches:
+      - main
+      - maintenance/**
+    paths-ignore:
+      - '**.pyi'
+      - '**.md'
+      - '**.rst'
+
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
+jobs:
+  build_packages:
+    name: Build Pixi packages
+    runs-on: ${{ matrix.runs-on }}
+    strategy:
+      fail-fast: false
+      matrix:
+        runs-on: [ubuntu-latest, macos-15]
+        package_variant: [asan, default]
+    if: github.repository == 'numpy/numpy'
+    steps:
+    - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+      with:
+        submodules: recursive
+        fetch-tags: true
+        persist-credentials: false
+
+    - uses: prefix-dev/setup-pixi@ba3bb36eb2066252b2363392b7739741bb777659 # v0.8.1
+      with:
+        pixi-version: v0.60.0
+        run-install: false
+
+    - name: Build
+      run: pixi build --path="pixi-packages/${{ matrix.package_variant }}"
--- a/pixi-packages/asan/pixi.toml
+++ b/pixi-packages/asan/pixi.toml
@@ -17,9 +17,9 @@ env.ASAN_OPTIONS = "detect_leaks=0:symbolize=1:strict_init_order=true:allocator_
 extra-args = ["-Csetup-args=-Db_sanitize=address", "-Csetup-args=-Dbuildtype=debug"] 
 
 [package.host-dependencies]
-python.git = "https://github.com/lucascolley/cpython"
+python.git = "https://github.com/python/cpython"
 python.subdirectory = "Tools/pixi-packages/asan"
-python.rev = "453e9eebd5e74ab58f3a0c43232e57d5e5e1020c"
+python.rev = "8bb5b6e8ce61da41b5affd5eb12d9cc46b5af448"
 
 meson-python = "*"
 cython = "*"
--- a/pixi-packages/asan/pixi.toml
+++ b/pixi-packages/asan/pixi.toml
@@ -1,0 +1,26 @@
+[workspace]
+channels = ["https://prefix.dev/conda-forge"]
+platforms = ["osx-arm64", "linux-64"]
+preview = ["pixi-build"]
+
+[package.build]
+source.path = "../.."
+
+[package.build.backend]
+name = "pixi-build-python"
+version = "*"
+
+[package.build.config]
+extra-input-globs = ["**/*.c.src"]
+compilers = ["c", "cxx"]
+env.ASAN_OPTIONS = "detect_leaks=0:symbolize=1:strict_init_order=true:allocator_may_return_null=1:use_sigaltstack=0"
+extra-args = ["-Csetup-args=-Db_sanitize=address", "-Csetup-args=-Dbuildtype=debug"] 
+
+[package.host-dependencies]
+python.git = "https://github.com/lucascolley/cpython"
+python.subdirectory = "Tools/pixi-packages/asan"
+python.rev = "375594b556fb6c41c0630c10b7e934fb89604712"
+
+meson-python = "*"
+cython = "*"
+uv = "*" # used to invoke the wheel build
--- a/pixi-packages/asan/pixi.toml
+++ b/pixi-packages/asan/pixi.toml
@@ -19,7 +19,7 @@ extra-args = ["-Csetup-args=-Db_sanitize=address", "-Csetup-args=-Dbuildtype=deb
 [package.host-dependencies]
 python.git = "https://github.com/lucascolley/cpython"
 python.subdirectory = "Tools/pixi-packages/asan"
-python.rev = "54274f2c4e017ef58b5cd73ce45f9277ee5c52f5"
+python.rev = "453e9eebd5e74ab58f3a0c43232e57d5e5e1020c"
 
 meson-python = "*"
 cython = "*"
--- a/pixi-packages/asan/pixi.toml
+++ b/pixi-packages/asan/pixi.toml
@@ -19,7 +19,7 @@ extra-args = ["-Csetup-args=-Db_sanitize=address", "-Csetup-args=-Dbuildtype=deb
 [package.host-dependencies]
 python.git = "https://github.com/lucascolley/cpython"
 python.subdirectory = "Tools/pixi-packages/asan"
-python.rev = "375594b556fb6c41c0630c10b7e934fb89604712"
+python.rev = "54274f2c4e017ef58b5cd73ce45f9277ee5c52f5"
 
 meson-python = "*"
 cython = "*"
--- a/pixi-packages/default/pixi.toml
+++ b/pixi-packages/default/pixi.toml
@@ -1,0 +1,21 @@
+[workspace]
+channels = ["https://prefix.dev/conda-forge"]
+platforms = ["osx-arm64", "linux-64"]
+preview = ["pixi-build"]
+
+[package.build]
+source.path = "../.."
+
+[package.build.backend]
+name = "pixi-build-python"
+version = "*"
+
+[package.build.config]
+extra-input-globs = ["**/*.c.src"]
+compilers = ["c", "cxx"]
+
+[package.host-dependencies]
+python = "*"
+meson-python = "*"
+cython = "*"
+uv = "*" # used to invoke the wheel build
--- a/doc/release/upcoming_changes/30381.new_feature.rst
+++ b/doc/release/upcoming_changes/30381.new_feature.rst
@@ -1,0 +1,11 @@
+Pixi package definitions
+------------------------
+Pixi package definitions have been added for different kinds
+of from-source builds of NumPy. These can be used in
+downstream Pixi workspaces via the ``pixi-build`` feature.
+
+Definitions for both ``default`` and AddressSanitizer-instrumented
+(``asan``) builds are available in the source code under the
+``pixi-packages/`` directory.
+
+``linux-64`` and ``osx-arm64`` platforms are supported.
--- a/pixi-packages/README.md
+++ b/pixi-packages/README.md
@@ -1,2 +1,37 @@
-- see https://github.com/scipy/scipy/pull/24066 for how these packages can be used in a Pixi workspace
-- combining directories blocked on https://github.com/prefix-dev/pixi/issues/2813
+# CPython Pixi packages
+
+This directory contains definitions for [Pixi packages](https://pixi.sh/latest/reference/pixi_manifest/#the-package-section)
+which can be built from the NumPy source code.
+
+Downstream developers can make use of these packages by adding them as Git dependencies in a
+[Pixi workspace](https://pixi.sh/latest/first_workspace/), like:
+
+```toml
+[dependencies]
+numpy = { git = "https://github.com/numpy/numpy", subdirectory = "pixi-packages/asan" }
+```
+
+This is particularly useful when developers need to build NumPy from source
+(for example, for an ASan-instrumented build), as it does not require any manual
+clone or build steps. Instead, Pixi will automatically handle both the build
+and installation of the package.
+
+See https://github.com/scipy/scipy/pull/24066 for a full example of downstream use.
+
+Each package definition is contained in a subdirectory.
+Currently defined package variants:
+
+- `default`
+- `asan`: ASan-instrumented build with `-Db_sanitize=address`
+
+## Maintenance
+
+- Keep host dependency requirements up to date
+- For dependencies on upstream CPython Pixi packages, keep the git revision at a compatible version
+
+## Opportunities for future improvement
+
+- More package variants (such as TSan, UBSan)
+- Support for Windows
+- Using a single `pixi.toml` for all package variants is blocked on https://github.com/prefix-dev/pixi/issues/2813
+- Consider pinning dependency versions to guard against upstream breakages over time
--- a/pixi-packages/README.md
+++ b/pixi-packages/README.md
@@ -1,0 +1,2 @@
+- see https://github.com/scipy/scipy/pull/24066 for how these packages can be used in a Pixi workspace
+- combining directories blocked on https://github.com/prefix-dev/pixi/issues/2813


pixi-packages/asan/pixi.toml
```diff
@@ -0,0 +1,24 @@
...
+extra-input-globs = ["**/*.c.src"]
+compilers = ["c", "cxx"]
+env.ASAN_OPTIONS = "detect_leaks=0:symbolize=1:strict_init_order=true:allocator_may_return_null=1:use_sigaltstack=0"
+extra-args = ["-Csetup-args=-Db_sanitize=address", "-Csetup-args=-Dbuildtype=debug"] 
+
+[package.host-dependencies]
+python = "*"
+meson-python = "*"
+cython = "*"
+uv = "*"
```
lucascolley: `uv` here is not strictly needed, it just means that the wheel build goes via `uv` instead of `pip`, which is probably a bit faster.

pixi-packages/asan/pixi.toml
```diff
@@ -0,0 +1,24 @@
...
+[package.build]
+source.path = "../.."
+
+[package.build.backend]
+name = "pixi-build-python"
+version = "*"
+
+[package.build.config]
+noarch = false
+extra-input-globs = ["**/*.c.src"]
```
lucascolley, line 14: might be able to remove this with a patch to https://github.com/prefix-dev/pixi-build-backends

EDIT: on second thought this is probably fine to keep as explicit

pixi-packages/asan/pixi.toml
```diff
@@ -0,0 +1,24 @@
+[workspace]
+channels = ["https://prefix.dev/conda-forge"]
+platforms = ["osx-arm64"]
```
lucascolley: would be good to cover ~linux and~ windows too

lucascolley (commented):
Currently separate manifests in different directories are required. This might change in the future pending https://github.com/prefix-dev/pixi/issues/2813.

lucascolley:
@ngoldbaum perhaps you would be interested in making a TSan version based on what I have here? I suppose to test that this is understandable and maintainable.

pixi-packages/asan/pixi.toml
```diff
@@ -0,0 +1,24 @@
...
+
+[package.build]
+source.path = "../.."
+
+[package.build.backend]
+name = "pixi-build-python"
+version = "*"
+
+[package.build.config]
+noarch = false
```
lucascolley: can probably remove this, I think it is implied by `compilers = ["c"]`

FFY00:
@lucascolley, could we extend this to `linux-x86_64` too? I feel like if we are gonna upstream the CPython recipes, we would want at least these two platforms. I can help with that if needed :blush:

lucascolley:
yes, go for it! I only have an osx-arm64 machine right now (although not really an excuse for linux with docker...) but will have access to a linux and a windows box next week. Feel free to beat me to it ☺️

ngoldbaum:
>  perhaps you would be interested in making a TSan version based on what I have here? I suppose to test that this is understandable and maintainable.

Sure, is there a CPython recipe I should be using? How do I use this recipe with the CPython recipe?

Maybe it'd help to have a README in the new pixi_packages folder that explains to luddites like me how to use the recipes to do useful development tasks.

lucascolley:
I'm figuring out the details for CPython now, will let you know when ready.

charris:
This will need a release note.

lucascolley:
@ngoldbaum 

> Sure, is there a CPython recipe I should be using? How do I use this recipe with the CPython recipe?

I pushed osx-arm64 recipes for default and ASan CPython at https://github.com/lucascolley/cpython/commit/375594b556fb6c41c0630c10b7e934fb89604712. I have updated this PR so that the NumPy ASan package builds against the CPython ASan package.

Hopefully it is clear how one would go about adding another build type like TSan:
- add linux support to the CPython recipes if needed, by adding `"linux-64"` to `workspace.platforms` in the `pixi.toml` files, seeing what fails on `pixi build`, and making the necessary adjustments
- add support for the build type in the CPython build script, https://github.com/lucascolley/cpython/blob/375594b556fb6c41c0630c10b7e934fb89604712/Tools/pixi-packages/build.sh
- add a new directory with `pixi.toml` and `recipe.yaml` for that build type in CPython, like https://github.com/lucascolley/cpython/blob/375594b556fb6c41c0630c10b7e934fb89604712/Tools/pixi-packages/asan/recipe.yaml
- add a new directory and `pixi.toml` for that build type in NumPy and point to the appropriate CPython directory, like https://github.com/lucascolley/numpy/blob/9f05918520d2579134758b63a45da9a7ebcacaa3/pixi-packages/asan/pixi.toml
- those packages can then be added as dependencies in a Pixi workspace, as in https://github.com/scipy/scipy/pull/24066.

> Maybe it'd help to have a README in the new pixi_packages folder that explains to luddites like me how to use the recipes to do useful development tasks.

Development tasks on which repo? For NumPy, we should probably wait until something like https://github.com/scipy/scipy/blob/main/pixi.toml exists, but they could also be added to an out-of-tree workspace like https://github.com/rgommers/pixi-dev-scipystack/blob/main/numpy/pixi.toml.

For SciPy, we should add some docs, yes, and it would probably be sufficient for a README here to link to those docs?

ngoldbaum:
> For SciPy, we should add some docs, yes, and it would probably be sufficient for a README here to link to those docs?

Sounds reasonable to me.

Thanks so much for working on this. I was hoping you would when you shared all this stuff earlier this year when it was beta.

lucascolley:
my pleasure — it has been super cool to work across both the development and user side of this!

lucascolley:
@FFY00 what information would be good to include on the CPython issue?

lucascolley:
> For NumPy, we should probably wait until something like https://github.com/scipy/scipy/blob/main/pixi.toml exists

gh-30387

lucascolley:
@FFY00 I went ahead and added linux-64 support at https://github.com/lucascolley/cpython/commit/54274f2c4e017ef58b5cd73ce45f9277ee5c52f5, https://github.com/lucascolley/numpy/commit/aa50cf47999591b0594785a7476763ee34769eb6, and https://github.com/scipy/scipy/pull/24066/commits/1d87a865d34294091319f5817bf0df968d6907c8. Tests now running at https://github.com/scipy/scipy/actions/runs/20012600510/job/57384436353.

FFY00:
Sorry for the delay getting back to you.

Thank you so much for all this work, it is extremely helpful!

> @FFY00 what information would be good to include on the CPython issue?

I opened an issue at https://github.com/python/cpython/issues/142466. Feel free to send a PR whenever you are ready!

pixi-packages/default/pixi.toml
```diff
@@ -0,0 +1,21 @@
...
+
+[package.build.config]
+extra-input-globs = ["**/*.c.src"]
+compilers = ["c", "cxx"]
+
+[package.host-dependencies]
+python = "*"
+meson-python = "*"
+cython = "*"
+uv = "*" # used to invoke the wheel build
```
ngoldbaum, line 21: I'm not really a big fan of splitting this up into multiple files - is there a reason why this can't look like scipy's configuration with a single `pixi.toml` in the root of the repository?

ngoldbaum: I'm also still confused how I actually use these definitions in-practice. It'd be really nice to have an explicit example in the readme. I don't think just linking to the SciPy PR is sufficient, sorry for saying that earlier, I hadn't actually tried to set up an environment.

If I want to set up an environment using this definition, what do I actually do? If I do `pixi run python -VV` inside this folder with your PR checked out, I get the output from my Pyenv Python build, not a conda Python build inside a pixi environment like I'd expect:

```
goldbaum at Nathans-MBP in ~/Documents/numpy/pixi-packages/default on pixi-build!
± pixi run "python -VV"
Python 3.14.2 free-threading build (main, Dec  8 2025, 15:41:12) [Clang 17.0.0 (clang-1700.4.4.1)]
```

lucascolley: > I'm not really a big fan of splitting this up into multiple files

as I mentioned [above](https://github.com/numpy/numpy/pull/30381#pullrequestreview-3544372676), combining the directories is blocked on https://github.com/prefix-dev/pixi/issues/2813

>  is there a reason why this can't look like scipy's configuration with a single `pixi.toml` in the root of the repository?

It's important to distinguish between the two concepts of package definitions and workspace definitions. https://github.com/scipy/scipy/blob/main/pixi.toml defines a workspace with many different environments, dependencies for those environments, tasks for those environments etc. but it does not define any Pixi packages that you can build from the SciPy source and use downstream.

This PR does the opposite for NumPy. While it does not define any workspaces with useful environments for developing NumPy itself, it provides definitions of different Pixi packages which can be built from the NumPy source code and used downstream.

> I'm also still confused how I actually use these definitions in-practice. It'd be really nice to have an explicit example in the readme. I don't think just linking to the SciPy PR is sufficient, sorry for saying that earlier, I hadn't actually tried to set up an environment.

Hopefully, given what I've just said, you understand why I opened gh-30387 in response to this question.

> If I do pixi run python -VV inside this folder with your PR checked out, I get the output from my Pyenv Python build, not a conda Python build inside a pixi environment like I'd expect

For that to be expected, there would need to be an environment defined in that workspace with Python as a dependency. So what you expected would come from adding:

```toml
[dependencies]
python.git = "https://github.com/lucascolley/cpython"
python.rev = "54274f2c4e017ef58b5cd73ce45f9277ee5c52f5"
```

which is implicitly equivalent to the following explicit environment definition:

```toml
[environments]
default = { features = ["default"] }

[feature.default.dependencies]
python.git = "https://github.com/lucascolley/cpython"
python.rev = "54274f2c4e017ef58b5cd73ce45f9277ee5c52f5"
```

Does that address your concerns? Just try to keep the environments of a workspace and the `package` that may be defined in that workspace separate in your mind for now. They can become interlinked (by adding `numpy = { path = "." }` to `dependencies`)[^1], but I'm not proposing to do that here.

[^1]: and hopefully more auto-magically in the future, but that is in development, https://github.com/prefix-dev/pixi/pull/4778

lucascolley: I think the best way for you to get to grips with these subtleties would be to work on gh-30387 :)

ngoldbaum: Thanks, that does help.

Why would someone want to depend on the `default` environment you define in this file? Can't someone just depend on numpy's git repository if they need that?

lucascolley: > Why would someone want to depend on the `default` environment? Can't someone just depend on numpy's git repository if they need that?

This _is_ the way to depend on NumPy's git repository (in Pixi _`package`_ land). Now, Pixi also supports PyPI, so you can indeed also depend on NumPy that way and have an editable PyPI dependency. That might be a meaningful difference depending on the use-case, there is some discussion of this at https://pixi.sh/latest/concepts/conda_pypi/. In general though, if you are using pixi-build, you will want to go all in on conda. Pixi can orchestrate recursive pixi-build steps cleverly, and throwing PyPI dependencies in the mix adds an unnecessary layer of complexity.

ngoldbaum: Thanks for explaining! I managed to get it to work now and I see more clearly how this works.

IMO the readme should say something like:

> You can write your own environment definitions that depend on these. For example, if you want to depend on an in-development build of NumPy, the `dependencies` section of your pixi.toml might look like:
>
> ```
> [dependencies]
> numpy.git = "https://github.com/numpy/numpy.git"
> numpy.subdirectory = "pixi-packages/default"
> numpy.rev = "pixi-build"
> python = "*"

You might also want to mention that right now you need `preview = ["pixi-build"]` as well (at least pixi 0.61 told me I did)

lucascolley: By the way, long term it would be great to have the 'default' package definition at the top level. Then all that changes for the user is to put the dependency on the NumPy repository under `dependencies` instead of `pypi-dependencies`.

ngoldbaum: Or even more explicit to start with:
 
> go to an empty directory, do `pixi init`, then add the following to the `pixi.toml`

lucascolley: > IMO the readme should say something like:
> 
> > You can write your own environment definitions that depend on these. For example, if you want to depend on an in-development build of NumPy, the `dependencies` section of your pixi.toml might look like:
> > ```
> > [dependencies]
> > numpy.git = "https://github.com/numpy/numpy.git"
> > numpy.subdirectory = "pixi-packages/default"
> > numpy.rev = "pixi-build"
> > python = "*"
> > ```
> 
> You might also want to mention that right now you need `preview = ["pixi-build"]` as well (at least pixi 0.61 told me I did)

💯 I was just planning to wait until the SciPy PR is close to finalised, because then we can copy the precise definition straight across.

ngoldbaum: I'm planning to add TSan definitions here and in CPython. I think we can merge this here without that but over in CPython it might be helpful to have that in-place before the PR you have open right now is merged. Would you be OK with me opening a PR against your fork of CPython so both TSan and ASan can be included in the initial version upstream?

lucascolley: > Would you be OK with me opening a PR against your fork of CPython so both TSan and ASan can be included in the initial version upstream?

yes, go for it!

ngoldbaum: Per Felipe over in CPython, I'm going to build off your work for now but will hold off to upstream it until your ASan PR is merged. Thankfully I suspect we'll be able to convince at least one core dev to hit the merge button reasonably quickly.

FFY00: The CPython PR looks like will be ready soon, I'll give it one week-ish to give time for others to object, but if nobody does, I'll merge it :smiley:

lucascolley:
> The CPython PR looks like will be ready soon, I'll give it one week-ish to give time for others to object, but if nobody does, I'll merge it 😃

🎉🎉

Would you have time to address Hugo's comment at https://github.com/python/cpython/pull/142469#discussion_r2606012931 @FFY00 ? I am running out of the ability to justify procrastinating my spare time away on this to myself 😅 .

FFY00:
Done. Could you just confirm the `llvm-tools` thing on macos, since IIRC you use that platform? Other than that, we are good :+1:

lucascolley:
replied at https://github.com/python/cpython/pull/142469#discussion_r2607655926

FFY00:
All good in CPython, I'll probably merge it on the weekend, or start of next week. Congrats @lucascolley!

pixi-packages/asan/pixi.toml
```diff
@@ -0,0 +1,26 @@
...
+[package.build.config]
+extra-input-globs = ["**/*.c.src"]
+compilers = ["c", "cxx"]
+env.ASAN_OPTIONS = "detect_leaks=0:symbolize=1:strict_init_order=true:allocator_may_return_null=1:use_sigaltstack=0"
+extra-args = ["-Csetup-args=-Db_sanitize=address", "-Csetup-args=-Dbuildtype=debug"] 
+
+[package.host-dependencies]
+python.git = "https://github.com/lucascolley/cpython"
+python.subdirectory = "Tools/pixi-packages/asan"
+python.rev = "453e9eebd5e74ab58f3a0c43232e57d5e5e1020c"
```
lucascolley:
```suggestion
python.rev = "9c811e98b9d0ab8f49c1938475ae3dac3aec3120"
```

lucascolley (commented):
TODO here:
- README (can borrow from CPython README)
- CI
- release note

Seems like we are happy punting Windows down the line

ngoldbaum:
> Seems like we are happy punting Windows down the line

Yeah that's fine I've never tried to do this exercise with sanitizers on Windows and am not aware of anyone else doing that

lucascolley:
I added a README and a simple CI job which invokes `pixi build` for each platform and variant.

I also added a release note.

.github/workflows/pixi-packages.yml
```diff
@@ -0,0 +1,39 @@
...
+        fetch-tags: true
+        persist-credentials: false
+
+    - uses: prefix-dev/setup-pixi@ba3bb36eb2066252b2363392b7739741bb777659 # v0.8.1
+      with:
+        pixi-version: v0.60.0
+        run-install: false
+
+    - name: Build
+      run: pixi build --path="pixi-packages/${{ matrix.package_variant }}"
```
lucascolley, line 39: do we want to add anything further to the CI job? Perhaps a sanity check that Python and NumPy at least import?

Not sure how much work that would be, maybe `pixi install` can help avoid too much extra boilerplate?

pixi-packages/asan/pixi.toml
lucascolley: this should be updated to point to `python/cpython` before merge

lucascolley:
```suggestion
python.git = "https://github.com/python/cpython"
python.subdirectory = "Tools/pixi-packages/asan"
python.rev = "8bb5b6e8ce61da41b5affd5eb12d9cc46b5af448"
```

lucascolley:
CI failures look unrelated

crusaderky:
I'm testing asan on my Linux box:
```yaml
[workspace]
channels = ["conda-forge"]
name = "cpython-pixi-test"
platforms = ["linux-64"]
preview = ["pixi-build"]

[dependencies]
python = { git = "https://github.com/python/cpython", subdirectory = "Tools/pixi-packages/asan" }
numpy = { git = "https://github.com/lucascolley/numpy", rev = "pixi-build", subdirectory = "pixi-packages/asan" }
```

It works, but it's extremely verbose, producing several hundreds of pages of output. Unsure if they are legit issues with numpy or just noise? Or am I using it incorrectly?
```bash
$ pixi s
$ python -c 'import numpy'

[... too many lines of  output to copy-paste. here is just one ...]
Indirect leak of 14 byte(s) in 1 object(s) allocated from:
    #0 0x782aed6fd9c7 in malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x782ae833e896 in PyArrayMethod_FromSpec_int ../numpy/_core/src/multiarray/array_method.c:481
    #2 0x782ae8594f9b in PyUFunc_AddLoopFromSpec_int ../numpy/_core/src/umath/dispatching.cpp:177
    #3 0x782ae86592e8 in init_ufunc(_object*, char const*, PyArray_DTypeMeta**, NPY_CASTING (*)(PyArrayMethodObject_tag*, PyArray_DTypeMeta* const*, _PyArray_Descr* const*, _PyArray_Descr**, long*), int (*)(PyArrayMethod_Context_tag*, char* const*, long const*, long const*, NpyAuxData_tag*), int, int, NPY_CASTING, NPY_ARRAYMETHOD_FLAGS, void*) ../numpy/_core/src/umath/stringdtype_ufuncs.cpp:2492

SUMMARY: AddressSanitizer: 35505 byte(s) leaked in 410 allocation(s).
```

.github/workflows/pixi-packages.yml
crusaderky: See my comment - 'python -c 'import numpy'` produces hundreds of pages of output worth of memory leaks, and exits with code 1.

lucascolley:
do you have something like https://github.com/lucascolley/scipy/blob/78445c90e2dc2560d8f6ae044fab0d6d0eaa08ea/pixi.toml#L598

crusaderky:
> do you have something like https://github.com/lucascolley/scipy/blob/78445c90e2dc2560d8f6ae044fab0d6d0eaa08ea/pixi.toml#L598

Ah yes, that quiets it down.

seberg (approved):
FWIW, looks good to me (I have not tried it out, but there are no users, so if there are problems follow-ups are easy enough anyway).

I think I'll just put it in, if e.g. Nathan has another comment it is after all easy to follow-up.

Thanks a lot for all this pixi work Lucas!

seberg:
Hmmm, fun CI failures. The MacOS ones are spurious matmul runtime warnings, the windows ones looked like new warnings generated by cython code.  Retriggered, but just out of curiosity...

seberg:
Sorry, nvm. would need a re-base for the new allowed warnings. And that is just not worth it.

PR #30385, opened 2025-12-06
MAINT, CI: avoid installing mypy in the ppc64le and armhf linux jobs

jorenham:
See e.g.:

- https://github.com/numpy/numpy/actions/runs/19978991810/job/57301923231?pr=30319
- https://github.com/numpy/numpy/actions/runs/19978991809/job/57301578333?pr=30319

Ref: https://github.com/numpy/numpy/pull/30319#issuecomment-3618965566

### DIFF
--- a/requirements/test_requirements.txt
+++ b/requirements/test_requirements.txt
@@ -1,7 +1,6 @@
 Cython
 wheel==0.38.1
-setuptools==65.5.1 ; python_version < '3.12'
-setuptools         ; python_version >= '3.12'
+setuptools
 hypothesis==6.142.2
 pytest==7.4.0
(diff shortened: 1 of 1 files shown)


WarrenWeckesser:
Thanks @jorenham.

The problem is fixed with the release of librt 0.7.2 (see https://github.com/mypyc/librt/issues/20), so this change shouldn't be necessary.

jorenham:
> Thanks @jorenham.
> 
> The problem is fixed with the release of librt 0.7.2 (see [mypyc/librt#20](https://github.com/mypyc/librt/issues/20)), so this change shouldn't be necessary.

Ah great; that's way better. Closing this then.

PR #30386, opened 2025-12-06
ENH: Change `assert_allclose` to accept `numpy.timedelta64` as `atol`

### LINKED ISSUE #30382: BUG: `np.testing.assert_allclose` raises `ValueError` when using `timedelta64` with `atol`
### Describe the issue:

When using `np.testing.assert_allclose` with `timedelta64` arrays and specifying `atol` as a `timedelta64`, a `ValueError` is raised due to an unsupported format code in the error message construction.

https://github.com/numpy/numpy/blob/8f9a6b2ebe6038bb866ca4a0c0cf4f31729cb50f/numpy/testing/_private/utils.py#L1769

According to the source code below, formatter for `np.timedelta64` seems not to be defined.

If it is not intentional, I am happy to open a PR to fix this.

https://github.com/numpy/numpy/blob/8f9a6b2ebe6038bb866ca4a0c0cf4f31729cb50f/numpy/_core/src/multiarray/scalartypes.c.src#L649-L652

This is partially related to #29619

```python
import numpy as np

a = np.array([1, 2], dtype="m8[s]")
b = np.array([3, 4], dtype="m8[s]")

np.testing.assert_allclose(a, b, atol=np.timedelta64(1, "s"))
```

### Error message:

```shell
  File "/home/riku/repos/personals/numpy/workspace/b.py", line 8, in <module>
    np.testing.assert_allclose(a, b, atol=np.timedelta64(1, "s"))
  File "/home/riku/repos/personals/numpy/build-install/usr/lib/python3.12/site-packages/numpy/testing/_private/utils.py", line 1769, in assert_allclose
    header = f'Not equal to tolerance rtol={rtol:g}, atol={atol:g}'

ValueError: Unknown format code 'g' for object of type 'str'
```

2.5.0.dev0+git20251205.8f9a6b2

### Runtime Environment:

_No response_

### Context for the issue:

_No response_

riku-sakamoto:
Closes #30382 

The `:g` float formatting caused a TypeError when `atol` was a `numpy.timedelta64` scalar. This patch falls back to default string formatting for `numpy.timedelta64` scalar.

### DIFF
--- a/numpy/testing/_private/utils.py
+++ b/numpy/testing/_private/utils.py
@@ -1687,7 +1687,7 @@ def assert_allclose(actual, desired, rtol=1e-7, atol=0, equal_nan=True,
         Array desired.
     rtol : float, optional
         Relative tolerance.
-    atol : float, optional
+    atol : float | np.timedelta64, optional
         Absolute tolerance.
     equal_nan : bool, optional.
         If True, NaNs will compare equal.
--- a/numpy/testing/_private/utils.py
+++ b/numpy/testing/_private/utils.py
@@ -1766,7 +1766,11 @@ def assert_allclose(actual, desired, rtol=1e-7, atol=0, equal_nan=True,
                                        equal_nan=equal_nan)
 
     actual, desired = np.asanyarray(actual), np.asanyarray(desired)
-    header = f'Not equal to tolerance rtol={rtol:g}, atol={atol:g}'
+    if isinstance(atol, np.timedelta64):
+        atol_str = str(atol)
+    else:
+        atol_str = f"{atol:g}"
+    header = f'Not equal to tolerance rtol={rtol:g}, atol={atol_str}'
     assert_array_compare(compare, actual, desired, err_msg=str(err_msg),
                          verbose=verbose, header=header, equal_nan=equal_nan,
                          strict=strict)
--- a/numpy/_core/tests/test_datetime.py
+++ b/numpy/_core/tests/test_datetime.py
@@ -2748,6 +2748,17 @@ class TestDateTime:
         results = inputs / divisor
         assert_array_equal(results, expected)
 
+    @pytest.mark.parametrize(
+        "atol", [np.timedelta64(1, "s"), np.timedelta64(1, "ms")]
+    )
+    def test_assert_all_close_with_timedelta_atol(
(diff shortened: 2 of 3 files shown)


jorenham:
The two CI failures are unrelated, see https://github.com/numpy/numpy/pull/30385

numpy/testing/_private/utils.py
jorenham, line 1: From a static typing perspective, `atol` will now accept `float | np.timedelta64`. So unless we're doing this in secret, it would help to update the `testing/_private_utils.pyi` stubs to reflect this.

Similarly, the docstring currently says that `atol` only accepts floats, and should probably also be updated then.

jorenham (commented):
Since the docs did not mention that `atol` supports `np.timedelta64` , I think it would be more appropriate to consider this to be an enhancement, rather than a bugfix. 

I can see how this would be handy to have, and the changes are subtle enough. So this is a +1 for me.

numpy/testing/_private/utils.py
riku-sakamoto: Thank you for the review!

I've updated `_private_utils.pyi` to allow `np.timedelta64` as `atol`.

Just to note, there are two overloads for `assert_allclose`: one for `_ArrayLikeTD64_co` and another for `_NumericArrayLike`.
Since `_NumericArrayLike` does not include arrays with `timedelta64` dtype, I updated only the overload for `_ArrayLikeTD64_co`.

I’ve also updated the docstring accordingly.

jorenham (approved):

riku-sakamoto:
From the discussion in #30385, it looks like the CI issue has been resolved.
Would it be alright if I rebase this branch to re-run the CI?

jorenham:
rebasing sounds good

seberg:
Seems OK to me, I do wonder a bit if there isn't a shorter solution (and ideally `:g` formatting would probably just work).
But these are testing utils, and I don't think it is too bad (nro do I have a quick idea that is better).

Thanks @riku-sakamoto. @jorenham, I think you should feel comfortable merging a PR like this :).

seberg:
Ah, if you care a lot, I guess we could add a `.. versionchanged::` but with this being in testing, I don't think it is particularly important and adds visual clutter.

Issue #29475, opened 2025-07-30
BUG: `np.ma.asarray` does not interrogate a foreign object for `__array__`. Should it?


ikrommyd:
### Describe the issue:

This is how the masked and not-masked `asarray` treats a numpy array
```py
In [1]: import numpy as np

In [2]: array = np.ma.array([1,2,3], mask=[False,True,False])

In [3]: np.asarray(array)
Out[3]: array([1, 2, 3])

In [4]: np.ma.asarray(array)
Out[4]:
masked_array(data=[1, --, 3],
             mask=[False,  True, False],
       fill_value=999999)
```
Now if I wrap my `array` object into a custom array class that implements `__array__`. `np.asarray` is fine with it but `np.ma.asarray` gives `RecursionError: maximum recursion depth exceeded` when you try to look at the array.
```py
In [5]: class Array:
   ...:     def __init__(self, data):
   ...:         self._data = data
   ...:     def __array__(self):
   ...:         return self._data
   ...:

In [6]: x = Array(array)

In [7]: np.asarray(x)
Out[7]: array([1, 2, 3])

In [8]: np.ma.asarray(x)
Out[8]: ---------------------------------------------------------------------------
RecursionError                            Traceback (most recent call last)
```
It appears that looking at the data is problematic
```py
In [9]: out = np.ma.asarray(x)

In [10]: out.mask
Out[10]: array([False,  True, False])

In [11]: out.data
Out[11]: ---------------------------------------------------------------------------
RecursionError                            Traceback (most recent call last)
```
Now of course, if I use the `__array__` method manually, `np.ma.asarray` is fine.
```py
In [12]: np.ma.asarray(x.__array__())
Out[12]:
masked_array(data=[1, --, 3],
             mask=[False,  True, False],
       fill_value=999999)
```
Therefore, my bug report/question is should `np.ma.asarray` look for a `__array__` to convert the object properly?

### Reproduce the code example:

```python
import numpy as np
array = np.ma.array([1,2,3], mask=[False,True,False])
class Array:
    def __init__(self, data):
        self._data = data
    def __array__(self):
        return self._data
x = Array(array)
print(np.asarray(x))
print(np.ma.asarray(x))
```

### Error message:

```shell
Traceback (most recent call last):
  File "<python-input-0>", line 10, in <module>
    print(np.ma.asarray(x))
    ~~~~~^^^^^^^^^^^^^^^^^^
[...]

PR #30388, opened 2025-12-07
BUG: Fix np.ma.asarray() to handle __array__ protocol properly

pavansai018:
### Issue #29475
`np.ma.asarray()` causes `RecursionError` when called on objects implementing `__array__` protocol.

### Reproduction
```python
import numpy as np

array = np.ma.array([1, 2, 3], mask=[False, True, False])

class Array:
    def __init__(self, data):
        self._data = data
    def __array__(self):
        return self._data

x = Array(array)
np.ma.asarray(x)  # RecursionError
```
**Fix:** Added `__array__` protocol check in `np.ma.asarray()` to use returned array data.

**Testing:** Added test cases verifying no recursion and proper handling of array-like objects.

### DIFF
--- a/numpy/ma/core.py
+++ b/numpy/ma/core.py
@@ -8597,6 +8597,17 @@ def asarray(a, dtype=None, order=None):
     <class 'numpy.ma.MaskedArray'>
 
     """
+    # Handle __array__ protocol
+    if hasattr(a, '__array__'):
+        try:
+            # Try to get the array
+            arr = a.__array__()
+            # Check if we got something different
+            if arr is not a:
+                a = arr
+        except Exception:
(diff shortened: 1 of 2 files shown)


mattip:
Did you run `spin test` locally? It seems from the failing doctests that something is broken here.

seberg:
Thanks, but this isn't the right place to do this, and it is also much more complicated than it should be.

I am honestly not sure we should be supporting this, but I think it is OK to support it here, because if we want to reject it, we still can reject it in `np.asarray()`.
But that comes to the point, you should identify exactly where the logic goes wrong, rather than tagging on something (too early).

The tests are also very verbose and should be condensed.

PR #30390, opened 2025-12-07
DOC: Correct wording usage such as a/an

star1327p:
Correct wording usage such as a/an. 

I searched for patterns such as `" a np"` and `" a nd"` in the GitHub codebase, because they are most likely to contain grammatical mistakes. Then I manually reviewed which ones needed to be changed. These patterns are not exhaustive, so a few issues of ("a" vs "an") still remain.

The correct usages are `"a np"` and `"an nd"` , because the phonetic spelling of the word starts with a vowel.
https://www.grammarly.com/blog/grammar/indefinite-articles-a-and-an/

I intentionally did not include `[skip actions][skip azp][skip cirrus]` in the commit, because one change is in the error message.


jakevdp:
Part of this depends on how you read it: e.g. I tend to read "a np.ndarray" as "a numpy-dot-ndarray", so "a" is gramatically the right choice. However, if you read it as "a en-pee-dot-ndarray", then I could see wanting to switch "a" to "an". But the ambiguity suggests that replacing "a" with "an" everywhere is probably not worth worrying about.

doc/source/reference/routines.ma.rst
```diff
@@ -247,7 +247,7 @@ Conversion operations
    ma.masked_where
 
 
-> to a ndarray
+> to an ndarray
 ~~~~~~~~~~~~~~
```
seberg:
```suggestion
~~~~~~~~~~~~~~~
```
FWIW, I guess for most of these its a small improvement. I honestly don't care what we do with `np.`, considering them to be read as "numpy..." would be just as well to me (I think I ).
But maybe we should just stick with the old one here for the sake of fewer changes?

jorenham:
> Part of this depends on how you read it: e.g. I tend to read "a np.ndarray" as "a numpy-dot-ndarray", so "a" is gramatically the right choice. However, if you read it as "a en-pee-dot-ndarray", then I could see wanting to switch "a" to "an". But the ambiguity suggests that replacing "a" with "an" everywhere is probably not worth worrying about.

Personally, I read `ndarray` as "en-dee-array", but `np.*` as "numpy dot". 
So I agree with "an ndarray", but I'd prefer keeping "a `np.spam`" as-is. An alternative would be to rewrite "a `np.spam`" as either "a `numpy.spam`" or just "a `spam`" to avoid ambiguity.

star1327p:
@jakevdp @seberg @jorenham Thanks all of you for the time reviewing this PR. I have kept "an ndarray", and reverted "an np.something" back to "a np.something" because it is pronounced as "a numpy". Hope this PR is finally ready to merge.

jorenham (approved):

jorenham:
Thanks @star1327p

PR #30391, opened 2025-12-08
TYP: ``random.Generator``: many fixes and improvements

jorenham:
This fixes the incompatible overlapping overload issues in the `random._generator` stubs, towards #27032. This additionally improves the `random.Generator` method return types so that they more accurately reflect the types returned at runtime.

The stub file is now also formatted with `ruff format` (if configured with `line-length = 130`), which should help reduce future diffs, and is just nice for DX in general.

I also changes the order of methods around, so that those with equivalent/similar callable signatures are grouped together. I know it's annoying to review, but this way it'll be less likely for equivalent method signatures to diverge if it'll be changed in the future, and makes working on these stubs a bit easier.


numpy/typing/tests/data/reveal/random.pyi
```diff
@@ -349,11 +349,11 @@ assert_type(def_gen.gumbel(0.5, 0.5), float)
 assert_type(def_gen.gumbel(0.5, 0.5, size=None), float)
 assert_type(def_gen.gumbel(0.5, 0.5, size=1), npt.NDArray[np.float64])
 assert_type(def_gen.gumbel(D_arr_0p5, 0.5), npt.NDArray[np.float64])
-assert_type(def_gen.gumbel(0.5, D_arr_0p5), npt.NDArray[np.float64])
+assert_type(def_gen.gumbel(0.5, D_arr_0p5), npt.NDArray[np.float64] | Any)
```
jorenham, line 352: These new `| Any` are needed to avoid those incompatible overlapping overloads. `Any` is by design backwards-compatible (that's the "gradual guarantee"), so it shouldn't be a problem in practice

numpy/typing/tests/data/reveal/random.pyi
```diff
@@ -515,107 +515,59 @@ assert_type(def_gen.integers(2, dtype=bool), bool)
...
-assert_type(def_gen.integers(0, 256, dtype="u1"), np.uint8)
-assert_type(def_gen.integers(255, dtype="u1", endpoint=True), np.uint8)
-assert_type(def_gen.integers(0, 255, dtype="u1", endpoint=True), np.uint8)
-assert_type(def_gen.integers(I_u1_low_like, 255, dtype="u1", endpoint=True), npt.NDArray[np.uint8])
-assert_type(def_gen.integers(I_u1_high_open, dtype="u1"), npt.NDArray[np.uint8])
-assert_type(def_gen.integers(I_u1_low, I_u1_high_open, dtype="u1"), npt.NDArray[np.uint8])
-assert_type(def_gen.integers(0, I_u1_high_open, dtype="u1"), npt.NDArray[np.uint8])
-assert_type(def_gen.integers(I_u1_high_closed, dtype="u1", endpoint=True), npt.NDArray[np.uint8])
-assert_type(def_gen.integers(I_u1_low, I_u1_high_closed, dtype="u1", endpoint=True), npt.NDArray[np.uint8])
-assert_type(def_gen.integers(0, I_u1_high_closed, dtype="u1", endpoint=True), npt.NDArray[np.uint8])
```
jorenham, line 553: I got rid of the "string as dtype" overloads, as they required 20+ overloads, making it infeasible to maintain. Now, passing e.g. `dtype="u1"` will instead return "Any"-ish types, making this a backwards-compatible change (according to the "gradual guarantee", that is).

charris:
Thanks Joren.

PR #30392, opened 2025-12-08
SIMD, BLD: Fix Highway target attribute build failure on ppc64

seiko2plus:
On GCC < 13, compiler target attributes cause a build failure with the `power9/vsx3` feature enabled:
```
from ../numpy/_core/src/umath/loops_trigonometric.dispatch.cpp:5:
hwy/ops/ppc_vsx-inl.h: In function ‘np::simd::{anonymous}::Vec<TLane> np::simd::{anonymous}::LoadN(const TLane*, size_t) [with TLane = float]’:
hwy/ops/ppc_vsx-inl.h:1200:19: error: inlining failed in call to ‘always_inline’ ‘hwy::N_PPC10::VFromD<D>
1200 | HWY_API VFromD<D> LoadN(D d, const T* HWY_RESTRICT p,
```

To fix this issue, a new Highway option `HWY_DISABLE_ATTR` has been introduced in Highway to explicitly disable compiler target attributes, as we rely on a source-based CPU dispatcher. Therefore, this patch updates to the latest Highway main. This patch also extends the native CI ppc64 build to include Clang and and sets testing against Python 3.12.

### DIFF
--- a/meson_cpu/ppc64/meson.build
+++ b/meson_cpu/ppc64/meson.build
@@ -3,15 +3,13 @@ mod_features = import('features')
 compiler_id = meson.get_compiler('c').get_id()
 
 VSX = mod_features.new(
-  'VSX', 1, args: '-mvsx',
+  'VSX', 1, args: ['-mvsx', '-DHWY_COMPILE_ONLY_STATIC', '-DHWY_DISABLE_ATTR'] +
+                  (compiler_id == 'clang' ? ['-maltivec'] : []),
   test_code: files(source_root + '/numpy/_core/src/_simd/checks/cpu_vsx.c')[0],
   extra_tests: {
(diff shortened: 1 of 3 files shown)


meson_cpu/ppc64/meson.build
```diff
@@ -3,15 +3,12 @@ mod_features = import('features')
 compiler_id = meson.get_compiler('c').get_id()
 
 VSX = mod_features.new(
-  'VSX', 1, args: '-mvsx',
+  'VSX', 1, args: ['-mvsx', '-DHWY_COMPILE_ONLY_STATIC', '-DHWY_DISABLE_ATTR'] + (compiler_id == 'clang' ? ['-maltivec'] : []),
```
charris: Is it possible to break this long line? At the "+" looks about right.

seiko2plus: done

charris:
Thanks Sayed.

PR #30393, opened 2025-12-08
BENCH: Add `--cpu-affinity` option to set CPU affinity for running the benchmark

seiko2plus:


mattip:
It is a shame we cannot pass through arguments to asv via `--`

seiko2plus:
> It is a shame we cannot pass through arguments to asv via --

The `nargs` `--` are currently passed to `git`, but I agree it would be better to pass them to `asv` instead; I simply followed the current approach, as the reason for explicitly specifying each `asv` option is unclear.

charris:
Thanks Sayed.

PR #30394, opened 2025-12-08
DOC: prefer passing scalar types to ``dtype=`` in examples

jorenham:
Besides the obvious advantages of consistent docs, passing scalar-types as `dtype=` is beneficial for static typing purposes, because in case of strings or builtin types the dtype can usually not be statically determined.

### DIFF
--- a/numpy/random/mtrand.pyx
+++ b/numpy/random/mtrand.pyx
@@ -719,7 +719,14 @@ cdef class RandomState:
             single value is returned.
         dtype : dtype, optional
             Desired dtype of the result. Byteorder must be native.
-            The default value is intp.
+            The default value is long.
+
+            .. warning::
+              This function defaults to the C-long dtype, which is 32bit on windows
+              and otherwise 64bit on 64bit platforms (and 32bit on 32bit ones).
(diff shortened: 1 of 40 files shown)


numpy/random/mtrand.pyx
```diff
@@ -719,14 +719,7 @@ cdef class RandomState:
...
             Desired dtype of the result. Byteorder must be native.
-            The default value is long.
-
-            .. warning::
-              This function defaults to the C-long dtype, which is 32bit on windows
-              and otherwise 64bit on 64bit platforms (and 32bit on 32bit ones).
-              Since NumPy 2.0, NumPy's default integer is 32bit on 32bit platforms
-              and 64bit on 64bit platforms.  Which corresponds to `np.intp`.
-              (`dtype=int` is not the same as in most NumPy functions.)
+            The default value is intp.
```
charris: I don't think this is correct:
```
        _dtype = np.dtype(dtype) if dtype is not int else np.dtype("long")
```
Note that the examples use RandomState, which should maintain the old implementations.

charris:
Thanks Joren.

PR #30396, opened 2025-12-08
TYP: simplified char code literals, ``ctypeslib``, and ``dtype.__new__``

jorenham:
This simplifies and deduplicates the (private) char code type-aliases and their use-sites, by applying the same C data-type assumptions as `numpy.dtypes`. 
This additionally adds support for complex `ctypes` to the `dtype` constructor (on non-Windows Python 3.14 platforms).
To keep things consistent, easy to reason about, and free of overlapping overloads errors, I also rewrote the `ctypeslib` stubs in terms of fixed-sized data types.

The datetime64 and timedelta64 char-code literals are now split up, based on their associated `.item()` type. I'm planning on using them in a follow-up so that we'll be able to distinguish between e.g. `np.dtype("M8[Y]")` and `np.dtype("M8[ns]")` as `dtype[datetime64[date]]` and `dtype[datetime64[int]]`, respectively.

Oh and the reason for why I pruned the literals, is because `Literal` can have a pretty significant impact on type-checker analysis times (when used in overloads, as we do), especially in case of pyright and mypy.


numpy/__init__.pyi
```diff
@@ -1357,34 +1349,63 @@ class dtype(Generic[_ScalarT_co], metaclass=_DTypeMeta):  # noqa: UP046
...
-    def __new__(
-        cls,
-        dtype: _CLongDoubleCodes,
-        align: builtins.bool = False,
-        copy: builtins.bool = False,
-        *,
-        metadata: dict[builtins.str, Any] = ...,
-    ) -> dtype[clongdouble]: ...
+    # `complexfloating` string-based representations and ctypes
+    if sys.version_info >= (3, 14) and sys.platform != "win32":
```
jorenham, line 1353: this condition matches the one from the typeshed stdlib `ctypes` stubs:
https://github.com/python/typeshed/blob/3e74d5c181ebbf006f32ea5717379edeff907191/stdlib/ctypes/__init__.pyi#L288-L296

numpy/__init__.pyi
```diff
@@ -1325,13 +1317,13 @@ class dtype(Generic[_ScalarT_co], metaclass=_DTypeMeta):  # noqa: UP046
         copy: builtins.bool = False,
         *,
         metadata: dict[builtins.str, Any] = ...,
-    ) -> dtype[long]: ...
+    ) -> dtype[int32 | int64]: ...
```
jorenham, line 1320: `long` is statically indistinguishable from e.g. `np.intp`, so this is more accurate

numpy/_core/src/highway
jorenham, line 1: woops

charris:
Thanks Joren.

PR #30403, opened 2025-12-09
SIMD, BLD: Fix Highway target attribute build failure on ppc64 (#30392)

charris:
Backport of #30392.

On GCC < 13, compiler target attributes cause a build failure with the `power9/vsx3` feature enabled:
```
from ../numpy/_core/src/umath/loops_trigonometric.dispatch.cpp:5:
hwy/ops/ppc_vsx-inl.h: In function ‘np::simd::{anonymous}::Vec<TLane> np::simd::{anonymous}::LoadN(const TLane*, size_t) [with TLane = float]’:
hwy/ops/ppc_vsx-inl.h:1200:19: error: inlining failed in call to ‘always_inline’ ‘hwy::N_PPC10::VFromD<D>
1200 | HWY_API VFromD<D> LoadN(D d, const T* HWY_RESTRICT p,
```

To fix this issue, a new Highway option `HWY_DISABLE_ATTR` has been introduced in Highway to explicitly disable compiler target attributes, as we rely on a source-based CPU dispatcher. Therefore, this patch updates to the latest Highway main. This patch also extends the native CI ppc64 build to include Clang and and sets testing against Python 3.12.

### DIFF
--- a/meson_cpu/ppc64/meson.build
+++ b/meson_cpu/ppc64/meson.build
@@ -3,15 +3,13 @@ mod_features = import('features')
 compiler_id = meson.get_compiler('c').get_id()
 
 VSX = mod_features.new(
-  'VSX', 1, args: '-mvsx',
+  'VSX', 1, args: ['-mvsx', '-DHWY_COMPILE_ONLY_STATIC', '-DHWY_DISABLE_ATTR'] +
+                  (compiler_id == 'clang' ? ['-maltivec'] : []),
   test_code: files(source_root + '/numpy/distutils/checks/cpu_vsx.c')[0],
   extra_tests: {
     'VSX_ASM': files(source_root + '/numpy/distutils/checks/extra_vsx_asm.c')[0]
   }
 )
-if compiler_id == 'clang'
-  VSX.update(args: ['-mvsx', '-maltivec'])
-endif
 VSX2 = mod_features.new(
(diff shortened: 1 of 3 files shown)


charris:
@seiko2plus This is a bit tricky as distutils is gone in main and relevant files moved. I've backported numpy/_core/src/_simd/checks/cpu_vsx.c from main, but that makes me nervous.

meson_cpu/ppc64/meson.build
```diff
@@ -3,15 +3,13 @@ mod_features = import('features')
 compiler_id = meson.get_compiler('c').get_id()
 
 VSX = mod_features.new(
-  'VSX', 1, args: '-mvsx',
-  test_code: files(source_root + '/numpy/distutils/checks/cpu_vsx.c')[0],
+  'VSX', 1, args: ['-mvsx', '-DHWY_COMPILE_ONLY_STATIC', '-DHWY_DISABLE_ATTR'] +
+                  (compiler_id == 'clang' ? ['-maltivec'] : []),
+  test_code: files(source_root + '/numpy/_core/src/_simd/checks/cpu_vsx.c')[0],
```
seiko2plus:
```suggestion
  test_code: files(source_root + '/numpy/distutils/checks/cpu_vsx.c')[0],
```
> I've backported numpy/_core/src/_simd/checks/cpu_vsx.c from main, but that makes me nervous.

@charris, Thank you for backporting it. I think the correct call is simply to use the distutils path.

PR #30404, opened 2025-12-09
BUG: Add missing return status check of NpyIter_EnableExternalLoop().

WarrenWeckesser:
Fix a small bug I just encountered:  calling `enable_external_loop()` on an `nditer` with `multi_index` enabled resulted in a `SystemError`.

Before this fix:

```
In [11]: np.nditer(([[10], [20]], [1, 2]), ['multi_index']).enable_external_loop()
ValueError: Iterator flag EXTERNAL_LOOP cannot be used if an index or multi-index is being tracked

SystemError                               Traceback (most recent call last)
Cell In[11], line 1
----> 1 np.nditer(([[10], [20]], [1, 2]), ['multi_index']).enable_external_loop()

SystemError: <method 'enable_external_loop' of 'numpy.nditer' objects> returned a result with an exception set
```

After:

```
In [1]: np.nditer(([[10], [20]], [1, 2]), ['multi_index']).enable_external_loop()
Cell In[1], line 1
----> 1 np.nditer(([[10], [20]], [1, 2]), ['multi_index']).enable_external_loop()

ValueError: Iterator flag EXTERNAL_LOOP cannot be used if an index or multi-index is being tracked
```

### DIFF
--- a/numpy/_core/src/multiarray/nditer_pywrap.c
+++ b/numpy/_core/src/multiarray/nditer_pywrap.c
@@ -1423,7 +1423,10 @@ npyiter_enable_external_loop(
         return NULL;
     }
 
-    NpyIter_EnableExternalLoop(self->iter);
+    if (NpyIter_EnableExternalLoop(self->iter) != NPY_SUCCEED) {
+        return NULL;
+    }
+
     /* EnableExternalLoop invalidates cached values */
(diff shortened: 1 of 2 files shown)


mattip:
Thanks @WarrenWeckesser

PR #30406, opened 2025-12-09
TST: add edge-case tests for np.negative and np.positive

karansingh012:
This PR adds additional edge-case tests for `np.negative` and `np.positive` in `test_umath.py`.

The tests cover:

- Empty arrays, ensuring shape and dtype are preserved for `np.negative`
- NaN and infinite values, including `-0.0` and `0.0`, with explicit expected results
- Basic integer and floating dtypes (`int8`, `int32`, `float32`, `float64`) and their dtype stability
- 0-D (scalar) vs 1-D arrays for `np.positive`

These tests are intended to strengthen coverage for the unary operators without changing any existing behavior.

Tests:

- `python -m pytest numpy/_core/tests/test_umath.py -q`

### DIFF
--- a/numpy/_core/tests/test_umath.py
+++ b/numpy/_core/tests/test_umath.py
@@ -2977,44 +2977,18 @@ class TestMinMax:
         assert_equal(d.max(), d[0])
         assert_equal(d.min(), d[0])
 
-    def test_negative_positive_edge_cases(self):
-        # Empty array: shape and dtype should be preserved
+    def test_negative_empty(self):
+        """np.negative on an empty array preserves dtype and shape."""
         arr = np.array([], dtype=float)
         res = np.negative(arr)
-        assert_equal(res.shape, arr.shape)
-        assert_equal(res.dtype, arr.dtype)
         assert_array_equal(res, arr)
 
-        # NaN and infinite values
+    def test_negative_nan_inf(self):
+        """np.negative on NaN and infinities behaves as expected."""
         arr = np.array([-np.inf, -1.0, -0.0, 0.0, 1.0, np.inf, np.nan])
         res = np.negative(arr)
-
-        expected_no_nan = np.array([np.inf, 1.0, 0.0, -0.0, -1.0, -np.inf])
-        assert_array_equal(res[:-1], expected_no_nan)
-        # Last element should remain NaN
-        assert np.isnan(arr[-1])
-        assert np.isnan(res[-1])
-
-        # Integer and float dtypes
-        for dtype in (np.int8, np.int32, np.float32, np.float64):
-            arr = np.array([-2, -1, 0, 1, 2], dtype=dtype)
-            res = np.negative(arr)
-            expected = np.array([2, 1, 0, -1, -2], dtype=dtype)
-            assert_equal(res.dtype, dtype)
-            assert_array_equal(res, expected)
-
-        # Positive should be a no-op
-        for dtype in (np.int8, np.int32, np.float32, np.float64):
-            arr = np.array([-2, -1, 0, 1, 2], dtype=dtype)
-            res = np.positive(arr)
-            assert_equal(res.dtype, dtype)
-            assert_array_equal(res, arr)
-
-        # Scalar and 1-D array
-        scalar = np.array(-5)
-        arr = np.array([-5])
-        assert_array_equal(np.positive(scalar), np.array(-5))
-        assert_array_equal(np.positive(arr), np.array([-5]))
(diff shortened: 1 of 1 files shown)


numpy/_core/tests/test_umath.py
```diff
@@ -2977,6 +2977,45 @@ def test_lower_align(self):
         assert_equal(d.max(), d[0])
         assert_equal(d.min(), d[0])
 
+    def test_negative_positive_edge_cases(self):
+        # Empty array: shape and dtype should be preserved
+        arr = np.array([], dtype=float)
+        res = np.negative(arr)
+        assert_equal(res.shape, arr.shape)
```
MaanasArora: Seems redundant to check shape, given `assert_array_equal`? (same in all below)

numpy/_core/tests/test_umath.py
```diff
@@ -2977,6 +2977,45 @@ def test_lower_align(self):
...
+        # Integer and float dtypes
+        for dtype in (np.int8, np.int32, np.float32, np.float64):
+            arr = np.array([-2, -1, 0, 1, 2], dtype=dtype)
+            res = np.negative(arr)
+            expected = np.array([2, 1, 0, -1, -2], dtype=dtype)
+            assert_equal(res.dtype, dtype)
+            assert_array_equal(res, expected)
+
+        # Positive should be a no-op
+        for dtype in (np.int8, np.int32, np.float32, np.float64):
```
MaanasArora: Something very close already exists in `TestPositive.test_(in)valid` in this file.

numpy/_core/tests/test_umath.py
```diff
@@ -2977,6 +2977,45 @@ def test_lower_align(self):
...
+        res = np.negative(arr)
+        assert_equal(res.shape, arr.shape)
+        assert_equal(res.dtype, arr.dtype)
+        assert_array_equal(res, arr)
+
+        # NaN and infinite values
+        arr = np.array([-np.inf, -1.0, -0.0, 0.0, 1.0, np.inf, np.nan])
+        res = np.negative(arr)
+
+        expected_no_nan = np.array([np.inf, 1.0, 0.0, -0.0, -1.0, -np.inf])
```
MaanasArora: Isn't positive zero `array_equal` to negative zero? I don't think this would catch that (not saying it should, just making the point that all this seems quite verbose given the edge case isn't tested).

MaanasArora (commented):
I might be missing context on the motivation, but not sure all these are edge cases, they seem to be testing quite basic functionality. If these are needed, should we make these more focused? (See `TestPositive` and `TestAbsoluteNegative` classes for duplication and if these really need to go there.)

numpy/_core/tests/test_umath.py
karansingh012: Thanks, that makes sense. I’ve dropped the explicit shape checks and now rely on assert_array_equal.

numpy/_core/tests/test_umath.py
karansingh012: Good point. I’ve removed the overlapping dtype/positive/scalar tests and replaced them with two focused tests for np.negative on empty arrays and on NaN/inf.

numpy/_core/tests/test_umath.py
karansingh012: Agreed, the previous version was too verbose without testing a distinct +0.0/-0.0 behaviour. I simplified it to a single np.negative test that checks the NaN/inf (including ±0.0) mapping.

karansingh012:
> I might be missing context on the motivation, but not sure all these are edge cases, they seem to be testing quite basic functionality. If these are needed, should we make these more focused? (See `TestPositive` and `TestAbsoluteNegative` classes for duplication and if these really need to go there.)

Thanks for the review and the pointers!

You’re right that the original version was testing quite a bit of basic behaviour and overlapping with `TestPositive`/`TestAbsoluteNegative`. I’ve simplified the PR now to just two focused tests for `np.negative` (empty array and NaN/±inf, including ±0.0), and dropped the overlapping dtype/positive/scalar cases.

Please let me know if you’d prefer these new tests to live inside one of the existing test classes instead – I’m happy to move them.

mattip:
Sorry, I just don't think this PR has enough added value to justify continuing the review and CI efforts. Closing. If you find an actual issue or unexpected outcome with these very simple ufuncs, please open an issue, and then we can discuss whether it is worth a PR. Opening PRs without a corresponding issue is rarely the right thing to do.

PR #30409, opened 2025-12-10
MAINT: clean up `setuptools` installs and `setup.py`-related leftovers

rgommers:
No longer used for anything after we dropped Python 3.11 support. There are a few example `setup.py` files in the repo, but it turns out we are not testing those (which is fine).

This takes care of one of the tasks in gh-30311.


rgommers:
There's always one last holdout .... I think we can replace `tools/swig/test/setup.py` with a `meson.build` file. 

EDIT: but it's a pain because there is no package, it just builds some extension modules in-place. So putting back one `setuptools` usage in the relevant CI job.

charris:
Thanks Ralf.

PR #30410, opened 2025-12-10
TST: skip deep recursion tests on TSAN and ASAN builds

kumaraditya303:
Relates #30085

Skip deep recursion tests which need large C stack on TSAN and ASAN builds, these tests would otherwise crash under sanitizers. 

cc @rgommers @ngoldbaum


rgommers (approved):
LGTM, thanks @kumaraditya303. The ASan/TSan detection assumes that CPython itself is also built with ASan/TSan and not only NumPy, but I think that's a fair assumption to make here (one gets weird issues without it, and we don't support such numpy-only builds really).

ngoldbaum:
Thanks @kumaraditya303! We should refactor some of the more complicated skip patterns we've grown like this, it's a really nice cleanup.

Issue #29000, opened 2025-05-18
Make `eig/eigvals` always return complex eigenvalues


ev-br:
For a real-valued input, NumPy's eigenvalue routines currently may return either real or complex eigenvalues: both `eig` and `eigvals` check for imaginary parts being exactly zero, and downcast the output if they are [1].

Other array libraries skip this handholding and always return complex eigenvalues. For instance,

In [10]: torch.linalg.eig(torch.arange(4).reshape(2, 2)*1.0).eigenvalues
Out[10]: tensor([-0.5616+0.j,  3.5616+0.j])

A recent array-api discussion [2], had a question if NumPy would consider changing this value-dependent behavior and always return a complex array for eigenvalues?

What is the downstream effect. Were we starting from scratch, I'd expect that doing `real_if_close` is not that taxing for a user---and in fact in the vast majority of use cases it's just not needed.

Since we're not starting from scratch, and to roughly assess the blast radius, I applied the following patch (hidden under the fold), and ran tests for scipy, scikit-learn and scikit-image. 

Here's the summary: 

**scipy**:

SciPy: 4 failures in scipy.signal, all look trivial to fix

FAILED signal/tests/test_dltisys.py::TestStateSpaceDisc::test_properties - AssertionError: dtypes do not match.
FAILED signal/tests/test_dltisys.py::TestTransferFunction::test_properties - AssertionError: dtypes do not match.
FAILED signal/tests/test_ltisys.py::TestStateSpace::test_properties - AssertionError: dtypes do not match.
FAILED signal/tests/test_ltisys.py::TestTransferFunction::test_properties - AssertionError: dtypes do not match.

**scikit-learn**:

all tests pass

**scikit-image**:

FAILED measure/tests/test_fit.py::test_ellipse_model_estimate - TypeError: unsupported operand type(s) for %=: 'numpy.complex128' and 'float'
FAILED measure/tests/test_fit.py::test_ellipse_parameter_stability - TypeError: unsupported operand type(s) for %=: 'numpy.complex128' and 'float'
FAILED measure/tests/test_fit.py::test_ellipse_model_estimate_from_data - TypeError: unsupported operand type(s) for %=: 'numpy.complex128' and 'float'
FAILED measure/tests/test_fit.py::test_ellipse_model_estimate_from_far_shifted_data - TypeError: unsupported operand type(s) for %=: 'numpy.complex128' and 'float'

[...]

PR #30411, opened 2025-12-10
ENH: linalg: return complex arrays from eigvals/eigvecs

ev-br:
A (simplest) companion to https://github.com/numpy/numpy/issues/29000 : change the dtype of `eig/eigvals` returns from _maybe float if starts align, else complex_ to always complex.

As an opinionated pitch from the long discussion in gh-29000:
- in general, eigenvalues of a non-symmetric matrix are complex
- they may or may not have zero imaginary parts
- NumPy goes out of its way to force-cast to reals if the imaginary parts happen to be zero

Hence this PR proposes to stop going out our way, and return what the math says: complex arrays.
_This is a breaking change_

What is the user impact?

Code search shows two kinds of affected usage:

- Users incorrectly assuming that their matrix has real eigenvalues.

This is typically a sign that the user code genuinely misses a case where the eigenvalues are not on a real line.
Example: https://github.com/scikit-image/scikit-image/issues/7013#issuecomment-1583203321
Note that this scikit-image issue is that the scikit-image code assumes real eigenvalues and fails where they are genuinely not.

The downstream fix could be along the lines of adding in the user code

```
if w.imag == 0:
     w = w.real
```

to bring the behavior to what it is today. If wanted, they can add a warning asking a user to report the reproducer, as the scikit-learn issue asks for.

- A long tail of one-off scripts which do `eigvals(covariance_matrix)`.

For a covariance matrix, we know that it's positive definite and the eigenvalues actually are real-values. Hence the fix is to use `eigh` instead.
I think the best we can do is to add a note to this effect.

-------------------------------

An opinionated summary of alternatives discussed in gh-29000:

- Make `eig` emit a warning of impending change.

The warning is extraneous and annoying for those users who are not affected or have already adapted.

- Add a kwarg to control the return type.

I frankly don't think it's very useful to users. Large, well-maintained users can just make the change. The long tail of incorrect usage is unlikely to be able to adapt twice (add the keyword, adapt to the change, wait for a couple of years, remove the keyword)

- Add a new (pair of) function(s) to eventually replace `eig` and `eigvals`.

Feels like a lot of churn across the ecosystem.


ev-br:
Here is a quick session of downstream testing: I ran test suites for SciPy, scikit-learn, scikit-image, networkx and pandas.

Short version: scipy has small issues which I'll take care of; scikit-image has a small issue which I'm volunteering to fix to restore the status quo; scikit-learn and networkx are not affected; pandas is most likely not affected. Testing notes are under the fold.

I'm happy to both run tests for other downstream projects (which ones?), and send downstream patches to restore the status quo after this PR lands (assuming that it is, of course).

<details>

```
Build / install patched numpy
-----------------------------

$ cd repos/numpy
$ python -m build .
$ pip install dist/numpy-2.5.0.dev0-cp312-cp312-linux_x86_64.whl
$ cd ~/temp



scikit-learn
------------

$ python -c'import sklearn; print(sklearn.__file__)'
/home/ev-br/.conda/envs/numpy-dev/lib/python3.12/site-packages/sklearn/__init__.py
$ pytest /home/ev-br/.conda/envs/numpy-dev/lib/python3.12/site-packages/sklearn/ -v

....

================================================= 37605 passed, 3945 skipped, 150 xfailed, 65 xpassed, 5402 warnings in 532.40s (0:08:52) =================================================


scikit-image
------------

$ python -c'import skimage; print(skimage.__file__)'
/home/ev-br/.conda/envs/numpy-dev/lib/python3.12/site-packages/skimage/__init__.py
$ pytest /home/ev-br/.conda/envs/numpy-dev/lib/python3.12/site-packages/skimage/ -v

.....

FAILED ../.conda/envs/numpy-dev/lib/python3.12/site-packages/skimage/io/tests/test_pil.py::test_imsave_filelike - ValueError: Unexpected warning: Saving I mode images as PNG is deprecated and will be removed in Pillow 13 (2026-10-15)
FAILED ../.conda/envs/numpy-dev/lib/python3.12/site-packages/skimage/io/tests/test_pil.py::test_all_mono - ValueError: Unexpected warning: Saving I mode images as PNG is deprecated and will be removed in Pillow 13 (2026-10-15)
FAILED ../.conda/envs/numpy-dev/lib/python3.12/site-packages/skimage/measure/tests/test_fit.py::test_ellipse_model_estimate - TypeError: unsupported operand type(s) for %=: 'numpy.complex128' and 'float'
FAILED ../.conda/envs/numpy-dev/lib/python3.12/site-packages/skimage/measure/tests/test_fit.py::test_ellipse_parameter_stability - TypeError: unsupported operand type(s) for %=: 'numpy.complex128' and 'float'
FAILED ../.conda/envs/numpy-dev/lib/python3.12/site-packages/skimage/measure/tests/test_fit.py::test_ellipse_model_estimate_from_data - TypeError: unsupported operand type(s) for %=: 'numpy.complex128' and 'float'
FAILED ../.conda/envs/numpy-dev/lib/python3.12/site-packages/skimage/measure/tests/test_fit.py::test_ellipse_model_estimate_from_far_shifted_data - TypeError: unsupported operand type(s) for %=: 'numpy.complex128' and 'float'
ERROR ../.conda/envs/numpy-dev/lib/python3.12/site-packages/skimage/io/tests/test_io.py::test_imread_http_url
===================================================== 6 failed, 8404 passed, 130 skipped, 289 warnings, 1 error in 134.77s (0:02:14) =====================================================



networkx
--------

$ python -c'import networkx; print(networkx.__file__)'
/home/ev-br/.conda/envs/numpy-dev/lib/python3.12/site-packages/networkx/__init__.py
$ pytest /home/ev-br/.conda/envs/numpy-dev/lib/python3.12/site-packages/networkx/ -v

......

==================================================================================== warnings summary ====================================================================================
.conda/envs/numpy-dev/lib/python3.12/site-packages/networkx/algorithms/link_analysis/tests/test_hits.py::TestHITS::test_hits_numpy
  /home/ev-br/.conda/envs/numpy-dev/lib/python3.12/site-packages/networkx/algorithms/link_analysis/hits_alg.py:229: ComplexWarning: Casting complex values to real discards the imaginary part
    hubs = dict(zip(G, map(float, h)))

.conda/envs/numpy-dev/lib/python3.12/site-packages/networkx/algorithms/link_analysis/tests/test_hits.py::TestHITS::test_hits_numpy
  /home/ev-br/.conda/envs/numpy-dev/lib/python3.12/site-packages/networkx/algorithms/link_analysis/hits_alg.py:230: ComplexWarning: Casting complex values to real discards the imaginary part
    authorities = dict(zip(G, map(float, a)))

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=========================================================== 6955 passed, 78 skipped, 1 xfailed, 2 warnings in 87.36s (0:01:27) ===========================================================


pandas
------

$ python -c'import pandas; print(pandas.__file__)'
/home/ev-br/.conda/envs/numpy-dev/lib/python3.12/site-packages/pandas/__init__.py
$ pytest /home/ev-br/.conda/envs/numpy-dev/lib/python3.12/site-packages/pandas/ -v

...

============================================================ 619 failed, 178346 passed, 26527 skipped, 1017 xfailed, 85 xpassed, 157 warnings, 783 errors in 1271.67s (0:21:11) =============================================================


But, all failures/erros seem to be IO related (files, html etc)
```

</details>

jorenham:
Coincidentally I was just working on the `eig[vals]` stubs yesterday, and it bothered me that it wasn't possible to accurately describe the return dtype in the stubs, causing them to become rather awkward:

https://github.com/numpy/numpy/blob/a64d310e17f67df3f7ef0383362c944eb8003e17/numpy/linalg/_linalg.pyi#L287-L299

So yea I like this :)

And on that note, it would help if you could also update the stubs accordingly now :)
(You'll have to rebase on main first though)

mattip:
Reformatted the release note

ev-br:
Updated to maintain strict backwards compatibility of root-finding routines. The reasoning is:  `np.roots` et al use of companion matrices is an implementation detail, so it's better to not leak it and keep strict backwards compat. This also limits the scope of the change.

<s>The remaining CI failure in `armhf` looks like an unrelated fluke.</s>And indeed, is gone on flushing the CI.

I tested locally that scipy and scikit-learn are not affected by this change (scikit-learn has some tests that rely on numpy's `complex_array > 0` returning True (?); this can be easily fixed but ISTM scikit-learn _ain't broken don't fix_)

The scikit-image fix is https://github.com/scikit-image/scikit-image/pull/8054

Apologies @mattip --- I rebased/force-pushed without checking your edits for the release note; tried recreating your edits manually. 

Types: @jorenham I've to admit I'm lost in the sea of overloads. What is the difference between `_ArrayLike[np.complex64]` and `_AsArrayC128` and `_ToArrayF64`.

jorenham:
> Types: @jorenham I've to admit I'm lost in the sea of overloads. What is the difference between `_ArrayLike[np.complex64]` and `_AsArrayC128` and `_ToArrayF64`.

unlike `_ArrayLike[...]`, `_AsArrayC128_1d` also accepts `list[complex]` (but not `list[float]` because lists are invariant), and `_ToArrayF64_1d` also accepts `Sequence[float]` (so also `list[int]` and `list[bool]` because of covariance).

jorenham (approved):
the stub changes are ok

(although the type-tests also need to be updated, see the mypy errors)

mattip:
Thanks @ev-br

mattip:
> I'm happy to both run tests for other downstream projects (which ones?), and send downstream patches to restore the status quo after this PR lands (assuming that it is, of course).

Let's keep an eye out for breaking issues

PR #28759, opened 2025-04-18
BLD: Add support for building iOS wheels

freakboy3742:
Adds iOS as a test and release target for NumPy.

**UPDATE: 17 July 2026** This PR is now ready for review. PR body reflects current status. 

In an environment where:
* Xcode and the iOS SDK is installed, with all licences accepted.
* cibuildwheel 4.1.0 is installed
* ninja is available on the path

This branch will build and test accelerated iOS simulator wheels:
```
CIBW_PLATFORM=ios CIBW_ENVIRONMENT="IPHONEOS_DEPLOYMENT_TARGET=17.0" cibuildwheel --arch arm64_iphonesimulator
```
You can also specify `arm64_iphoneos` for device wheels, or `x86_64_iphonesimulator` for simulator wheels on older hardware.

The build will also work without the `IPHONEOS_DEPLOYMENT_TARGET` definition, but with the older LP64 version of accelerate (this is the analog of the `MACOSX_DEPLOYMENT_TARGET=14.0` configuration, except that there isn't a pre-compiled openBLAS for iOS to fall back on). To build with a lower minimum iOS version, run:
```
CIBW_PLATFORM=ios CIBW_BUILD="cp3*" CIBW_CONFIG_SETTINGS="setup-args=-Duse-ilp64=false setup-args=-Dallow-noblas=true build-dir=build" cibuildwheel --archs arm64_iphonesimulator
```

Issue #29724, opened 2025-09-09
BUG: numpy.test on 2.3.2 fails with a bunch of assertion errors


AgilentGCMS:
### Describe the issue:

I just installed numpy 2.3.2 on top of python 3.11.11, compiled from source with intel oneapi 2024.2.1 compilers with
```
CC=icx CXX=icpx CFLAGS='-fveclib=none -fp-model=strict -march=skylake-avx512' FFLAGS='-fp-model=strict -march=skylake-avx512' CXXFLAGS='-fp-model=strict -march=skylake-avx512' python -m pip install --no-build-isolation -v . -Cbuild-dir=build -Csetup-args=-Dallow-noblas=false -Csetup-args=-Dblas-order=mkl -Csetup-args=-Dlapack-order=mkl -Csetup-args=-Dblas=mkl-dynamic-lp64-iomp -Csetup-args=-Dlapack=mkl-dynamic-lp64-iomp
```
Numpy test fails with the errors pasted below. This reminds me of the assertion errors I reported in #27111, which I just let be as recommended by others here. However, I'm not sure if the errors I see now can *also* be just ignored or they point to something more serious.

### Reproduce the code example:

```python
python -c 'import numpy; numpy.test()'
```

### Error message:

```shell
_________________________________________________________________ TestFloatExceptions.test_floating_exceptions[G] __________________________________________________________________

self = <test_numeric.TestFloatExceptions object at 0x1548c579ded0>, typecode = 'G'

    @pytest.mark.skipif(IS_WASM, reason="no wasm fp exception support")
    @pytest.mark.parametrize("typecode", np.typecodes["AllFloat"])
    def test_floating_exceptions(self, typecode):
        if 'bsd' in sys.platform and typecode in 'gG':
            pytest.skip(reason="Fallback impl for (c)longdouble may not raise "
                               "FPE errors as expected on BSD OSes, "
                               "see gh-24876, gh-23379")

        # Test basic arithmetic function errors
        with np.errstate(all='raise'):
            ftype = obj2sctype(typecode)
            if np.dtype(ftype).kind == 'f':
                # Get some extreme values for the type
                fi = np.finfo(ftype)
                ft_tiny = fi._machar.tiny
                ft_max = fi.max
                ft_eps = fi.eps
                underflow = 'underflow'
                divbyzero = 'divide by zero'
            else:
                # 'c', complex, corresponding real dtype
                rtype = type(ftype(0).real)
                fi = np.finfo(rtype)
[...]

PR #24876, opened 2023-10-07
BLD: fix math func feature checks, fix FreeBSD build, add CI job

rgommers:
Should fix the build on FreeBSD and other OSes that are not C99-compliant.

Closes gh-24873

PR #30412, opened 2025-12-10
BLD: Add Android support

### LINKED ISSUE #27698: Add Android Support

mhsmith:
Based on previous work:
* #28759
* #29770

I'm currently targeting the 2.3.x branch as part of a project I'm working on, but I'll rebase this PR on main once it's complete.

* Requires https://github.com/pypa/cibuildwheel/pull/2695
* Requires https://github.com/mesonbuild/meson-python/pull/824
* Requires https://github.com/chaquo/chaquopy/pull/1457
* Closes #27698


mhsmith:
The 2.3.2-1 wheels [here](https://chaquo.com/pypi-test/numpy/) were built using the following procedure:

* Use a Linux or Mac machine.
* Check out this PR.
* `cd` to the top-level directory of this repository.
* `git submodule update --init`
* Create and activate a Python 3.13 virtual environment.
* Run `pip install git+https://github.com/mhsmith/cibuildwheel@android-numpy`
* Run `cibuildwheel --only cp313-android_ARCH`, where `ARCH` is either `arm64_v8a` or `x86_64`.
* cibuildwheel will attempt to run the test suite on an Android emulator. If you don't want that, set the environment variable `CIBW_TEST_SKIP="*"`.

These wheels pass the entire test suite except for:

* A number of tests involving floating point exceptions which are expected to occur but don't. The discussion in [#29724](https://github.com/numpy/numpy/issues/29724) indicates this may be caused by slight differences between compilers.
* 3 tests called `test_legacy_pickle`, due to bugs in how the testbed handles .gz files. I have a fix for this in [python/cpython#142912](https://github.com/python/cpython/pull/142912), which I'll eventually include in the cibuildwheel PR.

On Chaquopy 17, [this workaround](https://github.com/chaquo/chaquopy/issues/1433#issuecomment-3736035239) is required.

rgommers (commented):
Thanks @mhsmith! It's fine to keep this PR open for reference if that's useful to you, however note that we're unlikely to do another 2.3.x release and even if we did, we'd want the changes to land in `main` first. 

The test suite changes all seem fine, I'd be happy to merge those in a separate PR to `main` (possibly together with the related changes for iOS, to avoid the duplication). 

The hardcoding of `longdouble` for Android (and iOS) also seems okay and fine to merge separately. 

The wheel build changes I'd prefer to not merge, at least until everything upstream is ready and then we can reassess.

mhsmith:
Thanks: I intend to create a set of similar PRs for numpy, pandas, scipy, and scikit-learn, so I can produce wheels of known versions for some Chaquopy users I'm working with. Once these PRs are fully working, I'll retarget them to the main branch of each project to enable them to release Android wheels on PyPI.

pyproject.toml
```diff
@@ -1,13 +1,15 @@
 [build-system]
 build-backend = "mesonpy"
 requires = [
-    "meson-python>=0.15.0",
+    # TODO: use an official meson-python version once
+    # https://github.com/mesonbuild/meson-python/pull/824 has been released.
+    "meson-python @ git+https://github.com/mesonbuild/meson-python@main",
```
rgommers: This change can be droped now

.github/workflows/macos.yml
```diff
@@ -22,7 +22,7 @@ jobs:
     name: macOS x86-64 conda
     # To enable this workflow on a fork, comment out:
     if: github.repository == 'numpy/numpy'
-    runs-on: macos-13
+    runs-on: macos-15-intel
```
rgommers, line 29: These changes are already in `main`

pyproject.toml
mhsmith: Done.

.github/workflows/macos.yml
mhsmith: This PR has now been rebased onto the main branch.

rgommers:
I'd still be happy to merge all the test changes separately. 

In case you haven't seen it yet, we now have [NEP 57 — NumPy platform support](https://numpy.org/neps/nep-0057-numpy-platform-support.html), defining CPython-like support tiers. It may be reasonable for this PR to aim to promote Android to Tier 3 by adding a CI job, similar to the current Emscripten one.

mhsmith:
> A number of tests involving floating point exceptions which are expected to occur but don't. 

These tests can be divided into two groups:

* Those that only fail for the `longdouble` type. This was previously seen on FreeBSD ([#24876](https://github.com/numpy/numpy/pull/24876)).
* Those involving division by zero, that fail for all floating point types. This was previously seen on macOS ([#19316](https://github.com/numpy/numpy/pull/19316)).

I've marked all of these tests as `xfail` on Android.

mhsmith:
This PR is now ready for review. Since the only supported way of building for Android is via cibuildwheel, I think it's enough to add a matrix entry in wheels.yml, rather than adding a whole separate workflow file.

The iOS PR #28759 is also complete, though it's currently blocked on https://github.com/numpy/meson/pull/25.

> In case you haven't seen it yet, we now have [NEP 57 — NumPy platform support](https://numpy.org/neps/nep-0057-numpy-platform-support.html), defining CPython-like support tiers. It may be reasonable for this PR to aim to promote Android to Tier 3 by adding a CI job, similar to the current Emscripten one.

Yes, I'd like to do that. As I understand it, the procedure would be to merge this PR to enable the CI job, and then submit a PR against NEP 57?

rgommers:
Working on integrating the iOS fix now. 

> As I understand it, the procedure would be to merge this PR to enable the CI job, and then submit a PR against NEP 57?

It'd be fine to do it in this same PR as well, might be even better since the merging of this PR is effectively the Tier 3 promotion.

> I think it's enough to add a matrix entry in wheels.yml, rather than adding a whole separate workflow file.

Agreed.

.github/workflows/wheels.yml
```diff
@@ -43,7 +43,11 @@ jobs:
           - [macos-14, macosx_arm64, openblas]
           - [windows-2022, win_amd64, ""]
           - [windows-11-arm, win_arm64, ""]
-        python: ["cp312"]
+          # GitHub Actions doesn’t currently support the Android emulator on any ARM
+          # runner. So we build for ARM on a non-ARM runner, which will skip the tests.
+          - [ubuntu-24.04, android_arm64_v8a]
+          - [ubuntu-24.04, android_x86_64]
+        python: ["cp313"]
```
rgommers, line 50: I thought about it for a bit; this is okay - it's usually set to the lowest supported Python version, but that's also covered by other jobs, so given that Android requires >=3.13, this is the easiest way to achieve that.

tools/wheels/cibw_before_build.sh
```diff
@@ -32,18 +34,26 @@ if [[ "$INSTALL_OPENBLAS" = "true" ]] ; then
...
-    # Copy scipy-openblas DLL's to a fixed location so we can point delvewheel
-    # at it in `repair_windows.sh` (needed only on Windows because of the lack
-    # of RPATH support).
-    if [[ $RUNNER_OS == "Windows" ]]; then
-        python <<EOF
+
+    if [[ $OPENBLAS == "android" ]]; then
+        python -m pip install \
+            --target $pkgconf_path \
+            --index-url https://chaquo.com/pypi-upstream/ \
```
rgommers, line 41: This is fine for this file; we wouldn't like it in `numpy/numpy-release`, but that has its own `cibw_before_build.sh` which doesn't have to be in sync with this one.

pyproject.toml
```diff
@@ -222,17 +231,32 @@ select = ["*-win32"]
...
-repair-wheel-command = ""
+[[tool.cibuildwheel.overrides]]
+select = ["*-android*", "*-ios*", "*-pyodide*"]
+before-test = ""
+
+# Any packages which appear both here and in test_requirements.txt should be kept in
+# sync.
+test-requires = [
+    "hypothesis==6.152.1",
+    "pytest==9.1.0",
```
rgommers: This has gone out of sync:
```suggestion
    "hypothesis==6.155.7",
    "pytest==9.1.1",
```

We'll change this over to `[dependency-groups]` in the near future anyway I think, and then we can unify these core test deps in a separate group.

rgommers (approved):
LGTM, thanks @mhsmith for getting this to the finish line! I'll fix the last minor thing and merge conflict now to see if CI is still happy.

rgommers:
Wheel build job times of 2 minutes for `android-arm64_v8a` (no tests run, hence very fast) and 11 minutes for `android-x86_64` are not a concern, easily fast enough. So we're all set here I think. Let's hit the green button here and add the Tier 3 entry in a follow-up PR.

Thanks again @mhsmith!

PR #30414, opened 2025-12-10
MAINT: Simplify array setstate by using general deallocation code

mhvk:
In preparation for (perhaps) having the array not necessarily allocate separate memory for sizes+strides and data, this simplifies one of the places where `ndarray` instance hacking takes place, viz., `__setstate__`. Here, one has to replace all bits from initialization of a new instance by pickle with the data actually stored in the pickle. Since that replacement requires what most of what deallocation does, with the exception of deallocating `self` itself, this PR splits off that bit in `arrayobject` and uses it in `methods`.

As part of the above, I made a few other simplifications in `__setstate__` (first commit).

### DIFF
--- a/numpy/_core/src/multiarray/arrayobject.c
+++ b/numpy/_core/src/multiarray/arrayobject.c
@@ -380,25 +380,37 @@ WARN_IN_DEALLOC(PyObject* warning, const char * msg) {
 
 /* array object functions */
 
-static void
-array_dealloc(PyArrayObject *self)
+/*
+ * Much of the actual work for dealloc, split off for use in __setstate__
+ * via clear_array_attributes function defined below.
+ * If not unraisable, will return -1 on error, 0 on success.
+ * If unraisable, always succeeds, though may print errors and warnings.
+ */
+static int
+_clear_array_attributes(PyArrayObject *self, npy_bool unraisable)
 {
     PyArrayObject_fields *fa = (PyArrayObject_fields *)self;
 
     if (_buffer_info_free(fa->_buffer_info, (PyObject *)self) < 0) {
+        if (!unraisable) {
+            return -1;
+        }
         PyErr_WriteUnraisable(NULL);
     }
+    fa->_buffer_info = NULL;
 
-    if (fa->weakreflist != NULL) {
-        PyObject_ClearWeakRefs((PyObject *)self);
-    }
     if (fa->base) {
         int retval;
         if (PyArray_FLAGS(self) & NPY_ARRAY_WRITEBACKIFCOPY)
         {
-            char const * msg = "WRITEBACKIFCOPY detected in array_dealloc. "
+            char const * msg = "WRITEBACKIFCOPY detected in clearing of array. "
                 " Required call to PyArray_ResolveWritebackIfCopy or "
                 "PyArray_DiscardWritebackIfCopy is missing.";
+            /* For non-dealloc case, we simply error */
+            if (!unraisable) {
+                PyErr_SetString(PyExc_RuntimeError, msg);
+                return -1;
+            }
             /*
              * prevent reaching 0 twice and thus recursing into dealloc.
              * Increasing sys.gettotalrefcount, but path should not be taken.
@@ -408,25 +420,34 @@ array_dealloc(PyArrayObject *self)
             retval = PyArray_ResolveWritebackIfCopy(self);
             if (retval < 0)
             {
-                PyErr_Print();
-                PyErr_Clear();
+                PyErr_WriteUnraisable(NULL);
             }
         }
         /*
          * If fa->base is non-NULL, it is something
          * to DECREF -- either a view or a buffer object
          */
-        Py_XDECREF(fa->base);
+        Py_CLEAR(fa->base);
     }
 
     if ((fa->flags & NPY_ARRAY_OWNDATA) && fa->data) {
         /* Free any internal references */
         if (PyDataType_REFCHK(fa->descr)) {
             if (PyArray_ClearArray(self) < 0) {
+                if (!unraisable) {
+                    return -1;
+                }
                 PyErr_WriteUnraisable(NULL);
             }
         }
+        /* mem_handler can be absent if NPY_ARRAY_OWNDATA arbitrarily set */
         if (fa->mem_handler == NULL) {
+            /* For non-dealloc case, we simply error */
+            if (!unraisable) {
+                PyErr_SetString(PyExc_RuntimeError,
+                                "no memory handler found but OWNDATA flag set");
+                return -1;
+            }
             if (npy_global_state.warn_if_no_mem_policy) {
                 char const *msg = "Trying to dealloc data, but a memory policy "
                     "is not set. If you take ownership of the data, you must "
@@ -442,14 +463,36 @@ array_dealloc(PyArrayObject *self)
                 nbytes = 1;
             }
             PyDataMem_UserFREE(fa->data, nbytes, fa->mem_handler);
-            Py_DECREF(fa->mem_handler);
+            Py_CLEAR(fa->mem_handler);
         }
+        fa->data = NULL;
     }
 
     /* must match allocation in PyArray_NewFromDescr */
     npy_free_cache_dim(fa->dimensions, 2 * fa->nd);
-    Py_DECREF(fa->descr);
+    fa->dimensions = NULL;
+    Py_CLEAR(fa->descr);
+    return 0;
+}
+
+static void
+array_dealloc(PyArrayObject *self)
+{
+    // NPY_TRUE flags that errors are unraisable.
+    int ret = _clear_array_attributes(self, NPY_TRUE);
+    assert(ret == 0);  // should always succeed if unraisable.
+    // Only done on actual deallocation, nothing allocated by numpy.
+    if (((PyArrayObject_fields *)self)->weakreflist != NULL) {
+        PyObject_ClearWeakRefs((PyObject *)self);
+    }
     Py_TYPE(self)->tp_free((PyObject *)self);
+}
+
+NPY_NO_EXPORT int
+clear_array_attributes(PyArrayObject *self)
+{
+    // NPY_FALSE flags that errors can be raised.
+    return _clear_array_attributes(self, NPY_FALSE);
 }
 
 /*NUMPY_API
--- a/numpy/_core/src/multiarray/arrayobject.c
+++ b/numpy/_core/src/multiarray/arrayobject.c
@@ -442,12 +442,6 @@ _clear_array_attributes(PyArrayObject *self, npy_bool unraisable)
         }
         /* mem_handler can be absent if NPY_ARRAY_OWNDATA arbitrarily set */
         if (fa->mem_handler == NULL) {
-            /* For non-dealloc case, we simply error */
-            if (!unraisable) {
-                PyErr_SetString(PyExc_RuntimeError,
-                                "no memory handler found but OWNDATA flag set");
-                return -1;
-            }
             if (npy_global_state.warn_if_no_mem_policy) {
                 char const *msg = "Trying to dealloc data, but a memory policy "
                     "is not set. If you take ownership of the data, you must "
--- a/numpy/_core/src/multiarray/arrayobject.c
+++ b/numpy/_core/src/multiarray/arrayobject.c
@@ -358,24 +358,19 @@ PyArray_ResolveWritebackIfCopy(PyArrayObject * self)
 /*********************** end C-API functions **********************/
 
 
-/* dealloc must not raise an error, best effort try to write
-   to stderr and clear the error
-*/
-
-static inline void
-WARN_IN_DEALLOC(PyObject* warning, const char * msg) {
-    if (PyErr_WarnEx(warning, msg, 1) < 0) {
-        PyObject * s;
-
-        s = PyUnicode_FromString("array_dealloc");
-        if (s) {
-            PyErr_WriteUnraisable(s);
-            Py_DECREF(s);
-        }
-        else {
-            PyErr_WriteUnraisable(Py_None);
-        }
+/*
+ * During dealloc we cannot propagate errors so if unraisable is set
+ * we simply print out the error message and convert the error into
+ * success (returning 0).
+ */
+static inline int
+write_and_clear_error_if_unraisable(int status, npy_bool unraisable)
+{
+    if (status < 0 && unraisable) {
+        PyErr_WriteUnraisable(npy_interned_str.array_dealloc);
+        return 0;
     }
+    return status;
 }
 
 /* array object functions */
@@ -392,23 +387,20 @@ _clear_array_attributes(PyArrayObject *self, npy_bool unraisable)
     PyArrayObject_fields *fa = (PyArrayObject_fields *)self;
 
     if (_buffer_info_free(fa->_buffer_info, (PyObject *)self) < 0) {
-        if (!unraisable) {
+        if (write_and_clear_error_if_unraisable(-1, unraisable) < 0) {
             return -1;
         }
-        PyErr_WriteUnraisable(NULL);
     }
     fa->_buffer_info = NULL;
 
     if (fa->base) {
-        int retval;
         if (PyArray_FLAGS(self) & NPY_ARRAY_WRITEBACKIFCOPY)
         {
             char const * msg = "WRITEBACKIFCOPY detected in clearing of array. "
                 " Required call to PyArray_ResolveWritebackIfCopy or "
                 "PyArray_DiscardWritebackIfCopy is missing.";
-            /* For non-dealloc case, we simply error */
-            if (!unraisable) {
-                PyErr_SetString(PyExc_RuntimeError, msg);
+            int retval = PyErr_WarnEx(PyExc_RuntimeWarning, msg, 1);
+            if (write_and_clear_error_if_unraisable(retval, unraisable) < 0) {
                 return -1;
             }
             /*
@@ -416,11 +408,9 @@ _clear_array_attributes(PyArrayObject *self, npy_bool unraisable)
              * Increasing sys.gettotalrefcount, but path should not be taken.
              */
             Py_INCREF(self);
-            WARN_IN_DEALLOC(PyExc_RuntimeWarning, msg);
             retval = PyArray_ResolveWritebackIfCopy(self);
-            if (retval < 0)
-            {
-                PyErr_WriteUnraisable(NULL);
+            if (write_and_clear_error_if_unraisable(retval, unraisable) < 0) {
+                return -1;
             }
         }
         /*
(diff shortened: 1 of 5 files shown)


numpy/_core/src/multiarray/arrayobject.c
```diff
@@ -358,56 +358,61 @@ PyArray_ResolveWritebackIfCopy(PyArrayObject * self)
...
+static int
+_dealloc_all_but_self(PyArrayObject *self, npy_bool unraisable)
 {
     PyArrayObject_fields *fa = (PyArrayObject_fields *)self;
+    int retval;
 
-    if (_buffer_info_free(fa->_buffer_info, (PyObject *)self) < 0) {
-        PyErr_WriteUnraisable(NULL);
+    if (fa->_buffer_info) {
+        if (_buffer_info_free(fa->_buffer_info, (PyObject *)self) < 0) {
```
mhvk: This part shows both the advantage and disadvantage of combining the deallocation from two different parts:
- advantage: `_buffer_info_free` was not called in the `__setstate__` path. Perhaps it can never be set? Or just introduced later? Better safe than sorry!
- disadvantage: bit spaghetti-codey to signal an error return only on one path. Also, we now have to make sure we clear any member that we deallocated (the assignment to NULL just below).

numpy/_core/src/multiarray/arrayobject.c
```diff
@@ -358,56 +358,61 @@ PyArray_ResolveWritebackIfCopy(PyArrayObject * self)
...
+            if (!unraisable) {
+                return -1;
+            }
+            PyErr_WriteUnraisable(NULL);
+        }
+        fa->_buffer_info = NULL;
     }
 
-    if (fa->weakreflist != NULL) {
```
mhvk, line 392: This should only be done for real `dealloc` (there is a guard against calling it when ref count != 0 in the code).

numpy/_core/src/multiarray/arrayobject.c
```diff
@@ -358,56 +358,61 @@ PyArray_ResolveWritebackIfCopy(PyArrayObject * self)
...
-    }
     if (fa->base) {
-        int retval;
         if (PyArray_FLAGS(self) & NPY_ARRAY_WRITEBACKIFCOPY)
         {
             char const * msg = "WRITEBACKIFCOPY detected in array_dealloc. "
                 " Required call to PyArray_ResolveWritebackIfCopy or "
                 "PyArray_DiscardWritebackIfCopy is missing.";
-            /*
-             * prevent reaching 0 twice and thus recursing into dealloc.
```
mhvk, line 403: I don't understand at all why this was there -- if there is a piece of code that tries to defref self again, it is a bug, and should be exposed. Given that tests pass without it, it doesn't seem to be necessary (any more?).

numpy/_core/src/multiarray/arrayobject.c
```diff
@@ -416,22 +421,32 @@ array_dealloc(PyArrayObject *self)
          * If fa->base is non-NULL, it is something
          * to DECREF -- either a view or a buffer object
          */
-        Py_XDECREF(fa->base);
+        Py_CLEAR(fa->base);
     }
-
-    if ((fa->flags & NPY_ARRAY_OWNDATA) && fa->data) {
+    else if ((fa->flags & NPY_ARRAY_OWNDATA) && fa->data) {
```
mhvk: `else if` since one cannot own data and have a base too.

numpy/_core/src/multiarray/arrayobject.c
```diff
@@ -416,22 +421,32 @@ array_dealloc(PyArrayObject *self)
...
+            }
             if (npy_thread_unsafe_state.warn_if_no_mem_policy) {
                 char const *msg = "Trying to dealloc data, but a memory policy "
                     "is not set. If you take ownership of the data, you must "
                     "set a base owning the data (e.g. a PyCapsule).";
-                WARN_IN_DEALLOC(PyExc_RuntimeWarning, msg);
+                retval = WARN_IN_DEALLOC(PyExc_RuntimeWarning, msg, unraisable);
+                assert(retval == 0);  // must be true if unraisable.
             }
             // Guess at malloc/free ???
```
mhvk, line 443: Not directly relevant here, but this seems pretty insane... I guess for deallocation one has little choice...

numpy/_core/src/multiarray/methods.c
```diff
@@ -2072,17 +2072,6 @@ array_setstate(PyArrayObject *self, PyObject *args)
         return NULL;
     }
 
-    /*
-     * Reassigning fa->descr messes with the reallocation strategy,
```
mhvk, line 2077: Yay, this all becomes irrelevant!

numpy/_core/src/multiarray/methods.c
```diff
@@ -2096,31 +2085,19 @@ array_setstate(PyArrayObject *self, PyObject *args)
...
-    npy_bool empty = NPY_FALSE;
-    nbytes = 1;
+    nbytes = typecode->elsize;
     for (int i = 0; i < nd; i++) {
         if (dimensions[i] < 0) {
             PyErr_SetString(PyExc_TypeError,
                     "impossible dimension while unpickling array");
             return NULL;
         }
-        if (dimensions[i] == 0) {
```
mhvk, line 2108: Unlike in `ctors`, here the code never replaced 0 by 1, so there is no need to separately track whether the result is empty: multiplication by 0 will just ensure that.

numpy/_core/src/multiarray/methods.c
```diff
@@ -2134,11 +2111,8 @@ array_setstate(PyArrayObject *self, PyObject *args)
 
         /* Backward compatibility with Python 2 NumPy pickles */
         if (PyUnicode_Check(rawdata)) {
-            PyObject *tmp;
-            tmp = PyUnicode_AsLatin1String(rawdata);
-            Py_DECREF(rawdata);
-            rawdata = tmp;
-            if (tmp == NULL) {
+            Py_SETREF(rawdata, PyUnicode_AsLatin1String(rawdata));
```
mhvk, line 2115: Small simplification

numpy/_core/src/multiarray/methods.c
```diff
@@ -2166,32 +2140,13 @@ array_setstate(PyArrayObject *self, PyObject *args)
             return NULL;
         }
     }
-
-    if ((PyArray_FLAGS(self) & NPY_ARRAY_OWNDATA)) {
```
mhvk, line 2171: The bigger cleanup. Note that this misses cleaning up `_buffer_info`, or the case where there is a `base`. Furthermore, it just clears `WRITEBACKIFCOPY`, while if that is set, presumably something should happen. (It may well be that none of those can happen, but better safe than sorry...)

numpy/_core/src/multiarray/methods.c
```diff
@@ -2221,11 +2176,8 @@ array_setstate(PyArrayObject *self, PyObject *args)
             if (num == 0) {
                 num = 1;
             }
-            /* Store the handler in case the default is modified */
```
mhvk, line 2225: Nice to avoid this strange dance with `mem_handler`.

mhvk (commented):
OK, tests pass now on this refactoring to bring deallocation attempts together. I added some comments in-line to help review.

numpy/_core/src/multiarray/arrayobject.c
seberg: > Perhaps it can never be set? Or just introduced later?

Both are true, since `__setstate__` only makes sense on a fresh array so a buffer cannot have been exported for it (I honestly don't think `__setstate__` is a good choice for our types, but that is a different issue).

numpy/_core/src/multiarray/arrayobject.c
seberg: It is probably never needed, but please don't remove it. If we remove it, I would rather just make the whole pass a full error, since the warning has been around forever (yes, it's not a raised error unfortunately, as it it will be printed).

The problem this protects against is if `PyArray_ResolveWritebackIfCopy` calls `Py_INCREC(self)` for any reason at all, things would blow up.
And while it may be that there is no path that will do an incref, it seems opaque enough to follow that leaking a global reference here is immaterial.
(Yeah, I am basically saying that I don't want to audit/promise that `PyArray_ResolveWritebackIfCopy` won't cause an `INCREF`.)

numpy/_core/src/multiarray/arrayobject.c
seberg: I don't think that is correct the update-if-copy path above has both set.

numpy/_core/src/multiarray/arrayobject.c
seberg: Not all _that_ insane, it allows downstream C extensions to create NumPy arrays in C that own their data but use an existing allocation.

It might be nice if those created a little capsule that owned the data instead, but that is a pretty big ABI break and honestly, I don't think this path is _that_ bad.

numpy/_core/src/multiarray/arrayobject.c
seberg: Can you re-factor this and remove the unraisable propagation?  This function has no reason to handle it, it can just propagate errors normally (it has be careful to be OK with an in-flight error, but it is, I think).

You can handle the `PyErr_WriteUnraisable` inside `dealloc` where it is obvious.
We could put the `array_dealloc` string into the static string data to avoid that error handling/NULL path, but it is also OK as it is (avoiding the `NULL` was there for older PyPy versions which is irrelevant.)

numpy/_core/src/multiarray/arrayobject.c
```diff
@@ -358,56 +358,61 @@ PyArray_ResolveWritebackIfCopy(PyArrayObject * self)
...
-static void
-array_dealloc(PyArrayObject *self)
+/*
+ * Much of the actual work for dealloc, split off for use in __setstate__
+ * via dealloc_all_but_self function defined below.
+ * If not unraisable, will return -1 on error, 0 on success.
+ * If unraisable, always succeeds.
+ */
+static int
+_dealloc_all_but_self(PyArrayObject *self, npy_bool unraisable)
```
seberg: I wonder if can find a nicer name "clear" might make sense. But we only use it twice, so doesn't matter much.

(`clear` is also a bit clearer on the fact that I think it now has to be setup to be OK to call twice due to error handling in `__setitem__`.)

numpy/_core/src/multiarray/arrayobject.c
```diff
@@ -442,16 +457,40 @@ array_dealloc(PyArrayObject *self)
...
-            Py_DECREF(fa->mem_handler);
+            Py_CLEAR(fa->mem_handler);
         }
+        fa->data = NULL;
     }
 
     /* must match allocation in PyArray_NewFromDescr */
-    npy_free_cache_dim(fa->dimensions, 2 * fa->nd);
-    Py_DECREF(fa->descr);
+    if (fa->dimensions != NULL) {
```
seberg: doesn't `npy_free_cache_dims` include the `!= NULL` check like typical `free` functions?

numpy/_core/src/multiarray/methods.c
```diff
@@ -2096,31 +2085,19 @@ array_setstate(PyArrayObject *self, PyObject *args)
...
-            empty = NPY_TRUE;
-        }
         overflowed = npy_mul_sizes_with_overflow(
                 &nbytes, nbytes, dimensions[i]);
         if (overflowed) {
             return PyErr_NoMemory();
         }
     }
-    overflowed = npy_mul_sizes_with_overflow(
-            &nbytes, nbytes, PyArray_ITEMSIZE(self));
```
seberg, line 2118: Hmmmm, the order here may be there intentionally for when `itemsize` is zero, we still need to limit the number of elements in the array to not overflow.

(For that path, even the "replace with 1" logic might be strictly speaking needed; iterations over a subset of axes shouldn't be able to cause overflows. All of this should be possible when unpickling an array created on a 64bit system on a 32bit one.)

numpy/_core/src/multiarray/methods.c
seberg: Yeah, none of this is remotely possible. If you call `__setstate__` on anything but a freshly created array during pickling, you are in trouble anyway.

numpy/_core/src/multiarray/methods.c
```diff
@@ -2166,32 +2140,13 @@ array_setstate(PyArrayObject *self, PyObject *args)
...
-    PyArray_CLEARFLAGS(self, NPY_ARRAY_WRITEBACKIFCOPY);
-
-    if (PyArray_DIMS(self) != NULL) {
-        npy_free_cache_dim_array(self);
-        fa->dimensions = NULL;
+    /*
+     * Get rid of everything on self, and then populate with pickle data.
+     */
+    if (dealloc_all_but_self(self) < 0) {
+        return NULL;
```
seberg, line 2148: OK, we assume that the only place where `__setitem__` is called is unpickling, which means that the _only_ thing that can happen to `self` on error return is that it is deleted (at this point, the array is completely broken!).

Of course, this can't actually fail in that context, because all of the paths that _might_ fail are actually unreachable.

seberg (commented):
A couple of comments, although by the time I reached the end of this, I am wondering if we can't just tighten error checks rather than taking into account things like a mutated data handler?

To be clear, I am not sure that this will simplify things quite enough to be worthwhile, but as such I like the idea of `__setstate__` being restrictive.

The fact is, we just created that array with `_reconstruct`, unfortunately, that might have caused a call to `__array_finalize__` which can do some crazy things (like reshaping), I don't think it can do many crazy things, such as:
* The data handler must be the default.
* The `OWNDATA` flag cannot have been unset.

I suppose technically, it could in theory do buffer exports, but it cannot have writebackifcopy (in a reasonable way).
Although, I would be willing to gamble on the buffer export thing and help that one person who it might hit to deal with it.

mattip:
Where are we in review of this PR? I see some comments that are not closed, and there is a merge conflict.

numpy/_core/src/multiarray/arrayobject.c
mhvk: I've gone for `clear_array_attributes` - better?

numpy/_core/src/multiarray/arrayobject.c
mhvk: I'm a little confused: as is, the behaviour is different depending on whether `unraisable` is set: if it is, the function keeps going, trying to deallocate everything, while if not, it bails on an error. Furthermore, if set, having no memory policy leads to a warning (instead of a failure), and warnings are treated more carefully.

I don't think I'm allowed to just bail if I'm in `dealloc`, so were you thinking of always continuing? But would I then need to clear any errors to ensure that there are no double errors any time?

The one advantage of what I have is that there is, by construction, no real change: for `__setstate__`, the error is fast, which presumably ends up percolating up so that `self` gets deallocated and the any remaining clearing is done.

Anyway, I would also prefer it if I could just bail on error, but am confused!

numpy/_core/src/multiarray/arrayobject.c
mhvk: My tendency here would be to bite the bullet and make this an error. But for now let me just put the code back.

numpy/_core/src/multiarray/arrayobject.c
mhvk: OK, reverted.

numpy/_core/src/multiarray/arrayobject.c
mhvk: Hmm, those extensions would get warning messages all the time, though, which seems pretty annoying. Anyway, out of scope here!

numpy/_core/src/multiarray/arrayobject.c
mhvk: I guess as long as `dealloc` is safe, it should indeed be OK: https://github.com/numpy/numpy/blob/f590d096927d9165979d3a7fb3b72a07364344f2/numpy/_core/src/multiarray/alloc.cpp#L191-L205

numpy/_core/src/multiarray/methods.c
mhvk: If that is the case, the code was wrong already: the only change I made, really, is to change the start from `nbytes = 1` to `nbytes = itemsize`, which allows removing the final multiplication with it.

If one wanted to have the same behaviour as in the constructor, one needs to ensure that for a dimension of zero, that does not get multiplied through.

Note that arguably there is a good case to factor out pieces of the constructor as well, and use those here, but maybe out of scope?

mhvk (commented):
@mattip - thanks for the reminder, I've now rebased and addressed most of the comments.

@seberg - the main remaining suggestion you made is to have the new function always just error, but I'm unsure how to deal with possibly multiple errors; see in-line.

numpy/_core/src/multiarray/arrayobject.c
seberg: Ah sorry, yeah... I guess it might be nice if it would execute everything either way, but that doens't make sense, since you would have to catch all the errors which is worse.

numpy/_core/src/multiarray/arrayobject.c
seberg: I'll be so free and remove the error path for this one.  This warning still defaults to `False`, so I think we have to make it a decision to default it to be True first (and maybe explain how to turn it off with the env variable within the warning).

---

I realize this doesn't actually matter at all, because the pickle path will never run into this.

numpy/_core/src/multiarray/arrayobject.c
```diff
@@ -408,25 +420,34 @@ array_dealloc(PyArrayObject *self)
...
             }
         }
+        /* mem_handler can be absent if NPY_ARRAY_OWNDATA arbitrarily set */
         if (fa->mem_handler == NULL) {
+            /* For non-dealloc case, we simply error */
+            if (!unraisable) {
+                PyErr_SetString(PyExc_RuntimeError,
+                                "no memory handler found but OWNDATA flag set");
+                return -1;
+            }
```
seberg:
```suggestion
```

numpy/_core/src/multiarray/methods.c
seberg: Yeah, sorry, I must have went through pretty quickly back then. It is strictly speaking wrong, but you are right that it was just as wrong before.

seberg (approved):
Thanks Marten, sorry that my review was a bit sloppy there asking for refactors that didn't make sense.

I still wonder if we actually want to use this clearing logic, but this is a nice cleanup and I am not sure that refactoring things there can be done really.

numpy/_core/src/multiarray/arrayobject.c
seberg: Ah, whoops, this means that the error is not raised when we the warning is raised, so not quite ideal, let me see...

seberg:
Sorry, my deleting those lines wasn't a good idea, but instead I refactored it to add a small helper so we can do:
```
                if (write_and_clear_error_if_unraisable(retval, unraisable) < 0) {
                    return -1;
                }
```
(And remove the `WARN_IN_DEALLOC` one instead.)  Not the prettiest, but I think it is a bit clearer than before and means that the warning behavior is just identical, which to me feels like the safer choice.

Anyway, I think it's all good, and if I come across it again I will merge, but if you have a brief look and merge that is even better.

mhvk:
@seberg - yes, I like your improvements, and good to get rid of `WARN_IN_DEALLOC`. Let's merge this.

PR #30417, opened 2025-12-11
MAINT: remove some old work-arounds and test skips for PyPy

mattip:
Related to #30416. Clean up some old PyPy work-arounds no longer needed since those versions have not been supported for more than two years.

Note these are no longer tested in CI, but I tested them locally.

### DIFF
--- a/numpy/_core/src/multiarray/stringdtype/dtype.c
+++ b/numpy/_core/src/multiarray/stringdtype/dtype.c
@@ -399,14 +399,7 @@ stringdtype_getitem(PyArray_StringDTypeObject *descr, char **dataptr)
         }
     }
     else {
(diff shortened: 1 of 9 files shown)


jorenham:
There are several other pypy-specific pytest skipif's and xfail's, e.g. https://github.com/numpy/numpy/blob/1ec4bd0f6031ae86a372394ea06e6e23ce90d3c1/numpy/_core/tests/test_umath.py#L4051

mattip:
All the others are still needed, see #30416 for more discussion. It could be that the outcome of that issue is to drop all future PyPy support, which can be done in a different PR

ngoldbaum (approved):
Thanks Matti!

PR #30418, opened 2025-12-11
BUG: fix double evaluation in PyArrayScalar_RETURN_BOOL_FROM_LONG

### LINKED ISSUE #30389: BUG: [IPC] Refcount corruption and crash in boolean memmap reads due to double evaluation in `PyArrayScalar_RETURN_BOOL_FROM_LONG`
### Describe the issue:

I am using `numpy.memmap` on a local file (located in `/dev/shm`) as an inter-process communication mechanism.

In the example setup, I create a `(400, 8000)` memmapped array at `/dev/shm/test.npy`.
The workflow is:

1. Run `writer.py`, which writes elements to the memmap one by one.
2. Run `reader.py`, which reads elements sequentially, waiting for each slot to be written before proceeding to the next.

When running the reader process, it crashes with the following error:
```
Reading round 0
Reading round 1
free(): invalid pointer
[1] 11203 abort (core dumped)
```

```python
# There are two python scripts writer.py and reader.py:
# writer.py
import time
import numpy as np

arr = np.memmap("/dev/shm/test.npy", dtype=np.bool_, mode="w+", shape=(400, 8000))

for i in range(arr.shape[0]):
    print(f"Writing round {i}")
    for j in range(arr.shape[1]):
        arr[i, j] = True
        time.sleep(0.0001)

# reader.py
import numpy as np

arr = np.memmap("/dev/shm/test.npy", dtype=np.bool_, mode="r", shape=(400, 8000))

for i in range(arr.shape[0]):
    print(f"Reading round {i}")
    for j in range(arr.shape[1]):
        while not arr[i, j]: pass
```

### Error message:

```shell
Reading round 0
Reading round 1
Reading round 2
Reading round 3
Reading round 4
Reading round 5
Reading round 6
Reading round 7
Reading round 8
Reading round 9
free(): invalid pointer
--Type <RET> for more, q to quit, c to continue without paging--

Thread 1 "python" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7c5d859 in __GI_abort () at abort.c:79
#2  0x00007ffff7cc8266 in __libc_message (action=action@entry=do_abort, fmt=fmt@entry=0x7ffff7df2298 "%s\n") at ../sysdeps/posix/libc_fatal.c:156
#3  0x00007ffff7cd02fc in malloc_printerr (str=str@entry=0x7ffff7df04c1 "free(): invalid pointer") at malloc.c:5347
#4  0x00007ffff7cd1b2c in _int_free (av=<optimized out>, p=<optimized out>, have_lock=0) at malloc.c:4173
#5  0x000055555572ff74 in _Py_DECREF () at /tmp/build/80754af9/python-split_1634043551344/work/Include/object.h:478
#6  _PyEval_EvalFrameDefault (f=<optimized out>, throwflag=<optimized out>) at /tmp/build/80754af9/python-split_1634043551344/work/Python/ceval.c:3079
#7  0x00005555557201f0 in PyEval_EvalFrameEx (throwflag=0, f=0x7ffff72d5440) at /tmp/build/80754af9/python-split_1634043551344/work/Python/ceval.c:741
#8  _PyEval_EvalCodeWithName (_co=<optimized out>, globals=<optimized out>, locals=<optimized out>, args=<optimized out>, argcount=<optimized out>, kwnames=0x0, kwargs=0x0, 
    kwcount=0, kwstep=2, defs=0x0, defcount=0, kwdefs=0x0, closure=0x0, name=0x0, qualname=0x0) at /tmp/build/80754af9/python-split_1634043551344/work/Python/ceval.c:4298
#9  0x0000555555721aa3 in PyEval_EvalCodeEx (closure=0x0, kwdefs=0x0, defcount=0, defs=0x0, kwcount=0, kws=0x0, argcount=0, args=0x0, locals=<optimized out>, 
    globals=<optimized out>, _co=<optimized out>) at /tmp/build/80754af9/python-split_1634043551344/work/Python/ceval.c:4327
#10 PyEval_EvalCode (co=<optimized out>, globals=<optimized out>, locals=<optimized out>) at /tmp/build/80754af9/python-split_1634043551344/work/Python/ceval.c:718
#11 0x0000555555795382 in run_eval_code_obj (co=0x7ffff723b030, globals=0x7ffff7314e40, locals=0x7ffff7314e40)
    at /tmp/build/80754af9/python-split_1634043551344/work/Python/pythonrun.c:1166
#12 0x00005555557a6202 in run_mod (mod=<optimized out>, filename=<optimized out>, globals=0x7ffff7314e40, locals=0x7ffff7314e40, flags=<optimized out>, arena=<optimized out>)
    at /tmp/build/80754af9/python-split_1634043551344/work/Python/pythonrun.c:1188
#13 0x00005555557a93ab in pyrun_file (fp=0x555555927d10, filename=0x7ffff71cdbf0, start=<optimized out>, globals=0x7ffff7314e40, locals=0x7ffff7314e40, closeit=1, 
    flags=0x7fffffffd8f8) at /tmp/build/80754af9/python-split_1634043551344/work/Python/pythonrun.c:1085
#14 0x00005555557a958f in pyrun_simple_file (flags=0x7fffffffd8f8, closeit=1, filename=0x7ffff71cdbf0, fp=0x555555927d10)
    at /tmp/build/80754af9/python-split_1634043551344/work/Python/pythonrun.c:439
#15 PyRun_SimpleFileExFlags (fp=0x555555927d10, filename=<optimized out>, closeit=1, flags=0x7fffffffd8f8)
    at /tmp/build/80754af9/python-split_1634043551344/work/Python/pythonrun.c:472
#16 0x00005555557a9a69 in pymain_run_file (cf=0x7fffffffd8f8, config=0x5555558f1720) at /tmp/build/80754af9/python-split_1634043551344/work/Modules/main.c:391
#17 pymain_run_python (exitcode=0x7fffffffd8f0) at /tmp/build/80754af9/python-split_1634043551344/work/Modules/main.c:616
#18 Py_RunMain () at /tmp/build/80754af9/python-split_1634043551344/work/Modules/main.c:695
#19 0x00005555557a9c69 in Py_BytesMain (argc=<optimized out>, argv=<optimized out>) at /tmp/build/80754af9/python-split_1634043551344/work/Modules/main.c:1127
#20 0x00007ffff7c5f083 in __libc_start_main (main=0x55555565c7d0 <main>, argc=2, argv=0x7fffffffdaf8, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, 
    stack_end=0x7fffffffdae8) at ../csu/libc-start.c:308
#21 0x000055555574b427 in _start ()
```

`import sys, numpy; print(numpy.__version__); print(sys.version)` shows:
```
1.24.4
3.8.12 (default, Oct 12 2021, 13:49:34) 
[GCC 7.5.0]
```

### Runtime Environment:

`import numpy; numpy.show_runtime()` shows:

```
[{'simd_extensions': {'baseline': ['SSE', 'SSE2', 'SSE3'],
                      'found': ['SSSE3',
                                'SSE41',
                                'POPCNT',
                                'SSE42',
                                'AVX',
                                'F16C',
                                'FMA3',
                                'AVX2'],
                      'not_found': ['AVX512F',
                                    'AVX512CD',
                                    'AVX512_KNL',
                                    'AVX512_KNM',
                                    'AVX512_SKX',
                                    'AVX512_CLX',
                                    'AVX512_CNL',
                                    'AVX512_ICL']}},
 {'architecture': 'Prescott',
  'filepath': '/home/yang/miniconda3/lib/python3.8/site-packages/numpy.libs/libopenblas64_p-r0-15028c96.3.21.so',
  'internal_api': 'openblas',
  'num_threads': 24,
  'prefix': 'libopenblas',
  'user_api': 'blas',
  'version': '0.3.21'}]
```

### Context for the issue:

After reviewing NumPy’s source code, I believe I have identified the root cause of the crash.

# Root Cause Analysis
When the reader process performs `arr[i, j]`, NumPy executes `PyArray_Scalar`, implemented in
`numpy/_core/src/multiarray/scalarapi.c` (see line ):
```
    if (type_num == NPY_BOOL) {
        PyArrayScalar_RETURN_BOOL_FROM_LONG(*(npy_bool*)data);
    }
```

Here, data points to the underlying memory inside the shared memmap region. Because the writer process may modify this region concurrently, the value may change between operations.

Expanding the macro reveals the issue.
`PyArrayScalar_RETURN_BOOL_FROM_LONG` is defined as:
```
#define PyArrayScalar_RETURN_BOOL_FROM_LONG(i)                  \
        return Py_INCREF(PyArrayScalar_FromLong(i)), \
                PyArrayScalar_FromLong(i)
```

And `PyArrayScalar_FromLong` returns one of NumPy’s two internal singleton objects for boolean scalars:
```
#define PyArrayScalar_FromLong(i) \
        ((PyObject *)(&(_PyArrayScalar_BoolValues[((i)!=0)])))
```

Thus, for a given boolean value, NumPy:
1. Reads `*(npy_bool*)data`
2. Maps it to one of the internal objects `_PyArrayScalar_BoolValues[False]` or `[True]`
3. Increments the reference count on the selected object
4. Reads `*(npy_bool*)data` **again**
5. Returns (possibly) a different singleton object, depending on whether the value changed between steps (1) and (4)

Because the shared memory is being modified concurrently by the writer process, the following race can occur:
* The first dereference sees `False`, so the macro increments the refcount of `_PyArrayScalar_BoolValues[False]`.
* Before the return expression runs, the writer changes the value to `True`.
* The macro returns `_PyArrayScalar_BoolValues[True]` instead.
* Later, `Py_DECREF` is called on the returned object (`True`), not the one whose refcount was incremented (`False`).

This causes the reference count of `_PyArrayScalar_BoolValues[True]` to reach zero incorrectly, leading to a crash such as:
```
free(): invalid pointer
Aborted (core dumped)
```

# Proposed Solution
The core issue is that `PyArrayScalar_RETURN_BOOL_FROM_LONG` is a macro, so it evaluates `data` twice. Rewriting this macro into a function might be too invasive, but a minimal fix can be applied directly in `scalarapi.c`.

Specifically, the boolean value can be safely stored before passing it to the macro:
```
    if (type_num == NPY_BOOL) {
        npy_bool val = *(npy_bool*)data;
        PyArrayScalar_RETURN_BOOL_FROM_LONG(val);
    }
```

# Remaining Question
`np.bool_` is implemented using two internal singleton objects.
My remaining question is:

**Is `np.bool_` the only dtype whose scalar-returning path relies on internal singleton objects and therefore is vulnerable to this kind of race?**

If other dtypes also use shared singleton scalar objects, similar issues may exist.

y7070:
Closes #30389

### DIFF
--- a/numpy/_core/tests/test_multiprocessing.py
+++ b/numpy/_core/tests/test_multiprocessing.py
@@ -1,0 +1,51 @@
+import multiprocessing
+from concurrent.futures import ProcessPoolExecutor
+from multiprocessing import shared_memory
+
+import pytest
+
+import numpy as np
+from numpy.testing import IS_WASM
+
+pytestmark = pytest.mark.thread_unsafe(
+    reason="tests in this module are explicitly multi-processed"
+)
+
+def bool_array_writer(shm_name, n):
+    # writer routine for test_read_write_bool_array
+    import time
+    shm = shared_memory.SharedMemory(name=shm_name)
+    arr = np.ndarray(n, dtype=np.bool_, buffer=shm.buf)
+    for i in range(n):
+        arr[i] = True
+        time.sleep(0.00001)
+
+def bool_array_reader(shm_name, n):
+    # reader routine for test_read_write_bool_array
+    shm = shared_memory.SharedMemory(name=shm_name)
+    arr = np.ndarray(n, dtype=np.bool_, buffer=shm.buf)
+    for i in range(n):
+        while not arr[i]:
+            pass
+
+@pytest.mark.skipif(IS_WASM,
+                    reason="WASM does not support _posixshmem")
+def test_read_write_bool_array():
+    # See: gh-30389
+    #
+    # Prior to Python 3.13, boolean scalar singletons (np.True / np.False) were
+    # regular reference-counted objects. Due to the double evaluation in
+    # PyArrayScalar_RETURN_BOOL_FROM_LONG, concurrent reads and writes of a
+    # boolean array could corrupt their refcounts, potentially causing a crash
+    # (e.g., `free(): invalid pointer`).
+    #
+    # This test creates a multi-process race between a writer and a reader to
+    # ensure that NumPy does not exhibit such failures.
+    n = 10000
+    shm = shared_memory.SharedMemory(create=True, size=n)
+    with ProcessPoolExecutor(max_workers=2) as executor:
+        f_writer = executor.submit(bool_array_writer, shm.name, n)
+        f_reader = executor.submit(bool_array_reader, shm.name, n)
+    shm.unlink()
+    f_writer.result()
+    f_reader.result()
--- a/numpy/_core/tests/test_multiprocessing.py
+++ b/numpy/_core/tests/test_multiprocessing.py
@@ -1,7 +1,3 @@
-import multiprocessing
-from concurrent.futures import ProcessPoolExecutor
-from multiprocessing import shared_memory
-
 import pytest
 
 import numpy as np
@@ -14,6 +10,7 @@ pytestmark = pytest.mark.thread_unsafe(
 def bool_array_writer(shm_name, n):
     # writer routine for test_read_write_bool_array
     import time
+    from multiprocessing import shared_memory
     shm = shared_memory.SharedMemory(name=shm_name)
     arr = np.ndarray(n, dtype=np.bool_, buffer=shm.buf)
     for i in range(n):
@@ -22,6 +19,7 @@ def bool_array_writer(shm_name, n):
 
 def bool_array_reader(shm_name, n):
     # reader routine for test_read_write_bool_array
+    from multiprocessing import shared_memory
     shm = shared_memory.SharedMemory(name=shm_name)
     arr = np.ndarray(n, dtype=np.bool_, buffer=shm.buf)
     for i in range(n):
@@ -41,6 +39,8 @@ def test_read_write_bool_array():
     #
     # This test creates a multi-process race between a writer and a reader to
     # ensure that NumPy does not exhibit such failures.
+    from concurrent.futures import ProcessPoolExecutor
+    from multiprocessing import shared_memory
     n = 10000
     shm = shared_memory.SharedMemory(create=True, size=n)
     with ProcessPoolExecutor(max_workers=2) as executor:
--- a/numpy/_core/include/numpy/arrayscalars.h
+++ b/numpy/_core/include/numpy/arrayscalars.h
@@ -173,9 +173,11 @@ typedef struct {
 #define PyArrayScalar_True ((PyObject *)(&(_PyArrayScalar_BoolValues[1])))
 #define PyArrayScalar_FromLong(i) \
         ((PyObject *)(&(_PyArrayScalar_BoolValues[((i)!=0)])))
-#define PyArrayScalar_RETURN_BOOL_FROM_LONG(i)                  \
-        return Py_INCREF(PyArrayScalar_FromLong(i)), \
-                PyArrayScalar_FromLong(i)
+#define PyArrayScalar_RETURN_BOOL_FROM_LONG(i) do {     \
+        PyObject *obj = PyArrayScalar_FromLong(i);      \
+        Py_INCREF(obj);                                 \
+        return obj;                                     \
+} while (0)
 #define PyArrayScalar_RETURN_FALSE              \
         return Py_INCREF(PyArrayScalar_False),  \
                 PyArrayScalar_False


numpy/_core/tests/test_multiprocessing.py
```diff
@@ -0,0 +1,51 @@
...
+
+def bool_array_writer(barrier, shm_name, n):
+    # writer routine for test_read_write_bool_array
+    import time
+    shm = shared_memory.SharedMemory(name=shm_name)
+    arr = np.ndarray(n, dtype=np.bool_, buffer=shm.buf)
+    barrier.wait()
+    for i in range(n):
+        arr[i] = True
+        time.sleep(0.00001)
```
ngoldbaum, line 18: Is this strictly necessary? I try to avoid writing tests that depend on timing.

numpy/_core/tests/test_multiprocessing.py
```diff
@@ -0,0 +1,51 @@
...
+    # ensure that NumPy does not exhibit such failures.
+    barrier = multiprocessing.Barrier(2)
+    n = 10000
+    shm = shared_memory.SharedMemory(create=True, size=n)
+    writer = Process(target=bool_array_writer, args=(barrier, shm.name, n))
+    reader = Process(target=bool_array_reader, args=(barrier, shm.name, n))
+    writer.start()
+    reader.start()
+    writer.join()
+    reader.join()
```
ngoldbaum: I'd probably write this using `ProcessPoolExecutor` to avoid this boilerplate and join the processes by construction with a `with` statement.

ngoldbaum (commented):
A couple minor comments. The C change looks correct to me.

numpy/_core/tests/test_multiprocessing.py
y7070: Thanks for pointing this out!

The small sleep helps ensure that the reader and writer stay interleaved rather than the writer running far ahead. Without it, the writer can finish a large portion of the loop before the reader observes any intermediate states, which makes the race much harder to trigger reproducibly.

The sleep interval here is intentionally very small and chosen together with n so that the test stays fast while still exercising the interleaving needed to expose the bug reliably.

numpy/_core/tests/test_multiprocessing.py
y7070: Got it — I’ll switch to ProcessPoolExecutor so the process setup and joining are handled automatically.

y7070:
rebased the commit history to keep only one commit :rocket:

ngoldbaum:
LGTM! @seberg would you mind looking at this one and merging if you agree.

seberg:
I am happy if you like the test and I guess you do. I was hoping to use advanced indexing and get away with just threading but I am not exactly sure how to design them best to make sure they actually result in a crash :).

So thanks @y7070, seems fine to just put it in as is.

PR #30422, opened 2025-12-11
ENH: Return rank 0 for empty matrices in matrix_rank

false200:
This PR fixes a bug in `np.linalg.matrix_rank` where empty matrices
(previously with 0 rows or 0 columns) would raise a ValueError due to 
attempting a reduction operation on a zero-size array.

Changes include:
- Return 0 for all empty matrices in `matrix_rank`.
- Added tests covering all empty matrix shapes to ensure the correct behavior.

Files modified:
- `numpy/linalg/_linalg.py`
- `numpy/linalg/tests/test_linalg.py`

Closes #30421

### DIFF
--- a/numpy/linalg/_linalg.py
+++ b/numpy/linalg/_linalg.py
@@ -2131,10 +2131,6 @@ def matrix_rank(A, tol=None, hermitian=False, *, rtol=None):
     if A.ndim < 2:
         return int(not all(A == 0))
 
-    # Handle empty matrices - rank is 0 for matrices with 0 rows or 0 columns
-    if A.size == 0:
-        return 0
-
     S = svd(A, compute_uv=False, hermitian=hermitian)
 
     if tol is None:
@@ -2142,7 +2138,7 @@ def matrix_rank(A, tol=None, hermitian=False, *, rtol=None):
             rtol = max(A.shape[-2:]) * finfo(S.dtype).eps
         else:
             rtol = asarray(rtol)[..., newaxis]
-        tol = S.max(axis=-1, keepdims=True) * rtol
+        tol = S.max(axis=-1, keepdims=True, initial = 0) * rtol
     else:
         tol = asarray(tol)[..., newaxis]
 
--- a/numpy/linalg/_linalg.py
+++ b/numpy/linalg/_linalg.py
@@ -2138,7 +2138,7 @@ def matrix_rank(A, tol=None, hermitian=False, *, rtol=None):
             rtol = max(A.shape[-2:]) * finfo(S.dtype).eps
         else:
             rtol = asarray(rtol)[..., newaxis]
-        tol = S.max(axis=-1, keepdims=True, initial = 0) * rtol
+        tol = S.max(axis=-1, keepdims=True, initial=0) * rtol
     else:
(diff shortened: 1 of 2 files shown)


numpy/linalg/_linalg.py
```diff
@@ -2130,6 +2130,11 @@ def matrix_rank(A, tol=None, hermitian=False, *, rtol=None):
     A = asarray(A)
     if A.ndim < 2:
         return int(not all(A == 0))
+
+    # Handle empty matrices - rank is 0 for matrices with 0 rows or 0 columns
+    if A.size == 0:
+        return 0
```
seberg: I wonder if we should just add `initial=0` which I assume also works. The tests you added are missing the case of `np.zeros((0, 5, 5))` where this return is clearly incorrect.

EDIT: Actually, while we are here, can you check if we don't return `np.intp(0)` on the last return?  We may want to adjust the return type for the other early return as well.

false200: Hi @seberg, thanks for the feedback!
Added initial=0 to the S.max(... ) call as suggested, and I also added the missing np.zeros((0, 5, 5)) test.

I ran the linalg test suite locally and everything looks good.
All relevant tests passed, only the usual SKIP/XFAIL cases showed up.

WhatsApp Image 2025-12-12 at 22 58 30_538d4365
Let me know if you'd like any additional tweaks!

false200: Hi @seberg , just a quick ping , all checks have passed now.

Would you mind taking a look for merge? Thanks!

seberg (approved):
Code looks good now (I might punt on backporting, but should be OK too).
The tests could be nicer with parametrization, but it is OK.

I am not quite 100% on whether zero is always expected, so wondering if you can argue that or know some prior art?

false200:
@seberg Thanks for the careful review and the questions
On the semantics, returning rank = 0 for empty matrices is consistent with standard linear algebra definitions and with prior art.
From a theoretical standpoint, rank is the dimension of the column (or row) space. For matrices with zero rows or zero columns, there are no columns (or no vectors at all), so the dimension of the span is 0 by definition. This is the convention used in linear algebra texts when discussing degenerate or empty linear maps.
From an SVD perspective, which matrix_rank is based on, an empty matrix has no singular values, so the count of singular values greater than the tolerance is naturally zero. Using initial=0 simply makes this behavior explicit and avoids the reduction error.
There is also prior art 
 - NumPy itself already returns 0 for degenerate inputs such as np.linalg.matrix_rank(np.zeros((n,))), which aligns with this interpretation.
 - MATLAB’s rank returns 0 for empty matrices (it treats the empty singular-value set as having no values above tolerance).
 - This is consistent with the rank–nullity theorem - if the number of columns is zero, rank must be zero.
 
 So this change doesn’t introduce a new definition so much as make the existing mathematical convention explicit and handle it robustly in code.
Happy to adjust wording or tests further if you’d like.

seberg:
Cool, thanks for confirming, agreed seems rather clear convention, so let's put this in, thanks.

Should be safe for backporting, but I would probably just consider it an enhancement.

false200:
@seberg Thanks for clarifying!

PR #30427, opened 2025-12-13
BUG: Fix GH#30349 einsum explicit-output ellipsis reduction when optimize=False

pavansai018:
## Summary

Fix `np.einsum` so that expressions with an ellipsis on the input and an explicit output that omits the ellipsis (e.g. `'i...->i'`) behave correctly when `optimize=False`. `optimize` should only affect performance, not correctness. This change makes
the non-optimized (`c_einsum`) path handle ellipsis-reduction the same way as the optimized path.

## Problem

For inputs like `x = np.ones((3, 3, 3), int)`, the expression:

```python
np.einsum("i...->i", x)
```

is expected to reduce (sum) over the ellipsis axes and return shape (3,). However, optimize=False raises an error, while optimize=True produces the correct result. This is a correctness bug because the optimize flag must not change semantics.

See: numpy/numpy#30349.

## What changed

Ensure the optimize=False path correctly parses/expands ellipsis for explicit-output expressions where the ellipsis is reduced away. Keep behavior identical between optimized and non-optimized execution paths.

Tests

Added/updated regression test covering `i...->i` with both `optimize=False` and `optimize=True`.
Fixes: numpy/numpy#30349

### DIFF
--- a/numpy/_core/einsumfunc.py
+++ b/numpy/_core/einsumfunc.py
@@ -1607,6 +1607,11 @@ def einsum(*operands, out=None, optimize=False, **kwargs):
     if optimize is False:
         if specified_out:
             kwargs['out'] = out
+        # FIX:GH#30349 normalize ellipsis parsing via python parser ---
+        if operands and isinstance(operands[0], str) and "..." in operands[0]:
+            input_sub, output_sub, parsed_ops = _parse_einsum_input(operands)
+            subscripts = input_sub + "->" + output_sub
+            return c_einsum(subscripts, *parsed_ops, **kwargs)
         return c_einsum(*operands, **kwargs)
 
     # Check the kwargs to avoid a more cryptic error later, without having to
--- a/numpy/_core/tests/test_einsum.py
+++ b/numpy/_core/tests/test_einsum.py
@@ -1349,3 +1349,16 @@ def test_einsum_chunking_precision():
 
     # At with GROWINNER 11 decimals succeed (larger will be less)
     assert_almost_equal(res, value, decimal=15)
+    
+def test_einsum_ellipsis_optimize_false_regression_30349():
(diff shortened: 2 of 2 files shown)


numpy/_core/einsumfunc.py
```diff
@@ -1607,6 +1607,11 @@ def einsum(*operands, out=None, optimize=False, **kwargs):
     if optimize is False:
         if specified_out:
             kwargs['out'] = out
+        # FIX:GH#30349 normalize ellipsis parsing via python parser ---
+        if operands and isinstance(operands[0], str) and "..." in operands[0]:
+            input_sub, output_sub, parsed_ops = _parse_einsum_input(operands)
+            subscripts = input_sub + "->" + output_sub
+            return c_einsum(subscripts, *parsed_ops, **kwargs)
```
mattip, line 1614: What would happen if you always take this path? Does it affect performance?

pavansai018: Yes — it adds python-level overhead to every call.

Even when there is no ellipsis bug to fix, we would now pay for:

- parsing the subscript string in Python,
- ellipsis/output construction logic,
- converting all operands via asanyarray, before entering c_einsum.

This overhead is most noticeable for small/medium arrays (where Python time is comparable to compute time).
For large contractions, the compute dominates, so the relative overhead is smaller — but it’s still unnecessary work for the common case.

seberg:
I am going to close this, since you have a second similar PR open where it feels the analysis of what is wrong is too superficial.
Maybe this is the right place on account of simplicity, but this may also be better handled in the C-code.

Also, I would like to see a test that reject the cases with dubious behavior that were identified in the discussion on the issue.

seberg:
Don't worry about just opening a new PR after thinking about it more. But it feels like two PRs that tackle problems superficially and need a lot of more thought are a bit much for a start.

PR #30428, opened 2025-12-13
WIP: Rewrite NumPy in Zig

jorenham:
... starting with a couple of ufunc kernels :)

But given all the talk about C++ and Rust, I thought that Zig could perhaps also be worth considering. So I [^1] wrote this little demo of what it could look like if Zig would be part of NumPy.
Note that I'm a C noob, and my Zig knowledge is even worse, so there's probably a lot of room for improvement here. 

It's a lot easier to integrate than rust (curerntly not even supported by meson), because all you need to compile zig is `pip install ziglang`. It's also quite a bit closer to C than Rust, and you can even include C headers in Zig ([example](https://gioele.dev/blog/calling-c-code-from-zig/)). In terms of type-safety, I'd say that it lies somewhere in between C and Rust. Zig is probably also easier to learn that Rust for those familiar with C/C++.

Anyway, don't take this too seriously; the title is clearly clickbait, and this is nothing but a quick little expirement ;)

[^1]: with a lot of help from copilot (GPT-5.2)


mattip:
Rather than putting experiments like this on the main repo, perhaps do this on your fork?

jorenham:
> Rather than putting experiments like this on the main repo, perhaps do this on your fork?

Oh no worries; I'm not planning on working on this any more (I meant "WIP" as in "should not be merged" rather than the literal "work in progress", which in hindsight is clearly confusing).

rgommers:
Nice to see what it'd take to use Zig. While I like it better than either C++ or Rust, integration-wise it won't work any time soon I think (Rust won't either).

> It's a lot easier to integrate than rust (curerntly not even supported by meson), 

Rust support in Meson isn't mature, but is present and should work fine for simpler cases/needs.

PR #30429, opened 2025-12-13
TYP: ``__init__.pyi``: alias ``builtins.bool`` as ``py_bool``

jorenham:
THis also simplifies `builtins.str` to `str`, and replaces `np.bool` with `bool_`, removing the need for the self-referential `import numpy as np`.


charris:
Thanks Joren.

PR #30431, opened 2025-12-14
DOC: Clarify in-place assignment description in basics.copies

### LINKED ISSUE #28780: DOC: Incorrect description of in-place operations

https://numpy.org/devdocs/reference/generated/numpy.ndarray.__iadd__.html

In the documentation for in-place operations (`__iadd__`, `__isub__`, etc.), it states "Return self+=value." However, an in-place operation (`self+=value`) is a statement, not an expression, so it does not have a return value.

The description should be `same as self+=value`, `perform self+=value` or something similar.

sriram2930:
Fixes #28780

The previous wording "no view or copy is created as the assignment happens in-place" was confusing and could be misinterpreted. 

This PR clarifies the language to explicitly state that the assignment modifies the array in-place without creating a view or copy.

### DIFF
--- a/doc/source/user/basics.copies.rst
+++ b/doc/source/user/basics.copies.rst
@@ -94,9 +94,8 @@ which in turn will not affect ``y`` at all::
     array([[3, 4, 5],
            [6, 7, 8]])
 
-It must be noted here that during the assignment of ``x[[1, 2]]`` no view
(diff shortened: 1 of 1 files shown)


jorenham:
I don't see the improvement to be honest. If another maintainer disagrees they can re-open this.

Issue #30405, opened 2025-12-09
TYP: broad `np.linalg` return types


autra:
### Describe the issue:

Followup of https://github.com/numpy/numpy/issues/27957 because the issue is not fixed for `numpy.linalg.inv`. 

`numpy.linalg.inv` apparently loses some typing information.

The following snippet is a test case demonstrating the issue. In my real code, I tried to pass inv_test to a function that only accepts `NDArray[np.float64]`. I'd expect an inversion to keep the same float size in type.

### Reproduce the code example:

```python
import numpy as np
import numpy.typing as npt

test: npt.NDArray[np.float64] = np.array([1.0, 2.0, 3.0], dtype=np.float64)
inv_test: npt.NDArray[np.float64] = np.linalg.inv(test)
```

### Error message:

```shell
test.py:5: error: Incompatible types in assignment (expression has type "ndarray[tuple[Any, ...], dtype[floating[Any]]]", variable has type "ndarray[tuple[Any, ...], dtype[float64]]")  [assignment]
Found 1 error in 1 file (checked 1 source file)
```

### Python and NumPy Versions:

numpy: 2.3.5
python: 3.14.0 (main, Oct  7 2025, 09:34:52) [GCC 14.3.0]

### Type-checker version and settings:

```
❯ mypy --version
mypy 1.19.0 (compiled: yes)
```

I used `mypy test.py` and `mypy --local-partial-types --allow-redefinition-new  test.py`, same result.

### Additional typing packages.

typing_extensions: 4.15.0

jorenham:
This is unrelated to #27957, which was only an issue for mypy users. This, on the other hand, isn't limited to mypy; other type-checkers will also complain here.

I can see that this is error is unexpected, and pretty annoying to deal with. But technically speaking, this isn't a regression, because this issue also existed prior to 2.2; it just wasn't reported as such. The `np.linalg.inv` stubs itself haven't changed (functionally) since 2021 ([src](https://github.com/numpy/numpy/commit/ff80d911d2ef3644b658cdb003f8fff37b5b101e#diff-5424f0cc3e9875e487091f53b666fc98f6acdb7ae77752bc0ec0d6ceed55c99aR83-R88)).

If you look at the `np.linalg.inv` stub definition ([src](https://github.com/numpy/numpy/blob/a48fb8562cce7e564f50f84f3717621b08edf490/numpy/linalg/_linalg.pyi#L160-L165)), you'll see that the return types are pretty broad:

```pyi
[...]

PR #30434, opened 2025-12-15
TYP: ``linalg.solve`` and ``tensorsolve``: improved return dtype specialization

jorenham:
towards #30405


charris:
Thanks Joren.

PR #30438, opened 2025-12-15
MAINT: Update msvc-allowed-warnings.txt

charris:
Line numbers have changed in numpy/random/_generator.pyx.c


charris:
Probably due to the Cython  3.2.3 release Dec 14. Can we use less specific warning entries to avoid this in the future?

mattip (approved):
LGTM, pending CI passing

PR #30442, opened 2025-12-16
BUG: Fix missing dtype reset in MaskedRecords.view() when dtype is ndarray subclass

sabasiddique1:
## Description
Fixes a bug in `MaskedRecords.view()` where the `dtype = None` assignment was missing when `dtype` is an ndarray subclass, causing both branches of an if/else statement to have identical code.

## Changes
- Added missing `dtype = None` assignment on line 367 in the `if issubclass(dtype, np.ndarray):` branch of `MaskedRecords.view()` method
- Added test `test_view_ndarray_subclass_resets_dtype()` to verify the fix
- This matches the reference implementation in `numpy/ma/core.py` (line 3272)

## Related Issue
Fixes #30441


Aniketsy:
@sabasiddique1  I noticed some lint errors, could you please fix them.

PR #30446, opened 2025-12-16
TYP: ``linalg.*norm``: shape-typing and transparent dtypes

jorenham:
So that's `linalg.norm`, `linalg.matrix_norm` , and `linalg.vector_norm`.

towards #30405


charris:
Needs rebase. Is the scipy error a scipy bug?

jorenham:
> Is the scipy error a scipy bug?

Yup: their `axis` is incorrectly annotated as `tuple[int]` instead of `tuple[int, int]`:
https://github.com/scipy/scipy/blob/fd30d0bd2644a2c25d01acfa538050b28452117f/scipy/_lib/_array_api.py#L427-L431

I'll open a PR for it

charris:
Thanks Joren. There is still an extra empty space, but I doubt you are through with that file.

PR #30447, opened 2025-12-16
DOC: clarify when `genfromtxt` produces a structured array

jkunimune:
The docstring for `genfromtxt` didn't make it clear when the output would be a structured array.  It kind of made it sound like it was only if when `names` is a sequence or a string, but in reality it's whenever `dtype` is structured _or_ `names` is not None.  Now it is stated how the structured outputs behave and when the output is and is not structured.


mattip:
Thanks @jkunimune

PR #30450, opened 2025-12-16
TYP: preserve type for split functions via structural Protocol

### LINKED ISSUE #23005: TYP: wrong return type when applying np.split to pd.DataFrame
### Describe the issue:

The annotation for the return type of `np.split` is `list[NDArray[Any]]`, but when a `pd.DataFrame` is passed, the return type is actually `list[pd.DataFrame]`.

The code example is valid, but `mypy` flags an error.

```python
import numpy as np
import pandas as pd

df = np.split(pd.DataFrame(), ())[0]
pd.melt(df)
```

### Error message:

```shell
example.py:5: error: Argument 1 to "melt" has incompatible type "ndarray[Any, dtype[Any]]"; expected "DataFrame"  [arg-type]
```

### Runtime information:
<details>
1.24.1
3.11.1 (tags/v3.11.1:a7a450f, Dec  6 2022, 19:58:39) [MSC v.1934 64 bit (AMD64)]
[{'simd_extensions': {'baseline': ['SSE', 'SSE2', 'SSE3'],
                      'found': ['SSSE3',
                                'SSE41',
                                'POPCNT',
                                'SSE42',
                                'AVX',
                                'F16C',
                                'FMA3',
                                'AVX2',
                                'AVX512F',
                                'AVX512CD',
                                'AVX512_SKX',
                                'AVX512_CLX',
                                'AVX512_CNL',
                                'AVX512_ICL'],
                      'not_found': []}},
 {'architecture': 'SkylakeX',
  'filepath': 'C:\\Users\\username\\AppData\\Local\\Packages\\PythonSoftwareFoundation.Python.3.11_qbz5n2kfra8p0\\LocalCache\\local-packages\\Python311\\site-packages\\numpy\\.libs\\libopenblas64__v0.3.21-gcc_10_3_0.dll',
  'internal_api': 'openblas',
  'num_threads': 8,
  'prefix': 'libopenblas',
  'user_api': 'blas',
  'version': '0.3.21'}]
None
</details>

janosh:
add overloads for `array_split`, `split`, `hsplit`, `vsplit`, and `dsplit` that preserve the input type when the input implements `swapaxes` and `__getitem__`.

this uses structural typing (Protocol) based on the methods actually used by these split functions internally - not `__array_function__`. the Protocol is:

```python
@type_check_only
class _SupportsSplitOps(Protocol):
    def swapaxes(self, axis1: Any, axis2: Any, copy: Any = ...) -> Any: ...
    def __getitem__(self, key: Any, /) -> Any: ...
```

objects like `pandas.DataFrame` that implement these methods (and return `Self` from them) will have their type preserved through split operations.

### example

```python
import numpy as np
import pandas as pd

df = pd.DataFrame({"a": [1, 2, 3, 4]})
splits = np.array_split(df, 2)
splits[0].columns  # before: error (ndarray has no attr "columns"), after: works
```

### why this works

the split functions use `swapaxes` and slicing (`__getitem__`) internally. for objects like DataFrame that return their own type from these operations, the split result preserves the type. this is pure duck-typing, not `__array_function__` dispatch.

closes #23005

### DIFF
--- a/numpy/lib/_shape_base_impl.pyi
+++ b/numpy/lib/_shape_base_impl.pyi
@@ -66,16 +66,15 @@ class _SupportsArrayWrap(Protocol):
     def __array_wrap__(self) -> _ArrayWrap: ...
 
 # Protocol for array-like objects that preserve their type through split operations.
-# Requires shape (with __getitem__ for axis access), ndim for dimensional checks
-# in hsplit/vsplit/dsplit, swapaxes for axis manipulation, and __getitem__ for
-# slicing. Examples: pandas DataFrame, xarray DataArray.
+# Requires shape for size, ndim for dimensional checks in hsplit/vsplit/dsplit,
+# swapaxes for axis manipulation, and __getitem__ for slicing.
 @type_check_only
 class _SupportsSplitOps(Protocol):
     @property
     def shape(self) -> tuple[int, ...]: ...
     @property
     def ndim(self) -> int: ...
-    def swapaxes(self, axis1: SupportsIndex, axis2: SupportsIndex, /) -> Self: ...
+    def swapaxes(self, axis1: int, axis2: int, /) -> Self: ...
     def __getitem__(self, key: Any, /) -> Self: ...
 
 ###
--- a/numpy/lib/_shape_base_impl.pyi
+++ b/numpy/lib/_shape_base_impl.pyi
@@ -1,5 +1,13 @@
 from collections.abc import Callable, Sequence
-from typing import Any, Concatenate, Protocol, SupportsIndex, overload, type_check_only
+from typing import (
+    Any,
+    Concatenate,
+    Protocol,
+    Self,
+    SupportsIndex,
+    overload,
+    type_check_only,
+)
 
 import numpy as np
 from numpy import (
@@ -57,6 +65,19 @@ class _SupportsArrayWrap(Protocol):
     @property
     def __array_wrap__(self) -> _ArrayWrap: ...
 
+# Protocol for array-like objects that preserve their type through split operations.
+# Requires shape (with __getitem__ for axis access), ndim for dimensional checks
+# in hsplit/vsplit/dsplit, swapaxes for axis manipulation, and __getitem__ for
+# slicing. Examples: pandas DataFrame, xarray DataArray.
+@type_check_only
+class _SupportsSplitOps(Protocol):
+    @property
+    def shape(self) -> tuple[int, ...]: ...
+    @property
+    def ndim(self) -> int: ...
+    def swapaxes(self, axis1: SupportsIndex, axis2: SupportsIndex, /) -> Self: ...
+    def __getitem__(self, key: Any, /) -> Self: ...
+
 ###
 
 def take_along_axis[ScalarT: np.generic](
@@ -113,6 +134,12 @@ def dstack[ScalarT: np.generic](tup: Sequence[_ArrayLike[ScalarT]]) -> NDArray[S
 def dstack(tup: Sequence[ArrayLike]) -> NDArray[Any]: ...
 
 @overload
+def array_split[SplitableT: _SupportsSplitOps](
+    ary: SplitableT,
+    indices_or_sections: _ShapeLike,
+    axis: SupportsIndex = 0,
+) -> list[SplitableT]: ...
+@overload
 def array_split[ScalarT: np.generic](
     ary: _ArrayLike[ScalarT],
     indices_or_sections: _ShapeLike,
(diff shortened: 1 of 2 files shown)


shoyer:
The design for `__array_function__` (see NEP-18) predates support for type checking in NumPy, but it isn't obvious to me that adding typing overloads for non-NumPy arrays is a good idea. This is particularly true for NumPy APIs acting on types like pandas.DataFrame which intentionally deviate from the design of NumPy and only implement an incomplete subset of its design.

Generally speaking my recommendation would be not to try to combine type checking and `__array_function__`. There is basically no way to guarantee type safety here given its highly dynamic design. The `__array_module__` API is a much saner (but more limited) approach here.

janosh:
thanks for the feedback! i updated the PR title and description - they were outdated.

to clarify: this PR does **not** use `__array_function__` at all. the approach is purely structural typing via a Protocol:

```python
class _SupportsSplitOps(Protocol):
    def swapaxes(self, axis1: Any, axis2: Any, copy: Any = ...) -> Any: ...
    def __getitem__(self, key: Any, /) -> Any: ...
```

the split functions internally use `swapaxes` and slicing. for objects like `DataFrame` that implement these methods and return `Self`, the type is preserved. this is just duck-typing - no `__array_function__` dispatch involved.

the Protocol matches what the implementation actually does, so it should be type-safe for any object that properly implements these methods.

shoyer:
OK, interesting. I'm still hesitant to recommend relying on this behavior, but it's been around in NumPy long enough that it's probaby not going to change.

The real challenge is that strictly following the implementation, we don't know what the return type should be -- it could be an entirely different type, principle.

I would still suggest that anybody concerned with type safety should avoid these implicitly overloaded NumPy APIs. In retrospect, they were clearly a design mistake.

jorenham (changes requested):
Thanks for this PR. 

Overall this looks fine to me. The shapes are retained, so even if the input arrays include static shape-typing information, this will still be sound.
One issue with this protocol is that it isn't complete. If you look at the `array_split` implementation, for example, you'll see 

```py
try:
    Ntotal = ary.shape[axis]
except AttributeError:
    Ntotal = len(ary)
```

But as it currently stands, there are objects assignable to `_SupportsSplitOps` that have neither a `shape` nor `__len__`. Please take a closer look at each implementation to see what other methods/attributes the `Protocol` should have.

Recently, NumPy dropped support for Python 3.11, so we're using PEP 695 syntax now, in favor of `TypeVar`. You'll see ruff complaining about this if you run `ruff check`.

This needs tests to verify that it is working as intended. We don't have Pandas available in our test environment, so you'll have to ducktype something (that'll also work at runtime) instead. You can find them under `typing/tests/data`

jorenham:
The primer error is unrelated

janosh:
@jorenham thanks for taking a look and the great feedback. i resolved merge conflicts and i think addressed all your suggestions

numpy/lib/_shape_base_impl.pyi
```diff
@@ -57,6 +65,19 @@ class _SupportsArrayWrap(Protocol):
...
+# Requires shape (with __getitem__ for axis access), ndim for dimensional checks
+# in hsplit/vsplit/dsplit, swapaxes for axis manipulation, and __getitem__ for
+# slicing. Examples: pandas DataFrame, xarray DataArray.
+@type_check_only
+class _SupportsSplitOps(Protocol):
+    @property
+    def shape(self) -> tuple[int, ...]: ...
+    @property
+    def ndim(self) -> int: ...
+    def swapaxes(self, axis1: SupportsIndex, axis2: SupportsIndex, /) -> Self: ...
```
jorenham: I'm not sure if this would cause any issues for pandas, but the `axis{1,2}` are in contravariant (input) positions, so it would be more general to annotate them as `: int`.

janosh: changed to `int`

jorenham (approved):

jorenham:
Before I merge this, did you verify that this indeed resolves the pandas issue from #23005?

janosh:
yes, i tested the changes here on https://github.com/janosh/matbench-discovery which uses `np.array_split` on `pd.DataFrame` a lot

jorenham:
Thanks @janosh; great use of structural typing :ok_hand:

janosh:
looks like this didn't make it into [v2.4.0](https://github.com/numpy/numpy/releases/tag/v2.4.0). will it be in the next patch?

jorenham:
Backporting this will be tricky because of the PEP 695 syntax unfortunately

janosh:
meaning it'll be in 2.5.0? [works for me](https://github.com/janosh/matbench-discovery/pull/299/commits/ae7ffd206d95e95c05f8d098071f4d1575abf5b7), just curious what to expect

jorenham:
> meaning it'll be in 2.5.0? [works for me](https://github.com/janosh/matbench-discovery/pull/299/commits/ae7ffd206d95e95c05f8d098071f4d1575abf5b7), just curious what to expect

Exactly :ok_hand:

PR #30451, opened 2025-12-16
MAINT: Fix np.interp to handle empty arrays

### LINKED ISSUE #30316: BUG: np.interp([], [], []) fails, when it could return an empty array

m4cd4r4:
Fixes #30316

When calling np.interp with empty arrays, it now returns an empty array instead of failing. Simple fix - just check if x is empty and return early.

Tested:
- np.interp([], [], []) now returns array([])
- Works with period parameter too

### DIFF
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -1640,6 +1640,11 @@ def interp(x, xp, fp, left=None, right=None, period=None):
     """
 
     fp = np.asarray(fp)
+    x = np.asarray(x)
+
+    # handle empty arrays
+    if x.size == 0:
+        return x.astype(fp.dtype)
 
     if np.iscomplexobj(fp):
         interp_func = compiled_interp_complex


charris:
Can you make an argument that this is better?

m4cd4r4:
Looking at this again, you're right to push back - my fix is incomplete.

It only handles empty `x`, but the original issue (#30316) is about `np.interp([], [], [])` where xp/fp are also empty. I didn't think that through properly.

@anntzer's use case in the issue (ridge detection returning zero points on some frames) makes sense to me - if you're interpolating at zero points, you should get zero results back, not an error. Same logic as `np.sum([])` returning 0.

I can either:
1. Expand this to handle empty xp/fp consistently + add tests
2. Close it if someone else wants to take a more comprehensive approach

Let me know which you'd prefer.

mattip:
This is a very minimalistic two line change and has no tests. The comment look suspiciously like it was AI generated, with no comment about using AI to help. I will close it. I hope the next attempt at a fix actually follows the contributor guidelines by fixing the issue and testing the fix.

PR #30455, opened 2025-12-17
CI: mypy_primer workaround for Pandas' fixed py311 mypy config

### LINKED ISSUE #230: "Type statement is only supported in Python 3.12 and greater"
We've recently dropped Python 3.11 in NumPy, and are now using PEP 695 syntax in the stubs (🎉), but because Pandas uses `python_version=3.11` ([src](https://github.com/pandas-dev/pandas/blob/bb474293da074cda27993bc5836367c59eb94585/pyproject.toml#L660)) in their mypy config, every mypy_primer run will now report the following:

```diff
pandas (https://github.com/pandas-dev/pandas)
- /home/runner/work/numpy/numpy/numpy_base/numpy/__init__.pyi:731: error: Type statement is only supported in Python 3.12 and greater  [syntax]
+ /home/runner/work/numpy/numpy/numpy_to_test/numpy/__init__.pyi:731: error: Type statement is only supported in Python 3.12 and greater  [syntax]
```

See e.g. https://github.com/numpy/numpy/pull/30443#issuecomment-3660516208

Is there a way to work around this on NumPy's side of things (preferably without excluding pandas), or maybe in mypy_primer itself?

jorenham:
An attempt at working around https://github.com/hauntsaninja/mypy_primer/issues/230

Closes hauntsaninja/mypy_primer#230

### DIFF
--- a/.github/workflows/mypy_primer.yml
+++ b/.github/workflows/mypy_primer.yml
@@ -65,6 +65,7 @@ jobs:
             --known-dependency-selector numpy \
             --old-prepend-path numpy_base --new-prepend-path numpy_to_test \
             --num-shards 1 --shard-index ${{ matrix.shard-index }} \
+            --additional-flags="--python-version=3.12" \
             --debug \
             --output concise \
             | tee diff_${{ matrix.shard-index }}.txt


jorenham:
No mypy_primer output, so this seems to resolve it :ok_hand:

seberg:
Seems like a good work-around, thanks Joren.

PR #30456, opened 2025-12-17
TYP: restore ``generic.__hash__``

### LINKED ISSUE #30445: TYP: [2.4.0rc1 regression] `np.generic` no longer `Hashable`
### Describe the issue:

The code below passes without error when using numpy 2.3.5, but gives the error message below on 2.4.0rc1.

cc @alexfikl

```python
from typing import Hashable

import numpy as np


def f(x: np.generic) -> Hashable:
    return x
```

### Error message:

```shell
/home/andreas/tmp/tnumpy.py:7:12 - error: Type "generic[Any]" is not assignable to return type "Hashable"
    "generic[Any]" is incompatible with protocol "Hashable"
      "__hash__" is an incompatible type
      "__hash__" is not defined as a ClassVar in protocol
        Type "None" is not assignable to type "() -> int" (reportReturnType)
1 error, 0 warnings, 0 notes
```

Python 3.14.2

```
basedpyright 1.36.1
based on pyright 1.1.407
```
no settings applied

_No response_

jorenham:
Closes #30445

### DIFF
--- a/numpy/__init__.pyi
+++ b/numpy/__init__.pyi
@@ -3662,6 +3662,12 @@ class generic(_ArrayOrScalarCommon, Generic[_ItemT_co]):
     @abstractmethod
     def __new__(cls, /, *args: Any, **kwargs: Any) -> Self: ...
 
+    # NOTE: Technically this doesn't exist at runtime, but it is unlikely to lead to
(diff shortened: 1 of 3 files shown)


charris:
Thanks Joren.

PR #30457, opened 2025-12-17
CI: have runtime jobs ignore ``tools/stubtest/**``

jorenham:
This avoids unnecessary CI jobs from being run on changes to the stubtest config, like in e.g. #30456.


charris:
Thanks Joren.

PR #30460, opened 2025-12-17
DEP: expire deprecation for ``finfo(None)``

jorenham:
towards #30440


charris:
Strictly speaking, there should be a test for the  error in `numpy/_core/tests/test_getlimits.py`.

jorenham:
> Strictly speaking, there should be a test for the error in `numpy/_core/tests/test_getlimits.py`.

sounds wise; I'll add it :ok_hand:

charris:
Thanks Joren.

PR #30462, opened 2025-12-17
DEP: remove ``_core.numerictypes.maximum_sctype``

jorenham:
towards #30440

Note that `numpy.maximum_sctype` was already removed in NumPy 2.0, so it was only accessible through `numpy._core.numerictypes`.

### DIFF
--- a/doc/release/upcoming_changes/30462.expired.rst
+++ b/doc/release/upcoming_changes/30462.expired.rst
@@ -1,0 +1,1 @@
+* ``numpy._core.numerictypes.maximum_sctype`` has been removed (deprecated since 2.0)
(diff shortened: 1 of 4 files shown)


doc/release/upcoming_changes/30462.expired.rst
jorenham, line 1: do we even need this :thinking: ? I mean, it's not public API; not even a little bit.

charris: Doesn't hurt to remove dead code.

jorenham: I meant the release note

PR #30463, opened 2025-12-17
DEP: expire deprecations in ``lib._shape_base_impl``

jorenham:
This removes `row_stack` and `lib._shape_base_impl.get_array_wrap`, both deprecated since 2.0.

towards #30440


jorenham:
I thought that `get_array_wrap` is private API only, so that's why I didn't include it in the release note. At least `numpy.get_array_wrap` raises an `AttributeError` on NumPy 2.3.

charris:
That is true, but it doesn't hurt to document the removal. Ideally, no one would  use any of the deprecated functions.

charris:
Thanks Joren.

PR #30466, opened 2025-12-18
TYP: ``linalg.cross``: shape-typing and improved dtype specialization

jorenham:
Note that shape-typing is only used for the input. If we were to also shape-type the output (1d/2d), it would quadruple the number of overloads ((?d, ?d) & (1d, 1d) & ({1,2}d, 2d) & (2d, {1,2}d)), which would result in 4x10=40 overloads.

towards #30405


charris:
Thanks Joren.

PR #30467, opened 2025-12-18
DEP: remove ``lib._npyio_impl.recfrom{txt,csv}``

jorenham:
Note that the `genfromtxt` deprecation for `encoding="bytes"` isn't included here, as I explained in https://github.com/numpy/numpy/issues/30440#issuecomment-3667327856

towards #30440


doc/release/upcoming_changes/30467.expired.rst
```diff
@@ -0,0 +1 @@
+* ``recfromtxt`` and ``recfromcsv`` have been removed from ``numpy.lib._npyio`` in favor of ``numpy.genfromtxt`` (deprecated since 2.0).
```
charris, line 1: These long lines can be broken as long as the continuation is indented. I will do that when editing the release notes, so this is just a heads up.

charris:
Thanks Joren.

PR #30471, opened 2025-12-18
MAINT: remove duplicate byte decode in ``loadtxt``

camriddell:
`numpy/lib/_npyio_impl.py::loadtxt` had repeated code for decoding the `delimiter[bytes]`. This appears to have been a mistake and these lines can be safely removed.

Here is a permalink to main that shows both instances of the duplicated code:
https://github.com/camriddell/numpy/blob/a63efa60f04003da994caec9f4b1fe1071c105f1/numpy/lib/_npyio_impl.py#L1368-L1382


jorenham (approved):
good catch; thanks!
good catch; thanks!

charris:
Thanks @camriddell .

PR #30472, opened 2025-12-18
DEP: remove unused configuration files for npymath and mlib

whyvineet:
This follows the discussion about whether npymath.ini / mlib.ini are still used. #30465


charris:
Let's see what happens. Thanks @whyvineet .

PR #30474, opened 2025-12-18
TYP: ``linalg.matmul``: shape-typing and improved dtype specialization

jorenham:
It has grown pretty beefy and now counts 43 overloads. Unfortunately there's no reliable method to reduce this. It helps that they can be grouping semantically, so it's not all too difficult to read IMO. 
All of these overloads are needed because the output type (scalar/array) depends on both input shape-types, while at the same time there are the NEP 50 dtype casting at play, resulting in a combinatorial explosion (not a very loud one luckily). Then there are also the 16 overloads that work around mypy's and pyright's incorrect (according to the typing spec) behavior regarding `Any` in overloads, breaking the should-be axiomatic "gradual guarantee" (there's a lot more to it than that, but I'll spare you the details).

towards #30405

### DIFF
--- a/numpy/linalg/_linalg.pyi
+++ b/numpy/linalg/_linalg.pyi
@@ -28,7 +28,6 @@ from numpy._typing import (
     _ArrayLikeInt_co,
     _ArrayLikeObject_co,
     _ArrayLikeTD64_co,
-    _ArrayLikeUInt_co,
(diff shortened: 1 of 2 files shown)


charris:
Thanks Joren. Annotations need for loops :)

PR #30475, opened 2025-12-19
MAINT: fix array size declarations in string_partition_resolve_descriptors

stratakis:
The string_partition_resolve_descriptors functions declared array parameters with size [3], but the code accesses indices beyond that range. The partition ufunc has 3 inputs + 3 outputs (6 operands) in string_ufuncs.cpp, and 2 inputs + 3 outputs (5 operands) in stringdtype_ufuncs.cpp.

This doesn't cause runtime issues since the actual arrays passed are correctly sized, however, the declarations serve as documentation and the mismatch could triggers static analysis warnings.

Found by Coverity static analysis.

### DIFF
--- a/numpy/_core/src/umath/string_ufuncs.cpp
+++ b/numpy/_core/src/umath/string_ufuncs.cpp
@@ -1125,9 +1125,9 @@ string_partition_promoter(PyObject *NPY_UNUSED(ufunc),
 static NPY_CASTING
 string_partition_resolve_descriptors(
         PyArrayMethodObject *self,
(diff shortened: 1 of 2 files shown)


seberg (approved):
Thanks @stratakis. Agreed, the number there doesn't matter at all... but if there is a number (maybe C++ wants it, dunno), then it should at least document the right thing :).

stratakis:
Rebased for a typo in the commit message

charris:
Thanks @stratakis .

PR #30479, opened 2025-12-19
TYP: ``linalg.multi_dot``: improved dtype specialization

jorenham:
This is the last `numpy.linalg` function that needed improving, towards #30405.

I'll write a release note for these improvements in a follow-up.


charris:
Thanks Joren.

PR #30443, opened 2025-12-16
TYP: ``linalg.outer``: shape-typing and transparent dtypes

jorenham:
towards #30405

PR #30480, opened 2025-12-19
TYP/DOC: release note for the ``linalg`` typing improvements

### LINKED ISSUE #30405: TYP: broad `np.linalg` return types

jorenham:
This summarizes #30407, #30430, #30433, #30434, #30443, #30446, #30466, #30474, #30479, and closes #30405.


seberg (approved):
Looks good, thanks Joren.  Please just merge. I do wonder if we want to add a `typing` category to the release notes, so not merging just in case you want to do that now.
(It should be pretty simple, IIRC, basically mainly modifying the `jinja` template.)

jorenham:
> I do wonder if we want to add a `typing` category to the release notes, so not merging just in case you want to do that now.

I was also wondering that myself. But I figured that we could also do that later, and then update this release note at that point.

charris:
Thanks Joren.

PR #30481, opened 2025-12-19
DOC: move ``linalg.outer`` from "Decompositions" to "Matrix and vector products"

jorenham:
The outer *product* is, surprisingly, a product. And as far as I'm aware, it's not a decomposition.


seberg:
Looks right, thanks Joren!

PR #30483, opened 2025-12-19
DOC: new "typing" changelog category

jorenham:
ref: https://github.com/numpy/numpy/pull/30480#pullrequestreview-3598763148


seberg:
Thanks a lot Joren, nice!  Looks right, but if not it would also be easy to follow up.  Doc build is done, so at least there is nothing that broke towncrier itself :).

PR #30484, opened 2025-12-19
TYP: ``argpartition`` shape-typing

jorenham:
This adds shape-typing support to 

- `np.argpartition`
- `np.ndarray.argpartition`
- `np.ma.masked_array.argpartition`

and syncs `np.argpartition` with `np.ndarray.argpartition` so that `np.argpartition` will now also reject passing an `order` argument for non-void arrays.


charris:
Thanks Joren.

PR #30487, opened 2025-12-20
REL: Prepare for the NumPy 2.4.0 release (2)

charris:
Don't run multiprocessing test for PyPy on Windows. PyPy uses _winapi which lacks UnmapViewOfFile. This is a known Python problem.


charris:
@mattip, this seems to be a bug on PyPy, which needs to work around the missing UnmapViewOfFile. I don't know if you want to bother with a fix. This didn't turn up in the backport because apparently there is not a PyPy test on windows.

PR #30489, opened 2025-12-20
ENH: use c11 atomics for `numpy`

kumaraditya303:
Experiment: use c11 atomics for numpy instead of hand-written atomic functions.


kumaraditya303:
cc @rgommers @ngoldbaum

meson.build
```diff
@@ -29,6 +29,7 @@ elif cc.get_id() == 'msvc'
     error('NumPy requires at least vc142 (default with Visual Studio 2019) ' + \
           'when building with MSVC')
   endif
+  add_project_arguments('/experimental:c11atomics', language: 'c')
```
ngoldbaum, line 32: @rgommers do you think the CI is sufficient to trust that our builds won't be affected by this? Or is there a way we can fire a test build using the wheel-building infrastructure?

FWIW, coveragepy has been doing this for a while and I don't think they've had any complaints: https://github.com/coveragepy/coveragepy/pull/2019

ngoldbaum (approved):
Nice simplification. I agree that relaxed loads should be fine - in the worst case there will be some blocking or duplication of work in code paths for cache misses.

Just to be extra sure, I wrote a little test script which doesn't produce any race reports on a TSan-instrumented build of 3.14:

```
import numpy as np
from numpy.testing._private.utils import run_threaded


dt = np.dtype(('i4', [('a', '<i2'), ('b', '<i2')]))


def worker(b):
    b.wait()
    hash(dt)


run_threaded(worker, 500, pass_barrier=True)
```

I'd like someone else who knows numpy's build infrastructure on Windows to weigh in about the new experimental build flag.

meson.build
rgommers: Our wheel build infra should be perfectly fine. Whether the option is robust enough I'm less sure about - MS(VC) is hard to trust with this stuff given how much they neglect C support. But, let's find out:) If the code changes look good to you, let's put it in. It's early in the release cycle, so there's enough time to catch it in case it won't work.

Let me just add a comment on MSVC versions.

rgommers (commented):
Nice, this is worth trying indeed. 

The `/experimental:c11atomics` was added in _VS 2022 version 17.5 Preview 2_, which I believe corresponds to toolchain version `19.35`. Right now the minimum we support is `19.20`, so before merging we should bump that version number on line 28 of the top-level `meson.build` file to `19.35`. That way we'll get a clear error message if the installed compiler is too old.

Once that is changed, this LGTM I think.

kumaraditya303:
> Once that is changed, this LGTM I think.

I have updated the msvc version.

ngoldbaum:
It occurs to me: let's add a release note for the MSVC minimum version bump too.

kumaraditya303:
> let's add a release note for the MSVC minimum version bump too.

Where do I need to add it? I am not familar with how numpy handles release notes.

mattip:
See the `README.rst` and other examples in `doc/release/upcoming_changes`

kumaraditya303:
Thanks, I added it to the release notes

ngoldbaum:
CI failures are due to transient network issues during the docs build and are unrelated. Let’s merge this and we’ll see if anyone running against our `main` branch reports issues.

PR #30498, opened 2025-12-22
DOC: ``numpy.select``: fix ``default`` parameter docstring

jorenham:
This indirectly led to the typing issue reported in #30497 that was introduce to #30246 (not to pass the blame; just trying to prevent it from happening again)


charris:
Thanks Joren.

PR #30500, opened 2025-12-22
REV: Revert part of #30164

jorenham:
The removal of this *unused* `SeedlessSequence` declaration (in #30164) broke every library that uses the `numpy.random` Cython API. Specifically, the wheels that were built against `numpy<2.4.0`. See https://github.com/scipy/scipy/issues/24215 for the full discussion.

### DIFF
--- a/.github/check-warnings/msvc-allowed-warnings.txt
+++ b/.github/check-warnings/msvc-allowed-warnings.txt
@@ -15,7 +15,7 @@
 C:\a\numpy\numpy\numpy\random\src\pcg64\pcg64.h(342): warning C4146: unary minus operator applied to unsigned type, result still unsigned
 D:\a\numpy\numpy\numpy\random\src\pcg64\pcg64.h(342): warning C4146: unary minus operator applied to unsigned type, result still unsigned
 cl : Command line warning D9025 : overriding '/arch:SSE2' with '/arch:AVX2'
-numpy/random/_generator.cp312-win32.pyd.p/numpy/random/_generator.pyx.c(26329): warning C4244: 'function': conversion from 'int64_t' to 'double', possible loss of data
-numpy/random/_generator.cp312-win32.pyd.p/numpy/random/_generator.pyx.c(38353): warning C4244: 'function': conversion from 'int64_t' to 'double', possible loss of data
-numpy/random/_generator.cp312-win_arm64.pyd.p/numpy/random/_generator.pyx.c(26329): warning C4244: 'function': conversion from 'int64_t' to 'double', possible loss of data
-numpy/random/_generator.cp312-win_arm64.pyd.p/numpy/random/_generator.pyx.c(38353): warning C4244: 'function': conversion from 'int64_t' to 'double', possible loss of data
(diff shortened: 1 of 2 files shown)


rgommers (approved):
Confirmed that this fixes the problem for SciPy.

bashtage:
Also breaks `randomgen`.  I don't think NumPy will ever directly use SeedlessSequence, but it was written with the purpose it is shows up downstream.

seberg:
Ack, that is too bad that Cython checks for it even though it never exists.

Let's probably just do this. I think the slightly better fix will be to just move this to the `.pyx` file. That way, Cython will find it at runtime for the moment but when compiling new Cython modules against this, it is gone from the `.pxd` so that in the future we can actually remove it completly.

rkern:
@bashtage `SeedlessSequence` was always a typo in the `.pxd` and was removed because the corresponding class didn't actually exist. `SeedlessSeedSequence` is the real class that was missing from the `.pxd` because of the typo.

In fixing this, we just didn't realize that downstream extension modules that linked in `np.random`'s Cython API would always access everything declared in the `.pxd` it saw at buildtime even if they didn't actually use it at runtime (because, it doesn't actually exist).

charris (approved):

jorenham:
> Let's probably just do this. I think the slightly better fix will be to just move this to the `.pyx` file. That way, Cython will find it at runtime for the moment but when compiling new Cython modules against this, it is gone from the `.pxd` so that in the future we can actually remove it completly.

That's something we could always do at a later point, no?

charris:
> in the future we can actually remove it completly.

That would be nice, we don't want to get stuck with an immortal typo.

seberg:
> That's something we could always do at a later point, no?

Yes, although the earlier we actually do it, the earlier the time comes when we _technically_ might break someone, but in practice nobody will notice that when we remove it :).

jorenham:
I'm up for trying it if no one objects

rgommers:
I would recommend against trying that for 2.4.1 - always best to do a straight revert, and do anything else in `main`.

seberg:
Let's just put it in, I think 2 lines with a clear way to test that it works doesn't matter for backporting, but 2.4.1 is basically identical anyway (normally I would prefer to not do this type of thin in a 2.4.x release though, while I don't think there would be a risk here).

But this whole thing is so risk free either way...

jorenham:
Ok, thanks for the help, all :)

maresb:
I just hit this, and so was wondering what happened with the plan to yank and release 2.4.1? Thanks!!!

jorenham:
> what happened with the plan to yank and release 2.4.1?

As far as I'm aware, that's still the plan.

rgommers:
@charris thanks for the 2.4.1 release! Are you planning to yank 2.4.0? Between the incoming links to this issue and https://github.com/scipy/scipy/issues/24215, this clearly affects quite a few packages, and it's the first accidental ABI break in quite a while I believe, so yanking is the right to do. There shouldn't be any downsides to doing that - anyone who used `== '2.4.0'` will continue to have access to that version.

charris:
@rgommers I'll yank 2.4.0,  but I am not convinced it will make much difference.

charris:
Yanked.

rgommers:
thanks!

PR #30502, opened 2025-12-22
MNT: use if constexpr for compile-time branch selection

r-devulap:
Closes  #30452


charris:
Ah, that's better. Updating all those submodules was a bit much :)

charris (approved):

r-devulap:
> Ah, that's better. Updating all those submodules was a bit much :)

goofed up git rebase, rusty after a long break :)

charris:
Thanks @r-devulap .

PR #30510, opened 2025-12-23
ENH: TSan pixi package definitions

ngoldbaum:
Followup for https://github.com/numpy/numpy/pull/30381.

Marked as draft for now because I'm running into https://github.com/prefix-dev/rattler-build/issues/2012 on my local development environment. I'm curious what happens on CI. We should also wait for the upstream CPython PR to be merged: https://github.com/python/cpython/pull/142872.

Also TODO: 
* update readme in `pixi-packages` directory.


pixi-packages/tsan-free-threading/pixi.toml
```diff
@@ -0,0 +1,26 @@
...
+[package.build.config]
+extra-input-globs = ["**/*.c.src"]
+compilers = ["c", "cxx"]
+env.TSAN_OPTIONS = "halt_on_error=0:allocator_may_return_null=1"
+extra-args = ["-Csetup-args=-Db_sanitize=thread", "-Csetup-args=-Dbuildtype=debug"]
+
+[package.host-dependencies]
+python.git = "https://github.com/crusaderky/cpython"
+python.subdirectory = "Tools/pixi-packages/tsan"
+python.rev = "eb0149f0b830651bffe757126f9d699859aa0fda"
```
crusaderky:
```suggestion
python.subdirectory = "Tools/pixi-packages/tsan-free-threading"
python.rev = "9a2a8963286077536f67b89a2123355aeb049a3d"
```

crusaderky:
https://github.com/ngoldbaum/numpy/pull/4

crusaderky:
One annoying side effect I'm observing (as of my PR linked above) is that this step takes 13 minutes on `tsan-freethreading`, vs. 1 minute on all other variants:
>  │ │ Compiling 405 .py files to .pyc

I get _why_ it takes so much longer, and I think it's definitely not worth adding (a lot of) extra complexity to somehow do the Cython compilation with python TSAN and the pyc compilation with regular python. Just wanted to flag it.

rgommers:
While Cython has the same issue, the `.pyc` compilation is Python bytecode, not Cython files. That seems unavoidable.

rgommers:
Just to make sure: you're using the regular (not debug) CPython build? If debug is enabled, it probably makes sense to turn it off - it's too slow, so probably only makes sense to use once one is working on CPython itself.

ngoldbaum:
We probably want `-g -O2`. That might need to be manually set in the cpython recipe via e.g. CFLAGS. I don't think Python's configure script has an option to get a debugoptimized build.

rgommers:
> We probably want `-g -O2`.

I was saying that you most likely _don't_ want `-g` for CPython by default when working on other projects (especially higher up in the stack like @crusaderky is for h5py), because it's so much slower. When doing a first pass over a new project, the interesting problems won't be helped much by using a debug build of CPython. If it seems likely that the problem is in CPython itself, crafting a reproducer and giving it to someone with CPython expertise (or then using `-g` if that seems helpful) who then does use that debug build is of course useful.

ngoldbaum:
> I was saying that you most likely don't want -g for CPython by default when working on other projects (especially higher up in the stack like @crusaderky is for h5py), because it's so much slower. When doing a first pass over a new project, the interesting problems won't be helped much by using a debug build of CPython. If it seems likely that the problem is in CPython itself, crafting a reproducer and giving it to someone with CPython expertise (or then using -g if that seems helpful) who then does use that debug build is of course useful.

Are you perhaps conflating `-g` in CFLAGS with [`--with-pydebug`](https://docs.python.org/3/using/configure.html#cmdoption-with-pydebug)? The latter implies the former but the former doesn't imply the latter.

I think we do want CPython builds with debug symbols, otherwise debugging issues that end up in CPython will be more difficult. I also don't think compiling with debug symbols adds much runtime overhead, it's all the Py_DEBUG checks that do that.

rgommers:
Ah yes, that makes sense, sorry for the noise.

crusaderky:
I mean, these recipes will only ever be used for CI, early testing and troubleshooting, so it makes sense that they all have `-g` enabled. I don't think anybody will care here about a 20% performance loss, except maybe people running benchmarks (at which point it's easy to just do a quick fork and change the flags).

crusaderky:
For the sake of test velocity and thoroughness, I've created https://github.com/crusaderky/cpython-pixi

crusaderky:
@isuruf just fixed Python on macos-15-intel. However, I'm now observing a failure building NumPy on the same VMs:

https://github.com/crusaderky/hdf5-pixi/actions/runs/21248797495/job/61144753910?pr=10
```
   │ │       ld: warning: dylib
   │ │       ($PREFIX/lib/libc++.dylib)
   │ │       was built for newer macOS version (10.13) than being linked (10.9)
   │ │       ld: warning: dylib
   │ │       ($BUILD_PREFIX/lib/clang/21/lib/darwin/libclang_rt.asan_osx_dynamic.dylib)
   │ │       was built for newer macOS version (10.13) than being linked (10.9)
   │ │       ld: fatal warning(s) induced error (-fatal_warnings)
   │ │       x86_64-apple-darwin13.4: error: linker command failed with exit code 1
   │ │       (use -v to see invocation)
   │ │       -----------
   │ │       ERROR: Linker x86_64-apple-darwin13.4.0-clang++ does not support
   │ │       sanitizer arguments ['-fsanitize=address']
```

linux-latest, linux-aarch64, and macos-latest work fine.
For clarity, MacOS Intel support is out of scope for this PR and I'd be happy to leave it as a follow-up.

isuruf:
You can change the target. `export MACOSX_DEPLOYMENT_TARGET=$(xcrun --show-sdk-platform-version)`

crusaderky:
> You can change the target. `export MACOSX_DEPLOYMENT_TARGET=$(xcrun --show-sdk-platform-version)`

I don't think you can set an env variable to the output of a script with pixi-build-python? 
https://prefix-dev.github.io/pixi-build-backends/backends/pixi-build-python/#env

crusaderky:
https://github.com/ngoldbaum/numpy/pull/5

pixi-packages/asan/pixi.toml
```diff
@@ -17,9 +17,9 @@ env.ASAN_OPTIONS = "detect_leaks=0:symbolize=1:strict_init_order=true:allocator_
 extra-args = ["-Csetup-args=-Db_sanitize=address", "-Csetup-args=-Dbuildtype=debug"] 
 
 [package.host-dependencies]
-python.git = "https://github.com/python/cpython"
+python.git = "https://github.com/crusaderky/cpython"
```
lucascolley:
```suggestion
python.git = "https://github.com/python/cpython"
```

pixi-packages/asan/pixi.toml
```diff
@@ -17,9 +17,9 @@ env.ASAN_OPTIONS = "detect_leaks=0:symbolize=1:strict_init_order=true:allocator_
...
 python.subdirectory = "Tools/pixi-packages/asan"
-python.rev = "8bb5b6e8ce61da41b5affd5eb12d9cc46b5af448"
+python.rev = "tsan"
```
lucascolley:
```suggestion
```

or maybe we do want to pin to a rev

crusaderky: I don't see a strong benefit in pinning?

crusaderky:
https://github.com/ngoldbaum/numpy/pull/6

pixi-packages/asan/pixi.toml
lucascolley: let's go without it for now 👍

lucascolley:
Just FYI I reported to the Pixi team a deadlock that can be experienced when an environment tries to depend on a CPython package both directly and via a NumPy package here. Possibly the problem will go away once we no longer pin to a branch but just a heads up that it is being investigated in case anyone hits it.

crusaderky:
> Just FYI I reported to the Pixi team a deadlock that can be experienced when an environment tries to depend on a CPython package both directly and via a NumPy package here. Possibly the problem will go away once we no longer pin to a branch but just a heads up that it is being investigated in case anyone hits it.

I've never experienced it; could you link your report?

lucascolley:
> could you link your report?

not public yet, once this PR is updated I will update the SciPy PR to use this and CPython main and see if it reproduces

crusaderky:
I'm seeing again 
```

   ╰─────────────────── (took 24 minutes)
  Error:   × failed to build 'numpy' from 'https://github.com/ngoldbaum/numpy@d8c5c2e52996bdde9fc25c0ba6234538e570f2c2 (subdir: pixi-packages/tsan-freethreading) (ref: tsan-ft-pixi)'
    ╰─▶   × No license files were copied
```
can't understand why it's back
REF https://github.com/prefix-dev/pixi/issues/5424

crusaderky:
Opened https://github.com/prefix-dev/pixi/issues/5557

ngoldbaum:
@crusaderky see https://github.com/numpy/numpy/issues/30888#issuecomment-3968945004

crusaderky:
> can you try removing the `rev` pin in [gh-30510](https://github.com/numpy/numpy/pull/30510) [@ngoldbaum](https://github.com/ngoldbaum) ? The workaround for [#30510 (comment)](https://github.com/numpy/numpy/pull/30510#issuecomment-3957465659) is to copy `LICENSE.txt` alongside the `pixi.toml` and add
> 
> [package]
> license-file = "LICENSE.txt"

This works, thank you @lucascolley 

@ngoldbaum https://github.com/ngoldbaum/numpy/pull/7 (also please merge from main)

ngoldbaum:
FWIW, the x86_64 TSAN pixi build is the slowest CI job in this PR, weighing in at 34 minutes. IMO that's a little too slow for this to be mergeable and run on every PR.

That said, the Ubuntu ARM64 build is a lot faster. Do we really need to do these pixi builds on so many different architectures?

crusaderky:
given how we very recently fixed macos-specific issues, and how macos intel is still currently broken, I would say yes. But I understand how most PR submitted don't care about this, nor carry a risk to break it. So maybe make it nightly + opt-in (with a tag?)

rgommers:
> So maybe make it nightly + opt-in (with a tag?)

We don't have/do that, that's more complexity than it's worth. I'd say keep as is, or just don't add it.

ngoldbaum:
I brought this up at the triage meeting today. There wasn't much appetite for adding expensive new CI jobs that won't benefit NumPy directly.

There was some appetite for adding some kind of cron testing. There are probably other jobs that don't need to run on *every* PR. If we did that, we'd also want some kind of automation that opens issues when cron tests fail, with details about the failure.

This is a decent-sized infrastructure project and this is in no way a request for you to do that to make this PR mergeable.

Given that downstream libraries want to use these recipes and not NumPy itself, the participants in the triage meeting seemed to agree that simply not testing them is also fine, since that downstream users will be motivated to send in fixes. For our own CI, the `cpython_sanity` docker images work fine, so I doubt we'd ever use this in CI.

rgommers:
> There was some appetite for adding some kind of cron testing. There are probably other jobs that don't need to run on _every_ PR. If we did that, we'd also want some kind of automation that opens issues when cron tests fail, with details about the failure.

I have very little interest in that. Whenever we did that in the past, failures weren't handled. We should focus on optimizing jobs instead if needed (caching, only running subsets of tests, etc.). Running cron jobs, and especially fragile ones, isn't attractive.

rgommers:
> simply not testing them is also fine, 

That's fine indeed

crusaderky:
> Whenever we did that in the past, failures weren't handled. We should focus on optimizing jobs instead if needed

I think there is a possible high value, mid-cost optimization here, which would benefit absolutely everyone in the free-threading ecosystem well beyond TSAN.

The TSAN build spends the vast majority of its time sequentially compiling .py files to .pyc. This is very slow, because the very code that compiles them is dragged down by TSAN.

Earlier in this thread I mentioned how it was thinkable to use a bespoke system to compile .pyc with a stock python interpreter and .so with the TSAN, but it would be a very substantial feat of engineering, and probably not worth the effort.

However, what if the build system (`meson-python`? `build`?) was changed to compile .py files to .pyc _in parallel_ using all available threads if it detects a free-threading interpreter? (why is it serial to begin with? the compiler shouldn't use CPython API anyway? Could it be parallel for _all_ python interpreters?)
This is clearly non-trivial as it opens up to race conditions, but I think it's a worthy exercise as it could potentially benefit every and all users that install a package from source tarballs.

Clearly, such an exercise would be well beyond the scope of this PR, but if it was the agreed way forward we could temporarily disable CI here and come back to it when it's ready.

ngoldbaum:
@crusaderky I think bytecode compilation is already parallelized: 

https://github.com/python/cpython/pull/99291
https://docs.python.org/3/library/compileall.html#cmdoption-compileall-j

You can try overriding `COMPILEALL_OPTS` to see if it has any impacts on local builds. Maybe something is broken in CPython's build configuration if it isn't actually parallelizing anything?

lucascolley:
shall we disable CI runs that are too slow and merge this, then?

rgommers:
+1 to merge with too expensive job(s) disabled.

Re byte-compiling in parallel, I think @crusaderky has a good idea. The links @ngoldbaum posted are for CPython itself, but don't look like they affect Meson. 

What Meson does is to collect all `.py` files that need to be byte-compiled, and put them in a single JSON manifest:

https://github.com/mesonbuild/meson/blob/64de45a99412c9145e5dece692d9da19bb268d11/mesonbuild/modules/python.py#L439-L444

and then those are compiled sequentially in a for-loop here:

https://github.com/mesonbuild/meson/blob/64de45a99412c9145e5dece692d9da19bb268d11/mesonbuild/scripts/pycompile.py#L14-L15

This is also the slowest step if you invoke `spin build` twice, if you bump up the verbosity you'll see the sequential force-recompiles:
```
...
Compiling '/Users/rgommers/code/pixi-dev/numpy/numpy/build/../build-install/usr/lib/python3.14/site-packages/numpy/_core/tests/examples/cython/setup.py'...
Compiling '/Users/rgommers/code/pixi-dev/numpy/numpy/build/../build-install/usr/lib/python3.14/site-packages/numpy/_core/tests/examples/limited_api/setup.py'...
```

For `spin`, where we invoke `meson install` directly, we should probably turn off recompiling completely, that will reduce the overhead a lot.

For a regular wheel build via `meson-python` though, byte-compiling isn't happening at all, because it's not calling `meson install` anymore. Instead, it's done by either the installer (`pip`/`uv`) or by the import system. See https://github.com/mesonbuild/meson-python/issues/706.

`pip`'s default is to byte-compile everything, and that can be turned off with the `--no-compile` option. Using that option will be a lot faster for TSan build envs.

`uv` skips byte-compiling by default though, which is faster. So if using `uv`, there's nothing left to optimize, byte-compiling happens only on demand at import time. See https://pythonspeed.com/articles/faster-pip-installs/.

Given that `meson install` is used pretty heavily in regular dev workflows, it seems worth trying to speed it up. It takes 2-3 seconds for NumPy for me, and 4-6 seconds for SciPy. It's currently the main overhead in running `spin build`. And most users won't go explicitly turn off byte-compilation.  One option would be to use a `ThreadPoolExecutor`, which will help only a little on a GIL-enabled interpreter, but should give a linear speedup on a free-threaded interpreter. And is probably the minimally invasive change, since a `ProcessPoolExecutor` would be much heavier and switching from `compile_file` to `compile_dir` would need to change the logic a fair amount. Collecting all files in a single list and then parallelizing over that is probably also the most performant option.

I did a quick test, adding `-Dpython.bytecompile=-1` to `meson setup` in the `spin` config. It reduces the time for a clean rebuild/install with `pixi r build` from 2.7 sec. to 1.0 sec. for me (on an M1 Macbook). That's a pretty good change to make! On a tight test loop including pytest test collection it's still significant: `pixi r test -- numpy/fft/` goes from 3.7 sec. to 2.0 sec.

@eli-schwartz given you wrote the byte-compile logic fairly recently, I'd be interested in hearing your thoughts.

lucascolley:
> uv skips byte-compiling by default though, which is faster. So if using uv, there's nothing left to optimize, byte-compiling happens only on demand at import time. See [pythonspeed.com/articles/faster-pip-installs](https://pythonspeed.com/articles/faster-pip-installs/).

On line 22134 of https://github.com/numpy/numpy/actions/runs/22587625551/job/65436838441?pr=30510#step:4:22135 I see

```
Compiling 406 .py files to .pyc
```

despite us using `uv` to invoke `mesonpy.build_wheel`.

lucascolley:
ah no, it's `rattler-build` doing the byte-code compilation in the "Packaging new files" step. There is an option to skip that at https://rattler-build.prefix.dev/latest/build_options/#python-options, but I'm not sure if that option is available via `pixi-build-python`, or indeed whether Pixi is happy to compile at import-time like `uv`.

eli-schwartz:
@rgommers I think it sounds fine to improve this with a ThreadPoolExecutor or a ProcessPoolExecutor (whichever one you think is faster in practice), and it's a very logical change to make.

I notice that the stdlib compileall uses a ProcessPoolExecutor by the way. (We aren't using the stdlib here because the stdlib has a long history of changing how compileall works, including options that don't actually work as expected on all versions of python that claim to support the option, but -- reimplementing the same thing ourselves should be relatively trivial and also portable.)

ngoldbaum:
I disabled the slow builds.

lucascolley:
(can follow-up on the byte code compilation once https://github.com/prefix-dev/pixi/pull/5737 is merged and released)

lucascolley:
I see in the logs locally and in the removed job (https://github.com/numpy/numpy/actions/runs/22587625551/job/65436838388?pr=30510#step:4:20648):

```
 │ │ DEBUG ../numpy/random/meson.build:177: WARNING: Trying to use ['thread'] sanitizer on Clang with b_lundef.
 │ │ DEBUG This will probably not work.
 │ │ DEBUG Try setting b_lundef to false instead.
```

is that expected @crusaderky @ngoldbaum ?

lucascolley:
The following diff successfully skips byte code compilation for me, using https://github.com/prefix-dev/pixi/pull/5737 from source:

```diff
diff --git a/pixi-packages/tsan-freethreading/pixi.toml b/pixi-packages/tsan-freethreading/pixi.toml
index 8c12c3c7a8..f70831b22b 100644
--- a/pixi-packages/tsan-freethreading/pixi.toml
+++ b/pixi-packages/tsan-freethreading/pixi.toml
@@ -12,13 +12,14 @@ source.path = "../.."

 [package.build.backend]
 name = "pixi-build-python"
-version = "*"
+path = "/Users/lucascolley/ghq/github.com/prefix-dev/pixi/crates/pixi_build_python"

 [package.build.config]
 extra-input-globs = ["**/*.c.src"]
 compilers = ["c", "cxx"]
 env.TSAN_OPTIONS = "halt_on_error=0:allocator_may_return_null=1"
 extra-args = ["-Csetup-args=-Db_sanitize=thread", "-Csetup-args=-Dbuildtype=debug"]
+skip-pyc-compilation = ["**/*.py"]

 [package.host-dependencies]
 python.git = "https://github.com/python/cpython"
```

ngoldbaum:
> is that expected ?

Yeah, I'm not 100% sure what that warning is getting at, but I've never run into issues with undefined symbols at link time. Maybe it has something to do with combining sanitizer and "normal" builds of libraries and relying on `LD_PRELOAD` to ensure the sanitizer runtime libraries are loaded? In either case, these builds don't hit linker errors.

I'll try your patch here as well to see how it helps the build times.

ngoldbaum:
Oh wait, I didn't see that you're using a PR branch. I'll just add a comment saying we should add this when that PR is available.

ngoldbaum:
I'll merge this tomorrow unless anyone thinks we should wait for whatever reason.

crusaderky:
> Given that `meson install` is used pretty heavily in regular dev workflows, it seems worth trying to speed it up. It takes 2-3 seconds for NumPy for me, and 4-6 seconds for SciPy. It's currently the main overhead in running `spin build`. And most users won't go explicitly turn off byte-compilation. 

On your local box, when you run `spin build` shouldn't you benefit from `__pycache__` from the previous build? Until you explicitly delete the `build` directory of course. Whereas on CI it start from scratch every time. Local rattler-build seems to start from scratch every time as well (maybe a missing feature in pixi?)

Issue #30494, opened 2025-12-21
BUG: Multithreaded scaling bottlenecks on free-threaded Python


ngoldbaum:
[This stackoverflow question](https://stackoverflow.com/questions/79851420/multithreading-becomes-much-slower-than-multiprocessing-in-free-threaded-python#79851420) seems to have identified two real scaling bottlenecks inside NumPy.

I build Python 3.15t from source using pyenv:

```
$ CFLAGS=-O2 pyenv install 3.15t-dev -k -g
...
$ pyenv global 3.15t-dev-debug
```

Then, in an up-to-date clone of the NumPy repo:

```
python -m pip install . -v -Cbuilddir=build -C'setup-args=-Dbuildtype=debugoptimized'
```

And finally run this script, which has to be named `mtmp.py`: https://gist.github.com/ngoldbaum/ff6428e2510e991247eba4c809dcc8ac. This is identical to what was posted on StackOverflow but with some code commented out.

On my M3 Macbook Pro, I get the following stdout running the script:

```
Inner loops 10, multithreading  time: 6.68 sec, result sum: 717434683.1879175
Inner loops 10, multiprocessing time: 4.86 sec, result sum: 717434683.1879175
```

If I run the script like so:

```
PYTHONPERFSUPPORT=1 samply record $(pyenv which python) mtmp.py
```

Then I get a profile output like this: https://share.firefox.dev/3Lde1Lm. I see two different scaling bottlenecks. First, there seem to be a mutex inside tracemalloc that has contention.

Second, the critical section I added to the array creation routines in https://github.com/numpy/numpy/pull/29394 is causing some contention in this case.

ngoldbaum:
If you're unfamiliar with how to read firefox profiles, here's a flame graph from one of the thread pool threads in the profile, which makes it clearer that most of the time is being spent waiting to acquire PyMutex locks in tracemalloc and in the array creation routines. 

<img width="1468" height="473" alt="Image" src="https://github.com/user-attachments/assets/23e964e4-0ed7-4fb7-a9a8-304e9eecee39" />

chris-eibl:
Here are numbers for Python 3.14.2 using [Python install manager](https://www.python.org/ftp/python/pymanager/python-manager-25.2.msix) and `numpy-2.4.0-cp314-cp314t-win_amd64.whl` on my 4 core Haswell i5-4570 on Windows 10:

`n_core = 1:`

```
[...]

PR #30514, opened 2025-12-24
ENH: use more fine-grained critical sections in array coercion internals

ngoldbaum:
Towards addressing #30494. Right now the critical section I remove here in `PyArray_FromAny_int` shows up in the profile for the script in that issue.

I added the critical section in 5a031d9b8822551711973e6ae220cd00817c9523 and this partially reverts that change. On reflection, it's not a good idea to introduce a scaling bottleneck here in service of a sort of wonky thing to do: mutating the operand of `np.array()` while the array is being created.

Instead, we should error in those cases. Like we already do without the critical section!

I also needed to add new, more fine-grained critical sections, in `PyArray_DiscoverDTypeAndShape_Recursive` and `PyArray_AssignFromCache_Recursive` to avoid data races due to use of the `PySequence_Fast` API.

I also updated the tests for this to allow the cases affected by this to raise errors instead of succeeding, since that success relies on introducing scaling bottlenecks for valid read-only uses.

Here's a Samply profile output run using this PR on the script from #30494 - I no longer see a scaling bottleneck inside the array coercion routines: https://share.firefox.dev/44KLZxs

### DIFF
--- a/numpy/_core/src/common/npy_pycompat.h
+++ b/numpy/_core/src/common/npy_pycompat.h
@@ -34,7 +34,7 @@
         }                                                               \
     }
 #else
-#define NPY_BEGIN_CRITICAL_SECTION_SEQUENCE_FAST(original) {
+#define NPY_BEGIN_CRITICAL_SECTION_SEQUENCE_FAST(original) { do { (void)(original); } while (0)
 #define NPY_END_CRITICAL_SECTION_SEQUENCE_FAST() }
 #endif
 
--- a/numpy/_core/src/multiarray/array_coercion.c
+++ b/numpy/_core/src/multiarray/array_coercion.c
@@ -1159,7 +1159,7 @@ PyArray_DiscoverDTypeAndShape_Recursive(
         return -1;
     }
 
-    int ret;
+    int ret = -1;
 
     NPY_BEGIN_CRITICAL_SECTION_SEQUENCE_FAST(obj);
 
@@ -1182,7 +1182,6 @@ PyArray_DiscoverDTypeAndShape_Recursive(
 
     /* Allow keyboard interrupts. See gh issue 18117. */
     if (PyErr_CheckSignals() < 0) {
-        ret = -1;
         goto finish;
     }
 
@@ -1203,7 +1202,6 @@ PyArray_DiscoverDTypeAndShape_Recursive(
                 flags, copy);
 
         if (max_dims < 0) {
-            ret = -1;
             goto finish;
         }
     }
--- a/numpy/_core/src/multiarray/array_coercion.c
+++ b/numpy/_core/src/multiarray/array_coercion.c
@@ -1209,7 +1209,7 @@ PyArray_DiscoverDTypeAndShape_Recursive(
     }
     ret = max_dims;
 
-  finish:
+  finish:;
 
     NPY_END_CRITICAL_SECTION_SEQUENCE_FAST();
 
--- a/numpy/_core/src/multiarray/array_coercion.c
+++ b/numpy/_core/src/multiarray/array_coercion.c
@@ -1159,6 +1159,10 @@ PyArray_DiscoverDTypeAndShape_Recursive(
         return -1;
     }
 
+    int ret;
+
+    NPY_BEGIN_CRITICAL_SECTION_SEQUENCE_FAST(obj);
+
     npy_intp size = PySequence_Fast_GET_SIZE(seq);
     PyObject **objects = PySequence_Fast_ITEMS(seq);
 
@@ -1166,17 +1170,20 @@ PyArray_DiscoverDTypeAndShape_Recursive(
                      out_shape, 1, &size, NPY_TRUE, flags) < 0) {
         /* But do update, if there this is a ragged case */
         *flags |= FOUND_RAGGED_ARRAY;
-        return max_dims;
+        ret = max_dims;
+        goto finish;
     }
     if (size == 0) {
         /* If the sequence is empty, this must be the last dimension */
         *flags |= MAX_DIMS_WAS_REACHED;
-        return curr_dims + 1;
+        ret = curr_dims + 1;
+        goto finish;
     }
 
     /* Allow keyboard interrupts. See gh issue 18117. */
     if (PyErr_CheckSignals() < 0) {
-        return -1;
+        ret = -1;
+        goto finish;
     }
 
     /*
@@ -1196,10 +1203,17 @@ PyArray_DiscoverDTypeAndShape_Recursive(
                 flags, copy);
 
         if (max_dims < 0) {
-            return -1;
+            ret = -1;
+            goto finish;
         }
     }
-    return max_dims;
+    ret = max_dims;
+
+  finish:
+
+    NPY_END_CRITICAL_SECTION_SEQUENCE_FAST();
+
+    return ret;
 }
 
 
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -1571,8 +1571,6 @@ PyArray_FromAny_int(PyObject *op, PyArray_Descr *in_descr,
         copy = 1;
     }
 
-    Py_BEGIN_CRITICAL_SECTION(op);
-
     ndim = PyArray_DiscoverDTypeAndShape(
             op, max_depth, dims, &cache, in_DType, in_descr, &dtype,
             copy, &was_copied_by__array__);
@@ -1741,7 +1739,6 @@ PyArray_FromAny_int(PyObject *op, PyArray_Descr *in_descr,
 cleanup:;
 
     Py_XDECREF(dtype);
-    Py_END_CRITICAL_SECTION();
     return (PyObject *)ret;
 }
 
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -519,7 +519,7 @@ PyArray_AssignFromCache_Recursive(
     else {
         assert(depth != ndim);
         npy_intp orig_length = PyArray_DIMS(self)[0];
-        int err = 0;
+        int err = 1;
         NPY_BEGIN_CRITICAL_SECTION_SEQUENCE_FAST(orig_seq);
         for (npy_intp i = 0; i < orig_length; i++) {
             // this macro takes *the argument* of PySequence_Fast, which is orig_seq;
@@ -531,7 +531,6 @@ PyArray_AssignFromCache_Recursive(
                 PyErr_SetString(PyExc_RuntimeError,
                                 "Inconsistent object during array creation? "
                                 "Content of sequences changed (length inconsistent).");
-                err = 1;
                 goto finish_critical_section;
             }
             else {
@@ -550,7 +549,6 @@ PyArray_AssignFromCache_Recursive(
                 char *item;
                 item = (PyArray_BYTES(self) + i * PyArray_STRIDES(self)[0]);
                 if (PyArray_Pack(PyArray_DESCR(self), item, item_pyvalue) < 0) {
-                    err = 1;
                     goto finish_critical_section;
                 }
                 /* If this was an array(-like) we still need to unlike int: */
@@ -562,24 +560,22 @@ PyArray_AssignFromCache_Recursive(
                 PyArrayObject *view;
                 view = (PyArrayObject *)array_item_asarray(self, i);
                 if (view == NULL) {
-                    err = 1;
                     goto finish_critical_section;
                 }
                 if (PyArray_AssignFromCache_Recursive(view, ndim, cache) < 0) {
                     Py_DECREF(view);
-                    err = 1;
                     goto finish_critical_section;
                 }
                 Py_DECREF(view);
             }
         }
+    err = 0;
       finish_critical_section:;
 
         NPY_END_CRITICAL_SECTION_SEQUENCE_FAST();
         if (err) {
             goto finish;
         }
-
 
     }
     ret = 0;
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -562,7 +562,8 @@ PyArray_AssignFromCache_Recursive(
             }
         }
 
-    finish_section:
+    finish_section:;
+
         NPY_END_CRITICAL_SECTION_SEQUENCE_FAST();
     }
     Py_DECREF(obj);
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -495,9 +495,11 @@ PyArray_AssignFromCache_Recursive(
 {
     /* Consume first cache element by extracting information and freeing it */
     PyObject *obj = (*cache)->arr_or_sequence;
+    PyObject *orig_obj = (*cache)->converted_obj;
     Py_INCREF(obj);
     npy_bool sequence = (*cache)->sequence;
     int depth = (*cache)->depth;
+    int ret = 0;
     *cache = npy_unlink_coercion_cache(*cache);
 
     /* The element is either a sequence, or an array */
@@ -510,14 +512,17 @@ PyArray_AssignFromCache_Recursive(
     }
     else {
         assert(depth != ndim);
+        // this macro thakes *the argument* of PySequence_Fast, which is orig_obj;
+        // not the object returned by PySequence_Fast, which is obj
+        NPY_BEGIN_CRITICAL_SECTION_SEQUENCE_FAST(orig_obj)
         npy_intp length = PySequence_Length(obj);
         if (length != PyArray_DIMS(self)[0]) {
             PyErr_SetString(PyExc_RuntimeError,
-                    "Inconsistent object during array creation? "
-                    "Content of sequences changed (length inconsistent).");
-            goto fail;
+                            "Inconsistent object during array creation? "
+                            "Content of sequences changed (length inconsistent).");
+            ret = -1;
+            goto finish_section;
         }
-
         for (npy_intp i = 0; i < length; i++) {
             PyObject *value = PySequence_Fast_GET_ITEM(obj, i);
 
@@ -533,7 +538,8 @@ PyArray_AssignFromCache_Recursive(
                 char *item;
                 item = (PyArray_BYTES(self) + i * PyArray_STRIDES(self)[0]);
                 if (PyArray_Pack(PyArray_DESCR(self), item, value) < 0) {
-                    goto fail;
+                    ret = -1;
+                    goto finish_section;
                 }
                 /* If this was an array(-like) we still need to unlike int: */
                 if (*cache != NULL && (*cache)->converted_obj == value) {
@@ -544,18 +550,23 @@ PyArray_AssignFromCache_Recursive(
                 PyArrayObject *view;
                 view = (PyArrayObject *)array_item_asarray(self, i);
                 if (view == NULL) {
-                    goto fail;
+                    ret = -1;
+                    goto finish_section;
                 }
                 if (PyArray_AssignFromCache_Recursive(view, ndim, cache) < 0) {
                     Py_DECREF(view);
-                    goto fail;
+                    ret = -1;
+                    goto finish_section;
                 }
                 Py_DECREF(view);
             }
         }
+
+    finish_section:
+        NPY_END_CRITICAL_SECTION_SEQUENCE_FAST();
     }
     Py_DECREF(obj);
-    return 0;
+    return ret;
 
   fail:
     Py_DECREF(obj);
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -520,7 +520,7 @@ PyArray_AssignFromCache_Recursive(
         assert(depth != ndim);
         npy_intp orig_length = PyArray_DIMS(self)[0];
         for (npy_intp i = 0; i < orig_length; i++) {
-            int critical_section_error = 0;
+            int err = 0;
             // this macro takes *the argument* of PySequence_Fast, which is orig_seq;
             // not the object returned by PySequence_Fast, which is obj
             NPY_BEGIN_CRITICAL_SECTION_SEQUENCE_FAST(orig_seq);
@@ -529,14 +529,14 @@ PyArray_AssignFromCache_Recursive(
                 PyErr_SetString(PyExc_RuntimeError,
                                 "Inconsistent object during array creation? "
                                 "Content of sequences changed (length inconsistent).");
-                critical_section_error = 1;
+                err = 1;
             }
             else {
                 Py_XSETREF(item_pyvalue, Py_NewRef(PySequence_Fast_GET_ITEM(obj, i)));
             }
             NPY_END_CRITICAL_SECTION_SEQUENCE_FAST();
 
-            if (critical_section_error) {
+            if (err) {
                 goto finish;
             }
 
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -522,7 +522,9 @@ PyArray_AssignFromCache_Recursive(
         for (npy_intp i = 0; i < orig_length; i++) {
             int err = 0;
             // this macro takes *the argument* of PySequence_Fast, which is orig_seq;
-            // not the object returned by PySequence_Fast, which is obj
+            // not the object returned by PySequence_Fast, which is a proxy object
+            // with its own per-object PyMutex lock.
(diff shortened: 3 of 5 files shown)


numpy/_core/tests/test_multithreading.py
```diff
@@ -319,9 +319,17 @@ def create_nditer(arrs):
...
-def test_arg_locking(kernel):
-    # should complete without failing or generating an error about an array size
+
+@pytest.mark.parametrize(
+    "kernel, outcome",
+    (
+        (np_broadcast, "error"),
+        (create_array, "error"),
+        (create_nditer, "success")
+    )
```
jorenham: micro-nit

```suggestion
        (create_nditer, "success"),
    ),
```

numpy/_core/tests/test_multithreading.py
```diff
@@ -319,9 +319,17 @@ def create_nditer(arrs):
...
+)
+def test_arg_locking(kernel, outcome):
+    # should complete without triggering races but may error
     # changing
```
jorenham:
```suggestion
```

jorenham (commented):
Makes sense :ok_hand: 

I picked some nits

ngoldbaum:
Ah there's a lower-level use of PySequence_Fast inside of `PyArray_DiscoverDTypeAndShape_Recursive` that needs a critical section. Let's see if that fixes the races here. And maybe taking many shorter-lived critical sections will scale better. 

If that doesn't fix the races, maybe we shouldn't be using `PySequence_Fast` at all in that code path - is it really so important to not use the regular C API and get strong references to objects while we iterate through a sequence?

ngoldbaum:
Ping @kumaraditya303 - I'd appreciate your thoughts on this.

numpy/_core/src/common/npy_pycompat.h
```diff
@@ -34,7 +34,7 @@
         }                                                               \
     }
 #else
-#define NPY_BEGIN_CRITICAL_SECTION_SEQUENCE_FAST(original) {
+#define NPY_BEGIN_CRITICAL_SECTION_SEQUENCE_FAST(original) { do { (void)(original); } while (0)
```
ngoldbaum, line 37: needed to suppress unused variable warnings - maybe there's a nicer way to spell this?

ngoldbaum:
I added the backport candidate label and milestoned this for 2.4.1 because it fixes a scaling issue that people will run into. All these critical sections are no-ops on the GIL-enabled build too.

numpy/_core/src/multiarray/ctors.c
```diff
@@ -510,14 +512,17 @@ PyArray_AssignFromCache_Recursive(
     }
     else {
         assert(depth != ndim);
+        // this macro thakes *the argument* of PySequence_Fast, which is orig_obj;
```
ngoldbaum:
```suggestion
        // this macro takes *the argument* of PySequence_Fast, which is orig_obj;
```

numpy/_core/src/multiarray/ctors.c
```diff
@@ -496,8 +496,10 @@ PyArray_AssignFromCache_Recursive(
     /* Consume first cache element by extracting information and freeing it */
     PyObject *obj = (*cache)->arr_or_sequence;
     Py_INCREF(obj);
+    PyObject *orig_obj = (*cache)->converted_obj;
```
mhvk: I'd call this differently, maybe `orig_seq` -- `orig_obj` seems super confusing in this context -- even more since `converted_obj` is assigned to it.

mhvk (commented):
Yes, generally would seem wise to do the critical sections as far down as possible.

Just one nit about a name, since there is a spelling error to correct anyway. But feel free to ignore.

numpy/_core/src/common/npy_pycompat.h
seberg: yeah, I am not sure there is, this seems good.

numpy/_core/src/multiarray/ctors.c
```diff
@@ -544,18 +550,24 @@ PyArray_AssignFromCache_Recursive(
...
                 if (PyArray_AssignFromCache_Recursive(view, ndim, cache) < 0) {
                     Py_DECREF(view);
-                    goto fail;
+                    ret = -1;
+                    goto finish_section;
                 }
                 Py_DECREF(view);
             }
         }
+
```
seberg: Since I guess you might do the small other nits.  I have one more:

I might slightly prefer to start with `ret = -1` as the default and add `ret = 0` before the `finish_section:` part.
(We could then use the `return ret` for the `fail:` too, although it would be nicer as `finish:`.)

seberg (approved):
Makes sense to me, thanks.  I think you can just merge, but maybe give me a ping if you want to backport it, since then it makes sense to make double sure there is no silly mistake.

numpy/_core/src/multiarray/ctors.c
```diff
@@ -510,14 +512,17 @@ PyArray_AssignFromCache_Recursive(
     }
     else {
         assert(depth != ndim);
+        // this macro thakes *the argument* of PySequence_Fast, which is orig_obj;
+        // not the object returned by PySequence_Fast, which is obj
+        NPY_BEGIN_CRITICAL_SECTION_SEQUENCE_FAST(orig_obj);
```
kumaraditya303: This can still crash and produce data races if the sequence gets mutated but the size remains same. Inside the loop, the items of sequences are accessed using borrowed references (`PySequence_Fast_GET_ITEM`) so if the critical section gets suspended and other thread sets a new object at current index then it can crash with use-after-free. 

You should be able to reproduce this case with a test which tries to create an `ndarray` from a list whose size is fixed but it getting concurrently modified by other thread with `list.__setitem__`.

ngoldbaum: To fix that would I need to use an API that returns a strong reference or could I just INCREF after getting the borrowed reference?

kumaraditya303: You can incref the item after getting the borrowed reference but it would cause a lot of reference counting contention. I don't think that would scale on free-threading builds.

ngoldbaum: I guess I'm trying to ask: how would you suggest fixing the issue?

kumaraditya303: You can try to further reduce the critical section and check the size after each iteration, so something like:
```
for (int i = 0; i = original_length; i++) {
NPY_BEGIN_CRITICAL_SECTION_SEQUENCE_FAST(orig_obj);
if (check length changed) {
    raise size changed exception
}
PyObject *value = PySequence_Fast_GET_ITEM(obj, i);
INCREF(value);
END_CRITICAL_SECTION();
// rest of loop
}
```

The strong reference is still needed though but I think with the reduced critical section, it would not cause a lot of contention.

ngoldbaum: I implemented your suggestion and added a new test but I was never able to trigger a race as you described. Do you have any idea why the test I modify in this PR doesn't trigger races using the old code that you said was buggy?

Either way, IMO it's way clearer with the critical section is as short as possible like I have it here. I also think doing the length check every iteration makes sense - I think it may be possible even on the GIL-enabled build for an unlucky thread switch to trigger issues that would be caught by the new repeated check.

ngoldbaum:
@seberg would you mind giving this another look to check the new changes to `PyArray_AssignFromCache_Recursive`? Do you still think this is OK to backport or foresee any risks or problems with doing the length checks multiple times?

numpy/_core/src/multiarray/ctors.c
```diff
@@ -493,33 +493,52 @@ NPY_NO_EXPORT int
...
-                    "Content of sequences changed (length inconsistent).");
-            goto fail;
-        }
+        npy_intp orig_length = PyArray_DIMS(self)[0];
+        for (npy_intp i = 0; i < orig_length; i++) {
+            int critical_section_error = 0;
+            // this macro takes *the argument* of PySequence_Fast, which is orig_seq;
+            // not the object returned by PySequence_Fast, which is obj
+            NPY_BEGIN_CRITICAL_SECTION_SEQUENCE_FAST(orig_seq);
+            npy_intp length = PySequence_Length(obj);
```
kumaraditya303: You can use `PySequence_Fast_GET_SIZE` here which would be faster.

numpy/_core/tests/test_multithreading.py
```diff
@@ -308,46 +308,76 @@ def func(index):
...
-    tasks.append(threading.Thread(target=mutate_list))
+    def replace_list_items(b):
+        nonlocal done
+        nonlocal arrs
+        b.wait()
+        rng = np.random.RandomState()
+        rng.seed(0x4d3d3d3)
+        while done < 4:
+            data = rng.randint(0, 1000, size=4)
+            print(data)
```
kumaraditya303:
```suggestion
```

numpy/_core/src/multiarray/ctors.c
```diff
@@ -493,33 +493,52 @@ NPY_NO_EXPORT int
...
-        npy_intp length = PySequence_Length(obj);
-        if (length != PyArray_DIMS(self)[0]) {
-            PyErr_SetString(PyExc_RuntimeError,
-                    "Inconsistent object during array creation? "
-                    "Content of sequences changed (length inconsistent).");
-            goto fail;
-        }
+        npy_intp orig_length = PyArray_DIMS(self)[0];
+        for (npy_intp i = 0; i < orig_length; i++) {
+            int critical_section_error = 0;
```
kumaraditya303: critical sections are nop on default builds and error can happen there as well, I suggest to name it generic like err

numpy/_core/tests/test_multithreading.py
```diff
@@ -308,46 +308,75 @@ def func(index):
...
+            arrs[data[0]] = data[1:]
 
-    [t.start() for t in tasks]
-    [t.join() for t in tasks]
+    for mutation_func in (replace_list_items, contract_and_expand_list):
+        b = threading.Barrier(5)
+        try:
+            with concurrent.futures.ThreadPoolExecutor(max_workers=5) as tpe:
+                tasks = [tpe.submit(read_arrs, b) for _ in range(4)]
+                tasks.append(tpe.submit(replace_list_items, b))
```
kumaraditya303: You should be passing `mutation_func` here, no?

kumaraditya303 (approved):
LGTM

ngoldbaum:
Rebased to fix the failing tests.

@seberg are you OK with this getting backported? I think it'd still be nice to do backport this even though we probably won't backport all the scaling fixes we're doing. This one fixes a new scaling issue in 2.4 so it's a regression fix in that sense.

numpy/_core/src/multiarray/array_coercion.c
```diff
@@ -1159,24 +1159,31 @@ PyArray_DiscoverDTypeAndShape_Recursive(
         return -1;
     }
 
+    int ret;
```
seberg:
```suggestion
    int ret = -1;
```
or do you trust the compiler warnings on this one?

numpy/_core/src/multiarray/ctors.c
```diff
@@ -493,33 +493,52 @@ NPY_NO_EXPORT int
...
-            PyErr_SetString(PyExc_RuntimeError,
-                    "Inconsistent object during array creation? "
-                    "Content of sequences changed (length inconsistent).");
-            goto fail;
-        }
+        npy_intp orig_length = PyArray_DIMS(self)[0];
+        for (npy_intp i = 0; i < orig_length; i++) {
+            int err = 0;
+            // this macro takes *the argument* of PySequence_Fast, which is orig_seq;
+            // not the object returned by PySequence_Fast, which is obj
```
seberg: You likely did, but can you explain why not `obj`?  If `PySequence_Fast` did convert to tuple, we'll have our own object anyway?
(Ideally in the comment, I guess.)

Also, can't we assume that a fast-sequence will not change its size if we hold a critical section on it?  (But then, I guess critical sections are not fool-proof, so maybe that is a silly thought.)

numpy/_core/src/multiarray/ctors.c
```diff
@@ -493,33 +493,52 @@ NPY_NO_EXPORT int
...
+            if (length != orig_length) {
+                PyErr_SetString(PyExc_RuntimeError,
+                                "Inconsistent object during array creation? "
+                                "Content of sequences changed (length inconsistent).");
+                err = 1;
+            }
+            else {
+                Py_XSETREF(item_pyvalue, Py_NewRef(PySequence_Fast_GET_ITEM(obj, i)));
+            }
+            NPY_END_CRITICAL_SECTION_SEQUENCE_FAST();
```
seberg: I am a bit confused is it really important for the relevant use-case to unlock the criticial section on every iteration?

seberg (approved):
I am slightly curious how this affects the uncontested case of:
```
l = [0] * 100_000
%timeit np.array(l)
```
but I am happy to trust you that this seems the better trade-off.  If the sequence doesn't contain trivial scalars, it certainly makes sense to unlock so other threads can access it (since I guess we have no RW-lock).

(I am still half wondering if we shouldn't just copy the sequences, but not worth trying here.)

numpy/_core/src/multiarray/ctors.c
ngoldbaum: When handed a list, PySequence_Fast returns a proxy object with its own per-object lock. If we passed `obj` in, then we'd lock the PySequence_Fast proxy object and not the original list.

We could check for this in the CRITICAL_SECTION_SEQUENCE_FAST macros but then we'd be doing a check every time we enter the critical section. This is an optimization that I copied from CPython where these macros originally came from.

> Also, can't we assume that a fast-sequence will not change its size if we hold a critical section on it? (But then, I guess critical sections are not fool-proof, so maybe that is a silly thought.)

It can't change while we're executing code in the body of the critical section. It also can't change at all if we never call back into the CPython C API.

Here we are calling back into the CPython C API, but we're using the low-level accessor macros which do no error checking and do not cause the critical section to be released. In principle a future version of Python could decide to change that but I doubt that would happen because the property that critical sections aren't released by low-level accessor macros is being used in lots of places.

The rule of thumb I have in my head is that a critical section can be released in exactly the same places where GIL switches are possible (e.g. if the interpreter enters a possibly blocking call). As long as we ensure there are no possibly blocking calls while we hold the critical section, then it behaves like a "real" lock and does not get released by code called inside the critical section body.

numpy/_core/src/multiarray/ctors.c
ngoldbaum: Yes, it's possible that the size gets changed by another thread while the critical section is released. It could happen any time outside of the critical section if the list is shared with other threads.

numpy/_core/src/multiarray/ctors.c
seberg: > When handed a list, PySequence_Fast returns a proxy object with its own per-object lock

Aha, I really thought that it just returns the original list in that case, not a proxy that can be mutated by mutating the original list...
Well, I'll let you decide whether you want to explain that.

numpy/_core/src/multiarray/ctors.c
seberg: Sorry, my question was that I am surprised that contention of the list objects with multiple threads is actually a very important use-case (i.e. we can't just lock the whole iteration).

ngoldbaum: That's the way it works right now, but in a very silly way where we unconditionally acquire a critical section for lists and tuples. That's the critical section I'm deleting in `PyArray_FromAny_int` down on line 1571 below here.

I observed contention in the benchmark from #30494 - you can see it in the screenshot from https://github.com/numpy/numpy/issues/30494#issuecomment-3679394646.

This PR fixes that contention because the critical sections are held for a much shorter period of time in this PR.

It also gives people an escape hatch if it does ever come up again: pass a tuple instead of a list.

seberg: OK, sorry for the noise, it wasn't (isn't) quite obvious why that script needs this use-case, but fine with it. The question would really be the opposite I guess: Would this make:
```
l = [0] * 100_000
%timeit np.array(l)
```
much slower in the uncontested case.  And if it doesn't all of this is moot anyway. If it is much slower, I suppose this may still be the more interesting thing (i.e. ensure we scale well on free-threading, at least for now that is hopefully how things are used)

ngoldbaum:
> I am slightly curious how this affects the uncontested case 

On current `main`, with 3.14.2t:

```
In [1]: l = [0] * 100_000

In [2]: %timeit np.array(l)
2.31 ms ± 5.96 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

On this PR with the same interpreter:

```
In [1]: l = [0] * 100_000

In [2]: %timeit np.array(l)
2.7 ms ± 3.99 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

On the GIL-enabled build, with current `main`:

```
In [1]: l = [0] * 100_000

In [2]: %timeit np.array(l)
2.18 ms ± 6.44 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

And this PR:

```
In [2]: %timeit np.array(l)
2.24 ms ± 9.87 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

So not nothing, but the impact is disproportionately worse on the free-threaded build.

I think I can improve things here - let me try to experiment a little.

seberg:
Well, to be fair, while very noticeable also not a super bad slowdown if the alternative is terrible congestion on some other cases. (The GIL build doesn't matter here of course)

ngoldbaum:
I posted an incorrect comment and then deleted it - sorry if you all got notifications.

ngoldbaum:
I played around with refactoring to amortize costs and couldn't find a solution that runs faster than this.

One thing we can do is document that using a tuple avoids the critical sections:

```
Preimported NumPy 2.5.0.dev0+git20260105.204d9a5 as np

In [1]: l = [0] * 100_000

In [2]: t = tuple(l)

In [3]: %timeit np.array(l)
2.4 ms ± 7.21 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [4]: %timeit np.array(t)
2.03 ms ± 3.43 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

ngoldbaum:
I opened https://github.com/numpy/numpy/issues/30597 as a longer-term followup task.

seberg:
So one more data point is this:
```
In [4]: l = [1] * 100_1000

In [5]: %timeit np.asarray(tuple(l))
30.6 ms ± 514 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [6]: %timeit np.asarray(l)
29.3 ms ± 171 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
on another computer (this was M1), I see more slowdown, but also <10%.  I think it really begs the question whether the optimization of not always converting to a tuple is even worthwhile (at least on free-threading).

Admittedly, this probably disregards what happens when we have lists that are so large that they do not fit into the cache, but I am not sure we need to optimize for those.

EDIT: OTOH, I guess the picture may be different for very small lists (although, it seems likely that array creation itself to be the lion share then).

ngoldbaum:
Good point. It'd be nice if there were a zero-copy way for a tuple to consume a list's entries.

ngoldbaum:
@colesbury and I chatted and arrived at moving the critical section outside the for loop. We're still only locking the innermost list in a recursive list of lists because only the innermost critical section in the call stack is locked. The difference is that now we keep the lock held while we're processing the whole list. That avoids all the critical section churn and converting a list to an ndarray has very little single-threaded overhead.

I think this is OK to backport now. @seberg are you ok with that?

numpy/_core/src/multiarray/ctors.c
```diff
@@ -532,34 +549,45 @@ PyArray_AssignFromCache_Recursive(
...
                 Py_DECREF(view);
             }
         }
+      finish_critical_section:;
+
+        NPY_END_CRITICAL_SECTION_SEQUENCE_FAST();
+        if (err) {
+            goto finish;
+        }
+
```
seberg, line 579:
```suggestion
```

numpy/_core/src/multiarray/ctors.c
```diff
@@ -532,34 +549,45 @@ PyArray_AssignFromCache_Recursive(
...
                 view = (PyArrayObject *)array_item_asarray(self, i);
                 if (view == NULL) {
-                    goto fail;
+                    err = 1;
+                    goto finish_critical_section;
                 }
                 if (PyArray_AssignFromCache_Recursive(view, ndim, cache) < 0) {
                     Py_DECREF(view);
-                    goto fail;
+                    err = 1;
```
seberg:
```suggestion
```

numpy/_core/src/multiarray/ctors.c
```diff
@@ -532,34 +549,45 @@ PyArray_AssignFromCache_Recursive(
...
+                if (*cache != NULL && (*cache)->converted_obj == item_pyvalue) {
                     *cache = npy_unlink_coercion_cache(*cache);
                 }
             }
             else {
                 PyArrayObject *view;
                 view = (PyArrayObject *)array_item_asarray(self, i);
                 if (view == NULL) {
-                    goto fail;
+                    err = 1;
```
seberg:
```suggestion
```

numpy/_core/src/multiarray/ctors.c
```diff
@@ -493,33 +493,50 @@ NPY_NO_EXPORT int
...
-            PyErr_SetString(PyExc_RuntimeError,
-                    "Inconsistent object during array creation? "
-                    "Content of sequences changed (length inconsistent).");
-            goto fail;
-        }
-
-        for (npy_intp i = 0; i < length; i++) {
-            PyObject *value = PySequence_Fast_GET_ITEM(obj, i);
+        npy_intp orig_length = PyArray_DIMS(self)[0];
+        int err = 0;
```
seberg:
```suggestion
        int err = 1;
```

numpy/_core/src/multiarray/ctors.c
```diff
@@ -532,34 +549,45 @@ PyArray_AssignFromCache_Recursive(
...
                 if (PyArray_AssignFromCache_Recursive(view, ndim, cache) < 0) {
                     Py_DECREF(view);
-                    goto fail;
+                    err = 1;
+                    goto finish_critical_section;
                 }
                 Py_DECREF(view);
             }
         }
+      finish_critical_section:;
```
seberg, line 573:
```suggestion
    err = 0;
      finish_critical_section:;
```

numpy/_core/src/multiarray/ctors.c
```diff
@@ -493,33 +493,50 @@ NPY_NO_EXPORT int
...
+            // this macro takes *the argument* of PySequence_Fast, which is orig_seq;
+            // not the object returned by PySequence_Fast, which is a proxy object
+            // with its own per-object PyMutex lock.
+            // We want to lock the list object exposed to users, not the proxy.
+            npy_intp length = PySequence_Fast_GET_SIZE(obj);
+            if (length != orig_length) {
+                PyErr_SetString(PyExc_RuntimeError,
+                                "Inconsistent object during array creation? "
+                                "Content of sequences changed (length inconsistent).");
+                err = 1;
```
seberg:
```suggestion
```

seberg (approved):
This seems the likely better trade-off to me, thanks. (Not that I care too much which way we go, happy to trust your guts.)

Silly nit/preference that I'll just apply .

And yes, this seems all simple enough to backport to me.

numpy/_core/src/multiarray/ctors.c
```diff
@@ -532,34 +549,45 @@ PyArray_AssignFromCache_Recursive(
                  */
                 char *item;
                 item = (PyArray_BYTES(self) + i * PyArray_STRIDES(self)[0]);
-                if (PyArray_Pack(PyArray_DESCR(self), item, value) < 0) {
-                    goto fail;
+                if (PyArray_Pack(PyArray_DESCR(self), item, item_pyvalue) < 0) {
+                    err = 1;
```
seberg:
```suggestion
```

numpy/_core/src/multiarray/ctors.c
seberg:
```suggestion
```

numpy/_core/src/multiarray/ctors.c
seberg:
```suggestion
        int err = 1;
```

seberg: (Btw. sorry about that spam. The reason is that the new github review experience seems very buggy suggestions for me... claiming the change can't be applied or just not showing the ability to add it)

ngoldbaum:
Thanks all for your help with this. @charris please feel free to ping me about the backport if there are issues.

PR #30518, opened 2025-12-26
CI: update debug CI to use more recent Python version

kumaraditya303:
Updates the debug CI to use Python 3.14. This has the added benefit of dumping C backtrace in case of a crash which is a new feature of faulthandler in 3.14.


charris:
Thanks @kumaraditya303 .

PR #30521, opened 2025-12-26
BUG: validate contraction axes in tensordot

vkverma9534:
tensordot performs a tensor contraction where each axis corresponds to a distinct summation index. Duplicate contraction axes (e.g. axes=([1, 1], [0, 0])) are mathematically invalid but currently fail later during an internal transpose, raising ValueError: axes don't match array. This PR adds early validation to reject duplicate axes explicitly, aligning the behavior with the definition of tensor contraction and other NumPy axis-handling APIs.

Current behavior:
Supplying duplicate contraction axes (e.g. axes=([1, 1], [0, 0])) causes tensordot to fail later during an internal transpose, raising ValueError: axes don't match array.

Expected behavior:
Duplicate contraction axes should be rejected explicitly, since tensor contraction requires distinct summation axes, and a clear input-validation error should be raised before any internal reshaping or transposition.

### DIFF
--- a/numpy/_core/numeric.py
+++ b/numpy/_core/numeric.py
@@ -1170,6 +1170,12 @@ def tensordot(a, b, axes=2):
         axes_b = [axes_b]
         nb = 1
 
+    if len(set(axes_a)) != len(axes_a):
+        raise ValueError("duplicate axes are not allowed in tensordot")
+    if len(set(axes_b)) != len(axes_b):
+        raise ValueError("duplicate axes are not allowed in tensordot")
+
+
     a, b = asarray(a), asarray(b)
     as_ = a.shape
     nda = a.ndim
--- a/numpy/_core/numeric.py
+++ b/numpy/_core/numeric.py
@@ -1022,7 +1022,8 @@ def tensordot(a, b, axes=2):
         * (2,) array_like
           Or, a list of axes to be summed over, first sequence applying to `a`,
           second to `b`. Both elements array_like must be of the same length.
-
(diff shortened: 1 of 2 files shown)


vkverma9534:
Exact case I used 
>>> a = np.ones((2, 3, 3))
>>> b = np.ones((3, 3, 4))
>>> np.tensordot(a, b, axes=([1, 1], [0, 0]))
ValueError: axes don't match array

vkverma9534:
Hello, following up with a gentle reminder on this PR. All checks are passing, and I’m open to any discussion or feedback. Thank you for your consideration.

ngoldbaum:
Seems fine to me.

Can you update the docstring to indicate that repeated entries in `axes` aren't allowed?

Also, IMO the explanation about `axes` in the note of the docstring is nigh-imcomprehensible. If you want to take a crack at making that more understandable by adding some examples that would be great.

vkverma9534:
@ngoldbaum  Thanks for the review!

Yes, agreed ,I’ll update the docstring to explicitly state that repeated entries in axes are not allowed.

I’ll also try to clarify the explanatory note by adding a couple of small concrete examples to make the contraction semantics easier to follow. I’ll push an update shortly.

vkverma9534:
@ngoldbaum 
I’ve added the docstring clarification and examples we talked about earlier and pushed the update. Please let me know if anything else would be helpful.

ngoldbaum (approved):
Thanks!

PR #30522, opened 2025-12-26
BUG: Ensure summed weights returned by np.average always are correct class

mhvk:
Over at astropy, it was noticed that if one passes in a `Quantity` weight to `np.average`, and asks for the summed weight to be returned, then the unit can be lost if the shapes of the weights and input array do not match (see https://github.com/astropy/astropy/issues/19054). This is because of a missing `subok=True` in a final broadcast -- which this PR fixes (how I wish I had insisted more that that would be default for `subok` when it was introduced...).

### DIFF
--- a/numpy/lib/tests/test_function_base.py
+++ b/numpy/lib/tests/test_function_base.py
@@ -497,6 +497,14 @@ class TestAverage:
 
         assert_equal(type(np.average(a)), subclass)
         assert_equal(type(np.average(a, weights=w)), subclass)
+        # Ensure a possibly returned sum of weights is correct too.
+        ra, rw = np.average(a, weights=w, returned=True)
+        assert_equal(type(ra), subclass)
+        assert_equal(type(rw), subclass)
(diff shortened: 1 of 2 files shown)


numpy/lib/tests/test_function_base.py
```diff
@@ -497,6 +497,14 @@ class subclass(np.ndarray):
 
         assert_equal(type(np.average(a)), subclass)
         assert_equal(type(np.average(a, weights=w)), subclass)
+        # Ensure a possibly returned sum of weights is correct too.
+        ra, rw = np.average(a, weights=w, returned=True)
+        assert_equal(type(ra), subclass)
+        assert_equal(type(rw), subclass)
+        # Even if it needs to be broadcast.
```
mhvk, line 504: This test fails without the `subok=True` argument.

charris (approved):

charris:
Thanks Marten.

mhvk:
That's quick! Thanks, Chuck.

PR #30524, opened 2025-12-26
ENH: Add dtype-based ufunc promotion

MaanasArora:
Adds a special type of promoter to the ufunc machinery that always matches when a specific dtype is found exactly in the operands. This is an approach to addressing quality-of-life issues with the current promotion registration, e.g. the need to register one loop multiple times for operand order, even if all promotion logic is essentially custom. 

Me and @seberg had a discussion about these issues and this seemed to be a good middle ground between a very strict hierarchy of promotion rules (what we have currently) and allowing more 'custom' matching logic. We lose the ability to enforce an order between promoters, so things are perhaps easier to break, though it seems it is not very hard to break promotion anyway :)

Open to discussion about better ways of doing something like this; I just wanted to get something concrete out for feedback! The main thing we might have to think about a bit more is the recursion / the 'priority' logic for this promoter and seeing if there's any edge cases there. So, marking this as draft, but if we go down this route this is basically ready to review. Thanks!


numpy/_core/code_generators/numpy_api.py
```diff
@@ -470,6 +470,8 @@ def get_annotations():
     'PyUFunc_GiveFloatingpointErrors':               (46, MinVersion("2.0")),
     # End 2.0 API
     'PyUFunc_AddLoopsFromSpecs':                     (47, MinVersion("2.4")),
+    # End 2.4 API
+    'PyUFunc_AddDTypeBasedPromoter':                 (48, MinVersion("2.5")),
```
MaanasArora: The name 'dtype-based promoter' sounds quite broad and generic... hopefully there's something better.

numpy/_core/src/umath/_scaled_float_dtype.c
```diff
@@ -1040,27 +1014,19 @@ sfloat_init_ufuncs(void) {
...
-        Py_DECREF(promoter);
-        return -1;
-    }
-    promoter_dtypes[0] = double_DType;
-    promoter_dtypes[1] = &PyArray_SFloatDType;
-    res = sfloat_add_loop("multiply", promoter_dtypes, promoter);
+
+    PyObject *ufunc = sfloat_get_ufunc("multiply");
+    res = PyUFunc_AddDTypeBasedPromoter(
+        (PyUFuncObject *)ufunc, &PyArray_SFloatDType, promoter);
```
MaanasArora: This is probably dangerous (this matches even for non-floats), so this is really just a test, happy to remove.

doc/source/reference/c-api/array.rst
```diff
@@ -1931,6 +1931,19 @@ with the rest of the ArrayMethod API.
    passed the operation and requested DType signatures and can mutate it to
    attempt a new search for a matching loop/promoter.
 
+.. c:function:: int PyUFunc_AddDTypeBasedPromoter( \
+                        PyUFuncObject *ufunc, PyArray_DTypeMeta *dtype, PyObject *promoter)
```
seberg: I might be tempted to shoe-horn it into the current promoter (by just allowing a non tuple).

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -317,11 +339,61 @@ resolve_implementation_info(PyUFuncObject *ufunc,
     Py_DECREF(tmp);
 #endif
 
+    Py_ssize_t size = PySequence_Length(ufunc->_dtype_promoters);
+    for (Py_ssize_t i = 0; i < size; i++) {
```
seberg: Hmmm, if we do this, we actually require a match (i.e. that dtype must match for one of the arguments).
Given that, I think that this type of match can be translated to a `(None, DType, None)` or `(DType, DType, None)` depending on how many arguments match?
In that case, it seems actually plausible to fold it into the normal hierarchy.

Otherwise, I am actually tempted to think that we should use these promoters only if the other promotion strategy fails (as they are more precise).

seberg (commented):
Thanks for looking into this, just a few quick thoughts right now.  A single dtype that is expected is probably really one of the most useful things to simplify things.

(That might actually be nicer than `sfloat` tests, we could use this for one of the string ufunc promoters.)

numpy/_core/src/umath/dispatching.cpp
MaanasArora: I think you're right. The only issue I can think of with folding it in is that there could be a lot of registrations for a single dtype, internally, and I wonder if we want that when, for example, we expose API to fetch the loops. E.g. for one dtype and nin=2, we have `(DType, None, None)`, `(None, DType, None)`, `(DType, DType, None)` - so three registrations. It might not be too bad if we explicitly describe this as a convenience though.

> Otherwise, I am actually tempted to think that we should use these promoters only if the other promotion strategy fails (as they are more precise).

Hmm, the difficult case might be something like `(None, None, None)` promoters, hopefully no one thinks of using those, but if they do, that would perhaps supersede any of these? That was my rationale, but I agree making these have precdence seems a bit aggressive.

One option could be to register in the normal hierarchy but store the promoters in different lists, i.e. when resolving, one reads the `dtype_promoters` and essentially replaces them with folded objects in the hierarchy (i.e, 1 dtype promoter is actually read as `2**nargs` promoters with the none replacements). That will fix both problems. It seems much more flexible for change too, though perhaps too complicated...

seberg: I think for `(None, None, None)` if provided, we should actually just call all of them in order as a last resort (lowest priority).
The problem with those is mainly that if there are two and one does nothing, it would be the same as "not matching" and the next one needs to be tried, but that needs some thought how to write well...

That was one of my thoughts when discussing, that just allowing this type of "None" to be a bit more convenient may be just as well and easier to explain.
(I.e. assume priority doesn't matter and users correctly return the input unchanged -- or a new return value -- when they don't match).

MaanasArora: Ah sorry yes, I was misunderstanding, thanks for explaining! So if the `(None) * nargs` is addressed, even in a follow-up, I think your suggestion of using these promoters only if a precise match is not found makes a lot of sense. I've implemented and pushed that now; it seemed more scalable, but still happy to just fold it in if you think that is better!

doc/source/reference/c-api/array.rst
MaanasArora: That sounds nice, thanks! Looking into doing that.

MaanasArora: This is done!

MaanasArora:
I added a test for `StringDType` multiply and shoe-horned the new promoter to `PyUFunc_AddPromoter`. I think this should be ready to review now. Thanks!

PR #30528, opened 2025-12-28
DOC: Add missing import in assert_array_equal docstring example

### LINKED ISSUE #30509: DOC: Missing `import numpy as np` in interactive example of `np.testing.assert_array_equal.html`

The interactive example for [`numpy.testing.assert_array_equal`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_equal.html) doesn't run on initial loading and raises `NameError: name 'np' is not defined`.

Image

Adding the import on the first line of the first cell solves this...

```
import numpy as np
```

I can take this on but checked a few other pages and it seems a common problem so I'm wondering if there is a more generic solution.

If `pyodide` is being used then it would perhaps be possible to [load the package](https://pyodide.org/en/stable/usage/loading-packages.html#loading-packages-with-pyodide-loadpackage) globally earlier in the configuration/setup. I doubt this will work...

```
await pyodide.loadPackage("numpy")
```

...but maybe

```
await pyodide.import("numpy", as="np")
```

but struggling to find the API reference for `pyodide.import` so that may not be possible either and I've not played with it much.

DanielDerefaka:
## Summary
Fixes #30509

The interactive example for `np.testing.assert_array_equal` in the web docs was raising a `NameError: name 'np' is not defined` because the `import numpy as np` statement was missing from the docstring examples.

## Changes
Added `>>> import numpy as np` at the beginning of the Examples section in `numpy/testing/_private/utils.py`.

## Test plan
- [x] Documentation change only - verified the example now runs correctly with the import

### DIFF
--- a/numpy/testing/_private/utils.py
+++ b/numpy/testing/_private/utils.py
@@ -1055,6 +1055,8 @@ def assert_array_equal(actual, desired, err_msg='', verbose=True, *,
 
     Examples
     --------
+    >>> import numpy as np
+
     The first assert does not raise an exception:
 
     >>> np.testing.assert_array_equal([1.0,2.33333,np.nan],


charris:
Documentation only, the minimal tests are sufficient.

DanielDerefaka:
Hi @charris, just checking if there's anything else needed for this to be merged?

jorenham:
All good; thanks @DanielDerefaka

PR #30530, opened 2025-12-28
TYP/STY: ``_core.defchararray`` formatting

jorenham:
I was (trying to) compare this with the numtype stubs to see if it needed updating, but the formatting made that pretty difficult, so here I am.


charris:
I think we want to get rid of character arrays in the long term.

charris:
Thanks Joren.

jorenham:
> I think we want to get rid of character arrays in the long term.

It's not deprecated yet, is it? I could smack a `@deprecated` with a `PendingDeprecationWarning` on it if it helps.

PR #30532, opened 2025-12-28
TYP: ``{lin,log,geom}space`` shape-typing

jorenham:
... for the (most commonly used) 1-d cases, that is. 

This also makes the return dtypes for the non-`float64` cases a bit more user-friendly (`floating` and `complexfloating` can be annoying to work with).

### DIFF
--- a/numpy/_core/function_base.pyi
+++ b/numpy/_core/function_base.pyi
@@ -1,5 +1,5 @@
 from _typeshed import Incomplete
-from typing import Literal as L, SupportsIndex, overload
+from typing import Any, Literal as L, SupportsIndex, overload
 
 import numpy as np
 from numpy._typing import (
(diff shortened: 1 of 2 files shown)


jorenham:
The `xarray` primer error is the reassignment of 2d array to 1d array, which mypy doesn't support (unless you run mypy with `--local-partial-types --allow-redefinition-new`): https://github.com/pydata/xarray/blob/26ccc7f8f014f29c551fd566a04d6e9f878c0b0b/xarray/tests/test_plot.py#L562-L564

And for `colour` the story is the same: https://github.com/colour-science/colour/blob/6746a9c1dc9b9018e3a10ccaf3e63305342ee49d/colour/temperature/tests/test_ohno2013.py#L95-L98

charris:
I take it you see this as a mypy problem?

jorenham:
> I take it you see this as a mypy problem?

Yup, and a solved one at that.

charris:
Thanks Joren.

PR #30534, opened 2025-12-28
TYP: move ``memmap`` stubs to ``_core.memmap``

jorenham:
This also narrows the generic shape type-parameter default to the gradual "any shape" type, and makes the constructor transparent w.r.t. passed `dtype` instances. The attributes have also been made final and slightly narrower in case of `mode`.


charris:
Thanks Joren.

PR #30535, opened 2025-12-28
TYP: inline declarations for ``empty``, ``zeros``, and ``ones``

jorenham:
It's less DRY, but the difference between callable protocols and "true" function types is non-negligible, and can lead to unexpected type-checker behavior. For IDE features (LSP stuff) this can also matter.


charris:
Thanks Joren.

PR #30536, opened 2025-12-28
TYP: ``bincount`` shape-typing

jorenham:
It always returns a 1d array, and with this the type-checkers also know.


charris:
Thanks Joren.

PR #30537, opened 2025-12-28
TYP: ``from{buffer,string,file}`` shape-typing

jorenham:
These three functions exclusively return 1d arrays, so let's put that in the stubs.


jorenham:
The new mypy primer error comes from https://github.com/pandas-dev/pandas/blob/main/pandas/io/stata.py#L1374-L1381.
It seems to be caused by another awkward inference issue, that could easily be resolved by explicitly annotating the relevant `dict`. Nothing we can do anything about from our side of things at least.

charris:
Thanks Joren.

PR #30538, opened 2025-12-29
BUG: Ensure stable data address in scalar __array_interface__

riku-sakamoto:
### Changes

* Implement `__array_interface__` for NumPy scalar types to avoid creating a temporary array.
* Set the read-only flag in the `__array_interface__` dict for NumPy scalars.

As a consequence of this change, ``numpy.ctypeslib.as_ctypes`` now raises ``TypeError`` when NumPy scalar types are passed directly.

Closes #30354

### DIFF
--- a/doc/release/upcoming_changes/30538.change.rst
+++ b/doc/release/upcoming_changes/30538.change.rst
@@ -9,11 +9,10 @@ Users who need to convert scalar types to ctypes should first convert them to an
 (e.g., ``numpy.asarray``) before passing them to ``numpy.ctypeslib.as_ctypes``.
 
 
-``__ref`` value in ``__array_interface__`` for scalar types is removed
-------------------------------------------------------------------------
-Previously, scalar types in NumPy had a ``__ref`` entry in their ``__array_interface__``
-dictionary, which pointed to the temporary array created to hold the scalar value.
-This behavior has been removed to avoid the issue `gh-30354 <https://github.com/numpy/numpy/issues/30354>`__.
-Now, scalar types will not have a ``__ref`` entry in their ``__array_interface__``.
-Instead, ``data`` entry in ``__array_interface__`` will directly point to the memory location of the scalar value and
-will be marked as readonly.
+``__array_interface__`` changes on scalars
+------------------------------------------
+Scalars now export the ``__array_interface__`` directly rather than including
+an array copy as a ``__ref`` entry. This means that scalars are now exported
+as read-only while they previously exported as writeable.
+The path via ``__ref__`` was undocumented and not consistently used even
+within NumPy itself.
--- a/doc/release/upcoming_changes/30538.change.rst
+++ b/doc/release/upcoming_changes/30538.change.rst
@@ -1,0 +1,8 @@
+``numpy.ctypeslib.as_ctypes`` now does not support scalar types
+----------------------------------------------------------------
+The function ``numpy.ctypeslib.as_ctypes`` has been updated to only accept ``numpy.ndarray``.
+Passing a scalar type (e.g., ``numpy.int32(5)``) will now raise a ``TypeError``.
+This change was made to avoid the issue `gh-30354 <https://github.com/numpy/numpy/issues/30354>`__
+and to enforce the readonly nature of scalar types in NumPy.
+Users who need to convert scalar types to ctypes should first convert them to an array
+(e.g., ``numpy.asarray``) before passing them to ``numpy.ctypeslib.as_ctypes``.
--- a/doc/release/upcoming_changes/30538.change.rst
+++ b/doc/release/upcoming_changes/30538.change.rst
@@ -14,5 +14,5 @@ Users who need to convert scalar types to ctypes should first convert them to an
 Scalars now export the ``__array_interface__`` directly rather than including
 an array copy as a ``__ref`` entry. This means that scalars are now exported
 as read-only while they previously exported as writeable.
-The path via ``__ref__`` was undocumented and not consistently used even
+The path via ``__ref`` was undocumented and not consistently used even
 within NumPy itself.
--- a/doc/release/upcoming_changes/30538.change.rst
+++ b/doc/release/upcoming_changes/30538.change.rst
@@ -4,5 +4,16 @@ The function ``numpy.ctypeslib.as_ctypes`` has been updated to only accept ``num
 Passing a scalar type (e.g., ``numpy.int32(5)``) will now raise a ``TypeError``.
 This change was made to avoid the issue `gh-30354 <https://github.com/numpy/numpy/issues/30354>`__
 and to enforce the readonly nature of scalar types in NumPy.
+The previous behavior relied on undocumented implicit temporary arrays and was not well-defined.
 Users who need to convert scalar types to ctypes should first convert them to an array
 (e.g., ``numpy.asarray``) before passing them to ``numpy.ctypeslib.as_ctypes``.
+
+
+``__ref`` value in ``__array_interface__`` for scalar types is removed
+------------------------------------------------------------------------
+Previously, scalar types in NumPy had a ``__ref`` entry in their ``__array_interface__``
+dictionary, which pointed to the temporary array created to hold the scalar value.
+This behavior has been removed to avoid the issue `gh-30354 <https://github.com/numpy/numpy/issues/30354>`__.
+Now, scalar types will not have a ``__ref`` entry in their ``__array_interface__``.
+Instead, ``data`` entry in ``__array_interface__`` will directly point to the memory location of the scalar value and
+will be marked as readonly.
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -3804,7 +3804,7 @@ array_typestr_get(PyArray_Descr *self)
 }
 
 
-PyObject *
+NPY_NO_EXPORT PyObject *
 array_protocol_descr_get(PyArray_Descr *self)
 {
     PyObject *res;
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -3797,6 +3797,42 @@ descr_subscript(PyArray_Descr *self, PyObject *op)
     }
 }
 
+static PyObject *
+array_typestr_get(PyArray_Descr *self)
+{
+    return arraydescr_protocol_typestr_get(self, NULL);
+}
+
+
+PyObject *
+array_protocol_descr_get(PyArray_Descr *self)
+{
+    PyObject *res;
+    PyObject *dobj;
+
+    res = arraydescr_protocol_descr_get(self, NULL);
+    if (res) {
+        return res;
+    }
+    PyErr_Clear();
+
+    /* get default */
+    dobj = PyTuple_New(2);
+    if (dobj == NULL) {
+        return NULL;
+    }
+    PyTuple_SET_ITEM(dobj, 0, PyUnicode_FromString(""));
+    PyTuple_SET_ITEM(dobj, 1, array_typestr_get(self));
+    res = PyList_New(1);
+    if (res == NULL) {
+        Py_DECREF(dobj);
+        return NULL;
+    }
+    PyList_SET_ITEM(res, 0, dobj);
+    return res;
+}
+
+
 static PySequenceMethods descr_as_sequence = {
     (lenfunc) descr_length,                  /* sq_length */
     (binaryfunc) NULL,                       /* sq_concat */
--- a/numpy/_core/src/multiarray/scalartypes.c.src
+++ b/numpy/_core/src/multiarray/scalartypes.c.src
@@ -1812,7 +1812,6 @@ gentype_interface_get(PyObject *self, void *NPY_UNUSED(ignored))
     inter = PyObject_GetAttrString((PyObject *)arr, "__array_interface__");
     if (inter == NULL) {
         Py_DECREF(arr);
-        Py_DECREF(inter);
         return NULL;        
     }
 
--- a/numpy/_core/src/multiarray/scalartypes.c.src
+++ b/numpy/_core/src/multiarray/scalartypes.c.src
@@ -1788,6 +1788,18 @@ gentype_shape_get(PyObject *NPY_UNUSED(self), void *NPY_UNUSED(ignored))
 
 
 static PyObject *
+gentype_dataptr_get(PyObject *self, int flags, void *NPY_UNUSED(ignored))
+{
+    int _flags = flags;
+    _flags &= ~(NPY_ARRAY_WRITEBACKIFCOPY | NPY_ARRAY_OWNDATA);
+    _flags |= NPY_ARRAY_NOTSWAPPED;
+
+    return Py_BuildValue(
+            "NO", PyLong_FromVoidPtr(scalar_value(self, NULL)),
+            ((flags & PyBUF_WRITEABLE) == PyBUF_WRITEABLE) ? Py_False : Py_True);
+}
+
+static PyObject *
 gentype_interface_get(PyObject *self, void *NPY_UNUSED(ignored))
 {
     PyArrayObject *arr;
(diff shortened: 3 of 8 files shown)


numpy/_core/src/multiarray/scalartypes.c.src
```diff
@@ -1798,15 +1810,31 @@ gentype_interface_get(PyObject *self, void *NPY_UNUSED(ignored))
         return NULL;
     }
     inter = PyObject_GetAttrString((PyObject *)arr, "__array_interface__");
-    if (inter != NULL) {
-        PyDict_SetItemString(inter, "__ref", (PyObject *)arr);
+    if (inter == NULL) {
+        Py_DECREF(arr);
+        Py_DECREF(inter);
```
ngoldbaum: Aren't you decrefing a pointer that has to be NULL here?

ngoldbaum: this also indicates this error path isn't tested - might be worth adding a test!

riku-sakamoto: > Aren't you decrefing a pointer that has to be NULL here?

Thank you for catching this. you're absolutely right.
I fixed the incorrect decref.

> this also indicates this error path isn't tested - might be worth adding a test!

This `if` branch seems to be unreachable because the `array` object always provides `__array_interface__`.
Instead of adding a direct test for this path, I improved the test coverage by:
* increasing the number of scalar types under test
* adding checks for unconversible types

numpy/_core/src/multiarray/scalartypes.c.src
```diff
@@ -1787,6 +1787,18 @@ gentype_shape_get(PyObject *NPY_UNUSED(self), void *NPY_UNUSED(ignored))
 }
 
 
+static PyObject *
+gentype_dataptr_get(PyObject *self, int flags, void *NPY_UNUSED(ignored))
+{
+    int _flags = flags;
+    _flags &= ~(NPY_ARRAY_WRITEBACKIFCOPY | NPY_ARRAY_OWNDATA);
+    _flags |= NPY_ARRAY_NOTSWAPPED;
```
ngoldbaum: This is all a no-op because `_flags` isn't used below. See my other comment below for speculation about why you defined this variable.

numpy/_core/src/multiarray/scalartypes.c.src
```diff
@@ -1787,6 +1787,18 @@ gentype_shape_get(PyObject *NPY_UNUSED(self), void *NPY_UNUSED(ignored))
...
+
+    return Py_BuildValue(
+            "NO", PyLong_FromVoidPtr(scalar_value(self, NULL)),
+            ((flags & PyBUF_WRITEABLE) == PyBUF_WRITEABLE) ? Py_False : Py_True);
```
ngoldbaum: Why are you checking `PyBUF_WRITEABLE` here - that isn't an ndarray flag. `PyBUF_WRITEABLE` is part of the [CPython buffer request API](https://docs.python.org/3/c-api/buffer.html#buffer-request-types) and will only show up in calls directly from CPython for handling the buffer protocol, which has nothing to do with this code as far as I can see. 

Looking at the implementation of `array_dataptr_get` I think you meant to use `_flags` here and not do the `PyBUF_WRITEABLE` check at all.

Also I don't think this check for read-only matters at all for scalars because scalars are always read-only:

```
>>> a = np.int32(5)
>>> a[:] = 7
    a[:] = 7
    ~^^^
TypeError: 'numpy.int32' object does not support item assignment
>>> a.data
<memory at 0x200012a6e50>
>>> a.data[:] = 7
  File "<python-input-5>", line 1, in <module>
    a.data[:] = 7
    ~~~~~~^^^
TypeError: cannot modify read-only memory
```

Maybe you can avoid all this messing with flags and just set statically that this is read-only.

Also why not set the other items in the `__array_interface__` dict like `shape` etc? Seems like there's no reason not to and it should be trivial to do that for scalars. I'd follow the ndarray implementation in `array_interface_get`.

ngoldbaum (dismissed):
There's some code that is doing questionable things and it only happens to work because scalars are read-only. See comments inline.

numpy/_core/src/multiarray/scalartypes.c.src
riku-sakamoto: Thank you for taking time to review my PR.

I’d like to briefly explain my original reasoning, and then align with your suggestions.


* Why I do not set read-only flag for scalar array?

As you point out, scalars are always read-only.
However, according to the code below in `ctypeslib`, if the array is read-only, `as_ctypes` raise an error. So, I decided to follow the way flags are created in `__array_struct__`.

https://github.com/numpy/numpy/blob/3556042ac2576a7bc4ec5a5bba9954fd07b4167e/numpy/ctypeslib/_ctypeslib.py#L594-L595


Maybe, is it better to set flag as writable ? (Or forbid `as_ctypes` for scalars altogether?)


* use of `PyBUF_WRITEABLE` flag

This was a clear mistake on my part. I confused ndarray's flags with buffer protocol's flags. As you pointed out, `PyBUF_WRITEABLE` is not appropriate here, and I will remove this entirely.


* `_flags` variable

You are also right here. I intended to use `_flags` to check writability, but the current code does not do that correctly. 
I am sorry for the confusion.


> why not set the other items in the __array_interface__ dict like shape etc? Seems like there's no reason not to and it should be trivial to do that for scalars. I'd follow the ndarray implementation in array_interface_get.

Sure! I will implement it following to `array_interface_get`.

riku-sakamoto: I've updated the PR accordingly except for the what kind of flags to set for scalar arrays.

numpy/_core/src/multiarray/scalartypes.c.src
```diff
@@ -1788,23 +1789,127 @@ gentype_shape_get(PyObject *NPY_UNUSED(self), void *NPY_UNUSED(ignored))
...
         return NULL;
     }
-    inter = PyObject_GetAttrString((PyObject *)arr, "__array_interface__");
-    if (inter != NULL) {
-        PyDict_SetItemString(inter, "__ref", (PyObject *)arr);
-    }
+
+    /* NOTE: sclar array is always readonly, but we need writable flags
+     * to use `ctypeslib.as_ctyps` via `__array_interface__`.
+     * So, we set the flag borrowing from the temporary array.
```
seberg: This doesn't really work out, I think. The content of the scalar is clearly readonly (except for void ones).
It seems that `__array_interface__` right now doesn't consider it to be readonly though, which is technically true, because it refers to `__ref` which is undocumented.

We have two options here:
1. Document that `__ref` thing
2. Change the flag to return `readonly` here (maybe always, also for void).

I would be tempted to do the second one but that might backfire in practice and we need to ask a few others.

In the second case, `ctypes` needs to work around whatever we do here. That may be just ignoring the readonly part or creating a copy explicitly.

Actually, it may make sense to simply error in `ctypeslib` because `ctypeslib` was clearly always _very_ broken with a reasonably high chance of generating bad results *without* an error.

riku-sakamoto: Thanks for the comment!

I agree with the second option. Under this option, we would not change anything in `ctypeslib.as_ctypes`.
As a result, scalar arrays would no longer be usable with `ctypeslib.as_ctypes`, but this behavior seems more robust.

>  we need to ask a few others.

Is there anyone I should ping about this, or would it be appropriate to bring this topic to the next triage meeting?

ngoldbaum: @mtsokol do you have any context about where your initial report about this came from?

mtsokol: It appeared in our custom compiler:
https://github.com/finch-tensor/finch-tensor-lite/pull/248/files#diff-70a950803da49a25141ebb1d3bd6a6e33d417e2c3e0c42b0a34ed662bd690b2cL211 - within code meant for serializing NumPy scalars to C. We ended up using 0-d arrays here so we won't be impacted if `as_ctypes` stops supporting scalars.

seberg: Yeah, just to be clear "asking around", I was just thinking of brining it up at the triage meeting and if there is a general consensus to try, try it (but not backport!).

I doubt we'll easily find someone who might actually be affected.  The problem is that the main use-case may be something like a Cython function that uses a typed memory view (like `double data[::1]`) that _could_ use a `const` (signalling read-only is OK), but doesn't...

And we should maybe think whether such Cython use-cases would hit this (I am not sure), because if they do, than that might be an unfortunately large change.

seberg: :facepalm: this is `__array_interface__` so doesn't matter for Cython of course, i.e. the chance of someone actually caring/noticing is a _lot_ smaller.

riku-sakamoto: Thanks everyone for the discussion.
It seems reasonable to try this out, but just in case, I’ve added it as a topic for the next triage meeting.

riku-sakamoto: I've updated the flags in `__array_interface__` for scalars, following the discussion in the last triage meeting.
I've also updated the tests accordingly.

I checked repositories under the numba organization to see whether `ctypeslib.as_ctypes` is used and did not find any usage.
For reference:
https://github.com/search?q=org%3Anumba+%22as_ctypes%28%22&type=code

riku-sakamoto:
@ngoldbaum @seberg 

Just a gentle ping.
I’ve updated the code based on previous triage mtg.
Please let me know if you have any remaining concerns!
I’m happy to address them.

ngoldbaum:
I think this needs a release note.

Can you also update the PR description to match what it actually does? The PR description will show up in the final merge commit.

ngoldbaum (commented):
The C API uses and reference counting all look correct to me.

riku-sakamoto:
@ngoldbaum 
Thank you for taking another look.
I've updated the PR description to clarify the behavior and added a release note.

doc/release/upcoming_changes/30538.change.rst
```diff
@@ -0,0 +1,8 @@
+``numpy.ctypeslib.as_ctypes`` now does not support scalar types
+----------------------------------------------------------------
+The function ``numpy.ctypeslib.as_ctypes`` has been updated to only accept ``numpy.ndarray``.
+Passing a scalar type (e.g., ``numpy.int32(5)``) will now raise a ``TypeError``.
+This change was made to avoid the issue `gh-30354 <https://github.com/numpy/numpy/issues/30354>`__
+and to enforce the readonly nature of scalar types in NumPy.
+Users who need to convert scalar types to ctypes should first convert them to an array
+(e.g., ``numpy.asarray``) before passing them to ``numpy.ctypeslib.as_ctypes``.
```
seberg, line 9: Might be worth pointing out that this was always shady? :).

Can you add a second note about the change in `__array_interface__`?  (or add both in one.)

numpy/_core/src/multiarray/scalartypes.c.src
```diff
@@ -1787,26 +1788,117 @@ gentype_shape_get(PyObject *NPY_UNUSED(self), void *NPY_UNUSED(ignored))
...
+    PyTuple_SET_ITEM(dobj, 0, PyUnicode_FromString(""));
+    PyTuple_SET_ITEM(dobj, 1, gentype_typestr_get(self));
+    res = PyList_New(1);
+    if (res == NULL) {
+        Py_DECREF(dobj);
+        return NULL;
+    }
+    PyList_SET_ITEM(res, 0, dobj);
+    return res;
+}
```
seberg: It would be cool to refactor these so that we don't repeat them in both files (i.e. by making them take the `PyArray_Descr *`).

But I can live without doing that at this point :).

seberg (approved):
LGTM, one nit is that we should also note the change in `__array_interface__` maybe?

I would prefer if we could refactor this so that we have a single function that build the array interface but gets `strides`, `shape`, `ptr` and `descr` passed in.
(Right now we have twice practically the identical code.)

But considering how long this has been sitting, I am also happy if we don't do that.

doc/release/upcoming_changes/30538.change.rst
```diff
@@ -0,0 +1,19 @@
...
+
+
+``__ref`` value in ``__array_interface__`` for scalar types is removed
+------------------------------------------------------------------------
+Previously, scalar types in NumPy had a ``__ref`` entry in their ``__array_interface__``
+dictionary, which pointed to the temporary array created to hold the scalar value.
+This behavior has been removed to avoid the issue `gh-30354 <https://github.com/numpy/numpy/issues/30354>`__.
+Now, scalar types will not have a ``__ref`` entry in their ``__array_interface__``.
+Instead, ``data`` entry in ``__array_interface__`` will directly point to the memory location of the scalar value and
+will be marked as readonly.
```
seberg: Thanks for the quick update! I was actually thinking the main thing is that we export it as readonly now.

(But maybe I'll just think about it a bit myself and push an update later)

numpy/_core/src/multiarray/scalartypes.c.src
riku-sakamoto: Thanks for pointing this out.

I've refactored the duplicated logic into a common helper function in `descriptor.c` and updated the relevant call to use it.

I left some functions, such as `gentype_shape_get`, unchanged since they are also used in other parts of `scalartypes.c.src`.
If you'd prefer to refactor those as well, I'm happy to follow up.

doc/release/upcoming_changes/30538.change.rst
riku-sakamoto: Thanks! That makes sense. Let me know if you'd like me to make any adjustments on my side.

seberg:
```suggestion
``__array_interface__`` changes on scalars
------------------------------------------
Scalars now export the ``__array_interface__`` directly rather than including
an array copy as a ``__ref`` entry. This means that scalars are now exported
as read-only while they previously exported as writeable.
The path via ``__ref__`` was undocumented and not consistently used even
within NumPy itself.
```

numpy/_core/src/multiarray/descriptor.c
```diff
@@ -3779,6 +3797,42 @@ descr_subscript(PyArray_Descr *self, PyObject *op)
...
 }
 
+static PyObject *
+array_typestr_get(PyArray_Descr *self)
+{
+    return arraydescr_protocol_typestr_get(self, NULL);
+}
+
+
+PyObject *
```
seberg:
```suggestion
NPY_NO_EXPORT PyObject *
```
(probably doesn't matter since it's in the header, but...

numpy/_core/src/multiarray/scalartypes.c.src
```diff
@@ -1787,26 +1788,82 @@ gentype_shape_get(PyObject *NPY_UNUSED(self), void *NPY_UNUSED(ignored))
...
+
+    obj = gentype_shape_get(self, NULL);
+    ret = PyDict_SetItemString(inter, "strides", obj);
+    Py_DECREF(obj);
+    if (ret < 0) {
+        Py_DECREF(inter);
+        return NULL;
+    }
+
+    obj = array_protocol_descr_get(PyArray_DescrFromScalar(self));
```
seberg: Hmm, thanks for the update, unfortunately `PyArray_DescrFromScalar` returns a new reference (and actually can in theory error). So we need to do refcounting here unfortunately.
(I was going to just "suggest" it, but it spans the next block also, a `goto finish:` that also cleans up `inter` may be nicer.

riku-sakamoto: Thanks for pointing it out.

I've updated this commit to handle the refcounting for `PyArray_DescrFromScalar`
in `scalartypes.c.src`:
https://github.com/numpy/numpy/pull/30538/commits/5632bf02e1a95ddb32d7328a34f57919d84c72ce

In addition, I also refactored the code to use a common helper when creating
`__array_interface__` for both arrays and scalars.

seberg (approved):
OK, thanks @riku-sakamoto for all the follow-ups and checking what it might break. Let's get this in.  (Small nit/comment: I think the building function would be slightly nicer to take the dtype/descr object to consolidate more of the logic there, but doens't matter if you feel about it, you can follow up).

If anyone finds this PR because the readonly or ctypes changes affect you, please let us know.  I can see this affecting downstream, and if it does we have to check what to do about `ctypes`.

(I know that using ctypes *is* quite typical, my assumption is that it is not all that typical to use it on scalars and since it was broken memory ownership wise, I am hoping it isn't used much -- but I realize that the e.g. the fact that we cache this type of small temporary arrays may have made this work in practice!)

PR #30541, opened 2025-12-29
TST: do not share arrays across threads in `test_ufunc_at_scalar_value_fastpath`

kumaraditya303:


ngoldbaum (approved):
This is in-line with other changes we did earlier this year in service of making the test suite runnable under pytest-run-parallel.

PR #30542, opened 2025-12-29
TYP: set ops shape-typing

jorenham:
This affects the return types of the following ``lib._arraysetops_impl`` functions:

- `unique_all`
- `unique_counts`
- `unique_inverse`
- `unique_values`
- `intersect1d`
- `setxor1d`
- `union1d`
- `setdiff1d`
- `ediff1d`  (not a set op, but oh well...)

This matches numtype


charris:
Thanks Joren.

PR #30545, opened 2025-12-29
TYP: ``histogram*``: shape-typing and dtype specialization

jorenham:
The histogram function stubs were mostly incomplete, giving little type-safety. This adds (many) overloads for `histogram_bin_edges`, `histogram`, and `histogramdd` for specific input combinations to return specialized dtypes. This additionally changes the returned array types to include their (often 1d) shape-types.

Don't let the many overloads scare you; because just like with death metal, once you can get past the machine-gun blastbeats and gutteral pig-squealing, then you'll find that there are actually a couple of notes hidden in there. So static typing is basically the death metal of Python (in a good way). I guess that must be why I like it then 🤔.

### DIFF
--- a/numpy/lib/_histograms_impl.pyi
+++ b/numpy/lib/_histograms_impl.pyi
@@ -8,6 +8,7 @@ from numpy._typing import (
     NDArray,
     _ArrayLike,
     _ArrayLikeComplex_co,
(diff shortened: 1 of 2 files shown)


charris:
Thanks Joren.

PR #30546, opened 2025-12-29
TYP/STY: ``lib._shape_base_impl`` formatting

jorenham:
For compatibility with ruff and the typing spec stub style guide (i.e. a 130 char limit)


charris:
Thanks Joren. I confess to finding long lines less readable, but to each their own.

Issue #28694, opened 2025-04-12
DOC: docs build warnings


lucascolley:
- [x] unterminated string literal

```console
WARNING: error while formatting arguments for numpy.__array_namespace_info__: Handler <function update_annotations_using_type_comments at 0x110762b60> for event 'autodoc-before-process-signature' threw an exception (exception: ('unterminated string literal (detected at line 2)', (2, 54)))
WARNING: error while formatting arguments for numpy.__array_namespace_info__: Handler <function update_annotations_using_type_comments at 0x110762b60> for event 'autodoc-before-process-signature' threw an exception (exception: ('unterminated string literal (detected at line 2)', (2, 54)))
```

- [ ] doxygen obsolete tags:

```console
warning: Tag 'OUTPUT_TEXT_DIRECTION' at line 15 of file 'build/doxygen/Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'CLANG_ASSISTED_PARSING' at line 149 of file 'build/doxygen/Doxyfile' belongs to an option that was not enabled at compile time.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u", or recompile doxygen with this feature enabled.
warning: Tag 'CLANG_OPTIONS' at line 150 of file 'build/doxygen/Doxyfile' belongs to an option that was not enabled at compile time.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u", or recompile doxygen with this feature enabled.
warning: Tag 'CLANG_DATABASE_PATH' at line 151 of file 'build/doxygen/Doxyfile' belongs to an option that was not enabled at compile time.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u", or recompile doxygen with this feature enabled.
warning: Tag 'COLS_IN_ALPHA_INDEX' at line 156 of file 'build/doxygen/Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'HTML_TIMESTAMP' at line 172 of file 'build/doxygen/Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'FORMULA_TRANSPARENT' at line 205 of file 'build/doxygen/Doxyfile' has become obsolete.
[...]

PR #30550, opened 2025-12-30
MAINT: Remove obsolete Doxygen tags from config

Harish3005-cloud:
Removed obsolete configuration tags from doc/source/doxyfile to fix build warnings. These tags are no longer supported in the version of Doxygen currently used by the build system.

Fixes build warnings reported in #28694.


charris:
What is the version currently used? I note that the changes here are not the same as what I see with Doxygen 1.14.0. Might be worth noting the version in a comment.

Harish3005-cloud:
The logs in the linked issue (#28694) indicate Doxygen version used: 1.13.2. I removed the tags that were explicitly flagged as obsolete in those specific build logs.

The file header currently lists 1.8.18. Would you like me to update that header comment to 1.13.2 to match the current build environment?

charris:
> Would you like me to update that header comment 

Yes.

charris:
Thanks @Harish3005-cloud .

PR #30556, opened 2026-01-01
MAINT: consistent ``__new__`` parameter naming

jorenham:
Pyright, for example, complains about ``__new__`` methods whose first parameter isn't named ``cls``.


charris:
Thanks Joren.

PR #30557, opened 2026-01-01
TYP: ``ma.core`` annotate the remaining unannotated functions

jorenham:
This adds annotations to the following functions:

- `ma.power`
- `ma.argsort`
- `ma.concatenate`
- `ma.diag` 
- `ma.left_shift`
- `ma.right_shift`
- `ma.transpose`
- `ma.reshape`
- `ma.resize`
- `ma.shape`
- `ma.diff`
- `ma.where`
- `ma.choose`
- `ma.round`
- `ma.inner`
- `ma.outer`
- `ma.correlate`
- `ma.convolve`
- `ma.fromflex`
- `ma.append`
- `ma.dot`

This also fixes `ma.sort` so that it'll will no longer falsely reject `axis=None`.


charris:
Thanks Joren.

PR #30558, opened 2026-01-01
TYP: ``ma.core.mvoid`` annotations

jorenham:
With this and #30557 we'll have 100% type coverage of the `numpy.ma.core` module.  
Note that there are still a bunch of unannotated functions in `ma.extras` and `ma.mrecords`, so  `numpy.ma` itself isn't yet at 100% yet.


charris:
Needs rebase.

charris:
Thanks Joren.

PR #30562, opened 2026-01-02
DOC: update security policy to be more explicit about bad assumptions

rgommers:
Also update the link in the README - we want to continue mentioning Tidelift, but point to https://github.com/numpy/.github/blob/main/SECURITY.md first so we can link the security page in our docs there.

Note that the removed https://www.python.org/dev/security link no longer contains anything useful; it used to contain concrete guidance similar to what this PR adds (more briefly) under an Important heading, see https://web.archive.org/web/20210108071933/https://www.python.org/dev/security/.

Hopefully this will reduce the number of invalid reports. Or at least it will make replying to them easier.


seberg (approved):
Thanks Ralf, looks good to me. Too bad, Python removed part of its guidance (but I am sure they had good reasons).

PR #30568, opened 2026-01-04
BUG: np.einsum('i...->i', optimize=False) errored out when optimize=True didn't

TomPOGET:
Fixes #30349

`np.einsum('i...->i', x)` works as expected when optimize=True but throws a ValueError when optimize=False,
with `x = np.array([[[1,2,3], [1,2,3], [1,2,3]], [[1,2,3], [1,2,3], [1,2,3]], [[1,2,3], [1,2,3], [1,2,3]]])`.

## NB

np.einsum('ijk->i', x) worked as expected with optimize=False, indicating an error in broadcasting with ellipses.

## Error message

  File "<python-input-0>", line 5, in <module>
    np.einsum('i...->i', x)
    ~~~~~~~~~^^^^^^^^^^^^^^```
File "[...]/venv/lib/python3.13/site-packages/numpy/_core/einsumfunc.py", line 1423, in einsum
    return c_einsum(*operands, **kwargs)
ValueError: output has more dimensions than subscripts given in einstein sum, but no '...' ellipsis provided to broadcast the extra dimensions.

## Fix

Deleted errors thrown for valid syntax; added the infered broadcasted dimensions to the dimensions of the operands, with label '0', so that 'i...->' is effectively processed the same way as 'ijk->' and so that it can use the same accelerations.

### DIFF
--- a/numpy/_core/src/multiarray/einsum.cpp
+++ b/numpy/_core/src/multiarray/einsum.cpp
@@ -228,15 +228,6 @@ parse_output_subscripts(char *subscripts, int length,
         }
     }
 
-    /* If no ellipsis was found there should be no broadcast dimensions. */
-    if (!ellipsis && ndim_broadcast > 0) {
-        PyErr_SetString(PyExc_ValueError,
-                        "output has more dimensions than subscripts "
-                        "given in einstein sum, but no '...' ellipsis "
-                        "provided to broadcast the extra dimensions.");
-        return -1;
-    }
-
     return ndim;
 }
 
@@ -281,13 +272,12 @@ get_single_op_view(PyArrayObject *op, char *labels,
             /* The next output label that's a broadcast dimension */
             for (; ibroadcast < ndim_output; ++ibroadcast) {
                 if (output_labels[ibroadcast] == 0) {
+                    /* Reduction */
                     break;
                 }
             }
             if (ibroadcast == ndim_output) {
-                PyErr_SetString(PyExc_ValueError,
-                        "output had too few broadcast dimensions");
-                return -1;
+                break;
             }
             new_dims[ibroadcast] = PyArray_DIM(op, idim);
             new_strides[ibroadcast] = PyArray_STRIDE(op, idim);
@@ -987,6 +977,18 @@ PyArray_EinsteinSum(char *subscripts, npy_intp nop,
                 goto fail;
             }
             iter_labels[ndim_iter++] = label;
+        }
+    }
+
+    /* Add broadcast dimensions (represented as zeros) that aren't in output */
+    if (memchr(output_labels, 0, ndim_output) == NULL && ndim_broadcast > 0) {
+        for (int i = 0; i < ndim_broadcast; ++i) {
+            if (ndim_iter >= NPY_MAXDIMS) {
+                PyErr_SetString(PyExc_ValueError,
+                            "too many subscripts in einsum");
+                goto fail;
+            }
+            iter_labels[ndim_iter++] = 0;
         }
     }
 
--- a/numpy/_core/tests/test_einsum.py
+++ b/numpy/_core/tests/test_einsum.py
@@ -830,6 +830,19 @@ class TestEinsum:
             assert_equal(np.einsum('...lmn,lmno->...o', A, B,
                                    optimize=opt), ref)  # used to raise error
 
+    def test_einsum_broadcast_30349(self):
+        # Issue 30349 related to broadcasting with ellipsis
+        # np.einsum with optimize=True was not handling broadcasting
+        # correctly when ellipsis were involved (e.g. 'i...->i')
+        x = np.array([[[1,2,3],[1,2,3],[1,2,3]],
+                    [[1,2,3],[1,2,3],[1,2,3]],
+                    [[1,2,3],[1,2,3],[1,2,3]]])
+        
+        print(np.einsum("i...->i", x, optimize=False))
+
+        assert_equal(np.einsum("i...->i", x, optimize=False), np.array([18, 18, 18]))
+        assert_equal(np.einsum("i...->i", x, optimize=True), np.einsum("i...->i", x, optimize=False))
+
     def test_einsum_fixedstridebug(self):
         # Issue #4485 obscure einsum bug
         # This case revealed a bug in nditer where it reported a stride
--- a/numpy/_core/tests/test_einsum.py
+++ b/numpy/_core/tests/test_einsum.py
@@ -832,18 +832,50 @@ class TestEinsum:
 
     def test_einsum_broadcast_30349(self):
         # Issue #30349 related to broadcasting with ellipsis
-        x = np.tile(np.array([1,2,3]), (3, 3, 1))
+        # Specific case that was failing
+        x = np.tile(np.array([1, 2, 3]), (3, 3, 1))
 
         assert_equal(np.einsum("i...->i", x, optimize=False), np.array([18, 18, 18]))
         assert_equal(np.einsum("i...->i", x, optimize=True), np.einsum("i...->i", x, optimize=False))
 
-        # Test other subscripts for same results with both paths
-        for subscript in {"...i->i", "ij...->i", "ij...->ij",
-                          "...ij->i", "...ij->ij", "i...j->i",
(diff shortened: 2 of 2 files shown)


TomPOGET:
Apologies, accidental click... I'll fix the description immediately

numpy/_core/src/multiarray/einsum.cpp
```diff
@@ -83,7 +83,6 @@ parse_operand_subscripts(char *subscripts, int length,
                              "in operand %d", iop);
                 return -1;
             }
-
```
jorenham, line 86: this seems unnecessary

numpy/_core/src/multiarray/einsum.cpp
```diff
@@ -1049,7 +1050,7 @@ PyArray_EinsteinSum(char *subscripts, npy_intp nop,
 
     /* Allocate the iterator */
     iter = NpyIter_AdvancedNew(nop+1, op, iter_flags, order, casting, op_flags,
-                               op_dtypes, ndim_iter, op_axes, NULL, 0);
+                            op_dtypes, ndim_iter, op_axes, NULL, 0);
```
jorenham: this seems unnecessary

numpy/_core/src/multiarray/einsum.cpp
TomPOGET: It was, I removed this change

numpy/_core/src/multiarray/einsum.cpp
TomPOGET: Removed

TomPOGET:
Fixed the linting.
The failure on the gcc stage doesn't seem directly linked to the content of the PR (failure to fetch some archives when setting up up the job)...? If it is I'm interested in any advice on how to fix it.

numpy/_core/tests/test_einsum.py
```diff
@@ -830,6 +830,19 @@ def test_einsum_broadcast(self):
...
 
+    def test_einsum_broadcast_30349(self):
+        # Issue 30349 related to broadcasting with ellipsis
+        # np.einsum with optimize=True was not handling broadcasting
+        # correctly when ellipsis were involved (e.g. 'i...->i')
+        x = np.array([[[1, 2, 3], [1, 2, 3], [1, 2, 3]],
+                    [[1, 2, 3], [1, 2, 3], [1, 2, 3]],
+                    [[1, 2, 3], [1, 2, 3], [1, 2, 3]]])
+
+        print(np.einsum("i...->i", x, optimize=False))
```
MaanasArora: We shouldn't have print statements in the test suite I think.

numpy/_core/tests/test_einsum.py
```diff
@@ -830,6 +830,19 @@ def test_einsum_broadcast(self):
...
+
+        assert_equal(np.einsum("i...->i", x, optimize=False), np.array([18, 18, 18]))
+        assert_equal(np.einsum("i...->i", x, optimize=True), np.einsum("i...->i", x, optimize=False))
```
MaanasArora, line 839: We should add several more examples. In that case, it might be good to simply parametrize and do something like (with parametrized subscript):

```python
result = np.einsum(subscript, a, optimize=True)
expected = np.einsum(subscript, a, optimize=False)
assert_equal(result, expected)
```

numpy/_core/tests/test_einsum.py
```diff
@@ -830,6 +830,19 @@ def test_einsum_broadcast(self):
             assert_equal(np.einsum('...lmn,lmno->...o', A, B,
                                    optimize=opt), ref)  # used to raise error
 
+    def test_einsum_broadcast_30349(self):
+        # Issue 30349 related to broadcasting with ellipsis
+        # np.einsum with optimize=True was not handling broadcasting
+        # correctly when ellipsis were involved (e.g. 'i...->i')
+        x = np.array([[[1, 2, 3], [1, 2, 3], [1, 2, 3]],
```
MaanasArora: Nit: this seems a bit verbose. Is it better to use `np.tile`?

numpy/_core/tests/test_einsum.py
```diff
@@ -830,6 +830,19 @@ def test_einsum_broadcast(self):
             assert_equal(np.einsum('...lmn,lmno->...o', A, B,
                                    optimize=opt), ref)  # used to raise error
 
+    def test_einsum_broadcast_30349(self):
+        # Issue 30349 related to broadcasting with ellipsis
+        # np.einsum with optimize=True was not handling broadcasting
+        # correctly when ellipsis were involved (e.g. 'i...->i')
```
MaanasArora: Nit: Is it nice to have such a long description? It's in the issue. Not sure.

And again, we should add a test, maybe a different one, to check that this still rejects the ambiguous cases in the issue.

MaanasArora (commented):
Thanks, this looks correct as far as I can tell! The idea of adding broadcast (zero) labels to the iterator makes sense to me - unless I'm missing anything, that follows along well the idea of 'copying' the broadcast indices from the inputs.

A maintainer can make the call about whether we want this in. If we do, since this is an extension IMO, it would be really good to add more tests (which we can parametrize if they get big). Also, I checked that this still rejects the `i...,i...->i` path with different dims, so that is good, but we should add a test for that and other such dubious cases.

numpy/_core/tests/test_einsum.py
TomPOGET: That... was a very ugly oversight. It will be fixed in the next commit, very shortly.

TomPOGET:
> Also, I checked that this still rejects the `i...,i...->i` path with different dims, so that is good, but we should add a test for that and other such dubious cases.

I'm surprised to read that; as I mentioned in the issue itself, this seemed like a different task and a potential breaking change so I thought wiser to leave it for another PR. It would also impact the optimize=True path of the function, which for now I didn't modify (I only tried to make optimize=False and optimize=True match, considering optimize=True as the correct one as per the issue).
Which is to say in this PR, on my end, the syntax np.einsum("i...,i...->i", x, x[0]), with both values of optimize, is not rejected (keeping the same definition for x, it gives [42, 42, 42]).

numpy/_core/tests/test_einsum.py
TomPOGET: Done

numpy/_core/tests/test_einsum.py
TomPOGET: I have added a few more examples; test_einsum_broadcast (lines 791 to 831) also covers some with more dimensions.

MaanasArora:
> I'm surprised to read that; as I mentioned in the issue itself, this seemed like a different task and a potential breaking change

I'm sorry, must be a mistake in testing - you're right, the `"i...,i...->i"` case is accepted in this PR regardless of the value of `optimize`. That's a little more of an ambiguous change IMO, so I'll defer this to maintainer review, but I'd lean towards not accepting that. To be clear - I'm only talking about the `optimize=False` case with this, in which case this PR changes behavior. We may want to keep it untouched even if `optimize=True` passes (basically, make no change in behavior for this subscript); at lest more discussion might be needed. How to do that: you may want to keep this to the single op case or add other restrictions to the if statement.

numpy/_core/tests/test_einsum.py
MaanasArora: Thanks, using `pytest.parametrize` might be nicer though.

TomPOGET: It's now the case. Apologies, I'm still a bit new to this.

TomPOGET:
I see... I'll wait for more discussion on whether to change the behavior of np.einsum("i...,i...->i", x, x[0]), and such, in this PR or another. I'll try to figure out how to do that in the meantime.

seberg:
> whether to change the behavior of np.einsum("i...,i...->i", x, x[0]),

Sorry this has been a while, but as is, the relaxation here seems very unclear to me to relax this for the non-optimizing path if we probably want to go towards disabling it in both.
To me only a single `...` case is pretty obvious here right now.  (I still wonder a bit if there isn't a good chance for allowing `?` syntax, but I am not sure if the `?` syntax is quite sufficient)
Without following up on that, I am not sure we should go ahead with this as is.

mattip:
Sorry, I am going to close since it is not clear we can accept this.

PR #30572, opened 2026-01-04
TYP: ``numpy.core`` stubs 

jorenham:
This is needed to get the typing coverage score to 100%. It's just too bad that there exists no mechanism similar to `@warnings.deprecate`  that can be used to statically mark entire modules as deprecated, so we can't have the cake and eat it too.

The other option (which I'd prefer TBH), would be to get rid of `numpy.core`. It has been deprecated two years ago in NumPy 2.0, after all. So... should we? Because if so, then I'd happily close this and rip `numpy.core` out instead.

### DIFF
--- a/numpy/core/__init__.pyi
+++ b/numpy/core/__init__.pyi
@@ -1,0 +1,45 @@
+# deprecated module
+
+from types import ModuleType
+
+from . import (
+    _dtype,
+    _dtype_ctypes,
+    _internal,
+    arrayprint,
+    defchararray,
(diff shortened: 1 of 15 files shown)


charris:
> to get rid of numpy.core

IIRC, it is to allow pickles from NumPy versions < 2 to be unpickled. So we probably want to keep it a bit longer than usual.

EDIT: But no one should be using it otherwise.

charris:
Thanks Joren. Might as well have this for the 100% :)

PR #26780, opened 2024-06-21
MAINT: use an atomic load/store and a mutex to initialize the argparse and runtime import caches

ngoldbaum:
This is a re-do of #26430.

It adds locking to the argparse cache and the cached runtime imports, making both thread safe.

Both caches leverage a new `npy_atomic.h` header that wraps C11 atomics and MSVC intrinsics to implement a uint8 atomic load. I think this is sufficient to work on all supported platforms but we'll see what CI says.

First, we do a regular load to see if the cache is initialized. If it is, we read from the cache and go on our way. If it isn't, we do an atomic load to serialize access and then immediately acquire a lock. Whichever thread wins the race to acquire the lock fills in the cache and the other threads wait until the lock is released, acquire the lock, see the cache is filled, and continue. Since the lock is only used before the cache is filled in, a global lock is OK I think.

I could probably refactor both uses to make use of an abstract single-initialization API that takes a function pointer, but given that there are only two usages I don't think the extra abstraction and needing to make all the cache initialization functions have the same function signature is worth the effort.

PR #30573, opened 2026-01-05
MAINT: cleanup global state of multiarray

kumaraditya303:
This PR cleans up the global state in multiarray, the imports are cached in `npy_runtime_imports_struct` so the PyObjects fields in `npy_thread_unsafe_state` are redundant. Also renames it to `npy_global_state` as the leftover state is thread safe and initialized on module import.


seberg:
Looks good, but if @ngoldbaum wants to make a quick comment that would be nice. Technically, these are thread unsafe if modified at runtime, but in practice they should indeed only be modified at import time and I guess for subinterpreters, this may still have to be changed a bit either way.

ngoldbaum (approved):
Thanks! I guess I forgot to clean up all the PyObject pointers in #26780.

numpy/_core/src/multiarray/multiarraymodule.h
```diff
@@ -6,60 +6,11 @@ extern "C" {
 #endif
 
 /*
- * A struct storing thread-unsafe global state for the _multiarray_umath
- * module. We should refactor so the global state is thread-safe,
- * e.g. by adding locking.
+ * A struct storing global state for the _multiarray_umath
+ * module. The state is initialized when the module is imported
+ * so no locking is necessary to access it.
```
ngoldbaum, line 11:
```suggestion
 * so no locking is necessary to access it.
 *
 * These globals will need to move to per-module state to
 * support reloading or subinterpreters.
```

Maybe worth adding this note for future readers who want to add support for subinterpreters - which should be much easier now that we've done a nice job of isolating all this state!

seberg (approved):
Also LGTM, thanks! Hadn't realized most of this was just deleting unused definitions.

seberg:
Hmmm, an issue in the i686 test with:
```

Current thread 0xf74a9700 (most recent call first):
  File "/__w/numpy/numpy/venv/lib/python3.12/site-packages/numpy/_core/tests/test_dtype.py", line 985 in test_list_recursion
```
maybe just Pythons recursion detection not being perfect? I'll try to restart it, I don't think it is related these changes (nor can I think of a very recent PR).

seberg:
Just putting this in, that `loongarch` test is taking ridiculously long to build, thanks @kumaraditya303.

(The above error might be interesting to investigate if we see it more often, but it went away.)

PR #30575, opened 2026-01-05
TST: expose identity hashtable as capsule for easier testing

kumaraditya303:
This PR exposes identity hashtable as capsule for easier testing. I plan to use it to write explicit multi-threaded tests for the lock-free implementation.


numpy/_core/src/multiarray/_multiarray_tests.c.src
```diff
@@ -923,89 +923,120 @@ get_all_cast_information(PyObject *NPY_UNUSED(mod), PyObject *NPY_UNUSED(args))
...
+
 
-/*
- * Helper to test the identity cache, takes a list of values and adds
- * all to the cache except the last key/value pair.  The last value is
- * ignored, instead the last key is looked up.
- * None is returned, if the key is not found.
- * If `replace` is True, duplicate entries are ignored when adding to the
- * hashtable.
- */
```
ngoldbaum, line 934: Can you add comments like this for the new functions you're defining? It's much easier for future readers to understand what the C testing code does if they have some docstrings

numpy/_core/tests/test_hashtable.py
```diff
@@ -1,36 +1,47 @@
...
+def test_identity_hashtable_replace(key_length):
+    ht = create_identity_hash(key_length)
+
+    key = tuple(object() for _ in range(key_length))
+    val1 = object()
+    val2 = object()
+
+    identity_hash_set_item(ht, key, val1)
+    got = identity_hash_get_item(ht, key)
+    assert got is val1
```
ngoldbaum, line 40: Can you also add the testing setup that was in the old `test_identity_hashtable`? Just to make sure the old test case continues to work.

numpy/_core/tests/test_hashtable.py
```diff
@@ -1,36 +1,47 @@
...
 
-    # add a new item with a key that is already used and a new value, this
-    # should error if replace is False, see gh-26690
-    new_key = (keys_vals[1][0], object())
-    keys_vals[0] = new_key
```
ngoldbaum, line 34: Please also make sure this case is tested still. It's not clear to me this is preserved in the new testing code.

ngoldbaum (commented):
Overall looks like a nice cleanup. My big comment is to try to make sure your edits to the tests don't accidentally delete tests for cases we explicitly added tests for before. Try to make your test changes be additive instead of changing what is being tested.

Also, in-practice, we don't actually use the `replace` functionality in NumPy at all. We could also remove that check in the hashtable internals, always raise errors on replacement, and delete the tests for `replace=True`.

I believe @seberg originally wrote these tests and might have opinions.

ngoldbaum:
#30580 fixes the test failures in this PR.

kumaraditya303:
> Overall looks like a nice cleanup. My big comment is to try to make sure your edits to the tests don't accidentally delete tests for cases we explicitly added tests for before. Try to make your test changes be additive instead of changing what is being tested.

I intentionally split the tests into two: `test_identity_hashtable_get_set` and `test_identity_hashtable_replace` because the lock-free implementation does not support replacement so the replacement option would go away in future with the second test. 

The previous test had duplicate entries mixed which required replacement however the newly added tests explicitly tests for duplicates and non duplicate separately.

numpy/_core/tests/test_hashtable.py
kumaraditya303: No, the previous setup had mixed duplicate entries which I am trying to avoid here.

seberg:
> I believe @seberg originally wrote these tests and might have opinions.

Yeah, I am sure. But I don't have much of an opinion.  Seems like a very reasonable thing to expose the API more fine grained via a capsule to improve tests.
(Of course one day we may yet want to just delete this custom hashtable construct.)

numpy/_core/src/multiarray/_multiarray_tests.c.src
kumaraditya303: done

numpy/_core/tests/test_hashtable.py
kumaraditya303: See test_identity_hashtable_replace which tests for the replacement case

ngoldbaum:
OK, let's bring this in. Thanks for explaining about the test changes.

PR #30576, opened 2026-01-05
TYP: move ``poly1d`` stubs from ``__init__.pyi`` to ``lib/_polynomial_impl.pyi``

jorenham:
For the same reasons as https://github.com/numpy/numpy/pull/29915

This also fixes the incompatible overlapping overloads of `poly1d.__call__`.


jorenham:
The scipy primer error is inevitable, so nothing we can do about it.

charris:
> The scipy primer error is inevitable,

Curious as to why.

jorenham:
> > The scipy primer error is inevitable,
> 
> Curious as to why.

SciPy basically copy-pasted the overloads for `poly1d.__call__` from numpy, including the incompatible overlapping overloads: https://github.com/scipy/scipy/blob/328ad15aceb98de0f42a5f744195eebed34f601c/scipy/special/_orthogonal.pyi#L287-L292

So by correcting them here, SciPy's overridden `__call__` method signature no longer matches, and mypy complains about that.

jorenham:
> > The scipy primer error is inevitable,
> 
> Curious as to why.

There also seems to be a minor issue in scipy's stubs (not scipy-stubs, but the ones bundled with scipy), where `orthopoly1d.__call__` returns `poly1d` instead of `orthopoly1d` (itself). But that *shouldn't* matter here (although you never know with mypy).

charris:
Thanks Joren.

PR #30579, opened 2026-01-05
TYP: move ``busdaycalendar`` stubs from ``__init__.pyi`` to ``_core/multiarray.pyi``

jorenham:
See #29915 for the motivation


charris:
Thanks Joren.

PR #30580, opened 2026-01-05
MAINT: avoid passing ints to functions that take double

### LINKED ISSUE #30578: BUG: type-unsafe Cython C functions calls in RNG interface
Cython 3.2.4 came out yesterday and our CI started failing:

```
Error: Disallowed warnings detected:
  numpy/random/_generator.cp312-win32.pyd.p/numpy/random/_generator.pyx.c(26344): warning C4244: 'function': conversion from 'int64_t' to 'double', possible loss of data
  numpy/random/_generator.cp312-win32.pyd.p/numpy/random/_generator.pyx.c(38368): warning C4244: 'function': conversion from 'int64_t' to 'double', possible loss of data
```

https://github.com/numpy/numpy/actions/runs/20717331262/job/59471608457

Someone should investigate whether this is a bug in Cython or Cython is now newly catching an issue in NumPy's RNG code.

ngoldbaum:
Fixes #30578 

See https://cython.readthedocs.io/en/latest/src/userguide/fusedtypes.html#type-checking-specializations if you're unfamiliar with the fused types specialization syntax I'm using.

### DIFF
--- a/numpy/random/_common.pxd
+++ b/numpy/random/_common.pxd
@@ -26,12 +26,15 @@ cdef enum ConstraintType:
     LEGACY_CONS_NON_NEGATIVE_INBOUNDS_LONG
 
 ctypedef ConstraintType constraint_type
+ctypedef fused double_or_int64:
+    double
+    int64_t
 
 cdef object benchmark(bitgen_t *bitgen, object lock, Py_ssize_t cnt, object method)
 cdef object random_raw(bitgen_t *bitgen, object lock, object size, object output)
 cdef object prepare_cffi(bitgen_t *bitgen)
 cdef object prepare_ctypes(bitgen_t *bitgen)
-cdef int check_constraint(double val, object name, constraint_type cons) except -1
+cdef int check_constraint(double_or_int64 val, object name, constraint_type cons) except -1
 cdef int check_array_constraint(np.ndarray val, object name, constraint_type cons) except -1
 
 cdef extern from "include/aligned_malloc.h":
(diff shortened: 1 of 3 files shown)


charris:
I'm a bit confused, as the documentation states that the  fused type with greatest precision is used. Which of double and int64 is the type with greatest precision? We want double, but the warning says int64.

ngoldbaum:
> I'm a bit confused, as the documentation states that the fused type with greatest precision is used.

What documentation are you referring to?

Fused types allow you to write code that can be called with any one of an enumerated set of types. Precision doesn't come into place in my use here.

In this PR, call sites that pass in ints will use the int version and call sites that use doubles will call the double version. There are two different functions for each case in the generated C code.

Maybe the docs you're reading are talking about a fused type like `cython.integral` that can be any of the integer types? In that case precision does matter.

charris:
My impression was that what is wanted is a cast of int64 to double, with maybe a check that the value is preserved.

ngoldbaum:
I’m implementing @rkern’s suggestion from [here](https://github.com/numpy/numpy/issues/30578#issuecomment-3711612258).

> But I think it could be worth robustifying both, using np.arange() in the first instance and adding an integer check_constraint() variant in the second.

This is exactly that suggestion. I could have simply casted to double, but that could overflow and IMO this is better.

rkern (approved):
LGTM! Thank you!

seberg:
LGTM also, thanks all!

PR #30582, opened 2026-01-05
TYP: gradual generic scalar item type parameter defaults

jorenham:
Previously,  `_: np.number` would default to `_: np.number[Any, int | float | complex]`, which is equivalent to `_: np.number[Any, complex]` in static-typing-land. This would cause type-checkers to reject the following:

```py
def f() -> np.number: ...
x: np.number[Any, int | float] = np.number  # <--
```

It was rejected because `int | float | complex` is not assignable to `int | float`.

By changing the default type-parameter of `np.number` from `int | float | complex` to `Any`, type-checkers will interpret `_: np.number` as `_: np.number[Any]`. That way the example above will no longer be rejected.

The type-parameter defaults of `np.floating`, `np.flexible`, and `np.character` have also been changed in the same way.

### DIFF
--- a/numpy/__init__.pyi
+++ b/numpy/__init__.pyi
@@ -714,15 +714,10 @@ _NBitT2 = TypeVar("_NBitT2", bound=NBitBase, default=_NBitT1)  # pyright: ignore
 
 _ItemT_co = TypeVar("_ItemT_co", default=Any, covariant=True)
 _BoolItemT_co = TypeVar("_BoolItemT_co", bound=py_bool, default=py_bool, covariant=True)
-_NumberItemT_co = TypeVar("_NumberItemT_co", bound=complex, default=int | float | complex, covariant=True)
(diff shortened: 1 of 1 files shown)


jorenham:
I'm not sure why this resolves those `overload-overlap` errors in scipy-stubs ([here](https://github.com/scipy/scipy-stubs/blob/ff8029f171bd11fbb96e20a108c1d8c5adac3bce/scipy-stubs/sparse/linalg/_interface.pyi#L452-L457) and [here](https://github.com/scipy/scipy-stubs/blob/ff8029f171bd11fbb96e20a108c1d8c5adac3bce/scipy-stubs/stats/_morestats.pyi#L390-L407)), but I guess it confirms that those mypy errors are/were false positives (pyright doesn't report the same error, for example).

charris:
Thanks Joren.

PR #30583, opened 2026-01-05
ENH: add C memory leak regression tests

giampaolo:
Hello,
I'm the author of [psutil](https://github.com/giampaolo/psutil), and I recently released a package to detect memory leaks in C extensions called [psleak](https://github.com/giampaolo/psleak) (see [blog post](https://gmpy.dev/blog/2025/psutil-heap-introspection-apis)).

This is a draft PR showing how psleak could be integrated into NumPy's test suite as a regression tool that guarantees that no leaks are introduced by the C functions. It only covers a limited subset of NumPy APIs, and is mainly meant to illustrate the approach rather than provide exhaustive coverage. 

I’m not sure whether this is desirable or fits NumPy's testing philosophy, but since NumPy is mostly written in C it seemed like a good candidate to explore this idea.

### DIFF
--- a/numpy/tests/test_memleaks.py
+++ b/numpy/tests/test_memleaks.py
@@ -1,11 +1,9 @@
 import os
-import numpy as np
 
 import pytest
+from psleak import Checkers, LeakTest as LT, MemoryLeakTestCase
 
-from psleak import LeakTest as LT
-from psleak import MemoryLeakTestCase
-from psleak import Checkers
+import numpy as np
 
 api = {
     # array creation
--- a/numpy/tests/test_memleaks.py
+++ b/numpy/tests/test_memleaks.py
@@ -5,6 +5,11 @@ import numpy as np
 psleak = pytest.importorskip("psleak")
 LT = psleak.LeakTest
 
+# config
+psleak.MemoryLeakTestCase.verbosity = 1
+psleak.MemoryLeakTestCase.checkers = psleak.Checkers.exclude("gcgarbage")
+psleak.MemoryLeakTestCase.retries = 30  # minimize flaky failures
+
 api = {
     # array creation
     "zeros": LT(np.zeros, 10),
@@ -105,9 +110,6 @@ api = {
 
 
 class TestNumpyLeaks(psleak.MemoryLeakTestCase):
-    verbosity = 1
-    checkers = psleak.Checkers.exclude("gcgarbage")
-
     @classmethod
     def auto_generate(cls):
         return api
--- a/numpy/tests/test_memleaks.py
+++ b/numpy/tests/test_memleaks.py
@@ -1,9 +1,9 @@
-import os
-
 import pytest
-from psleak import Checkers, LeakTest as LT, MemoryLeakTestCase
 
 import numpy as np
+
+psleak = pytest.importorskip("psleak")
+LT = psleak.LeakTest
 
 api = {
     # array creation
@@ -104,12 +104,9 @@ api = {
 }
 
 
-@pytest.mark.skipif(
-    "PYTHONMALLOC" not in os.environ, reason="PYTHONMALLOC not set"
-)
-class TestNumpyLeaks(MemoryLeakTestCase):
+class TestNumpyLeaks(psleak.MemoryLeakTestCase):
     verbosity = 1
-    checkers = Checkers.exclude("gcgarbage")
+    checkers = psleak.Checkers.exclude("gcgarbage")
 
     @classmethod
     def auto_generate(cls):
--- a/numpy/tests/test_memleaks.py
+++ b/numpy/tests/test_memleaks.py
@@ -110,6 +110,7 @@ api = {
     "PYTHONMALLOC" not in os.environ, reason="PYTHONMALLOC not set"
 )
 class TestNumpyLeaks(MemoryLeakTestCase):
+    # foo
     verbosity = 0
     checkers = Checkers.exclude("gcgarbage")
 
--- a/numpy/tests/test_memleaks.py
+++ b/numpy/tests/test_memleaks.py
@@ -110,8 +110,7 @@ api = {
     "PYTHONMALLOC" not in os.environ, reason="PYTHONMALLOC not set"
 )
 class TestNumpyLeaks(MemoryLeakTestCase):
-    # foo
-    verbosity = 0
+    verbosity = 1
     checkers = Checkers.exclude("gcgarbage")
 
     @classmethod
--- a/numpy/tests/test_memleaks.py
+++ b/numpy/tests/test_memleaks.py
@@ -1,0 +1,118 @@
+import os
+import numpy as np
+
+import pytest
+
+from psleak import LeakTest as LT
+from psleak import MemoryLeakTestCase
+from psleak import Checkers
+
+api = {
+    # array creation
+    "zeros": LT(np.zeros, 10),
+    "ones": LT(np.ones, 10),
+    "empty": LT(np.empty, 10),
+    "full": LT(np.full, 10, 7),
+    "arange": LT(np.arange, 10),
+    "linspace": LT(np.linspace, 0, 1, 10),
+    "eye": LT(np.eye, 3),
+    "identity": LT(np.identity, 3),
+    "diag": LT(np.diag, [1, 2, 3]),
+    "copy": LT(np.copy, np.arange(5)),
+    # dtype / casting
+    "astype": LT(np.ndarray.astype, np.arange(5), np.float64),
+    # basic math / ufuncs
+    "add": LT(np.add, 1, 2),
+    "subtract": LT(np.subtract, 5, 3),
+    "multiply": LT(np.multiply, 3, 4),
+    "divide": LT(np.divide, 10, 2),
+    "mod": LT(np.mod, 10, 3),
+    "power": LT(np.power, 2, 3),
+    "sqrt": LT(np.sqrt, 16.0),
+    "exp": LT(np.exp, 1.5),
+    "log": LT(np.log, 10.0),
+    "log2": LT(np.log2, 8.0),
+    "log10": LT(np.log10, 100.0),
+    "sin": LT(np.sin, 0.5),
+    "cos": LT(np.cos, 0.5),
+    "tan": LT(np.tan, 0.5),
+    "arcsin": LT(np.arcsin, 0.5),
+    "arccos": LT(np.arccos, 0.5),
+    "arctan": LT(np.arctan, 0.5),
+    # comparisons / boolean
+    "greater": LT(np.greater, 3, 2),
+    "greater_equal": LT(np.greater_equal, 2, 2),
+    "less": LT(np.less, 1, 2),
+    "less_equal": LT(np.less_equal, 2, 2),
+    "equal": LT(np.equal, 3, 3),
+    "not_equal": LT(np.not_equal, 3, 2),
+    "logical_and": LT(np.logical_and, True, False),
+    "logical_or": LT(np.logical_or, True, False),
+    "logical_not": LT(np.logical_not, True),
+    "logical_xor": LT(np.logical_xor, True, False),
+    # reductions
+    "sum": LT(np.sum, [1, 2, 3, 4]),
+    "mean": LT(np.mean, [1, 2, 3, 4]),
+    "max": LT(np.max, [1, 2, 3, 4]),
+    "min": LT(np.min, [1, 2, 3, 4]),
+    "argmax": LT(np.argmax, [1, 3, 2]),
+    "argmin": LT(np.argmin, [1, 3, 2]),
+    "prod": LT(np.prod, [1, 2, 3]),
+    "std": LT(np.std, [1, 2, 3, 4]),
+    "var": LT(np.var, [1, 2, 3, 4]),
+    # shape manipulation
+    "reshape": LT(np.reshape, np.arange(9), (3, 3)),
+    "transpose": LT(np.transpose, np.arange(9).reshape(3, 3)),
+    "ravel": LT(np.ravel, np.arange(9).reshape(3, 3)),
+    "squeeze": LT(np.squeeze, np.zeros((1, 3, 1))),
+    "expand_dims": LT(np.expand_dims, np.arange(3), 0),
+    "flatten": LT(np.ndarray.flatten, np.arange(9).reshape(3, 3)),
+    # indexing / selection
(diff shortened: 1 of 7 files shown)


jorenham:
Thanks! Since you called this a draft, I went ahead and marked it as such; I hope you don't mind :)

jorenham:
The windows Cython failures are unrelated, see https://github.com/numpy/numpy/issues/30578

ngoldbaum:
Please feel free to shout if you need a maintainer to trigger CI. As an alternative, if you can find a way to merge a small NumPy PR separate from this one, that will allow you to run CI from a PR without approval from a maintainer.

numpy/tests/test_memleaks.py
```diff
@@ -0,0 +1,116 @@
+import os
+
+import pytest
+from psleak import Checkers, LeakTest as LT, MemoryLeakTestCase
```
ngoldbaum: you either need to make this import optional or add psleak as a test dependency in the test requirements file. I’m not sure where else it might be needed besides that.

giampaolo:
> Please feel free to shout if you need a maintainer to trigger CI. As an alternative, if you can find a way to merge a small NumPy PR separate from this one, that will allow you to run CI from a PR without approval from a maintainer.

Hi. At this stage I.m afraid I’ll have to shout quite a bit 🙂
Would it be possible to temporarily enable CI to run on every commit somehow? That would speed things up significantly. Right now I'm in the process of getting the CI scripts to pass, but there are many of them and they're fairly complex, so I'm very much in a trial-and-error phase.

I’m also unsure about the best approach going forward, specifically whether to:

- use a dedicated test_memleaks.yml workflow that only runs on Linux, or
- run the memory-leak tests on all platforms as part of the existing cibw_test_command.sh script.

I don't know NumPy internals, but if the C code is pretty much the same regardless of the platform (UNIX vs. Windows), then running tests on Linux only probably makes more sense.

seberg:
I have to admit, I would be more interested in ASAN+leak-check runs if they are not insanely slow (the main problem with valgrind here is that it is slow enough that it isn't fun to run).
As that would allow finding new leaks rather than just once covered by regression tests (which I think is more interesting).

Although, we do have a bit of API with raw allocations where you do need valgrind/sanitizers to really detect it (as opposed to reference count leaks that are more directly detectable).

---

Doesn't mean that this may not be an interesting path to at least strengthen the regression tests.

giampaolo:
> Although, we do have a bit of API with raw allocations where you do need valgrind/sanitizers to really detect it (as opposed to reference count leaks that are more directly detectable).

Note: psleak doesn't use [sys.getrefcount](https://docs.python.org/dev/library/sys.html#sys.getrefcount) or similar. It's more low level: it monitors what gets allocated on the heap via `malloc()` and related functions. It does so by using glibc's [mallinfo2](https://man7.org/linux/man-pages/man3/mallinfo2.3.html) (on Linux). As long as tests are run with `PYTHONMALLOC=malloc` env var set (which bypasses pymalloc, the internal cPython allocator), psleak will be able to detect things such as  `malloc()` without `free()`,  `PyMem_Malloc()` without `PyMem_Free()`, missing `Py_DECREF`,  missing `Py_CLEAR` etc., which is typically what you care about in C extension modules.

I don't know Valgrind in details, but in terms of pure "leak detection" `mallinfo2()` should grant the same level of accuracy. Or in other words, when you test a C function, psleak tells you you forgot to free memory _somewhere_, even if it doesn't tell you where exactly. 

Or to put it yet another way: (in this context) Valgrind is more about debugging, psleak is more about getting a failure.

ngoldbaum:
On reflection, I’m in agreement with Sebastian: it’d probably be easier and less overhead to update our existing address sanitizer testing to no longer disable leak detection.

Grep the codebase for `ASAN_OPTIONS` to see places where we set up ASan testing. You could also try the new pixi build definitions that were added to the repo recently to aid SciPy’s ASan testing.

Have you compared psleak with ASan memory leak detection?

As for including psleak in the numpy tests, we do have a decent amount of platform-specific code, particularly in the SIMD routines, but I’d probably start only looking at Linux, particularly if you’re working on Linux locally. If you can show that it’s a useful thing to include that finds real issues missed by other tools then it’s probably worth including. But I would want you to check that it finds issues missed by ASan, otherwise why not use our existing ASan tests to check these issues?

I don’t have permissions to give you CI access. Maybe if you notice a real memory leak as you’re working on this, you could send in a PR for that? Any small merged PR will give you sufficient permissions to run CI on this PR. It also might make sense to try smaller contributions before trying a big project that involves adding a new test dependency.

seberg:
Btw. as a general thing about `psleak` itself. It could be interesting to do what I did with `pytest-valgrind`.  I.e. add machinery to run before and after tests to find leaks.

Unfortunately, a very plausible/likely problem with this is that `pytest-valgrind` works well mostly because valgrind knows if things are still reachable.
This is one annoyance with `pytest-leaks`: a lot of testing machinery tends to "leak" some objects itself when running tests (even multiple times) and one needs to clean that up very well to be able to detect a real leak.

OTOH, this is likely not a problem when looking only at a complete test suite run (i.e. this is what we would do with ASAN, I think). `pytest-valgrind` is extra slow, because it looks at each test individually. That is great for tracking down a problem, but not all that important if we can run ASAN on every PR.

giampaolo:
@ngoldbaum wrote:

> Have you compared psleak with ASan memory leak detection?
> I would want you to check that it finds issues missed by ASan, otherwise why not use our existing ASan tests to check these issues?

To be honest I don't know what ASan does exactly, or what it can do in the context of NumPy specifically. 
psleak is more about detecting regressions, not root causes. It's more about integrating it into a test case that tells "this C function used to be fine, now it leaks".

Can ASan be instructed to do that? Can it be integrated in a test case? (again, I don't know)

ngoldbaum:
> Can ASan be instructed to do that? Can it be integrated in a test case? (again, I don't know)

I'm not sure how it compares since I'm not familiar with psleak. If you want to learn more about leak checking with address sanitizer, maybe take a look at https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer. If you want to run the NumPy test suite with ASan, take a look at https://numpy.org/doc/2.4/dev/development_advanced_debugging.html#compiler-sanitizers or the compiler sanitizer CI configurations in the .github folder.

ASan works by instrumenting the compiled binary. The leak checker overrides malloc, free, and other low-level memory management routines to add extra checks. This adds overhead but not too much, so it's feasible to run the entire NumPy test suite under ASan.

I don't know offhand if ASan would detect leaks in the new tests you've added without adding any new Python test dependencies. Would you mind checking that? If psleak does find issues that you insert by-hand but ASan misses them, that's a strong argument for including psleak tests.

Note that as I said above, currently our ASan tests disable leak detection. If we want to use ASan for this, we'll likely need to add new suppression files for known leaks in e.g. CPython.

seberg:
I think the main question if we can instrument ASan to do a leak check at runtime in principle. With valgrind that is possible, with gcc's leak sanitizer there seems to be: `__lsan_do_leak_check` to do this.

Hooking into `malloc`, I assume `psleak` does just that: a targeted instrumentation that can be done around individual tests.

If the clang leak sanitizer doesn't have `__lsan_do_leak_check` I still somewhat suspect that isn't too bad if it runs on every PR. You will know relatively well what might have caused the leak (and you can still escalate to valgrind or `psleak` which will tell you also what got leaked and can narrow down better _where_ it got leaked).

Although running `__lsan_do_leak_check` (or `psleak` mechanism) rather than instrumenting the complete test suite run might simplify suppressions a lot (i.e. because some leaks happen at import or so). But we won't know if that is a problem or not until we didn't try to just sit down and see if creating the suppressions is a rabbit hole or not.

rgommers:
Re running CI: that doesn't seem useful on this PR, we don't need to run 72 jobs repeatedly to test one new one. The standard way of working on a CI job is to enable it on one's one fork (that's one simple DEBUG commit away, triggering on the branch name and skipping any `if:` conditions) and iterate on it until it's ready. Then push it here for discussion/review with maintainers.

> I don’t have permissions to give you CI access. 

Aside from being undesirable, it isn't even possible to change this setting on a per-user or per-PR basis.

rgommers:
Thanks for sharing @giampaolo. More testing/debugging tools for these kinds of issues are certainly useful.

My first impression is that, given the coarse-grained approach and the need to call the functionality under test in the extension module many times in a loop, it's more useful as a focused debugging tool (analogously to plots with increasing memory use for some code snippets we sometimes get) rather than being suitable to run in CI. There's a lot of things implemented in C, so if the test here repeats calls 30x, it's going to be pretty slow to run in a comprehensive fashion.

> [_from the blog post_] If heap_used or mmap_used values increase consistently, you've found a native leak.

Or a reference cycle, right?

giampaolo:
Hi @rgommers, and thanks for chiming in.

> My first impression is that, given the coarse-grained approach and the need to call the functionality under test in the extension module many times in a loop, it's more useful as a focused debugging tool (analogously to plots with increasing memory use for some code snippets we sometimes get) rather than being suitable to run in CI. 

Actually, it's the opposite. You give psleak a set of C functions to test (possibly the entire C API). If one of them leak memory you get a failure. So it's de-facto a regression testing tool specifically meant to be **integrated into CI**. 

The intended goal is to catch cases where a C function allocates memory without freeing it. A classic example is forgetting `Py_DECREF`:


```c
PyObject *leak_list(PyObject *self, PyObject *args) {
    PyObject *obj = PyList_New(0);
    if (!obj)
        return NULL;

    // Do something with the list
    ...

    // Forget to decref it
    // Py_DECREF(obj);   // <--- Whoops! Here's the leak!
    
    Py_RETURN_NONE;
}
```

You write a test case for it:

```python
import psleak
import my_c_extension

class TestCase(psleak.MemoryLeakTestCase):
    def test_leaky_list(self):
        self.execute(my_c_extension.leak_list)
```

Then run it, and you will see a failure like this:

```
$ PYTHONMALLOC=malloc python3 -m pytest test_case.py
============================= test session starts =============================
collected 1 item                                                                                                                                                                    

test_case.py:7: in test_leaky_list
    self.execute(my_c_extension.leak_list)
psleak.py:730: in execute
    self._check_mem(
psleak.py:635: in _check_mem
    raise MemoryLeakError(msg)
E   psleak.MemoryLeakError: memory kept increasing after 5 runs
E   Run # 1: heap=+6384                                         (calls=  50, avg/call=+127)
E   Run # 2: heap=+7184                                         (calls=  75, avg/call=+95)
E   Run # 3: heap=+11248                                        (calls= 100, avg/call=+112)
E   Run # 4: heap=+11984                                        (calls= 125, avg/call=+95)
E   Run # 5: heap=+16048                                        (calls= 150, avg/call=+106)

=========================== short test summary info ===========================
FAILED test_case.py.py::TestCase::test_leak_leaky_list - psleak.MemoryLeakError: memory kept increasing after 5 runs
============================== 1 failed in 0.12s ==============================
```

What happened is that the missing `Py_DECREF` caused a constant heap-memory growth across multiple function calls, psleak detected it and triggered a **unit test failure**. 

Other detectable leak patterns that come to mind are:

```c
void *ptr = malloc(4096)                      // forget to free(ptr)
PyObject *obj = PyList_New(0);                // forget to Py_DECREF(obj)
PyObject *str = PyUnicode_FromString("hi");   // forget to Py_DECREF(str)
PyObject *caps = PyCapsule_New(ptr, NULL, NULL);  // forget Py_DECREF(caps)
char *buf = PyMem_Malloc(4096);               // forget to PyMem_Free(buf)
void *ptr = PyObject_Malloc(4096);            // forget to PyObject_Free(ptr)
```

If any of these patterns exist in NumPy, psleak will detect them and report a failure.
To give a rough idea, here's a grep from NumPy code base: 

```
~/svn/numpy$ grep -rE --include='*.{c,cpp}' 'PyMem_Malloc|PyMem_RawMalloc|PyMemNew|PyObject_Malloc|PyUnicode_FromString|PyList_New|PyDict_New|PyCapsule_New|alloc\(' | wc -l
636
```

Each of these 636 calls represent potential allocation sites that become leaks if not properly freed. Compare that to 3437 decref/clear/free calls:

```
~/svn/numpy$ grep -rE --include='*.{c,cpp}' 'Py_DECREF|Py_XDECREF|Py_CLEAR|PyMem_Free|free\(' . | wc -l
3437
```

That means roughly 1 allocation for every 5-6 cleanups. Plenty of room for human error in such a large C codebase.

> There's a lot of things implemented in C, so if the test here repeats calls 30x, it's going to be pretty slow to run in a comprehensive fashion.

Actually no. In my branch covering around 100 NumPy APIs, the full memleak test suite runs in ~3.5 seconds (see [test_memleaks.py](https://github.com/giampaolo/numpy/blob/98cbdaed1d0c0670aa85b104f40954f3d23db7ed/numpy/tests/test_memleaks.py
)). I haven't tried the ASan approach suggested by @ngoldbaum yet, but I expect ASan to be *a lot* slower than psleak.
Reason: psleak does not hook into `malloc()` like ASan or Valgrind do. It simply calls the C function N times, and then asks the OS how much memory is allocated in the heap (via [mallinfo2](https://man7.org/linux/man-pages/man3/mallinfo2.3.html) on Linux). That's as fast as querying process RSS. Things get slower only when the tested C function has a leak: in that case the function will be repeated more times to make sure the leak is not a false positive.

> Or a reference cycle, right?

Correct, reference cycles (both C-level and Python-level) can also be detected. There's even 2 test cases for it in psleak:
- C cycle: https://github.com/giampaolo/psleak/blob/d719860/tests/test_ext.c#L376-L389
- Python cycle: https://github.com/giampaolo/psleak/blob/d719860/tests/test_python_leaks.py#L39-L52

If you skim through that `test_ext.c` file you'll see other patterns which psleak is able to recognize.

> The standard way of working on a CI job is to enable it on one's one fork

You're absolutely right. My bad! I will keep working on my branch until CI turns green, and then report back here to continue discussion.

Thanks again for the feedback.

rgommers:
Thanks for the detailed answer @giampaolo. I think the main thing that sticks out is that you expect to be writing _separate_ tests for this. That is going to lead to duplication as well as have limited coverage. I assume in my previous reply that it would run with the existing test suite - slower, but much easier to maintain and with comprehensive coverage.

seberg:
Especially since we have the lsan job now, I don't think there is much appetite for this for now, thanks.

giampaolo:
Totally understandable. Sorry for keeping this stagnating.

PR #30585, opened 2026-01-06
BLD: Avoiding conflict with pygit2 for static build

shadchin:
`pygit2` also defines this symbol - https://github.com/libgit2/pygit2/blob/master/src/pygit2.c#L411

If you build statically, there will be a name conflict. Common practice is to write `static PyMethodDef module_methods[] = ...`.

Related https://github.com/libgit2/pygit2/pull/1446


seberg (approved):
Thanks, this shouldn't be used (although it would be nice to change the default linking to avoid this type of thing).

PR #30595, opened 2026-01-06
BLD: Avoiding conflict with pygit2 for static build 

charris:
Backport of #30585.

`pygit2` also defines this symbol - https://github.com/libgit2/pygit2/blob/master/src/pygit2.c#L411

If you build statically, there will be a name conflict. Common practice is to write `static PyMethodDef module_methods[] = ...`.

Related https://github.com/libgit2/pygit2/pull/1446


charris:
The test failure is unrelated.

PR #30599, opened 2026-01-07
DOC: Numpy for Matlab users - Add omitted 'np.' in examples missing it

SF-PVL:
The **NumPy for MATLAB users** documentation page was inconsistent in whether `np.` was being used in the NumPy column. I've added it to all the cases where it was missing.

Left it out for 2 cases "eigs", "signal.resample" and "cg". These functions do not seem to exist in NumPy. "signal.resample" and "cg" are in SciPy, but "eigs" is in neither. Left these as-is for now, as they don't work with or without the `np.`.

[skip ci]

### DIFF
--- a/doc/source/user/numpy-for-matlab-users.rst
+++ b/doc/source/user/numpy-for-matlab-users.rst
@@ -95,7 +95,7 @@ General purpose equivalents
      - Notes
 
    * - ``help func``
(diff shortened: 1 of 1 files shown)


jorenham:
None of these are numpy functions. They're either from ipython or scipy

PR #30601, opened 2026-01-07
BUG: Undo result type change of quantile/percentile but keep ``q`` precision

seberg:
Defer cast to input type to the end of the computation by passing through the `weak_q` information.

This fixes gh-30586, but I am not sure it is great to backport. Although, compared to the other PR it should just align the nan-functions fully and otherwise revert anything that isn't just a precision fix.

---

If we don't want to backport this, I would be tempted to just revert the original PR for 2.5, TBH.

**~This still needs tests, although~ it passes all tests added in gh-29105.**

### DIFF
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -4227,6 +4227,7 @@ def percentile(a,
     if a.dtype.kind == "c":
         raise TypeError("a must be an array of real numbers")
 
+    weak_q = type(q) in (int, float)  # use weak promotion for final result type
     q = np.true_divide(q, 100, out=...)
     if not _quantile_is_valid(q):
         raise ValueError("Percentiles must be in the range [0, 100]")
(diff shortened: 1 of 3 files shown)


seberg:
(One tiny effect here is that previously, for arrays of longdouble, i.e. high precision, the `q` computation also used that higher precision. Since it now sticks to it's own default precision until the end -- we do lose that part, but I doubt there is much point in trying that hard...)

charris:
I will put this before 2.4.1.

mhvk (approved):
This looks good. The direct test for `in (int, float)` make me wonder, though, if this isn't something that can be done with `_array_converter`? But you probably considered that, since you added that... Anyway, if that doesn't work for some reason, or is just too big an overhaul for this fixup, fine to go with this.

charris:
Thanks Sebastian.

seberg:
Yeah, I considered it and thought it doesn't matter. `_array_converter` also solves the `__array_wrap__` which I guess doesn't apply (at least not for this fix).
OTOH, it might still be a neat thing, just to try to remove explicit type checks...

seberg:
Dang it, I forgot to undo one line of code and the tests didn't notice!

PR #30604, opened 2026-01-07
DEP: remove `numpy.chararray`

jorenham:
This was decided in today's triage meeting.

See #30605 for the *actual* deprecation of `chararray` itself.


seberg:
Thanks Joren! this has been around since 2.0, so I think it's safe to finalize (especially early in 2.5).

PR #30606, opened 2026-01-07
TST: remove ``python<3.12`` tests

jorenham:
We probably missed these because of the double negation :upside_down_face:


charris:
Thanks Joren.

PR #30553, opened 2025-12-31
BLD: update vendored Meson to 1.9.2

rgommers:
Pulls in https://github.com/numpy/meson/pull/24

(draft until CI is green, that PR is merged, and this PR retargeted to the merge commit)

PR #30608, opened 2026-01-08
BLD: update vendored Meson to 1.9.2

charris:
Backport of #30553. It is best to keep the current release branch in sync with main to avoid accidental version changes.


rgommers (approved):
Makes sense, thanks Chuck. 

The 1.9.2 merge was complicated and I think I broke something somewhere, but nothing that can be hit by a `numpy` user/packager AFAIK, so let's trust our CI.

PR #30609, opened 2026-01-08
CI: build numpy with 2 jobs for faster sanitizers CI

kumaraditya303:
This reduces the build time by 1-2 mins, not a lot but overtime this would save a lot of CI time.

### DIFF
--- a/.github/workflows/compiler_sanitizers.yml
+++ b/.github/workflows/compiler_sanitizers.yml
@@ -69,7 +69,7 @@ jobs:
         pip uninstall -y pytest-xdist
     - name: Build
       run:
-        python -m spin build -j4 -- -Db_sanitize=address,undefined -Db_lundef=false
+        python -m spin build -j2 -- -Db_sanitize=address,undefined -Db_lundef=false
     - name: Test
       run: |
         # pass -s to pytest to see ASAN errors and warnings, otherwise pytest captures them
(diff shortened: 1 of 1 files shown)


rgommers:
Hmm this looks like it should be slower. By default `spin build` uses all cores (`2n+2` processes actually, which may or may not be faster than using `n` processes, that depends), and there are 4 available. You had `-j4` first, why switch to `-j2`?

kumaraditya303:
> You had -j4 first, why switch to -j2?

-j2 is faster than -j4 on CI so I switched to it, other CI workflows use -j2 as well. On this PR clang_ASAN_UBSAN takes 3:35 and on https://github.com/numpy/numpy/actions/runs/20825233286/job/59824692846?pr=30593 it takes 4:58 so almost 1.5 minutes speedup.

rgommers:
I suspect a large part is run to run variation, since the test phase is also 3 min slower in that linked log. There doesn't seem much difference between `-j4` and `-j2` here:

With `-j4`:
image

With `-j2`:

image

That said, I can imagine fewer cores could help specifically for sanitizers because of lowered memory usage. So let's go with this.

rgommers (approved):
Thanks @kumaraditya303

charris:
Code scanning is complaining that the spin upgrade isn't pinned with a hash.

rgommers:
@charris that's not related to this PR, and I'm not seeing that warning so quickly. Can you share a link to a CI log?

PR #30611, opened 2026-01-08
TST: remove broken & outdated deprecation tests

jorenham:
These test classes did not start with `Test*`, so pytest ignored them. Both were testing deprecations that have already been expired, so I removed them.


charris (approved):

charris:
Thanks Joren.

PR #30615, opened 2026-01-09
BUG: np.take out dtype

### LINKED ISSUE #25588: BUG: np.take cast to out argument
### Describe the issue:

`np.take(a, indices, out=out)` fails if `out.dtype` is not exactly `a.dtype`. The doc says to `out`:
> It should be of the appropriate shape and dtype.

But what is *appropriate*?

```python
import numpy as np
a = np.arange(3).astype(np.int32)
indices = np.arange(2)
out = np.zeros_like(indices, dtype=np.int64)
np.take(a, indices, out=out)
```

### Error message:

```shell
File ~/github/numpy/build-install/usr/lib/python3.11/site-packages/numpy/_core/fromnumeric.py:59, in _wrapfunc(obj, method, *args, **kwds)
     58 try:
---> 59     return bound(*args, **kwds)
     60 except TypeError:
     61     # A TypeError occurs if the object does have such a method in its
     62     # class, but its signature is not identical to that of NumPy's. This
   (...)
     66     # Call _wrapit from within the except clause to ensure a potential
     67     # exception has a traceback chain.

TypeError: Cannot cast array data from dtype('int64') to dtype('int32') according to the rule 'safe'

During handling of the above exception, another exception occurred:

Cell In[4], line 1
----> 1 np.take(a, indices, out=out)

File ~/github/numpy/build-install/usr/lib/python3.11/site-packages/numpy/_core/fromnumeric.py:192, in take(a, indices, axis, out, mode)
     95 @array_function_dispatch(_take_dispatcher)
     96 def take(a, indices, axis=None, out=None, mode='raise'):
     97     """
     98     Take elements from an array along an axis.
     99 
   (...)
    190            [5, 7]])
    191     """
--> 192     return _wrapfunc(a, 'take', indices, axis=axis, out=out, mode=mode)

File ~/github/numpy/build-install/usr/lib/python3.11/site-packages/numpy/_core/fromnumeric.py:68, in _wrapfunc(obj, method, *args, **kwds)
     59     return bound(*args, **kwds)
     60 except TypeError:
     61     # A TypeError occurs if the object does have such a method in its
     62     # class, but its signature is not identical to that of NumPy's. This
   (...)
     66     # Call _wrapit from within the except clause to ensure a potential
     67     # exception has a traceback chain.
---> 68     return _wrapit(obj, method, *args, **kwds)

File ~/github/numpy/build-install/usr/lib/python3.11/site-packages/numpy/_core/fromnumeric.py:45, in _wrapit(obj, method, *args, **kwds)
     43 except AttributeError:
     44     wrap = None
---> 45 result = getattr(asarray(obj), method)(*args, **kwds)
     46 if wrap:
     47     if not isinstance(result, mu.ndarray):

TypeError: Cannot cast array data from dtype('int64') to dtype('int32') according to the rule 'safe'
```

'numpy_version': '2.0.0.dev0+git20240115.bbdd595'

### Runtime Environment:

not important

### Context for the issue:

`out = np.take(a, indices)` works, but passing the `out` arg explicitly saves intermediate memory and could be a tiny bit faster.

himanow:
### Description
This PR addresses the issue in `np.take` where an error was raised when input and output dtypes were different during casting. 

Key changes:
- Allows 'same-kind' casting to proceed in `np.take`.
- Emits a `DeprecationWarning` for other casting types to maintain backward compatibility while signaling future changes.

This is my first contribution to NumPy, so I'm opening this as a **draft** to ensure the implementation and formatting align with the project's standards. I'd appreciate any feedback on the code or the CI results.

### Fixes
Closes #25588

### DIFF
--- a/numpy/_core/src/multiarray/item_selection.c
+++ b/numpy/_core/src/multiarray/item_selection.c
@@ -312,9 +312,9 @@ PyArray_TakeFrom(PyArrayObject *self0, PyObject *indices0, int axis,
         dtype = PyArray_DESCR(self);
         Py_INCREF(dtype);
         out_dtype = PyArray_DESCR(out);
+        Py_INCREF(out_dtype);
         if (dtype == out_dtype) {
-            obj = (PyArrayObject *)out;
-            Py_INCREF(obj);
+            obj = (PyArrayObject *)PyArray_FromArray(out, dtype, flags);
         }
         else {
             if (PyArray_CanCastTypeTo(dtype, out_dtype, NPY_SAME_KIND_CASTING) == 0) {
@@ -329,7 +329,6 @@ PyArray_TakeFrom(PyArrayObject *self0, PyObject *indices0, int axis,
             }
             flags |= NPY_ARRAY_FORCECAST;
             obj = (PyArrayObject *)PyArray_FromArray(out, dtype, flags);
-            Py_INCREF(obj);
         }
         if (obj == NULL) {
             goto fail;
--- a/numpy/_core/src/multiarray/item_selection.c
+++ b/numpy/_core/src/multiarray/item_selection.c
@@ -232,7 +232,7 @@ NPY_NO_EXPORT PyObject *
 PyArray_TakeFrom(PyArrayObject *self0, PyObject *indices0, int axis,
                  PyArrayObject *out, NPY_CLIPMODE clipmode)
 {
-    PyArray_Descr *dtype;
+    PyArray_Descr *dtype, *out_dtype;
     PyArrayObject *obj = NULL, *self, *indices;
     npy_intp nd, i, n, m, max_item, chunk, itemsize, nelem;
     npy_intp shape[NPY_MAXDIMS];
@@ -311,7 +311,26 @@ PyArray_TakeFrom(PyArrayObject *self0, PyObject *indices0, int axis,
         }
         dtype = PyArray_DESCR(self);
         Py_INCREF(dtype);
-        obj = (PyArrayObject *)PyArray_FromArray(out, dtype, flags);
+        out_dtype = PyArray_DESCR(out);
+        if (dtype == out_dtype) {
+            obj = (PyArrayObject *)out;
+            Py_INCREF(obj);
+        }
+        else {
+            if (PyArray_CanCastTypeTo(dtype, out_dtype, NPY_SAME_KIND_CASTING) == 0) {
+                if (DEPRECATE(
+                            "Implicit casting of output to a different kind is "
+                            "deprecated. "
+                            "In a future version, this will result in an error. Please "
+                            "ensure the output has the same-kind type as the input.") <
+                    0) {
+                    goto fail;
+                }
+            }
+            flags |= NPY_ARRAY_FORCECAST;
+            obj = (PyArrayObject *)PyArray_FromArray(out, dtype, flags);
+            Py_INCREF(obj);
+        }
         if (obj == NULL) {
             goto fail;
         }
--- a/numpy/_core/src/multiarray/item_selection.c
+++ b/numpy/_core/src/multiarray/item_selection.c
@@ -312,12 +312,11 @@ PyArray_TakeFrom(PyArrayObject *self0, PyObject *indices0, int axis,
         dtype = PyArray_DESCR(self);
         out_dtype = PyArray_DESCR(out);
         if (dtype != out_dtype) {
-            /*Deprecated NumPy 2.5, 2026-01*/
-            if (PyArray_CanCastTypeTo(dtype, out_dtype, NPY_SAME_KIND_CASTING) == 0) {
+            /* Deprecated NumPy 2.5, 2026-01 */
+            if (!PyArray_CanCastTypeTo(dtype, out_dtype, NPY_SAME_KIND_CASTING)) {
                 if (DEPRECATE(
-                            "Implicit casting of output to a different kind is "
-                            "deprecated. "
-                            "In a future version, this will result in an error. (Deprecated NumPy 2.5)") <
+                        "Implicit casting of output to a different kind is deprecated."
+                        "In a future version, this will result in an error. (Deprecated NumPy 2.5)") <
                     0) {
                     goto fail;
                 }
@@ -325,7 +324,6 @@ PyArray_TakeFrom(PyArrayObject *self0, PyObject *indices0, int axis,
             flags |= NPY_ARRAY_FORCECAST;
         }
         Py_INCREF(dtype);
-        Py_INCREF(out_dtype);
         obj = (PyArrayObject *)PyArray_FromArray(out, dtype, flags);
         if (obj == NULL) {
             goto fail;
--- a/numpy/_core/src/multiarray/item_selection.c
+++ b/numpy/_core/src/multiarray/item_selection.c
@@ -310,26 +310,23 @@ PyArray_TakeFrom(PyArrayObject *self0, PyObject *indices0, int axis,
             flags |= NPY_ARRAY_ENSURECOPY;
         }
         dtype = PyArray_DESCR(self);
-        Py_INCREF(dtype);
         out_dtype = PyArray_DESCR(out);
-        Py_INCREF(out_dtype);
-        if (dtype == out_dtype) {
-            obj = (PyArrayObject *)PyArray_FromArray(out, dtype, flags);
-        }
-        else {
+        if (dtype != out_dtype) {
+            /*Deprecated NumPy 2.5, 2026-01*/
             if (PyArray_CanCastTypeTo(dtype, out_dtype, NPY_SAME_KIND_CASTING) == 0) {
                 if (DEPRECATE(
                             "Implicit casting of output to a different kind is "
                             "deprecated. "
-                            "In a future version, this will result in an error. Please "
-                            "ensure the output has the same-kind type as the input.") <
+                            "In a future version, this will result in an error. (Deprecated NumPy 2.5)") <
                     0) {
                     goto fail;
                 }
             }
             flags |= NPY_ARRAY_FORCECAST;
-            obj = (PyArrayObject *)PyArray_FromArray(out, dtype, flags);
         }
+        Py_INCREF(dtype);
+        Py_INCREF(out_dtype);
+        obj = (PyArrayObject *)PyArray_FromArray(out, dtype, flags);
         if (obj == NULL) {
             goto fail;
         }
--- a/numpy/_core/src/multiarray/item_selection.c
+++ b/numpy/_core/src/multiarray/item_selection.c
@@ -315,7 +315,7 @@ PyArray_TakeFrom(PyArrayObject *self0, PyObject *indices0, int axis,
             /* Deprecated NumPy 2.5, 2026-01 */
             if (!PyArray_CanCastTypeTo(dtype, out_dtype, NPY_SAME_KIND_CASTING)) {
                 if (DEPRECATE(
-                        "Implicit casting of output to a different kind is deprecated."
+                        "Implicit casting of output to a different kind is deprecated. "
                         "In a future version, this will result in an error. (Deprecated NumPy 2.5)") <
                     0) {
                     goto fail;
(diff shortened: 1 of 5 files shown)


jorenham:
CI isn't happy

himanow:
I rebased my branch onto the latest main, and all CI/CD checks are now passing

numpy/_core/tests/test_regression.py
```diff
@@ -1100,9 +1100,7 @@ def test_compress_small_type(self):
         b = np.zeros((2, 1), dtype=np.single)
         try:
             a.compress([True, False], axis=1, out=b)
-            raise AssertionError("compress with an out which cannot be "
-                                 "safely casted should not return "
-                                 "successfully")
+            assert_equal(b, np.array([[1.0], [3.0]]))
```
seberg: Can you please explain this change?

numpy/_core/tests/test_item_selection.py
```diff
@@ -81,6 +81,17 @@ def test_empty_partition(self):
...
+    def test_out_dtype(self):
+        # In reference to github issue #25588
+        a = np.arange(3).astype(np.int32)
+        indices = np.arange(2)
+        out = np.zeros_like(indices, dtype=np.int64)
+        np.take(a, indices, out=out)
+        assert_array_equal(a[indices], out)
+        diffrent_dtype_out = np.zeros_like(indices, dtype=np.uint32)
+        with pytest.warns(DeprecationWarning):
+            np.take(a, indices, out=diffrent_dtype_out)
```
seberg: Please move this test to `test_deprecations` and use the pattern used there.

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -311,7 +311,25 @@ PyArray_TakeFrom(PyArrayObject *self0, PyObject *indices0, int axis,
...
+        if (dtype == out_dtype) {
+            obj = (PyArrayObject *)PyArray_FromArray(out, dtype, flags);
+        }
+        else {
+            if (PyArray_CanCastTypeTo(dtype, out_dtype, NPY_SAME_KIND_CASTING) == 0) {
+                if (DEPRECATE(
+                            "Implicit casting of output to a different kind is "
+                            "deprecated. "
+                            "In a future version, this will result in an error. Please "
+                            "ensure the output has the same-kind type as the input.") <
```
seberg: Please add the comments we usually add to say when the deprecation happened (before it) and also inside the deprecation itself, such as `(deprecated NumPy 2.5)`.

The last sentence feels like unnecessary to me.  (I should think once more if we shouldn't just use safe casting, although then one might be tempted to ask for a `casting=` kwarg.)

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -311,7 +311,25 @@ PyArray_TakeFrom(PyArrayObject *self0, PyObject *indices0, int axis,
         }
         dtype = PyArray_DESCR(self);
         Py_INCREF(dtype);
```
seberg: This needs cleaning up, you are inserting code but that code interacts closely with this so you can't insert code between these two lines.
I.e. the `dtype` reference can be lost on error.

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -311,7 +311,25 @@ PyArray_TakeFrom(PyArrayObject *self0, PyObject *indices0, int axis,
...
+                            "Implicit casting of output to a different kind is "
+                            "deprecated. "
+                            "In a future version, this will result in an error. Please "
+                            "ensure the output has the same-kind type as the input.") <
+                    0) {
+                    goto fail;
+                }
+            }
+            flags |= NPY_ARRAY_FORCECAST;
+            obj = (PyArrayObject *)PyArray_FromArray(out, dtype, flags);
```
seberg: This is the _exact_ same code as the first branch, except for a flag that is irrelevant in the first branch.

himanow: Fixed in commit 0029b81.

numpy/_core/src/multiarray/item_selection.c
himanow: Fixed in 0029b81. Is it safe to insert code between `dtype = PyArray_DESCR(self); `and `Py_INCREF(dtype);`? Also, is the line `Py_INCREF(out_dtype);` unnecessary or redundant here?

numpy/_core/src/multiarray/item_selection.c
himanow: Fixed in commit 0029b81. I removed the last sentence.

numpy/_core/tests/test_item_selection.py
himanow: Fixed in commit 0029b81. Could you please check if the implementation and coding style are correct?

numpy/_core/tests/test_regression.py
himanow: The compress method uses `PyArray_TakeFrom` internally. Since it no longer raises an error after my changes, I have updated the test to verify that the output values match the expected results.

himanow:
Thank you for the detailed feedback. I have pushed the fixes in the latest commit.

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -310,7 +310,22 @@ PyArray_TakeFrom(PyArrayObject *self0, PyObject *indices0, int axis,
...
+                            "deprecated. "
+                            "In a future version, this will result in an error. (Deprecated NumPy 2.5)") <
+                    0) {
+                    goto fail;
+                }
+            }
+            flags |= NPY_ARRAY_FORCECAST;
+        }
         Py_INCREF(dtype);
+        Py_INCREF(out_dtype);
```
seberg: This incref isn't necessary here.

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -310,7 +310,22 @@ PyArray_TakeFrom(PyArrayObject *self0, PyObject *indices0, int axis,
...
         }
         dtype = PyArray_DESCR(self);
+        out_dtype = PyArray_DESCR(out);
+        if (dtype != out_dtype) {
+            /*Deprecated NumPy 2.5, 2026-01*/
+            if (PyArray_CanCastTypeTo(dtype, out_dtype, NPY_SAME_KIND_CASTING) == 0) {
+                if (DEPRECATE(
+                            "Implicit casting of output to a different kind is "
+                            "deprecated. "
+                            "In a future version, this will result in an error. (Deprecated NumPy 2.5)") <
```
seberg: Can you format this a bit prettier? (e.g. start at smaller indent, don't break the line when there is no `\n` anyway.

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -310,7 +310,22 @@ PyArray_TakeFrom(PyArrayObject *self0, PyObject *indices0, int axis,
             flags |= NPY_ARRAY_ENSURECOPY;
         }
         dtype = PyArray_DESCR(self);
+        out_dtype = PyArray_DESCR(out);
+        if (dtype != out_dtype) {
+            /*Deprecated NumPy 2.5, 2026-01*/
```
seberg:
```suggestion
            /* Deprecated NumPy 2.5, 2026-01 */
```

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -310,7 +310,22 @@ PyArray_TakeFrom(PyArrayObject *self0, PyObject *indices0, int axis,
...
+            if (PyArray_CanCastTypeTo(dtype, out_dtype, NPY_SAME_KIND_CASTING) == 0) {
```
seberg: Most code uses !PyArray_... for this type of pattern, so I would stick to it here too.  (The `< 0` is very common for errors.)

numpy/_core/tests/test_regression.py
seberg: OK, I guess this is fine to generalize. We may want to tighten the casting from same-kind at some point, but I don't think we have to worry about it now.

numpy/_core/tests/test_deprecations.py
```diff
@@ -342,3 +342,17 @@ class TestTooManyArgsExtremum(_DeprecationTestCase):
...
+    message = "Implicit casting of output to a different kind."
+
+    def test_out_dtype_deprecated(self):
+        a = np.arange(3).astype(np.int32)
+        indices = np.arange(2)
+        diffrent_dtype_out = np.zeros_like(indices, dtype=np.uint32)
+
+        self.assert_deprecated(
+            np.take, args=(a, indices), kwargs={"out": diffrent_dtype_out}
+        )
```
seberg: `different_dtype_out` or just shorten (small typo).  This is good as is, but while I wrote this args/kwargs, I actually like the pattern we use a lot more to just use a lambda, i.e. passing `lambda: np.take(a, indices, out=out)` (it even ends up shorter!)

numpy/_core/tests/test_item_selection.py
```diff
@@ -81,6 +81,14 @@ def test_empty_partition(self):
 
         assert_array_equal(a, a_original)
 
+    def test_out_dtype(self):
+        # In reference to github issue #25588
+        a = np.arange(3).astype(np.int32)
+        indices = np.arange(2)
+        out = np.zeros_like(indices, dtype=np.int64)
+        np.take(a, indices, out=out)
+        assert_array_equal(a[indices], out)
```
seberg, line 93: The code change looks like it is fine, but maybe you can parametrize the test to also test casting to a float output?  Right now it does a same-kind cast (if we look at it the wrong way around).

seberg (commented):
Thanks a few nits, mainly the refcount addition is incorrect and I think it would be good to add a test that goes across kind boundaries (if you looked at it the other way).

numpy/_core/src/multiarray/item_selection.c
himanow: "Updated as suggested."

numpy/_core/src/multiarray/item_selection.c
himanow: Fixed in 6f40af4e2.

numpy/_core/src/multiarray/item_selection.c
himanow: Fixed in 6f40af4e2.

numpy/_core/tests/test_item_selection.py
himanow: Fixed in 6f40af4e2. Updated the tests using pytest.mark.parametrize. I’m now testing casting from int32 input to various output dtypes: [np.int8, np.int16, np.int32, np.int64, np.float16, np.float32, np.float64, np.longdouble].

numpy/_core/tests/test_deprecations.py
himanow: Fixed in 6f40af4e2. I've renamed the variable to different_dtype_out. I also switched to the lambda pattern as you suggested—it definitely looks much cleaner and more concise.

numpy/_core/src/multiarray/item_selection.c
himanow: Fixed in 6f40af4e2. I've updated this to follow the !PyArray_... pattern.

himanow:
Thanks for the thorough review. I’ve updated the PR and addressed each comment.

himanow:
I have resolved the merge conflicts and rebased the branch onto the latest main.
CI is still in progress, but the PR is ready for further review.
I will keep monitoring the test results. Thank you.

himanow:
Hi @seberg , sorry for the ping, but I'm just checking in to see if there's anything else I should update for this PR. I’d appreciate any feedback when you have a moment"

seberg:
Sorry, I forgot about this.  We have conflicts now, and it would make sense to add a very brief (single bullet point is OK) release note for the new deprecation.
But yeah, I think this is good and I can make sure to just do those follow ups myself some time.

himanow:
Thank you! I appreciate you taking care of the conflicts and the release note. I look forward to it.

PR #30617, opened 2026-01-09
ENH: Add mask comparison support to `assert_array_equal` and `assert_equal`

### LINKED ISSUE #30564: ENH: equal_mask flag for assert_array_equal

Working with masked arrays, unit tests often need to check if a function returns the expected  masked array.
However, doing it with assert_array_equal is error prone because masked dimensions will not be compared.
Hence this will succeed:
```
ret = tested_function() # suppose returned np.ma.MaskedArray([4.0], mask=[True])
assert_array_equal(ret, np.array([3.0]))
```

May an `equal_mask` flag could be added so that the function compares that, for each item, both or none are masked

prathamhole14:
Fixes: #30564

I added `equal_mask` keyword arg to `assert_array_compare` allowing it to be passed in  `assert_array_equal` and `assert_equal`. I believe this should be done for other functions like `assert_array_almost_equal` which uses `assert_array_compare` internally, please let me know if this is required. 

Apologies in advance, I haven’t read the contribution guidelines yet. I’ll read them ASAP and make the necessary changes.

### DIFF
--- a/doc/release/upcoming_changes/30617.new_feature.rst
+++ b/doc/release/upcoming_changes/30617.new_feature.rst
@@ -1,0 +1,5 @@
+``equal_mask`` argument for masked array testing
+--------------------------------------------------
+A new ``equal_mask`` parameter has been added to `assert_equal`, `assert_array_equal`,
+and `assert_array_compare`. When ``True``, masks must match during comparison;
+masked positions are ignored. Default is ``False`` for backward compatibility.
--- a/doc/release/upcoming_changes/30617.new_feature.rst
+++ b/doc/release/upcoming_changes/30617.new_feature.rst
@@ -1,5 +1,5 @@
 ``equal_mask`` argument for masked array testing
---------------------------------------------------
+------------------------------------------------
 A new ``equal_mask`` parameter has been added to `assert_equal`, `assert_array_equal`,
 and `assert_array_compare`. When ``True``, masks must match during comparison;
 masked positions are ignored. Default is ``False`` for backward compatibility.
--- a/numpy/testing/_private/utils.py
+++ b/numpy/testing/_private/utils.py
@@ -253,7 +253,8 @@ def build_err_msg(arrays, err_msg, header='Items are not equal:',
     return '\n'.join(msg)
 
 
-def assert_equal(actual, desired, err_msg='', verbose=True, *, strict=False):
+def assert_equal(actual, desired, err_msg='', verbose=True, *, strict=False,
+                 equal_mask=False):
     """
     Raises an AssertionError if two objects are not equal.
 
@@ -283,6 +284,13 @@ def assert_equal(actual, desired, err_msg='', verbose=True, *, strict=False):
         parameter has no effect.
 
         .. versionadded:: 2.0.0
+
+    equal_mask : bool, optional
+        If True, masked arrays must have matching masks. When False (default),
+        mask differences are ignored for backward compatibility. Only applies
+        when comparing masked arrays.
+
+        .. versionadded:: 2.2.0
 
     Raises
     ------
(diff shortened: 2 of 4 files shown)


prathamhole14:
Please have a look at this PR and let me know the changes required.

doc/release/upcoming_changes/30617.new_feature.rst
```diff
@@ -0,0 +1,5 @@
+``equal_mask`` argument for masked array testing
+--------------------------------------------------
```
jorenham:
```suggestion
``equal_mask`` argument for masked array testing
------------------------------------------------
```

numpy/testing/_private/utils.py
```diff
@@ -284,6 +285,13 @@ def assert_equal(actual, desired, err_msg='', verbose=True, *, strict=False):
 
         .. versionadded:: 2.0.0
 
+    equal_mask : bool, optional
+        If True, masked arrays must have matching masks. When False (default),
+        mask differences are ignored for backward compatibility. Only applies
+        when comparing masked arrays.
+
+        .. versionadded:: 2.2.0
```
jorenham:
```suggestion
        .. versionadded:: 2.5.0
```

numpy/testing/_private/utils.py
```diff
@@ -1035,6 +1084,13 @@ def assert_array_equal(actual, desired, err_msg='', verbose=True, *,
 
         .. versionadded:: 1.24.0
 
+    equal_mask : bool, optional
+        If True, treat ``numpy.ma.MaskedArray`` mask positions as part of the
+        comparison: an AssertionError is raised when masks differ, and masked
+        positions are ignored when comparing the underlying data. When False
+        (default), mask differences are ignored and masked entries are filled
+        before comparison.
+
```
jorenham, line 1093:
```suggestion
        before comparison.

        .. versionadded:: 2.5.0
```

jorenham (commented):
I gave this a quick look and left some preliminary review suggestions. There's quite a lot of code, and reviewing this thoroughly is going to be pretty time consuming, so it might take us a while to get to it.

But it's important to note that we've been trying to move away from masked arrays lately, rather than add more support for it, as that would make it even more difficult. And seeing as this might send of the wrong signal in that aspect, So although this seems like a useful feature in isolation, I'm just not sure if this is actually something that we should add, given the context.

prathamhole14:
> But it's important to note that we've been trying to move away from masked arrays lately, rather than add more support for it, as that would make it even more difficult. And seeing as this might send of the wrong signal in that aspect, So although this seems like a useful feature in isolation, I'm just not sure if this is actually something that we should add, given the context.

That makes sense, it works for me.

mattip:
Is there a reason not to use `np.ma.allclose` or `np.ma.allequal`?

prathamhole14 (commented):
Sorry for the delayed response! @mattip yes we can use `np.ma.allequal(a, b, fill_value=False)` to catch mask differences, but when it fails you just get `AssertionError` with no details about what's wrong or where.
 
The advantage of `equal_mask` in `assert_array_equal` is you get helpful error messages showing exactly which indices have mask mismatches, making debugging much easier.
 
That said, I get @jorenham's point about not adding more masked array support. Happy to close this if it doesn't fit with NumPy's direction going forward.

mattip:
Can you make `np.ma.allequal`'s `AssertionError` more helpful instead of modifying `assert_array_equal`?

prathamhole14:
> Can you make `np.ma.allequal`'s AssertionError more helpful

Hi @mattip , this can be done but few thing to consider, `np.ma.allequal` and `np.ma.allclose` they both return bool (not assertion error like `np.ma.assert_array_equal` and `np.testing.assert_array_compare`).




If we where to still add the "detailed" error for `np.ma.allequal` then should this be done by adding a kwarg eg `verbose` as the users don't expect it to give out assertion error? (I am not in favor of this as it will make the function inconsistent of the return value).

I think the better approach would be to add a `masked_equal`/`fill_value` like argument to `np.ma.assert_array_compare` family.  Please let me know what do you think of this approach.

mattip:
I wonder how [marray](https://github.com/mdhaber/marray) handles this, and if we shouldn't be closing more masked array issues/PRs with "NumPy's masked array support is not complete. Use marray instead"

mattip:
I think adding np.ma.assert_array_equal like np.assert_array_equal is the best way forward

PR #30619, opened 2026-01-09
Fix memory leak in np.eye (gh-30618)

Abu018:
Replace problematic flat iterator assignment m[:M-k].flat[i::M+1] = 1 with explicit index computation to avoid memory leak.

The fix computes diagonal indices explicitly using arange, similar to the approach used in diagflat function.


jorenham:
This does not fix the underlying memory, but only circumvents it. If `m[:M-k].flat[i::M+1] = 1` is what's causing the leak, then that should be investigated further to find the actual source of the problem. It's a valid use-case after all. So `np.eye` itself is not what needs fixing; it's just an example of how this leak can be triggered.

PR #30621, opened 2026-01-09
Fix memory leak in np.eye (gh-30618)

Abu018:
Replace problematic flat iterator assignment m[:M-k].flat[i::M+1] = 1 with explicit index computation to avoid memory leak.

The fix computes diagonal indices explicitly using arange, similar to the approach used in diagflat function.


seberg:
Thanks, but this has long been fixed, the new release should happen on the weekend. And the fix is to fix the flat iterator, not to avoid using it.

jorenham:
This is also an exact duplicate of #30619 ...

PR #30633, opened 2026-01-10
TYP: ``dtype``: unit-based ``datetime64``/``timedelta64`` specialization

jorenham:
This adds dtype constructor overloads for `datetime64` and `timedelta64` strings with specific units. Type-checkers will now infer `dtype("datetime64[Y]")` as `dtype[datetime64[date]]` (previously `dtype[datetime64[Any]]`) and `dtype("timedelta64[ns]")` as `dtype[timedelta64[int]]` (previously `dtype[timedelta64[Any]]`). This can help improve type-safety in case mixed units are used. It's also helpful to know which internal representation will be used for a particular unit by hovering the dtype variable in your IDE.

This already happens in the scalar-type constructors, btw.

### DIFF
--- a/numpy/__init__.pyi
+++ b/numpy/__init__.pyi
@@ -97,7 +97,6 @@ from numpy._typing import (  # type: ignore[deprecated]
     _ULongLongCodes,
     _LongDoubleCodes,
     _CLongDoubleCodes,
-    _DT64Codes,
     _TD64Codes,
     _StrCodes,
     _BytesCodes,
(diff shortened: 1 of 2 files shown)


jorenham:
The new primer errors for `static_frame` are caused by mypy inferring `frozenset((DT64_MONTH, DT64_NS, ...))` as `frozenset[object]`, which is pretty strange if you ask me, because `frozenset[np.dtype[np.datetime64[Any]]]` would clearly be better in this case.
Luckily, `static_frame` can easily work around this by adding one or two annotations.

So these errors are are consequence of mypy's strange inference here rather than an issue on our side.

charris (approved):

charris:
Thanks Joren.

PR #30637, opened 2026-01-11
MAINT: avoid possible race condition by not touching `os.environ` on import

### LINKED ISSUE #21223: BUG: Calling `import numpy` at the same time in two different threads can lead to a race-condition
### Describe the issue:

Calling `import numpy` at the same time in two different threads can lead to a race-condition
This happens for example with Xpra when loading the encoder nvjpeg: 
```
2022-03-20 12:54:59,298  cannot load enc_nvjpeg (nvjpeg encoder)
  File "<pythondir>/lib/python3.9/site-packages/xpra/codecs/loader.py", line 52, in codec_import_check
    ic =  __import__(class_module, {}, {}, classnames)
  File "xpra/codecs/nvjpeg/encoder.pyx", line 8, in init xpra.codecs.nvjpeg.encoder
  File "<pythondir>/lib/python3.9/site-packages/numpy/__init__.py", line 150, in <module>
    from . import core
  File "<pythondir>/lib/python3.9/site-packages/numpy/core/__init__.py", line 51, in <module>
    del os.environ[envkey]
  File "<pythondir>/lib/python3.9/os.py", line 695, in __delitem__
    raise KeyError(key) from None
KeyError: 'OPENBLAS_MAIN_FREE'
```

The problem seems to come from numpy directly.

Here the environment variable OPENBLAS_MAIN_FREE is set:
https://github.com/numpy/numpy/blob/maintenance/1.21.x/numpy/core/__init__.py#L18
and short after that it is deleted
https://github.com/numpy/numpy/blob/maintenance/1.21.x/numpy/core/__init__.py#L51
But this deletion fails ... 

To me this looks like a threading issue in numpy. A lock would need to be set here.

```python
Here the environment variable OPENBLAS_MAIN_FREE is set:
https://github.com/numpy/numpy/blob/maintenance/1.21.x/numpy/core/__init__.py#L18
and short after that it is deleted
https://github.com/numpy/numpy/blob/maintenance/1.21.x/numpy/core/__init__.py#L51

If two threads call this funtion at the same time we get a race-condition.
The deletion fails ...
```

### Error message:

```shell
File "<pythondir>/lib/python3.9/site-packages/numpy/core/__init__.py", line 51, in <module>
    del os.environ[envkey]
  File "<pythondir>/lib/python3.9/os.py", line 695, in __delitem__
    raise KeyError(key) from None
KeyError: 'OPENBLAS_MAIN_FREE'
```
```


numpy 1.21.3
Python 3.9.6

### LINKED ISSUE #30627: BUG: os.environ manipluation during import can lead to race condition import failures
### Describe the issue:

this one is twisty, bear with me as I don't know where to _start_ but I do have it figured out:

https://github.com/numpy/numpy/blob/b91f4568c86beafb0337e33cdb982a16a2fca64e/numpy/_core/__init__.py#L84-L86

in particular does an unconditional `del os.envion[XXX]` which can fail if os.environ has been modified by another thread while this numpy import was happening.  leaving the numpy import in an incomplete state until importlib finishes unwinding as it makes an ImportError.  "hilarity" ensues if anything accesses numpy in this state (intended to be impossible for anything to get a reference to the incomplete numpy... but... rabbit hole: my rabbit hole sunk deeper and led to the creation of https://github.com/python/cpython/issues/143650 and a CPython bugfix - so it _can_ happen).

Workaround: import numpy before any thread that'd otherwise do such an import is spawned.

why is modifying os.environ tricky?  it's a global and in this case, test fixtures and tests of all forms like to do things such as mock or monkeypatch it either in its entirety or just saving and restoring its contents.  that, racing with an import that is expecting exclusive consistent access to os.environ, is non-deterministic.

if it hurts "don't do that" is valid here - but applies equally to all sides sadly. 😅 

### this is the real ask for this unusual issue:

**Can there be a better way** to plum those extension module load+init time only (right?) process environment `*BLAS_MAIN_FREE` settings into the `.multiarray` import presumably to be seen by native extension modules more directly?  I realize this is hard and fundamental perhaps to how BLAS native code "works".  But `numpy` could **instead call `os.putenv(key, "1")` and `os.unsetenv(key)`** in the cleanup to be more direct and not involve Python's `os.environ` proxy.  as I'm _hoping_ no Python level code needs to see these temporary environment variable changes in the higher level `os.environ` proxy dict right? (ever hopeful?)

Hard to reproduce as it's concurrency related, but this does appear to have been the root cause until CPython's own bugfix is in and widely deployed (slow).

Originally observed on Python 3.13.11 and numpy 2.3.5

1. we had a pytest plugin that was spawning a thread that transitively wound up triggering a numpy import.
2. when that plugin was enabled while running pytest on code that otherwise has nothing to do with numpy and thus does not import numpy itself... it showed up as flakiness during pytest post-test teardown... because our pytest conftest.py fixture did _another import_ of something during teardown that used numpy and had a dataclass `numpy.ndarray` type annotation but occasional the numpy module would be incomplete raise a mystery `AttributeError`.

thankfully this appears to only happen in wacky CI test scenarios.  code that actually _uses_ numpy outside of tests tends to import it up front and not leave that until later sometimes from two threads at once while another diddles with the environment.

rgommers:
We need to set the `OPENBLAS_MAIN_FREE` environment variable before the OpenBLAS initialization runs, however we don't need to touch `os.environ` for that. Deleting a key from `os.environ` can lead to a race condition when numpy is imported in parallel in multiple threads (may lead to a `KeyError`), while `os.unsetenv` always succeeds. See gh-30627 for context.

Also clean up `GOTOBLAS_MAIN_FREE`, that's a deprecated env var that has the same function as `OPENBLAS_MAIN_FREE`. There won't be any GotoBLAS or early OpenBLAS versions in use anymore that only have the GOTOBLAS flavor.

Closes gh-30627
Closes gh-21223

### DIFF
--- a/numpy/_core/__init__.py
+++ b/numpy/_core/__init__.py
@@ -13,9 +13,11 @@ from numpy.version import version as __version__
 # disables OpenBLAS affinity setting of the main thread that limits
 # python threads or processes to one core
 env_added = []
-for envkey in ['OPENBLAS_MAIN_FREE', 'GOTOBLAS_MAIN_FREE']:
+for envkey in ['OPENBLAS_MAIN_FREE']:
     if envkey not in os.environ:
-        os.environ[envkey] = '1'
+        # Note: using `putenv` (and `unsetenv` further down) instead of updating
+        # `os.environ` on purpose to avoid a race condition, see gh-30627.
+        os.putenv(envkey, '1')
         env_added.append(envkey)
 
 try:
@@ -83,7 +85,7 @@ Original error was: {exc}
     raise ImportError(msg) from exc
 finally:
     for envkey in env_added:
-        del os.environ[envkey]
+        os.unsetenv(envkey)
 del envkey
 del env_added
 del os


mattip:
Thanks @rgommers

PR #30638, opened 2026-01-12
DEV: Add Pixi workspace configuration

Aniketsy:
#30387 

This PR adds a `pixi.toml `to provide a Pixi-based developer environment for NumPy.

This work proceeds from @rgommers  earlier Pixi setup and extends to cover NumPy’s current development workflows.
- Still left with documentation.

Please let me know if there’s anything that should be adjusted or improved.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -1,0 +1,178 @@
+[workspace]
+name = "numpy"
+version = "2.5.0.dev0"
+description = "Fundamental package for array computing in Python"
+authors = ["NumPy Developers <numpy-dev@python.org>"]
+channels = ["https://prefix.dev/conda-forge"]
+platforms = ["osx-arm64", "linux-64", "win-64"]
+
+### Environments ###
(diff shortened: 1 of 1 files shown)


Aniketsy:
Sorry, I just noticed that we already have a `pixi-package`. I added this as a separate file, should I move these changes into pixi-package instead?

rgommers:
Thanks @Aniketsy, however this PR is too early (see https://github.com/numpy/numpy/issues/30387#issuecomment-3737952971). It seems like it's in the right direction, but there's no lock file nor usage in CI, (both of which are needed for a PR), so it'd be better to iterate on this in your own fork to make CI pass and then use it locally for a while to ensure things actually work as expected.

Aniketsy:
Thanks, that’s clear. I’ll add a lock file and integrate it with CI, then continue iterating on this in my fork and use it locally.

PR #30639, opened 2026-01-12
DOC: Update legacy pareto documentation to match Pareto II

Aniketsy:
closes #30250

### DIFF
--- a/numpy/random/mtrand.pyx
+++ b/numpy/random/mtrand.pyx
@@ -2421,11 +2421,14 @@ cdef class RandomState:
 
         Notes
         -----
-        The probability density for the Pareto distribution is
+        The probability density for the Pareto II distribution is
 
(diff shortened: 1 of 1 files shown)


jorenham (commented):
Looks good.

Random thought: It might help to mention that it's also often called "Lomax" (technically there are some differences, but those don't really matter here)

Aniketsy:
Sure, I’ll update this.

@jorenham Thoughts on this, should we add a separate line like
`The Pareto II distribution is also often called the "Lomax" distribution,` or is it better to mention it inline, e.g.
The probability density for the Pareto II or `(Lomax)` distribution is …?

jorenham:
At the start of the docstring it's mentioned in-line, so that should also be enough here AFAIK.

mattip:
Thanks @Aniketsy

PR #30640, opened 2026-01-12
use clang instead of gcc

wojiushixiaobai:
This allows a speed-up of all QEMU targets by using a statically built clang that runs with the native architecture of the runner rather than calling gcc through QEMU.

Requirement:
- "scipy-openblas32>=0.3.30.359.2"

This PR is mostly used as a PoC for this trick and see wether it would see adoption, what are the limitations and how it could be improved.

Clang used in this PR is built & published in the following repo: https://github.com/mayeut/static-clang-images

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -37,6 +37,10 @@ jobs:
         buildplat:
           - [ubuntu-22.04, manylinux_x86_64, ""]
           - [ubuntu-22.04, musllinux_x86_64, ""]
+          - [ubuntu-22.04, manylinux_ppc64le, ""]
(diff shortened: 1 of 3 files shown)


rgommers:
Thanks for sharing @wojiushixiaobai. 

> This PR is mostly used as a PoC for this trick and see wether it would see adoption, what are the limitations and how it could be improved.

We're happy to give feedback. Regarding adoption, I don't think it's likely we want to use it to ship wheels for new platforms without native runners. We do use QEMU in several jobs, which may be a better target if there's benefit there: https://github.com/numpy/numpy/blob/main/.github/workflows/linux_qemu.yml

PR #30642, opened 2026-01-12
DOC: clarify return behavior of in-place ndarray operations

### LINKED ISSUE #28780: DOC: Incorrect description of in-place operations

saksham-1020:
Fixes #28780

Clarifies that in-place arithmetic special methods modify the array
and return self, avoiding confusion with augmented assignment syntax.

### DIFF
--- a/doc/source/reference/arrays.ndarray.rst
+++ b/doc/source/reference/arrays.ndarray.rst
@@ -535,6 +535,12 @@ Arithmetic, in-place:
    ndarray.__ior__
    ndarray.__ixor__
 
+In-place arithmetic methods modify the array itself and return the
+modified array. Although the corresponding operators are written using
+augmented assignment syntax (for example, ``a += b``), the underlying
+special methods return ``self`` after performing the operation.
+
+
 .. warning::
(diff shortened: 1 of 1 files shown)


jorenham:
#28780 is about the specific augmented operator method documentation, but you have modified that of the `ndarray` class.
And even if these docs were to be moved there, it still would not address the issue, because this simply restates the commonly known fact that these so-called "in-place operators" operate in-place (kinda hard to miss tbh).
So that's why I'm going to close this.

saksham-1020:
Thanks for the clarification!  
That makes sense — I misunderstood the scope of the issue.  
I’ll close this out and look for a more targeted contribution.  
Appreciate the feedback.

PR #30644, opened 2026-01-12
DEP: Deprecate numpy.fix in favor of numpy.trunc

skyvanguard:
## Summary

Deprecate `numpy.fix` as it provides identical functionality to `numpy.trunc` but is slower and is not part of the Array API standard.

### Motivation

As noted in #30096:
- `numpy.fix` and `numpy.trunc` produce identical results
- `numpy.trunc` is ~2.5x faster than `numpy.fix`
- The Array API standard defines `trunc` but not `fix`
- The name `trunc` is clearer and more standard

### Changes

- Add `DeprecationWarning` to `numpy.fix()`
- Update docstring with `.. deprecated::` directive
- Add tests to verify deprecation warning is emitted
- Update existing tests to suppress deprecation warnings during functional tests
- Add release note about deprecation

### Performance comparison

```python
In [1]: arr = np.random.random(1_000_000)*1000

In [2]: %timeit np.trunc(arr)
426 μs ± 15.1 μs per loop

In [3]: %timeit np.fix(arr)
1.06 ms ± 3.34 μs per loop

In [4]: np.all(np.trunc(arr) == np.fix(arr))
Out[4]: np.True_
```

Closes #30096


jorenham:
Could you also remove the `category=PendingDeprecationWarning` from the stubs `@deprecated`?

https://github.com/numpy/numpy/blob/89776e23e188438637b636651ff3ce0df4bd2097/numpy/lib/_ufunclike_impl.pyi#L14-L25

numpy/lib/tests/test_ufunclike.py
jorenham, line 1: `@pytest.mark.filterwarnings` is the preferred way of suppressing warnings in the tests

numpy/lib/_ufunclike_impl.py
```diff
@@ -56,6 +62,13 @@ def fix(x, out=None):
     array([ 2.,  2., -2., -2.])
 
     """
+    # Deprecated in NumPy 2.3, 2025-01-12
```
jorenham: you're a year behind ;)

numpy/lib/_ufunclike_impl.py
```diff
@@ -18,6 +20,10 @@ def fix(x, out=None):
     """
     Round to nearest integer towards zero.
 
+    .. deprecated:: 2.3
```
jorenham:
```suggestion
    .. deprecated:: 2.5
```

jorenham (changes requested):

skyvanguard:
Done! Removed `category=PendingDeprecationWarning` from the stubs and updated the deprecation message to match the runtime warning.

skyvanguard:
Fixed! Updated to NumPy 2.5 and 2026. Thanks for catching that 😅

skyvanguard:
Updated tests to use `@pytest.mark.filterwarnings` instead of `warnings.catch_warnings()`. Thanks for the guidance!

numpy/typing/tests/data/pass/ufunclike.py
```diff
@@ -1,9 +1,13 @@
 from __future__ import annotations
 
+import warnings
 from typing import Any
 
 import numpy as np
 
+# Suppress DeprecationWarning for np.fix during runtime testing
+warnings.filterwarnings("ignore", "numpy.fix is deprecated", DeprecationWarning)
```
jorenham: please use `@pytest.mark.filterwarnings` here too

jorenham (approved):

jorenham:
Thanks @skyvanguard

PR #30647, opened 2026-01-13
BUG: unrelated error raised when a dtype's `__setstate__` is called with an invalid state tuple size

### LINKED ISSUE #30476: BUG: Uninitialized variables in arraydescr_setstate when unpickling malformed dtype
### Describe the issue:

Our downstream static analysis scan identified uninitialized variable usage in arraydescr_setstate() in [descriptor.c](https://github.com/numpy/numpy/blob/ccaf198024f42148d334ea1c28e736a9fd3bee81/numpy/_core/src/multiarray/descriptor.c#L2867).

When unpickling a dtype with a malformed tuple size, the [default case](https://github.com/numpy/numpy/blob/ccaf198024f42148d334ea1c28e736a9fd3bee81/numpy/_core/src/multiarray/descriptor.c#L2934) in the switch statement does not return an error, allowing code to continue with uninitialized endian_obj, fields, and subarray variables.

Exact coverity output:

<details>

<summary> snippet </summary>
```

Defect type: UNINIT
1. numpy-2.3.4/numpy/_core/src/multiarray/descriptor.c:2864:5: var_decl: Declaring variable "fields" without initializer.
13. numpy-2.3.4/numpy/_core/src/multiarray/descriptor.c:2949:9: uninit_use: Using uninitialized value "fields".
 2947|   
 2948|       if (version == 1 || version == 0) {
 2949|->         if (fields != Py_None) {
 2950|               PyObject *key, *list;
 2951|               key = PyLong_FromLong(-1);

Defect type: UNINIT
1. numpy-2.3.4/numpy/_core/src/multiarray/descriptor.c:2863:5: var_decl: Declaring variable "endian_obj" without initializer.
16. numpy-2.3.4/numpy/_core/src/multiarray/descriptor.c:2971:5: uninit_use_in_call: Using uninitialized value "(PyObject *)endian_obj" when calling "_Py_TYPE".
17. /usr/include/python3.14/object.h:270:9: read_value: Reading value "ob".
  2969|   
  2970|       /* Parse endian */
  2971|->     if (PyUnicode_Check(endian_obj) || PyBytes_Check(endian_obj)) {
  2972|           PyObject *tmp = NULL;
  2973|           char *str;

Defect type: UNINIT
1. numpy-2.3.4/numpy/_core/src/multiarray/descriptor.c:2864:5: var_decl: Declaring variable "fields" without initializer.
20. numpy-2.3.4/numpy/_core/src/multiarray/descriptor.c:3003:5: uninit_use: Using uninitialized value "fields".
  3001|       }
  3002|   
  3003|->     if ((fields == Py_None && names != Py_None) ||
  3004|           (names == Py_None && fields != Py_None)) {
  3005|           PyErr_Format(PyExc_ValueError,

Defect type: UNINIT
1. numpy-2.3.4/numpy/_core/src/multiarray/descriptor.c:2864:5: var_decl: Declaring variable "subarray" without initializer.
31. numpy-2.3.4/numpy/_core/src/multiarray/descriptor.c:3033:5: uninit_use: Using uninitialized value "subarray".
  3031|       self->subarray = NULL;
  3032|   
  3033|->     if (subarray != Py_None) {
  3034|           PyObject *subarray_shape;
  3035|   

```
</details>

```python
A malformed pickle with a tuple size outside 5-9, where the first element is a valid version number (0-4)
```

### Error message:

```shell

```

Python 3.14, Numpy 2.3.4

### Runtime Environment:

_No response_

### Context for the issue:

_No response_

ikrommyd:
Closes https://github.com/numpy/numpy/issues/30476

My understanding here is that the switch statement checks only specific sizes of the input state tuple `PyTuple_GET_SIZE(PyTuple_GET_ITEM(args,0))`. If none of those sizes match, then it does this:
https://github.com/numpy/numpy/blob/60057c05a349a6d2281ffaf7a7627a1ae8732c63/numpy/_core/src/multiarray/descriptor.c#L2919-L2927
So if it's size is greater than 5, it tries to convert the tuple into a long with `PyLong_AsLong(PyTuple_GET_ITEM(args, 0))` which can't work right? `PyLong_AsLong` returns -1 on failure (the else block sets the version to -1 too) so we will get a `ValueError: can't handle version -1 of numpy.dtype pickle` from right below:
https://github.com/numpy/numpy/blob/60057c05a349a6d2281ffaf7a7627a1ae8732c63/numpy/_core/src/multiarray/descriptor.c#L2933-L2937

That appears like an unrelated error though and it looks to me like a proper error should be raised if the state tuple is not one of the acceptable sizes.

Before:
```py
>>> import numpy as np
>>> dt = np.dtype([('a', 'i4'), ('b', 'f8')])
>>> dt.__setstate__((6,7,8))
    dt.__setstate__((6,7,8))
    ~~~~~~~~~~~~~~~^^^^^^^^^
ValueError: can't handle version -1 of numpy.dtype pickle
```
After:
```py
>>> import numpy as np
>>> dt = np.dtype([('a', 'i4'), ('b', 'f8')])
>>> dt.__setstate__((6,7,8))
  File "<python-input-4>", line 1, in <module>
    dt.__setstate__((6,7,8))
    ~~~~~~~~~~~~~~~^^^^^^^^^
ValueError: invalid input tuple size 3 in numpy.dtype setstate
```

### DIFF
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -844,7 +844,7 @@ _try_convert_from_inherit_tuple(PyArray_Descr *type, PyObject *newobj)
         return (PyArray_Descr *)Py_NotImplemented;
     }
     if (!PyDataType_ISLEGACY(type) || !PyDataType_ISLEGACY(conv)) {
-        /* 
+        /*
          * This specification should probably be never supported, but
          * certainly not for new-style DTypes.
          */
@@ -1951,7 +1951,7 @@ NPY_NO_EXPORT PyArray_Descr *
 PyArray_DescrNew(PyArray_Descr *base_descr)
 {
     if (!PyDataType_ISLEGACY(base_descr)) {
-        /* 
+        /*
          * The main use of this function is mutating strings, so probably
          * disallowing this is fine in practice.
          */
@@ -2917,13 +2917,10 @@ arraydescr_setstate(_PyArray_LegacyDescr *self, PyObject *args)
         }
         break;
     default:
-        /* raise an error */
-        if (PyTuple_GET_SIZE(PyTuple_GET_ITEM(args,0)) > 5) {
-            version = PyLong_AsLong(PyTuple_GET_ITEM(args, 0));
-        }
-        else {
-            version = -1;
-        }
+        PyErr_Format(PyExc_ValueError,
+                     "invalid tuple size %zd in numpy.dtype setstate",
+                     PyTuple_GET_SIZE(PyTuple_GET_ITEM(args, 0)));
+        return NULL;
     }
 
     /*
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -2917,9 +2917,9 @@ arraydescr_setstate(_PyArray_LegacyDescr *self, PyObject *args)
         }
         break;
     default:
-        PyErr_Format(PyExc_ValueError,
-                     "invalid input tuple size %zd in numpy.dtype setstate",
-                     PyTuple_GET_SIZE(PyTuple_GET_ITEM(args, 0)));
+        PyErr_SetString(PyExc_ValueError,
+                        "Invalid state while unpickling. Is the pickle corrupted "
+                        "or created with a newer NumPy version?");
         return NULL;
     }
 
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -2918,7 +2918,7 @@ arraydescr_setstate(_PyArray_LegacyDescr *self, PyObject *args)
         break;
     default:
         PyErr_Format(PyExc_ValueError,
-                     "invalid tuple size %zd in numpy.dtype setstate",
+                     "invalid input tuple size %zd in numpy.dtype setstate",
                      PyTuple_GET_SIZE(PyTuple_GET_ITEM(args, 0)));
         return NULL;
     }
--- a/numpy/_core/tests/test_dtype.py
+++ b/numpy/_core/tests/test_dtype.py
@@ -1341,6 +1341,29 @@ class TestPickling:
             assert roundtrip_dt == dt
             assert hash(roundtrip_dt) == pre_pickle_hash
 
+    @pytest.mark.parametrize('dt', [
+        np.dtype([('a', 'i4'), ('b', 'f8')]),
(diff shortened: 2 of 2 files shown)


ikrommyd:
Please let me know if I did not understand something properly from that code path or whether you want this addressed in another way 😃

ngoldbaum:
Git blame says the code has been written this way for 17 years: https://github.com/numpy/numpy/commit/7c0908e9bfc084ef730d5b8845b8528187ca6e6a to support pickle format 1.3, which is very old now.

In practice this probably can't be hit without a corrupt pickle file or something like that, which explains why it took a static analysis tool like coverity to catch that the PyLong_AsLong call is a silly thing to do in the generic case.

I think this seems correct to me and I sort of doubt real-world pickle files ever hit this case, so I think it's probably low-risk to merge. Maybe really really old pickle files would be broken somehow?

That said, I'll wait for another maintainer to weigh in

seberg:
Right, I think you need a corrupt pickle (at which point pretty much all is lost either way).

The slight downside of this change is that if the tuple was ever made larger, then we would get a less good error (the previous version would try print "can't support version 5").
Since I doubt it'll happen... I might be tempted to just prepare for the eventuality anyway and  make the error in the direction of: invalid state while unpickling, is the pickle corrupted or created with a newer NumPy version?

I don't actually see us making the tuple larger either way, since I would create a new unpickling function rather than going via `__setstate__`...

ikrommyd:
I switched the error message to what you proposed. The previous error would always be `can't handle version -1 of numpy.dtype pickle` no matter the size because `PyLong_AsLong(PyTuple_GET_ITEM(args, 0)` would always fail. This reads to me as if the state was fine and it's just the version that's wrong. "invalid state" sounds better to me at least because it tells you that something's wrong with the state entirely.

seberg:
No, it wouldn't fail in any sane reason for hitting it: We increased the length of the tuple and of course the first entry would still be the version in that case (just a higher version).

seberg:
Ah, sorry... yes, you are right it would always fail cryptically because it forgot one level of unpacking. The intention was to report "this is a newer version" of course.

EDIT: On the up-side. That mistake actually means that the code is 100% safe, this effectively just improves the error message.

ikrommyd:
Yeah exactly. I assume coverity that was used for the original report does not know that PyLong_AsLong returns -1 on failure and that the code would just stop a few lines later (unintentionally).

seberg:
Long arch is taking forever again (but IIRC there is an open PR), thanks @ikrommyd.

PR #30656, opened 2026-01-15
DOC:clarify fromnumeric wrapper behaviour

Nandika32:
This PR adds a short explanation to the module-level docstring in
`numpy/_core/fromnumeric.py` to clarify how wrapper functions are exposed
and why warnings may be emitted when accessing them via `np.<function>`.
No functional changes are made.


jorenham:
The PR description does not match the changes. Furthermore, `numpy.core` is a deprecated module (since NumPy 2.0) and should not be used, so I'm going to close this.

PR #30660, opened 2026-01-16
ENH: Generalize split function to split matrix into sub-matrices.

### LINKED ISSUE #15380: Feature request: generalizing split
Is there a way the split function could be generalized? I think it would be useful to split matrices into a grid of submatrices, for example.

I'm currently trying to implement such a function in python using numpy, and I can't seem to find an idiomatic way to do it. Could someone point me to some pointers on how to handle different axes sizes? Has someone tried to generalize strides? If someone could link some resources, which I'm having a hard time finding, that would be appreciated!

KoningR:
Closes numpy#15380

This PR supersedes numpy#15810, which was closed due to age.

### DIFF
--- a/doc/release/upcoming_changes/30660.new_feature.rst
+++ b/doc/release/upcoming_changes/30660.new_feature.rst
@@ -1,0 +1,20 @@
+`numpy.split` and `numpy.array_split` now accept a tuple for ``axis``
+---------------------------------------------------------------------
+Both `numpy.split` and `numpy.array_split` now accept a tuple
+for the ``axis`` parameter. This tuple denotes the axes over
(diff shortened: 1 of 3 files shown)


jorenham:
Thanks for this. 
But note that, given the size of this PR, it might be a while until one of us maintainers finds the time to review this.

mattip:
The original issue does not provide a clear motivation for this. It seems this collapses N calls to `split` into one call, but other than convenience is there any memory or speed savings? The issue has a comment 
> I think we would be happy to accept a pull request which generalizes the current split functionality. (I assume that the code will not be super complex)

I am not sure thinking about the actual output is really all that simple.

PR #30662, opened 2026-01-16
BUG: fix time-of-check vs time-of-use bug in npy_hashtable.c

### LINKED ISSUE #30658: BUG: Possible bug in `npy_hashtable.c`
### Bug in `PyArrayIdentityHash_GetItem()`

I think there's a bug in the new `npy_hashtable.c`, although I'm not 100% confident.

1) If `find_item()` doesn't find a matching entry, it returns the address of the empty entry where the key would be inserted.
2) `PyArrayIdentityHash_GetItem()` then calls `FT_ATOMIC_LOAD_PTR_ACQUIRE(tb_item[0])` to load the value

Normally, if the key is missing, then (2) will return `NULL`. However, if there's a concurrent write to the table, then that entry may be filled in with a *different* key and value, and the `FT_ATOMIC_LOAD_PTR_ACQUIRE(tb_item[0])` could return the incorrect value.

https://github.com/numpy/numpy/blob/508943d1070d8f4e0712b98db2379d49304e4e38/numpy/_core/src/common/npy_hashtable.c#L306-L311

### Resizing behavior

Not exactly a bug, but I don't understand why we have logic for shrinking the table. Aren't entries only inserted, never deleted? Shrinking the table also would break the guarantees about memory usage for the `Py_GIL_DISABLED` build. Currently, it uses at most 2x the memory from saving old buckets due to doubling in size when resizing. Shrinking and growing would break this.

https://github.com/numpy/numpy/blob/508943d1070d8f4e0712b98db2379d49304e4e38/numpy/_core/src/common/npy_hashtable.c#L202-L217

cc @kumaraditya303 @ngoldbaum 

Possibly related to the crash seen in https://github.com/numpy/numpy/pull/30653#issuecomment-3752431364

main branch

### Runtime Environment:

_No response_

See https://github.com/numpy/numpy/pull/30653#issuecomment-3752431364 (possibly related)

kumaraditya303:
Fixes #30658

### DIFF
--- a/numpy/_core/src/common/npy_hashtable.c
+++ b/numpy/_core/src/common/npy_hashtable.c
@@ -105,7 +105,7 @@ identity_list_hash(PyObject *const *v, int len)
 
 static inline PyObject **
 find_item_buckets(struct buckets *buckets, int key_len, PyObject *const *key,
-                PyObject **pvalue)
+                  PyObject **pvalue)
 {
     Py_hash_t hash = identity_list_hash(key, key_len);
     npy_uintp perturb = (npy_uintp)hash;
--- a/numpy/_core/src/common/npy_hashtable.c
+++ b/numpy/_core/src/common/npy_hashtable.c
@@ -104,7 +104,8 @@ identity_list_hash(PyObject *const *v, int len)
 
 
 static inline PyObject **
-find_item_buckets(struct buckets *buckets, int key_len, PyObject *const *key)
+find_item_buckets(struct buckets *buckets, int key_len, PyObject *const *key,
+                PyObject **pvalue)
 {
     Py_hash_t hash = identity_list_hash(key, key_len);
     npy_uintp perturb = (npy_uintp)hash;
@@ -113,8 +114,11 @@ find_item_buckets(struct buckets *buckets, int key_len, PyObject *const *key)
 
     while (1) {
         PyObject **item = &(buckets->array[bucket * (key_len + 1)]);
-        PyObject *value = FT_ATOMIC_LOAD_PTR_ACQUIRE(item[0]);
-        if (value == NULL) {
+        PyObject *val = FT_ATOMIC_LOAD_PTR_ACQUIRE(item[0]);
+        if (pvalue != NULL) {
+            *pvalue = val;
+        }
+        if (val == NULL) {
             /* The item is not in the cache; return the empty bucket */
             return item;
         }
@@ -130,10 +134,10 @@ find_item_buckets(struct buckets *buckets, int key_len, PyObject *const *key)
 
 
 static inline PyObject **
-find_item(PyArrayIdentityHash const *tb, PyObject *const *key)
+find_item(PyArrayIdentityHash const *tb, PyObject *const *key, PyObject **pvalue)
 {
     struct buckets *buckets = FT_ATOMIC_LOAD_PTR_ACQUIRE(tb->buckets);
-    return find_item_buckets(buckets, tb->key_len, key);
+    return find_item_buckets(buckets, tb->key_len, key, pvalue);
 }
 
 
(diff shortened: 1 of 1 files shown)


numpy/_core/src/common/npy_hashtable.c
```diff
@@ -196,26 +200,16 @@ _resize_if_necessary(PyArrayIdentityHash *tb)
...
-        while ((old_buckets->nelem + 8) * 2 < new_size / 2) {
-            /*
-             * Should possibly be improved.  However, we assume that we
-             * almost never shrink.  Still if we do, do not shrink as much
-             * as possible to avoid growing right away.
-             */
-            new_size /= 2;
-        }
-        assert(new_size >= 4);
-    }
```
ngoldbaum, line 217: We removed the unused ability to replace items and this is leftover from that, so it's totally safe to delete

numpy/_core/src/common/npy_hashtable.c
```diff
@@ -104,7 +104,8 @@ identity_list_hash(PyObject *const *v, int len)
 
 
 static inline PyObject **
-find_item_buckets(struct buckets *buckets, int key_len, PyObject *const *key)
+find_item_buckets(struct buckets *buckets, int key_len, PyObject *const *key,
+                PyObject **pvalue)
```
colesbury:
```suggestion
find_item_buckets(struct buckets *buckets, int key_len, PyObject *const *key,
                  PyObject **pvalue)
```

numpy/_core/src/common/npy_hashtable.c
```diff
@@ -306,6 +300,7 @@ PyArrayIdentityHash_SetItemDefault(PyArrayIdentityHash *tb,
 NPY_NO_EXPORT PyObject *
 PyArrayIdentityHash_GetItem(PyArrayIdentityHash *tb, PyObject *const *key)
 {
-    PyObject **tb_item = find_item(tb, key);
-    return FT_ATOMIC_LOAD_PTR_ACQUIRE(tb_item[0]);
```
ngoldbaum, line 310: Currently, we're doing *two* atomic loads of `tb_item[0]`. One inside of `find_item` and one here. That's incorrect because there's no way to guarantee that e.g. the table might have been resized in another thread, leaving `tb_item[0]` here as an invalid pointer to freed memory.

This PR fixes that by passing the object we want to return into `find_item` and returning the value loaded from the one and only atomic load, which is guaranteed to be loaded from a valid pointer due to the atomic stores in the setter implementation.

Is that right? I'd appreciate you checking my understanding for me.

ngoldbaum (commented):
I edited the description to link the issue and so it gets closed when this is merged.

numpy/_core/src/common/npy_hashtable.c
```diff
@@ -196,26 +200,16 @@ _resize_if_necessary(PyArrayIdentityHash *tb)
...
-             * almost never shrink.  Still if we do, do not shrink as much
-             * as possible to avoid growing right away.
-             */
-            new_size /= 2;
-        }
-        assert(new_size >= 4);
-    }
-    if (new_size == prev_size) {
+
+    if (new_size == 0) {
```
colesbury: The condition here on `new_size == 0` is a bit confusing to me. I think it would be easier to understand as:

```
if ((old_buckets->nelem + 1) * 2 <= old_buckets->size) {
    /* No resize necessary if load factor is not more than 0.5 */
    return 0;
}

/* Double in size */
npy_intp new_size = old_buckets->size * 2;
...
```

colesbury (commented):
A few minor comments below, but the logic LGTM

numpy/_core/src/common/npy_hashtable.c
kumaraditya303: Yes two atomic loads was incorrect because another thread can set an item at tb_item  between the two loads which would be different than what we checked against our key. 

> the table might have been resized in another thread, leaving tb_item[0] here as an invalid pointer to freed memory.

No, the old table is not freed immediately in free-threading so that does not happen.

numpy/_core/src/common/npy_hashtable.c
kumaraditya303: yes, it was leftover

numpy/_core/src/common/npy_hashtable.c
kumaraditya303: done

numpy/_core/src/common/npy_hashtable.c
ngoldbaum: > No, the old table is not freed immediately in free-threading so that does not happen.

Thanks for clarifying, makes sense.

ngoldbaum:
Thanks for the follow-through here Kumar! I'll keep my eyes peeled for more crashes but hopefully this fixes it.

Issue #30648, opened 2026-01-14
BUG: data races in `_buffer_get_info` when multiple threads try to access buffer of ndarray


kumaraditya303:
### Describe the issue:

Run the reproducer using a TSAN instrumented build of CPython and numpy.

### Reproduce the code example:

```python
import numpy as np
from threading import Thread, Barrier
import inspect
import random
arr = np.array([1, 2, 3, 4, 5])

flags = [inspect.BufferFlags.STRIDED, inspect.BufferFlags.READ]

barrier = Barrier(4)
def func():
    barrier.wait()
    arr.__buffer__(random.choice(flags))

threads = [Thread(target=func) for _ in range(4)]
for t in threads:
    t.start()
for t in threads:
    t.join()
```

### Error message:

```shell
WARNING: ThreadSanitizer: data race (pid=6155)
  Read of size 8 at 0x000128060100 by thread T1:
    #0 _buffer_get_info <null> (_multiarray_umath.cpython-314t-darwin.so:arm64+0x37c1d8)
    #1 array_getbuffer <null> (_multiarray_umath.cpython-314t-darwin.so:arm64+0x37c5b8)
    #2 _PyMemoryView_FromBufferProc memoryobject.c:838 (libpython3.14t.dylib:arm64+0x122c1c)
    #3 buffer typeobject.c:9688 (libpython3.14t.dylib:arm64+0x1a48b0)
    #4 wrapperdescr_call descrobject.c:570 (libpython3.14t.dylib:arm64+0xa17ac)
    #5 _PyObject_MakeTpCall call.c:242 (libpython3.14t.dylib:arm64+0x8a3ac)
    #6 PyObject_Vectorcall call.c:327 (libpython3.14t.dylib:arm64+0x8af9c)
    #7 _PyEval_EvalFrameDefault generated_cases.c.h:1619 (libpython3.14t.dylib:arm64+0x27cc00)
    #8 _PyEval_Vector ceval.c:2083 (libpython3.14t.dylib:arm64+0x278928)
    #9 _PyFunction_Vectorcall call.c (libpython3.14t.dylib:arm64+0x8b54c)
    #10 method_vectorcall classobject.c:73 (libpython3.14t.dylib:arm64+0x8f9c0)
    #11 context_run context.c:728 (libpython3.14t.dylib:arm64+0x2c3d18)
[...]

PR #30666, opened 2026-01-17
BUG/TST: Fix TSAN race condition in _buffer_get_info and add TSAN test

sagerkudrick:
Adds a thread-safety test for the buffer protocol to help catch potential race conditions

- Motivation:
   - This is related to Issue #30648, which discusses thread safety concerns in the buffer protocol

- What this PR does:
   - Adds a small test for buffer protocol operations across multiple threads
   - The goal is to trigger TSAN / sanitizer CI to see if any race conditions are flagged 

- Notes: 
   - This is a CI/test PR

Thanks!

### DIFF
--- a/numpy/_core/src/multiarray/buffer.c
+++ b/numpy/_core/src/multiarray/buffer.c
@@ -692,6 +692,16 @@ _buffer_get_info(void **buffer_info_cache_ptr, PyObject *obj, int flags)
 {
     _buffer_info_t *info = NULL;
     _buffer_info_t *stored_info;  /* First currently stored buffer info */
+    void *cache_snapshot;
+
+#ifdef Py_GIL_DISABLED 
+    /* Atomic read for free-threaded Python to avoid race conditions */
+    cache_snapshot = _Py_atomic_load_ptr(buffer_info_cache_ptr);
+#else
+    /* GIL protects us in traditional Python builds */
+    cache_snapshot = *buffer_info_cache_ptr;
+#endif 
+
 
     if (_buffer_info_untag(*buffer_info_cache_ptr, &stored_info, obj) < 0) {
         return NULL;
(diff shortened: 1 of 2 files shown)


sagerkudrick:
``.github/workflows/compiler_sanitizers.yml`` now fails with this test, with this logged:

``==== 870 passed, 14 skipped, 38 deselected, 2 xfailed in 111.45s (0:01:51) =====
ThreadSanitizer: reported 8 warnings
Error: Process completed with exit code 66.``

Failed CLI:

Screenshot 2026-01-17 020923

TSAN warnings (which fail the suite):

image

TSAN's race conditions:

image

sagerkudrick:
Regardless of pytest/strict=False, TSAN would exit with code 66 at the process level and cause the suite to fail. 

I moved the TSAN test into a subprocess specifically so it can't override pytest and exit with 66, but there's no race condition in this scenario, likely as it doesn't inherent TSAN.

Edit: #[7862d88](https://github.com/numpy/numpy/pull/30666/commits/7862d8814a8870cfebb01cbe58e5b5e5cb5b2210)

1. I've added a test for the TSAN race condition
2. I've updated _buffer_info_free in buffer.c to explicitly handle free-threaded access. When GIL is enabled (normal usage/instalation), it should operate as normal, as we explicitly check for GIL being disabled to do anything new to avoid unnecessary overhead. When GIL is disabled for free-threading, that's when we handle the race condition.
3. Originally, the CI test failed prior to the _buffer_info_free update because of the race condition. It now passes successfully with this update

sagerkudrick:
Modified it so it still tests non-TSAN as a sanity check, ensuring XFAIL isn't a false negative. It'll successfully complete on non-TSAN systems, and XFAIL on TSAN systems.

image

seberg:
@kumaraditya303 do you think this more maximal approach may be worthwhile?  I want to still merge your PR first, because I think it can be backported, while I don't want to backport this type of atomics (just so much easier to get wrong).

kumaraditya303:
> do you think this more maximal approach may be worthwhile? I want to still merge your PR first, because I think it can be backported, while I don't want to backport this type of atomics (just so much easier to get wrong).

This is not correct, if we want to use atomics then any old buffer info tag objects needs to be freed using something like QSBR because other threads might be using them while it can get concurrently freed.

seberg:
Thanks for the note. I think I'll close the PR for now, we can discuss/reconsider a new approach though.

Issue #24530, opened 2023-08-24
BUG: ufuncs created with np.frompyfunc do not recognize identity in reduce in numpy 1.25.X


jakevdp:
### Describe the issue:

Starting in numpy 1.25.0, it appears that the `ufunc.reduce` method does not recognize the `identity` argument for ufuncs created with `np.frompyfunc`

### Reproduce the code example:

```python
import numpy as np
import operator

print(np.__version__)

x = np.arange(10)
mask = x % 2 == 0
print(np.add.reduce(x, where=mask))

my_add = np.frompyfunc(operator.add, nin=2, nout=1, identity=0)
print(my_add.reduce(x, where=mask))
```

### Error message:

```shell
Result in numpy 1.24.4:

1.24.4
20
20
```
Result in numpy 1.25.0:
```
1.25.0
20
```

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-f018ed213832> in <cell line: 9>()
      7 
      8 my_add = np.frompyfunc(np.add, nin=2, nout=1, identity=0)
----> 9 print(my_add.reduce(x, where=mask))

ValueError: reduction operation 'add (vectorized)' does not have an identity, so to use a where mask one has to specify 'initial'
```
```

### Runtime information:

```
>>> import sys, numpy; print(numpy.__version__); print(sys.version)
1.25.0
3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0]
>>> print(numpy.show_runtime())
[{'numpy_version': '1.25.0',
  'python': '3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0]',
  'uname': uname_result(system='Linux', node='e87715ee78c7', release='5.15.109+', version='#1 SMP Fri Jun 9 10:57:30 UTC 2023', machine='x86_64')},
 {'simd_extensions': {'baseline': ['SSE', 'SSE2', 'SSE3'],
                      'found': ['SSSE3',
                                'SSE41',
                                'POPCNT',
                                'SSE42',
                                'AVX',
                                'F16C',
                                'FMA3',
                                'AVX2'],
                      'not_found': ['AVX512F',
                                    'AVX512CD',
                                    'AVX512_KNL',
                                    'AVX512_KNM',
                                    'AVX512_SKX',
                                    'AVX512_CLX',
                                    'AVX512_CNL',
                                    'AVX512_ICL']}},
 {'architecture': 'Haswell',
[...]

PR #30670, opened 2026-01-19
ENH: infer nin/nout for frompyfunc

danielzgtg:
This eliminates redundantly typing `nin`/`nout` especially when obvious with `lambda`.

# Before

```console
home@daniel-desktop3:~$ python3
Python 3.13.7 (main, Nov 24 2025, 20:51:28) [GCC 15.2.0] on linux
>>> import numpy as np
>>> np.frompyfunc(lambda x: x, 1, 1).nin
1
>>> np.frompyfunc(lambda x: x, 1, 1).nout
1
>>> np.frompyfunc(lambda x: x + y, 2, 1).nin
2
>>> np.frompyfunc(lambda x: x + y, 2, 1).nout
1
>>> np.frompyfunc(lambda x: x)
    np.frompyfunc(lambda x: x)
    ~~~~~~~~~~~~~^^^^^^^^^^^^^
TypeError: frompyfunc() missing required argument 'nin' (pos 2)
>>> np.frompyfunc(lambda x, y: x + y)
    np.frompyfunc(lambda x, y: x + y)
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
TypeError: frompyfunc() missing required argument 'nin' (pos 2)
>>> 
```

# After

```console
(venv) home@daniel-desktop3:~/CLionProjects/numpy$ spin python
Invoking `build` prior to invoking Python:
$ /home/home/CLionProjects/numpy/venv/bin/python3 vendored-meson/meson/meson.py compile -C build
INFO: autodetecting backend as ninja
INFO: calculating backend command to run: /home/home/CLionProjects/numpy/venv/bin/ninja -C /home/home/CLionProjects/numpy/build
ninja: Entering directory `/home/home/CLionProjects/numpy/build'
[1/1] Generating numpy/generate-version with a custom command
Saving version to numpy/version.py
$ /home/home/CLionProjects/numpy/venv/bin/python3 vendored-meson/meson/meson.py install --only-changed -C build --destdir ../build-install
$ export PYTHONPATH="/home/home/CLionProjects/numpy/build-install/usr/lib/python3/dist-packages"
🐍 Launching Python with PYTHONPATH="/home/home/CLionProjects/numpy/build-install/usr/lib/python3/dist-packages"
$ /home/home/CLionProjects/numpy/venv/bin/python3 -P
Python 3.13.7 (main, Nov 24 2025, 20:51:28) [GCC 15.2.0] on linux
>>> import numpy as np
>>> np.frompyfunc(lambda x: x).nin
1
>>> np.frompyfunc(lambda x: x).nout
1
>>> np.frompyfunc(lambda x, y: x + y).nin
2
>>> np.frompyfunc(lambda x, y: x + y).nout
1
>>> 
```


numpy/_core/multiarray.pyi
jorenham, line 1: The defaults are incorrect now, and should be `-1` instead.

numpy/_core/multiarray.pyi
```diff
@@ -893,72 +893,72 @@ def fromstring(
...
+    nout: L[1] = 1,
     *,
     identity: IdentityT,
 ) -> _PyFunc_Nin2_Nout1[ReturnT, IdentityT]: ...
 @overload
 def frompyfunc[ReturnT, NInT: int](
     func: Callable[..., ReturnT], /,
-    nin: NInT,
-    nout: L[1],
+    nin: NInT = -1,
```
jorenham: This cannot have a default because otherwise `NInT` might go unresolved

jorenham (changes requested):
Thanks; sounds like a good improvement to me (assuming that it works). 

This will need runtime test, e.g. to verify that non-function callables like 

```py
class F:
    def __call__(self, x):
        return x

assert np.frompyfunc(F(), 1, 1)(42) == 42
```

still work.

numpy/_core/multiarray.pyi
danielzgtg: I reverted all NInt/NoutT lines

numpy/_core/multiarray.pyi
danielzgtg: Can you be more specific please as to which lines of defaults are incorrect?

jorenham: every line that doens't have -1 as default

danielzgtg: Below, you requested `nin: NInT = -1` removed, and `nin: SupportsIndex = -1` is still there.

The remaining ones are `L[1] = 1` and `L[2] = 2`. `L` is `typing.Literal`. Are you sure you want `L[2] = -1`?

jorenham: `_: L[2] = -1` would be incorrect too, it should be `_: L[2, -1] = -1`, so that it describes the case where `_` is either the default (`-1`), or manually specified as `2`

danielzgtg: Ah, that `L[2, -1]` makes sense. Thank you!

jorenham:
The docstring signatures also need to be updated

https://github.com/numpy/numpy/blob/4e2d3f0cdca877aa7e9eca9f089fa4076cda1eee/numpy/_core/_add_newdocs.py#L4886-L4891

That'll (magically) also fix the stubtest errors :)

numpy/_core/tests/test_multiarray.py
```diff
@@ -11107,3 +11107,10 @@ def test_add_newdoc_function_signature(self, func, parameter_names):
         sig = inspect.signature(func)
         assert sig.parameters
         assert tuple(sig.parameters) == parameter_names
+
+def test_frompyfunc_nonfunction_callable():
+    class F:
+        def __call__(self, x):
+            return x
+
+    assert np.frompyfunc(F(), 1, 1)(42) == 42
```
jorenham, line 11157: What will happen without `nin`?

danielzgtg: The new automatic inferencing doesn't work for arbitrary callables. A `IndexError: nin must be positive` will result for `np.frompyfunc(F())`. If requested, I can improve the error message via another `PyArg_ParseTupleAndKeywords`.

 Most users only care about functions incl. lambdas. Using `len(inspect.signature(F()).parameters)` supports more callables, but that's more complicated to import and call from C.

jorenham (approved):
Ok, all good from me on the typing side. It might be good for another maintainer to look at this before merging though.

numpy/_core/src/umath/umathmodule.c
```diff
@@ -78,14 +78,25 @@ ufunc_frompyfunc(PyObject *NPY_UNUSED(dummy), PyObject *args, PyObject *kwds) {
...
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "O|ii$O:frompyfunc", kwlist,
                 &function, &nin, &nout, &identity)) {
         return NULL;
     }
     if (!PyCallable_Check(function)) {
         PyErr_SetString(PyExc_TypeError, "function must be callable");
         return NULL;
     }
+    if (nin < 0 && PyFunction_Check(function)) {
+        nin = ((PyCodeObject *) ((PyFunctionObject *) function)->func_code)->co_argcount;
```
seberg: Hmmm, this feels not ideal from the perspective of supporting other Python implementations?

It seems OK to guess this but I would rather use `inspect` in some manner (yes, via Python, not the C-API).

danielzgtg: Reimplemented in Python with the `inspect` code I wrote above

numpy/_core/tests/test_multiarray.py
danielzgtg: Added tests for without `nin`. `inspect` is used now, see below.

mattip:
Seems to be a net positive. Do we have any idea how this might break current usage? Trying to prevent a regression like #24530

PR #30672, opened 2026-01-19
DEV: make `spin mypy` run `spin build`

danielzgtg:
For consistency with `spin test`. I wondered why `spin mypy` had been ignoring my changes and printing stale results.

### DIFF
--- a/.spin/cmds.py
+++ b/.spin/cmds.py
@@ -544,6 +544,7 @@ def ipython(*, ipython_args, build_dir):
 def mypy(ctx):
     """🦆 Run Mypy tests for NumPy
     """
+    ctx.invoke(build)
     env = os.environ
     env['NPY_RUN_MYPY_IN_TESTSUITE'] = '1'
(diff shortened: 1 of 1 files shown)


jorenham:
Personally I just run `spin stubtest && spin mypy` as combo, but this would indeed make things easier. I'm guessing that it didn't build before because in CI it uses a custom, presumably faster, `spin build` command:

https://github.com/numpy/numpy/blob/4e2d3f0cdca877aa7e9eca9f089fa4076cda1eee/.github/workflows/mypy.yml#L81-L85

danielzgtg:
> a custom, presumably faster, `spin build` command

That's still supported. My new `ctx.invoke(build)` is still an incremental build, so it only takes 1 second before realizing everything's already built by that faster build command thus skippable.

jorenham:
We do the same thing with `spin stubtest`, so it should be fine. Thanks!

PR #30673, opened 2026-01-19
BUG: fix alloc tls destruction on gcc

kumaraditya303:
Addresses the issue discussed at https://github.com/numpy/numpy/pull/30499#issuecomment-3765645237

### DIFF
--- a/numpy/_core/src/multiarray/alloc.cpp
+++ b/numpy/_core/src/multiarray/alloc.cpp
@@ -51,6 +51,12 @@ typedef struct {
     void * ptrs[NCACHE];
 } cache_bucket;
 
+// See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61991
+// gcc has a bug where if the thread_local variable is unused
+// then its destructor may not get called at thread exit.
+// to work around this we define consolidate both datacache
+// and dimcache into a single thread_local variable so
+// that there is no issue of unused thread_local variables.
 typedef struct thread_local_cache {
     cache_bucket datacache[NBUCKETS];
     cache_bucket dimcache[NBUCKETS_DIM];
--- a/numpy/_core/src/multiarray/alloc.cpp
+++ b/numpy/_core/src/multiarray/alloc.cpp
@@ -51,11 +51,10 @@ typedef struct {
     void * ptrs[NCACHE];
 } cache_bucket;
 
-static NPY_TLS cache_bucket datacache[NBUCKETS];
-static NPY_TLS cache_bucket dimcache[NBUCKETS_DIM];
-
-typedef struct cache_destructor {
-    ~cache_destructor() {
+typedef struct thread_local_cache {
+    cache_bucket datacache[NBUCKETS];
+    cache_bucket dimcache[NBUCKETS_DIM];
+    ~thread_local_cache() {
         for (npy_uint i = 0; i < NBUCKETS; ++i) {
             while (datacache[i].available > 0) {
                 free(datacache[i].ptrs[--datacache[i].available]);
@@ -67,9 +66,12 @@ typedef struct cache_destructor {
             }
         }
     }
-} cache_destructor;
+} thread_local_cache;
 
-static NPY_TLS cache_destructor tls_cache_destructor;
+static NPY_TLS thread_local_cache tls_cache;
+
+#define datacache (tls_cache.datacache)
+#define dimcache (tls_cache.dimcache)
 
 /*
  * This function tells whether NumPy attempts to call `madvise` with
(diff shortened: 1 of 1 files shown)


seberg:
Thanks for finding the solution.  I am wondering if we should add a short comment about it, but not super important.
Ping @ngoldbaum if just because it might be interesting to him.

I do wonder if the `armhf` failure may be real though :(.

kumaraditya303:
> Thanks for finding the solution. I am wondering if we should add a short comment about it, but not super important.

I added a comment with link to the gcc issue.

ngoldbaum:
It looks like the armhf tests are using gcc 13.3, in case it helps. The tests crash on `test_arrayprint` which has a multithreaded test, so I bet this is related.

seberg:
Hmmm, could it be that we have to explicitly initialize `available` to zero now!?  (just wondering what the heck could in theory fail...)

kumaraditya303:
It failed again with a crash. Not sure how to debug it on such a platform.

kumaraditya303:
I am guessing it might be because `sizeof(tls_cache) = 66560UL` which might be too large for a single allocation or something on armhf

numpy/_core/src/multiarray/alloc.cpp
```diff
@@ -51,11 +51,16 @@ typedef struct {
...
+// See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61991
+// gcc has a bug where if the thread_local variable is unused
+// then its destructor may not get called at thread exit.
+// to work around this we define consolidate both datacache
+// and dimcache into a single thread_local variable so
+// that there is no issue of unused thread_local variables.
+typedef struct thread_local_cache {
+    cache_bucket datacache[NBUCKETS];
+    cache_bucket dimcache[NBUCKETS_DIM];
+    ~thread_local_cache() {
```
seberg:
```suggestion

    thread_local_cache() {
        for (npy_uint i = 0; i < NBUCKETS; ++i) {
            datacache[i].available = 0;
        }
        for (npy_uint i = 0; i < NBUCKETS_DIM; ++i) {
            dimcache[i].available = 0;
        }
    }

    ~thread_local_cache() {
```
Let's just see if that works, even if it is a random try maybe.

FWIW, I also have 13.3.0 locally and everything is good (also in valgrind `test_multithreaded.py` is fine with this).

seberg:
> I am guessing it might be because sizeof(tls_cache) = 66560UL which might be too large for a single allocation or something on armhf

Ohhh, that would make sense, but in that case we would have to split it up? Debugging it certainly won't be fun :(.

ngoldbaum:
> Not sure how to debug it on such a platform.

I've used the tmate action to debug stuff like this on a github runner: https://github.com/mxschmitt/action-tmate

Otherwise I think you have an ARM mac, right? You should be able to run Linux in a VM or use orbstack  or some other docker environment to run ARM Linux docker containers locally on your Mac to get the armhf container the CI job uses.

kumaraditya303:
It passed with splitting up the cache so it was indeed the size issue.

numpy/_core/src/multiarray/alloc.cpp
```diff
@@ -51,10 +51,16 @@ typedef struct {
     void * ptrs[NCACHE];
 } cache_bucket;
 
-static NPY_TLS cache_bucket datacache[NBUCKETS];
-static NPY_TLS cache_bucket dimcache[NBUCKETS_DIM];
+static NPY_TLS cache_bucket _datacache[NBUCKETS];
+static NPY_TLS cache_bucket _dimcache[NBUCKETS_DIM];
 
 typedef struct cache_destructor {
```
ngoldbaum, line 68: Can we get a comment here about the gcc and armhf issues you ran into?

kumaraditya303: added

ngoldbaum (approved):
Looks good. I'd appreciate it if @seberg can confirm this fixes the leaks he saw in valgrind.

seberg:
I'll test it once more in a bit (it worked with the earlier version, so I think it should be OK).

This large allocation erroring does make me wonder whether it would be cleaner to allocate this so that thread-local is just the pointer to the cache object?!

seberg (approved):
OK, I can confirm that valgrind is happy with this (and so is GCC 13.3.0).

(If nobody has a thought/opinion if an allocation is better, I'll probably just put it in soon. -- I am happy to follow up myself if we should allocate.)

charris:
Backport?

ngoldbaum:
No, https://github.com/numpy/numpy/pull/30499 wasn't and shouldn't be backported.

kumaraditya303:
> This large allocation erroring does make me wonder whether it would be cleaner to allocate this so that thread-local is just the pointer to the cache object?!

That might be a little slower because statically allocated blocks have better locality than block allocated on heap, given that it is not an issue on major platforms I don't think we should change it.

seberg:
Thanks for the multiple follow-ups on this one @kumaraditya303.  Surprisingly weird/tricky/subtle!

PR #30678, opened 2026-01-20
MNT: Improving binsearch explanation

amcandio:
This is a maintenance PR as a follow up from https://github.com/numpy/numpy/pull/30517 regarding binary search explanation.


charris:
Might treat the subtle bit. Something like:
```
// length == floor(length / 2)  + ceil(length / 2), hence
// length -= floor(length / 2)  sets length to ceil(length / 2)
```

charris:
Thanks @amcandio .

PR #30680, opened 2026-01-20
BUG: Fix some bugs found via valgrind

seberg:
A hodgepodge of reference leaks, mostly small or rather irrelevant stuff (although the unique one is large).

Maybe valgrind has it's ups, it did find a Python issue as well, of course that issue was already fixed in 3.14.2 :).

EDIT: Also note, I have not run valgrind on f2py tests here.

EDIT3: There is also a small leak in `numpy/_core/tests/test_dtype.py::TestUserDType::test_custom_structured_dtype_errors`. I am ignoring it (`names` is not cleaned up on the last error, because user dtypes are never fully cleaned up and it is too late, but I don't feel like hard-coding partial cleanup there, if would have to do the full cleanup there).

Since I have no shame, I used this ridiculous script to parallelize.  Your favorite LLM will write a prettier one though easily :).

<details>

```
#!/usr/bin/env python

"""
NOTE: To open the files with actual failures, try something like:

find . -type f ! -name "*valgrind-out.txt" | xargs -I {} sh -c "grep -q -E '(FAILURES|ERRORS)' {} && gnome-text-editor {} &"
"""

from concurrent.futures import ProcessPoolExecutor
import os
import subprocess

os.environ["PYTHONMALLOC"] = "malloc"  # for the spawned processes

test_files = []

DIR = os.getcwd()
FOLDER = "/home/sebastianb/forks/numpy/"

for root, dirs, files in os.walk(FOLDER + "numpy/"):
    files = [f for f in files if f.startswith("test_") and f.endswith(".py")]
    root = root.removeprefix(FOLDER + "numpy/")
    test_files.extend([root + "/" + f for f in files])

for f in test_files[:]:
    # These are particularly slow, so start early.
    if "test_multiarray" in f or "ma/test_core" in f:
        test_files.remove(f)
        test_files.insert(0, f)


setup = """
export PYTHONPATH="/home/sebastianb/forks/numpy/build-install/usr/lib/python3.14/site-packages"
/home/sebastianb/miniforge3/envs/numpy-dev/bin/python3.14 -P -c 'import numpy'
"""

cmd = "PYTHONMALLOC=malloc valgrind --errors-for-leak-kinds=definite --show-leak-kinds=definite --log-file={log_file} "
cmd += '''python -P -mpytest -vv {test_file} -m "not slow"'''
cmd += " --valgrind --valgrind-log={log_file}"


def process(file):
    # Yeah, should use proper pathlib paths :)
    os.chdir(DIR)
    output = os.path.abspath("output/" + file.replace("/", "_").replace(".py", ".txt"))
    log = output.replace(".txt", "-valgrind-out.txt")

    with open(output, "w") as out_file:
        os.chdir("/home/sebastianb/forks/numpy/build-install/usr/lib/python3.14/site-packages")
        actual_command = cmd.format(log_file=log, test_file="numpy/" + file, FOLDER=FOLDER)
        print(actual_command)
        subprocess.run(setup, shell=True)
        subprocess.run(actual_command, shell=True, stdout=out_file, stderr=out_file)
        print("    done file:", file)


def run():
    with ProcessPoolExecutor(16) as executor:
        result = executor.map(process, test_files)

    print(list(result))

if __name__ == "__main__":
    run()
```

### DIFF
--- a/numpy/_core/src/multiarray/mapping.c
+++ b/numpy/_core/src/multiarray/mapping.c
@@ -270,7 +270,7 @@ prepare_index_noarray(int array_ndims, npy_intp *array_dims, PyObject *index,
     npy_intp n;
 
     PyObject *obj = NULL;
-    PyArrayObject *arr;
+    PyArrayObject *arr = NULL;  // free'd on error use Py_CLEAR to decref.
 
     int index_type = 0;
     int ellipsis_pos = -1;
@@ -440,7 +440,6 @@ prepare_index_noarray(int array_ndims, npy_intp *array_dims, PyObject *index,
                     && DEPRECATE("Invalid non-array indices for iterator objects are deprecated and will be "
                                  "removed in a future version. (Deprecated NumPy 2.4)") < 0) {
                     Py_DECREF(tmp_arr);
-                    Py_DECREF(arr);
                     goto failed_building_indices;
                 }
                 Py_DECREF(tmp_arr);
@@ -490,6 +489,7 @@ prepare_index_noarray(int array_ndims, npy_intp *array_dims, PyObject *index,
                     index_type = HAS_BOOL;
                     indices[curr_idx].type = HAS_BOOL;
                     indices[curr_idx].object = (PyObject *)arr;
+                    arr = NULL;  // Reference moved, clean up for error path.
 
                     /* keep track anyway, just to be complete */
                     used_ndim = array_ndims;
@@ -524,7 +524,7 @@ prepare_index_noarray(int array_ndims, npy_intp *array_dims, PyObject *index,
                 indices[curr_idx].value = n;
                 indices[curr_idx].object = PyArray_Zeros(1, &n,
                                             PyArray_DescrFromType(NPY_INTP), 0);
-                Py_DECREF(arr);
+                Py_CLEAR(arr);
 
                 if (indices[curr_idx].object == NULL) {
                     goto failed_building_indices;
@@ -542,7 +542,6 @@ prepare_index_noarray(int array_ndims, npy_intp *array_dims, PyObject *index,
             n = _nonzero_indices((PyObject *)arr, nonzero_result);
 
             if (n < 0) {
-                Py_DECREF(arr);
                 goto failed_building_indices;
             }
 
@@ -553,7 +552,6 @@ prepare_index_noarray(int array_ndims, npy_intp *array_dims, PyObject *index,
                 for (i=0; i < n; i++) {
                     Py_DECREF(nonzero_result[i]);
                 }
-                Py_DECREF(arr);
                 goto failed_building_indices;
             }
 
@@ -567,7 +565,7 @@ prepare_index_noarray(int array_ndims, npy_intp *array_dims, PyObject *index,
                 used_ndim += 1;
                 curr_idx += 1;
             }
-            Py_DECREF(arr);
+            Py_CLEAR(arr);
 
             /* All added indices have 1 dimension */
             if (fancy_ndim < 1) {
@@ -588,7 +586,7 @@ prepare_index_noarray(int array_ndims, npy_intp *array_dims, PyObject *index,
                  */
                 npy_intp ind = PyArray_PyIntAsIntp((PyObject *)arr);
 
-                Py_DECREF(arr);
+                Py_CLEAR(arr);
                 if (error_converting(ind)) {
                     goto failed_building_indices;
                 }
@@ -604,15 +602,17 @@ prepare_index_noarray(int array_ndims, npy_intp *array_dims, PyObject *index,
                 }
             }
 
+            if (fancy_ndim < PyArray_NDIM(arr)) {
+                fancy_ndim = PyArray_NDIM(arr);
+            }
+
             index_type |= HAS_FANCY;
             indices[curr_idx].type = HAS_FANCY;
             indices[curr_idx].value = -1;
             indices[curr_idx].object = (PyObject *)arr;
+            arr = NULL;  // Reference moved, clean up for error path.
 
             used_ndim += 1;
-            if (fancy_ndim < PyArray_NDIM(arr)) {
-                fancy_ndim = PyArray_NDIM(arr);
-            }
             curr_idx += 1;
             continue;
         }
@@ -633,7 +633,6 @@ prepare_index_noarray(int array_ndims, npy_intp *array_dims, PyObject *index,
                     is_flatiter_object ? "" : ", numpy.newaxis (`None`)"
                 );
         }
-        Py_DECREF(arr);
         goto failed_building_indices;
     }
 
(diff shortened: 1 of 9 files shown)


numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -2406,6 +2407,7 @@ array_fromfile(PyObject *NPY_UNUSED(ignored), PyObject *args, PyObject *keywds)
     }
     if (npy_fseek(fp, offset, SEEK_CUR) != 0) {
         PyErr_SetFromErrno(PyExc_OSError);
+        Py_XDECREF(type);
```
seberg, line 2410: NOTE: This one *wasn't* found by valgrind, but be me looking for a leak (that turned out elsewhere).

(I.e. this line is untested, so please double check me.)

numpy/_core/src/multiarray/unique.cpp
```diff
@@ -508,9 +508,11 @@ array__unique_hash(PyObject *NPY_UNUSED(module),
         auto type = PyArray_TYPE(arr);
         // we only support data types present in our unique_funcs map
         if (unique_funcs.find(type) == unique_funcs.end()) {
-            Py_RETURN_NOTIMPLEMENTED;
+            result = Py_NewRef(Py_NotImplemented);
```
kumaraditya303, line 511: Py_NotImplemented is immortal so incref has no effect

seberg: Wouldn't it still make the global refcount be wrong (at least on older Python versions)?

kumaraditya303: immortal objects were implemented in 3.12 and numpy only supports 3.12+ so the global refcount is correct on debug builds.

charris:
@kumaraditya303 If backported there is still 3.11 to consider.

numpy/_core/src/multiarray/unique.cpp
seberg: Ah good point. So in this case (since we should backport), it might make some sense (maybe?).
But in general, I guess you are right, we should start to roll with the immortality.

numpy/_core/src/multiarray/mapping.c
```diff
@@ -440,6 +440,7 @@ prepare_index_noarray(int array_ndims, npy_intp *array_dims, PyObject *index,
                     && DEPRECATE("Invalid non-array indices for iterator objects are deprecated and will be "
                                  "removed in a future version. (Deprecated NumPy 2.4)") < 0) {
                     Py_DECREF(tmp_arr);
+                    Py_DECREF(arr);
```
ngoldbaum: I think we're also missing a `Py_DECREF(arr)` in the error path for `PyObject_IsTrue` below. Maybe it would be easier to keep track if `failed_building_indices` had a `Py_XDECREF(arr)` in it.

numpy/_core/src/multiarray/multiarraymodule.c
ngoldbaum: Isn't the `Py_XDECREF(type)` in the `PyArg_ParseTupleAndKeywords` error path incorrect? I don't actually know what happens if the converter succeeds but somehow `PyArg_ParseTupleAndKeywords` fails, so maybe it is needed.

Other than that, the refcounting for type looks correct to me in this PR.

numpy/_core/src/multiarray/refcount.c
```diff
@@ -124,7 +124,7 @@ PyArray_ZeroContiguousBuffer(
      * and only arrays which own their memory should clear it.
      */
     int aligned = PyArray_ISALIGNED(arr);
-    if (PyArray_ISCONTIGUOUS(arr)) {
+    if (PyArray_ISCONTIGUOUS(arr) || PyArray_IS_F_CONTIGUOUS(arr)) {
```
ngoldbaum, line 127: I don't follow the logic for this one - can you elaborate a bit?

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -2734,6 +2734,7 @@ PyUFunc_Accumulate(PyUFuncObject *ufunc, PyArrayObject *arr, PyArrayObject *out,
     npy_intp fixed_strides[3];
     if (need_outer_iterator) {
         NpyIter_GetInnerFixedStrideArray(iter, fixed_strides);
+        fixed_strides[2] = fixed_strides[0];
```
ngoldbaum: Lift this out of the if/else since it's repeated in both branches.

I also don't understand why this assignment is needed - a comment might be helpful to aid future readers.

numpy/_core/tests/test_arrayobject.py
```diff
@@ -73,3 +75,21 @@ def test_array_wrap(subclass_self, subclass_arr):
     # Non 0-D array can't be converted to scalar, so we ignore that
     arr1d = np.array([3], dtype=np.int8).view(subclass_arr)
     assert type(arr.__array_wrap__(arr1d, None, True)) is type(arr)
+
+
+@pytest.mark.skipif(not HAS_REFCOUNT, reason="Python lacks refcounts")
+def test_cleanup_with_refs_non_contig():
```
ngoldbaum, line 81: I guess this test found the `ZeroContiguousBuffer` leak?

ngoldbaum (commented):
Found myself with some time between meetings and did a code review pass. Mostly looks great but I spotted a couple issues.

numpy/_core/src/multiarray/refcount.c
seberg: I just forgot this when writing.  If we have a contiguous array, even if it is fortran order, we can just do a single 1-D iteration.
This "fixed" the test that failed, but of course not the bug (below), which is why I added that new test.

numpy/_core/tests/test_arrayobject.py
seberg: `PyArray_ClearArray` but right (I didn't touch the `ZeroContiguousBuffer` code).

numpy/_core/src/multiarray/multiarraymodule.c
seberg: I have an inkling that there was a way to do cleanup with converters, but no they don't, the XDECREF there is correct.

numpy/_core/src/multiarray/mapping.c
seberg: Good thought. Need to use `Py_CLEAR` rather than `DECREF` on the happy paths, but that seems still a bit better.  Also least this way the failure would be a double DECREF, which is probably a more reliable failure.

numpy/_core/src/umath/ufunc_object.c
seberg: OK, let me move it with a comment. Accumulate is an in-place op, all in-place ops do this (since they use the `a, b, out=a`.

ngoldbaum:
Thanks @seberg!

PR #30683, opened 2026-01-20
MAINT: update pins in requirements files

ngoldbaum:
I updated all the pins except for `spin`, where I simply deleted them. IMO spin is stable enough it's OK to run with unpinned versions.

The most substantial change is to accommodate ruff rule ISC004 "Unparenthesized implicit string concatenation in collection": https://docs.astral.sh/ruff/rules/implicit-string-concatenation-in-collection-literal/

Punting on the docs pins, see #30684


numpy/_core/tests/test_stringdtype.py
```diff
@@ -1457,7 +1457,7 @@ def test_unary(string_array, unicode_array, function_name):
     "strip",
     "lstrip",
     "rstrip",
-    "replace"
+    "replace",
```
ngoldbaum, line 1460: 🙃 

Perhaps concerningly: fixing this didn't cause the correspond test below to fail. I unfortunately don't have time to dive in to understand why though.

ngoldbaum:
Maybe we should be running dependabot on these pins, if that's possible?

charris:
> if that's possible?

Should be, we could also add the tools directory. I'll put up a PR for requirements.

charris:
See #30685. Dependabot can be manually triggered.

numpy/testing/_private/utils.py
```diff
@@ -2848,6 +2848,8 @@ def wrapper(*args, **kwargs):
             pytest.skip("Pyston disables recursion checking")
         if IS_WASM:
             pytest.skip("WASM has limited stack size")
+        if not IS_64BIT:
+            pytest.skip("32 bit Python has limited stack size")
```
ngoldbaum, line 2852: To avoid a segfault on the `i686-cp312` test job: https://github.com/numpy/numpy/actions/runs/21183357907/job/60931865175?pr=30683

I suspect that the different pytest version is leading to a slightly different Python stack structure. I could try to understand more deeply too, but I don't think it's worth spending a bunch of time on an issue that only happens on 32 bit Python in a stress-test that we know causes issues in some situations.

charris:
Thanks Nathan. I note that you have fixed the ruff problems the dependabot update showed.

PR #30692, opened 2026-01-20
MAINT: update OpenBLAS

mattip:
OpenBLAS released 0.3.31, and I created scipy-openblas-0.3.31.22.0. Let's try it out. Note that scipy/scipy#24342 with 0.3.30.443 (some 20 commits behind this one) seems to have some problem with 32-bit linux. Not clear whether that is an OpenBLAS problem or a SciPy problem.

### DIFF
--- a/requirements/ci32_requirements.txt
+++ b/requirements/ci32_requirements.txt
@@ -1,3 +1,3 @@
 spin
 # Keep this in sync with ci_requirements.txt
-scipy-openblas32==0.3.30.0.7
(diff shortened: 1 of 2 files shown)


charris:
The failing test warns that `numpy-release` will also need an update.

charris:
Thanks Matti.

danielzgtg:
After rebasing onto your change, my #30670 and someone's #30707 fail Windows CI with an error message related to BLAS.

PR #30696, opened 2026-01-21
MAINT: Fix typo 'paramterize' in test_multiarray.py

weibo-gao:
This PR fixes a typo in the comments of `test_multiarray.py`.
`paramterize` -> `parameterize`

This correction improves keyword searchability (grep) for the term "parameterize" within the codebase.


charris:
Thanks @weibo-gao . There seems to be no universally accepted form of this word, Parameterize is a common variant, sometimes with "s" instead of "z".

PR #30698, opened 2026-01-21
TST: fix limited API example in tests for latest Cython

rgommers:
Cython 3.3 will require a minimum target of 3.9 for the limited API. There is no reason for us to still test 3.6/3.7

Remove `#define Py_LIMITED_API` from a C source file, that's not a good habit and it requires manually keeping the version in sync between the C file and the meson.build file.

Also fix warnings like these:
```
WARNING: Project specifies no minimum version but uses features which were added in versions:
 * 1.3.0: {'limited_api arg in python.extension_module'}
```
by requiring a recent Meson version.


charris (approved):

charris:
Thanks Ralf.

Issue #30704, opened 2026-01-21
Support for the new Limited API and Stable ABI in Python 3.15 (PEP 803/809)


rgommers:
_Note: [PEP 803](https://peps.python.org/pep-0803) and [PEP 809](https://peps.python.org/pep-0809) - which are two competing PEPs with large overlap to define a new Stable ABI that will work with both GIL-enabled and free-threaded interpreters - are still in Draft state. This issue tries to summarize what would be needed if one of them got accepted._

NumPy doesn't use the Limited C API and Stable ABI itself, but does support other projects targeting the Stable ABI while using the NumPy headers. Support for that was implemented about two years ago in gh-25531, shipped in NumPy 2.0. It isn't widely used yet, but seems to work fine. 

I worked on PyWavelets using the Stable ABI recently (https://github.com/PyWavelets/pywt/pull/828), and decided to try that with PEP 803 and CPython 3.15a5, given that there are branches of Cython ([cython#7399](https://github.com/cython/cython/issues/7399)) and meson-python ([mgorny/meson-python#freethreading-limited-api](https://github.com/mgorny/meson-python/tree/freethreading-limited-api)).

After doing the plumbing to make PyWavelets use the new abi3t, I got an unpleasant surprise from NumPy:
```
     ....
     [24/32] Compiling C object pywt/_extensions/_pywt.abi3.so.p/meson-generated__pywt.c.o
      samu: job failed: cc -Ipywt/_extensions/_pywt.abi3.so.p [snip] D_Py_OPAQUE_PYOBJECT -DPy_LIMITED_API=0x030f0000 -o pywt/_extensions/_pywt.abi3.so.p/meson-generated__pywt.c.o -c pywt/_extensions/_pywt.abi3.so.p/_pywt.c
      In file included from /home/rgommers/code/tmp/python-builds/py315t-21jan/lib/python3.15t/site-packages/numpy/_core/include/numpy/ndarrayobject.h:12,
                       from /home/rgommers/code/tmp/python-builds/py315t-21jan/lib/python3.15t/site-packages/numpy/_core/include/numpy/arrayobject.h:5,
                       from pywt/_extensions/_pywt.abi3.so.p/_pywt.c:1164:
      /home/rgommers/code/tmp/python-builds/py315t-21jan/lib/python3.15t/site-packages/numpy/_core/include/numpy/ndarraytypes.h:654:9: error: unknown type name 'PyObject_HEAD'; did you mean 'PyObject_DEL'? 
```

[...]

PR #30705, opened 2026-01-21
MAINT: replace ob_type access with Py_TYPE in PyArray_CheckExact

ngoldbaum:
Directly accessing `ob_type` like this will not be possible in a future ABI where PyObject is opaque (https://github.com/numpy/numpy/issues/30704). Py_TYPE is in the stable ABI starting in Python 3.14 and has been available outside the stable ABI forever, so we can use that instead to avoid accessing CPython internal details.


seberg:
I guess we still use it in a few places in C code, could update, but no need here (as this is interesting for public API).

Let's put this in, thanks Nathan.

PR #30709, opened 2026-01-22
BUG: Fix some bugs found via valgrind (#30680)

charris:
Backport of #30680.

A hodgepodge of reference leaks, mostly small or rather irrelevant stuff (although the unique one is large).

Maybe valgrind has it's ups, it did find a Python issue as well, of course that issue was already fixed in 3.14.2 :).

EDIT: Also note, I have not run valgrind on f2py tests here.

EDIT3: There is also a small leak in `numpy/_core/tests/test_dtype.py::TestUserDType::test_custom_structured_dtype_errors`. I am ignoring it (`names` is not cleaned up on the last error, because user dtypes are never fully cleaned up and it is too late, but I don't feel like hard-coding partial cleanup there, if would have to do the full cleanup there).

Since I have no shame, I used this ridiculous script to parallelize.  Your favorite LLM will write a prettier one though easily :).

<details>

```
#!/usr/bin/env python

"""
NOTE: To open the files with actual failures, try something like:

find . -type f ! -name "*valgrind-out.txt" | xargs -I {} sh -c "grep -q -E '(FAILURES|ERRORS)' {} && gnome-text-editor {} &"
"""

from concurrent.futures import ProcessPoolExecutor
import os
import subprocess

os.environ["PYTHONMALLOC"] = "malloc"  # for the spawned processes

test_files = []

DIR = os.getcwd()
FOLDER = "/home/sebastianb/forks/numpy/"

for root, dirs, files in os.walk(FOLDER + "numpy/"):
    files = [f for f in files if f.startswith("test_") and f.endswith(".py")]
    root = root.removeprefix(FOLDER + "numpy/")
    test_files.extend([root + "/" + f for f in files])

for f in test_files[:]:
    # These are particularly slow, so start early.
    if "test_multiarray" in f or "ma/test_core" in f:
        test_files.remove(f)
        test_files.insert(0, f)


setup = """
export PYTHONPATH="/home/sebastianb/forks/numpy/build-install/usr/lib/python3.14/site-packages"
/home/sebastianb/miniforge3/envs/numpy-dev/bin/python3.14 -P -c 'import numpy'
"""

cmd = "PYTHONMALLOC=malloc valgrind --errors-for-leak-kinds=definite --show-leak-kinds=definite --log-file={log_file} "
cmd += '''python -P -mpytest -vv {test_file} -m "not slow"'''
cmd += " --valgrind --valgrind-log={log_file}"


def process(file):
    # Yeah, should use proper pathlib paths :)
    os.chdir(DIR)
    output = os.path.abspath("output/" + file.replace("/", "_").replace(".py", ".txt"))
    log = output.replace(".txt", "-valgrind-out.txt")

    with open(output, "w") as out_file:
        os.chdir("/home/sebastianb/forks/numpy/build-install/usr/lib/python3.14/site-packages")
        actual_command = cmd.format(log_file=log, test_file="numpy/" + file, FOLDER=FOLDER)
        print(actual_command)
        subprocess.run(setup, shell=True)
        subprocess.run(actual_command, shell=True, stdout=out_file, stderr=out_file)
        print("    done file:", file)


def run():
    with ProcessPoolExecutor(16) as executor:
        result = executor.map(process, test_files)

    print(list(result))

if __name__ == "__main__":
    run()
```

### DIFF
--- a/numpy/_core/src/multiarray/item_selection.c
+++ b/numpy/_core/src/multiarray/item_selection.c
@@ -1371,7 +1371,6 @@ fail:
     if (needcopy) {
         PyArray_ClearBuffer(odescr, buffer, elsize, N, 1);
         PyDataMem_UserFREE(buffer, N * elsize, mem_handler);
-        Py_DECREF(odescr);
     }
     if (ret < 0 && !PyErr_Occurred()) {
         /* Out of memory during sorting or buffer creation */
@@ -1384,6 +1383,7 @@ fail:
     if (PyErr_Occurred() && ret == 0) {
         ret = -1;
     }
+    Py_XDECREF(odescr);
     Py_DECREF(it);
     Py_DECREF(mem_handler);
     NPY_cast_info_xfree(&to_cast_info);
@@ -1594,7 +1594,6 @@ fail:
     if (needcopy) {
         PyArray_ClearBuffer(odescr, valbuffer, elsize, N, 1);
         PyDataMem_UserFREE(valbuffer, N * elsize, mem_handler);
-        Py_DECREF(odescr);
     }
     PyDataMem_UserFREE(idxbuffer, N * sizeof(npy_intp), mem_handler);
     if (ret < 0) {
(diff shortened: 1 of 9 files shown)


charris:
MacOS 3,14t seems to have gone missing.

PR #30710, opened 2026-01-22
BUG: Fixup the quantile promotion fixup (promoted to 64bit always)

seberg:
Sorry, I incorrectly reverted a line and made things maybe worse :(. Basically it now *always* returned float64 (at least for most methods).

I am surprised that we had not tests for this at all.

I suppose this needs backporting, which isn't great, but...


charris (approved):

charris:
Thanks Sebastian.

PR #30711, opened 2026-01-22
TYP: `RandomState`: various fixes

jorenham:
This fixes many `overlapping-overload` errors (towards #27032) in a backwards-compatible manner. I also reformatted `random/mtrand.pyi` to make it compatible with `ruff format` and to have a style that's consistent with the rest of the stubs.


charris:
Thanks Joren.

PR #30714, opened 2026-01-22
TYP: ``fft.[r]fftfreq`` shape-typing and precise dtypes

jorenham:
This was relatively easy (8 overloads) because `fftfreq` only accepts arrays and python scalars, not "true" array-likes:

```pycon
>>> np.fft.fftfreq(2, np.array([1,2]))
array([ 0.  , -0.25])
>>> np.fft.fftfreq(2, [1,2])
    np.fft.fftfreq(2, [1,2])
    ~~~~~~~~~~~~~~^^^^^^^^^^
  File "/home/joren/.cache/uv/archive-v0/_u3DEwvQ5LAEKPpZu8vx_/lib/python3.14/site-packages/numpy/fft/_helper.py", line 170, in fftfreq
    val = 1.0 / (n * d)
          ~~~~^~~~~~~~~
TypeError: unsupported operand type(s) for /: 'float' and 'list'
```

and that's great because this way we *always* are able to determine the shape-type of the input array (or scalar). Otherwise, we'd have to separately overload each of the first couple of dimensions. That quickly becomes a mess, and doesn't work for deeply nested lists and such (probably rare, but still).
/rant

Oh yea, this also solves all two `overlap-overload` errors in `fft._helper`; towards #27032


charris:
Thanks Joren.

PR #30716, opened 2026-01-23
BUG: Fix MaskedRecords.view dtype reset for ndarray subclasses

gamila-wisam:
Summary

This PR fixes a bug in MaskedRecords.view() where the dtype reset was missing when viewing as an ndarray subclass. The current implementation contains identical code in both branches of the issubclass(dtype, np.ndarray) conditional, which prevents the dtype from being reset properly.

What was wrong?

When dtype is an ndarray subclass, the code should set dtype = None after creating the view, mirroring the behavior in numpy/ma/core.py. Without this, subsequent logic may behave incorrectly and the conditional becomes ineffective.

Fix

Added the missing dtype = None assignment in the issubclass(dtype, np.ndarray) branch.

Test

Added a new test test_view_ndarray_subclass_resets_dtype to ensure:

The view returns the correct subclass type

The dtype is preserved

Mask and fill values remain consistent

Fixes

Closes #30441

### DIFF
--- a/numpy/ma/core.py
+++ b/numpy/ma/core.py
@@ -3276,6 +3276,16 @@ class MaskedArray(ndarray):
         # also make the mask be a view (so attr changes to the view's
         # mask do no affect original object's mask)
         # (especially important to avoid affecting np.masked singleton)
+        # ensure subclass views preserve mask and fill_value
+        
+        if not hasattr(output, "_mask"):
+            output._mask = self._mask
+        if not hasattr(output, "_fill_value"):
+            output._fill_value = self._fill_value
+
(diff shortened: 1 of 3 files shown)


Aniketsy:
@gamila-wisam  small suggestion-- maybe check if someone is already working on the issue before opening a PR, as there’s already one open.

gamila-wisam:
Hi @Aniketsy, Thanks for the note! I saw the existing PR but it’s been inactive for a while, so I tried an alternative approach to explore the fix from a different angle, Happy to iterate or coordinate if needed.

PR #30721, opened 2026-01-24
DOC: clarify nesting limit of _FiniteNestedSequence

Anirudh-kasthuri:
While using _FiniteNestedSequence, I ran into confusing behavior with deeply nested lists not being accepted by type checkers.

Reading the alias definition shows that it is intentionally limited to four levels of nesting, but this limitation is not documented. This PR adds a short comment to make that constraint explicit so users don’t have to infer it from the source.

No behavior changes, documentation only.

### DIFF
--- a/numpy/_typing/_array_like.py
+++ b/numpy/_typing/_array_like.py
@@ -37,6 +37,7 @@ class _SupportsArrayFunc(Protocol):
 
 
 # TODO: Wait until mypy supports recursive objects in combination with typevars
+# Limited to 4 nesting levels - e.g. [[[[1]]]] works, [[[[[1]]]]] doesn't
 type _FiniteNestedSequence[T] = (
     T
     | Sequence[T]


jorenham:
> While using _FiniteNestedSequence,

what for?

jorenham:
The only place that still uses `_FiniteNestedSequence` is `ndenumerate.__init__`, so I'd rather get rid of it altogether.

Anirudh-kasthuri:
Ha, fair point! Didn't realize it was only used in one spot.

I can close this and work on removing it entirely if you think that's better. Just point me in the right direction for what ndenumerate should use instead and I'll put together a cleanup PR.

Seems like that'd be more valuable than documenting something that's on the way out anyway.

jorenham:
> Ha, fair point! Didn't realize it was only used in one spot.
> 
> I can close this and work on removing it entirely if you think that's better. Just point me in the right direction for what ndenumerate should use instead and I'll put together a cleanup PR.
> 
> Seems like that'd be more valuable than documenting something that's on the way out anyway.

`_NestedSequence`

Anirudh-kasthuri:
Got it, switching to _NestedSequence. I'll close this and open a cleanup PR. Thanks!

PR #30729, opened 2026-01-26
TYP: Type-checking with Pyrefly 

jorenham:
Currently only mypy type-checks the stubs. But since mypy is rather slow and has several open bugs, it could be to use idea to add type-checker. Pyrefly is still in beta, but I've had good experiences with it in scipy-stubs, and it's significantly faster than mypy. Another important advantage is that, unlike mypy, it has/is an LSP.

The next step will be to also use pyrefly for type-testing, i.e. type-checking the stuff under `numpy/typing/tests`


jorenham:
cc @MarcoGorelli

MarcoGorelli (commented):
nice!

I noticed that
```
WARN `chararray` is deprecated [deprecated]
 --> numpy/char/__init__.pyi:7:5
  |
7 |     chararray,
  |     ---------
  |
  The chararray class is deprecated and will be removed in a future release. Use an ndarray with a string or bytes dtype instead.
Warning: `chararray` is deprecated
```
is in the logs, is that worth silencing?

Encouraging that this only takes 2 seconds anyway

Also, it's nice that pyrefly flags something like
```py
class Foo:
    bar: tuple[int, ...]

class Bar(Foo):
    bar: tuple[int, int]
```
which does look like a safety concern is `bar` were written to, even though pyright/mypy don't flag anything

jorenham:
> I noticed that
> 
> ```
> WARN `chararray` is deprecated [deprecated]
>  --> numpy/char/__init__.pyi:7:5
>   |
> 7 |     chararray,
>   |     ---------
>   |
>   The chararray class is deprecated and will be removed in a future release. Use an ndarray with a string or bytes dtype instead.
> Warning: `chararray` is deprecated
> ```
> 
> is in the logs, is that worth silencing?

Hmm yea, you're right: https://github.com/numpy/numpy/actions/runs/21372731006/job/61521224077?pr=30729

But are you able to reproduce that locally? Because I don't see it :thinking: 

```bash
$ uvx pyrefly check
 INFO Checking project configured at `/home/joren/Workspace/numpy/pyproject.toml`
 INFO 0 errors (238 suppressed)
```

```bash
$ spin pyrefly
$ pyrefly check
 INFO Checking project configured at `/home/joren/Workspace/numpy/pyproject.toml`
 INFO 0 errors (238 suppressed)
```

jorenham:
If a `# pyrefly:ignore[deprecated]` the reported line and then run `uvx pyrefly check --remove-unused-ignores`, it removes it :man_shrugging:

MarcoGorelli:
yeah i get
```console
$ spin run pyrefly check 
$ /home/marcogorelli/miniconda3/envs/numpy-dev/bin/python3.12 vendored-meson/meson/meson.py compile -C build
$ /home/marcogorelli/miniconda3/envs/numpy-dev/bin/python3.12 vendored-meson/meson/meson.py install --only-changed -C build --destdir ../build-install
 WARN PYTHONPATH environment variable is set to `/home/marcogorelli/numpy-dev/build-install/usr/lib/python3.12/site-packages`. Checks in other environments may not include these paths.
 INFO Checking project configured at `/home/marcogorelli/numpy-dev/pyproject.toml`
 INFO 0 errors (238 suppressed)
```

for my understanding, any reason to use `uvx` here instead of `spin run pyrefly check`?

jorenham:
> yeah i get
> 
> ```
> $ spin run pyrefly check 
> $ /home/marcogorelli/miniconda3/envs/numpy-dev/bin/python3.12 vendored-meson/meson/meson.py compile -C build
> $ /home/marcogorelli/miniconda3/envs/numpy-dev/bin/python3.12 vendored-meson/meson/meson.py install --only-changed -C build --destdir ../build-install
>  WARN PYTHONPATH environment variable is set to `/home/marcogorelli/numpy-dev/build-install/usr/lib/python3.12/site-packages`. Checks in other environments may not include these paths.
>  INFO Checking project configured at `/home/marcogorelli/numpy-dev/pyproject.toml`
>  INFO 0 errors (238 suppressed)
> ```

Hmm, strange... I opened https://github.com/facebook/pyrefly/issues/2258

> for my understanding, any reason to use `uvx` here instead of `spin run pyrefly check`?

It avoids having to install most dependencies, so it's just a performance thing.

BTW, `spin pyrefly` also works :)

jorenham:
@connernilsen [figured out](https://github.com/facebook/pyrefly/issues/2258#issuecomment-3857274038) was was going on with the deprecation warning:

> [@jorenham](https://github.com/jorenham) Okay, this is a really weird way for an out of date branch to present :(
> 
> The `main` branch of `numpy` [added the `@deprecated` decorator to `chararray`](https://github.com/numpy/numpy/blob/0d7986494b39ace565afda3de68be528ddade602/numpy/_core/defchararray.pyi#L89-L93), and the `checkout` action seems to attempt to rebase your changes on top of that. Because of that, in CI you'll get the error because you're looking at the rebased version, but locally, you don't have those changes yet.
> 
> If you rebase/merge with `numpy/main`, you'll see the deprecated warning. I think the checkout action is a bit silent about doing the rebase, which makes this a little more confusing than it needed to be, but I think everything is working as expected.

And indeed, after rebasing on main, the `deprecated` warning also turned up locally. I've added the appropriate ignore comment, so now there are no warnings (or errors) anymore, and this is ready to merge AFAIK.

jakevdp:
Sorry to abuse this PR thread, but since you're exploring pyrefly here I thought you might have looked at this. I've found that with the latest pyrefly and latest NumPy, this code raises an error:
```python
import numpy as np
print(np.iinfo(np.int32).max)
```
```
$ uv run pyrefly check test.py 
ERROR No matching overload found for function `numpy._core.getlimits.iinfo.__init__` called with arguments: (type[signedinteger[_32Bit]]) [no-matching-overload]
 --> test.py:2:15
  |
2 | print(np.iinfo(np.int32).max)
  |               ^^^^^^^^^^
  |
  Possible overloads:
  (int_type: _HasDType[dtype[_IntegerT_co]] | _HasNumPyDType[dtype[_IntegerT_co]] | dtype[_IntegerT_co] | _IntegerT_co | type[_IntegerT_co]) -> None [closest match]
  (int_type: Literal['<n', '=n', '>n', 'int', 'int_', 'intp', 'n', '|n'] | int | type[int]) -> None
  (int_type: _Int8Codes) -> None
  (int_type: _UInt8Codes) -> None
  (int_type: _Int16Codes) -> None
  (int_type: _UInt16Codes) -> None
  (int_type: _Int32Codes) -> None
  (int_type: _UInt32Codes) -> None
  (int_type: _Int64Codes) -> None
  (int_type: _UInt64Codes) -> None
  (int_type: str) -> None
 INFO 1 error
```
Do you think this is a pyrefly bug, a NumPy bug, or is it a pattern I should stop using?

jorenham:
> ```python
> import numpy as np
> print(np.iinfo(np.int32).max)
> ```

This issue seems to only occur with pyrefly. When I run this through mypy, (based)pyright, ty, they all report 0 errors. So could you raise this at pyrefly?

jakevdp:
> So could you raise this at pyrefly?

https://github.com/facebook/pyrefly/issues/2398

mattip (approved):
LGTM, we discussed this in the triage meeting and it got a 👍

seberg:
Yeah, sounds good, just need to remove that accidental submodule update.

jorenham:
> Yeah, sounds good, just need to remove that accidental submodule update.

done

---

The CI failure seems unrelated

seberg (approved):

seberg:
Thanks Joren! Will be curious if we end up with a long list of type checkers soon :).

PR #30667, opened 2026-01-17
BUG: fix thread safety of `array_getbuffer`

kumaraditya303:
Fixes https://github.com/numpy/numpy/issues/30648

PR #30736, opened 2026-01-27
BUG: fix thread safety of `array_getbuffer` (#30667)

charris:
Backport of #30667.

Add critical sections around buffer info creation/mutation.

Fixes https://github.com/numpy/numpy/issues/30648

### DIFF
--- a/numpy/_core/tests/test_multithreading.py
+++ b/numpy/_core/tests/test_multithreading.py
@@ -375,3 +375,29 @@ def test_arg_locking(kernel, outcome):
         finally:
             if len(tasks) < 5:
                 b.abort()
+
+def test_array__buffer__thread_safety():
+    import inspect
+    arr = np.arange(1000)
+    flags = [inspect.BufferFlags.STRIDED, inspect.BufferFlags.READ]
+
+    def func(b):
+        b.wait()
+        for i in range(100):
+            arr.__buffer__(flags[i % 2])
+
+    run_threaded(func, max_workers=8, pass_barrier=True)
+
+def test_void_dtype__buffer__thread_safety():
+    import inspect
+    dt = np.dtype([('name', np.str_, 16), ('grades', np.float64, (2,))])
+    x = np.array(('ndarray_scalar', (1.2, 3.0)), dtype=dt)[()]
+    assert isinstance(x, np.void)
+    flags = [inspect.BufferFlags.STRIDES, inspect.BufferFlags.READ]
+
+    def func(b):
+        b.wait()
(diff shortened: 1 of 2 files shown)


charris:
Looks like this needs an inspect fixup. @jorenham IIRC, you were involved in that.

charris:
Actually, looks like the problem is Python 3.11.

jorenham:
> Looks like this needs an inspect fixup. @jorenham IIRC, you were involved in that.

`inspect.BufferFlags` was added in Python 3.12 (https://docs.python.org/3/library/inspect.html#inspect.BufferFlags). It's an `IntEnum`, and looking at the [source code](https://github.com/python/cpython/blob/a76e4cd62dd68e7cbe86e37e6ed988495a646b66/Lib/inspect.py#L3352-L3371), `STRIDED` can be replaced with `25`, `STRIDES` with `24`, and `READ` with `256`, i.e. `flags=[25, 256]` and `flags=[24, 256]` for the first and second added tests, respectively.

numpy/_core/tests/test_multithreading.py
```diff
@@ -375,3 +375,29 @@ def replace_list_items(b):
         finally:
             if len(tasks) < 5:
                 b.abort()
+
+def test_array__buffer__thread_safety():
+    import inspect
+    arr = np.arange(1000)
+    flags = [inspect.BufferFlags.STRIDED, inspect.BufferFlags.READ]
```
jorenham:
```suggestion
    flags = [25, 256]
```

numpy/_core/tests/test_multithreading.py
```diff
@@ -375,3 +375,29 @@ def replace_list_items(b):
...
+            arr.__buffer__(flags[i % 2])
+
+    run_threaded(func, max_workers=8, pass_barrier=True)
+
+def test_void_dtype__buffer__thread_safety():
+    import inspect
+    dt = np.dtype([('name', np.str_, 16), ('grades', np.float64, (2,))])
+    x = np.array(('ndarray_scalar', (1.2, 3.0)), dtype=dt)[()]
+    assert isinstance(x, np.void)
+    flags = [inspect.BufferFlags.STRIDES, inspect.BufferFlags.READ]
```
jorenham, line 399:
```suggestion
    flags = [24, 256]
```

charris:
Thanks Joren, let's give that a shot.

jorenham:
Hmmm, I guess we should just disable these tests completely on Python<3.12

PR #30333, opened 2025-11-29
BEG, MAINT: Begin NumPy 2.5 development

charris:
- Create 2.5.0-notes.rst
- Update release.rst
- Update pyproject.toml
- Update cversions.txt
- Update numpyconfig.h
- Update numpy/_core/meson.build
- Delete release fragments

PR #30740, opened 2026-01-28
MAINT: correct typo in NumPy 2.5 C API version

ngoldbaum:
@charris typo'd this in #30333. The `NPY_2_5_API_VERSION` macro should match the `C_API_VERSION` definition in `numpy/_core/meson.build` but right now they're out of sync:

https://github.com/numpy/numpy/blob/2ccea713e1bbb447666c00932c1020aa2572d4c2/numpy/_core/meson.build#L55

I hit this writing a test that did `-DNPY_TARGET_VERSION=NPY_2_5_API_VERSION`, which triggered an error about the FEATURE_VERSION being higher than the version in NumPy's headers.


seberg:
Thanks, Nathan. Yeah, I don't think 2.5 is it's own C-API version quite yet :).

PR #30741, opened 2026-01-28
CI: add smoke test CI coverage for Python 3.15

ngoldbaum:
This is a little earlier than we normally do this, but I'd like to have some Python 3.15 CI coverage to better test my fixes for #30704.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -67,7 +67,7 @@ jobs:
       MESON_ARGS: "-Dallow-noblas=true -Dcpu-baseline=none -Dcpu-dispatch=none"
     strategy:
       matrix:
-        version: ["3.12", "3.13", "3.14", "3.14t"]
+        version: ["3.12", "3.13", "3.14", "3.14t", "3.15-dev", "3.15t-dev"]
     steps:
     - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
       with:
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -67,7 +67,7 @@ jobs:
       MESON_ARGS: "-Dallow-noblas=true -Dcpu-baseline=none -Dcpu-dispatch=none"
     strategy:
       matrix:
-        version: ["3.12", "3.13", "3.14", "3.14t", "3.15-dev", "3.15t-dev"]
+        version: ["3.12", "3.14t"]
     steps:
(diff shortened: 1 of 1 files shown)


andyfaff:
Do you have a branch/PR for the fixes? Would that be a better place to have the CI change?

ngoldbaum:
> Do you have a branch/PR for the fixes? Would that be a better place to have the CI change?

We could wait for those, but I'd also like to know if NumPy is breaking Python 3.15 support in CI. Right now the `main` branch works with Python 3.15a5 and I'd like to know if either us or upstream breaks anything.

I'm happy to commit to fixing any issues that come up before the betas. We could also make these checks non-voting if anyone is concerned about that.

andyfaff:
It's more that the past approach has been to wait for beta phase, so that we don't experience difficulties if the alpha phase give us lots of CI fails. (the community always chirps up at the start of alpha asking us to test against that). It just adds some stability.
That being said, if there's big changes upstream that could break things in a major way, it's good to get ahead of that. Could we mark those matrix entries as experimental, so that they don't break CI. See the code block in https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstrategyfail-fast.

seberg (approved):
Thanks, seems all good, might be just as well to fix things along the way (and we can disable it if it gets annoying).  But generally happy to get this in.

That said, the "smoke test" category is growing a bit much maybe?  It might make sense to just keep 2 entries (say 3.12 and 3.14t?) and move all to a new step.
I am not sure it helps a lot with the tests being spread far, but the idea for smoke-tests is to not waste CI on failures.

EDIT: Yeah, in the past we didn't do it, but Nathan is around and wants to stay ahead, I see no big reason not to (unless it gets noisy).

**EDIT2: Actually... it will also be much easier to *ignore* if this doesn't block the other jobs, so I think it is important to restructure.**

charris:
> so I think it is important to restructure.

What do you have in mind?

seberg:
I mean that it would be good to move these new (and maybe some old ones) out of the `smoke-test` and into a different category.  That way we run fewer tests on failures, but mainly if 3.15 fails, it doesn't block all the other tests.

ngoldbaum:
I split off the 3.13, 3.14, and both 3.15 runs into a new job. Now the smoke tests only run on 3.12 and 3.14t under the theory that issues on 3.14t would also probably happen on 3.14.

seberg (approved):
Thanks Nathan! I think splitting it up like this makes sense. If anyone disagrees, I am happy to follow up.

PR #30742, opened 2026-01-29
Update dependabot.yml - add cooldown days parameter

clintonsteiner:
Add cooldown to guard against supply chain attacks

Dependency cooldowns are a free, easy, and incredibly effective way to mitigate the large majority of open source supply chain attacks. More individual projects should apply cooldowns (via tools like Dependabot and Renovate) to their dependencies, and packaging ecosystems should invest in first-class support for cooldowns directly in their package managers.

https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,8 @@ updates:
     directory: /
     schedule:
       interval: daily
+    cooldown:
+      default-days: 7
     commit-message:
       prefix: "MAINT"
     labels:
@@ -15,6 +17,8 @@ updates:
     directory: /requirements
     schedule:
       interval: daily
+    cooldown:
+      default-days: 7
     commit-message:
       prefix: "MAINT"
     labels:


charris:
Is it possible to override? For example, we like to release after new python version releases as soon as they are supported by cibuildwheel. Not that we couldn't wait for a few days, but it would be nice to know if things could be speeded up in some cases.

clintonsteiner:
Not possible to override but shouldn't be the worst thing in the world.
Does protect you from vulnerabilities as well, similar to the npm attack

seberg:
Yeah, I think we should probably just do this. I think we can manually do Python if need be, this is good for all the actions (and by now those are surprisingly many, I think).
I am a bit curious about the opposite of not getting an important fix quickly :).

andyfaff:
AFAICT this setting only applies to how quickly dependabot suggests an update *after* a release has been made. We can manually make them earlier if we want to.

charris:
Thanks @clintonsteiner .

PR #25359, opened 2023-12-09
RFC: Heap type migration

koubaa:
Demonstrate migration of one type to heap types.

Related to #18586 , as heap types are essential to eliminate global state and allow embedded numpy to be usable from multiple independent subinterpreters. These subinterpreters do not share the same module state and may not be prevented from running in parallel by the GIL.

Obviously - this PR does not remove all global state from numpy. Rather, it shows an example of how this might work. It is clear that in order to remove global state, the globally exported type structs must be removed from the public API, requiring a new API version. Migrating all types to the heap can cause quite a lot of churn, and so it may make sense to do it incrementally, one type at a time. But... if that is done, then there will be N API breakages, one for each type - which is quite a lot.

#25358 takes a small part of this PR, showing how the API can change without much churn. I propose that this is done for all types in one shot, and then after that heap type migration can occur one at a time, without any API changes along the way.

PR #30744, opened 2026-01-30
API: hide PyArray_Descr fields on internal builds

ngoldbaum:
Towards fixing #30704.

This makes the `PyArray_Descr` typedef an opaque struct. I then update NumPy to use field accessor macros instead of directly accessing any fields.

I also added a few macros that didn't yet exist:

* PyDataType_SET_FLAGS
* PyDataType_TYPE
* PyDataType_KIND
* PyDataType_BYTEORDER
* PyDataType_TYPEOBJ

Also new macro to get the fields, if they're needed:

* PyDataType_GET_TYPE_INFO

This is a pretty big diff but almost all the changes outside the headers are mechanical.

The biggest non-trivial bits are the set-up of the descriptor object's getset table, which is based off of `offsetof` on the descriptor type as well as a few other misc spots, see inline comments.


seberg:
Sorry, this became a bit silly detailed (most of which you would be doing as next steps anyway).  Either way, I think this isn't quite the right approach TBH.
That said of course we need all of these accessor inline functions, etc. and I don't mind adopting them internally (but we also don't have yet, IMO).

The main problem is of course that you can't use the version API switch at all, because that would disable support NumPy<2.4, and you don't want to wait 3 years to start using this (unless there really wasn't another way, but of course there is).

We clearly want to add this type of accessors. I would even be OK to only introduce `PyDataType_FIELDS()` and use that in most cases, but the other static inline functions are just one liners around that anyway, introducing them is good.

What I would suggest (and I think did briefly on the issue) is this:
1. Detect whether we are compiling with the new opaque API.  We can conditionally/also have an `NPY_OPAQUE_OBJECTS=1` define to force this.
   * Yes, it _would_ be nice to force users to transition if they compile with a 2.5 API or so, but I am not sure how to best pull that off.  (The main problem is that over the years at some point the 2.5 API will be the default, so there is a question how to tell users that they should transition or set `NPY_OPAQUE_OBJECTS=0`)
2. Do what you have here, introduce accessors, including the `_PyArray_Descr_fields` struct, however:
   * In a normal build, this structure *will* include the `PyObject_HEAD` and `PyArray_Descr` can just be an alias.
   * In an opaque API build it doesn't include the head and `PyArray_Descr` is fully opaque.
3. We introduce the static inline `PyArrayDescr_FIELDS()` which does nothing on a normal build but on an opaque build, it either:
   * Calls an API function, or looks up the actual runtime field offset from our API table (the big change here being that you can't access the fields without importing/initializing the API table in that compilation unit).
     The approach here really depends on what Python provides, it would be nice if Python would just provide some fast "this is a base type, so just skip PyObject_HEAD".
5. Yeah, maybe do introduce all of these helpers things that use `PyArrayDescr_FIELDS` internally.  `PyArrayDescr_FIELDS` should maybe not be API directly exposed to the user anyway.

Now, I would like to see a world where we ask users to transition to not use direct field access and I do think we could tie that to the 2.5 API!
But the mechanism for that would have to be the `NO_DEPRECATED_API`(?) version macro.  I.e. when _that_ is `>=2.5` (or the actual feature version is) we hide the fields from the main object.  When it is set to `<2.5` we give a warning at compile time.

(That said, I am not sure that we should start giving a warning quite yet. Downstream seems to tend to insist on compiling with old NumPy versions, meaning that they cannot adopt the new accessor functions immediately.)

ngoldbaum:
> Either way, I think this isn't quite the right approach TBH.

That's cool, thanks for the detailed comments! I'm going to try a different approach following your suggestions.

ngoldbaum:
@seberg can you give my latest commit a look? I think it implements your suggestions, although a couple notes:

> When it is set to <2.5 we give a warning at compile time.

I don't actually see to generate a compiler warning - how would I do that?

> We introduce the static inline PyArrayDescr_FIELDS()

There's already a field in the struct named `fields` and a corresponding `PyDataType_FIELDS`, so I called this function `PyDataType_FIELDS_STRUCT`. I think it has to be public because user dtypes probably need it, otherwise we would need to make setter macros for all the fields public.

numpy/_core/include/numpy/ndarraytypes.h
```diff
@@ -679,6 +683,21 @@ typedef struct {
...
+typedef PyArray_Descr_fields PyArray_Descr;
+#else
+/*
+ * Public version of the Descriptor struct
+ * To hide implementation details, we only expose
+ * the Python struct HEAD.
+ */
+typedef struct _PyArray_Descr {
+    PyObject_HEAD
+} PyArray_Descr;
```
seberg: I don't think there is a reason to even expose `PyObject_HEAD`? Maybe might as well make it fully opaque?
(I.e. we can't expose it when compiling with both ft and not probably?)

numpy/_core/include/numpy/npy_2_compat.h
```diff
@@ -142,22 +142,62 @@ PyArray_ImportNumPyAPI(void)
 #if NPY_FEATURE_VERSION >= NPY_2_0_API_VERSION || NPY_ABI_VERSION < 0x02000000
     /* Compiling for 1.x or 2.x only, direct field access is OK: */
 
+    static inline PyArray_Descr_fields *
+    PyDataType_FIELDS_STRUCT(const PyArray_Descr *dtype)
+    {
+        return (PyArray_Descr_fields *)dtype;
+    }
```
seberg: Hmmm, should also define this when compiling for 1.x and 2.x.  FWIW, I think this definition always works _except_ on an ABI compatible build.

(But I guess the idea is to add that code path in a follow-up PR).

numpy/_core/include/numpy/npy_2_compat.h
```diff
@@ -142,22 +142,62 @@ PyArray_ImportNumPyAPI(void)
...
+    PyDataType_SET_TYPEOBJ(PyArray_Descr *dtype, PyTypeObject *typeobj)
+    {
+        PyDataType_FIELDS_STRUCT(dtype)->typeobj = typeobj;
+    }
+
+    static inline void
+    PyDataType_SET_TYPENUM(PyArray_Descr *dtype, int type_num)
+    {
+        PyDataType_FIELDS_STRUCT(dtype)->type_num = type_num;
+    }
```
seberg: I am not sure why this needs to be in an `#ifdef` block.  That said, I am fine to keep it inside the `npy_2_compat.h` header so that if someone insists, they can copy paste the definitions over to start using this earlier.

seberg (commented):
Ah, true about the fields name clash. The idea for "fields" as a name, is probably just from `PyArrayObject_fields` and I think that name is fully private.

We can also borrow (a variant of?) the Python C-API name of "`*_GetItemData`", I think that may be better anyway.

If we just keep it public, we don't even need the new setters in principle (but yeah, user-dtypes would have to modify code again if we ever modify the struct. So having setters has advantages).

> I don't actually see to generate a compiler warning - how would I do that?

You have to write the message manually since right now we are not using this mechanism.  At the very end of `ndarraytypes.h` there seems to be a HOWTO for how it used to work.

We may want to discuss timing/need of this warning (maybe @mattip you have a thought?).
Right now downstream may not be able to fully avoid it, because they may need to compile against older NumPy versions for older Python versions at least (a header only numpy package would be sweet one day).

numpy/_core/include/numpy/ndarraytypes.h
```diff
@@ -447,9 +447,9 @@ typedef void (PyArray_VectorUnaryFunc)(void *, void *, npy_intp, void *,
  * is bumped. It used to be the separator.
  */
 typedef int (PyArray_ScanFunc)(FILE *fp, void *dptr,
-                               char *ignore, struct _PyArray_Descr *);
+                               char *ignore,  struct _PyArray_Descr *);
 typedef int (PyArray_FromStrFunc)(char *s, void *dptr, char **endptr,
-                                  struct _PyArray_Descr *);
+                                   struct _PyArray_Descr *);
```
mattip: Revert these formatting changes?

numpy/__init__.cython-30.pxd
```diff
@@ -226,19 +226,31 @@ cdef extern from "numpy/arrayobject.h":
...
+            return PyDataType_TYPEOBJ(self)
+
+        @property
+        cdef inline char kind(self) noexcept nogil:
+            return PyDataType_KIND(self)
+
+        @property
+        cdef inline char type(self) noexcept nogil:
+            return PyDataType_TYPE(self)
+
```
mattip, line 241: Yay for property accessors 😄. We should at some point remove the cython<3 `__init__.pxd` and rename this.

mattip (commented):
Nice. At some point I would like to revive #25359 to move to heap types, as a first step toward using the limited API internally. This change is only tangentially related, and I don't think it makes the move to heap types/limited API harder, maybe even slightly easier.

numpy/__init__.cython-30.pxd
ngoldbaum: I opened https://github.com/numpy/numpy/pull/30770 to start the ball rolling on that. IMO we shouldn't rename this file, otherwise downstream will see confusing errors.

numpy/_core/include/numpy/ndarraytypes.h
ngoldbaum: This works! Thanks for the suggestion.

numpy/_core/include/numpy/npy_2_compat.h
ngoldbaum: > (But I guess the idea is to add that code path in a follow-up PR).

Yup, I want to make all the structs opaque before I do that.

ngoldbaum:
I sent a message to the mailing list. I'd appreciate replies there indicating how to handle the deprecation, if any.

numpy/_core/include/numpy/npy_2_compat.h
```diff
@@ -166,7 +197,7 @@ PyArray_ImportNumPyAPI(void)
             }                                                  \
             return ((_PyArray_LegacyDescr *)dtype)->field;     \
         }
-#else  /* compiling for both 1.x and 2.x */
+#    else /* compiling for both 1.x and 2.x */
```
ngoldbaum, line 200:
```suggestion
#else /* compiling for both 1.x and 2.x */
```

Whitespace change my editor did without me asking

numpy/_core/src/multiarray/stringdtype/dtype.c
```diff
@@ -119,14 +119,15 @@ new_stringdtype_instance(PyObject *na_object, int coerce)
...
-    base->type = NPY_VSTRINGLTR;
+    PyArray_Descr_fields *fields = PyDataType_GET_ITEM_DATA(base);
+    fields->elsize = SIZEOF_NPY_PACKED_STATIC_STRING;
+    fields->alignment = ALIGNOF_NPY_PACKED_STATIC_STRING;
+    fields->flags |= NPY_NEEDS_INIT;
+    fields->flags |= NPY_LIST_PICKLE;
+    fields->flags |= NPY_ITEM_REFCOUNT;
+    fields->type_num = NPY_VSTRING;
+    fields->kind = NPY_VSTRINGLTR;
+    fields->type = NPY_VSTRINGLTR;
```
ngoldbaum, line 130: Rather than exposing more setters I think this is the pattern we should suggest for user dtypes. They'll always need to know what the layout of the descriptor fields are.

numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -4287,8 +4287,8 @@ PyArray_DescrFromType(int type)
         if (ret == NULL) {
             return NULL;
         }
-        ret->elsize = 1;
-        ret->type = NPY_CHARLTR;
+        PyDataType_SET_ELSIZE(ret, 1);
+        _PyDataType_SET_TYPE(ret, NPY_CHARLTR);
```
ngoldbaum, line 4291: I needed this private setter exactly once, right here. I could also use the fields struct 🤷

numpy/_core/src/multiarray/descriptor.c
```diff
@@ -2054,22 +2053,22 @@ arraydescr_dealloc(PyArray_Descr *self)
...
     {"itemsize",
-        T_PYSSIZET, offsetof(PyArray_Descr, elsize), READONLY, NULL},
+        T_PYSSIZET, offsetof(_PyArray_LegacyDescr, elsize), READONLY, NULL},
     {"alignment",
-        T_PYSSIZET, offsetof(PyArray_Descr, alignment), READONLY, NULL},
+        T_PYSSIZET, offsetof(_PyArray_LegacyDescr, alignment), READONLY, NULL},
     {"flags",
 #if NPY_SIZEOF_LONGLONG == 8
-        T_ULONGLONG, offsetof(PyArray_Descr, flags), READONLY, NULL},
+        T_ULONGLONG, offsetof(_PyArray_LegacyDescr, flags), READONLY, NULL},
```
ngoldbaum, line 2071: This works but is definitely a little weird. Not sure if there's a clearer way to spell this (and is `LegacyDescr` the right typedef to use?)

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -3831,7 +3831,7 @@ format_longfloat(PyObject *NPY_UNUSED(dummy), PyObject *args, PyObject *kwds)
  */
 static int _is_user_defined_string_array(PyArrayObject* array)
 {
-    if (NPY_DT_is_user_defined(PyArray_DESCR(array))) {
+    if (NPY_DT_is_user_defined(NPY_DTYPE(PyArray_DESCR(array)))) {
```
ngoldbaum, line 3834: this was a bug that just happened to work because the field names are the same on `PyArray_Descr` and `DTypeMeta`.

numpy/_core/src/multiarray/usertypes.c
```diff
@@ -179,7 +179,7 @@ PyArray_RegisterDataType(PyArray_DescrProto *descr_proto)
         return -1;
     }
     descr_proto->type_num = -1;
-    if (PyDataType_ISUNSIZED(descr_proto)) {
+    if (descr_proto->elsize == 0) {
```
ngoldbaum, line 182: this is the same as what `PyDataType_ISUNSIZED` used to do. We can't use that function anymore because of the type information in the new static inline functions.

numpy/_core/src/umath/_scaled_float_dtype.c
```diff
@@ -139,14 +139,15 @@ NPY_DType_Slots sfloat_slots = {
...
-    .scaling = 1,
+static PyArray_SFloatDescr SFloatSingleton = {
+        .base={
+                .byteorder = '|',  /* do not bother with byte-swapping... */
+                .flags = NPY_USE_GETITEM|NPY_USE_SETITEM,
+                .type_num = -1,
+                .elsize = sizeof(double),
+                .alignment = NPY_ALIGNOF(double),
+        },
+    .scaling = 1
```
ngoldbaum, line 150: I think I edited this because of a design we discarded and I can safely revert this change.

numpy/_core/src/umath/string_ufuncs.cpp
```diff
@@ -762,7 +762,8 @@ string_addition_resolve_descriptors(
         Py_DECREF(loop_descrs[1]);
         return _NPY_ERROR_OCCURRED_IN_CAST;
     }
-    loop_descrs[2]->elsize += loop_descrs[1]->elsize;
+    npy_intp elsize_2 = PyDataType_ELSIZE(loop_descrs[2]);
+    PyDataType_SET_ELSIZE(loop_descrs[2], elsize_2 + PyDataType_ELSIZE(loop_descrs[1]));
```
ngoldbaum, line 766: This in-place addition is a little gross with the macros 🤷

numpy/_core/include/numpy/ndarraytypes.h
```diff
@@ -608,11 +608,19 @@ typedef struct {
                                 NPY_ITEM_IS_POINTER | NPY_ITEM_REFCOUNT | \
                                 NPY_NEEDS_INIT | NPY_NEEDS_PYAPI)
 
-#if NPY_FEATURE_VERSION >= NPY_2_0_API_VERSION
 /*
- * Public version of the Descriptor struct as of 2.x
```
ngoldbaum, line 613: should restore this comment inside the #if

numpy/__init__.cython-30.pxd
```diff
@@ -226,19 +226,31 @@ cdef extern from "numpy/arrayobject.h":
     ctypedef struct PyArray_Descr:
         pass
 
-    ctypedef class numpy.dtype [object PyArray_Descr, check_size ignore]:
-        # Use PyDataType_* macros when possible, however there are no macros
-        # for accessing some of the fields, so some are defined.
-        cdef PyTypeObject* typeobj
-        cdef char kind
-        cdef char type
+    ctypedef class numpy.dtype [object PyArray_Descr_fields, check_size ignore]:
```
da-woods, line 229: I don't know if you need them for this particular use-case (I think you possibly do) but I implemented property setters in https://github.com/cython/cython/pull/7505

That'll likely need Cython 3.3

(It currently isn't part of my big freethreading-limited-api-preview branch. It probably could be though).

ngoldbaum:
This is out-of-date and Kumar took this a different direction. Closing.

PR #30746, opened 2026-01-30
DOC: fix missing numpy import in emath documentation example

agnuspaul98:
Adds the missing NumPy import to the numpy.emath documentation example.
Without this import, the example raises a NameError when executed.


mattip:
Thanks @agnuspaul98

PR #30751, opened 2026-01-31
Fix ctypes conclusion list indentation in docs

### LINKED ISSUE #30750: DOC: Incorrect indentation in the documentation

In the [Using Python as glue](https://numpy.org/doc/1.25/user/c-info.python-as-glue.html#id4) documentation page, some bullet points in the ctypes/Conclusion section are not rendering correctly.

This appears to be caused by incorrect indentation in the source RST file.

Screenshot: 

Image

_No response_

sabasiddique1:
Fix #30750 
## Summary
- fix incorrect indentation in the ctypes Conclusion list
- ensure the list renders as a single flat bullet list

### DIFF
--- a/doc/source/user/c-info.python-as-glue.rst
+++ b/doc/source/user/c-info.python-as-glue.rst
@@ -825,19 +825,13 @@ Conclusion
 Using ctypes is a powerful way to connect Python with arbitrary
 C-code. Its advantages for extending Python include
 
-- clean separation of C code from Python code
-
-    - no need to learn a new syntax except Python and C
-
-    - allows reuse of C code
-
-    - functionality in shared libraries written for other purposes can be
-      obtained with a simple Python wrapper and search for the library.
-
-
-- easy integration with NumPy through the ctypes attribute
-
-- full argument checking with the ndpointer class factory
+* clean separation of C code from Python code
+* no need to learn a new syntax except Python and C
+* allows reuse of C code
+* functionality in shared libraries written for other purposes can be
+  obtained with a simple Python wrapper and search for the library.
+* easy integration with NumPy through the ctypes attribute
+* full argument checking with the ndpointer class factory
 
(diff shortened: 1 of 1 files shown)


sabasiddique1:
Screenshot 2026-01-31 at 4 59 13 PM

Verified locally: the ctypes Conclusion list now renders as a single flat bullet list (see screenshot).
Please let me know if you prefer a different RST style or formatting convention for small doc fixes.

jorenham:
Could you rebase your branch onto upstream `main`? It's 608 commits behind

doc/source/user/c-info.python-as-glue.rst
```diff
@@ -758,17 +758,11 @@ Using ctypes is a powerful way to connect Python with arbitrary
...
-    - allows reuse of C code
-
-    - functionality in shared libraries written for other purposes can be
-      obtained with a simple Python wrapper and search for the library.
-
-
+- no need to learn a new syntax except Python and C
+- allows reuse of C code
+- functionality in shared libraries written for other purposes can be
+  obtained with a simple Python wrapper and search for the library.
```
melissawm: I think the original was a nested list, so the correct solution is the following:


```suggestion
- no need to learn a new syntax except Python and C
  - allows reuse of C code
  - functionality in shared libraries written for other purposes can be
    obtained with a simple Python wrapper and search for the library.

```

melissawm (dismissed):

melissawm:
I'm sorry, I think I was distracted and misled you here. The rst syntax can be tricky, in this case using `-` for the list items does not work. See https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#lists-and-quote-like-blocks

So we need to exchange the `-` for asterisks in all items of this list. Sorry again! This will fix the build issue.

mattip:
Thanks @sabasiddique1

PR #30754, opened 2026-01-31
TYP: Fix class annotations using names shadowed by a class member

### LINKED ISSUE #30752: TYP: types in a class body that use names shadowed by a class method confuse ty
### Describe the issue:

Better explain with an example

```py
# file test.py
class dtype: ...
class ndarray:
    def method(self) -> dtype: ...  # defined before the dtype property
    @property
    def dtype(self): ...
    def method2(self) -> dtype: ...  # defined after the dtype property

print(ndarray.method.__annotations__)
print(ndarray.method2.__annotations__)
```

```console
$ uvx --isolated --python=3.13 python test.py
{'return': <class '__main__.dtype'>}
{'return': <property object at 0x7e4a7e223510>}
```

```console
$ uvx --isolated --python=3.14 python test.py
{'return': <property object at 0x787a8e5a4a90>}
{'return': <property object at 0x787a8e5a4a90>}
```

So starting with Python 3.14, the `dtype` annotation of methods defined anywhere in the class body will point to `ndarray.dtype` property instead of the `dtype` class. While mypy models the <3.14 behavior, ty chose to model the >=3.14 behavior.

I'll prepare a PR to fix these annotations so that work with all type checkers and across Python versions.

```python
import numpy as np

a = np.zeros(3).tolist()
reveal_type(a)
```

### Error message:

```shell
ty: Revealed type: `list[Unknown]`
```

2.4.1
3.13.11 (main, Dec  6 2025, 08:52:51) [GCC 11.4.0]

ty: 0.0.14

_No response_

hamdanal:
Fixes #30752


numpy/lib/_nanfunctions_impl.py
```diff
@@ -1603,9 +1603,9 @@ def _nanquantile_unchecked(
 
 
 def _nanquantile_ureduce_func(
-        a: np.array,
-        q: np.array,
-        weights: np.array,
+        a: np.ndarray,
+        q: np.ndarray,
+        weights: np.ndarray,
```
hamdanal, line 1608: Drive-by fix. This was the only remaining `invalid-type-form` error reported by ty after all other changes in this PR

jorenham: nice catch, but kinda weird that mypy didn't pick up on this

jorenham (commented):
Looks good, thanks.

But could you add a quick inline comment next to the `py_*` imports so that we don't remove it again in the future?

numpy/lib/_nanfunctions_impl.py
hamdanal: Yeah looks weird indeed. I'll probably try to minimize a reproducer and report it to mypy.

jorenham:
CI is failing because the pyright type-completeness dropped for some reason. So there is something in this PR that causes pyright to not be able to completely determine the type of a commonly used thing, such as dtype or ndarray.

hamdanal:
> CI is failing because the pyright type-completeness dropped for some reason. So there is something in this PR that causes pyright to not be able to completely determine the type of a commonly used thing, such as dtype or ndarray.

I am not familiar with pyright's type completeness. Do you know if it can tell what symbols are the root cause of the problem?

jorenham:
> > CI is failing because the pyright type-completeness dropped for some reason. So there is something in this PR that causes pyright to not be able to completely determine the type of a commonly used thing, such as dtype or ndarray.
> 
> I am not familiar with pyright's type completeness. Do you know if it can tell what symbols are the root cause of the problem?

I took a look, and it turns out that it's caused by the lack of generic type arguments passed to `_dtype` in `__init__.pyi`. Those should instead be `_dtype[Any]` or `_dtype[generic]`

hamdanal:
Made `_dtype` an actual alias of `dtype` so that they have the same type parameters with proper defaults.

jorenham (approved):

jorenham:
Thanks @hamdanal

PR #30756, opened 2026-01-31
BUG: Fix some leaks found via LeakSanitizer

mayeut:
The ref leak in `PyUFunc_AddLoopFromSpec_int` was the one reported when running `python -c 'import numpy'` with AddressSanitizer+LeakSanitizer enabled.
The other ones were found by searching for the same pattern.


seberg (approved):
Thanks, LGTM. (if anyone is curious, I always look at the valgrind output only for actual tests, which means init code isn't necessarily covered).

Issue #15859, opened 2020-03-28
Documenting single-character type codes


adeak:
I could find at least five places where single-character type codes appear, and some of these don't correspond to the same total collection of types. In other words some types are missing in some places, and I can't tell how much of these differences is on purpose.

  1. Where I originally started from: type coercion tables from `from numpy.testing import print_coercion_tables`
  2. dtypes reference https://docs.scipy.org/doc/numpy/reference/arrays.dtypes.html
  3. scalar types reference https://docs.scipy.org/doc/numpy/reference/arrays.scalars.html#built-in-scalar-types
  4. `np.typename` https://docs.scipy.org/doc/numpy/reference/generated/numpy.typename.html
  5. `np.typecodes` (a dict)

I'm aware that 1. is not documentation, and not exactly public. I know from https://github.com/numpy/numpy/issues/7869#issuecomment-234872466 that 2. and 3. on my list don't have to agree, though I don't understand this in detail. 4. seems documented and references 5., so I guess those are public (although I only bumped into `np.typename` by accident).

The dtype codes referenced by each of the above (as of 1.17.4 on linux), respectively:
  1. `set('? b h i l q p B H I L Q P e f d g F D G S U V O M m'.split())` (copied from the output of the import)
  2. `set('Hd')` (examples for "one-character strings"), `set('?bBiufcmMOSaUV')` (comprehensive-looking "array-protocol type strings") 
  3. `set('?bhilqpBHILQPefdgFDGO')` along with the flexible types `'S#'`, `'U#'`, `'V#'`
  4. `set(np.typename.__globals__['_namefromtype'])`, same as `set('S1 ? b B h H i I l L q Q f d g F D G S U V O'.split())`
  5. `np.typecodes['All']` i.e. `'?bhilqpBHILQPefdgFDGSUVOMm'`

Observations:
  * Among these 1. and 5. are the most comprehensive and they contain the same 26 codes. 
  * Set 3. is missing `{'m', 'M'}`, i.e. `np.typecodes['Datetime']`
  * Set 4. is further missing `{'P', 'e', 'p'}`: I believe these are intp/uintp and half-precision. It also contains `S1` but "character" is not really a separate type, so it's OK for this to be missing from the other codes. I understand if intp/uintp don't have an item in `np.typename`, because these might be platform-specific and they are (always?) aliases to other concrete integer types.
[...]

PR #30758, opened 2026-01-31
DOC: Cleanup type documentation

Noxaster:
**Motivation**
`np.typename` was relying on a hardcoded dictionary (`_namefromtype`), which is hard to maintain and outdated. It failed for newer types like datetime (`M`, `m`) and returned the wrong name for bytes (`S` was mapped to 'string').

Also, removed the incorrect 'c' entry from the array-protocol docs, which was confusingly listed as a complex type (see #15859).

**Changes**
- **Refactor**: Rewrote `np.typename(char)` to simply delegate to `np.dtype(char).name`. This creates a single source of truth and handles all valid NumPy types automatically.
- **Docs**: Fixed `arrays.dtypes.rst`.
- **Tests**: Added `TestTypename` in `numpy/lib/tests/test_type_check.py` covering:
    - Basic types (`f`, `i`, etc.)
    - Datetime types (`M`, `m`) - *Fixes regression*
    - Bytes (`S`) - *Fixes incorrect return value*
    - Invalid codes.

**Related to**
See #15859

### DIFF
--- a/doc/source/reference/arrays.dtypes.rst
+++ b/doc/source/reference/arrays.dtypes.rst
@@ -258,23 +258,23 @@ Array-protocol type strings (see :ref:`arrays.interface`)
    which represents boolean.  The item size must correspond to an existing
    type, or an error will be raised.  The supported kinds are
 
-   =================  ========================
-   ``'?'``, ``'b1'``  boolean
-   ``'b'``            (signed) byte
-   ``'B'``            unsigned byte
-   ``'i'``            (signed) integer
-   ``'I'``            unsigned integer
-   ``'f'``            floating-point
-   ``'m'``            timedelta
-   ``'F'``, ``'c8'``  complex single precision
-   ``'D'``, ``'c16'`` complex double precision
-   ``'G'``, ``'c32'`` complex long double precision
-   ``'M'``            datetime
-   ``'O'``            (Python) objects
-   ``'S'``            zero-terminated bytes (not recommended)
-   ``'U'``            Unicode string
-   ``'V'``            raw data (:class:`void`)
-   =================  ========================
+   ==================  ========================
+   ``'?'``, ``'b1'``   boolean
+   ``'b'``             (signed) byte
+   ``'B'``             unsigned byte
+   ``'i'``             (signed) integer
+   ``'I'``             unsigned integer
+   ``'f'``             floating-point
+   ``'m'``             timedelta
+   ``'F'``, ``'c8'``   complex single precision
+   ``'D'``, ``'c16'``  complex double precision
+   ``'G'``, ``'c32'``  complex long double precision
+   ``'M'``             datetime
+   ``'O'``             (Python) objects
+   ``'S'``             zero-terminated bytes (not recommended)
+   ``'U'``             Unicode string
+   ``'V'``             raw data (:class:`void`)
+   ==================  ========================
 
    .. admonition:: Example
 
--- a/doc/source/reference/arrays.dtypes.rst
+++ b/doc/source/reference/arrays.dtypes.rst
@@ -276,8 +276,6 @@ Array-protocol type strings (see :ref:`arrays.interface`)
    ``'D'``             complex double precision
    ``'g'``             long precision
    ``'G'``             complex long double precision
-   ``'m'``             timedelta
-   ``'M'``             datetime
    ``'O'``             (Python) objects
    ``'S'``             zero-terminated bytes (not recommended)
    ``'U'``             Unicode string
--- a/doc/source/reference/arrays.dtypes.rst
+++ b/doc/source/reference/arrays.dtypes.rst
@@ -263,15 +263,15 @@ Array-protocol type strings (see :ref:`arrays.interface`)
    ``'b'``            (signed) byte
    ``'B'``            unsigned byte
    ``'i'``            (signed) integer
-   ``'u'``            unsigned integer
+   ``'I'``            unsigned integer
    ``'f'``            floating-point
    ``'m'``            timedelta
-   ``'F'``            complex single precision
-   ``'D'``            complex double precision
-   ``'G'``            complex long double precision
+   ``'F'``, ``'c8'``  complex single precision
+   ``'D'``, ``'c16'`` complex double precision
+   ``'G'``, ``'c32'`` complex long double precision
    ``'M'``            datetime
    ``'O'``            (Python) objects
-   ``'S'``, ``'a'``   zero-terminated bytes (not recommended)
+   ``'S'``            zero-terminated bytes (not recommended)
    ``'U'``            Unicode string
    ``'V'``            raw data (:class:`void`)
    =================  ========================
--- a/doc/source/reference/arrays.dtypes.rst
+++ b/doc/source/reference/arrays.dtypes.rst
@@ -265,8 +265,10 @@ Array-protocol type strings (see :ref:`arrays.interface`)
    ``'i'``            (signed) integer
    ``'u'``            unsigned integer
    ``'f'``            floating-point
-   ``'c'``            complex-floating point
    ``'m'``            timedelta
+   ``'F'``            complex single precision
+   ``'D'``            complex double precision
+   ``'G'``            complex long double precision
    ``'M'``            datetime
    ``'O'``            (Python) objects
    ``'S'``, ``'a'``   zero-terminated bytes (not recommended)
--- a/doc/source/reference/arrays.dtypes.rst
+++ b/doc/source/reference/arrays.dtypes.rst
@@ -259,16 +259,24 @@ Array-protocol type strings (see :ref:`arrays.interface`)
    type, or an error will be raised.  The supported kinds are
 
    ==================  ========================
-   ``'?'``, ``'b1'``   boolean
+   ``'?'``             boolean
    ``'b'``             (signed) byte
    ``'B'``             unsigned byte
+   ``'h'``             (signed) short
+   ``'H'``             unsigned short
    ``'i'``             (signed) integer
    ``'I'``             unsigned integer
-   ``'f'``             floating-point
+   ``'l'``             (signed) long integer
+   ``'L'``             unsigned long integer
+   ``'q'``             (signed) long long integer
+   ``'Q'``             unsigned long long integer
+   ``'f'``             single precision
+   ``'F'``             complex single precision
+   ``'d'``             double precision
+   ``'D'``             complex double precision
+   ``'g'``             long precision
+   ``'G'``             complex long double precision
    ``'m'``             timedelta
-   ``'F'``, ``'c8'``   complex single precision
-   ``'D'``, ``'c16'``  complex double precision
-   ``'G'``, ``'c32'``  complex long double precision
    ``'M'``             datetime
    ``'O'``             (Python) objects
    ``'S'``             zero-terminated bytes (not recommended)
--- a/numpy/lib/_type_check_impl.py
+++ b/numpy/lib/_type_check_impl.py
@@ -588,7 +588,7 @@ def typename(char):
     try:
         return dtype(char).name
     except TypeError as e:
-        raise TypeError(f"{char} does not represent a valid dtype") from None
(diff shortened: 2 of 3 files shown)


numpy/lib/_type_check_impl.py
```diff
@@ -604,36 +580,15 @@ def typename(char):
...
+    >>> type(np.typename('S'))
+    <class 'str'>
+    >>> np.typename('S')
+    'bytes'
     """
-    return _namefromtype[char]
+    try:
+        return dtype(char).name
+    except TypeError as e:
+        raise TypeError(f"{char} does not represent a valid dtype") from None
```
jorenham:
```suggestion
        raise TypeError(f"{char!r} does not represent a valid dtype") from None
```

some quotes might be nice

numpy/lib/_type_check_impl.py
```diff
@@ -604,36 +580,15 @@ def typename(char):
...
-    q  :  long long integer
-
+    >>> type(np.typename('S'))
+    <class 'str'>
+    >>> np.typename('S')
+    'bytes'
     """
-    return _namefromtype[char]
+    try:
+        return dtype(char).name
```
jorenham: This is backwards incompatible:

```py
>>> import numpy as np
>>> np.typename("F")
'complex single precision'
>>> np.dtype("F").name
'complex64'
```

numpy/lib/_type_check_impl.py
Noxaster: Thanks for your suggestion!

numpy/lib/_type_check_impl.py
Noxaster: Make sense. I've compared `.dtype(...).name` against the old `_namefromtype` dictionary. The changes generally fall into three categories:
- Same meaning but different spelling (backward incompatible): `h`, `H`, `i`, `I`, `l`, `L`, `q`, `Q`, `f`, `d`, `g`, `F`, `D`, `G` (Mostly C-style/Verbose -> Bit-width/Standard)
- Different meaning (or significant naming shift): `S1`, `b`, `B`, `S`, `U`
- Same spelling: `?`, `V`, `O`

I plan to restore the legacy mapping for `F`, `D`, `G` (complex types) and `S1` (`'character'` -> `'bytes8'`) to preserve backward compatibility.

However, for `b` (`'signed char'` -> `'int8'`) and `B` (`'unsigned char'` -> `'uint8'`), I feel the new names are significant improvements and less confusing in data type identification.

And one question regarding the rest of the first category (e.g., `h`, `i`, `f`): Do you prefer strict backward compatibility (keeping C-style names like 'short', 'float'), or is adopting standardized bit-width names ('int16', 'float32') acceptable here?

jorenham: Sure, that could work.

numpy/lib/_type_check_impl.py
```diff
@@ -604,36 +584,26 @@ def typename(char):
...
-    d  :  double precision
-    g  :  long precision
-    f  :  single precision
-    i  :  integer
-    h  :  short
-    l  :  long integer
-    q  :  long long integer
-
+    >>> np.typename('S')
+    'bytes'
```
jorenham: Should be `'string'`

numpy/lib/_type_check_impl.py
```diff
@@ -604,36 +584,26 @@ def typename(char):
...
+    'complex single precision'
     """
-    return _namefromtype[char]
+    # Preserve backward compatibility
+    _legacy_mapping = {
+            'F': 'complex single precision',
+            'D': 'complex double precision',
+            'G': 'complex long double precision',
+            'S1': 'character',
+    }
```
jorenham: You're missing a bunch of others. Almost all of the previous ones in the `_namefromtype` dict actually.

So I recommend reverting all code changes, and limiting this PR to the doc fixes

jorenham (changes requested):

numpy/lib/tests/test_type_check.py
jorenham, line 1: this is unrelated now

Noxaster: Thanks for catching that!

doc/source/reference/arrays.dtypes.rst
```diff
@@ -258,21 +258,23 @@ Array-protocol type strings (see :ref:`arrays.interface`)
...
-   ``'U'``            Unicode string
-   ``'V'``            raw data (:class:`void`)
-   =================  ========================
+   ==================  ========================
+   ``'?'``, ``'b1'``   boolean
+   ``'b'``             (signed) byte
+   ``'B'``             unsigned byte
+   ``'i'``             (signed) integer
+   ``'I'``             unsigned integer
+   ``'f'``             floating-point
```
jorenham: you added complex double and long double, but forgot about the regular double (`d`) and long double (`g`).


```pycon
>>> np.typename("D")
'complex double precision'
>>> np.typename("d")
'double precision'
>>> np.typename("g")
'long precision'
```

doc/source/reference/arrays.dtypes.rst
```diff
@@ -258,21 +258,23 @@ Array-protocol type strings (see :ref:`arrays.interface`)
...
-   ``'c'``            complex-floating point
-   ``'m'``            timedelta
-   ``'M'``            datetime
-   ``'O'``            (Python) objects
-   ``'S'``, ``'a'``   zero-terminated bytes (not recommended)
-   ``'U'``            Unicode string
-   ``'V'``            raw data (:class:`void`)
-   =================  ========================
+   ==================  ========================
+   ``'?'``, ``'b1'``   boolean
```
jorenham: `b1` doesn't work

```pycon
>>> np.typename("b1")
    np.typename("b1")
    ~~~~~~~~~~~^^^^^^
  File "/home/joren/.cache/uv/archive-v0/jTWoQtjQfzJawFTGIbCll/lib/python3.14/site-packages/numpy/lib/_type_check_impl.py", line 636, in typename
    return _namefromtype[char]
           ~~~~~~~~~~~~~^^^^^^
KeyError: 'b1'
```

```suggestion
   ``'?'``             boolean
```

doc/source/reference/arrays.dtypes.rst
```diff
@@ -258,21 +258,23 @@ Array-protocol type strings (see :ref:`arrays.interface`)
...
+   ``'b'``             (signed) byte
+   ``'B'``             unsigned byte
```
jorenham, line 264: We might as well add the other integer types (short, long, long long) while we're at it.

doc/source/reference/arrays.dtypes.rst
```diff
@@ -258,21 +258,23 @@ Array-protocol type strings (see :ref:`arrays.interface`)
...
+   ``'i'``             (signed) integer
+   ``'I'``             unsigned integer
+   ``'f'``             floating-point
+   ``'m'``             timedelta
```
jorenham: why did you move this up? it should be next to `M`

doc/source/reference/arrays.dtypes.rst
```diff
@@ -258,21 +258,23 @@ Array-protocol type strings (see :ref:`arrays.interface`)
...
+   ``'F'``, ``'c8'``   complex single precision
+   ``'D'``, ``'c16'``  complex double precision
+   ``'G'``, ``'c32'``  complex long double precision
```
jorenham: `c8`, `c16` and `c32` don't work...

```
>>> np.typename("c8")
    np.typename("c8")
    ~~~~~~~~~~~^^^^^^
  File "/home/joren/.cache/uv/archive-v0/jTWoQtjQfzJawFTGIbCll/lib/python3.14/site-packages/numpy/lib/_type_check_impl.py", line 636, in typename
    return _namefromtype[char]
           ~~~~~~~~~~~~~^^^^^^
KeyError: 'c8'
>>> np.typename("c16")
    np.typename("c16")
    ~~~~~~~~~~~^^^^^^^
  File "/home/joren/.cache/uv/archive-v0/jTWoQtjQfzJawFTGIbCll/lib/python3.14/site-packages/numpy/lib/_type_check_impl.py", line 636, in typename
    return _namefromtype[char]
           ~~~~~~~~~~~~~^^^^^^
KeyError: 'c16'
>>> np.typename("c32")
    np.typename("c32")
    ~~~~~~~~~~~^^^^^^^
  File "/home/joren/.cache/uv/archive-v0/jTWoQtjQfzJawFTGIbCll/lib/python3.14/site-packages/numpy/lib/_type_check_impl.py", line 636, in typename
    return _namefromtype[char]
           ~~~~~~~~~~~~~^^^^^^
KeyError: 'c32'
```

doc/source/reference/arrays.dtypes.rst
```diff
@@ -258,21 +258,23 @@ Array-protocol type strings (see :ref:`arrays.interface`)
...
-   ``'S'``, ``'a'``   zero-terminated bytes (not recommended)
-   ``'U'``            Unicode string
-   ``'V'``            raw data (:class:`void`)
-   =================  ========================
+   ==================  ========================
+   ``'?'``, ``'b1'``   boolean
+   ``'b'``             (signed) byte
+   ``'B'``             unsigned byte
+   ``'i'``             (signed) integer
+   ``'I'``             unsigned integer
```
jorenham, line 268: This is wrong now... Why did you change `u` to `I`?

jorenham (changes requested):
These documentation changes are for a large part incorrect and incomplete. This is becoming rather frustrating and disproportionately time-consuming to review... 

Please be mindful of the reviewer's time in the future, and take critically look at the code before you push your changes.

doc/source/reference/arrays.dtypes.rst
Noxaster: It turns out `u` causes a KeyError in the legacy dictionary, while `I` is the correct legacy code for 'unsigned integer'.
```bash
>>> np.typename('u')
  File "/home/noxaster/numpy/build-install/usr/lib/python3/dist-packages/numpy/lib/_type_check_impl.py", line 636, in typename
    return _namefromtype[char]
           ~~~~~~~~~~~~~^^^^^^
KeyError: 'u'
```
```bash
>>> np.typename('I')
'unsigned integer'
```
So I used `I` to ensure the table is technically accurate regarding typename support.

Noxaster:
Apologies for the back-and-forth. I didn't realize strict adherence to `np.typename` was the priority here.
I've updated the PR to:
- Dropped unsupported aliases: Removed `b1`, `c8`, `c16`, etc.
- Added missing legacy types: Included `d`, `g`, and the extra integer types (`h`/`H`, `l`/`L`, `q`/`Q`) as you suggested.
- Moved `m` next to `M`.
- Swapped `u` -> `I` (since `u` isn't supported by typename).

doc/source/reference/arrays.dtypes.rst
jorenham: Oh yea you're right. Sorry about the noise.

doc/source/reference/arrays.dtypes.rst
```diff
@@ -258,21 +258,31 @@ Array-protocol type strings (see :ref:`arrays.interface`)
...
+   ``'q'``             (signed) long long integer
+   ``'Q'``             unsigned long long integer
+   ``'f'``             single precision
+   ``'F'``             complex single precision
+   ``'d'``             double precision
+   ``'D'``             complex double precision
+   ``'g'``             long precision
+   ``'G'``             complex long double precision
+   ``'m'``             timedelta
+   ``'M'``             datetime
```
jorenham: As it turns out, both `np.typename("m")` and `np.typename("M")` will raise a `KeyError`. So lets's remove it here.

```suggestion
```

jorenham (commented):
This function is a big mess haha. Maybe we should just deprecate it (not in this PR or something). I left one small suggestion. Apart from that, it looks fine now.

doc/source/reference/arrays.dtypes.rst
Noxaster: Applied. Thanks!

jorenham (approved):

jorenham:
thanks @Noxaster

PR #30763, opened 2026-02-01
MAINT: update scipy-openblas to one with risc-V

mattip:
This version of scipy-openblas has risc-V wheels. Towards verfying NumPy on riscv64 (see #30216). `numpy-release` has already been updated.


charris:
Thanks Matti.

PR #30764, opened 2026-02-01
MAINT: drop pypy

### LINKED ISSUE #30416: MAINT: drop support for PyPy

mattip:
Drop PyPy. Closes #30416

I searched for all PyPy, PYPY, and pypy. Perhaps there are accommodations I missed, like calls to `gc.collect()`, in tests, that were not marked as needed for PyPy.

Hopefully putting this in one PR will make it possible to revert if PyPy development restarts.


jorenham:
Hmm, apparently scipy relies on `numpy.testing.IS_PYPY` in their tests. Maybe we should keep that one around for a while longer?

mattip:
I restored numpy.testing.IS_PYPY, and submitted scipy/scipy#24491 to detach `IS_PYPY`. We may want to wait for a SciPy release (or two?) to actually remove `IS_PYPY` here.

seberg (approved):
Thanks Matti, LGTM and seems complete. There is nothing else here, right? (I.e. this is ready to just merge? Then go ahead, or I'll do it).

mattip:
Yes, this is ready to merge from my perspective

jorenham:
Thanks Matti :)

rgommers:
> We may want to wait for a SciPy release (or two?) to actually remove `IS_PYPY` here.

That's the absolute minimum, but more importantly: this should now be considered public API, and `break_cycles` as well. The problem is `from ._private.utils import *` in `numpy/testing/__init__.py`, which leaks things that I don't think we intended to leak.

PR #30770, opened 2026-02-02
MAINT: disable Cython bindings for Cython < 3.0.0.

ngoldbaum:
According to [the cython docs](https://cython.readthedocs.io/en/latest/src/userguide/sharing_declarations.html#versioning), if there is a `__init__cython-30.pxd`, Cython > 3.0 will ignore a `__init__.pxd`. 

In an attempt to give downstream users a useful error message, instead of just deleting `__init__.pxd` or renaming `__init__cython-30.pxd`, I left behind `__init__.pxd` but made it generate a build error if it is ever used.

I also updated the section on using the C API that covers Cython. The old content still talked about distutils, so I took the opportunity to update it to talk about meson.


ngoldbaum:
@da-woods I'm curious about your opinion on this. Is there a better way to drop support for building against the NumPy headers and older Cython versions that you might use instead of this?

numpy/__init__.pxd
```diff
@@ -1,10 +1,18 @@
...
+#
+# Only Cython versions older than 3.0.0 will attempt to process this
+# header. It is left unchanged from its state when we dropped support
+# for Cython < 3.0 but using it will generate a build error when
+# compiling the generated C file due to the error in the C header
+# included below.
+#
+
+cdef extern from "numpy/_fail_builds_with_old_cython.h":
+    pass
```
da-woods:
```suggestion
cdef extern from *:
    """
    #error "The Cython NumPy pxd wrappers require Cython 3.0.0 or newer"
    """
```

You can do this and then don't have to add an extra file.

da-woods:
> Is there a better way to drop support for building against the NumPy headers and older Cython versions that you might use instead of this?

I think this is fine. But if I did it then I'd probably delete all of `__init__.pxd` except for the error. That'll hopefully avoid any possibility of people thinking that it's the real file my mistake. But that's just because I think it's clearer - the basic idea is fine I think.

---------

The other thing you could do is to use `CYTHON_HEX_VERSION`. So just have one `__init__.pxd` but inject something like

```
#if !defined(CYTHON_HEX_VERSION) || CYTHON_HEX_VERSION < 0x03000000
#error "The Cython NumPy pxd wrappers require Cython 3.0.0 or newer"
#endif
```

With that said, I don't think Cython 0.29 would successfully compile the 3.0 pxd file so you might not even get far enough for people to see the error message.  So it probably won't work for this case.

ngoldbaum:
> But if I did it then I'd probably delete all of __init__.pxd except for the error. That'll hopefully avoid any possibility of people thinking that it's the real file my mistake. But that's just because I think it's clearer - the basic idea is fine I think.

I tried this, but on a nontrivial project I end up seeing Cython compilation errors due to missing definitions in the cython headers. So to make compilation proceed to the point where the `#error` macro gets processed by a C compiler, I need to make sure all the cython compilation succeeds. That means keeping the content of `__init__.pxd` the same except for the new include that triggers the C compilation error.

I was hoping there's a way to do this in pure Cython so I can trigger this error in the Cython compilation step, but it sounds like there isn't.

da-woods:
You could do something like

```
DEF err = int('Cython minimum version is 3.0')
```

That immediately generates an error (i.e. it essentially just does `eval` on the right-hand side of the expression).

```
ddd.pyx:1:13: Error in compile-time expression: ValueError: invalid literal for int() with base 10: 'Cython minimum version is 3.0'
```

I couldn't quickly think of a version that generates a much nicer error message. But it at least moves the error message earlier.

ngoldbaum:
Thanks! I also opened https://github.com/cython/cython/issues/7496 to track this feature upstream, IMO something like it would be nice to have in the future.

doc/release/upcoming_changes/30770.compatibility.rst
```diff
@@ -0,0 +1,28 @@
+Cython support
+--------------
+
+NumPy's Cython headers (accessed via ``cimport numpy``) now require
+Cython 3.0 or newer to build. If you try to a project that depends on
```
ngoldbaum: "try to a"

doc/source/user/c-info.python-as-glue.rst
```diff
@@ -159,30 +159,63 @@ work with multidimensional arrays.
...
+  project(
+     'my_filter',
+     'c', 'cython',
+     version: '0.0.1',
+     license: 'MIT',
+  )
+
+  cython = find_program('cython')
+  py = import('python').find_installation(pure: false)
+  py_dep = py.dependency()
```
rgommers: You don't need this line

doc/source/user/c-info.python-as-glue.rst
```diff
@@ -159,30 +159,63 @@ work with multidimensional arrays.
...
+
+  numpy_nodepr_api = ['-DNPY_NO_DEPRECATED_API=NPY_2_0_API_VERSION']
+
+  np_dep = declare_dependency(dependencies: dependency('numpy'),
+                              compile_args: numpy_nodepr_api)
+
+  pyx_files = ['yourmod.pyx']
+
+  foreach pyx_file: pyx_files
```
rgommers: No need to make it more complicated with a `foreach` loop here. Can remove this as well as the `pyx_files` variable and put `'yourmod.pyx'`as the first argument to `py.extension_module()`

doc/source/user/c-info.python-as-glue.rst
```diff
@@ -159,30 +159,63 @@ work with multidimensional arrays.
...
+    py.extension_module(
+      pyx_file,
+      dependencies: [np_dep],
+      install: true
```
rgommers: Would be useful to add `subdir: 'my_filter'` here, since it looks like the example is creating a `my_filter` Python package.

rgommers (changes requested):
Thanks Nathan! LGTM modulo some minor issues with the new example.

ngoldbaum:
@rgommers thanks for the suggestions! I went ahead and made sure the whole thing is end-to-end runnable, which required adding a `__init__.py` file. I think this is ready now.

rgommers (approved):
LGTM, but may need to hit the mailing list, so I won't hit the green button just yet.

ngoldbaum:
> but may need to hit the mailing list

I sent a message to the mailing list to elicit more feedback. I also said I'd merge the PR next week if no one objects.

ngoldbaum:
No one objected, so I'm planning to merge this as soon as the CI passes.

If anyone in the future reads this because of the compiler error introduced in this PR, please open an issue if updating to use Cython 3.0.0 or newer is impossible for some reason.

PR #8315, opened 2016-11-25
Add out of bound check for as_strided

uchida:
`numpy.lib.stride_trick.as_strided` is useful but dangerous function, it may lead out of bounds memory access when one give illegal shape and strides.

This pull request checks out of bounds by calculating/comparing actual bytes of original and returned arrays in memory, and raises `ValueError` when out of bounds.

PR #30771, opened 2026-02-02
ENH: add a variant of as_strided with bound checks.

Carreau:
This is a bit safer than as_strided, and can allow to catch typoes or mistake.

This also tries to lookup both the base array when using on a view; I'm not sure we want to do that; but it's easy; this comes from an internal discussion at work, so I include it – I can ask for the actual use case if necessary.

Consider this Pull request more a proof of concept for discussion, as
this could also be flag for as_strided, that either `skip checks | raise a warning | error`

I found #8315 from 10 years ago, but did not found any discussion on the ML.

### DIFF
--- a/numpy/lib/_stride_tricks_impl.py
+++ b/numpy/lib/_stride_tricks_impl.py
@@ -157,13 +157,13 @@ def as_strided(
         if view_low < base_low:
             raise ValueError(
                 f"Given shape and strides would access memory out of bounds. "
-                f"View starts {base_low - view_low} before starts"
+                f"View starts {base_low - view_low} bytes before lowest address"
             )
 
         if view_high > base_high:
             raise ValueError(
                 f"Given shape and strides would access memory out of bounds. "
-                f"View ends {view_high - base_high} after ends"
+                f"View ends {view_high - base_high} bytes after highest address"
             )
 
     return view
--- a/numpy/lib/_stride_tricks_impl.py
+++ b/numpy/lib/_stride_tricks_impl.py
@@ -8,7 +8,7 @@ import numpy as np
 from numpy._core.numeric import normalize_axis_tuple
 from numpy._core.overrides import array_function_dispatch, set_module
 
-__all__ = ['broadcast_to', 'broadcast_arrays', 'broadcast_shapes']
+__all__ = ["broadcast_to", "broadcast_arrays", "broadcast_shapes", "as_strided_checked"]
 
 
 class DummyArray:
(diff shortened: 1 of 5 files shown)


seberg:
While it seems nice, is there much of a niche if you subtract use-cases that can make use of `sliding_window_view`?

Carreau:
I've already suggested that in an internal discussion before opening this PR; and my understanding is that it does not work for the given use case.

seberg:
I wonder what others think, but to me it still seems very niche (although strictly good to do bounds check). I wonder if we should transition to a `check_bounds=True` via a warning on the existing function if we care a lot (force those that actually want out-of-bounds access to pass `check_bounds=False` in the future).

EDIT: But yeah, transitioning is a bit annoying... And might want to start with just adding the kwarg, but not doing an actual transition.

Carreau:
I tend to prefer a function with a different name so I can just swap in import and call it a day instead of reviewing each use case and have this kwarg dance. But up to you.

Carreau:
Let me know also if it's ok to look at the bytes of the base array, or if that's too much. 

I'll change this to use a kwarg.

numpy/lib/_stride_tricks_impl.py
```diff
@@ -87,27 +99,73 @@ def as_strided(x, shape=None, strides=None, subok=False, writeable=True):
...
 
-    view = _maybe_view_as_subclass(x, array)
+    view = _maybe_view_as_subclass(base, array)
 
     if view.flags.writeable and not writeable:
         view.flags.writeable = False
 
+    if check_bounds:
+        while base.base is not None:
+            base = base.base
```
seberg, line 152: I am OK with walking the base. But you have to refine it so that you also stop when `not isinstance(base.base, np.ndarray)`.  I suppose since `None` isn't an instance of ndarray and there is probably no reason to fret over a few ns here that would be sufficient also.

Marked as triage-review. (with it now beings just a new kwarg, I don't really see why not just add it.)

(I suppose unrelated, but wonder if we should add a `.. note::` that tells people to prefer `sliding_window_view`.  Such a note seems a bit missing.)

Carreau: > (I suppose unrelated, but wonder if we should add a `.. note::` that tells people to prefer `sliding_window_view`. Such a note seems a bit missing.)

It is in `See also` section:

```
    lib.stride_tricks.sliding_window_view :
        userfriendly and safe function for a creation of sliding window views.

```

seberg:
We decided to just add this, but need to ahve a pass over the implementation though.

Carreau:
No hurry; let me know if you need any changes and I'll be happy to implement them.

numpy/lib/_stride_tricks_impl.py
```diff
@@ -87,27 +99,73 @@ def as_strided(x, shape=None, strides=None, subok=False, writeable=True):
...
+        while isinstance(base.base, np.ndarray):
+            base = base.base
+
+        base_low, base_high = byte_bounds(base)
+        view_low, view_high = byte_bounds(view)
+
+        if view_low < base_low:
+            raise ValueError(
+                f"Given shape and strides would access memory out of bounds. "
+                f"View starts {base_low - view_low} before starts"
```
seberg: "before starts" and "after ends" reads a bit funny, but doesn't matter.

seberg (approved):
Thanks! I am happy, the most I could complain is that the tests are verbose and that may also just be a good thing :).

Good to merge, I'll wait briefly in case someone notices it now, but adding a keyword-only argument is safe enough that we don't need much discussion.

numpy/lib/_stride_tricks_impl.py
Carreau:
```suggestion
                f"View starts {base_low - view_low} bytes before lowest address"
```

Maybe ?

numpy/lib/_stride_tricks_impl.py
```diff
@@ -87,27 +99,73 @@ def as_strided(x, shape=None, strides=None, subok=False, writeable=True):
...
+            )
+
+        if view_high > base_high:
+            raise ValueError(
+                f"Given shape and strides would access memory out of bounds. "
+                f"View ends {view_high - base_high} after ends"
```
Carreau:
```suggestion
                f"View ends {view_high - base_high} bytes after highest address"
```

numpy/lib/_stride_tricks_impl.py
Carreau: we can also do `X bytes before beginning of array`/`X bytes after end of array`, or use `underlying memory region` instead of array; up to you.

Carreau:
I can also try to decrease the number of test, and use more `@parameterize` if you wish.

numpy/lib/_stride_tricks_impl.py
seberg: Sounds, good, let's just go with this.

seberg:
> I can also try to decrease the number of test, and use more @parameterize if you wish.

Nah, I'm good thanks, I don't think it matters much here either way.

Carreau:
Just poking to make sure this does not fall under the radar.

seberg:
Whoops, thanks @Carreau!  I had brought it up once and nobody had a strong opinion, IIRC, so in it goes :).

Carreau:
Thanks; and no worries; items a small thing just good to have it for next release (and i'm always fan of having less PRs open on any repos)

PR #30774, opened 2026-02-03
DEP: deprecate numpy.typename

### LINKED ISSUE #30762: DEP: Deprecating `typename`
The names returned by [`np.typename`](https://numpy.org/doc/stable/reference/generated/numpy.typename.html) are rather outdated and inconsistent:

```pycon
>>> import numpy as np
>>> np.typename("b"), np.typename("B")
('signed char', 'unsigned char')
>>> np.typename("i"), np.typename("I")
('integer', 'unsigned integer')
>>> np.typename("g"), np.typename("G")
('long precision', 'complex long double precision')
>>> np.typename("S"), np.typename("U")
('string', 'unicode')
```

These names don't match `dtype.name` or [Array API dtype names](https://data-apis.org/array-api/latest/API_specification/generated/array_api.info.dtypes.html). 

Additionaly, it lacks support for newer dtype codes such as `h`, `m`, `M`, and `T`. Attempting to use these will result in a `KeyError`, even though a `ValueError` would have been more appropriate:

```pycon
>>> import numpy as np
>>> np.typename("m")
    np.typename("m")
    ~~~~~~~~~~~^^^^^
  File "/home/joren/.cache/uv/archive-v0/jTWoQtjQfzJawFTGIbCll/lib/python3.14/site-packages/numpy/lib/_type_check_impl.py", line 636, in typename
    return _namefromtype[char]
           ~~~~~~~~~~~~~^^^^^^
KeyError: 'm'
>>> 
```

The [docs](https://numpy.org/doc/stable/reference/generated/numpy.typename.html) have also been incorrect for a long time now, but this was recently fixed in #30758.

A quick code search shows [85 hits](https://github.com/search?q=%2F%28%5C%28%7C%5C%5B%7C+%29np%5C.typename%5C%28%2F+language%3APython++NOT+is%3Aarchived+NOT+is%3Afork+NOT+path%3A%2F%5Enumpy%5C%2F%2F+NOT+path%3A%2Fsite-packages%5C%2F%2F++NOT+path%3A%2Fnumpy%5C%2Flib%5C%2F%2F++NOT+path%3A%2F%5C.pyi%24%2F&type=code), and most of them look like copies of the numpy source itself.

So, long story short, I think we should get rid of `typename`, and deprecate it in 2.5. Thoughts?

antareepsarkar:
resolves #30762 

This PR deprecates `numpy.typename` in favor of `numpy.dtype.name`.

### DIFF
--- a/numpy/conftest.py
+++ b/numpy/conftest.py
@@ -185,6 +185,7 @@ if HAVE_SCPDT:
                 "NumPy warning suppression and assertion utilities are deprecated.",
                 "numpy.fix is deprecated",  # fix -> trunc
                 "The chararray class is deprecated",  # char.chararray
+                "numpy.typename is deprecated",  # typename -> dtype.name
         ]
         msg = "|".join(msgs)
 
--- a/numpy/conftest.py
+++ b/numpy/conftest.py
@@ -185,7 +185,6 @@ if HAVE_SCPDT:
                 "NumPy warning suppression and assertion utilities are deprecated.",
                 "numpy.fix is deprecated",  # fix -> trunc
                 "The chararray class is deprecated",  # char.chararray
-                "numpy.typename is deprecated",  # typename -> dtype.name
         ]
         msg = "|".join(msgs)
 
--- a/numpy/lib/_type_check_impl.py
+++ b/numpy/lib/_type_check_impl.py
@@ -637,6 +637,7 @@ def typename(char):
     q  :  long long integer
 
     """
+    # Deprecated in NumPy 2.5, 2026-02-03
     warnings.warn(
         "numpy.typename is deprecated. Use numpy.dtype.name instead.",
         DeprecationWarning,
--- a/numpy/lib/_type_check_impl.py
+++ b/numpy/lib/_type_check_impl.py
@@ -642,8 +642,7 @@ def typename(char):
     warnings.warn(
         "numpy.typename is deprecated. Use numpy.dtype.name instead.",
         DeprecationWarning,
-        skip_file_prefixes=(os.path.dirname(__file__),),
-        stacklevel=2,
+        skip_file_prefixes=(os.path.dirname(__file__),)
     )
     return _namefromtype[char]
 
--- a/numpy/lib/_type_check_impl.py
+++ b/numpy/lib/_type_check_impl.py
@@ -2,6 +2,7 @@
 
 """
 import functools
+import warnings
 
 __all__ = ['iscomplexobj', 'isrealobj', 'imag', 'iscomplex',
            'isreal', 'nan_to_num', 'real', 'real_if_close',
@@ -587,6 +588,9 @@ def typename(char):
     """
     Return a description for the given data type code.
 
+    .. deprecated:: 2.5
+        `numpy.typename` is deprecated. Use `numpy.dtype.name` instead.
+
     Parameters
     ----------
     char : str
@@ -633,6 +637,11 @@ def typename(char):
     q  :  long long integer
 
     """
+    warnings.warn(
+        "numpy.typename is deprecated. Use numpy.dtype.name instead.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
     return _namefromtype[char]
 
 #-----------------------------------------------------------------------------
--- a/numpy/lib/_type_check_impl.py
+++ b/numpy/lib/_type_check_impl.py
@@ -2,7 +2,6 @@
 
 """
 import functools
-import os
 import warnings
 
 __all__ = ['iscomplexobj', 'isrealobj', 'imag', 'iscomplex',
@@ -642,7 +641,7 @@ def typename(char):
     warnings.warn(
         "numpy.typename is deprecated. Use numpy.dtype.name instead.",
         DeprecationWarning,
-        skip_file_prefixes=(os.path.dirname(__file__),)
+        stacklevel=2
     )
     return _namefromtype[char]
 
--- a/numpy/lib/_type_check_impl.py
+++ b/numpy/lib/_type_check_impl.py
@@ -2,6 +2,7 @@
 
 """
 import functools
+import os
 import warnings
 
 __all__ = ['iscomplexobj', 'isrealobj', 'imag', 'iscomplex',
@@ -641,6 +642,7 @@ def typename(char):
     warnings.warn(
         "numpy.typename is deprecated. Use numpy.dtype.name instead.",
         DeprecationWarning,
+        skip_file_prefixes=(os.path.dirname(__file__),),
         stacklevel=2,
     )
     return _namefromtype[char]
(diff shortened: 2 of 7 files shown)


jorenham:
deprecations should not change behavior; that defeats its purpose

antareepsarkar:
Probably, the test did not fail because of the changes.
Is the last commit okay?

numpy/lib/_type_check_impl.py
```diff
@@ -633,6 +637,11 @@ def typename(char):
     q  :  long long integer
 
     """
+    warnings.warn(
+        "numpy.typename is deprecated. Use numpy.dtype.name instead.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
```
jorenham, line 645:
```suggestion
	# Deprecated in NumPy 2.5, 2026-02-03
    warnings.warn(
        "numpy.typename is deprecated. Use numpy.dtype.name instead.",
        DeprecationWarning,
        stacklevel=2,
    )
```

numpy/lib/tests/test_type_check.py
```diff
@@ -471,3 +473,14 @@ def test_basic(self):
...
+class TestTypenameDeprecation:
+    """Test that numpy.typename emits a DeprecationWarning."""
+
+    def test_typename_emits_deprecation_warning(self):
+        a = ['S1', '?', 'B', 'D', 'G', 'F', 'I', 'H', 'L', 'O', 'Q',
+             'S', 'U', 'V', 'b', 'd', 'g', 'f', 'i', 'h', 'l', 'q']
+
+        for c in a:
+            with pytest.warns(DeprecationWarning, match="numpy.typename is deprecated"):
+                np.typename(c)
```
jorenham: There's no need to check all inputs AFAIK. But I wouldn't mind if you want to do so anyway, although in that case it would be cleaner to refactor this as a parameterized test.

jorenham (commented):
Apart form some minor suggestions, this looks good to me :)

numpy/lib/tests/test_type_check.py
antareepsarkar: Getting back soon after doing the changes locally.

antareepsarkar: I did the changes locally as the tests needed to be changed, and there may have been errors if I did it directly.

numpy/lib/_type_check_impl.py
tylerjereddy: Should we start encouraging i.e., `skip_file_prefixes=(os.path.dirname(__file__),)` now that `3.11` has been dropped?

jorenham: yea good idea, I forgot we could do that

antareepsarkar: Should I change it?

jorenham: yes please

numpy/lib/_type_check_impl.py
```diff
@@ -633,6 +638,13 @@ def typename(char):
     q  :  long long integer
 
     """
+    # Deprecated in NumPy 2.5, 2026-02-03
+    warnings.warn(
+        "numpy.typename is deprecated. Use numpy.dtype.name instead.",
+        DeprecationWarning,
+        skip_file_prefixes=(os.path.dirname(__file__),),
+        stacklevel=2,
```
jorenham: I don't think we need the `stacklevel` anymore now

numpy/lib/_type_check_impl.py
```diff
@@ -633,6 +638,13 @@ def typename(char):
     q  :  long long integer
 
     """
+    # Deprecated in NumPy 2.5, 2026-02-03
+    warnings.warn(
+        "numpy.typename is deprecated. Use numpy.dtype.name instead.",
+        DeprecationWarning,
+        skip_file_prefixes=(os.path.dirname(__file__),),
```
jorenham:
```suggestion
        skip_file_prefixes=(os.path.dirname(__file__)),
```

antareepsarkar: This is giving a `TypeError`, saying that the argument must be a tuple. I may have made some mistake, so please check once.

jorenham: Oh sorry I didn't realize this was a tuple

numpy/lib/_type_check_impl.py
antareepsarkar: Should we change the test requiring `stacklevel`?

jorenham: Hmm, those failures are unfortunate... But it seems like we should #30358 separately, because if we change the tests to switch from `stacklevel` to `skip_file_prefixes`, then we'd also have to update every `warn` call in the codebase. So that should be done in a dedicated PR.
So I'm afraid that we should should then revert to just use `stacklevel` after all, sorry about that.

antareepsarkar:
Are the changes okay? Is something still left?

numpy/lib/tests/test_type_check.py
```diff
@@ -471,3 +473,12 @@ def test_basic(self):
...
         assert_all(isrealobj(b))
+
+class TestTypenameDeprecation:
+    """Test that numpy.typename emits a DeprecationWarning."""
+
+    def test_typename_emits_deprecation_warning(self):
+        with pytest.warns(DeprecationWarning, match="numpy.typename is deprecated"):
+            np.typename("S1")
+        with pytest.warns(DeprecationWarning, match="numpy.typename is deprecated"):
+            np.typename("h")
```
mattip: Deprecation tests should live in `numpy/_core/tests/test_deprecations.py`

numpy/conftest.py
```diff
@@ -185,6 +185,7 @@ def warnings_errors_and_rng(test=None):
                 "NumPy warning suppression and assertion utilities are deprecated.",
                 "numpy.fix is deprecated",  # fix -> trunc
                 "The chararray class is deprecated",  # char.chararray
+                "numpy.typename is deprecated",  # typename -> dtype.name
```
mattip: At some point we should revisit this list and see if any can be removed in another PR.

Are there tests that need the `typename` filter? Can't they use `dtype.name` instead?

mattip (commented):
Two small comments

numpy/conftest.py
antareepsarkar: It seems that I added it by mistake. It is not required as no tests require `typename`. 
Just removed it in the latest commit.

antareepsarkar: Circle CI failed. Because of the removal?

mattip: Yes, sorry I misled you. We need the warning suppression for doctests

mattip:
LGTM. Is this too small to hit the mailing list before deprecation?

jorenham:
> Is this too small to hit the mailing list before deprecation?

I suppose it wouldn't hurt, but I don't expect anyone would care about this one tbh :yum:

antareepsarkar:
So, should it go to the mailing list?

jorenham:
Let's just see what happens. Thanks @antareepsarkar

PR #30775, opened 2026-02-03
DOC: Use ``spin`` for lint checks

ganesh-k13:
### Use ``spin`` for lint checks

Updated doc to use ``spin`` instead of exposing the tools script

### Generated Output

https://output.circle-artifacts.com/output/job/b3ecb9d4-d2ab-493c-adf9-a8fee1ab7398/artifacts/0/doc/build/html/dev/development_environment.html#running-linting


mattip:
Thanks @ganesh-k13

PR #30776, opened 2026-02-03
API: Make PyArrayObject opaque on internal builds

ngoldbaum:
Towards fixing https://github.com/numpy/numpy/issues/30704.

After #30705 we don't access any fields in the PyObject header, so it's safe to make this opaque. Only doing it for internal builds for now but in a future PR I'll also do this for opaque PyObject builds.

In the future we can also make the public definition opaque and deprecate the old definition.


ngoldbaum:
closing, see https://github.com/numpy/numpy/pull/30744#issuecomment-4099703492

PR #30778, opened 2026-02-04
API: expose accessors for ufunc object members and use them internally

ngoldbaum:
Towards fixing #30704. Also see #30744.

Opening as a draft because docs and discussion are still TODO.

This makes `PyUFuncObject` an opaque pointer internally, moves the fields to a new `PyUFunc_fields` struct, and exposes accessor macros for all fields and setter macros for a subset of fields needed by the NumPy implementation. I also added `PyUFunc_GET_ITEM_DATA` to get the fields struct. We may decide to simply expose `GET_ITEM_DATA` instead of the setters. We also probably don't need to expose getters for members we deem to be private.

Opening for initial review and as a basis for a performance test on Windows.


ngoldbaum:

PR #30781, opened 2026-02-05
Add modular matrix inversion utility to numpy.matrixlib

Akshay-Ramesh-VITC:
This PR proposes a new utility function inv_mod under numpy.matrixlib for computing the inverse of a square integer matrix modulo an integer m, when such an inverse exists.

The implementation follows the standard adjugate–determinant identity:

𝐴^{−1}≡(det(𝐴)^{−1})⋅adj(𝐴)(mod 𝑚)

and is intended for educational, algebraic, and cryptographic use cases that operate over modular arithmetic rather than real or complex fields.

While NumPy’s linear algebra functionality focuses on numerical computation over ℝ and ℂ, numpy.matrixlib already provides higher-level matrix utilities that are not strictly LAPACK-backed.

There are recurring use cases in: classical cryptography (e.g. Hill cipher), coding theory, modular linear algebra education, where users require matrix inversion over ℤₘ rather than floating-point fields. Currently, users must either reimplement this logic themselves or rely on symbolic libraries for otherwise simple numeric workflows.

This PR introduces a small, self-contained utility that: operates purely on integer arrays, does not alter existing numerical linear algebra APIs, is explicitly scoped to modular arithmetic.

**Proposed API**
np.matrixlib.inv_mod(A, m)

Parameters:

A: Square array_like of intergers
m: Interger Modulus

Returns:
ndarray: inverse of A modulo m

Raises:
ValueError if A is not invertible modulo m

**Code Example:**
import numpy as np

A = np.array([[3, 3],
              [2, 5]])

invA = np.matrixlib.inv_mod(A, 26)

assert np.all((A @ invA) % 26 == np.eye(2) % 26)

**Implementation Details**

Determinant is computed via np.linalg.det and rounded to handle floating-point artifacts.
1. Invertibility is checked using gcd(det(A), m) == 1.
2. The adjugate matrix is constructed using minors and cofactors.
3. Final result is reduced modulo m.
The function is implemented entirely in Python and does not depend on LAPACK or BLAS routines.

**Limitations**

Intended for small matrices (educational / cryptographic use).
1. Not optimized for large n.
2. This is not a replacement for numerical inversion routines in np.linalg.
These constraints are documented in the docstring.

Tests Added

Inversion of valid 2×2 matrices modulo 26
Verification that A · A⁻¹ ≡ I (mod m)
Failure cases when gcd(det(A), m) ≠ 1
All tests pass locally.

**Scope Considerations**
This functionality is deliberately placed under numpy.matrixlib rather than numpy.linalg to avoid conflating numerical linear algebra over fields with modular arithmetic over rings. If maintainers feel this functionality is better suited as an external package, this PR can serve as a reference implementation and API discussion starter.

**Related Work**
SymPy supports modular matrix inversion via Matrix.inv_mod(m). This PR provides a NumPy-array–centric alternative for users who do not require symbolic computation.

### DIFF
--- a/numpy/matrixlib/modular.py
+++ b/numpy/matrixlib/modular.py
@@ -1,0 +1,20 @@
+import numpy as np
+from math import gcd
+
+def inv_mod(A, m):
+    A = np.asarray(A, dtype=int)
+    n = A.shape[0]
+
+    det = int(round(np.linalg.det(A))) % m
+    if gcd(det, m) != 1:
+        raise ValueError("Matrix not invertible modulo m")
+
+    det_inv = pow(det, -1, m)
+
+    cof = np.zeros((n, n), dtype=int)
+    for i in range(n):
+        for j in range(n):
+            minor = np.delete(np.delete(A, i, 0), j, 1)
+            cof[i, j] = ((-1)**(i+j)) * int(round(np.linalg.det(minor)))
+
+    return (det_inv * cof.T) % m


rkern:
Please take a look at our contributing guidelines linked below. This does not follow our guidelines in several respects.

Before trying to fix this PR to meet these guidelines, however, I do suggest waiting for feedback on the issue proposing this functionality. I suspect that we do not want this functionality in the first place.

PR #30783, opened 2026-02-05
Add unit test for inv_mod function in matrixlib

Akshay-Ramesh-VITC:
Companion unit test for the inv_mod function defined in the matrixlib/modular.py


rkern:
Please do not submit tiny "companion PRs". These all should have been in one PR.

PR #30794, opened 2026-02-06
BUG: Handle empty x array in np.interp

### LINKED ISSUE #30316: BUG: np.interp([], [], []) fails, when it could return an empty array

Mr-Neutr0n:
## Summary

- Fixes `np.interp([], [], [])` raising `ValueError` instead of returning an empty array
- Adds early check for empty `x` arrays and returns empty array with appropriate dtype

## Details

When `np.interp` is called with empty arrays (e.g., `np.interp([], [], [])`), it currently raises:

```
```

However, the correct behavior should be to return an empty array (which is clearly the expected result for interpolating zero points).

## Fix

The fix adds an early check at the beginning of the function:

```python
x = np.asarray(x)
xp = np.asarray(xp)
fp = np.asarray(fp)

# Handle empty x array: return empty array with appropriate dtype
if x.size == 0:
    if np.iscomplexobj(fp):
        return np.empty(x.shape, dtype=np.complex128)
    else:
        return np.empty(x.shape, dtype=np.float64)
```

This preserves the shape of `x` while returning an empty array with the correct dtype.

## Test plan

- [x] `np.interp([], [], [])` now returns `array([], dtype=float64)` instead of raising an error
- [x] Complex interpolation: `np.interp([], [], [1j])` returns `array([], dtype=complex128)`

Fixes #30316

### DIFF
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -1638,8 +1638,16 @@ def interp(x, xp, fp, left=None, right=None, period=None):
     array([0.+1.j , 1.+1.5j])
 
     """
-
+    x = np.asarray(x)
+    xp = np.asarray(xp)
     fp = np.asarray(fp)
+
+    # Handle empty x array: return empty array with appropriate dtype
+    if x.size == 0:
+        if np.iscomplexobj(fp):
+            return np.empty(x.shape, dtype=np.complex128)
+        else:
+            return np.empty(x.shape, dtype=np.float64)
 
     if np.iscomplexobj(fp):
         interp_func = compiled_interp_complex


ikrommyd:
As a user, I think I'd honestly like to get an error like "hey what are you doing?" if I called `np.interp([], [], [])` cause I don't really see how one would get to that point and want to continue silently.

ikrommyd:
Also scipy's interpolators error if you try to create them with empty x,y arrays so that would be an inconsistency.

Mr-Neutr0n:
Thanks for the feedback @ikrommyd! You make a good point about consistency with scipy's behavior.

Looking at the original issue #30316, the reporter expected an empty array return (similar to how `np.interp([], [1,2], [1,2])` already returns `[]`). However, your point about scipy raising errors and catching potential bugs is valid.

Would you prefer:
1. **Raise a clear error** when `xp` or `fp` are empty (matching scipy behavior)
2. **Keep the empty array return** for consistency with how `np.interp` handles empty `x`

I'm happy to update the implementation either way. What do the maintainers think?

jorenham:
> What do the maintainers think?

That I prefer talking to a human ;)

hxrikp1729:
@jorenham I am a human. But yes sometimes i do allow my personal JARVIS to do some replies. Can opt out of that here if its a concern

Mr-Neutr0n:
@jorenham Are we safe to merge?

jorenham:
> @jorenham Are we safe to merge?

No not really actually. As @ikrommyd correctly noted, there's also something to be said for having it raise for empty input. So this should probably go through the [mailing list](https://mail.python.org/archives/list/numpy-discussion@python.org/) first.

ikrommyd:
FWIW, I think that it's probably best to error like it's doing now but perhaps with a slightly better error message than `ValueError: array of sample points is empty` (which is also a very valid error message anyways). If the user wants to write for example some data analysis code where empty array chunks pass through interpolation, then it's very easy for them to add the 2 line if statement you added here manually. For reference, the scipy errors are way worse and can definitely be improved (but that's an issue for scipy)
```py
In [1]: from scipy import interpolate

In [2]: interpolate.interp1d([], [])
Cell In[2], line 1
----> 1 interpolate.interp1d([], [])

File ~/.pyenv/versions/3.14.3/lib/python3.14/site-packages/scipy/interpolate/_interpolate.py:333, in interp1d.__init__(self, x, y, kind, axis, copy, bounds_error, fill_value, assume_sorted)
    331 # Interpolation goes internally along the first axis
    332 self.y = y
--> 333 self._y = self._reshape_yi(self.y)
    334 self.x = x
    335 del y, x  # clean up namespace to prevent misuse; use attributes

File ~/.pyenv/versions/3.14.3/lib/python3.14/site-packages/scipy/interpolate/_polyint.py:118, in _Interpolator1D._reshape_yi(self, yi, check)
    115     ok_shape = (f"{self._y_extra_shape[-self._y_axis:]!r} + (N,) + "
    116                 f"{self._y_extra_shape[:-self._y_axis]!r}")
    117     raise ValueError(f"Data must be of shape {ok_shape}")
--> 118 return yi.reshape((yi.shape[0], -1))

ValueError: cannot reshape array of size 0 into shape (0,newaxis)

In [3]: interpolate.make_interp_spline([], [])
IndexError                                Traceback (most recent call last)
Cell In[3], line 1
----> 1 interpolate.make_interp_spline([], [])

File ~/.pyenv/versions/3.14.3/lib/python3.14/site-packages/scipy/interpolate/_bsplines.py:1611, in make_interp_spline(x, y, k, t, bc_type, axis, check_finite)
   1609         t = _periodic_knots(x, k)
   1610     else:
-> 1611         t = _not_a_knot(x, k)
   1612 else:
   1613     t = _augknt(x, k)

File ~/.pyenv/versions/3.14.3/lib/python3.14/site-packages/scipy/interpolate/_bsplines.py:1076, in _not_a_knot(x, k)
   1073     t = (x[1:] + x[:-1]) / 2
   1075 t = t[k2:-k2]
-> 1076 t = np.r_[(x[0],)*(k+1), t, (x[-1],)*(k+1)]
   1077 return t

IndexError: index 0 is out of bounds for axis 0 with size 0

In [4]: interpolate.make_interp_spline([], [], 1)
IndexError                                Traceback (most recent call last)
Cell In[4], line 1
----> 1 interpolate.make_interp_spline([], [], 1)

File ~/.pyenv/versions/3.14.3/lib/python3.14/site-packages/scipy/interpolate/_bsplines.py:1592, in make_interp_spline(x, y, k, t, bc_type, axis, check_finite)
   1590 if not (deriv_l is None and deriv_r is None):
   1591     raise ValueError("Too much info for k=1: bc_type can only be None.")
-> 1592 t = np.r_[x[0], x, x[-1]]
   1593 c = np.asarray(y)
   1594 c = np.ascontiguousarray(c, dtype=_get_dtype(c.dtype))

IndexError: index 0 is out of bounds for axis 0 with size 0
```

Mr-Neutr0n:
Fair enough — sounds like this needs more discussion before any code changes. If the consensus ends up being that an error is the right behavior here, this PR wouldn't be the right approach anyway. Feel free to close this one out, and I can revisit if needed after the mailing list discussion settles on a direction.

jorenham:
> after the mailing list discussion settles on a direction.

It'd help if you'd actually start one

Mr-Neutr0n:
Hey, fair enough — I'll get a thread going on the mailing list. Seems like there's a few directions this could go (return empty vs raise) so makes sense to hash that out there first before committing to an approach here.

Mr-Neutr0n:
Closing this — sounds like this needs a mailing list discussion first to settle on the right direction. Happy to revisit once there's consensus on the approach.

jorenham:
> sounds like this needs a mailing list discussion first to settle on the right direction

yes; that's why I suggested you do so (two times now, making this the third)

PR #30797, opened 2026-02-06
API: make array scalars opaque internally and expose accessors

ngoldbaum:
Towards fixing https://github.com/numpy/numpy/issues/30704. Also see https://github.com/numpy/numpy/pull/30744.

This makes the structs backing the scalar objects opaque in the internal API. I also added a few accessors that were missing but for the most part we can use `PyArrayScalar_VAL` for this.

Because the bytes scalar extends `PyBytesObject`, I had to adjust some of the codegen to account for the fact that the NumPy headers don't define `PyBytesObject`.

### DIFF
--- a/numpy/_core/code_generators/genapi.py
+++ b/numpy/_core/code_generators/genapi.py
@@ -402,7 +402,7 @@ class BoolValuesApi:
 
     def internal_define(self):
         astr = """\
-extern NPY_NO_EXPORT PyBoolScalarObject _PyArrayScalar_BoolValues[2];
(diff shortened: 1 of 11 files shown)


seberg:
I am slightly of the opinion that this direct access is rather unnecessary, because I think using `PyArray_Pack()` to get the value is on much more solid footing without a ridiculous overhead (i.e. it works for user scalars that may not be able to expose this all that cleanly).
(Yes, for structured dtypes or large strings, the copy might not be ideal, but...)

But, if we do the internal refactor, then there is probably no real reason not to expose it publicly.

ngoldbaum:

PR #30761, opened 2026-02-01
BUG: Fix some leaks found in f2py via LeakSanitizer

mayeut:
This fixes leaks reported when running the f2py test suite with AddressSanitizer+LeakSanitizer enabled.

Given the discussion in #30583, I'm thinking of proposing a follow-up PR which adds CI AddressSanitizer+LeakSanitizer for Linux x86_64 if there is interest (at the moment,  AddressSanitizer is only ran in CI on macOS arm64 if I'm not mistaken).

PR #30800, opened 2026-02-07
CI: add Linux x86_64 clang ASAN+LSAN job

mayeut:
This PR adds a Linux x86_64 clang AddressSanitizer+LeakSanitizer job as discussed in https://github.com/numpy/numpy/pull/30761#issuecomment-3837059875 and #30583

I'll try to address the various comments found in #30583 regarding leak checks in this description.

For now, AddressSanitizer is only ran on macOS arm64 so it does make sense to add a job for Linux x86_64 which likely covers a much larger share of users and uses different optimisations not yet covered by existing AddressSanitizer tests. LeakSanitizer should not add much overhead to testing time compared to AddressSanitizer tests.

Limitations:
- CPython itself does not seem to be tested with `detect_leaks=1` and thus, leaks in CPython might be revealed here on CPython upgrades.
- The actual leak check is only done once the interpreter exits and the (native only) stack trace does not help tracking which tests leak. As mentioned by @seberg in https://github.com/numpy/numpy/pull/30583#issuecomment-3715109379, if this runs on every PR then you will know relatively well what might have caused the leak as it would be related to code changed in the PR itself.
- It's much easier to skip some tests rather than to try and update the suppression file (there are 4 tests skipped in this PR, at least one of those should be a leak mentioned in https://github.com/numpy/numpy/pull/30680#issue-3832871611) which mean those are not covered by AddressSanitizer, for this specific job.

Remarks:
- I didn't try to use `__lsan_do_leak_check`/`__lsan_do_recoverable_leak_check` to check if it could help narrow down which specific tests are actually leaking. This could be investigated as a future enhancement.
- I would expect more than 2 leaks for initialize_static_globals, furthermore, when running the full test suite, those do not appear (as opposed to running `python -c "import numpy"` or some subprocess tests in the test suite).
- Trying to use `ghcr.io/nascheme/numpy-asan:3.14t` shows many leaks and at this point, I did not check any of those and felt it was better to get something working reliably with the GIL-enabled interpreter rather than nothing.
- Running on Linux aarch64 (I do use this locally in a docker container on macOS at home where I do opensource contributions in my spare time) is really slow compared to x86_64 (which I'm more used to at work).
- Skipping tests is only done based on the presence of `LSAN_OPTIONS` in environment variables. I did not want to overcomplicate things to determine if leak sanitizer was actually enabled. This also means that if the suppression file evolves to be empty, it shall probably shall be kept, rather than removed, with no updates to the job or, this skip mechanism should be reworked.
- The leaks fixed by #30756 and #30761 were found while setting up this job in my fork.

### DIFF
--- a/.github/workflows/compiler_sanitizers.yml
+++ b/.github/workflows/compiler_sanitizers.yml
@@ -108,3 +108,85 @@ jobs:
           python -m spin test \
           `find numpy -name "test*.py" | xargs grep -E -l "import threading|ThreadPoolExecutor" | tr '\n' ' '` \
            -- -v -s --timeout=600 --durations=10
+
+  clang_ASAN:
+    # To enable this workflow on a fork, comment out:
+    if: github.repository == 'numpy/numpy'
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+      with:
+        submodules: recursive
+        fetch-tags: true
+        persist-credentials: false
+
+    - name: Set up pyenv
+      run: |
+        git clone https://github.com/pyenv/pyenv.git "$HOME/.pyenv"
+        PYENV_ROOT="$HOME/.pyenv"
+        PYENV_BIN="$PYENV_ROOT/bin"
+        PYENV_SHIMS="$PYENV_ROOT/shims"
+        echo "$PYENV_BIN" >> $GITHUB_PATH
+        echo "$PYENV_SHIMS" >> $GITHUB_PATH
+        echo "PYENV_ROOT=$PYENV_ROOT" >> $GITHUB_ENV
+
+    - name: Check pyenv is working
+      run: pyenv --version
+
+    - name: Install python & numpy build dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get -yq install \
+          libb2-dev \
+          libbz2-dev \
+          libffi-dev \
+          libgdbm-dev \
+          libgdbm-compat-dev \
+          liblzma-dev \
+          libncurses5-dev \
+          libopenblas-dev \
+          libreadline6-dev \
+          libsqlite3-dev \
+          libssl-dev \
+          libzstd-dev \
+          lzma-dev \
+          ninja-build \
+          tk-dev \
+          uuid-dev \
+          zlib1g-dev
+
+    - name: Install clang 21
+      run: |
+        curl -fsSLo /tmp/llvm.sh https://apt.llvm.org/llvm.sh
+        chmod +x /tmp/llvm.sh
+        sudo /tmp/llvm.sh 21
+        CC=/usr/bin/clang-21
+        CXX=/usr/bin/clang++-21
+        echo "CC=$CC" >> $GITHUB_ENV
+        echo "CXX=$CXX" >> $GITHUB_ENV
+        sudo update-alternatives --install /usr/bin/cc cc $CC 100
+        sudo update-alternatives --install /usr/bin/c++ c++ $CXX 100
+
+    - name: Build Python with address sanitizer
+      run: |
+        CONFIGURE_OPTS="--with-address-sanitizer" pyenv install 3.14
+        pyenv global 3.14
+
+    - name: Install dependencies
+      run: |
+        pip install -r requirements/build_requirements.txt
+        pip install -r requirements/ci_requirements.txt
+        pip install -r requirements/test_requirements.txt
+        # xdist captures stdout/stderr, but we want the ASAN output
+        pip uninstall -y pytest-xdist
+
+    - name: Build
+      run: python -m spin build -j4 -- -Db_sanitize=address,leak
+
+    - name: Test
+      run: |
+        # pass -s to pytest to see ASAN errors and warnings, otherwise pytest captures them
+        export ASAN_OPTIONS="detect_leaks=1:symbolize=1:strict_init_order=true:allocator_may_return_null=1:use_sigaltstack=0"
+        export LSAN_OPTIONS="suppressions=$GITHUB_WORKSPACE/tools/ci/lsan_suppressions.txt"
+        python -m spin test -- -v -s --timeout=600 --durations=10
--- a/.github/workflows/compiler_sanitizers.yml
+++ b/.github/workflows/compiler_sanitizers.yml
@@ -156,6 +156,8 @@ jobs:
           uuid-dev \
           zlib1g-dev
 
+    # TODO install as a regular ubuntu package in the previous step once
+    # ubuntu 26.04 is available in GHA
     - name: Install clang 21
       run: |
         curl -fsSLo /tmp/llvm.sh https://apt.llvm.org/llvm.sh
--- a/numpy/_core/tests/test_dtype.py
+++ b/numpy/_core/tests/test_dtype.py
@@ -2,6 +2,7 @@ import contextlib
 import ctypes
 import inspect
 import operator
+import os
 import pickle
 import sys
 import types
@@ -1518,6 +1519,7 @@ class TestFromDTypeAttribute:
         with pytest.raises(ValueError):
             np.dtype(dt_instance)
 
+    @pytest.mark.xfail("LSAN_OPTIONS" in os.environ, reason="known leak", run=False)
     def test_void_subtype(self):
         class dt(np.void):
             # This code path is fully untested before, so it is unclear
@@ -1897,6 +1899,7 @@ class TestUserDType:
     @pytest.mark.thread_unsafe(
         reason="crashes when GIL disabled, dtype setup is thread-unsafe",
     )
+    @pytest.mark.xfail("LSAN_OPTIONS" in os.environ, reason="known leak", run=False)
     def test_custom_structured_dtype(self):
         class mytype:
             pass
@@ -1920,6 +1923,7 @@ class TestUserDType:
     @pytest.mark.thread_unsafe(
         reason="crashes when GIL disabled, dtype setup is thread-unsafe",
     )
+    @pytest.mark.xfail("LSAN_OPTIONS" in os.environ, reason="known leak", run=False)
     def test_custom_structured_dtype_errors(self):
         class mytype:
             pass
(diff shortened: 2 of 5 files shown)


ngoldbaum:
Neat, thanks for working on this! I'll try to take a close look at this next week sometime.

.github/workflows/compiler_sanitizers.yml
jorenham, line 1: This looks like it could take a while to install everything. So I'm guessing that caching things could help with that.

ngoldbaum:
Sorry for dropping this! I'll look this over today.

.github/workflows/compiler_sanitizers.yml
ngoldbaum: It looks like the slow part of the gcc_ASAN test is actually running the tests, not installing everything. So caching won't help much and it will eat into our project-wide cache budget.

.github/workflows/compiler_sanitizers.yml
```diff
@@ -108,3 +108,73 @@ jobs:
           python -m spin test \
           `find numpy -name "test*.py" | xargs grep -E -l "import threading|ThreadPoolExecutor" | tr '\n' ' '` \
            -- -v -s --timeout=600 --durations=10
+
+  gcc_ASAN:
```
ngoldbaum: Sorry for taking two weeks to notice this.

For a while we were using the GCC compiler sanitizers in CI but [turned them off](https://github.com/numpy/numpy/issues/25875) after encountering mysterious flaky failures. Perhaps things have improved since then?

If I were setting this up from scratch on Linux, I'd use an LLVM toolchain.

I have a vague feeling that the tests will also run faster under LLVM's ASan implementation. Of course it's apples-to-oranges because of the different CPU and OS but the Mac ASan builds using LLVM run the tests in 8.5 minutes but the GCC ASan tests run in 15 minutes.

Thankfully, we have a neat fix for this already. @nascheme set up ASan docker images that have a CPython build using an LLVM toolchain, see https://github.com/nascheme/cpython_sanity. You should be able to follow the set up of the docker images for the TSan tests, which also use the `cpython_sanity` images and  an LLVM toolchain. This also nicely sidesteps the need to build CPython from source under TSan and the docker images already have a build environment installed.

mayeut: I haven't seen flakiness running gcc but in any case moving to clang allows to reduce run time a bit, mostly numpy build time, [see this workflow run in my fork](https://github.com/mayeut/numpy/actions/runs/22297062608/job/64495816980).
I can move to clang in this PR if this is the preferred option.

As for the Asan docker image, it is not usable without further investigations into the leaks reported when running the free-threaded python build which is the only one available in those images as mentioned in my 3rd remark in the description: "Trying to use ghcr.io/nascheme/numpy-asan:3.14t shows many leaks and at this point, I did not check any of those and felt it was better to get something working reliably with the GIL-enabled interpreter rather than nothing."
I'm trying to narrow down the tests causing those leaks but it will take weeks given the time I can spend on this in my spare time.

ngoldbaum: > which is the only one available in those images as mentioned in my 3rd remark in the description

Sorry, missed this.

I'd like to fix these issues on the free-threaded build, but no worries if you're not interested.

> I can move to clang in this PR if this is the preferred option.

Let's do that.

mayeut: > Sorry, missed this.

no worries

> I'd like to fix these issues on the free-threaded build, but no worries if you're not interested.

I'll try & post the list of test files that shows leak even if I can't narrow it down to individual tests just yet. This first pass should be faster and can help further investigations by others. Maybe this is better posted as a new issue in the bug tracker.

> Let's do that.

done. rebased on main and added clang-21 installation.

ngoldbaum:
This now comes in at 25 minutes, which is only about a minute longer than our longest QEMU-based test jobs.

I think there are plenty of opportunities to speed this up and I'm OK with adding another long test job - particularly one that will help us avoid memory leaks in CI.

@charris @rgommers @seberg what do you all think?

seberg:
I agree, I think this is awesome to have on regular CI, thanks for all the work!

.github/workflows/compiler_sanitizers.yml
```diff
@@ -108,3 +108,85 @@ jobs:
...
+        # xdist captures stdout/stderr, but we want the ASAN output
+        pip uninstall -y pytest-xdist
+
+    - name: Build
+      run: python -m spin build -j4 -- -Db_sanitize=address,leak
+
+    - name: Test
+      run: |
+        # pass -s to pytest to see ASAN errors and warnings, otherwise pytest captures them
+        export ASAN_OPTIONS="detect_leaks=1:symbolize=1:strict_init_order=true:allocator_may_return_null=1:use_sigaltstack=0"
```
ngoldbaum, line 192: First time I've seen `use_sigaltstack` or `strict_init_order`. Should we maybe be enabling in other sanitizer jobs?

ngoldbaum (approved):
I looked over the CI logs and didn't see anything besides clang-21 adding some compiler warnings, which isn't something to worry about here. One question inline.

rgommers:
Agreed, sounds like a good idea to me - I don't see a problem with one more heavy job. Thanks for doing this.

.github/workflows/compiler_sanitizers.yml
```diff
@@ -108,3 +108,85 @@ jobs:
...
+          libzstd-dev \
+          lzma-dev \
+          ninja-build \
+          tk-dev \
+          uuid-dev \
+          zlib1g-dev
+
+    - name: Install clang 21
+      run: |
+        curl -fsSLo /tmp/llvm.sh https://apt.llvm.org/llvm.sh
```
rgommers, line 163: This looks a bit questionable, if acceptable. We should replace it as soon as Clang 21 packages are available from Ubuntu though, it looks like that's in the 26.04 image which should arrive in a month or two. Maybe add a comment?

.github/workflows/compiler_sanitizers.yml
```diff
@@ -108,3 +108,85 @@ jobs:
...
+        CC=/usr/bin/clang-21
+        CXX=/usr/bin/clang++-21
+        echo "CC=$CC" >> $GITHUB_ENV
+        echo "CXX=$CXX" >> $GITHUB_ENV
+        sudo update-alternatives --install /usr/bin/cc cc $CC 100
+        sudo update-alternatives --install /usr/bin/c++ c++ $CXX 100
+
+    - name: Build Python with address sanitizer
+      run: |
+        CONFIGURE_OPTS="--with-address-sanitizer" pyenv install 3.14
```
rgommers, line 175: The line numbers in the suppressions file might change with bug fix releases. Might be better to use `3.14.3` here?

rgommers (approved):
Overall LGTM, approving - just two small suggestions

.github/workflows/compiler_sanitizers.yml
mayeut: I've never used them myself before and never saw them either. They come from a grep on ASAN_OPTIONS on the numpy code base which lead me to the following line:
https://github.com/numpy/numpy/blob/089ceb74a8d7c7e96c34dc01d930693d90e6b0f5/pixi-packages/asan/pixi.toml#L16

Given it seemed to be used for the linux-x64 build in their, I've copied the same options in the new job.

.github/workflows/compiler_sanitizers.yml
mayeut: The line numbers in the suppressions file are just illustrative (the same goes for numpy line numbers in there).
They are not used to determine suppressions, it's just a comment.
I used the same major.minor minus free-threading as what's found in https://github.com/nascheme/cpython_sanity (although cpython 3.14 images have not been rebuilt in a while so stuck to 3.14.0 for now).
I see both pros and cons in pinning python to an exact release. In the pros, obviously stability of the job, it won't fail because a new cpython version is available. On the other hand, I think it would be best to test with the latest patch version as is done in most of the other jobs.

.github/workflows/compiler_sanitizers.yml
mayeut: comment added.

> This looks a bit questionable, if acceptable.

What kind of issue are you worried about ?

ngoldbaum: I don't want to speak for Ralf, but I guess just depending on upstream binaries instead of system binaries is a little odd. Maybe also the security implications of downloading and executing binaries like this?

ngoldbaum:
Let's bring this in and see if it is noisy or, even better, starts telling us when we write memory leaks.

Thanks so much for the contribution @mayeut and also for your work on manylinux.

.github/workflows/compiler_sanitizers.yml
nascheme: I'm in the process of building images for 3.14.3.  They will also be built automatically when there is a new release, which will be nice.

.github/workflows/compiler_sanitizers.yml
rgommers: Yes, just a general security principle. The more random downloads there are, with no pinning or hashes, the wider the trust circle. If it's the best option for right now, then sure it's not a major issue. But when we can replace it with trusting Ubuntu packages, we should.

mayeut: I don't think this could be considered as a major supply chain security risk given the source (not so random, upstream LLVM repos) and the fact that's this is only a test job which does not expose sensitive secrets and does not produce any artefacts. It could have been better by manually adding the repo and the public key rather than an unchecked script download but I did not considered the risk enough to warrant for the extra complexity.

I agree however that this can count as an exemple of bad habit regarding supply chain security risk.

@nascheme published some new images including a GIL-enabled ASAN python build. See https://github.com/numpy/numpy/pull/30905 for an update of this job. This moves the supply chain security risk to what's been accepted for the TSAN job.

rgommers: Yep, agreed with all your thoughts - thanks for the follow-up!

nascheme: I wonder if would be good if I change my docker images to use Ubuntu 25 rather than 24.  The 25 release comes with Clang 21 as an Ubuntu package.  For testing with TSAN or ASAN, perhaps it's better to be running on the newest stable OS version that's supported.

rgommers: That does seem useful indeed, newer Clang versions can help. But no hurry to change it right now on account of NumPy.

PR #30803, opened 2026-02-08
ENH: ``numpy.where``: don't silently truncate Python scalars

### LINKED ISSUE #27006: `where` with out-of-bounds Python integers in Numpy 2
With Numpy 2 I've observed that `where` seems to behave in a (now strange) way with out-of-bounds Python integers compared to other functions.

The first argument of `where` does not participate in type promotion, so if using a Python integer as `x1` or `x2` and an array as the other argument, the output should be an array with the data type of the Numpy array argument.

However, unlike every ufunc with binary type promotion, `where` permits an out-of-bounds Python integer.

For example,
```
In [24]: np.where(np.asarray([True, False], dtype="?"), -1, np.asarray(0, dtype="u1"))
Out[24]: array([255,   0], dtype=uint8)
```

where for any other function
```
In [28]: np.add(np.asarray(0, dtype="u1"), -1)
OverflowError                             Traceback (most recent call last)
Cell In[28], line 1
----> 1 np.add(np.asarray(0, dtype="u1"), -1)

OverflowError: Python integer -1 out of bounds for uint8

In [29]: np.clip(np.asarray(0, dtype=np.uint8), -1, np.asarray(0, dtype=np.uint8))
OverflowError                             Traceback (most recent call last)
...
OverflowError: Python integer -1 out of bounds for uint8
```

Given that not even assignment now permits out-of-bounds Python integers, it seems like an oversight.

SpecLad:
Since NumPy 2.0, code like the following will cause a scalar value to be truncated:

    >>> np.where([False], np.array([1], dtype=np.uint8), 1000)
    array([232], dtype=uint8)

(in NumPy 1.x, it would instead output a uint16 array)

This is dangerous, and inconsistent with other NumPy functions, e.g.:

    >>> np.multiply(np.array([1], dtype=np.uint8), 1000)
        np.multiply(np.array([1], dtype=np.uint8), 1000)
        ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    OverflowError: Python integer 1000 out of bounds for uint8

Use `npy_update_operand_for_scalar` to bring the behavior of `where` in line with other functions.

Closes #27006.

<!--         ----------------------------------------------------------------
                MAKE SURE YOUR PR GETS THE ATTENTION IT DESERVES!
                ----------------------------------------------------------------

*  FORMAT IT RIGHT:
      https://www.numpy.org/devdocs/dev/development_workflow.html#writing-the-commit-message

*  IF IT'S A NEW FEATURE OR API CHANGE, TEST THE WATERS:
      https://www.numpy.org/devdocs/dev/development_workflow.html#get-the-mailing-list-s-opinion

*  HIT ALL THE GUIDELINES:
      https://numpy.org/devdocs/dev/index.html#guidelines

*  WHAT TO DO IF WE HAVEN'T GOTTEN BACK TO YOU:
      https://www.numpy.org/devdocs/dev/development_workflow.html#getting-your-pr-reviewed
-->

### DIFF
--- a/doc/release/upcoming_changes/30803.compatibility.rst
+++ b/doc/release/upcoming_changes/30803.compatibility.rst
@@ -1,0 +1,8 @@
+``numpy.where`` no longer truncates Python integers
+---------------------------------------------------
+
+Previously, if the ``x`` or ``y`` argument of ``numpy.where`` was a Python integer
+that was out of range of the output type, it would be silently truncated.
+Now, an `OverflowError` will be raised instead.
+
+This change also applies to the underlying C API function ``PyArray_Where``.
--- a/numpy/_core/src/multiarray/multiarraymodule.c
+++ b/numpy/_core/src/multiarray/multiarraymodule.c
@@ -3275,6 +3275,21 @@ PyArray_Where(PyObject *condition, PyObject *x, PyObject *y)
     if (common_dt == NULL) {
         goto fail;
     }
+
+    if (PyArray_FLAGS(ax) & NPY_ARRAY_WAS_PYTHON_LITERAL) {
+        if (npy_update_operand_for_scalar(&ax, x, common_dt, NPY_SAFE_CASTING) < 0) {
+            goto fail;
+        }
+        op_in[2] = ax;
+    }
+
+    if (PyArray_FLAGS(ay) & NPY_ARRAY_WAS_PYTHON_LITERAL) {
+        if (npy_update_operand_for_scalar(&ay, y, common_dt, NPY_SAFE_CASTING) < 0) {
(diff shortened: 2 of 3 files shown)


ngoldbaum:
Ping @seberg since this is related to scalar promotion changes.

seberg (approved):
Thanks, sorry for not looking at it closer earlier.  I hadn't realized that we already fixed the promotion part, but just the error was missing.

LGTM (Staring at this now, I am wondering if we may want to push this into the iterator even, but not sure that is good and it isn't relevant here).

PR #30807, opened 2026-02-09
MAINT: typo 'seperators' to 'separators' as part of variable name

thecaptain789:
Fix typo in variable name:  ->


ngoldbaum:
This is vendored code that shouldn't be updated like this. Closing.

PR #30808, opened 2026-02-09
BUG: Fix f2py OpenMP support with Meson backend

abslock128w:
Fixes #30804
**Problem**
On Python 3.12+, f2py with OpenMP-enabled Fortran code fails in two ways:

1.Import error: ``ImportError: Library not loaded: @rpath/libgomp.1.dylib``
2.The Meson backend doesn't configure OpenMP dependencies or runtime library paths correctly.

**This PR**

1.Auto-detects OpenMP flags (``-fopenmp``, ``-qopenmp``) in compiler arguments
2.Adds OpenMP dependency with fallback when Meson can't find it automatically
3.Configures ``install_rpath`` to include OpenMP library directories
4.Supports both GCC (``libgomp``) and Intel (``libiomp5``) compilers

**Changea**
``numpy/f2py/_backends/_meson.py:`` Added OpenMP detection, dependency fallback, and rpath configuration.
``numpy/f2py/_backends/meson.build.template:``Added ``${openmp_declaration} and ${rpath}`` substitution variables.

### DIFF
--- a/numpy/f2py/_backends/_meson.py
+++ b/numpy/f2py/_backends/_meson.py
@@ -47,6 +47,11 @@ class MesonTemplate:
             f"'{x}'" if not (x.startswith("'") and x.endswith("'")) else x
             for x in fortran_args
         ]
+               self.has_openmp = self._detect_openmp(fortran_args)
+        self.openmp_lib_dir = None
+        if self.has_openmp:
+            self.openmp_lib_dir = self._find_openmp_library()
+
         self.pipeline = [
             self.initialize_template,
             self.sources_substitution,
@@ -55,10 +60,65 @@ class MesonTemplate:
             self.include_substitution,
             self.libraries_substitution,
             self.fortran_args_substitution,
-        ]
+            self.rpath_substitution,  # NEW: Add rpath handling
+        ]    
         self.build_type = build_type
         self.python_exe = python_exe
         self.indent = " " * 21
+    
+    def _detect_openmp(self, fortran_args: list[str]) -> bool:
+        """Detect if OpenMP flags are present in fortran_args"""
+        openmp_flags = ['-fopenmp', '-qopenmp', '-openmp', '/Qopenmp']
+        return any(flag in fortran_args for flag in openmp_flags)
+    
+    def _find_openmp_library(self) -> Path | None:
+        """Try to find the OpenMP library directory from the compiler"""
+        try:
+            # Determine which compiler we're using
+            fc = os.environ.get('FC', 'gfortran')
+            
+            # Try to get library path from compiler
+            if 'gfortran' in fc or 'gcc' in fc:
+                # For GCC/gfortran, find libgomp
+                result = subprocess.run(
+                    [fc, '--print-file-name=libgomp.so'],
+                    capture_output=True,
+                    text=True,
+                    check=False
+                )
+                if result.returncode == 0 and result.stdout.strip():
+                    lib_path = Path(result.stdout.strip())
+                    if lib_path.exists():
+                        return lib_path.parent
+                
+                # Try macOS dylib
+                result = subprocess.run(
+                    [fc, '--print-file-name=libgomp.dylib'],
+                    capture_output=True,
+                    text=True,
+                    check=False
+                )
+                if result.returncode == 0 and result.stdout.strip():
+                    lib_path = Path(result.stdout.strip())
+                    if lib_path.exists():
+                        return lib_path.parent
+                        
+            elif 'ifort' in fc or 'ifx' in fc:
+                # For Intel Fortran, find libiomp5
+                result = subprocess.run(
+                    [fc, '--print-file-name=libiomp5.so'],
+                    capture_output=True,
+                    text=True,
+                    check=False
+                )
+                if result.returncode == 0 and result.stdout.strip():
+                    lib_path = Path(result.stdout.strip())
+                    if lib_path.exists():
+                        return lib_path.parent
+        except Exception:
+            pass
+        return None
+        
 
     def meson_build_template(self) -> str:
         if not self.build_template_path.is_file():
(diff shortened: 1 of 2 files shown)


abslock128w:
The commit shud fix both GCC and Intel compiler OpenMP issues. @AgilentGCMS - could you please test with Intel ifx compiler to confirm the Intel linking issue is resolved?"

numpy/f2py/_backends/_meson.py
```diff
@@ -55,10 +60,66 @@ def __init__(
...
+    def _detect_openmp(self, fortran_args: list[str]) -> bool:
+        """Detect if OpenMP flags are present in fortran_args"""
+        openmp_flags = ['-fopenmp', '-qopenmp', '-openmp', '/Qopenmp']
+        return any(flag in fortran_args for flag in openmp_flags)
+    
+    def _find_openmp_library(self) -> Path | None:
+        """Try to find the OpenMP library directory from the compiler"""
+        try:
+            # Determine which compiler we're using
+            fc = os.environ.get('FC', 'gfortran')
```
tylerjereddy, line 79: I suppose this environment variable is indeed common to use, although this seems potentially brittle?

numpy/f2py/_backends/_meson.py
```diff
@@ -55,10 +60,66 @@ def __init__(
...
+                )
+                if result.returncode == 0 and result.stdout.strip():
+                    lib_path = Path(result.stdout.strip())
+                    if lib_path.exists():
+                        return lib_path.parent
+                        
+            elif 'ifort' in fc or 'ifx' in fc:
+                # For Intel Fortran, find libiomp5
+                result = subprocess.run(
+                    [fc, '--print-file-name=libiomp5.so'],
```
tylerjereddy, line 110: Do we really want to look for a specific "version" of the shared object like this re: portability/durability?

tylerjereddy (commented):
I think @HaoZeke started commenting on `f2py`-related things recently again so they're likely the best person to review.

I suppose `f2py` is rather separate from the rest of NumPy in various regards, though I should point out that core NumPy does, as of recently, have some `meson`-related OpenMP-handling infrastructure in `numpy/_core/meson.build`. Since that is for C or C++ (I think..) and the main library, I suppose there may not be much potential infrastructure overlap, but did want to point that out just in case it matters during review. I don't know if i.e., one would want to discourage using a separate OpenMP "version"/"type" for main lib vs. `f2py` (probably "ok" to be separate since they're quite separate conceptually?).

numpy/f2py/_backends/_meson.py
```diff
@@ -55,10 +60,66 @@ def __init__(
...
+                    text=True,
+                    check=False
+                )
+                if result.returncode == 0 and result.stdout.strip():
+                    lib_path = Path(result.stdout.strip())
+                    if lib_path.exists():
+                        return lib_path.parent
+        except Exception:
+            pass
+        return None
```
HaoZeke, line 121: None of this should be required, if `meson` has a bug finding `openmp` for `fortran` compiled projects (doubtful) then that should be handled upstream..

numpy/f2py/_backends/_meson.py
```diff
@@ -86,9 +147,77 @@ def objects_substitution(self) -> None:
...
+        link_args.append(f"'-l{openmp_lib}'")
+
+        if is_intel and compile_args:
+            link_args.extend(compile_args)
+
+        lines.append(f"    link_args: [{', '.join(link_args)}],")
+        lines.append("  )")
+        lines.append("endif")
+
+        return "\n".join(lines)
```
HaoZeke, line 220: Also seems like a local environment issue, or a bug with `meson`. Afterall, with the right `pkgconf` setup a normal `meson.build` with OpenMP should work..

numpy/f2py/_backends/_meson.py
```diff
@@ -125,6 +254,36 @@ def fortran_args_substitution(self) -> None:
...
+        """Force Fortran as link language when OpenMP is used with Intel compilers"""
+        # Intel compilers need Fortran linker for OpenMP, not C linker
+        is_intel = any('ifort' in str(arg) or 'ifx' in str(arg) or 'qopenmp' in str(arg).lower() 
+                       for arg in self.fortran_args)
+
+        if self.has_openmp and is_intel:
+            self.substitutions["override_options"] = f"{self.indent}override_options: ['fortran_link_language=fortran'],"
+        else:
+            self.substitutions["override_options"] = ""
+
```
HaoZeke, line 286: Definitely not a good idea.

HaoZeke (changes requested):
Please open an issue with a reproducer first, and then we'll need to debug if this is a local environment issue or indeed something wrong with `meson` before trying to patch `f2py` for this. The main reason to switch to `meson` was to not handle these compiler specific issues.

numpy/f2py/_backends/_meson.py
AgilentGCMS: Since meson build doesn't support the `--compiler` and `--fcompiler` options of distutils, setting `FC`, `CC`, and `CXX` is, AFAIK, the _only_ way to specify a compiler. So anyone who is building with meson would _have_ to set one or more of those environment variables. So I would say this is robust, at least as robust as meson is.

numpy/f2py/_backends/_meson.py
AgilentGCMS: `libiomp5.so` has been [Intel's OpenMP library](https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2023-1/openmp-support-libraries.html#GUID-4B60BCA6-1948-4EB0-A9F3-D56AA1C683F7) since 2015, so I think it's fairly stable. Although one shouldn't *need* to link explicitly with `-liomp5` if one compiles with `-qopenmp`, for some reason I've found that `f2py` always needs that explicit link line.

AgilentGCMS:
> Please open an issue with a reproducer first, and then we'll need to debug if this is a local environment issue or indeed something wrong with `meson` before trying to patch `f2py` for this. The main reason to switch to `meson` was to not handle these compiler specific issues.

What's the next step? I have access to Intel compilers on multiple platforms, and meson-building f2py with openmp directives has never worked on any of them. So I am willing to test changes if needed.

In general I question the wisdom of hiding everything with `meson`, which was the supposed purpose of switching. Most people who use `numpy` don't muck with `f2py`, and anyone who _wants_ to wrap Fortran code with `f2py` is likely to know the compiler settings that work on their specific platform. In this respect I find distutil's _possibility_ of specifying compilers and flags much more flexible and easy to use. Automatic is fine as long as there is a manual override.

numpy/f2py/_backends/_meson.py
AgilentGCMS: I would suspect a `meson` bug. Building with `distutils` for python 3.11.x in the same local (compiler) environment works.

AgilentGCMS:
> The commit shud fix both GCC and Intel compiler OpenMP issues. @AgilentGCMS - could you please test with Intel ifx compiler to confirm the Intel linking issue is resolved?"

It isn't. I've posted the errors in #30804

abslock128w:
> > The commit shud fix both GCC and Intel compiler OpenMP issues. @AgilentGCMS - could you please test with Intel ifx compiler to confirm the Intel linking issue is resolved?"
> 
> It isn't. I've posted the errors in #30804

interesting will look into it...
but again this probably isint a PR i wanna focus on since it could just be a meson bug,
and also the whole point of meson is prolly to avoid having to maintain compiler specific workarounds so...
but ill see to it!

numpy/f2py/_backends/_meson.py
abslock128w: true that hmm
this duplicates meson's job
meson shud automatically do with openmp
and shud do the same even for the rpath

abslock128w: > true that hmm 
this duplicates meson's job 
meson shud automatically work with openmp

charris:
@abslock128w I'm going to close this. Has there been any resolution?

PR #30811, opened 2026-02-10
DEP: Add `VisibleDeprecationWarning` for `intent(inplace)` usage

### LINKED ISSUE #30809: DEP: Deprecate intent(inplace) with a VisibleDeprecationWarning due to user-facing change
gh-29929 changed `intent(inplace)` in a user-facing way, because the old way did *very* bad things with NumPy array internals, meaning that user data could easily be corrupted.

But it was also a possible subtle behavior change, which isn't great either.  It would be good to add a `VisibleDeprecationWarning` on it (it can be around for a long time):
* For users, it will inform them that behavior may have changed. They should make sure to pass in an array of the correct dtype.
* For maintainers, it should say that they should (eventually) migrate to `intent(in, out)`.

From the user-point of view, at some point `intent(inplace)` could just be the same behavior as `intent(in, out)` (after which the warning could shift to just deprecate `intent(inplace)` at code generation time).

abslock128w:
Fixes #30809

## Problem

Pull request gh-29929 changed ``intent(inplace)` behavior to fix unsafe array manipulation that could corrupt user data. The old implementation used `swap_arrays()` which directly modified NumPy array internals, silently changing dtypes when incorrect types were passed.

This fix lacked user notification about the behavior change.

## This PR

Adds a `VisibleDeprecationWarning` when `intent(inplace)` is used at runtime, informing users to:
- Verify arrays have correct dtype
- Migrate to `intent(in, out)` for explicit behavior

## Changes

- **numpy/f2py/src/fortranobject.c**: Added deprecation check in `array_from_pyobj()` after alignment validation (~line 1015). Emits `VisibleDeprecationWarning` when `F2PY_INTENT_INPLACE` flag is detected.


numpy/f2py/src/fortranobject.c
```diff
@@ -988,6 +1012,19 @@ ndarray_from_pyobj(const int type_num,
...
+                    "intent(inplace) behavior has changed in NumPy and may be "
+                    "deprecated in a future release. The previous implementation "
+                    "was unsafe and could corrupt data. Ensure your arrays have "
+                    "the correct dtype, or migrate to intent(in, out) for "
+                    "explicit behavior. See gh-30809 for details.",
+                    1) < 0) {
+                    Py_DECREF(descr);
+                    return NULL;
+                }
+            }
```
seberg: Please move it inside and add a test I would assume that the current tests even fail if run locally. Also why did you change the `meson.build.template` (please revert that)?

A test should test both the error and the warning (that is turn the warning into an error).

I don't want to deprecate this *always*, just when the cast is needed also.  The message can focus on that part (but I suspect we have to iterate once more after moving+tests anyway).

abslock128w: Hey there thanks for pointing it out!
will make the suggested changes.

jorenham:
Please ensure that the test pass locally before committing your changes. Then, after CI turns green, and only after that, will we be able to review.

abslock128w:
> Please ensure that the test pass locally before committing your changes. Then, after CI turns green, and only after that, will we be able to review.

my bad i totally forgot about this 
will run test and fix the CI asap

jorenham:
CI is still failing :sweat_smile:

abslock128w:
Nah nah I made some changes by mistake 🥹(indentation error).. so reverted all of it back
Will Fix it...

abslock128w:
hey there  @jorenham 
for some odd reason i am unable to build numpy on my laptop i have been trying to do so idk whats wrong
either there's a bug with my laptop or environment issue or idk=_=
imma try smth else

charris:
@abslock128w I'm going to close this. Feel free to reopen if you intend to pursue this.

Issue #30813, opened 2026-02-10
[ENH] Remove remnants of Python 2 buffer protocol from NumPy internals


ngoldbaum:
I'm noticing this because I'm touching the NumPy headers to fix #30704. That includes touching `PyArray_Chunk`.

It took me quite a while to understand what `PyArray_Chunk` and `PyArray_BufferConverter` are doing, because they're based on the *very* old "single-segment" buffer protocol. See [PEP 3118](https://peps.python.org/pep-3118/) for more historical details.

The docs for `PyArray_FromBuffer` also talk about a single-segment buffer protocol, but that function has supported the "new" buffer protocol for a very long time.

`PyArray_BufferConverter` is publicly exposed, it's also used inside `array_new`, so it is somewhat load-bearing unfortunately.

As far as I can see, `PyArray_BufferConverter` has zero uses outside NumPy in public code on Github outside of copies of NumPy or wrappers around the entire NumPy C API: https://github.com/search?q=PyArray_BufferConverter&type=code&p=2. Maybe we can just deprecate the public version, along with `PyArray_Chunk`?

We should also update the docs do no longer talk about the Python 2 buffer protocol.

Possibly related issues:

https://github.com/numpy/numpy/issues/23305
https://github.com/numpy/numpy/issues/15450

charris:
That sounds good to me. If there is a problem, now would be when to find it.

seberg:
This function is used for reinterpreting buffer imports that ignore the original type (frombuffer/new constructor). Talking about the old buffer protocol is wrong of course...

But I am OK to deprecate it. It is unlikely NumPy users need it and non-NumPy users should use the buffer protocol directly.

sharmavikas18:
I would like to work on this issue.

PR #30814, opened 2026-02-10
API: Make PyArray_Chunk opaque on internal builds

ngoldbaum:
I didn't expose any accessors because I'm not sure this struct should be or needs to be public at all. c.f. #30813.

### DIFF
--- a/numpy/_core/include/numpy/ndarraytypes.h
+++ b/numpy/_core/include/numpy/ndarraytypes.h
@@ -855,7 +855,13 @@ typedef struct {
         void *ptr;
         npy_intp len;
         int flags;
-} PyArray_Chunk;
+} PyArray_Chunk_fields;
+
+#if defined(NPY_INTERNAL_BUILD) && (NPY_INTERNAL_BUILD)
+typedef struct PyArray_Chunk PyArray_Chunk;
+#else
+typedef PyArray_Chunk_fields PyArray_Chunk;
+#endif
 
 typedef struct {
     NPY_DATETIMEUNIT base;
--- a/numpy/_core/src/multiarray/arrayobject.c
+++ b/numpy/_core/src/multiarray/arrayobject.c
@@ -1113,7 +1113,7 @@ array_new(PyTypeObject *subtype, PyObject *args, PyObject *kwds)
     int itemsize;
     PyArray_Dims dims = {NULL, 0};
     PyArray_Dims strides = {NULL, -1};
-    PyArray_Chunk buffer;
+    PyArray_Chunk_fields buffer;
     npy_longlong offset = 0;
     NPY_ORDER order = NPY_CORDER;
     int is_f_order = 0;
--- a/numpy/_core/src/multiarray/conversion_utils.c
+++ b/numpy/_core/src/multiarray/conversion_utils.c
@@ -297,9 +297,11 @@ PyArray_BufferConverter(PyObject *obj, PyArray_Chunk *buf)
 {
     Py_buffer view;
 
-    buf->ptr = NULL;
-    buf->flags = NPY_ARRAY_BEHAVED;
-    buf->base = NULL;
+    PyArray_Chunk_fields *_buf = (PyArray_Chunk_fields *)buf;
+
+    _buf->ptr = NULL;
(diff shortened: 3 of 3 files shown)


charris (approved):

numpy/_core/include/numpy/ndarraytypes.h
```diff
@@ -855,7 +855,13 @@ typedef struct {
...
         npy_intp len;
         int flags;
-} PyArray_Chunk;
+} PyArray_Chunk_fields;
+
+#if defined(NPY_INTERNAL_BUILD) && (NPY_INTERNAL_BUILD)
+typedef struct PyArray_Chunk PyArray_Chunk;
+#else
+typedef PyArray_Chunk_fields PyArray_Chunk;
+#endif
```
seberg, line 864: Sorry, to some degree, I want to just keep going here (or the other PRs), but I a bit unsure what the aim is.

If the aim is to be able to compile downstream with the new stable ABI that is missing `PyObject_HEAD`, I would tend to simply not define `PyArray_Chunk` if `PyObject_HEAD` isn't defined. (Or I guess define it to an `#error "harharhar PyArray_BufferConvert can't be used with this build"`)

If the goal is to compile NumPy with this (I am doubtful about this mid-term, but OK with preparing it long term), then we need a version of `_fields` that doesn't include `PyObject_HEAD`.
(Of course the alternative to just preparing is to put a deprecation warning now and making it "error only" in a year or so.)

seberg: (But maybe this `_fields` definition is a way to prepare for all the `_fields` to omit `PyObject_HEAD` in a follow-up?)


EDIT: And to be clear, I am also OK with just saying "carry on" even if this is all a bit fuzzy but feels to you like it will help!

ngoldbaum: > And to be clear, I am also OK with just saying "carry on" even if this is all a bit fuzzy but feels to you like it will help!

I appreciate it!

Keep in mind that all of this work is in support of end-to-end tests of PEP 803 and is not intended to be mergeable.

For this PR I was mostly just trying to understand what the heck `PyArray_Chunk` even is, given the poor docs and that its based on concepts that were removed in Python 3 from the Python C API. I agree that it probably makes sense just not to expose this struct and `PyArray_BufferConverter` in an opaque pyobject limited API, should such a thing happen in reality.

FWIW, my plan (roughly) is:

* Make NumPy compilable and runnable with all PyObject-extending structs opaque in internal builds, adding accessors and setters as needed. This assumes that NumPy itself is a reasonably complete consumer of the NumPy C API.
* Octopus merge all the PRs
* Change all the NPY_INTERNAL_BUILD macros I added in the various draft PRs I have open to check `_Py_OPAQUE_PYOBJECT` instead. Possibly revert all the changes to numpy internals if we're worried about performance overhead of static inline functions.
    * Note that I tried testing this on Windows but got inconclusive, random benchmark results on some of the ufunc benchmarks and a bunch of the `bench_ufunc_strides` benchmarks doing things like comparing a commit with a no-op commit that has the same content.
* Actually try building a real-world Cython project that depends on NumPy using the opaque pyobject ABI.

Also, @kumaraditya303 is going to help out with this and may take over for me completely.

ngoldbaum: I missed two important steps above:

* Ifdef-out the `PyObject_HEAD` part of the fields structs on opaque pyobject builds.
* Add C API functions that return pointers to the field structs, usable on the opaque pyobject build and add ifdefs for all the field accessor inline functions to call the API function on opaque pyobject builds.

seberg: OK, cool. To me starting with the changes in NumPy internals (like these here) doesn't seem very necessary for the sake of making the headers `_Py_OPAQUE_PYOBJECT` compatible, but not complaining.
(Although, yeah, that testing all offset functions/macros 100% might be a bit hard without a pretty complete user as a basis.)

funtimefranky1530-coder (commented):
**ok**

ngoldbaum:

PR #30819, opened 2026-02-11
CI/BUG: add native jobs for s390x, fix bug in `pack_inner`

### LINKED ISSUE #30817: CI: use IBM self-hosted runner for native s390x tests
s390x runners are also available at https://github.com/IBM/actionspz. We would like to use them as well to get rid of cross-compiling and running tests under qemu for s390x too. I could raise a PR to incorporate the required workflow adjustments when s390x runners are added.

@mkumatag @rgommers

AlekseiNikiforovIBM:
Rename workflow linux-ppc64le.yml into linux-actionspz.yml.
Copy and update ppc64le workflow for s390x.
Disable qemu-based workflows for s390x.

Fix big endian issue in `pack_inner`.

Closes: https://github.com/numpy/numpy/issues/30817

@mkumatag @rgommers

### DIFF
--- a/.github/workflows/linux-ppc64le.yml
+++ b/.github/workflows/linux-ppc64le.yml
@@ -75,9 +75,9 @@ jobs:
     - name: Install clang
       if: matrix.config.compiler == 'clang'
       run: |
-        sudo apt install -y clang
-        echo CC=clang >> $GITHUB_ENV
-        echo CXX=clang++ >> $GITHUB_ENV
+        sudo apt install -y clang-20
+        echo CC=clang-20 >> $GITHUB_ENV
+        echo CXX=clang++-20 >> $GITHUB_ENV
 
     - name: Meson Build
       run: |
--- a/.github/workflows/linux-ppc64le.yml
+++ b/.github/workflows/linux-ppc64le.yml
@@ -1,4 +1,4 @@
-name: Native ppc64le Linux Test
+name: Native ppc64le and s390x Linux Tests
 
 on:
   pull_request:
@@ -20,21 +20,41 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  native_ppc64le:
-    # This job runs only in the main NumPy repository.
-    # It requires a native ppc64le GHA runner, which is not available on forks.
+  native_ppc64le_and_s390x:
+    # These jobs runs only in the main NumPy repository.
+    # It requires a native ppc64le and s390x GHA runners, which are not available on forks.
     # For more details, see: https://github.com/numpy/numpy/issues/29125
     if: github.repository == 'numpy/numpy'
-    runs-on: ubuntu-24.04-ppc64le-p10
+    runs-on: ${{ matrix.config.runner }}
 
     strategy:
       fail-fast: false
       matrix:
         config:
-          - name: "GCC"
+          - name: "GCC (ppc64le)"
             args: "-Dallow-noblas=false"
-          - name: "clang"
+            runner: ubuntu-24.04-ppc64le-p10
+            compiler: "gcc"
+          - name: "clang (ppc64le)"
             args: "-Dallow-noblas=false"
+            runner: ubuntu-24.04-ppc64le-p10
+            compiler: "clang"
+          - name: "GCC (s390x)"
+            args: "-Dallow-noblas=false"
+            runner: ubuntu-24.04-s390x
+            compiler: "gcc"
+          - name: "clang (s390x)"
+            args: "-Dallow-noblas=false"
+            runner: ubuntu-24.04-s390x
+            compiler: "clang"
+          - name: "GCC with zvector (s390x)"
+            args: "-Dallow-noblas=false -Dcpu-baseline=vx"
+            runner: ubuntu-24.04-s390x
+            compiler: "gcc"
+          - name: "clang with zvector (s390x)"
+            args: "-Dallow-noblas=false -Dcpu-baseline=vx"
+            runner: ubuntu-24.04-s390x
+            compiler: "clang"
 
     name: "${{ matrix.config.name }}"
     steps:
@@ -53,7 +73,7 @@ jobs:
         echo "/home/runner/.local/bin" >> $GITHUB_PATH
 
     - name: Install clang
-      if: matrix.config.name == 'clang'
+      if: matrix.config.compiler == 'clang'
       run: |
         sudo apt install -y clang
         export CC=clang
--- a/.github/workflows/linux-ppc64le.yml
+++ b/.github/workflows/linux-ppc64le.yml
@@ -76,8 +76,8 @@ jobs:
       if: matrix.config.compiler == 'clang'
       run: |
         sudo apt install -y clang
-        export CC=clang
-        export CXX=clang++
+        echo CC=clang >> $GITHUB_ENV
+        echo CXX=clang++ >> $GITHUB_ENV
 
     - name: Meson Build
       run: |
--- a/.github/workflows/linux-ppc64le.yml
+++ b/.github/workflows/linux-ppc64le.yml
@@ -38,7 +38,7 @@ jobs:
             args: "-Dallow-noblas=false"
             runner: ubuntu-24.04-s390x
           - name: "GCC with zvector (s390x)"
-            args: "-Dallow-noblas=false -Dcpu-baseline=vx"
+            args: "-Dallow-noblas=false -Dcpu-baseline=vxe2"
             runner: ubuntu-24.04-s390x
 
     name: "${{ matrix.config.name }}"
@@ -84,7 +84,7 @@ jobs:
             args: "-Dallow-noblas=false"
             runner: ubuntu-24.04-s390x
           - name: "clang with zvector (s390x)"
-            args: "-Dallow-noblas=false -Dcpu-baseline=vx"
+            args: "-Dallow-noblas=false -Dcpu-baseline=vxe2"
             runner: ubuntu-24.04-s390x
 
     name: "${{ matrix.config.name }}"
--- a/.github/workflows/linux-ppc64le.yml
+++ b/.github/workflows/linux-ppc64le.yml
@@ -1,4 +1,4 @@
-name: Native ppc64le and s390x Linux Tests
+name: Linux ppc64le and s390x native tests
 
 on:
   pull_request:
--- a/.github/workflows/linux-ppc64le.yml
+++ b/.github/workflows/linux-ppc64le.yml
@@ -43,18 +43,10 @@ jobs:
             args: "-Dallow-noblas=false"
             runner: ubuntu-24.04-s390x
             compiler: "gcc"
-          - name: "clang (s390x)"
-            args: "-Dallow-noblas=false"
-            runner: ubuntu-24.04-s390x
-            compiler: "clang"
           - name: "GCC with zvector (s390x)"
             args: "-Dallow-noblas=false -Dcpu-baseline=vx"
             runner: ubuntu-24.04-s390x
             compiler: "gcc"
-          - name: "clang with zvector (s390x)"
-            args: "-Dallow-noblas=false -Dcpu-baseline=vx"
-            runner: ubuntu-24.04-s390x
-            compiler: "clang"
 
     name: "${{ matrix.config.name }}"
     steps:
--- a/.github/workflows/linux-ppc64le.yml
+++ b/.github/workflows/linux-ppc64le.yml
@@ -34,19 +34,12 @@ jobs:
           - name: "GCC (ppc64le)"
             args: "-Dallow-noblas=false"
             runner: ubuntu-24.04-ppc64le-p10
-            compiler: "gcc"
-          - name: "clang (ppc64le)"
-            args: "-Dallow-noblas=false"
-            runner: ubuntu-24.04-ppc64le-p10
-            compiler: "clang"
           - name: "GCC (s390x)"
             args: "-Dallow-noblas=false"
             runner: ubuntu-24.04-s390x
-            compiler: "gcc"
           - name: "GCC with zvector (s390x)"
             args: "-Dallow-noblas=false -Dcpu-baseline=vx"
             runner: ubuntu-24.04-s390x
-            compiler: "gcc"
 
     name: "${{ matrix.config.name }}"
     steps:
@@ -64,13 +57,6 @@ jobs:
         pip install -r requirements/build_requirements.txt -r requirements/test_requirements.txt
         echo "/home/runner/.local/bin" >> $GITHUB_PATH
 
-    - name: Install clang
-      if: matrix.config.compiler == 'clang'
-      run: |
-        sudo apt install -y clang-20
-        echo CC=clang-20 >> $GITHUB_ENV
-        echo CXX=clang++-20 >> $GITHUB_ENV
-
     - name: Meson Build
       run: |
         spin build -- ${{ matrix.config.args }}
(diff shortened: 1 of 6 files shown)


mkumatag:
@sandeepgupta12 ptal

AlekseiNikiforovIBM:
So, currently clang tests are failing on ppc64le and s390x, and clang and gcc tests with forced vectorization fail on s390x.

Can we proceed with these workflows now or should tests be fixed first? I'm already looking into s390x failures.

AlekseiNikiforovIBM:
This change fixes s390x tests with forced zvector:

https://github.com/numpy/numpy/pull/30819/changes/663107d270340ea4e5f2f1bfe9fcf66d3c6939fb

I could also split it into a separate PR if you'd like.

Now there are only some failures when clang is used.

.github/workflows/linux-ppc64le.yml
```diff
@@ -0,0 +1,127 @@
...
+    # It requires a native ppc64le GHA runner, which is not available on forks.
+    # For more details, see: https://github.com/numpy/numpy/issues/29125
+    if: github.repository == 'numpy/numpy'
+    runs-on: ubuntu-24.04-ppc64le-p10
+
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - name: "GCC (ppc64le)"
```
rgommers: This whole job matrix is `ppc64le`, so this name change seem a bit unnecessary

.github/workflows/linux-actionspz.yml
```diff
@@ -0,0 +1,127 @@
...
+        pip install --upgrade pip
+        pip install -r requirements/build_requirements.txt -r requirements/test_requirements.txt
+        echo "/home/runner/.local/bin" >> $GITHUB_PATH
+
+    - name: Install clang
+      if: matrix.config.name == 'clang (ppc64le)'
+      run: |
+        sudo apt install -y clang
+        echo CC=clang >> $GITHUB_ENV
+        echo CXX=clang++ >> $GITHUB_ENV
```
rgommers: Good catch, `export` doesn't survive across job steps

.github/workflows/linux-ppc64le.yml
```diff
@@ -0,0 +1,127 @@
...
+      matrix:
+        config:
+          - name: "GCC (s390x)"
+            args: "-Dallow-noblas=false"
+          - name: "clang (s390x)"
+            args: "-Dallow-noblas=false"
+          - name: "GCC with zvector (s390x)"
+            args: "-Dallow-noblas=false -Dcpu-baseline=vx"
+          - name: "clang with zvector (s390x)"
+            args: "-Dallow-noblas=false -Dcpu-baseline=vx"
```
rgommers: We had two jobs under QEMU; four seems a little excessive - can you do `GCC` and `Clang with zvector` here (or vice versa?

rgommers (commented):
Thanks @AlekseiNikiforovIBM. This is a bit tricky to review (let's blame the GitHub UI), so it'd be great if you could revert the name change and leave it at `linux-ppc64le.yml` for now. That will make the diff much smaller, and it'll be visible what you tweaked in the ppc64le job.

rgommers:
> I could also split it into a separate PR if you'd like.

A single PR seems fine to me, thanks.

.github/workflows/linux-ppc64le.yml
AlekseiNikiforovIBM: It looks like it takes title of workflow file, and then job name. Without this change there will be two jobs named "Native ppc64le and s390x Linux Test / GCC", one for ppc64le and one for s390x.

.github/workflows/linux-ppc64le.yml
AlekseiNikiforovIBM: They produce different results, so I think it's better to have all 4 of them. Considering issues with clang, maybe clang jobs should be dropped until all issues when compiling with clang are resolved?

sandeepgupta12: With clang 18, ppc64le tests are passing. See here- https://github.com/numpy/numpy/actions/runs/21948675141/job/63393053952

AlekseiNikiforovIBM: I don't see clang-18 being used there. I see gcc. From linked build log:
```
C compiler for the host machine: cc (gcc 13.3.0 "cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0")
C linker for the host machine: cc ld.bfd 2.42
C++ compiler for the host machine: c++ (gcc 13.3.0 "c++ (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0")
C++ linker for the host machine: c++ ld.bfd 2.42
```

Clang-20 logs for comparison:
```
C compiler for the host machine: clang-20 (clang 20.1.2 "Ubuntu clang version 20.1.2 (0ubuntu1~24.04.2)")
C linker for the host machine: clang-20 ld.bfd 2.42
C++ compiler for the host machine: clang++-20 (clang 20.1.2 "Ubuntu clang version 20.1.2 (0ubuntu1~24.04.2)")
C++ linker for the host machine: clang++-20 ld.bfd 2.42
```

AlekseiNikiforovIBM:
I've disabled clang builds and tests for s390x while there are still too many issues to be fixed. I think enabling them later would be a good choice.

.github/workflows/linux-ppc64le.yml
```diff
@@ -53,11 +65,11 @@ jobs:
         echo "/home/runner/.local/bin" >> $GITHUB_PATH
 
     - name: Install clang
-      if: matrix.config.name == 'clang'
+      if: matrix.config.compiler == 'clang'
       run: |
-        sudo apt install -y clang
-        export CC=clang
-        export CXX=clang++
+        sudo apt install -y clang-20
```
sandeepgupta12: @AlekseiNikiforovIBM, could you please revert the “Install clang” change since this is being skipped for s390?

AlekseiNikiforovIBM:
I've updated PR.

I've used docker and fedora:43 image to build and test numpy with clang. Fedora has newer clang, and maybe clang is patched differently there, so there are no miscompilations detected by testsuite.

Only 1 test failed and was needed to be fixed: `test_cpu_features.py::Test_ZARCH_Features::test_features`.

`AT_HWCAP` is no longer decoded in glibc output:
https://sourceware.org/git/?p=glibc.git;a=commit;h=e5893e6349541d871e8a25120bca014551d13ff5
Now it's provided as a hex number:
```
$ LD_SHOW_AUXV=1 /bin/true | grep AT_HWCAP
AT_HWCAP:             0x67ffff
```
Luckily, on s390x a copy of output is available in `/proc/cpuinfo` and I've switched to using it:
```
$ grep feature /proc/cpuinfo 
features        : esan3 zarch stfle msa ldisp eimm dfp edat etf3eh highgprs te vx vxd vxe gs vxe2 vxp sort dflt pcimio sie 
```

Unfortunately, I don't think similar output is available for ppc64le. Hex output of `AT_HWCAP` would be needed to be decoded for ppc64le when it'd be detected, or some other approach would have to be used.

I've also bumped zvector build to enable everything up to VXE2.

There still were 65 failing tests for ppc64le when building with clang:
https://github.com/numpy/numpy/actions/runs/21990958320/job/63537634918?pr=30819
```
2026-02-13T14:52:11.1751562Z = 65 failed, 47875 passed, 344 skipped, 2843 deselected, 34 xfailed, 3 xpassed in 256.65s (0:04:16) =
```

I disabled ppc64le clang build for now. It'd be better to have those tests fixed or disabled first, or some other build environment used.

@rgommers please take one more look.

.github/workflows/linux-ppc64le.yml
```diff
@@ -1,4 +1,4 @@
-name: Native ppc64le Linux Test
+name: Linux ppc64le and s390x native tests
```
seiko2plus:
```suggestion
name: Linux IBM
```
No need for duplicates, along with the matrix label that holds CPU architecture + compiler name is quite enough.

.github/workflows/linux-ppc64le.yml
```diff
@@ -20,21 +20,26 @@ concurrency:
...
+          - name: "GCC (ppc64le)"
             args: "-Dallow-noblas=false"
-          - name: "clang"
+            runner: ubuntu-24.04-ppc64le-p10
+          - name: "GCC (s390x)"
             args: "-Dallow-noblas=false"
+            runner: ubuntu-24.04-s390x
+          - name: "GCC with zvector (s390x)"
+            args: "-Dallow-noblas=false -Dcpu-baseline=vxe2"
+            runner: ubuntu-24.04-s390x
```
seiko2plus:
```suggestion
          - name: "s390x/gcc - baseline(Z15/VXE2)"
            args: "-Dallow-noblas=false -Dcpu-baseline=vxe2"
            runner: ubuntu-24.04-s390x
```

`vxe2` belongs to z15, so it's better to be specific.

.github/workflows/linux-ppc64le.yml
```diff
@@ -70,3 +68,69 @@ jobs:
...
+    runs-on: ${{ matrix.config.runner }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - name: "clang (s390x)"
+            args: "-Dallow-noblas=false"
+            runner: ubuntu-24.04-s390x
+          - name: "clang with zvector (s390x)"
```
seiko2plus:
```suggestion
          - name: "s390x/clang - baseline(Z15/VXE2)"
```
similar to gcc

.github/workflows/linux-ppc64le.yml
```diff
@@ -70,3 +68,69 @@ jobs:
...
+    # It requires a native ppc64le and s390x GHA runners, which are not available on forks.
+    # For more details, see: https://github.com/numpy/numpy/issues/29125
+    if: github.repository == 'numpy/numpy'
+    runs-on: ${{ matrix.config.runner }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - name: "clang (s390x)"
```
seiko2plus:
```suggestion
          - name: "s390x/clang - baseline(default)"
```
On zystem and by default, NumPy doesn't enable zvector. We may need to change that—maybe raising the baseline to z12 CPU features? So the correct name for now should be:

```
          - name: "s390x/clang - baseline(default)"
```
or 
```
          - name: "s390x/clang - baseline(none)"
```

.github/workflows/linux-ppc64le.yml
seiko2plus:
```suggestion
         - name: "ppc64le/gcc - baseline(default)"
```
I agree with @AlekseiNikiforovIBM. We need to be specific about the job name since this workflow supports both architectures. I would like to be more specific about the baseline since we enabled multiple baseline configurations.

numpy/_core/src/multiarray/compiled_base.c
```diff
@@ -1632,6 +1652,11 @@ pack_inner(const char *inptr,
                 va = npyv_rev64_u8(va);
```
seiko2plus, line 1652: I think we can do better than just a fix. We can avoid using `npyv_rev64_u8` on bigendian if the required order is `PACK_ORDER_BIG`. So we reverse if the required order is `PACK_ORDER_BIG` and byte order is `NPY_LITTLE_ENDIAN`, and we swap if the byte order is `NPY_BIG_ENDIAN`.

seiko2plus (approved):
Thank you for the fix and extending the CI. The work is good to me. I left some comments below — not necessary to be accomplished, and they can be done in another PR. Regarding the SIMD optimization comment I left, it's for reference — it can be done within this PR or within another PR using the current C SIMD interface or during the transition of moving to Google Highway.

seiko2plus:
> Luckily, on s390x a copy of output is available in /proc/cpuinfo and I've switched to using it:

It's actually better to rely on `cpuinfo` in the testing unit since CPU feature detection depends on `auxv`. So validating `auxv` using the kernel's final report is a good idea, but I'm not sure if this would be friendly with chroots over QEMU BFMTs. However, it looks good to me.

> I disabled ppc64le clang build for now. It'd be better to have those tests fixed or disabled first, or some other build environment used.

It would be great to have a minimal reproducer, and you can open a separate issue for it. It could be a clang bug!

AlekseiNikiforovIBM:
Thanks! I've incorporated new names for workflow and jobs.

I've also rebased this PR due to `pyrefly` in test requirements: it fails to build on s390x.

AlekseiNikiforovIBM:
> It would be great to have a minimal reproducer, and you can open a separate issue for it. It could be a clang bug!

Yes, it could be a clang bug. I've seen similar failures on s390x on Ubuntu 24.04 with clang 20. But on Fedora 43 with clang 20 it didn't reproduce for me. That's why I'm using docker and Fedora 43 for clang builds on s390x.

I don't have a ppc64le machine at the moment to debug this issue. I'll have to leave it to others.

sandeepgupta12:
> > It would be great to have a minimal reproducer, and you can open a separate issue for it. It could be a clang bug!
> 
> Yes, it could be a clang bug. I've seen similar failures on s390x on Ubuntu 24.04 with clang 20. But on Fedora 43 with clang 20 it didn't reproduce for me. That's why I'm using docker and Fedora 43 for clang builds on s390x.
> 
> I don't have a ppc64le machine at the moment to debug this issue. I'll have to leave it to others.

I will give it a try for ppc64le.

AlekseiNikiforovIBM:
I've rebased this PR on new `main` due to previous conflict in `requirements/test_requirements.txt`

AlekseiNikiforovIBM:
@sandeepgupta12 has found that adding `-fno-strict-aliasing` resolves issues on ppc64le, and using clang-20 works on s390x.

The need for `-fno-strict-aliasing` can be investigated separately, I think.

AlekseiNikiforovIBM:
It looks like numpy is compiled with `-fno-strict-aliasing`, but on-demand compiled additional modules were compiled without `-fno-strict-aliasing`. I've added this option there and now it's no longer needed to be specified in CFLAGS for ppc64le and s390x.

I've also fixed one clang warning on s390x about undefined behaviour in negative number shifting.

numpy/_core/src/multiarray/compiled_base.c
AlekseiNikiforovIBM: How would that be done? I don't see how this place can be improved. Maybe such change could be added later in a separate PR?

AlekseiNikiforovIBM:
Is anything blocking this PR?

numpy/_core/src/multiarray/compiled_base.c
seiko2plus: What I mean is simply no need to swap what's already mistakenly swapped! So this fix is like performing "!!". The core of the problem is in:
https://github.com/numpy/numpy/blob/245e20bee21a603308d588942f792b0ad9d80e2b/numpy/_core/src/multiarray/compiled_base.c#L1595-L1599

A fix should be as simple as:
```C
            if ((order == PACK_ORDER_BIG) != (NPY_BYTE_ORDER == NPY_BIG_ENDIAN)) {
                v0 = npyv_rev64_u8(v0);
                v1 = npyv_rev64_u8(v1);
                v2 = npyv_rev64_u8(v2);
                v3 = npyv_rev64_u8(v3);
            }
```

Am I missing something in here?

seiko2plus:
> Is anything blocking this PR?

My apologies. It was the bottleneck on my part. I merged this PR to speed things up and land the fix, and since the changes wouldn't affect performance on non-bigendian archs, I think it's fine to apply the suggestion I made in here https://github.com/numpy/numpy/pull/30819#discussion_r3028068103 in another PR.


> It looks like numpy is compiled with -fno-strict-aliasing, but on-demand compiled additional modules were compiled without -fno-strict-aliasing. I've added this option there and now it's no longer needed to be specified in CFLAGS for ppc64le and s390x.

The fix is good to me. Yes, the flag should be passed to the C/C++ source already defined in https://github.com/numpy/numpy/blob/245e20bee21a603308d588942f792b0ad9d80e2b/meson.build#L48-L51.


Thank you, @AlekseiNikiforovIBM.

numpy/_core/src/multiarray/compiled_base.c
AlekseiNikiforovIBM: Yes, it doesn't work as you'd expect it due to `npyv_tobits_b8` calls.

It looks like `PACK_ORDER_BIG` is bit-order parameter, not byte-order parameter.

Let's take values from testcase I'm getting:
```
v[0]: 0x00 0x00 0x00 0x00 0x01 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x01 0x00 0x00 0x01
v[1]: 0x00 0x00 0x00 0x00 0x01 0x00 0x01 0x00 0x00 0x00 0x00 0x00 0x01 0x00 0x01 0x01
v[2]: 0x00 0x00 0x00 0x00 0x01 0x01 0x00 0x00 0x00 0x00 0x00 0x00 0x01 0x01 0x00 0x01
v[3]: 0x00 0x00 0x00 0x00 0x01 0x01 0x01 0x00 0x00 0x00 0x00 0x00 0x01 0x01 0x01 0x01
bb[0]: 0x0000000000009010
bb[1]: 0x000000000000d050
bb[2]: 0x000000000000b030
bb[3]: 0x000000000000f070
arr[0]: 0xf070b030d0509010
```

If excessive byteswapping is done where suggested and removed from other places, result would be:
```
v[0]: 0x00 0x00 0x00 0x01 0x00 0x00 0x00 0x00 0x01 0x00 0x00 0x01 0x00 0x00 0x00 0x00
v[1]: 0x00 0x01 0x00 0x01 0x00 0x00 0x00 0x00 0x01 0x01 0x00 0x01 0x00 0x00 0x00 0x00
v[2]: 0x00 0x00 0x01 0x01 0x00 0x00 0x00 0x00 0x01 0x00 0x01 0x01 0x00 0x00 0x00 0x00
v[3]: 0x00 0x01 0x01 0x01 0x00 0x00 0x00 0x00 0x01 0x01 0x01 0x01 0x00 0x00 0x00 0x00
bb[0]: 0x0000000000000908
bb[1]: 0x0000000000000b0a
bb[2]: 0x0000000000000d0c
bb[3]: 0x0000000000000f0e
```

I don't see how from those `bb` values I can easily get correct value of `arr[0]`.

PR #30664, opened 2026-01-16
DOC: add python-level parallelism guide to NumPy fundamentals

mariam851:
This PR adds a new section to the NumPy User Guide titled "Parallel Programming in NumPy" under the NumPy Fundamentals heading.

As discussed in issue #30597, there was a need for high-level documentation regarding:

When NumPy releases the GIL and how it benefits multithreading.

Practical guidance on using concurrent.futures.ThreadPoolExecutor vs multiprocessing.

Clarification on Vectorization vs. Parallelism to prevent common performance pitfalls.

Brief mention of free-threading (PEP 703) to future-proof the documentation.

Changes
Modified: doc/source/user/basics.rst -> Registered the new page in the toctree.

Created: doc/source/user/basics.parallel.rst -> Added the detailed conceptual guide and code examples.

PR #30655, opened 2026-01-15
DOC: add introductory overview of parallelism in NumPy

Nandika32:
### What does this PR do?

This PR adds a new introductory documentation page on parallelism in NumPy under
the "NumPy Fundamentals" section of the user guide.

The goal is to provide a high-level overview explaining what parallelism means
in the context of NumPy, without giving performance tuning advice.

### Why is this change needed?

There is currently no single, beginner-friendly entry point explaining how
parallelism relates to NumPy, the GIL, and when users should (or should not)
expect parallel execution.

This page is intentionally scoped to avoid implementation-specific details.

### Scope and limitations

 *Introductory overview only
 *No performance claims or benchmarks
 *No recommendations for specific libraries or techniques

PR #30820, opened 2026-02-11
DOC: add Python-level parallelism tips section

### LINKED ISSUE #30597: DOC: Add python-level parallelism tips to the docs

Right now the docs don't really have any suggestions on how to write performant code with NumPy. With the free-threaded build, there's increasing pressure to support multithreaded use but both multiprocessing and multithreading have seen lots of use with NumPy out in the real world for a long time.

I guess we could also tell people "just use dask!" and maybe these docs should suggest dask but I also think documenting how to work with standard library parallelism primitives is important.

Also, while working on https://github.com/numpy/numpy/issues/30494 we've been assembling some knowledge about ways to improve performance under multithreading. We need a place to add these tips in a way that's visible to users.

A new section in the user guide with content about how to use multiple CPU cores to work with numpy arrays. It probably makes sense to cover both multithreading and multiprocessing.

This can include both general guidance for writing performant code as well as tips for micro-optimizations like I point out here: https://github.com/numpy/numpy/pull/30514#issuecomment-3716554540.

I think it should also probably include general tips about how to write parallel code, including advising to minimize shared state and how to avoid contention around shared state with threads or lots of overhead due to pickle with multiprocessing.

riku-sakamoto:
### Add Python level parallelism tips section

Closes #30597

This PR includes the required content in #30597

Overview of changes made:

- Add a section on general tips for both multiprocessing and multithreading with NumPy
- Add code examples for both multiprocessing and multithreading using Python standard libraries to calculate Mandelbrot set
- ~~Add some micro tips for writing performant NumPy code~~
- Add a brief introduction to Dask as a third-party library for parallel computing with NumPy

Notes:

Two PRs currently exist that address the same issue. (#30664 and #30655)
Both PRs have been inactive for some time and they seems not to resolve all requests in the issue.  So, to move forward on #30597, I've created this new PR.
If preferred, I am also happy to merge their efforts.


doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,386 @@
...
+
+* Use dynamic task allocation where tasks are assigned to processes as they become available, rather than pre-allocating tasks.
+* Check ``chunksize`` parameter to ensure that tasks are neither too small (causing excessive overhead) nor too large (leading to load imbalance).
+
+
+Multithreading
+-----------------
+
+Multithreading allows multiple threads to run within the same process, sharing the same memory space.
+Starting with Python 3.13, a free-threading build of Python is available. When combined with libraries that are explicitly designed to be thread-safe, this can enable true parallel execution with threads.
```
ngoldbaum:
```suggestion
Starting with Python 3.13, a free-threaded build of Python is available. When combined with libraries that are explicitly designed to be thread-safe, this can enable true parallel execution with threads.
```

I'd also maybe link to https://py-free-threading.github.io as well.

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,386 @@
...
+The main pros and cons of multithreading are as follows:
+
+Pros
+++++
+* Lower memory usage since threads share the same memory space
+* Easier communication between threads
+
+Cons
+++++
+* Possibility of race conditions when accessing shared data accidentally
```
ngoldbaum:
```suggestion
* Possibility of race conditions when mutating shared data simultaneously with reads.
```

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,386 @@
...
+* Limited performance improvement if using Python libraries are not thread-safe or have limited support for free-threaded Python builds
+
+General tips
+++++++++++++
+
+The following are general tips for utilizing multithreading.
```
ngoldbaum, line 147: @lysnikolaou is in the process of writing a document that overlaps with this for the CPython docs. Once that exists we should link to it here. Pinging him to let him know about this PR.

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,386 @@
...
+
+Avoid race conditions
+~~~~~~~~~~~~~~~~~~~~~
+
+Race conditions occur when multiple threads update shared data simultaneously, leading to unpredictable results.
+To avoid race conditions, consider the following strategies:
+
+* Use thread-safe data structures or synchronization primitives like locks, semaphores, or condition variables to manage access to shared data.
+* Minimize the amount of shared data between threads by designing your program to use thread-local storage or by passing data explicitly to threads.
```
ngoldbaum: I'd put this and the point below *before* talking about thread-safe data structures.

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,386 @@
...
+
+The following are general tips for utilizing multithreading.
+
+Avoid race conditions
+~~~~~~~~~~~~~~~~~~~~~
+
+Race conditions occur when multiple threads update shared data simultaneously, leading to unpredictable results.
+To avoid race conditions, consider the following strategies:
+
+* Use thread-safe data structures or synchronization primitives like locks, semaphores, or condition variables to manage access to shared data.
```
ngoldbaum: I would bring up that locks mean the possibility of deadlocks to emphasize that it should only be done with utmost care.

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,386 @@
...
+        real = np.linspace(xmin, xmax, nx, dtype=np.float64)
+        imag = np.linspace(ymin, ymax, ny, dtype=np.float64)
+
+        SHARED_readonly_arr = (real[:, np.newaxis] + 1j * imag[np.newaxis, :]).ravel()
+        SHARED_readonly_arr.flags.writeable = False
+
+        SHARED_updating_steps = np.zeros(SHARED_readonly_arr.shape, dtype=np.int64)
+
+        mandelbrot_set(SHARED_readonly_arr.size, max_iter, n_workers)
+        mandelbrot_image = SHARED_updating_steps.reshape((nx, ny))
```
ngoldbaum, line 420: I don't know if you're aware that I wrote a very similar example here: https://py-free-threading.github.io/examples/mandelbrot/

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,386 @@
...
+    # Using a list
+    arr_list = np.array([1, 2, 3, 4, 5])
+
+    # Using a tuple
+    arr_tuple = np.array((1, 2, 3, 4, 5))
+
+
+This is reported in https://github.com/numpy/numpy/pull/30514#issuecomment-3716554540
+
+Using tuples avoids critical sections in the array creation code path, resulting in faster execution. Critical sections are portions of code that must be executed by only one thread at a time to prevent data corruption. In NumPy, critical sections are managed using ``NPY_BEGIN_CRITICAL_SECTION_SEQUENCE_FAST`` and ``NPY_END_CRITICAL_SECTION_SEQUENCE_FAST``. You can refer to those macros if necessary for more details.
```
ngoldbaum: This whole section isn't true. Not your fault - my fault for not correcting that in the issue. 

It's not actually very easy to get data as tuples initially and building tuples is very slow. Converting to a tuple and then an array is slower than just converting a list to an array. 

We also fixed the scaling and performance issues that this is suggesting to work around. 

I'd just delete the whole subsection.

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,386 @@
...
+
+.. code-block:: python
+
+    import numpy as np
+
+    # Avoiding module attribute lookups
+    sin = np.sin
+
+    arr = array([1, 2, 3, 4, 5])
+    value = sin(arr)
```
ngoldbaum: I also feel like we shouldn't document this microoptimization. There's also room to avoid contention by fixing this in CPython. IMO we shouldn't document bugs.

I'd also just delete this section.

Again, sorry for not correcting that in the issue.

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,386 @@
...
+
+Third Party Libraries for Multi-Core Processing
+===============================================
+
+In many practical scenarios, third-party libraries can provide more convenient and efficient solutions than using Python's standard libraries.
+We will briefly introduce some third party libraries at the end of this introduction.
+
+
+Dask
+----
```
ngoldbaum, line 434: @crusaderky are you ok with this language, as a dask maintainer?

ngoldbaum (commented):
Nice!

No need to integrate the other PRs. I'm suspicious they were largely generated by AI bots (apologies if they weren't!)

I left comments inline and pinged a couple people.

doc/source/user/basics.performant_code.rst
riku-sakamoto: Thanks. I've also mentioned https://py-free-threading.github.io/ in this section.

doc/source/user/basics.performant_code.rst
riku-sakamoto: I've changed the order so that the last bullet mention thread-safe data structures.

doc/source/user/basics.performant_code.rst
riku-sakamoto: I've added the possibility of deadlocks caused by these synchronization mechanism.

doc/source/user/basics.performant_code.rst
riku-sakamoto: Thanks for pointing this out. I've deleted this section.

I've also overlooked the comment of https://github.com/numpy/numpy/pull/30514#issuecomment-3717883421

doc/source/user/basics.performant_code.rst
riku-sakamoto: I've deleted this section.

doc/source/user/basics.performant_code.rst
riku-sakamoto: I wasn’t aware of this example.
It seems to be more detailed than the example I included. 

I've added a link to this example in this section.
Would you prefer that I replace my example with this one as well?

ngoldbaum: No that's OK, just offered it in case there was anything in there that seemed useful. 

While writing that documentation I ran into the fact that it's sometimes annoying to get multiprocessing working because of issues around pickleability of worker function arguments or shared data. That's why I used joblib for the multiprocessing example.

That limitation might actually be worth calling out as a con of multiprocessing - multithreading with a ThreadPoolExecutor "just works" on the first try way more often.

riku-sakamoto: Thank you for the clarification.

I agree that pickling requirements are also the downsides of multiprocessing.
I think it would be worth mentioning `joblib` as a possible solution for handling pickling-related issues, so I've added a section `pickling considerations` in `General tips` of multiprocessing.

Calling use of `joblib` as general tips seems a little bit debatable, but I believe it is worth  to mention it. I am happy to revise this if necessary.

doc/source/user/basics.performant_code.rst
riku-sakamoto: According to the CPython Issue ( https://github.com/python/cpython/issues/142518 ), the process of writing documentation about free threading seems almost finished. So, I've added the link below to the docs!
https://docs.python.org/3.15/library/threadsafety.html#thread-safety-guarantees

ngoldbaum:
@dpdani apologies for the random ping, but I thought that reviewing this might be in your wheelhouse? Feel free to ignore though.

dpdani:
@ngoldbaum was just going out, I'll definitely have a look tomorrow 👍

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,362 @@
...
+***************************************************
+Writing Performant NumPy Code with Multi-Core CPUs
+***************************************************
+
+
+Introduction
+================
+
+NumPy is designed for high performance numerical computing in Python by leveraging vectorized operations.
+However, vectorization does not always fully utilize the capabilities of multi-core processors. To realize multiprocessing and multithreading, additional strategies are necessary.
```
dpdani:
```suggestion
However, vectorization does not always fully utilize the capabilities of multi-core processors. To exploit parallelism, additional strategies are necessary.
```

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,362 @@
...
+
+* Use process pools to reuse existing processes instead of creating new ones for each task. `concurrent.futures.ProcessPoolExecutor` provides this feature.
+* Select appropriate startup methods. On Unix-like systems, the ``fork`` method can be faster than ``spawn`` or ``forkserver`` because it duplicates the parent process's memory space. However, ``fork`` may cause potential issues when using combined with threads or certain libraries. See the `multiprocessing documentation <https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods>`__ for more details.
+
+
+Reduce communication overhead
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Inter-process communication (IPC) can introduce significant overhead due to data serialization and transfer between processes. In Python, only picklable objects are allowed to be passed between processes.
+Due to this nature, multiprocessing is not suitable for the program which needs to synthesize data between processes frequently.
```
dpdani:
```suggestion
Due to this limitation, multiprocessing is not suitable for programs which needs to serialize data between processes frequently.
```

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,362 @@
...
+
+To reduce communication overhead, consider the following strategies:
+
+* Minimize the amount of data transferred between processes.
+* Use shared memory constructs such as `multiprocessing.shared_memory`, `multiprocessing.Array` or `multiprocessing.Value` for large data that needs to be accessed by multiple processes.
+* Set appropriate ``chunksize`` when using methods like ``map`` or ``imap`` in process pools. ``chunksize`` determines the number of tasks assigned to each process at a time, which can help balance the trade-off between task granularity and communication overhead.
+
+
+Balance processing load
```
dpdani: This is also true for multithreading. Maybe move this out of the multiprocessing section?

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,362 @@
...
+
+
+.. _basics.performant_code.third_party_libraries:
+
+
+Third Party Libraries for Multi-Core Processing
+===============================================
+
+In many practical scenarios, third-party libraries can provide more convenient and efficient solutions than using Python's standard libraries.
+We will briefly introduce some third party libraries at the end of this introduction.
```
dpdani: Can probably lose this sentence since the list starts immediately afterwards.

```suggestion
```

dpdani (commented):
Hi there 👋

This was good to read, thank you for writing it 👍

Just a few nitpicks for your consideration.

doc/source/user/basics.performant_code.rst
riku-sakamoto: Thank you for the suggestion!
I created a new section `Common tips for both multiprocessing and multithreading` and moved this part there.

riku-sakamoto:
@dpdani 
Thank you for the review!
I've applied the suggested changes.



Also, I've added linebreaks to make the text more readable. (originally requested in another PR's comment (https://github.com/numpy/numpy/pull/31282#discussion_r3111221022) )

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,416 @@
...
+Install NumPy from Source
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+NumPy includes ongoing work to improve compatibility and thread-safety
+when used with free-threaded Python builds.
+(See https://github.com/numpy/numpy/issues/30494 for details.)
+
+Thus, to utilize multithreading parallelism fully, we may need to build NumPy from source.
+Please refer to the :ref:`Building from source to use NumPy <building-from-source>` section
+for detailed instructions.
```
ngoldbaum: IMO we shouldn't include this in user-facing documentation, let's just delete this whole section.

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,416 @@
...
+
+    pyenv install 3.14t-dev
+    pyenv virtualenv 3.14t-dev numpy-multithreading
+    pyenv activate numpy-multithreading
+
+
+According to the `Python documentation <https://docs.python.org/3/howto/free-threading-python.html>`__, there are several ways to verify if your Python build is free-threaded. 
+
+* Run ``python -VV`` in your terminal and check ``free-threading build`` is shown
+* Check the value of `sys._is_gil_enabled()` in a Python shell, which should return `False`.
```
ngoldbaum, line 345: Rather than giving people all this information (and these days I'd probably point to uv before pyenv), I'd just point people to the free-threaded guide's installation instructions, which tells people how to do this with whatever package manager they'd like to use.

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,416 @@
...
+Dask
+----
+
+Dask is an open-source library that provides parallel compuing features
+not only for a single machine but also for a cluster of machines.
+It also provides ``DaskArray`` which has a similar API to NumPy's ``ndarray``. 
+If you are familiar with NumPy, you can easily get started with ``DaskArray``.
+
+* Dask Documentaion: https://docs.dask.org/en/stable/
+* Dask GitHub Repository: https://github.com/dask/dask
```
ngoldbaum, line 442: I'd add a section on joblib as well.

I would also probably link to `threadpoolctl`, which has hooks to prevent CPU oversubscription: https://github.com/joblib/threadpoolctl

ngoldbaum (commented):
IMO this is a good addition to the docs and something where I think it will be helpful for the NumPy docs to have an opinionated take on.

I have a couple opinions inline on a second pass through the text.

I'm also going to ping in a few other people just to make sure I have buy-in from the docs community and to get one more set of eyes on the text.

@itamarst is there any chance you could look this over, given your recent work on scikit-learn maybe you'll have some insight as well. Itamar can also tell you about his recent experiences with CPU oversubscription causing headaches.

@melissawm @rossbar just to get an opinion from another NumPy maintainer with strong opinions about docs: are you OK with the NumPy docs having a Python parallelism quickstart guide? Is this an OK place to put it?

charris:
I think it is appropriate. Parallelism is certainly an important topic in scientific computing, and having some guidance in place avoids document chasing for those who want to use it.

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,416 @@
...
+Starting with Python 3.13, a free-threaded build of Python is available.
+When combined with libraries that are explicitly designed to be thread-safe,
+this can enable true parallel execution with threads.
+For details on free-threaded Python builds, see the
+`Python Free-Threading Guide <https://py-free-threading.github.io/>`__.
+
+
+As a high-level API, Python provides the `concurrent.futures.ThreadPoolExecutor` class 
+for thread-based parallelism.
+Python also provides the `concurrent.futures.InterpreterPoolExecutor`,
```
melissawm: This class does not get picked up by sphinx, maybe due to the CircleCI setup being [frozen on 3.12](https://github.com/numpy/numpy/blob/ea9cae77b41c531b8badbf5b7548e2029f6c13be/.circleci/config.yml#L12).

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,416 @@
...
+
+Inter-process communication (IPC) can introduce significant overhead due to data serialization and transfer between processes. In Python, only picklable objects are allowed to be passed between processes.
+Due to this limitation, multiprocessing is not suitable for programs which need to serialize data between processes frequently.
+
+To reduce communication overhead, consider the following strategies:
+
+* Minimize the amount of data transferred between processes.
+* Use shared memory constructs such as `multiprocessing.shared_memory`, `multiprocessing.Array`
+  or `multiprocessing.Value` for large data that needs to be accessed by multiple processes.
+* Set appropriate ``chunksize`` when using methods like ``map`` or ``imap`` in process pools.
```
melissawm: As someone who is not familiar with multiprocessing/multithreaded development, I was a bit confused as to where this (the chunksize arg in particular) would come into play until I saw the Mandelbrot example below. Maybe mention here that you will show an example further ahead? If this is too basic and you expect the readers of this doc to already understand how to use this, feel free to ignore me.

melissawm (approved):
Hi @ngoldbaum - I think this is super relevant and placed at exactly the right spot, thanks!

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,416 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Process creation has a higher overhead compared to thread creation due to the need to initialize a new Python interpreter and memory space.
+To mitigate this overhead, consider the following strategies:
+
+* Use process pools to reuse existing processes instead of creating new ones for each task.
+  `concurrent.futures.ProcessPoolExecutor` provides this feature.
+* Select appropriate startup methods. On Unix-like systems, the ``fork`` method can be faster
+  than ``spawn`` or ``forkserver`` because it duplicates the parent process's memory space.
+  However, ``fork`` may cause potential issues when using combined with threads or certain libraries.
```
itamarst: This is an understatement, I would say it's _fundamentally and irretrievably_ broken and should never be used. This is why Python changed the default away from `fork` in 3.14.

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,416 @@
...
+The main pros and cons of multithreading are as follows:
+
+Pros
+++++
+* Lower memory usage since threads share the same memory space
+* Easier communication between threads
+
+Cons
+++++
+* Possibility of race conditions when mutating shared data simultaneously with reads.
```
itamarst: Maybe rephrase as "mutating shared data simultaneously with reads in other threads"

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,416 @@
...
+
+General tips
+++++++++++++
+
+The following are general tips for utilizing multithreading.
+For more details on thread safety guarantees for built-in types
+in Python's free-threaded build, see the Python documentation
+on `Thread Safety Guarantees <https://docs.python.org/3.15/library/threadsafety.html#thread-safety-guarantees>`__.
+
+
```
itamarst, line 153: Might be worth saying something like:

> NumPy's linear algebra APIs (TODO exact reference, it's... `np.linalg` and `@` maybe?) may have their own thread pool, provided by the underlying BLAS library. If you have your own thread pool that is using all the cores on your machine, and you are using you will want to set the number of threads in BLAS to 1, which you can do with the `threadpoolctl` library (TODO with link). Otherwise you may end up using too many threads, slowing down your program.

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,416 @@
...
+  for serialization and can handle a wider range of Python objects than the standard ``pickle`` module.
+  See the ``joblib`` documentaion on `Serialization of un-picklable objects <https://joblib.readthedocs.io/en/latest/auto_examples/serialization_and_wrappers.html>`__ for more details.
+
+
+
+Multithreading
+-----------------
+
+Multithreading allows multiple threads to run within the same process, sharing the same memory space.
+Starting with Python 3.13, a free-threaded build of Python is available.
```
itamarst: Maybe 3.14? Since that's first stable one.

doc/source/user/basics.performant_code.rst
```diff
@@ -0,0 +1,416 @@
...
+  since they reduce the need for explicit synchronization.
+* Use thread-safe data structures or synchronization primitives like locks, semaphores,
+  or condition variables to manage access to shared data.
+  Note that improper use of these synchronization mechanisms can cause deadlocks,
+  so they should be used with care.
+
+
+Common tips for both multiprocessing and multithreading
+-------------------------------------------------------
+
```
itamarst, line 193: Set the pool size to `joblib.cpu_count()`.

Python's `cpu_count()` or the improved `process_cpu_count()` are OK, but they will give the wrong answer in containers (Docker, Kubernetes). `joblib.cpu_count()` does the right thing there (it's basically `loky.cpu_count()`, if you want to recommend that instead).

doc/source/user/basics.performant_code.rst
dpdani: I don't think that's an accurate representation of `fork`. It definitely has use cases where it's the proper and better choice, and it's not broken. The fact that it's not easy to use correctly, and thus no longer a default, doesn't change that.

itamarst: It's very often broken whenever you have threads. And many libraries start threads automatically on import, `numpy` amongst them in some configurations, often unbeknownst to users.

For example, if you're using an libgomp OpenMP-build of OpenBLAS you're going to get hangs on fork(): https://github.com/OpenMathLib/OpenBLAS/blob/1bd74ad3d1e8d21f86d1a6be35abfcdf27c0208a/driver/others/memory_qalloc.c#L295

From the libgomp bug description I suspect this is the case for libgomp OpenMP users in general.

Basically `fork()` is an invitation to mysterious "my software deadlocks twice a day and I have no idea why", and "someone else used my software in a slightly different environment and now it hangs twice a day and I have no idea why", and hanging is the best case failure. 

People who have the knowledge to look out for these things don't need to be told about `fork`. People who don't know to look (the audience for intro docs) should be told not to use `fork`.

riku-sakamoto:
Thank you everyone for reviewing this!
I'll address the comments this weekend.

doc/source/user/basics.performant_code.rst
riku-sakamoto: I agree! I've deleted it.

doc/source/user/basics.performant_code.rst
riku-sakamoto: I've left the link to free-threaded guide's installation instructions only and deleted the part about `pyenv`.

doc/source/user/basics.performant_code.rst
riku-sakamoto: Thank you for the suggestion!
I've added sections about `joblib` and `threadpoolctl` with links to their documentation and GitHub repositories.

doc/source/user/basics.performant_code.rst
riku-sakamoto: Thank you for the information!
As a temporary workaround, I've set a link explicitly to the documentation of `concurrent.futures.InterpreterPoolExecutor` in the Python standard library documentation.

doc/source/user/basics.performant_code.rst
riku-sakamoto: I agree. I've added a sentence at the beginning of `General tips` section to mention that there will be an example later in the document.

doc/source/user/basics.performant_code.rst
riku-sakamoto: I've tweaked the wording about `fork` to make it more clear that it can lead to deadlocks or crashes and it is not recommended unless users know it well.

doc/source/user/basics.performant_code.rst
riku-sakamoto: Thanks. I've rephrased it.

doc/source/user/basics.performant_code.rst
riku-sakamoto: Thanks for the suggestion!
I've added a section about `Avoid CPU oversubscription` and metioned `threadpoolctl` there.

doc/source/user/basics.performant_code.rst
riku-sakamoto: I wrote it as `Python 3.13` since I refered to the [PEP703](https://peps.python.org/pep-0703/).
But, as you suggested, according to the Python3.14 documentation below, stable support for free-threaded Python is added in Python 3.14, so I've tweaked the wording.

* https://docs.python.org/3/whatsnew/3.14.html#free-threaded-mode-improvements
* https://docs.python.org/3/whatsnew/3.14.html#free-threaded-python-is-officially-supported

doc/source/user/basics.performant_code.rst
riku-sakamoto: I've added a new section about `Determine the correct number of cpus` and mentioned `joblib.cpu_count()` there.

ngoldbaum (approved):
Thanks @riku-sakamoto! Let's bring this in and we can leave further tweaks to followups.

PR #27345, opened 2024-09-04
ENH: Better handling of nan values in np.unique for multidimensional arrays

caph1993:
Enhancement of ``equal_nan`` option in ``np.unique`` for multi-dimensional data
-----------------------------------------------------
fixes gh-23286.
fixes gh-20873.

This PR affects the ``np.unique`` function when ``equal_nan`` is set to true and when the input is either a multi-dimensional array with given axis (not None) or a structured 1-dimensional array.

Before this PR, the results of the following two expressions are the same:
```
np.unique([[0, np.nan], [0, np.nan], [np.nan, 1]], equal_nan=True, axis=0) # Expression 1
np.unique([[0, np.nan], [0, np.nan], [np.nan, 1]], equal_nan=False, axis=0) # Expression 2
```
Their result is ``[[0, np.nan], [0, np.nan], [np.nan, 1]]``, which is the result one would expect only when ``equal_nan`` is ``False``. In other words,  ``equal_nan`` is basically ignored and assumed as ``False`` (but not explicitly in the code) for multi-dimensional arrays and 1-dimensional structured arrays. The same behavior is observed if the arrays are 1-dimensional with structured dtype instead of 2-dimensional.

After this PR, the first expression returns ``[[0, np.nan], [np.nan, 1]]`` while the second returns the old result, which is more meaningful and resolves some exisiting github issues. This behavior matches the semantics of "nans being equal", in the sense that the function is essentially treating the comparison ``(1, np.nan) == (1, np.nan)`` as if it was true. The same good behavior is observed after this PR if the arrays are 1-dimensional with structured dtype instead of 2-dimensional.

Issue #28920, opened 2025-05-07
BUG: `array_equal(x, y, equal_nan=True)` raises for structured arrays


LasseBlaauwbroek:
### Describe the issue:

Given structured arrays (example below), running `array_equal(x, y, equal_nan=True)` does not work.

Analysis:
1. The code correctly deduces that structured arrays may include NaNs. Therefore, it uses `isnan(x)` to compare NaNs.
2.  This happens because structured arrays may include types that cannot be NaN, which `isnan(x)` flags as such and raises.

### Reproduce the code example:

```python
import numpy as np

x = np.array((1,), dtype=[('f0', np.int32)])
y = np.array((1,), dtype=[('f0', np.int32)])
np.array_equal(x, y, equal_nan=True)
```

### Error message:

```shell
Traceback (most recent call last):
  File "<..>/test5.py", line 5, in <module>
    np.array_equal(x, y, equal_nan=True)
  File "<..>/venv/lib/python3.11/site-packages/numpy/_core/numeric.py", line 2569, in array_equal
    a1nan, a2nan = isnan(a1), isnan(a2)
                   ^^^^^^^^^
TypeError: ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

### Python and NumPy Versions:

2.1.3
3.11.3 (main, Apr 18 2023, 16:19:41) [GCC 12.2.1 20230201]

### Runtime Environment:

[{'numpy_version': '2.1.3',
  'python': '3.11.3 (main, Apr 18 2023, 16:19:41) [GCC 12.2.1 20230201]',
  'uname': uname_result(system='Linux', node='lasse-work', release='6.14.4-arch1-2', version='#1 SMP PREEMPT_DYNAMIC Tue, 29 Apr 2025 09:23:13 +0000', machine='x86_64')},
 {'simd_extensions': {'baseline': ['SSE', 'SSE2', 'SSE3'],
                      'found': ['SSSE3',
                                'SSE41',
                                'POPCNT',
                                'SSE42',
                                'AVX',
                                'F16C',
                                'FMA3',
                                'AVX2',
                                'AVX512F',
                                'AVX512CD',
                                'AVX512_SKX',
                                'AVX512_CLX',
                                'AVX512_CNL',
                                'AVX512_ICL'],
                      'not_found': ['AVX512_KNL', 'AVX512_KNM']}}]

ngoldbaum:
[...]

PR #30822, opened 2026-02-11
BUG: Unique with nans and with axis

### LINKED ISSUE #20873: BUG: `unique` with NaNs and along an axis inconsistent with flat version
### Describe the issue:

Running `np.unique` on a flat array with multiple `NaN`s only returns one `NaN`. Running along an axis that has duplicated rows containing `NaN`s returns both rows however.

I suspect this is related to https://github.com/numpy/numpy/issues/20326 - whatever the decision here I think the result should be consistent between `axis=None` and `axis=0`.

```python
import numpy as np

print(np.unique([np.nan, np.nan, 2.]))
# [ 2. nan]

a = np.array([[np.nan, 2.],
              [np.nan, 2.]])
print(np.unique(a))
# [ 2. nan]
print(np.unique(a, axis=0))
# [[nan  2.]
#  [nan  2.]]
#
# Expected:
# [[nan  2.]]
```

### Error message:

_No response_

1.22.1
3.9.7 (default, Sep 16 2021, 08:50:36) 
[Clang 10.0.0 ]

### LINKED ISSUE #23286: BUG: np.unique incorrect output given input with NaNs when equal_nan=True and axis is not None
### Describe the issue:

np.unique fails to remove duplicate NaNs even though equal_nan=True when axis is not None. In the code example below, the output is
```
[ 0.  1.  2. nan nan]
```
but one should expect
```
[ 0.  1.  2. nan]
```
This bug does not occur when axis is None.

```python
import numpy as np

a = np.array([0., np.nan, 2., np.nan, 2., 1., 0., 1., 2., 0.])
print(np.unique(a, equal_nan=True, axis=0))
```

### Error message:

_No response_

### Runtime information:

```
1.24.2
3.9.13 (main, Oct 13 2022, 16:12:19) 
[Clang 12.0.0 ]
WARNING: `threadpoolctl` not found in system! Install it by `pip install threadpoolctl`. Once installed, try `np.show_runtime` again for more detailed build information
[{'simd_extensions': {'baseline': ['NEON', 'NEON_FP16', 'NEON_VFPV4', 'ASIMD'],
                      'not_found': ['ASIMDFHM']}}]
None
```

### Context for the issue:

_No response_

caph1993:
Support for ``equal_nan`` option in ``np.unique`` when ``axis!=None``
-----------------------------------------------------
fixes gh-23286.
fixes gh-20873.

This PR affects the ``np.unique`` function when ``equal_nan`` is set to true and when the input is either a multi-dimensional array with given axis (not None) or a structured 1-dimensional array.

Before this PR, the results of the following two expressions are the same:
```
np.unique([[0, np.nan], [0, np.nan], [np.nan, 1]], equal_nan=True, axis=0) # Expression 1
np.unique([[0, np.nan], [0, np.nan], [np.nan, 1]], equal_nan=False, axis=0) # Expression 2
```

This PR is an alternative to #27345 . It's simpler and does the job.


caph1993:
Fixes gh-29336 too

caph1993:
If `np.array_equal(x, y, equal_nan=True)` is ever implemented for structured arrays, as suggested in gh-28920, the code for unique could be greatly simplified. The big if-elif-else of this PR would simply be reduced to one instruction: `mask[1:] = ~np.array_equal(aux[1:], aux[:-1], equal_nan=equal_nan)`.

caph1993:
@caph1993 do you intend on completing this in the spirit of the review comment?
> I encourage you to re-submit this PR with a simpler implementation, focusing only on fixing the bug.

_Originally posted by @mattip in https://github.com/numpy/numpy/issues/27345#issuecomment-3802203584_

I submitted this alternative simpler PR. Can you please aprove the workflows to verify that it can be merged?

caph1993:
Could someone help triage this PR and suggest appropriate labels?

PR #30825, opened 2026-02-12
MAINT: update pythoncapi-compat for PyUnstable_SetImmortal()

kumaraditya303:
This is needed for https://github.com/numpy/numpy/issues/30494 so that we can use `PyUnstable_SetImmortal` on 3.14.


ngoldbaum:
Thanks Kumar!

PR #30826, opened 2026-02-12
ENH: make PyDataMem_DefaultHandler and default extobj capsule immortal in free-threading

kumaraditya303:
Towards fixing #30494 

With this, multit-threading is almost ~30% faster than multi-processing on mac.
```console
❯ samply record python -X perf mtmp.py
Inner loops 10, multithreading  time: 2.76 sec, result sum: 717434683.1879176
Inner loops 10, multiprocessing time: 3.90 sec, result sum: 717434683.1879176
```

### DIFF
--- a/numpy/_core/src/multiarray/multiarraymodule.c
+++ b/numpy/_core/src/multiarray/multiarraymodule.c
@@ -5170,7 +5170,11 @@ _multiarray_umath_exec(PyObject *m) {
     if (PyDataMem_DefaultHandler == NULL) {
         return -1;
     }
-
+#ifdef Py_GIL_DISABLED
+    int rc = PyUnstable_SetImmortal(PyDataMem_DefaultHandler);
+    assert(rc == 1);
+    (void)rc;
+#endif
     /*
      * Initialize the context-local current handler
      * with the default PyDataMem_Handler capsule.
--- a/numpy/_core/src/multiarray/multiarraymodule.c
+++ b/numpy/_core/src/multiarray/multiarraymodule.c
@@ -5171,9 +5171,11 @@ _multiarray_umath_exec(PyObject *m) {
         return -1;
     }
 #ifdef Py_GIL_DISABLED
-    int rc = PyUnstable_SetImmortal(PyDataMem_DefaultHandler);
-    assert(rc == 1);
-    (void)rc;
+    if (PyUnstable_SetImmortal(PyDataMem_DefaultHandler) == 0) {
+        PyErr_SetString(PyExc_RuntimeError,
+                        "Could not mark memory handler capsule as immortal");
+        return -1;
+    }
 #endif
     /*
      * Initialize the context-local current handler
--- a/numpy/_core/src/umath/extobj.c
+++ b/numpy/_core/src/umath/extobj.c
@@ -15,6 +15,7 @@
 #include "numpy/ufuncobject.h"
 
 #include "common.h"
+#include "npy_pycompat.h"
 
 
 #define UFUNC_ERR_IGNORE 0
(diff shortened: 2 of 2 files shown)


ngoldbaum:
Thanks Kumar!

Can you fix [this issue](https://github.com/python/pythoncapi-compat/pull/164/changes#r2799946915) before we merge this? Otherwise CI will fail whenever the github actions runners pick up alpha6.

ngoldbaum:
I think I also made the extobj handler immortal in my original attempt, something like this:

```
diff --git a/numpy/_core/src/umath/extobj.c b/numpy/_core/src/umath/extobj.c
index 91b0b4c62d..5f3fa46678 100644
--- a/numpy/_core/src/umath/extobj.c
+++ b/numpy/_core/src/umath/extobj.c
@@ -8,7 +8,7 @@
 #include "npy_config.h"

 #include "npy_argparse.h"
-
+#include "npy_pycompat.h"
 #include "conversion_utils.h"

 #include "extobj.h"
@@ -145,6 +145,13 @@ init_extobj(void)
     if (npy_static_pydata.default_extobj_capsule == NULL) {
         return -1;
     }
+
+    if (PyUnstable_SetImmortal(npy_static_pydata.default_extobj_capsule) == 0) {
+        PyErr_SetString(PyExc_RuntimeError,
+                        "Could not mark extobj capsule as immortal");
+        return -1;
+    }
+
     npy_static_pydata.npy_extobj_contextvar = PyContextVar_New(
             "numpy.ufunc.extobj", npy_static_pydata.default_extobj_capsule);
     if (npy_static_pydata.npy_extobj_contextvar == NULL) {
```

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -5170,7 +5170,11 @@ _multiarray_umath_exec(PyObject *m) {
     if (PyDataMem_DefaultHandler == NULL) {
         return -1;
     }
-
+#ifdef Py_GIL_DISABLED
+    int rc = PyUnstable_SetImmortal(PyDataMem_DefaultHandler);
+    assert(rc == 1);
+    (void)rc;
+#endif
```
ngoldbaum, line 5179: Why only do it on the free-threaded build?

Also I slightly prefer registering failure here as a runtime error:

```
diff --git a/numpy/_core/src/multiarray/multiarraymodule.c b/numpy/_core/src/multiarray/multiarraymodule.c
index 2f381a1c7a..2b6e96f952 100644
--- a/numpy/_core/src/multiarray/multiarraymodule.c
+++ b/numpy/_core/src/multiarray/multiarraymodule.c
@@ -5180,6 +5180,12 @@ _multiarray_umath_exec(PyObject *m) {
         return -1;
     }

+    if (PyUnstable_SetImmortal(npy_static_pydata.default_extobj_capsule) == 0) {
+        PyErr_SetString(PyExc_RuntimeError,
+                        "Could not mark memory handler capsule as immortal");
+        return -1;
+    }
+
     if (initumath(m) != 0) {
         return -1;
     }
```

ngoldbaum:
> Can you fix [this issue](https://github.com/python/pythoncapi-compat/pull/164/changes#r2799946915) before we merge this? Otherwise CI will fail whenever the github actions runners pick up alpha6.

Never mind, it seems `SetImmotal` didn't make it into alpha 6. So that means if we merge this PR, building against CPython's `main` branch will be broken until alpha 7. But also a one-line hacky patch to the pythoncapi-compat headers works as a quick fix if anyone cares and runs into that.

ngoldbaum:
> With this, multit-threading is almost ~30% faster than multi-processing on mac.

Given that there aren't any remaining bottlenecks in the benchmark script from #30494, you can feel free to add "fixes #30494" to the PR description so it'll be auto-closed when this is merged.

numpy/_core/src/multiarray/multiarraymodule.c
kumaraditya303: > Why only do it on the free-threaded build?

No strong reason, just that making it immortal on default build has little to no effect.

kumaraditya303: Added error handling.

ngoldbaum:
No one else commented and I think this is a straight-up win that shouldn't have any downsides. These objects are already immortal so this isn't a memory leak.

Merging, thank Kumar!

PR #30829, opened 2026-02-13
TYP: ``matlib``: missing extended precision imports

jorenham:
I found these while working on https://github.com/jorenham/typestats: `float96` and `complex192` are ilisted in `numpy.matlib.__all__` but not imported. (we export both `float96 and `float128` in the stubs to avoid bugs of the "but it works on my machine" kind)


charris:
Thanks Joren.

PR #30835, opened 2026-02-15
ENH: Add minmax function for single-pass min/max computation

### LINKED ISSUE #9836: ENH: minmax function
Would be really handy to have a function that computes and returns both the minimum and maximum of an array. This should be substantially faster than calling `min` and then `max` on the same array.

FWIW here's an [SO question]( https://stackoverflow.com/q/12200580 ) from some time ago that has gotten a fair bit of attention.

veeceey:
Adds `np.minmax()` that returns `(min, max)` as a tuple, addressing #9836.

**What it does:**
- New public function `np.minmax(a, axis=None, out=None, keepdims=False)`
- Returns `(min, max)` tuple, consistent with how `np.divmod` returns two values
- Supports `axis`, `keepdims`, and `out` parameters matching `np.min`/`np.max`
- Propagates NaN values (same as `np.min`/`np.max`)
- Uses the name `minmax` to stay consistent with C++, MATLAB, Julia, and Ruby conventions as discussed in #9836

**Implementation approach:**
The current implementation delegates to `np.minimum.reduce` and `np.maximum.reduce`, which gives us full dtype support and correct axis handling for free. This establishes the public API surface so that a future C-level or SIMD-optimized true single-pass implementation (like the one in numpy-minmax or cyminmax) can be dropped in behind this same API without any user-facing changes.

**What's included:**
- `np.minmax()` in `numpy/_core/fromnumeric.py` with full docstring
- Helper `_minmax()` in `numpy/_core/_methods.py`
- Type stubs in `fromnumeric.pyi` and `__init__.pyi`
- Tests covering basic usage, axis, keepdims, NaN propagation, and empty arrays

Closes #9836


mattip:
This is not what we were looking for. The problem is not adding to the API (which is what this PR solves) rather making a `minmax` performant in a way that can also be implemented by user-defined dtypes.

Closing, please reopen with justification if you think otherwise.

PR #30838, opened 2026-02-16
ENH: Added kernel and subsystem info to `show_config`

ganesh-k13:
## Changes:
<details><summary>Before</summary>
<p>

```python
>>> np.show_config
...
Machine Information:
  build:
    cpu: aarch64
    endian: little
    family: aarch64
    system: darwin
  host:
    cpu: aarch64
    endian: little
    family: aarch64
    system: darwin
...

```

</p>
</details> 

<details><summary>After</summary>
<p>

```python
>>> np.show_config
...
Machine Information:
  build:
    cpu: aarch64
    endian: little
    family: aarch64
    kernel: xnu
    subsystem: macos
    system: darwin
  host:
    cpu: aarch64
    endian: little
    family: aarch64
    kernel: xnu
    subsystem: macos
    system: darwin
...
```

</p>
</details> 

## References
- https://mesonbuild.com/Reference-manual_builtin_build_machine.html#build_machinekernel
- https://mesonbuild.com/Reference-manual_builtin_host_machine.html#build_machinekernel

## Notes
requires `meson>=1.2.0`, but NumPy pins to `>=0.13.1`. Not sure if this will be a problem for some systems


ganesh-k13:
Pyodide build fails with `kernel` not found although the version is 1.9. [ref](https://github.com/numpy/numpy/actions/runs/22049880891/job/63705587462?pr=30838#step:3:564). Not sure why

charris:
> Not sure why

I saw that in another PR, which I merged anyway and it didn't persist IIRC.

numpy/meson.build
```diff
@@ -406,6 +406,13 @@ foreach name, machine : machines
   conf_data.set(name + '_CPU_FAMILY', machine.cpu_family())
   conf_data.set(name + '_CPU_ENDIAN', machine.endian())
   conf_data.set(name + '_CPU_SYSTEM', machine.system())
+  if meson.version().version_compare('>=1.2.0') and not meson.is_cross_build()
```
ganesh-k13, line 409: So there are two ways to solve this problem, one would be to add this guard for `emscripten` builds or, we add kernel and subsystem in https://github.com/numpy/numpy/blob/552c46f7b114466c760bb63fba4ae1e4842f5f53/tools/ci/emscripten/emscripten.meson.cross#L11-L15

ganesh-k13: Query: Which is better and what would the values be? Based on my search, it seems `kernel=emscripten` and `subsystem=''` would be correct?

mattip:
What is the motivation for showing the build system kernel? Have you found a need for that?

ganesh-k13:
An example where this would be of help is `system: darwin`: This could mean `macOS` or `iOS`.

That being said, this did not come as an immediate need and I found it when I saw `macOS` was missing in `show_config` and `show_runtime`. Also this puts it inline with Meson provided configs.

mattip:
Ok, that explains system, but why is the kernel used in the build system important?

ganesh-k13:
ah, you mean it's fine for the host system info but not needed for the build system? That makes sense, thought of keeping parity between them, but ok to keep it only for host if that's better.

mattip:
In `show_config()` the `host` is not the current running host, it is the host used to build NumPy. Those values are hard-coded into the `__config__.py` file shipped with NumPy. I think the `system` info is also redundant, it can be found in the `dist-info/WHEEL` tag.

In general I think `show_config` is of limited value, `show_runtime` is a better source of info.

ganesh-k13:
ahh ok, got it. I was assuming it's the running host for some reason even though it's filled in at compile time, my bad, I'll close this for now.

Tangential to this, I remember seeing an issue where there was a proposal to combine `show_config` and `show_runtime` is that worth doing now?

PR #30747, opened 2026-01-30
MAINT: Update x86-simd-sort to latest

r-devulap:
Porting https://github.com/numpy/x86-simd-sort/pull/225/changes/4273453816197dd2672a4ab92b301138b423e21b for https://github.com/numpy/x86-simd-sort/issues/224. Fixes build failure on GCC 16.0.1 when using LTO.

PR #30841, opened 2026-02-16
MAINT: Synchronize 2.4.x submodules with main

charris:
Backport #30747 and #30825

This helps avoid accidental modifications of the submodules.

- Update pythoncapi-compat
- Update x86-simd-sort


charris:
@r-devulap @kumaraditya303 Could these cause problems with Python 3.11?

r-devulap:
@charris I don't think so.

r-devulap (approved):
LGTM.

mattip:
Thanks @charris as always for these backports.

PR #30843, opened 2026-02-17
ENH: add np.minmax() for combined min/max reduction

### LINKED ISSUE #9836: ENH: minmax function

veeceey:
Adds `np.minmax()` that returns `(min, max)` in a single function call, addressing a long-standing request from #9836 (57 upvotes).

**Motivation:** computing both min and max is extremely common (normalization, range checks, extent calculations) but currently requires two separate calls that each scan the entire array. This PR establishes the public API so that a follow-up C-level single-pass implementation can be swapped in behind it.

**API:**
```python
mn, mx = np.minmax(a, axis=None, keepdims=False, initial=..., where=...)
```

The `initial` parameter accepts either a scalar (used for both reductions) or a `(min_initial, max_initial)` tuple for separate identity values — useful for empty slices:

```python
>>> np.minmax([], initial=(np.inf, -np.inf))
(inf, -inf)
```

Supports `axis`, `keepdims`, `initial`, and `where` matching the signatures of `np.min`/`np.max`. The name `minmax` follows the convention used by C++ (`std::minmax`), MATLAB, Julia, and Ruby.

**Current benchmarks** (M3 MacBook, 10M float64 elements):

| Operation | Median time |
|---|---|
| `np.min(a) + np.max(a)` | 2.43 ms |
| `np.min(a)` alone | 1.19 ms |
| `np.minmax(a)` (this PR) | ~2.4 ms |

The current implementation delegates to `minimum.reduce` + `maximum.reduce` internally (two passes), same as `np.ptp`. A follow-up PR can add a single-pass C/Highway kernel behind the same API to bring `minmax` close to the single-pass lower bound (~1.2 ms for 10M elements, roughly 2x faster than calling min+max separately).

**Changes:**
- `numpy/_core/_methods.py` — `_minmax()` helper
- `numpy/_core/fromnumeric.py` — public `minmax()` with full docstring
- `numpy/_core/fromnumeric.pyi` — type stubs
- `numpy/__init__.py` / `numpy/__init__.pyi` — exports
- `numpy/_core/tests/test_minmax.py` — comprehensive tests

Closes #9836

Signed-off-by: Varun Chawla <varun_6april@hotmail.com>


ikrommyd:
https://github.com/numpy/numpy/pull/30835

mattip:
This is not what we were looking for. The problem is not adding to the API (which is what this PR solves) rather making a minmax performant in a way that can also be implemented by user-defined dtypes.

Closing, please reopen with justification if you think otherwise.

veeceey:
Thanks @ikrommyd for pointing out #30835 - that's actually my earlier PR as well. #30835 was the initial implementation, and this PR (#30843) is a refined version addressing some design feedback. I'll close #30835 in favor of this one to avoid confusion.

mattip:
Please stop submitting the same thing hoping for different outcomes without actually relating to reviewer comments or the actual issue

PR #30845, opened 2026-02-18
ENH: Add TypeError for user-defined dtypes in dot, inner, vdot, and correlateError message for dot family

abslock128w:
as discussed in the issue, this is the first step better error messages 
for the dot family...

### Problem
`np.dot`, `np.inner`, `np.vdot`, and `np.correlate` hit a wall with 
user-defined dtypes because `PyArray_ObjectType` rejects non-legacy dtypes, 
that give the users a confusing error.

### Fix
added a `typenum >= NPY_NTYPES_LEGACY` check in the five affected C functions 
in `multiarraymodule.c`:
- `PyArray_InnerProduct`
- `PyArray_MatrixProduct2`
- `PyArray_Correlate`
- `PyArray_Correlate2`
- `array_vdot`

each one now raises a clear `TypeError` pointing to `@`, `np.vecdot`, 
`np.matvec`, or `np.vecmat` as alternatives.

### Test
tested locally with `ml_dtypes.bfloat16`:
```python
import numpy as np
import ml_dtypes
a = np.array([1.0, 2.0], dtype=ml_dtypes.bfloat16)
b = np.array([3.0, 4.0], dtype=ml_dtypes.bfloat16)
np.dot(a, b)
# TypeError: dot() does not support user-defined dtypes. 
# Consider using '@' (matmul), np.vecdot, np.matvec, or np.vecmat instead.

### DIFF
--- a/numpy/_core/src/multiarray/multiarraymodule.c
+++ b/numpy/_core/src/multiarray/multiarraymodule.c
@@ -894,6 +894,12 @@ PyArray_InnerProduct(PyObject *op1, PyObject *op2)
     if (typenum == NPY_NOTYPE) {
         return NULL;
     }
+    if (typenum >= NPY_NTYPES_LEGACY) {
+    PyErr_SetString(PyExc_TypeError,
+        "dot() does not support user-defined dtypes. "
+        "Consider using '@' (matmul), np.vecdot, np.matvec, or np.vecmat instead.");
+    return NULL;
+    }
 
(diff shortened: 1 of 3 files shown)


seberg:
Thanks, but this approach isn't quite right.  You have to check for the dot function being `NULL` explicitly.

seberg:
Thanks for the PR, this was fixed more thoroughly by supporting them, while we already have an error for missing implementations.

PR #30846, opened 2026-02-18
ENH: use raw memory allocation APIs under free-threading

### LINKED ISSUE #30494: BUG: Multithreaded scaling bottlenecks on free-threaded Python

kumaraditya303:
Fixes https://github.com/numpy/numpy/issues/30494

This PR changes `alloc.cpp` to use raw memory allocation APIs from CPython so that it can use `mimalloc` under 3.15 free-threaded builds. I have also removed the unnecessary `NPY_BEGIN_THREADS_DEF` calls to release the thread state which just adds overhead. For reference CPython also does not releases the thread state when allocating memory so this aligns numpy with CPython.

### DIFF
--- a/doc/release/upcoming_changes/30846.compatibility.rst
+++ b/doc/release/upcoming_changes/30846.compatibility.rst
@@ -1,4 +1,4 @@
 Default memory allocator change
-===============================
+-------------------------------
 NumPy now uses ``PyMem_RawMalloc`` and ``PyMem_RawFree`` as the default memory allocator,
 instead of system's ``malloc`` and ``free`` directly.
--- a/doc/release/upcoming_changes/30846.compatibility.rst
+++ b/doc/release/upcoming_changes/30846.compatibility.rst
@@ -1,0 +1,4 @@
+Default memory allocator change
+===============================
+NumPy now uses ``PyMem_RawMalloc`` and ``PyMem_RawFree`` as the default memory allocator,
+instead of system's ``malloc`` and ``free`` directly.
--- a/doc/release/upcoming_changes/30846.performance.rst
+++ b/doc/release/upcoming_changes/30846.performance.rst
@@ -1,0 +1,18 @@
+Improved scaling of ufuncs on free-threading
+============================================
+
+NumPy's ufuncs now scale significantly better on free-threading builds
+of CPython due to the following optimizations:
+
+* **Lock-free dispatch table:** The ufuncs dispatch table is now
+  implemented as a lock-free concurrent hash map, allowing multiple threads
+  to call ufuncs without contention.
+
+* **Immortal shared objects:** Certain shared objects, such as global memory
+  handlers, have been made immortal. This effectively reduces reference
+  counting contention across threads.
+
+* **Optimized memory allocation:** NumPy now utilizes ``PyMem_RawMalloc`` and
+  ``PyMem_RawFree`` for memory allocation. On Python 3.15 and newer, this
+  leverages ``mimalloc`` and significantly reduces memory allocation overhead
+  in multi-threaded workloads.
--- a/doc/release/upcoming_changes/30846.performance.rst
+++ b/doc/release/upcoming_changes/30846.performance.rst
@@ -1,5 +1,5 @@
 Improved scaling of ufuncs on free-threading
-============================================
+--------------------------------------------
 
 NumPy's ufuncs now scale significantly better on free-threading builds
 of CPython due to the following optimizations:
--- a/numpy/_core/src/multiarray/alloc.cpp
+++ b/numpy/_core/src/multiarray/alloc.cpp
@@ -252,7 +252,7 @@ npy_alloc_cache_dim(npy_uintp sz)
         sz = 2;
     }
     return _npy_alloc_cache(sz, sizeof(npy_intp), NBUCKETS_DIM, dimcache,
-                            &PyArray_malloc);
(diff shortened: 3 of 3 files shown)


numpy/_core/src/multiarray/alloc.cpp
```diff
@@ -41,6 +41,17 @@ extern "C" {
...
+#ifdef Py_GIL_DISABLED
+#define raw_malloc PyMem_RawMalloc
+#define raw_calloc PyMem_RawCalloc
+#define raw_realloc PyMem_RawRealloc
+#define raw_free PyMem_RawFree
+#else
+#define raw_malloc malloc
+#define raw_calloc calloc
+#define raw_realloc realloc
+#define raw_free free
```
seberg: I am curious what would happen if we just always use `PyMem_RawMalloc`? On current Python versions we'll get normal malloc/free unless the Python was configured explicitly with mimalloc (a custom build)?

I.e. I am getting a bit bolder with thinking that users shouldn't use `realloc` or so on NumPy arrays (now that we have the allocator API!).
And if we do that, even those that might have a chance that at least they break not by upgrading the NumPy versions, but rather when recompiling everything for Python.
(Which is a bit nicer, IMO, that is of course unless they use the stable API.)

And with that, maybe we should just do that (with a brief release note)?

seberg: (Plus, we already have `PyArray_free`, etc. macros that people may be using and maps to the `PyMem_Raw*` ones)

kumaraditya303: > And with that, maybe we should just do that (with a brief release note)?

Initially I wanted do the change for both builds but because of compat concerns on issue, I restricted it to free-threading only. IMO we should do the change for both versions.

seberg: Yeah, I am concerned about compat. But with a few arguments, I now think it should be harmless enough and likely worthwhile :).
If it would change behavior on a (non free-threaded) release build of Python I might pause for a bit. But I don't think it even does that.

kumaraditya303: Yes, by default PyMem_RawMalloc is just an alias for malloc so it is the same on normal builds.

kumaraditya303: Added a release note.

doc/release/upcoming_changes/30846.compatibility.rst
ngoldbaum, line 1: This is great but also consider adding another release note describing all the improvements you and I have merged over the past couple months to improve ufunc performance on the free-threaded build. I think you were planning to write that anyway and if you include it in the 2.5 release notes it will get a little more visibility. You could even add a link to a blog post if you end up writing one before the 2.5 release.

kumaraditya303: Added it.

ngoldbaum (approved):

doc/release/upcoming_changes/30846.compatibility.rst
```diff
@@ -0,0 +1,4 @@
+Default memory allocator change
+===============================
```
seberg:
```suggestion
-------------------------------
```

doc/release/upcoming_changes/30846.performance.rst
```diff
@@ -0,0 +1,18 @@
+Improved scaling of ufuncs on free-threading
+============================================
```
seberg:
```suggestion
--------------------------------------------
```

seberg (approved):
LGTM, since I think Nathan also doesn't think this affects things default Python builds _and_ it is unlikely to affect more than a hand-few of projects, let's get it in.

(I'll apply a tiny change, since I think the release notes section will get mangled up otherwise.)

ngoldbaum:
Pulling this in, thanks Kumar!

If anyone ends up here because of CI breakage on the short timescales or mysterious crashes on longer timescales after 2.5.0 comes out, please open an issue.

PR #30851, opened 2026-02-19
DOC: improve eig return value release note

mattip:
Continued from #30411, cc @ev-br 

Improve the release note and fix the code snippet (from [numpy-discussion mailing list post](https://mail.python.org/archives/list/numpy-discussion@python.org/message/DFXIIFTJMLQWVWKJBHV7EPUGAQSH4B4O/))

### DIFF
--- a/doc/release/upcoming_changes/30411.compatibility.rst
+++ b/doc/release/upcoming_changes/30411.compatibility.rst
@@ -1,20 +1,18 @@
-* `linalg.eig` and `linalg.eigvals` always return complex arrays. Previously, the return
-   values were depening on whether the eigenvalues happen to lie on the real line (which, for
-   a general, non-symmetric matrix, is not guaranteed).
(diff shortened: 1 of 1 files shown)


doc/release/upcoming_changes/30411.compatibility.rst
```diff
@@ -1,20 +1,18 @@
...
    w = eigvals(a)
-   if w.imag == 0:    # this is what NumPy used to do
-   w = w.real
-   ```
-
-   - if your matrix is symmetrix/hermitian, use `eigh` and `eigvalsh` instead of `eig` and `eigvals`:
-     these are guaranteed to return real values. (A common case is covariance matrices, which are
-     symmetric and positive definite by construction)
-
+   if np.any(w.imag == 0):  # this is what NumPy used to do
```
mattip, line 11: Thanks @jni

seberg (approved):
Reads well, nice improvement, thanks Matti!

Issue #29349, opened 2025-07-09
BUG? np.ma.flatten_structured_array infinite recursion when field contains iterable object


Smith-cc:
When using np.ma.flatten_structured_array on a structured array that has a field of object type containing an iterable (like a list or str), it causes infinite recursion and crashes with RecursionError: maximum recursion depth exceeded.

I would like to confirm: **Is this the expected behavior**, **or should `flatten_structured_array` guard against nested iterables like str to prevent infinite recursion?**

```python
# Example
import numpy as np

ndtype = [('col1', 'U1'), ('col2', object)]
arr = np.array([('A', ['oops'])], dtype=ndtype)

np.ma.flatten_structured_array(arr)
```
I think the root cause [source code](https://github.com/numpy/numpy/blob/v2.3.0/numpy/ma/core.py#L2559-L2612)
```python
def flatten_sequence(iterable):
    for elm in iter(iterable):
        if hasattr(elm, '__iter__'):
            yield from flatten_sequence(elm)
        else:
            yield elm
```
does not distinguish between strings and containers — so str are recursively iterated character-by-character, causing infinite recursion if nested in an object field.

The same problem also occurs below：
```python
import numpy as np

mask = np.array([(0, 0), (0, '1')], dtype=[('a', bool), ('b', bool)]) # Ok
x = np.ma.flatten_mask(mask)

mask = np.array([(0, 0), (0, '1')], dtype=[('a', bool), ('b', object)])
x = np.ma.flatten_mask(mask) # Error
```
My concern is, should structured arrays with object fields containing containers be rejected or handled differently? Or is it the user’s responsibility to sanitize such fields before calling these functions?

seberg:
It _should_ not recurse into object fields at all, I think.  It will probably do the wrong thing anyway for *any* masked array.
That said, I wouldn't be surprised if someone just uses this for non-masked arrays and gets away with it, possibly even wanting the recursion behavior.

So, to my eyes, it seems like a mess.  `numpy.lib.recfunctions.structured_to_unstructured` is probably sane (probably, because `recfunctions` is also a little used darker corner).  But it won't work with masked arrays. But you could apply it to data and mask seperately, which should work fine.

Smith-cc:
[...]

PR #30855, opened 2026-02-20
BUG: fix infinite recursion in np.ma.flatten_structured_array

antareepsarkar:
Towards resolving #29349 

Avoids recursion for both strings and objects.

### DIFF
--- a/numpy/ma/core.py
+++ b/numpy/ma/core.py
@@ -2590,7 +2590,7 @@ def flatten_structured_array(a):
 
         """
         for elm in iter(iterable):
-            if hasattr(elm, '__iter__'):
+            if not np.ndim(elm) == 0:
                 yield from flatten_sequence(elm)
             else:
                 yield elm
--- a/numpy/ma/core.py
+++ b/numpy/ma/core.py
@@ -2590,7 +2590,7 @@ def flatten_structured_array(a):
 
         """
         for elm in iter(iterable):
-            if not np.ndim(elm) == 0:
+            if hasattr(elm, "__iter__") and not isinstance(elm, (str, bytes)):
                 yield from flatten_sequence(elm)
             else:
                 yield elm
--- a/numpy/ma/tests/test_core.py
+++ b/numpy/ma/tests/test_core.py
@@ -998,6 +998,13 @@ class TestMaskedArray:
         control = np.array([[[1., 1.], ], [[2., 2.], ]], dtype=float)
         assert_equal(test, control)
         assert_equal(test.dtype, control.dtype)
+        # for strings
+        ndtype = [('a', 'U5'), ('b', [('c', 'U5')])]
+        arr = np.array([('NumPy', ('array',)), ('array', ('numpy',))], dtype=ndtype)
+        test = flatten_structured_array(arr)
+        control = np.array([['NumPy', 'array'], ['array', 'numpy']], dtype='U5')
+        assert_equal(test, control)
+        assert_equal(test.dtype, control.dtype)
 
     def test_void0d(self):
         # Test creating a mvoid object


seberg:
Thanks!  This is always a bit trickier with regressions in masked arrays and not being quite sure how this is used.

I actually, think a slightly _more_ maximal approach may be nice, but with a release note and no backporting.
If this is important to you to backport, I might try to go more minimal and just skip recursion into strings for now.

The more maximal approach would be to do the following (sorry, it is a bit more complicated, in the sense of changing the pattern!):
1. Rather than flattening each _element_, flatten the `arr.dtype` (if it is structured).  There is a function that does this in `iotools`: `flatten_dtype`.
2. We can just use `dtype = np.result_type()` on the flattened dtype.
3. We have to create the final array, without moving at least a bit to C, I suspect the easiest thing will be to iterate the fields and assign one by one, like:
   * `for i, name in enumerate(dtype.names): out[i+something] = input[name]`
   * Unfortunately, subarray fields are a bit tricky and need some thought there.
   
This approach will also change the result dtype, e.g. for a struct array of `float32` where the dtype will now still be `float32` (not just for dtypes that include objects as you have it).

The `numpy.lib.recfunctions` probably has some related functionality, although similar to masked arrays, some of the code may not be something we should adopt exactly.

antareepsarkar:
> Thanks! This is always a bit trickier with regressions in masked arrays and not being quite sure how this is used.
> 
> I actually, think a slightly _more_ maximal approach may be nice, but with a release note and no backporting. If this is important to you to backport, I might try to go more minimal and just skip recursion into strings for now.
> 
> The more maximal approach would be to do the following (sorry, it is a bit more complicated, in the sense of changing the pattern!):
> 
> 1. Rather than flattening each _element_, flatten the `arr.dtype` (if it is structured).  There is a function that does this in `iotools`: `flatten_dtype`.
> 2. We can just use `dtype = np.result_type()` on the flattened dtype.
> 3. We have to create the final array, without moving at least a bit to C, I suspect the easiest thing will be to iterate the fields and assign one by one, like:
>    
>    * `for i, name in enumerate(dtype.names): out[i+something] = input[name]`
>    * Unfortunately, subarray fields are a bit tricky and need some thought there.
> 
> This approach will also change the result dtype, e.g. for a struct array of `float32` where the dtype will now still be `float32` (not just for dtypes that include objects as you have it).
> 
> The `numpy.lib.recfunctions` probably has some related functionality, although similar to masked arrays, some of the code may not be something we should adopt exactly.

Thanks for replying. 
Should I implement the maximal approach you mentioned?

antareepsarkar:
@seberg @charris 
The last changes skip recursion for strings. Maybe, it can be just used for backporting now.
Then, another PR with the maximal approach mentioned can be done.

ngoldbaum:
https://github.com/numpy/numpy/pull/30872 fixes the test failures seen here, a rebase should fix it.

numpy/ma/core.py
```diff
@@ -2590,7 +2590,7 @@ def flatten_sequence(iterable):
 
         """
         for elm in iter(iterable):
-            if hasattr(elm, '__iter__'):
+            if not np.ndim(elm) == 0:
```
seberg: If we do this minimal thing, I am tempted to just add `not isinstance(elem, (str, bytes))`, TBH.
`np.ndim()` may be slow and I don't consider this the true fix either way, so I am probably more happy with a practical 98% fix than something that looks like a neat fix but isn't really.

seberg:
Thanks, let's get this minimal approach in then!  FWIW, I think the real fix is still to revamp the approach to walk the dtype fields.

PR #30858, opened 2026-02-20
MAINT: bump ``ruff`` to ``0.15.2``

jorenham:
ref https://github.com/numpy/numpy/pull/30856#issuecomment-3936493429

The default ruleset changed in 0.15.2, so I had to tweak the config a bit to avoid 2510 ruff errors. 
It's probably a good idea that we (gradually) address some of these, but doing so here would be out of scope. For reference, here's the overview:

```
642     PLR2044 [*] empty-comment
376     RUF012  [ ] mutable-class-default
163     RUF059  [ ] unused-unpacked-variable
116     PLR0124 [ ] comparison-with-itself
108     RET504  [ ] unnecessary-assign
 94     FURB167 [*] regex-flag-alias
 89     RUF022  [-] unsorted-dunder-all
 88     BLE001  [ ] blind-except
 79     PYI047  [ ] unused-private-type-alias
 54     PYI015  [*] assignment-default-in-stub
 53     SIM102  [ ] collapsible-if
 47     SIM118  [-] in-dict-keys
 42     RUF010  [*] explicit-f-string-type-conversion
 42     SIM201  [ ] negate-equal-op
 39     PLR0402 [*] manual-from-import
 37     PYI042  [ ] snake-case-type-alias
 35     RUF100  [*] unused-noqa
 33     SIM117  [-] multiple-with-statements
 29     PLW0602 [ ] global-variable-not-assigned
 29     PYI010  [*] non-empty-stub-body
 25     PYI026  [*] type-alias-without-annotation
 21     DTZ001  [ ] call-datetime-without-tzinfo
 21     PYI046  [ ] unused-private-protocol
 20     PT031   [ ] pytest-warns-with-multiple-statements
 17     S110    [ ] try-except-pass
 16     SIM115  [ ] open-file-with-context-handler
 14     EXE001  [ ] shebang-not-executable
 14     TRY004  [ ] type-check-without-type-error
 13     PYI017  [ ] complex-assignment-in-stub
 13     PYI034  [ ] non-self-return-type
 13     TRY002  [ ] raise-vanilla-class
 12     PLW1510 [ ] subprocess-run-without-check
 11     PT014   [ ] pytest-duplicate-parametrize-test-cases
 11     PYI052  [ ] unannotated-assignment-in-stub
 10     PLC0206 [ ] dict-index-missing-items
  9     TRY300  [ ] try-consider-else
  5     D419    [ ] empty-docstring
  5     FURB105 [*] print-empty-string
  5     PLR1704 [ ] redefined-argument-from-local
  5     PLW0177 [ ] nan-comparison
  5     SIM114  [*] if-with-same-arms
  4     DTZ005  [ ] call-datetime-now-without-tzinfo
  4     PLR1730 [*] if-stmt-min-max
  4     S102    [ ] exec-builtin
  3     N999    [ ] invalid-module-name
  3     PLW0127 [ ] self-assigning-variable
  3     PYI064  [*] redundant-final-literal
  3     SIM103  [ ] needless-bool
  3     SIM208  [*] double-negation
  2     FURB122 [*] for-loop-writes
  2     PYI029  [*] str-or-repr-defined-in-stub
  2     PYI032  [*] any-eq-ne-annotation
  2     PYI063  [ ] pep484-style-positional-only-parameter
  2     PYI066  [ ] bad-version-info-order
  1     DTZ011  [ ] call-date-today
  1     PLR1711 [*] useless-return
  1     PLR1733 [*] unnecessary-dict-index-lookup
  1     PLW0406 [ ] import-self
  1     PYI002  [ ] complex-if-statement-in-stub
  1     PYI018  [ ] unused-private-type-var
  1     PYI045  [ ] iter-method-return-iterable
  1     PYI055  [*] unnecessary-type-union
  1     PYI059  [ ] generic-not-last-base-class
  1     RUF046  [*] unnecessary-cast-to-int
  1     SIM202  [ ] negate-not-equal-op
  1     TRY201  [ ] verbose-raise
  1     TRY203  [ ] useless-try-except
Found 2510 errors.
[*] 1103 fixable with the `--fix` option (406 hidden fixes can be enabled with the `--unsafe-fixes` option).
```

(The `PLR2044` is intentional and helps make the stubs a bit less unreadable, so we should ignore that one)

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -45,7 +45,7 @@ dependencies:
   - breathe>4.33.0
   # For linting
   - cython-lint
-  - ruff=0.14.7
+  - ruff=0.15.2
   - gitpython
   # Used in some tests
   - cffi
--- a/numpy/_core/numeric.pyi
+++ b/numpy/_core/numeric.pyi
@@ -1046,9 +1046,11 @@ def isfortran(a: ndarray | generic) -> py_bool: ...
 def argwhere(a: ArrayLike) -> _Array2D[np.intp]: ...
 def flatnonzero(a: ArrayLike) -> _Array1D[np.intp]: ...
 
+# NOTE: we ignore UP047 because inlining `_AnyScalarT` would result in a lot of code duplication
+
 # keep in sync with `convolve` and `ma.core.correlate`
 @overload
-def correlate(
+def correlate(  # noqa: UP047
     a: _ArrayLike1D[_AnyNumericScalarT], v: _ArrayLike1D[_AnyNumericScalarT], mode: _CorrelateMode = "valid"
 ) -> _Array1D[_AnyNumericScalarT]: ...
 @overload
@@ -1068,7 +1070,7 @@ def correlate(
 
 # keep in sync with `correlate`
 @overload
-def convolve(
+def convolve(  # noqa: UP047
     a: _ArrayLike1D[_AnyNumericScalarT], v: _ArrayLike1D[_AnyNumericScalarT], mode: _CorrelateMode = "valid"
 ) -> _Array1D[_AnyNumericScalarT]: ...
 @overload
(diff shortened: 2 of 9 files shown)


charris:
The failing test is unrelated.

charris:
Thanks Joren.

PR #30859, opened 2026-02-21
ENH: add out parameter to numpy.fromfile for memory-stable I/O (#30777)

### LINKED ISSUE #30777: ENH: add out parameter to numpy.fromfile for memory-stable I/O
## Context

I am working with large-scale 3D LiDAR point cloud datasets (like nuScenes) on HPC clusters. My workflow involves reading millions of small binary files using `np.fromfile`.

## The Problem

Even when variables are overwritten, repeated calls to `np.fromfile` cause a continuous linear growth in memory (observed by the server jobs manager), leading to OOM (Out of Memory) kills by the job scheduler. This is likely due to memory fragmentation in the system allocator (glibc malloc) caused by the constant allocation/deallocation of new `ndarrays`.

A very simple code that was able to overload the memory:

```
import os
import numpy as np
import gc 

files = os.listdir('PATH')
qnt_reads = 100_000_000

# Caminho base para evitar f-string repetitiva e longa
base_path = 'PATH'
num_files = len(files)

for i in tqdm(range(qnt_reads)):
    file_path = os.path.join(base_path, files[i % num_files])
   
    points = np.fromfile(file_path, dtype=np.float32)

    del points
   
    if i % 1000 == 0:
        gc.collect()
```

## Proposed Feature 

Add an `out` parameter to `numpy.fromfile`, allowing users to provide a pre-allocated array (or buffer).

Rishabh-git10:
Closes #30777.

### Context
When reading high-frequency small binary files via `numpy.fromfile`, a new array is allocated each time, which can lead to system allocator fragmentation or OOM kills on HPC clusters. 

### What was changed
This PR adds an `out` parameter to `numpy.fromfile`, allowing users to read data directly into pre-allocated memory buffers, completely bypassing `malloc`.

1. Modified `PyArg_ParseTupleAndKeywords` in `multiarraymodule.c` to accept an `out` object (`$OO`).
2. Added C-contiguous, writeable, and shape/dtype size-matching validation for the `out` buffer.
3. Implemented direct `fread` into `PyArray_DATA(out)` if the buffer is provided and valid.
4. Added `test_fromfile_out` to `test_arrayobject.py` to verify memory sharing and exceptions.

### Proof of memory reuse:
```python
import numpy as np
import tempfile
import os

expected_data = np.array([1.5, 2.5, 3.5, 4.5], dtype=np.float32)
fd, filepath = tempfile.mkstemp()
with os.fdopen(fd, 'wb') as f:
    f.write(expected_data.tobytes())

buffer = np.empty(4, dtype=np.float32)
arr_out = np.fromfile(filepath, dtype=np.float32, out=buffer)

print(f"arr_out is buffer: {arr_out is buffer}") 
# True
print(f"Shares memory: {np.shares_memory(buffer, arr_out)}") 
# True

os.remove(filepath)

### CODE REVIEW

**Comment by Rishabh-git10** [comment]
I was reading through the later half of the discussion in #30777 regarding whether `out=` makes sense for `fromfile` given the 'unknown size' and consistency constraints mentioned by @seberg and @ikrommyd.
I went ahead and implemented the C-level bypass and tests anyway so that we have a concrete implementation to evaluate rather than just theory. If the consensus is that this pollutes the API or causes too many edge cases with count, I completely understand if we decide to close this! Just wanted to put the code on the table to help make the final decision easier.

File: numpy/_core/tests/test_arrayobject.py [hunk 1]
```diff
@@ -93,3 +93,34 @@ def test_cleanup_with_refs_non_contig():
...
+    
+    expected = np.array([1.5, 2.5, 3.5, 4.5], dtype=np.float32)
+    fd, filepath = tempfile.mkstemp()
+    try:
+        with os.fdopen(fd, 'wb') as f:
+            f.write(expected.tobytes())
+        
+        # Standard out mapping
+        buffer = np.empty(4, dtype=np.float32)
+        res = np.fromfile(filepath, dtype=np.float32, out=buffer)
```
seberg: I don't think this makes sense, if you pass `out=` there is no point in passing `dtype=`.

numpy/_core/tests/test_arrayobject.py
```diff
@@ -93,3 +93,34 @@ def test_cleanup_with_refs_non_contig():
...
+        assert res is buffer
+        assert_array_equal(res, expected)
+
+        # Using 'count' with 'out'
+        buffer2 = np.empty(4, dtype=np.float32)
+        res2 = np.fromfile(filepath, dtype=np.float32, count=2, out=buffer2)
```
seberg: I also don't agree with this. If `out` is passes the `count` should probably just be identical.
There may be an argument that passing `count=len(out)` has the meaning of "it's OK if the file is bigger".

numpy/_core/tests/test_arrayobject.py
```diff
@@ -93,3 +93,34 @@ def test_cleanup_with_refs_non_contig():
...
+        assert res2 is buffer2
+        assert_array_equal(res2[:2], expected[:2])
+        
+        # Error on non-contiguous array
+        buffer3 = np.empty(8, dtype=np.float32)[::2]
```
seberg, line 127: Another interesting thing is whether `out` is also 1-D, I think it probably should be.

seberg (commented):
Thanks, I am honestly not sure I find it valuable enough. The context you mention is of course just the reason for the discussion, it isn't an actual reason.

The only reason would be to try and add `out=` to most functions in NumPy, which is generally fair, but pretty niche here.

So not sure where it'll go. I may need another maintainer being slightly in favor to feel like pursuing. But, I added some comments anyway to the tests, since I think the approach wouldn't be quite right.

numpy/_core/tests/test_arrayobject.py
Rishabh-git10: Yup, I removed it from the tests and added a strict C-level guard (`!PyArray_EquivTypes`) to reject the call if a user explicitly passes a `dtype` that doesn't match the `out` buffer for any silent buffer overflows.

numpy/_core/tests/test_arrayobject.py
Rishabh-git10: Makes sense. I added a validation step (`nin != -1 && nin != PyrArray_SIZE(out)`) that forces any explicitly provided `count` to match the buffer size. Added tests to verify the `ValueError` if they mismatch.

numpy/_core/tests/test_arrayobject.py
Rishabh-git10: Since `fromfile` historically only returns flat arrays, I enforced this by adding a `PyArray_NDIM(out) != 1` check to reject any multi-dimensional buffers.

Rishabh-git10:
Pushed a commit addressing all these edge cases (details in the inline replies above).
Let's see what the other maintainers think about this to decide whether to pursue it further.

PR #30861, opened 2026-02-22
STY: enable ``RUF012``  (``mutable-class-default``)

jorenham:
It seems like a good rule to enable:
https://docs.astral.sh/ruff/rules/mutable-class-default/


charris:
Needs rebase.

charris (approved):

charris:
Thanks Joren.

PR #30867, opened 2026-02-24
ENH: replace std::unordered_* with absl::flat_hash_* in np.unique

math-hiyoko:
close #30866


math-hiyoko:
Current commit (1af7b45) fails because vendoring process of abseil doesn't exist yet.

rgommers (changes requested):
Thanks for working on this performance improvement @math-hiyoko. `unique` performance is important, so in principle this is very welcome.

That said, a dependency on Abseil that isn't header-only (which this isn't AFAICT) is a nonstarter. If you can think of a way to make this header-only, it's still very debatable whether we'd want that. I'll comment more on gh-30866.

A performance-focused PR needs some benchmarks. It'd be useful to add them to the PR description, that may inform the decision on whether to merge it.

PR #30869, opened 2026-02-24
BUG: fix uint input for triu_indices and deprecate non-int

antareepsarkar:
resolves #29488 
Does not raise an error when any or all of of `n`, `m` and `k` are unsigned integers.

### DIFF
--- a/numpy/lib/_twodim_base_impl.py
+++ b/numpy/lib/_twodim_base_impl.py
@@ -25,6 +25,7 @@ from numpy._core.numeric import (
     nonzero,
     ones,
     promote_types,
+    unsignedinteger,
     where,
     zeros,
 )
@@ -388,20 +389,19 @@ def diagflat(v, k=0):
     return conv.wrap(res)
 
 def _to_int(n):
-    a = int(n)
-    dtype_not_int = True
 
-    if hasattr(n, "dtype") and n.dtype.kind in "iu":
-        dtype_not_int = False
-
-    if dtype_not_int and not isinstance(n, int):
+    try:
+        n = operator.index(n)
+    except TypeError:
+        # Deprecated NumPy 2.5, 2026-02
         warnings.warn(
             (f"Cannot convert {type(n).__name__} safely to an integer."
-             "This will raise an error in future versions(Deprecated NumPy 2.4)"),
+             "This will raise an error in future versions (Deprecated NumPy 2.5)"),
             DeprecationWarning,
             skip_file_prefixes=(os.path.dirname(__file__),),
         )
-    return a
+
+    return n
 
 @finalize_array_function_like
 @set_module('numpy')
@@ -1159,8 +1159,8 @@ def triu_indices(n, k=0, m=None):
 
     """
 
-    if hasattr(k, "dtype") and k.dtype.kind == "u":
-        k = int(k)
+    if isinstance(k, unsignedinteger):
+        k = operator.index(k)
 
     tri_ = ~tri(n, m, k=k - 1, dtype=bool)
 
--- a/numpy/lib/_twodim_base_impl.py
+++ b/numpy/lib/_twodim_base_impl.py
@@ -3,6 +3,7 @@
 """
 import functools
 import operator
+import os
 import warnings
 
 from numpy._core import iinfo, overrides
@@ -386,13 +387,21 @@ def diagflat(v, k=0):
 
     return conv.wrap(res)
 
-def _warn_if_float(n, stacklevel=3):
-    if isinstance(n, float):
+def _to_int(n):
+    a = int(n)
+    dtype_not_int = True
+
+    if hasattr(n, "dtype") and n.dtype.kind in "iu":
+        dtype_not_int = False
+
+    if dtype_not_int and not isinstance(n, int):
         warnings.warn(
-            f"Found {n} as an argument, floats as an input have been deprecated",
+            (f"Cannot convert {type(n).__name__} safely to an integer."
+             "This will raise an error in future versions(Deprecated NumPy 2.4)"),
             DeprecationWarning,
-            stacklevel=stacklevel
+            skip_file_prefixes=(os.path.dirname(__file__),),
         )
+    return a
 
 @finalize_array_function_like
 @set_module('numpy')
@@ -437,20 +446,20 @@ def tri(N, M=None, k=0, dtype=float, *, like=None):
            [1.,  1.,  0.,  0.,  0.]])
 
     """
-    try:
-        N = operator.index(N)
-    except TypeError:
-        _warn_if_float(N, stacklevel=4)
+
+    if not isinstance(N, int):
+        N = _to_int(N)
+
     if like is not None:
         return _tri_with_like(like, N, M=M, k=k, dtype=dtype)
 
     if M is None:
         M = N
-    else:
-        try:
-            M = operator.index(M)
-        except TypeError:
-            _warn_if_float(M, stacklevel=4)
+    elif not isinstance(M, int):
+        M = _to_int(M)
+
+    if not isinstance(k, int):
+        k = _to_int(k)
 
     m = greater_equal.outer(arange(N, dtype=_min_int(0, N)),
                             arange(-k, M - k, dtype=_min_int(-k, M - k)))
@@ -1150,10 +1159,8 @@ def triu_indices(n, k=0, m=None):
 
     """
 
-    try:
-        k = operator.index(k)
-    except TypeError:
-        _warn_if_float(k)
+    if hasattr(k, "dtype") and k.dtype.kind == "u":
+        k = int(k)
 
     tri_ = ~tri(n, m, k=k - 1, dtype=bool)
 
--- a/numpy/lib/_twodim_base_impl.py
+++ b/numpy/lib/_twodim_base_impl.py
@@ -20,12 +20,12 @@ from numpy._core.numeric import (
     int16,
     int32,
     int64,
+    integer,
     intp,
     multiply,
     nonzero,
     ones,
     promote_types,
-    unsignedinteger,
     where,
     zeros,
 )
@@ -388,21 +388,6 @@ def diagflat(v, k=0):
 
     return conv.wrap(res)
 
-def _to_int(n):
-
-    try:
-        n = operator.index(n)
-    except TypeError:
-        # Deprecated NumPy 2.5, 2026-02
-        warnings.warn(
-            (f"Cannot convert {type(n).__name__} safely to an integer."
-             "This will raise an error in future versions (Deprecated NumPy 2.5)"),
-            DeprecationWarning,
-            skip_file_prefixes=(os.path.dirname(__file__),),
-        )
-
-    return n
-
 @finalize_array_function_like
 @set_module('numpy')
 def tri(N, M=None, k=0, dtype=float, *, like=None):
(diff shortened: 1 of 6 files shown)


numpy/lib/_twodim_base_impl.py
```diff
@@ -19,7 +19,9 @@
     int32,
     int64,
     intp,
+    issubdtype,
     multiply,
+    ndim,
     nonzero,
     ones,
     promote_types,
```
jorenham: `unsignedinteger` can also be imported from `numpy._core.numeric`

numpy/lib/_twodim_base_impl.py
```diff
@@ -1132,6 +1134,14 @@ def triu_indices(n, k=0, m=None):
            [ 12,  13,  14,  -1]])
 
     """
+    if hasattr(n, "dtype") and issubdtype(n.dtype, unsignedinteger):
+        if ndim(n) == 0:
```
jorenham: nit: there's a ruff rule (currently disabled, but still) that would complain about collapsible if statements like these: https://docs.astral.sh/ruff/rules/collapsible-if/

numpy/lib/_twodim_base_impl.py
```diff
@@ -1132,6 +1135,18 @@ def triu_indices(n, k=0, m=None):
...
+    def convert_if_unsigned(val):
+        if (hasattr(val, "dtype") and issubdtype(val.dtype, unsignedinteger)
+        and ndim(val) == 0):
+            return operator.index(val)
+
+        return val
+
+    m = convert_if_unsigned(m)
+    n = convert_if_unsigned(n)
+    k = convert_if_unsigned(k)
```
ngoldbaum: Generally we don't like to force users who aren't affected by issues to pay the cost of fixing an issue. It seems better to fix this another way...

numpy/lib/_twodim_base_impl.py
```diff
@@ -1132,6 +1135,18 @@ def triu_indices(n, k=0, m=None):
            [ 12,  13,  14,  -1]])
 
     """
+
+    def convert_if_unsigned(val):
+        if (hasattr(val, "dtype") and issubdtype(val.dtype, unsignedinteger)
+        and ndim(val) == 0):
+            return operator.index(val)
```
ngoldbaum: just a general note: this function should be defined at module level. You only do stuff like this if you're defining a closure that captures values in the enclusing scope. This function doesn't do that.

ngoldbaum: I guess this is basically @seberg's solution from https://github.com/numpy/numpy/issues/29488#issuecomment-3140325900 but it's not quite as clean as he envisioned.

antareepsarkar: I get it and agree that the code should have been cleaner. Please tell me what approach can I take for this solution.

ngoldbaum: sorry, not sure offhand

antareepsarkar: Actually, I did not see any way of solving this without checking for the `dtype` (without changing the current behavior).
And, in the first commit I did not define any function but just checked for `n`'s and `m`'s `dtype`. But, then I thought that if `k` is an unsigned integer, it will again raise an error.
So, I defined the function to reduce the lines of code.
Now, if I remove the function and check the `dtype` of `n`, `m` and `k` because these kinds of conditional statements are used in some places, will it be okay?

seberg: Maybe see the other PR, you could even pick up there.  It was already very good, there was just one question that we either have to decide to break:
```
np.tri(3.5)
```
or add a `DeprecationWarning`.  I would probably not add a helper, except maybe to give the warning itself (because we need it in a few places here).  All it needs is a try/except.

It would be awesome if you could take the branch from the other closed PR and add commits to implement the DeprecationWarning.

seberg:
@antareepsarkar can you please add a test (ideally in `test_deprecations.py`) for the now deprecated behavior.
You'll find that the current code doesn't do the same thing as the old one.  Also, `float` isn't special at all here, you should use `int()` and if that doesn't fail give a generic warning (I forgot what we used to use in years past for this type of things, but something along "cannot convert {type(val).__name__} safely to an integer, this will raise an error in future versions (Deprecated NumPy 2.4)").

It isn't actually about floats as such, float is just the most likely kind of object coming in hitting this.

Also add a comment `# Deprecated NumPy 2.4, 2026-02-26` or similar, that is just to make it easier to keep track of deprecations.

antareepsarkar:
> @antareepsarkar can you please add a test (ideally in `test_deprecations.py`) for the now deprecated behavior. You'll find that the current code doesn't do the same thing as the old one. Also, `float` isn't special at all here, you should use `int()` and if that doesn't fail give a generic warning (I forgot what we used to use in years past for this type of things, but something along "cannot convert {type(val).**name**} safely to an integer, this will raise an error in future versions (Deprecated NumPy 2.4)").
> 
> It isn't actually about floats as such, float is just the most likely kind of object coming in hitting this.
> 
> Also add a comment `# Deprecated NumPy 2.4, 2026-02-26` or similar, that is just to make it easier to keep track of deprecations.

@seberg 
I wanted to ask about the `stacklevel`. Since, we are deprecating, we need to warn for all functions, `triu_indices`, `tril_indices`, etc. All these functions call `tri`. Again, `tri` can be called by itself. 
Probably, patching `tri` is the best thing. But, I'm getting confused about what should the `stacklevel` be?

seberg:
There is actually a new mechanism that I think we can use now: [`skip_file_prefixes=`](https://docs.python.org/3/library/warnings.html#warnings.warn).

Otherwise, it would be 2 typically, then +1 for every layer of helper function (the smallest one we can find), so in this case 3.  But, the skip function is nicer (and the only thing that I _might_ worry about is if I thought it was a warning that could happen in very hot paths, this isn't).

antareepsarkar:
> There is actually a new mechanism that I think we can use now: [`skip_file_prefixes=`](https://docs.python.org/3/library/warnings.html#warnings.warn).
> 
> Otherwise, it would be 2 typically, then +1 for every layer of helper function (the smallest one we can find), so in this case 3. But, the skip function is nicer (and the only thing that I _might_ worry about is if I thought it was a warning that could happen in very hot paths, this isn't).

Yes, but warnings at present require `stacklevel` to be present. So, I can change the behaviour to allow `skip_file_prefixes`(in this PR?) or just let `stacklevel` be `3`. Just asking to you so that I can be sure what to do before starting.

seberg:
Is this from the `ruff` check or another test? I think the `ruff` check should not complain when `skip_file_prefixes` is used, but we used to have a very old custom check, which I suspect we can just retire now.

So yeah, I think we can go with `skip_file_prefixes` if you like, at the risk that you (or I, or someone else) might have to make one more tweak here or in another PR before tests pass.

antareepsarkar:
@seberg
I did the changes. I have added `skip_file_prefixes`(should that be done in another PR?) in `test_warnings.py`.
Please review and tell if anything should be changed.

numpy/lib/_twodim_base_impl.py
```diff
@@ -385,6 +387,21 @@ def diagflat(v, k=0):
...
+    a = int(n)
+    dtype_not_int = True
+
+    if hasattr(n, "dtype") and n.dtype.kind in "iu":
+        dtype_not_int = False
+
+    if dtype_not_int and not isinstance(n, int):
+        warnings.warn(
+            (f"Cannot convert {type(n).__name__} safely to an integer."
+             "This will raise an error in future versions(Deprecated NumPy 2.4)"),
```
seberg:
```suggestion
             "This will raise an error in future versions (Deprecated NumPy 2.5)"),
```

numpy/lib/_twodim_base_impl.py
```diff
@@ -385,6 +387,21 @@ def diagflat(v, k=0):
...
     return conv.wrap(res)
 
+def _to_int(n):
+    a = int(n)
+    dtype_not_int = True
+
+    if hasattr(n, "dtype") and n.dtype.kind in "iu":
+        dtype_not_int = False
+
+    if dtype_not_int and not isinstance(n, int):
```
seberg: Nit, add a `# Deprecated NumPy 2.5, 2026-...`

numpy/lib/_twodim_base_impl.py
```diff
@@ -385,6 +387,21 @@ def diagflat(v, k=0):
 
     return conv.wrap(res)
 
+def _to_int(n):
+    a = int(n)
+    dtype_not_int = True
+
+    if hasattr(n, "dtype") and n.dtype.kind in "iu":
+        dtype_not_int = False
```
seberg: I suspect you have a reason, but why did you get away from `operator.index?`

What I wanted to nudge towards (sorry, sometimes I try to be a bit fuzzy), was to do:
```
try:
    n = operator.index(n)
except TypeError:
    n = int(n)
    # successfully converted, but not via `operator.index()`
    warnings.warn(...)
```

seberg (commented):
Thanks, great you tracked down how to keep CI passing!  I would like to go back to the old `operator.index` pattern with a try/except though.

numpy/lib/_twodim_base_impl.py
antareepsarkar: I thought you wanted to use `int()` rather than `operator.index` https://github.com/numpy/numpy/pull/30869#issuecomment-3964676755.
I will use `operator.index` now. The thing is, if we do `int(n)`, it takes the floor of floats. But, previously the ceil was being taken i.e. now `np.tri(3.14)` will be `np.tri(3)` but before the change it was `np.tri(4)`. So, it will be kind of a breaking change. 
I just want to be sure from you what needs to be done in this case.

antareepsarkar: Actually, I think we can just warn in the except block and not do anything as we just want to warn as of now. Anyways, it's upto you to decide what should be done.

seberg: Ah! I had not realized that the `arange` led to the ceil being taken.

Sorry, probably the earlier iteration was actually the closest and I didn't think right.  If `operator.index()` fails, we should just keep doing exactly what we were doing before.
(And I think that was right as nothing at all, I guess...  sorry for slipping into thinking that wasn't right)

antareepsarkar: Thanks.
Now, I have kind of a clear picture of what to do.

antareepsarkar:
@seberg 
Probably, the last commit has some problems(for example, if I put a string as an input, it shows both the warning and the error). 

So, as you said, I will make things as before.

Please tell me whether this needs a release note or not. And, does both the fix for `unsigned integers` and deprecation need a release note.

numpy/lib/_twodim_base_impl.py
```diff
@@ -1132,6 +1158,10 @@ def triu_indices(n, k=0, m=None):
            [ 12,  13,  14,  -1]])
 
     """
+
+    if isinstance(k, unsignedinteger):
```
seberg: This `unsignedinteger` check is new, it looks a bit like a left-over?

As to your note about giving warning + error.  Great observation!  I could live with it, but you are right it isn't great in the, maybe unlikely, case anyone actually relies on the `TypeError`.
This would not have been a problem if `int()` wasn't a regression due to `ceil`.

I suppose we could change things to do something like:
```
try:
    n = operator.index(n)
    k = operator.index(k)
except TypeError:
    deprecation_warning = True
else:
    deprecation_warning = False

<old code at least until it converts>

if deprecation_warning:
    _give_int_warning()
```
Although if we want to print the type, we would have to remember _which_ failed and the above doesn't do it.

Maybe you want to try that and decide for yourself if it it doesn't get too annoying?  These functions themself fairly simply, so I think it would be fine.

(We have been iterating a quite quickly, I probably should slow down a bit :))

antareepsarkar:
Sorry, just forgot to put the comment for deprecated.
I think that using list will not affect speed. There are only 3 iterations at max.

Please tell me whether things need to be changed.

antareepsarkar:
@seberg 
Sorry, just forgot to put the comment for deprecated.
I think that using list will not affect speed. There are only 3 iterations at max.

Please tell me whether things need to be changed.

numpy/lib/_twodim_base_impl.py
```diff
@@ -1132,6 +1160,11 @@ def triu_indices(n, k=0, m=None):
            [ 12,  13,  14,  -1]])
 
     """
+
+    # To make sure unsigned integers, etc. get converted to int
+    if isinstance(k, integer):
+        k = operator.index(k)
```
seberg, line 1164: If you just do the try/except dance here, then that seems just as well?  The worst case would be that the deprecation warning is given twice.

(that will also mean we'll start to reject e.g. NumPy booleans just like in `tri` itself. -- for better or worse)

numpy/lib/_twodim_base_impl.py
```diff
@@ -429,18 +431,44 @@ def tri(N, M=None, k=0, dtype=float, *, like=None):
...
+        w.append(k)
 
     m = greater_equal.outer(arange(N, dtype=_min_int(0, N)),
                             arange(-k, M - k, dtype=_min_int(-k, M - k)))
 
     # Avoid making a copy if the requested type is already bool
     m = m.astype(dtype, copy=False)
 
+    # warn for all deprecated
+    for i in w:
```
seberg: Thanks, FWIW, it seems fine to me to just the single first warning.  Which then allows you to write it as:
```
warning_for_type = False
try: ...
except:
    warning_for_type = warning_for_type or type(N)
```

seberg (commented):
Thanks, two comments, but honestly this is fine and if you prefer I might just push something (it's my problem if I have the urge to change it to something I feel is slightly prettier :)).

antareepsarkar:
So, I just did the changes you told. I am just trying to avoid the double warning for `np.triu_indices`.

antareepsarkar:
@seberg 
Are the changes okay?

seberg (approved):
Yeah, thanks @antareepsarkar. let's go with this.

PR #30871, opened 2026-02-24
MAINT: remove download-wheels.py and update release-requirements.txt

rgommers:
We're releasing from `numpy/numpy-release` with trusted publishing, hence no more manual shuffling of wheels.

The main motivation of this change is that the script pulled in extra dependencies like `urllib3`, which are triggering some (non-applicable) security scanner alerts.

There is probably more to do here, but I want to limit it to this change, and see how much of the security alerts it cleans up.

[ci skip]


andyfaff:
This looks ok to me.

charris (approved):

charris:
Thanks Ralf.

PR #30872, opened 2026-02-24
TYP/CI: separate ``typing_requirements.txt``

jorenham:
For (runtime) testing there is no need to install mypy and pyrefly, but because they were listed in `test_requirements.tst`, they are installed in more workflows that needed. 
By splitting off the the type-checker requirements into a new `typing_requirements.txt`, we can be more precise about this. The hope is that this will also resolve the armhf CI issues noted in https://github.com/numpy/numpy/pull/30863#issuecomment-3947960648

### DIFF
--- a/.github/workflows/mypy_primer.yml
+++ b/.github/workflows/mypy_primer.yml
@@ -41,7 +41,7 @@ jobs:
         shell: bash
         run: |
           cd numpy_to_test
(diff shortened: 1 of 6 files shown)


ganesh-k13:
This change is in https://github.com/numpy/numpy/pull/26255 as well, I can re-apply this

charris:
Might add an explanation in the new file so that it doesn't get merged back into test_requirements at some point.

charris:
Thanks Joren.

PR #30877, opened 2026-02-25
BUG: Fix buffer overrun in CPU baseline validation

stratakis:
Use sizeof() - 1 to exclude the string literal's null terminator when copying feature names into baseline_failure, preventing a buffer overrun and embedded nulls that truncated the error message.

Uncovered by Coverity static analysis

Fixes: #30477


seberg:
Thanks, but isn't the `*fptr = 0` still wrong, i.e. the `if` condition should be changed to `fptr > buffer` and that branch already has the right write?

stratakis:
> Thanks, but isn't the `*fptr = 0` still wrong, i.e. the `if` condition should be changed to `fptr > buffer` and that branch already has the right write?

Right. Rebased.

seberg (approved):
Thanks, should go in once tests pass.  Out of lazyness, I just pushed a fix here for that new warning in MSVC, no idea why it started showing up now.

From what I can tell briefly, there is on nice way to test this so...

PR #30879, opened 2026-02-25
ENH: Test ._is_numeric not .char in np.testing.assert_equal

hawkinsp:
This PR makes assert_equal correctly handle NaN tests for user-defined float types.

This fixes an issue found when attempting to port ml_dtypes to use the NumPy 2 user dtype APIs (https://github.com/jax-ml/ml_dtypes/pull/360).

@seberg

### DIFF
--- a/numpy/testing/_private/utils.py
+++ b/numpy/testing/_private/utils.py
@@ -744,7 +744,7 @@ def assert_array_compare(comparison, x, y, err_msg='', verbose=True, header='',
     ox, oy = x, y
 
     def isnumber(x):
-        return x.dtype.char in '?bhilqpBHILQPefdgFDG'
+        return x.dtype.kind in 'biufc'
 
     def istime(x):
         return x.dtype.char in "Mm"
--- a/numpy/testing/_private/utils.py
+++ b/numpy/testing/_private/utils.py
@@ -744,7 +744,7 @@ def assert_array_compare(comparison, x, y, err_msg='', verbose=True, header='',
     ox, oy = x, y
 
     def isnumber(x):
-        return x.dtype.kind in 'biufc'
+        return type(x.dtype)._is_numeric
 
     def istime(x):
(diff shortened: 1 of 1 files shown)


seberg:
Hmmm, maybe we can use `type(dtype)._is_numeric` and then set `NPY_DT_NUMERIC` flag over there that is a bit private, and it would be nice to expose that as something like `isinstance(dtype, numpy.dtypes.NumericDType)` or so, but that is fine.

That might be a bit more robust (i.e. should work even if you use a random character) -- although, I am not sure if you need something around `isnan` here too...

ngoldbaum:
> Hmmm, maybe we can use type(dtype)._is_numeric and then set NPY_DT_NUMERIC flag over there that is a bit private

@hawkinsp FWIW I added `NPY_DT_NUMERIC` for uses like this. I also agree with sebastian it would be really nice to have a better story for registering dtypes in an ontology or hierarchy that's centrally managed by NumPy.

This is very load-bearing code in the ecosystem so I'm a little nervous about changing behavior for legacy dtypes. But we can always add a check for an `_is_numeric` attribute, since I don't think this is performance-sensitive code, at least not at the level of one `getattr` call.

numpy/testing/_private/utils.py
```diff
@@ -744,7 +744,7 @@ def assert_array_compare(comparison, x, y, err_msg='', verbose=True, header='',
     ox, oy = x, y
 
     def isnumber(x):
-        return x.dtype.char in '?bhilqpBHILQPefdgFDG'
+        return x.dtype.kind in 'biufc'
```
seberg:
```suggestion
        return type(x.dtype)._is_numeric
```
Let's just try it, if tests pass, I think it should be fine.

seberg:
Thanks, let's try with this, since Nathan also thought it should be a good solution.

PR #26130, opened 2024-03-25
ENH: Optimize performance of np.atleast_1d

eendebakpt:
In this PR we improve the performance of `np.atleast_1d` for the single argument case (and slightly improve for the other cases). 
The `np.atleast_1d` is used in many other methods as part of the input pre-processing ([Github search for np.atleast_1d](https://github.com/search?q=np.atleast_1d+language%3APython&type=code&l=Python)).

Benchmark
```
atleast_1d(1): Mean +- std dev: [main] 489 ns +- 16 ns -> [pr] 422 ns +- 9 ns: 1.16x faster
atleast_1d(x): Mean +- std dev: [main] 224 ns +- 9 ns -> [pr] 159 ns +- 7 ns: 1.41x faster
atleast_1d(1, x ): Mean +- std dev: [main] 592 ns +- 26 ns -> [pr] 605 ns +- 24 ns: 1.02x slower
polyadd: Mean +- std dev: [main] 2.50 us +- 0.11 us -> [pr] 2.26 us +- 0.08 us: 1.10x faster

Geometric mean: 1.15x faster
```

<details><summary>Benchmark script</summary>

```
import pyperf

runner = pyperf.Runner()

setup="""
import numpy as np
from numpy.polynomial import Polynomial

x = np.array([2, 3, 4])
y = np.array([2, 3])

p = Polynomial([1, 2, 3.])
"""   
runner.timeit(name=f"atleast_1d(1)", stmt=f"np.atleast_1d(1)", setup=setup)
runner.timeit(name=f"atleast_1d(x)", stmt=f"np.atleast_1d(x)", setup=setup)
runner.timeit(name=f"atleast_1d(1, x )", stmt=f"np.atleast_1d(1, x )", setup=setup)
runner.timeit(name=f"polyadd", stmt=f"np.polyadd(x, y)", setup=setup)
```
</details>

Issue #2613, opened 2012-10-19
Reduce doesn't call inner loop for scalars. (Trac #2020)


numpy-gitbot:
_Original ticket http://projects.scipy.org/numpy/ticket/2020 on 2012-01-15 by @charris, assigned to unknown._

This isn't a problem for current ufuncs, but suppose there was a binary ufunc addrounded(a, b) that added the rounded values of a and b. Then addrounded.reduce(1.5) should be 1, but currently it will just return the element 1.5 without calling the inner loop that would round the result. On the other hand, it would work fine with addrounded.reduce([1.5])

seberg:
Closing, seems like a bridge/enhancement to cross when someone actually needs to add such a ufunc.  If we really need it, I think it could be done with a flag once gh-20970 is in.

I actually suspect it already works so long you make sure that the reduction has an identity, but if it has not, we cannot call the innerloop anyway.

PR #30880, opened 2026-02-25
ENH: Increase np.atleast_2d and np.atleast_3d performance

benediktjohannes:
This pull request is a features #26130 by @eendebakpt (credits to him for the great work 👍 ) and should improve the performance of `np.atleast_2d` and `np.atleast_3d`.

I've not tested the changes, but it seems quite obvious to me from looking at the code that these should be (at least in most cases) positive (and in general more positive than negative, but I'm not one hundred percent sure, so please correct me if I'm mistaken, thanks).

Credits to: @eendebakpt
Contributed by Benedikt Johannes

### DIFF
--- a/numpy/_core/shape_base.py
+++ b/numpy/_core/shape_base.py
@@ -125,7 +125,7 @@ def atleast_2d(*arys):
         if result.ndim == 0:
             result = result.reshape(1, 1)
         elif result.ndim == 1:
-            result = ary[_nx.newaxis, :]
+            result = result[_nx.newaxis, :]
         res.append(result)
     return tuple(res)
 
@@ -195,10 +195,10 @@ def atleast_3d(*arys):
         result = asanyarray(ary)
         if result.ndim == 0:
             result = result.reshape(1, 1, 1)
-        elif ary.ndim == 1:
-            result = ary[_nx.newaxis, :, _nx.newaxis]
-        elif ary.ndim == 2:
-            result = ary[:, :, _nx.newaxis]
+        elif result.ndim == 1:
+            result = result[_nx.newaxis, :, _nx.newaxis]
+        elif result.ndim == 2:
+            result = result[:, :, _nx.newaxis]
         res.append(result)
     return tuple(res)
 
--- a/numpy/_core/shape_base.py
+++ b/numpy/_core/shape_base.py
@@ -118,8 +118,8 @@ def atleast_2d(*arys):
         result = asanyarray(arys[0])
         if result.ndim == 0:
             result = result.reshape(1, 1)
-        elif ary.ndim == 1:
-            result = ary[_nx.newaxis, :]
+        elif result.ndim == 1:
+            result = result[_nx.newaxis, :]
         return result
     res = []
     for ary in arys:
@@ -191,10 +191,10 @@ def atleast_3d(*arys):
         result = asanyarray(arys[0])
         if result.ndim == 0:
             result = result.reshape(1, 1, 1)
-        elif ary.ndim == 1:
-            result = ary[_nx.newaxis, :, _nx.newaxis]
+        elif result.ndim == 1:
+            result = result[_nx.newaxis, :, _nx.newaxis]
         elif ary.ndim == 2:
-            result = ary[:, :, _nx.newaxis]
+            result = result[:, :, _nx.newaxis]
         return result
     res = []
     for ary in arys:
--- a/numpy/_core/shape_base.py
+++ b/numpy/_core/shape_base.py
@@ -114,20 +114,20 @@ def atleast_2d(*arys):
     (array([[1]]), array([[1, 2]]), array([[1, 2]]))
 
     """
+    if len(arys) == 1:
+        result = asanyarray(arys[0])
+        if result.ndim == 0:
+            result = result.reshape(1, 1)
+        return result
     res = []
     for ary in arys:
-        ary = asanyarray(ary)
-        if ary.ndim == 0:
-            result = ary.reshape(1, 1)
-        elif ary.ndim == 1:
+        result = asanyarray(ary)
+        if result.ndim == 0:
+            result = result.reshape(1, 1)
+        elif result.ndim == 1:
             result = ary[_nx.newaxis, :]
-        else:
-            result = ary
         res.append(result)
-    if len(res) == 1:
-        return res[0]
-    else:
-        return tuple(res)
+    return tuple(res)
 
 
 def _atleast_3d_dispatcher(*arys):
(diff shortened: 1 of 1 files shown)


ngoldbaum:
> I've not tested the changes

It looks like your change introduces new failures to the unit tests (see the freebsd tests on circleci, you'll see the same thing if you run the unit tests locally).

Please make sure the unit tests locally pass before updating this PR.

Also since this PR is supposed to improve performance, please actually test it and show that it does indeed improve performance on a real-world usecase.

> This pull request is a features https://github.com/numpy/numpy/issues/2613

I'm not sure what this is supposed to mean but I'm guessing you accidentally linked the wrong issue.

benediktjohannes:
Hi, thank you very much for the feedback!

You're right, I made a mistake at linking the issue - then this all makes sense because the other PR actually already showed the performance increase (that's why I didn't test it). And I'll make sure to correct unit tests as soon as possible. 👍

benediktjohannes:
I guess that adding performance tests is not necessary in this case because it's nearly the same as in the (now correctly linked) PR and there the performance increases were shown, please also add your opinion on this @eendebakpt and thanks again for the great work! Thank you! 👍 

Should be ready for re-review, thanks!

ngoldbaum:
> I guess that adding performance tests is not necessary in this case because it's nearly the same as in the (now correctly linked) PR and there the performance increases were shown

I'm sorry, no. Please run a benchmark for the code change you're proposing. I'd be surprised if there were performance differences but IMO it's not worth skipping the check. As the person who is submitting the PR it's incumbent on you to do the necessary due diligence.

benediktjohannes:
Thanks for the feedback! I'll have a look on this tomorrow (because I'll need to install numpy at first).

> I'd be surprised if there were performance differences

Probably you meant this in comparison to the PR of @eendebakpt, didn't you? Because if you meant it in comparison to the original, I'd be rather surprised. But probably this was meant in comparison to the other PR because you said in the next sentence that you wouldn't like skipping these, so probably you meant it that way.

I just wanted to ask this so that my effort isn't useless if you think that there is no difference to the original. Thanks!

benediktjohannes:
But checks seem to fail again, so I'll have to fix something else probably at first

ngoldbaum:
> Probably you meant this in comparison to the PR of @eendebakpt, didn't you? Because if you meant it in comparison to the original, I'd be rather surprised. But probably this was meant in comparison to the other PR because you said in the next sentence that you wouldn't like skipping these, so probably you meant it that way.

I mean I want you to compare this code change you're proposing here to the current `main` branch. The test should use `atleast_2d` and `atleast_3d`. You can base it off of the benchmark from the other PR.

> But checks seem to fail again, so I'll have to fix something else probably at first

Yes, this is why we generally suggest to test PRs locally before pushing code to github.

> because I'll need to install numpy at first

In general, please don't send in code changes to projects without testing them locally first. As you saw here, there can sometimes be surprising fallout for even simple code changes.

benediktjohannes:
I've got the results for the benchmark! 👍 

Script: `numpytest.py`

    import pyperf

    runner = pyperf.Runner()

    setup="""
    import numpy as np
    from numpy.polynomial import Polynomial

    x = np.array([2, 3, 4])
    y = np.array([2, 3])

    p = Polynomial([1, 2, 3.])
    """   
    runner.timeit(name=f"atleast_2d(1)", stmt=f"np.atleast_2d(1)", setup=setup)
    runner.timeit(name=f"atleast_2d(x)", stmt=f"np.atleast_2d(x)", setup=setup)
    runner.timeit(name=f"atleast_2d(1, x )", stmt=f"np.atleast_2d(1, x )", setup=setup)

    runner.timeit(name=f"atleast_3d(1)", stmt=f"np.atleast_3d(1)", setup=setup)
    runner.timeit(name=f"atleast_3d(x)", stmt=f"np.atleast_3d(x)", setup=setup)
    runner.timeit(name=f"atleast_3d(1, x )", stmt=f"np.atleast_3d(1, x )", setup=setup)

    runner.timeit(name=f"polyadd", stmt=f"np.polyadd(x, y)", setup=setup)

(Credits to @eendebakpt for the great script👍)

Results:

Old (`numpy original`):

    atleast_2d(1): Mean +- std dev: 728 ns +- 21 ns
    atleast_2d(x): Mean +- std dev: 575 ns +- 25 ns
    atleast_2d(1, x ): Mean +- std dev: 1.16 us +- 0.07 us

    atleast_3d(1): Mean +- std dev: 761 ns +- 42 ns
    atleast_3d(x): Mean +- std dev: 635 ns +- 63 ns
    atleast_3d(1, x ): Mean +- std dev: 1.17 us +- 0.04 us

    polyadd: Mean +- std dev: 2.88 us +- 0.13 us

New (`numpy by Benedikt Johannes`):

    atleast_2d(1): Mean +- std dev: 662 ns +- 43 ns
    atleast_2d(x): Mean +- std dev: 488 ns +- 22 ns
    atleast_2d(1, x ): Mean +- std dev: 1.12 us +- 0.05 us

    atleast_3d(1): Mean +- std dev: 671 ns +- 22 ns
    atleast_3d(x): Mean +- std dev: 512 ns +- 24 ns
    atleast_3d(1, x ): Mean +- std dev: 1.15 us +- 0.06 us

    polyadd: Mean +- std dev: 2.80 us +- 0.08 us

benediktjohannes:
Test failure is unrealted (can't install build environment)

benediktjohannes:
Should I write release notes 2.5 for this change, @ngoldbaum? Or is this done by the team? I only know this from CPython.

benediktjohannes:
Because it says:

`Beyond changes to a functions docstring and possible description in the general documentation, if your change introduces any user-facing modifications they may need to be mentioned in the release notes. To add your change to the release notes, you need to create a short file with a summary and place it in doc/release/upcoming_changes. The file doc/release/upcoming_changes/README.rst details the format and filename conventions.`

And the change is kind of user facing in my opinion because the performance is changed and users might notice this (and the latest news document also includes performance improvements).

ngoldbaum:
minor performance improvements like this don't need a release note

benediktjohannes:
All checks pass, do I need to use any command like /merge?

benediktjohannes:
So without the unrelated one

ngoldbaum:
Please be patient, I added the milestone so this won't get lost before the next major release this summer.

benediktjohannes:
Thank you!

benediktjohannes:
But one question: Is everything finished that I got to do? So can I just wait for the summer without doing anything for this?

ngoldbaum:
Maybe sooner, this PR has only been open for a day or so. The project doesn't move terribly quickly and people do code review as their time and interest permits.

I was also planning to look closer at this at some point. Usually we like to leave PRs that make performance improvements and add new features open for at least a few days.

ngoldbaum:
Also, IMO this PR shouldn't be backported, so in practice very few people will be able to use this until NumPy 2.5 comes out this summer.

benediktjohannes:
Ah, okay, thank you very much for the Information! Don’t feel stressed about it!

eendebakpt (approved):
Looks good to me.

numpy/_core/shape_base.py
```diff
@@ -114,20 +114,22 @@ def atleast_2d(*arys):
...
-            result = ary.reshape(1, 1)
-        elif ary.ndim == 1:
-            result = ary[_nx.newaxis, :]
-        else:
-            result = ary
+        result = asanyarray(ary)
+        if result.ndim == 0:
+            result = result.reshape(1, 1)
+        elif result.ndim == 1:
+            result = result[_nx.newaxis, :]
```
eendebakpt, line 130: This could be `np.reshape(-1, 1)` i think. It is quite a bit faster.

ngoldbaum:
It's been seven months so it might make sense to make a new PR if the OP has gone on to other projects.

benediktjohannes:
@ngoldbaum @eendebakpt I‘m still here 🙂👍 I always wanted to get this done and merged, it’s just that I didn’t Post anything here anymore because You Said I should not be so pushy, so no Problem, we can add this Suggestion by @eendebakpt, thank you by the way for the good idea

numpy/_core/shape_base.py
benediktjohannes: Thanks for the idea! Does this apply to 2d and 3d as well?

benediktjohannes:
@eendebakpt thanks for the Like, but I think you didn’t See the comment above concerning my question to your Suggestion

numpy/_core/shape_base.py
eendebakpt: Not sure, but you could try

PR #30885, opened 2026-02-25
BUG: Fix busdaycalendar's handling of a bool array weekmask.

### LINKED ISSUE #30883: BUG: weekmask parameter in busdaycalendar is not accepting array of booleans .
### Describe the issue:

the busdaycalendar is not working properly when an array of boolean is given to weekmask. I'm using the example from the numpy page.

```python
import numpy as np

np.busdaycalendar(
    holidays=['2011-07-01', '2011-07-04', '2011-07-17'], weekmask=np.array([ True,  True, True,  True,  True, False, False]))
```

### Error message:

```shell
  File "<input>", line 1, in <module>
TypeError: 'numpy.bool' object cannot be interpreted as an intege
```

python 3.14
numpy 2.4.2

### Runtime Environment:

_No response_

I'm developing a library to handle day-count conventions and generate cash flows. This is affecting my joint calendars implementation.

WarrenWeckesser:
Closes gh-30883.

This change is simple but it has implications beyond the problem observed in gh-30883.  With this change, when `weekmask` is a sequence, the boolean array is created by evaluating the "truthiness" of the elements in the sequence.  This might be more flexible than desired.

For example, with this PR, we have the following behavior:

```
In [1]: mask = ["x", "x", "x", "x", "", "", ""]

In [2]: cal = np.busdaycalendar(weekmask=mask)

In [3]: cal.weekmask
Out[3]: array([ True,  True,  True,  True, False, False, False])
```

Before this change, that sequence of strings would result in a `TypeError`:

```
In [13]: np.__version__
Out[13]: '2.4.2'

In [14]: mask = ["x", "x", "x", "x", "", "", ""]

In [15]: cal = np.busdaycalendar(weekmask=mask)
Cell In[15], line 1
----> 1 cal = np.busdaycalendar(weekmask=mask)

TypeError: 'str' object cannot be interpreted as an integer
```

### DIFF
--- a/numpy/_core/src/multiarray/datetime_busdaycal.c
+++ b/numpy/_core/src/multiarray/datetime_busdaycal.c
@@ -159,15 +159,15 @@ invalid_weekmask_string:
             int i;
 
             for (i = 0; i < 7; ++i) {
-                long val;
+                int val;
                 PyObject *f = PySequence_GetItem(obj, i);
                 if (f == NULL) {
                     Py_DECREF(obj);
                     return 0;
                 }
 
-                val = PyLong_AsLong(f);
-                if (error_converting(val)) {
+                val = PyObject_IsTrue(f);
+                if (val == -1) {
                     Py_DECREF(f);
                     Py_DECREF(obj);
                     return 0;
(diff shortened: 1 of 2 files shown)


seberg (approved):
Thanks Warren!  I can't of a nice way to allow this but remove teh caveat so the caveat seems OK to me.
(Converting to a NumPy bool might work, but doesn't reall seem worth it here.  Doesn't mean I have any confidence that this functionality should be used, hehe ;).)

(It's a tricky world, with our bools not being integer subclasses, unlike Python bools.)

ngoldbaum:
Thanks @WarrenWeckesser!

PR #30887, opened 2026-02-26
DOC: clarify float-scalar handling in array_split

### LINKED ISSUE #29494: DOC:  Clarify float-to-int truncation behavior in numpy.array_split documentation

The current [documentation](https://numpy.org/devdocs/reference/generated/numpy.array_split.html#numpy.array_split) for `numpy.array_split` states:

> The only difference between these functions is that `array_split` allows _indices_or_sections_ to be an integer that does not equally divide the axis.

This description implies that `indices_or_sections` should be an `integer`, but it does not mention that floating-point numbers (e.g., 3.9) are accepted and **silently truncated** to integers via `int()`. This leads to potentially surprising behavior:
```
np.array_split(arr, 3.9)  # Treated as 3 sections
np.array_split(arr, 4.0)  # Treated as 4 sections
```
Moreover, this behavior is inconsistent with `np.split`, which raises a ValueError when given a float, making the difference between the two functions less about "equal division" and more about input type tolerance.

Add a clear note in the `numpy.array_split` documentation explaining the implicit conversion of float inputs to integers. For example:

> **Note**: If _indices_or_sections_ is a floating-point number, it will be truncated towards zero (equivalent to `int()`) before splitting. For example, `3.9` is treated as `3`. To avoid ambiguity, it is recommended to pass integer values.

LarytheLord:
## Summary
- clarify `numpy.array_split` docs for scalar `indices_or_sections`
- document that floating-point scalars are truncated toward zero via `int(indices_or_sections)` before splitting

Closes #29494.

## Validation
- `python3 -m compileall numpy/lib/_shape_base_impl.py`

### DIFF
--- a/numpy/lib/_shape_base_impl.py
+++ b/numpy/lib/_shape_base_impl.py
@@ -728,6 +728,9 @@ def array_split(ary, indices_or_sections, axis=0):
     into n sections, it returns l % n sub-arrays of size l//n + 1
     and the rest of size l//n.
 
+    If `indices_or_sections` is a floating-point scalar, it is truncated
+    toward zero through ``int(indices_or_sections)`` before splitting.
+
     See Also
     --------
     split : Split array into multiple sub-arrays of equal size.


LarytheLord:
Closing this to keep scope tight and reduce reviewer load. I’m focusing on #30894 only.

PR #30889, opened 2026-02-26
CI: unpin CPython in pixi CI and revert SetImmortal changes

### LINKED ISSUE #30888: CI: pixi CI is failing due to upstream changes
See e.g. https://github.com/numpy/numpy/actions/runs/22453396194/job/65041383464

@lucascolley thinks this might be related to recent changes in the compiler-rt-feedstock - ping @h-vetinari.

ngoldbaum:
Fixes #30888 (I hope)

### DIFF
--- a/pixi-packages/asan/pixi.toml
+++ b/pixi-packages/asan/pixi.toml
@@ -19,7 +19,7 @@ extra-args = ["-Csetup-args=-Db_sanitize=address", "-Csetup-args=-Dbuildtype=deb
 [package.host-dependencies]
 python.git = "https://github.com/python/cpython"
 python.subdirectory = "Tools/pixi-packages/asan"
-python.rev = "8bb5b6e8ce61da41b5affd5eb12d9cc46b5af448"
+python.rev = "main"
 
 meson-python = "*"
 cython = "*"
--- a/pixi-packages/asan/pixi.toml
+++ b/pixi-packages/asan/pixi.toml
@@ -19,7 +19,7 @@ extra-args = ["-Csetup-args=-Db_sanitize=address", "-Csetup-args=-Dbuildtype=deb
 [package.host-dependencies]
 python.git = "https://github.com/python/cpython"
 python.subdirectory = "Tools/pixi-packages/asan"
-python.rev = "main"
+python.rev = "v3.15.0a6"
 
 meson-python = "*"
 cython = "*"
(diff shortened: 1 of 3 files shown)


pixi-packages/asan/pixi.toml
```diff
@@ -19,7 +19,7 @@ extra-args = ["-Csetup-args=-Db_sanitize=address", "-Csetup-args=-Dbuildtype=deb
 [package.host-dependencies]
 python.git = "https://github.com/python/cpython"
 python.subdirectory = "Tools/pixi-packages/asan"
-python.rev = "8bb5b6e8ce61da41b5affd5eb12d9cc46b5af448"
+python.rev = "v3.15.0a6"
```
lucascolley: seems to work
```suggestion
```

ngoldbaum:
Unfortunately we might also have to back out #30825 and #30826 until alpha7 comes out.

ngoldbaum:
> Unfortunately we might also have to back out https://github.com/numpy/numpy/pull/30825 and https://github.com/numpy/numpy/pull/30826 until alpha7 comes out.

Yeah, looks like it. Ping @kumaraditya303 FYI.

ngoldbaum:
I think I'm just going to ahead and merge this to unbreak CI and add a reminder to myself to un-revert the `SetImmortal` changes when 3.15a7 comes out.

@kumaraditya303 let me know if you'd prefer I handle this a different way.

kumaraditya303:
> let me know if you'd prefer I handle this a different way.

I don't think it is a good idea to remove the fixed python rev and use the latest one, it might break again if something changes upstream. You could instead just pin it to https://github.com/python/cpython/commit/81484c5629f9d8b698a2be58f850a0e4efe48ed3 which is the immediate parent commit of https://github.com/python/cpython/commit/347d3594d3993282c3ae5a6872f36a71fbfbc734 which added the unstable APIs and keep the immortal changes.

ngoldbaum:
> I don't think it is a good idea to remove the fixed python rev and use the latest one, it might break again if something changes upstream. You could instead just pin it to https://github.com/python/cpython/commit/81484c5629f9d8b698a2be58f850a0e4efe48ed3 which is the immediate parent commit of https://github.com/python/cpython/commit/347d3594d3993282c3ae5a6872f36a71fbfbc734 which added the unstable APIs and keep the immortal changes.

Thanks! I'll try that today.

Issue #30886, opened 2026-02-26
CI: Tracking issue for using ``spin`` in CI


ganesh-k13:
## CI changes to use ``spin``

Following places still use legacy commands that needs to be fixed:

| Status | Workflow | File | PR |
|--------|--------|--------|--------|
| Done | `lint` | [linux.yml](https://github.com/numpy/numpy/blob/main/.github/workflows/linux.yml) | https://github.com/numpy/numpy/pull/25103
| PR Raised | linx.debug |[linux.yml](https://github.com/numpy/numpy/blob/main/.github/workflows/linux.yml) | https://github.com/numpy/numpy/pull/30868
| | linux.full |[linux.yml](https://github.com/numpy/numpy/blob/main/.github/workflows/linux.yml) |
| | linux.sdist | [linux.yml](https://github.com/numpy/numpy/blob/main/.github/workflows/linux.yml) |
| | linux.array_api_tests | [linux.yml](https://github.com/numpy/numpy/blob/main/.github/workflows/linux.yml) | |
| | linux.Linux_Python_312_32bit_full | [linux.yml](https://github.com/numpy/numpy/blob/main/.github/workflows/linux.yml) | |
| ?? (direct pytest) | cygwin.cygwin_build_test | [cygwin.yml](https://github.com/numpy/numpy/blob/main/.github/workflows/cygwin.yml) | |
| ?? (direct pytest) | linux_simd.intel_sde_avx512 | [linux_simd.yml](https://github.com/numpy/numpy/blob/main/.github/workflows/linux_simd.yml) | |
| ?? (direct pytest) | linux_simd.intel_sde_spr | [linux_simd.yml](https://github.com/numpy/numpy/blob/main/.github/workflows/linux_simd.yml) | |
| | windows.msvc_python32bit_no_openblas | [windows.yml](https://github.com/numpy/numpy/blob/main/.github/workflows/windows.yml) | | 
| | windows.msvc_python64bit_openblas | [windows.yml](https://github.com/numpy/numpy/blob/main/.github/workflows/windows.yml) | | 

### Notes
- items marked ?? needs to be validated before raising PRs

mattip:
I think we should intentionally leave `pip install -e .` in at least one build/test PR to prove that it still works as expected.

ganesh-k13:
Makes sense @mattip . I'll leave the `linux.full` test to use `pip install -e .` and maybe use the coverage tests to test the openblas flag.

rgommers:
[...]

PR #30891, opened 2026-02-27
CI: Use spin test for linux.full job

ganesh-k13:
## Use spin test for linux.full job

- Part of https://github.com/numpy/numpy/issues/30886

## Notes
- Have added one item to #30886 for doing code coverage that will cover both C and Python


ganesh-k13:
I'll close this for now as the test can use pytest along with pip install, will remove the coverage comments as part of coverage PR.

PR #30893, opened 2026-02-27
DOC: clarify ``NpzFile`` handling of non-``.npy`` archive members

### LINKED ISSUE #26004: DOC:  Npz files are documented to ignore npy - but they don't

Npy files are silently loaded https://github.com/numpy/numpy/blob/82eca7bae8249f2576d28dae0800147546d0604d/numpy/lib/_npyio_impl.py#L199C1-L200C37

My suggestion would be to ignore them or even better raise an exception.

LarytheLord:
## Summary
- update `numpy.lib.npyio.NpzFile` docs to match current runtime behavior
- clarify that `.npy` members are loaded as arrays, while non-`.npy` members are returned as raw bytes
- clarify `files` attribute wording: it lists all archive members, with `.npy` suffix removed where present

Closes #26004.

## Validation
- `python3 -m compileall numpy/lib/_npyio_impl.py`

### DIFF
--- a/numpy/lib/_npyio_impl.py
+++ b/numpy/lib/_npyio_impl.py
@@ -121,19 +121,21 @@ class NpzFile(Mapping):
     archive provided on construction.
 
     `NpzFile` is used to load files in the NumPy ``.npz`` data archive
-    format. It assumes that files in the archive have a ``.npy`` extension,
-    other files are ignored.
+    format. Files in the archive with a ``.npy`` extension are loaded as
+    NumPy arrays; other files can be read as raw bytes.
 
     The arrays and file strings are lazily loaded on either
     getitem access using ``obj['key']`` or attribute lookup using
-    ``obj.f.key``. A list of all files (without ``.npy`` extensions) can
+    ``obj.f.key``. A list of all files (with ``.npy`` extensions removed where
+    present) can
     be obtained with ``obj.files`` and the ZipFile object itself using
     ``obj.zip``.
 
     Attributes
     ----------
     files : list of str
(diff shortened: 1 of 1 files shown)


jorenham:
I'm not sure what `python3 -m compileall numpy/lib/_npyio_impl.py` does, but we use `spin` for building numpy and the docs: https://numpy.org/devdocs/dev/spin.html

LarytheLord:
Thanks for the note. You’re right: `compileall` was only a quick syntax sanity check on my side, not a NumPy build/doc validation step. I’ll use `spin` for validation notes on NumPy PRs going forward.

LarytheLord:
CI note: the current `GCC` / `riscv64` failures appear infra-related (`Could not resolve ports.ubuntu.com` during apt install), before test/build steps run. This does not look specific to the doc change in this PR.

LarytheLord:
Closing this to reduce review load after opening too many PRs too quickly. I apologize for the noise.

rkern:
> I'm not sure what `python3 -m compileall numpy/lib/_npyio_impl.py` does, but we use `spin` for building numpy and the docs: https://numpy.org/devdocs/dev/spin.html

It's a Claudism. It likes to check that the Python it wrote is syntactically valid with `compileall`.

LarytheLord:
Fair point, and thanks for calling it out.

I should have validated this using NumPy’s documented `spin` workflow and reported that, instead of using a generic syntax check. I’ve stopped the batch-PR behavior and I’m being more deliberate with project-specific validation and communication.

PR #30896, opened 2026-02-27
DOC: clarify ``nan_to_num(copy=...)`` behavior in NumPy 2.x

### LINKED ISSUE #29045: DOC: `np.nan_to_num(..., copy=)` uses NumPy 1.x description
### Describe the issue:

In NumPy 2.0>, the stricter semantics for the `copy=False` keyword result in a `ValueError` in `np.nan_to_num` when the input is not already a NumPy array or when a copy is required due to dtype mismatch.

This breaks previously valid and widely used patterns that worked in NumPy 1.x.

```python
import numpy as np

data = [float("nan"), float("inf"), 1.0]
np.nan_to_num(data, copy=False)  # raises ValueError in NumPy 2.0
```

### Error message:

```shell
ValueError: Unable to avoid copy while creating an array as requested.
If using `np.array(obj, copy=False)` replace it with `np.asarray(obj)` to allow a copy when needed (no behavior change in NumPy 1.x).
```

* NumPy version: 2.0.x
* Python version: 3.10+
* OS: Any

### Runtime Environment:

_No response_

#### **Cause**

Internally, `nan_to_num` uses:

```python
x = np.array(x, subok=True, copy=copy)
```

In NumPy 2.0, this raises an error if a copy is needed but `copy=False` was passed.

#### **Proposed Fix**

Instead of forcing the user to pre-convert input using `np.asarray(data)` before passing it to `nan_to_num`, the function should handle this gracefully by adapting its internal logic.

Proposed replacement:

```python
if copy is False:
    x = np.asarray(x)
else:
    x = np.array(x, subok=True, copy=copy)
```

This preserves the original intent (`copy if needed`) and avoids unnecessary burden on users to refactor existing code.

#### **Additional Context**

The [migration guide](https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword) recommends replacing `np.array(..., copy=False)` with `np.asarray(...)` when the goal is "copy if needed". This change would make `nan_to_num` robust across versions and eliminate the need for callers to pre-process data explicitly.

LarytheLord:
## Summary
- clarify `nan_to_num(copy=...)` docs to match NumPy 2.x `copy` semantics
- document all accepted modes: `True` (always copy), `None` (copy if needed), and `False` (never copy)
- explicitly note that `copy=False` raises `ValueError` when a copy is required

## Why
Issue #29045 notes that the current wording reads like NumPy 1.x behavior and can mislead users when `copy=False` is used with non-array inputs in NumPy 2.x.

Closes #29045

### DIFF
--- a/numpy/lib/_type_check_impl.py
+++ b/numpy/lib/_type_check_impl.py
@@ -394,11 +394,16 @@ def nan_to_num(x, copy=True, nan=0.0, posinf=None, neginf=None):
     ----------
     x : scalar or array_like
         Input data.
-    copy : bool, optional
-        Whether to create a copy of `x` (True) or to replace values
-        in-place (False). The in-place operation only occurs if
-        casting to an array does not require a copy.
-        Default is True.
+    copy : {bool, None}, optional
+        Controls what kind of copy is made when coercing `x` to an array.
+
+        - ``True`` (default): always create a copy.
+        - ``None``: only create a copy if needed.
+        - ``False``: never create a copy. If a copy is required, a
+          ``ValueError`` is raised.
+
+        In-place replacement only occurs when casting to an array does not
+        require a copy.
     nan : int, float, or bool or array_like of int, float, or bool, optional
         Values to be used to fill NaN values. If no values are passed
         then NaN values will be replaced with 0.0.


numpy/lib/_type_check_impl.py
```diff
@@ -394,11 +394,16 @@ def nan_to_num(x, copy=True, nan=0.0, posinf=None, neginf=None):
     ----------
     x : scalar or array_like
         Input data.
-    copy : bool, optional
-        Whether to create a copy of `x` (True) or to replace values
-        in-place (False). The in-place operation only occurs if
-        casting to an array does not require a copy.
-        Default is True.
+    copy : {bool, None}, optional
```
jorenham, line 397: The `optional` bit usually already means `None`.

```suggestion
    copy : bool, optional
```

LarytheLord:
Closing this to reduce review load after opening too many PRs too quickly. I apologize for the noise.

Issue #26932, opened 2024-07-13
DOC: `spin` command for coverage doesn't work


paddyroddy:
### Issue with current documentation:

https://github.com/numpy/numpy/blob/c1d1ff4548209ad12e3bae794eeb4aadedaefc14/doc/source/dev/index.rst?plain=1#L210

When I run this command locally

> spin test --coverage

`Error: needs valid configuration in [.spin.toml], [spin.toml] or [pyproject.toml]`

`spin test` does work.

### Idea or request for content:

Either suggest a replacement for this command to `spin` or suggest the `pytest` command use in CI https://github.com/numpy/numpy/blob/c1d1ff4548209ad12e3bae794eeb4aadedaefc14/.github/workflows/linux.yml#L148

rgommers:
This would be nice to fix indeed. There's an open PR for it: gh-24992. Cc @ganesh-k13

ganesh-k13:
Thanks for pointing to this Ralf. Will try to get https://github.com/numpy/numpy/issues/26977 resolved before merging https://github.com/numpy/numpy/pull/24992

ganesh-k13:
Coverage now works by doing `spin test --coverage`

Issue #19853, opened 2021-09-08
CI: Add code coverage for Python code


seberg:
It would be nice to have python code coverage as part of CI (same as C code coverage).  Unlike the C code coverage, it should be even easier to get very good coverage (since impossible to test error paths are fairly common in C).

I think it would be generally good.  One thing where I am noticing it now is that `loadtxt` is terrible test coverage, and code coverage of the existing code would also help with replacing that code (I can run that local for that concrete use-case).  It might also identify some sprint-able/LGTM style small contributions.

Plus, I think it might simplify review a bit: Instead of asking to make sure that error paths are covered, we can ask to make sure code coverage annotations are gone, giving a clearer picture of how many tests are missing (and why that is not good).

BvB93:
Codecov currently has a github actions plugin (https://github.com/codecov/codecov-action); it might be worth using.

ganesh-k13:
I have experimented on @BvB93 GitHub action and it seems to provide a lot of features. Here is an example of python coverage: https://github.com/ganeshkathiresan/CI-19853-NumPy-Codecov/pull/2

I have integrated the badge feature and a PR comment feature. There are also stuff like minimum %age and lots more. It looks like this:

![Screenshot_20210918_123855](https://user-images.githubusercontent.com/20969920/133879840-062ea0af-18be-406e-99fe-1ee1737f0c68.png)

Let me know if this is what you were looking for.

BvB93:
The codecov comments are currently disabled (per [.codecov.yml](https://github.com/numpy/numpy/blob/main/.codecov.yml), but I think the inline warnings for untested codepaths should still appear, right?
![Untitled](https://user-images.githubusercontent.com/43369155/133894247-ef49d06c-d523-45a3-9711-dac6bb334773.png)

seberg:
Yeah, I assume the settings we use for C just translate and we mainly have to set up the a runner for python coverage that uploads it?  I personally don't see much use in the comments and they can feel spammy, neither do we need the badge (we used to have it for C even).  Basically, we will have the annotations, and if the codecov reports show some things to look at, we can make an issue and link to the website (which also has the full reports).

mattip:
Github's upload action can now upload a (single) html page (no separate formatting css or js files). It might be nice to have a `spin coverage` command that would run `pytest --coverage` and upload the results, extra points for getting c level gcov as well

PR #30898, opened 2026-02-27
Fix `os.path.commonprefix` deprecation

hugovk:
`os.path.commonprefix` is deprecated in Python 3.15:

* https://docs.python.org/3.15/deprecations/
* https://docs.python.org/3/library/os.path.html#os.path.commonprefix
* https://sethmlarson.dev/deprecate-confusing-apis-like-os-path-commonprefix
* [python/cpython#74453](https://github.com/python/cpython/issues/74453) & [python/cpython#144436](https://github.com/python/cpython/pull/144436)

### DIFF
--- a/tools/c_coverage/c_coverage_report.py
+++ b/tools/c_coverage/c_coverage_report.py
@@ -84,7 +84,7 @@ class SourceFiles:
             if self.prefix is None:
                 self.prefix = path
             else:
-                self.prefix = os.path.commonprefix([self.prefix, path])
+                self.prefix = os.path.commonpath([self.prefix, path])
         return self.files[path]
 
     def clean_path(self, path):


ngoldbaum:
It's fine to merge this but FWIW the whole C coverage folder is dead code, at least as far as our CI is concerned.

hugovk:
Should we delete the `c_coverage` directory?

https://github.com/numpy/numpy/tree/main/tools/c_coverage

mattip:
I would like to have a `spin coverage` command that runs the tests including python and c level coverage. Wouldn't removing the directory would make that task harder?

mattip:
See also #26932, #19853.

ngoldbaum:
Ah good, I missed that. If you think this code will be useful for future plans we should merge this.

mattip:
Thanks @hugovk

PR #30901, opened 2026-02-27
CI: pin pixi asan build to 3.15.0a6 and re-introduce SetImmortal bindings

ngoldbaum:
Alternate fix for #30888 (see #30889).


ngoldbaum:
Seems to work, let's merge this so the performance fix is available in `main`. The failure in the riscv64 CI is unrelated. Thanks for the suggestion @kumaraditya303.

PR #30902, opened 2026-02-27
BUG: Fix tracemalloc tracking of allocations

### LINKED ISSUE #30900: BUG: tracemalloc reporting higher memory usage in dev numpy
### Describe the issue:

tracemalloc with numpy 2.4.2 reports allocated bytes that are roughly the size of an allocated array. In dev numpy the reported allocations are roughly 2x the array size.

```python
import tracemalloc

import numpy as np


tracemalloc.start()
arr = np.zeros((1000, 1000, 1000), dtype="f8")
_, peak = tracemalloc.get_traced_memory()
tracemalloc.stop()

print(f"{peak=}")
print(f"{arr.nbytes=}")
```

### Error message:

```shell
No error reported but when run with numpy 2.4.2 (python 3.12.4) the output is as expected:

peak=8000000168
arr.nbytes=8000000000

with development numpy (commit 316f598b06fde24b69318e70f894ec88d6b03ce9) the output is

peak=16000000168
arr.nbytes=8000000000
```

Tested with:
- python 3.12.4
- numpy either 2.4.2 or 316f598b06fde24b69318e70f894ec88d6b03ce9

### Runtime Environment:

For 316f598b06fde24b69318e70f894ec88d6b03ce9:
```
[{'numpy_version': '2.5.0.dev0+git20260222.0063401',
  'python': '3.12.4 (main, Nov 26 2024, 13:35:36) [Clang 14.0.0 '
            '(clang-1400.0.29.102)]',
  'uname': uname_result(system='Darwin', node='XXXXX', release='25.3.0', version='Darwin Kernel Version 25.3.0: Wed Jan 28 20:53:15 PST 2026; root:xnu-12377.81.4~5/RELEASE_ARM64_T6000', machine='arm64')},
                      'not_found': ['ASIMDFHM', 'SVE']}},
```
For 2.4.2:
```
[{'numpy_version': '2.4.2',
  'python': '3.12.4 (main, Nov 26 2024, 13:35:36) [Clang 14.0.0 '
            '(clang-1400.0.29.102)]',
  'uname': uname_result(system='Darwin', node='XXXXX', release='25.3.0', version='Darwin Kernel Version 25.3.0: Wed Jan 28 20:53:15 PST 2026; root:xnu-12377.81.4~5/RELEASE_ARM64_T6000', machine='arm64')},
                      'not_found': ['ASIMDFHM']}},
```

We have some package that runs tests against dev numpy. This package has tests that check for memory usage against some thresholds (to check for upstream changes that might impact memory usage). The tests started failing a few days ago. More details: https://github.com/spacetelescope/stcal/issues/525

braingram:
Prior to 316f598b06fde24b69318e70f894ec88d6b03ce9 memory allocations were tracked in the NPY_TRACE_DOMAIN domain. With the switch to the tracemalloc enabled allocator (PyMem_RawMalloc etc) memory was being tracked in both the python library standard PYMEM_DOMAIN_RAW domain and the NPY_TRACE_DOMAIN domain resulting in double tracking and reporting of allocations.

This PR untracks the allocations from the PYMEM_DOMAIN_RAW domain to restore the old behavior and fix the double counting.

Fixes: #30900

### DIFF
--- a/numpy/_core/src/multiarray/alloc.cpp
+++ b/numpy/_core/src/multiarray/alloc.cpp
@@ -294,6 +294,7 @@ PyDataMem_NEW(size_t size)
 
     assert(size != 0);
     result = PyMem_RawMalloc(size);
+    PyTraceMalloc_Untrack(PYMEM_DOMAIN_RAW, (npy_uintp)result);
     int ret = PyTraceMalloc_Track(NPY_TRACE_DOMAIN, (npy_uintp)result, size);
     if (ret == -1) {
         PyMem_RawFree(result);
@@ -311,6 +312,7 @@ PyDataMem_NEW_ZEROED(size_t nmemb, size_t size)
     void *result;
 
     result = PyMem_RawCalloc(nmemb, size);
+    PyTraceMalloc_Untrack(PYMEM_DOMAIN_RAW, (npy_uintp)result);
     int ret = PyTraceMalloc_Track(NPY_TRACE_DOMAIN, (npy_uintp)result, nmemb * size);
     if (ret == -1) {
         PyMem_RawFree(result);
@@ -340,6 +342,7 @@ PyDataMem_RENEW(void *ptr, size_t size)
     assert(size != 0);
     PyTraceMalloc_Untrack(NPY_TRACE_DOMAIN, (npy_uintp)ptr);
     result = PyMem_RawRealloc(ptr, size);
+    PyTraceMalloc_Untrack(PYMEM_DOMAIN_RAW, (npy_uintp)result);
     int ret = PyTraceMalloc_Track(NPY_TRACE_DOMAIN, (npy_uintp)result, size);
     if (ret == -1) {
         PyMem_RawFree(result);
@@ -428,6 +431,7 @@ PyDataMem_UserNEW(size_t size, PyObject *mem_handler)
     }
     assert(size != 0);
     result = handler->allocator.malloc(handler->allocator.ctx, size);
+    PyTraceMalloc_Untrack(PYMEM_DOMAIN_RAW, (npy_uintp)result);
     int ret = PyTraceMalloc_Track(NPY_TRACE_DOMAIN, (npy_uintp)result, size);
     if (ret == -1) {
         handler->allocator.free(handler->allocator.ctx, result, size);
@@ -446,6 +450,7 @@ PyDataMem_UserNEW_ZEROED(size_t nmemb, size_t size, PyObject *mem_handler)
         return NULL;
     }
     result = handler->allocator.calloc(handler->allocator.ctx, nmemb, size);
+    PyTraceMalloc_Untrack(PYMEM_DOMAIN_RAW, (npy_uintp)result);
     int ret = PyTraceMalloc_Track(NPY_TRACE_DOMAIN, (npy_uintp)result, nmemb * size);
     if (ret == -1) {
         handler->allocator.free(handler->allocator.ctx, result, size);
@@ -482,6 +487,7 @@ PyDataMem_UserRENEW(void *ptr, size_t size, PyObject *mem_handler)
     assert(size != 0);
     PyTraceMalloc_Untrack(NPY_TRACE_DOMAIN, (npy_uintp)ptr);
     result = handler->allocator.realloc(handler->allocator.ctx, ptr, size);
+    PyTraceMalloc_Untrack(PYMEM_DOMAIN_RAW, (npy_uintp)result);
     int ret = PyTraceMalloc_Track(NPY_TRACE_DOMAIN, (npy_uintp)result, size);
     if (ret == -1) {
         handler->allocator.free(handler->allocator.ctx, result, size);
--- a/numpy/_core/src/multiarray/alloc.cpp
+++ b/numpy/_core/src/multiarray/alloc.cpp
@@ -387,7 +387,8 @@ default_malloc(void *NPY_UNUSED(ctx), size_t size)
         return NULL;
     }
     // untrack the allocation from tracemalloc
-    int ret = PyTraceMalloc_Untrack(PYMEM_DOMAIN_RAW, (npy_uintp)result);
+    // ignore return value, since we'd early return either way
+    PyTraceMalloc_Untrack(PYMEM_DOMAIN_RAW, (npy_uintp)result);
     return result;
 }
 
--- a/numpy/_core/src/multiarray/alloc.cpp
+++ b/numpy/_core/src/multiarray/alloc.cpp
@@ -529,6 +529,7 @@ PyDataMem_UserFREE(void *ptr, size_t size, PyObject *mem_handler)
                      "Could not get pointer to 'mem_handler' from PyCapsule");
         return;
     }
+    // ignore -2 return when tracemalloc is disabled
     PyTraceMalloc_Untrack(NPY_TRACE_DOMAIN, (npy_uintp)ptr);
     handler->allocator.free(handler->allocator.ctx, ptr, size);
 }
--- a/numpy/_core/src/multiarray/alloc.cpp
+++ b/numpy/_core/src/multiarray/alloc.cpp
@@ -388,10 +388,6 @@ default_malloc(void *NPY_UNUSED(ctx), size_t size)
     }
     // untrack the allocation from tracemalloc
     int ret = PyTraceMalloc_Untrack(PYMEM_DOMAIN_RAW, (npy_uintp)result);
-    if (ret == -2) {
-        // tracemalloc is disabled
-        return result;
-    }
     return result;
 }
 
--- a/numpy/_core/src/multiarray/alloc.cpp
+++ b/numpy/_core/src/multiarray/alloc.cpp
@@ -294,8 +294,15 @@ PyDataMem_NEW(size_t size)
 
     assert(size != 0);
     result = PyMem_RawMalloc(size);
-    PyTraceMalloc_Untrack(PYMEM_DOMAIN_RAW, (npy_uintp)result);
-    int ret = PyTraceMalloc_Track(NPY_TRACE_DOMAIN, (npy_uintp)result, size);
+    if (result == NULL) {
+        return NULL;
+    }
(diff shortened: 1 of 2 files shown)


ngoldbaum:
It'd also be nice to have a test for this. Clearly we don't since we never noticed the issue.

braingram:
> It'd also be nice to have a test for this. Clearly we don't since we never noticed the issue.

The minimal example from the issue might work. Any recommendation for where it should live?

ngoldbaum:
How about numpy/_core/tests/test_multiarray.py, right next to `test_malloc_fails`.

braingram:
> numpy/_core/tests/test_multiarray.py

Thanks! I added a test in https://github.com/numpy/numpy/pull/30902/commits/8cad8405bfb3c3a16a99b32ad71715fdd6486a4b

It's a bit different from the minimal example from the issue to:
- avoid turning off tracemalloc if for some reason (perhaps a pytest plugin etc) tracemalloc is enabled prior to the test
- it's structured closer to the example in the docs: https://numpy.org/doc/stable/reference/c-api/data_memory.html#example-of-memory-tracing-with-np-lib-tracemalloc-domain

I avoided expanding it to use the `np.lib.tracemalloc_domain` (to try and limit the scope of the changes) but that might be doable with a pytest parameterization. Let me know if you'd like me to expand it to test the custom domain in addition to the standard one.

numpy/_core/tests/test_multiarray.py
```diff
@@ -1040,6 +1041,19 @@ def test_malloc_fails(self):
         with assert_raises(np._core._exceptions._ArrayMemoryError):
             np.empty(np.iinfo(np.intp).max, dtype=np.uint8)
 
+    def test_tracemalloc(self):
```
ngoldbaum, line 1045: Can you mark this test as thread-unsafe? The tracemalloc Python API isn’t thread-safe.

braingram: Does the maker in https://github.com/numpy/numpy/commit/fbe8d4642a30521543d3d5fbc878e89b12417d3f look right?

kumaraditya303 (approved):
Thanks, LGTM

numpy/_core/src/multiarray/alloc.cpp
```diff
@@ -340,6 +342,7 @@ PyDataMem_RENEW(void *ptr, size_t size)
     assert(size != 0);
     PyTraceMalloc_Untrack(NPY_TRACE_DOMAIN, (npy_uintp)ptr);
```
seberg: Sorry, random notice that this looks subtly wrong. If `result == NULL` shouldn't do any untracking.

Anyway, will ignore it here, but in case you want to follow up.

seberg (approved):
Looks good to me too, thanks.  A bit too bad we track+untrack now, but I'll guess it isn't a big deal...

I'll let Nathan put it in in case he has thoughts (or just put it in myself tomorrow or so).

ngoldbaum:
Isn't it only correct to untrack allocations from `PYMEM_DOMAIN_RAW` if the allocating function really is one of the `PyMemRaw` functions? Maybe we have to add a check (using NPY_UNLIKELY so it gets optimized away in most cases) before untracking to make sure the memory handler really is the default memory handler?

Also as Sebastian pointed out, and it is somewhat of a pre-existing issue, but both `PyTraceMalloc_Track` and `PyTraceMalloc_Untrack` can fail so we should be adding error checking for all calls into the tracemalloc API. I'd prefer not to add new calls without error checking, but it would be nice to update all of the other calls as well.

@braingram if you're not interested in addressing these issues, just say so and one of will take over.

ngoldbaum:
One way to save some runtime overhead might be to store pointers to functions that call the tracemalloc tracking and untracking functions, and then call that function instead of the `PyTraceMalloc` API directly. That said, the memory handler and allocator structs are public API. Maybe you could store it in the `ctx` though?

braingram:
Thanks for the reviews.

I'm happy to have someone take this over because it sounds like there are other issues that folks would like to address as part of this change. I pushed up one more commit in an attempt to address the comments but I suspect some of this requires further discussion.

Going off the docs for [PyTraceMalloc_Untrack](https://docs.python.org/3/c-api/memory.html#c.PyTraceMalloc_Untrack) the return code is either:
- `-2` if tracemalloc isn't enabled
- `0` otherwise, for succesful untracking or for an already untracked block (where nothing is done)

So maybe several of these functions could return early, successfully, for `-2` since there's no point in attempting to track if tracemalloc isn't enabled.

As pointed out by @seberg in https://github.com/numpy/numpy/pull/30902#discussion_r2870999859 renew is a bit different since it doesn't appear to correctly handle a failed reallocation (on main). If `result = PyMem_RawRealloc(ptr, size);` returns `result == NULL`, `ptr` is still valid (based on the [docs](https://docs.python.org/3/c-api/memory.html#c.PyMem_RawRealloc)). However at that point in the function `ptr` has already been untracked from `NPY_TRACE_DOMAIN`. Possibly the untracking could be moved after the realloc? I think this would mean for a successful reallocation that we're untracking a pointer to memory freed by python. (while looking for examples of `PyDataMem_RENEW` I found one in scipy](https://github.com/scipy/scipy/blob/0a0a989d71f5fad44a62a1c378be80bb0ea02619/scipy/ndimage/src/_ni_label.pyx#L361-L363) that doesn't appear to handle a possible `NULL` result and the numpy docs don't mention what `NULL` means here).

As @ngoldbaum pointed out the `PyDataMem_User*` functions, when the allocators are changed from the defaults, might not use `PYMEM_DOMAIN_RAW`. In this case `PyTraceMalloc_Untrack` should do nothing but also calling it shouldn't be needed. Reading: 
https://github.com/numpy/numpy/blob/20f444313823bd3160d1305c64ab73f4aee530d8/numpy/_core/src/multiarray/alloc.cpp#L351-L353
the description isn't quite right since after https://github.com/numpy/numpy/commit/316f598b06fde24b69318e70f894ec88d6b03ce9 that function is interacting with tracemalloc. Perhaps it makes sense to untrack the pointers within `default_malloc/etc`?

I applied the above changes in 84b8bb220d08cba47967476c1d95b43afb656aae
- handling `-2` return codes from `PyTraceMalloc_Untrack`
- moving the `PyTraceMalloc_Untrack` for `NPY_TRACE_DOMAIN` in reallocation functions to after the reallocation and only when it was successful
- moving the `PyTraceMalloc_Untrack` for `PYMEM_DOMAIN_RAW` inside `default_malloc/calloc/realloc` and removing the untracks from `PyDataMem_User*`

ngoldbaum:
I think this is now completely correct, but I have a couple touch-ups I'm going to push directly to this branch.

Thanks @braingram for your awesome work. If you ever want to mess around more in NumPy internals please feel free to keep on sending in more high quality PRs even when we don't accidentally break your numpy-dev tests!

@seberg if you agree, I'd appreciate it if you could hit the merge button since this PR changed a little bit since you last looked at it.

ngoldbaum (approved):

numpy/_core/src/multiarray/alloc.cpp
```diff
@@ -354,7 +380,19 @@ PyDataMem_RENEW(void *ptr, size_t size)
...
+        // alloc failed, nothing more to do
+        return NULL;
+    }
+    // untrack the allocation from tracemalloc
+    int ret = PyTraceMalloc_Untrack(PYMEM_DOMAIN_RAW, (npy_uintp)result);
+    if (ret == -2) {
+        // tracemalloc is disabled
+        return result;
+    }
+    return result;
```
seberg: A lot of logic for not much (in the end, I guess untrack on 0 is even OK), since in this case we have no further logic in the same function, unfortunately.

seberg, line 392: Seems like a bit of an unnecessary dance, but maybe it's nice to just keep the pattern.

seberg (approved):
Thanks Nathan, looks good.

The one thing I do wonder is whether we should avoid the `Untrack()` call when the cache is used?

If we want to do that, I think one approach that might also end up nicer is to add `static inline PyMem_RawMallocNoTrack/PyMem_RawCallocNoTrack/PyMemRawReallocNoTrack` functions that include the untracking.

But I am not exactly sure, and this looks fine.

numpy/_core/src/multiarray/alloc.cpp
```diff
@@ -354,7 +380,19 @@ PyDataMem_RENEW(void *ptr, size_t size)
...
+    if (result == NULL) {
+        // alloc failed, nothing more to do
+        return NULL;
+    }
+    // untrack the allocation from tracemalloc
+    int ret = PyTraceMalloc_Untrack(PYMEM_DOMAIN_RAW, (npy_uintp)result);
+    if (ret == -2) {
+        // tracemalloc is disabled
+        return result;
+    }
```
ngoldbaum:
```suggestion
```

numpy/_core/src/multiarray/alloc.cpp
```diff
@@ -354,7 +380,15 @@ PyDataMem_RENEW(void *ptr, size_t size)
...
 {
-    return _npy_alloc_cache(size, 1, NBUCKETS, datacache, &PyMem_RawMalloc);
+    void *result;
+    result = _npy_alloc_cache(size, 1, NBUCKETS, datacache, &PyMem_RawMalloc);
+    if (result == NULL) {
+        // alloc failed, nothing more to do
+        return NULL;
+    }
+    // untrack the allocation from tracemalloc
+    int ret = PyTraceMalloc_Untrack(PYMEM_DOMAIN_RAW, (npy_uintp)result);
```
ngoldbaum:
```suggestion
    // ignore return value, since we'd early return either way
    PyTraceMalloc_Untrack(PYMEM_DOMAIN_RAW, (npy_uintp)result);
```

numpy/_core/src/multiarray/alloc.cpp
ngoldbaum: Good point, I deleted the check and added a comment explaining why we're not checking the return value for future readers.

ngoldbaum:
Thanks again @braingram!

Issue #22635, opened 2022-11-21
DOC: Remove info specific to Python 2 from the latest docs version


InessaPawson:
Remove mentions of `python2` (without space). 
Please note that the files with release notes should not be modified.
For context, refer to https://github.com/numpy/numpy/issues/22400.

SANTHOSH-MAMIDISETTI:
@InessaPawson  is it closed ?,  I mean is it resolved ?

mattip:
` git grep "ython *2" | grep -v doc/changelog | grep -v doc/source/release |grep -v distutils | grep -v "dropping-python2.7"` still shows some results that may need some cleanup. Although some of those are valid mentions and should not be removed. I excluded distutils since it is deprecated.

JostMigenda:
I started looking at this during yesterday's EuroSciPy sprint day. Currently, on the main branch, I find 13 results for the suggested query:

```bash
% git grep "python *2" | grep -v doc/changelog | grep -v doc/source/release |grep -v distutils | grep -v "dropping-python2.7"   
doc/neps/nep-0012-missing-data.rst:    /home/mwiebe/virtualenvs/dev/lib/python2.7/site-packages/numpy/core/fromnumeric.py:2374: RuntimeWarning: invalid value encountered in double_scalars
doc/source/reference/arrays.nditer.cython.rst:    $ gcc -shared -pthread -fPIC -fwrapv -O2 -Wall -I/usr/include/python2.7 -fno-strict-aliasing -o sum_squares.so sum_squares.c
numpy/_core/src/multiarray/dtypemeta.c:    // NPY_STRING is the legacy python2 name
numpy/_core/tests/test_regression.py:            # (original, python2_pickle, koi8r_validity)
numpy/lib/_format_impl.py:    Required as e.g. ZipExtFile in python 2.6 can return less data than
numpy/lib/_format_impl.py:        # would-block, python2 file will truncate, probably nothing can be
numpy/lib/tests/test_format.py:def test_python2_python3_interoperability():
numpy/lib/tests/test_format.py:    fname = 'win64python2.npy'
numpy/lib/tests/test_format.py:def test_pickle_python2_python3():
numpy/linalg/lapack_lite/README.rst:appropriate scrubbing. As of writing, **this is only available for python 2.7**,
numpy/linalg/lapack_lite/clapack_scrub.py:#!/usr/bin/env python2.7
numpy/linalg/lapack_lite/make_lite.py:#!/usr/bin/env python2.7
```

[...]

PR #30904, opened 2026-02-28
Remove reference to Python 2 from count_nonzero() docstring

ChristinaDec:
Related to https://github.com/numpy/numpy/issues/22635 - removal of references to Python 2.  I edited the count_nonzero() docstring to remove reference to Python 2.   The function name itself is still based on Python 2's legacy method __nonzero__ , but that's more of an implicit reference.


mattip:
Thanks @ChristinaDec

PR #30908, opened 2026-03-01
BUG: Fix reference leaks and NULL pointer dereferences

eendebakpt:
The PR addresses some reference leaks and NULL pointer dereferences in uncommon paths (mostly failures due to out-of-memory). 

The issues addressed have been found using Claude.


jorenham:
there's an accidental submodule update

seberg (approved):
Thanks, changes look all fine. Hard to cover mostly since most indeed can't fail in practice.
(Except maybe that one dtype refcount leak, but unless we do something like `pytest-leaks` it's just whack-a-mole.)

PR #30910, opened 2026-03-02
BUG: fft: Fix handling of out parameter for hfft, ifft2 and irfft2.

WarrenWeckesser:
Closes gh-30909.


mattip:
Thanks @WarrenWeckesser

PR #30913, opened 2026-03-02
MAINT: fix two minor issues noticed when touching the C API setup

ngoldbaum:
@kumaraditya303 and I are playing with the opaque PyObject API, which required updating the C API. I noticed the error message in `verify_c_api_version.py` is out of date. We also happened to trigger an unusual error in the user dtype setup and I noticed the error message is formatted incorrectly.


charris:
Thanks Nathan.

PR #30918, opened 2026-03-03
BUG: fix type issues in uses if PyDataType macros

ngoldbaum:
@kumaraditya303 and I ran into these working on support for the opaque PyObject build.

These `PyDataType` macros expect a `PyArray_Descr *` argument, but in all three cases happen to work because the macros use direct struct field accesses and all these structs share field names.


mattip:
Thanks @ngoldbaum

PR #30920, opened 2026-03-04
DOC: Remove Python 2 array-protocol strings type note

ChristinaDec:
Related to https://github.com/numpy/numpy/issues/22635 - removal of references to Python 2.  I edited the "Note on string types" in the doc on array-protocol type strings to remove Python 2.
Within the edited content, I also found and removed a reference to type "a" for zero-terminated bytes.  This type was removed from the list of supported types last month as part of merged PR: https://github.com/numpy/numpy/pull/30758#ref-issue-3882906912.

### DIFF
--- a/doc/source/reference/arrays.dtypes.rst
+++ b/doc/source/reference/arrays.dtypes.rst
@@ -298,10 +298,10 @@ Array-protocol type strings (see :ref:`arrays.interface`)
 
    .. admonition:: Note on string types
 
-    For backward compatibility with existing code originally written to support
-    Python 2, ``S`` and ``a`` typestrings are zero-terminated bytes.
-    For unicode strings, use ``U``, `numpy.str_`.  For signed bytes that do not
-    need zero-termination ``b`` or ``i1`` can be used.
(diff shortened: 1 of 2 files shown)


charris:
This should probably mention the new StringDType ("T"). @ngoldbaum Ping.

seberg:
`T` cannot be supported here (although I am seriously wondering if we shouldn't add a `numpy_dtype` field and just use that if present).

But I am also wondering if this note even adds anything, should we just delete it completely?

ChristinaDec:
> But I am also wondering if this note even adds anything, should we just delete it completely?

It contains much the same information as the "supported kinds" table before it and nothing really noteworthy.  It is one more thing to update, along with the table, and that seems to not be happening reliably anyway.  Deletion seems to make sense.

seberg:
If you agree, want to just apply that?

charris:
The circleci failure is at
`/home/circleci/repo/doc/source/reference/arrays.scalars.rst:330: WARNING: undefined label: 'string-dtype-note'`
Removing the reference should fix it.

seberg (approved):
Thanks @ChristinaDec!

PR #30927, opened 2026-03-04
BUG: Fix matrix_rank crash on empty matrices

### LINKED ISSUE #30915: BUG: numpy.linalg.matrix_rank crashes with ValueError on empty matrices [DeepTest]
### Describe the issue:

`numpy.linalg.matrix_rank` raises a `ValueError` when given a matrix with a zero-sized dimension (e.g., shape `(0,
  3)`). The rank of an empty matrix is mathematically 0, and other linalg functions (`pinv`, `cond`) already handle
  this via the `_is_empty_2d` guard — `matrix_rank` is missing it.

```python
import numpy as np

   a = np.empty((0, 3))
   np.linalg.matrix_rank(a)  # Expected: 0
```

### Error message:

```shell
  File "C:\Users\madanm\git\DeepTest\repo_benchmarks\numpy-DeepTests\standalone\crash_v2\score5\linalg_matrix_rank_empty_matrix.py", line 24, in <module>
    rank = np.linalg.matrix_rank(a)
  File "C:\Users\madanm\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.13_qbz5n2kfra8p0\LocalCache\local-packages\Python313\site-packages\numpy\linalg\_linalg.py", line 2165, in matrix_rank
    tol = S.max(axis=-1, keepdims=True) * rtol
          ~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\madanm\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.13_qbz5n2kfra8p0\LocalCache\local-packages\Python313\site-packages\numpy\_core\_methods.py", line 43, in _amax
    return umr_maximum(a, axis, None, out, keepdims, initial, where)
ValueError: zero-size array to reduction operation maximum which has no identity
```

2.3.5

### Runtime Environment:

_No response_

Using DeepTest

thakoreh:
## Summary
Fixes #30915 

`numpy.linalg.matrix_rank` now correctly handles empty matrices (e.g., shape `(0, 3)`) by returning 0 instead of raising `ValueError`.

## Problem
Previously, calling `matrix_rank` on an empty matrix would crash with:
```
ValueError: zero-size array to reduction operation maximum which has no identity
```

This occurred because `S.max()` fails on empty arrays.

## Solution
Added the same `_is_empty_2d` guard that other linalg functions (`pinv`, `cond`) already use. The rank of an empty matrix is mathematically 0, so we return `zeros(A.shape[:-2], dtype=intp)`.

## Testing
Added comprehensive tests for:
- Empty matrices: (0, 3), (3, 0), (0, 0)
- Stacked empty matrices: (2, 0, 3)  
- Empty matrix with custom tolerance

All existing tests continue to pass.

## Backward Compatibility
✓ No breaking changes - only fixes a crash case
✓ All existing functionality preserved

### DIFF
--- a/numpy/linalg/_linalg.py
+++ b/numpy/linalg/_linalg.py
@@ -2130,6 +2130,10 @@ def matrix_rank(A, tol=None, hermitian=False, *, rtol=None):
     if A.ndim < 2:
         return int(not all(A == 0))
 
+    # Handle empty matrices - rank is 0
+    if _is_empty_2d(A):
+        return zeros(A.shape[:-2], dtype=intp)
+
     S = svd(A, compute_uv=False, hermitian=hermitian)
 
     if tol is None:
--- a/numpy/linalg/tests/test_linalg.py
+++ b/numpy/linalg/tests/test_linalg.py
@@ -1688,6 +1688,16 @@ class TestMatrixRank:
         assert_equal(4, matrix_rank(I, hermitian=True, tol=0.99e-8))
         assert_equal(3, matrix_rank(I, hermitian=True, tol=1.01e-8))
 
+    def test_empty_matrix(self):
+        # Empty matrices have rank 0
+        assert_equal(matrix_rank(np.empty((0, 3))), 0)
+        assert_equal(matrix_rank(np.empty((3, 0))), 0)
+        assert_equal(matrix_rank(np.empty((0, 0))), 0)
+        # Stacked empty matrices
+        assert_equal(matrix_rank(np.empty((2, 0, 3))), np.array([0, 0]))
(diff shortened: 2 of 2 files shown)


seberg:
Please compare with the PR in https://github.com/numpy/numpy/pull/30422 which solved most of this, but missed a branch.

And on a general note that we should just ask always: We would prefer to know about any (AI) tool use.  We haven't spelled things out, but a "sprintable" label is also to learn and e.g. fully automated PRs to solve them could also be triggered by maintainers.

PR #30928, opened 2026-03-04
ENH: Replace PyTyple_Pack with PyTuple_FromArray

eendebakpt:
Replace `PyTyple_Pack` (which uses variadic arguments) with the more performant `PyTuple_FromArray`.

Benchmark using
```
import time
import sys
# about 100 ms for python start and numpy import
import numpy as np
x = np.array([1., 2.])
y = np.array([0., 0.])


outer = 10
inner = 3
iterations = 200_000

for _ in range(outer):
    t00=0
    for _ in range(inner):
        t0 = time.perf_counter()
        for _ in range(iterations):
            np.add.reduce(x)
        dt = time.perf_counter() - t0
        t00 += dt
   
    print(f'{1e9*t00/(inner*iterations):.2f} [ns] / operation')
print(sys.executable)
print(np)
```
shows an improvement of 764 to 731 ns / operation.

PR created with help from Claude

### DIFF
--- a/numpy/_core/src/umath/ufunc_object.c
+++ b/numpy/_core/src/umath/ufunc_object.c
@@ -1384,7 +1384,7 @@ _parse_axes_arg(PyUFuncObject *ufunc, int op_core_num_dims[], PyObject *axes,
             Py_INCREF(op_axes_tuple);
         }
         else if (op_ncore == 1) {
-            op_axes_tuple = PyTuple_Pack(1, op_axes_tuple);
+            op_axes_tuple = PyTuple_FromArray(&op_axes_tuple, 1);
             if (op_axes_tuple == NULL) {
                 return -1;
             }
@@ -3412,7 +3412,7 @@ _set_full_args_out(int nout, PyObject *out_obj, ufunc_full_args *full_args)
             return 0;
         }
         /* Can be an array if it only has one output */
-        full_args->out = PyTuple_Pack(1, out_obj);
+        full_args->out = PyTuple_FromArray(&out_obj, 1);
         if (full_args->out == NULL) {
             return -1;
         }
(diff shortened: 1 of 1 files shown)


seberg:
Sounds a bit like you might refactor this further to remove this again?  I am surprised it actually is noticable seems like `PyTuple_Pack` is suprisingly slow (due to varargs?!).

Anyway, I don't mind using these patterns if it has a modest gain and the change is trivial, thanks @eendebakpt.

SwayamInSync:
This seems to be a fairly new symbol present in upstream main, and not present in 3.13, 3.14 versions of cpython.
Should we atleast keep a fallback for older version support? @seberg

seberg:
@SwayamInSync we test against older versions, it's part of capi-compat, so it's fine: https://github.com/python/pythoncapi-compat/

SwayamInSync:
> @SwayamInSync we test against older versions, it's part of capi-compat, so it's fine: [python/pythoncapi-compat](https://github.com/python/pythoncapi-compat/?rgh-link-date=2026-03-06T12%3A00%3A33Z)

Interesting inside quaddtype we also use capi-compat and with latest cpython (main branch 3.15-dev) runtime registration is failing.
I will look into in detail

seberg:
You probably just have to update the the submodule, would be my guess.

EDIT: Or is building NumPy from source in CI failing?!  But then I guess it would still seem like a submodule issue, just weirder...

SwayamInSync:
>You probably just have to update the the submodule, would be my guess.

Yeah already set to `main` (using meson warp it clones the from the given checkpoint)

> Or is building NumPy from source in CI failing?

The build works, its at runtime during import (dtype registration) it fails with undefined symbol `PyTuple_FromArray`
```
Original error was: /home/miniconda3/envs/quaddtype/lib/python3.14/site-packages/numpy/_core/_multiarray_umath.cpython-314-x86_64-linux-gnu.so: undefined symbol: PyTuple_FromArray
```

seberg:
@SwayamInSync weird, but this file includes the header, so I still somewhat think the submodule may not be up to date.
I.e. that is the expected error in that case, because otherwise the header replaces it with a `static inline` function (so the symbol is never searched).

EDIT: But if the header is outdated, it would be missing and C would just give a warning and make it an implicitly function here leading to an error at loading time.

SwayamInSync:
Oh yes it was outdated local submodule 🥲 Sorry for the trouble, my bad!!

PR #30929, opened 2026-03-04
MAINT: Refactor PyUFunc_GenericReduction

eendebakpt:
Extract the code to prepare the `axis` into a helper method. This is in preparation of PRs to improve performance numpy ufuncs.


seberg:
Could even move it futher to conversion utils, but probably not worthwhile here.  Seems indeed a bit nicer and if it helps you with further work, let's put this in, thanks.

seberg (approved):

PR #30930, opened 2026-03-04
ENH: Improve performance of npy_parse_arguments

eendebakpt:
We eliminate the variadic argument parsing in `npy_parse_arguments` to improve performance. In the PR the variadic arguments are handled at compile time.

Benchmark using
```
import time
import sys
import numpy as np
x = np.array([1., 2.])
y = np.array([0., 0.])

outer = 10
inner = 3
iterations = 200_000

for _ in range(outer):
    t00=0
    for _ in range(inner):
        t0 = time.perf_counter()
        for _ in range(iterations):
            np.add.reduce(x, axis=0)
        dt = time.perf_counter() - t0
        t00 += dt
   
    print(f'{1e9*t00/(inner*iterations):.2f} [ns] / operation')
print(sys.executable)
print(np)
```
shows an improvement of 787 to 762 ns / operation.

PR created with help from Claude

### DIFF
--- a/numpy/_core/src/multiarray/methods.c
+++ b/numpy/_core/src/multiarray/methods.c
@@ -126,11 +126,10 @@ array_take(PyArrayObject *self,
     NPY_PREPARE_ARGPARSER;
 
     if (npy_parse_arguments("take", args, len_args, kwnames,
-            "indices", NULL, &indices,
-            "|axis", &PyArray_AxisConverter, &dimension,
-            "|out", &PyArray_OutputConverter, &out,
-            "|mode", &PyArray_ClipmodeConverter, &mode,
-            NULL, NULL, NULL) < 0) {
+            {"indices", NULL, &indices},
+            {"|axis", &PyArray_AxisConverter, &dimension},
+            {"|out", &PyArray_OutputConverter, &out},
+            {"|mode", &PyArray_ClipmodeConverter, &mode}) < 0) {
         return NULL;
     }
 
@@ -225,8 +224,7 @@ array_squeeze(PyArrayObject *self,
     NPY_PREPARE_ARGPARSER;
 
     if (npy_parse_arguments("squeeze", args, len_args, kwnames,
-            "|axis", NULL, &axis_in,
-            NULL, NULL, NULL) < 0) {
+            {"|axis", NULL, &axis_in}) < 0) {
         return NULL;
     }
 
@@ -253,9 +251,8 @@ array_view(PyArrayObject *self,
     NPY_PREPARE_ARGPARSER;
 
     if (npy_parse_arguments("view", args, len_args, kwnames,
-            "|dtype", NULL, &out_dtype,
-            "|type", NULL, &out_type,
-            NULL, NULL, NULL) < 0) {
+            {"|dtype", NULL, &out_dtype},
+            {"|type", NULL, &out_type}) < 0) {
         return NULL;
     }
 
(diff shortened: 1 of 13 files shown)


seberg:
A lot of churn, unfortunately. But I like the fact that it removes the need for `NULL` termination and don't mind the new `{}`.

I didn't realize that varargs were slowish, TBH.  I am fine with this, but would be nice for @mattip or @ngoldbaum to give an OK.

numpy/_core/src/multiarray/methods.c
```diff
@@ -1433,12 +1425,12 @@ array_argsort(PyArrayObject *self,
...
-            "|kind", &PyArray_SortkindConverter, &sortkind,
-            "|order", NULL, &order,
-            "$stable", &PyArray_OptionalBoolConverter, &stable,
-//            "$descending", &PyArray_OptionalBoolConverter, &descending,
-            NULL, NULL, NULL) < 0) {
+            {"|axis", &PyArray_AxisConverter, &axis},
+            {"|kind", &PyArray_SortkindConverter, &sortkind},
+            {"|order", NULL, &order},
+            {"$stable", &PyArray_OptionalBoolConverter, &stable}
+            // {"$descending", &PyArray_OptionalBoolConverter, &descending}
```
ngoldbaum, line 1433:
```suggestion
            // TODO: add descending sorts, gh-14728
            // {"$descending", &PyArray_OptionalBoolConverter, &descending}
```

numpy/_core/src/multiarray/unique.cpp
```diff
@@ -493,11 +493,13 @@ array__unique_hash(PyObject *NPY_UNUSED(module),
...
     NPY_PREPARE_ARGPARSER;
-    if (npy_parse_arguments("_unique_hash", args, len_args, kwnames,
-                            "arr", &PyArray_Converter, &arr,
-                            "|equal_nan",  &PyArray_BoolConverter, &equal_nan,
-                            NULL, NULL, NULL
-                            ) < 0
+    npy_arg_spec parse_specs[] = {
+        {"arr", (void *)&PyArray_Converter, &arr},
+        {"|equal_nan", (void *)&PyArray_BoolConverter, &equal_nan},
+    };
```
ngoldbaum: I guess C++ requires stack-allocating the specs like this?

Maybe add a note to the docstring in `npy_argparse.c` that you have to use a slightly different pattern for C++ code?

ngoldbaum (approved):
Definitely a nice cleanup. I left two very minor comments.

numpy/_core/src/multiarray/unique.cpp
ngoldbaum: Maybe the compiler does it automatically, but you could also make this array of structs `static` I think.

seberg: I suppose it is `const` but not `static` since the last parsing result is on the stack.
(Would have to reshuffle that if we want to optimize things, since the rest is indeed static and only read once at first use.)

FWIW, I hadn't noticed this C++ discrepancy, which I don't like (if just because it is different...).
Brainstorming a bit with the rubber duck, how about we make the macro something like:
```
#if defined(__cplusplus)
    ([&]() {
        const npy_arg_spec parse_specs[]{__VA_ARGS__};
        return ...;
    }())
#else
   ...
#endif
```

Which should work with C++17 to create the scope and we require C++17 (I guess it's likely C++11, but not sure either way).

eendebakpt: I agree we should be able to have the C and C++ code identical, so I made some changes based on the suggestion. I will try to run a benchmark for the c++ code later.

seberg (approved):
Thanks @eendebakpt, I'll be an optimist and trust the C++ compilers to be able to inline this well enough.
But if you find something else, let's follow up.

PR #30931, opened 2026-03-04
BUG: Fix `np.dot` to allow user dtypes

MaanasArora:
This is a start to fix `PyArray_MatrixProduct2` for user-defined dtypes, as a bug was reported in https://github.com/jax-ml/ml_dtypes/pull/360. This is a draft as I've not tested this outside of a dummy dtype. I will try to use the linked PR with this, but wanted to get this out if anyone has comments. ping @seberg - thanks!

There also seem to be other methods with the same issue, so we might need a helper function long run. Hopefully this stays minimal on more testing, though not sure if even this is small enough to backport.

### DIFF
--- a/numpy/_core/src/multiarray/common.c
+++ b/numpy/_core/src/multiarray/common.c
@@ -381,22 +381,16 @@ _may_have_objects(PyArray_Descr *dtype)
  */
 NPY_NO_EXPORT PyArrayObject *
 new_array_for_sum(PyArrayObject *ap1, PyArrayObject *ap2, PyArrayObject* out,
-                  int nd, npy_intp dimensions[], int typenum, PyArray_Descr *descr,
-                  PyArrayObject **result)
+                  int nd, npy_intp dimensions[], PyArray_Descr *descr, PyArrayObject **result)
 {
     PyArrayObject *out_buf;
 
     if (out) {
         int d;
-        int typenum_check = typenum;
-        if (typenum == NPY_NOTYPE) {
-            // user-defined dtype
-            typenum_check = descr->type_num;
-        }
 
         /* verify that out is usable */
         if (PyArray_NDIM(out) != nd ||
-            PyArray_TYPE(out) != typenum_check ||
+            PyArray_TYPE(out) != descr->type_num ||
             !PyArray_ISCARRAY(out)) {
             PyErr_SetString(PyExc_ValueError,
                 "output array is not acceptable (must have the right datatype, "
@@ -458,19 +452,11 @@ new_array_for_sum(PyArrayObject *ap1, PyArrayObject *ap2, PyArrayObject* out,
             subtype = Py_TYPE(ap1);
         }
 
-        // to accomodate user-defined dtypes, use descriptor when typenum is NPY_NOTYPE
-        if (typenum == NPY_NOTYPE) {
-            out_buf = (PyArrayObject *)PyArray_NewFromDescr(subtype, descr, nd, dimensions,
-                                                            NULL, NULL, 0,
-                                                            (PyObject *)
-                                                            (prior2 > prior1 ? ap2 : ap1));
-        }
-        else {
-            out_buf = (PyArrayObject *)PyArray_New(subtype, nd, dimensions,
-                                                   typenum, NULL, NULL, 0, 0,
-                                                   (PyObject *)
-                                                   (prior2 > prior1 ? ap2 : ap1));
-        }
+        Py_INCREF(descr);
+        out_buf = (PyArrayObject *)PyArray_NewFromDescr(subtype, descr, nd, dimensions,
+                                                        NULL, NULL, 0,
+                                                        (PyObject *)
+                                                        (prior2 > prior1 ? ap2 : ap1));
 
         if (out_buf != NULL && result) {
             Py_INCREF(out_buf);
--- a/numpy/_core/src/multiarray/common.c
+++ b/numpy/_core/src/multiarray/common.c
@@ -460,7 +460,6 @@ new_array_for_sum(PyArrayObject *ap1, PyArrayObject *ap2, PyArrayObject* out,
 
         // to accomodate user-defined dtypes, use descriptor when typenum is NPY_NOTYPE
         if (typenum == NPY_NOTYPE) {
-            Py_INCREF(descr);
             out_buf = (PyArrayObject *)PyArray_NewFromDescr(subtype, descr, nd, dimensions,
                                                             NULL, NULL, 0,
                                                             (PyObject *)
--- a/numpy/_core/src/multiarray/common.c
+++ b/numpy/_core/src/multiarray/common.c
@@ -381,16 +381,22 @@ _may_have_objects(PyArray_Descr *dtype)
  */
 NPY_NO_EXPORT PyArrayObject *
 new_array_for_sum(PyArrayObject *ap1, PyArrayObject *ap2, PyArrayObject* out,
-                  int nd, npy_intp dimensions[], int typenum, PyArrayObject **result)
+                  int nd, npy_intp dimensions[], int typenum, PyArray_Descr *descr,
+                  PyArrayObject **result)
 {
     PyArrayObject *out_buf;
 
     if (out) {
         int d;
+        int typenum_check = typenum;
+        if (typenum == NPY_NOTYPE) {
+            // user-defined dtype
+            typenum_check = descr->type_num;
+        }
 
         /* verify that out is usable */
         if (PyArray_NDIM(out) != nd ||
-            PyArray_TYPE(out) != typenum ||
+            PyArray_TYPE(out) != typenum_check ||
             !PyArray_ISCARRAY(out)) {
             PyErr_SetString(PyExc_ValueError,
                 "output array is not acceptable (must have the right datatype, "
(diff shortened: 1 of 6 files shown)


numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -983,18 +983,34 @@ PyArray_MatrixProduct2(PyObject *op1, PyObject *op2, PyArrayObject* out)
     npy_intp dimensions[NPY_MAXDIMS];
     PyArray_DotFunc *dot;
     PyArray_Descr *typec = NULL;
+    int user_type = 0;
     NPY_BEGIN_THREADS_DEF;
 
-    typenum = PyArray_ObjectType(op1, NPY_NOTYPE);
-    if (typenum == NPY_NOTYPE) {
+    if (PyArray_DTypeFromObject(op1, NPY_MAXDIMS, &typec) < 0) {
```
seberg, line 986: Thanks! Wow, I must have missed the fact that I added that function at some point!?
I think the last `&typec` can actually just be used in a second call to continue promotion, giving the same behavior as the old one?

In that case the replacement here is actually straight-forward (I somehow thought it may not be).

(We don't have to do anything more then, since `dot == NULL` is already checked for below.)

MaanasArora: > I think the last &typec can actually just be used in a second call to continue promotion, giving the same behavior as the old one?

Thanks, yeah basically! I think that keeps this minimal - I'll do that.

Though, I found a pretty bad silent bug with this, which is that `new_array_from_sum` actually uses the `typenum` to construct the array, so it doesn't take a descr right now. Changing that will make this a bit bigger unfortunately, but still pretty straightforward luckily. I'll push soon!

ngoldbaum:
ping @SwayamInSync this seems like something you might be interested in

numpy/_core/src/multiarray/multiarraymodule.c
MaanasArora: Pushed the bugfix! I checked and the dot test on the `ml_dtypes` PR passes with this change.

I now tried using `PyArray_DTypeFromObject` twice for promotion, but not sure if there is a bug somewhere because the output `descr` causes a segmentation fault on this test. Looking into where that might be.

Of course, just using the first dtype is probably not acceptable (!?) so we might have to rely on `common_dtype`?

seberg: Yeah, we could just create two arrays and then promote/use CommonDType, that would be nicer. That might be slightly less quick for some weirder inputs, but probably we shouldn't care.
I have to think whether there are weird corner cases (I can't think of any right now, but somehow I think there was _something_ slightly trickier).

MaanasArora: Nevermind, it seems our code might be fine, `CommonDType` causes the same crash. It seems the issue might be with `ml_dtypes`, given on gdb, the second call for two complexes returns the float builtin dtype for some reason. I pushed just with the second call, but will investigate later if there's more to worry about. I'll mark this as ready for review, it's close at least!

Edit: yeah not sure what's going on with this, actually! Only on adding the second call (for promotion), it seems the whole operation breaks, at multiple steps. Will look more later.

MaanasArora: Sorry, that was wrong, there was simply a refcounting issue. Works now - should be ready to review!

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -985,16 +985,37 @@ PyArray_MatrixProduct2(PyObject *op1, PyObject *op2, PyArrayObject* out)
...
+    if (PyArray_DTypeFromObject(op2, NPY_MAXDIMS, &typec) < 0) {
+        Py_XDECREF(typec);
         return NULL;
     }
 
-    typec = PyArray_DescrFromType(typenum);
+    if (typec != NULL) {
+        if (NPY_DT_is_user_defined(typec)) {
+            // we steal a reference in new_array_for_sum
+            Py_INCREF(typec);
```
seberg: Can we change this and not steal the reference (it tends to be confusing).

Plus, I think there is no reason +not to adopt the same pattern to all of these places meaning that `typenum` should not be used anywhere (i.e. not in `new_array_for_sum`).

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -985,16 +985,37 @@ PyArray_MatrixProduct2(PyObject *op1, PyObject *op2, PyArrayObject* out)
...
+            typenum = NPY_NOTYPE;
+        }
+        else {
+            Py_DECREF(typec);
+            typec = NULL;
+        }
+    }
+    if (typec == NULL) {
```
seberg, line 994: You _do_ need this branch, but you don't need to keep this fallback, all it needs is `typec = PyArray_DescrFromType(NPY_DEFAULT_TYPE)` (to get the default float64 for empty list inputs for example).

seberg (commented):
Thanks, I am very happy to do this `PyArray_DTypeFromObject` but I think it makes sense here to apply it accross all functions that end up using `new_array_for_sum`.

numpy/_core/src/multiarray/multiarraymodule.c
MaanasArora: Both done, thank you!

numpy/_core/src/multiarray/multiarraymodule.c
MaanasArora: Sorry, a bit confused, won't that include non-user dtypes? Which is fine, but then this can be simplified more than branching like this? In that case, this is done!

numpy/_core/src/multiarray/common.c
```diff
@@ -390,7 +390,7 @@ new_array_for_sum(PyArrayObject *ap1, PyArrayObject *ap2, PyArrayObject* out,
 
         /* verify that out is usable */
         if (PyArray_NDIM(out) != nd ||
-            PyArray_TYPE(out) != typenum ||
+            PyArray_TYPE(out) != descr->type_num ||
```
seberg:
```suggestion
            !PyArray_EquivTypes(PyArray_DESCR(out), descr) ||
```

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -985,23 +989,25 @@ PyArray_MatrixProduct2(PyObject *op1, PyObject *op2, PyArrayObject* out)
...
+    if (PyArray_DTypeFromObject(op2, NPY_MAXDIMS, &typec) < 0) {
+        Py_XDECREF(typec);
         return NULL;
     }
 
-    typec = PyArray_DescrFromType(typenum);
+    if (typec == NULL) {
+        typec = PyArray_DescrFromType(NPY_DEFAULT_TYPE);
+    }
     if (typec == NULL) {
```
seberg: This branch is now dead.

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -985,23 +989,25 @@ PyArray_MatrixProduct2(PyObject *op1, PyObject *op2, PyArrayObject* out)
...
         if (!PyErr_Occurred()) {
             PyErr_SetString(PyExc_TypeError,
                             "Cannot find a common data type.");
         }
         return NULL;
     }
+    typenum = typec->type_num;
```
seberg, line 998: Hmmm, this makes me realize that using the `type_num` previously, meant that we automatically avoid non-native byte-order.
This is now lost, I think we have to insert an `NPY_DT_ensure_canonical()` call or so.

Would be nice to see if we have a test, it might need something as strange as:
```
np.array([1, 2, 3], ">f8").dot([[], [], []])
```

seberg (commented):
There is still a segfault, hopefully something small. But I also noticed 2 smaller things.
Thanks for looking at all of it now!

numpy/_core/src/multiarray/multiarraymodule.c
MaanasArora: Removed, thanks!

numpy/_core/src/multiarray/multiarraymodule.c
MaanasArora: Thanks for reviewing so quickly! Ah right - I'll test this out, but first I pushed your other comments and segfault fix (`vdot`) as I couldn't reproduce it locally. Was just a missing NULL initialization, I think.

MaanasArora: You were right, behavior did change for non-native byte order! There wasn't a test before, but I added one, and also pushed calls for `NPY_DT_ensure_canonical` in each function.

This is a bit repetitive, but it already was - would a helper function be nice? Maybe follow-up?

seberg: I think it's fine, thanks!  Because if we clean this up, we may want to completely restructure it (and live with the slight behavior changes)...

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -1397,45 +1402,51 @@ NPY_NO_EXPORT PyObject *
...
+    if (PyArray_DTypeFromObject(op2, NPY_MAXDIMS, &typec) < 0) {
+        Py_XDECREF(typec);
         return NULL;
     }
 
-    typec = PyArray_DescrFromType(typenum);
+    if (typec == NULL) {
+        typec = PyArray_DescrFromType(NPY_DEFAULT_TYPE);
+    }
+    typec = NPY_DT_CALL_ensure_canonical(typec);
```
seberg:
```suggestion
    Py_SETREF(typec, NPY_DT_CALL_ensure_canonical(typec));
```
I think, ref counting is hard :(.

Did you add it to all of these code paths yet?

EDIT: ooops, was only paying half attention. suggestion was silly typoed first.

MaanasArora: Yeah, ensure canonical is in all the code paths now I think!

It does change the refcounts, though :(

Did you mean to have a source for the SETREF?

MaanasArora: Neat suggestion, thanks! I pushed a commit with it for all code paths.

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -2610,16 +2620,18 @@ array_vdot(PyObject *NPY_UNUSED(dummy), PyObject *const *args, Py_ssize_t len_ar
...
+    if (PyArray_DTypeFromObject(op2, NPY_MAXDIMS, &type) < 0) {
+        Py_XDECREF(type);
         return NULL;
     }
 
-    type = PyArray_DescrFromType(typenum);
+    if (type == NULL) {
+        type = PyArray_DescrFromType(NPY_DEFAULT_TYPE);
+    }
+
```
seberg, line 2635:
```suggestion
   Py_SETREF(typec, NPY_DT_CALL_ensure_canonical(typec));

```

seberg (approved):

SwayamInSync (approved):
I'm sorry coming late, this looks good to me.
Thanks @MaanasArora for working on this

seberg (approved):
Thanks for the final fixes, let's get this in.  (It wouldn't hurt to add more byte-swap related tests, but I don't want to block on that -- all modified code paths include `NPY_DT_CALL_ensure_canonical`.)

PR #30685, opened 2026-01-20
MAINT: Run dependabot on the requirements files

charris:
If this works, we may want to run on the tools files as well.

PR #30935, opened 2026-03-05
CI: CI hardening

scottshambaugh:
Inspired by recent CI/CD attacks: https://www.stepsecurity.io/blog/hackerbot-claw-github-actions-exploitation#summary-of-results

This adds some hardening to the CI pipeline:
* Checksums for direct intel downloads and versioned containers (rolling containers not checksummed)
* Pinned versions for some currently unpinned downloads
* Remove the [`gerrymanoim/pr-prefix-labeler`](https://github.com/gerrymanoim/pr-prefix-labeler) action and replace it with inlined logic that does the same thing. While this action was checksum verified and looks safe to me right now, `GITHUB_TOKEN` should not be passed externally and this is pretty minimal functionality to replicate. The label config is kept in the current `.github/pr-prefix-labeler.yml`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,7 +12,6 @@ updates:
       - "03 - Maintenance"
     ignore:
       - dependency-name: "bus1/cabuild"
-updates:
   - package-ecosystem: pip
     directory: /requirements
     schedule:
--- a/.github/workflows/compiler_sanitizers.yml
+++ b/.github/workflows/compiler_sanitizers.yml
@@ -83,7 +83,7 @@ jobs:
     if: github.repository == 'numpy/numpy'
     runs-on: ubuntu-latest
     container:
-        image: ghcr.io/nascheme/numpy-tsan:3.14t
+        image: ghcr.io/nascheme/numpy-tsan:3.14t@sha256:1ec427e2e480cc373d0fecbf21b8ac590fb94119fb81d18489945cd0afd04dd3
         options: --shm-size=2g # increase memory for large matrix ops
 
     steps:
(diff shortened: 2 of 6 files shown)


mattip:
Thanks. I assume dependabot will not update the links and checksums. Could you add comments with links/directions how to update?

scottshambaugh:
Added instructions for those downloads.

It looks like the dependabot config is actually broken and not running for github actions right now - there are two `updates` blocks so the second is overwriting the first. I fixed that but didn't up-rev the actions, as it's probably a good verification to see if those get picked up once this is merged. For example, `actions/upload-artifact` has v7.0.0 released instead of the v6.0.0 that's pinned.

This bug was introduced in https://github.com/numpy/numpy/pull/30685 (@charris FYI)

scottshambaugh:
Looks like this triggered dependabot on a bunch of actions, and the PRs are being automatically labeled. Seems to be working!

charris:
@scottshambaugh dependabot ran on merging the updated yml file. You can check the generated PRs and see what more needs to be added (if needed) in a new PR.

scottshambaugh:
The dependabot PRs are updating the sha and versioning comments as expected, and it's caught all 5 actions that I saw as needing an upgrade. Everything looks good to me

mattip:
Hey @scottshambaugh, ~this~ Something seems to have broken the PR labeler. Something about yaml not being found? [Here is an example](https://github.com/numpy/numpy/actions/runs/29438550441/job/87431283578) run that seems to work but fails

scottshambaugh:
That looks like my fault, thanks for the catch @mattip! PR here: https://github.com/numpy/numpy/pull/32011

mattip:
Thanks for the PR

PR #30939, opened 2026-03-05
ENH: improve UFuncOutputCastingError message to explain resolved dtype

### LINKED ISSUE #14828: Incorrect error message when adding i8 array to u8 array
When attempting to add an `ndarray` of type `int64` to an `ndarray` of type `uint64`, `numpy` throws a `TypeError`:

    TypeError: Cannot cast ufunc add output from dtype('float64') to dtype('uint64') with casting rule 'same_kind'

The error message incorrectly identifies the type of the first array as `float64`, when it is actually `int64`.

### Reproducing code example:

```python
import numpy as np
x = np.arange(5, dtype='u8')
y = np.arange(5, dtype='i8')
x += y
```

### Error message:

    TypeError: Cannot cast ufunc add output from dtype('float64') to dtype('uint64') with casting rule 'same_kind'

### Numpy/Python version information:

Tested with Python 3:

    1.16.4 3.6.8 (default, Oct  7 2019, 12:59:55) 
    [GCC 8.3.0]

Also with Python 2:

    ('1.15.1', '2.7.15+ (default, Oct  7 2019, 17:39:04) \n[GCC 7.4.0]')

Bortlesboat:
## Summary
- Improves the `_UFuncOutputCastingError` message to explain that the unexpected dtype (e.g. `float64`) is the resolved result type of the ufunc loop, not an input type
- Updates the quickstart doc example to match

Before:
```
Cannot cast ufunc 'add' output from dtype('float64') to dtype('uint64') with casting rule 'same_kind'
```

After:
```
Cannot cast ufunc 'add' output from dtype('float64') to dtype('uint64') with casting rule 'same_kind'. This means the resolved dtype for the ufunc result is dtype('float64'), which cannot be safely stored in an output array of dtype dtype('uint64').
```

Closes #14828

### DIFF
--- a/doc/source/user/quickstart.rst
+++ b/doc/source/user/quickstart.rst
@@ -359,7 +359,7 @@ existing array rather than create a new one.
     >>> a += b  # b is not automatically converted to integer type
     Traceback (most recent call last):
         ...
-    numpy._core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'add' output from dtype('float64') to dtype('int64') with casting rule 'same_kind'
+    numpy._core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'add' output from dtype('float64') to dtype('int64') with casting rule 'same_kind'. This means the resolved dtype for the ufunc result is dtype('float64'), which cannot be safely stored in an output array of dtype dtype('int64').
 
 When operating with arrays of different types, the type of the resulting
 array corresponds to the more general or precise one (a behavior known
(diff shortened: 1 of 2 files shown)


ngoldbaum:
The prior attempt (https://github.com/numpy/numpy/pull/14843) ended up much more complicated than this. Also, in the end, @mattip [requested](https://github.com/numpy/numpy/pull/14843#issuecomment-2213972820) to initiate a discussion about updated language. I don't think anyone did that.

I think this is very wordy. This is also a very user-facing error message (I've definitely seen it!) and I don't think it's helpful to add visual noise by adding lots of text to error messages where the current message already contains context. This just repeats the context in a more wordy (albeit easier to understand) way.

I'm going to go ahead and close this. If you're interested in improving the error message, the best way to take that on is to restart the discussion in #14828 about this. Or by starting a discussion on the mailing list to solicit suggestions.

PR #30941, opened 2026-03-05
DOC: clarify from_dlpack GPU example with CPU fallback

mitre88:
## Summary
- updates the DLPack interoperability note to clarify why the GPU example fails by default
- keeps the failure case and adds a CPU fallback example using `device='cpu'`
- clarifies that NumPy arrays are CPU-only while cross-library DLPack exchange can still work for non-NumPy GPU arrays

## Context
Closes #30936


seberg:
Honestly, after thinking about it.  With 2 more PRs that look just automated, I'll just close this for now.  you can go through manually once and re-open or so.

PR #30942, opened 2026-03-05
DOC: align __array__ signature with keyword-only copy

### LINKED ISSUE #30907: DOC: signature of `__array__`

On https://numpy.org/devdocs/user/basics.interoperability.html#dunder-array-interface it is stated that the signature of `__array__` should be `__array__(self, dtype=None, copy=None)`, i.e. arguments are either positional or keyword-based.
However, `numpy.ndarray.__array__` has `__array__([dtype, ]*, copy=None)`, i.e. with keyword-only `copy`, see https://numpy.org/devdocs/reference/generated/numpy.ndarray.__array__.html#numpy.ndarray.__array__

This means that `np.array(a,int,True)` works for a user-defined `a` that follows the documentation but not for `numpy.ndarray`.

Note that this was mentioned in https://github.com/numpy/numpy/issues/30873, but I understand the `nonfix` as related to the typing issue.

There are two options:
1. Change the implementation of `numpy.ndarray.__array__`: Not a DOC change and enforcing a keyword argument was probably a conscious decision for reasons unknown to me.
2. Update the documentation and suggest a to have the same signature as `numpy.ndarray.__array__`.

All code where the `copy` argument is given via a keyword will continue to work in both cases.

mitre88:
## Summary
- update `doc/source/user/basics.interoperability.rst` to document `__array__` as
  `__array__(self, dtype=None, *, copy=None)`
- add a short clarifying sentence that `copy` is keyword-only, matching `ndarray.__array__`

This aligns the interoperability guide with the current `ndarray.__array__` signature and avoids confusion around positional `copy` arguments.

Closes #30907

### DIFF
--- a/doc/source/user/basics.interoperability.rst
+++ b/doc/source/user/basics.interoperability.rst
@@ -139,7 +139,9 @@ new ndarray object. This is not optimal, as coercing arrays into ndarrays may
 cause performance problems or create the need for copies and loss of metadata,
 as the original object and any attributes/behavior it may have had, is lost.
 
-The signature of the method should be ``__array__(self, dtype=None, copy=None)``.
+The signature of the method should be
+``__array__(self, dtype=None, *, copy=None)``.
+The ``copy`` argument is keyword-only, matching ``ndarray.__array__``.
 If a passed ``dtype`` isn't ``None`` and different than the object's data type,
 a casting should happen to a specified type. If ``copy`` is ``None``, a copy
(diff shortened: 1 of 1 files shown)


seberg:
Closing, IMO I don't think this is a big enough improvement for a PR that seems fully automated (the user has 99+ PRs in a single day).  And considering it is fully automated, I am not sure I want to iterate here.

Issue #30926, opened 2026-03-04
DOC: f2py: status about conda windows compilers is outdated


jeanwsr:
### Issue with current documentation:

The doc says that 
https://github.com/numpy/numpy/blob/9259ab7643426dd8efda5ca818d262146ee7e446/doc/source/f2py/windows/index.rst?plain=1#L60-L62

which is not a problem now in 2026. The windows compilers in conda should be up-to-date enough, see [conda-forge/conda-forge.github.io/issues/1044](https://github.com/conda-forge/conda-forge.github.io/issues/1044) or [conda-forge/packages/gfortran_win-64](https://anaconda.org/channels/conda-forge/packages/gfortran_win-64/overview).

mattip:
@jeanwsr want to make a PR to remove that outdated information?

jeanwsr:
> [@jeanwsr](https://github.com/jeanwsr) want to make a PR to remove that outdated information?

I'm willing to, but currently I can only confirm and update the status of gnu compilers in conda. I don't know much about MSVC/flang which are also part of [conda-forge compilers](https://conda-forge.org/docs/maintainer/infrastructure/#compilers-supplied-by-conda-forge). Maybe an expert of f2py/conda-forge can update the whole page.

PR #30952, opened 2026-03-06
DOC: f2py: update info about conda compilers

jeanwsr:
Minimal fix of #30926.

### DIFF
--- a/doc/source/f2py/windows/conda.rst
+++ b/doc/source/f2py/windows/conda.rst
@@ -25,10 +25,4 @@ Now we will setup a ``conda`` environment.
 
 ``conda`` pulls packages from ``msys2``, however, the UX is sufficiently different enough to warrant a separate discussion.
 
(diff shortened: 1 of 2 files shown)


HaoZeke (approved):

Nikita2106-05:
This PR provides a minimal fix for #30926 by clarifying the behavior of cpu-baseline="none" in the SIMD build options:

Only a single generic code path is generated when cpu-baseline="none".
Runtime CPU dispatch is not performed in this case.
This differs from the default cpu-baseline="min", which generates multiple CPU-specific code paths and selects the fastest one at runtime.
Users can still use cpu-dispatch to enable additional optimized code paths for supported CPUs.

rgommers (approved):
LGTM too. What remains is still a little odd (why would we mention `scoop`?), but that's fine, this is strictly an improvement. So in it goes.

PR #30953, opened 2026-03-06
DOC: remove a redundant build command

SpecLad:
Both `environment.yml` and `build_requirements.txt` already include spin, so there is no need to install it again.


mattip:
Thanks @SpecLad

PR #30954, opened 2026-03-07
MAINT: change ruff version to 0.15.4

antareepsarkar:
`ruff` errored in my conda environment.
`linter_requirements.txt` already uses `0.15.4`, so may be `environment.yml` should also use it.


charris:
does that fix the problem

antareepsarkar:
Yes. I updated the environment after the change and all checks passed.

charris:
Thanks @antareepsarkar .

PR #30956, opened 2026-03-07
ENH: Added steps to lib.stride_tricks.sliding_window_view

joe-saronic:
Adds a new `steps` argument to `sliding_window_view` to enable non-unit window strides. Doc example and unit tests show intended usage.

Hoping this is sufficiently non-controversial to not require a mailing list message.


joe-saronic:
To pre-empt the inevitable question, my commits are registered to my personal account, but am using my corporate account to actually push. That is why there appear to be two authors here. They are both me, and there is no conflict of interest or anything like that going on here.

seberg:
> Hoping this is sufficiently non-controversial to not require a mailing list message.

It's not a huge deal, but IIRC when adding it we somewhat thought there isn't much point since you can do `sliding_window_view(...)[::2]`, so it seemed easy enough to just not integrate it.

(I.e. I am not sure I am convinced yet, the mailing list might help get support. But maybe others like it. I don't have a problem with it, I just feel an example may be just as well.)

joe-saronic:
> > 
> 
> It's not a huge deal, but IIRC when adding it we somewhat thought there isn't much point since you can do `sliding_window_view(...)[::2]`, so it seemed easy enough to just not integrate it.

Good call. I made this very late at night when my brain thought this was an easier solution than just slicing the input. I'm going to replace this with an update to in the docs instead. This particular approach has too many corner cases IMO, including the inability to handle duplicate axes cleanly.

PR #30957, opened 2026-03-07
BUG: fix type issues in uses if PyDataType macros

charris:
Backport of #30918.

@kumaraditya303 and I ran into these working on support for the opaque PyObject build.

These `PyDataType` macros expect a `PyArray_Descr *` argument, but in all three cases happen to work because the macros use direct struct field accesses and all these structs share field names.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -354,7 +354,7 @@ jobs:
       run: |
         pip install -r requirements/build_requirements.txt
         pip install -r requirements/test_requirements.txt
-        pip install vulture
+        pip install "vulture!=2.15"
     - name: Build and install NumPy
       run: |
         # Install using the fastest way to build (no BLAS, no SIMD)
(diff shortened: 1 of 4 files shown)


pllim:
Hi. Do you think this could cause downstream failure as such?

* https://github.com/astropy/astropy/issues/19422

mattip:
Thanks @plim for the heads up. I can't actually see what versions of numpy and cython are getting installed, cibuildwheel is swallowing the stdout in those jobs. For others: the failure is in the cython `__Pyx_Get_FusedFunction_Type`, It is calling `Py_INCREF(tp)`, and the error is `note: expected ‘PyObject *’ {aka ‘struct _object *’} but argument is of type ‘PyTypeObject *’ {aka ‘struct _typeobject *’}`. This looks like a Cython problem? Maybe fallout from cython/cython#7556, see [my comment](https://github.com/cython/cython/pull/7556#pullrequestreview-3951812191)

pllim:
Sorry for the noise, I think it was on our end after all (https://github.com/astropy/astropy/pull/19429/changes).

pllim:
Oh actually, I see https://github.com/cython/cython/pull/7576 now. Well, I am glad upstream is fixing it anyway.

PR #30933, opened 2026-03-04
DOC: add AI policy

mattip:
## PR Summary

This is prompted by the [mailing list discussion](https://mail.python.org/archives/list/numpy-discussion@python.org/thread/LAR7P3KQWHWAIKYSHTS2MY7X4HUBVA3L/) about AI. It is a copy of the [scipy policy](https://scipy.github.io/devdocs/dev/conduct/ai_policy.html#ai-agents) with a few small changes in wording, and adding a reference to @melissawm's great list of policies. The copyright section will be the target of another PR, we could leave it out if it proves to be the focus of discussion here. Once this PR starts to settle, I will propose another (based on the reformatting in #30932) to add AI disclosure checkboxes to the PR template.

PR #30963, opened 2026-03-08
DOC: AI policy exception for Dependabot

jorenham:
After merging https://github.com/numpy/numpy/pull/30933 I realized that Dependabot might also be considered to be an "AI Agent", seeing as it's not unlikely that it uses AI for *something*, or otherwise will do so in the near future.


seberg:
Honestly, let's just not go there, IMO. First of all, isn't AI, second of all it is *configured* on the project itself, so went through a normal contribution and signed off by maintainers.

Detailing such tiny exceptions (that, IMO, aren't even exceptions), doesn't seem helpful?

PR #30964, opened 2026-03-08
DOC: Consolidate Fortran compiler instructions into F2PY docs

### LINKED ISSUE #28187: DOC: Duplication of Fortran installation instructions

Currently there are two places in the documentation which talks about installing a Fortran compiler on Windows:
- The [building from source page](https://numpy.org/doc/stable/building/index.html)
- The [F2PY and Windows page](https://numpy.org/doc/stable/f2py/windows/index.html)
As noted in https://github.com/numpy/numpy/pull/28141#pullrequestreview-2561035821 I think its best to have all the Fortran-Windows-Compiler stuff in the F2Py docs, and link out to it from the build documentation page.

_No response_

HaoZeke:
Move detailed Fortran compiler setup for Windows from the main building docs to the F2PY Windows documentation. The building page now links to the F2PY docs for Fortran-specific details, reducing duplication. FWIW it's just a restructure, I haven't had time yet to actually revisit these docs.

Closes #28187

### DIFF
--- a/doc/source/building/index.rst
+++ b/doc/source/building/index.rst
@@ -140,66 +140,26 @@ your system.
   .. tab-item:: Windows
     :sync: windows
 
-    On Windows, the use of a Fortran compiler is more tricky than on other
-    platforms, because MSVC does not support Fortran, and gfortran and MSVC
-    can't be used together. If you don't need to run the ``f2py`` tests, simply
-    using MSVC is easiest. Otherwise, you will need one of these sets of
-    compilers:
-
-    1. MSVC + Intel Fortran (``ifort``)
-    2. Intel compilers (``icc``, ``ifort``)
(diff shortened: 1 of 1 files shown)


mattip:
This seems like an improvement. There is more to do to clean up the windows build instructions since llvm + flang seems like what we should be promoting, but good enough for now. 

Why is there an added test and source file in the PR?

HaoZeke:
> This seems like an improvement. There is more to do to clean up the windows build instructions since llvm + flang seems like what we should be promoting, but good enough for now.
> 
> Why is there an added test and source file in the PR?

My bad, too many branches, too little sleep 😅 . Fixed

mattip:
Thanks @HaoZeke

PR #30965, opened 2026-03-08
BUG: Show allocatable arrays in `dir()` for f2py modules

### LINKED ISSUE #27696: BUG: Allocatable array variables don't show up in `dir`
@HaoZeke Thank you very much for the fix!

I installed using
```
pip install git+https://github.com/numpy/numpy.git@refs/pull/27695/merge
```

Yes, it is including the `lanedata` module.  Interestingly, the allocatable module variable `theta`, however, is not listed by `dir`
```
In [9]: dir(laneemden._solver.lanedata)
Out[9]: 
['__call__',
 '__class__',
 '__delattr__',
 '__dir__',
 '__doc__',
 '__eq__',
 '__format__',
 '__ge__',
 '__getattribute__',
 '__getstate__',
 '__gt__',
 '__hash__',
 '__init__',
 '__init_subclass__',
 '__le__',
 '__lt__',
 '__ne__',
 '__new__',
 '__reduce__',
 '__reduce_ex__',
 '__repr__',
 '__setattr__',
 '__sizeof__',
 '__str__',
 '__subclasshook__',
 'maxdata',
 'ndata']
```
but it is accessible, and returns `None` if not allocated and returns the array when allocated.  Even when allocated, however, it is not listed.
```
In [11]: laneemden._solver.lanedata.<TAB>
maxdata ndata   
In [11]:
```
This may not be new behaviour (I do not recall having tried this before), but is a bit curious, and might be confusing.

Before allocating
```
In [5]: laneemden._solver.lanedata.theta

In [6]:
```
and after allocating (`laneemden.solver.test()`)
```
In [10]: laneemden._solver.lanedata.theta
Out[10]: 
array([[ 1.00000000e+00,  0.00000000e+00],
       [ 9.99999999e-01, -2.03450521e-05],
       [ 9.99999998e-01, -4.06901040e-05],
       ...,
       [ 5.09606183e-06, -4.24312354e-02],
       [ 2.50628767e-06, -4.24304844e-02],
       [-8.34406605e-08, -4.24297334e-02]])

In [11]:
```

_Originally posted by @2sn in https://github.com/numpy/numpy/issues/27695#issuecomment-2453721148_

HaoZeke:
Add a `__dir__` method to `PyFortranObject` that includes all defs names (routines, variables, and allocatable arrays), not just the keys in the object dict. Previously, allocatable arrays were accessible via getattr but invisible to `dir()` because `PyFortranObject_New` only populated the dict with routines and non-allocatable variables.

Closes #27696

cc @2sn (sorry for the long delay)

### DIFF
--- a/doc/release/upcoming_changes/30965.improvement.rst
+++ b/doc/release/upcoming_changes/30965.improvement.rst
@@ -1,0 +1,4 @@
+``f2py`` modules now show allocatable arrays in ``dir()``
+---------------------------------------------------------
+Allocatable module variables wrapped by ``f2py`` now appear in ``dir()``
+output, matching their accessibility by name.
--- a/numpy/f2py/src/fortranobject.c
+++ b/numpy/f2py/src/fortranobject.c
@@ -576,6 +576,47 @@ fortran_repr(PyFortranObject *fp)
     return repr;
 }
 
+static PyObject *
+fortran_dir(PyFortranObject *fp, PyObject *Py_UNUSED(args))
+{
+    int i;
+    PyObject *dir_list = PyDict_Keys(fp->dict);
+    if (dir_list == NULL) {
+        return NULL;
+    }
+    for (i = 0; i < fp->len; i++) {
+        PyObject *name = PyUnicode_FromString(fp->defs[i].name);
+        if (name == NULL) {
+            Py_DECREF(dir_list);
+            return NULL;
+        }
+        int contains = PySequence_Contains(dir_list, name);
+        if (contains == -1) {
+            Py_DECREF(name);
+            Py_DECREF(dir_list);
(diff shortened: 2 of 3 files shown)


mattip:
Makes sense and works nicely. Why is the test in `test_regressions`? Did this work once and then stop working?

HaoZeke:
> Makes sense and works nicely. 

Thanks for the test ^_^

> Why is the test in `test_regressions`? Did this work once and then stop working?

Ah, no, there was an allocatable array so I reused it while testing, but I've moved it now.

mattip:
Thanks @HaoZeke. Still a bit weird that the test uses a file from the src/regressions directory, but that is a small paper cut.

PR #30967, opened 2026-03-08
BUG: Fix F77/F90 line wrapping generating invalid continuations

HaoZeke:
Fix the F77 wrapping condition to avoid empty remainders and fix the F90 cleanup to properly remove both the empty continuation line and the trailing ampersand on the preceding line.

Closes #28474.

cc @joshuashaffer


charris:
Thanks @HaoZeke .

PR #30969, opened 2026-03-08
BUG: f2py: replace conflicting external decls with interface blocks in F90 callback wrappers

HaoZeke:
When f2py generates Fortran 90 wrapper subroutines for functions that have both callback arguments and assumed-shape arrays, the wrapper previously emitted bare ``external`` declarations for the callbacks. This conflicts with the interface blocks for those same callbacks that appear inside the nested interface for the wrapped function.

Closes #20157.

Strict Fortran compilers (Intel ifort/ifx >= 2019) correctly reject this as a standard violation: a dummy procedure cannot be both declared ``external`` and given an explicit interface in the same scope.

The fix has two parts:
- At the outer wrapper subroutine level, emit the callback's interface block instead of ``external``.
- Inside the nested interface block for the wrapped function, strip redundant ``external`` lines for callbacks that already have inline interface blocks.

I will add that this is a "blind" fix, in that I don't have the intel/arm compilers handy (nor the disk space to get them rn). cc @akesandgren or @stepannassyr or @boegel since they're on the issue.


pearu:
Could you add a test that ensures legacy F77 support is not broken after this change?

PR #30970, opened 2026-03-08
MAINT: Deprecate PyArray_BufferConverter and remove Python 2 buffer protocol remnants

### LINKED ISSUE #30813: [ENH] Remove remnants of Python 2 buffer protocol from NumPy internals

sharmavikas18:
…l remnants

Description
Fixes #30813
This PR removes remaining remnants of the legacy Python 2 buffer protocol from NumPy internals and moves related structures toward the modern Python buffer protocol. The changes isolate legacy structures from the public API, introduce internal definitions where still required for compatibility, and formally deprecate legacy conversion helpers.

These updates reduce technical debt while preserving source and binary compatibility for downstream projects.
Changes
Public header modernization
numpy/_core/include/numpy/ndarraytypes.h
Make PyArray_Chunk opaque when building NumPy internally (NPY_INTERNAL_BUILD).
Prevent internal code from relying on the legacy public definition while maintaining compatibility for external extensions.
Internal compatibility layer
numpy/_core/src/multiarray/common.h
Added a binary-compatible internal definition of:

struct _PyArray_Chunk
Ensures internal code paths such as the ndarray constructor (array_new) continue to work during the transition.
Deprecation of legacy converters
numpy/_core/src/multiarray/conversion_utils.c

Added a deprecation warning to:

PyArray_BufferConverter

Marked as deprecated starting in NumPy 2.5.

Users are encouraged to use the standard Python buffer protocol instead.

Documentation updates

doc/source/reference/c-api/array.rst

Updated documentation for PyArray_FromBuffer.

Removed references to the legacy __buffer__ attribute and the single-segment buffer protocol.

Added:

.. deprecated:: 2.5

for PyArray_BufferConverter.

Verification

Confirmed np.ndarray(..., buffer=...) triggers the expected deprecation warning.

Verified np.frombuffer continues to work normally without warnings.

Ensured existing internal usage of PyArray_Chunk remains functional.

Motivation
These changes remove remaining dependencies on legacy Python 2 buffer protocol mechanisms while preserving compatibility with existing internal code paths. This reduces technical debt and moves NumPy further toward exclusive use of the modern Python buffer protocol.

### DIFF
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -595,14 +595,13 @@ From other objects
         PyObject* buf, PyArray_Descr* dtype, npy_intp count, npy_intp offset)
 
     Construct a one-dimensional ndarray of a single type from an
-    object, ``buf``, that exports the (single-segment) buffer protocol
-    (or has an attribute __buffer\__ that returns an object that
-    exports the buffer protocol). A writeable buffer will be tried
-    first followed by a read- only buffer. The :c:data:`NPY_ARRAY_WRITEABLE`
+    object, ``buf``, that exports the modern buffer protocol.
+    A writeable buffer will be tried
+    first followed by a read-only buffer. The :c:data:`NPY_ARRAY_WRITEABLE`
     flag of the returned array will reflect which one was
     successful. The data is assumed to start at ``offset`` bytes from
     the start of the memory location for the object. The type of the
-    data in the buffer will be interpreted depending on the data- type
+    data in the buffer will be interpreted depending on the data-type
     descriptor, ``dtype.`` If ``count`` is negative then it will be
     determined from the size of the buffer and the requested itemsize,
     otherwise, ``count`` represents how many elements should be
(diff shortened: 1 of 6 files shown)


mattip:
This PR seems to be created by an AI-agent without disclosure, which violates our new [AI guidelines](https://numpy.org/devdocs/dev/ai_policy.html#disclosure). Closing. Please reopen and accept my apologies if this is incorrect.

PR #30971, opened 2026-03-08
BUG: f2py: fix allocatable character arrays in module data

### LINKED ISSUE #21674: BUG: Allocation of character arrays in F2PY
I've recently come across this issue while using f2py in numpy 1.20.3. As a minimum working example, I have the following simple fortran module which declares two allocatable arrays (`charge` is real type and `names` is character). The subroutine "main" will allocate them:

```
! -*- f90 -*-

      module inputs
      implicit none
              real*8, dimension(:,:), target, allocatable :: charge
              integer is_alloc
              character(len=10), allocatable, dimension(:) :: names
      end module inputs

      subroutine main(N)
      use inputs
      implicit none

      integer ii, jj, N
      
      IF (allocated(charge)) THEN
         deallocate(charge)
      end if
      allocate(charge(N, N))
      allocate(character(len=10) :: names(N)) !, source="TEST")

      END subroutine
```

I can build this by running `python setup.py build_ext --inplace` with the following setup.py file (having put the above code in test.f90):
```
from numpy.distutils.core import Extension
from numpy.distutils.core import setup
from numpy.distutils.misc_util import Configuration


F90_COMBINED = "test.f90"

calc_ext = Extension(
    name="alloc",
    sources=[F90_COMBINED],
    include_dirs=['.'],
    libraries=['gsl', 'gslcblas'],
    extra_f90_compile_args=['-Wargument-mismatch'],
)

def configuration(parent_package='', top_path=None):

    config = Configuration('', parent_package, top_path,
                           ext_modules=[calc_ext])
    return config

if __name__ == "__main__":
    setup(configuration=configuration)
```

This builds a module called `alloc`. Here's a screencap of what happens if I try to access each allocated array.
image

Any suggestions would be greatly appreciated!

_Originally posted by @aelanman in https://github.com/numpy/numpy/issues/10027#issuecomment-988511053_

HaoZeke:
Happily I was wrong in https://github.com/numpy/numpy/issues/21674#issuecomment-2481897399 and this wasn't so bad.

Uses `get_descr_from_type_and_elsize` for allocatable arrays in `fortran_getattr`, matching the non-allocatable path. Without this, `NPY_STRING` arrays get `itemsize=0` which triggers `ValueError`.

Closes #21674


charris:
Thanks @HaoZeke .

PR #30976, opened 2026-03-09
MAINT: Avoid warning on no-cblas build

seberg:
Oddly enough, I never made to have to have cblas in a local build and this warning started showing up.


mattip:
Which warning?

seberg:
Ah sorry, `typenum` is only used withing the `#ifdef` block, so it is an unused variable.

mattip:
Thanks @seberg

PR #30966, opened 2026-03-08
MAINT: Use `npy_cfloat/npy_cdouble/npy_clongdouble` in `f2py`

HaoZeke:
Replaces custom complex struct definitions (struct {T r,i;}) with the `npy_cfloat`, `npy_cdouble`, `npy_clongdouble` types. Updates accessor patterns from .r/.i to `npy_creal/npy_cimag/npy_csetreal/npy_csetimag`. Replaces `{0.0,0.0}` initializers with `npy_cpack()`.

The `TRYCOMPLEXPYARRAYTEMPLATE` macro now uses helper functions to extract real/imag parts as doubles, since the `npy` accessor functions are type-specific while the macro is type-generic.

Closes #19996

PR #30983, opened 2026-03-09
BUG: f2py: restore .r/.i field access on complex types via union typedef

HaoZeke:
#30966 changed `f2py`'s complex_double/complex_float/complex_long_double typedefs from `struct {T r,i;}` to direct `npy_cdouble`/`npy_cfloat`/ `npy_clongdouble` aliases.

and forgot to check downstream. In this case it broke scipy https://github.com/scipy/scipy/issues/24775 because .pyf callstatements use `(a+k)->r` and `(a+k)->i` on complex_double* pointers, and npy_cdouble (double _Complex) has no .r/.i fields as reported by @lucascolley (thanks!)

Fix by using a union containing an anonymous struct (for .r/.i backward compat) and the native npy_* type (for layout guarantee):

    typedef union { struct {double r,i;}; npy_cdouble _npy; } complex_double;

Internal `_from_pyobj` functions and `pyobj_from` macros now access the `_npy` member for npy_cset*/npy_creal*/npy_cimag* calls, while `rules.py` and `callstatement .r/.i` patterns work unchanged through the anonymous struct.

Closes scipy/scipy#24775. Tested on SciPy this time 😅

### DIFF
--- a/numpy/f2py/cfuncs.py
+++ b/numpy/f2py/cfuncs.py
@@ -89,9 +89,9 @@ typedefs['long_double'] = """
 typedef long double long_double;
 #endif
 """
-typedefs['complex_long_double'] = 'typedef npy_clongdouble complex_long_double;'
-typedefs['complex_float'] = 'typedef npy_cfloat complex_float;'
-typedefs['complex_double'] = 'typedef npy_cdouble complex_double;'
+typedefs['complex_long_double'] = 'typedef union { struct {long double r,i;}; npy_clongdouble _npy; } complex_long_double;'
+typedefs['complex_float'] = 'typedef union { struct {float r,i;}; npy_cfloat _npy; } complex_float;'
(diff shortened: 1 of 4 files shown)


seberg:
@HaoZeke at this point, I am not actually sure that we gain anything with this change? It seemed like a good cleanup either way, but if it is public API, is a union really any better than just keeping the original weird typedef?

HaoZeke:
> @HaoZeke at this point, I am not actually sure that we gain anything with this change? It seemed like a good cleanup either way, but if it is public API, is a union really any better than just keeping the original weird typedef?

I'm hoping to basically spend some time on the weekend migrating downstream away from the old `.r/.i` stuff and then we can deprecate and cleanup in a few releases..

seberg:
Ok, lets try this then, thanks.

PR #30984, opened 2026-03-09
ENH: Add internal real/imag ufuncs and use them for the attrs

seberg:
**This is a work in progress, although already functional**

I am trying to figure out how to restructure `.imag` and `.real` to:
* Allow complex dtypes not defined inside NumPy
* Ideally, not hard-code that they must be views
  - *Fix `object` dtype behavior* (a bit?) by not returning views.
  - It was also brought up by Warren.
* Define one part of a future `finfo` (the ability to fetch the corresponding real type)
* (while not breaking current strange behavior)

The approach here is to define _internal_ `imag` and `real` ufuncs, plus we need a few small tweaks to allow those ufuncs to return actual `view`s.

One wrinkle is that `arr.imag = 0` works and right now is a bit hacked, because I wasn't sure I like a `set_imag` ufunc (but maybe that is the better solution?).
There is also a wrinkle that if `arr.real`/`arr.imag` still need weird special handling for backcompat since a ufunc really can't return a broadcasted array of zeros.

(The code looks wrong, but it is mainly boilerplate ufunc code+docs.)

Generally, I think this is a reasonable approach to make it a ufuncs, although there may be some details to hash out. Ping @ngoldbaum and @mhvk in case you have some thoughts already. 

EDIT: Closes gh-1740

*EDIT: Should have mentioned that I tried to use an agent to write some of the ufunc code, but it didn't do a job I liked, so there is practically nothing left of it.*

### DIFF
--- a/doc/release/upcoming_changes/30984.capi.rst
+++ b/doc/release/upcoming_changes/30984.capi.rst
@@ -1,0 +1,4 @@
+* It is now possible to register ``".real"`` and ``".imag"``
+  ArrayMethods via ``PyUFunc_AddLoopsFromSpecs``.  These will
+  be used for ``.imag`` and ``.real`` and should (presumably)
+  set ``*view_offset`` in their ``resolve_descriptors`` function.
--- a/doc/release/upcoming_changes/30984.capi.rst
+++ b/doc/release/upcoming_changes/30984.capi.rst
@@ -1,4 +1,5 @@
-* It is now possible to register ``".real"`` and ``".imag"``
+* It is now possible to register ``"real"`` and ``"imag"``
   ArrayMethods via ``PyUFunc_AddLoopsFromSpecs``.  These will
-  be used for ``imag`` and ``real`` and should (presumably)
-  set ``*view_offset`` in their ``resolve_descriptors`` function.
+  be used for ``imag`` and ``real`` and should normally
+  set ``*view_offset`` in their ``resolve_descriptors`` function
+  to allow the array attributes to return views.
--- a/doc/release/upcoming_changes/30984.capi.rst
+++ b/doc/release/upcoming_changes/30984.capi.rst
@@ -1,4 +1,4 @@
 * It is now possible to register ``".real"`` and ``".imag"``
   ArrayMethods via ``PyUFunc_AddLoopsFromSpecs``.  These will
-  be used for ``.imag`` and ``.real`` and should (presumably)
+  be used for ``imag`` and ``real`` and should (presumably)
   set ``*view_offset`` in their ``resolve_descriptors`` function.
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -1899,8 +1899,9 @@ with the rest of the ArrayMethod API.
             entry points, ``(module ':')? (object '.')* name``, with ``numpy``
             the default module. Examples: ``sin``, ``strings.str_len``,
             ``numpy.strings:str_len``.
-            ``"sort"``, ``"argsort"``, ``".real"``, ``".imag"`` and specific names
-            as they do not correspond directly to ufuncs.
+            Note that some names are supported but do not directly correspond
+            to ufuncs: ``"sort"``, ``"argsort"``, ``"real"``, ``"imag"``.
+            (These do use ufunc-likes or even ufuncs internally.)
 
         .. c:member:: PyArrayMethod_Spec *spec
 
@@ -1960,7 +1961,7 @@ with the rest of the ArrayMethod API.
         skipping a recursion protection step.
         This mainly allows the promoter to add new loop to the ufunc that must
         now match instead of the promoter itself.
-        (Normally, a promoter must modify)
+        (Normally, a promoter must modify the DTypes help find the right loop.)
 
 .. c:function:: int PyUFunc_GiveFloatingpointErrors( \
                         const char *name, int fpe_errors)
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -1942,6 +1942,22 @@ with the rest of the ArrayMethod API.
    attempt a search for a new loop or promoter that can accomplish the operation
    by casting the inputs to the "promoted" DTypes.
 
+    A promoter should honor ``signature[]`` (if set). A promoter must return ``-1``
+    on failure. A Python error may be set but is not required (a general error is
+    set in either paths, although the original error is chained).
+    A promoter must return ``0`` or ``1`` on success.  NumPy normally checks that
+    ``new_op_dtypes`` are different from ``op_dtypes`` to prevent recursion.
+    This check is skipped if the promoter returns ``1``, which allows the promoter
+    to add a new loop (when adding a new loop, ``new_op_dtypes`` should be identical
+    to ``op_dtypes``).
+
+    .. versionchanged:: 2.5
+        After 2.5 a return of ``1`` indicates that the promoter was successful
+        skipping a recursion protection step.
+        This mainly allows the promoter to add new loop to the ufunc that must
+        now match instead of the promoter itself.
+        (Normally, a promoter must modify)
+
 .. c:function:: int PyUFunc_GiveFloatingpointErrors( \
                         const char *name, int fpe_errors)
 
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -1899,6 +1899,8 @@ with the rest of the ArrayMethod API.
             entry points, ``(module ':')? (object '.')* name``, with ``numpy``
             the default module. Examples: ``sin``, ``strings.str_len``,
             ``numpy.strings:str_len``.
+            ``"sort"``, ``"argsort"``, ``".real"``, ``".imag"`` and specific names
+            as they do not correspond directly to ufuncs.
 
         .. c:member:: PyArrayMethod_Spec *spec
 
@@ -1912,6 +1914,8 @@ with the rest of the ArrayMethod API.
     Add multiple loops to ufuncs from ArrayMethod specs. This also
     handles the registration of methods for the ufunc-like functions
     ``sort`` and ``argsort``. See :ref:`array-methods-sorting` for details.
+    As well as for the array attributes ``.real`` and ``.imag`` needed
+    for user defined complex DTypes (with ``".real"`` and ``".imag"`` as names).
 
     The ``slots`` argument must be a  NULL-terminated array of
     `PyUFunc_LoopSlot` (see above), which give the name of the
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -1914,9 +1914,9 @@ with the rest of the ArrayMethod API.
 
     Add multiple loops to ufuncs from ArrayMethod specs. This also
     handles the registration of methods for the ufunc-like functions
-    ``sort`` and ``argsort``. See :ref:`array-methods-sorting` for details.
-    As well as for the array attributes ``.real`` and ``.imag`` needed
-    for user defined complex DTypes (with ``".real"`` and ``".imag"`` as names).
+    ``sort`` and ``argsort`` (see :ref:`array-methods-sorting` for details),
+    as well as for the array attributes ``.real`` and ``.imag`` needed
+    for user defined complex DTypes (with ``"real"`` and ``"imag"`` as names).
 
     The ``slots`` argument must be a  NULL-terminated array of
     `PyUFunc_LoopSlot` (see above), which give the name of the
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -2666,7 +2666,7 @@ add_newdoc('numpy._core.multiarray', 'ndarray', ('imag',
     The imaginary part of the array.
 
     Returns a view into the original array for complex arrays.
-    For non-compelx ones, returns a zero array of the same dtype.
+    For non-complex arrays, returns a zero array of the same dtype.
     For ``object`` arrays returns elementwise ``.imag`` or ``0``
     if ``.imag`` is undefined.
 
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -2665,6 +2665,11 @@ add_newdoc('numpy._core.multiarray', 'ndarray', ('imag',
     """
     The imaginary part of the array.
 
+    Returns a view into the original array for complex arrays.
+    For non-compelx ones, returns a zero array of the same dtype.
+    For ``object`` arrays returns elementwise ``.imag`` or ``0``
+    if ``.imag`` is undefined.
+
     Examples
     --------
     >>> import numpy as np
@@ -2855,6 +2860,9 @@ add_newdoc('numpy._core.multiarray', 'ndarray', ('ndim',
 add_newdoc('numpy._core.multiarray', 'ndarray', ('real',
     """
     The real part of the array.
+
+    Usually returns a view into the original array, but returns
+    elementwise ``.real`` for arrays of objects.
 
     Examples
     --------
--- a/numpy/_core/code_generators/ufunc_docstrings.py
+++ b/numpy/_core/code_generators/ufunc_docstrings.py
@@ -4544,8 +4544,8 @@ add_newdoc('numpy._core.umath', 'real',
 
     Notes
     -----
-    This ufunc is used internally in the implement of the `ndarray.real`
-    attribute and `np.real` function which should be used instead.
+    This ufunc is used internally to implement the `ndarray.real`
+    attribute and the `np.real` function. It should not be used directly.
 
     """)
 
@@ -4573,8 +4573,8 @@ add_newdoc('numpy._core.umath', 'imag',
 
     Notes
     -----
-    This ufunc is used internally in the implement of the `ndarray.imag`
-    attribute and `np.imag` function which should be used instead.
+    This ufunc is used internally to implement the `ndarray.imag`
+    attribute and the `np.imag` function. It should not be used directly.
 
     """)
 
--- a/numpy/_core/code_generators/ufunc_docstrings.py
+++ b/numpy/_core/code_generators/ufunc_docstrings.py
@@ -4524,10 +4524,7 @@ add_newdoc('numpy._core.umath', 'bitwise_count',
 
 add_newdoc('numpy._core.umath', 'real',
     """
-    Returns the real part of the elements in the array. Unlike most ufuncs,
-    the return may be a view into the original array.
-
-    Unlike typical ufuncs, the return is typically a view into the original array.
+    Returns the real part of the elements in the array.
 
     Parameters
     ----------
@@ -4547,8 +4544,8 @@ add_newdoc('numpy._core.umath', 'real',
 
     Notes
     -----
-    This ufunc is used to implement the `ndarray.real` attribute which is usually
-    preferred.
+    This ufunc is used internally in the implement of the `ndarray.real`
+    attribute and `np.real` function which should be used instead.
 
     """)
 
@@ -4576,11 +4573,8 @@ add_newdoc('numpy._core.umath', 'imag',
 
     Notes
     -----
-    This ufunc is used to implement the `ndarray.imag` attribute which is usually
-    preferred. Unlike the attribute, this ufunc raises an error for non-complex
-    inputs.
-    The attribute itself additionally returns a broadcast, read-only array of zeros
-    for known real dtypes.
+    This ufunc is used internally in the implement of the `ndarray.imag`
+    attribute and `np.imag` function which should be used instead.
 
     """)
 
--- a/numpy/_core/code_generators/ufunc_docstrings.py
+++ b/numpy/_core/code_generators/ufunc_docstrings.py
@@ -4522,6 +4522,68 @@ add_newdoc('numpy._core.umath', 'bitwise_count',
 
     """)
 
+add_newdoc('numpy._core.umath', 'real',
+    """
+    Returns the real part of the elements in the array. Unlike most ufuncs,
+    the return may be a view into the original array.
+
+    Unlike typical ufuncs, the return is typically a view into the original array.
+
+    Parameters
+    ----------
+    x : array_like
+    $PARAMS
+
+    Returns
+    -------
+    y : ndarray
+        Real part of input array.
+        $OUT_SCALAR_1
+
+    See Also
+    --------
+    ndarray.real
+    ndarray.imag
+
+    Notes
+    -----
+    This ufunc is used to implement the `ndarray.real` attribute which is usually
+    preferred.
+
+    """)
+
+add_newdoc('numpy._core.umath', 'imag',
+    """
+    Returns the imaginary part of the elements in the array.
+
+    Unlike typical ufuncs, the return is typically a view into the original array.
+
+    Parameters
+    ----------
+    x : array_like
+    $PARAMS
+
+    Returns
+    -------
+    y : ndarray
+        Complex part of input array or zeros.
+        $OUT_SCALAR_1
+
+    See Also
+    --------
+    ndarray.imag
+    ndarray.real
+
+    Notes
+    -----
+    This ufunc is used to implement the `ndarray.imag` attribute which is usually
+    preferred. Unlike the attribute, this ufunc raises an error for non-complex
+    inputs.
+    The attribute itself additionally returns a broadcast, read-only array of zeros
+    for known real dtypes.
+
+    """)
+
 add_newdoc('numpy._core.umath', 'str_len',
     """
     Returns the length of each element. For byte strings,
--- a/numpy/_core/src/multiarray/array_method.c
+++ b/numpy/_core/src/multiarray/array_method.c
@@ -86,7 +86,7 @@ default_resolve_descriptors(
      * abstract ones or unspecified outputs).  We can use the common-dtype
      * operation to provide a default here.
      */
-    if (method->casting == NPY_NO_CASTING && (method->flags & _NPY_METH_IS_CAST)) {
+    if (method->casting == NPY_NO_CASTING) {
         /*
          * By (current) definition no-casting should imply viewable.  This
          * is currently indicated for example for object to object cast.
@@ -129,9 +129,9 @@ is_contiguous(
  * param move_references UNUSED -- listed below but doxygen doesn't see as a parameter
  * @param strides Array of step sizes for each dimension of the arrays involved
  * @param out_loop Output pointer to the function that will perform the strided loop.
- * @param out_transferdata Output pointer to auxiliary data (if any)
+ * @param out_transferdata Output pointer to auxiliary data (if any) 
  *        needed by the out_loop function.
- * @param flags Output pointer to additional flags (if any)
+ * @param flags Output pointer to additional flags (if any) 
  *        needed by the out_loop function
  * @returns 0 on success -1 on failure.
  */
--- a/numpy/_core/src/multiarray/array_method.c
+++ b/numpy/_core/src/multiarray/array_method.c
@@ -86,7 +86,7 @@ default_resolve_descriptors(
      * abstract ones or unspecified outputs).  We can use the common-dtype
      * operation to provide a default here.
      */
-    if (method->casting == NPY_NO_CASTING) {
+    if (method->casting == NPY_NO_CASTING && (method->flags & _NPY_METH_IS_CAST)) {
         /*
          * By (current) definition no-casting should imply viewable.  This
          * is currently indicated for example for object to object cast.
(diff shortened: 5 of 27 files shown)


numpy/_core/src/umath/real_imag_ufuncs.cpp
```diff
@@ -0,0 +1,218 @@
...
+    npy_intp istride = strides[0];
+    npy_intp ostride = strides[1];
+
+    while (N--) {
+        real_type value;
+        if constexpr (real_part) {
+            value = *reinterpret_cast<real_type *>(in);
+        }
+        else {
+            value = *reinterpret_cast<real_type *>(in + sizeof(real_type));
```
mhvk: The addition should be done outside the loop (even if we do not exercise it!).

mhvk (commented):
@seberg - I like the idea of the ufuncs, but two worries.
1. doing `.real` and `.imag` happens a *lot* and is time-critical, did you check the performance? (e.g., most of my code dealing with complex numbers has `def power(z): return z.real**2+z.imag**2`). I fear you may have to stick with having the standard view taking as the default...
2.  I am less enchanted by the "magic" that now sometimes outputs can be views. I think these functions should behave like all ufuncs, e.g.,`np.positive`, and thus return a copy by default.

On (2), of course we do want to allow views and avoid actual work in that case, but could we introduce an actual signal that that is wanted (and that would thus work for `np.positive` as well)? At the risk of repurposing too much, maybe `casting='view'` or giving `out=...` an extra meaning, that a view is OK?

mhvk:
p.s. The broadcasted zero makes me wonder whether a ufunc is in fact the right option...

seberg:
> p.s. The broadcasted zero makes me wonder whether a ufunc is in fact the right option...

Right, that is annoying. My angle right now was that a custom solution may be a bit much (we could have a flag or a `get_imag_constant()` slot).
Rather, I was thinking that for user-dtypes we could apply this trick in the future if they indicate `real` (via a flag and/or new abstract dtype).
For those it is obvious, and honestly `string_arr.imag` returning zeros is pretty ridiculous (sure datetimes are a bit tricky because are they real?), but then, I dunno if I think this zeros trick is extremely vital?

> I am less enchanted by the "magic" that now sometimes outputs can be views.

Yeah, fair, I wanted to start from the side of trying everything in the ufunc (it seems conceptually nice if it works).
We could indeed add a `view_ok` style parameter to the ufunc.  The other thing, I think I will try first for now is to side-step the ufunc a bit more (also to reduce overhead, which sure is pretty significant – although of course still a fast ufunc):
* Reach into the ufunc to avoid calling it for the view-compatible paths.
* Or make `real/imag` special, possibly _still_ have a ufunc for the non-view paths but make it reach into the DType for the implementation.
* (We could avoid the ufunc even for the full path, but I think that is not a good idea to do here yet, without an imperative need.)

(I honestly think it isn't all that bad to add overhead, even doubling(!), which is what I expect the first approach might do in the worst case. Because once you have an expression like yours, even doubling it is still <10% of the run-time. And that is for tiny arrays.)

*EDIT:* OK, after trying the first a bit. I don't really like reaching into the dispatching part, so let's try the second...

seberg:
OK, I did the iteration of moving it _onto_ to the DType itself:
* Registration now via `".imag"` and `".real"` similar to the sorts.  (I forced the sorts to match `"numpy:sort"` exactly, just for simplicity.)
* Removed the UFunc handling for views, but tweaked the promotion very slightly in order to allow a `promoter` that adds loops to the ufunc from the DType slot.
  (I always meant promoters being able to do this, but without this, the recursion guard might stop that from happening.)
  * I.e. the `object` path still calls into the UFunc, but the UFunc doesn't "own" the loops (sure, you could add _additional_ loops to it that wouldn't affect `.real` and `.imag`).
* The actual attributes, now access the `resolve_descriptors` directly.  It is 20% slower (40ns -> 49ns) for me. But that is a *very* small thing (it might also go away if looking at things a bit closer).
  *EDIT: Actually, at least practically all of this is attributable to the `view` creation itself taking longer on current `main`.*

---

Again, still needs tests, but maybe this is the right approach.  Maybe the ufuncs don't even need to exist, but it is a fairly convenient way for the new `object` path (in principle, it would be nice to call the ufunc core directly here, but that requires more refactor).

~Still needs tests at least and clearly fixes!~

EDIT: Still needs at least release notes (it is quite substantial with object arrays). But if we think this is a reasonable path, then I think it is far along now.

MaanasArora:
I went over this very briefly and it looks quite nice! The user dtype API in particular benefits from having ufuncs anyway, I think. Sorts and argsorts are similar insofar they're inplace, so having these too on the dtype fits well!

I'm still wrapping my head around this a bit, but yes, putting it on the dtype made it nicer as the ufuncs have less "responsibility" for the view path now. I think it is fine for the ufunc to return views even if atypical to start with; this is mostly internal after all.

About a signal that a view is wanted, `casting="view"` seems like a natural extension. But it is not likely to be many ufuncs that typically return views so maybe we can get away with them being a bit special (at least in the short run)?

Edit: Going to take a deeper look soon!

doc/source/reference/c-api/array.rst
```diff
@@ -1899,6 +1899,8 @@ with the rest of the ArrayMethod API.
             entry points, ``(module ':')? (object '.')* name``, with ``numpy``
             the default module. Examples: ``sin``, ``strings.str_len``,
             ``numpy.strings:str_len``.
+            ``"sort"``, ``"argsort"``, ``".real"``, ``".imag"`` and specific names
```
mhvk: Bigger: should this just be `real` and `imag`? This since we do have `np.real` and `np.imag` functions, and presumably those are affected as well. (Indeed, I think those in principle could just be replaced by the new ufuncs, no? In which case there would be no need to mention this here...)

Smaller: did you mean to write "are" instead of "and"?  Though perhaps write instead,
```
Note that some names do not correspond directly to ufuncs names, but
instead to functions that use ufuncs internally:
``"sort"``, ``"argsort"``, ``".real"``, ``".imag"`` 
```

doc/source/reference/c-api/array.rst
```diff
@@ -1912,6 +1914,8 @@ with the rest of the ArrayMethod API.
     Add multiple loops to ufuncs from ArrayMethod specs. This also
     handles the registration of methods for the ufunc-like functions
     ``sort`` and ``argsort``. See :ref:`array-methods-sorting` for details.
+    As well as for the array attributes ``.real`` and ``.imag`` needed
+    for user defined complex DTypes (with ``".real"`` and ``".imag"`` as names).
```
mhvk: Seeing it again, I think even if we don't directly replace `np.real` and `np.imag`, I'd be in favour of just calling it `real` and `imag`, without the dots.

doc/source/reference/c-api/array.rst
```diff
@@ -1942,6 +1946,22 @@ with the rest of the ArrayMethod API.
...
+    This check is skipped if the promoter returns ``1``, which allows the promoter
+    to add a new loop (when adding a new loop, ``new_op_dtypes`` should be identical
+    to ``op_dtypes``).
+
+    .. versionchanged:: 2.5
+        After 2.5 a return of ``1`` indicates that the promoter was successful
+        skipping a recursion protection step.
+        This mainly allows the promoter to add new loop to the ufunc that must
+        now match instead of the promoter itself.
+        (Normally, a promoter must modify)
```
mhvk: Maybe `(Normally, a promotor must modify the descriptors to help find an implementation.)

numpy/_core/code_generators/ufunc_docstrings.py
```diff
@@ -4522,6 +4522,68 @@ def add_newdoc(place, name, doc):
 
     """)
 
+add_newdoc('numpy._core.umath', 'real',
+    """
+    Returns the real part of the elements in the array. Unlike most ufuncs,
```
mhvk: Delete `Unlike...` since it is written just below too.

numpy/_core/code_generators/ufunc_docstrings.py
```diff
@@ -4522,6 +4522,68 @@ def add_newdoc(place, name, doc):
...
+        $OUT_SCALAR_1
+
+    See Also
+    --------
+    ndarray.real
+    ndarray.imag
+
+    Notes
+    -----
+    This ufunc is used to implement the `ndarray.real` attribute which is usually
```
mhvk: Maybe 
```
This ufunc is used to implement the `ndarray.real` attribute and the `np.real` function.
Those are generally preferred.
```

numpy/_core/code_generators/ufunc_docstrings.py
```diff
@@ -4522,6 +4522,68 @@ def add_newdoc(place, name, doc):
...
+        $OUT_SCALAR_1
+
+    See Also
+    --------
+    ndarray.imag
+    ndarray.real
+
+    Notes
+    -----
+    This ufunc is used to implement the `ndarray.imag` attribute which is usually
```
mhvk: So this would become,
```
This ufunc is used to implement the `ndarray.real` attribute and the `np.imag` function.
Those are generally preferred. A difference is that this ufunc raises an error for
non-complex data, while the attribute and function  return a broadcast, read-only
array of zeros for known real dtypes.
```

numpy/_core/src/multiarray/array_method.c
```diff
@@ -86,7 +86,7 @@ default_resolve_descriptors(
      * abstract ones or unspecified outputs).  We can use the common-dtype
      * operation to provide a default here.
      */
-    if (method->casting == NPY_NO_CASTING) {
+    if (method->casting == NPY_NO_CASTING && (method->flags & _NPY_METH_IS_CAST)) {
         /*
          * By (current) definition no-casting should imply viewable.  This
```
mhvk, line 91: Should this comment be adjusted? Things will change if the flag is not set (but I'm not sure this can have been used...)

Maybe it should be 
```
By (current) definition a cast that does not need casting should imply
it can be done with a view. This is ...
```

numpy/_core/src/multiarray/convert_datatype.c
```diff
@@ -930,7 +930,7 @@ PyArray_CastDescrToDType(PyArray_Descr *descr, PyArray_DTypeMeta *given_DType)
         Py_INCREF(descr);
         return descr;
     }
-    if (!NPY_DT_is_parametric(given_DType)) {
+    if (!NPY_DT_is_parametric(given_DType) && !NPY_DT_is_abstract(given_DType)) {
```
mhvk, line 933: I guess this was an oversight before?

numpy/_core/src/multiarray/getset.c
```diff
@@ -573,132 +574,103 @@ array_base_get(PyArrayObject *self, void *NPY_UNUSED(ignored))
...
-        default:
-            PyErr_Format(PyExc_ValueError,
-                     "Cannot convert complex type number %d to float",
-                     PyArray_DESCR(self)->type_num);
-            return NULL;
 
-    }
-    type = PyArray_DescrFromType(float_type_num);
-    if (type == NULL) {
+static PyObject *
```
mhvk, line 584: It might be good to add a short comment, in particular to note that on failure, no exception is set.

numpy/_core/src/umath/_scaled_float_dtype.c
```diff
@@ -1026,8 +1026,9 @@ sfloat_init_ufuncs(void) {
     PyUFunc_LoopSlot loops[] = {
         {"multiply", &multiply_spec},
         {"_core._multiarray_umath.add", &add_spec},
-        {"numpy:sort", &sort_spec},
-        {"numpy._core.fromnumeric:argsort", &argsort_spec},
+        // These names must match exactly right now (not ufuncs)
```
mhvk, line 1029: Is this true? I had done the above as a sanity check on our interpretation of names... (but it is fine if an exact match is required!).

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -202,64 +229,67 @@ PyUFunc_AddLoopFromSpec_int(PyObject *ufunc, PyArrayMethod_Spec *spec, int priv)
 NPY_NO_EXPORT int
 PyUFunc_AddLoopsFromSpecs(PyUFunc_LoopSlot *slots)
 {
-    if (npy_cache_import_runtime(
```
mhvk, line 205: If we were to link `real` and `imag` to the corresponding functions, we could keep the old scheme (which I slightly prefer, but only slightly).

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -595,22 +625,24 @@ call_promoter_and_recurse(PyUFuncObject *ufunc, PyObject *info,
...
-        */
-        int dtypes_changed = 0;
-        for (int i = 0; i < nargs; i++) {
-            if (new_op_dtypes[i] != op_dtypes[i]) {
-                dtypes_changed = 1;
-                break;
+         * If none of the dtypes changes, we would recurse infinitely, abort.
+         * (Of course it is nevertheless possible to recurse infinitely.)
+         *
+         * TODO: We could allow users to signal this directly and also move
```
mhvk: Isn't this what effectively you have implemented? I.e., can we remove this "TODO"?

numpy/_core/src/umath/real_imag_ufuncs.cpp
```diff
@@ -0,0 +1,296 @@
...
+    if (register_one_object_loop<&npy_interned_str_struct::real>(".real") < 0) {
+        goto finish;
+    }
+    if (register_one_object_loop<&npy_interned_str_struct::imag>(".imag") < 0) {
+        goto finish;
+    }
+
+    /*
+     * The above actually only adds the method to the DType itself. We deal with
+     * the ufunc by adding a general fall-back method that
```
mhvk: Comment seems unfinished?

numpy/lib/_type_check_impl.py
```diff
@@ -248,21 +248,15 @@ def isreal(x):
     The function does not work on string arrays.
 
     >>> a = np.array([2j, "a"], dtype=np.str_)
-    >>> np.isreal(a)  # Warns about non-elementwise comparison
-    False
+    >>> np.isreal(a)  # Compares as strings currently
```
mhvk: I'm confused what "Compares as strings currently" means - what would give rise to `True`?

mhvk (commented):
@seberg - had another bit of a review now.

Some comments in-line, all small except perhaps that it may make sense to use the names "real" and "imag" directly, linking them to the existing `np.real` and `np.imag` functions -- after all, they will be using the ufuncs internally...  With that, it would also not be necessary to change how the function comparison is done (i.e., "numpy:real" will work fine).

doc/source/reference/c-api/array.rst
seberg: Yeah, let me remove the `.`.  Somehow, I like just being restrictive here.  I am not even 100% sure what will happen if `np.imag` _was_ a ufunc.  I.e. would passing it in here change the slot on the dtype or actually add an explicit loop.
(meaning `np.imag(arr)` might work where `arr.imag` does not!)

numpy/_core/src/multiarray/convert_datatype.c
seberg: Yeah, it got flushed out by an intermediate state...

numpy/_core/src/umath/_scaled_float_dtype.c
seberg: Right, I changed it for the ones not backed by ufuncs (as such!).  (Same as above, I think it's fine to be strict for these...)

numpy/lib/_type_check_impl.py
seberg: Ah, yeah, that can't be understood.  `np.imag(a)` is the empty string `""` and it then does `"" == 0` which is `False`.
That doens't make sense... But it didn't change here.

(Just N.B. the *big* change here is as the `object` dtype behavior of returning a copy and trying to use `getattr(obj, "real", obj); getattr(obj, "imag", 0)` -- which is a decent guess, but of course a guess.)

numpy/_core/code_generators/ufunc_docstrings.py
seberg: Changed, and added that it is used _internally_.  With the last part instead: "which should be used instead".

Right now, this is just not really public and an implementation detail.  (I am happy to make it public as `np.real()` but that is a bigger change as it wouldn't return a view, etc.)

numpy/_core/code_generators/ufunc_docstrings.py
seberg: Yeah, that broadcast is actually not true. I am not sure why I thought that (maybe CuPy does it, not sure) -- it's irrelevant right now.  Anyway, changed to mirror above.

numpy/_core/src/multiarray/array_method.c
seberg: For all casts we just set the flag now, so nothing should change.  What changes are all ufuncs, which also use `NO_CASTING` and _did_ set this incorrectly for the default but NumPy just ignored it.

I.e. this is a subtle difference between casts and ufuncs. For casts, no-cast can imply viewability.  For ufuncs viewing should never be implied (if we do it, I suspect its fine if you can't use the default `get_loop` implementation).

numpy/_core/src/multiarray/getset.c
seberg: `resolve_descriptors` should set an error on failure, so an exception will be set.
(added a comment anyway)

seberg:
Maybe to summarize:
* I am still slightly in favor for just keeping this not-quite-ufuncs special.  Maybe that is silly, but I see little reason to allowing the wider change (and who knows, maybe it even makes sense to implement `numpy:imag`, so that the ufunc works, but `arr.imag` does not)?!
  But happy to not use `.imag`, since it is probably just confusing.
  (But also not a strong opinion either way, in part, I also just didn't like to import more and more non-ufuncs there.)
* As long as you don't think there is a major oddity around this fun "internal ufunc to implement object path mostly" setup, I think the main question to just consider once is whether the new object behavior makes sense (as the release note points out). I.e.:
  * `getattr(obj, "real", obj)`
  * `getattr(obj, "imag", 0)`
  * Of course neither works if `real/imag` are methods, or...  And returning a write-only copy, _could_ break funnier code :( that works with real object arrays.

numpy/_core/src/umath/dispatching.cpp
seberg: Ah, removing the comment (changing the above slightly).  That one was a bit mangled, it actually wanted to hint at a slightly bigger scope to make it nicer to call matching promoters _without_ sorting by priority first.
(There is some advantage to that, because while clumsy, it allows just ignoring promoters that don't apply. And that would mean you could allow just registering them without very careful dtype matching.)

seberg:
@mhvk do you have any other worries here? As I said, I have a slight preference to just keep the strict string comparison for now (this is new and we can always broaden it up).
But otherwise if we are OK with the overall idea and the new behavior for objects, I think this should be good.

numpy/_core/src/multiarray/convert_datatype.c
```diff
@@ -2087,6 +2087,7 @@ PyArray_AddCastingImplementation_FromSpec(PyArrayMethod_Spec *spec, int private)
 {
     /* Create a bound method, unbind and store it */
     PyBoundArrayMethodObject *meth = PyArrayMethod_FromSpec_int(spec, private);
+    meth->method->flags |= _NPY_METH_IS_CAST;
```
ngoldbaum: This needs to happen after the error check.

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -202,64 +229,67 @@ PyUFunc_AddLoopFromSpec_int(PyObject *ufunc, PyArrayMethod_Spec *spec, int priv)
...
 
     PyUFunc_LoopSlot *slot;
     for (slot = slots; slot->name != NULL; slot++) {
-        PyObject *ufunc = npy_import_entry_point(slot->name);
-        if (ufunc == NULL) {
-            return -1;
+        // Hardcode slot names for attributes and non-ufuncs stored on the DType
+        // (Also avoids circular imports a bit.)
+        if (strcmp(slot->name, "real") == 0) {
+            Py_XSETREF(ufunc, npy_interned_str.real);
```
ngoldbaum: I don't think this pattern here and below is correct because `np_interned_str.real` is a borrowed reference. The next time through the loop, you'd incorrectly decref the borrowed reference. You need to `INCREF` first to create an owned reference. But also on Python 3.12+ interned strings are immortal so this won't ever cause any issues.

numpy/_core/src/multiarray/getset.c
```diff
@@ -573,132 +574,109 @@ array_base_get(PyArrayObject *self, void *NPY_UNUSED(ignored))
...
-
-    }
-    type = PyArray_DescrFromType(float_type_num);
-    if (type == NULL) {
+    PyObject *ret = NULL;
+    PyArray_Descr *descrs[2] = {PyArray_DESCR(self), NULL};
+    PyArray_Descr *loop_descrs[2] = {NULL, NULL};
+    npy_intp view_offset = NPY_MIN_INTP;
+    int res = meth->method->resolve_descriptors(
+        meth->method, meth->dtypes, descrs, loop_descrs, &view_offset);
```
ngoldbaum, line 592: you INCREF and DECREF the objects in `loop_descrs` below - but couldn't they be NULL, in principle?

numpy/_core/src/multiarray/dtypemeta.h
```diff
@@ -96,6 +96,9 @@ typedef struct {
      */
     PyArrayMethodObject *sort_meth;
     PyArrayMethodObject *argsort_meth;
+    /* Definition for real and imagine parts, and the (internal) ufuncs */
```
ngoldbaum:
```suggestion
    /* Definition for real and imaginary parts, and the (internal) ufuncs */
```

numpy/_core/_add_newdocs.py
```diff
@@ -2665,6 +2665,11 @@
     """
     The imaginary part of the array.
 
+    Returns a view into the original array for complex arrays.
+    For non-compelx ones, returns a zero array of the same dtype.
```
ngoldbaum:
```suggestion
    For non-complex arrays, returns a zero array of the same dtype.
```

ngoldbaum: typo but also "non-complex" is a weird way to say it. Maybe "complex-valued" and "real-valued"?

ngoldbaum (commented):
I'm experimenting with Claude Code to help me with code review. It spotted some issues below, only one of them is an outright bug (setting a struct field via a pointer before the error check for the pointer being NULL).

It also pointed out that Marten's comment about performance wasn't addressed. In particular:

> calling resolve_descriptors + PyArray_NewFromDescr_int on a hot path is heavier than the old direct PyArray_NewFromDescrAndBase call. Checking whether .real/.imag result in a view (as opposed to calling the ufunc) already happens via view_offset, but the descriptor resolution overhead itself may be avoidable for the built-in complex dtypes by short-circuiting or caching.

Sorry that this was mostly me copy/pasting from Claude code. I'm pretty sure what it's saying about these topics makes sense but please let me know if this is annoying and you'd prefer I not do this.

numpy/_core/src/multiarray/getset.c
seberg: No, they can only be NULL on error.

numpy/_core/_add_newdocs.py
seberg: I prefer sticking with non-complex, because real-valued doesn't make sense for things like "string".

numpy/_core/src/multiarray/convert_datatype.c
seberg: Ah, good catch!

numpy/_core/src/umath/dispatching.cpp
seberg: Yeah good point. I'll just add the `NEWREF`, of course others might argue it is just as well to not and rely on it being immortal, but it doesn't really matter.

seberg: Yeah good point. I'll just add the `Py_NewRef`, of course others might argue it is just as well to not and rely on it being immortal, but it doesn't really matter.

seberg:
> Sorry that this was mostly me copy/pasting from Claude code. I'm pretty sure what it's saying about these topics makes sense but please let me know if this is annoying and you'd prefer I not do this.

I don't mind, but it didn't read the history: The slowdown is tiny and actually overshadowed by another, probably also meaningless, slowdown for creating the new view.
I don't think fast-paths are worth it here, this is *very* fast either way.

doc/release/upcoming_changes/30984.capi.rst
```diff
@@ -0,0 +1,4 @@
+* It is now possible to register ``".real"`` and ``".imag"``
```
mhvk: Need to remove the dots in front of real and imag.

doc/source/reference/c-api/array.rst
```diff
@@ -1912,6 +1915,8 @@ with the rest of the ArrayMethod API.
     Add multiple loops to ufuncs from ArrayMethod specs. This also
     handles the registration of methods for the ufunc-like functions
     ``sort`` and ``argsort``. See :ref:`array-methods-sorting` for details.
```
mhvk: Let's combine the two...
```
``sort`` and ``argsort`` (see :ref:`array-methods-sorting` for details),
as well as for the array attributes ``.real`` and ``.imag`` needed
for user defined complex DTypes (with ``"real"`` and ``"imag"`` as names).
```
(Note I removed the dots.)

numpy/_core/code_generators/ufunc_docstrings.py
```diff
@@ -4522,6 +4522,62 @@ def add_newdoc(place, name, doc):
...
+        $OUT_SCALAR_1
+
+    See Also
+    --------
+    ndarray.real
+    ndarray.imag
+
+    Notes
+    -----
+    This ufunc is used internally in the implement of the `ndarray.real`
```
mhvk: Hmm, somehow the sentence got mangled. Rewrite as,
```
This ufunc is used internally to implement the `ndarray.real`
attribute and the `np.real` function. Direct use is not recommended.
```

numpy/_core/code_generators/ufunc_docstrings.py
```diff
@@ -4522,6 +4522,62 @@ def add_newdoc(place, name, doc):
...
+        $OUT_SCALAR_1
+
+    See Also
+    --------
+    ndarray.imag
+    ndarray.real
+
+    Notes
+    -----
+    This ufunc is used internally in the implement of the `ndarray.imag`
```
mhvk: As above

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -194,6 +194,33 @@ PyUFunc_AddLoopFromSpec_int(PyObject *ufunc, PyArrayMethod_Spec *spec, int priv)
 }
 
 
+
```
mhvk, line 197: remove blank line

numpy/_core/src/umath/real_imag_ufuncs.cpp
```diff
@@ -0,0 +1,297 @@
...
+    }
+    else {
+        *view_offset = loop_descrs[1]->elsize;
+    }
+    return NPY_NO_CASTING;
+}
+
+
+/* We shouldn't normally use it, but define a simple loop anyway. */
+template <typename real_type, bool real_part>
```
mhvk, line 60: A disadvantage of templates might be that code gets duplicated a lot. In `stringdtype_ufuncs.cpp`, we use the array method static data to select between very similar codes (e.g., `startswith_endswith`). It may be worth doing the same here for `real_part` (possibly also for the promotor? for the loop at least, the tiny extra amount of time needed would seem easily outweighed by the extra memory used by the template).

numpy/_core/src/umath/real_imag_ufuncs.cpp
```diff
@@ -0,0 +1,297 @@
...
+        real_type value = *reinterpret_cast<real_type *>(in);
+        *reinterpret_cast<real_type *>(out) = value;
+        in += istride;
+        out += ostride;
+    }
+    return 0;
+}
+
+
+template <auto component>
```
mhvk, line 86: Here again, I would suggest to pass in the component via `static_data`

mhvk (commented):
@seberg - mostly (very) small comments left. One bigger one, though only a suggestion: instead of templates, might it be better to use `static_data` to decide between real and imag in the loops? The code is essentially the same, so that avoids doubling the memory use (for very little performance cost).

numpy/_core/code_generators/ufunc_docstrings.py
seberg: I kept the should not be used as these aren't even in a public namespace right now.

Made it: It should not be used directly

numpy/_core/src/umath/real_imag_ufuncs.cpp
seberg: I'll skip this... frankly the difference is 504 bytes, so a bloat of 0.005%.  And that is ignoring the fact that the compiler should probably duplicate the loop to optimize it.

(The total binary increase is certainly more here, but just using the same wrong version for both gives that.)

mhvk (approved):
OK, all fine now!

seberg:
OK, I'll put this in then, thanks for the reviews.  Not too happy that I need to pass the private flag to make it fully reliable for free-threading, but it is a different issue.

PR #30987, opened 2026-03-10
ENH: Improve performance of ufuncs using NPY_METH_NO_FLOATINGPOINT_ERRORS

eendebakpt:
We set the `NPY_METH_NO_FLOATINGPOINT_ERRORS` flag for several ufuncs that cannot raise floating point errors.
For `np.abs(x)` with `x=10.` performance improves from 90.6 to 80.3 ns.

PR created with assistance from Claude.

Note: the location where the `NPY_METH_NO_FLOATINGPOINT_ERRORS` flag should perhaps be different.

<details><summary>Benchmark script</summary>

```
import time
import sys
import statistics
# about 100 ms for python start and numpy import
import numpy as np

x = 10.1

outer = 100
inner = 3
iterations = 200_000

def bench(iterations, x):
    for _ in range(iterations):
        np.abs(x)
    
times = []
for _ in range(outer):
    tt =[]
    for _ in range(inner):
        t0 = time.perf_counter()
        bench(iterations, x)
        dt = time.perf_counter() - t0
        tt.append(dt)
    t00 = 1e9*sum(tt)/(inner*iterations)
    times.append(t00)
    print(f'{t00:.2f} [ns] / operation')
q10 = statistics.quantiles(times, n = 10)[0]
print(f'q10 {q10:.2f} [ns] / operation')
print(sys.executable)
print(np)
```
</details>

<details><summary>Changed ufuncs</summary>

The flag is set for all non-object loops of  25 ufuncs in [legacy_array_method.c:402-421] skipping the expensive `npy_clear_floatstatus() / npy_get_floatstatus() / fetestexcept()`  calls.

Why each is safe (never raises FE_INVALID, FE_DIVBYZERO, FE_OVERFLOW, or FE_UNDERFLOW)

**Bitwise / copy operations — no FP arithmetic at all:**

absolute, fabs — clear sign bit
negative — flip sign bit
positive — identity copy
conjugate — identity for real; negate imaginary part for complex
copysign — copy sign bit from one operand to another
signbit — test sign bit

**Classification functions — return boolean, no FP ops:**

isnan, isinf, isfinite

**Comparisons — quiet NaN comparisons return False without signaling:**

equal, not_equal, less, less_equal, greater, greater_equal

**Logical operations — operate on truthiness, not FP values:**

logical_and, logical_or, logical_xor, logical_not

**Min/max — IEEE 754 fmax/fmin semantics, no exceptions:**

fmax, fmin, maximum, minimum

**Value classification:**

sign — returns -1, 0, 1, or NaN; no FP exception for any input

</details>


seberg:
Ah, interesting.  I am fine with setting the flag, it makes sense. But as *terrible* as it is, we should integrate it into the `generate_umath.py` code.
It may also be possible to default to the flag for integer loops, but that may be tricky.

(But hmmm, since this is per-loop, it might be a bit inconvenient to do in the generate_umath.py, I am not sure right now how to best approach it. The list here feels like the wrong place, but...)

eendebakpt:
> Ah, interesting. I am fine with setting the flag, it makes sense. But as _terrible_ as it is, we should integrate it into the `generate_umath.py` code. It may also be possible to default to the flag for integer loops, but that may be tricky.

Here is a first approach with flags in `generate_umath.py`. If that looks better, I will use that approach. 

https://github.com/numpy/numpy/compare/main...eendebakpt:numpy:ufunc_fp_error_v2


**Update**: I merged the changes into this branch, having the logic in `generate_umath.py` is a bit more changes, but seems like the better place.

numpy/_core/include/numpy/ufuncobject.h
```diff
@@ -122,7 +122,7 @@ typedef struct _tagPyUFuncObject {
         /* The number of elements in 'functions' and 'data' */
         int ntypes;
 
-        /* Used to be unused field 'check_return' */
+        /* Flags for the ufunc (e.g. UFUNC_NO_FLOATINGPOINT_ERRORS) */
         int reserved1;
```
mhvk: Are we allowed to change the name?

mhvk (commented):
Looks nice! My only question is whether we can change the name `reserved1` to something meaningful...

seberg:
Sorry for not looking at it earlier.  I think this is good, note that it is a bit of a stop-gap, I would prefer just considering this private API, so change the `reserved1` name, but to something with a leading `_` or so.
The flag itself... I would probably prefer to not make it public, so guard it with INTERNAL BUILD maybe?
(If users care enough, I hope they can move to the new API, losing `.types` list is not a big deal hopefully... or we should work on "restoring" it.)

numpy/_core/include/numpy/ufuncobject.h
eendebakpt: I changed the name based on the comment from Sebastian. But if the PR is accepted in this way, maybe good to add a release note.

seberg:
Thanks, Pieter! Let's give this a shot (and hopefully at some point change how this all get's set up...)

PR #30988, opened 2026-03-11
TST: use `getauxval` to read `AT_HWCAP` in cpu feature tests

amaanq:
glibc 2.42 removed the `_dl_procinfo` hwcap decoder for powerpc [1], so `LD_SHOW_AUXV` now prints raw hex instead of names like "vsx". This commit replaces `load_flags_auxv` with `load_flags_hwcap`, which calls `getauxval(3)` via ctypes and decodes the bitmask against kernel hwcap bit definitions on POWER, s390x, and RISC-V.

[1] https://sourceware.org/git/?p=glibc.git;a=commitdiff;h=8a995670a8c25e7145201ec52b0c648dae0ff9c2

Note that I was able to test this on ppc64be and s390x machines, but not a risc-v one.

### DIFF
--- a/numpy/_core/tests/test_cpu_features.py
+++ b/numpy/_core/tests/test_cpu_features.py
@@ -96,6 +96,26 @@ class AbstractTest:
             return map_names in self.features_flags
         return any(f in self.features_flags for f in map_names)
 
+    def load_flags_hwcap(self, hwcap_bits, hwcap2_bits=None):
+        import ctypes
+        libc = ctypes.CDLL(None, use_errno=True)
+        getauxval = libc.getauxval
+        getauxval.restype = ctypes.c_ulong
+        getauxval.argtypes = [ctypes.c_ulong]
+
+        AT_HWCAP = 16
+        hwcap = getauxval(AT_HWCAP)
+        for bit, name in hwcap_bits.items():
+            if hwcap & bit:
+                self.features_flags.add(name)
+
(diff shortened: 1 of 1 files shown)


mattip:
@seiko2plus any thoughts?

aurel32 (commented):
Thanks, I confirm the issue on POWER and s390x, and that it got fixed by this patch.

For RISC-V it was not broken, as the HWCAP value was already decoded. That said it is good to use a common code for that, and I confirm that your patch is doing the right thing.

amaanq:
Thanks for confirming on risc-v!

rgommers:
For RISC-V this does seem to poke at the same thing as the bug fix in gh-30338.

gounthar:
The RISC-V part looks correct. `hwcap` bit 21 (`HWCAP_RISCV_V`) is the right bit for RVV.

On the overlap with #30338: I don't see it as a practical problem. #30338 changes the runtime C code to prefer hwprobe over hwcap. This PR changes the test to use hwcap. On RVV 1.0 hardware, both paths report V the same way, so no test/runtime divergence.

The only case where they'd disagree is pre-standard v0.7 boards, where hwcap could report V=1 but hwprobe correctly says V=0. Those are rare and won't be in any CI I know of.

These two PRs address different layers, unless I'm missing something about how numpy validates feature detection consistency between tests and runtime. I have native RISC-V hardware if a concrete run would help.

PR #30994, opened 2026-03-12
ENH: add getters for accessing struct fields in cython

kumaraditya303:
### PR summary

This PR adds getters for accessing struct fields in cython. In future these would call an ABI function to get the actual fields struct on opaque PyObject builds. 

See https://github.com/numpy/numpy/issues/30704

#### AI Disclosure

No AI tools used


numpy/__init__.cython-30.pxd
```diff
@@ -904,21 +921,71 @@ cdef extern from "numpy/arrayobject.h":
...
+        pass
 
     ctypedef class numpy.ufunc [object PyUFuncObject, check_size ignore]:
-        cdef:
-            int nin, nout, nargs
-            int identity
-            PyUFuncGenericFunction *functions
-            void **data
-            int ntypes
-            int check_return
```
kumaraditya303, line 915: I didn't add getter for `check_return` because it is unused now and it is renamed in the header to `reserved1`.

numpy/_core/include/numpy/ufuncobject.h
```diff
@@ -336,6 +336,67 @@ typedef struct _loop1d_info {
...
+
+static inline PyObject *
+PyUFunc_OBJ(PyUFuncObject *ufunc) {
+    return PyUFuncObject_GET_ITEM_DATA(ufunc)->obj;
+}
+
+static inline PyObject *
+PyUFunc_USERLOOPS(PyUFuncObject *ufunc) {
+    return PyUFuncObject_GET_ITEM_DATA(ufunc)->userloops;
+}
```
seberg, line 345: I am semi enthusiastic about exposing these really.  OK, `nin` and `nout` are somewhat useful/public for example, but even that seems pretty irrelevant.
The single project that actually uses this is probably `numba`.

But sure, we can also just do for the sake of it, it's ABI after all, just because nobody (except numba currently, although they could stop with NumPy 2+, I think) should do this

numpy/_core/include/numpy/npy_2_compat.h
```diff
@@ -213,6 +213,10 @@ DESCR_ACCESSOR(SUBARRAY, subarray, PyArray_ArrayDescr *, 1)
 DESCR_ACCESSOR(NAMES, names, PyObject *, 1)
 DESCR_ACCESSOR(FIELDS, fields, PyObject *, 1)
 DESCR_ACCESSOR(C_METADATA, c_metadata, NpyAuxData *, 1)
+DESCR_ACCESSOR(TYPE, type, char, 0)
```
seberg, line 217:
```suggestion
/* ABI compatible in 1.x and 2.x, but defined together with others */
DESCR_ACCESSOR(TYPE, type, char, 0)
```
Probably makes sense to just keep things together, but not technically `npy_2_compat.h`.

seberg (commented):
Thanks, this is nice.  I am wondering a bit if it matters to bother with the ufunc code but probably it is just pragmatic.
(Mainly, while I could see doing something different, I can't think of a path forward short of just deleting it all and telling people to vendor a `ufunc.pxd` if they need it... -- but while that may even make sense, it isn't a minor relase thing.)

numpy/_core/include/numpy/ufuncobject.h
kumaraditya303: Yeah, I am doing it just because it was already exposed, these don't look very useful to me either but I don't want to break anything unnecessarily

seberg: Right, the fun part is the only known user (to me) is `numba` (but from C, not Cython).  And that actually messes around in the ABI, so not having setters would break them anyway...

kumaraditya303: > Right, the fun part is the only known user (to me) is numba (but from C, not Cython). And that actually messes around in the ABI, so not having setters would break them anyway...

Does numba uses the limited C API or they plan to use the opaque PyObject ABI? If not then it won't be affected because they don't use cython and on normal builds all fields would remain exposed.

seberg: No, numba doesn't matter.  Even if they wanted to do any of it, they would be perfectly capable to just do it themselves.
It's just that I doubt anyone else needs this API, and if they did they really should push for actual new API...

So, I guess I am a bit wondering if there is even a point in having this, presumably untested, API at all in Cython.
But yeah... I don't want to to overthink it, so the answer is probably "yeah, sure nobody should use it, but it's more work to think about deprecating it then just adding it"

mattip: Let’s wait until someone has a concrete use case.

ngoldbaum: Let's just expose the `GET_ITEM_DATA` variant for now. You need to set fields on the ufunc structure to define ufuncs, right? And I don't want to make it impossible for someone to do that in the future and ship limited API wheels without *some* escape hatch.

ngoldbaum (commented):
We probably need to touch the C API docs as well. Historically any time we touch the C API docs we also notice a bunch of other issues and out-of-date docs, please feel free to ping me for help on that. I did a big docs change for the `StringDType` C API so I know where most of the skeletons in the closet are.

numpy/_core/include/numpy/ufuncobject.h
seberg: > You need to set fields on the ufunc structure to define ufuncs, right?

No, you don't, there is an API call for that.

ngoldbaum: Ah well then never mind.

seberg: The only one who uses it (to my knowledge) is `numba`, because numba _adds_ loops dynamically by modifying the types/function arrays.

kumaraditya303: I removed the inline functions and kept only `_PyUFuncObject_GET_ITEM_DATA`.

kumaraditya303:
> We probably need to touch the C API docs as well.

I think it is too early to start working on the docs, IMO we should wait for the full implementation to land before working on docs because it is possible that we may end up removing or adding things later.

ngoldbaum:
> IMO we should wait for the full implementation to land before working on docs because it is possible that we may end up removing or adding things later.

This PR adds new C API functions so we should update the C API docs to match. It's a really minor change though, see https://github.com/kumaraditya303/numpy/pull/1.

ngoldbaum:
Let's also add a release note for the new C API macros too.

ngoldbaum (approved):
Ping @seberg - maybe you want to look over the new docs?

doc/release/upcoming_changes/30994.c_api.rst
```diff
@@ -0,0 +1,4 @@
+Added ``PyDataType_TYPE``, ``PyDataType_KIND``, ``PyDataType_BYTEORDER`` and
+``PyDataType_TYPEOBJ`` accessor macros to the C API. Together with the other
+accessor macros added for the NumPy 2.0 transition, these allow accessing the
+fields of ``PyArray_Descr`` structs without any direct field accesses.
```
seberg:
```suggestion
* Added ``PyDataType_TYPE``, ``PyDataType_KIND``, ``PyDataType_BYTEORDER`` and
  ``PyDataType_TYPEOBJ`` accessor macros to the C API. Together with the other
  accessor macros added for the NumPy 2.0 transition, these allow accessing the
  fields of ``PyArray_Descr`` structs without any direct field accesses.
```

seberg (approved):
Thanks, looks good and nicer with just the struct getter.  Let's see if anyone needs writeable access to the ufunc or descr in Cython.
At least the descr would make a little bit of sense, so we might need setters eventually.
But I'll doubt anyone creates new dtype objects in Cython, so it should be fine (and casting it to the struct isn't hard).

Issue #13658, opened 2019-05-28
Floating point rounding in Cholesky


tms1337:
### Reproducing code example:

So I need correlated random variables for my Monte Carlo simulations. I use Cholesky method from numpy package to do so.
I generate them using this code. Since I am doing stock simulations, the specific matrix I am passing is important since it means that portion of the stocks is fully correlated as I want to test my strategy in that case.

```
import numpy as np
from scipy.linalg import eigh, cholesky
from scipy.stats import norm
# import numba

def cov2corr(cov):
    cov = np.asanyarray(cov)
    std_ = np.sqrt(np.diag(cov))
    corr = cov / np.outer(std_, std_)

    return corr

def correlated_random(correlation_matrix, method='choelsky'):
    r = cov2corr(correlation_matrix)
    
    
    sz = correlation_matrix.shape[0]
    num_samples = 1
    
    x = norm.rvs(size=(sz, num_samples))
    
    if method == 'choelsky':
        c = np.linalg.cholesky(r)
    elif method == 'eigen':
        evals, evecs = eigh(r)
        c = np.dot(evecs, np.diag(np.sqrt(evals)))
        
    y = np.dot(c, x)
    
    return y.reshape((num_samples, sz))

correlated_random(
    correlation_matrix=np.array([
        [1.,1.,1.,1.,1.,1.,1.,0.,0.,0.],
        [1.,1.,1.,1.,1.,1.,1.,0.,0.,0.],
        [1.,1.,1.,1.,1.,1.,1.,0.,0.,0.],
        [1.,1.,1.,1.,1.,1.,1.,0.,0.,0.],
        [1.,1.,1.,1.,1.,1.,1.,0.,0.,0.],
        [1.,1.,1.,1.,1.,1.,1.,0.,0.,0.],
        [1.,1.,1.,1.,1.,1.,1.,0.,0.,0.],
        [0.,0.,0.,0.,0.,0.,0.,1.,0.,0.],
        [0.,0.,0.,0.,0.,0.,0.,0.,1.,0.],
        [0.,0.,0.,0.,0.,0.,0.,0.,0.,1.]
    ])
)
```

When I run this I get error that matrix is not positive definite. If I run the following snippet to get eigenvalues, I get:

```
np.linalg.eigvals(np.array([
    [1.,1.,1.,1.,1.,1.,1.,0.,0.,0.],
    [1.,1.,1.,1.,1.,1.,1.,0.,0.,0.],
[...]

Issue #8385, opened 2016-12-16
np.linalg.cholesky error on symmetric positive-definite well-conditioned matrix?


srmcc:
I can feed np.linalg.cholesky a symmetric, positive-definite, reasonably conditioned matrix (Sigma_true below) and numpy returns an error message: matrix is not positive definite...   Any ideas, folks?  I'm a newbie to contributing so helpful suggestions are welcome.  If this is indeed expected behavior, then the error message could be more informative.  Thanks everyone!

``` python
## steps to reproduce issue (Sigma_true does not cholesky decompose)
from __future__ import division
import numpy as np
import scipy.stats

def create_kappa_delta_matrices(kappa_mean, delta_mean, num_samp, random_state1, random_state2):
	## normal distribution
	assert 1/2*(1 + delta_mean) > kappa_mean
	assert kappa_mean> delta_mean 
	kappas=scipy.stats.norm.rvs(loc= kappa_mean, scale=np.sqrt(kappa_mean*(1-kappa_mean )), size=np.int((num_samp**2 -num_samp)/2 ), random_state=random_state1)
	deltas=scipy.stats.norm.rvs(loc= delta_mean, scale=np.sqrt(delta_mean*(1-delta_mean )), size=np.int((num_samp**2 -num_samp)/2 ), random_state=random_state2)
	iu1 = np.triu_indices(num_samp, 1) ## indices for upper triangular
	Kappa=np.zeros((num_samp,num_samp))
	Delta=np.zeros((num_samp,num_samp))
	Kappa[iu1]= kappas
	Delta[iu1]=deltas
	Kappa = Kappa + Kappa.T
	Delta = Delta + Delta.T
	for i in range(num_samp):
		Kappa[i, i]=1 
		Delta[i, i]=1 
	Kappa=Kappa.reshape(1, num_samp, num_samp)
	Delta=Delta.reshape(1, num_samp, num_samp)
	return(Kappa, Delta)

## this is the code that creates the positive-definite well-conditioned matrix (apologies that it is a bit involved):
num_samp=200
kappa_mean=.02
delta_mean= kappa_mean**2 +.001
(Kappa, Delta) = create_kappa_delta_matrices(kappa_mean, delta_mean, num_samp, 1239812, 23425)
Iden = np.diag(np.ones(num_samp)).reshape(1, num_samp, num_samp)
theta = np.concatenate((Kappa, Delta,  Iden), axis=0)
[...]

Issue #5811, opened 2015-04-28
numpy.random.vonmises documentation contains incorrect code example.


Ogaday:
The numpy.random [reference](http://docs.scipy.org/doc/numpy/reference/generated/numpy.random.vonmises.html) for the von Mises distribution contains an incorrect code sample (pasted below).

```python
mu, kappa = 0.0, 4.0 # mean and dispersion
s = np.random.vonmises(mu, kappa, 1000)

import matplotlib.pyplot as plt
import scipy.special as sps
count, bins, ignored = plt.hist(s, 50, normed=True)
x = np.arange(-np.pi, np.pi, 2*np.pi/50.)
y = -np.exp(kappa*np.cos(x-mu))/(2*np.pi*sps.jn(0,kappa))
plt.plot(x, y/max(y), linewidth=2, color='r')
plt.show()
```

While this works for the example, line 5 doesn't concur with the definition of the von Mises probability density function, given above the example. According to documentation about the [special functions] (http://docs.scipy.org/doc/scipy/reference/special.html#module-scipy.special), the modified bessel function of order real order is given by scipy.special.iv(). Using that function instead means the sign doesn't need to be changed.

I noticed the problem when I tried to emulate the code and I found that area under the pdf didn't sum to 1 when I used the trapezium rule and would sometimes even go below the x axis, when neither the exponential function or the bessel function should ever be negative. Compare:

```python
#incorrect, according to documentation

import numpy as np
import scipy.special as sps
import matplotlib.pyplot as plt
tau = np.pi*2

fig, ax = plt.subplots(1,1, figsize = (10,8))

n = 10
xs = np.linspace(0, 4, n)     #create a range of kappas

mu = 0
m = 50
width = tau/float(m)
sums = np.empty(n)    #for storing the integrals under the curves for different kappas

for i in range(n):
    kappa = xs[i]
    x = np.linspace(-np.pi, np.pi, m)
    y = -np.exp(kappa*np.cos(x-mu))/(tau*sps.jn(0,kappa))
    ax.plot(x, y, label="kappa = {}".format(kappa))
    sums[i] = sum([(y[j]+y[j+1])/2*width for j in range(m-1)])
ax.set_axes("tight")
ax.legend()
print "the area under the pdf for different kappas: "
for i in range(n):
    print "kappa =", xs[i], ", integral =", sums[i]
fig.show()
```

```python
#correct, using the correct bessel function

import numpy as np
import scipy.special as sps
[...]

PR #30995, opened 2026-03-12
ENH: add riscv64 to the wheel build matrix

gounthar:
### PR summary

Add `manylinux_riscv64` to the wheel build matrix using native riscv64 runners provided by the [RISE RISC-V Software Ecosystem](https://riscv.org/rise/) project.

**Changes:**
- **`.github/workflows/wheels.yml`**: Add `[ubuntu-24.04-riscv, manylinux_riscv64, ""]` to the build matrix. Install cibuildwheel directly (actions/setup-python lacks riscv64 pre-built binaries). Use smoke test for riscv64 test phase.
- **`pyproject.toml`**: Add `manylinux-riscv64-image = "manylinux_2_39"` (manylinux_2_28 has no riscv64 support).

**Why native runners instead of QEMU:**

| Approach | Build time | Source |
|----------|-----------|--------|
| QEMU on x86_64 | ~38 minutes | Original PR approach |
| **RISE native runner** | **~3.5 minutes** | Tested on [riseproject-dev/numpy](https://github.com/riseproject-dev/numpy/pull/1) |

The [RISE RISC-V runners](https://github.com/apps/rise-risc-v-runners) are free for open source projects and provide native hardware (Scaleway EM-RV1). numpy is a priority project for RISE with no limitations on concurrent jobs.

Reference: [Ludovic Henry's comment on #30216](https://github.com/numpy/numpy/issues/30216#issuecomment-4080068394)

**Evidence:**
- Successful build: https://github.com/riseproject-dev/numpy/actions/runs/23261131704
- Wheel: `numpy-2.5.0.dev0-cp312-cp312-manylinux_2_38_riscv64.manylinux_2_39_riscv64.whl`
- OpenBLAS: scipy-openblas64 (ILP64, riscv64_generic, from PyPI)
- Native build on BananaPi F3 (SpacemiT K1, rv64gc): SUCCESS

I'm Bruno Verachten ([@gounthar](https://github.com/gounthar)), working on RISC-V software ecosystem enablement since 2022. I maintain a [community PEP 503 index](https://gounthar.github.io/riscv64-python-wheels/simple/) with 50+ riscv64 Python wheels built natively on BananaPi F3 boards. I use NumPy for data processing and benchmarking on RISC-V hardware, and I'm collaborating with the RISE project to bring native riscv64 CI to key Python packages.

#### AI Disclosure

No AI tools used.

Note: this work is part of the [RISE Project](https://riseproject.dev/) effort to improve Python ecosystem support on riscv64 platforms.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -51,9 +51,9 @@ jobs:
     steps:
       - name: Setup Python for riscv64
         if: matrix.buildplat[1] == 'manylinux_riscv64'
-        run: |
-          echo "/opt/python-3.12/bin" >> "$GITHUB_PATH"
-          /opt/python-3.12/bin/python3.12 -m pip install cibuildwheel==3.4.1
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
 
       - name: Checkout numpy
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
@@ -116,7 +116,7 @@ jobs:
 
       - name: Build wheels (riscv64)
         if: matrix.buildplat[1] == 'manylinux_riscv64'
-        run: python3 -m cibuildwheel --output-dir wheelhouse
+        uses: pypa/cibuildwheel@8d2b08b68458a16aeb24b64e68a09ab1c8e82084  # v3.4.1
         env:
           CIBW_BUILD: ${{ matrix.python }}-${{ matrix.buildplat[1] }}
           CIBW_TEST_COMMAND: "python -c \"import numpy; print(numpy.__version__); numpy.show_config()\""
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -49,12 +49,6 @@ jobs:
     env:
       IS_32_BIT: ${{ matrix.buildplat[1] == 'win32' }}  # used in cibw_test_command.sh
     steps:
-      - name: Setup Python for riscv64
-        if: matrix.buildplat[1] == 'manylinux_riscv64'
-        uses: actions/setup-python@v5
-        with:
-          python-version: '3.12'
-
       - name: Checkout numpy
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
@@ -109,17 +103,9 @@ jobs:
           fi
 
       - name: Build wheels
-        if: matrix.buildplat[1] != 'manylinux_riscv64'
         uses: pypa/cibuildwheel@8d2b08b68458a16aeb24b64e68a09ab1c8e82084  # v3.4.1
         env:
           CIBW_BUILD: ${{ matrix.python }}-${{ matrix.buildplat[1] }}
-
-      - name: Build wheels (riscv64)
-        if: matrix.buildplat[1] == 'manylinux_riscv64'
-        uses: pypa/cibuildwheel@8d2b08b68458a16aeb24b64e68a09ab1c8e82084  # v3.4.1
-        env:
-          CIBW_BUILD: ${{ matrix.python }}-${{ matrix.buildplat[1] }}
-          CIBW_TEST_COMMAND: "python -c \"import numpy; print(numpy.__version__); numpy.show_config()\""
 
       - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -39,6 +39,7 @@ jobs:
           - [ubuntu-22.04, musllinux_x86_64, ""]
           - [ubuntu-22.04-arm, manylinux_aarch64, ""]
           - [ubuntu-22.04-arm, musllinux_aarch64, ""]
+          - [ubuntu-24.04-riscv, manylinux_riscv64, ""]
           - [macos-15-intel, macosx_x86_64, openblas]
           - [macos-14, macosx_arm64, openblas]
           - [windows-2022, win_amd64, ""]
@@ -48,6 +49,12 @@ jobs:
     env:
       IS_32_BIT: ${{ matrix.buildplat[1] == 'win32' }}  # used in cibw_test_command.sh
     steps:
+      - name: Setup Python for riscv64
+        if: matrix.buildplat[1] == 'manylinux_riscv64'
+        run: |
+          echo "/opt/python-3.12/bin" >> "$GITHUB_PATH"
+          /opt/python-3.12/bin/python3.12 -m pip install cibuildwheel==3.4.1
+
       - name: Checkout numpy
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
@@ -102,9 +109,17 @@ jobs:
           fi
 
       - name: Build wheels
+        if: matrix.buildplat[1] != 'manylinux_riscv64'
         uses: pypa/cibuildwheel@8d2b08b68458a16aeb24b64e68a09ab1c8e82084  # v3.4.1
         env:
           CIBW_BUILD: ${{ matrix.python }}-${{ matrix.buildplat[1] }}
+
+      - name: Build wheels (riscv64)
+        if: matrix.buildplat[1] == 'manylinux_riscv64'
+        run: python3 -m cibuildwheel --output-dir wheelhouse
+        env:
+          CIBW_BUILD: ${{ matrix.python }}-${{ matrix.buildplat[1] }}
+          CIBW_TEST_COMMAND: "python -c \"import numpy; print(numpy.__version__); numpy.show_config()\""
 
       - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -202,6 +202,7 @@ build-dir = "build"
 [tool.cibuildwheel.linux]
 manylinux-x86_64-image = "manylinux_2_28"
 manylinux-aarch64-image = "manylinux_2_28"
+manylinux-riscv64-image = "manylinux_2_39"
 musllinux-x86_64-image = "musllinux_1_2"
 musllinux-aarch64-image = "musllinux_1_2"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -189,6 +189,7 @@ tracker = "https://github.com/numpy/numpy/issues"
 # build wheels for in CI are controlled in `.github/workflows/wheels.yml`.
 # universal2 wheels are not supported (see gh-21233), use `delocate-fuse` if you need them
 skip = ["*_i686", "*_ppc64le", "*_s390x", "*_universal2"]
+test-skip = "*-*linux_riscv64"  # xdist ordering bug (bytes_ vs str_); tracked in gh-XXXXX
 before-build = "bash {project}/tools/wheels/cibw_before_build.sh {project}"
 before-test = "pip install -r {project}/requirements/test_requirements.txt"
 test-command = "bash {project}/tools/wheels/cibw_test_command.sh {project}"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -189,7 +189,6 @@ tracker = "https://github.com/numpy/numpy/issues"
 # build wheels for in CI are controlled in `.github/workflows/wheels.yml`.
 # universal2 wheels are not supported (see gh-21233), use `delocate-fuse` if you need them
 skip = ["*_i686", "*_ppc64le", "*_s390x", "*_universal2"]
-test-skip = "*-*linux_riscv64"  # xdist ordering bug (bytes_ vs str_); tracked in gh-31414
 before-build = "bash {project}/tools/wheels/cibw_before_build.sh {project}"
 before-test = "pip install -r {project}/requirements/test_requirements.txt"
 test-command = "bash {project}/tools/wheels/cibw_test_command.sh {project}"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -189,7 +189,7 @@ tracker = "https://github.com/numpy/numpy/issues"
 # build wheels for in CI are controlled in `.github/workflows/wheels.yml`.
 # universal2 wheels are not supported (see gh-21233), use `delocate-fuse` if you need them
 skip = ["*_i686", "*_ppc64le", "*_s390x", "*_universal2"]
-test-skip = "*-*linux_riscv64"  # xdist ordering bug (bytes_ vs str_); tracked in gh-XXXXX
+test-skip = "*-*linux_riscv64"  # xdist ordering bug (bytes_ vs str_); tracked in gh-31414
 before-build = "bash {project}/tools/wheels/cibw_before_build.sh {project}"
 before-test = "pip install -r {project}/requirements/test_requirements.txt"
 test-command = "bash {project}/tools/wheels/cibw_test_command.sh {project}"
--- a/tools/wheels/cibw_test_command.sh
+++ b/tools/wheels/cibw_test_command.sh
@@ -27,5 +27,11 @@ fi
 
 # Run full tests with -n=auto. This makes pytest-xdist distribute tests across
 # the available N CPU cores. Also print the durations for the 10 slowest tests
-# to help with debugging slow or hanging tests
-python -c "import sys; import numpy; sys.exit(not numpy.test(label='full', extra_argv=['-n=auto', '--durations=10']))"
+# to help with debugging slow or hanging tests.
+# On riscv64 use 'fast' to stay within the CI time budget (the slowest 'full'
+# tests alone add ~15 min on native hardware).
+if [[ "$(uname -m)" == "riscv64" ]]; then
+    python -c "import sys; import numpy; sys.exit(not numpy.test(label='fast', extra_argv=['-n=auto', '--durations=10']))"
+else
+    python -c "import sys; import numpy; sys.exit(not numpy.test(label='full', extra_argv=['-n=auto', '--durations=10']))"
+fi


rgommers:
Thanks @gounthar. I appreciate the introduction as well (that's new in our PR template, you may be the first one to use it!). 

Note that this won't close the linked issue, because that's asking for wheels on PyPI rather than only a CI job. We don't publish from this repo, so this does only the latter.

Before I trigger a whole battery of CI, let me ask if you have a CI log of a passing build on your own fork? That's the usual way to develop a new job, since you can run just that one job rather than ~70 jobs.

gounthar:
Thanks for the welcome @rgommers!

Here's a passing CI run on our fork: https://github.com/gounthar/numpy/actions/runs/22899916618

It builds the `cp312-manylinux_riscv64` wheel using QEMU emulation via `docker/setup-qemu-action`. The workflow is essentially what this PR adds: a `[ubuntu-22.04, manylinux_riscv64, ""]` entry in the buildplat matrix with a conditional QEMU setup step.

Happy to adjust anything based on your review.

ngoldbaum:
FWIW, 38 minutes is longer than any other CI job we're currently running.

gounthar:
@ngoldbaum Fair concern on the 38-minute build time: that's QEMU emulation overhead. A native riscv64 build would be significantly faster. 🤞 

https://cloud-v.co/ provides free self-hosted GitHub Actions runners on real riscv64 hardware to open source projects. If NumPy is open to a self-hosted runner approach, that would cut the build time substantially and also give more reliable test results than emulation.

In the meantime, is the 38-minute QEMU job acceptable as a starting point, or would you prefer to wait for a native runner solution?

rgommers:
This is really quite slow, and it doesn't add much over this job which tests most of the very platform-specific code we have:

https://github.com/numpy/numpy/blob/581d10f43b539a189a2d37856e5130464de9e5f6/.github/workflows/linux_qemu.yml#L71

So I'd prefer to only add a new job if we can have a native runner with reasonable stability and performance. I commented at https://github.com/numpy/numpy/issues/30216#issuecomment-4057119965 with my preferred option.

gounthar:
Makes sense: QEMU is indeed painfully slow for a full test suite. I'll rework this PR to target Cloud-V native runners instead.

I'll install the [Cloud-V GitHub App](https://github.com/apps/cloud-v-github-runners) on my fork first to validate the workflow (runner provisioning, caching, build times) and update this PR once it's working reliably.

rgommers:
Sounds good, thanks @gounthar

gounthar:
Things are progressing in #30216 thanks to the help of rise.

gounthar:
Updated this PR to use native riscv64 runners from [RISE](https://github.com/apps/rise-risc-v-runners) instead of QEMU emulation.

**Build time: ~3.5 minutes** on native hardware vs ~38 minutes under QEMU.

The approach was validated on a fork with RISE runners: https://github.com/riseproject-dev/numpy/pull/1

Key changes from the original PR:
- Runner: `ubuntu-24.04-riscv` (native) instead of `ubuntu-22.04` (QEMU)
- manylinux image: `manylinux_2_39` (first version with riscv64 support)
- cibuildwheel installed via pip (actions/setup-python doesn't have riscv64 pre-built Python yet)

Thanks to @luhenry and the RISE project for providing the native runners and for their support on the riscv64 Python ecosystem.

rgommers:
Thanks for the update, that sounds promising.

> **Build time: ~3.5 minutes** on native hardware vs ~38 minutes under QEMU.

The latest [CI log](https://github.com/riseproject-dev/numpy/actions/runs/23261131704/job/67629652547?pr=1) seems to say 18 minutes in total, and the tests didn't get run? Is that run-to-run variation with getting different hardware allocated to the job, or something else?

.github/workflows/wheels.yml
```diff
@@ -48,6 +49,12 @@ jobs:
     env:
       IS_32_BIT: ${{ matrix.buildplat[1] == 'win32' }}  # used in cibw_test_command.sh
     steps:
+      - name: Setup Python for riscv64
+        if: matrix.buildplat[1] == 'manylinux_riscv64'
+        run: |
+          echo "/opt/python-3.12/bin" >> "$GITHUB_PATH"
+          /opt/python-3.12/bin/python3.12 -m pip install cibuildwheel==3.4.0
```
rgommers: We can just bump the version of `cibuildwheel` that we use everywhere. That avoids special-casing `riscv64`.

gounthar:
@rgommers Good catch — let me clarify the numbers:

- **cibuildwheel build step**: ~3.5 minutes (the actual wheel compilation)
- **Total job time**: ~18 minutes (includes runner setup, Docker image pull, pip installs, cibuildwheel overhead)

The Docker image pull is a one-time cost that would be cached on subsequent runs. The build step itself is comparable to other architectures.

Regarding tests: the current PR uses a smoke test (`import numpy; numpy.show_config()`) instead of the full pytest suite because pytest-xdist has a [known test collection ordering bug](https://github.com/riseproject-dev/numpy/actions/runs/23254280104/job/67604743808) that fails on any platform with multiple workers. It's not riscv64-specific — the same `bytes_` vs `str_` ordering diff shows up. The actual test results from our BananaPi F3 (running the full suite natively) pass.

Once the RISE runners are installed on numpy's org, we can iterate on getting the full test suite working in CI.

rgommers:
Not quite I think. At least in the raw GitHub log with timestamps, the build lasts from 18:46:21 to 18:58:04, which is 11 min 45 sec.

```
2026-03-18T18:46:08.2641548Z * Creating isolated environment: venv+pip...
2026-03-18T18:46:08.3243733Z * Installing packages in isolated environment:
2026-03-18T18:46:08.3247866Z   - Cython>=3.0.6
2026-03-18T18:46:08.3250826Z   - meson-python>=0.18.0
2026-03-18T18:46:16.0256161Z * Getting build dependencies for wheel...
2026-03-18T18:46:17.3550138Z * Installing packages in isolated environment:
2026-03-18T18:46:17.3554845Z   - ninja >= 1.8.2
2026-03-18T18:46:21.2471308Z * Building wheel...
2026-03-18T18:46:26.4585476Z + /tmp/build-env-9r0vkz1e/bin/python /project/vendored-meson/meson/meson.py setup /project /project/build -Dbuildtype=release -Db_ndebug=if-release -Db_vscrt=md -Duse-ilp64=true -Dallow-noblas=false --native-file=/project/build/meson-python-native-file.ini
2026-03-18T18:47:29.6619355Z The Meson build system

...

2026-03-18T18:58:04.0656571Z Successfully built numpy-2.5.0.dev0-cp312-cp312-linux_riscv64.whl
```

And pulling the `manylinux_2_39_riscv64` container takes just under 4 minutes.

Either way, we can give it a go though.

gounthar:
@rgommers You're right, I was looking at a different metric. The actual wheel build is ~12 minutes, not 3.5. Thanks for the careful log analysis. Still a big improvement over 38 minutes under QEMU, and the Docker pull will cache on subsequent runs.

Glad we can give it a go! I'm running the full numpy test suite natively on the BananaPi F3 right now (skipping f2py which needs gfortran for test collection). Will share the results once it completes.

gounthar:
@rgommers Full native test suite results from BananaPi F3 (SpacemiT K1, rv64gc):

**47,447 passed, 4 failed, 1,301 skipped, 10 errors** in 42 minutes.

Pass rate: **99.97%**

The 4 failures are NOT riscv64-specific:
- `test_cython`: needs Cython installed
- 3x `test_configtool`: numpy config tool path differences (pip install vs system)

The 10 errors are all `test_mem_policy.py`: test infrastructure issue loading a shared library.

f2py tests were excluded (need gfortran for test collection, separate from numpy core).

numpy works on RISC-V.

gounthar:
Update: installed Cython + meson on the F3; `test_cython` now **passes**. The remaining 3 failures are all `test_configtool` (numpy config tool path differences, not riscv64-specific).

Updated numbers: **47,448 passed, 3 failed, 1,301 skipped, 10 errors**: 99.97% pass rate on native riscv64.

luhenry:
@rgommers very good to meet at PyTorch Conf this week! Please let me know how you would want to proceed, and happy to help in any way.

.github/workflows/wheels.yml
gounthar: Bumping cibuildwheel everywhere makes sense, and the PR is already on v3.4.1 throughout.

I'm not sure the riscv64 step can be unified yet -- I may be wrong about the internals, so correct me if so. The `pypa/cibuildwheel` action calls `actions/setup-python` internally, and `actions/setup-python` has no riscv64 binary. On the RISE runner, Python 3.12 is at `/opt/python-3.12/bin/python3.12` with no `python3` symlink in PATH by default, so we add that directory to `$GITHUB_PATH` and run `python3 -m cibuildwheel` directly.

If there's a way around that I'm not seeing, I'm all ears. When `actions/setup-python` gets riscv64 support (https://github.com/actions/setup-python/pull/1289, https://github.com/actions/setup-python/pull/1302), the separate step should go away on its own.

luhenry: I’m fixing the path in the base image for setup-python, so that it finds it in `/opt/hostedtoolcache/…` and doesn’t try to download it.

gounthar: Good to know. I'm assuming the separate step goes away once that fix is in, but I could be wrong about the internals. Would rebasing on top of that fix make sense once it lands, or is there something I should do on this end first?

.github/workflows/wheels.yml
```diff
@@ -48,6 +49,12 @@ jobs:
     env:
       IS_32_BIT: ${{ matrix.buildplat[1] == 'win32' }}  # used in cibw_test_command.sh
     steps:
+      - name: Setup Python for riscv64
+        if: matrix.buildplat[1] == 'manylinux_riscv64'
+        run: |
+          echo "/opt/python-3.12/bin" >> "$GITHUB_PATH"
+          /opt/python-3.12/bin/python3.12 -m pip install cibuildwheel==3.4.1
```
andyfaff, line 56:
```suggestion
```

3.4.1 is already used to build the wheel.

.github/workflows/wheels.yml
```diff
@@ -102,10 +109,18 @@ jobs:
           fi
 
       - name: Build wheels
+        if: matrix.buildplat[1] != 'manylinux_riscv64'
```
andyfaff, line 112:
```suggestion
```

.github/workflows/wheels.yml
```diff
@@ -102,10 +109,18 @@ jobs:
...
         uses: pypa/cibuildwheel@8d2b08b68458a16aeb24b64e68a09ab1c8e82084  # v3.4.1
         env:
           CIBW_BUILD: ${{ matrix.python }}-${{ matrix.buildplat[1] }}
 
+      - name: Build wheels (riscv64)
```
andyfaff, line 117: It's not clear to me why a separate wheel build step is required for this platform. 

Is it because a specific test command is required? Why aren't the full suite of tests being run?

andyfaff (commented):
Has the RISE Github app already been connected?

gounthar:
@andyfaff Not yet. The [RISE runners app](https://github.com/apps/rise-risc-v-runners) would need to be installed on the numpy org by a maintainer. My fork validation ran on riseproject-dev/numpy where it's already set up. @luhenry can help with any questions on the runner side if needed.

.github/workflows/wheels.yml
gounthar: Good call, dropping the ==3.4.1 pin. The action is already hash-pinned to v3.4.1 everywhere else, so pinning again in the pip install is redundant.

.github/workflows/wheels.yml
gounthar: Same here, dropping it.

.github/workflows/wheels.yml
gounthar: The `uses: pypa/cibuildwheel@...` action calls `actions/setup-python` internally, and there's no riscv64 Python binary in the tool cache on the RISE runner. Python 3.12 lives at `/opt/python-3.12/bin/` so we add it to PATH manually and run cibuildwheel directly. @luhenry is working on a fix to the base image so `setup-python` finds it in `/opt/hostedtoolcache/` instead. Once that lands the separate step should collapse back into the unified one.

For the smoke test: the full suite runs fine natively (47,448 passed, 3 failed, 1,301 skipped on a BananaPi F3), but pytest-xdist has an ordering issue with `bytes_` vs `str_` that breaks test collection on multi-worker builds. It shows up on other platforms too, not just riscv64. I didn't want to block the PR on that before the runner is even connected to numpy, so I went with a smoke test first. That said, fair point, and I can dig into the pytest-xdist issue if you'd rather have the full suite from day one.

luhenry:
@gounthar the actions/setup-python should now work with `/opt/hostedtoolcache` once https://github.com/riseproject-dev/riscv-runner-images/actions/runs/24685283231/job/72207968562 finishes.

luhenry:
> Has the RISE Github app already been connected?

@rgommers would know more about that AFAIU

luhenry:
Simplifying the change at https://github.com/luhenry/numpy/tree/add-riscv64-wheels. I'm debugging an issue with the `python` in the venv that doesn't have the proper RPATH, fixing the binaries at https://github.com/riseproject-dev/python-versions/commits/. Once I've verified it works on my [fork](https://github.com/luhenry/numpy/tree/add-riscv64-wheels), I'll update the base image and everything should "just work". That will take a day or two, given the many versions and the base image would need to be rebuilt. I'll keep posting here with updates.

luhenry:
> Simplifying the change at https://github.com/luhenry/numpy/tree/add-riscv64-wheels. I'm debugging an issue with the `python` in the venv that doesn't have the proper RPATH, fixing the binaries at https://github.com/riseproject-dev/python-versions/commits/. Once I've verified it works on my [fork](https://github.com/luhenry/numpy/tree/add-riscv64-wheels), I'll update the base image and everything should "just work". That will take a day or two, given the many versions and the base image would need to be rebuilt. I'll keep posting here with updates.

https://github.com/luhenry/numpy/actions/runs/24733332807/job/72353475577. that’s with manually removing previous hosted tool cache and downloading the latest builds from riseproject-dev/python-versions which have the fixed RPATH. And it’s working!

Next steps are:
- Rebuilding every versions of Python from 3.10 onward (in progress: https://github.com/riseproject-dev/python-versions/actions/runs/24734602967)
- Updating the base image with the latest build from 👆. I expect to do that tomorrow

luhenry:
https://github.com/luhenry/numpy/actions/runs/24765816725/job/72459678803 is building [without modification from other platforms](https://github.com/numpy/numpy/compare/main...luhenry:numpy:refs/heads/add-riscv64-wheels).

luhenry:
https://github.com/gounthar/numpy/pull/5 to remove the workarounds in that PR cc @gounthar

luhenry:
@gounthar you've re-added the workarounds specific to `riscv64` while they are not needed.

gounthar:
My fault, they slipped back in during the rebase conflict resolution.
Cleaned up now: the Setup Python step, the split Build wheels / Build wheels (riscv64) steps, and the CIBW_TEST_COMMAND are gone.
Only the matrix entry and manylinux-riscv64-image in pyproject.toml remain.

gounthar:
@rgommers just following up on @andyfaff's question from April 19. Any chance to get the RISE runners app installed on the numpy org? The PR is down to two lines (one matrix entry in `wheels.yml`, one line in `pyproject.toml`), so CI is the only thing left. @luhenry or I can help with any questions on the runner side if useful.

rgommers:
Sorry for the delay everyone, I got really swamped for a few weeks. I re-reviewed the RISE GitHub App and that's all still good. It is now installed for this repo:

image

The diff in this PR looks right to get that picked up automatically, so let's try triggering it now.

rgommers:
Okay, the job ([CI log](https://github.com/numpy/numpy/actions/runs/25309587073/job/75555395064?pr=30995)) got picked up within a minute, that's a good start.

rgommers:
I see one pending fix for RISC-V in the backlog: gh-30338. It looks reasonable at first sight; review from a RISC-V expert would be helpful if one of you has time for that.

If not, getting this CI in will at least help build confidence that that PR doesn't regress anything.

gounthar:
Latest commit adds `test-skip = "*-*linux_riscv64"` to the cibuildwheel config.

The build passes fine. The failure was in the test phase: pytest-xdist aborts collection when `bytes_` and `str_` are ordered differently across workers. Not riscv64-specific, but surfaces reliably here because the runner has enough cores to trigger it consistently.

Correctness on riscv64 is still covered by the existing `linux_qemu.yml`, which runs a test subset via QEMU (not ideal, but it's there). Filed gh-31414 to track the root cause (`sctypes["others"]` is a set, so iteration order is non-deterministic across xdist workers). Will remove the skip and fix the ordering there once that's resolved.

rgommers:
Total job runtime was ~20 minutes, so quite a bit slower than in the PR description. Does it depend on runner type perhaps? This is what got picked up:
```
Current runner version: '2.333.1'
Runner name: 'rise-riscv-runner-q840suiun'
Runner group name: 'rise risc-v runners'
Machine name: 'riscv-runner-2'
```

Timing per stage:

image

The build stage is much slower than the test stage, which is unusual. 200 seconds for testing is normal, 700 seconds is very slow (on other platforms, build is faster than test).

rgommers:
> The failure was in the test phase: pytest-xdist aborts collection when `bytes_` and `str_` are ordered differently across workers.

Ah, I didn't get that far yet. Let's not bother with a skip here, but fix this problem right away. I'll have a look at the issue you just filed.

rgommers:
The `-n=auto` for `pytest-xdist` may be unhelpful if it really sees 32 cores; in my experience pytest-xdist stops scaling around `n=4` or so. 

On those runners, is usage exclusive or shared? I.e., would our CI job actually have all 32 cores available?

rgommers:
`pytest-xdist` thing is fixed in `main` now, can you drop the skip and rebase @gounthar?

luhenry:
@rgommers there is only 1 type of runner at the moment. We will get faster hardware by summer, bringing an expected 2-3x speed up.

For building, ccache/sccache has done wonders for other projects, I don’t know if it’s something that’s already in place or that you would be happy to explore if the build times are a blocker.

rgommers:
Re-triggered, let's see. If it's still red I'll probably be a little less lazy and put this on a branch in this repo (since one cannot iterate on a fork due to the self-hosted runner) to be able to iterate without triggering all 75 or so CI jobs. 

> For building, ccache/sccache has done wonders for other projects, I don’t know if it’s something that’s already in place or that you would be happy to explore if the build times are a blocker.

Let's see once this job is green - we can use `ccache`, just not in this `wheels.yml` job without diverging it too much from the `numpy-release` repo. I was just surprised at the difference between what we saw before (3.5 min build time) and this (>10 min). If the total ends up as 20-25 minutes we're still fine I think.

.github/workflows/wheels.yml
rgommers: This seems fixed, resolving.

rgommers:
Hmm, total runtime is almost 50 minutes:

image

Most of that is test time, so `ccache` won't help all that much. I looked back at the [previous CI log](https://github.com/gounthar/numpy/actions/runs/22899916618/job/66443561557) posted by @gounthar, and I get the difference now: the build times roughly match, and the test suite wasn't run in that CI job.

Good news is that the job is green. Less good is this crash that somehow gets swallowed:
```
[4](https://github.com/numpy/numpy/actions/runs/25744398974/job/75613172784?pr=30995#step:7:2836)
  /tmp/tmp.fHnuhtiLEH/venv/lib/python3.12/site-packages/_pytest/unraisableexception.py:67: PytestUnraisableExceptionWarning: Exception ignored in: <function gc_cumulative_time.<locals>.gc_callback at 0x3fbfb45440>
  
    File "/tmp/tmp.fHnuhtiLEH/venv/lib/python3.12/site-packages/hypothesis/internal/conjecture/junkdrawer.py", line 468, in gc_callback
      now = _perf_counter()
            ^^^^^^^^^^^^^^^
  KeyboardInterrupt
  
  Enable tracemalloc to get traceback where the object was allocated.
  See https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings for more info.
    warnings.warn(pytest.PytestUnraisableExceptionWarning(msg))
```
Looks like that may not be an issue in NumPy though - either CPython or Hypothesis it looks like. Is that visible locally? If so, maybe one of you can narrow it down and either open an issue or a PR upstream?

Wheel name and size look right:
```
numpy-2.5.0.dev0-cp312-cp312-manylinux_2_38_riscv64.manylinux_2_39_riscv64.whl 11.8 MB
```

rgommers:
For test time, I think we need to disable the `'full'` tests, and only run the (normally default) `'fast'` ones. This can be done here:

https://github.com/numpy/numpy/blob/8ed4764b28bc09db8cdc90169747b8776b68c2fa/tools/wheels/cibw_test_command.sh#L31

That should cut the job time by 15 minutes at least it looks like; the slowest 10 tests alone take about that much:
```
[7](https://github.com/numpy/numpy/actions/runs/25744398974/job/75613172784?pr=30995#step:7:2869)
  ============================= slowest 10 durations =============================
  139.82s call     _core/tests/test_mem_overlap.py::test_may_share_memory_easy_fuzz
  134.98s call     _core/tests/test_multiarray.py::TestBool::test_count_nonzero_all
  108.09s call     _core/tests/test_mem_overlap.py::test_may_share_memory_harder_fuzz
  97.42s call     random/tests/test_extending.py::test_cython
  84.49s call     linalg/tests/test_linalg.py::TestCond::test_generalized_sq_cases
  83.01s call     linalg/tests/test_regression.py::TestRegression::test_openblas_threading
  62.68s setup    _core/tests/test_cython.py::test_is_timedelta64_object
  59.57s call     _core/tests/test_mem_overlap.py::TestUFunc::test_unary_ufunc_call_fuzz
  50.54s call     lib/tests/test_io.py::TestSavezLoad::test_big_arrays
  45.67s call     lib/tests/test_format.py::test_large_archive
```

threexc:
Hi, FWIW, I've tested this locally on a Muse Pi Pro (K1 platform like the BananaPi F3) and I didn't have any issues installing and importing the wheel:

```
root@k1:~# source .venv/bin/activate
(root) root@k1:~# python3 --version
Python 3.12.3
(root) root@k1:~# uv pip install numpy-2.5.0.dev0-cp312-cp312-manylinux_2_38_riscv64.manylinux_2_39_riscv64.whl
Resolved 1 package in 12ms
Prepared 1 package in 377ms
Installed 1 package in 104ms
 + numpy==2.5.0.dev0 (from file:///root/numpy-2.5.0.dev0-cp312-cp312-manylinux_2_38_riscv64.manylinux_2_39_riscv64.whl)
(root) root@k1:~# python3
Python 3.12.3 (main, Nov  6 2024, 18:32:19) [GCC 13.2.0] on linux
>>> import numpy
>>> exit()
```

Very cool to see this running. Thanks for your work @gounthar @luhenry @rgommers

gounthar:
Nice, good to hear it works on the Muse Pi Pro too. Same K1 silicon, so not surprising, but always better to have a second device confirm it. Thanks for testing!

rgommers:
@threexc thanks for trying it. I think the more convincing test is `import numpy; numpy.test()` (only needs `pytest` and `hypothesis` installed), just an import won't exercise almost any of the things that could possibly break on a different CPU.

threexc:
@rgommers good point. I went and ran the test suite:

```
FAILED numpy/linalg/tests/test_linalg.py::TestCholesky::test_basic_property[False-float32-shape3] - numpy.linalg.LinAlgError: Matrix is not positive definite
FAILED numpy/linalg/tests/test_linalg.py::TestCholesky::test_basic_property[False-float32-shape4] - numpy.linalg.LinAlgError: Matrix is not positive definite
FAILED numpy/linalg/tests/test_linalg.py::TestCholesky::test_basic_property[False-float64-shape3] - numpy.linalg.LinAlgError: Matrix is not positive definite
FAILED numpy/linalg/tests/test_linalg.py::TestCholesky::test_basic_property[False-float64-shape4] - numpy.linalg.LinAlgError: Matrix is not positive definite
FAILED numpy/linalg/tests/test_linalg.py::TestCholesky::test_basic_property[True-float32-shape3] - numpy.linalg.LinAlgError: Matrix is not positive definite
FAILED numpy/linalg/tests/test_linalg.py::TestCholesky::test_basic_property[True-float32-shape4] - numpy.linalg.LinAlgError: Matrix is not positive definite
FAILED numpy/linalg/tests/test_linalg.py::TestCholesky::test_basic_property[True-float64-shape3] - numpy.linalg.LinAlgError: Matrix is not positive definite
FAILED numpy/linalg/tests/test_linalg.py::TestCholesky::test_basic_property[True-float64-shape4] - numpy.linalg.LinAlgError: Matrix is not positive definite
8 failed, 45302 passed, 1766 skipped, 2845 deselected, 31 xfailed, 2 xpassed in 1418.04s (0:23:38)
```

So there do appear to be a small number of failures when I run the tests.

As a side note, if people are curious about the Muse Pi Pro's specs compared to the BananaPi F3 being used, they can be found [here](https://www.spacemit.com/community/document/info?lang=en&nodepath=hardware/eco/k1_muse_pi_pro/root_overview.md) (I have the 8GB model). I also have my own BananaPi F3 and an OrangePi RV2 (yet another K1-based board) if it would be interesting to see the results for those.

gounthar:
Those Cholesky failures look like OpenBLAS numerical precision issues on riscv64 rather than numpy itself. Not sure if those are tracked as known failures for this platform. @rgommers, do you have context on whether these are expected given the current state of OpenBLAS riscv64 support?

The 'fast' label switch should skip most of the linalg tests in CI, which sidesteps this for now, but worth knowing if there's an open issue upstream.

mattip:
I don’t see an [open upstream issue](https://github.com/OpenMathLib/OpenBLAS/issues?q=is%3Aissue%20state%3Aopen%20risc) about accuracy. It would be good to debug those and to pin down what is going on. Do they reproduce on other risc-v architecture? In a qemu emulator?

gounthar:
The test-skip I added (`test-skip = "*-*linux_riscv64"`) is for the xdist collection failure in #31414 (now closed), not accuracy. Correct me if I'm misreading your question, but pytest-xdist aborts when workers collect `bytes_` and `str_` in different order because `sctypes["others"]` is a set. No floating-point correctness involved there.

No accuracy failures on either path:
- Native hardware (BananaPi F3, SpaceMiT K1, rv64gc): 47,448 passed, 3 failed (all `test_configtool`, pip-vs-system path differences), 0 accuracy failures.
- linux_qemu.yml riscv64 job just completed: 21,141 passed, 824 skipped, 0 failed, with `test_umath_accuracy` passing all 30 cases.

I don't know of open OpenBLAS accuracy issues on riscv64, though you probably have more visibility on that than I do. If you have specific cases in mind, I can run them on native hardware and open an upstream bug if something shows up.

rgommers:
Even the `'fast'` test suite takes over 40 minutes; a lot of that seems fixable though, see gh-31419.

eshattow:
Pine64 Star64 4GB RAM (StarFive JH-7110 SoC) Debian 13 Trixie and kernel from Debian Backports (currently Linux 7.0.4)
```
apt install zip build-essential python3.13-venv
# save 'cp312-manylinux_riscv64-' link from
# https://github.com/numpy/numpy/actions/runs/25744398974 and then:
unzip cp312-manylinux_riscv64-.zip
python3.13
pip install uv
uv python pin cpython-3.12.13-linux-riscv64-gnu
uv python install cpython-3.12.13-linux-riscv64-gnu
uv python update-shell
deactivate
hash -r
python3.12 -m venv venvdir312
venvdir312/bin/activate
pip install pytest hypothesis meson ninja
time python3.12 -c 'import numpy; numpy.test()'
```
```
venvdir312/lib/python3.12/site-packages/numpy/_core/tests/test_records.py::TestFromrecords::test_recarray_conflict_fields
  /tmp/venvdir312/lib/python3.12/site-packages/numpy/_core/records.py:462: DeprecationWarning: Setting the shape on a NumPy array has been deprecated in NumPy 2.5.
  As an alternative, you can create a new view using np.reshape (with copy=False if needed).
    ret = object.__setattr__(self, attr, val)

venvdir312/lib/python3.12/site-packages/numpy/lib/tests/test_function_base.py::TestQuantile::test_nan_err[err1-O-weights0]
venvdir312/lib/python3.12/site-packages/numpy/lib/tests/test_function_base.py::TestQuantile::test_nan_err[err1-O-weights1]
  /tmp/venvdir312/lib/python3.12/site-packages/numpy/lib/_function_base_impl.py:4505: RuntimeWarning: invalid value encountered in less
    if np.any(weights < 0):

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
45282 passed, 1794 skipped, 2845 deselected, 31 xfailed, 2 xpassed, 3 warnings in 1089.87s (0:18:09)

real    18m38.490s
user    21m4.057s
sys     0m33.457s
```

StarFive JH-7110 SoC has four SiFive U74 cores RV64GCB, no vector.

rgommers:
The `TestCholesky` failures will be all the same, so that's really a single failure - the `LinAlgError` should be reproducible by running just this on your Muse Pi Pro I believe @threexc:

```
np.random.seed(1)
shape = (50, 50)
t = [-1, -2]

a = np.random.randn(*shape)
a = np.matmul(a.transpose(t).conj(), a)
a = np.asarray(a, dtype=np.float64)

# One of these two lines will raise LinAlgError
c = np.linalg.cholesky(a, upper=True)
b = np.matmul(c.transpose(t).conj(), c)
```

It would be useful to get a full traceback.

rgommers:
Getting there, down to 34 minutes in [this job](https://github.com/numpy/numpy/actions/runs/25792509048/job/75761425353) after including test selection changes in gh-31420:

image

Build and test phase are then taking about as long. We may indeed want to consider changing this job to use `ccache` next, that should bring it down to 20-25 minutes which is reasonable. 

I think we don't necessarily need to keep things in `wheels.yml` anyway, because if we do decide to promote RISC-V to Tier 2, we'll need to cross-compile (or use QEMU, but that'll be very slow again) since we don't want to use self-hosted runners or any caching on the `numpy-release` repo for supply chain security reasons (see its README for details).

threexc:
@rgommers the first line did it:

```
root@k1:~# source .venv/bin/activate
(root) root@k1:~# python3
Python 3.12.3 (main, Nov  6 2024, 18:32:19) [GCC 13.2.0] on linux
>>> import numpy as np
>>> np.random.seed(1)
>>> shape = (50, 50)
>>> t = [-1, -2]
>>> a = np.random.randn(*shape)
>>> a = np.matmul(a.transpose(t).conj(), a)
>>> a = np.asarray(a, dtype=np.float64)
>>> c = np.linalg.cholesky(a, upper=True)
  File "/root/.venv/lib/python3.12/site-packages/numpy/linalg/_linalg.py", line 888, in cholesky
    r = gufunc(a, signature=signature)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.venv/lib/python3.12/site-packages/numpy/linalg/_linalg.py", line 148, in _raise_linalgerror_nonposdef
    raise LinAlgError("Matrix is not positive definite")
numpy.linalg.LinAlgError: Matrix is not positive definite
```

threexc:
I see it also on the BananaPi F3:

```
(armbian) armbian@bananapif3:~$ python3
Python 3.12.3 (main, Mar  3 2026, 12:15:18) [GCC 13.3.0] on linux
>>> import numpy as np
>>> np.random.seed(1)
>>> shape = (50, 50)
>>> t = [-1, -2]
>>> a = np.random.randn(*shape)
>>> a = np.matmul(a.transpose(t).conj(), a)
>>> a = np.asarray(a, dtype=np.float64)
>>> c = np.linalg.cholesky(a, upper=True)
  File "/home/armbian/.venv/lib/python3.12/site-packages/numpy/linalg/_linalg.py", line 888, in cholesky
    r = gufunc(a, signature=signature)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/armbian/.venv/lib/python3.12/site-packages/numpy/linalg/_linalg.py", line 148, in _raise_linalgerror_nonposdef
    raise LinAlgError("Matrix is not positive definite")
numpy.linalg.LinAlgError: Matrix is not positive definite
>>> np.__version__
'2.5.0.dev0+git20260512.d7735ce'
```

rgommers:
That `linalg` issue could still be in NumPy (due to differences in the random numbers generated or the `np.matmul` precision), or in OpenBLAS. The test is sensitive to numerical precision issues, see gh-8385 and gh-13658 for much older occurrences of this same error.

kashyapc:
I ran this on some new RISC-V hardware,SpacemiT K3. I ran this in a Fedora 43 container.

That was my setup:
```
$ python3.12 -m venv numpy_venv_3.12
$ source numpy_venv_3.12/bin/activate
$ pip install --upgrade pip
$ pip install meson pytest hypothesis
$ unzip cp312-manylinux_riscv64-.zip -d numpy_artifact
$ pip install numpy_artifact/*.whl
```

Ran NumPy's bundled test suite; it failed with a bunch of tests
```
$ { time python3.12 -c 'import numpy; numpy.test()'; } 
NumPy version 2.5.0.dev0+git20260512.d7735ce
NumPy CPU features:   RVV*
[...]
../numpy_venv_3.12/lib64/python3.12/site-packages/numpy/linalg/_linalg.py:888: in cholesky
    r = gufunc(a, signature=signature)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        a          = array([[[10.73433257, -4.87682749, -3.1554569 , -4.59975348,
          2.10303633, -1.24351932,  3.03068619, -4.149521...      ,  0.        ,
          0.        ,  0.        ,  0.        ,  0.        ,
         -7.0796601 ,  9.4854158 ]]])
        gufunc     = <ufunc 'cholesky_up'>
        result_t   = <class 'numpy.float64'>
        signature  = 'd->d'
        t          = <class 'numpy.float64'>
        upper      = True
        wrap       = <built-in method __array_wrap__ of numpy.ndarray object at 0x3f97825fb0>
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

err = 'invalid value', flag = 8

    def _raise_linalgerror_nonposdef(err, flag):
>       raise LinAlgError("Matrix is not positive definite")
E       numpy.linalg.LinAlgError: Matrix is not positive definite

err        = 'invalid value'
flag       = 8

../numpy_venv_3.12/lib64/python3.12/site-packages/numpy/linalg/_linalg.py:148: LinAlgError
=============================== warnings summary ===============================
_core/tests/test_records.py::TestFromrecords::test_recarray_conflict_fields
  /root/numpy_venv_3.12/lib64/python3.12/site-packages/numpy/_core/records.py:462: DeprecationWarning: Setting the shape on a NumPy array has been deprecated in NumPy 2.5.
  As an alternative, you can create a new view using np.reshape (with copy=False if needed).
    ret = object.__setattr__(self, attr, val)

lib/tests/test_function_base.py::TestQuantile::test_nan_err[err1-O-weights0]
lib/tests/test_function_base.py::TestQuantile::test_nan_err[err1-O-weights1]
  /root/numpy_venv_3.12/lib64/python3.12/site-packages/numpy/lib/_function_base_impl.py:4505: RuntimeWarning: invalid value encountered in less
    if np.any(weights < 0):

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=========================== short test summary info ============================
FAILED linalg/tests/test_linalg.py::TestCholesky::test_basic_property[False-float32-shape3]
FAILED linalg/tests/test_linalg.py::TestCholesky::test_basic_property[False-float32-shape4]
FAILED linalg/tests/test_linalg.py::TestCholesky::test_basic_property[False-float64-shape3]
FAILED linalg/tests/test_linalg.py::TestCholesky::test_basic_property[False-float64-shape4]
FAILED linalg/tests/test_linalg.py::TestCholesky::test_basic_property[True-float32-shape3]
FAILED linalg/tests/test_linalg.py::TestCholesky::test_basic_property[True-float32-shape4]
FAILED linalg/tests/test_linalg.py::TestCholesky::test_basic_property[True-float64-shape3]
FAILED linalg/tests/test_linalg.py::TestCholesky::test_basic_property[True-float64-shape4]
ERROR _core/tests/test_mem_policy.py::test_set_policy - subprocess.CalledProc...
ERROR _core/tests/test_mem_policy.py::test_default_policy_singleton - subproc...
ERROR _core/tests/test_mem_policy.py::test_policy_propagation - subprocess.Ca...
ERROR _core/tests/test_mem_policy.py::test_context_locality - subprocess.Call...
ERROR _core/tests/test_mem_policy.py::test_thread_locality - subprocess.Calle...
ERROR _core/tests/test_mem_policy.py::test_switch_owner[0] - subprocess.Calle...
ERROR _core/tests/test_mem_policy.py::test_switch_owner[1] - subprocess.Calle...
ERROR _core/tests/test_mem_policy.py::test_switch_owner[None] - subprocess.Ca...
ERROR _core/tests/test_mem_policy.py::test_owner_is_base - subprocess.CalledP...
8 failed, 45267 passed, 1792 skipped, 2845 deselected, 31 xfailed, 2 xpassed, 3 warnings, 9 errors in 478.95s (0:07:58)
```

I've attached the full log here (232KB):
[numpy_test_results_on_k3.log](https://github.com/user-attachments/files/27708882/numpy_test_results_on_k3.log)

rgommers:
Thanks @kashyapc! Could you post a traceback on one of the `test_mem_policy.py` ones (I assume they're all the same)?  EDIT: the link to the log file doesn't work for me

gounthar:
Reproduced on my BananaPi F3 (SpaceMiT K1, rv64gc). Walking back my earlier "0 accuracy failures" comment, that was based on a different setup (system numpy with system OpenBLAS), not this PR's wheel. Sorry for the noise.

Running rgommers' minimal repro with the wheel from runs/25744398974:

```
numpy 2.5.0.dev0+git20260512.d7735ce
BLAS: scipy-openblas 0.3.33.0.0 (DYNAMIC_ARCH, riscv64_generic)

a is symmetric: True
min eigenvalue = -2.779433e+00      # mathematically should be >= 0
max eigenvalue = 1.949916e+02
LinAlgError: Matrix is not positive definite
```

Same machine, swapping in system numpy 2.4.3 with system OpenBLAS 0.3.29 (libopenblas0 0.3.29+ds-3, Debian 13 trixie): cholesky succeeds.

So cholesky is correctly rejecting a matrix that the matmul step computed wrong. The bug is upstream of cholesky, in OpenBLAS GEMM/SYRK on K1. Either a regression between 0.3.29 and 0.3.33, or a packaging difference between Debian's build and scipy-openblas's. Both targets compile to riscv64_generic, so it's not vector-related.

Lines up with what eshattow showed: Pine64 Star64 (StarFive JH-7110, SiFive U74) doesn't reproduce this, while threexc's Muse Pi Pro and my F3 (both K1) do. K1-specific so far.

I can bisect 0.3.30 through 0.3.32 and open an upstream OpenBLAS issue once I've narrowed the range. Will report back.

gounthar:
Bisected on the BananaPi F3 (LD_PRELOAD swap of `libscipy_openblas64_*.so`, same numpy wheel throughout):

| scipy-openblas64 | embedded OpenBLAS | min eigval | cholesky |
| --- | --- | --- | --- |
| 0.3.31.22.1 | 0.3.31.dev | +1.405e-02 | OK |
| 0.3.31.126.1 | 0.3.31.dev | +1.405e-02 | OK |
| 0.3.31.126.5 | 0.3.31.dev | +1.405e-02 | OK |
| 0.3.31.159.0 | 0.3.31.159.0 | +1.405e-02 | OK |
| 0.3.31.188.0 | 0.3.31.188.0 | +1.405e-02 | OK |
| 0.3.33.0.0 | 0.3.33.0.0 | -2.779e+00 | LinAlgError |

Clean break between OpenBLAS 0.3.31 and 0.3.33 (no 0.3.32 in scipy-openblas64 PyPI). The K1 has no RVV, so DYNAMIC_ARCH dispatches to riscv64_generic kernels. 250 commits between the v0.3.31 and v0.3.33 tags.

Filing an upstream OpenBLAS issue with the repro and this bisect range. Will link back here.

gounthar:
Upstream issue filed: OpenMathLib/OpenBLAS#5811

threexc:
> Bisected on the BananaPi F3 (LD_PRELOAD swap of `libscipy_openblas64_*.so`, same numpy wheel throughout):
> scipy-openblas64 	embedded OpenBLAS 	min eigval 	cholesky
> 0.3.31.22.1 	0.3.31.dev 	+1.405e-02 	OK
> 0.3.31.126.1 	0.3.31.dev 	+1.405e-02 	OK
> 0.3.31.126.5 	0.3.31.dev 	+1.405e-02 	OK
> 0.3.31.159.0 	0.3.31.159.0 	+1.405e-02 	OK
> 0.3.31.188.0 	0.3.31.188.0 	+1.405e-02 	OK
> 0.3.33.0.0 	0.3.33.0.0 	-2.779e+00 	LinAlgError
> 
> Clean break between OpenBLAS 0.3.31 and 0.3.33 (no 0.3.32 in scipy-openblas64 PyPI). The K1 has no RVV, so DYNAMIC_ARCH dispatches to riscv64_generic kernels. 250 commits between the v0.3.31 and v0.3.33 tags.
> 
> Filing an upstream OpenBLAS issue with the repro and this bisect range. Will link back here.

What do you mean by "no RVV"? The K1 is [advertised](https://docs.banana-pi.org/en/BPI-F3/SpacemiT_K1) as RVA22 + RVV 1.0.

rgommers:
> Getting there, down to 34 minutes in [this job](https://github.com/numpy/numpy/actions/runs/25792509048/job/75761425353) after including test selection changes in [gh-31420](https://github.com/numpy/numpy/pull/31420):
>
> Build and test phase are then taking about as long. We may indeed want to consider changing this job to use `ccache` next, that should bring it down to 20-25 minutes which is reasonable.

Of course if 34 minutes is tolerance for a little while, we can also decide to just merge this as is now, so we get some hands-on experience. If new hardware is coming in the summer that's 2-3x faster, that will address the performance and it'll be fine with no further changes needed.

Currently the other two slowest jobs:
- _Linux Qemu tests / riscv64_: varies between 21 and 31 min. in logs I checked
-  _Linux SIMD tests / old gcc_: 24 min.

We can turn off the Qemu riscv64 test if this new one is stable, so the overall extra runtime before that new hardware arrives only takes a minor hit.

There's something to say for any of these options:
1. Merge as is, wait for new hardware
2. Merge as is, immediately work on a follow-up PR to move the job away from `wheels.yml` and use `ccache` 
3. Don't merge because 34 minutes is too slow, work on a job with `ccache` now

@andyfaff any preferences?
@luhenry any preferences taking into account the load on your infra?

andyfaff:
I think options 2 or 3 are preferable .

luhenry:
> @luhenry any preferences taking into account the load on your infra?

Looking at the [data](https://github.com/numpy/numpy/actions/metrics/usage?dateRangeType=DATE_RANGE_TYPE_LAST_30_DAYS&filters=workflow_file_name%3Awheels.yml+runner_runtime%3Alinux), there are around ~300-350 wheels.yml workflow runs / 30day. At 30min per run, the load on the infra infrastructure would be negligible. (we are currently at the 300-500 runs/day and the infra is far from peak capacity).

> There's something to say for any of these options:
> 1. Merge as is, wait for new hardware
> 2. Merge as is, immediately work on a follow-up PR to move the job away from `wheels.yml` and use `ccache`
> 3. Don't merge because 34 minutes is too slow, work on a job with `ccache` now

To me, the main question is what _you_, as maintainers, are happy to take on. We should make the load as small as possible for everyday's maintainers and contributors, and riscv64 shouldn't be a pain point. I don't think we are in such a rush (we all want to see it integrated!) that we can't work on some workarounds before merging (split out of wheels.yml? ccache? all of the above?).

Integrating `ccache` would in my experience be quite a focused change with:
1. use https://github.com/hendrikmuhs/ccache-action to manage the cache save/restore to GHA Cache
2. setup ccache _inside_ the container: `CIBW_BEFORE_ALL_LINUX: 'yum install -y ccache'` should do?
3. setup the ccache environment variables:
```
CIBW_ENVIRONMENT_LINUX: >-
  CC="ccache gcc"
  CXX="ccache g++"
  CCACHE_DIR=/host/tmp/ccache
  CCACHE_COMPILERCHECK=content
  CCACHE_BASEDIR=/project
```
4. profit?

It's worth trying out, I can give it a try today or tomorrow.

rgommers:
Okay, option (3) it is:) 

> 1. use https://github.com/hendrikmuhs/ccache-action to manage the cache save/restore to GHA Cache

Interesting, I didn't see that before. We normally use regular https://github.com/actions/cache and set `CCACHE_DIR`, which this action seems to be doing under the hood. We only have one usage in NumPy right now, [here](https://github.com/numpy/numpy/blob/ff11fb225c01c59880094d4ce659ff0731a64fc8/.github/workflows/macos.yml#L42-L50) in `macos.yml`. And that looks outdated/suboptimal, we want to run the job on merges to `main` and only store caches there I think. 

In SciPy we recently overhauled this and made sure the behavior was optimal for cache hits and storage consumption. It uses a generic composite action ([this one](https://github.com/scipy/scipy/blob/930870e57a31074e8f80e5f35a30d74bd1386563/.github/ccache/action.yml)) and then there's still two job steps per job that uses `ccache`, plus setting `CCACHE_DIR`:

https://github.com/scipy/scipy/blob/930870e57a31074e8f80e5f35a30d74bd1386563/.github/workflows/linux.yml#L162-L166

https://github.com/scipy/scipy/blob/930870e57a31074e8f80e5f35a30d74bd1386563/.github/workflows/linux.yml#L205-L209

If https://github.com/hendrikmuhs/ccache-action works the same, I think we can use that. 




> 2. setup ccache _inside_ the container: `CIBW_BEFORE_ALL_LINUX: 'yum install -y ccache'` should do?

Agreed.

> 3. setup the ccache environment variables:

That isn't even necessary (except for `CCACHE_DIR` probably). As long as `ccache` is on the PATH, `meson` will use it automatically.

> It's worth trying out, I can give it a try today or tomorrow.

That would be great, thanks. I'd suggest creating a new `linux_riscv64.yml`, with contents stripped down from `wheels.yml` plus the extra `push: branches: main` trigger.

luhenry:
> Interesting, I didn't see that before. We normally use regular https://github.com/actions/cache and set `CCACHE_DIR`, which this action seems to be doing under the hood. We only have one usage in NumPy right now, [here](https://github.com/numpy/numpy/blob/ff11fb225c01c59880094d4ce659ff0731a64fc8/.github/workflows/macos.yml#L42-L50) in `macos.yml`. And that looks outdated/suboptimal, we want to run the job on merges to `main` and only store caches there I think.

The main advantage I saw to this https://github.com/hendrikmuhs/ccache-action is that you can selectively save. You can save on workflows to main, and never save on workflows on pull requests for example. This way you have a good "hot" cache but don't pollute it with many from pull requests.

> That isn't even necessary (except for CCACHE_DIR probably). As long as ccache is on the PATH, meson will use it automatically.

We would want to mount the ccache dir from the host and so need to point the ccache inside the container to that directory. So if meson automatically picks it up it's a lot easier indeed.

gounthar:
You're right, I had this completely wrong. The K1 X60 cores do implement RVV 1.0 (the ISA string in /proc/cpuinfo shows `_v_` along with the Zve/Zvfh/Zvkt extensions, VLEN=256).

Re-checked with `OPENBLAS_VERBOSE=2` on the F3:

```
=== scipy-openblas64==0.3.31.188.0 ===
Core: riscv64_zvl256b

=== scipy-openblas64==0.3.33.0.0 ===
Core: riscv64_zvl256b
```

So OpenBLAS DYNAMIC_ARCH dispatches to the riscv64_zvl256b kernels on the K1, not generic. The bisect range is unchanged (0.3.31.188.0 OK, 0.3.33.0.0 broken), but the regression is somewhere in the vector kernels (or something they call), not the scalar generic path.

That actually fits the earlier observation that the Pine64 Star64 (SiFive U74, no RVV, dispatches to riscv64_generic) does not reproduce. Same regression range, different kernel path, different result, and the failing path is the vectorized one.

Updating OpenMathLib/OpenBLAS#5811 with the corrected dispatch info.

kashyapc:
> Thanks @kashyapc! Could you post a traceback on one of the `test_mem_policy.py` ones (I assume they're all the same)? EDIT: the link to the log file doesn't work for me

@rgommers Hey, can you please try again the link in my [comment here](https://github.com/numpy/numpy/pull/30995#issuecomment-4440939372).  (I fixed the formatting, but the same link worked for me.)

As for traceback, I'll let you know once I get it.

gounthar:
The 8 `TestCholesky::test_basic_property` failures in @kashyapc's K3 run are a known OpenBLAS regression, not a numpy or wheel defect. I filed it as OpenMathLib/OpenBLAS#5811: DGEMM/SYRK regressed between scipy-openblas64 0.3.31 and 0.3.33 on the riscv64 vector kernel path, so `A.T @ A` comes back with a negative eigenvalue and Cholesky correctly rejects the result as non-PSD. It reproduces on the K1 (zvl256b dispatch) and the K3 signature matches exactly. I bisected it to OpenBLAS git history between v0.3.31 and v0.3.33; the repro and bisect range are in that issue.

The 9 `test_mem_policy.py` errors look like a different thing. They're `subprocess.CalledProcessError` on the in-subprocess library load, which is the path @rgommers is already digging into, not the linalg math.

So nothing in this PR's diff is implicated by either bucket: the build is clean and the job is green. The Cholesky path will stay red on K1/K3 hardware until the OpenBLAS fix lands, and I'm tracking that at #5811.

rgommers:
> The 9 `test_mem_policy.py` errors look like a different thing. They're `subprocess.CalledProcessError` on the in-subprocess library load, which is the path @rgommers is already digging into, not the linalg math.

The relevant part of the traceback is this:
```
>           raise CalledProcessError(retcode, cmd)
E           subprocess.CalledProcessError: Command '['meson', 'setup', '--vsenv', '..', '--native-file=/tmp/pytest-of-root/pytest-0/test_set_policy0/mem_policy/.mesonpy-native-file.ini']' returned non-zero exit status 1.

cmd        = ['meson', 'setup', '--vsenv', '..', '--native-file=/tmp/pytest-of-root/pytest-0/test_set_policy0/mem_policy/.mesonpy-native-file.ini']
kwargs     = {'cwd': PosixPath('/tmp/pytest-of-root/pytest-0/test_set_policy0/mem_policy/build')}
popenargs  = (['meson', 'setup', '--vsenv', '..', '--native-file=/tmp/pytest-of-root/pytest-0/test_set_policy0/mem_policy/.mesonpy-native-file.ini'],)
retcode    = 1

/usr/lib64/python3.12/subprocess.py:413: CalledProcessError
---------------------------- Captured stdout setup -----------------------------
The Meson build system
Version: 1.11.1
Source dir: /tmp/pytest-of-root/pytest-0/test_set_policy0/mem_policy
Build dir: /tmp/pytest-of-root/pytest-0/test_set_policy0/mem_policy/build
Build type: native build
Project name: foo
Project version: undefined
C compiler for the host machine: cc (gcc 15.2.1 "cc (GCC) 15.2.1 20260123 (Red Hat 15.2.1-7)")
C linker for the host machine: cc ld.bfd 2.45.1-4
Host machine cpu family: riscv64
Host machine cpu: riscv64
Program python found: YES (/root/numpy_venv_3.12/bin/python3.12)
Found pkg-config: YES (/usr/bin/pkg-config) 2.3.0
Run-time dependency python found: NO  (tried pkgconfig and sysconfig)

../meson.build:3:3: ERROR: Python dependency not found

A full log can be found at /tmp/pytest-of-root/pytest-0/test_set_policy0/mem_policy/build/meson-logs/meson-log.txt
```

At first sight that looks like an incomplete Python install, e.g. `python-config` is missing, or `python-dev` which provides `Python.h` (my guess is the latter). I don't think it requires a change in `numpy`.

gounthar:
Confirming @rgommers's read with a data point from native riscv64. On the BPI-F3 (SpaceMiT K1, Armbian trixie), the same cp312 wheel from this PR's CI (`2.5.0.dev0+git20260512.d7735ce`) runs `test_mem_policy.py` clean: 9 passed, 1 skipped in 18.7s, native, not QEMU. That is the exact set of tests that errored on @kashyapc's K3 run.

The only difference is the Python: the venv here uses a standalone 3.12 that ships `Python.h`, so `meson` resolves the `python` dependency. The Fedora container python3.12 had no devel package, which is what the `meson setup` failure was. So this is environment, not riscv64 or the wheel, and nothing for numpy to change.

luhenry:
Submitted a PR for adding ccache support: https://github.com/numpy/numpy/pull/31474

luhenry:
Given feedback at [1] and [2], I suggest we add a `linux_riscv64.yml` file and extract that wheel building steps into it. @gounthar want to have a look, or should I do it? Happy either way. If you want to do it, simply copy/paste what I've done at https://github.com/numpy/numpy/pull/31474, all good for me.

[1]:
https://github.com/numpy/numpy/pull/31474#issuecomment-4503246797
[2]:
https://github.com/numpy/numpy/pull/31474#issuecomment-4506413527

gounthar:
Go for it, Ludovic. You've already got the working version in #31474, so it makes sense for you to extract it rather than me copying it over. Once linux_riscv64.yml is up I'll run it against the Banana Pi F3 (native K1) and review. Thanks for offering either way.

andyfaff:
Thank you for being understanding, https://github.com/numpy/numpy/blob/main/.github/workflows/linux-ibm.yml is an example to base the file off.

rgommers:
gh-31488 was merged and incorporated most of this PR, so I'll close this now. Thanks all!

PR #30996, opened 2026-03-12
ENH: replace direct access of struct fields with macros and static inline functions

kumaraditya303:
### PR summary

Replace direct access of structs fields with macros and static inline functions. 

#### AI Disclosure

No AI tools used


ngoldbaum (approved):
These should all compile to direct field accesses, except on future opaque PyObject builds. It's possible some compilers will fail to inline the static inline functions, but IMO we shouldn't be working around that, especially when we're already using static inline functions in a lot of these examples anyway.

seberg (approved):
Thanks, nice cleanups/fixes!

PR #31000, opened 2026-03-13
BUG: Raise BufferError instead of RuntimeError in from_dlpack

### LINKED ISSUE #30937: BUG: Should `np.from_dlpack` raise BufferError instead of ValueError on unsupported dl_device values?
### Describe the issue:

Relates to #30341 and #30936

In the [from_dlpack](https://data-apis.org/array-api/latest/API_specification/generated/array_api.from_dlpack.html#from-dlpack) docs it says:

> BufferError – The __dlpack__ and __dlpack_device__ methods on the input array may raise BufferError when the data cannot be exported as DLPack (e.g., incompatible dtype, strides, or device).

I read the meaning as: incompatible dtype, incompatible strides, or incompatible device - which would suggest a BufferError should be raised when device is incompatible?

I understand `BufferError` may seem like an odd choice (https://github.com/numpy/numpy/issues/30341#issuecomment-3599629104) but just raising the question from my interpretation of the docs.

```python
x_torch = torch.arange(5, device='cuda')
np.from_dlpack(x_torch)
```

### Error message:

```shell
RuntimeError: Unsupported device in DLTensor.
```

2.4.2
3.14.2 | packaged by conda-forge | (main, Feb  4 2026, 15:49:49) [GCC 14.3.0]

### Runtime Environment:

_No response_

https://github.com/scikit-learn/scikit-learn/pull/32755#issuecomment-3970642029

joshuaswanson:
`from_dlpack` currently raises `RuntimeError` when it encounters an unsupported device, dtype, or dimension count in the incoming DLPack tensor. Both the DLPack spec and the Array API spec say this should be a `BufferError`.

This only changes the four `RuntimeError`s on the consumer side (`from_dlpack` in `dlpack.c`). The producer-side `RuntimeError` in `__dlpack__` for unsupported streams is left as-is, and `ValueError`s for invalid user arguments (like `device="gpu"`) are also untouched.

Closes #30937

### DIFF
--- a/doc/release/upcoming_changes/30937.compatibility.rst
+++ b/doc/release/upcoming_changes/30937.compatibility.rst
@@ -1,0 +1,8 @@
+``from_dlpack`` raises ``BufferError`` instead of ``RuntimeError``
+------------------------------------------------------------------
+
+``np.from_dlpack`` now raises ``BufferError`` instead of ``RuntimeError``
+when the incoming DLPack tensor has an unsupported device, dtype, or
+exceeds the maximum number of dimensions. This aligns with the DLPack
+and Array API specifications, which recommend ``BufferError`` for data
+that cannot be imported.
--- a/numpy/_core/src/multiarray/dlpack.c
+++ b/numpy/_core/src/multiarray/dlpack.c
@@ -577,7 +577,7 @@ from_dlpack(PyObject *NPY_UNUSED(self),
 
     const int ndim = dl_tensor.ndim;
     if (ndim > NPY_MAXDIMS) {
-        PyErr_SetString(PyExc_RuntimeError,
+        PyErr_SetString(PyExc_BufferError,
                 "maxdims of DLPack tensor is higher than the supported "
                 "maxdims.");
         Py_DECREF(capsule);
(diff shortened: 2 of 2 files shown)


doc/release/upcoming_changes/30937.compatibility.rst
```diff
@@ -0,0 +1,8 @@
+``from_dlpack`` raises ``BufferError`` instead of ``RuntimeError``
+------------------------------------------------------------------
+
+``np.from_dlpack`` now raises ``BufferError`` instead of ``RuntimeError``
+when the incoming DLPack tensor has an unsupported device, dtype, or
```
tylerjereddy: If that's really what happens on this branch, why is `spin test -t numpy/_core/tests/test_dlpack.py` passing given a test case (`test_device`) that does:

```python
187         with pytest.raises(ValueError):
188             np.from_dlpack(x, device="gpu")
```

tylerjereddy (commented):
It does seem to do the right thing for a reproducer similar to the one in the issue on this branch:

```python
import numpy as np
import torch

x_torch = torch.arange(5, device='mps')
np.from_dlpack(x_torch)
```

```
  File "/Users/treddy/github_projects/numpy/probe.py", line 5, in <module>
    np.from_dlpack(x_torch)
BufferError: Unsupported device in DLTensor.
```

Is there a clean way to write a lightweight regression test that doesn't add external deps? I messed around with it for a bit, and even with mocking it seemed annoying to do.

joshuaswanson:
I couldn't find a clean way either. Building a fake DLManagedTensor with ctypes is possible but pretty fragile since it depends on the exact C struct layout. The dtype and byte-swapping error paths are already covered by `test_invalid_dtype` and `test_invalid_byte_swapping` (both expect `BufferError`), so the change is partially tested. Happy to add a ctypes-based mock if you think it's worth it, but I'd understand if you'd rather just merge without the device-specific test.

seberg:
Yeah, dunno... Let's just do this, I think it is rather annoying to test. There isn't much point in adding one of these as a dependency.
I think the only way would be to create a (bogus) DLtensor with a wrong device.

Thanks @joshuaswanson.

PR #31004, opened 2026-03-13
ENH: make bound array method immortal to reduce reference counting contention

kumaraditya303:
### PR summary

Running the script from https://github.com/numpy/numpy/issues/30494 under `perf c2c` found that there is a lot of reference counting contention in `PyArray_GetCastingImpl`. This PR makes the array methods immortal which removes the contention. 

See below for the `perf c2c` report (high RmtHitm indicates high cacheline contention):

```console

Cacheline 0x200025d4d00
  ----- HITM -----  ------- Store Refs ------  ------- CL --------                      ---------- cycles ----------    Total       cpu                                                                            Shared
  RmtHitm  LclHitm   L1 Hit  L1 Miss      N/A    Off  Node  PA cnt        Code address  rmt hitm  lcl hitm      load  records       cnt                       Symbol                                               Object         Source:Line  Node
+   6.54%    6.54%    0.00%    0.00%    0.00%   0x10     0       1      0x7069f8d74d85       397         0       656       19        10  [.] PyArray_GetCastingImpl   _multiarray_umath.cpython-315td-x86_64-linux-gnu.so  refcount.h:274       0
+   0.93%    0.93%    0.00%    0.00%    0.00%   0x10     0       1      0x7069f8d76757        51         0         0        4         3  [.] PyArray_CheckCastSafety  _multiarray_umath.cpython-315td-x86_64-linux-gnu.so  refcount.h:345       0
+  42.06%   42.06%    0.00%    0.00%    0.00%   0x1c     0       1      0x7069f8d5cd09       776         0       445      122        45  [.] NPY_cast_info_xfree      _multiarray_umath.cpython-315td-x86_64-linux-gnu.so  pyatomic_gcc.h:367   0
+  40.19%   40.19%    0.00%    0.00%    0.00%   0x1c     0       1      0x7069f8d74d94       204         0       112      138        48  [.] PyArray_GetCastingImpl   _multiarray_umath.cpython-315td-x86_64-linux-gnu.so  pyatomic_gcc.h:367   0
+   2.80%    2.80%    0.00%    0.00%    0.00%   0x1c     0       1      0x7069f8d76740       520         0       147       15        10  [.] PyArray_CheckCastSafety  _multiarray_umath.cpython-315td-x86_64-linux-gnu.so  pyatomic_gcc.h:367   0
+   1.87%    1.87%    0.00%    0.00%    0.00%   0x1c     0       1      0x7069f8d74d71       696         0      1573        5         3  [.] PyArray_GetCastingImpl   _multiarray_umath.cpython-315td-x86_64-linux-gnu.so  pyatomic_gcc.h:367   0
+   2.80%    2.80%    0.00%    0.00%    0.00%   0x20     0       1      0x7069f8d74d8e       479         0       443        9         6  [.] PyArray_GetCastingImpl   _multiarray_umath.cpython-315td-x86_64-linux-gnu.so  pyatomic_gcc.h:63    0
+   2.80%    2.80%    0.00%    0.00%    0.00%   0x20     0       1      0x7069f9dd8fe0       138         0         0        9         6  [.] _Py_DecRefSharedDebug    libpython3.15td.so.1.0                               pyatomic_gcc.h:130   0

```

### DIFF
--- a/numpy/_core/src/multiarray/array_method.c
+++ b/numpy/_core/src/multiarray/array_method.c
@@ -129,9 +129,9 @@ is_contiguous(
  * param move_references UNUSED -- listed below but doxygen doesn't see as a parameter
  * @param strides Array of step sizes for each dimension of the arrays involved
  * @param out_loop Output pointer to the function that will perform the strided loop.
- * @param out_transferdata Output pointer to auxiliary data (if any) 
+ * @param out_transferdata Output pointer to auxiliary data (if any)
  *        needed by the out_loop function.
- * @param flags Output pointer to additional flags (if any) 
+ * @param flags Output pointer to additional flags (if any)
  *        needed by the out_loop function
  * @returns 0 on success -1 on failure.
  */
@@ -485,7 +485,9 @@ PyArrayMethod_FromSpec_int(PyArrayMethod_Spec *spec, int private)
         return NULL;
     }
     strcpy(res->method->name, spec->name);
-
+#ifdef Py_GIL_DISABLED
+    PyUnstable_SetImmortal((PyObject *)res->method);
+#endif
     return res;
 }
 
--- a/numpy/_core/src/multiarray/array_method.c
+++ b/numpy/_core/src/multiarray/array_method.c
@@ -486,6 +486,9 @@ PyArrayMethod_FromSpec_int(PyArrayMethod_Spec *spec, int private)
     }
     strcpy(res->method->name, spec->name);
 #ifdef Py_GIL_DISABLED
+    // Mark immortal to reduce reference count contention in PyArray_GetCastingImpl
+    // If we ever allow replacing ArrayMethod objects or cleanup it DTypes or ufuncs, this may need to be reconsidered.
+    // An alternative that might help is to store cast methods in a PyArrayIdentityHash instead of a dict.
     PyUnstable_SetImmortal((PyObject *)res->method);
 #endif
     return res;


seberg:
Sorry, maybe it's obvious from your timings, but how much of a difference does it make? I have long accepted that DTypes are effectively immortal; except maybe for some future subclass patterns (Python is actually missing the capabilities of making them mortal in general to begin with).

But, I am a bit unsure about making this blanket immortal, if just out of a principle that we try to e.g. make `ufunc`s clean up for the most part (except builtin NumPy ones for which we leak the docs).

(For the specific code line it seems like making sure that a cast function is never replaced and thus using a borrowed reference rather than `GetItemRef` might also work? Although, I am sure there are other paths where we can't strictly rely on borrowed refs.)

seberg:
That said, if it's clearly worthwhile, we can just do it. But maybe document it? It would also be easy enough to allow flagging that one wants it to be mortal in the future (not particularly likely that it is important, but you never know).

kumaraditya303:
> Sorry, maybe it's obvious from your timings, but how much of a difference does it make? I have long accepted that DTypes are effectively immortal; except maybe for some future subclass patterns (Python is actually missing the capabilities of making them mortal in general to begin with).

The speed up varies depending on the platform but I measure 5-10% improvement on mac and linux.

kumaraditya303:
> Although, I am sure there are other paths where we can't strictly rely on borrowed refs.)

Using borrowed reference is not safe because `ensure_castingimpl_exists` follows the pattern of check-and-act, if between checking and adding an entry the critical section gets released then some thread can access freed cast and the pointer would become dangling, there are ways to workaround that but that would require more changes than just making it immortal in free-threading where it is relevant (the regular builds remain unaffected).

seberg:
> 5-10% improvement

That's pretty nice.

> if between checking and adding an entry the critical section gets released then some thread can access freed cast

Not sure I am following. Isn't the only problem that it still uses `PyDict_SetItem` rather than `PyDict_SetItemRef`?  I.e. I don't think we really allow mutating the dict, except adding things and `PyDict_SetItemRef` makes that safe, I think?

EDIT: Sorry, I meant `SetItemDefault`.

kumaraditya303:
> Not sure I am following. Isn't the only problem that it still uses PyDict_SetItem rather than PyDict_SetItemRef? I.e. I don't think we really allow mutating the dict, except adding things and PyDict_SetItemRef makes that safe, I think?

There is no `PyDict_SetItemRef` in C-API AFAIK, the issue is that multiple threads could try to set the cast-impl for the same key and in that case only one would succeed in the end, it can cause consistency issue where different threads could be using different objects for same castimpl.

seberg:
Sorry, I meant to copy paste `PyDict_SetDefaultRef`, `pythoncapi-compat` seems to include it.

kumaraditya303:
> Sorry, I meant to copy paste PyDict_SetDefaultRef, pythoncapi-compat seems to include it.

Yes, `PyDict_SetDefaultRef` can be used which would avoid this issue but I haven't checked if there are more such places where these issues can happen.

seberg:
>  but I haven't checked if there are more such places where these issues can happen.

I can help there: there aren't.  There is currently no API that allows replacing these. But, I dunno... if such API _existed_, should we actually prefer leaking under the assumption that if they get replaced then surely not more than once (i.e. immortality is OK in practice)?

kumaraditya303:
> I can help there: there aren't. There is currently no API that allows replacing these. But, I dunno... if such API existed, should we actually prefer leaking under the assumption that if they get replaced then surely not more than once (i.e. immortality is OK in practice)?

If we can assume that there is no way to replace entries then I think it would be better if we can remove the dictionary based implementation altogether and instead use `PyArrayIdentityHash`. It would be much easier to reason about because it does no reference counting and `PyArrayIdentityHash_SetItemDefault` is atomic and would scale much better than the current implementation.

ngoldbaum:
> then I think it would be better if we can remove the dictionary based implementation altogether and instead use PyArrayIdentityHash

Are you interested in working on that? Or should we merge this to get an easy win now?

kumaraditya303:
> Are you interested in working on that? Or should we merge this to get an easy win now?

I think we should merge this for now.

numpy/_core/src/multiarray/array_method.c
```diff
@@ -485,7 +485,9 @@ PyArrayMethod_FromSpec_int(PyArrayMethod_Spec *spec, int private)
         return NULL;
     }
     strcpy(res->method->name, spec->name);
-
+#ifdef Py_GIL_DISABLED
```
ngoldbaum, line 488:
```suggestion
#ifdef Py_GIL_DISABLED
    // Mark immortal to reduce reference count contention in PyArray_GetCastingImpl
    // If we ever allow replacing ArrayMethod objects or cleanup it DTypes or ufuncs, this may need to be reconsidered.
    // An alternative that might help is to store cast methods in a PyArrayIdentityHash instead of a dict.
```

ngoldbaum: I think that summarizes the discussion that happened here - @seberg, is leaving this comment behind sufficient for you?

seberg: I can live with it.  It also disables full cleanup of ufuncs and DTypes (DTypes are unlikely to be cleaned up).
(But yeah, since your motivation is casts seems fine.)

ngoldbaum: I added a note about ufunc and dtype cleanup

ngoldbaum:
Thanks @kumaraditya303!

PR #31008, opened 2026-03-13
MAINT: Remove COLS_IN_ALPHA_INDEX from Doxyfile.

charris:
This removes a warning and should make no difference in the output, Doxygen has been ignoring the line since about version 1.9.x.

AI: Grok suggested removal of the line was the simplest solution.

[skip cirrus] [skip actions] [skip azp]


seberg:
Can't find official release notes, but from a brief search it just seems right. Thanks.

PR #31009, opened 2026-03-13
ENH: Refactor ufunc implementation to avoid intermediate tuples

eendebakpt:
### PR summary

We replace the `ufunc_full_args` structure which contains tuples with a struct with plain C arrays. The input and output arguments are now allocated in the same way is the `signature`, `operands`, `operand_DTypes` and `operation_descrs`.

With this we can:

- Avoid construction of intermediate tuples in the common case
- Use borrowed references on the input arguments (for the common case)
- In future PRs: use borrowed references on the `operand_DTypes` (and perhaps more cases)

On a FT build this improves performance:
```
x + x: Mean +- std dev: [main] 470 ns +- 1 ns -> [pr] 446 ns +- 4 ns: 1.06x faster
np.abs(x): Mean +- std dev: [main] 424 ns +- 9 ns -> [pr] 401 ns +- 8 ns: 1.06x faster
np.cos(x): Mean +- std dev: [main] 422 ns +- 13 ns -> [pr] 397 ns +- 10 ns: 1.06x faster
np.add.reduce(x): Mean +- std dev: [main] 1.04 us +- 0.02 us -> [pr] 1.01 us +- 0.03 us: 1.03x faster
np.add.accumulate(x): Mean +- std dev: [main] 560 ns +- 11 ns -> [pr] 548 ns +- 6 ns: 1.02x faster

Geometric mean: 1.05x faster
```
<details><summary>Script</summary>

```
import pyperf

runner = pyperf.Runner()

setup = """
import numpy as np
x = np.array([1., 2.])
y = np.array([-1., 12.])
"""

runner.timeit(name="x + x", stmt="x + y", setup=setup)
runner.timeit(name="np.abs(x)", stmt="np.abs(x)", setup=setup)
runner.timeit(name="np.cos(x)", stmt="np.cos(x)", setup=setup)
runner.timeit(name="np.add.reduce(x)", stmt="np.add.reduce(x)", setup=setup)
runner.timeit(name="np.add.accumulate(x)", stmt="np.add.accumulate(x)", setup=setup)

```
</details>

#### AI Disclosure

Claude code was used heavily in creating the PR.


numpy/_core/src/umath/ufunc_object.c
```diff
@@ -84,24 +84,55 @@
...
+        Py_XDECREF(array[i]);
+    }
+}
+
+static inline void
+_decref_pyobjects(PyObject *const *array, npy_intp size) {
+    for (npy_intp i = 0; i < size; i++) {
+        Py_DECREF(array[i]);
+    }
+}
```
seberg: Makes sense as a helper, we can probably use it in a few places. Ideally can we move this to `common.h` or so, There is also a `multi_DECREF()` in `mapping.c` that can be merged with this.

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -84,24 +84,55 @@
...
-    PyObject *in;   /* The input arguments to the ufunc, a tuple */
-    PyObject *out;  /* The output arguments, a tuple. If no non-None outputs are
-                       provided, then this is NULL. */
-} ufunc_full_args;
+    PyObject *const *in;   /* Array of input arguments (borrowed or owned refs, immutable) */
+    PyObject **out;  /* Array of output arguments (owned refs). If no non-None outputs are
+                        provided, then this is NULL. */
+    npy_intp nin;    /* Number of input arguments */
+    npy_intp nout;   /* Number of output arguments (0 if out is NULL) */
+} ufunc_full_args_light;
```
seberg: I am still wondering if `full_args` can't just be a `PyObject **`, but I may have to see how it pans out.
The is a bit silly for sure though, the `_light`, makes no sense. `original_args` might be best, since even "full" was never very too the point.

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -84,24 +84,55 @@
...
+/*
+ * Release all owned output references held by a ufunc_full_args_light
+ * and reset the output count.
+ */
+static inline void
+ufunc_full_args_light_clear_out(ufunc_full_args_light *full_args)
+{
+    _xdecref_pyobjects(full_args->out, full_args->nout);
+    full_args->nout = 0;
+}
```
seberg: This is also weird.  If we need a helper to clean up full-args, it should just be a single helper that deals with everything.

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -3403,18 +3435,24 @@ _set_full_args_out(int nout, PyObject *out_obj, ufunc_full_args *full_args)
...
         else {
-            Py_INCREF(out_obj);
-            full_args->out = out_obj;
+            /* Populate full_args_light.out if provided (owned refs) */
+            if (full_args_light->out != NULL) {
+                for (int i = 0; i < nout; i++) {
+                    PyObject *item = PyTuple_GET_ITEM(out_obj, i);
+                    full_args_light->out[i] = Py_NewRef(item);
+                }
+                full_args_light->nout = nout;
```
seberg: Why are the output arguments owned suddenly?

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4477,24 +4530,27 @@ ufunc_generic_fastcall(PyUFuncObject *ufunc,
...
+    if (outer) {
+        /* outer modifies inputs, so we need owned refs in scratch space */
+        full_args_light.in = args_scratch;
+        for (int i = 0; i < nin; i++) {
+            ((PyObject **)full_args_light.in)[i] = Py_NewRef(args[i]);
+        }
+    }
+    else {
+        /* Borrowed references directly from vectorcall args */
+        full_args_light.in = args;
```
seberg: Does it matter with the borrowed vs. not.  Most of the time, I suspect we'll touch the refcount anyway, so just copying them over always is maybe just as well?

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4600,8 +4655,17 @@ ufunc_generic_fastcall(PyUFuncObject *ufunc,
             if (out_obj == Py_Ellipsis) {
                 return_scalar = NPY_FALSE;
             }
-            else if (_set_full_args_out(nout, out_obj, &full_args) < 0) {
-                goto fail;
+            else {
+                full_args_light.out = args_scratch + (outer ? nin : 0);
```
seberg: This looks like it doesn't make sense here. It should be merged with the entry code.

seberg: Why not just use `nout == 0` to signal no outputs passed.  It's always going to be correct and aligns code paths clearer.
(Yes, `.out != NULL` then, but that shouldn't matter?)

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -3621,36 +3656,41 @@ PyUFunc_GenericReduction(PyUFuncObject *ufunc,
...
+            full_args_light.out, full_args_light.nout,
+            wheremask_obj, args, len_args, kwnames, &override);
     if (errval) {
         return NULL;
     }
     else if (override) {
-        Py_XDECREF(full_args.in);
-        Py_XDECREF(full_args.out);
+        _xdecref_pyobjects(in_args, full_args_light.nin);
+        ufunc_full_args_light_clear_out(&full_args_light);
```
seberg: Why owning reference(s) here and elsewhere?

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4477,24 +4530,27 @@ ufunc_generic_fastcall(PyUFuncObject *ufunc,
         goto fail;
```
seberg: This fail goto looks like it should explode cleanup? Although, I am surprised we wouldn't have a single test excercising it.

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -84,24 +84,55 @@
...
 typedef struct {
-    PyObject *in;   /* The input arguments to the ufunc, a tuple */
-    PyObject *out;  /* The output arguments, a tuple. If no non-None outputs are
-                       provided, then this is NULL. */
-} ufunc_full_args;
+    PyObject *const *in;   /* Array of input arguments (borrowed or owned refs, immutable) */
+    PyObject **out;  /* Array of output arguments (owned refs). If no non-None outputs are
+                        provided, then this is NULL. */
+    npy_intp nin;    /* Number of input arguments */
+    npy_intp nout;   /* Number of output arguments (0 if out is NULL) */
```
seberg: integer is enough here.

numpy/_core/src/umath/override.h
```diff
@@ -6,7 +6,9 @@
 
 NPY_NO_EXPORT int
 PyUFunc_CheckOverride(PyUFuncObject *ufunc, char *method,
-        PyObject *in_args, PyObject *out_args, PyObject *wheremask_obj,
+        PyObject *const *in_args, int nin,
+        PyObject *const *out_args, int nout,
```
seberg, line 10: Maybe we should just move the struct definition and pass it?  (Maybe even merge with `NpyUFuncContext`, but may want to split out `ufunc` there then.)

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -3740,8 +3780,9 @@ PyUFunc_GenericReduction(PyUFuncObject *ufunc,
     Py_DECREF(signature[2]);
 
     Py_DECREF(mp);
-    Py_XDECREF(full_args.in);
-    Py_XDECREF(full_args.out);
+    _xdecref_pyobjects(in_args, full_args_light.nin);
+    ufunc_full_args_light_clear_out(&full_args_light);
+    Py_XDECREF(out_tuple);
```
seberg: `out_tuple` isn't ever used...

seberg (commented):
As I had mentioned offline, I like this, but I think we need to make it pretty to be worthwhile.
So a few comments, maybe they make sense, or maybe I'll push some changes if I feel like it soon.

(Part of me wonders also if the in-args should just always be a copy.)

numpy/_core/src/umath/ufunc_object.c
eendebakpt: The path where `outer` is true is a bit more complicated than the common path, this is now handled a bit better (but still, depending on `outer` the handling is slightly different)

numpy/_core/src/umath/ufunc_object.c
eendebakpt: Converted to plain `PyObject **`.

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4440,61 +4454,63 @@ ufunc_generic_fastcall(PyUFuncObject *ufunc,
...
     /*
      * Scratch space for operands, dtypes, etc.  Note that operands and
      * operation_descrs may hold an entry for the wheremask.
+     * Also includes space for ufunc_input (nin slots, used by outer) and
+     * ufunc_output (nout slots).
      */
-    NPY_ALLOC_WORKSPACE(scratch_objs, void *, UFUNC_STACK_NARGS * 4 + 2, nop * 4 + 2);
+    NPY_ALLOC_WORKSPACE(scratch_objs, void *,
+                        UFUNC_STACK_NARGS * 4 + 2,
+                        nop * 4 + 2 + nop);
```
```suggestion
    NPY_ALLOC_WORKSPACE(scratch_objs, void *, UFUNC_STACK_NARGS * 5 + 2, nop * 5 + 2);
```

numpy/_core/src/umath/override.c
```diff
@@ -113,9 +113,14 @@ initialize_normal_kwds(PyObject *out_args,
         }
     }
 
-    if (out_args != NULL) {
+    if (out_args != NULL && nout > 0) {
```
mhvk: Can `nout`  be zero if `out_args` is defined? (If this is defensive programming, add a comment, or perhaps just do `assert nout > 0` on the next line).

mhvk (commented):
@eendebakpt - Is there a reason this is still a draft? It looks very good to me! As I've always rather disliked the needless creation of input and output tuples, I'm very happy to see that removed here...

eendebakpt:
> @eendebakpt - Is there a reason this is still a draft? It looks very good to me! As I've always rather disliked the needless creation of input and output tuples, I'm very happy to see that removed here...

No real reason it was still draft. With these refactorings it is not always clear where to stop (doing more refactoring can be good, but also makes the diff larger), and I had not made my mind up entirely.

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4517,9 +4533,9 @@ ufunc_generic_fastcall(PyUFuncObject *ufunc,
             else {
                 tmp = Py_None;
             }
-            Py_INCREF(tmp);
-            PyTuple_SET_ITEM(full_args.out, i-nin, tmp);
+            ufunc_output[i-nin] = Py_NewRef(tmp);
         }
+        nout_args = nout;
```
seberg: Claude was right on this one, you have to update it within the loop, so that `goto fail` does the right thing (or before the `goto fail`, but doubt that is nicer).

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4446,61 +4460,63 @@ ufunc_generic_fastcall(PyUFuncObject *ufunc,
...
     /*
      * Scratch space for operands, dtypes, etc.  Note that operands and
      * operation_descrs may hold an entry for the wheremask.
+     * Also includes space for ufunc_input (nin slots, used by outer) and
+     * ufunc_output (nout slots).
      */
-    NPY_ALLOC_WORKSPACE(scratch_objs, void *, UFUNC_STACK_NARGS * 4 + 2, nop * 4 + 2);
+    NPY_ALLOC_WORKSPACE(scratch_objs, void *,
+                        UFUNC_STACK_NARGS * 4 + 2,
+                        nop * 4 + 2 + nop);
```
seberg: Might just write at as `nop * 5`, since it's just a + anyway?

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4606,8 +4621,18 @@ ufunc_generic_fastcall(PyUFuncObject *ufunc,
             if (out_obj == Py_Ellipsis) {
                 return_scalar = NPY_FALSE;
             }
-            else if (_set_full_args_out(nout, out_obj, &full_args) < 0) {
-                goto fail;
+            else {
+                ufunc_output = args_scratch + nin;
+                int n = _set_full_args_out(nout, out_obj, ufunc_output);
```
seberg: I might suggesting moving that `ufunc_output = NULL` logic into the the function (and maybe rename it).
Just because this is duplicated... I admit it getting set to NULL may be a tad cryptic, but then it can just keep returning 0 and -1 and is concise here?

One thing is that `nout_args` is really just `nout` _except_ when `ufunc_output == NULL` and some places even rely on that.

(The other nit is that the `_set_full_args_out` name is a bit outdated, but I don't really care.)

seberg (commented):
Asked claude to nitpick (it found a bug) and scrolled through and have two nitpicks on my own (claude some more, but even I thought they were overly nitpicky and/or unrelated).

Part of that "maybe `NULL`" and "may be owned" is annoying, but it seems fine.

Overall, I think I am getting happy and we could just go and put it in. It's not the prettiest, but the existing tuple isn't either really...

seberg:
I guess this keeps getting conflicts :(, want to try to push it over?  There is also some talk about whether we can get stable API working well enough and if not vital, this might actually be somewhat nice for that (as it removes some relatively hot-tuple building).

numpy/_core/src/multiarray/common.h
```diff
@@ -21,6 +21,23 @@ extern "C" {
 
 #define error_converting(x)  (((x) == -1) && PyErr_Occurred())
 
+static inline void
+multi_DECREF(PyObject *const *objects, npy_intp n)
+{
+    assert(n == 0 || objects != NULL);
```
mhvk, line 27: Make this the same as the one below?

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -3754,22 +3759,23 @@ _set_full_args_out(int nout, PyObject *out_obj, ufunc_full_args *full_args)
             return -1;
         }
         if (tuple_all_none(out_obj)) {
```
mhvk, line 3790: In principle, there might be a bit more speed gain to be had by doing this checking for all arguments being None on the fly in the loop below.

mhvk (approved):
@eendebakpt - looked back here again because of a comment. Two totally nitpicky comments if you still need to do a rebase anyway.

A question would be if it might not make more sense to keep in and out together in a single array (even if one has pointers to different parts), but, really, what you have is great and very clear, so the best thing is probably just to get it in!

numpy/_core/src/umath/ufunc_object.c
eendebakpt: In that way if all are None we then have an extra incref/decref pair for all arguments. For the common case (out not set) it does not matter either way.

mhvk: Fair enough, though if a tuple is present, it is nearly guaranteed not to be all-None.

eendebakpt:
> A question would be if it might not make more sense to keep in and out together in a single array (even if one has pointers to different parts), but, really, what you have is great and very clear, so the best thing is probably just to get it in!
The two are allocated together in `args_scratch` already. Are do you mean something else?

seberg:
Well, let me get this in before it diverges again (and probably annoying someone else with merge conflict :)).

seberg:
>  The two are allocated together in args_scratch already. Are do you mean something else?

Yeah, I think the question is more whether we could just drag around a single array for in+out.  Conceptually, I like that more, but I realize there were a few paths were you needed the distinction.

mhvk:
> The two are allocated together in `args_scratch` already. Are do you mean something else?

Ah, yes, I see now that for the regular ufunc call they indeed are! I think the reduction case was different, that's why this came up. Anyway, I like it how it is!

PR #31017, opened 2026-03-14
TYP: fix ``numpy.strings`` ufunc stubs

jorenham:
### PR summary

The comparison ufuncs in `numpy.strings` were annotated as functions rather than full-fledged ufuncs. This caused type-checkers to complain when doing valid things like `np.strings.equal.reduce(...)`. Since these ufuncs are the very same as the ones in the main numpy namespace, re-exporting them from there fixes it. 
Unfurtunately, the ufunc signatures are currently not  individually typable (I've hacked around this in numtype, but for numpy itself that's a bit too hacky to port over for my taste). This means that the call signatures of these fixed ufuncs aren't as specific anymore, and will no longer return a `NDArray[np.bool]`, but rather a `NDArray[Any]`. This shouldn't be a problem for backwards compatibility though, because of the [gradual guarantee](https://typing.python.org/en/latest/spec/concepts.html#the-gradual-guarantee).

#### AI Disclosure

I'm pretty sure I'm a human.


charris:
Thanks Joren.

PR #31021, opened 2026-03-16
DOC: document caveats of ndarray.resize on 3.14 and newer

### LINKED ISSUE #30991: BUG: ndarray.resize() always throws a ValueError with Python 3.14.3 and numpy 2.4.3
### Describe the issue:

I'm building Python 3.14.3 from scratch on a RHEL 8 workstation with GCC 12.2 and installing numpy 2.4.3 through PIP.  Simply trying to execute the example code in the `ndarray.resize()` documentation leads to a value error.

This seems related to https://github.com/numpy/numpy/pull/30278.

It appears that I'm still getting an additional reference count coming down the call stack and the new method being used (```PyUnstable_Object_IsUniquelyReferenced()```) is claiming it isn't unique, though I have not modified the numpy source to verify this is the case.

I'm building Python with:
```
installPrefix=/path/to/python
./configure --libdir=$installPrefix/Python-3.14.3/lib \
            --enable-shared \
            --enable-optimizations --with-lto \
            --prefix=$installPrefix/Python-3.14.3
```

```python
import numpy as np
a = np.array([[0, 1], [2, 3]], order='C')
a.resize((2, 1))
```

### Error message:

```shell
    a.resize((2, 1))
    ~~~~~~~~^^^^^^^^
ValueError: cannot resize an array that references or is referenced
by another array in this way.
Use the np.resize function or refcheck=False
```

2.4.3
3.14.3 (main, Mar 11 2026, 15:21:20) [GCC 12.2.0]

### Runtime Environment:

[{'numpy_version': '2.4.3',
  'python': '3.14.3 (main, Mar 11 2026, 15:21:20) [GCC 12.2.0]',
  'uname': uname_result(system='Linux', node='calvin.create.hpc.mil', release='4.18.0-553.104.1.el8_10.x86_64', version='#1 SMP Tue Feb 3 13:44:58 EST 2026', machine='x86_64')},
                      'found': ['X86_V3', 'X86_V4'],
                      'not_found': ['AVX512_ICL', 'AVX512_SPR']}},
 {'architecture': 'SkylakeX',
  'filepath': '/home/mcnally/Development/ptoolsrte/0.7.0/packages/Python-3.14.3/lib/python3.14/site-packages/numpy.libs/libscipy_openblas64_-ff84a88b.so',
  'internal_api': 'openblas',
  'num_threads': 64,
  'user_api': 'blas',
  'version': '0.3.31.dev'}]

Quite a bit of code that previously worked fine now generates errors.  This was found when upgrading our dependency stack to the Python and numpy versions listed above.  We were previously on Python 3.9 and numpy 1.25.

ngoldbaum:
Closes #30991.

This doesn't "fix" the regression, because I don't think there's a way to do that without introducing new bugs or depending on interpreter internals and/or trying to argue for yet another `PyUnstable` C API function we can use.

I'm also a little concerned that the old check is actually incorrect for arrays created by C extensions: the check on 3.13 and older allows resizing an array created directly via e.g. `PyArray_FromAny` with exactly one reference.

IMO it's better to keep the logic of the check the same and simply improve the documentation and error messages.

Now when the reference count is exactly 2, the error message on 3.14 and newer explains that this might be a false positive.

I added tests which trigger the behaviors we expect to happen on different Python versions. I also deleted some out-of-date docs about an unused `order` argument and added some notes that resizing in-place can lead to memory fragmentation and worse overall memory usage.

Also looking for any ideas for ways to make the error messages and docstrings less wordy.

### DIFF
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -2254,12 +2254,19 @@ Shape Manipulation
     a different total number of elements then the old shape. If reallocation is
     necessary, then *self* must own its data, have *self* - ``>base==NULL``,
     have *self* - ``>weakrefs==NULL``, and (unless refcheck is 0) not be
-    referenced by any other array.  The fortran argument can be
-    :c:data:`NPY_ANYORDER`, :c:data:`NPY_CORDER`, or
-    :c:data:`NPY_FORTRANORDER`.  It currently has no effect. Eventually it
-    could be used to determine how the resize operation should view the data
-    when constructing a differently-dimensioned array.  Returns None on success
-    and NULL on error.
+    referenced by any other array.
+
+    On Python 3.13 and older, the check allows objects with exactly one
+    reference to be resized, because it is impossible to differentiate between
+    an array with one reference created via an extension and a uniquely
+    referenced array defined in a Python function. On Python 3.14 and newer,
+    the array must be uniquely referenced.
+
+    Resizing arrays in-place can often lead to memory fragmentation and should
+    be avoided if the goal is to reclaim memory. Create a new array and copy the
+    relevant subset of data over instead.
+
+    Returns None on success and NULL on error.
 
 .. c:function:: PyObject* PyArray_Transpose( \
         PyArrayObject* self, PyArray_Dims* permute)
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -2254,11 +2254,12 @@ Shape Manipulation
     a different total number of elements then the old shape. If reallocation is
     necessary, then *self* must own its data, have *self* - ``>base==NULL``,
     have *self* - ``>weakrefs==NULL``, and (unless refcheck is 0) not be
-    referenced by any other array.
+    referenced by any other array. The *fortran* argument has no effect.
 
-    On Python 3.13 and older, the check allows objects with exactly one
-    reference to be reallocated in-place. On Python 3.14 and newer, the array
-    must be uniquely referenced. See the Python 3.14 `What's New entry
+    On Python 3.13 and older, the check allows uniquely referenced objects and
+    objects with exactly one reference to be reallocated in-place. On Python
+    3.14 and newer, the array must be uniquely referenced. See the Python 3.14
+    `What's New entry
     <https://docs.python.org/3/whatsnew/3.14.html#whatsnew314-refcount>`_ on
     this topic for more information on why there is a behavior difference.
 
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -2257,13 +2257,18 @@ Shape Manipulation
     referenced by any other array.
 
     On Python 3.13 and older, the check allows objects with exactly one
-    reference to be resized, because it is impossible to differentiate between
-    an array with one reference created via an extension and a uniquely
-    referenced array defined in a Python function. On Python 3.14 and newer,
-    the array must be uniquely referenced.
+    reference to be reallocated in-place. On Python 3.14 and newer, the array
+    must be uniquely referenced. See the Python 3.14 `What's New entry
+    <https://docs.python.org/3/whatsnew/3.14.html#whatsnew314-refcount>`_ on
+    this topic for more information on why there is a behavior difference.
 
-    Resizing arrays in-place can often lead to memory fragmentation and should
-    be avoided. If the goal is to reclaim over-allocated memory, alternatives are to create a view or a copy of just the desired data, or using two passes to build the array: one to cheaply determine the shape and another to allocate and fill. Benchmark your use case to determine what is optimum. You may be surprised to find ``resize`` actually slows down or bloats your application.
+    Reallocating arrays in-place can often lead to memory fragmentation and
+    should be avoided. If the goal is to reclaim over-allocated memory,
+    alternatives are to create a view or a copy of just the desired data, or
+    using two passes to build the array: one to cheaply determine the shape and
+    another to allocate and fill. Benchmark your use case to determine what is
+    optimum. You may be surprised to find ``resize`` actually slows down or
+    bloats your application.
 
     Returns None on success and NULL on error.
 
@@ -4267,9 +4272,9 @@ Memory management
 .. c:function:: int PyArray_ResolveWritebackIfCopy(PyArrayObject* obj)
 
     If ``obj->flags`` has :c:data:`NPY_ARRAY_WRITEBACKIFCOPY`, this function
-    clears the flags, `DECREF` s
-    `obj->base` and makes it writeable, and sets ``obj->base`` to NULL. It then
-    copies ``obj->data`` to `obj->base->data`, and returns the error state of
+    clears the flags, ``DECREF`` s
+    ``obj->base`` and makes it writeable, and sets ``obj->base`` to NULL. It then
+    copies ``obj->data`` to ``obj->base->data``, and returns the error state of
     the copy operation. This is the opposite of
     :c:func:`PyArray_SetWritebackIfCopyBase`. Usually this is called once
     you are finished with ``obj``, just before ``Py_DECREF(obj)``. It may be called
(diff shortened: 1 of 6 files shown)


ngoldbaum:
Marking as a draft while I consider Sam's suggestion: https://github.com/numpy/numpy/issues/30991#issuecomment-4070264086

doc/source/reference/c-api/array.rst
```diff
@@ -2254,12 +2254,19 @@ Shape Manipulation
...
-    could be used to determine how the resize operation should view the data
-    when constructing a differently-dimensioned array.  Returns None on success
-    and NULL on error.
+    referenced by any other array.
+
+    On Python 3.13 and older, the check allows objects with exactly one
+    reference to be resized, because it is impossible to differentiate between
+    an array with one reference created via an extension and a uniquely
+    referenced array defined in a Python function. On Python 3.14 and newer,
+    the array must be uniquely referenced.
```
mattip: I find this a bit confusing. Maybe "On Python 3.13 and older, the check allows objects with exactly one reference to be reallocated in-place. On Python 3.14 and newer, the array must be uniquely referenced, which is getting progressively more complicated to detect.

doc/source/reference/c-api/array.rst
```diff
@@ -2254,12 +2254,19 @@ Shape Manipulation
...
+
+    Resizing arrays in-place can often lead to memory fragmentation and should
```
mattip:
```suggestion
    Reallocating arrays in-place can often lead to memory fragmentation and should
```

doc/source/reference/c-api/array.rst
```diff
@@ -2254,12 +2254,19 @@ Shape Manipulation
...
+    be avoided if the goal is to reclaim memory. Create a new array and copy the
+    relevant subset of data over instead.
```
mattip:
```suggestion
    be avoided. If the goal is to reclaim over-allocated memory, alternatives are to create a view or a copy of just the desired data, or using two passes to build the array: one to cheaply determine the shape and another to allocate and fill. Benchmark your use case to determine what is optimum. You may be surprised to find ``resize`` actually slows down or bloats your application.
```

mattip (commented):
Thanks. I think this is useful even if it turns out we can work around CPython refcounting fragility. We may need to adjust the post-3.13 comment depending on the outcome of the other work.

numpy/_core/src/multiarray/shape.c
```diff
@@ -77,26 +77,40 @@ PyArray_Resize_int(PyArrayObject *self, PyArray_Dims *newshape, int refcheck)
...
             if (!PyUnstable_Object_IsUniquelyReferenced((PyObject *)self)) {
+                if (Py_REFCNT(self) == 2) {
+                    PyErr_SetString(
+                            PyExc_ValueError,
+                            "cannot resize an array that may be referenced "
+                            "by another object in this way.\n"
+                            "It is possible that the array is not referenced by "
+                            "another object and this is a false positive.\n"
+                            "If you are sure that the array is uniquely referenced, "
+                            "set refcheck=False to disable this check, otherwise use np.resize.");
```
mattip: A bit briefer?
```suggestion
                            "cannot resize an array that may be referenced "
                            "by another object.\n"
                            "It is possible that this is a false positive.\n"
                            "If you are sure that the array is uniquely referenced, "
                            "set refcheck=False.");
```

numpy/_core/src/multiarray/shape.c
```diff
@@ -175,6 +189,12 @@ PyArray_Resize_int(PyArrayObject *self, PyArray_Dims *newshape, int refcheck)
  * array and it is contiguous.  If refcheck is 0, then the reference count is
  * not checked and assumed to be 1.  You still must own this data and have no
  * weak-references and no base object.
+ *
+ * On Python 3.13 and older, the check allows objects with exactly one
+ * reference to be resized, because it is impossible to differentiate between
+ * an array with one reference created via an extension and a uniquely
+ * referenced array defined in a Python function.
```
mattip: Maybe like in the docs

On Python 3.13 and older, the check allows objects with exactly one reference to be reallocate in-place. On newer versions ...

numpy/_core/_add_newdocs.py
```diff
@@ -4190,12 +4185,25 @@ def _array_method_doc(name: str, params: str, doc: str) -> None:
     Only contiguous arrays (data elements consecutive in memory) can be
     resized.
 
+    Resizing arrays in-place can increase memory fragmentation. For that reason,
+    it is often preferable to allocate new memory for the result by calling
+    ``np.resize`` instead. This can reduce overall memory usage, even in
+    situations where one might expect to avoid wasting memory by resizing
+    in-place.
+
```
mattip, line 4196: Maybe rework this as suggested in the array.rst comment. Resizing -> Reallocating, and suggesting alternative strategies.

mattip (commented):
Thanks!

ngoldbaum:
I think I responded to all your comments @mattip

doc/source/reference/c-api/array.rst
```diff
@@ -2254,12 +2254,23 @@ Shape Manipulation
     a different total number of elements then the old shape. If reallocation is
     necessary, then *self* must own its data, have *self* - ``>base==NULL``,
     have *self* - ``>weakrefs==NULL``, and (unless refcheck is 0) not be
-    referenced by any other array.  The fortran argument can be
```
mhvk, line 2257: Should still note at the end that `fortran` has no effect.

numpy/_core/src/multiarray/shape.c
```diff
@@ -175,6 +188,17 @@ PyArray_Resize_int(PyArrayObject *self, PyArray_Dims *newshape, int refcheck)
  * array and it is contiguous.  If refcheck is 0, then the reference count is
  * not checked and assumed to be 1.  You still must own this data and have no
  * weak-references and no base object.
+ *
+ * On Python 3.13 and older, the check allows objects with exactly one
+ * reference to be reallocated in-place. On Python 3.14 and newer, the array
+ * must be uniquely referenced. In some cases this can lead to spurious
+ * ValueErrors on Python 3.14.
+ *
+ * On Python 3.13 and older, the check allows objects with exactly one
```
mhvk: Strange partial duplication with text above. Merge?

numpy/_core/tests/test_multiarray.py
```diff
@@ -6377,6 +6379,23 @@ def test_check_reference(self):
...
+            a = np.array([[0, 1], [2, 3]], order='C')
+            a.resize((2, 1))
+        """)
+        try:
+            subprocess.check_output([sys.executable, "-c", code],
+                                    stderr=subprocess.STDOUT, text=True)
+        except subprocess.CalledProcessError as e:
+            assert sys.version_info >= (3, 14)
+            assert "ValueError" in e.stdout
+            assert "It is possible that this is a false positive." in e.stdout
```
mhvk, line 6397: I think you want an
```
else:
    raise AssertionError("Unexpected success of resize check.")
```
(i.e., we should know if things start to succeed...)

mhvk (commented):
Yes, I think documenting behaviour is preferred over trying to make `.resize` work magically. Some small comments in-line.

ngoldbaum:
I think I responded to all review comments.

numpy/_core/src/multiarray/shape.c
```diff
@@ -175,6 +188,12 @@ PyArray_Resize_int(PyArrayObject *self, PyArray_Dims *newshape, int refcheck)
  * array and it is contiguous.  If refcheck is 0, then the reference count is
  * not checked and assumed to be 1.  You still must own this data and have no
  * weak-references and no base object.
+ *
+ * On Python 3.13 and older, the check allows objects with exactly one
+ * reference to be reallocated in-place. On Python 3.14 and newer, the array
+ * must be uniquely referenced. In some cases this can lead to spurious
+ * ValueErrors on Python 3.14.
```
mattip:
```suggestion
 * ValueErrors.
```

mattip (approved):
On tiny nit, but can be merged without it.

charris:
Thanks Nathan.

PR #18683, opened 2021-03-27
WIP: Make .item() and .astype(object) return NumPy scalars for datetime and timedelta

h-vetinari:
This is to avoid implicitly casting e.g. datetime64 to (python-)datetimes when
using operations - like `.astype(object)` - that are expected not to change the
type. Fixes #12550.

Co-Authored-By: Sebastian Berg <sebastian@sipsolutions.net>

--------------

The diff was [provided](https://github.com/numpy/numpy/issues/12550#issuecomment-780203343) by @seberg, but is likely incomplete (see link). Opening this PR to hopefully start iterating towards completion. Help welcome - I'll happily give push access to my fork for those who want to contribute. 

CC @jbrockmendel

Issue #7619, opened 2016-05-11
BUG: .item() on a 0-dimensional datetime64[ns] array yields an integer


shoyer:
This is obviously not useful:
```
In [4]: np.array(np.datetime64('2000-01-01', 'us')).item()
Out[4]: datetime.datetime(2000, 1, 1, 0, 0)

In [5]: np.array(np.datetime64('2000-01-01', 's')).item()
Out[5]: datetime.datetime(2000, 1, 1, 0, 0)

In [6]: np.array(np.datetime64('2000-01-01', 'ns')).item()
Out[6]: 946684800000000000
```

The underlying problem is that not all datetime64 types can be represented in `datetime.datetime` objects, which have fixed us precision.

We should either:

1. Raise `TypeError` in all cases when calling `.item()` on a datetime64[ns] array.
2. Convert to `datetime.datetime` if it can be done safely, raise `ValueError` otherwise.

My preference is for 2.

gerritholl:
The [behaviour on `float128`](https://github.com/numpy/numpy/issues/8547) is also peculiar.

OmerJog:
Is there any progress with fixing this issue?

eric-wieser:
@OmerJog: We can't "fix" it until we can come to a decision about what the correct behavior is.

h-vetinari:
> @eric-wieser: We can't "fix" it until we can come to a decision about what the correct behavior is.

Another option in addition to the OP:
3. don't return `datetime.datetime`, but rather a `np.datetime64`

I know this is a large breaking change, but it is still the best behaviour to follow IMO. `datetime.datetime` are not well-embedded nor compatible in many different (performance-relevant) ways with numpy/pandas etc. Just rip off the old crusty bandaid (which is not to say that the change cannot be carefully planned/managed/executed).

eric-wieser:
@h-vetinari's suggestion is not a terrible one, and is at least somewhat consistent with how `longdouble` behaves. On the other hand, there's little point calling `item` in the first place if you're happy with a numpy scalar.

Which of the following does the suggestion apply to?

* All datetime types
* All datetime types with higher precision than `datetime.datetime`
* Any datetime _value_ that falls out of range of a `datetime.datetime`

h-vetinari:
My suggestion applies to all datetime types**. And the ramifications (intentionally) don't stop at `.item` cf. #12550.

[...]

PR #31023, opened 2026-03-17
BUG/API: return numpy scalars instead of ints from datetime64/timedelta64.astype(object)

### LINKED ISSUE #12550: BUG: astype(object) downcasts for datetime-dtype
The dtype `np.object_` is the most general catch-all type to contain arbitrary python objects. As such, `.astype(object)` should preferably not upcast, but certainly never *down*cast:

### Reproducing code example:

```python
>>> import numpy as np
>>> arr = np.array([10 ** 18], dtype='M8[ns]')
>>> arr
array(['2001-09-09T01:46:40.000000000'], dtype='datetime64[ns]')
>>> arr[0]
numpy.datetime64('2001-09-09T01:46:40.000000000')
>>> arr.astype(object)
array([1000000000000000000], dtype=object)
>>> arr.astype(object)[0]
1000000000000000000
```

The expected outcome for the last two lines is:
```
>>> arr.astype(object)
array([numpy.datetime64('2001-09-09T01:46:40.000000000')], dtype=object)
>>> arr.astype(object)[0]
numpy.datetime64('2001-09-09T01:46:40.000000000')
```
as well as
```
>>> arr.astype(object)[0] == arr[0]
True
```

### Numpy/Python version information:

```
numpy: 1.15.4
python: 3.7.1
```

jbrockmendel:
- [x] closes #12550

cc @seberg 

Claude wrote this, I reviewed it manually.  But my C-fu is weak, so I'm skirting the bounds of socially-acceptable AI use.  My feelings won't be hurt if reviewers think I'm over the line.

TLDR: `dt64array.astype(object)[0]` give np.datetime64 objects instead of integers (the point of #12550).

Making `float(d64obj)` and `int(dt64obj)` raise is just because claude says its necessary to avoid recursion and a segfault.


mattip:
This is along the lines of [@seberg's comment](https://github.com/numpy/numpy/issues/12550#issuecomment-780203343), [your comment](https://github.com/numpy/numpy/issues/12550#issuecomment-1019592151) and #18683, glad to see it no longer segfaults/recurses. I don't really mind that you used AI as long as we can talk to a human, thanks for the honest disclosure. 

Also related to #7619. 

This definitely needs a release note and is quite a large change. We could sweep it under the rug as a "bugfix", but I think it is bigger than that. How can we best communicate the change to users? How would it affect Pandas?

seberg:
N.B.: https://github.com/numpy/numpy/issues/30985 makes a point that mixing Python and NumPy timedelta/datetime currently works a little bit sometimes on grounds of the path removed here.  Not sure it matters but maybe you have a quick thought @jbrockmendel?

But, I am positively surprised things work this much, the guard against recursion is correct (and maybe an acceptable regression, even if it returned the value for `ns` unit before, since that was converted via int previously; we could restore that behavior with a `DeprecationWarning` ideally, maybe...).

I am wondering if removing this means that we may need a ufunc(?) to convert, but raise an error probably if it fails?  OTOH, considering that `ns` is the default unit and it converts nonsensical maybe it doesn't matter?

(There is one weird(?) possibility, we could in theory make `.item()` try to convert to the Python `datetime`, but change the `.astype(object)` cast _specifically_ to stop doing this. But of course that would mean that `.item()` could fail, which seems strange.)

jbrockmendel:
> How would it affect Pandas?

Once our minimum-version catches up, it would let us rip out some special-casing we currently have to deal with the current behavior.

It's _possible_ to keep the current behavior for coarser resolutions and only change it for a) microsecond cases that are out of bounds for pydatetime and b) finer resolutions.  But thats an ugly enough API that i made a judgement call to handle all cases the same.  If going the other way is what it takes to get merged though, I'll change it.

jbrockmendel:
Thanks for your patience.  Just updated.

> [...] we could in theory make `.item()` try to convert to the Python datetime, [...]

I'd advocate against that since it would mean a different type depending on your resolution and whether you are OOB for stdlib datetime.

> I am wondering if removing this means that we may need a ufunc(?) to convert, but raise an error probably if it fails?

IIUC you're suggesting that `.astype(dt.datetime)` would convert to an object array of datetime objects?  I see the appeal, but I think that opens the door to a lot of other similar-if-you-squint cases.  Do you implement the same thing for decimal.Decimal?

> We could sweep it under the rug as a "bugfix", but I think it is bigger than that. How can we best communicate the change to users?

I don't have any bright ideas.  Could just wait for a major release I guess.

PR #31025, opened 2026-03-17
BUG: fix np.astype with copy=False for scalars

antareepsarkar:
### PR summary
resolves #29455 

Since a 0-D array is made for scalars, this PR checks whether the same array is returned or not; if true, then the same object is returned. Also includes tests.

This is my first time trying to modify C code, so I may have missed something.       

#### AI Disclosure
No AI or LLM was used for the PR.

### DIFF
--- a/numpy/_core/src/multiarray/scalartypes.c.src
+++ b/numpy/_core/src/multiarray/scalartypes.c.src
@@ -2201,7 +2201,7 @@ gentype_byteswap(PyObject *self, PyObject *args, PyObject *kwds)
  *
  * #name = take, getfield, put, repeat, tofile, mean, trace, diagonal, clip,
  *         std, var, sum, cumsum, prod, cumprod, compress, sort, argsort,
- *         round, argmax, argmin, max, min, any, all, astype, resize,
+ *         round, argmax, argmin, max, min, any, all, resize,
  *         reshape, choose, tobytes, copy, searchsorted, view,
  *         flatten, ravel, squeeze#
  */
@@ -2252,6 +2252,32 @@ static PyObject *
     return obj;
 }
 /**end repeat**/
+
+static PyObject *
+gentype_astype(PyObject *self, PyObject *args, PyObject *kwds)
+{
+    if (PyObject_IsInstance(self, (PyObject *)&PyVoidArrType_Type)) {
+        // guard for structured data types
+        return gentype_generic_method(self, args, kwds, "astype");
+    }
+    PyArray_Descr *from = PyArray_DescrFromScalar(self);
+    if (from == NULL) {
+        return NULL;
+    }
+    PyArray_Descr *to = NULL;
+    if (PyTuple_GET_SIZE(args) >= 1
+            && !PyArray_DescrConverter(PyTuple_GET_ITEM(args, 0), &to)) {
+        Py_DECREF(from);
+        return NULL;
+    }
+    int equiv = (to != NULL) && PyArray_EquivTypes(from, to) && Py_TYPE(self) == (PyTypeObject *)to->typeobj;
+    Py_DECREF(from);
+    Py_XDECREF(to);
+    if (equiv) {
+        return Py_NewRef(self);
+    }
+    return gentype_generic_method(self, args, kwds, "astype");
+}
 
 static PyObject *
 voidtype_getfield(PyVoidScalarObject *self, PyObject *args, PyObject *kwds)
--- a/numpy/_core/tests/test_scalar_methods.py
+++ b/numpy/_core/tests/test_scalar_methods.py
@@ -135,6 +135,36 @@ class TestIsInteger:
                 continue
             assert not value.is_integer()
 
+class TestAsType:
+    def test_copy(self) -> None:
+        a = np.float64(4.5)
+        b = a.astype(np.float64, copy=False)
+        assert a is b
+        b = np.astype(a, np.int64, copy=False)
+        assert id(a) != id(b)
+        b = a.astype(np.float64, copy=True)
+        assert a is b
+
+class TestAsTypeStructuredCopy:
+    # Regression test for gh-31025: astype on a structured scalar
+    # must return an independent object, not a mutable view into the parent array.
+    def test_copy_true_is_independent(self) -> None:
+        a = np.array([('AB',), ('CD',)], dtype=[('name', 'U5')])
+        b = a[1]
+
+        c = b.astype(b, copy=True)
+        assert c is not b
+        c['name'] = 'ZZ'
+        assert a[1]['name'] == 'CD'
+
+    def test_default_copy_is_independent(self) -> None:
+        # copy defaults to True, so the same guarantee holds without the kwarg
+        a = np.array([('AB',), ('CD',)], dtype=[('name', 'U5')])
+        b = a[1]
+
+        c = b.astype(b)
+        c['name'] = 'ZZ'
+        assert a[1]['name'] == 'CD'
 
 class TestClassGetItem:
     @pytest.mark.parametrize("cls", [
--- a/environment.yml
+++ b/environment.yml
@@ -24,7 +24,7 @@ dependencies:
   - pytest-xdist
   - hypothesis
   # For type annotations
-  - mypy==2.3.0
+  - mypy==2.3.1
   - orjson # makes mypy faster
   # For building docs
   - sphinx>=8.0
--- a/requirements/typing_requirements.txt
+++ b/requirements/typing_requirements.txt
@@ -3,5 +3,5 @@
 -r test_requirements.txt
 -r hypothesis_requirements.txt
 
-mypy==2.3.0
+mypy==2.3.1
 pyrefly==1.2.0


eendebakpt:
There is also https://github.com/numpy/numpy/pull/29643. I will have a look at both a bit later.

antareepsarkar:
> There is also #29643.

Actually, I did not know about it as it is not linked to the issue.

eendebakpt:
@antareepsarkar Thanks for working on this. Your PR addresses the bug, but has two issues:

* The `gentype_generic_method` is not only used for `astype` but many other methods (about 36). The `strcmp` would slow down all the other methods. This could be addressed by creating a dedicated method for `astype` (as in in the other PR)
* Even though the same element is returned, still a copy is made (e.g. the `PyObject_CallObject`/`PyObject_Call` is executed)

If you just want to fix the bug, I suggest we change the documentation to make the statement in the docstring a bit more relaxed.

Even better would be the fix the issue and make performance a bit better. For that you can continue along the lines of the other PR (the author is ok with continuing from there, see https://github.com/numpy/numpy/pull/29643#issuecomment-3862694926)

antareepsarkar:
> Even better would be the fix the issue and make performance a bit better. For that you can continue along the lines of the other PR (the author is ok with continuing from there, see [#29643 (comment)](https://github.com/numpy/numpy/pull/29643#issuecomment-3862694926))

I think that I should only fix the bug (ofcourse by declaring a separate method) . I do not want to do the improvement because I think I can leave something behind while implementing it. Again, the idea came from you, so you will be able to do it better.

I think if we just fix the bug, there will be no regression.

antareepsarkar:
> Even better would be the fix the issue and make performance a bit better. For that you can continue along the lines of the other PR (the author is ok with continuing from there, see [#29643 (comment)](https://github.com/numpy/numpy/pull/29643#issuecomment-3862694926))

I think I should only fix the bug (through a new method). Since, the improvement idea came from you, you will be able to do it better.

I think that fixing the bug will not create a regression so it should be ok.

eendebakpt:
@antareepsarkar I think the correct way to solve the bug is via the other PR (which both fixes the bug and improves performance). If you want to pick that up, go ahead! If not I will pick it up

antareepsarkar:
> @antareepsarkar I think the correct way to solve the bug is via the other PR (which both fixes the bug and improves performance). If you want to pick that up, go ahead! If not I will pick it up

Okay, give me some time. I will do something to also improve along the lines of the other PR.

antareepsarkar:
Is the test failure related?

numpy/_core/tests/test_scalar_methods.py
```diff
@@ -135,6 +135,24 @@ def test_false(self, code: str) -> None:
                 continue
             assert not value.is_integer()
 
+class TestAsType:
+    @pytest.mark.parametrize("val", [3, 4, 5])
```
eendebakpt, line 139: Is the parameterization needed? (the values are nothing special, so just 3 would be enough)

numpy/_core/tests/test_scalar_methods.py
```diff
@@ -135,6 +135,24 @@ def test_false(self, code: str) -> None:
...
+    def test_copy(self, val: type[np.number]) -> None:
+        a = np.float64(val)
+        b = a.astype(np.float64, copy=False)
+        assert id(a) == id(b)
+        b = np.astype(a, np.int64, copy=False)
+        assert id(a) != id(b)
+        b = a.astype(np.float64, copy=True)
+        assert id(a) != id(b)
```
eendebakpt, line 147: Could we instead update the documentation? Since float scalars are immutable we can just return the same object as an optimization. This is also what python does:

```
import copy
x = 1.23
id(copy.copy(x)) == id(x) # True
```

numpy/_core/src/multiarray/scalartypes.c.src
```diff
@@ -2253,6 +2253,42 @@ static PyObject *
 }
 /**end repeat**/
 
+static PyObject *
+gentype_astype(PyObject *self, PyObject *args, PyObject *kwds)
+{
+    int copy = 0;
```
eendebakpt, line 2259: The implementation is quite complex now. How about the approach of #29643 with `PyArray_DescrFromScalar` replaced by `PyArray_DescrConverter`?

antareepsarkar: I will try to implement it.

numpy/_core/tests/test_scalar_methods.py
antareepsarkar: No, not needed actually. Will get rid of this.

numpy/_core/tests/test_scalar_methods.py
antareepsarkar: Not sure. But, probably this can be done.

antareepsarkar:
I will change the things and try to make it better.

antareepsarkar:
I have changed the things.

I can't use `PyArray_DescrConverter` instead of `PyArray_DescrFromScalar`.

The thing is, `PyArray_DescrConverter` cannot convert `np.str_("abc")` or things like that. It is good at converting data types.

For example:
```py
a = np.array('abc', (np.str_, 2))
a[()].astype(np.bool)
```
will error with `PyArray_DescrConverter`.

After this change, immutables do not return a copy if the data types match.

antareepsarkar:
@eendebakpt 
And, sorry for the long delay. From now, I will make changes quickly.

numpy/_core/src/multiarray/scalartypes.c.src
```diff
@@ -2253,6 +2253,38 @@ static PyObject *
...
+static PyObject *
+gentype_astype(PyObject *self, PyObject *args, PyObject *kwds)
+{
+    PyObject *temp = NULL;
+    if (kwds != NULL){
+        PyDict_GetItemStringRef(kwds, "copy", &temp);
+    }      
+    PyArray_Descr *dtype1 = PyArray_DescrFromScalar(self);
+    assert(dtype1 != NULL);
+    if((PyDataType_NAMES(dtype1) != NULL) && (temp == NULL)){
```
eendebakpt: This skips the guard of `copy` is set, which can lead to issues.

I think he following fails for the current PR
```
class TestAsTypeStructuredCopy:
    # Regression test for gh-31025: astype on a structured scalar
    # must return an independent object, not a mutable view into the parent array.
    def test_copy_true_is_independent(self) -> None:
        a = np.array([('AB',), ('CD',)], dtype=[('name', 'U5')])
        b = a[1]

        c = b.astype(b, copy=True)
        assert c is not b                 # must be a real copy
        c['name'] = 'ZZ'
        assert a[1]['name'] == 'CD'       # parent must be untouched

    def test_default_copy_is_independent(self) -> None:
        # copy defaults to True, so the same guarantee holds without the kwarg
        a = np.array([('AB',), ('CD',)], dtype=[('name', 'U5')])
        b = a[1]

        c = b.astype(b)
        c['name'] = 'ZZ'
        assert a[1]['name'] == 'CD'
```

Something like the following might work
```
static PyObject *
gentype_astype(PyObject *self, PyObject *args, PyObject *kwds)
{
    if (PyObject_IsInstance(self, (PyObject *)&PyVoidArrType_Type)) {
        return gentype_generic_method(self, args, kwds, "astype");
    }
    PyArray_Descr *from = PyArray_DescrFromScalar(self);
    if (from == NULL) {
        return NULL;
    }
    PyArray_Descr *to = NULL;
    if (PyTuple_GET_SIZE(args) >= 1
            && !PyArray_DescrConverter(PyTuple_GET_ITEM(args, 0), &to)) {
        Py_DECREF(from);
        return NULL;
    }
    int equiv = (to != NULL) && PyArray_EquivTypes(from, to);
    Py_DECREF(from);
    Py_XDECREF(to);
    if (equiv) {
        return Py_NewRef(self);
    }
    return gentype_generic_method(self, args, kwds, "astype");
}
```

antareepsarkar:
Changes done.

numpy/_core/src/multiarray/scalartypes.c.src
```diff
@@ -2253,6 +2253,32 @@ static PyObject *
 }
 /**end repeat**/
 
+static PyObject *
+gentype_astype(PyObject *self, PyObject *args, PyObject *kwds)
```
eendebakpt: I have some more corner cases.

* `type(np.longdouble(1).astype(np.float64))` is `np.float64` on Windows. With the PR it is `np.longdouble` I think. Maybe fixed by adding a check `Py_TYPE(self) == (PyTypeObject *)to->typeobj`.
* What about:
```
class MyFloat64(np.float64):
    pass

type(MyFloat64(1.5).astype(np.float64))
type(MyFloat64(1.5).astype(np.float64, subok=False))
```

numpy/_core/tests/test_scalar_methods.py
```diff
@@ -135,6 +135,36 @@ def test_false(self, code: str) -> None:
                 continue
             assert not value.is_integer()
 
+class TestAsType:
+    def test_copy(self) -> None:
+        a = np.float64(4.5)
+        b = a.astype(np.float64, copy=False)
+        assert id(a) == id(b)
```
eendebakpt:
```suggestion
        assert a is b
```

eendebakpt (commented):
This turns out to be tricky to get right. Maybe we end up with too many checks to make this worthwhile, but in any case well get some better tests

antareepsarkar:
Changes done.


```
class MyFloat64(np.float64):
    pass

type(MyFloat64(1.5).astype(np.float64))
type(MyFloat64(1.5).astype(np.float64, subok=False))
```
For the above example:
The `main` branch (without the changes of the PR) returns the same thing as this PR does, thus this PR changes nothing.
Changing this would require a release note, as it changes the way `astype` works.
Thus, it's better to leave it.

And, the Circle CI error does not seem to come from the changes.

numpy/_core/src/multiarray/scalartypes.c.src
```diff
@@ -2253,6 +2253,32 @@ static PyObject *
...
+    if (PyTuple_GET_SIZE(args) >= 1
+            && !PyArray_DescrConverter(PyTuple_GET_ITEM(args, 0), &to)) {
+        Py_DECREF(from);
+        return NULL;
+    }
+    int equiv = (to != NULL) && PyArray_EquivTypes(from, to) && Py_TYPE(self) == (PyTypeObject *)to->typeobj;
+    Py_DECREF(from);
+    Py_XDECREF(to);
+    if (equiv) {
+        return Py_NewRef(self);
```
eendebakpt: This can return without inspecting `kwds`. That means that `a.astype(np.float64, bogus=3)` can now return, where before it would raise an error.

numpy/_core/tests/test_scalar_methods.py
```diff
@@ -135,6 +135,36 @@ def test_false(self, code: str) -> None:
                 continue
             assert not value.is_integer()
 
+class TestAsType:
+    def test_copy(self) -> None:
+        a = np.float64(4.5)
+        b = a.astype(np.float64, copy=False)
+        assert a is b
+        b = np.astype(a, np.int64, copy=False)
+        assert id(a) != id(b)
```
eendebakpt:
```suggestion
        assert a != b
```

numpy/_core/src/multiarray/scalartypes.c.src
```diff
@@ -2253,6 +2253,32 @@ static PyObject *
 }
 /**end repeat**/
 
+static PyObject *
+gentype_astype(PyObject *self, PyObject *args, PyObject *kwds)
+{
+    if (PyObject_IsInstance(self, (PyObject *)&PyVoidArrType_Type)) {
```
eendebakpt: `PyObject_IsInstance` can return -1 with an exception set. Maybe use `PyArray_IsScalar`?

eendebakpt (commented):
@antareepsarkar The code here is more complex than I anticipated. I would suggest to keep only the tests, but skip the optimization for now. 

Historically there have been reasons to do this optimization, but I suspect the performance  gain is smaller on recent main branch (it would have to be benchmarked in any case).

numpy/_core/src/multiarray/scalartypes.c.src
eendebakpt: Or skip the guard here, and add `&& from->type_num != NPY_VOID` to the line `int equiv = ...` below.

PR #31028, opened 2026-03-18
DOC: Small darkmode theme fixes

neofelis2X:
### PR summary
I noticed some irregularities when switching the documentation to dark mode. This PR should fix them.

Changes:
- Index card icons now have correct color and background in dark mode
- Index card icons SVGs have been cleaned up
  - two of them contained embedded pixel data instead of curves
  - two of them had different margins then the other two
  - one of them was not centered
- Theme switcher now has correct font-color in dark mode
- The `apple-touch-icon.png` was blurry and has been re-exported

That's all errors I noticed. If you know any other problems with the theme / styling, please let me know and I will have a look at it.

#### Screenshot
new / old

Screenshot 2026-03-17 at 23 04 09

Screenshot 2026-03-17 at 23 04 27

Hi, I'm a designer and hobby-coder. When I don't write in C, numpy is probably my favourite python library, mostly in combination with OpenGL or for working with 3D data. I have done some open source contributions before, usually visual, graphic and ui related stuff.

#### AI Disclosure
No AI was used in any way or form.


seberg (approved):
LGTM, and all fine.  I am not sure why the svgs now have an `xml` comment, but doesn't matter either way (fixing them is great!).

I don't recall if there was a specific reason for the old settings, but I doubt we discussed it much so unless anyone else has a different opinion, should just put it in soon.

melissawm (approved):
Looks much better, thanks @neofelis2X !

melissawm:
The svg headers look like they could be different - I did some research and Google tells me what we had before is actually the recommended/modern format, and this PR introduces a strict DTD header. I don't think it changes much for us or for regular browsers but just wanted to throw the info out there.

neofelis2X:
The xml declaration is just the standard form how most applications export svgs. (Adobe Illustrator, Affinity Designer, Inkscape) They are xml files after all. The browser doesn't need the xml declaration, hence why "svg optimisers" usually remove it.
If you want I can run them through an optimiser, which would also make them 50-70% smaller.

melissawm:
I think that would be great @neofelis2X , I'd appreciate it thanks!

seberg:
Awesome, thanks for the followup even!

NitinSomisetty:
Hi, I noticed the live docs at numpy.org/doc/stable still show the old CSS with `color: black` on the version switcher, causing it to be illegible in dark mode. Is there a plan to redeploy the docs with this fix? Happy to help if needed!

mattip:
stable is the 2.4 version, we don't typically rebuild docs after the 2.4.0 release.

melissawm:
I can fix manually just for accessibility purposes.

PR #31032, opened 2026-03-19
BUG: Allow explicit `out=None`  for `ufunc` overrides.

roytsmart:
### PR summary

This PR fixes #31030 by not deleting `out` at the end of `initialize_normal_kwds()`.

### DIFF
--- a/doc/release/upcoming_changes/31032.improvement.rst
+++ b/doc/release/upcoming_changes/31032.improvement.rst
@@ -1,0 +1,8 @@
+Allow explicit `out=None` for `ufunc` overrides if the `where` keyword is present
+---------------------------------------------------------------------------------
+Numpy now emits the warning:
+``UserWarning: 'where' used without 'out', expect unitialized memory in output.
+If this is intentional, use out=None.``
+
+This change allows `out=None` to be propagated through ``__array_ufunc__()``
+if the `where` keyword is also present.
--- a/numpy/_core/tests/test_umath.py
+++ b/numpy/_core/tests/test_umath.py
@@ -3663,17 +3663,17 @@ class TestSpecialMethods:
         res = np.multiply.reduce(a, None, out=(None,), dtype=None)
         assert_equal(res[4], {'axis': None, 'dtype': None})
         res = np.multiply.reduce(a, 0, None, None, False, 2, True)
-        assert_equal(res[4], {'axis': 0, 'dtype': None, 'keepdims': False,
-                              'initial': 2, 'where': True})
+        assert_equal(res[4], {'axis': 0, 'dtype': None, 'out': None,
+                              'keepdims': False, 'initial': 2, 'where': True})
         # np._NoValue ignored for initial
         res = np.multiply.reduce(a, 0, None, None, False,
                                  np._NoValue, True)
-        assert_equal(res[4], {'axis': 0, 'dtype': None, 'keepdims': False,
-                              'where': True})
+        assert_equal(res[4], {'axis': 0, 'dtype': None, 'out': None,
+                              'keepdims': False, 'where': True})
         # None kept for initial, True for where.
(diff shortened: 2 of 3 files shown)


seberg:
Hmmm, until now we explicitly protected from it, so this might break someones `__array_ufunc__` implementation in theory (I haven't looked if any of the big ones -- e.g.  dask/astropy/cupy would notice).

But, I can't think of a different way to ensure the warning isn't given.  But we could at least make it conditional on `where` to (very much) reduce the potential blast radius (since `where=` isn't used a lot).

Can you also check if the docs need to be updated about this?  And maybe add a brief release note.

EDIT: Also clearly, the tests are failing, and need to be adapted to test whatever new behavior we have.

roytsmart:
@ngoldbaum, do you know why the tests are failing? I thought `where` would be in `normal_kwds`, but the test failure seems to indicate that it's not, unless I'm making a silly mistake.

seberg:
I think the reason for the failure is that `out` is passed by position and not kwarg, so it is never part of the `dict` from the start.  I dunno if it matters much, I guess you would instead have to add `out=None` to the kwargs (but then if there are multiple out should it be `(None, None)`?
(I would be tempted to pass in the info that `where` is passed, as it is available, but more of an opinion. Just to avoid the `Contains`)

doc/release/upcoming_changes/31032.improvement.rst
```diff
@@ -0,0 +1,8 @@
+Allow explicit `out=None` for `ufunc` overrides if the `where` keyword is present
+---------------------------------------------------------------------------------
+Numpy now emits the warning:
+``UserWarning: 'where' used without 'out', expect unitialized memory in output.
```
tylerjereddy, line 4: `unitialized ==> uninitialized`, but I suppose the spelling also exists in the actual warning issued by NumPy, so maybe not.

ngoldbaum: 🙃 

We should probably fix that typo in the warning text too

roytsmart: You don't think that's appropriate to do in this PR, right?

ngoldbaum: You could do it here. We'll probably want to backport this and it will make that a little easier.

ngoldbaum:
One thing that might also help is to debug a local build of numpy with gdb or lldb:

https://numpy.org/devdocs/dev/development_advanced_debugging.html#c-debuggers

I'll try to take a closer look to understand what's going wrong. Also ping @mattip - this is fixing some fallout from merging #29813 for projects that implement `__array_ufunc__`.

roytsmart:
Thanks @ngoldbaum, I will get that set up.

I'm surprised `wheremask_obj == NULL` didn't work, I need to read the code further to determine how the override knows if `where` is passed,.

numpy/_core/tests/test_umath.py
```diff
@@ -3663,17 +3663,17 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
         res = np.multiply.reduce(a, None, out=(None,), dtype=None)
         assert_equal(res[4], {'axis': None, 'dtype': None})
         res = np.multiply.reduce(a, 0, None, None, False, 2, True)
```
mhvk, line 3665: In this test and the ones below, both `out` and `where` are passed as positional parameters, so it seems correct that a warning would later be emitted.

But see my more general comment.

mhvk (commented):
@roytsmart - I think the test failures actually make some sense given what the warning was trying to do, but worry this is the wrong approach, as `__array_ufunc__` implementations will expect `out` to be a tuple. See my comment at https://github.com/numpy/numpy/issues/31030#issuecomment-4103153283

PR #31035, opened 2026-03-19
BUG: fix FNV-1a 64-bit selection by using NPY_SIZEOF_UINTP

### LINKED ISSUE #31034: BUG: 64-bit FNV-1a path is never used due to undefined NPY_SIZEOF_SIZE_T
### Describe the issue:

In the FNV-1a hash implementation, the 64-bit variant is never selected.

In the following code:

https://github.com/numpy/numpy/blob/621413ba39caa072d4e1b9e0e8ca78ba434a6f01/numpy/_core/src/multiarray/fnv.c#L73-L85

the selection between 32-bit and 64-bit implementations depends on `NPY_SIZEOF_SIZE_T`. However, this macro does not appear to be defined, so the condition always falls back to the 32-bit version (`npy_fnv1a_32`).

As a result, even on 64-bit platforms, the 64-bit FNV-1a implementation is never used.

Since `npy_intp` is typically defined to match `size_t`, it may be more appropriate to use `NPY_SIZEOF_INTP` instead.

```python
import numpy as np

# This issue is in the C implementation and cannot be directly reproduced
# from Python. It can be observed by inspecting the compiled code path
# or preprocessing output, where NPY_SIZEOF_SIZE_T is not defined and
# the 32-bit path is always selected.
```

### Error message:

```shell

```

2.5.0.dev0+git20260319.b9ccff1
3.14.3 (main, Feb  3 2026, 15:32:20) [Clang 17.0.0 (clang-1700.6.3.2)]

### Runtime Environment:

_No response_

_No response_

math-hiyoko:
### PR summary

close #31034

<details><summary>benchmark result</summary>

| Change   | Before [621413ba] <main>   | After [b9ccff1e] <fix_size_t>   |   Ratio | Benchmark (Parameter)                                                                |
|----------|----------------------------|---------------------------------|---------|--------------------------------------------------------------------------------------|
| -        | 807±30μs                   | 746±10μs                        |    0.92 | bench_lib.Unique.time_unique_values(200000, 90.0, 0.2, <class 'numpy.float64'>)      |
| -        | 1.64±0.02ms                | 1.52±0.01ms                     |    0.92 | bench_lib.UniqueIntegers.time_unique_values(100000, 5000, <class 'numpy.int64'>)     |
| -        | 1.26±0.01ms                | 1.16±0.02ms                     |    0.92 | bench_lib.UniqueIntegers.time_unique_values(100000, 5000, <class 'numpy.uint32'>)    |
| -        | 14.8±0.1ms                 | 13.6±0.1ms                      |    0.92 | bench_lib.UniqueIntegers.time_unique_values(1000000, 5000, <class 'numpy.int64'>)    |
| -        | 3.38±0.05μs                | 3.10±0.03μs                     |    0.92 | bench_lib.UniqueIntegers.time_unique_values(200, 25, <class 'numpy.int64'>)          |
| -        | 1.04±0.01ms                | 936±9μs                         |    0.9  | bench_lib.UniqueIntegers.time_unique_values(100000, 5000, <class 'numpy.int16'>)     |
| -        | 9.87±0.07ms                | 8.75±0.07ms                     |    0.89 | bench_lib.UniqueIntegers.time_unique_values(1000000, 5000, <class 'numpy.int16'>)    |
| -        | 11.8±0.1ms                 | 10.5±0.09ms                     |    0.89 | bench_lib.UniqueIntegers.time_unique_values(1000000, 5000, <class 'numpy.uint32'>)   |
| -        | 37.1±0.1ms                 | 29.4±0.1ms                      |    0.79 | bench_lib.UniqueIntegers.time_unique_values(1000000, 250000, <class 'numpy.int64'>)  |
| -        | 29.5±0.2ms                 | 22.6±0.2ms                      |    0.77 | bench_lib.UniqueIntegers.time_unique_values(1000000, 250000, <class 'numpy.uint32'>) |

</details>

### DIFF
--- a/numpy/_core/src/multiarray/fnv.c
+++ b/numpy/_core/src/multiarray/fnv.c
@@ -21,7 +21,7 @@
   https://github.com/lcn2/fnv/blob/b7fcbee95538ee6a15744e756e7e7f1c02862cb0/hash_32a.c
 */
 npy_uint32 
-npy_fnv1a_32(const void *buf, size_t len, npy_uint32 hval)
+npy_fnv1a_32(const void *buf, npy_intp len, npy_uint32 hval)
 {
     const unsigned char *bp = (const unsigned char *)buf;  /* start of buffer */
     const unsigned char *be = bp + len;                    /* beyond end of buffer */
@@ -48,7 +48,7 @@ npy_fnv1a_32(const void *buf, size_t len, npy_uint32 hval)
   https://github.com/lcn2/fnv/blob/b7fcbee95538ee6a15744e756e7e7f1c02862cb0/hash_64a.c
 */
 npy_uint64 
-npy_fnv1a_64(const void *buf, size_t len, npy_uint64 hval)
+npy_fnv1a_64(const void *buf, npy_intp len, npy_uint64 hval)
 {
     const unsigned char *bp = (const unsigned char *)buf;  /* start of buffer */
     const unsigned char *be = bp + len;                    /* beyond end of buffer */
@@ -71,15 +71,15 @@ npy_fnv1a_64(const void *buf, size_t len, npy_uint64 hval)
 }
 
 /*
- * Compute a size_t FNV-1a hash of the given data
- * This will use 32-bit or 64-bit hash depending on the size of size_t
+ * Compute a npy_intp FNV-1a hash of the given data
+ * This will use 32-bit or 64-bit hash depending on the size of npy_intp
  */
-size_t 
-npy_fnv1a(const void *buf, size_t len)
+npy_intp 
+npy_fnv1a(const void *buf, npy_intp len)
 {
-#if NPY_SIZEOF_SIZE_T == 8
-    return (size_t)npy_fnv1a_64(buf, len, FNV1A_64_INIT);
-#else /* NPY_SIZEOF_SIZE_T == 4 */
-    return (size_t)npy_fnv1a_32(buf, len, FNV1A_32_INIT);
+#if NPY_SIZEOF_INTP == 8
+    return (npy_intp)npy_fnv1a_64(buf, len, FNV1A_64_INIT);
+#else /* NPY_SIZEOF_INTP == 4 */
+    return (npy_intp)npy_fnv1a_32(buf, len, FNV1A_32_INIT);
 #endif
 }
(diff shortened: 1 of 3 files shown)


numpy/_core/src/multiarray/fnv.c
```diff
@@ -71,15 +71,15 @@ npy_fnv1a_64(const void *buf, size_t len, npy_uint64 hval)
...
-size_t 
-npy_fnv1a(const void *buf, size_t len)
+npy_intp 
+npy_fnv1a(const void *buf, npy_intp len)
 {
-#if NPY_SIZEOF_SIZE_T == 8
-    return (size_t)npy_fnv1a_64(buf, len, FNV1A_64_INIT);
-#else /* NPY_SIZEOF_SIZE_T == 4 */
-    return (size_t)npy_fnv1a_32(buf, len, FNV1A_32_INIT);
+#if NPY_SIZEOF_INTP == 8
```
seberg: FWIW, I think it's fine to only change this and keep `size_t` throughout.  (can add a comment that intp is the same size as `size_t`).
`uintp` feels a bit more natural, but it doesn't matter much.  Some of the calculations below also seem unrelated, but I guess they also don't really matter (since they don't seem to generate new warnings).

(if this was C++, constexpr would be nice I guess, but dunno it's worth to change that.)

Anyway, this also looks completely fine, though.

math-hiyoko: Thanks for the review.

I updated the condition to use `NPY_SIZEOF_UINTP` and reverted the unrelated changes.

ngoldbaum:
I don't think the loongarch64 or ppc64le failures are real.

That said, I think maybe the cp312-macosx_286_64 failure is real, maybe?

```
delocate.libsana.DelocationError: Library dependencies do not satisfy target MacOS version 10.13:
  /private/var/folders/9n/h1d0pgxn4f5bbpg9mrrft9140000gn/T/tmpyq6zeudq/wheel/numpy/.dylibs/libunwind.1.0.dylib has a minimum target of 14.0
  Set the environment variable 'MACOSX_DEPLOYMENT_TARGET=14.0' to update minimum supported macOS for this wheel.
```

Or maybe you just got unlucky and this started breaking on your PR for some reason.

Can you try rebasing on current `main` to see if it goes away?

Sorry you're hitting issues in our CI. I think this change looks correct to me and great catch - nice to see such huge speedups!

ngoldbaum:
Also maybe add a release note for the performance improvement? No need to if you don't want to though.

ngoldbaum:
> That said, I think maybe the cp312-macosx_286_64 failure is real, maybe?

nah, it's happening on other PRs. This is unrelated. Sorry our CI is so flaky!

ngoldbaum (approved):

seberg:
Fun that we have SIZEOF_UINTP, but not for most types... but since we have it, LGTM, thanks.

PR #31040, opened 2026-03-20
BUG: fix np.sinc(np.inf)

antareepsarkar:
### PR summary
resolves #31027 

#### AI Disclosure
No AI or LLM use.

### DIFF
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -3734,8 +3734,9 @@ def sinc(x):
     Return the normalized sinc function.
 
     The sinc function is equal to :math:`\sin(\pi x)/(\pi x)` for any argument
-    :math:`x\ne 0`. ``sinc(0)`` takes the limit value 1, making ``sinc`` not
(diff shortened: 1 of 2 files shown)


numpy/lib/_function_base_impl.py
```diff
@@ -3810,7 +3811,8 @@ def sinc(x):
     # Hope that 1e-20 is sufficient for objects...
     eps = np.finfo(x.dtype).eps if x.dtype.kind == "f" else 1e-20
     y = where(x, x, eps)
-    return sin(y) / y
+    z = where(abs(y) != np.inf, y, 0)
```
ngoldbaum, line 3814: I don't think we want to add a second call to `where` here that all callers will need to pay for just to handle this corner case.

Did you try benchmarking? I bet this hurts performance.

antareepsarkar: Yes, I understand. I also wanted to not include calls to expensive functions like `where`, but thought that since one call is made for a corner case, another would be just fine.

Okay, I get it, will try something else.

PR #31043, opened 2026-03-21
ENH: add dense integer fast path for np.unique counts/inverse

AsymptotaX:
### PR summary
This PR adds a dense-range fast path to `np.unique` for 1-D integer/bool arrays with `return_counts=True` and/or `return_inverse=True` (when `return_index=False`).

Instead of always using the generic sort-based path (`O(n log n)`), it uses a bincount-based counting table (`O(n + k`), where `k` is the covered value range) when a heuristic determines that the range is small relative to the input size. This preserves existing semantics while significantly speeding up repeated-label workloads such as factorization, grouping, and frequency counting.

The optimization has `O(k)` temporary memory usage and is guarded to avoid excessive allocation. All unsupported or non-beneficial cases fall back to the existing implementation.

### Env
mac mini m4
Python 3.13

### Benchmarks (1e6 elements, dense integer labels)

| Operation | Old (s) | New (s) | Speedup |
|----------|--------:|--------:|--------:|
| groupby (runtime) | 0.032371 | 0.003108 | 10.4× |
| groupby (tree)    | 0.032300 | 0.003116 | 10.4× |
| counts (runtime)  | 0.006392 | 0.001388 | 4.6×  |
| counts (tree)     | 0.006341 | 0.001415 | 4.5×  |

### DIFF
--- a/numpy/lib/_arraysetops_impl.py
+++ b/numpy/lib/_arraysetops_impl.py
@@ -136,6 +136,64 @@ def _unpack_tuple(x):
         return x
 
 
+def _unique1d_dense_ints(
+    ar, return_inverse=False, return_counts=False, *, inverse_shape=None
+):
+    """
+    Fast path for dense integer/bool arrays when `return_index` is not needed.
+    """
+    if ar.size == 0 or ar.dtype.kind not in "uib":
+        return NotImplemented
+
+    ar_min = int(np.min(ar))
+    ar_max = int(np.max(ar))
+    ar_range = ar_max - ar_min
+
+    # `bincount` allocates one slot for every value in the covered range.
+    if ar_range < 0 or ar_range > np.iinfo(np.intp).max:
+        return NotImplemented
+    if ar_range > 6 * ar.size:
+        return NotImplemented
+
+    ar_min = ar.dtype.type(ar_min)
+
+    if ar_range == 0:
+        uniq = np.array([ar_min], dtype=ar.dtype)
+        ret = (uniq,)
+        if return_inverse:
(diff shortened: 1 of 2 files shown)


mattip:
Please show the benchmarks you used, and also run the benchmark suite.

AsymptotaX:
#### Benchmark command:

```bash
cd benchmarks
asv continuous main HEAD -b bench_lib.UniqueIntegers.time_unique_counts
``` 
shows a small number of regressions on very small inputs, while dense medium/large integer cases improve substantially.

#### Selected regressions

| Scenario | Before | After | Ratio |
|---|---:|---:|---:|
| N=200, k=125, int16 | 4.97 μs | 6.63 μs | 1.33x |
| N=200, k=25, int16 | 4.83 μs | 6.40 μs | 1.33x |
| N=1e6, k=25, int16 | 1.47 ms | 1.89 ms | 1.28x |

#### Selected improvements

| Scenario | Before | After | Ratio |
|---|---:|---:|---:|
| N=1e6, k=25, uint32 | 2.76 ms | 2.06 ms | 0.75x |
| N=100k, k=125, int16 | 211 μs | 138 μs | 0.65x |
| N=1e6, k=5000, int64 | 12.4 ms | 2.36 ms | 0.19x |
| N=1e6, k=125, uint8 | 21.0 ms | 1.65 ms | 0.08x |

Possible follow-up: if we want to address the regression cases, we could add a small-size cutoff for `_unique1d_dense_ints`. The regressions appear concentrated in cases where fast-path setup cost dominates.

PR #31045, opened 2026-03-22
BUG: avoid warning on ufunc with where=True and no output

mhvk:
Alternative fix for gh-31030. This does not change that `out=None` is removed from what is passed to the ufunc, but ensures that if from the ufunc one calls the original function with the original keyword arguments, there will only be a warning about a missing `out` argument if `where!=True`. Hence, the net effect is that the warning introduced in gh-29813 will no longer fire if there is no risk of uninitialized output.

(No AI was used.)


mattip:
Makes sense, but it does not solve the problem around `__arrray_ufunc__` argument passing.

mattip:
Thanks @mhvk

PR #31054, opened 2026-03-23
BUG: np.ones/ones_like raise ValueError for datetime64 with generic unit

Riccardo-stack:
Fixes a bug where np.ones(1, "M8") and np.ones_like(np.empty(1, "M8")) silently create an invalid datetime64 array with generic unit instead of raising an error immediately.
#30903

# **Root cause**: 
_ones()_ and _ones_like()_ fill the allocated array using multiarray.copyto(a, 1, casting='unsafe'). The raw Python integer 1 is copied directly into memory, bypassing NumPy's datetime type-checking pipeline entirely. The error is only triggered later — e.g. when printing — because that's when the C-level conversion (NpyDatetime_ConvertDatetime64ToDatetimeStruct) finally inspects the value.

By contrast, np.array(1, dtype="M8") correctly raises ValueError immediately because it goes through the standard type-checking path.

# **Fix**: 
Wrap the fill value in _array(1, dtype=a.dtype)_ before calling _copyto_. This forces the same validation path and raises ValueError at the point of creation. A TypeError catch is added for dtypes like _void_ that legitimately cannot convert integer 1 to bytes, preserving their existing behavior. The same fix is applied to _ones_like()_.

Built locally using spin build and ran the existing test suite:
```
spin test -t numpy/_core/tests/test_numeric.py::TestCreationFuncs
# 9 passed
```
And manually verified the fix.

## **First time committer introduction**
Hi! I'm Riccardo, a Python developer learning to contribute to open-source projects. I've recently started learning NumPy for my university course. This is one of my first contributions, and I found this issue quite interesting and that's why I tried to solve it

AI Disclosure
I used Antigravity (Google DeepMind) as an AI assistant to help me understand the issue, trace through the relevant source files (including the C-level _datetime.c_), and reason about the correct fix. 

**The final code changes were written and applied by me**. All interactions with the NumPy repository (commits, push, PR submission) are performed by me directly.

### DIFF
--- a/numpy/_core/numeric.py
+++ b/numpy/_core/numeric.py
@@ -230,7 +230,11 @@ def ones(shape, dtype=None, order='C', *, device=None, like=None):
         )
 
     a = empty(shape, dtype, order, device=device)
-    multiarray.copyto(a, 1, casting='unsafe')
+    try:
+        fill_value = array(1, dtype=a.dtype)
+    except TypeError:
+        fill_value = 1
+    multiarray.copyto(a, fill_value, casting='unsafe')
     return a
 
 
@@ -310,7 +314,11 @@ def ones_like(
     res = empty_like(
         a, dtype=dtype, order=order, subok=subok, shape=shape, device=device
     )
-    multiarray.copyto(res, 1, casting='unsafe')
+    try:
+        fill_value = array(1, dtype=res.dtype)
+    except TypeError:
+        fill_value = 1
+    multiarray.copyto(res, fill_value, casting='unsafe')
     return res
 
 


Riccardo-stack:
# Question
So the CI test revealed 34 failures with basically the same root cause, that is some NumPy tests call np.ones with generic datetime64 (M8) dtype and expect it to succeed.
Specifically:

**- test_iter_copy_casts** 
iterates over all type codes and creates arrays via np.ones(1000, dtype=...) including generic M8 — these are used as inputs for testing iterator casting behavior.

**- test_ufunc_types**
 similarly creates _np.ones((3,3), t)_ for every type in each ufunc's .types, which includes generic M8 for ufuncs like _add_, _subtract_, _maximum_, etc.
Both tests rely on _np.ones(N, "M8")_ succeeding — they don't care about the datetime value being valid, they just need an array of that dtype for testing other operations.

So my is approach too aggressive, but i see some kind of asimmetry, because _np.array_ is never called in these tests, and thus immediately raises **ValueError** (like I implementet with the fix);
On the other hand, if _np.ones_ raises the error immediately it generates problems in the tests.

**Looking forward on your advise on how to proceed and if you find the fix useful!**

riku-sakamoto:
I’m not sure this change properly addresses the issue, because the try-except block seems to handle the symptom rather than fixing the root cause.  I am afraid that it may also have some negative impact on performance.

It might be worth checking the C implementation (e.g., the copyto method) and find out how datetime arrays are initialized.

Riccardo-stack:
Thank you for the feedback! I agree that fixing this at the C level might be the right way to go, but I don't have experience with C and thus I'm not able to implement the proper fix for this. I will close this PR so that someonez can pick it up. 

Thanks again for the help and guidance!

PR #31061, opened 2026-03-24
BUG: propagate out=None to __array_ufunc__ dispatch kwargs

alvinttang:
## Summary

Fixes #31030.

When a user passes `out=None` explicitly to a ufunc that has a subclass
override (e.g. `np.log(duck_array, where=mask, out=None)`), the
`out=None` was silently dropped before `__array_ufunc__` was called.
This caused subclass implementations that forward `**kwargs` back to the
underlying ufunc to lose the `out=None` signal, triggering a spurious
`UserWarning: 'where' used without 'out'` on the re-entrant call.

### Root cause

`initialize_normal_kwds` in `numpy/_core/src/umath/override.c` builds
the keyword dict that is passed to `__array_ufunc__`.  When no real
output arrays are present (`out_args == NULL`), the function
**unconditionally deleted** any `out` key that had been copied from the
original `kwnames`.  This erased an explicit `out=None` before dispatch.

### Fix

In the `out_args == NULL` branch, preserve an `out` entry whose value
is `None` (the user's explicit intent).  A non-`None` value is still
removed to maintain the previous behaviour for unexpected cases.

## Changes

- `numpy/_core/src/umath/override.c` — fix `initialize_normal_kwds` to
  keep `out=None` in the normalized keyword dict.
- `numpy/_core/tests/test_ufunc.py` — add regression test
  `test_where_out_none_propagated_to_array_ufunc` that uses a
  `DuckArray` subclass to verify `out=None` reaches `__array_ufunc__`
  and that no `UserWarning` is raised.

## Test plan

- [ ] New regression test `TestUfunc::test_where_out_none_propagated_to_array_ufunc` passes
- [ ] Existing `test_where_warns` continues to pass (no regression on the warning itself)
- [ ] `python -m pytest numpy/_core/tests/test_ufunc.py -k "where" -v`

### DIFF
--- a/numpy/_core/src/umath/override.c
+++ b/numpy/_core/src/umath/override.c
@@ -121,13 +121,30 @@ initialize_normal_kwds(PyObject *out_args,
         }
     }
     else {
-        /* Ensure that `out` is not present. */
-        int res = PyDict_Contains(normal_kwds, npy_interned_str.out);
+        /*
+         * No normalized outputs (out_args is NULL), which means either no
+         * `out` was passed or `out=None` was passed explicitly.  If the user
+         * passed `out=None` it will already be in normal_kwds (copied from
+         * kwnames above); preserve it so that __array_ufunc__ implementations
+         * can forward it to the ufunc and suppress the "where without out"
+         * warning.  Only delete `out` from the dict when its value is not
+         * None (which would indicate a stale entry that should not be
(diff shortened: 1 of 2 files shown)


mattip:
We prefer to talk to a human, see the PR template which you ignored.

PR #31062, opened 2026-03-24
BUG: fix ndarray.resize() spurious ValueError on Python 3.14 (gh-30991)

### LINKED ISSUE #30991: BUG: ndarray.resize() always throws a ValueError with Python 3.14.3 and numpy 2.4.3

alvinttang:
## Summary

Closes #30991.

The previous fix in #31021 improved error messages but did not eliminate the spurious `ValueError`. This PR implements the actual fix by distinguishing between **optimized frames** (regular Python function bodies, where Python 3.14 uses borrowed references and does not bump the refcount) and **non-optimized callers** (module scope, `exec()`, Cython extensions — which still increment the refcount via the normal call protocol).

### Root cause

Python 3.14 changed reference-counting for function-local variables ([What's New](https://docs.python.org/3/whatsnew/3.14.html#whatsnew314-refcount)): optimized frames use borrowed references around method calls, so `Py_REFCNT(self)` stays at 1. But non-optimized contexts (module scope, Cython, `exec`) still do a normal `INCREF`/`DECREF` around the call, producing `Py_REFCNT == 2` even when the array is the sole owner — making `PyUnstable_Object_IsUniquelyReferenced` return false incorrectly.

### Fix

In `PyArray_Resize_int`, on Python >= 3.14:

1. Retrieve the caller's frame via `PyEval_GetFrame` (stable public API).
2. Check `PyFrame_GetCode(frame)->co_flags & CO_OPTIMIZED` (stable public API, `cpython/code.h`).
3. If the caller is in an **optimized frame**, use `PyUnstable_Object_IsUniquelyReferenced` — the most accurate check.
4. If the caller is in a **non-optimized frame**, fall back to `Py_REFCNT(self) > 2` — the pre-3.14 heuristic that correctly handles module scope and Cython.

All APIs used (`PyEval_GetFrame`, `PyFrame_GetCode`, `CO_OPTIMIZED`) are available in the stable CPython C API.

### Changes

- `numpy/_core/src/multiarray/shape.c`: implement frame-aware refcheck dispatch
- `numpy/_core/tests/test_multiarray.py`: `test_check_reference_module_scope` now asserts the resize **succeeds** on all Python versions (not just raises a nicer error)
- `numpy/_core/tests/test_cython.py`: `test_resize_refcheck` now asserts the Cython-scope resize **succeeds**; removed the Python 3.14-only skip
- `numpy/_core/_add_newdocs.py`: document the actual behavior change

## Reproduction

Before this fix (Python 3.14, module scope):
```python
import numpy as np
a = np.array([[0, 1], [2, 3]], order='C')
a.resize((2, 1))
# ValueError: cannot resize an array that may be referenced by another object.
```

After this fix: the above succeeds as expected.

## Test plan

- [ ] `test_check_reference_module_scope` — module-scope resize succeeds on all Python versions
- [ ] `test_resize_refcheck` — Cython-scope resize succeeds on all Python versions  
- [ ] `test_check_reference` — resize with `y = x` (genuine shared ref) still raises `ValueError`
- [ ] `test_check_reference_2` — second genuine-ref test still raises

### DIFF
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -4199,8 +4199,12 @@ _array_method_doc('resize', "*new_shape, refcheck=True",
     reallocate the memory.
 
     On Python 3.13 and older, the check allows objects with exactly one
-    reference to be reallocated in-place. On Python 3.14 and newer, the array
-    must be uniquely referenced. See [1]_ for more details.
+    reference to be reallocated in-place. On Python 3.14 and newer, the check
+    uses ``PyUnstable_Object_IsUniquelyReferenced`` when the array is resized
+    from within a regular Python function (an *optimized frame*), and falls
+    back to the pre-3.14 heuristic for module-level code, Cython extensions,
+    and other non-optimized callers. This avoids the spurious
+    ``ValueError`` that occurred in those contexts. See [1]_ for more details.
 
     If you are sure that you have not shared the memory for this array with
     another Python object, then you may safely set `refcheck` to False.
--- a/numpy/_core/src/multiarray/shape.c
+++ b/numpy/_core/src/multiarray/shape.c
@@ -90,22 +90,31 @@ PyArray_Resize_int(PyArrayObject *self, PyArray_Dims *newshape, int refcheck)
         if (refcheck) {
 #if PY_VERSION_HEX >= 0x030E00B0
             // Python 3.14 changed reference counting semantics for function-
-            // local variables. There is no way to tell if the calling function
-            // has been optimized (because it might be implemented in C or Cython)
+            // local variables (PEP 703 / gh-30991): optimized frames use
+            // borrowed references, so calling a method on a local variable
+            // does NOT bump its refcount.  Non-optimized frames (module
+            // scope, exec(), Cython, etc.) still increment the refcount
+            // around the call, producing a spurious refcount of 2.
             //
-            // Instead, warn if the refcount is exactly 2 that this might be a
-            // false positive
-            if (!PyUnstable_Object_IsUniquelyReferenced((PyObject *)self)) {
-                if (Py_REFCNT(self) == 2) {
-                    PyErr_SetString(
-                            PyExc_ValueError,
-                            "cannot resize an array that may be referenced "
-                            "by another object.\n"
-                            "It is possible that this is a false positive.\n"
-                            "If you are sure that the array is uniquely referenced, "
-                            "set refcheck=False.");
-                    return -1;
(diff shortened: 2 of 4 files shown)


seberg:
Closing, what is your motivation to reject the decision from the issue?

(Also as it looks just automated without a comment, if disclosed.)

PR #31063, opened 2026-03-24
BUG: result_type treats builtin type objects as concrete instead of abstract (#31037)

gambletan:
## Summary
- `np.result_type(np.int32, int)` returned `int64` instead of `int32`, inconsistent with `np.result_type(np.int32(1), 1)` which returns `int32` under NEP 50 weak-type promotion
- Root cause: Python builtin type objects (`int`, `float`, `complex`, `bool`) fell through to `PyArray_DescrConverter` which mapped them to concrete default dtypes, bypassing weak promotion
- Fix: intercept `PyLong_Type`, `PyFloat_Type`, `PyComplex_Type`, `PyBool_Type` and create representative 0-D arrays with Python-scalar flags, matching how instances are handled

## Test plan
- Added `test_result_type_builtin_types_as_abstract` verifying symmetry between type objects and instances
- All existing `test_result_type` tests pass

Closes #31037


mattip:
You ignored the PR template. I suspect this was AI generated. If not please reopen and fill out the template.

PR #31064, opened 2026-03-24
BUG: fix dangling pointer in PyArray_FromInterface (gh-31036)

gambletan:
## Summary

Fixes #31036.

Objects that implement `__array_interface__` as a *property* (returning a fresh dict each call) but do **not** implement `__array_struct__` were silently producing corrupt array data.

### Root cause

`PyArray_FromInterface` reads the `data` pointer from the interface dict, calls `PyArray_NewFromDescrAndBase(... base=origin)`, then does `Py_DECREF(iface)`.

NumPy scalars embed a 0-d array under the `__ref` key of the interface dict to keep the scalar's buffer alive. When the dict was a temporary (returned by a property), its refcount hit zero at the `Py_DECREF`, freeing `__ref` and leaving the new array with a **dangling data pointer**.

### Fix

When the data comes from the pointer-tuple path (`iface['data'] = (ptr, readonly_flag)`), build a `(origin, iface)` tuple as the `base` instead of using `origin` alone. This keeps both the calling object *and* the interface dict alive for the entire lifetime of the array — the same pattern already used in `PyArray_FromStructInterface` with its `refs` tuple.

### Reproducer (from the issue)

```python
import numpy as np, gc

class MyObj:
    value = np.int64(2)
    def __rsub__(self, other):
        return self.value.__rsub__(other)
    @property
    def __array_interface__(self):
        return self.value.__array_interface__

o = MyObj()
result = np.int64(1) - o
gc.collect()
assert result == -1, f"got {result}"   # was failing with 0
```

### Tests

- `numpy/_core/tests/test_array_interface.py::test_array_interface_without_array_struct` — new pure-Python regression test
- Updated `TestArrayCreationCopyArgument::test_array_interfaces` in `test_multiarray.py` to reflect that `res.base` is now a `(origin, iface_dict)` tuple (both are kept alive); the array values and copy semantics are unchanged

### DIFF
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -2176,6 +2176,7 @@ PyArray_FromInterface(PyObject *origin)
     PyObject *iface = NULL;
     PyObject *attr = NULL;
     PyObject *base = NULL;
+    PyObject *refs = NULL;  /* owned tuple to keep iface alive when base=origin */
     PyArrayObject *ret;
     PyArray_Descr *dtype = NULL;
     char *data = NULL;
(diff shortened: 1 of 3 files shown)


mattip:

PR #31066, opened 2026-03-24
BUG: fix fill_value propagation in mvoid.__getitem__

### LINKED ISSUE #28450: BUG: `MaskedArray.fill_value` not consistently preserved in structured voids with array-valued fields
### Describe the issue:

If a masked array of a structured void type with an array-valued field is built, the `.fill_value` differs between the following two cases (see the example):
1. First accessing the array item (of `np.ma.core.mvoid` type), then accessing the array-valued field, and
2. First accessing the array-valued field (returning a masked array of scalar data type), then accessing the array item.

While both chains of item access results in a new masked array of the shape of the array-valued field, they have different `.fill_value`s; the first form results in a `.fill_value` which:
- Is array-valued, which does not seem consistent with masked arrays usually behaves, and
- Neglects the `.fill_value` of the parent array, substituting it with the default value for the data type as given by `.get_fill_value()`.

Beside `numpy 2.2`, the bug is also consistently reproduced in `numpy 1.24` on my machine (MacOS 14.6.1 on M3 chip). Presumably, `numpy` versions in between and before are also affected. Since this also has to do with `.fill_value` inconsistencies on differing orders of operations, #22141 may be a related issue.

```python
from numpy.ma import MaskedArray


# This masked array contains a structured void with a single field which is array-valued
ma = MaskedArray([((1,),)], dtype=[('a', 'i', (1,))], fill_value=-1)

# 1. Item access -> field access:
# `.fill_value` becomes an array and falls back to the default
subma_1 = ma[0]['a']
# >>> subma_1
# masked_array(data=[1],
#              mask=[False],
#        fill_value=[999999],
#             dtype=int32)
assert subma_1.fill_value.shape == (1,)
assert subma_1.fill_value[0] != -1

# 2. Field access -> item access:
# `.fill_value` remains a scalar and retains the specified value
subma_2 = ma['a'][0]
# >>> subma_1
# masked_array(data=[1],
#              mask=[False],
#        fill_value=-1,
#             dtype=int32)
assert subma_1.fill_value.shape == ()
assert subma_1.fill_value[0] == -1
```

### Error message:

```shell

```

```
2.2.2
3.13.2 (main, Feb  4 2025, 14:51:09) [Clang 16.0.0 (clang-1600.0.26.6)]
```

### Runtime Environment:

_No response_

### Context for the issue:

It is understood that this is an edge case, and the `.fill_value` is mostly not of concern to end-users. Still, it may be of use to document it here, and of interest in revealing other possible inconsistencies.

Tomas-rep:
Ensure that structured masked arrays correctly propagate the fill_value when accessing a field from an mvoid object (e.g., ma[0]['field']). Previously, accessing an array-valued field from an mvoid would lose the custom fill_value and revert to the default. Closes #28450

### PR summary
 The issue was located in mvoid.__getitem__. When a structured row was accessed, the mvoid constructor was called without passing the existing fill_value. This caused the resulting object to drop the custom fill_value and revert to the default. I updated the constructor call to explicitly propagate self._fill_value to the new mvoid instance.

 I'm a Bsc Student in Computer Science and Engineering, currently finishing the third year. I was already familiar with numpy due to some of my previous projects, so when I was told that I needed to make a contribution to an open-source project, numpy seemed to fit what I was looking for.

#### AI Disclosure
 No AI tools used

### DIFF
--- a/numpy/ma/tests/test_core.py
+++ b/numpy/ma/tests/test_core.py
@@ -4071,6 +4071,19 @@ class TestMaskedArrayMethods:
         assert_equal(xd.mask, x.diagonal().mask)
         assert_equal(xd.data, x.diagonal().data)
 
+    def test_mvoid_fill_value_propagation(self):
+            data = [((1,),)]
+            dtype = [('a', 'i', (1,))]
+            fill_value = -1
+            
+            ma = np.ma.MaskedArray(data, dtype=dtype, fill_value=fill_value)
+            
+            subma_1 = ma[0]['a']
+            subma_2 = ma['a'][0]
+            
+            assert_equal(subma_1.fill_value, fill_value)
+            assert_equal(subma_1.fill_value, subma_2.fill_value)
+
 
 class TestMaskedArrayMathMethods:
     def _create_data(self):
--- a/numpy/ma/tests/test_core.py
+++ b/numpy/ma/tests/test_core.py
@@ -4074,13 +4074,13 @@ class TestMaskedArrayMethods:
     def test_mvoid_fill_value_propagation(self):
         data = [((1,),)]
         dtype = [('a', 'i', (1,))]
-        fill_value = -1
-            
+        fill_value = 42
+
         ma = np.ma.MaskedArray(data, dtype=dtype, fill_value=fill_value)
-            
+
         subma_1 = ma[0]['a']
         subma_2 = ma['a'][0]
-            
+
         assert_equal(subma_1.fill_value, fill_value)
         assert_equal(subma_1.fill_value, subma_2.fill_value)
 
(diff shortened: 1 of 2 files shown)


numpy/ma/tests/test_core.py
```diff
@@ -4071,6 +4071,19 @@ def test_diagonal_view(self):
         assert_equal(xd.mask, x.diagonal().mask)
         assert_equal(xd.data, x.diagonal().data)
 
+    def test_mvoid_fill_value_propagation(self):
+        data = [((1,),)]
+        dtype = [('a', 'i', (1,))]
+        fill_value = -1
```
mattip: use some other value, like 42, to prove it is not being set to -1 by default

mattip (commented):
Seems OK. I am not sure this is useful at all, but it is an odd corner in the returned value.

PR #31067, opened 2026-03-24
MAINT,BUG: refactor ``PyArray_DescrFromScalar`` for parametric support

MaartenBaert:
### PR summary

When PyArray_DescrFromScalar is called on a scalar corresponding to a user-defined parametric dtype, it silently discards the parameters of this dtype and replaces them with the defaults.

This shows up for example when the `astype` member function of the scalar type is called, which internally tries to construct a 0-D array as an intermediate step. This array ends up having the wrong dtype (parameters are lost) and as a result the conversion can unexpectedly break.

The problem was discovered during the development of a user-defined fixed-point dtype, where it ended up discarding the 'number of fractional bits' parameter, completely breaking the conversion. We currently work around this by overriding the `astype` member function of the scalar type.

It currently also shows up in a more subtle way in [numpy_quaddtype](https://github.com/numpy/numpy-quaddtype) where it silently discards the backend parameter, resulting in a temporary conversion to the default SLEEF backend. This is relatively harmless but still unexpected.

I did not see an obvious way to trigger this issue with the SFloat user-defined dtype that is used in the numpy test suite, because it doesn't have a corresponding scalar type. Instead, the added tests are simply there in an attempt to catch potential regressions caused by this change, and are entirely optional.

#### AI Disclosure

Claude Sonnet 4.6 was used to identify and fix the bug. I have reviewed the rationale and proposed fix, and they seem reasonable to me. However I have limited knowledge of numpy internals and can't judge whether this is the most appropriate solution.

The full AI-generated description of the bug and proposed solution can be found here:
https://gist.github.com/MaartenBaert/3cf4a926f139b0ec551d23bc5920ecc7

### DIFF
--- a/doc/release/upcoming_changes/31067.c_api.rst
+++ b/doc/release/upcoming_changes/31067.c_api.rst
@@ -1,0 +1,8 @@
+``PyArray_DescrFromScalar`` now preserves parametric dtype information
+----------------------------------------------------------------------
+``PyArray_DescrFromScalar`` now correctly returns the full dtype descriptor for
+scalars of user-defined parametric data types, including any dtype parameters.
+Previously, parameters were silently discarded, which could cause incorrect
+results in operations like ``astype`` on scalar objects. Internally, the
+function now delegates to ``discover_descr_from_pyobject``, which handles
+parametric dtypes correctly.
--- a/numpy/_core/src/multiarray/dtypemeta.c
+++ b/numpy/_core/src/multiarray/dtypemeta.c
@@ -573,15 +573,10 @@ discover_datetime_and_timedelta_from_pyobject(
         PyArray_DTypeMeta *cls, PyObject *obj) {
     if (PyArray_IsScalar(obj, Datetime) ||
             PyArray_IsScalar(obj, Timedelta)) {
-        PyArray_DatetimeMetaData *meta;
-        PyArray_Descr *descr = PyArray_DescrFromScalar(obj);
-        meta = get_datetime_metadata_from_dtype(descr);
-        if (meta == NULL) {
-            return NULL;
-        }
-        PyArray_Descr *new_descr = create_datetime_dtype(cls->type_num, meta);
-        Py_DECREF(descr);
-        return new_descr;
+        /* Extract metadata directly from the scalar object. */
+        PyArray_DatetimeMetaData *meta =
+                &((PyDatetimeScalarObject *)obj)->obmeta;
+        return create_datetime_dtype(cls->type_num, meta);
     }
     else {
         return find_object_datetime_type(obj, cls->type_num);
--- a/numpy/_core/src/multiarray/scalarapi.c
+++ b/numpy/_core/src/multiarray/scalarapi.c
@@ -402,67 +402,37 @@ PyArray_DescrFromTypeObject(PyObject *type)
 NPY_NO_EXPORT PyArray_Descr *
 PyArray_DescrFromScalar(PyObject *sc)
 {
-    int type_num;
-    PyArray_Descr *descr;
-
-    if (PyArray_IsScalar(sc, Void)) {
-        descr = (PyArray_Descr *)((PyVoidScalarObject *)sc)->descr;
-        Py_INCREF(descr);
-        return descr;
+    /*
+     * Look up the DType directly from the scalar's type.  This avoids calling
+     * the NPY_DT_default_descr slot (via PyArray_GetDefaultDescr), which for
+     * parametric dtypes may raise an error or return an incorrect stub.
+     * Once we have the DType class, discover_descr_from_pyobject extracts the
+     * correct instance-specific descriptor (handling void, datetime, string,
+     * and new-style user-defined parametric dtypes correctly).
+     */
+    PyArray_DTypeMeta *DType =
+            (PyArray_DTypeMeta *)PyArray_DiscoverDTypeFromScalarType(Py_TYPE(sc));
+    if (DType != NULL) {
+        PyArray_Descr *result = NPY_DT_CALL_discover_descr_from_pyobject(DType, sc);
+        Py_DECREF(DType);
+        return result;
     }
 
-    if (PyArray_IsScalar(sc, Datetime) || PyArray_IsScalar(sc, Timedelta)) {
-        PyArray_DatetimeMetaData *dt_data;
-
-        if (PyArray_IsScalar(sc, Datetime)) {
-            descr = PyArray_DescrNewFromType(NPY_DATETIME);
-        }
-        else {
-            /* Timedelta */
-            descr = PyArray_DescrNewFromType(NPY_TIMEDELTA);
-        }
-        if (descr == NULL) {
-            return NULL;
-        }
-        dt_data = &(((PyArray_DatetimeDTypeMetaData *)((_PyArray_LegacyDescr *)descr)->c_metadata)->meta);
-        memcpy(dt_data, &((PyDatetimeScalarObject *)sc)->obmeta,
-               sizeof(PyArray_DatetimeMetaData));
-
-        return descr;
-    }
-
-    descr = PyArray_DescrFromTypeObject((PyObject *)Py_TYPE(sc));
+    /*
+     * Fallback for scalar subclasses that are not directly in the scalar-type
+     * registry (e.g. a Python subclass of np.float64).  These are always
+     * legacy non-parametric dtypes, so PyArray_DescrFromTypeObject is safe:
+     * it walks the MRO to find the registered base type and calls
+     * PyArray_GetDefaultDescr, which works correctly for non-parametric dtypes.
+     */
+    PyArray_Descr *descr = PyArray_DescrFromTypeObject((PyObject *)Py_TYPE(sc));
     if (descr == NULL) {
         return NULL;
     }
-    if (PyDataType_ISLEGACY(descr) && PyDataType_ISUNSIZED(descr)) {
-        PyArray_DESCR_REPLACE(descr);
-        if (descr == NULL) {
-            return NULL;
-        }
-        type_num = descr->type_num;
-        if (type_num == NPY_STRING) {
-            descr->elsize = PyBytes_GET_SIZE(sc);
-        }
-        else if (type_num == NPY_UNICODE) {
-            descr->elsize = PyUnicode_GET_LENGTH(sc) * 4;
-        }
-        else {
-            _PyArray_LegacyDescr *ldescr = (_PyArray_LegacyDescr *)descr;
-            PyArray_Descr *dtype;
-            dtype = (PyArray_Descr *)PyObject_GetAttrString(sc, "dtype");
-            if (dtype != NULL) {
-                descr->elsize = dtype->elsize;
-                ldescr->fields = PyDataType_FIELDS(dtype);
-                Py_XINCREF(ldescr->fields);
-                ldescr->names = PyDataType_NAMES(dtype);
-                Py_XINCREF(ldescr->names);
-                Py_DECREF(dtype);
-            }
-            PyErr_Clear();
-        }
-    }
-    return descr;
+    DType = NPY_DTYPE(descr);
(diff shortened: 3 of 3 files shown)


numpy/_core/tests/test_custom_dtypes.py
```diff
@@ -391,3 +391,56 @@ def test_type_pickle():
...
+        intermediate scaling used to store the data, confirming that
+        PyArray_DescrFromScalar returns a descriptor that encodes the
+        correct scaling.
+        """
+        values = [1.0, 2.0, 3.0]
+        for scaling in [0.5, 1.0, 2.0, -3.0]:
+            arr = np.array(values, dtype=SF(scaling))
+            result = arr.astype(np.float64)
+            np.testing.assert_array_equal(result, values,
+                err_msg=f"scaling={scaling}: astype(float64) gave wrong result")
```
seberg: I don't understand these tests. You were talking about scalars, no? Am I missing where any of these tests use scalars?

EDIt: To be clear, there may not be a way to write a test with this at all.  But we could make one that uses `quaddtype` (even if that might not run in CI right now).

numpy/_core/src/multiarray/scalarapi.c
```diff
@@ -431,6 +431,29 @@ PyArray_DescrFromScalar(PyObject *sc)
...
+            PyArray_DTypeMeta *DType = (PyArray_DTypeMeta *)DType_obj;
+            if (!NPY_DT_is_legacy(DType)) {
+                PyArray_Descr *result =
+                    NPY_DT_CALL_discover_descr_from_pyobject(DType, sc);
+                Py_DECREF(DType_obj);
+                return result;
+            }
+            Py_DECREF(DType_obj);
+        }
+    }
```
seberg, line 419: Part of the reason why the function doesn't have this was is that I had hoped to not force subclassing `np.generic` although we never added a new `np.scalar` abstract base class (for the purpose of having something for `isinstance()`).

(I.e. this is just unchanged, with the hope to basically not use it, but of course it is used for some `np.generic` related code paths.)

It would be nice to untangle this a bit, though...  But I have to think about how that is best.

```
PyObject *DType_obj = PyArray_DiscoverDTypeFromScalarType(Py_TYPE(sc));
```
should work _mostly_ but not always, because of crazy things like subclasses, that is why some other paths actually call _this_ function as a backup solution, which means they will end up calling things twice, a bit awkward but maybe OK...

So when that fails, it seems fine to use `PyArray_DescrFromTypeObject`, _but_ I would much prefer if you just used:
```
descr = PyArray_DescrFromTypeObject((PyObject *)Py_TYPE(sc));
Py_INCREF(NPY_DTYPE(descr));
DType = NPY_DTYPE(descr);
```
or so, then just call `NPY_DT_CALL_discover_descr_from_pyobject` unconditionally.

(That is, we delete everything else in this function!)

I _think_ that will broaden up this function even more than it already is, but maybe that is OK.  There is a chance that this is very slightly slower for builtin NumPy dtypes in which case we could swap the order. But overall, I wouldn't worry about it.

seberg (commented):
Thanks, the history behind is that I had hoped to never touch this stuff and you would rather just not subclass `generic` at least for parametric dtypes...

Because really, also getting the scalar value can be dubious (i.e. `generic` forces a very strict ABI on you that is IMO not helpful at all).

---

But... pragmatic improvements like here are good and what we need. It would just be nice to try to reduce the complexity, rather than add to it.

numpy/_core/tests/test_custom_dtypes.py
MaartenBaert: Unfortunately SFloat doesn't have a scalar, so this can't be tested directly. Instead the tests try to verify that we at least haven't broken things for the array case. Honestly this is a very weak test and I don't like it either, but implementing a whole scalar SFloat type just for this test seemed overkill. If you want I can also drop the tests entirely.

I have verified that the fix actually solves the problem for our custom fixed-point type.

seberg: We can write one using quaddtype, but yeah this tests right now is completely unrelated to the code change, so there is little point hoenstly.

numpy/_core/src/multiarray/scalarapi.c
MaartenBaert: By 'delete everything else', I assume you mean removing the entire `if (PyDataType_ISLEGACY(descr) && PyDataType_ISUNSIZED(descr)) {` block?

AI analysis claims this will break for datetime/timedelta because discover_datetime_and_timedelta_from_pyobject (dtypemeta.c:577) calls back into PyArray_DescrFromScalar which would create infinite recursion. But maybe it is wrong, or I didn't understand your intent correctly.

Full analysis, if you want to read it: https://gist.github.com/MaartenBaert/b77e664b1b86a4fb585ef1686151705d

seberg: > AI analysis claims this will break for datetime/timedelta because discover_datetime_and_timedelta_from_pyobject (dtypemeta.c:577) calls back into PyArray_DescrFromScalar which would create infinite recursion. But maybe it is wrong, or I didn't understand your intent correctly.

Maybe, but in that case we can fix it there to prevent the recursion. I expected this to be the case for most dtypes, but you are making logically circular here already, I want to break that cycle.
If moving logic over to `dtypemeta.c` is needed for that, then that is fine.

(It seems most logic is duplicated, apparently, I wanted to avoid doing subtle changes here... but maybe it's time to bite the bullet and try to get away with it.)

ngoldbaum:
IMO we could add quaddtype as a test dependency of NumPy's CI rather than going out of our way to make the scaled float test dtype work for tests.

seberg:
I was just doing that in gh-30798 (very minimally).  So for this PR, we can just assume it'll be there, I hope.
(We can hash out the details there, it probably makes sense to add to more runs, but I didn't have a better idea, and was OK with being rather minimal, but wanted to avoid possibly compiling.)

MaartenBaert:
@seberg I've attempted to simplify PyArray_DescrFromScalar like you suggested, though this required rewriting discover_datetime_and_timedelta_from_pyobject to extract metadata directly from the scalar's obmeta field instead of calling back into PyArray_DescrFromScalar. I hope this approach is okay.

Edit: This doesn't work! This doesn't break anything in Numpy, but it just reintroduced the `astype` bug for my custom dtype. Looks like I can't avoid going through PyArray_DiscoverDTypeFromScalarType.

MaartenBaert:
Second attempt. This does properly solve the problem for my custom dtype, and is probably what you meant in the first place, I just didn't understand your comment correctly.

numpy/_core/src/multiarray/dtypemeta.c
```diff
@@ -573,15 +573,10 @@ discover_datetime_and_timedelta_from_pyobject(
...
-        if (meta == NULL) {
-            return NULL;
-        }
-        PyArray_Descr *new_descr = create_datetime_dtype(cls->type_num, meta);
-        Py_DECREF(descr);
-        return new_descr;
+        /* Extract metadata directly from the scalar object. */
+        PyArray_DatetimeMetaData *meta =
+                &((PyDatetimeScalarObject *)obj)->obmeta;
+        return create_datetime_dtype(cls->type_num, meta);
```
MaartenBaert, line 579: I realize this looks extremely sketchy because it relies on PyDatetimeScalarObject and PyTimedeltaScalarObject having identical memory layout such that the `obmeta` field is at the same offset. However the old code in PyArray_DescrFromScalar did the exact same thing:
```
memcpy(dt_data, &((PyDatetimeScalarObject *)sc)->obmeta,
               sizeof(PyArray_DatetimeMetaData));
```
Perhaps this was a bug. Would it be better to handle the Datetime and Timedelta cases separately to avoid this?

seberg: Yeah, this is fine, I am sure most of the code bakes this in, and it's not like the ABI can change reasonably.

numpy/_core/tests/test_custom_dtypes.py
MaartenBaert: I removed them.

numpy/_core/src/multiarray/scalarapi.c
MaartenBaert: I think this is fixed now, please confirm whether this is what you intended.

numpy/_core/src/multiarray/scalarapi.c
```diff
@@ -402,67 +402,40 @@ PyArray_DescrFromTypeObject(PyObject *type)
...
-        if (type_num == NPY_STRING) {
-            descr->elsize = PyBytes_GET_SIZE(sc);
-        }
-        else if (type_num == NPY_UNICODE) {
-            descr->elsize = PyUnicode_GET_LENGTH(sc) * 4;
-        }
-        else {
-            _PyArray_LegacyDescr *ldescr = (_PyArray_LegacyDescr *)descr;
-            PyArray_Descr *dtype;
-            dtype = (PyArray_Descr *)PyObject_GetAttrString(sc, "dtype");
```
seberg, line 453: Just a note: This path is a bit scary, but I can't see it actually ever being triggered. User dtypes shouldn't be unsized, but all dtypes that can be (void, string, unicode) are already explicitly handled earlier.

So I am pretty sure this is safe.

numpy/_core/src/multiarray/scalarapi.c
```diff
@@ -402,67 +402,40 @@ PyArray_DescrFromTypeObject(PyObject *type)
...
-            PyErr_Clear();
-        }
-    }
-    return descr;
+    DType = NPY_DTYPE(descr);
+    Py_INCREF(DType);
+    Py_DECREF(descr);
+
+    PyArray_Descr *result = NPY_DT_CALL_discover_descr_from_pyobject(DType, sc);
+    Py_DECREF(DType);
```
seberg:
```suggestion
    DType = NPY_DTYPE(descr);
    PyArray_Descr *result = NPY_DT_CALL_discover_descr_from_pyobject(DType, sc);
    Py_DECREF(descr);
```
Just a nit, it's fine to borrow the DType here.  Or move `PyArray_DescrFromTypeObject` into a `DType == NULL` path above.  But it's so short, I don't mind duplicating `NPY_DT_CALL_discover_descr_from_pyobject`.

seberg (approved):
Yeah, this is what I meant.  I think we can go with this, it would be nice to have a quaddtype test, but that is broken right now :(.

Would you be up to add a brief `c_api` release note? Maybe just to point out that `PyArray_DescrFromScalar` now works more broadly and relies directly on `discover_descr_from_pyobject`.

doc/release/upcoming_changes/31067.c_api.rst
```diff
@@ -0,0 +1,8 @@
+``PyArray_DescrFromScalar`` now preserves parametric dtype information
+----------------------------------------------------------------------
+``PyArray_DescrFromScalar`` now correctly returns the full dtype descriptor for
+scalars of user-defined parametric data types, including any dtype parameters.
+Previously, parameters were silently discarded, which could cause incorrect
+results in operations like ``astype`` on scalar objects. Internally, the
+function now delegates to ``discover_descr_from_pyobject``, which handles
+parametric dtypes correctly.
```
MaartenBaert, line 8: Is this okay?

MaartenBaert:
The CI failure looks spurious to me. I don't see why this commit would specifically break openblas32 and nothing else.

ngoldbaum:
I restarted the CI and it seems to have gotten past it.

Unfortunately the NumPy CI is kind of flaky. If you see a failure that looks unrelated, chances are it is.  A maintainer will flag it if it's a real issue.

doc/release/upcoming_changes/31067.c_api.rst
seberg: Yeah, I think it points out to a change in behavior well enough. It's not like I expect anyone to seriously use this API anyway.

seberg:
Yeah, I had also restarted CI once :).  Thanks, let's give this a shot, then, hopefully there isn't some weird subclass fallout, but I don't _think_ there should be a change in behavior, even for weird user-dtypes.

PR #31068, opened 2026-03-24
ENH: Improve ufunc performance by caching the loop function

eendebakpt:
### PR summary

The `PyUFunc_DefaultLegacyInnerLoopSelector` is quite slow and called for every ufunc execution (for details see  https://github.com/numpy/numpy/pull/31018).

In this PR we cache the selected loop on the `PyArrayMethodObject`. Benchmark results:
```
np.abs(-2.2): Mean +- std dev: [main_fix] 129 ns +- 2 ns -> [pr_fix] 110 ns +- 6 ns: 1.18x faster
np.cos(-2.2): Mean +- std dev: [main_fix] 121 ns +- 9 ns -> [pr_fix] 108 ns +- 8 ns: 1.12x faster
np.abs(np.float64(-2.2)): Mean +- std dev: [main_fix] 173 ns +- 5 ns -> [pr_fix] 155 ns +- 8 ns: 1.12x faster
np.cos(np.float64(-2.2)): Mean +- std dev: [main_fix] 160 ns +- 7 ns -> [pr_fix] 148 ns +- 7 ns: 1.09x faster
np.abs(x): Mean +- std dev: [main_fix] 445 ns +- 3 ns -> [pr_fix] 420 ns +- 4 ns: 1.06x faster
np.cos(x): Mean +- std dev: [main_fix] 442 ns +- 12 ns -> [pr_fix] 426 ns +- 12 ns: 1.04x faster
np.add.reduce(x): Mean +- std dev: [main_fix] 1.11 us +- 0.03 us -> [pr_fix] 1.10 us +- 0.04 us: 1.01x faster
np.add.accumulate(x): Mean +- std dev: [main_fix] 612 ns +- 26 ns -> [pr_fix] 579 ns +- 17 ns: 1.06x faster

Geometric mean: 1.08x faster
```

For reviewers: in the first commit legacy ufunc loops and cached (core of this PR). In the second commit we move the caching to the level of strided loops (not necessary for performance, but seems nicer). In the third commit the freelist for `NpyAuxData` is removed to simplify code (it is no longer needed).

We could probably refactor `generate_umath.py` to generate strided loops directly. In that way we could avoid the additional fields on `PyArrayMethodObject`. The amount of changes would be larger though, and we maybe would have to maintain two sets of generator (for backwards compatibility reasons). If desired, I can explore this option.

#### AI Disclosure

Claude code was used in analysing the performance patterns and suggested to cache the ufunc loops. Significant parts of the code in the PR have been written by the tools.


numpy/_core/src/umath/legacy_array_method.c
```diff
@@ -32,61 +32,26 @@ typedef struct {
...
     data->user_data = user_data;
     data->pyerr_check = pyerr_check;
     return (NpyAuxData *)data;
 }
 
-#undef NPY_LOOP_DATA_CACHE_SIZE
-
 /*
- * This is a thin wrapper around the legacy loop signature.
+ * Thin wrapper around the legacy loop using heap-allocated auxdata.
```
eendebakpt:
```suggestion
 * This is a thin wrapper around the legacy loop signature.
```

mhvk (commented):
This seems a rather nice improvement, arguably quite a bit clearer than what we had. It  looks OK, but I'm sufficiently unsure I understand all details, so would be better for @seberg to have a look too.

Just to show my ignorance: the net effect of this PR is to *move* caching to the method, rather than add caching right? This since the loop data cache is removed.

seberg:
Yeah, we had discussed this. There are two things that are slightly relevant (and why this wasn't always the case):
1. `numba` adds loops on the fly, however, adding loops doesn't matter for _caching_, so long we still do the old loop lookup on cache failure.
2. We have a `PyUFunc_ReplaceLoopBySignature` which this will break. Now, we probably have to just break that eventually and in most cases it'll still work if done early on (NumPy may do some ufunc calls at import time, though).
   If Marten agrees, I would be up for just breaking it in some cases. Of course it would be possible to try and see if there is a legacy loop that cached the previous version and clear the cache.  But that might be a bit annoying.

seberg:
Hmmm, actually one solution we could do is to allow `AddLoopFromSpec` to add a:
```
#define NPY_METH_replace   <INTEGER>
```
slot (or flag, but I think I actually like a slot number) to the spec.  Then `AddLoopFromSpec` can actually replace the entry.

We would have to put a fat warning that replacing loops is thread-unsafe (truly!) right now, but I am OK with that.
That way we introduce a clear path to replace existing loops, and as there are no *real* users, I think that should be sufficient.

Maybe that is too much here, but I can see trying to do that quickly soon to not worry about it here.
@kumaraditya303 this is what I meant when I always said that we really want to allow replacing `ArrayMethod` in ufunc dispatching.  It would be nice if we could figure out a way to make that thread-safe on free-threading in a sane way.
In practice, such things should happen at import time, etc. and one could mark API as thread unsafe, but at least `import` is hard, because it is not *that* weird to import in a threaded environment.

kumaraditya303:
> Maybe that is too much here, but I can see trying to do that quickly soon to not worry about it here.
@kumaraditya303 this is what I meant when I always said that we really want to allow replacing ArrayMethod in ufunc dispatching.

Just to be clear you mean to allows replacing values on the `ufunc->_dispatch_cache` `PyArrayIdentityHash` table?
That is possible to do but would use a lot of memory in the current implementation, you would have to replace the entire backing `buckets` table on every replacement. If CPython exposes `PyMem_FreeDelayed` API in future then it can be done without significant memory use.

eendebakpt:
> Just to show my ignorance: the net effect of this PR is to _move_ caching to the method, rather than add caching right? This since the loop data cache is removed.

The loop data is stored in `ufunc->functions`/ `ufunc->types`. On main the correct loop is determined for each ufunc call using `PyUFunc_DefaultLegacyInnerLoopSelector`. In this PR for each dtype combination the correct loop is stored in the `PyArrayMethodObject`, which saves a call to `PyUFunc_DefaultLegacyInnerLoopSelector`.

However, the old data structure is not removed for two reasons: i) on initialization the structure is used to set the correct fields on the `PyArrayMethodObject` ii) at runtime the old structure (which one could call a cache) is used for the special int comparisons.

Since the old data structure is not removed, the cache is not purely moved. However, for practical purposes (e.g. all normal ufunc calls) it has moved to the `PyArrayMethodObject`.

numpy/_core/src/umath/legacy_array_method.c
```diff
@@ -218,17 +206,39 @@ get_wrapped_legacy_ufunc_loop(PyArrayMethod_Context *context,
...
+     */
+    *flags = context->method->flags & NPY_METH_RUNTIME_FLAGS;
+    if (context->method->cached_loop != NULL) {
+        *out_loop = &call_cached_loop;
+        *out_transferdata = NULL;
+        return 0;
+    }
+
+    /*
+     * Fallback: resolve loop at call time and heap-allocate auxdata.
```
seberg: So, I think we can just delete this fallback completely.  By the time you reach here, you have constructed the ArrayMethod.
Thus you already set up the loop cache (at this point, not really a cache anymore) when you reach here.
(That is also nice, because it removes that auxdata dance completely.)

I.e. if we fail to set up the new version, I think that is perfectly OK to give a hard error and thus much simplifying this code.
(As that removes a whole lot of the old setup code that you kept around.)

While none of this is thread-safe, it never was anyway, so I think the "look up the loop and set change it" approach is sufficient for backcompat.

(Which makes this a very nice cleanup!)

eendebakpt: This is still used for the special int comparisons. I agree it would be nice to remove this altogether, but the change would be larger. 
(I looked at it briefly, but iirc there was no easy handling of the special int comparisons. perhaps it can be done though, I would have to take a second look)

seberg: Ah, hmmmm. Maybe it's OK if we do a bit of crazy things here and:
1. Manually fetch the original int loop (for that specific special integer one).
2. Set it as `self->wrapped_meth` (just hard override after creation)
3. Directly use `self->wrapped_meth->get_loop()`.

Since this is all internal, the huge amount of deletion is worth a little bit of a hack, I think.
(Maybe we could also dynamically look up the correct loop, but it seems a bit much effort to be worth it.)

eendebakpt: Also the user loops still required the fallback. I let Claude handle those as well, the result is 

https://github.com/numpy/numpy/compare/main...eendebakpt:ufunc_loop_cache_v3

I have not reviewed it myself yet, but it does clean up. I will probably merge the changes into this branch unless we find some issues with the approach.

seberg: Argg, OK, yeah, the userloop via dict logic is ridiculous, I forgot about that.  So yeah I can see that this would be a problem (one that would be nice to deprecate in the crazier forms, although it's impossible who might rely on it for some weird code).

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -4827,6 +4827,33 @@ PyUFunc_ReplaceLoopBySignature(PyUFuncObject *func,
...
+
+        /*
+         * Keep the wrapping ArrayMethod's cached_loop in sync.  Best-effort:
+         * if the lookup fails, the legacy ``functions[]`` slot is still
+         * updated above and the next reduction-cache rebuild would pick it up.
+         */
+        if (func->_loops != NULL) {
+            PyArray_DTypeMeta *sig_dtypes[NPY_MAXARGS];
+            for (j = 0; j < func->nargs; j++) {
+                sig_dtypes[j] = PyArray_DTypeFromTypeNum(signature[j]);
```
KRRT7, line 4949: `PyArray_DTypeFromTypeNum` returns a new reference here, but `PyArray_TupleFromItems(..., make_null_none=0)` increments each item again rather than stealing those references. The temporary `sig_dtypes[j]` references are never decref'ed, so every successful `PyUFunc_ReplaceLoopBySignature` call leaks one DType reference per ufunc argument. This block should either decref the `sig_dtypes` entries after tuple construction or use a helper/path that steals the new references.

KRRT7 (commented):
There is a reference leak in the loop-replacement cache update path.

numpy/_core/src/umath/ufunc_object.c
eendebakpt: Here only the builtin dtypes can be returned which are immortal, so this is no real issue.

Since this path has no performance impact I did add the decref's anyway.

numpy/_core/src/umath/legacy_array_method.c
```diff
@@ -21,20 +21,62 @@
...
+#ifdef Py_GIL_DISABLED
+#include <stdatomic.h>
+#define FT_ATOMIC_LOAD_PTR_ACQUIRE(ptr) \
+    atomic_load_explicit((_Atomic(void *) *)&(ptr), memory_order_acquire)
+#define FT_ATOMIC_STORE_PTR_RELEASE(ptr, val) \
+    atomic_store_explicit((_Atomic(void *) *)&(ptr), (void *)(val), memory_order_release)
+#else
+#define FT_ATOMIC_LOAD_PTR_ACQUIRE(ptr) (ptr)
+#define FT_ATOMIC_STORE_PTR_RELEASE(ptr, val) (ptr) = (val)
+#endif
```
seberg: OK, but maybe you can put this to somewhere central to not duplicate?

numpy/_core/src/umath/ufunc_object.c
seberg: Yeah, need to clean up I think, but also true that in practice dtypes are effectively immortal either way.  (a goto pattern would be nicer prolly, but not going to harp it).

numpy/_core/src/umath/legacy_array_method.c
```diff
@@ -404,15 +458,27 @@ PyArray_NewLegacyWrappingArrayMethod(PyUFuncObject *ufunc,
         }
         if (PyUFunc_DefaultLegacyInnerLoopSelector(
                 ufunc, descrs, &loop, &user_data, &needs_api) < 0) {
+            /* Only ignore a missing loop (a TypeError, `_UFuncNoLoopError`) */
+            if (!PyErr_ExceptionMatches(PyExc_TypeError)) {
```
seberg: Might as well check for `_UFuncNoLoopError`, I guess?

seberg (commented):
Sorry, let's try to move this again (please ping me, maybe even on slack).  Maybe it's a bit minor but in a sense I like this.

One thing that my bot said beyond the comments is that `PyUFunc_FromFuncAndData` isn't copied and thus mutable.  That is now more incorrect than correct since we don't usually look at it.

seberg:
Thanks @eendebakpt let's give this a shot, if there is a fallout we may have to see (I don't really see it, but someone may do weird things)...

On the one hand, I don't love the additional complexity, but conceptually it is an improvement I think and the speedup isn't nothing, so let's give it a shot.

Issue #29117, opened 2025-06-02
BUG: masked array: nanmean(), nanstd(), nanpercentile () raised ValueError in edge cases (all values masked)


orionlee:
### Describe the issue:

(Not sure if it's considered a bug or a feature request.)
For masked array, in the edge case where all values are masked, `nanmean`, `nanstd`, `nanpercentile` would raise `ValueError: output array is read-only`.

### Reproduce the code example:

```python
import numpy as np
print("numpy", np.__version__)

vals, mask = np.array([np.nan, 3]), [True, True]

vals_ma = np.ma.MaskedArray(vals, mask=mask)

def nanpercentile90(a):
    return np.nanpercentile(a, 90)

for f in [np.nanmean, np.mean, np.nanstd, np.std, np.nanmedian, np.median, np.nanmax, np.max, nanpercentile90]:
    v = vals_ma
    try:
        res = f(v)
        print(f.__name__, type(v).__name__, res)
    except Exception as e:
        print(f.__name__, type(v).__name__, f"{type(e).__name__}: {e}")
```

### Error message:

```shell
Stack trace for `nanmean` case:

/usr/local/lib/python3.11/dist-packages/numpy/lib/_nanfunctions_impl.py in nanmean(a, axis, dtype, out, keepdims, where)
   1062     tot = np.sum(arr, axis=axis, dtype=dtype, out=out, keepdims=keepdims,
   1063                  where=where)
-> 1064     avg = _divide_by_count(tot, cnt, out=out)
   1065 
   1066     isbad = (cnt == 0)

/usr/local/lib/python3.11/dist-packages/numpy/lib/_nanfunctions_impl.py in _divide_by_count(a, b, out)
    230         if isinstance(a, np.ndarray):
    231             if out is None:
--> 232                 return np.divide(a, b, out=a, casting='unsafe')
    233             else:
    234                 return np.divide(a, b, out=out, casting='unsafe')

ValueError: output array is read-only
```

### Python and NumPy Versions:

The issue also existed in v1.x branch
```
2.0.2
3.11.12 (main, Apr  9 2025, 08:55:54) [GCC 11.4.0]
```

### Context for the issue:

The issue is identified in the context of binning / down sampling masked timeseries data. The error showed up in the edge case that all values of a bin are masked. 

See https://github.com/astropy/astropy/pull/18188#issuecomment-2929073478

PR #31069, opened 2026-03-25
BUG: Fix ValueError in nanfunctions when MaskedArray is fully masked

prathamhole14:
### PR summary
I ran into this problem and there is a equivalent issue #29117 related to this.

- `nanmean`, `nanstd` throw `ValueError: output array is read-only`, as it was trying to write to the buffer which has no permisson to write to (when everything is nan). Fix (for `nanmean`): removing `out=a` from `np.divide(a, b, out=a, casting='unsafe')` so that division is not done inplace. Subsequently same done for `nanstd`.

 - `nanpercentile`  [np.isnan(arr1d)](https://github.com/numpy/numpy/blob/7baf0e7c71dfc2de9fa5022cf51cb4873efd4285/numpy/lib/_nanfunctions_impl.py#L173) on a MaskedArray returns a MaskedArray. The masked positions in that result have fill_value=False,  and `np.asarray(c)` strips the outer mask and gives boolean data. This data is `or`ed with mask to get the "real" mask so `np.nonzero(c)` would skip them. 

 ```
vals, mask = np.array([np.nan, 3]), [True, True]
vals_ma = np.ma.MaskedArray(vals, mask=mask)
``` 

#### AI Disclosure

No AI tools used
Edit: Later used it to write release note and comments.

### DIFF
--- a/doc/release/upcoming_changes/31069.improvement.rst
+++ b/doc/release/upcoming_changes/31069.improvement.rst
@@ -1,0 +1,3 @@
+``nanfunctions`` now support fully masked ``MaskedArray`` inputs
+----------------------------------------------------------------
+Functions like ``np.nanmean``, ``np.nanstd``, and ``np.nanpercentile`` no longer raise a ``ValueError`` when given a ``MaskedArray`` where all values are masked. For fully masked arrays, ``nanmean`` and ``nanstd`` now correctly return a masked array/constant instead of throwing an error, while ``nanpercentile`` returns ``NaN``.
--- a/doc/release/upcoming_changes/31069.improvement.rst
+++ b/doc/release/upcoming_changes/31069.improvement.rst
@@ -1,6 +1,12 @@
 ``nanfunctions`` now support fully masked ``MaskedArray`` inputs
 ----------------------------------------------------------------
-Functions like ``np.nanmean``, ``np.nanstd``, ``np.nanvar``, and ``np.nanpercentile`` no longer raise a ``ValueError`` when given a ``MaskedArray`` where all values are masked. For fully masked arrays, ``nanmean``, ``nanstd``, and ``nanvar`` now correctly return a masked array/constant instead of throwing an error, while ``nanpercentile`` returns ``NaN``.
+Functions like ``np.nanmean``, ``np.nanstd``, ``np.nanvar``, and
+``np.nanpercentile`` no longer raise a ``ValueError`` when given a
+``MaskedArray`` where all values are masked.
+
+For fully masked arrays, ``nanmean``, ``nanstd``, and ``nanvar`` now correctly
+return a masked array/constant instead of throwing an error, while
+``nanpercentile`` returns ``NaN``.
 
 This also changes behavior for *partially* masked arrays: masked values are now
 also treated as missing, the same as ``NaN``. For example,
--- a/doc/release/upcoming_changes/31069.improvement.rst
+++ b/doc/release/upcoming_changes/31069.improvement.rst
@@ -1,3 +1,9 @@
 ``nanfunctions`` now support fully masked ``MaskedArray`` inputs
 ----------------------------------------------------------------
-Functions like ``np.nanmean``, ``np.nanstd``, and ``np.nanpercentile`` no longer raise a ``ValueError`` when given a ``MaskedArray`` where all values are masked. For fully masked arrays, ``nanmean`` and ``nanstd`` now correctly return a masked array/constant instead of throwing an error, while ``nanpercentile`` returns ``NaN``.
+Functions like ``np.nanmean``, ``np.nanstd``, ``np.nanvar``, and ``np.nanpercentile`` no longer raise a ``ValueError`` when given a ``MaskedArray`` where all values are masked. For fully masked arrays, ``nanmean``, ``nanstd``, and ``nanvar`` now correctly return a masked array/constant instead of throwing an error, while ``nanpercentile`` returns ``NaN``.
+
+This also changes behavior for *partially* masked arrays: masked values are now
+also treated as missing, the same as ``NaN``. For example,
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by prathamhole14** [comment]
I forgot to lint. I will do it after the PR is reviewed. I have mentioned the issue number as references in the code should it be the PR number instead?

**Comment by jorenham** [comment]
the linter exists to make reviewing easier; so it's better to do it *before* we review it

File: numpy/lib/tests/test_nanfunctions.py [hunk 1]
```diff
@@ -1436,3 +1436,21 @@ def test_memmap_takes_fast_route(tmpdir):
...
+
+def test_masked_array_all_masked():
+    """
+    Test that `nan*` functions do not raise a `ValueError` when passed a
+    `MaskedArray` where all values are masked (gh-29117).
+    """
+    vals_ma = np.ma.MaskedArray([np.nan, 3], mask=[True, True])
+
+    with warnings.catch_warnings():
+        warnings.simplefilter('ignore', RuntimeWarning)
```
jorenham: it's cleaner to use the `pytest.mark` decorator for this

prathamhole14:
I have added release note since this is a user facing change, let me know if I should remove it. Thanks

mattip:
closing/reopening to get full CI to run

doc/release/upcoming_changes/31069.improvement.rst
jorenham, line 1: This also changes behavior for partially masked arrays, not just fully masked arrays, e.g. `nanmedian` of `[1,2,3, nan]` with `2` masked now returns `2.0` where it would previously return `masked`, so that should also be mentioned in the relnote.

numpy/lib/tests/test_nanfunctions.py
```diff
@@ -1447,3 +1436,19 @@ def test_memmap_takes_fast_route(tmpdir):
         # For completeness, same for nanmin.
         with pytest.raises(ValueError, match="reduction operation fmin"):
             np.nanmin(mm, out=np.zeros(2))
+
+
+@pytest.mark.filterwarnings("ignore:All-NaN slice:RuntimeWarning")
```
jorenham: it's better to use `pytest.warns` instead of `@pytest.mark.filterwarnings`

numpy/lib/tests/test_nanfunctions.py
```diff
@@ -1447,3 +1436,19 @@ def test_memmap_takes_fast_route(tmpdir):
...
+def test_masked_array_all_masked():
+    """
+    Test that `nan*` functions do not raise a `ValueError` when passed a
+    `MaskedArray` where all values are masked (gh-29117).
+    """
+    vals_ma = np.ma.MaskedArray([np.nan, 3], mask=[True, True])
+
+    for f in [np.nanmean, np.nanstd]:
```
jorenham: it's cleaner to parametrize the test instead of using a for loop like this.

Also, `np.nanvar` and `np.median` are missing

prathamhole14: Thank you for the reviews, I think you mean `np.nanmedian`?

jorenham: ah yea (sorry nan)

doc/release/upcoming_changes/31069.improvement.rst
```diff
@@ -0,0 +1,9 @@
+``nanfunctions`` now support fully masked ``MaskedArray`` inputs
+----------------------------------------------------------------
+Functions like ``np.nanmean``, ``np.nanstd``, ``np.nanvar``, and ``np.nanpercentile`` no longer raise a ``ValueError`` when given a ``MaskedArray`` where all values are masked. For fully masked arrays, ``nanmean``, ``nanstd``, and ``nanvar`` now correctly return a masked array/constant instead of throwing an error, while ``nanpercentile`` returns ``NaN``.
```
jorenham: this could use a linebreak or two

prathamhole14:
I think this can be merged now!

numpy/lib/_nanfunctions_impl.py
```diff
@@ -172,6 +172,10 @@ def _remove_nan_1d(arr1d, second_arr1d=None, overwrite_input=False):
     else:
         c = np.isnan(arr1d)
 
+    # For masked arrays, also treat masked values as if they are NaN (gh-29117).
+    if isinstance(arr1d, np.ma.MaskedArray):
+        c = np.asarray(c) | np.ma.getmaskarray(arr1d)
+
```
mattip: I don't think we should be adding any isintance(.., np.ma.MaskedArray) in non-specific masked array code. We need to think of another approach.

prathamhole14: ah right the `ValueError` output array is read-only and it comes from `_divide_by_count` which calls `np.divide(a, b, out=a)` and `nanstd` which calls `np.sqrt(var, out=var)`. Both assume that reduction returns a writable array.

numpy/lib/tests/test_nanfunctions.py
```diff
@@ -1447,3 +1447,22 @@ def test_memmap_takes_fast_route(tmpdir):
...
+    # the fallback must not promote the result to float64
+    assert_equal(res.dtype, np.float32)
+
+
+@pytest.mark.parametrize("f", [np.nanmean, np.nanstd, np.nanvar])
+def test_all_masked_maskedarray(f):
+    # gh-29117: an all-masked MaskedArray reduces to the read-only
+    # `np.ma.masked`, which these used to try to divide into
+    vals_ma = np.ma.MaskedArray([np.nan, 3], mask=[True, True])
+    assert_(np.ma.is_masked(f(vals_ma)))
```
mattip: masked array tests belong under numpy/ma/tests, not here

mattip (approved):
LGTM, just need to move the test

mattip:
Thanks @prathamhole14

PR #31071, opened 2026-03-25
ENH: convert PyArrayDTypeMeta_Type to a heap type

mattip:
### PR summary

Towards #31026. Convert PyArray_DTypeMeta to a heap type ~that uses multiphase initialization~. I am still worried about performance. I ran `spin bench -c` and the results were all over the place: some slower, some faster, by up to 50%. I tried reversing the commits and running again, the results seemed to be consistent, but still all over the place. Here is an example for just `bench_itemselection`:

| Change   | Before [7baf0e7c] <main>   | After [d2bbb722] <dtype-meta>   |   Ratio | Benchmark (Parameter)                                                    |
|----------|----------------------------|---------------------------------|---------|--------------------------------------------------------------------------|
| +        | 3.73±0.01μs                | 4.06±0.2μs                      |    1.09 | bench_itemselection.Take.time_contiguous((1000, 3), 'wrap', 'complex64') |
| -        | 9.93±0.08μs                | 9.41±0.02μs                     |    0.95 | bench_itemselection.Take.time_contiguous((2, 1000, 1), 'wrap', 'O')      |
| -        | 23.0±0.2μs                 | 17.7±0.2μs                      |    0.77 | bench_itemselection.Take.time_contiguous((1000, 1), 'clip', 'i,O')       |
| -        | 32.5±0.1μs                 | 24.5±0.06μs                     |    0.75 | bench_itemselection.Take.time_contiguous((1000, 3), 'clip', 'i,O')       |
| -        | 45.5±0.5μs                 | 33.8±0.3μs                      |    0.74 | bench_itemselection.Take.time_contiguous((2, 1000, 1), 'clip', 'i,O')    |

 Change   | Before [d2bbb722] <dtype-meta>   | After [7baf0e7c] <main>   |   Ratio | Benchmark (Parameter)                                                        |
|----------|----------------------------------|---------------------------|---------|------------------------------------------------------------------------------|
| +        | 17.5±0.3μs                       | 23.2±0.3μs                |    1.33 | bench_itemselection.Take.time_contiguous((1000, 1), 'clip', 'i,O')           |
| +        | 33.9±0.4μs                       | 45.0±0.8μs                |    1.33 | bench_itemselection.Take.time_contiguous((2, 1000, 1), 'clip', 'i,O')        |
| +        | 25.3±0.7μs                       | 32.6±0.4μs                |    1.29 | bench_itemselection.Take.time_contiguous((1000, 3), 'clip', 'i,O')           |
| -        | 808±10ns                         | 764±20ns                  |    0.95 | bench_itemselection.PutMask.time_sparse(True, 'int64')                       |
| -        | 1.27±0.01μs                      | 1.20±0.01μs               |    0.95 | bench_itemselection.Take.time_contiguous((1000, 1), 'wrap', 'complex128')    |
| -        | 1.16±0.01μs                      | 1.10±0μs                  |    0.95 | bench_itemselection.Take.time_contiguous((1000, 1), 'wrap', 'float64')       |
| -        | 2.28±0.05μs                      | 2.16±0.01μs               |    0.95 | bench_itemselection.Take.time_contiguous((2, 1000, 1), 'clip', 'complex128') |
| -        | 943±7ns                          | 891±6ns                   |    0.94 | bench_itemselection.PutMask.time_dense(True, 'float32')                      |
| -        | 948±6ns                          | 891±30ns                  |    0.94 | bench_itemselection.PutMask.time_sparse(False, 'float16')                    |
| -        | 837±9ns                          | 786±10ns                  |    0.94 | bench_itemselection.PutMask.time_sparse(True, 'float16')                     |
| -        | 842±6ns                          | 794±7ns                   |    0.94 | bench_itemselection.PutMask.time_sparse(True, 'int16')                       |
| -        | 969±10ns                         | 901±20ns                  |    0.93 | bench_itemselection.PutMask.time_dense(True, 'int16')                        |
| -        | 4.51±0.4μs                       | 3.88±0μs                  |    0.86 | bench_itemselection.Take.time_contiguous((1000, 3), 'clip', 'int16')         |
| -        | 4.62±0.02μs                      | 3.87±0.01μs               |    0.84 | bench_itemselection.Take.time_contiguous((1000, 3), 'clip', 'float16')       |
| -        | 3.89±0.4μs                       | 3.28±0.05μs               |    0.84 | bench_itemselection.Take.time_contiguous((1000, 3), 'raise', 'float64')      |

#### AI Disclosure
I used claude code to help me. I went over every change to make sure it was required.

### DIFF
--- a/numpy/_core/include/numpy/ndarraytypes.h
+++ b/numpy/_core/include/numpy/ndarraytypes.h
@@ -1876,9 +1876,8 @@ typedef struct {
 } PyArray_StringDTypeObject;
 
 /*
- * PyArray_DTypeMeta related definitions.
- *
- * As of now, this API is preliminary and will be extended as necessary.
+ * PyArray_DTypeMeta related definitions. Only for internal use, should be moved
+ * away from public headers eventually
  */
 #if defined(NPY_INTERNAL_BUILD) && NPY_INTERNAL_BUILD
     /*
@@ -1887,18 +1886,20 @@ typedef struct {
      * Part of this (at least the size) is expected to be public API without
      * further modifications.
      */
-    /* TODO: Make this definition public in the API, as soon as its settled */
-    NPY_NO_EXPORT extern PyTypeObject PyArrayDTypeMeta_Type;
+    /*
+     * This is made public in _public_dtype_api_table.h via the PyArray_API table,
+     * but there it is a dereferenced pointer
+     */
+    NPY_NO_EXPORT extern PyTypeObject *PyArrayDTypeMeta_Type;
 
     /*
      * While NumPy DTypes would not need to be heap types the plan is to
      * make DTypes available in Python at which point they will be heap types.
+     *
      * Since we also wish to add fields to the DType class, this looks like
      * a typical instance definition, but with PyHeapTypeObject instead of
      * only the PyObject_HEAD.
-     * This must only be exposed very extremely careful consideration, since
-     * it is a fairly complex construct which may be better to allow
-     * refactoring of.
+     * This is exposed in dtype_api.h
      */
     typedef struct {
         PyHeapTypeObject super;
--- a/numpy/_core/src/multiarray/dtypemeta.c
+++ b/numpy/_core/src/multiarray/dtypemeta.c
@@ -1148,7 +1148,7 @@ dtypemeta_wrap_legacy_descriptor(
      */
     static PyArray_DTypeMeta prototype = {
         {{
-            PyVarObject_HEAD_INIT(&PyArrayDTypeMeta_Type, 0)
+            PyVarObject_HEAD_INIT(NULL, 0)
             .tp_name = NULL,  /* set below */
             .tp_basicsize = sizeof(_PyArray_LegacyDescr),
             .tp_flags = Py_TPFLAGS_DEFAULT,
@@ -1160,6 +1160,7 @@ dtypemeta_wrap_legacy_descriptor(
         /* Further fields are not common between DTypes */
     };
     memcpy(dtype_class, &prototype, sizeof(PyArray_DTypeMeta));
+    Py_SET_TYPE((PyTypeObject *)dtype_class, PyArrayDTypeMeta_Type);
     /* Fix name and superclass of the Type*/
     ((PyTypeObject *)dtype_class)->tp_name = name;
     ((PyTypeObject *)dtype_class)->tp_base = dtype_super_class,
@@ -1310,6 +1311,28 @@ dtypemeta_get_is_numeric(PyArray_DTypeMeta *self, void *NPY_UNUSED(ignored)) {
     return PyBool_FromLong(NPY_DT_is_numeric(self));
 }
 
+static PyObject *
+dtypemeta_getattro(PyObject *self, PyObject *name)
+{
+    /*
+     * Legacy (non-heap) DType types store their module in tp_name as
+     * "module.classname" (e.g. "numpy.dtypes.TimeDelta64DType"), but their
+     * MRO does not include PyType_Type so type.__module__'s data descriptor
+     * is never found during attribute lookup, causing __module__ to fall back
+     * to the string stored in _DTypeMeta's tp_dict by PyType_FromSpecWithBases.
+     * Intercept here and compute the correct module from tp_name.
+     * TODO: remove this after converting all legacy DTypes to heap types.
+     */
+    if (PyUnicode_CompareWithASCIIString(name, "__module__") == 0
+            && !(((PyTypeObject *)self)->tp_flags & Py_TPFLAGS_HEAPTYPE)) {
+        const char *tp_name = ((PyTypeObject *)self)->tp_name;
+        const char *dot = strrchr(tp_name, '.');
+        assert(dot != NULL);
+        return PyUnicode_FromStringAndSize(tp_name, dot - tp_name);
+    }
+    return PyType_Type.tp_getattro(self, name);
+}
+
 /*
  * Simple exposed information, defined for each DType (class).
  */
@@ -1368,23 +1391,41 @@ initialize_legacy_dtypemeta_aliases(_PyArray_LegacyDescr **_builtin_descrs) {
     _Void_dtype = NPY_DTYPE(_builtin_descrs[NPY_VOID]);
 }
 
-NPY_NO_EXPORT PyTypeObject PyArrayDTypeMeta_Type = {
-    PyVarObject_HEAD_INIT(NULL, 0)
-    .tp_name = "numpy._DTypeMeta",
-    .tp_basicsize = sizeof(PyArray_DTypeMeta),
-    .tp_dealloc = (destructor)dtypemeta_dealloc,
+static PyType_Slot dtypemeta_slots[] = {
+    {Py_tp_dealloc, (void *)dtypemeta_dealloc},
     /* Types are garbage collected (see dtypemeta_is_gc documentation) */
-    .tp_flags = Py_TPFLAGS_DEFAULT | Py_TPFLAGS_HAVE_GC,
-    .tp_doc = "Preliminary NumPy API: The Type of NumPy DTypes (metaclass)",
-    .tp_traverse = (traverseproc)dtypemeta_traverse,
-    .tp_members = dtypemeta_members,
-    .tp_getset = dtypemeta_getset,
-    .tp_base = NULL,  /* set to PyType_Type at import time */
-    .tp_init = (initproc)dtypemeta_init,
-    .tp_alloc = dtypemeta_alloc,
-    .tp_new = dtypemeta_new,
-    .tp_is_gc = dtypemeta_is_gc,
+    {Py_tp_traverse, (void *)dtypemeta_traverse},
+    {Py_tp_members, (void *)dtypemeta_members},
+    {Py_tp_getset, (void *)dtypemeta_getset},
+    {Py_tp_getattro, (void *)dtypemeta_getattro},
+    {Py_tp_init, (void *)dtypemeta_init},
+    {Py_tp_alloc, (void *)dtypemeta_alloc},
+    {Py_tp_new, (void *)dtypemeta_new},
+    {Py_tp_is_gc, (void *)dtypemeta_is_gc},
+    {Py_tp_doc, (void *)"Preliminary NumPy API: The Type of NumPy DTypes (metaclass)"},
+    {0, NULL},
 };
+
+static PyType_Spec dtypemeta_spec = {
+    .name = "numpy._DTypeMeta",
+    .basicsize = sizeof(PyArray_DTypeMeta),
+    .flags = Py_TPFLAGS_DEFAULT | Py_TPFLAGS_HAVE_GC,
+    .slots = dtypemeta_slots,
+};
+
+NPY_NO_EXPORT PyTypeObject *PyArrayDTypeMeta_Type = NULL;
+
+NPY_NO_EXPORT int
+PyArrayDTypeMeta_Type_init(void) {
+    PyObject *bases = PyTuple_Pack(1, &PyType_Type);
+    if (bases == NULL) {
+        return -1;
+    }
+    PyArrayDTypeMeta_Type = (PyTypeObject *)PyType_FromSpecWithBases(
+            &dtypemeta_spec, bases);
+    Py_DECREF(bases);
+    return (PyArrayDTypeMeta_Type != NULL) ? 0 : -1;
+}
 
 PyArray_DTypeMeta *_Bool_dtype = NULL;
 PyArray_DTypeMeta *_Byte_dtype = NULL;
--- a/numpy/_core/src/multiarray/dtypemeta.c
+++ b/numpy/_core/src/multiarray/dtypemeta.c
@@ -1409,7 +1409,7 @@ static PyType_Slot dtypemeta_slots[] = {
 static PyType_Spec dtypemeta_spec = {
     .name = "numpy._DTypeMeta",
     .basicsize = sizeof(PyArray_DTypeMeta),
-    .flags = Py_TPFLAGS_DEFAULT | Py_TPFLAGS_HAVE_GC,
+    .flags = Py_TPFLAGS_DEFAULT | Py_TPFLAGS_HAVE_GC | Py_TPFLAGS_IMMUTABLETYPE,
     .slots = dtypemeta_slots,
 };
 
--- a/numpy/_core/src/multiarray/stringdtype/dtype.c
+++ b/numpy/_core/src/multiarray/stringdtype/dtype.c
@@ -1011,7 +1011,7 @@ init_string_dtype(void)
     };
 
     /* Loaded dynamically, so needs to be set here: */
-    ((PyObject *)&PyArray_StringDType)->ob_type = &PyArrayDTypeMeta_Type;
+    ((PyObject *)&PyArray_StringDType)->ob_type = PyArrayDTypeMeta_Type;
     ((PyTypeObject *)&PyArray_StringDType)->tp_base = &PyArrayDescr_Type;
     if (PyType_Ready((PyTypeObject *)&PyArray_StringDType) < 0) {
         return -1;
--- a/numpy/_core/src/multiarray/abstractdtypes.c
+++ b/numpy/_core/src/multiarray/abstractdtypes.c
@@ -85,12 +85,15 @@ discover_descriptor_from_pycomplex(
 NPY_NO_EXPORT int
 initialize_and_map_pytypes_to_dtypes()
 {
+    Py_SET_TYPE((PyTypeObject *)&PyArray_IntAbstractDType, PyArrayDTypeMeta_Type);
     if (PyType_Ready((PyTypeObject *)&PyArray_IntAbstractDType) < 0) {
         return -1;
     }
+    Py_SET_TYPE((PyTypeObject *)&PyArray_FloatAbstractDType, PyArrayDTypeMeta_Type);
     if (PyType_Ready((PyTypeObject *)&PyArray_FloatAbstractDType) < 0) {
         return -1;
     }
+    Py_SET_TYPE((PyTypeObject *)&PyArray_ComplexAbstractDType, PyArrayDTypeMeta_Type);
     if (PyType_Ready((PyTypeObject *)&PyArray_ComplexAbstractDType) < 0) {
         return -1;
     }
@@ -101,18 +104,21 @@ initialize_and_map_pytypes_to_dtypes()
     ((PyTypeObject *)&PyArray_PyLongDType)->tp_base =
         (PyTypeObject *)&PyArray_IntAbstractDType;
     PyArray_PyLongDType.scalar_type = &PyLong_Type;
+    Py_SET_TYPE((PyTypeObject *)&PyArray_PyLongDType, PyArrayDTypeMeta_Type);
     if (PyType_Ready((PyTypeObject *)&PyArray_PyLongDType) < 0) {
         return -1;
     }
     ((PyTypeObject *)&PyArray_PyFloatDType)->tp_base =
         (PyTypeObject *)&PyArray_FloatAbstractDType;
     PyArray_PyFloatDType.scalar_type = &PyFloat_Type;
+    Py_SET_TYPE((PyTypeObject *)&PyArray_PyFloatDType, PyArrayDTypeMeta_Type);
     if (PyType_Ready((PyTypeObject *)&PyArray_PyFloatDType) < 0) {
         return -1;
     }
     ((PyTypeObject *)&PyArray_PyComplexDType)->tp_base =
         (PyTypeObject *)&PyArray_ComplexAbstractDType;
     PyArray_PyComplexDType.scalar_type = &PyComplex_Type;
+    Py_SET_TYPE((PyTypeObject *)&PyArray_PyComplexDType, PyArrayDTypeMeta_Type);
     if (PyType_Ready((PyTypeObject *)&PyArray_PyComplexDType) < 0) {
         return -1;
     }
@@ -292,7 +298,7 @@ complex_common_dtype(PyArray_DTypeMeta *cls, PyArray_DTypeMeta *other)
  * Here, also define types corresponding to the python scalars.
  */
 NPY_NO_EXPORT PyArray_DTypeMeta PyArray_IntAbstractDType = {{{
-        PyVarObject_HEAD_INIT(&PyArrayDTypeMeta_Type, 0)
+        PyVarObject_HEAD_INIT(NULL, 0)
         .tp_name = "numpy.dtypes._IntegerAbstractDType",
         .tp_base = &PyArrayDescr_Type,
         .tp_basicsize = sizeof(PyArray_Descr),
@@ -309,7 +315,7 @@ NPY_DType_Slots pylongdtype_slots = {
 };
 
 NPY_NO_EXPORT PyArray_DTypeMeta PyArray_PyLongDType = {{{
-        PyVarObject_HEAD_INIT(&PyArrayDTypeMeta_Type, 0)
+        PyVarObject_HEAD_INIT(NULL, 0)
         .tp_name = "numpy.dtypes._PyLongDType",
         .tp_base = NULL,  /* set in initialize_and_map_pytypes_to_dtypes */
         .tp_basicsize = sizeof(PyArray_Descr),
@@ -321,7 +327,7 @@ NPY_NO_EXPORT PyArray_DTypeMeta PyArray_PyLongDType = {{{
 };
 
 NPY_NO_EXPORT PyArray_DTypeMeta PyArray_FloatAbstractDType = {{{
-        PyVarObject_HEAD_INIT(&PyArrayDTypeMeta_Type, 0)
+        PyVarObject_HEAD_INIT(NULL, 0)
         .tp_name = "numpy.dtypes._FloatAbstractDType",
         .tp_base = &PyArrayDescr_Type,
         .tp_basicsize = sizeof(PyArray_Descr),
@@ -338,7 +344,7 @@ NPY_DType_Slots pyfloatdtype_slots = {
 };
 
 NPY_NO_EXPORT PyArray_DTypeMeta PyArray_PyFloatDType = {{{
-        PyVarObject_HEAD_INIT(&PyArrayDTypeMeta_Type, 0)
+        PyVarObject_HEAD_INIT(NULL, 0)
         .tp_name = "numpy.dtypes._PyFloatDType",
         .tp_base = NULL,  /* set in initialize_and_map_pytypes_to_dtypes */
         .tp_basicsize = sizeof(PyArray_Descr),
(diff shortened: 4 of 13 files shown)


numpy/_core/include/numpy/ndarraytypes.h
```diff
@@ -1887,18 +1886,20 @@ typedef struct {
      * Part of this (at least the size) is expected to be public API without
      * further modifications.
      */
-    /* TODO: Make this definition public in the API, as soon as its settled */
-    NPY_NO_EXPORT extern PyTypeObject PyArrayDTypeMeta_Type;
+    /*
+     * This is made public in _public_dtype_api_table.h via the PyArray_API table,
+     * but there it is a dereferenced pointer
+     */
+    NPY_NO_EXPORT extern PyTypeObject *PyArrayDTypeMeta_Type;
```
mattip, line 1893: Note this is behind `NPY_INTERNAL_BUILD`. I think it is still here for historical reasons, but maybe we should be `#pragma warn`-ing that we will move it out of these headers soon. The definition of ~`PyHeapTypeObject`~ `PyArray_DTypeMeta` is duplicated in `dtype_api.h`

Edit PyHeapTypeObject -> PyArray_DTypeMeta

numpy/_core/src/multiarray/stringdtype/dtype.c
```diff
@@ -1011,7 +1011,7 @@ init_string_dtype(void)
     };
 
     /* Loaded dynamically, so needs to be set here: */
-    ((PyObject *)&PyArray_StringDType)->ob_type = &PyArrayDTypeMeta_Type;
+    ((PyObject *)&PyArray_StringDType)->ob_type = PyArrayDTypeMeta_Type;
```
mattip, line 1014: Lots of this PR are these changes: `&PyArrayDTypeMeta_Type` becomes `PyArrayDTypeMeta_Type` since `PyArrayDTypeMeta_Type` is now a pointer.

numpy/_core/src/multiarray/dtypemeta.c
```diff
@@ -1310,6 +1311,28 @@ dtypemeta_get_is_numeric(PyArray_DTypeMeta *self, void *NPY_UNUSED(ignored)) {
...
+     */
+    if (PyUnicode_CompareWithASCIIString(name, "__module__") == 0
+            && !(((PyTypeObject *)self)->tp_flags & Py_TPFLAGS_HEAPTYPE)) {
+        const char *tp_name = ((PyTypeObject *)self)->tp_name;
+        const char *dot = strrchr(tp_name, '.');
+        assert(dot != NULL);
+        return PyUnicode_FromStringAndSize(tp_name, dot - tp_name);
+    }
+    return PyType_Type.tp_getattro(self, name);
+}
```
mattip, line 1334: claude helped me with this solution to getting the `__module__` name be consistent with current behaviour when the metaclass is a heap type but the real type is still a static type.

numpy/_core/src/multiarray/dtypemeta.c
```diff
@@ -1148,7 +1148,7 @@ dtypemeta_wrap_legacy_descriptor(
      */
     static PyArray_DTypeMeta prototype = {
         {{
-            PyVarObject_HEAD_INIT(&PyArrayDTypeMeta_Type, 0)
+            PyVarObject_HEAD_INIT(NULL, 0)
```
mattip, line 1151: Since `PyArrayDTypeMeta_Type` is now a pointer, it cannot be used in the macro. So use NULL here and `Py_SET_TYPE` below.

mattip (commented):
A few clarifications.

mattip:
I tried just the benchmarks that showed the most slowdown,
```
spin bench -c -t bench_ufunc_strides.UnaryFP --cpu-affinity 0
```

which consistently reproduced the slowdown for float16 ufuncs:

| Change   | Before [7baf0e7c] <main>   | After [d2bbb722] <dtype-meta>   |   Ratio | Benchmark (Parameter)                                                             |
|----------|----------------------------|---------------------------------|---------|-----------------------------------------------------------------------------------|
| +        | 4.40±0.3ms                 | 6.69±0.04ms                     |    1.52 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sqrt'>, 1, 1, 'e')                 |
| +        | 4.40±0.1ms                 | 6.65±0.07ms                     |    1.51 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sqrt'>, 4, 1, 'e')                 |
| +        | 4.39±0.3ms                 | 6.58±0.03ms                     |    1.5  | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'fabs'>, 1, 1, 'e')                 |
| +        | 4.41±0.3ms                 | 6.57±0.03ms                     |    1.49 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'ceil'>, 1, 1, 'e')                 |
| +        | 4.39±0.3ms                 | 6.56±0.4ms                      |    1.49 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'ceil'>, 4, 2, 'e')                 |
| +        | 4.45±0.3ms                 | 6.66±0.2ms                      |    1.49 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'fabs'>, 4, 2, 'e')                 |
| +        | 5.17±0.09ms                | 7.70±0.04ms                     |    1.49 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sqrt'>, 4, 2, 'e')          |
| +        | 4.45±0.3ms                 | 6.59±0.02ms                     |    1.48 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'degrees'>, 1, 1, 'e')              |
| +        | 4.45±0.4ms                 | 6.57±0.04ms                     |    1.48 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'degrees'>, 4, 1, 'e')              |
| +        | 4.45±0.04ms                | 6.61±0.3ms                      |    1.48 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'degrees'>, 4, 2, 'e')              |
| +        | 4.64±0.07ms                | 6.87±0.08ms                     |    1.48 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'floor'>, 4, 1, 'e')                |
| +        | 4.47±0.05ms                | 6.63±0.5ms                      |    1.48 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sqrt'>, 4, 2, 'e')                 |
| +        | 5.20±0.5ms                 | 7.70±0.05ms                     |    1.48 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'sqrt'>, 4, 1, 'e')          |
| +        | 4.45±0.04ms                | 6.56±0.01ms                     |    1.47 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'fabs'>, 4, 1, 'e')                 |
| +        | 4.64±0.3ms                 | 6.81±0.5ms                      |    1.47 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'floor'>, 4, 2, 'e')                |
| +        | 4.45±0.4ms                 | 6.56±0.02ms                     |    1.47 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'rad2deg'>, 1, 2, 'e')              |
| +        | 4.50±0.7ms                 | 6.56±0.02ms                     |    1.46 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'rad2deg'>, 1, 1, 'e')              |
| +        | 4.64±0.3ms                 | 6.77±0.03ms                     |    1.46 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'trunc'>, 1, 2, 'e')                |
| +        | 5.01±0.5ms                 | 7.26±0.5ms                      |    1.45 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'fabs'>, 4, 1, 'e')          |
| +        | 5.14±0.4ms                 | 7.23±0.1ms                      |    1.41 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'ceil'>, 4, 1, 'e')          |
| +        | 4.77±0.4ms                 | 6.63±0.08ms                     |    1.39 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sqrt'>, 1, 2, 'e')                 |
| +        | 5.03±0.5ms                 | 6.88±0.08ms                     |    1.37 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'floor'>, 1, 2, 'e')                |
| +        | 5.13±0.4ms                 | 6.55±0.03ms                     |    1.28 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'ceil'>, 1, 2, 'e')                 |
| +        | 5.18±0.3ms                 | 6.56±0.03ms                     |    1.26 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'degrees'>, 1, 2, 'e')              |
| +        | 5.87±0.06ms                | 6.50±0.03ms                     |    1.11 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'log'>, 1, 1, 'd')           |
| +        | 8.27±0.2ms                 | 8.73±0.08ms                     |    1.06 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'log1p'>, 4, 1, 'd')         |
| -        | 7.17±0.1ms                 | 6.81±0.04ms                     |    0.95 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'cos'>, 1, 1, 'e')           |
| -        | 1.85±0.03ms                | 1.74±0.02ms                     |    0.94 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'square'>, 1, 2, 'd')        |
| -        | 1.90±0.05ms                | 1.77±0.02ms                     |    0.93 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'square'>, 1, 2, 'd')               |
| -        | 1.54±0.05ms                | 1.42±0.02ms                     |    0.92 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'rint'>, 4, 2, 'f')                 |
| -        | 1.56±0.06ms                | 1.44±0ms                        |    0.92 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'conjugate'> (0), 4, 2, 'f') |
| -        | 6.70±0.2ms                 | 6.18±0.05ms                     |    0.92 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'exp2'>, 4, 1, 'e')          |
| -        | 1.94±0.06ms                | 1.77±0ms                        |    0.91 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'conjugate'> (1), 1, 2, 'd') |
| -        | 1.54±0.03ms                | 1.39±0.02ms                     |    0.9  | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'rint'>, 4, 2, 'f')          |
| -        | 3.88±0.2ms                 | 3.34±0.01ms                     |    0.86 | bench_ufunc_strides.UnaryFPSpecial.time_unary(<ufunc 'log2'>, 4, 2, 'f')          |

kumaraditya303:
> Convert PyArray_DTypeMeta to a heap type that uses multiphase initialization. I am still worried about performance. 

Heap types and multiphase initialization are two different things, I don't think you are doing multi-phase init in this PR. 

A few thoughts on this:

- Heap types are mortal so every object allocated needs to do reference counting, this will affect performance on ft so you may want to make the type immortal.
- While we are doing the conversion it would be good to remove the need for `PyHeapTypeObject` for this.
- You can probably gain some performance back depending on workload by making the type immutable like it was when it was static type.

mattip:
Thanks for taking a look.

> I don't think you are doing multi-phase init in this PR.

I [switched `PyType_Ready` for `PyArrayDTypeMeta_Type_init`](https://github.com/numpy/numpy/pull/31071/changes#diff-71aadf0228c9cd61ea260748af1dcd4a726eea8476b8cbd667367466c1e39077L4995).

> you may want to make the type immortal

I [essentially did](https://github.com/numpy/numpy/pull/31071/changes#diff-c7a7b47cc7364674777daee63c40581084f8070077929f96d9576d2a58b0d8d6R1424), since the global `PyArrayDTypeMeta_Type` is never decrefed. When we get to supporting subinterpreters we will have to carefully examine the lifetimes of all the global elements in the API data structs.

> it would be good to remove the need for `PyHeapTypeObject` for this

Which "this"? Maybe you are relating to my comment in the PR which mixed up `PyHeapTypeObject` and `PyArray_DTypeMeta`? I could break out the changes to make `PyArrayDTypeMeta_Type` into a pointer and remove the entire `#if defined(NPY_INTERNAL_BUILD) && NPY_INTERNAL_BUILD` clause from `numpy/_core/include/numpy/ndarraytypes.h` into a separate PR, these changes are fallout from this PR but should be harmless.

> making the type immutable

Good idea, added this flag. Unfortunately it did not help the decreased performance on some ufunc calls:
```
spin bench -c -t bench_ufunc_strides.UnaryFP.time_unary --cpu-affinity 0
```
| Change   | Before [7baf0e7c] <main>   | After [654fdd74] <dtype-meta>   |   Ratio | Benchmark (Parameter)                                                 |
|----------|----------------------------|---------------------------------|---------|-----------------------------------------------------------------------|
| +        | 4.47±0.01ms                | 6.80±0.3ms                      |    1.52 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'degrees'>, 4, 2, 'e')  |
| +        | 4.46±0.3ms                 | 6.74±0.08ms                     |    1.51 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'fabs'>, 4, 2, 'e')     |
| +        | 4.49±0.3ms                 | 6.74±0.03ms                     |    1.5  | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'rad2deg'>, 4, 2, 'e')  |
| +        | 4.48±0.5ms                 | 6.67±0.09ms                     |    1.49 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'fabs'>, 1, 2, 'e')     |
| +        | 4.53±0.3ms                 | 6.74±0.1ms                      |    1.49 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'rad2deg'>, 1, 1, 'e')  |
| +        | 4.48±0.02ms                | 6.69±0.03ms                     |    1.49 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'rad2deg'>, 4, 1, 'e')  |
| +        | 4.50±0.01ms                | 6.70±0.08ms                     |    1.49 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sqrt'>, 4, 1, 'e')     |
| +        | 4.51±0.04ms                | 6.70±0.1ms                      |    1.48 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'degrees'>, 1, 2, 'e')  |
| +        | 4.50±0.1ms                 | 6.60±0.2ms                      |    1.47 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'fabs'>, 1, 1, 'e')     |
| +        | 4.51±0.4ms                 | 6.60±0.5ms                      |    1.46 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'sqrt'>, 1, 1, 'e')     |
| +        | 4.77±0.1ms                 | 6.91±0.06ms                     |    1.45 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'floor'>, 4, 1, 'e')    |
| +        | 4.75±0.4ms                 | 6.73±0.2ms                      |    1.42 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'ceil'>, 1, 2, 'e')     |
| +        | 4.85±0.4ms                 | 6.79±0.07ms                     |    1.4  | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'fabs'>, 4, 1, 'e')     |
| -        | 6.72±0.04ms                | 6.27±0.01ms                     |    0.93 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'log2'>, 1, 1, 'e')     |
| -        | 4.07±0.2ms                 | 3.70±0.02ms                     |    0.91 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'rint'>, 4, 2, 'd')     |
| -        | 2.26±0.2ms                 | 2.02±0.01ms                     |    0.89 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'negative'>, 1, 2, 'd') |
| -        | 1.83±0.08ms                | 1.63±0.04ms                     |    0.89 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'rint'>, 4, 2, 'f')     |
| -        | 4.82±0.5ms                 | 4.24±0.05ms                     |    0.88 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'log2'>, 4, 1, 'd')     |
| -        | 2.30±0.06ms                | 2.02±0.01ms                     |    0.88 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'square'>, 1, 2, 'd')   |
| -        | 2.45±0.2ms                 | 2.11±0.02ms                     |    0.86 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'rad2deg'>, 1, 2, 'd')  |
| -        | 2.37±0.2ms                 | 2.01±0.03ms                     |    0.85 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'absolute'>, 1, 2, 'd') |
| -        | 2.37±0.09ms                | 2.01±0.01ms                     |    0.85 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'rint'>, 1, 2, 'd')     |
| -        | 2.41±0.3ms                 | 2.03±0.01ms                     |    0.84 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'positive'>, 1, 2, 'd') |
| -        | 1.05±0.06ms                | 825±6μs                         |    0.79 | bench_ufunc_strides.UnaryFP.time_unary(<ufunc 'rint'>, 4, 1, 'f')     |

The strange thing about this regression is that `PyArray_DTypeMeta` is not involved after the singleton dtype type instances are created, so I don't see how it can impact performance like this, except for the getattr function now added to the meta type. Maybe the benchmarks are sensitive to memory layout?
```

**Comment by ngoldbaum** [comment]
Just to keep the terminology straight: Multi-phase initialization is a concept for [module objects](https://docs.python.org/3/c-api/extension-modules.html#multi-phase-initialization). It doesn't have anything to do with initializing type objects.

**Comment by mattip** [comment]
Thanks for the clarification. So any change to heap types will change PyType_Ready to use an init?

File: numpy/_core/src/multiarray/stringdtype/dtype.c (hunk 2 above)
Comment (seberg) [reply numpy/_core/src/multiarray/stringdtype/dtype.c:1014]: Wondering a bit if we can't to a dance to keep the & even if it is via a weird macro?

That is mainly because we can't change the public API and that is set up with a macro to use `&PyArrayDTypeMeta_Type` even though it is kinda not static...

So this diverges internal and public API which is a bit tedious.  I don't have a big opinion, I agree that pointers are nicer, but...

File: numpy/_core/src/multiarray/dtypemeta.c (hunk 3 above)
Comment (seberg) [reply numpy/_core/src/multiarray/dtypemeta.c:1334]: Sorry, -1 on this. Claude came up with a bad solution.

Not sure what is best, what DTypes are still static after this or can be both? ~I thought this would make them all heap types?!~

Even if both are mixed, i assume you can write a `__module__` getter that may have to directly look up from the `__dict__` for heap-types.
But a custom `getattro` is a nuclear option, IMO, that really should only be used for strange types that need to do dynamic things.
(This isn't likek module `__getattr__` that only get's called if the entry is missing from the dict, as I understand it.)

I am honestly a little bit confused why any of this matters at all, does making this a heap-type modify the actual DTypes (DTypeMeta instances) to be proper heaptypes now even though they still use `PyType_Ready`?

File: numpy/_core/src/multiarray/dtypemeta.c [hunk 5]
```diff
@@ -1368,24 +1391,42 @@ initialize_legacy_dtypemeta_aliases(_PyArray_LegacyDescr **_builtin_descrs) {
...
+PyArrayDTypeMeta_Type_init(void) {
+    PyObject *bases = PyTuple_Pack(1, &PyType_Type);
+    if (bases == NULL) {
+        return -1;
+    }
+    PyArrayDTypeMeta_Type = (PyTypeObject *)PyType_FromSpecWithBases(
+            &dtypemeta_spec, bases);
+    Py_DECREF(bases);
+    return (PyArrayDTypeMeta_Type != NULL) ? 0 : -1;
+}
```
seberg, line 1428: Nice!

seberg (commented):
Please ignore the timings and all discussions about it?

This changes the **metatype**, I can't think of anything that would modify its refcount, the only thing I could fathom changes is the `isinstance()`/type checks and those usually hit `==` fast-paths probably (also we don't usually check for the meta type isntances, I expect).
Either way, ufunc changes would be in the range of a few `ns` at best, so `ms` changes are clearly unrelated noise.
(I dunno if there is anything where many `ns` changes could even add up much in the numpy API for *this*)

The interesting part will be to change the init for `PyArray_PyFloatDType`, but speed wise, I also expect that to be largely irrelevant.

Either way, not approving yet, because the `getattro` use would be a desperate solution to me and I doubt we have to go to such desperation.

numpy/_core/src/multiarray/stringdtype/dtype.c
mattip: What if we use a different name `PyArrayDTypeMeta_TypeP` internally, and add the internal name to the external API, like
```
NPY_NO_EXPORT extern PyTypeObject *PyArrayDTypeMeta_TypeP;
//what we have now
#define PyArrayDTypeMeta_Type (*(PyTypeObject *)(PyArray_API + 320)[0])
//what we add
#define PyArrayDTypeMeta_TypeP (PyTypeObject *)(PyArray_API + 320)[0]
```

Edit: please bike shed the name

numpy/_core/src/multiarray/dtypemeta.c
mattip: No need to be sorry, claude doesn't care. This change modifies the `__module__` lookup path for the static dtypes with a heap-type metaclass, and if I recall correctly changed the return value to cut off the `dtype`: i.e. `numpy.int32` instead of `numpy.dtype.int32`. Once we move the static dtypes to heaptypes, we can remove this code. What would a `__module__`-specific getter look like?

seberg: Yeah, I just don't like overriding `tp_getattro`.

> What would a `__module__`-specific getter look like?

Add a `tp_getset` slot to the `dtypemeta_getset` for `__module__` that has exactly the same code (hopefully).

> i.e. `numpy.int32` instead of `numpy.dtype.int32`

might be something like `numpy.dtypes` vs. just `numpy` (or even empty).  Might be good to check that the `__name__` is correct for e.g. `numpy.dtypes.Int32DType`?

If `__module__` is confused, it seems likely `__name__` might also be and include the module part?  Maybe check once, dunno if we have a test or not (although it seems likely that we do).

numpy/_core/src/multiarray/stringdtype/dtype.c
seberg: I am fine with adding `TypeP` or `TypePtr` whatever you like. Dunno if I care about public API unless we add it for all types, but also happy to just add it, it is all just repetitive after all.

EDIT: Hmmm, I was wondering what Pyhon uses, but I am not sure there is any object that uses a pointer? (i guess they are all static if exposed though).

numpy/_core/src/multiarray/dtypemeta.c
```diff
@@ -1368,24 +1391,42 @@ initialize_legacy_dtypemeta_aliases(_PyArray_LegacyDescr **_builtin_descrs) {
...
-    .tp_doc = "Preliminary NumPy API: The Type of NumPy DTypes (metaclass)",
-    .tp_traverse = (traverseproc)dtypemeta_traverse,
-    .tp_members = dtypemeta_members,
-    .tp_getset = dtypemeta_getset,
-    .tp_base = NULL,  /* set to PyType_Type at import time */
-    .tp_init = (initproc)dtypemeta_init,
-    .tp_alloc = dtypemeta_alloc,
-    .tp_new = dtypemeta_new,
-    .tp_is_gc = dtypemeta_is_gc,
+    {Py_tp_traverse, (void *)dtypemeta_traverse},
```
da-woods, line 1397: You'll need to modify this slightly:

> Heap types ([Py_TPFLAGS_HEAPTYPE](https://docs.python.org/3/c-api/typeobj.html#c.Py_TPFLAGS_HEAPTYPE)) must visit their type with:
>
> `Py_VISIT(Py_TYPE(self));`

https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_traverse

(You don't need to do an equivalent modification of `tp_clear` though)

numpy/_core/src/multiarray/dtypemeta.c
```diff
@@ -1368,24 +1391,42 @@ initialize_legacy_dtypemeta_aliases(_PyArray_LegacyDescr **_builtin_descrs) {
     _Void_dtype = NPY_DTYPE(_builtin_descrs[NPY_VOID]);
 }
 
-NPY_NO_EXPORT PyTypeObject PyArrayDTypeMeta_Type = {
-    PyVarObject_HEAD_INIT(NULL, 0)
-    .tp_name = "numpy._DTypeMeta",
-    .tp_basicsize = sizeof(PyArray_DTypeMeta),
-    .tp_dealloc = (destructor)dtypemeta_dealloc,
+static PyType_Slot dtypemeta_slots[] = {
+    {Py_tp_dealloc, (void *)dtypemeta_dealloc},
```
da-woods, line 1395: You need to modify this slightly

> If the type is heap allocated ([Py_TPFLAGS_HEAPTYPE](https://docs.python.org/3/c-api/typeobj.html#c.Py_TPFLAGS_HEAPTYPE)), the deallocator should release the owned reference to its type object (via [Py_DECREF()](https://docs.python.org/3/c-api/refcounting.html#c.Py_DECREF)) after calling the type deallocator.

https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_dealloc

numpy/_core/src/multiarray/dtypemeta.c
```diff
@@ -1310,6 +1311,28 @@ dtypemeta_get_is_numeric(PyArray_DTypeMeta *self, void *NPY_UNUSED(ignored)) {
...
 }
 
+static PyObject *
+dtypemeta_getattro(PyObject *self, PyObject *name)
+{
+    /*
+     * Legacy (non-heap) DType types store their module in tp_name as
+     * "module.classname" (e.g. "numpy.dtypes.TimeDelta64DType"), but their
+     * MRO does not include PyType_Type so type.__module__'s data descriptor
+     * is never found during attribute lookup, causing __module__ to fall back
```
da-woods, line 1321: I don't think the comment here is accurate.

CPython's `type_module` function just has explicitly different behaviour for heaptypes and non-heaptypes (https://github.com/python/cpython/blob/24db78c5329dd405460bfdf76df380ced6231353/Objects/typeobject.c#L1606)

Either way, I agree with the recommendation to make it a specific  getter.

PR #31077, opened 2026-03-26
BUG: np.load(..., mmap_mode=...) raises ValueError for .npz

### LINKED ISSUE #28795: BUG: `load` of a npz file silently ignores argument `mmap_mode`
### Describe the issue:

Already mentioned in numpy/numpy#5976.

If I create a `.npz` file with `np.savez`, and then read it with `np.load(..., mmap_mode=...)` with some non-default `mmap_mode`, the file is not actually memmapped.

I think `np.load` should raise an exception if `mmap_mode` is set and the file is a `.npz` archive.

There's a package on PyPI to memmap npz archives (https://github.com/f0k/mmnpz) so there is no need to implement the functionality in numpy, raising an error is sufficient.

```python
import numpy as np

x = np.arange(10)
np.save('file.npy', x)
np.savez('file.npz', x=x)

x_ = np.load('file.npy', mmap_mode='r')
assert x_.__class__.__name__ == 'memmap'

arch = np.load('file.npz', mmap_mode='r')
assert arch['x'].__class__.__name__ == 'memmap', 'not an memmap!'
```

### Error message:

```shell
AssertionError                            Traceback (most recent call last)
Cell In[60], line 11
      8 assert x_.__class__.__name__ == 'memmap'
     10 arch = np.load('file.npz', mmap_mode='r')
---> 11 assert arch['x'].__class__.__name__ == 'memmap', 'not an memmap!'

AssertionError: not an memmap!
```

2.2.1
3.13.1 | packaged by conda-forge | (main, Jan 13 2025, 09:45:31) [Clang 18.1.8 ]

### Runtime Environment:

[{'numpy_version': '2.2.1',
  'python': '3.13.1 | packaged by conda-forge | (main, Jan 13 2025, 09:45:31) '
            '[Clang 18.1.8 ]',
  'uname': uname_result(system='Darwin', node='MacBookPro.lan', release='24.3.0', version='Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:16 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6000', machine='arm64')},
                      'found': ['ASIMDHP'],
                      'not_found': ['ASIMDFHM']}}]

### Context for the issue:

It's unexpected behavior and wasted some of my time.

ricas28:
### PR summary

Previously, `mmap_mode` could be passed to `np.load` on `.npz` (compressed) files, which is not supported and could silently fail or behave incorrectly.

This commit:

- Keeps the roundtrip tests for `.npz` intact.
- Adds a test to verify that `np.load(..., mmap_mode='r')` raises `ValueError` for `.npz` files.
- Skips `test_mmap` in `TestSavezLoad` subclass since mmap is not supported for `.npz` files.

Fixes numpy#28795  
See also numpy#5976

Hi! I am a third-year Computer Engineering student. This is my first contribution to NumPy.  
I have been using NumPy for university projects and personal practice.  
I wanted to contribute by fixing the behavior of `np.load` with `mmap_mode` on `.npz` files, as I noticed it was a simple task to start contributing to NumPy.

#### AI Disclosure

I used AI (ChatGPT) only to navigate and understand which files were relevant to change.  
All code and test changes were written by me.

### DIFF
--- a/numpy/lib/tests/test_io.py
+++ b/numpy/lib/tests/test_io.py
@@ -228,6 +228,17 @@ class TestSavezLoad(RoundtripTest):
                 assert len(npz["test2"]) == 10
                 assert npz["metadata"] == b"Name: Test"
 
+    def test_mmap(self):
+        """"
+        Verify that np.load raises a ValueError when mmap_mode 
+        is used with .npz files, as mmap is not supported for .npz.
+        """
+        with temppath(prefix="numpy_test_npz_mmap_", suffix=".npz") as tmp:
+            np.savez(tmp, x=np.arange(10))
+
+            with pytest.raises(ValueError):
+                np.load(tmp, mmap_mode="r")
+
     @pytest.mark.skipif(not IS_64BIT, reason="Needs 64bit platform")
     @pytest.mark.slow
     @pytest.mark.thread_unsafe(reason="crashes with low memory")
--- a/doc/release/upcoming_changes/31077.changes.rst
+++ b/doc/release/upcoming_changes/31077.changes.rst
@@ -1,0 +1,6 @@
+``np.load`` raises ``ValueError`` for ``mmap_mode`` with ``.npz`` files
+---------------------------------------------------------------------
+
+``np.load`` previously allowed using ``mmap_mode`` when loading ``.npz`` files.
+Since memory mapping is not supported for this format, it now raises a
+``ValueError`` when ``mmap_mode`` is specified.
--- a/numpy/lib/_npyio_impl.py
+++ b/numpy/lib/_npyio_impl.py
@@ -466,6 +466,9 @@ def load(file, mmap_mode=None, allow_pickle=False, fix_imports=True,
         fid.seek(-min(N, len(magic)), 1)  # back-up
         if magic.startswith((_ZIP_PREFIX, _ZIP_SUFFIX)):
             # zip-file (assume .npz)
+            if mmap_mode is not None:
+                raise ValueError("mmap_mode is not supported when loading .npz files.")
+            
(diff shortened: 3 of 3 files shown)


jorenham:
Thanks for contributing :)

Raising an error where it previously didn't is a behavior change, and should be accompanied with a release note. So could you add one?

numpy/lib/tests/test_io.py
```diff
@@ -212,6 +212,10 @@ def roundtrip(self, *args, **kwargs):
                 arr_reloaded.fid.close()
                 os.remove(arr_reloaded.fid.name)
 
+    @pytest.mark.skip(reason="mmap not supported for .npz files")
+    def test_mmap(self):
+        pass
```
tylerjereddy: Whoa do we really need this?

Maybe my brain is missing the point or something. At first glance I see an always-skipped no-op test.

tylerjereddy (changes requested):

tylerjereddy:
> should be accompanied with a release note.

Is NumPy doing release notes for behavior changes that are bug fixes now?

jorenham:
> > should be accompanied with a release note.
> 
> Is NumPy doing release notes for behavior changes that are bug fixes now?

When it can cause a `ValueError` to be raised in cases where it previously didn't, then I think it's a good idea to mention that in the release notes.

numpy/lib/tests/test_io.py
ricas28: Good point. Instead of skipping the inherited test, I’ve overridden test_mmap in TestSavezLoad to assert that a ValueError is raised when using mmap_mode with .npz files. This way the test now validates the intended behavior rather than being skipped.

PR #31080, opened 2026-03-27
BUG - Fix #30342: correct min_scalar_type boundary promotion

Manufra05:
### PR summary
The min_scalar_type function used approximate floating point literals for float64, float32 and float16 maximums. This caused values very close to the limits to be incorrectly promoted to higher precision types.

This patch replaces literals with exact C constants (FLT_MAX and DBL_MAX), and replaces the half float max with NPY_MAX_HALF, correcting the boundary logic to ensure proper type detection. I also added tests to make sure the previous problem didn't sustain, except for the cases where Python's rounding prevails.

My name is Manuel and, as most students, I have used numpy for a different set of reasons.
This contribution is part of a discipline at my college, being mandatory to fix a bug and add a feature to a open-source project, and as I have used numpy before, I thought it was the perfect choice.

#### AI Disclosure
No AI tools used

### DIFF
--- a/numpy/_core/tests/test_multiarray.py
+++ b/numpy/_core/tests/test_multiarray.py
@@ -8327,6 +8327,19 @@ class TestMinScalarType:
         wanted = np.dtype('O')
         assert_equal(wanted, dt)
 
+    def test_float_max_boundaries(self):
+        f16_max = np.finfo(np.float16).max
+        f32_max = np.finfo(np.float32).max
+        f64_max = np.finfo(np.float64).max
+        assert_equal(np.min_scalar_type(f16_max), np.dtype('float16'))
+        assert_equal(np.min_scalar_type(-f16_max), np.dtype('float16'))
+        assert_equal(np.min_scalar_type(65504.2), np.dtype('float32')) 
+        assert_equal(np.min_scalar_type(f32_max), np.dtype('float32'))
+        assert_equal(np.min_scalar_type(-f32_max), np.dtype('float32'))
+        assert_equal(np.min_scalar_type(3.5e38), np.dtype('float64'))
+        assert_equal(np.min_scalar_type(f64_max), np.dtype('float64'))
+        assert_equal(np.min_scalar_type(-f64_max), np.dtype('float64'))
+
 
 from numpy._core._internal import _dtype_from_pep3118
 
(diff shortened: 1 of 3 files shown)


numpy/_core/tests/test_multiarray.py
```diff
@@ -8360,6 +8327,19 @@ def test_object(self):
...
+        f32_max = np.finfo(np.float32).max
+        f64_max = np.finfo(np.float64).max
+        assert_equal(np.min_scalar_type(f16_max), np.dtype('float16'))
+        assert_equal(np.min_scalar_type(-f16_max), np.dtype('float16'))
+        assert_equal(np.min_scalar_type(65504.2), np.dtype('float32')) 
+        assert_equal(np.min_scalar_type(f32_max), np.dtype('float32'))
+        assert_equal(np.min_scalar_type(-f32_max), np.dtype('float32'))
+        assert_equal(np.min_scalar_type(3.5e38), np.dtype('float64'))
+        assert_equal(np.min_scalar_type(f64_max), np.dtype('float64'))
+        assert_equal(np.min_scalar_type(-f64_max), np.dtype('float64'))
```
tylerjereddy, line 8342: I'm not actually seeing a test failure here when I revert the source patches locally on ARM Mac?

And when I build NumPy `main` from source locally, the reproducer doesn't seem to hold anymore:

```python
>>> import numpy as np
>>> float_32_max = np.finfo("float32").max
>>> np.min_scalar_type(float_32_max)
dtype('float32')
```

Maybe things got a bit better by accident somewhere else? The diff itself looks sensible enough to me, though this function wasn't a priority it seems according to https://github.com/numpy/numpy/issues/30342#issuecomment-3601572627.

I guess because of floating point spacing and stuff, this also works on `main`:

```python
>>> np.min_scalar_type(float_32_max + 1000.)
dtype('float32')
>>> (float_32_max + 1000.) > float_32_max
np.False_
```

The diff proposed in this PR does seem to make sense, but might be nice if the testsuite reflected a tangible improvement over current behavior?

Manufra05: Thank you for the feedback. I'll change the tests in order for them  to have the exposed case(s) on #30342 explicitly and commit.

WarrenWeckesser: > And when I build NumPy main from source locally, the reproducer doesn't seem to hold anymore:

In your example, `float_32_max` is a `float32`.  To reproduce the problem, you need to pass in a `float64` instance whose value is close to the `float32` max:

```
In [2]: np.__version__
Out[2]: '2.5.0.dev0+git20260402.245e20b'

In [3]: float_32_max = np.finfo("float32").max

In [4]: np.min_scalar_type(np.float64(float_32_max))  # should be dtype('float32')
Out[4]: dtype('float64')
```

tylerjereddy:
> In your example, float_32_max is a float32

Fair enough--I think the original comment about the original regression test not failing on source reversion was fair though.

PR #31085, opened 2026-03-28
BUG: raise OverflowError on datetime64 unit conversion overflow

### LINKED ISSUE #16352: ENH: overflow-safe astype for datetime64/timedelta64 unit conversion
```
import numpy as np
from pandas._libs.tslibs.conversion import ensure_datetime64ns
arr = np.array(["2367-12-31 12:00:00"], dtype="datetime64[h]")

>>> arr.astype("datetime64[ns]")
array(['1783-06-11T12:25:26.290448384'], dtype='datetime64[ns]')

>>> ensure_datetime64ns(arr)
[...]
pandas._libs.tslibs.np_datetime.OutOfBoundsDatetime: Out of bounds nanosecond timestamp: 2367-12-31 12:00:00
```

Is there a compelling reason numpy doesnt do this?

### LINKED ISSUE #22346: BUG: dt64[s].astype("M8[m]") overflow
### Describe the issue:

I'm used to getting silent overflows when converting to higher-resolution units.  This one is new to me.

```python
import numpy as np
dt = np.datetime64(np.iinfo(np.int64).min + 1, "s")
expected = np.datetime64('-292277022657-01-27T08:29')
dt.astype("M8[m]")

>>> dt.astype("M8[m]")
numpy.datetime64('292277026596-12-04T15:29')
```

### Error message:

_No response_

numpy 1.23.1
py3.9.13 Mac

### Context for the issue:

_No response_

jbrockmendel:
Claude wrote this, I reviewed it manually.  But as with #31023, my C-fu is weak.  The prompt was pretty directly to adapt the pandas implementation.

closes #16352
closes #22346

### DIFF
--- a/numpy/_core/src/multiarray/_datetime.h
+++ b/numpy/_core/src/multiarray/_datetime.h
@@ -352,7 +352,8 @@ timedelta_hash(PyArray_DatetimeMetaData *meta, npy_timedelta td);
  */
 static inline int
 _datetime_scale_with_overflow_check(
-        npy_int64 *dt, npy_int64 num, npy_int64 denom)
+        npy_int64 *dt, npy_int64 num, npy_int64 denom,
+        const char *type_name)
 {
     if (*dt == NPY_DATETIME_NAT) {
         return 0;
@@ -361,9 +362,9 @@ _datetime_scale_with_overflow_check(
     npy_int64 neg_limit = (NPY_MAX_INT64 - denom + 1) / num;
 
     if (*dt > pos_limit || *dt < -neg_limit) {
-        PyErr_SetString(PyExc_OverflowError,
+        PyErr_Format(PyExc_OverflowError,
                 "Overflow when converting between "
-                "datetime64 units");
+                "%s units", type_name);
         return -1;
     }
     if (*dt < 0) {
--- a/numpy/_core/src/multiarray/_datetime.h
+++ b/numpy/_core/src/multiarray/_datetime.h
@@ -335,4 +335,44 @@ datetime_hash(PyArray_DatetimeMetaData *meta, npy_datetime dt);
 NPY_NO_EXPORT npy_hash_t
 timedelta_hash(PyArray_DatetimeMetaData *meta, npy_timedelta td);
 
+/*
+ * Scale a datetime or timedelta value by num/denom, checking for overflow.
+ *
+ * Positive values compute *dt * num / denom.
+ * Negative values compute (*dt * num - (denom - 1)) / denom to round
+ * toward negative infinity.
+ *
+ * NPY_DATETIME_NAT is NPY_MIN_INT64 (i.e. -NPY_MAX_INT64 - 1).
+ * The asymmetric neg_limit formula ensures that a valid *dt * num never
+ * produces NPY_MIN_INT64, which would be misinterpreted as NaT.
+ *
+ * NaT values pass through unchanged.
+ *
+ * Returns 0 on success, -1 on overflow (with PyExc_OverflowError set).
+ */
+static inline int
+_datetime_scale_with_overflow_check(
+        npy_int64 *dt, npy_int64 num, npy_int64 denom)
+{
+    if (*dt == NPY_DATETIME_NAT) {
+        return 0;
+    }
+    npy_int64 pos_limit = NPY_MAX_INT64 / num;
+    npy_int64 neg_limit = (NPY_MAX_INT64 - denom + 1) / num;
+
+    if (*dt > pos_limit || *dt < -neg_limit) {
+        PyErr_SetString(PyExc_OverflowError,
+                "Overflow when converting between "
+                "datetime64 units");
+        return -1;
+    }
+    if (*dt < 0) {
+        *dt = (*dt * num - (denom - 1)) / denom;
+    }
+    else {
+        *dt = *dt * num / denom;
+    }
+    return 0;
+}
+
 #endif  /* NUMPY_CORE_SRC_MULTIARRAY__DATETIME_H_ */
--- a/numpy/_core/src/multiarray/datetime.c
+++ b/numpy/_core/src/multiarray/datetime.c
@@ -444,7 +444,16 @@ NpyDatetime_ConvertDatetime64ToDatetimeStruct(
         return -1;
     }
 
-    /* TODO: Change to a mechanism that avoids the potential overflow */
+    /* Check for overflow before multiplying by meta->num */
+    if (meta->num > 1) {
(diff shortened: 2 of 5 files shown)


ngoldbaum:
Let me know if you'd like a hand fixing the tests.

numpy/_core/src/multiarray/dtype_transfer.c
```diff
@@ -854,10 +869,25 @@ _aligned_strided_to_strided_datetime_cast(
     npy_int64 num = d->num, denom = d->denom;
     npy_int64 dt;
 
+    /*
+     * Precompute overflow boundaries for the dt * num multiplication.
+     * Positive values compute dt * num / denom (no subtraction risk).
+     * Negative values compute (dt * num - (denom - 1)) / denom, so
+     * the limit must also account for the (denom - 1) subtraction.
+     */
```
ngoldbaum: Maybe add to this comment that `NPY_DATETIME_NAT` is NPY_MIN_INT64, which is `-NPY_MAX_INT64 - 1`, and the formula below avoids generating a value that overflows to `NPY_DATETIME_NAT`.

Maybe also add a test that the exact boundary value triggers an OverflowError. e.g. right now on current `main`:

```
>>> np.datetime64(9223372036, "s").astype('datetime64[ns]')
np.datetime64('2262-04-11T23:47:16.000000000')
>>> np.datetime64(9223372037, "s").astype('datetime64[ns]')
np.datetime64('1677-09-21T00:12:43.290448384')
```

numpy/_core/tests/test_datetime.py
```diff
@@ -957,6 +957,53 @@ def cast2():
...
+        arr_ok = np.array(["2000-01-01", "2020-06-15"], dtype="datetime64[s]")
+        result_arr = arr_ok.astype("datetime64[ns]")
+        expected = np.array(["2000-01-01", "2020-06-15"], dtype="datetime64[ns]")
+        assert_equal(result_arr, expected)
+
+        # NaT should pass through without raising
+        arr_nat = np.array(["NaT", "2020-01-01"], dtype="datetime64[s]")
+        result_nat = arr_nat.astype("datetime64[ns]")
+        assert np.isnat(result_nat[0])
+        assert result_nat[1] == np.datetime64("2020-01-01", "ns")
```
ngoldbaum, line 1030: I'm pretty sure that none of these tests exercise the code path where `meta->num != 1`. Maybe add one that converts to a unit of "2s" or something like that?

numpy/_core/src/multiarray/datetime.c
```diff
@@ -444,7 +444,16 @@ NpyDatetime_ConvertDatetime64ToDatetimeStruct(
...
-    /* TODO: Change to a mechanism that avoids the potential overflow */
+    /* Check for overflow before multiplying by meta->num */
+    if (meta->num > 1) {
+        npy_int64 overflow_limit = NPY_MAX_INT64 / meta->num;
+        if (dt > overflow_limit || dt < -overflow_limit) {
+            PyErr_SetString(PyExc_OverflowError,
+                    "Overflow when converting between "
+                    "datetime64 units");
+            return -1;
+        }
```
ngoldbaum, line 452: Not a big deal in the grand scheme of things, but one issue with AI-generated contributions is there's a forest-for-the-trees problem. We have `npy_mul_with_overflow` variants in `templ_common.h.src`. They're only defined for non-bitwidth C types but you could add an int64 variant and use that instead. Those functions use clang/gcc intrinsics if they're available and avoids division if not.

Maybe we should have an AGENTS.md file with suggestions like that? It'd probably have to be a mailing list discussion though because people have strong feelings.

The other code path in `dtype_transfer.c` is different from generic signed integer overflow detection because datetimes have a special NaT value, so you can't use the generic signed integer overflow utility functions NumPy has. But here you can, I think.

ngoldbaum (commented):
I asked claude Opus to review this PR and it had a couple minor suggestions that I passed on. I think the overflow checking code is correct.

I guess Pandas doesn't really care about it, but I'm pretty sure timedeltas have the exact same bug, see datetime.c lines 3179 to 3185. Maybe Claude could fix that too? Also totally OK to punt on that if you don't want to do it.

jbrockmendel:
I had my claude implement your claude's suggestions.  

> I guess Pandas doesn't really care about it, but I'm pretty sure timedeltas have the exact same bug

for timedeltas we wrap the dt64 code and catch any exception to re-raise with a td64-specific message

ngoldbaum (approved):
LGTM! This is mergeable as is. The test failure is unrelated (#31129).

I have one comment to reduce the code and comment duplication a little. Define the following helper in a header shared by `datetime.c` and `dtype_transfer.c`:

```C
static inline int
_datetime_scale_with_overflow_check(
        npy_int64 *dt, npy_int64 num, npy_int64 denom)
{
    if (*dt == NPY_DATETIME_NAT) {
        return 0;
    }
    /*
     * long comment that's repeated twice in the PR
     */
    npy_int64 pos_limit = NPY_MAX_INT64 / num;
    npy_int64 neg_limit = (NPY_MAX_INT64 - denom + 1) / num;

    if (*dt > pos_limit || *dt < -neg_limit) {
        PyErr_SetString(PyExc_OverflowError,
                "Overflow when converting between "
                "datetime64 units");
        return -1;
    }
    if (*dt < 0) {
        *dt = (*dt * num - (denom - 1)) / denom;
    }
    else {
        *dt = *dt * num / denom;
    }
    return 0;
}
```

and then use it to reduce the boilerplate

(totally untested, please double-check)

ngoldbaum:
I guess it also needs a parameter for whether it's a datetime or a timedelta conversion...

numpy/_core/src/multiarray/_datetime.h
```diff
@@ -335,4 +335,44 @@ datetime_hash(PyArray_DatetimeMetaData *meta, npy_datetime dt);
...
+    if (*dt == NPY_DATETIME_NAT) {
+        return 0;
+    }
+    npy_int64 pos_limit = NPY_MAX_INT64 / num;
+    npy_int64 neg_limit = (NPY_MAX_INT64 - denom + 1) / num;
+
+    if (*dt > pos_limit || *dt < -neg_limit) {
+        PyErr_SetString(PyExc_OverflowError,
+                "Overflow when converting between "
+                "datetime64 units");
```
ngoldbaum: It could be a timedelta too. You need a parameter to format this error message correctly.

ngoldbaum (approved):
Thanks for working with me on this! Looks like a clear improvement now and I don't see any remaining issues.

PR #31091, opened 2026-03-29
ENH: Implement opaque PyObject ABI support 

kumaraditya303:
### PR summary

This PR adds support for building opaque PyObject stable ABI extensions using NumPy. 

#### AI Disclosure

Claude was used to generate the tests.


ngoldbaum:
Sorry for adding and removing the milestone repeatedly. I know we talked about this yesterday at the meeting, but it occurs to me that maybe it's reasonable for NumPy 2.5 to support Python 3.15, e.g. by uploading wheels to PyPI? If so, it would be logical to include this as well.

NumPy 2.5 is scheduled to be released this summer, after the ABI for Python 3.15 is frozen, so it should be perfectly fine to build and upload NumPy wheels using a CPython beta.

I know that has implications for the release workflow, so ping @rgommers.

charris:
We don't upload wheels to PyPI for new Python releases until the rc release and cibuildwheel support for it.

EDIT: it is fine to build the wheels, just not to upload them. For instance, we build wheels for pyodide, but we don't upload them.

numpy/_core/include/numpy/ndarraytypes.h
```diff
@@ -1771,166 +1805,202 @@ PyArray_CLEARFLAGS(PyArrayObject *arr, int flags)
...
-                           * flags are CONTIGUOUS (1), F_CONTIGUOUS (2),
-                           * ALIGNED (0x100), NOTSWAPPED (0x200), and
-                           * WRITEABLE (0x400).  ARR_HAS_DESCR (0x800)
-                           * states that arrdescr field is present in
-                           * structure
-                           */
+#define PyArray_ITER_NEXT(it) do { \
+        _PyAIT(it)->index++; \
+        if (_PyAIT(it)->nd_m1 == 0) { \
+                _PyArray_ITER_NEXT1(_PyAIT(it)); \
```
ngoldbaum: subtle bug: `_PyArray_ITER_NEXT1` *also* does `_PyAIT(it)`, which in the opaque PyObject build means that `_PyArrayIter_GET_ITEM_DATA()` gets called on the result of another call to `_PyArrayIter_GET_ITEM_DATA()`. That means there the `sizeof(PyObject)` offset gets applied twice.

You could make some of these private macros static inline functions so you get type errors for bugs like this. Of course that's a bigger change...

ngoldbaum: No idea how one might test this code path, but I think that indicates this code path isn't tested in your abi3t test. Maybe Claude can help generate more test cases for the full API?

numpy/_core/include/numpy/ndarraytypes.h
```diff
@@ -1771,166 +1805,202 @@ PyArray_CLEARFLAGS(PyArrayObject *arr, int flags)
...
-typedef struct npy_unpacked_static_string {
-    size_t size;
-    const char *buf;
-} npy_static_string;
+#define PyArray_MultiIter_GOTO(multi, dest) do {                            \
+        int __npy_mi;                                                       \
+        for (__npy_mi=0; __npy_mi < _PyMIT(multi)->numiter; __npy_mi++) {   \
+                PyArray_ITER_GOTO(_PyMIT(multi)->iters[__npy_mi], dest);    \
+        }                                                                   \
+        _PyMIT(multi)->index = _PyMIT(multi)->iters[0]->index;              \
```
ngoldbaum: Here and elsewhere where you access the `iters` member of `PyArrayMultiIterObject` in expressions like `_PyMIT(multi)->iters[0]`, I think it needs to be `_PyAIT(_PyMIT(multi)->iters[0])` instead, otherwise anyone using this macro will see a compilation error.

ngoldbaum: (also indicates it's worth adding a test for this API)

numpy/_core/code_generators/numpy_api.py
```diff
@@ -409,6 +409,13 @@ def get_annotations():
     # End 2.0 API
     # NpyIterGetTransferFlags (slot 223) added.
     # End 2.3 API
+    '_PyArray_GET_ITEM_DATA':                  (369, MinVersion("2.5")),
+    '_PyArrayIter_GET_ITEM_DATA':              (370, MinVersion("2.5")),
+    '_PyArray_LegacyDescr_GET_ITEM_DATA':      (371, MinVersion("2.5")),
+    '_PyDataType_GET_ITEM_DATA':               (372, MinVersion("2.5")),
+    '_PyArrayMultiIter_GET_ITEM_DATA':         (373, MinVersion("2.5")),
+    '_PyArrayNeighborhoodIter_GET_ITEM_DATA':  (374, MinVersion("2.5")),
+    '_PyDatetimeScalarObject_GetUnit':         (375,),
```
ngoldbaum:
```suggestion
    '_PyDatetimeScalarObject_GetUnit':         (375, MinVersion("2.5")),
```

Unless you did that intentionally for some reason?

numpy/__init__.cython-30.pxd
```diff
@@ -1119,28 +1116,32 @@ cdef inline bint is_datetime64_object(object obj) noexcept:
     return PyObject_TypeCheck(obj, &PyDatetimeArrType_Type)
 
 
-cdef inline npy_datetime get_datetime64_value(object obj) noexcept nogil:
+cdef inline npy_datetime get_datetime64_value(object obj) noexcept:
```
ngoldbaum: Removing `nogil` is an API change. One way to avoid that is to move the C API call inside a `with gil` block (e.g. re-acquring the GIL). See https://cython.readthedocs.io/en/latest/src/userguide/nogil.html#releasing-and-reacquiring-the-gil.

This also applies to the other functions below where you removed `nogil` annotations.

ngoldbaum (commented):
I asked Claude to review this PR. I unfortunately don't have the mental patience to actually read this whole diff in one sitting.

It spotted a couple subtle issues in the iterator macros on the opaque pyobject build. Up to you, but I think it might be worth adding tests for the C iterator APIs under the opaque pyobject builds given the complexity of these macros.

But maybe also check for open-source uses first by searching github for items in the C API to see if it's worth doing.

rgommers:
> NumPy 2.5 is scheduled to be released this summer, after the ABI for Python 3.15 is frozen, so it should be perfectly fine to build and upload NumPy wheels using a CPython beta.

What Chuck said - only after 3.15.0rc1. There's nothing in this PR that requires this though, and there's no problem merging this PR and having support in the headers in 2.5.0.

numpy/_core/code_generators/numpy_api.py
seberg: I might slightly prefer if we make it a `GetMetadata` or so since the meta also includes a factor and we really should need both.
Not sure why this isn't exposed, for some reason pandas must not be needing (or maybe it is a latent bug there because nobody uses it).

Or actually, maybe a direct getter for the time_fields, so we can also fetch the `value` (i.e. the `nogil` comment below)?

(Not that I wouldn't prefer if downstream just extract values via API calls, but if the `nogil` is there for a reason, then that seems better than adding a `with gil:` unless it is a giant mess)

numpy/_core/code_generators/numpy_api.py
```diff
@@ -409,6 +409,13 @@ def get_annotations():
     # End 2.0 API
     # NpyIterGetTransferFlags (slot 223) added.
     # End 2.3 API
+    '_PyArray_GET_ITEM_DATA':                  (369, MinVersion("2.5")),
+    '_PyArrayIter_GET_ITEM_DATA':              (370, MinVersion("2.5")),
+    '_PyArray_LegacyDescr_GET_ITEM_DATA':      (371, MinVersion("2.5")),
+    '_PyDataType_GET_ITEM_DATA':               (372, MinVersion("2.5")),
+    '_PyArrayMultiIter_GET_ITEM_DATA':         (373, MinVersion("2.5")),
+    '_PyArrayNeighborhoodIter_GET_ITEM_DATA':  (374, MinVersion("2.5")),
```
seberg, line 417: These functions are all identical, since they are all base-classes.  I don't really see a need for trying to future-proof with per-object versions?

(To me it seems like it may be neater if these had a different name to avoid the `#define` them rather than need `#undef`. If we do the above that would happen automatically.)

numpy/_core/include/numpy/ndarraytypes.h
seberg: Getting rid of the casts in general would be nice (but it does seem like something to do in a follow-up and there was that MSVC inlining discussion, IIRC)...

On the up-side, I think _if_ we are on an opaque build and this macro is used then this is already a function and should already give type errors.
So hopefully, we only have a risk of API gaps that are easy to fix and at least not as bad bugs.

numpy/_core/include/numpy/ndarraytypes.h
```diff
@@ -1771,166 +1805,202 @@ PyArray_CLEARFLAGS(PyArrayObject *arr, int flags)
...
-} PyArrayInterface;
+#define PyArray_ITER_NOTDONE(it) (_PyAIT(it)->index < _PyAIT(it)->size)
 
 
-/****************************************
- * NpyString
- *
- * Types used by the NpyString API.
- ****************************************/
+#define _PyMIT(m) (_PyArrayMultiIter_GET_ITEM_DATA((PyArrayMultiIterObject *)m))
```
seberg, line 1922: Do we need to include the cast here?  I.e. yeah, it's an API change to not cast for the user, but it is better anyway.
(Related to the safety discussion above, if the new API is a bit safer w.r.t. to this that seems fine, it would be nice generally, but not really here.)

seberg (commented):
Didn't scan too closely for bugs, but I think Nathan got them (and honestly, I am not all that worried because all of these look like they should be compile time errors downstream and then even a minor release can fix them when they come up -- doesn't mean it isn't better to do it now :)).

Overall, the diff seems much more churn than it actually is, IMO, so this should be fast to wrap up.

numpy/_core/code_generators/numpy_api.py
kumaraditya303: > These functions are all identical, since they are all base-classes. I don't really see a need for trying to future-proof with per-object versions?

These are identical currently but we need it incase the object header size changes for one of them if for example the layout switches to PyVarObject.

numpy/_core/include/numpy/ndarraytypes.h
kumaraditya303: Fixed, I added some tests for this with Claude

numpy/_core/code_generators/numpy_api.py
kumaraditya303: This does not has min version set because this is used by numpy internal cython modules and those have a target version less than 2.5 set, it was easier to just not add min version for this.

numpy/__init__.cython-30.pxd
kumaraditya303: fixed

numpy/_core/include/numpy/ndarraytypes.h
kumaraditya303: Yes otherwise users can see new cast warnings

numpy/_core/code_generators/numpy_api.py
seberg: Hmmm, would be an ABI break on the normal build, so feels like a far future.  Of course for some of these (not most), a VarObject could indeed be neat (multi-iter and the array object itself maybe mostly).
I don't care too much, I guess I still lean that it seems like duplication for little gain, but I'll let Nathan make a decision.

numpy/_core/include/numpy/ndarraytypes.h
seberg: Right, but is that so worrisome?  Once the user adds this cast, their code will be fine however they build it (and they get clearer type safety)?

kumaraditya303: Adding cast here is a bit nicer as user don't have to add it, I can change it but this is the same behaviour as `Py_INCREF` and other APIs.

numpy/_core/include/numpy/numpyconfig.h
```diff
@@ -119,6 +119,8 @@
 #elif defined(NPY_TARGET_VERSION) && NPY_TARGET_VERSION
     /* user provided a target version, use it */
     #define NPY_FEATURE_VERSION NPY_TARGET_VERSION
+#elif defined(Py_TARGET_ABI3T) && NPY_TARGET_VERSION < NPY_2_5_API_VERSION
+    #error "NumPy 2.5 or later is required when compiling with opaque PyObject"
```
seberg: The other things were all opinionated things mostly, but one thing that occurred to me:

Will CPython provide a function to apply this offset easily or can we write a small static-inline version!?

Adding the function(s) to the API likely makes sense either way, but this limitation means nobody can actually use this for 2-3 years.
If a static inline version is easy, then you can start using it immediately (for the Python versions that support it not sure what that means constraint wise).

kumaraditya303: > Will CPython provide a function to apply this offset easily or can we write a small static-inline version!?

Which offset? `sizeof(PyObject)`?

seberg: Yes, that is all you need in practice after all.  I expect you can easily figure it out, but I just didn't think about it much yet if it'll end up pretty.

kumaraditya303: > Yes, that is all you need in practice after all. I expect you can easily figure it out, but I just didn't think about it much yet if it'll end up pretty.

CPython will not expose that offset, you can do some tricks in stable ABI to find that but then you are breaking the contract of limited API assuming that offset is same of for all objects. This is the reason stable ABI requires all object layout to be PEP 697 so that user doesn't need to know `sizeof(PyObject)`.

seberg: Right, my understanding is that: `PyObject_GetTypeData(almost_anything, type(alomst_anything))` will give you the offset you need.
Storing that statically in the generated part of the header is easy (we already store the API table and the version there).
With that, adding a static-inline function is also easy.  Now, we can still add the API table function and prefer that when the `NPY_TARGET_VERSION` is high enough (so that this grows out eventually).

So I think that is *very* worthwhile, because otherwise it'll be 2+ years before anyone can use this.  That is, unless nobody can use it for 2 years because of CPython adoption anyway (but I don't think that is true).


EDIT: 

> but then you are breaking the contract of limited API assuming that offset is same of for all objects.

I think that isn't to worry about: `PyObject_HEAD` is clearly the same for all of *our* objects which are all objects that matter.

kumaraditya303: > With that, adding a static-inline function is also easy. Now, we can still add the API table function and prefer that when the NPY_TARGET_VERSION is high enough (so that this grows out eventually).

Who would be the user for the static inline functions? 

> So I think that is very worthwhile, because otherwise it'll be 2+ years before anyone can use this. That is, unless nobody can use it for 2 years because of CPython adoption anyway (but I don't think that is true).

Can you elaborate why it will be 2+ years before anyone can use this? 3.15 is due release in October IIRC so once that get's released anyone can start using this.

seberg: Because any larger package wants to support old NumPy versions a couple of years back.

ngoldbaum: In practice, people target very old numpy APIs. At minimum, they have to target the oldest numpy they support.

That said, I think people could set the target API to the NumPy 2.5 API just for opaque PyObject builds. But also that adds build complexity.

seberg: Right, my understanding is that this type of wheel/build will also be used on older Python versions.
I.e. pywavelet wants to upload a single wheel for everything which then will need to support older NumPy versions.

But maybe that is thinking is wrong (I am really not sure), and pywavelet needs to upload a new wheel for 3.15 and greater anyway, in which case it wouldn't matter.

kumaraditya303: To be clear abi3t is not backwards compatible, it cannot run on 3.14 or older so if anyone wants to use it they still need to build wheels for older versions manually.

ngoldbaum: > But maybe that is thinking is wrong (I am really not sure), and pywavelet needs to upload a new wheel for 3.15 and greater anyway, in which case it wouldn't matter.

Yeah as Kumar said, abi3t covers all python builds 3.15 and newer, older builds will still need abi3 and/or version-specific wheels. It's probably no biggie to require the NumPy 2.5 API for those builds beyond adding some build complexity.

seberg: Ack, sorry then.  Then it doesn't matter and to some degree the only thing would be if we want to just default `NPY_TARGET_VERSION` to something higher (but not sure it is problematic).

seberg: Hmmm, reading it: https://peps.python.org/pep-0803/#shims-for-compatibility-with-cpython-3-14 says that it is plausible but doesn't exist? But indeed not clear to me if the machinery is likely to come (and if it goes back only to 3.14, maybe nobody will care much about it as well).

kumaraditya303: It is unlikely to be implemented if you ask me :)

neutrinoceros: > think people could set the target API to the NumPy 2.5 API just for opaque PyObject builds.

As a downstream maintainer of a handful large-ish packages, that is exactly what I intend to do for abi3t wheels, FTR.

neutrinoceros: Some additional thoughts: I initially thought there would be a small point of friction[^1] where users requesting older versions of numpy might get runtime errors from a cp315-abi3t wheel, but this actually seems very unlikely to happen if numpy 2.5.0 is both the first version to support abi3t _and_ 3.15, as is currently the plan (AFAIU). Errors could then only happen if an older numpy was installed _from source_, which I hope is rare enough that it's no big deal.

[^1]: and even then, it seemed acceptable to me

numpy/_core/code_generators/numpy_api.py
ngoldbaum: I'm happy with the duplication if it gives us some future flexibility.

numpy/__init__.cython-30.pxd
```diff
@@ -862,14 +863,10 @@ cdef extern from "numpy/arrayscalars.h":
         pass
 
     ctypedef struct PyDatetimeScalarObject:
-        # PyObject_HEAD
-        npy_datetime obval
-        PyArray_DatetimeMetaData obmeta
+        pass
```
ngoldbaum: I tried compiling Pandas against this branch and I get a compiler error in Pandas related to the changes to these structs:

```
     cython -M --fast-fail -3 -Xfreethreading_compatible=true --include-dir /Users/goldbaum/Documents/pandas/.mesonpy-dvqz12__/pandas/_libs/tslibs '-X always_allow_keywords=true' --shared=pandas._libs._cyutility /Users/goldbaum/Documents/pandas/pandas/_libs/tslibs/np_datetime.pyx -o pandas/_libs/tslibs/np_datetime.cpython-314-darwin.so.p/pandas/_libs/tslibs/np_datetime.pyx.c

      Error compiling Cython file:
      ------------------------------------------------------------
      ...

      cdef NPY_DATETIMEUNIT get_datetime64_unit(object obj) noexcept nogil:
          """
          returns the unit part of the dtype for a numpy datetime64 object.
          """
          return <NPY_DATETIMEUNIT>(<PyDatetimeScalarObject*>obj).obmeta.base
                                    ^
      ------------------------------------------------------------
      /Users/goldbaum/Documents/pandas/pandas/_libs/tslibs/np_datetime.pyx:78:30: Cannot convert 'PyDatetimeScalarObject *' to Python object
      [12/159] Compiling Cython source /Users/goldbaum/Documents/pandas/pandas/_libs/tslibs/dtypes.pyx
      [13/159] Generating _cyutility.c with a custom command
      [14/159] Compiling Cython source /Users/goldbaum/Documents/pandas/pandas/_libs/tslibs/conversion.pyx
      FAILED: [code=1] pandas/_libs/tslibs/conversion.cpython-314-darwin.so.p/pandas/_libs/tslibs/conversion.pyx.c
      cython -M --fast-fail -3 -Xfreethreading_compatible=true --include-dir /Users/goldbaum/Documents/pandas/.mesonpy-dvqz12__/pandas/_libs/tslibs '-X always_allow_keywords=true' --shared=pandas._libs._cyutility /Users/goldbaum/Documents/pandas/pandas/_libs/tslibs/conversion.pyx -o pandas/_libs/tslibs/conversion.cpython-314-darwin.so.p/pandas/_libs/tslibs/conversion.pyx.c

      Error compiling Cython file:
      ------------------------------------------------------------
      ...
              return convert_str_to_tsobject(ts, tz, dayfirst, yearfirst)

          if checknull_with_nat_and_na(ts):
              obj.value = NPY_NAT
          elif cnp.is_datetime64_object(ts):
              num = (<PyDatetimeScalarObject*>ts).obmeta.num
                     ^
      ------------------------------------------------------------
      /Users/goldbaum/Documents/pandas/pandas/_libs/tslibs/conversion.pyx:384:15: Cannot convert 'PyDatetimeScalarObject *' to Python object
```

So I don't think we can unconditionally change these definitions without providing an API to migrate to. We might need to ask David Woods for help if it's not clear how to write the Cython bindings without relying on conditional compilation or other discouraged constructs in Cython (see https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#conditional-compilation).

Maybe we need new getter functions for `obval` as well as `obmeta`?

numpy/__init__.cython-30.pxd
```diff
@@ -1126,21 +1123,27 @@ cdef inline npy_datetime get_datetime64_value(object obj) noexcept nogil:
     Note that to interpret this as a datetime, the corresponding unit is
     also needed.  That can be found using `get_datetime64_unit`.
     """
-    return (<PyDatetimeScalarObject*>obj).obval
+    cdef npy_datetime value
+    with gil:
+        PyArray_ScalarAsCtype(obj, &value)
```
ngoldbaum: I know I suggested doing this, but ideally we wouldn't need to acquire the GIL on the normal build. `PyAray_ScalarAsCtype` does a lot more than just a field access, so this might cause a significant performance regression in Pandas for non-opaque builds. Another reason to add functions to access these fields?

numpy/_core/code_generators/generate_numpy_api.py
```diff
@@ -58,7 +58,7 @@
 #include "numpy/_public_dtype_api_table.h"
 
 #if !defined(NO_IMPORT_ARRAY) && !defined(NO_IMPORT)
-static int
+static inline int
```
ngoldbaum, line 61: This is kind-of load bearing code ecosystem-wide. Is this change actually necessary for this PR? If not, maybe propose it in a separate PR?

It might also need a release note because it changes how user code will compile `import_array`.

numpy/_core/include/numpy/arrayscalars.h
```diff
@@ -162,6 +163,34 @@ typedef struct {
...
+typedef struct PyCDoubleScalarObject PyCDoubleScalarObject;
+typedef struct PyCLongDoubleScalarObject PyCLongDoubleScalarObject;
+typedef struct PyObjectScalarObject PyObjectScalarObject;
+typedef struct PyDatetimeScalarObject PyDatetimeScalarObject;
+typedef struct PyTimedeltaScalarObject PyTimedeltaScalarObject;
+typedef struct PyScalarObject PyScalarObject;
+typedef struct PyVoidScalarObject PyVoidScalarObject;
+#endif
+
+#define PyStringScalarObject PyBytesObject
```
ngoldbaum: I think it was buggy before, but this should be moved inside an `ifndef Py_LIMITED_API` block, because `PyBytesObject` isn't exposed in the limited API.

numpy/_core/include/numpy/numpyconfig.h
ngoldbaum: I think this error is unreachable if `NPY_TARGET_VERSION` is defined and set to a truthy value, since the elif branch before this one will fire.

ngoldbaum (commented):
I asked Claude to give it another once-over and spotted a couple more issues.

I also tried building this branch against Pandas and saw compilation errors. We'll need to fix those before this can be merged.

numpy/__init__.cython-30.pxd
kumaraditya303: > cdef NPY_DATETIMEUNIT get_datetime64_unit(object obj) noexcept nogil:
          """
          returns the unit part of the dtype for a numpy datetime64 object.
          """
          return <NPY_DATETIMEUNIT>(<PyDatetimeScalarObject*>obj).obmeta.base

~This looks like from older version of this branch, are you trying with the latest commit?~

Never mind, this is from pandas internals.

numpy/__init__.cython-30.pxd
kumaraditya303: fixed by adding `_PyTimedeltaScalarObject_GetValue`

numpy/_core/code_generators/generate_numpy_api.py
kumaraditya303: If I do not do this then I get compilation error because of unused function:
```
numpy/_core/__multiarray_api.h:1560:1: error: ‘_import_array’ defined but not used [-Werror=unused-function]
 1560 | _import_array(void)
      | ^~~~~~~~~~~~~
```

numpy/__init__.cython-30.pxd
kumaraditya303: I fixed it by adding it back, so that projects like pandas which directly access the struct can continue working. This also means that anyone trying to access in limited API would see compilation errors. I don't think there is an easy way to add it for limited API, @da-woods?

numpy/_core/include/numpy/arrayscalars.h
kumaraditya303: fixed

numpy/_core/include/numpy/numpyconfig.h
kumaraditya303: fixed

numpy/_core/code_generators/generate_numpy_api.py
seberg: I saw this and was wondering a bit as well.  But I _think_ it is actually an improvement.

Even if someone uses NumPy symbols indirectly, they need to include the NumPy headers, so they would have the function available if needed.
And calling an incompatible earlier definition would not work out, I think (and is pretty implausible).

If we are still a bit squirmish (which I can understand), let's just make sure to include this in 2.5.0.

numpy/_core/code_generators/numpy_api.py
seberg: Fair, I guess I have a silly compulsion to say that I still think it seems like unnecessary churn and code/binary duplication (unless it will be an established pattern across the ecosystem).

In practice it will be years before we can take advantage of the flexibility to ease transition (everyone uses the new API).  And even then I can't think of a reason why I would want to for any of our objects.

numpy/__init__.cython-30.pxd
da-woods: There's no reason not to give the struct contents in the pxd file - it doesn't generate any code until you try to use it.

It won't actually be possible to use the fields of these structs when using the opaque limited API. I don't currently think there's much way around that.

----------------------

Possibly:

I added an argument to `cython.cast(..., objstruct_cast=True)` which gets the objstruct of an opaque type by doing `PyObject_GetTypeData`. That'd require unmerged and unreleased Cython versions. It would also be a bit dodgy because the original objects *aren't* actually opaque, but given the object layout you *might* get away with it. I think Cython might stop you from doing it though because it'll know that it isn't safe.

numpy/_core/code_generators/generate_numpy_api.py
ngoldbaum: It looks like the OpenBLAS CI on this PR are hitting errors related to this, because it compiles with -Werror. 

```
numpy/_core/__multiarray_api.h:1560:1: error: ‘_import_array’ defined but not used [-Werror=unused-function]
 1560 | _import_array(void)
      | ^~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

Any way to add this without generating new unused function warnings?

kumaraditya303: > Any way to add this without generating new unused function warnings?

I had marked the function as inline as that avoids this warning, shall I add that back?

ngoldbaum:
I added `needs-release-note` because it's a new feature so we should at least describe it in the release notes.

We have zero narrative docs about how to use the NumPy C API with the Python limited API. I'm going to open a new issue tracking that and we can describe abi3t builds there.

numpy/_core/code_generators/generate_numpy_api.py
kumaraditya303: https://github.com/numpy/numpy/pull/31258 adds it.

ngoldbaum (approved):
I’m planning to merge this next week, let me know if anyone else wants a chance to review.

numpy/_core/src/multiarray/arrayobject.c
```diff
@@ -1275,3 +1275,70 @@ NPY_NO_EXPORT PyTypeObject PyArray_Type = {
     .tp_getset = array_getsetlist,
     .tp_new = (newfunc)array_new,
 };
+#undef _PyDataType_GET_ITEM_DATA
+/*NUMPY_API*/
+NPY_NO_EXPORT PyArray_Descr_fields *
+_PyDataType_GET_ITEM_DATA(const PyArray_Descr *dtype)
```
da-woods, line 1281: There's two potential issues with this pattern: 

1. I think it's really a violation of the c and c++ strict aliasing rules and so probably undefined behaviour. It's probably not problematic undefined behaviour (since the compiler probably can't see the access through the "other" type in opaque code). But I decided against it as a solution to a similar problem in Cython for now. The problem remains unsolved though so you should probably just do it. But be aware...
2. If you somehow manage to by-pass the macro in non-opaque code this definitely ends up with the wrong answer. You could do that by taking a function pointer for example.

kumaraditya303: > I think it's really a violation of the c and c++ strict aliasing rules and so probably undefined behaviour. It's probably not problematic undefined behaviour (since the compiler probably can't see the access through the "other" type in opaque code).

The pointer is first cast to `char *` which avoid strict aliasing violation because `char *` is allowed to alias any other object pointer. 

> If you somehow manage to by-pass the macro in non-opaque code this definitely ends up with the wrong answer. You could do that by taking a function pointer for example.


I don't think that is an issue as that should never happen.

da-woods: > The pointer is first cast to char * which avoid strict aliasing violation because char * is allowed to alias any other object pointer.

I don't going via a cast to `char *` avoids it - the issue is accessing the same data through seemingly unrelated structures.  The idea of allowing a cast to `char *` to to provide a way to reading the binary representation of an object rather than to allow any cast through `A -> char* -> B`.

As I say, I don't think it will be a problem in practice but it makes me a bit nervous

kumaraditya303: I've added a comment about this in code.

seberg: @da-woods we were wondering if our reasoning yesterday in the NumPy meeting adds up here.

I.e. our current thought is that UB comes from operation ordering or assumption that a change to one type cannot affect another.
In NumPy, we use the same type for access.  In extensions, we call a function (which shouldn't re-order, I think?) and even if we do that twice after this access is through the same type.

I guess there are two scenarios I am not 100% sure about, but somewhat assume are safe:
```
fields *f1 = GetField(x);
fields *f2 = GetField(x);
// use f1 and f2, compiler should realize they might be the same, no?
```
and:
```
fields *f1 = GetField(x);
// use f1
ModifyX(x);  // API call modifying f1.
// use f1
```
where I would think the compiler cannot assume `ModifyX` must not have side-effects on `f1`?  (but I am not sure, not sure if strict aliasing UB freedom is confusing, but it isn't something we have to think about often enough...)

da-woods: >  our current thought is that UB comes from operation ordering or assumption that a change to one type cannot affect another.

I think that's the reasoning behind it.

> In NumPy, we use the same type for access. In extensions, we call a function (which shouldn't re-order, I think?) and even if we do that twice after this access is through the same type.

I think this is saying:
* in Numpy you use the non-opaque type
* in an extension, you access the same data via the opaque "fields-only" type
so although they're being accessed through different types, the compiler doesn't have visibility across both and so it's probably safe.

That was my conclusion too (at least in practice). I don't think the C standard really takes a view on dynamic libraries. But I think the opportunity for unsafe optimizations has mostly gone.

-----------------------------

For the first scenario I'm pretty confident it realizes that they might be the same.

For the second scenario I think it'd depend on how much it could see into `ModifyX`. If `ModifyX` is completely opaque to the compiler then it has to assume that anything could happen. For example, we know that `ModifyX` could call `GetField` and thus it can change `f1`.

It's more interesting if the compiler knows what's in `ModifyX`. E.g. if it can effectively inline it to

```
fields *f1 = GetField(x);
// use f1
// [inlined ModifyX]
x.some_int_value = 10;
// [end  inlined ModifyX]
// use f1
```

then it probably will assume that `f1` is unchanged.  I think the "extension module dynamic loading separation" should prevent that in practice. Because if it doesn't then you've included knowledge that means it can't be an `abi3t` module. But I don't think the C standard promises that.

colesbury: While I agree that this is technically UB, I think that:

1) It's unlikely to cause a problem in practice
2) I think it's unavoidable if we want to maintain source compatibility AND support opaque PyObjects

The standards compliant way would be to have `PyArray_Descr` contain a `PyArray_Descr_fields` struct, e.g.:

```c
typedef struct {
        PyTypeObject *typeobj;
        /* kind for this type */
        char kind;
        ...
} PyArray_Descr_fields;

typedef struct {
        PyObject_HEAD
        PyArray_Descr_fields fields;
} PyArray_Descr;
```

But this breaks source compatibility because you'd need to write something like `descr->fields.kind` instead of `descr->kind`.

I don't think [anonymous structs](https://gcc.gnu.org/onlinedocs/gcc/Unnamed-Fields.html) solves this problem either both because they are not C++ compatible and because C11 requires that the struct doesn't have a name.

ngoldbaum:
3.15b1 is coming soon and I'd like to enable downstream testing of this, so I'm going to merge this.

charris:
`test_limited_opaque`  is causing test failures.

ngoldbaum:
Sorry for the overly hasty merge. It probably needs to be updated for 3.15.0b1, feel free to merge a commit disabling the test and either me or Kumar will send in a fix soon.

kumaraditya303:
Created https://github.com/numpy/numpy/pull/31403 for fixing it

PR #31092, opened 2026-03-29
ENH, PERF: allocate memory as part of the array object for scalars

mhvk:
@seberg, @eendebakpt: this is an updated (and greatly improved) version of #29878, making use of the fact that in previous PRs I removed the explicit dependence on how data were allocated.

### Details

Allocate memory for the dimensions and strides, as well as small amounts of data, as part of the object, to avoid the overhead of multiple allocations (with the allocation of tracked memory for data especially large).

Note that the above is done only for standard arrays, though I think in principle it could be extended to subttypes with a suitable check.

Also, allocating space for data is done only for the standard allocator.

### Timings
```
%timeit np.empty(())
133->68.2 ns
a = 1.6

%timeit np.array(a)
151->83 ns

%timeit np.add(a, a)
581->423 ns

b = np.array(a)
%timeit np.add(b, b)
293->225 ns
```

No AI was used.

### DIFF
--- a/numpy/_core/include/numpy/ndarraytypes.h
+++ b/numpy/_core/include/numpy/ndarraytypes.h
@@ -173,7 +173,7 @@ enum NPY_TYPECHAR {
  * should be downstream compatible, but the actual algorithms used may be
  * different than before. The new approach should be more flexible and easier
  * to update.
- * 
+ *
  * Names with a leading underscore are private, and should only be used
  * internally by NumPy.
  *
@@ -187,7 +187,7 @@ typedef enum {
         NPY_HEAPSORT = 1,
         NPY_MERGESORT = 2,
         NPY_STABLESORT = 2,
-        // new style names 
+        // new style names
         _NPY_SORT_HEAPSORT = 1,
         NPY_SORT_DEFAULT = 0,
         NPY_SORT_STABLE = 2,
@@ -914,12 +914,20 @@ typedef int (PyArray_FinalizeFunc)(PyArrayObject *, PyObject *);
  */
 
 /*
- * If set, the array owns the data: it will be free'd when the array
- * is deleted.
+ * If set, the array owns the data: if the data is not stored on the
+ * instance, it will be free'd when the array is deleted.
  *
  * This flag may be tested for in PyArray_FLAGS(arr).
  */
 #define NPY_ARRAY_OWNDATA         0x0004
+
+/*
+ * If set, the data is stored on the array instance.
+ *
+ * This flag may be tested for in PyArray_FLAGS(arr).
+ * (MAYBE: allow it to be requested as well?)
+ */
+#define NPY_ARRAY_DATAONINSTANCE  0x0008
 
 /*
  * An array never has the next four set; they're only used as parameter
@@ -1957,7 +1965,7 @@ typedef struct {
  * #endif
  * #ifndef NUMPY_CORE_INCLUDE_NUMPY_NPY_1_7_DEPRECATED_API_H_
  * #define NUMPY_CORE_INCLUDE_NUMPY_NPY_1_7_DEPRECATED_API_H_
- * 
+ *
  * #ifndef NPY_NO_DEPRECATED_API
  * #if defined(_WIN32)
  * #define _WARN___STR2__(x) #x
--- a/numpy/_core/src/multiarray/alloc.h
+++ b/numpy/_core/src/multiarray/alloc.h
@@ -44,7 +44,12 @@ npy_free_cache_dim_obj(PyArray_Dims dims)
 static inline void
 npy_free_cache_dim_array(PyArrayObject * arr)
 {
-    npy_free_cache_dim(PyArray_DIMS(arr), PyArray_NDIM(arr));
+    /* deallocate if not part of the array instance */
+    if ((PyArray_DIMS(arr) != NULL)
+        && (PyArray_DIMS(arr) !=
+            (npy_intp *)((char*)arr + Py_TYPE(arr)->tp_basicsize))) {
+        npy_free_cache_dim(PyArray_DIMS(arr), PyArray_NDIM(arr));
+    }
 }
 
 extern PyDataMem_Handler default_handler;
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -744,10 +744,46 @@ PyArray_NewFromDescr_int(
         }
     }
     /*
+     * Check dimensions and find total array size `nbytes`.
+     */
+    int is_zero = 0;
+    for (int i = 0; i < nd; i++) {
+        if (dims[i] == 0) {
+            /*
+             * Continue calculating the max size "as if" this were 1
+             * to get the proper overflow error
+             */
+            is_zero = 1;
+            continue;
+        }
+        if (dims[i] < 0) {
+            PyErr_SetString(PyExc_ValueError,
+                            "negative dimensions are not allowed");
+            Py_DECREF(descr);
+            return NULL;
+        }
+        /*
+         * Care needs to be taken to avoid integer overflow when multiplying
+         * the dimensions together to get the total size of the array.
+         */
+        if (npy_mul_sizes_with_overflow(&nbytes, nbytes, dims[i])) {
+            PyErr_SetString(PyExc_ValueError,
+                            "array is too big; `arr.size * arr.dtype.itemsize` "
+                            "is larger than the maximum possible size.");
+            Py_DECREF(descr);
+            return NULL;
+        }
+    }
+    if (is_zero) {
+        nbytes = 0;
+    }
+    /*
      * Create the array instance.
      *
      * For standard arrays, we allocate extra memory for the dimensions and
-     * strides, to avoid the overhead of another allocation.
+     * strides, as well as the data themselves if small enough (and the
+     * default memory handler is used), to avoid the overhead of allocating
+     * tracked memory with PyDataMem_UserNew.
      *
      * We do this bypassing the standard tp_alloc, basically copying the
      * relevant code from PyType_GenericAlloc in Objects/typeobject.c.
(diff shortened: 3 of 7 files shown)


numpy/_core/src/multiarray/alloc.h
```diff
@@ -44,7 +44,12 @@ npy_free_cache_dim_obj(PyArray_Dims dims)
 static inline void
 npy_free_cache_dim_array(PyArrayObject * arr)
 {
-    npy_free_cache_dim(PyArray_DIMS(arr), PyArray_NDIM(arr));
+    /* deallocate if not part of the array instance */
+    if ((PyArray_DIMS(arr) != NULL)
+        && (PyArray_DIMS(arr) !=
```
mhvk, line 49: Possibly, there should be another flag for whether the strides and dimensions are stored on the array, but it seemed overkill.

numpy/_core/include/numpy/ndarraytypes.h
```diff
@@ -916,13 +914,21 @@ typedef int (PyArray_FinalizeFunc)(PyArrayObject *, PyObject *);
...
  */
 #define NPY_ARRAY_OWNDATA         0x0004
 
+/*
+ * If set, the data is stored on the array instance.
+ *
+ * This flag may be tested for in PyArray_FLAGS(arr).
+ * (MAYBE: allow it to be requested as well?)
+ */
+#define NPY_ARRAY_DATAONINSTANCE  0x0008
```
eendebakpt, line 930: This is not strictly needed (where the data is allocated is an implementation detail, in any case for the default memory handler). But fine in having this in the interface. If we keep this, can we add a test for it?

numpy/_core/src/multiarray/ctors.c
```diff
@@ -743,20 +743,121 @@ PyArray_NewFromDescr_int(
...
+            size = data_offset + (nbytes == 0? 1 : nbytes);
+        }
+        /* Allocate the object and extra space */
+        char *alloc = PyObject_Malloc(size);
+        if (alloc == NULL) {
+            return PyErr_NoMemory();
+        }
+        /* PyType_GenericAlloc zeroes the extra bytes, but we don't need to. */
+        PyObject_Init((PyObject *)alloc, subtype);
+        if (alloc == NULL) {
```
eendebakpt: `alloc` cannot be zero here. The decref should be moved to the check a few lines above

numpy/_core/src/multiarray/ctors.c
```diff
@@ -778,50 +879,19 @@ PyArray_NewFromDescr_int(
...
-             * the dimensions together to get the total size of the array.
-             */
-            if (npy_mul_sizes_with_overflow(&nbytes, nbytes, fa->dimensions[i])) {
-                PyErr_SetString(PyExc_ValueError,
-                        "array is too big; `arr.size * arr.dtype.itemsize` "
-                        "is larger than the maximum possible size.");
+            fa->dimensions = npy_alloc_cache_dim(2 * nd);
+            if (fa->dimensions == NULL) {
+                PyErr_NoMemory();
                 goto fail;
```
eendebakpt, line 887: This failure path can leak the memory handler.

numpy/_core/src/multiarray/ctors.c
```diff
@@ -837,6 +907,14 @@ PyArray_NewFromDescr_int(
             PyArray_UpdateFlags((PyArrayObject *)fa,
                     NPY_ARRAY_C_CONTIGUOUS|NPY_ARRAY_F_CONTIGUOUS);
         }
+        if (is_zero) {
+            nbytes = 0;
```
eendebakpt: was already set on line 778
```suggestion
            assert(nbytes == 0);
```

eendebakpt (commented):
Some comments and small things, but overall I am +1 on this

There are a couple of cases (pickling, reshapes) where we could make more use of the inline data, dims and strides, but since these are deprecated or not performance critical I suggest we do not add them here.

numpy/_core/include/numpy/ndarraytypes.h
mhvk: Indeed, this should be an implementation detail, and in principle one could also tell that the data is on the instance from the fact that `mem_handler=NULL` -- unfortunately, in the code elsewhere that is taken to mean that someone messed around with the instance and just set `data` (it is `free'd` in the deallocator).

That said, I agree this does not need to be public; I just need to have some way to know the data are on the instance internally.

I actually wondered whether one could adjust the default memory handler to somehow keep track of this, but am not sure how... (it doesn't have access to the instance, after all).

numpy/_core/src/multiarray/ctors.c
```diff
@@ -998,7 +1081,6 @@ PyArray_NewFromDescr_int(
 
  fail:
     NPY_traverse_info_xfree(&fill_zero_info);
-    Py_XDECREF(fa->mem_handler);
```
mhvk, line 1001: This was a mistake in the failure path - `mem_handler` gets decref'd in the deallocator too.

numpy/_core/src/multiarray/ctors.c
mhvk: Thanks, done!

numpy/_core/src/multiarray/ctors.c
mhvk: Indeed, now dealt with.

numpy/_core/src/multiarray/ctors.c
mhvk: Thanks, that was a silly one - solved.

mhvk:
OK, now split into 2 commits, one for storing the dimensions and strides on the object (which has only a modest benefit for performance, see commit message), and the second for also storing small amounts of data.

I'm still not quite happy with having to keep track of whether or not the data is on the instance via a flag; suggestions for alternatives appreciated! Ideally it would somehow be possible to let the default allocator do some of the work, by registering the address of the data-on-instance as memory that does not need freeing (so `renew` would just allocate new memory).

numpy/_core/src/multiarray/ctors.c
```diff
@@ -743,20 +743,121 @@ PyArray_NewFromDescr_int(
...
+        }
+        /* Allocate the object and extra space */
+        char *alloc = PyObject_Malloc(size);
+        if (alloc == NULL) {
+            Py_XDECREF(mem_handler);
+            Py_DECREF(descr);
+            return PyErr_NoMemory();
+        }
+        /* PyType_GenericAlloc zeroes the extra bytes, but we don't need to. */
+        fa = (PyArrayObject_fields *)PyObject_Init((PyObject *)alloc, subtype);
```
ngoldbaum, line 830: this can fail to due to memory exhaustion, might as well add a NULL check just in case

numpy/_core/src/multiarray/ctors.c
```diff
@@ -743,20 +743,121 @@ PyArray_NewFromDescr_int(
...
+            size = data_offset + (nbytes == 0 ? 1 : nbytes);
+        }
+        /* Allocate the object and extra space */
+        char *alloc = PyObject_Malloc(size);
+        if (alloc == NULL) {
+            Py_XDECREF(mem_handler);
+            Py_DECREF(descr);
+            return PyErr_NoMemory();
+        }
+        /* PyType_GenericAlloc zeroes the extra bytes, but we don't need to. */
```
ngoldbaum, line 829:
```suggestion
        /* PyType_GenericAlloc zeroes the extra bytes, but we don't need to
           because all fields are explicitly initialized below */
```

for future readers, in case new fields ever get added?

numpy/_core/src/multiarray/alloc.h
ngoldbaum: `NPY_ARRAY_DIMSONINSTANCE` seems clearer to me than the pointer math that happens here but I agree it's not a big deal

numpy/_core/src/multiarray/shape.c
```diff
@@ -129,10 +109,52 @@ PyArray_Resize_int(PyArrayObject *self, PyArray_Dims *newshape, int refcheck)
...
+                                           handler);
+        }
+        else if (PyArray_CHKFLAGS(self, NPY_ARRAY_DATAONINSTANCE)) {
+            /*
+             * The data was stored on instance, so has no reference.
+             * If the data still fits, just keep it, otherwise allocate.
+             * TODO: make this a plain else; see comment in array_dealloc.
+             */
+            if (newnbytes <= oldnbytes) {
+                new_data = PyArray_DATA(self);
```
ngoldbaum, line 126: If we get here and `PyArray_NDIM(self) != new_nd`, then we might get into a state where the dimensions are stored outside the array but the data are still stored on-instance. Maybe there should be a `PyArray_NDIM(self) == new_nd` check here?

numpy/_core/src/multiarray/ctors.c
```diff
@@ -743,20 +743,121 @@ PyArray_NewFromDescr_int(
...
+            return NULL;
+        }
+    }
+    if (is_zero) {
+        nbytes = 0;
+    }
+    /*
+     * Create the array instance.
+     *
+     * For standard arrays, we allocate extra memory for the dimensions and
```
ngoldbaum, line 783: I would say "for ndarrays (but not subtypes)" instead of 'standard arrays'.

numpy/_core/src/multiarray/ctors.c
```diff
@@ -743,20 +743,121 @@ PyArray_NewFromDescr_int(
...
+            return NULL;
+        }
+    }
 
-    fa = (PyArrayObject_fields *) subtype->tp_alloc(subtype, 0);
-    if (fa == NULL) {
-        Py_DECREF(descr);
-        return NULL;
+    if (subtype == &PyArray_Type) {
+        size_t size = subtype->tp_basicsize;
```
ngoldbaum, line 807: On 64 bit architectures, this is 96 bytes, which is enough space for 12 64 bit data elements. So this optimization applies for small arrays as well as scalars. Unless I'm missing something that restricts it to just scalars.

The PR description and title talk about scalars, so maybe that behavior is unintentional.

ngoldbaum (commented):
I left some minor comments below. I asked Claude Code to do a review pass and most of these are generated from the report is made. All the wording is from me and I tried to validate everything I commented about.

mhvk:
See gh-31108 for an alternative implementation that uses a "hacked" allocator to deal with possible resize, etc., but avoids having a flag. It has the advantage that it needs far less change elsewhere in the code base, thus giving hope that if people use `mem_handler` outside of numpy, things will just continue to work.

numpy/_core/src/multiarray/ctors.c
mhvk: Sorry, originally I only did scalars, but now it is for small arrays as well, indeed.

numpy/_core/src/multiarray/ctors.c
mhvk: Agreed. I've done this in the other PR, though, since I think that ended up the better approach.

numpy/_core/src/multiarray/ctors.c
mhvk: This cannot fail - the allocation has already been made above. Python also does not check:
https://github.com/python/cpython/blob/9e5b8383724211d14165a32c0e7682e56e13843a/Objects/object.c#L531-L540

PR #31095, opened 2026-03-29
TYP: deprecate unitless ``timedelta64`` construction

jorenham:
This is a follow-up of #29619, addressing the static typing side of unitless `timedelta64` construction. I'll statically deprecate the int-like overloads of the arithmetic operator in a follow-up, 

#### AI Disclosure


charris:
Thanks Joren.

PR #31096, opened 2026-03-29
ENH, PERF: allocate dims and strides on array object by default.

mhvk:
This is a subset of #31092, only allocating space for the dimensions and strides. The effect is much less big than when one also includes data, but still a nice benefit. Mostly split out since I need to understand where a memory leak comes from... But perhaps also good to just make the PRs smaller.

### Details
Allocate memory for the dimensions and strides as part of the object, to avoid the overhead of an extra allocation.  This will only become useless memory if the array shape is set (now deprecated) or the array is resized in-place (frowned upon).

Note that the above is done only for standard arrays, though I think in principle it could be extended to subttypes with a suitable check.

Also, allocating space for data is done only for the standard allocator.

### Timings
```
%timeit np.empty(())
133->127 ns

a = 1.6
%timeit np.array(a)
151->137 ns

%timeit np.add(a, a)
581->557 ns

b = np.array(a)
%timeit np.add(b, b)
293->279 ns
```

No AI was used.

### DIFF
--- a/numpy/_core/src/multiarray/alloc.h
+++ b/numpy/_core/src/multiarray/alloc.h
@@ -44,7 +44,12 @@ npy_free_cache_dim_obj(PyArray_Dims dims)
 static inline void
 npy_free_cache_dim_array(PyArrayObject * arr)
 {
-    npy_free_cache_dim(PyArray_DIMS(arr), PyArray_NDIM(arr));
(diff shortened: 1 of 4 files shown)


mhvk:
Close in favour of #31092, which now no longer has memory leaks.

PR #31097, opened 2026-03-29
TST: Add tests for PyUFunc_ReplaceLoopBySignature

eendebakpt:
### PR summary

Add C test helpers (`replace_loop`/`restore_loop`) that replace a ufunc's float64 loop with one that always outputs 42.0.

Python tests verify:

- Replaced loop is actually called (output is 42.0, not normal result)
- Replacement works after the ufunc has been called (caches populated)
- Original loop is correctly restored

#### AI Disclosure

Claude was instructed to create a test based on https://github.com/numpy/numpy/pull/31068#issuecomment-4141269180. Afterwards the test was manually cleaned and made more specific.


numpy/_core/tests/test_umath.py
```diff
@@ -5111,6 +5111,25 @@ def test_bad_legacy_gufunc_silent_errors(x1):
         ncu_tests.always_error_gufunc(x1, 0.0)
 
 
+class TestReplaceLoopBySignature:
+    """Tests for PyUFunc_ReplaceLoopBySignature C API."""
+
+    def test_replace_loop(self):
```
seberg, line 5118:
```suggestion
    @pytest.mark.thread_unsafe(reason="modifies ufunc within test")
    def test_replace_loop(self):
```

seberg (approved):
Thanks for adding this test, probably best to have if we want to touch things here (and even if we deprecated easy to remove).

A bit scary to use a real ufunc, but should be OK.  (Although if there is a testing one around, and you feel like it...)

Either way, the only real thing is that this test isn't thread safe, so needs the marker.

seberg (approved):
Thanks, if things change and the test is outdated it should be easy to follow-up!

PR #31099, opened 2026-03-29
ENH: Improve performance of PyArray_FromAny and cleanup PyArray_ExtractDTypeAndDescriptor

eendebakpt:
### PR summary

Add a fast path in `PyArray_FromAny()` when the object passed is a numpy array and no dtype, flags, depth, or context constraints are given. The method is called internally from a few places, so multiple methods benefit.

The `PyArray_ExtractDTypeAndDescriptor` cannot fail, we simplify the code calling this method.

Part of #31098

**Results: main vs branch**

| Benchmark | main | branch | Speedup | Hits fast path? |
|---|---|---|---|---|
| `np.copyto(dst, src)` | 179 ns | 128 ns | **1.39x** | YES |
| `np.where(cond, a, b)` | 604 ns | 486 ns | **1.24x** | YES (3 inputs) |
| `np.sin(a, out=out)` | 209 ns | 196 ns | **1.06x** | YES (out arg) |
| `np.sum(4x4)` | 1.49 us | 1.43 us | **1.04x** | YES |
| `np.max(array_4)` | 1.44 us | 1.39 us | **1.04x** | YES |
| `np.prod(array_4)` | 1.38 us | 1.32 us | **1.04x** | YES |
| `np.sum(array_4)` | 1.42 us | 1.38 us | **1.02x** | YES |
| `np.any(bool_4)` | 1.29 us | 1.26 us | **1.02x** | YES |
| `np.sin(array_4)` | 260 ns | 260 ns | ~1.00x | NO |
| `np.count_nonzero(4)` | 123 ns | 123 ns | ~1.00x | NO (has dtype) |
| **Geometric mean** | | | **1.08x** | |

#### AI Disclosure

Claude was used in profiling the numpy `np.sum`, identifying the possible performance improvements and creating of the initial commit. The simplification of `PyArray_ExtractDTypeAndDescriptor` was identified manually, but call sites have been updated by Claude.


numpy/_core/src/multiarray/descriptor.c
ngoldbaum, line 1448: Looking closer, I think the following change is also possible:

```
diff --git a/numpy/_core/src/multiarray/descriptor.c b/numpy/_core/src/multiarr>
index e712ff5372..32b3ef44eb 100644
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -1413,7 +1413,8 @@ descr_is_legacy_parametric_instance(PyArray_Descr *descr,
     /* Flexible descr with generic time unit (which can be adapted) */
     if (PyDataType_ISDATETIME(descr)) {
         PyArray_DatetimeMetaData *meta;
-        meta = get_datetime_metadata_from_dtype(descr);
+        _PyArray_LegacyDescr *ldescr = (_PyArray_LegacyDescr *)descr;
+        meta =  &(((PyArray_DatetimeDTypeMetaData *)ldescr->c_metadata)->meta);
         if (meta->base == NPY_FR_GENERIC) {
             return 1;
         }
```

(i.e. inining part of `get_datetime_metadata_from_dtype`, to avoid a second `ISDATETIME` check.)

Nice catch that this can't fail.

ngoldbaum (commented):
Seems worth doing to me. I got a little tripped up where I commented about whether it's legal to get rid of the error checking for `PyArray_ExtractDTypeAndDescriptor`, but I think I agree it can't fail.

Adding the new fast path makes sense to me. I didn't think carefully about whether there are any downsides to not getting the dtype in corner cases.

ngoldbaum (approved):

seberg:
LGTM, but even if it costs a nanosecond, I think this fast-path should live in `PyArray_FromAny_int`, since we don't want to lose it when moving functions over :).

(I suspect that moving it down into the dtype discovery function would be too heavy weight to be worthwhile.)

(You could keep a `descr == NULL` that skips the dtype extraction step if it seems worthwhile. I'll assume the other few integer ops about maxdims are irrelevant in practice.)

eendebakpt:
> LGTM, but even if it costs a nanosecond, I think this fast-path should live in `PyArray_FromAny_int`, since we don't want to lose it when moving functions over :).

I moved the fast path into the `PyArray_FromAny_int` and removed some more unused `context` arguments. Benchmarks have not really changed.

numpy/_core/src/multiarray/ctors.c
```diff
@@ -1520,19 +1519,18 @@ NPY_NO_EXPORT PyObject *
 PyArray_FromAny(PyObject *op, PyArray_Descr *newtype, int min_depth,
                 int max_depth, int flags, PyObject *context)
 {
+    if (context != NULL) {
+        PyErr_SetString(PyExc_RuntimeError, "'context' must be NULL");
```
seberg, line 1523:
```suggestion
        PyErr_SetString(PyExc_RuntimeError, "'context' must be NULL");
        Py_XDECREF(newtype);
```

numpy/_core/src/multiarray/ctors.c
```diff
@@ -1817,27 +1821,25 @@ NPY_NO_EXPORT PyObject *
 PyArray_CheckFromAny(PyObject *op, PyArray_Descr *descr, int min_depth,
                      int max_depth, int requirements, PyObject *context)
 {
+    if (context != NULL) {
+        PyErr_SetString(PyExc_RuntimeError, "'context' must be NULL");
+        return NULL;
```
seberg, line 1828:
```suggestion
        Py_XDECREF(descr);
        return NULL;
```

seberg (approved):
Thanks, yeah, nice followup to remove it from the private API and move the error to the public one.
We do need to steal the reference on failure, but that is a trivial thing to apply.

PR #31100, opened 2026-03-29
MAINT: use pointers to dtype dtypes instead of direct object access

mattip:
### PR summary
Pulled out of #31071.
- Add pointers to the dtype type objects.
- Use the pointers internally
- Create some NUMPY_2_5 api of the pointers, re-casting the existing export table.

This will make it easier to move to heap types.

#### AI Disclosure
Claude helped me fix the fallout of the rough regex replacements I did. I only used AI to find the cause of compilation failures and typos, but fixed the problems myself.

Once this goes in I will rebase #31071 on top of it

### DIFF
--- a/doc/source/reference/c-api/types-and-structures.rst
+++ b/doc/source/reference/c-api/types-and-structures.rst
@@ -807,9 +807,9 @@ PyArrayMethod_Context and PyArrayMethod_Spec
 PyArray_DTypeMeta and PyArrayDTypeMeta_Spec
 -------------------------------------------
 
-.. c:var:: PyTypeObject PyArrayDTypeMeta_Type
+.. c:var:: PyTypeObject PyArrayDTypeMeta_TypePtr
 
-   The python type object corresponding to :c:type:`PyArray_DTypeMeta`.
+   A pointer to the python type object corresponding to :c:type:`PyArray_DTypeMeta`.
 
 .. c:type:: PyArray_DTypeMeta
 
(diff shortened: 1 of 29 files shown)


doc/source/reference/c-api/types-and-structures.rst
```diff
@@ -919,25 +919,27 @@ Exposed DTypes classes (``PyArray_DTypeMeta`` objects)
...
+* ``PyArray_FloatAbstractDTypePtr``
+* ``PyArray_ComplexAbstractDTypePtr``
 
 .. warning::
     As of NumPy 2.0, the *only* valid use for these DTypes is registering a
     promoter conveniently to e.g. match "any integers" (and subclass checks).
-    Because of this, they are not exposed to Python.
+    Because of this, they are not exposed to Python. From NumPy 2.0 to 2.5 they
+    were exposed without the `Ptr` suffix, at some point the non-pointer version
+    will be removed.
```
ngoldbaum, line 942: I don't think this is enough for someone who is implementing a user DType to understand what to do about this change.

To make that concrete - how should e.g. numpy_quaddtype adapt to this API change? It's not completely clear to me what to do based on these docs changes. The text you've added here also isn't very greppable for someone who is searching for the old name in the docs.

numpy/_core/include/numpy/ndarraytypes.h
```diff
@@ -1887,7 +1887,11 @@ typedef struct {
      * Part of this (at least the size) is expected to be public API without
      * further modifications.
      */
-    /* TODO: Make this definition public in the API, as soon as its settled */
+    NPY_NO_EXPORT extern PyTypeObject *PyArrayDTypeMeta_TypePtr;
+    /*
+     * This is made public in _public_dtype_api_table.h via the PyArray_API table,
+     * where it is a dereferenced pointer. Maybe deprecate at some point.
```
ngoldbaum, line 1893: given that isn't not deprecated, we probably still need to document it...

seberg:
Maybe I didn't think about it enough on the other PR? Because while it seems generally nicer if these were exposed as the pointers, I dunno if there is much of a need in practice `&(*(...))`?

One thing is that it isn't just the DTypes, if we have a reason to ask users to switch, then it makes sense for all the other types as well (array, scalars, iterators, ...).
I don't hate this, but Nathan is right, it is churn for users, and right now I am not sure there is any gain except a faint feeling that it is a pointer, so it would be nicer to use it as one...

All I didn't love is mixing the two patterns (`..._Type` sometimes being a reference and sometimes not).
But... for that purpose renaming `PyArrayDTypeMeta_Type` to `PyArrayDTypeMeta_TypePtr` (and _maybe_ adding a `#define PyArrayDTypeMeta_Type (*PyArrayDTypeMeta_TypePtr)` macro) is sufficient.
FWIW, if you don't find mixing it weird, then I am OK with just doing what you have in the other PR...

PR #31105, opened 2026-03-30
MAINT: f2py: Stop setting re._MAXCACHE to 50.

WarrenWeckesser:
The value has been set to 50 for more than 20 years.  Computers generally have much more memory now than they did 20 years ago, so we should be able to let `re` use the default.

Closes gh-31104.

#### AI Disclosure

No AI tools used.

### DIFF
--- a/numpy/f2py/crackfortran.py
+++ b/numpy/f2py/crackfortran.py
@@ -243,7 +243,6 @@ def outmess(line, flag=1):
         sys.stdout.write(line)
 
 
-re._MAXCACHE = 50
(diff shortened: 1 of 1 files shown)


tylerjereddy:
That seems obviously correct at this point indeed.

mattip:
Thanks @WarrenWeckesser.

The docs build CI could not find the scipy intersphinx reference. Hopefully that is a once-off failure.

charris:
I tagged this for backport, seems pretty safe.

PR #31108, opened 2026-03-31
ENH, PERF: small arrays/dimensions on array object - with special allocator                           

mhvk:
This is an alternative to gh-31092 (and shares essentially the same first commit).

### Details
This PR stores the dimensions, strides, and small arrays on the instance, and uses a new allocator to deal with possible resizes. This is less intrusive than gh-31092 in the rest of the code, as the custom memory allocator can be used as normal. But the performance is slightly worse because of slower checks that freeing memory is not needed.

### Timings
Timings relative to having dimensions, strides and data not on the object (results for gh-31092 in parenthesis)
```
%timeit np.empty(())
133->73 ns (68)

a = 1.6
%timeit np.array(a)
151->92 ns (83)

%timeit np.add(a, a)
581->431 ns (423)

b = np.array(a)
%timeit np.add(b, b)
293->248 ns (225)
```

EDIT: storing the strides helps also for arrays that are views (so independent of how data allocation is done):
```
a = np.arange(10.)

%timeit a.view()
49.9 -> 36 ns

%timeit a.reshape((2, 5))
114 -> 97 ns

b = np.ones((5, 2))
%timeit b[0]
67 -> 49 ns

```

### DIFF
--- a/numpy/_core/src/multiarray/alloc.cpp
+++ b/numpy/_core/src/multiarray/alloc.cpp
@@ -468,6 +468,97 @@ PyDataMem_Handler default_handler = {
 PyObject *PyDataMem_DefaultHandler;
 PyObject *current_handler;
 
+// For small allocations, we store data on the instance, with the size
+// in bytes stored right in front.
+// If the data is resized to something bigger, we allocate new memory.
+// To distinguish this case, it will also have extra space in front
+// for a size, but set to -1.
+// Note: malloc and calloc are only included for completeness. They
+// are not used in the numpy code
+static inline void *
+on_instance_malloc(void *NPY_UNUSED(ctx), size_t size)
+{
+    const size_t extra = sizeof(npy_intp*);
+    void *base = PyDataMem_NEW(size + extra);
+    if (base == NULL) {
+        return NULL;
+    }
+    *(npy_intp *)base = -1;
+    return (void*)((char*)base + extra);
+}
+
+static inline void *
+on_instance_calloc(void *NPY_UNUSED(ctx), size_t nelem, size_t elsize)
+{
+    const size_t extra = sizeof(npy_intp*);
+    if (npy_mul_with_overflow_size_t(&nelem, nelem, elsize)) {
+        return NULL;
+    }
+    void *base = PyDataMem_NEW_ZEROED(nelem + extra, 1);
+    if (base == NULL) {
+        return NULL;
+    }
+    *(npy_intp *)base = -1;
+    return (void*)((char*)base + extra);
+}
+
+static inline void
+on_instance_free(void *ctx, void *ptr, size_t size)
+{
+    const size_t extra = sizeof(npy_intp*);
+    void *oldbase = (void*)((char*)ptr - extra);
+    if (*(npy_intp*)oldbase < 0) {
+        // Use PyDataMem_FREE to ensure tracking skipped in UserFree.
+        PyDataMem_FREE(oldbase);
+    }
+    // else: data is on instance, nothing to free.
+}
+
+static inline void *
+on_instance_realloc(void *ctx, void *ptr, size_t new_size)
+{
+    void *newbase;
+    const size_t extra = sizeof(npy_intp*);
+    void *oldbase = (void*)((char*)ptr - extra);
+    if (*(npy_intp*)oldbase >= 0) {
+        npy_uintp old_size = *(npy_uintp*)oldbase;
+        if (new_size <= old_size) {
+            *(npy_intp*)oldbase = new_size;  // new size fits, just store it.
+            return ptr;
+        }
+        newbase = PyDataMem_NEW(new_size + extra);
+        if (newbase == NULL) {
+            // malloc failed, nothing more to do.
+            return NULL;
+        }
+        // Move existing data over, past storage for size.
+        memmove((char*)newbase + extra, (char*)ptr, old_size);
+    }
+    else {
+        // Not on instance, so default reallocation, but with extra space.
+        newbase = PyDataMem_RENEW(oldbase, new_size + extra);
+        if (newbase == NULL) {
+            return NULL;
+        }
+    }
+    *(npy_intp*)newbase = -1;
+    return (void *)((char*)newbase + extra);
+}
+
+/* Memory handler global default */
+PyDataMem_Handler on_instance_handler = {
+    "on_instance_allocator",
+    1,
+    {
+        NULL,            /* ctx */
+        on_instance_malloc,  /* not used normally */
+        on_instance_calloc,  /* not used normally */
+        on_instance_realloc, /* realloc */
+        on_instance_free     /* free */
+    }
+};
+
+
 int uo_index=0;   /* user_override index */
 
 /* Wrappers for the default or any user-assigned PyDataMem_Handler */
@@ -485,6 +576,10 @@ PyDataMem_UserNEW(size_t size, PyObject *mem_handler)
     result = handler->allocator.malloc(handler->allocator.ctx, size);
     if (result == NULL) {
         return NULL;
+    }
+    if (handler == &on_instance_handler) {
+        // Since data can be on instance, tracking is done by the handler.
+        return result;
     }
     int ret = PyTraceMalloc_Track(NPY_TRACE_DOMAIN, (npy_uintp)result, size);
     if (ret == -1) {
@@ -507,6 +602,10 @@ PyDataMem_UserNEW_ZEROED(size_t nmemb, size_t size, PyObject *mem_handler)
     if (result == NULL) {
         return NULL;
     }
+    if (handler == &on_instance_handler) {
+        // Since data can be on instance, tracking is done by the handler.
+        return result;
+    }
     int ret = PyTraceMalloc_Track(NPY_TRACE_DOMAIN, (npy_uintp)result, nmemb * size);
     if (ret == -1) {
         handler->allocator.free(handler->allocator.ctx, result, size);
@@ -524,6 +623,11 @@ PyDataMem_UserFREE(void *ptr, size_t size, PyObject *mem_handler)
     if (handler == NULL) {
         WARN_NO_RETURN(PyExc_RuntimeWarning,
                      "Could not get pointer to 'mem_handler' from PyCapsule");
+        return;
+    }
+    if (handler == &on_instance_handler) {
+        // Since data can be on instance, tracking is done by the handler.
+        handler->allocator.free(handler->allocator.ctx, ptr, size);
         return;
     }
     // ignore -2 return when tracemalloc is disabled
@@ -546,6 +650,10 @@ PyDataMem_UserRENEW(void *ptr, size_t size, PyObject *mem_handler)
     if (result == NULL) {
         // ptr is still valid here
         return NULL;
+    }
+    if (handler == &on_instance_handler) {
+        // Since data can be on instance, tracking is done by the handler.
+        return result;
     }
     int ret = PyTraceMalloc_Untrack(NPY_TRACE_DOMAIN, (npy_uintp)ptr);
     if (ret == -2) {
--- a/numpy/_core/src/multiarray/arrayobject.c
+++ b/numpy/_core/src/multiarray/arrayobject.c
@@ -455,7 +455,7 @@ _clear_array_attributes(PyArrayObject *self, npy_bool unraisable)
     }
 
     /* must match allocation in PyArray_NewFromDescr */
-    npy_free_cache_dim(fa->dimensions, 2 * fa->nd);
+    npy_free_cache_dim_array(self); // frees it if not on instance.
     fa->dimensions = NULL;
     Py_CLEAR(fa->descr);
     return 0;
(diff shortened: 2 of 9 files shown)


mhvk:
Hmm, the macOS_arm64 failure is strange, and possible related to the `imag` ufunc,
```
PARALLEL FAILED numpy/_core/tests/test_arrayobject.py::test_real_imag_ufunc_minimal[imag-imag] - RecursionError: Stack overflow (used 16364 kB) in __subclasscheck__
```

numpy/_core/src/multiarray/alloc.cpp
```diff
@@ -468,6 +468,89 @@ PyDataMem_Handler default_handler = {
...
+    void *base = PyDataMem_NEW(size + extra);
+    *(npy_intp *)base = -1;
+    return (void*)((char*)base + extra);
+}
+
+static inline void *
+on_instance_calloc(void *NPY_UNUSED(ctx), size_t nelem, size_t elsize)
+{
+    const size_t extra = sizeof(npy_intp*);
+    size_t sz = nelem * elsize + extra;
```
ngoldbaum: This can overflow and signed integer overflow is UB. I think you can use `npy_mul_sizes_with_overflow` instead here to do the overflow checking, but it's possible there are circular dependency issues I'm not noticing.

numpy/_core/src/multiarray/alloc.cpp
```diff
@@ -468,6 +468,89 @@ PyDataMem_Handler default_handler = {
...
+
+static inline void
+on_instance_free(void *ctx, void *ptr, size_t size)
+{
+    const size_t extra = sizeof(npy_intp*);
+    void *oldbase = (void*)((char*)ptr - extra);
+    if (*(npy_intp*)oldbase < 0) {
+        // Use PyDataMem_FREE to ensure tracking skipped in UserFree.
+        PyDataMem_FREE(oldbase);
+    }
```
ngoldbaum, line 513:
```suggestion
    }
    // else: data is on instance, nothing to free -- OK
```

Just to make it clearer that there is another case, it's just a no-op right now.

numpy/_core/src/multiarray/arrayobject.c
ngoldbaum, line 452: I think this might need to be made unconditional (e.g. lift it outside the OWNDATA if block), otherwise you might leak the `mem_handler` if allocating an array fails before the  OWNDATA flag is set. That can happen if `fill_zer_info.func` fails.

ngoldbaum (commented):
I did a code review focusing on correctness with some help from Claude Code, see inline comments.

Also, are the timings in the description from this PR or from the old one? The description implies they're from the old PR but I think that might be a typo.

numpy/_core/src/multiarray/alloc.cpp
```diff
@@ -468,6 +468,89 @@ PyDataMem_Handler default_handler = {
...
+// If the data is resized to something bigger, we allocate new memory.
+// To distinguish this case, it will also have extra space in front
+// for a size, but set to -1.
+// Note: malloc and calloc are only included for completeness. They
+// are not used in the numpy code
+static inline void *
+on_instance_malloc(void *NPY_UNUSED(ctx), size_t size)
+{
+    const size_t extra = sizeof(npy_intp*);
+    void *base = PyDataMem_NEW(size + extra);
```
ngoldbaum, line 482: You don't check for failure here and a couple other places you call this function below.

mhvk:
Just a quick reply: the timings on top give both the results for this PR and the old one (latter in parentheses); this one is slightly less efficient because of the need to "free" the allocation (which gets a no-op obviously). But I personally think it is the better approach as it is less disruptive, even if hacky at the memory allocator level.

numpy/_core/src/multiarray/alloc.cpp
mhvk: These are all `size_t`, while `npy_mul_sizes_with_overflow` is for `npy_intp`. In the routines above, there is also a simple multiplication done...

mhvk: Ah, never mind, there is a `npy_mul_sizes_with_overflow_size_t` as well... Using that now.

numpy/_core/src/multiarray/alloc.cpp
mhvk: Done

numpy/_core/src/multiarray/alloc.cpp
mhvk: Oops, thanks, and one sees the risks of writing something that does not get used.

I now added tests for using the on-instance handler as well.

ngoldbaum:
I opened https://github.com/numpy/numpy/issues/31112 to track the free-threaded CI failure, it's unrelated to this.

mhvk:
@ngoldbaum - thanks for the review! I think I address all the comments.

One general question (also to @eendebakpt and @seberg): between the two approaches, one with flags on the instance to tell where data is, or, as here, a custom (and slightly hacky) allocator, which do you prefer? I ask to avoid having to update two PRs... My own preference is for the approach here, with an allocator, as I think with this approach whether or not data is on the instance becomes truly an implementation detail.

ngoldbaum:
We talked a little about this general idea at the triage meeting. @seiko2plus pointed out that these benchmarks aren't terribly realistic - do you have a motivation here in some other project where this leads to real speedups outside of toy benchmarks? It'd be nice to also get info about your testing setup.

For example, on free-threaded Python 3.15 and newer, `PyMem_RawMalloc` uses mimalloc, which may already do a good job of avoiding syscalls for small allocations. Maybe try your microbenchmarks on 3.15t?

mhvk:
@ngoldbaum - the benchmark for me that counts is the ufunc call on a small array -- trying to reduce the overheads for those (and array scalars). This also helps astropy's `Quantity` which is used a lot with small arrays and people complain about it being very slow.

Note that the overhead is only partially allocating more memory -- even now numpy uses a cache for small allocations. I think the bigger overhead is actually the tracking that is done, i.e., the hoops to jump through to support different memory allocators.

mhvk:
I mentioned that this was part of an effort to speed up ufuncs, but perhaps worth noting the preceding preparatory work: main preceding one was adding a fast-path for scalars in single-argument ufuncs (#29819), and preparatory work included sanitizing how array state changes are handled (#29925, #29929, #30414) .

eendebakpt:
Like Marten I also care about the performance of ufuncs for small arrays and scalars. On my system

| Benchmark | merge base | PR #31108 | Change |
|---|---|---|---|
| `np.abs(x)` | 323 ns | 260 ns | 1.24x faster |
| `np.cos(x)` | 318 ns | 258 ns | 1.23x faster |
| `np.add.accumulate(x)` | 412 ns | 352 ns | 1.17x faster |
| `np.add.reduce(x)` | 724 ns | 669 ns | 1.08x faster |
| `np.abs(-2.2)` | 83.2 ns | 87.1 ns | 1.05x slower |
| `np.cos(-2.2)` | 74.1 ns | 77.1 ns | 1.04x slower |
| `np.abs(np.float64(-2.2))` | 120 ns | 123 ns | 1.02x slower |
| `np.cos(np.float64(-2.2))` | 111 ns | 113 ns | 1.01x slower |
| Geometric mean | (ref) | 1.07x faster | |

(Python 3.13.3 installed via uv on Linux)

The performance increase of this PR is a few percent, but combined with other improvements to the ufuncs I hope we can improve performance by a factor 2 (see https://github.com/numpy/numpy/issues/24345#issuecomment-4138412542).

The custom memory allocators  indeed also matter for performance. The `PyCapsule` interface is part of the public API so that is probably fixed, but internally we can still improve it a bit.

mhvk:
@eendebakpt - your comment about the `PyCapsule` made me wonder if it was worth trying to avoid that too. It really only affects free-ing. As a test, if I avoid calling `PyDataMem_UserFree` in the deallocation code, there is a slight reduction, especially for the ufunc call involving a scalar, which makes multiple temporary arrays. Comparing with this PR as is,
```
%timeit np.empty(())
75.5 -> 69.8 ns

a = 1.6
%timeit np.array(a)
95.7->87.5 ns

%timeit np.add(a, a)
434 -> 416 ns

b = np.array(a)
%timeit np.add(b, b)
248->238 ns
```
Possibly worth it...

mhvk:
Hmm, on second thought, I think that the transformation of a scalar in an array in the ufunc can also be dealt with by creating a read-only temporary array that has the scalar as its base. That may be even faster. So, right now it seems to me that special-casing in the deallocation code is adding complexity for too little benefit--- the one thing I like about the approach here is that it is fairly surgical, adding a new memory handler and only changing the allocation code.

mhvk:
The remaining two failures seem unrelated. Both state
```
..\numpy\meson.build:185:4: ERROR: Problem encountered: No BLAS library detected! Install one, or use the `allow-noblas` build option (note, this may be up to 100x slower for some linear algebra operations).
```

seberg:
I am a bit torn about the approach, this is nicer about `realloc` special casing especially. But other than that I think I actually lean towards the other approach.

I.e. the NumPy memory handler is in that case a bit special and we assume it is fine to special case e.g. `realloc` there. 
(Yeah, I suppose I'll dismiss anyone using the `mem_handler->realloc()` directly as a reasonable use-case to call on a NumPy array. There seems very little point in that.)

(For the other PR, I am wondering if it doesn't make sense to just leave the `mem_handler` as the default one, even if we are not technically using it. But maybe it doesn't really matter.)

> Hmm, on second thought, I think that the transformation of a scalar in an array in the ufunc can also be dealt with by creating a read-only temporary array that has the scalar as its base.

This could be interesting if it gets us the more interesting chunk here?  I don't love using the scalar raw data too much, but it may be the practical path for a quick scalar->0d array path.
(One reason is, we do leak such arrays at least via `__array_wrap__` which we shouldn't worry much about, but it may be best to not leak very weird ones.)

> This also helps astropy's `Quantity` which is used a lot with small arrays and people complain about it being very slow.

Might need to figure out to extend the approach to Python only subclasses at least maybe? (There will be at least a new `__dict__` field that needs to be `NULL`'d, though.)

mhvk:
@seberg - hmm, maybe the approach for the data needs more thought--I had not thought about just hacking the default allocator (though like that with my new allocator there is a way to tell where the data is).

Do you think it makes sense to split off the first commit, which puts the dimensions and strides on the instance? I.e., reopen gh-31096, but with a slightly updated commit? That has smaller speed-up benefits, but also seems much more obviously the right thing to do (and helps every array allocation, including taking views, etc.).

seberg:
> Do you think it makes sense to split off the first commit, which puts the dimensions and strides on the instance? 

Yeah, likely, but up you; if just to make the second part simpler (also even there a brief release note may make sense)?  I suppose that is a pretty small speed improvement (I have no idea how much right now).
(That improvement shouldn't really do anything for scalars, since they are 0-D and don't need to allocate anything...)

> just hacking the default allocator

Ah interesting, maybe that would be nicer.  I was actually just thinking that it seems fine if we special case things when the default allocator is used.

> trying to reduce the overheads for those (and array scalars).

On the general discussion. I tend to think that speed-ups for small arrays are probably quite worthwhile (i.e. not uncommon in real world code even if that is often not code that is timed carefully) of course we need to balance complexity; and yeah it is a matter of opinion.
(One related thing we should possibly check is whether the small allocations cache even matters if `mimalloc` is used in Python.)

seberg:
Hmmmmpf, as much as I hate it.  Don't we have to wait a bit with all of this due that f2py nonsense?
I.e. that terrible swapping is compiled into the downstream project, so technically downstream has to recompile with the new f2py version, which means breaking ABI (if this is used).

Maybe this is all incredibly niche, but it might make sense to delay at least for 2.6 here :/?

mhvk:
> Hmmmmpf, as much as I hate it. Don't we have to wait a bit with all of this due that f2py nonsense? I.e. that terrible swapping is compiled into the downstream project, so technically downstream has to recompile with the new f2py version, which means breaking ABI (if this is used).

Yikes, I think you are right... Indeed, just to be sure I checked with the `inplace` routine that #29929 added, compiling it with my regular system numpy. On current main, it works fine, with this PR, it does not (it's bad in giving all-zero results, not an error message or core dump...).

If you think waiting for 2.6 would address that, then I am fine with that. It certainly is a corner case.

Even better of course would be able to recognize this is happening... I had a look with `nm *so` at what is in the f2py produced module, but see nothing related to numpy, not even `import_array()` - but I'm not really sure how that works anyway. In particular, would we be able to recognize an `f2py` module inside `import_array`?

mhvk:
Probably worth noting that storing the strides and dimensions on the instance affects all arrays; it makes views and reshapes a bit cheaper -- see timings I added on top.

seberg:
I think `import_array()` must be generated somewhere, but not sure where.  Either way, there won't be a way to recognize it from NumPy, though.

eendebakpt:
> Hmmmmpf, as much as I hate it. Don't we have to wait a bit with all of this due that f2py nonsense? I.e. that terrible swapping is compiled into the downstream project, so technically downstream has to recompile with the new f2py version, which means breaking ABI (if this is used).
> 
> Maybe this is all incredibly niche, but it might make sense to delay at least for 2.6 here :/?

Is there something we can already do in 2.5 that would make the change easier in 2.6?

seberg:
> Is there something we can already do in 2.5 that would make the change easier in 2.6?

No, we already did that all we can do now is wait (and decide how long we need to wait).

PR #31119, opened 2026-04-02
DEP: expire deprecation for null handling of `context` and `return_scalar` in `arraywrap`

nakul-krishnakumar:
### PR summary

Progresses #30440

#### AI Disclosure

No AI tools used

### DIFF
--- a/numpy/_core/src/multiarray/arraywrap.c
+++ b/numpy/_core/src/multiarray/arraywrap.c
@@ -137,7 +137,6 @@ npy_apply_wrap(
     PyObject *res = NULL;
     PyObject *new_wrap = NULL;
     PyArrayObject *arr = NULL;
-    PyObject *err_type, *err_value, *traceback;
 
     /* If provided, we prefer the actual out objects wrap: */
     if (original_out != NULL && original_out != Py_None) {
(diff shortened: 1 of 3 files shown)


nakul-krishnakumar:
I am not sure about when to add the release notes, it would be really helpful if anybody could guide me on that. Thank you!

jorenham:
> I am not sure about when to add the release notes, it would be really helpful if anybody could guide me on that. Thank you!

Expired deprecations indeed require a release note. See https://github.com/numpy/numpy/blob/main/doc/release/upcoming_changes/README.rst for details.

nakul-krishnakumar:
@jorenham Thank you for the reference, I have added the release notes.

numpy/_core/src/multiarray/arraywrap.c
```diff
@@ -236,59 +235,6 @@ npy_apply_wrap(
...
-     */
-    PyErr_Fetch(&err_type, &err_value, &traceback);
-    res = PyObject_CallFunctionObjArgs(wrap, arr, NULL);
-    if (res == NULL) {
-        Py_DECREF(err_type);
-        Py_XDECREF(err_value);
-        Py_XDECREF(traceback);
-        goto finish;
-    }
-
```
JosephMehdiyev, line 277: It should a throw error here for example

numpy/_core/src/multiarray/arraywrap.c
```diff
@@ -236,59 +235,6 @@ npy_apply_wrap(
             wrap, arr, py_context,
             (return_scalar && PyArray_NDIM(arr) == 0) ? Py_True : Py_False,
             NULL);
-    if (res != NULL) {
-        goto finish;
-    }
```
JosephMehdiyev, line 241: why is this deleted? The result will be the same since all the code between the `finish`  label is deleted, but instead of warning,  code should throw an error I think. So this code should stay (after adding the error logic). New error tests instead of warning tests are needed too

seberg:
Thanks, this LGTM.  But, I have a bit of a feeling that people might have managed to blissfully not notice this, e.g. torch here: https://github.com/pytorch/pytorch/blob/main/torch/_tensor.py#L1259

So that makes me wonder a bit if we should at least only bump it to a `VisibleDeprecationWarning`, and maybe also open a torch issue and wait til after 2.5 (since that is very soon) to merge.

I.e. as much as I would like to get this done... there seems little rush here.

nakul-krishnakumar:
> Thanks, this LGTM. But, I have a bit of a feeling that people might have managed to blissfully not notice this, e.g. torch here: https://github.com/pytorch/pytorch/blob/main/torch/_tensor.py#L1259
> 
> So that makes me wonder a bit if we should at least only bump it to a `VisibleDeprecationWarning`, and maybe also open a torch issue and wait til after 2.5 (since that is very soon) to merge.
> 
> I.e. as much as I would like to get this done... there seems little rush here.

That makes sense, thanks for the suggestion.
I will go ahead and open a PR for converting it to `VisibleDeprecationWarning` and open an issue with PyTorch to check whether this behavior is relied upon there. 
Does that sound good?

seberg:
Yes, that would be great!  Could also convert this PR, but either way is good.

PR #31120, opened 2026-04-02
TYP: ``cum{sum,prod}`` and ``cumulative_{sum,prod}`` shape-typing

jorenham:
This adds shape-typing support for 

- `cumsum`
- `cumulative_sum`
- `cumprod`
- `cumulative_prod`

For `axis=None` (default), these always return a 1-d array. Otherwise, when `axis` is given, the shape-type is preserved.

This came up in https://github.com/python/mypy/issues/18343#issuecomment-4169055346

CC @tyralla

#### AI Disclosure


tyralla:
Cool.  I checked it in the context where I encountered the problem.  Works as expected now.  Thanks!

charris:
Thanks Joren.

PR #31121, opened 2026-04-02
MAINT: Rename ppc64le and s390x workflow

AlekseiNikiforovIBM:
### PR summary
Workflow file no longer contains only ppc64le jobs. Use more common name.

Previously the renaming was postponed to make review of changes easier.

Follow up to #30819.

#### AI Disclosure
No AI tools used


AlekseiNikiforovIBM:
@rgommers @seiko2plus , please take a look at this follow up PR. Keeping current file name or some other file name would also work for me.

charris:
close/reopen

seiko2plus:
@AlekseiNikiforovIBM, It's fine to rename the workflow file to a more convenient name that fits both IBM architectures, but I'm not sure about your chosen name. `linux-actionspz.yml` seems a bit mysterious to me; why don't you just call it `linux-ibm.yml` so it can align with the label checks names?

AlekseiNikiforovIBM:
> @AlekseiNikiforovIBM, It's fine to rename the workflow file to a more convenient name that fits both IBM architectures, but I'm not sure about your chosen name. `linux-actionspz.yml` seems a bit mysterious to me; why don't you just call it `linux-ibm.yml` so it can align with the label checks names?

Done. Thanks for feedback.

seiko2plus:
@AlekseiNikiforovIBM, Thank you.

PR #31126, opened 2026-04-02
PERF: branchless calendar algorithms for datetime conversions

jbrockmendel:
Claude wrote this, i reviewed manually.  As with my other PRs, my C-fu is weak.

This is explicitly modeled on https://github.com/pandas-dev/pandas/pull/64662.

cc @seberg

### DIFF
--- a/numpy/_core/src/multiarray/datetime.c
+++ b/numpy/_core/src/multiarray/datetime.c
@@ -116,64 +116,27 @@ is_leapyear(npy_int64 year)
 
 /*
  * Calculates the days offset from the 1970 epoch.
+ *
+ * Adapted from Hinnant's days_from_civil algorithm (public domain).
+ * See: https://howardhinnant.github.io/date_algorithms.html#days_from_civil
+ *
+ * The March-1 epoch trick places the leap day at the end of the year,
+ * eliminating special cases for February.
  */
 NPY_NO_EXPORT npy_int64
 get_datetimestruct_days(const npy_datetimestruct *dts)
 {
-    int i, month;
-    npy_int64 year, days = 0;
-    int *month_lengths;
+    npy_int64 y = dts->year - (dts->month <= 2);
+    npy_int64 era = (y >= 0 ? y : y - 399) / 400;
+    npy_uint32 yoe = (npy_uint32)(y - era * 400);           /* [0, 399] */
+    npy_uint32 doy =
+        (153 * (npy_uint32)(dts->month > 2 ? dts->month - 3
+                                           : dts->month + 9) + 2) /
+            5 +
+        (npy_uint32)dts->day - 1;                            /* [0, 365] */
+    npy_uint32 doe = yoe * 365 + yoe / 4 - yoe / 100 + doy; /* [0, 146096] */
 
-    year = dts->year - 1970;
-    days = year * 365;
-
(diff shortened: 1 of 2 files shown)


ngoldbaum:
Can we have a little more context why we want this? Speed, correctness, something else?

jbrockmendel:
The benefit is performance.  I see a 5% speedup in ns->D conversion, 18% in ns->M.

charris:
close/reopen

ngoldbaum:
Can we get some concrete benchmarks demonstrating the speedup? Can you also verify that our test coverage for these conversions is good? I'm a little nervous about completely changing the implementation of a poorly tested codepath.

If you know of an expert we can ping to review this, that might also help assuage concerns about correctness.

jbrockmendel:
> Can we get some concrete benchmarks demonstrating the speedup?

```
In [3]: arr = np.arange("2000-01-01", "2010-01-01", dtype="datetime64[s]").astype("datetime64[ns]")

In [4]: %timeit arr.astype("datetime64[M]")
3.55 s ± 25.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)  # <- main
2.95 s ± 58.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)  # <- PR

In [5]: %timeit arr.astype("datetime64[D]")
416 ms ± 99.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)  # <- main
350 ms ± 32.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)  # <- PR
```

> Can you also verify that our test coverage for these conversions is good?

claude says "The existing 470 tests already cover pre-1970 dates, leap year boundaries, year 0/negatives, Y2038, and int64 limits"

> I'm a little nervous about completely changing the implementation of a poorly tested codepath.

That's similar to the comment in pandas asking me to open this PR.  The hope is that by keeping the implementations aligned we can benefit from each others' coverage.

FWIW my understanding is that the same algorithm is used in c++20's chrono implementation.  We just can't use it directly because they don't support int64 years.

ngoldbaum:
I added a triage review label to make sure we talk about this at next week's triage meeting. Feel free to come, see [the calendar](https://scientific-python.org/calendars/).

> FWIW my understanding is that the same algorithm is used in c++20's chrono implementation.

Oh nice, it might also help if you add a link to an implementation of this as a comment. Someone could also try comparing the two implementations.

seberg:
Yeah, I think it probably makes sense to just follow pandas, it seems like the sources are trustworthy, although reviewing that it is identical code isn't all that easy (bots help a bit with confidence, I guess).
It seems the GCC version: https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=libstdc%2B%2B-v3/include/std/chrono;h=10e868e5a036924177cf952a81f4bce10c497106;hb=HEAD#l1678
uses the same/very similar approach, but very slightly different constants (no full understanding here, but it seems related to the valid range of years).

seberg (approved):
OK, hard to follow, but I think it aligns with other implementations and the tests seem to cover the maximum values fine, so I think they are relatively good.

Thanks @jbrockmendel, let's get this in.  There is definitely value in keeping pandas/numpy aligned.

PR #31133, opened 2026-04-03
TYP: ``inner`` shape-typing and transparent dtypes

jorenham:
`numpy.inner` was effectively untyped, because doing so would require some tricky shape-typing acrobatics. 
I finally got around to it, and added shape-typing support for mixed 1d and 2d inputs, overloads for pure-python 1d array-likes (i.e. flat sequences) with mixed scalar-types, and special-casing for object dtypes. 
Those weird first two `_ArrayJustND` overloads that have `Never` in their input shape types are workarounds for mypy and pyright, which (still) don't treat overloads as dictated by the typing spec (grumble).

#### AI Disclosure


charris:
Thanks Joren.

PR #31134, opened 2026-04-03
TYP: ``fromiter`` shape-typing and more dtype overloads

jorenham:
`numpy.fromiter` always returns a 1-d array, and now the stubs reflect this.
This also adds overloads for specific non-numpy-like `dtype` types such as `dtype=float` and `dtype=None`. Note that because of the static overlap in `bool`, `int`, `float`, and `complex`,  the output types of these overloads must be compatible, which is what the `| Any` is for.

#### AI Disclosure

### DIFF
--- a/numpy/_core/multiarray.pyi
+++ b/numpy/_core/multiarray.pyi
@@ -994,23 +994,65 @@ def fromfile(
     like: _SupportsArrayFunc | None = ...,
 ) -> _Array1D[Any]: ...
 
-@overload
+#
+@overload  # dtype=<known>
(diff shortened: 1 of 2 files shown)


jorenham:
Ah, `_ScalarLike` is apparently too narrow as `Iterable` element type

charris:
Thanks Joren. I assume you are thinking about further fixes here.

jorenham:
> Ah, `_ScalarLike` is apparently too narrow as `Iterable` element type

I guess I'll fix that later then :stuck_out_tongue:

PR #31135, opened 2026-04-03
TYP: ``asarray`` shape-typing

jorenham:
This adds several overloads to `asarray` to pass along the input shape-type if known, or deduce it in case of 1-d and 2-d sequences of scalars. Overloads were also added for (sequences of) sequences of specific python scalars, so that e.g. `np.asarray([1, 2])` will return a `np.ndarray[tuple[int], np.dtype[np.int_]]` -- a 1-d `int_` array.

#### AI Disclosure


charris:
How should I read the mypy_primer output? In particular, the colors red and green and the difference between "note" and "error".

charris:
Thanks Joren. You seem happy with this, so in it goes.

jorenham:
> How should I read the mypy_primer output? In particular, the colors red and green and the difference between "note" and "error".

If a particular code line is both red and green then only the reporting has changed; so nothing to worry about.

For pandas, there's a new error - "redundant cast" - which is a good thing, because casting was a workaround here, and now now they'll be able to remove this workaround

For xarray there was a similar workaround, but they used a `# type: ignore`, which is now no longer necessary.

So the bottom line is that this resolves two issues.

PR #31142, opened 2026-04-04
TYP: ``dot`` shape-typing and transparent dtypes

jorenham:
This is pretty similar to the shape-typing improvements for `np.inner` from #31133. But since `np.dot` shape-type signatures is more complicated, and because there are already 19 overloads now, it's a more incomplete shape-typing solution than for `np.inner`. But the hope is that these overloads should cover its most common use cases.

#### AI Disclosure


charris:
Thanks Joren.

PR #31143, opened 2026-04-04
TYP: ``nonzero`` and single-arg ``where`` shape-typing

jorenham:
`nonzero`, `ndarray.nonzero`, and `where` (single arg) return a tuple of 1d `intp` arrays, where the number of tuple elements is the number of input dimensions. This adds shape-aware overloads to these functions for the first three `ndim` cases.

#### AI Disclosure


jorenham:
The new mypy primer error for pandas is correct; `np.where` raises if you pass it a bare integer, so passing `int | ndarray[tuple[Any, ...], dtype[Any]]` is unsafe, and should indeed be reported as an error.

charris:
Needs rebase.

charris:
Thanks Joren.

PR #31148, opened 2026-04-04
TYP: ``fromregex`` shape-typing

jorenham:
It always returns a 1d array, so it's a straighforward change.

#### AI Disclosure


charris:
Thanks Joren.

PR #31150, opened 2026-04-04
BUG: Fix matvec/vecmat in-place aliasing (out=input produces zeros)

Ijtihed:
### PR summary

Fixes #31144.

`np.matvec(A, b, out=b)` returns all zeros because `matvec` and `vecmat` are missing the `~NPY_ITER_OVERLAP_ASSUME_ELEMENTWISE` output flag that `matmul` already has. Without it the iterator ends up skipping the overlap check when `out` is the same array as an input so no temporary copy is made and BLAS `gemv` clobbers the input buffer.

This PR adds the same fix to `matvec` and `vecmat` and adds regression testing for both.

Hi, I'm Ijtihed Kilani ([ijtihed.com](https://ijtihed.com)); I'm an early-career software engineer based in Helsinki, Finland. I work mainly on physics-informed simulations for robotics and autonomy, currently at Supercell working on new games. I've used NumPy heavily for linear algebra in simulation code/studies.

#### AI Disclosure

This fix was made with help from Claude in Cursor. I used it to look over any assumptions I might have missed and draft the basic code changes which I then looked over, then rewrote and verified. All changes were reviewed and verified by me before submission.

Made with [Cursor](https://cursor.com)


jorenham (commented):
Thanks for your contribution. 

Could you revert the file permission changes?

Ijtihed:
> Thanks for your contribution.
> 
> Could you revert the file permission changes?

Done. I hope everything was done properly 😄

numpy/_core/tests/test_multiarray.py
```diff
@@ -3618,6 +3618,24 @@ def test_matmul_out(self):
...
+        assert_equal(c, expected)
+
+    def test_vecmat_out(self):
+        # overlapping memory
+        a = np.arange(18, dtype=float).reshape(2, 3, 3)
+        b = np.arange(6, dtype=float).reshape(2, 3)
+        expected = np.vecmat(b, a)
+        c = np.vecmat(b, a, out=b)
+        assert_(c is b)
+        assert_equal(c, expected)
```
tylerjereddy: Both tests do indeed fail before and pass after the patch. 

Note for other NumPy devs--the original problem was only reproducible for me on x86_64 Linux with OpenBLAS, not with ARM Mac/Accelerate, in case that's relevant.

numpy/_core/tests/test_multiarray.py
```diff
@@ -3618,6 +3618,24 @@ def test_matmul_out(self):
...
         assert_equal(c, b)
 
+    def test_matvec_out(self):
+        # overlapping memory
+        a = np.arange(18, dtype=float).reshape(2, 3, 3)
+        b = np.arange(6, dtype=float).reshape(2, 3)
+        expected = np.matvec(a, b)
+        c = np.matvec(a, b, out=b)
+        assert_(c is b)
+        assert_equal(c, expected)
```
tylerjereddy: I suppose we could try to parametrize the two new tests, but the complexity of the argument reversal probably negates any potential simplification.

I guess these are both very similar to the original `test_matmul_out()` above.

Minor/debatable points might be preference for `assert` over the old `assert_` that should no longer be needed in new tests, and maybe `assert_allclose` given we're comparing floats here (bit of a stretch that it matters, but if they do or could someday follow different codepaths...).

Ijtihed: happy to switch to `assert` and `assert_allclose` if you think that might be better. I was (like you mentioned) just matching the style of `test_matmul_out`

Ijtihed: should I make the edits or leave as is? im okay with either

RadiusFighter:
Thanks for your contribution, Ijtihed! This fix for the in-place aliasing issue in `matvec` and `vecmat` is crucial, and adding regression tests will help ensure stability. Looking forward to reviewing your PR!

numpy/_core/tests/test_multiarray.py
seberg: You could for example parametrize it with a `lambda A, x: np.vecmat(A, x)` and `lambda A, x: np.matvec(x, A)`.

I would also prefer if you move these to `test_ufunc.py`, there seem to be more tests for these there, and that file is at least slightly smaller.

But otherwise this looks good and really we could basically put it in as is.

Ijtihed: should be done now, thanks for the quick feedback

seberg (approved):
Thanks, LGTM now and an important fix if someone uses in-place operators here.
(Just waiting for CI.)

PR #31151, opened 2026-04-04
TYP: ``mean`` and ``ndarray.mean`` shape-typing and improved dtypes

jorenham:
This significantly improves `np.mean` and `np.ndarray.mean`. These now have support for shape-typing, and have improved input-dependent return dtypes.

This also fixes the incompatible overlapping overloads in `np.mean`, towards #27032

Once this is reviewed and merged, I'll do the same for `std` and `var`.

#### AI Disclosure


charris:
Thanks Joren.

PR #31153, opened 2026-04-04
MAINT: bump ``mypy`` to ``1.20.0``

jorenham:
Mypy 1.20 includes a change in stubtest behavior that makes it stricter in its treatment of ufuncs, which required changing several `np.char` and `np.string`  functions into ufuncs (because that's what they are). It's good that these ufuncs are now correctly annotated as such, but the downside is that their callable signatures aren't as specific as they were before. This shouldn't be a problem for downstream users because of the [*gradual guarantee*](https://typing.python.org/en/latest/spec/concepts.html#the-gradual-guarantee).

Conversely, stubtest will now treat `@type_check_only` stubclasses tranparently as their base class. That's why we no longer need most of the stubtest "allowlist" entries. This will also make it possible to improve static typing for ufuncs, because we'll now be able to create `@type_check_only` `ufunc` subtypes in the stubs for individual (or groups of) ufuncs with specialized callable signatures.  See https://github.com/python/mypy/pull/20352 for the details.

It looks like there also have been changes in the way mypy deals with (reflected) binops, so i had to update some of the type-tests for scalar arithmetic. 

release notes: https://mypy.readthedocs.io/en/stable/changelog.html

#### AI Disclosure
N/A


charris:
Thanks Joren.

PR #31154, opened 2026-04-04
TYP: ``std`` and ``var`` shape-typing and improved dtypes

jorenham:
This follows #31151, and uses the same overload pattern as for `np[.ndarray].mean` and applies it to `np[.ndarray].std`  and `np[.ndarray].var` .

If I have time tomorrow, I'll also use this overload pattern to improve the other similar aggregation functions: `sum`, `prod`, `amax`, `amin`, and `ptp`. 

I'm also planning to write a release note that summarizes the shape-typing improvements sometime soon (in a separate PR). 

#### AI Disclosure

N/A


charris:
Thanks Joren.

PR #31158, opened 2026-04-05
YAML file Python venv version change

imredavid64-glitch:
I need to do this introduction, but first hello, and I am a daily user of NumPy for all kinds of analysis and data use.

AI was used to check over the versions of Python, as if 3.14 is in development and if 3.12 is older or not, also a small YAML code check.

Onto the pull request!

So there is in the .circleci folder/config.yml file, a small error in the virtual environment, as it is 3.12 Python, and I understand that the GIL removal made it different for NumPy, so instead use the YAML code below:

- run:
          name: build NumPy (Python 3.13 No-GIL)
          command: |
            # 1. Create a venv using the free-threaded interpreter
            python3.13t -m venv venv
            . venv/bin/activate
            
            # 2. Set the 'Magic' variable to disable the GIL globally in this shell
            export PYTHON_GIL=0
            
            # 3. Install requirements 
            pip install --upgrade pip
            pip install --progress-bar=off -r requirements/test_requirements.txt \
                -r requirements/build_requirements.txt \
                -r requirements/ci_requirements.txt
            
            # 4. Install NumPy with No-GIL optimizations enabled
            # 'spin' will detect the 3.13t interpreter and build accordingly
            spin build --with-scipy-openblas=64 -j 2

      - run:
          name: Verify GIL Status
          command: |
            . venv/bin/activate
            # This short script confirms if the GIL is actually off
            python -c "import sys; print(f'GIL Enabled: {sys._is_gil_enabled()}')"

To explain a bit on this code:

python3.13t: This is a specific binary. In the 3.13 era, there are often two versions: python3.13 (standard GIL) and python3.13t (free-threaded). Using the t version is the "best practice" for No-GIL work.

export PYTHON_GIL=0: This is your manual override. Even with the 3.13t build, Python might re-enable the GIL if it detects an "old" C-extension that isn't ready. This force-disables it.

sys._is_gil_enabled(): This is the new internal function added in 3.13 to let you prove the lock is actually gone.

Again, it isn't needed to implement this, but just a small reccomendation, since to prepare for 3.14 since there are entire new interpreters, which can make NumPy even faster. But since the venv is still using the old 3.12, it won't assist in multi-tasking like with the old Python 3.12 interpreter only allowing mostly single core, but with 3.13 and even 3.14, it uses all cores, making sufficient quicker operations. Also, since the old GIL used more Python manual work, it uses more of a single core that is 100% busy, but with 3.13 and 3.14 is uses more of all the other cores you have. Thus, without this old GIL, NumPy won't have to make it's own system for quicker calculations, it can just rely on Python.


jorenham:

PR #31161, opened 2026-04-05
TYP: ``tile``: accept numpy scalars and arrays as second argument

jorenham:
This fixes an issue in `np.tile` where it incorrectly rejected numpy `integer` scalars or arrays as second (`reps`) argument.

ref: https://github.com/numpy/numpy/pull/31160#issuecomment-4189492521

I'm not sure if there'll be a 2.4.5, so I marked this as backport candidate just in case there will be. It's not a very big issue though, so it's also fine to instead wait for 2.5.0 if that's easier.

#### AI Disclosure
N/A

### DIFF
--- a/numpy/lib/_shape_base_impl.pyi
+++ b/numpy/lib/_shape_base_impl.pyi
@@ -18,6 +18,7 @@ from numpy._typing import (
     _ArrayLikeBool_co,
     _ArrayLikeComplex_co,
     _ArrayLikeFloat_co,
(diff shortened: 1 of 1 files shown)


charris:
There will definitly be a 2.4.5 and maybe a 2.4.6

charris:
Thanks Joren.

jorenham:
I'll backport this one @charris

PR #31163, opened 2026-04-05
TYP: ``[a]min``, ``[a]max``, ``ptp``: shape-typing and improved dtypes

jorenham:
This follows #31154 and uses a similar overload pattern to improve the following functions and methods:

- `np.min`, `np.amin`, `np.nanmin`
- `np.max`, `np.amax`, `np.nanmax`
- `np.ndarray.min`
- `np.ndarray.max`
- `np.ptp`

This also fixed the incompatible overlapping overloads of `amax`, `amin`, and `ptp`, towards #27032.

#### AI Disclosure
N/A


jorenham:
Ah woops, I forgot to add overloads for comparable types that aren't array-likes. Pandas, for example, uses `np.max` on a list of `np.dtype` instances. 
Apparently any sequence with either a `__ge__` or `__le__` method should be accepted.

jorenham:
I think the pandas error in `pandas/core/dtypes/dtypes.py` is caused by that mypy overload bug, where it selects the wrong overload for `Any` input. I'll add a workaround for that.

The other errors (pandas and static-frame) look right to me. As in: they're caused by actual typing issues in pandas in static-frame.

charris:
Needs a rebase.

charris:
Thanks Joren.

PR #31164, opened 2026-04-06
TYP: ``ndarray.cum{prod,sum}`` shape-typing and improved dtypes

jorenham:
In #31120 I forgot to also shape-type-ify the corresponding `cumsum` and `cumprod` methods of `ndarray` (and `ma.MaskedArray`).

#### AI Disclosure
N/A


charris:
Thanks Joren.

PR #31166, opened 2026-04-06
TYP: ``arg{min,max}`` shape-typing

jorenham:
Dual to #31163, this improves the shape-typing of 

- `np.arg{min,max}`
- `np.ndarray.arg{min,max}`
- `np.ma.MaskedArray.arg{min,max}`

#### AI Disclosure
N/A


jorenham:
The primer errors for `pandas` are correct: `np.intp` and `builtins.int` are incompatible types, so this is a typing issue in pandas.

charris:
Thanks Joren.

PR #31169, opened 2026-04-06
TYP: ``partition`` shape-typing

jorenham:
Here's yet another shape-typing improvement. This time, it's for `np.partition`. It's pretty straightforward, because if `axis` is given (or the default) then the shape is preserved. Otherwise, if `axis=None`, then it always returns a 1d array.

#### AI Disclosure
N/A


charris:
Thanks Joren.

PR #31170, opened 2026-04-06
TYP: ``all`` and ``any`` shape-typing

jorenham:
``np.any``, ``np.all``, and the corresponding `ndarray` methods will now return the same input shape when `keepdims=True`, and will no longer return `Any` in case `axis` is not `None`. 

#### AI Disclosure
N/A


jorenham:
The new `unused-ignore` errors are good to see.

But there are also some other errors. I looked into them, and they don't seem to indicate a problem on our side.

- pandas: 
  - a call to `np.all` with `axis: tuple[int, ...]` is used as argument to `__and__`, but that `__and__` does not accept `NDArray[np.bool]`, even though that's what it could be at runtime 
  - the other error correct, because `np.bool` and `builtins.bool` are indeed incompatible types, so that's a typing issue in pandas.
- static-frame: all errors seem to be caused by incorrect callable annotations in static-frame itself.

charris:
Needs rebase.

charris:
Thanks Joren.

PR #31171, opened 2026-04-06
TYP: fix ``np.shape`` assignability issue for python lists

jorenham:
It was working fine when a list literal (e.g. `[1, 2]`) was passed, but some type-checkers would reject it when not passed directly, e.g. `a: list[int] = [1, 2]; np.shape(a)`. This has the do with `list` having an invariant generic type parameter. The solution is to use a type variable, so that the item type can "vary" in the function scope.

#### AI Disclosure
N/A


charris:
Needs rebase. I'm suspecting backports are getting more difficult at this point :)

jorenham:
Yea, especially because of the PEP 695 syntax stuff. I'll backport this one myself

charris:
Thanks Joren.

PR #31173, opened 2026-04-07
DEV: avoid byte-compiling every .py file on a `spin build` call.

rgommers:
This makes a `spin build` on an already-built numpy more than twice faster. Which also applies to `spin test`, and every other command that depends on the build step.

See https://github.com/numpy/numpy/pull/30510#issuecomment-4097010127 for details.

#### AI Disclosure

No AI tools used

### DIFF
--- a/.spin/cmds.py
+++ b/.spin/cmds.py
@@ -77,10 +77,14 @@ def changelog(token, revision_range):
     help="Build with pre-installed scipy-openblas32 or scipy-openblas64 wheel"
 )
 @spin.util.extend_command(spin.cmds.meson.build)
-def build(*, parent_callback, with_scipy_openblas, **kwargs):
+def build(*, parent_callback, meson_args, with_scipy_openblas, **kwargs):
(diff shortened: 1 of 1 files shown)


mattip:
LGTM. Could this become the default for meson in an upstream PR/issue?

mattip (approved):

rgommers:
> Could this become the default for meson in an upstream PR/issue?

I plan to open a `spin` PR as well to make this the default, after giving this a bit of time to make sure there's no unforeseen issues (I don't expect it). And then for `meson` we can't turn it off (that'd be backwards-incompatible and not suitable for all use cases), but we can speed it up as discussed in the conversation linked in the PR description.

seberg:
Let's give this a try then, thanks.

ngoldbaum:
I didn't notice this until I was doing some end-of-month reporting. Lo and behold - no more long pause for no-op `spin build` anymore. Thanks so much for fixing this!!!

PR #31175, opened 2026-04-07
MAINT: allow reading auxv output from recent glibc versions on s390x

AlekseiNikiforovIBM:
### PR summary
During review of https://github.com/numpy/numpy/pull/30819 it was mentioned (https://github.com/numpy/numpy/pull/30819#issuecomment-3937113785) that sticking to processing of output of `auxv` could be preferable due to qemu-user.

This change updates parsing of `auxv` output to accept new format of output values and provides necessary values for s390x.

It was tested on Fedora 43 s390x.

#### AI Disclosure
No AI tools used.

### DIFF
--- a/numpy/_core/tests/test_cpu_features.py
+++ b/numpy/_core/tests/test_cpu_features.py
@@ -69,6 +69,7 @@ class AbstractTest:
     features_groups = {}
     features_map = {}
     features_flags = set()
+    features_hwcap = dict()
 
     def load_flags(self):
         # a hook
@@ -116,9 +117,14 @@ class AbstractTest:
                 continue
             hwcap_value = [s.strip() for s in at.split(b':', 1)]
             if len(hwcap_value) == 2:
-                self.features_flags = self.features_flags.union(
-                    hwcap_value[1].upper().decode().split()
-                )
+                hwcap_decoded = hwcap_value[1].upper().decode().split()
+                if len(hwcap_decoded) == 1 and hwcap_decoded[0].startswith("0X"):
+                    hwcap_int = int(hwcap_decoded[0], 16)
+                    for k, v in self.features_hwcap.items():
+                        if (hwcap_int & k) != 0:
+                            self.features_flags.add(v)
+                else:
+                    self.features_flags = self.features_flags.union(hwcap_decoded)
 
 @pytest.mark.skipif(
     sys.platform == 'emscripten',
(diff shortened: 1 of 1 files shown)


AlekseiNikiforovIBM:
@rgommers @seiko2plus , please take a look at this PR.

charris:
I am inclined to put this in. @seiko2plus @rgommers comment?

numpy/_core/tests/test_cpu_features.py
```diff
@@ -116,9 +117,14 @@ def load_flags_auxv(self):
                 continue
             hwcap_value = [s.strip() for s in at.split(b':', 1)]
             if len(hwcap_value) == 2:
-                self.features_flags = self.features_flags.union(
-                    hwcap_value[1].upper().decode().split()
-                )
+                hwcap_decoded = hwcap_value[1].upper().decode().split()
+                if len(hwcap_decoded) == 1 and hwcap_decoded[0].startswith("0X"):
```
charris, line 127: This condition could use a comment to explain it, `startswith("0X")` seems a bit generic. I assume the length being 1 is also important.

EDIT: I also wonder if the base class is the best place for this.

AlekseiNikiforovIBM: I've added comment about this format. It is not s390x-specific format. Power and RISCV also parse it and could reuse this code. Actually, RISCV already does parse it. I probably can update RISCV. For Power I do not have correct flag values to check for.

Edit: other platforms don't parse output of `LD_SHOW_AUXV=1 /bin/true`, but if they ever do, they'll also benefit from this code.

AlekseiNikiforovIBM: I've updated RISCV to also reuse this code, but it needs testing.

AlekseiNikiforovIBM: Tests passed for RISCV, but they ran on `riscv64/ubuntu:22.04`. To actually test this change, something like `fedora:43` is needed.

rgommers:
> I am inclined to put this in. @seiko2plus @rgommers comment?

I'd defer to @seiko2plus here, he's the real expert. And for s390x patches, I'd lean towards liberally accepting them from real-world users and IBM maintainers. The RVV part seems small here, and the rest is s390x specific.

AlekseiNikiforovIBM:
This PR is still waiting to get merged or for feedback.

PR #31176, opened 2026-04-07
ENH: Removed remnants of Python 2 buffer protocol from NumPy internals

### LINKED ISSUE #30813: [ENH] Remove remnants of Python 2 buffer protocol from NumPy internals

Anushreebasics:
fixes #30813 

Hi NumPy team, I’m a first-time contributor. I use NumPy regularly for array-heavy Python work, mainly for data processing and numerical experimentation. While working with NumPy internals/C-API docs, I noticed issue #30813 around legacy Python 2 buffer-protocol language and wanted to help modernize and clarify that path. This contribution is my attempt to reduce confusion for future contributors and keep the API/docs aligned with current behavior.

#### AI Disclosure
no AI tools were used

### DIFF
--- a/numpy/_core/include/numpy/ndarraytypes.h
+++ b/numpy/_core/include/numpy/ndarraytypes.h
@@ -849,8 +849,6 @@ typedef struct tagPyArrayObject {
  * compatible with multiple NumPy versions.
  */
 
-/* Mirrors buffer object to ptr */
-
 typedef struct {
         PyObject_HEAD
         PyObject *base;
--- a/numpy/_core/src/multiarray/conversion_utils.c
+++ b/numpy/_core/src/multiarray/conversion_utils.c
@@ -280,11 +280,11 @@ PyArray_AsTypeCopyConverter(PyObject *obj, NPY_ASTYPECOPYMODE *copymode)
     return NPY_SUCCEED;
 }
 
-/*NUMPY_API
+/*
  * Get buffer chunk from object
  *
- * this function takes a Python object which exposes the (single-segment)
- * buffer interface and returns a pointer to the data segment
+ * this function takes a Python object which exposes the buffer interface
+ * and returns a pointer to the data segment
  *
  * You should increment the reference count by one of buf->base
  * if you will hang on to a reference
@@ -293,7 +293,7 @@ PyArray_AsTypeCopyConverter(PyObject *obj, NPY_ASTYPECOPYMODE *copymode)
  * memory...
  */
 NPY_NO_EXPORT int
-PyArray_BufferConverter(PyObject *obj, PyArray_Chunk *buf)
+PyArray_BufferConverterInternal(PyObject *obj, PyArray_Chunk *buf)
 {
     Py_buffer view;
 
@@ -333,6 +333,18 @@ PyArray_BufferConverter(PyObject *obj, PyArray_Chunk *buf)
         buf->base = obj;
     }
     return NPY_SUCCEED;
+}
+
+/*NUMPY_API
+ * Deprecated API wrapper around the internal buffer converter.
+ */
+NPY_NO_EXPORT int
+PyArray_BufferConverter(PyObject *obj, PyArray_Chunk *buf)
+{
+    if (DEPRECATE("PyArray_BufferConverter is deprecated and will be removed in a future version. Use the standard Python buffer protocol instead.") < 0) {
+        return NPY_FAIL;
+    }
+    return PyArray_BufferConverterInternal(obj, buf);
 }
 
 /*NUMPY_API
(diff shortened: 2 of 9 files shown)


Anushreebasics:
## Summary

It makes three main changes:

- Adds a [DeprecationWarning] when [PyArray_BufferConverter] is used.

- Moves internal ndarray construction to a private [_PyArray_Chunk] typedef so internal code no longer depends on the public alias.

- Updates the C-API reference docs for [PyArray_BufferConverter] and [PyArray_Chunk] to describe the modern buffer-protocol behavior instead of the old single-segment/Python 2 framing.

@ngoldbaum @jorenham please review

numpy/_core/include/numpy/ndarraytypes.h
```diff
@@ -849,15 +849,29 @@ typedef struct tagPyArrayObject {
...
+ */
+#if defined(NPY_INTERNAL_BUILD) && NPY_INTERNAL_BUILD
+typedef struct _PyArray_Chunk {
+        PyObject_HEAD
+        PyObject *base;
+        void *ptr;
+        npy_intp len;
+        int flags;
+} _PyArray_Chunk;
+#else
```
ngoldbaum: Why add this indirection if you're not going to deprecate using the struct or change it? I'd just keep it around and use the public version internally.

ngoldbaum: You could also figure out a way to signal to users that this struct is deprecated. In that case it makes sense to add the indirection.

numpy/_core/src/multiarray/conversion_utils.c
```diff
@@ -335,6 +335,18 @@ PyArray_BufferConverter(PyObject *obj, PyArray_Chunk *buf)
...
 }
 
+/*NUMPY_API
+ * Deprecated API wrapper around the internal buffer converter.
+ */
+NPY_NO_EXPORT int
+PyArray_BufferConverter(PyObject *obj, _PyArray_Chunk *buf)
+{
+    if (DEPRECATE("PyArray_BufferConverter is deprecated and will be removed in a future version. Use the standard Python buffer protocol instead.") < 0) {
+        return NPY_FAIL;
```
ngoldbaum, line 345: there should be a test for this deprecation

ngoldbaum (commented):
Left some comments. FWIW it's OK if you used AI for PRs like this, you just need to disclose it.

Anushreebasics:
@ngoldbaum please review the changes

ngoldbaum (approved):
Looks good! I marked it for the 2.5 release. Before this can be merged it needs a release note though. See `doc/release/upcoming_changes/README.rst` for instructions.

Anushreebasics:
@ngoldbaum I have added the release notes, please review the changes

ngoldbaum:
I think this is OK to merge. Maybe @seberg could take a look and hit the merge button though?

numpy/_core/src/multiarray/conversion_utils.c
seberg: Sorry for not looking earlier, if you are up for an iteration.

It would be good to add a `(Deprecated NumPy 2.5)` as well as a `/* Deprecated NumPy 2.5, 2026-04 */` style comment to this (breaking the long line would be nice also).

The deprecation test-case is nice. And I don't mind keeping it where it is, although it would be nice to cover the error path by making sure `DeprecationWarning` is raised an error.
(That is part of why we have a `test_deprecated`, moving the test there may be the nicest solution.)

Either way, all of these are indeed more nits and we could probably merge without as well.

ngoldbaum:
@Anushreebasics sorry for dropping this - it looks like there's a merge conflict now. Would you mind fixing that?

seberg:
I think I commented on a resolved convo, so it got lost.  But it would be good to test add `/* Deprecated NumPy 2.6, 2026-05 */` or so (and maybe also `(Deprecated NumPy 2.6)` in the message as well as breaking the long line.
It would also be nice for the test to exercise the error path, but it's simple enough.

Either way, both are smaller things if Nathan prefers to just go ahead that is fine with me, but if there is a conflict to resolve...

PR #31177, opened 2026-04-07
DEP: Undo deprecation for ``np.dtype()`` signature used by old pickles

### LINKED ISSUE #30786: BUG: Numpy 2.4.2 raises warning when unpickling
### Describe the issue:

Numpy shouldn’t warn when reading a pickle file it wrote.

```python
import pickle
import subprocess

url = "https://github.com/scverse/scanpy/raw/refs/heads/4c7c5c0e8bacc958370fcb0988495a6bf8410add/tests/_data/objs_t_test.pkl"
with urllib.request.urlopen(url) as con:
    true_scores_t_test, true_names_t_test = np.load(con, allow_pickle=True)
```

### Error message:

```shell
numpy.exceptions.VisibleDeprecationWarning: dtype(): align should be passed as Python or NumPy boolean but got `align=0`. Did you mean to pass a tuple to create a subarray type? (Deprecated NumPy 2.4)
```

2.4.2
3.14.0 (main, Oct 14 2025, 21:10:22) [Clang 20.1.4 ]

### Runtime Environment:

_No response_

_No response_

seberg:
Some (very) old pickles pass align=0, copy=1 as integers rather than bools.
This just assumes no user will really do that, and allows it.

It may be possible to also use `copyreg` to work around it, but since it is a rather specific signature this seems easier/just as well.

(No special tooling)

Closes gh-30786

### DIFF
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -2553,13 +2553,21 @@ arraydescr_new(PyTypeObject *subtype,
         /*
          * In the future, reject non Python (or NumPy) boolean, including integers to avoid any
          * possibility of thinking that an integer alignment makes sense here.
+         * We omit the case of `oalign == 0` and `ocopy == 1` if there are exact ints.
+         * This can fail, in which case res is -1 and we enter the deprecation path.
          */
+        int res = 0;
+        int overflow;
         if (!PyBool_Check(oalign) && !PyArray_IsScalar(oalign, Bool) && !(
                 // Some old pickles use 0, 1 exactly, assume no user passes it
                 // (It may also be possible to use `copyreg` instead.)
-                PyLong_CheckExact(oalign) &&  PyLong_AsLong(oalign) == 0 &&
-                ocopy != NULL && PyLong_CheckExact(ocopy) && PyLong_AsLong(ocopy) == 1)) {
+                PyLong_CheckExact(oalign) && (res = PyLong_IsZero(oalign)) == 1 &&
+                ocopy != NULL && PyLong_CheckExact(ocopy) &&
+                (res = PyLong_AsLongAndOverflow(ocopy, &overflow)) == 1)) {
             /* Deprecated 2025-07-01: NumPy 2.4 */
+            if (res == -1 && PyErr_Occurred()) {
+                return NULL;  // Should actually be impossible (as inputs are `long`)
+            }
             if (PyErr_WarnFormat(npy_static_pydata.VisibleDeprecationWarning, 1,
                         "dtype(): align should be passed as Python or NumPy boolean but got `align=%.100R`. "
                         "Did you mean to pass a tuple to create a subarray type? (Deprecated NumPy 2.4)",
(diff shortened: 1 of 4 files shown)


ngoldbaum:
This fixes #30786, right? Ping @flying-sheep.

seberg:
Yeah, I just didn't check if the long discussion on docs warranted leaving it open.

EDIT: but probably just as well to do.

numpy/_core/src/multiarray/descriptor.c
```diff
@@ -2537,20 +2538,27 @@ arraydescr_new(PyTypeObject *subtype,
...
         /*
          * In the future, reject non Python (or NumPy) boolean, including integers to avoid any
          * possibility of thinking that an integer alignment makes sense here.
          */
-        if (!PyBool_Check(oalign) && !PyArray_IsScalar(oalign, Bool)) {
+        if (!PyBool_Check(oalign) && !PyArray_IsScalar(oalign, Bool) && !(
+                // Some old pickles use 0, 1 exactly, assume no user passes it
+                // (It may also be possible to use `copyreg` instead.)
+                PyLong_CheckExact(oalign) &&  PyLong_AsLong(oalign) == 0 &&
+                ocopy != NULL && PyLong_CheckExact(ocopy) && PyLong_AsLong(ocopy) == 1)) {
```
ngoldbaum: `PyLong_AsLong` can fail and set `OverflowError`: https://docs.python.org/3/c-api/long.html#c.PyLong_AsLong. Ditto for all the other `PyLong_As` functions. It also looks like `PyLong_AsLong` is soft-deprecated - maybe use `PyLong_AsInt` instead?

ngoldbaum (commented):
Storing old test data in a pickle file seems like a reasonable use-case so I'm happy to un-deprecate this. I spotted some incorrect C API uses below (with my own eyes, no Claude this time 😜).

It also maybe needs a release note? Kinda on the bubble. It is changing how the deprecations works a little though so that seems worth mentioning.

numpy/_core/src/multiarray/descriptor.c
seberg: Ack... yeah, annoying to add error handling here, but clearly needed...

PyLong_AsLong isn't soft depreated of course, `PyLong_AS_LONG` is.

seberg: Well, I pushed a pretty annoying fix (annoying, because of how the control flow is).

I don't think this can ever fail, so I just made it a critical error, but I don't like baking this in when it isn't guaranteed in the C-API docs.

(The logic is really strange, because if it fails `res == -1` the if branch _is_ taken, tested locally by removing the `PyLong_CheckExact` which makes it possible to fail them)

seberg:
I dunno, I was about to amend the 2.4.0 release notes to add a note that this corner case affecting pickles was solved in 2.4.5.  But it feels a bit weird also?

OTOH, I don't think a release note in 2.5 seems helpful. This eases a pain in 2.4-2.4.4 (2.4.5 should get this as a backport, I think), so the only interesting point would be where you actually run into the pickle loading failure...

I'll add another comment to the original PR, I think that is the most likely place for someone actually to find the info...

ngoldbaum (approved):

gdementen:
@seberg Thanks a lot for your work

charris:
Thanks Sebastian.

PR #31184, opened 2026-04-07
BUG: Use dict to store ufunc loops (fixes thread-safety)

### LINKED ISSUE #31112: BUG: flaky test under pytest-run-parallel
```
==================================== ERRORS ====================================
___________ ERROR at call of test_real_imag_ufunc_minimal[imag-imag] ___________

ufunc = <ufunc 'imag'>, attr = 'imag'

    @pytest.mark.parametrize("ufunc,attr", [
        (np._core.umath.real, "real"), (np._core.umath.imag, "imag")])
    def test_real_imag_ufunc_minimal(ufunc, attr):
        with pytest.raises(TypeError):
            ufunc(np.array([1, 2, 3]))  # non-complex or object raises
    
        arr = np.array([1 + 2j, 3 + 4j])
>       res = ufunc(arr)
              ^^^^^^^^^^
E       RecursionError: Stack overflow (used 16364 kB) in __subclasscheck__

arr        = array([1.+2.j, 3.+4.j])
attr       = 'imag'
ufunc      = <ufunc 'imag'>
```

As seen here: https://github.com/numpy/numpy/actions/runs/23796435549/job/69344855092?pr=31108#step:9:47471

Almost certainly fallout from merging #30984. I'm going to try to reproduce this locally and understand what's happening.

Ping @seberg @kumaraditya303

seberg:
Change `_loops` to use a dict instead to be able to use `setdefault` for adding new loops.

A dict is also ordered, although we need to use `.items()` to iterate it, otherwise things get nicer over-all (although the key is repeated for now, to simplify passing around `info`).

Closes gh-31112

### DIFF
--- a/numpy/_core/include/numpy/ufuncobject.h
+++ b/numpy/_core/include/numpy/ufuncobject.h
@@ -232,6 +232,10 @@ typedef struct _tagPyUFuncObject {
          */
         PyUFunc_ProcessCoreDimsFunc *process_core_dims_func;
     #endif
+    #if NPY_INTERNAL_BUILD && Py_GIL_DISABLED
+        /* Private mutex, currently for thread-safety while adding loops */
+        PyMutex _mutex;
+    #endif
 } PyUFuncObject_fields;
 
 typedef PyUFuncObject_fields PyUFuncObject;
--- a/numpy/_core/include/numpy/ufuncobject.h
+++ b/numpy/_core/include/numpy/ufuncobject.h
@@ -223,7 +223,7 @@ typedef struct _tagPyUFuncObject {
     #if NPY_FEATURE_VERSION >= NPY_1_22_API_VERSION
         /* New private fields related to dispatching */
         void *_dispatch_cache;
-        /* A PyListObject of `(tuple of DTypes, ArrayMethod/Promoter)` */
+        /* Ordered dict `tuple of DTypes -> (tuple of DTypes, ArrayMethod/Promoter)` */
         PyObject *_loops;
     #endif
     #if NPY_FEATURE_VERSION >= NPY_2_1_API_VERSION
@@ -231,10 +231,6 @@ typedef struct _tagPyUFuncObject {
          * Optional function to process core dimensions of a gufunc.
          */
         PyUFunc_ProcessCoreDimsFunc *process_core_dims_func;
-    #endif
-    #if NPY_INTERNAL_BUILD && Py_GIL_DISABLED
-        /* Private mutex, currently for thread-safety while adding loops */
-        PyMutex _mutex;
     #endif
 } PyUFuncObject_fields;
 
--- a/numpy/_core/src/umath/dispatching.cpp
+++ b/numpy/_core/src/umath/dispatching.cpp
@@ -114,39 +114,50 @@ PyUFunc_AddLoop(PyUFuncObject *ufunc, PyObject *info, int ignore_duplicate)
         return -1;
     }
 
+    Py_ssize_t length = -1;
+    int ret = -1;
+#ifdef Py_GIL_DISABLED
+    PyMutex_Lock(&ufunc->_mutex);
+#endif
     if (ufunc->_loops == NULL) {
         ufunc->_loops = PyList_New(0);
         if (ufunc->_loops == NULL) {
-            return -1;
+            goto finish;
         }
     }
 
-    PyObject *loops = ufunc->_loops;
-    Py_ssize_t length = PyList_Size(loops);
+    length = PyList_Size(ufunc->_loops);
     for (Py_ssize_t i = 0; i < length; i++) {
-        PyObject *item = PyList_GetItemRef(loops, i);
+        PyObject *item = PyList_GetItemRef(ufunc->_loops, i);
         PyObject *cur_DType_tuple = PyTuple_GetItem(item, 0);
         Py_DECREF(item);
         int cmp = PyObject_RichCompareBool(cur_DType_tuple, DType_tuple, Py_EQ);
         if (cmp < 0) {
-            return -1;
+            goto finish;
         }
         if (cmp == 0) {
             continue;
         }
         if (ignore_duplicate) {
-            return 0;
+            ret = 0;
+            goto finish;
         }
         PyErr_Format(PyExc_TypeError,
                 "A loop/promoter has already been registered with '%s' for %R",
                 ufunc_get_name_cstr(ufunc), DType_tuple);
-        return -1;
+        goto finish;
     }
 
-    if (PyList_Append(loops, info) < 0) {
-        return -1;
+    if (PyList_Append(ufunc->_loops, info) < 0) {
+        goto finish;
     }
-    return 0;
+    ret = 0;
+
+    finish:
+#ifdef Py_GIL_DISABLED
+    PyMutex_Unlock(&ufunc->_mutex);
+#endif
+    return ret;
 }
 
 
--- a/numpy/_core/src/umath/dispatching.cpp
+++ b/numpy/_core/src/umath/dispatching.cpp
@@ -6,8 +6,8 @@
  *
  * - operand_DTypes:  The datatypes as passed in by the user.
  * - signature: The DTypes fixed by the user with `dtype=` or `signature=`.
- * - ufunc._loops: A list of all ArrayMethods and promoters, it contains
- *   tuples `(dtypes, ArrayMethod)` or `(dtypes, promoter)`.
+ * - ufunc._loops: Ordered dict of all ArrayMethods and promoters, mapping
+ *   `dtypes` to tuples `(dtypes, ArrayMethod)` or `(dtypes, promoter)`.
  * - ufunc._dispatch_cache: A cache to store previous promotion and/or
  *   dispatching results.
  * - The actual arrays are used to support the old code paths where necessary.
@@ -70,8 +70,8 @@ promote_and_get_info_and_ufuncimpl(PyUFuncObject *ufunc,
 
 
 /**
- * Function to add a new loop to the ufunc.  This mainly appends it to the
- * list (as it currently is just a list).
+ * Function to add a new loop to the ufunc.  This adds it to the
+ * _loops dict keyed by the DType tuple.
  *
  * @param ufunc The universal function to add the loop to.
  * @param info The tuple (dtype_tuple, ArrayMethod/promoter).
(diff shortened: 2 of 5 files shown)


seberg:
While a mutex/critical section should be OK here (although I forgot to replace the `PySequence_Fast_GET_ITEM` with a `PyList_GetItemRef`), replacing it with a dict seems a bit nicer (as it has the builtin `setdefault` which is nicely atomic and even informs if things were added).

The downside is that the best solution for the actual promotion is probably just to call `tuple(dict)`, but since normally we hit the cache that should be fine probably.

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -332,7 +317,13 @@ resolve_implementation_info(PyUFuncObject *ufunc,
         PyObject **out_info)
 {
     int nin = ufunc->nin, nargs = ufunc->nargs;
-    Py_ssize_t size = PySequence_Length(ufunc->_loops);
+    int ret = -1;
+    /* PyDict_Values returns a snapshot, safe against concurrent additions. */
```
seberg, line 314: Yeah, I used cursor/claude (although had to tell it to modify almost everything to my liking... just because you set out to make something thread-safe, doesn't mean it quite remembered that part, hehe), had to tell it off to reduce it's comment load, but this seems pretty OK ;).

One note: I decided to keep the `tuple -> (tuple, meth/promoter)` for now.  Just like that one `// borrow-ref OK` comment this is currently not designed to allow deletion and it would just add churn here...
(I am not even sure if we want to ever delete things, rather than just put it somewhere else... even if we allow replacing, it shouldn't happen often and immortalizing may be pragmatic.)

~EDIT: As a note, I suspect all is good. But I'll go once more myself tomorrow morning to see if there aren't ref-counting issues or so.~ (done)

mhvk: So, the value of the dict effectively contains a tuple `(key, real-value)`? Or can the tuple of dtypes possibly be different? If not, then my sense would be to remove the key from the value, to help the logic: a `dict` keyed by the dtypes that returns a promotor or method seems much clearer conceptually.

That said, I think it makes sense to not make that change here if it is more intrusive, to keep this PR focussed on thread safety.

mhvk (commented):
Overall, this looks nice! Indeed, the `dict` seems a more logical solution to the problem.

But it does seem to me that, perhaps in follow-up, we should move to have a `dict` that just has the promoter or array method as its value.

numpy/_core/src/umath/dispatching.cpp
seberg: Right, the thing is that `info` currently ends up in the cache (where the `key` _isn't_ identical anymore) as borrowed.

So it looks awkward, but changing it seems really awkward too.

The practical refactor might be different: Right now there is an awkward distinction between "BoundArrayMethod" and the "ArrayMethod", if we remove it the need for the tuple should just go away (of course the DTypes will still be duplicated, but it shouldn't feel as awkward ;)).

(The backstory is to accept that DTypes are immortal -- with a plausible exception for some future dynamically created subclasses of a superclass DType which would have no loops registered for the subclass.  At the time that felt, and was pointed out, like we should keep it plausible to make DTypes mortal in practice and that requires the split. But it was never practical -- Python doesn't even have the necessary technology to do it nicely/correctly!)

mhvk: OK, that makes sense -- and getting rid of bound vs unbound (a difference I don't fully understand...) sounds like a plan!

seberg: Too much information, so feel free to ignore :).

> and getting rid of bound vs unbound (a difference I don't fully understand...)

Like a bound method:  A bound method keeps `self` alive, a `BoundArrayMethod` keeps the DTypes (i.e. the multiple "self") alive.

With that distinction it was _plausible_ to at least clean up many/most cases (still annoying as the `key` needs to be something like a weakref).

Without... Maybe it is still plausible if maybe more awkward; but _all_ cases would require technology Python doesn't have (i.e. [ephemerons](https://en.wikipedia.org/wiki/Ephemeron)).

mhvk: In a variation, too much ignorance, so feel free to ignore :)

But why do array methods need to keep dtypes alive at all? Isn't it only the other way around, an otherwise unused array method should just disappear if all dtypes that use it have disappeared?  Or, trying not to be completely ignorant, is it because the array method has internal links to the dtypes it works for? (I really should re-read the relevant NEPs and documentation and try to make improve/summarize myself, so people like me who have an incomplete idea can understand how it all works...)

seberg: Since I started to try to explain it, I'll continue. But to be clear: The only thing that matters is if there is a real reason to think that cleanup matters...
I am not 100% sure, but to me it seems rather implausible (it still works for dynamically created DTypes if you only register their loops via a superclass DTypes -- although at least casts don't work for that pattern right now).

I doubt the NEPs help this is just unwieldy to think about but a pretty small detail in a sense...

---

We need the DTypes when executing any `ArrayMethod` at some point (we also pass it in to the `resolve_descriptors`, etc. but we need it for more things before also).
So at that time you need them guaranteed alive in _some_ form, and those DTypes *aren't* the ones passed in by the user:
1. Because promotion might have happened.
2. If output DTypes differ from inputs, then those need to be kept somewhere and at least in principle such an `ArrayMethod` (or ufunc loop) should keep the output DType alive as long as it is callable.

Now, I think at the time I mostly said that 2 is the thing that just can't work (except maybe for casts).

The thing you can possibly make work, is if the `ArrayMethod`s can become "invalid". I don't think I really thought about it at the time (as it breaks 2).
Having a `BoundArrayMethod` solve the need for having to "invalidate" an `ArrayMethod` (whether we can be sure to clean it up or not), while allowing us to have an object that represents the full thing.

Now, the only truly correct path is if the ufunc loop (effectively) keeps it's DTypes alive but is cleaned up when it becomes inaccessible.  If the `ArrayMethod` doesn't own the DTypes this can work for many cases (e.g. if all input and output DTypes are identical and maybe casts -- if all are the same, a `weakref` can be used).

But for all other cases you need an Ephemeron to spell something like: As long as the input DType is alive, this `ArrayMethod`/`key` is alive and keeps all the (output) DTypes alive.
(If Python _had_ an ephemeron aware GC, the cyclic GC could then figure out if a DType becomes truly inaccessible.)

But as you can (hopefully) see, you do need a split if you want cleanup in the generic case.  Right now it is a (barely used/existing) `BoundArrayMethod` vs. `ArrayMethod`.  If the `ArrayMethod` knew it's DTypes it but didn't keep it alive, then it would mean that it needs to have an "invalid" state.
And technically, if you expose it to Python you still kinda need the `BoundArrayMethod` as at least in that case it _should_ guarantee staying valid!

(I suppose I could also store the `BoundArrayMethod` as the info for now, but it is also a step in the "we'll never implement cleanup anyway" direction ;).)

---

Anyway, this is too much and I dunno if it can be understood :). In the end, the question is really if there actually is some reason to care about plausible generic cleanup...  My gut feeling is "no" as it is just really impractical, but _if_ there is a sensible reason, then keeping the current split may be better to not move in a direction that would make it even harder.

mhvk: Thanks, yes, the output `dtype` is a tricky one. Conceptually, I think it doesn't strictly need to be on the array method:
- For dtype casts, the casts themselves could be a dict of methods keyed by output dtype, so that the cast machinery rather than the array method becomes responsible for keeping the references.
- For ufuncs, I guess the method would need to be the value in a dict keyed by both input and output dtype, and anything that is incomplete has to be handled by a promoter.

But, really, there is a lot to be said for a arraymethod specification to be self-describing, which does mean it is logical to have references to the dtypes.

And I think you are right that throw-away dtypes that also have their own loops are an unlikely prospect, so in practice there really should not be a problem.

numpy/_core/src/umath/wrapping_array_method.c
```diff
@@ -249,28 +249,21 @@ PyUFunc_AddWrappingLoop(PyObject *ufunc_obj,
...
-        }
-        wrapped_meth = (PyArrayMethodObject *)PyTuple_GET_ITEM(item, 1);
-        if (!PyObject_TypeCheck(wrapped_meth, &PyArrayMethod_Type)) {
+    PyObject *existing_info;
+    if (PyDict_GetItemRef(ufunc->_loops, wrapped_dt_tuple, &existing_info) < 0) {
+        goto finish;
+    }
+    if (existing_info != NULL) {
+        PyObject *existing_meth = PyTuple_GET_ITEM(existing_info, 1);
+        Py_DECREF(existing_info);
```
ngoldbaum: I think it's possible this decref could cause the last reference to `existing_meth` to go away. Since you have a borrowed reference, that means existing_meth is possibly invalid after here. There might be larger architectural reasons why that's not possible, so a comment explaining might be nice. You could also move the `existing_inf` decref to line 272 or below, to ensure `existing_meth` stays alive until its last use.

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -5249,18 +5243,13 @@ PyUFunc_RegisterLoopForType(PyUFuncObject *ufunc,
...
-            continue;
-        }
-        PyObject *registered = PyTuple_GET_ITEM(item, 1);
+    PyObject *existing_item;
+    if (PyDict_GetItemRef(ufunc->_loops, signature_tuple, &existing_item) < 0) {
+        goto fail;
+    }
+    if (existing_item != NULL) {
+        PyObject *registered = PyTuple_GET_ITEM(existing_item, 1);
+        Py_DECREF(existing_item);
```
ngoldbaum, line 5256: same pattern here, either leave a comment explaining why this is safe or move the decref below the last use of `registered`.

ngoldbaum (commented):
Spotted a couple minor issues (with the help of Claude). Also ping @kumaraditya303 to give this a once-over.

Also can you update the PR description so this gets merged with a more relevant commit message?

ngoldbaum:
I tried to trigger the crash on my Mac for about 5 minutes running the command in https://github.com/numpy/numpy/issues/31112#issuecomment-4170815158 and didn't see any failures. Hard to prove a negative but I think we can conclude this avoids the race that causes the crash. `main` crashes about about 30 seconds of looping on that test on the same Mac.

numpy/_core/src/umath/ufunc_object.c
seberg, line 5256:
```suggestion
        Py_DECREF(existing_item);  // currently borrowed from _loops
```
Right, the old code just used borrowed refs, I like `GetItemRef` (also due to the code pattern), but that makes this looks even more dangerous :).

numpy/_core/include/numpy/ufuncobject.h
```diff
@@ -223,7 +223,7 @@ typedef struct _tagPyUFuncObject {
     #if NPY_FEATURE_VERSION >= NPY_1_22_API_VERSION
         /* New private fields related to dispatching */
         void *_dispatch_cache;
-        /* A PyListObject of `(tuple of DTypes, ArrayMethod/Promoter)` */
+        /* Ordered dict `tuple of DTypes -> (tuple of DTypes, ArrayMethod/Promoter)` */
```
kumaraditya303, line 226: I find it little confusing because there is a separate collections.OrderdDict, dicts are insertion ordered by default so maybe just remove the "ordered"?

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -115,37 +115,22 @@ PyUFunc_AddLoop(PyUFuncObject *ufunc, PyObject *info, int ignore_duplicate)
     }
 
     if (ufunc->_loops == NULL) {
-        ufunc->_loops = PyList_New(0);
+        ufunc->_loops = PyDict_New();
```
kumaraditya303: This can still data race, can we just eagerly create this dict where ufunc object is allocated?

numpy/_core/src/umath/wrapping_array_method.c
kumaraditya303: The dict holds a strong ref to `existing_info` so this is safe, but it would still be nice to move the decref.

numpy/_core/src/umath/ufunc_object.c
seberg: Yeah, nicer to reorganize, claude is right, that claude didn't do this perfectly ;) (even if fine as the old code also relied on borrowing being OK)

numpy/_core/src/umath/dispatching.cpp
seberg: Hmmm, it already _is_ in the normal path and now I am unsure if this is better to keep around for some crazy reason...
(at least for backporting)

seberg: Whatever, let's just delete it and not backport the PR, it just fixes free-threading stuff that nobody hit in practice.

numpy/_core/include/numpy/ufuncobject.h
seberg: True, but I want to remind whoever is about to change it that the order may matter.

numpy/_core/src/umath/ufunc_object.c
ngoldbaum: I think there's still an issue - the `DECREF` has to be after the last use of `registered` on line 5259 below.

ngoldbaum (approved):
One last comment, I think there's still a refcounting issue in one of the code paths I commented on. Also the PR description still needs a rewrite before merging. Approving though since these are just touchup-level changes.

kumaraditya303 (approved):
LGTM, thanks for fixing this

ngoldbaum:
Thanks Sebastian!

PR #31195, opened 2026-04-10
MAINT: rename parameters in ``tensordot`` function

PelicanQ:
### PR summary

The function signature of `tensordot` called the two inputs x1 and x2 but docs call them a and b.

You can see this here
(https://numpy.org/doc/stable/reference/generated/numpy.linalg.tensordot.html#numpy.linalg.tensordot)
#### AI Disclosure

No AI tools used

### DIFF
--- a/numpy/linalg/_linalg.py
+++ b/numpy/linalg/_linalg.py
@@ -3400,13 +3400,13 @@ def matmul(x1, x2, /):
 
 # tensordot
 
-def _tensordot_dispatcher(x1, x2, /, *, axes=None):
-    return (x1, x2)
+def _tensordot_dispatcher(a, b, /, *, axes=None):
+    return (a, b)
 
 
 @array_function_dispatch(_tensordot_dispatcher)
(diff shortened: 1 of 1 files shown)


jorenham:
They're positional only so their names don't matter, but I guess it was indeed a bit weird.

BTW, you ignored the PR template, and therefore also the AI disclosure, which we require.

PelicanQ:
This is what I meant

image
image

jorenham:
Ah of course, I forgot that those also show up in the docs.

---

> BTW, you ignored the PR template, and therefore also the AI disclosure, which we require.

The PR template you ignored contains the following comment:

> If AI was used in the preparation of this pull request, please disclose
> the tool(s) used, how they were used, and specify what code or text is AI generated.
> If no AI tools were used, please write "No AI tools used" in this section. Read our
> policy on AI generated code at
> https://numpy.org/devdocs/dev/ai_policy.html.
> 
> In particular, all interaction is to be done by humans, including submission of PRs.

So please do.

PR #31196, opened 2026-04-10
BUG: fix ptp nan in masked arrays

### LINKED ISSUE #27674: BUG:  numpy 2 ptp returns nan when applied to masked arrays
### Describe the issue:

Before numpy 2 np.ptp(masked_array) worked as expected and where the masked values are ignored. Starting from numpy 2 the ptp function now returns nan instead.

```python
import numpy as np
arr = np.ma.masked_invalid([np.nan, 5, 10, 5, 10])
np.ptp(arr)
masked_array(data=nan,
             mask=False,
       fill_value=1e+20)
# expected the same as
arr.ptp()
np.float64(5.0)
```

### Error message:

_No response_

numpy 2.1.2
Python 3.13.0

### Runtime Environment:

_No response_

### Context for the issue:

The issue is not a problem b/c one can still use `np.ma.ptp`. However, it seems inconsistent with all other functions that when applied to a masked array, will ignore masked values.

Anarion-zuo:
### PR summary

Resolves #27674 .

This PR fixes a regression in `np.ptp` for non-`ndarray` inputs, specifically masked arrays. In NumPy 2, `np.ptp(masked_array)` could return `nan` instead of matching `MaskedArray.ptp()` and ignoring masked invalid values.

The fix updates `np.ptp` to preserve subclass-aware behavior for non-plain `ndarray` inputs by computing the result through `np.max` and `np.min`, while keeping the existing fast path for plain `ndarray` inputs. This restores the expected masked-array behavior without changing the standard ndarray implementation.

The PR also adds regression coverage for masked arrays, including the reported `masked_invalid([np.nan, 5, 10, 5, 10])` case. While validating the change, matrix-specific `ptp` behavior needed an adjustment as well, so `matrix.ptp()` is now computed directly from `ndarray.max`/`ndarray.min` to preserve its existing return-shape semantics.

Validation:
- Focused tests:
  - `spin test numpy/ma/tests/test_core.py numpy/matrixlib/tests/test_defmatrix.py -- -q -k "test_ptp or test_ptp_masked_invalid_matches_masked_array_method"`
  - Result: `3 passed`
- Full test run:

The full suite passed except for the pre-existing environment-specific failure:
- `numpy/_core/tests/test_cpu_features.py::Test_X86_Features::test_features`

### DIFF
--- a/numpy/_core/fromnumeric.py
+++ b/numpy/_core/fromnumeric.py
@@ -3000,6 +3000,14 @@ def ptp(a, axis=None, out=None, keepdims=np._NoValue):
     kwargs = {}
     if keepdims is not np._NoValue:
         kwargs['keepdims'] = keepdims
+
+    if type(a) is not mu.ndarray:
+        return um.subtract(
+            max(a, axis=axis, **kwargs),
+            min(a, axis=axis, **kwargs),
+            out=out,
+        )
+
     return _methods._ptp(a, axis=axis, out=out, **kwargs)
 
 
--- a/numpy/ma/tests/test_core.py
+++ b/numpy/ma/tests/test_core.py
@@ -4144,14 +4144,21 @@ class TestMaskedArrayMathMethods:
         _, X, _, m, mx, mX, _, _, _, _ = self._create_data()
         (n, m) = X.shape
         assert_equal(mx.ptp(), np.ptp(mx.compressed()))
+        assert_equal(np.ptp(mx), mx.ptp())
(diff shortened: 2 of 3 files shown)


Anarion-zuo:
@charris All CI passed.

jorenham:
You ignored the PR template and did not disclose that you used AI. The PR description is clearly AI written. So following our [AI policy](https://github.com/numpy/numpy/blob/main/doc/source/dev/ai_policy.rst), I'm going to close this.

PR #31213, opened 2026-04-11
BUG: raise warning when `np.timedelta64()` is called

### LINKED ISSUE #31103: BUG: `np.timedelta64()` and `np.datetime64()` does not raise any warning
### Describe the issue:

`generic` unit in `timedelta64` dtype is deprecated in the PR gh-29619, but `np.timedelta64()` do not raise `DeprecationWarning` when called.
This is same for `np.datetime64()`, which returns `NaT` without unit.

Current behavior:

```python
np.timedelta64() # no warning, returns 0 timedelta with generic unit
np.datetime64() # no warning, returns NaT without unit
```

It might be good to set some specific unit such as "ns" as default unit.

I'd like to mention @seberg 's comment here for reference.

* https://github.com/numpy/numpy/pull/29619#issuecomment-4150436618

```
Rethinking it, there some subtleties. I.e. np.array([np.timedelta64(3, "ns")], dtype="m") needs to keep working the way it currently does.
That is in principle not so hard, but in practice the path it currently takes is still np.dtype("m") -> unitless singleton -> type(uniteless singleton). But if np.dtype("m") there gives a warning, this conversion needs to happen later.
```

* https://github.com/numpy/numpy/pull/29619#issuecomment-4150990234

```
One thing I am not sure about is, if you were to change the default to ns for np.dtype("m") then some things might change behavior without a warning.
```

```python
import numpy as np

np.timedelta64() # no warning, returns 0 timedelta with generic unit
np.datetime64() # no warning, returns NaT without unit
```

### Error message:

```shell

```

2.5.0.dev0+git20260329.8568ed8

### Runtime Environment:

_No response_

This issue is reported and based on the discussion in gh-26919.

riku-sakamoto:
### PR summary

Closes #31103.

* Raise a DeprecationWarning when following codes are called without specific unit
  * `np.timedelta()`
  * `np.datetime()`
  * `np.datetime("NaT")`

#### AI Disclosure

No AI tools are used.


numpy/_core/tests/test_deprecations.py
```diff
@@ -408,6 +408,9 @@ def test_raise_warning_for_operation_with_generic_unit(
     ):
         self.assert_deprecated(op, args=(value, generic_value))
 
+    def test_raise_warning_for_default_constructor(self):
+        self.assert_deprecated(lambda: np.timedelta64())
```
seberg, line 412:
```suggestion
        self.assert_deprecated(lambda: np.timedelta64())
        self.assert_deprecated(lambda: np.datetime64())
```

seberg (approved):
Thanks, let's hope that this doesn't cause pain due to the default construction, but I'll guess it shouldn't be too bad.

riku-sakamoto:
@seberg 

Thank you for the review!
Would you prefer to include changes about raising a warning about `np.datetime64()` ?
If so, I'll add it and update the PR. (maybe need a release note as well?)

seberg:
Ah, thanks, sorry I missed that you didn't warn it because it was NaT and it is a bit less important.  Since I think we warn almost everywhere else, maybe we might as well try it?

I would think the old release note sufficiently captures things, but either way is fine.

riku-sakamoto:
@seberg 
Thank you as always!
I've updated to raise a warning when `np.datetime64()` or `np.timedelta64("NaT")` is called with no unit.

seberg (approved):
Thanks @riku-sakamoto, let's try this.

If this makes the deprecation worse for downstream, at least around `NaT` scalar creation where generic units make a bit more sense let us know.
(But since we didn't make an exception for 0 timedelta right now either, I think, this does seem reasonable to me.)

lucascolley:
@seberg I think the example of `np.datetime64('nat')` at https://numpy.org/devdocs/reference/arrays.datetime.html#basic-datetimes needs to be updated too.

PR #31217, opened 2026-04-11
TYP: ``pad`` shape-typing

jorenham:
The input `ndim` for `np.pad` are the same as the output `ndim`, so this is a pretty straightforward improvement.

Fun fact: "pad" in dutch means "toad".

#### AI Disclosure
N/A


charris:
Thanks Joren.

PR #31218, opened 2026-04-11
TYP: ``ix_`` shape-typing and fix for boolean input

jorenham:
Calling `np.ix_` with N arrays returns an N-tuple of N-d arrays with same dtype as input, **except in case of `bool`, because that is upcasted to `int_`**. This exception is what wasn't correct before, because `bool` input were annotated to return `bool` output.

Since this boolean behavior is also mentioned in the docs, I'm thinking of partially backporting this.

#### AI Disclosure
N/A


charris:
Thanks Joren.

PR #31219, opened 2026-04-11
TYP: ``diag_indices[_from]`` shape-typing

jorenham:
The output tuple type size now depends on the number of input dimensions, and the individually returned index arrays are now typed as 1-d instead of N-d.

#### AI Disclosure
N/A


charris:
Need rebase.

charris:
Thanks Joren.

PR #31220, opened 2026-04-11
TYP: ``expand_dims`` shape-typing

jorenham:
This adds shape typing overloads for `np.expand_dims` for `{0,1,2}-d` input arrays with axis as either an int, empty tuple, 1-tuple, or 2-tuple.

#### AI Disclosure
N/A


charris:
Thanks Joren.

Issue #24897, opened 2023-10-10
ENH: No longer auto-convert array scalars to numpy scalars in ufuncs (and elsewhere)


mhvk:
### Proposed new feature or change:

This may come too late, but numpy 2.0 may be a good opportunity to remove an irritating wart, that `ufunc(array_scalar_arguments)` returns a numpy scalar rather than an array scalar. There is a very long-standing PR (#14489) that introduces an option to avoid this (using `out=...`), but perhaps it is time to just ditch the conversion to scalar? 

See #13015 for discussion. As a specific example of why it is unhandy, it mentions:
> #13100 raises a case where `np.fix` resorts to calling `np.asanyarray(np.ceil(x, out=out))` in order to ensure that the result is an array. Unfortunately, this has a number of draw-backs:
> 
> * It discards duck-types, like `dask` arrays
> * It changes the dtype of 0d object arrays containing array-likes

The [first comment](https://github.com/numpy/numpy/issues/13105#issuecomment-471382946) on the issue by @rgommers suggests just removing the cast to scalar. To me, this still makes the most sense.

seberg:
**Reposting from gh-14489, not everything aged perfectly:**

An alternative thing that would be a big change, but not as big as removing scalars one, would be this: https://github.com/seberg/numpy/pull/new/try-0d-preservation that I tried to explore a bit today.

The proposal would be to:
* Ensure that 0-d arrays are preserved in ufuncs, `np.add(0d_array, 2)` will return a 0d array.  (And all other functions, this only has ufuncs right now!)
* Reductions would return scalars if `axis=None`.
* We may or may not need a (internal) helper `arr, wrap = asarray_and_wrap(obj)` that returns a function to do the "decaying".

(The tail of that is not super trivial, quite a few test failures, scipy also relies on scalars in a few places for example, but it looks harmless).  But run that branch with:
```
NPY_NUMPY_2_BEHAVIOR=1
```
to try.  A fun little caveat: `np.matmul(vector, vector)` should be a scalar so is a bit hackish still using the old behavior due to that.  

---

I never truly believed that we can pull of *removing* scalars or always converting them to arrays in ufuncs.  But maybe I am just too far on the side of "change as little as possible".

The alternative is what I say above.  That means making basically scalars something like NumPy duck-types with an implicit `__array_wrap__` that unwraps them, but they have a lower array-priority (one way to make a lot of our Python code be able to retain scalars when they should).

[...]

PR #31221, opened 2026-04-11
ENH: add optional flag to preserve 0D arrays as ndarray objects

Breno-Favaretto:
## Summary

This PR introduces an opt-in mechanism to control 0-dimensional NumPy array
return behavior in `PyArray_Return`.

A new global flag (`NPY_KEEP_0D`) is added to control whether 0D arrays are
converted to Python scalars or preserved as `ndarray` objects.

By default, NumPy behavior is unchanged: 0D arrays are still converted to
scalars. When enabled, 0D arrays are preserved as `ndarray`.

A Python-level context manager `numpy.keep_0D` is exposed to allow scoped
control of this behavior.

## API additions

- `numpy.keep_0D` context manager
- internal flag `NPY_KEEP_0D`
- C API setter `PyArray_SetKeep0D`

## Behavior change

No default behavior changes are introduced. This is a fully opt-in feature.

## Motivation

In some workflows, especially when working with higher-level array logic,
preserving 0D arrays as `ndarray` objects is desirable to avoid implicit
scalar conversion.

## Notes

This introduces a global state flag affecting array return behavior in
`PyArray_Return`. Care should be taken when using this in multithreaded
contexts.

---
## Issues

https://github.com/scientific-python/summit-2025/issues/38
https://github.com/numpy/numpy/issues/31145
---

## AI Disclosure

AI tools were used to assist in drafting and structuring code comments and
documentation. All implementation and final decisions were made by the author.


jorenham:
New features proposals should go through the mailing list first.

Breno-Favaretto:
Thanks for the clarification. Where would be the best place to start the discussion for this feature proposal?

rkern:
Please don't merge `main` until there's a need to. This is probably going to be a substantial discussion before this PR goes anywhere, and we don't want to keep spending CI cycles when nothing has actually changed.

Breno-Favaretto:
Thanks for the heads up, I won’t do that again.

seberg:
I am going to close this, as you have linked, you seem to be aware of the discussion, including this: https://github.com/numpy/numpy/issues/24897 which also includes a large branch of mine to do this type of things.

I.e. this is (a) vastly more complicated, (b) there is already a lot of art here. Now that has stalled, and I am happy if someone looks at it. But this isn't a simple project at the least.

PR #31223, opened 2026-04-12
TYP: ``real_if_close`` and ``nan_to_num`` shape-typing

jorenham:
Both are elementwise functions (for the most part), so shape-typing is pretty straightforward because the shapes aren't modified. 

For `np.nan_to_num` I also added some additional overloads for specific non-numpy array-likes. I only did this for `nan_to_num` because it, unlike `real_to_close`, return bare scalars for 0-d input.

```
In [1]: np.real_if_close(1)
Out[1]: array(1)

In [2]: np.nan_to_num(1)
Out[2]: np.int64(1)
```

And since `ArrayLike` and `_SupportsArray` accept both scalars and arrays, this often resulted in `Any` as return type. The new overloads cover the (I think) most commonly used "array-like" cases, for which `nan_to_num` will now return fully static array or scalar types with specialized dtypes.

#### AI Disclosure
N/A

### DIFF
--- a/numpy/lib/_type_check_impl.pyi
+++ b/numpy/lib/_type_check_impl.pyi
@@ -13,6 +13,7 @@ from numpy._typing import (
     _ArrayLike,
     _NestedSequence,
     _ScalarLike_co,
+    _Shape,
     _SupportsArray,
 )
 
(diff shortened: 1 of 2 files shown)


charris:
Thanks Joren.

PR #31224, opened 2026-04-12
TYP: ``lib.NumpyVersion`` final instance attributes and pos-only dunder params

jorenham:
Just dotting some "i"'s.

#### AI Disclosure
N/A


charris:
Thanks Joren.

PR #31225, opened 2026-04-12
TYP: ``fft.[i]fftshift`` shape-typing

jorenham:
`np.fft.fftshift` and `np.fft.ifftshift` retain shapes (but not `ndarray` subtype), which makes this pretty straighforward.

#### AI Disclosure
N/A


charris:
Thanks Joren. I am a bit curious why float16 was there in the first place.

jorenham:
> I am a bit curious why float16 was there in the first place.

Yea, I was wondering that myself :thinking:

PR #31234, opened 2026-04-13
BUG,DEP: Change logic for ``arr.view()`` to use ``_set_dtype``

### LINKED ISSUE #31192: Potential oversight from gh-29575: self-triggered deprecation warnings
I'm working on adjusting astropy following gh-29575 . From where I'm standing now it appears possible to hit the deprecation warning even without user code explicitly trying to set the `dtype` attribute of a ndarray subclass. Here's a reproducer
```py
from astropy.table import Column

c = Column([1,2,3])

c.view("i8")
```

run as `python -Werror t.py`

```python-traceback
  File "/Users/clm/dev/gh/neutrinoceros/astropy-compatibility-checks/numpy-dev/_astropy/t.py", line 5, in <module>
    c.view("i8")
    ~~~~~~^^^^^^
  File "/Users/clm/dev/gh/neutrinoceros/astropy-compatibility-checks/numpy-dev/_astropy/astropy/table/column.py", line 1269, in __setattr__
    super().__setattr__(item, value)
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
DeprecationWarning: Setting the dtype on a NumPy array has been deprecated in NumPy 2.4.
Instead of changing the dtype on an array x, create a new array with x.view(new_dtype)
```

This traceback unfortunately doesn't tell me much about where `__setattr__` is actually being called from, so I suspect it may be coming from a low level extension. I checked that no parent class to `Column`, other than `ndarray`, defines a `view` or a `__setattr__` method. My current conclusion is that these components are all inherited from `ndarray`, though I can't really reconcile this with the fact that using a *plain* `ndarray` instead of a `Column` in my example avoids the warning.
I'm not excluding that the problem may lie in astropy, though I'm currently stuck in my inquiry. Any help would be most welcome.

seberg:
This modifies the logic for `arr.view()` to:
1. Use `arr._set_dtype(new_dtype)` for subclasses to have a simple way to migrate away from 2.
2. Keep using `arr.dtype = new_dtype` if dtype is a setter property.
3. If neither is defined, we now call `__array_finalize__` with a new array that has the new dtype+shape. (Previously `__array_finalize__` was skipped, a subclass that wanted to know this happened had to implememt a `dtype` attribute setter).

The first two should be completely safe, with the only downside that we have to support `_set_dtype()` indefinitely :(.
(Maybe a NumPy 3 can switch over to just using step 3 always, but that would be a big transition).

OTOH, calling `__array_finalize__` with mismatching dtypes _could_ fail admitted.  I.e. `MaskedArrays` would fail for non-trivial cases (same itemsize, no structured dtypes) and even the trivial ones may give "invalid value" warnings due to the `_fill_value`.

However, this is also arguably a fix.  If a class dealt with it before correctly, it implemented `.dtype` attribute setting.  And if not and it stumbles on it, it was likely subtly broken around `dtype` handling with views.

(Claude used to execute much of this, but everything is very precisely how I want it.)

Closes gh-31192

Also CC @mhvk, you might be interested and have the right eye to look closely at this.

### DIFF
--- a/numpy/_core/src/multiarray/convert.c
+++ b/numpy/_core/src/multiarray/convert.c
@@ -6,6 +6,7 @@
 #include <structmember.h>
 
 #include "npy_config.h"
+#include "npy_pycompat.h"  // PyObject_GetOptionalAttr
 
 #include "numpy/arrayobject.h"
 #include "numpy/arrayscalars.h"
--- a/numpy/_core/src/multiarray/convert.c
+++ b/numpy/_core/src/multiarray/convert.c
@@ -24,6 +24,8 @@
 #include "convert.h"
 #include "array_coercion.h"
 #include "refcount.h"
+#include "getset.h"
+#include "npy_static_data.h"
 
 #if defined(HAVE_FALLOCATE) && defined(__linux__)
 #include <fcntl.h>
@@ -352,17 +354,17 @@ PyArray_ToString(PyArrayObject *self, NPY_ORDER order)
         || (PyArray_IS_F_CONTIGUOUS(self) && (order == NPY_FORTRANORDER))) {
         return PyBytes_FromStringAndSize(PyArray_DATA(self), (Py_ssize_t) numbytes);
     }
-    
+
     /* Avoid Ravel where possible for fewer copies. */
-    if (!PyDataType_REFCHK(PyArray_DESCR(self)) && 
+    if (!PyDataType_REFCHK(PyArray_DESCR(self)) &&
         ((PyArray_DESCR(self)->flags & NPY_NEEDS_INIT) == 0)) {
-        
+
         /* Allocate final Bytes Object */
         ret = PyBytes_FromStringAndSize(NULL, (Py_ssize_t) numbytes);
         if (ret == NULL) {
             return NULL;
         }
-        
+
         /* Writable Buffer */
         char* dest = PyBytes_AS_STRING(ret);
 
@@ -388,14 +390,14 @@ PyArray_ToString(PyArrayObject *self, NPY_ORDER order)
             Py_DECREF(ret);
             return NULL;
         }
-        
+
         /* Copy directly from source to destination with proper ordering */
         if (PyArray_CopyInto(dest_array, self) < 0) {
             Py_DECREF(dest_array);
             Py_DECREF(ret);
             return NULL;
         }
-        
+
         Py_DECREF(dest_array);
         return ret;
 
@@ -406,7 +408,7 @@ PyArray_ToString(PyArrayObject *self, NPY_ORDER order)
     if (contig == NULL) {
         return NULL;
     }
-    
+
     ret = PyBytes_FromStringAndSize(PyArray_DATA(contig), numbytes);
     Py_DECREF(contig);
     return ret;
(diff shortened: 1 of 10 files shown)


numpy/_core/tests/test_deprecations.py
```diff
@@ -259,6 +258,25 @@ def test_deprecated_shape_set(self):
         x = np.eye(2)
         self.assert_deprecated(setattr, args=(x, "shape", (4, 1)))
 
+class TestDeprecatedViewDtypePropertySetter(_DeprecationTestCase):
```
mhvk, line 263: Extra empty line..,

numpy/_core/src/multiarray/convert.c
```diff
@@ -545,26 +548,129 @@ PyArray_View(PyArrayObject *self, PyArray_Descr *type, PyTypeObject *pytype)
...
+    if (subtype != &PyArray_Type) {
+        PyObject *sub_set_dtype;
+        if (PyObject_GetOptionalAttr(
+                (PyObject *)subtype,
+                npy_interned_str._set_dtype, &sub_set_dtype) < 0) {
+            goto finish;
+        }
+        use_set_dtype = (sub_set_dtype != NULL &&
+                sub_set_dtype != npy_static_pydata.ndarray_set_dtype);
+        Py_XDECREF(sub_set_dtype);
```
mhvk, line 587: I guess in principle, one could keep `sub_set_dtype` and call it below (and same for `sub_dtype_descr`), but arguably what you have is clearer, so fine to have it like this too.

mhvk (commented):
This looks nice.

I wonder if we should not just move most of the `dtype` changing stuff here, i.e., here do the checks and pre-calculate the new shape and strides, and create an array directly with those. An advantage of that would be that we actually move away from setting attributes after the fact, and thus do not bake in relying on `_set_dtype` (or indeed `_set_shape`; instead, for full control, a subclass would need to override `.view()`, as in fact we do in astropy's `MaskedNDArray` class...). 

Arguably, this is out of scope for this PR, though... (and really nothing stopping us from doing it afterwards, as long as we do it before a release, so that `.view()` doesn't have to check `_set_dtype`)

numpy/_core/src/multiarray/convert.c
seberg: Ah, didn't do it for now.  For the `_set_dtype` function it's nice.  For the descriptor, it feels a bit raw.

Also removed the `tp_descr_set` check though... For Python implementation that is always implemented anyway.  We should have to make it a `try/except`.

seberg:
>  a subclass would need to override `.view()`, as in fact we do in astropy's MaskedNDArray class...).

Hmmm, the internal `_set_dtype()` call path doesn't matter. But what does matter a bit is keeping `_set_dtype` around indefinitely.

Let's see:
* in the long run, we could even just call `__array_finalize__` directly in principle. But we cannot transition there right now (because we don't know that this is OK).
* I thought calling `.view()` might not work, because of recursion.  But that is OK, since `ndarray.view()` can avoid the recursion.
  So this is actually neat, with two potential caveats:
  * We basically have to assume `.view()` is correctly implemented (for C-API uses of `PyArray_View()` which would now start calling `.view()` as well).
  * I am not sure how tedious it is to modify a current `.view` subclass implementation, because to avoid the new `DeprecationWarning` it must avoid `subclass_intstance.view(new_dtype)`, it might have to do `subclass_instance.view(np.ndarray).view(new_dtype).view(type(subclass_instance))`.
    (The problem is that `super().view(new_dtype)` would hit the deprecation path, I think?  I.e. we don't actually _know_ that `.view()` is fixed or not.)

(Maybe I'll make up my mind about the `view` idea later, right now I am tempted to be lazy and go with this, but that may be mostly the laziness speaking because this path is at least obvious.)

seberg:
Would be nice to make a decision here soon probably, since it changes how e.g. astropy downstream can react to the deprecation.

eendebakpt (approved):
It would have been nice to have simpler logic here (and maybe we can in the future), but this seems a good fix for now.

mhvk:
@seberg, @eendebakpt - I'm having a try to see if inside `PyArray_View` we cannot just avoid setting the dtype altogether. So far, it is looking pretty good, just giving errors in `MaskedArray` that I think are similar to those here. Maybe hold off on merging for another day or two?

mhvk (approved):
See #31271 for my (still not quite complete) alternative. As I note there, I think it can fairly easily be follow-up of this instead of really an alternative, so in that sense happy to just merge this first. So, I'll approve, but won't yet merge, since perhaps you want to have a look at #31271 first and have a better idea...

seberg:
OK, after thinking about it.  I'll put this in, I dunno if there is even anyone who'll need to implement `_set_dtype` (besides MaskedArrays) -- also since I think astropy kinda needs it.
But if we figure out the other PR by 2.5, we should put it in.

mhvk:
@seberg - more detailed report upcoming, but unfortunately this turned out to break things in astropy unrelated to setting the dtype (I think...) -- see https://github.com/astropy/astropy/issues/19579

seberg:
arggg :(.  But that smells like calling `__array_finalize__` with the new dtype broke things.  Which means we cannot hope to migrate to a world where `__array_finalize__` is a nice way to implement this.

mhvk:
Indeed, I just had a look too and the problem is `__array_finalize__`, which now has the new dtype. I think it may actually be a trivial fix, where also in the implementation of `Masked.view()`, we separate out viewing as a new class from setting the dtype... Which I think is what we need to move to anyway.

seberg:
> we separate out viewing as a new class from setting the dtype... Which I think is what we need to move to anyway.

Yes, but we can't cleanly transition, I guess?  I.e. this would break older `astropy` and maybe others...
So as much as I would like to go there, the simple thing might be to just stick with the bad "don't call `__array_finalize__`.
(In which case, it does beg the question if calling `.view()` isn't better does.)

The one slightly crazy alternative could be to wrap it into a `try/except` (with a `DeprecationWarning`), but I am not sure how much we can trust things to fail reliably when `__array_finalize__` vs. silently doing the wrong thing.

PR #31237, opened 2026-04-13
BUG: Remove thread-local allocator cache implementation

ngoldbaum:
### PR summary

Fixes #31179.

This reverts #30499 and #30673 in favor of the old single global allocator cache on the GIL-enabled build and no cache on the free-threaded build. Since #30499 was merged, CPython has changed `PyMy_RawMalloc` to use mimalloc instead of the system allocator on the free-threaded build. That means we can rely on mimalloc to cache allocations.

The thread-local cache we have currently is complicated and leads to observed races in our CI (#31179).

Using uv's build of Python, I consistently see better allocation performance on the GIL-enabled build:

```

| Change   | Before [7a0dfadc] <main>   | After [be63b9f2] <rm-alloc-cache-ft>   |   Ratio | Benchmark (Parameter)                                    |
|----------|----------------------------|----------------------------------------|---------|----------------------------------------------------------|
| -        | 2.89±0.03ms                | 2.76±0.02ms                            |    0.96 | bench_alloc_cache.SmallArrayCreation.time_full_loop(128) |
| -        | 2.59±0.02ms                | 2.50±0.04ms                            |    0.96 | bench_alloc_cache.SmallArrayCreation.time_full_loop(64)  |
| -        | 2.91±0.03ms                | 2.76±0.02ms                            |    0.95 | bench_alloc_cache.SmallArrayCreation.time_full_loop(256) |
| -        | 2.63±0.03ms                | 2.48±0.02ms                            |    0.94 | bench_alloc_cache.SmallArrayCreation.time_full_loop(16)  |
| -        | 2.62±0.02ms                | 2.47±0.03ms                            |    0.94 | bench_alloc_cache.SmallArrayCreation.time_full_loop(4)   |

```

I see no significant performance change on the free-threaded build.

#### AI Disclosure

I used Claude to help generate the new benchmark. Happy to leave it out if you don't think it's useful.

### DIFF
--- a/numpy/_core/src/multiarray/alloc.c
+++ b/numpy/_core/src/multiarray/alloc.c
@@ -27,9 +27,11 @@
 #endif
 #endif
 
-/* Do not enable the alloc cache if the GIL is disabled, or if ASAN or MSAN
- * instrumentation is enabled. The cache makes ASAN use-after-free or MSAN
- * use-of-uninitialized-memory warnings less useful. */
+/* Do not enable the alloc cache on the free-threaded build, or if ASAN or
+ * MSAN instrumentation is enabled. CPython uses mimalloc on the free-threaded
+ * build, which we trust to cache allocations better than we can. The cache
+ * makes ASAN use-after-free or MSAN use-of-uninitialized-memory warnings less
+ * useful. */
 #ifdef Py_GIL_DISABLED
 #    define USE_ALLOC_CACHE 0
 #elif defined(__has_feature)
--- a/numpy/_core/src/multiarray/alloc.c
+++ b/numpy/_core/src/multiarray/alloc.c
@@ -27,13 +27,17 @@
 #endif
 #endif
 
-/* Do not enable the alloc cache on the free-threaded build, or if ASAN or
- * MSAN instrumentation is enabled. CPython uses mimalloc on the free-threaded
- * build, which we trust to cache allocations better than we can. The cache
- * makes ASAN use-after-free or MSAN use-of-uninitialized-memory warnings less
- * useful. */
+
+/*
+ * CPython uses mimalloc on the free-threaded build, which we trust to cache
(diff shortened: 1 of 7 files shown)


numpy/_core/src/multiarray/alloc.c
```diff
@@ -29,10 +28,17 @@ extern "C" {
...
 
-/* Do not enable the alloc cache if ASAN or MSAN instrumentation is enabled.
- * The cache makes ASAN use-after-free or MSAN
- * use-of-uninitialized-memory warnings less useful. */
-#if defined(__has_feature)
+/*
+ * CPython uses mimalloc on the free-threaded build, which we trust to cache
+ * allocations better than we can.
+ */
+#ifdef Py_GIL_DISABLED
```
seberg, line 35: Should we maybe just do `|| PYTHON_USING_MIMALLOC` (whatever that exact spelling is)?

seberg (approved):
Thanks for adding benchmarks even.  Cool if this just works out for free-threading!

mhvk (approved):
This looks good. Agreed with @seberg that we might as well also not use our home-brew cache generally if a better method is available. Even better if that is linked to a python version -- if so, it will be more obvious when we can just remove the cache altogether.

numpy/_core/src/multiarray/alloc.c
ngoldbaum: The interpreter only uses mimalloc for the `PyMem_Raw` on the free-threaded build, so the way I have it spelled here is correct I think.

ngoldbaum:
Two approvals so I'm gonna merge this, thanks all! Also thanks @kumaraditya303 for working on this. No shame that we ended up just backing out the two PRs since it led you to making the default allocator mimalloc on the free-threaded build, which benefits all free-threaded applications.

PR #31238, opened 2026-04-13
BUG: Fix Stirling correction sign error in the BTPE algorithm

k88hudson:
Hi there, I came across this when comparing different implementations of the BTPE algorithm. Like some other languages, numpy inherits an error from section 5.3 in the [original 1988 paper by Kachitvichyanukul & Schmeiser](https://dl.acm.org/doi/pdf/10.1145/42372.42381): the acceptance bound adds all four Stirling terms, but this is incorrect: the corrections for x1 = y+1 and w = n-y+1 should be subtracted.

As far as I know the authors of GSL were the first to notice it, you can see their comment [here](https://github.com/ampl/gsl/blob/fb0cfbc82b5e9e4d22444efce9241924f4aefb83/randist/binomial_tpe.c#L339-L361). My collaborator and I have independently checked the equations, as well as maintainers of Julia's implementation (you can see the fix [here](https://github.com/JuliaStats/Distributions.jl/pull/2053)).

Happy to adjust the comment or provide more details if it would be helpful!


rkern:
Thank you! Per [our compatibility guarantees](https://numpy.org/neps/nep-0019-rng-policy.html) for the RNG code, we will need to copy the old implementation over to `legacy-distributions.c` first so that `RandomState` remains bug-for-bug compatible.

This does change the RNG stream. With a bit of searching, I can find a PCG64 state that will reliably provoke a difference, so we can devise a regression test, at least.

```python
import numpy as np

def get_vulnerable_bg():
    STATE = {'bit_generator': 'PCG64', 'state': {'state': 339225526786748945562563845880185242573, 'inc': 114135179160287400024908587472913682319}, 'has_uint32': 0, 'uinteger': 0}
    bg = np.random.PCG64()
    bg.state = STATE
    return bg

rng = np.random.Generator(get_vulnerable_bg())
assert rng.binomial(500, 0.5) == 238
rs = np.random.RandomState(get_vulnerable_bg())
assert rs.binomial(500, 0.5) == 227
```

The important thing is that the first two floats it pulls are `[0.9521737146091229, 0.934540472511796]`. The first one ensures that the logic goes to Step52. The second is within a narrow window induced by the first draw such that rejection test flips result due to the change. With the new logic, the draw of 227 is rejected and moves on to the next draw to get to 238. In the old logic, 227 is accepted (and 238 is the following draw).

AI disclosure: Claude did help me identify good binomial parameters to try and then the windows for the variate draws. Actually hunting for the state was more due to blind searching of differences in the `binomial()` outputs on my own, but Claude's analysis does help validate that these values are indeed exercising the changed logic.

[edit: I copy-pasted the numbers to the wrong sources and explained it backwards. The above edited code and explanation is now correct]

ngoldbaum:
Let us know if you need a hand responding to Robert's comments - it can be a bit overwhelming getting NumPy PRs mergeable if you're not used to the process.

k88hudson:
Thanks for the guidance/investigation @rkern and the offer @ngoldbaum, I'll take a look and see how far I get / let you know

k88hudson:
Ok, I looked at some previous commits and tried doing the legacy implementation for RandomState + some regression tests. I am not sure if I am missing something but I think I got the opposite values (Generator=238, RandomState=227). That is super cool how you were able to find that :)

rkern:
Yes, apologies. I copy-pasted in the wrong order. I will edit my comment so there isn't misleading history.

ngoldbaum:
Just a heads up you'll need to add a release note too. See https://github.com/numpy/numpy/blob/main/doc/release/upcoming_changes/README.rst for instructions.

k88hudson:
You know what, as I was writing the release notes I think I see another problem: it looks like the leading constant should be `13860`, but the code has the 6 and the 8 swapped: 
https://github.com/numpy/numpy/blob/67a64d172eb95fa51a1c719bb227fa0033fcedef/numpy/random/src/distributions/distributions.c#L746-L752

Compare with page 4 of https://dl.acm.org/doi/pdf/10.1145/42372.42381
image

Assuming I'm not missing something, do you want me to fix that here or should we handle it separately?

rkern:
Hmm, yes. `13860.0` also matches the BTPE sources from Numeric's RANLIB that this was originally sourced from, but this [error has been there from the beginning](https://github.com/numpy/numpy/blob/1e7839d74149ebbd7113f15379b23a42b6a62b01/scipy/corelib/mtrand/distributions.c#L300-L303). Our code review in those days was nonexistent. Thank you for your excellent eyes!

It's possible that that state will no longer work (but it also still might).

rkern:
Yes, I believe it still works.

k88hudson:
Yes I think we're in luck!

k88hudson:
Let me know if the release notes look ok / you want me to do any commit rewriting

ngoldbaum:
Out of curiosity: did you find both bugs using Mark I Human eyeballs or with an AI model or some other kind of tool assistance?

doc/release/upcoming_changes/31238.compatibility.rst
```diff
@@ -0,0 +1,22 @@
...
+step of the BTPE algorithm used by `numpy.random.Generator.binomial`
+have been corrected:
+
+* The third and fourth error terms were added rather than
+  subtracted. This sign error was inherited from section 5.3 of the
+  original 1988 paper by Kachitvichyanukul & Schmeiser, which incorrectly
+  adds all four terms.
+
+* The leading coefficient had a digit-swap typo (``13680`` instead of ``13860``)
+  that was introduced when this code was first ported from Numeric's RANLIB.
```
rkern: Just to correct myself, looking at the paper again gave me the appropriate flashback: I did not actually port this from RANLIB directly, IIRC. I went back to the paper because I did not want to reproduce the somewhat unclear licensing of this source from ACM TOMS (at the time, there wasn't a clear license statement). That's why the `goto` labels follow the paper's labeling more closely and why this typo happened.

Not a super-important detail, all told, but I don't want my mis-claim about the provenance to raise any flags for anyone looking for license boo-boos.

k88hudson:
@ngoldbaum For the initial issue, a bit of both: I built a suite of benchmarks and comparison/analysis infrastructure which uses Claude/Opus 4.6, which my collaborators and I combine with somewhat painstaking human review of the original papers. One of the things I look for is any differences between a set of reference implementations: the infra surfaced a comment in the GSL code about the error in the paper, and we found this so surprising that we started looking around at other languages to see if it was inherited elsewhere (it was also in R, Julia). My collaborator @jasonasher also worked through the derivations by hand to independently confirm the fix.

For the second bug, I just noticed it by chance double checking the final diff against the paper. But I imagine I would have caught it with the comparison infra if numpy had been part of my original reference set!

doc/release/upcoming_changes/31238.compatibility.rst
k88hudson: Ok thanks. I'll update the comment 👍

rkern:
My review with Claude also noted that the `legacy_random_multinomial` just calls `random_multinomial` (only in order to change the integer type used for the output), and `random_multinomial` calls `random_binomial`. So we should copy the `random_multinomial` implementation over to `legacy_random_multinomial` and make it use `legacy_random_binomial` as well.

FWIW, there is a distributional test that we can do if we wanted to. It's just somewhat slow, having to draw a few hundred million samples. Probably wouldn't do it in our test suite, but it's nice to know it's detectable. Claude shows that if we use `n=300, 0.4`, we're in a sweet spot, and if we focus on the specific regions of the tails affected by `Step52` (`[86, 99] ∪ [141, 154]`, inclusive), we can amplify the signal.

image

k88hudson:
> My review with Claude also noted that the `legacy_random_multinomial` just calls `random_multinomial` (only in order to change the integer type used for the output), and `random_multinomial` calls `random_binomial`. So we should copy the `random_multinomial` implementation over to `legacy_random_multinomial` and make it use `legacy_random_binomial` as well.

Ah makes sense, good catch!

doc/release/upcoming_changes/31238.compatibility.rst
```diff
@@ -0,0 +1,22 @@
...
+
+* The third and fourth error terms were added rather than
+  subtracted. This sign error was inherited from section 5.3 of the
+  original 1988 paper by Kachitvichyanukul & Schmeiser, which incorrectly
+  adds all four terms.
+
+* The leading coefficient had a digit-swap typo (``13680`` instead of ``13860``)
+  that was introduced in the initial implementation.
+
+As a result, ``Generator.binomial`` may now return different samples for the
```
rkern: Worth noting `multinomial` as well.

rkern (approved):
Just the one comment on the release note. Otherwise, I think we're in good shape on my end!

@ngoldbaum Would you mind also stamping your approval when you think this is ready?

k88hudson:
Release notes updated. Thank you @rkern!

doc/release/upcoming_changes/31238.compatibility.rst
rkern: I meant here, primarily, to note that `Generator.multinomial` has changed, not just that `RandomState.multinomial` is _unchanged_.

rkern: You can use `[skip ci]` for doc-only changes like this.

ngoldbaum (approved):
Thanks @k88hudson and please let us know if you catch any other ancient typos like this or otherwise want to contribute, this was great!

PR #31242, opened 2026-04-14
MAINT: pass sorted=False in unique_{all,counts,inverse} to match docstring

### LINKED ISSUE #31241: BUG: `unique_all/counts/inverse` source omits `sorted=False` despite docstring claim of equivalence
### Describe the issue

The Array API helpers `np.unique_all`, `np.unique_counts`, and `np.unique_inverse` document themselves as equivalent to `np.unique(..., sorted=False)`, but their implementations call `np.unique(...)` without passing `sorted=False`, so they inherit the default `sorted=True`.

The companion function `np.unique_values` was updated in NumPy 2.3 to correctly pass `sorted=False` (see its `versionchanged::` note). The other three helpers were apparently missed.

### Reproducing the discrepancy

```python
import numpy as np
import inspect
print(np.__version__)  # 2.3.3
src = inspect.getsource(np.unique_counts)
assert "sorted=False" not in src  # passes — docstring says sorted=False, source omits it
```

### Source

`numpy/lib/_arraysetops_impl.py` (current `main`):

```python
def unique_counts(x):
    """
    ...
    This function is an Array API compatible alternative to::

        np.unique(x, return_counts=True, equal_nan=False, sorted=False)
    ...
    """
    result = unique(
        x,
        return_index=False,
        return_inverse=False,
        return_counts=True,
        equal_nan=False,
        # sorted=False is missing here
    )
```

For comparison, `unique_values` does it correctly:

```python
def unique_values(x):
    return unique(
        x,
        return_index=False,
        return_inverse=False,
        return_counts=False,
        equal_nan=False,
        sorted=False,
    )
```

`unique_all` and `unique_inverse` have the same omission as `unique_counts`.

### Impact

Observable behavior is unchanged today: `np.unique(..., sorted=False)` still sorts when any of `return_index/inverse/counts` is True (the fast hash path currently only triggers for the values-only case). However:

1. The source contradicts the docstring's stated equivalence.
2. If a future NumPy adds a hash-based fast path for these return modes (analogous to what 2.3 did for `unique_values`), these helpers will silently keep sorting because they're still asking for it.
3. The Array API spec does not require sorted output, and the existing `.. note::` in each docstring already permits the future change.

### Reproduce the equivalence claim being false

```python
>>> np.unique([3, 1, 2], sorted=False)        # genuinely unsorted
array([2, 1, 3])
>>> np.unique_values([3, 1, 2])               # also unsorted (correctly mirrors above)
array([2, 1, 3])
>>> np.unique_counts([3, 1, 2])               # docstring claims sorted=False, but…
UniqueCountsResult(values=array([1, 2, 3]), counts=array([1, 1, 1]))   # sorted
```

### Proposed fix

Add `sorted=False` to the `unique(...)` call in `unique_all`, `unique_counts`, and `unique_inverse`. Three one-line changes. PR to follow.

### NumPy/Python version information

- numpy 2.3.3
- Python 3 / CPython
- macOS (Darwin 25.3.0)

komaksym:
## Summary

`np.unique_all`, `np.unique_counts`, and `np.unique_inverse` document themselves as equivalent to `np.unique(..., sorted=False)`, but their implementations don't pass `sorted=False` to the underlying `np.unique` call, so they inherit the default `sorted=True`.

This PR adds the missing `sorted=False` argument to all three calls, mirroring what was already done for `np.unique_values` in NumPy 2.3.

Fixes #31241

## Behavior change

None observable today. `np.unique(..., sorted=False)` currently still returns sorted output when `return_index/inverse/counts` is True — only the values-only case takes the unsorted hash path. The existing `.. note::` in each docstring already covers this:

> This function currently always returns a sorted result, however, this could change in any NumPy minor release.

The patch makes the source honor the documented contract, so a future hash-path optimization for these return modes will be picked up automatically without further code changes.

## Test plan

- [ ] CI passes existing `_arraysetops_impl` tests (output unchanged)
- [ ] No docstring changes required

### DIFF
--- a/numpy/lib/_arraysetops_impl.py
+++ b/numpy/lib/_arraysetops_impl.py
@@ -493,6 +493,7 @@ def unique_all(x):
         return_inverse=True,
         return_counts=True,
         equal_nan=False,
+        sorted=False,
     )
     return UniqueAllResult(*result)
 
@@ -549,6 +550,7 @@ def unique_counts(x):
         return_inverse=False,
         return_counts=True,
         equal_nan=False,
+        sorted=False,
     )
     return UniqueCountsResult(*result)
 
@@ -606,6 +608,7 @@ def unique_inverse(x):
         return_inverse=True,
         return_counts=False,
         equal_nan=False,
+        sorted=False,
     )
     return UniqueInverseResult(*result)
 


jorenham:

PR #31244, opened 2026-04-14
BUG: add ``sorted=False`` to ``unique_all``, ``unique_counts``, ``unique_inverse``

Jah-yee:
Good day,

The docstrings claim equivalence to `np.unique(..., sorted=False)`, but the source was missing the `sorted=False` parameter. This fix adds it to `unique_all`, `unique_counts`, and `unique_inverse` to match `unique_values`, which was fixed in NumPy 2.3.

See: numpy/numpy/issues/31241

Thank you for your work on this project. I hope this small fix is helpful. Please let me know if there's anything to adjust.

Warmly, RoomWithOutRoof

### DIFF
--- a/numpy/lib/_arraysetops_impl.py
+++ b/numpy/lib/_arraysetops_impl.py
@@ -493,6 +493,7 @@ def unique_all(x):
         return_inverse=True,
         return_counts=True,
         equal_nan=False,
+        sorted=False,
     )
     return UniqueAllResult(*result)
 
@@ -549,6 +550,7 @@ def unique_counts(x):
         return_inverse=False,
         return_counts=True,
         equal_nan=False,
+        sorted=False,
     )
     return UniqueCountsResult(*result)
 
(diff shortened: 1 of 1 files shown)


ngoldbaum:
The linter failure is unrelated to this PR.

That said, it'd be nice to have tests that demonstrate that these keywords are being passed. It looks like none of our existing tests demonstrate that and that means this could regress again in the future.

seberg:
I think the reason is that until recently, we didn't have a `sorted=False` option.  I agree with Nathan though: it would be good to have a test (under the assumption that a reliable test is likely be possible even with a smallish dataset).

ngoldbaum:
This user recently was banned by CPython and astral for AI-generated spamming. Just FYI.

https://github.com/python/cpython/pull/148939#issuecomment-4308615641

seberg:
In that case, I'll close this for now.  If you follow-up, please post a comment and we'll reopen.  We are getting too many PRs like this these days.

PR #31247, opened 2026-04-15
ENH: added Support for scalar operations against python datetime and …

Anushreebasics:
…timedelta
fixes #29201 

### PR summary

- Updated ufunc argument conversion so Python datetime-like scalars are no longer kept as object scalars when mixed with datetime/timedelta ndarrays.

- The coercion is targeted to mixed scalar/array operations and preserves scalar-only behavior.

- Added regression tests for the reported ndarray cases.

#### AI Disclosure
"No AI tools used"

### DIFF
--- a/numpy/_core/src/umath/ufunc_type_resolution.c
+++ b/numpy/_core/src/umath/ufunc_type_resolution.c
@@ -742,6 +742,103 @@ timedelta_dtype_with_copied_meta(PyArray_Descr *dtype)
 }
 
 /*
+ * Creates a new NPY_DATETIME dtype, copying the datetime metadata
+ * from the given dtype.
+ */
+static PyArray_Descr *
+datetime_dtype_with_copied_meta(PyArray_Descr *dtype)
+{
+    PyArray_Descr *ret;
+    PyArray_DatetimeMetaData *dst, *src;
+    PyArray_DatetimeDTypeMetaData *dst_dtmd, *src_dtmd;
+
+    ret = PyArray_DescrNewFromType(NPY_DATETIME);
+    if (ret == NULL) {
+        return NULL;
+    }
+
+    src_dtmd = (PyArray_DatetimeDTypeMetaData *)((_PyArray_LegacyDescr *)dtype)->c_metadata;
+    dst_dtmd = (PyArray_DatetimeDTypeMetaData *)((_PyArray_LegacyDescr *)ret)->c_metadata;
+    src = &(src_dtmd->meta);
+    dst = &(dst_dtmd->meta);
+
+    *dst = *src;
+
+    return ret;
+}
+
+/*
+ * For add/subtract resolvers only: if one operand is a datetime/timedelta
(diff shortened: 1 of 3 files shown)


Anushreebasics:
@jorenham @ngoldbaum please review

ngoldbaum:
@riku-sakamoto since this is related to datetimes and timedeltas and you recently worked on them, I'd appreciate your eyes here.

And since this touches ufunc internals, maybe @mhvk?

riku-sakamoto:
@Anushreebasics 

IMO, it seems more appropriate that this kind of timedelta/datetime specific type resolution should be implemented in `PyUFunc_AdditionTypeResolver` and `PyUFunc_SubtractionTypeResolver` to minimize the risk of breaking other ufuncs.

https://github.com/numpy/numpy/blob/41439af3ab401f5f533452e333f15d1e4650c115/numpy/_core/src/umath/ufunc_type_resolution.c#L744-L777

Anushreebasics:
@riku-sakamoto please review

#### Summary of changes

- Removed the datetime/timedelta-specific coercion from the generic ufunc argument path in [ufunc_object.c:620]

- Added resolver-scoped handling only for add/sub in [ufunc_type_resolution.c:770]

- Hooked that logic into: 
[ufunc_type_resolution.c:857]
[ufunc_type_resolution.c:1049]

numpy/_core/src/umath/ufunc_type_resolution.c
```diff
@@ -766,6 +863,13 @@ PyUFunc_AdditionTypeResolver(PyUFuncObject *ufunc,
     int type_num1, type_num2;
     int i;
 
+    if (maybe_coerce_object_datetime_scalar_operand(&operands[0], operands[1]) < 0) {
```
mhvk, line 866: If I understand correctly, this only matters if one or the other operand is a numpy datetime, so I think we can put this a bit further down, and thus not slow down the much more common no-datetime case. 

In fact, down in the `if` statements about `type_num1 == NPY_TIMEDELTA`, etc., you already know that one type is a numpy type, so you only have to check the other.

mhvk (commented):
This looks reasonable, but I think can be moved slightly down so that it doesn't impact performance of regular floats, etc. (unless I misunderstand that there is always at least one numpy time involved; see in-line comment).

Ping also @seberg, since this would seem more easily solved in the new dtype machinery, but not sure whether that is possible/easy here.

seberg:
> since this would seem more easily solved in the new dtype machinery, but not sure whether that is possible/easy here.

Yeah, not sure I like this approach enough.  We have to plausible ones:
1. We special case datetime scalars more similar to int/float/complex scalars earlier on.  That may actually already be possible (I honestly am not sure).  So a "new dtype" territory.
   * I.e. for this check how comparison with Python integers is specialized.
2. Or, we actually decide that `np.array(datetime.datetime.now())` should return a NumPy datetime array (but of course unit promotion wise that will behave differently).

PR #31250, opened 2026-04-15
BUG: fix np.linalg.pinv(rtol=None) for integer input

### LINKED ISSUE #30917: BUG: numpy.linalg.pinv crashes with ValueError when called with rtol=None on an integer-dtype array [DeepTest]
### Describe the issue:

`np.linalg.pinv(a, rtol=None)` raises `ValueError` when `a` has an integer dtype (e.g. int32, int64). The `rtol=None` code path is the Array API standard default, so it should work with any numeric input.

Suggested Fix:
```python
● --- a/numpy/linalg/_linalg.py
   +++ b/numpy/linalg/_linalg.py
   @@ -2323,7 +2323,7 @@ def pinv(a, rcond=None, hermitian=False, *, rtol=_NoValue):
        if rcond is None:
            if rtol is _NoValue:
                rcond = 1e-15
            elif rtol is None:
   -            rcond = max(a.shape[-2:]) * finfo(a.dtype).eps
   +            rcond = max(a.shape[-2:]) * finfo(_commonType(a)[1]).eps
            else:
                rcond = rtol
```

```python
import numpy as np

# A valid integer matrix
a = np.array([[1, 2], [3, 4]], dtype=np.int32)

# This should compute the pseudo-inverse (pinv with default rcond works fine)
result = np.linalg.pinv(a, rtol=None)
print("pinv:", result)
```

### Error message:

```shell
  File "C:\Users\madanm\git\DeepTest\repo_benchmarks\numpy-DeepTests\standalone\crash_v2\score4\linalg_pinv_rtol_none_integer_dtype.py", line 26, in <module>
    result = np.linalg.pinv(a, rtol=None)
  File "C:\Users\madanm\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.13_qbz5n2kfra8p0\LocalCache\local-packages\Python313\site-packages\numpy\linalg\_linalg.py", line 2266, in pinv
    rcond = max(a.shape[-2:]) * finfo(a.dtype).eps
                                ~~~~~^^^^^^^^^
  File "C:\Users\madanm\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.13_qbz5n2kfra8p0\LocalCache\local-packages\Python313\site-packages\numpy\_core\getlimits.py", line 526, in __new__
    raise ValueError(f"data type {dtype!r} not inexact")
ValueError: data type <class 'numpy.int32'> not inexact
```

2.3.5

### Runtime Environment:

_No response_

Using DeepTest

erhan1209:
## Summary
- fix `np.linalg.pinv(..., rtol=None)` for integer-dtype inputs
- compute the default tolerance from the promoted inexact linalg dtype instead of the original integer dtype
- add a regression test covering integer input with `rtol=None`

## Testing
- `python -m py_compile numpy/linalg/_linalg.py numpy/linalg/tests/test_linalg.py`
- verified the patched `pinv` returns the same result as default `pinv(a)` for an `int32` matrix

### PR summary

Fix `np.linalg.pinv(..., rtol=None)` for integer-dtype inputs.

When `rtol=None`, `pinv` currently computes the default tolerance using
`finfo(a.dtype).eps`. That fails for integer arrays because `finfo` requires an
inexact dtype, so calls such as `np.linalg.pinv(np.array([[1, 2], [3, 4]], dtype=np.int32), rtol=None)`
raise `ValueError`.

This PR computes the default tolerance from the promoted inexact linalg dtype
instead, which matches the actual computation path used by `pinv`. A regression
test is included to verify that integer input with `rtol=None` behaves the same
as the default `pinv(a)` call.

Closes gh-30917.

I’m a NumPy user contributing this fix after noticing that `np.linalg.pinv`
fails on integer input when `rtol=None`. I traced the issue to the tolerance
calculation and put together a small fix with a regression test.

#### AI Disclosure

AI tool used: OpenAI Codex.

How it was used:
- to inspect the local checkout and identify a suitable issue to fix
- to propose and apply the code change
- to draft the regression test
- to help write this PR text

AI-generated content:
- the code change in `numpy/linalg/_linalg.py`
- the regression test added to `numpy/linalg/tests/test_linalg.py`
- this PR description text

All review, validation, and final submission decisions are made by me.

### DIFF
--- a/numpy/linalg/_linalg.py
+++ b/numpy/linalg/_linalg.py
@@ -2238,7 +2238,7 @@ def pinv(a, rcond=None, hermitian=False, *, rtol=_NoValue):
         if rtol is _NoValue:
             rcond = 1e-15
         elif rtol is None:
-            rcond = max(a.shape[-2:]) * finfo(a.dtype).eps
+            rcond = max(a.shape[-2:]) * finfo(_commonType(a)[1]).eps
         else:
             rcond = rtol
     elif rtol is not _NoValue:
--- a/numpy/linalg/tests/test_linalg.py
+++ b/numpy/linalg/tests/test_linalg.py
@@ -917,6 +917,15 @@ def test_pinv_rtol_arg():
         np.linalg.pinv(a, rcond=0.5, rtol=0.5)
 
 
+def test_pinv_integer_input_rtol_none():
+    a = np.array([[1, 2], [3, 4]], dtype=np.int32)
+
+    assert_almost_equal(
+        np.linalg.pinv(a, rtol=None),
+        np.linalg.pinv(a),
+    )
+
+
 class DetCases(LinalgSquareTestCase, LinalgGeneralizedSquareTestCase):
 
     def do(self, a, b, tags):


numpy/linalg/_linalg.py
```diff
@@ -2238,7 +2238,7 @@ def pinv(a, rcond=None, hermitian=False, *, rtol=_NoValue):
         if rtol is _NoValue:
             rcond = 1e-15
         elif rtol is None:
-            rcond = max(a.shape[-2:]) * finfo(a.dtype).eps
+            rcond = max(a.shape[-2:]) * finfo(_commonType(a)[1]).eps
```
seberg, line 2241: OK, I think we end up calling `svd` which also calls `_commonType`.

But since we are touching this here... Can you also fix the `_is_empty_2d` path down there and add a test for that fix as well?
Right now the result dtype diverges from the `svd` path below.

robrui:
Hi @erhan1209, I was looking into gh-30917 and opened a duplicate PR before noticing this one — sorry about that.

I also looked at @seberg’s comment about the `_is_empty_2d` path. My understanding is that the empty-input path should use the same result dtype as the normal SVD path, rather than keeping `a.dtype`.

So maybe this could address that part:

```python

_, result_t = _commonType(a)

res = empty(a.shape[:-2] + (n, m), dtype=result_t)

PR #31251, opened 2026-04-15
BUG: Make a bug sweep for missing error handling, etc.

seberg:
I asked claude to make a bug sweep and this was most of it (there were two or three weird things that it found but they are trickier and also pretty irrelevant).

Fully audited by hand and the more complex changes massaged to look nicer.

None of these are very worrying bugs, half of them just missing out of memory.  But while it had missed some directly next to things, all of them were good fixes.

(I have intentionally not done too much maintenance like moving cleanup into a `goto fail` here.)

### DIFF
--- a/numpy/_core/src/umath/wrapping_array_method.c
+++ b/numpy/_core/src/umath/wrapping_array_method.c
@@ -206,7 +206,7 @@ wrapping_method_get_identity_function(
     int res = context->method->wrapped_meth->get_reduction_initial(
             &orig_context, reduction_is_empty, item);
     for (int i = 0; i < nin + nout; i++) {
-        Py_DECREF(orig_descrs);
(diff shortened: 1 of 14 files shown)


numpy/_core/src/umath/wrapping_array_method.c
```diff
@@ -206,7 +206,7 @@ wrapping_method_get_identity_function(
     int res = context->method->wrapped_meth->get_reduction_initial(
             &orig_context, reduction_is_empty, item);
     for (int i = 0; i < nin + nout; i++) {
-        Py_DECREF(orig_descrs);
+        Py_DECREF(orig_descrs[i]);
```
charris, line 209: Whoa :)

charris (approved):

charris:
Thanks Sebastian.

PR #31256, opened 2026-04-16
ENH: Allow downstream to register for DLPack support

seberg:
This basically allows `ml_dtypes` to add something like:
```
np.dtypes.register_dlpack_dtype((4, 16), np.dtype("bfloat16"))
```
after defining their bfloat16 dtype.  After that import and export will work.  (And of course the same for all other dtypes.)

This is a pragmatic version just using dicts internally.  The code is slightly longer then I had hoped, but straight forward.

Now, generally, I don't like registering on the dtype instances, but:
* It seems simple and focused enough that it isn't a big long-term burden.
* The more proper "put it on the DType class" style is actually a bit more complex because it would need either need two full methods (and a loop to ask all DTypes or so) or assumptions around dtypes being non-parametric.

In the end, basically, I suspect this is good enough for all we need and not bad enough that need to worry a lot that it might stand in the way one day.

(I had an agent execute much of this, but all my design and all audited, just not all typed out by hand really...)

### DIFF
--- a/numpy/_core/src/multiarray/dlpack.c
+++ b/numpy/_core/src/multiarray/dlpack.c
@@ -69,11 +69,10 @@ dlpack_dtype_registry_lookup(uint8_t code, uint8_t bits)
             npy_static_pydata.dlpack_dtype_registry, key, &reg_val);
     Py_DECREF(key);
     if (gres < 0) {
-        Py_XDECREF(reg_val);
         return NULL;
     }
     if (gres == 0) {
-        PyErr_SetString(PyExc_RuntimeError,
+        PyErr_SetString(PyExc_BufferError,
                 "Unsupported dtype in DLTensor.");
         return NULL;
     }
@@ -518,7 +517,7 @@ array_dlpack(PyArrayObject *self,
     }
 
     if (stream != Py_None) {
-        PyErr_SetString(PyExc_RuntimeError,
+        PyErr_SetString(PyExc_ValueError,
                 "NumPy only supports stream=None.");
         return NULL;
     }
@@ -823,11 +822,16 @@ _register_dlpack_dtype(PyObject *NPY_UNUSED(self), PyObject *args)
         goto finish;
     }
 
-    if (code < 0 || code > 255 || bits_l < 8 || bits_l > 255
-            || (bits_l % 8) != 0) {
+    /* Sanity check code and bits, if DLPack relaxes this we can do this also. */
+    if (code < 0 || code > 255) {
         PyErr_SetString(PyExc_ValueError,
-                "register_dlpack_dtype: DLPack code must be in 0..255 and "
-                "bits must be a multiple of 8 in 8..255.");
+                "register_dlpack_dtype: DLPack code must be in 0..255.");
+        goto finish;
+    }
+    if (descr->elsize > 255/8 || descr->elsize * 8 != bits_l) {
+        PyErr_SetString(PyExc_ValueError,
+                "register_dlpack_dtype: number of bits must match the "
+                "dtypes and be <=255.");
         goto finish;
     }
 
@@ -898,8 +902,8 @@ _dlpack_registry_replace(PyObject *NPY_UNUSED(self), PyObject *args)
     }
 
     /* Replace the currently used dicts in place. */
-    npy_static_pydata.dlpack_dtype_registry = Py_NewRef(imp);
-    npy_static_pydata.dlpack_export_registry = Py_NewRef(exp);
+    Py_SETREF(npy_static_pydata.dlpack_dtype_registry, Py_NewRef(imp));
+    Py_SETREF(npy_static_pydata.dlpack_export_registry, Py_NewRef(exp));
     return ret;
 }
 
--- a/numpy/_core/src/multiarray/dlpack.c
+++ b/numpy/_core/src/multiarray/dlpack.c
@@ -54,7 +54,7 @@ dlpack_export_registry_lookup(PyArray_Descr *dtype,
 
 
 /*
- * Return a registered dtype or raise an error if non is found.
+ * Return a registered dtype or raise an error if none is found.
  */
 static PyArray_Descr *
 dlpack_dtype_registry_lookup(uint8_t code, uint8_t bits)
@@ -338,8 +338,8 @@ fill_dl_tensor_information(
         if (!reg_res) {
             PyErr_SetString(PyExc_BufferError,
                     "DLPack only supports signed/unsigned integers, float "
-                    "and complex dtypes (or dtypes registered via "
-                    "by importing external packages).");
+                    "and complex dtypes (or dtypes registered by third-party "
+                    "packages).");
             return -1;
         }
     }
--- a/numpy/_core/src/multiarray/dlpack.c
+++ b/numpy/_core/src/multiarray/dlpack.c
@@ -818,8 +818,8 @@ _register_dlpack_dtype(PyObject *NPY_UNUSED(self), PyObject *args)
 
     long code = 0;
     long bits_l = 0;
-    if (!PyArg_ParseTuple(args, "(ll)O&:register_dlpack_dtype", &code, &bits_l,
-            PyArray_DescrConverter, &descr)) {
+    if (!PyArg_ParseTuple(args, "(ll)O!:register_dlpack_dtype", &code, &bits_l,
+            &PyArrayDescr_Type, &descr)) {
         goto finish;
     }
 
@@ -874,7 +874,6 @@ _register_dlpack_dtype(PyObject *NPY_UNUSED(self), PyObject *args)
     ret = Py_NewRef(Py_None);
 
 finish:
-    Py_XDECREF((PyObject *)descr);
     Py_XDECREF(dlpack_tuple);
     Py_XDECREF(original);
     return ret;
--- a/numpy/_core/src/multiarray/dlpack.c
+++ b/numpy/_core/src/multiarray/dlpack.c
@@ -813,7 +813,8 @@ _register_dlpack_dtype(PyObject *NPY_UNUSED(self), PyObject *args)
     PyObject *ret = NULL;
     PyArray_Descr *descr = NULL;
     PyObject *dlpack_tuple = NULL;
-    PyObject *original = NULL;  // Original value if in dict.
+    PyObject *original_tuple = NULL;  // Existing dlpack tuple for export
+    PyObject *original_descr = NULL;  // Existing dtype for import
 
     long code = 0;
     long bits_l = 0;
@@ -828,10 +829,11 @@ _register_dlpack_dtype(PyObject *NPY_UNUSED(self), PyObject *args)
                 "register_dlpack_dtype: DLPack code must be in 0..255.");
         goto finish;
     }
+    // Check bits fit into 255 bytes via elsize to avoid elsize * 8 overflow.
     if (descr->elsize > 255/8 || descr->elsize * 8 != bits_l) {
         PyErr_SetString(PyExc_ValueError,
                 "register_dlpack_dtype: number of bits must match the "
-                "dtypes and be <=255.");
+                "dtype's elsize and be <=255.");
         goto finish;
     }
 
@@ -842,13 +844,13 @@ _register_dlpack_dtype(PyObject *NPY_UNUSED(self), PyObject *args)
 
     int set_res = PyDict_SetDefaultRef(
             npy_static_pydata.dlpack_export_registry, (PyObject *)descr, dlpack_tuple,
-            &original);
+            &original_tuple);
     if (set_res < 0) {
         goto finish;
     }
     else if (set_res == 1) {
         /* Key was present, allow if the value is equal: */
-        int exp_same = PyObject_RichCompareBool(original, dlpack_tuple, Py_EQ);
+        int exp_same = PyObject_RichCompareBool(original_tuple, dlpack_tuple, Py_EQ);
         if (exp_same < 0) {
             goto finish;
         }
@@ -860,18 +862,17 @@ _register_dlpack_dtype(PyObject *NPY_UNUSED(self), PyObject *args)
         }
     }
 
-    Py_CLEAR(original);  // re-use original.
     if (PyDict_SetDefaultRef(
             npy_static_pydata.dlpack_dtype_registry, dlpack_tuple, (PyObject *)descr,
-            &original) < 0) {
+            &original_descr) < 0) {
         goto finish;
     }
-    if ((PyObject *)descr != original) {
+    if ((PyObject *)descr != original_descr) {
         PyErr_Format(PyExc_ValueError,
             "register_dlpack_dtype: the same (code, bits) already maps to a "
             "%R which is not identical (it may be equal). "
             "The dtype->(code, bits) was, however, established.",
-            original);
+            original_descr);
         goto finish;
     }
 
@@ -879,7 +880,8 @@ _register_dlpack_dtype(PyObject *NPY_UNUSED(self), PyObject *args)
 
 finish:
     Py_XDECREF(dlpack_tuple);
-    Py_XDECREF(original);
+    Py_XDECREF(original_tuple);
+    Py_XDECREF(original_descr);
     return ret;
 }
 
--- a/numpy/_core/src/multiarray/dlpack.c
+++ b/numpy/_core/src/multiarray/dlpack.c
@@ -6,10 +6,86 @@
 
 #include "dlpack/dlpack.h"
 #include "numpy/arrayobject.h"
+#include "npy_pycompat.h"
 #include "npy_argparse.h"
 #include "npy_dlpack.h"
 #include "npy_static_data.h"
+#include "common.h"
 #include "conversion_utils.h"
+#include "descriptor.h"
+
+
+/*
+ * Find user-registered DLPack dtype mapping (1 if found, -1 on error).
+ */
+static int
+dlpack_export_registry_lookup(PyArray_Descr *dtype,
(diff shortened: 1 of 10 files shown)


numpy/dtypes.pyi
```diff
@@ -622,3 +624,5 @@ class StringDType(  # type: ignore[misc]
     def isalignedstruct(self) -> L[False]: ...
     @property
     def isnative(self) -> L[True]: ...
+
+def register_dlpack_dtype(dlpack_key: tuple[int, int], dtype: _DTypeLike[ScalarT], /) -> None: ...
```
jorenham: can `dtype` be any "dtype-like" here, or should it be an actual `np.dtype` instance?

seberg: Ah, yeah... I allowed it to convert to a dtype... but honestly, I dunno why I did, this should be a strict for being a dtype instance. (which I guess means I can just write `np.dtype` in the types)

numpy/_core/src/multiarray/dlpack.c
```diff
@@ -6,10 +6,86 @@
...
+    }
+    *out_code = (uint8_t)c;
+    *out_bits = (uint8_t)b;
+    Py_DECREF(val);
+    return 1;
+}
+
+
+/*
+ * Return a registered dtype or raise an error if non is found.
```
tylerjereddy:
```suggestion
 * Return a registered dtype or raise an error if none is found.
```

(just reading for curiosity)

numpy/_core/src/multiarray/dlpack.c
```diff
@@ -254,10 +330,18 @@ fill_dl_tensor_information(
...
+        int reg_res = dlpack_export_registry_lookup(dtype,
+                &managed_dtype.code, &managed_dtype.bits);
+        if (reg_res < 0) {
+            return -1;
+        }
+        if (!reg_res) {
+            PyErr_SetString(PyExc_BufferError,
+                    "DLPack only supports signed/unsigned integers, float "
+                    "and complex dtypes (or dtypes registered via "
+                    "by importing external packages).");
```
tylerjereddy: maybe slight reword on the "via by" ?

numpy/_core/tests/test_dlpack.py
```diff
@@ -186,3 +186,63 @@ def test_device(self):
...
+            np.dtypes.register_dlpack_dtype(key, dtype)
+
+    def test_register_idempotent(self):
+        dt = np.dtype(np.float16)
+        np.dtypes.register_dlpack_dtype((4, 16), dt)
+        np.dtypes.register_dlpack_dtype((4, 16), dt)
+
+    def test_roundtrip(self, dtype=np.dtype("S1")):  # noqa: B008
+        # Register "S1" as kDLFloat8_e3m4 == 7
+        # (use of kwarg ensure singleton in free-threading)
```
tylerjereddy, line 227: hmm, that seems pretty tricky to do right!

numpy/_core/tests/test_dlpack.py
```diff
@@ -186,3 +186,63 @@ def test_device(self):
...
+        np.dtypes.register_dlpack_dtype((7, 8), dtype)
+        x = np.array([1.0, 2.0], dtype="S1")
+        y = np.from_dlpack(x)
+        assert y.dtype == "S1"
+        assert_array_equal(x, y)
+
+    def test_register_conflict(self):
+        np.dtypes.register_dlpack_dtype((4, 16), np.dtype(np.float16))
+        with pytest.raises(ValueError, match="already exported"):
+            np.dtypes.register_dlpack_dtype((5, 16), np.dtype(np.float16))
```
tylerjereddy, line 237: This is based on the same object or the same properties of the dtype? I guess if someone tries hard enough to encode the same dtype "behavior" in different objects that's their problem (and not necessarily a problem anyway, apart from sanity)

numpy/_core/tests/test_dlpack.py
```diff
@@ -186,3 +186,63 @@ def test_device(self):
...
+        a = np.array(["12", "23"])
+        with pytest.raises(BufferError):
+            np.from_dlpack(a)  # dtype not yet registered
+
+        with pytest.raises(ValueError, match="already maps"):
+            np.dtypes.register_dlpack_dtype((4, 16), np.dtype("S2"))
+
+        # But... accept that this now does get exported (but won't roundtrip)
+        arr = np.from_dlpack(np.array(["12", "23"], dtype="S2"))
+        assert arr.dtype == np.float16
```
tylerjereddy, line 248: A bit weird I guess--on `main` it is: `BufferError: DLPack only supports signed/unsigned integers, float and complex dtypes.`.

No point in i.e., warning since nobody pays attention to warnings anyway?

Is there a good reason that someone would want to do this on purpose or is it most likely to happen by mistake in a control flow?

numpy/dtypes.py
```diff
@@ -23,8 +23,49 @@
 """
 
 # See doc/source/reference/routines.dtypes.rst for module-level docs
+__all__ = ["register_dlpack_dtype"]
 
-__all__ = []
+
+def register_dlpack_dtype(dlpack_key, dtype):
```
tylerjereddy: maybe dumb question--why the positional-only enforcement in the docs below but not the function sig proper? I guess it is in the `pyi` prototype as well, so might be safe to add

numpy/_core/tests/test_dlpack.py
seberg: Not in normal use, I think.  `np.dtype("bfloat16")` does return a singleton.  It also is only important during registration to avoid the error.  `"S1"` is just a _weird_ choice (but I chose it because it is weird) :).

numpy/_core/tests/test_dlpack.py
seberg: I dunno... I thought I'd just allow this in theory just as a (maybe silly) hedge towards two competing `bfloat16` implementations.
(Although even then, it might already work, due to the dict lookup in practice.)

numpy/dtypes.py
seberg: Honestly, no good reason, I'll change it...

EDIT: Added and removed the the duplication (as this is Python defined shouldn't be needed).

numpy/_core/tests/test_dlpack.py
seberg: Right, you can definitely register stuff that would break dlpack with dtype that are not supported by NumPy itself.  (up to the point of crashing things)

numpy/_core/src/multiarray/dlpack.c
seberg: Made it: ` (or dtypes registered by third-party packages).`

ngoldbaum:
@hawkinsp @jakevdp any chance one of you could weigh in on this from the perspective of the jax and ml_dtypes projects?

numpy/_core/src/multiarray/dlpack.c
```diff
@@ -719,3 +808,98 @@ from_dlpack(PyObject *NPY_UNUSED(self),
...
+    PyObject *ret = PyTuple_Pack(2,
+        npy_static_pydata.dlpack_dtype_registry,
+        npy_static_pydata.dlpack_export_registry);
+    if (ret == NULL) {
+        return ret;
+    }
+
+    /* Replace the currently used dicts in place. */
+    npy_static_pydata.dlpack_dtype_registry = Py_NewRef(imp);
+    npy_static_pydata.dlpack_export_registry = Py_NewRef(exp);
```
ngoldbaum:
```suggestion
    Py_SETREF(npy_static_pydata.dlpack_dtype_registry, Py_NewRef(imp));
    Py_SETREF(npy_static_pydata.dlpack_export_registry = Py_NewRef(exp));
```

Reference leak, I think?

numpy/_core/src/multiarray/dlpack.c
```diff
@@ -6,10 +6,86 @@
...
+    PyObject *reg_val = NULL;
+    int gres = PyDict_GetItemRef(
+            npy_static_pydata.dlpack_dtype_registry, key, &reg_val);
+    Py_DECREF(key);
+    if (gres < 0) {
+        Py_XDECREF(reg_val);
+        return NULL;
+    }
+    if (gres == 0) {
+        PyErr_SetString(PyExc_RuntimeError,
```
ngoldbaum: Should this be a `BufferError` instead? I'm not terribly familiar with this code, but it seems to be the error type used elsewhere in this compilation unit (except for one spot in `array_dlpack`, which might also be a bug).

numpy/_core/src/multiarray/dlpack.c
```diff
@@ -719,3 +808,98 @@ from_dlpack(PyObject *NPY_UNUSED(self),
...
+
+    long code = 0;
+    long bits_l = 0;
+    if (!PyArg_ParseTuple(args, "(ll)O!:register_dlpack_dtype", &code, &bits_l,
+            &PyArrayDescr_Type, &descr)) {
+        goto finish;
+    }
+
+    if (code < 0 || code > 255 || bits_l < 8 || bits_l > 255
+            || (bits_l % 8) != 0) {
```
ngoldbaum: Should there also be a check that `bits_l / 8 == descr->elsize`?

numpy/_core/src/multiarray/dlpack.c
```diff
@@ -719,3 +808,98 @@ from_dlpack(PyObject *NPY_UNUSED(self),
...
+            &original) < 0) {
+        goto finish;
+    }
+    if ((PyObject *)descr != original) {
+        PyErr_Format(PyExc_ValueError,
+            "register_dlpack_dtype: the same (code, bits) already maps to a "
+            "%R which is not identical (it may be equal). "
+            "The dtype->(code, bits) was, however, established.",
+            original);
+        goto finish;
```
ngoldbaum, line 876: It's a little weird that you leave the `dlpack_export_registry` mutated in this failure case.

IMO it would be cleaner to store the `npy_static_pydata` state as a tuple of (frozen?)dicts, so there's only one atomic unit of state to update and you don't have issues like this. Then you can use a copy-on-write pattern for the whole atomic unit. I think you also maybe need to use atomic loads and stores?

You could also use a critical section or a global PyMutex lock (the latter is probably easier to reason about given the C API calls you need to do). IMO this registration dance shouldn't matter for multithreaded performance - most of the time it'll happen at import time for a package like ml_dtypes.

numpy/_core/src/multiarray/dlpack.c
```diff
@@ -719,3 +808,98 @@ from_dlpack(PyObject *NPY_UNUSED(self),
...
+    if (!PyArg_ParseTuple(args, "(ll)O!:register_dlpack_dtype", &code, &bits_l,
+            &PyArrayDescr_Type, &descr)) {
+        goto finish;
+    }
+
+    if (code < 0 || code > 255 || bits_l < 8 || bits_l > 255
+            || (bits_l % 8) != 0) {
+        PyErr_SetString(PyExc_ValueError,
+                "register_dlpack_dtype: DLPack code must be in 0..255 and "
+                "bits must be a multiple of 8 in 8..255.");
```
ngoldbaum: The 255s in here are a bit of a lie, since multiples of 8 only go up to 248!

numpy/_core/src/multiarray/dlpack.c
```diff
@@ -6,10 +6,86 @@
...
+    PyObject *key = PyArray_IntTupleFromIntp(2, key_vals);
+    if (key == NULL) {
+        return NULL;
+    }
+    PyObject *reg_val = NULL;
+    int gres = PyDict_GetItemRef(
+            npy_static_pydata.dlpack_dtype_registry, key, &reg_val);
+    Py_DECREF(key);
+    if (gres < 0) {
+        Py_XDECREF(reg_val);
```
ngoldbaum: I guess the C API docs are a little ambiguous about this, but I think `reg_val` is always NULL in this case, so you can delete this. I'll update the CPython docs once I get an upstream clarification, but I'm pretty sure the docs are just imprecise.

numpy/_core/tests/test_dlpack.py
```diff
@@ -186,3 +186,63 @@ def test_device(self):
             x.__dlpack__(dl_device=(10, 0))
         with pytest.raises(ValueError):
             np.from_dlpack(x, device="gpu")
+
+
+class TestRegisterDlpackDtype:
+    @pytest.fixture(scope="class", autouse=True)
```
ngoldbaum, line 193: Can you mark this test as thread-unsafe, because you're intentionally mutating shared global state?

ngoldbaum (commented):
I asked Claude to look this over and it spotted some issues, see below.

numpy/_core/src/multiarray/dlpack.c
seberg: Hmmmm, I am not even sure I care for this check.  I was considering it, but then felt a bit it was a user problem dlpack ever allows this to mismatch.

numpy/_core/src/multiarray/dlpack.c
seberg: I had left this kinda intentional: This way you can actually have two `bfloat16` and a with a `try/except` you won't get yours be the "default" one, but it still can be exported as dlpack.

Not that any of this matters, if it did, I would really want a stronger error.

Anwyway, I don't actually care much, but you are right.  If we don't want to allow this we need to swap both atomically or have a lock and remove the item again.
(Let me see how it looks, I still don't really mind the first thing, although I agree it is a bit funny.)

numpy/_core/tests/test_dlpack.py
seberg: The fixture runs in the class scope so before pytest-run-parallel would start threads, I think?

numpy/_core/src/multiarray/dlpack.c
ngoldbaum: I personally am fine with checks the help users from shooting themselves in the foot.

numpy/_core/src/multiarray/dlpack.c
ngoldbaum: Ah that makes sense. I agree this is a weird corner case and it's unlikely that people will be dynamically defining these types at runtime in a manner where these thread safety concerns matter.

That said, I'd prefer to err on the side of simpler solutions for dealing with global state while still ensuring thread safety on the free-threaded build. That would make me want to choose to store the state as a tuple of two dicts and then use a copy-on-write pattern.

Also keep in mind that `Py_SETREF` isn't atomic, so I think you'll need to use explicit atomic writes if you want to make the current implementation thread-safe on the free-threaded build.

numpy/_core/tests/test_dlpack.py
ngoldbaum: No, that's not how pytest fixtures work. All this means is that the fixture is shared between all tests in the class. You're still updating global state. If the tests are run in a thread pool that global state transition could happen while some tests are running in other threads.

seberg: > If the tests are run in a thread pool that global state transition could happen while some tests are running in other threads.

Sorry, I am actually curious about this: The individual tests are thread-safe, so the downside of marking it is that we lose testing that.
OTOH, yes, clearly the fixture isn't thread-safe.  But right now we thread only on the individual test level, so this shouldn't matter.

I can see that it would be nice to thread across tests as well! But it might be great if the test runner could then be aware of the different level that is unsafe (the fixture level is unsafe but the individual tests should be tested threaded -- maybe a bit like an `xdist` group).

Unless I am missing something gigantic obvious here?

numpy/_core/src/multiarray/dlpack.c
ngoldbaum: The doc clarification got merged upstream

numpy/_core/src/multiarray/dlpack.c
seberg: > That said, I'd prefer to err on the side of simpler solutions for dealing with global state while still ensuring thread safety on the free-threaded build.

Well, if we are OK with the slightly odd `The dtype->(code, bits) was, however, established.` then this is already the simplest solution? We can rely on `PyDict_SetDefaultRef()` being atomic with no need for any CoW (except testing).
Unless we care about mutating both atomically together, but I don't really see that adding a lock just for that is useful?

numpy/_core/src/multiarray/dlpack.c
seberg: Yeah good spot, that should not have happened. I'll change the other one to a `ValueError` (that one is bad user input, not an inability to export/import the data as such).

(I'll see if I can't add a test...)

numpy/_core/src/multiarray/dlpack.c
seberg: Yeah, I was once thinking we should use bits slightly more broadly in dlpack... but we can think about relaxing this when that actually happens.

numpy/_core/tests/test_dlpack.py
ngoldbaum: I'm sorry, I was misunderstanding what was happening. I just tried running this branch under pytest-run-parallel with a TSan build, which I was pretty sure would lead to data races based on what I said above. But, no races. I think *I* was misunderstanding what class-scope fixtures do.

Anyway, sorry for the bother!

numpy/_core/src/multiarray/dlpack.c
ngoldbaum: OK, fine as-is.

ngoldbaum (approved):
Thanks for bearing with me!

seberg:
Thanks for the review, I guess we can put this in.  I'll open an `ml_dtypes` issue to actually use it!

The test failures were the spurious failures from the slightly different Python 3.15 API from what was expected in the PR.

Issue #31254, opened 2026-04-15
BUG: SIGSEGV on non-ASCII strings for ``arrayflags_setitem`` and ``PyArray_ToFile``


mjbommar:
### Describe the issue:

Two unchecked uses of `PyUnicode_AsASCIIString` can trivially trigger SIGSEGVs.  Both follow the same pattern - any non-ASCII chars in the flag or format string will break downstream calls or copies.  

I'm guessing the easiest solution is probably to throw a ValueError or UnicodeXError or something similar to hint to the user.



### Reproduce the code example:

```python
>>> import numpy
>>> numpy.array([1]).flags['µ'] = True
(SIGSEGV)

>>> import numpy
>>> arr = numpy.zeros(1)
>>> arr.tofile(open("/tmp/x", "wb"), sep=" ", format="µ%d")
(SIGSEGV)
```

### Error message:

```shell
# Bug 1: arrayflags_setitem (flagsobject.c:588)                                                                     

Trigger: `np.array([1]).flags["µ"] = True`

                                                          
  Thread 1 "python3" received signal SIGSEGV, Segmentation fault.
  0x00007ffff69ccf1e in arrayflags_setitem () from .../numpy/_core/_multiarray_umath.cpython-314-x86_64-linux-gnu.so
  #0  0x00007ffff69ccf1e in arrayflags_setitem ()                                        
  #1  0x0000000001842d17 in _TAIL_CALL_STORE_SUBSCR ()    
  #2  0x00000000018ab8e5 in _PyEval_Vector ()             
  #3  0x00000000018ab5e6 in PyEval_EvalCode ()            
  #4  0x00000000018caeb2 in run_mod ()                    
  #5  0x0000000001a4564d in pyrun_file ()                                                                           
  #6  0x0000000001a44dc2 in _PyRun_SimpleFileObject ()                                                              
  #7  0x0000000001988b1c in _PyRun_AnyFileObject ()       
  #8  0x0000000001988ab8 in pymain_run_file_obj ()        
  #9  0x00000000019889b5 in pymain_run_file ()            
  #10 0x000000000199705b in Py_RunMain ()                                                                                                                                                                                               
  #11 0x0000000001996cad in pymain_main ()                
  #12 0x0000000001996ac6 in main ()                       

  (registers at fault)                                    
  rax  0x0                  <-- NULL                      
  rdi  0x7ffff74a0140                                     
  rip  0x7ffff69ccf1e <arrayflags_setitem+254>       
     


[...]

PR #31257, opened 2026-04-16
BUG: handle non-ASCII flags and tofile format strings

mjbommar:
### PR summary

As described in #31254, non-ASCII strings are not handled 
safely by `arrayflags_setitem` and `PyArray_ToFile`.

What to do with Unicode strings elsewhere upstream / downstream
is a complicated question that has design consequences, so 
@WarrenWeckesser suggested reasonable error handling in the bug
report comments to provide the user with useful, non-segfaulting
execptions.

This PR simply implements those recommendations and adds coverage
to ensure `KeyError` / `ValueError` are properly raised when 
ecnountered.

#### First-time contributor introduction
Hi there! Numeric/numarray user who has been on the long journey
after switching from Matlab and Fortran a long time ago...thank you
for all of your work - I know how much went into 2.0!

#### AI Disclosure

The original C issue was detected using Claude Agent SDK and a variety of
tools like semgrep / ast-grep / libclang.  

Claude Code (Opus 4.6) and Codex CLI were used to prepare the PR branch,
although the credit for the proposed fixes themselves really go to 
@WarrenWeckesser.

PS: sorry again for earlier auto-submission.

### DIFF
--- a/doc/release/upcoming_changes/31257.change.rst
+++ b/doc/release/upcoming_changes/31257.change.rst
@@ -2,4 +2,4 @@ Non-ASCII inputs no longer crash ``flags`` assignment or ``tofile``
 -------------------------------------------------------------------
 Passing a non-ASCII flag key to ``ndarray.flags[...] = ...`` now raises
 ``KeyError`` instead of crashing. Passing a non-ASCII ``format`` argument to
-`numpy.ndarray.tofile` now raises ``ValueError`` instead of crashing.
+``numpy.ndarray.tofile`` now raises ``UnicodeEncodeError`` instead of crashing.
--- a/doc/release/upcoming_changes/31257.change.rst
+++ b/doc/release/upcoming_changes/31257.change.rst
@@ -1,5 +1,0 @@
-Non-ASCII inputs no longer crash ``flags`` assignment or ``tofile``
--------------------------------------------------------------------
-Passing a non-ASCII flag key to ``ndarray.flags[...] = ...`` now raises
-``KeyError`` instead of crashing. Passing a non-ASCII ``format`` argument to
-``numpy.ndarray.tofile`` now raises ``UnicodeEncodeError`` instead of crashing.
--- a/doc/release/upcoming_changes/31257.change.rst
+++ b/doc/release/upcoming_changes/31257.change.rst
@@ -1,0 +1,5 @@
+Non-ASCII inputs no longer crash ``flags`` assignment or ``tofile``
+-------------------------------------------------------------------
+Passing a non-ASCII flag key to ``ndarray.flags[...] = ...`` now raises
+``KeyError`` instead of crashing. Passing a non-ASCII ``format`` argument to
+`numpy.ndarray.tofile` now raises ``ValueError`` instead of crashing.
--- a/numpy/_core/src/multiarray/convert.c
+++ b/numpy/_core/src/multiarray/convert.c
@@ -299,6 +299,16 @@ PyArray_ToFile(PyArrayObject *self, FILE *fp, char *sep, char *format)
                 }
             }
             byteobj = PyUnicode_AsASCIIString(strobj);
+            if (byteobj == NULL) {
+                Py_DECREF(strobj);
+                Py_DECREF(it);
+                if (n4 != 0) {
+                    PyErr_SetString(PyExc_ValueError,
+                            "The `format` parameter must contain only ASCII "
+                            "characters.");
+                }
+                return -1;
+            }
             NPY_BEGIN_ALLOW_THREADS;
             n2 = PyBytes_GET_SIZE(byteobj);
             n = fwrite(PyBytes_AS_STRING(byteobj), 1, n2, fp);
--- a/numpy/_core/src/multiarray/convert.c
+++ b/numpy/_core/src/multiarray/convert.c
@@ -141,7 +141,7 @@ PyArray_ToFile(PyArrayObject *self, FILE *fp, char *sep, char *format)
     npy_intp n, n2;
     size_t n3, n4;
     PyArrayIterObject *it;
-    PyObject *obj, *strobj, *tupobj, *byteobj;
+    PyObject *obj, *strobj, *tupobj, *byteobj, *formatobj = NULL;
 
     n3 = (sep ? strlen((const char *)sep) : 0);
     if (n3 == 0) {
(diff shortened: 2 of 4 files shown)


jorenham:
We prefer to talk to humans. You ignored the PR template, and did not disclose AI usage.

mjbommar:
Very sorry @jorenham , codex sub agent ignored instructions not to execute gh CLI commands...

I think the patch contents itself were already specified by @WarrenWeckesser in the original report that I made yesterday.  I reviewed the code, ran the original MRE to confirm the patches fixed the issue, and checked that build/test are green after.

If you re-open it, I will manually edit this to clean up the template.  Otherwise
I'm just going to create clutter opening another PR to fix as @WarrenWeckesser requested

numpy/_core/src/multiarray/convert.c
```diff
@@ -299,6 +299,16 @@ PyArray_ToFile(PyArrayObject *self, FILE *fp, char *sep, char *format)
                 }
             }
             byteobj = PyUnicode_AsASCIIString(strobj);
+            if (byteobj == NULL) {
+                Py_DECREF(strobj);
+                Py_DECREF(it);
+                if (n4 != 0) {
```
tylerjereddy, line 305: Note that the full NumPy testsuite passes for me locally on this branch if I simplify this conditional to always execute per diff below. If this is genuinely needed, it might be helpful to enforce the requirement with a test. If there's a good reason it is needed, but can't be tested, perhaps an explanatory comment might at least make sense.

```diff
--- a/numpy/_core/src/multiarray/convert.c
+++ b/numpy/_core/src/multiarray/convert.c
@@ -302,7 +302,7 @@ PyArray_ToFile(PyArrayObject *self, FILE *fp, char *sep, char *format)
             if (byteobj == NULL) {
                 Py_DECREF(strobj);
                 Py_DECREF(it);
-                if (n4 != 0) {
+                if (1) {
                     PyErr_SetString(PyExc_ValueError,
                             "The `format` parameter must contain only ASCII "
                             "characters.");
```

mjbommar: I'll add a test to catch this branch (e.g., `str()`)

mjbommar: @tylerjereddy  - OK, should see coverage on that branch with the extra test I pushed

seberg: I would be tempted to just stick with the unicode error either way.  Sure, it's not necessary immediately clear it comes from the format itself, but I suspect it's clear enough?

mjbommar: @seberg I'd defer to you and @WarrenWeckesser, who originally proposed it.

WarrenWeckesser: It turns out it is probably better to stick with the unicode error as @seberg suggests.  I hadn't considered Unicode string arrays in my initial comments in #31254.  As it is now in the PR, if the array contains non-ASCII Unicode, and `format="%s"`, the error message in the `ValueError` is not correct:

```
In [25]: np.__version__
Out[25]: '2.5.0.dev0+git20260416.dedf7c2'

In [26]: s = np.array(['λ123'], dtype='<U4')

In [27]: s.tofile('junk.dat', sep="; ", format="%s")
Cell In[27], line 1
----> 1 s.tofile('junk.dat', sep="; ", format="%s")

ValueError: The `format` parameter must contain only ASCII characters.
```

mjbommar:
OK, hopefully this hits the mark now.  Here's a summary of everything that's in the PR now:

  - [x] Fixed `arrayflags_setitem` so non-ASCII flag keys no longer crash and instead follow the existing `KeyError` path.
  - [x] Fixed `ndarray.tofile()` so non-ASCII `format` strings no longer crash and now raise `ValueError`.
  - [x] Preserved `UnicodeEncodeError` for non-ASCII Unicode array element data in `tofile()`, instead of misreporting it as a `format` error.
  - [x] Covered the reviewed `format="%s"` Unicode-array case so it now raises `UnicodeEncodeError` rather than a misleading `ValueError`.
  - [x] Added regression tests for non-ASCII flag keys, non-ASCII `format`, non-ASCII element data without `format`, and non-ASCII element data with ASCII `format="%s"`.
  - [x] Added a release note describing the crash fixes and resulting exception behavior.

If you want me to clean up the branch so it's clean without the commit history, just let me know and I'll push a single clean commit.  Wasn't sure if that will erase the code review or break the GH UI now.  I'm afraid to do anything these days with all the changes they made...

numpy/_core/src/multiarray/convert.c
mjbommar: @WarrenWeckesser @seberg OK, see note below, hopefully this is what you were thinking for the format case

seberg: Can we please just use the error as-is?  I don't think it is worth going through the trouble to validate the format string explicitly just to give it an explicit `ValueError`.

doc/release/upcoming_changes/31257.change.rst
ngoldbaum, line 1: total nit but this should probably be labeled as an "improvement" rather than a "change"

numpy/_core/src/multiarray/convert.c
```diff
@@ -318,13 +329,15 @@ PyArray_ToFile(PyArrayObject *self, FILE *fp, char *sep, char *format)
...
                     return -1;
                 }
             }
             Py_DECREF(strobj);
             PyArray_ITER_NEXT(it);
         }
+        Py_XDECREF(formatobj);
         Py_DECREF(it);
     }
     return 0;
```
ngoldbaum, line 343: Another way to handle the repeated cleanup code is to add a `cleanup` goto label below here that does `Py_XDECREF` on all the objects that need cleanup in error paths and then use `goto cleanup` in error paths above.

You can even define the return value like `int ret = -1` in the function header and then set `ret = 0` here and be sure then `return ret` in the cleanup code to unify the error paths and the happy path.

You also need to make sure that you initialize all the variables to NULL before the first use. The cost of this approach is some extra NULL checks which IMO is worth it for the improved maintainability and readability.

IMO that's better style for CPython C API code (despite the use of goto) because it avoids duplicate or almost-duplicate cleanup code which can sometimes be tricky to reason about.

ngoldbaum (commented):
I triggered CI but unfortunately github actions is having issues and there are lots of unrelated test failures. Someone is probably going to need to re-trigger.

I left a comment below about C API style and another nit about how you classified the release note. Please take or leave them as you see fit.

seberg:
Close re-open, since even another rerun failed... let's see if it works with that.

ngoldbaum:
Should we perhaps merge this as-is? My last comment is more of a style thing.

seberg:
> Should we perhaps merge this as-is?

Sounds good.  (I might be tempted to just delete the release note, as I think we don't really have a habit of it for bug-fixes, but fine either way.)

seberg:
Let me just do that.  Thanks @mjbommar, I agree a `goto` pattern would be much clearer, but it is also OK to do this as a bug-fix.

seberg:
Test failure is unrelated (I think a rebase would fix it).  Thanks!

PR #31258, opened 2026-04-16
ENH: mark _import_array function as inline

kumaraditya303:
### PR summary

Marked function as inline to avoid unused function warning.

#### AI Disclosure

Not used.


ngoldbaum:
This came out of https://github.com/numpy/numpy/pull/31091, I thought it made sense to upstream this independently of that big change.

ngoldbaum:
@seberg do you think this needs a release note?

seberg:
I am happy without it, since I am very sure it shouldn't break anything (but happy to break it out so we don't backport something we are only _pretty_ sure about).

seberg (approved):

Issue #17286, opened 2020-09-10
ENH: Implement a maxlag like feature for `np.correlate`


seberg:
The old PR gh-5978 wanted to add a `maxlag` functionality to our correlate/convolve function, so that correlations with a smaller shift can be conveniently calculated.

In general, this is probably a good feature to add, but the PR is outdated and always had a few outstanding issues.  We had discussed this briefly, and one option may be to implement this, but potentially start in SciPy since it has the more full featured versions and any API additions in NumPy should probably end up in SciPy anyway.

PR #5978, opened 2015-06-17
ENH: adding maxlag mode to convolve and correlate

HoniSanders:
What was troubling me is that numpy.correlate does not have a maxlag feature. This means that even if I only want to see correlations between two time series with lags between -100 and +100 ms, for example, it will still calculate the correlation for every lag between -20000 and +20000 ms (which is the length of the time series). This (theoretically) gives a 200x performance hit! Is it possible that I could contribute this feature?

I have introduced this question as a [numpy issue](https://github.com/numpy/numpy/issues/5954), a [scipy issue](https://github.com/scipy/scipy/issues/4940) and on the [scipy-dev list](http://mail.scipy.org/pipermail/scipy-dev/2015-June/020757.html).  It seems the best place to start is with numpy.correlate, so that is what I am requesting.  

This is my first experience with contributing to open-source software, so any pointers are appreciated. 

Previous discussion of this functionality can be found at [another discussion on numpy correlate (and convolution)](http://numpy-discussion.10968.n7.nabble.com/another-discussion-on-numpy-correlate-and-convolution-td32925.html).  Other issues related to correlate functions include [ENH: Fold fftconvolve into convolve/correlate functions as a parameter #2651](https://github.com/scipy/scipy/issues/2651), [Use FFT in np.correlate/convolve? (Trac #1260) #1858](https://github.com/numpy/numpy/issues/1858), and [normalized cross-correlation (Trac #1714) #2310](https://github.com/numpy/numpy/issues/2310).

I have added extra options for the "mode" parameter, namely an int or an int tuple that determine for which lags the correlation/convolution should be calculated.  These correspond to the values in the vector that the same tuple would generate if used as an argument to the numpy.arange() function.  

Here are the specific issues that could use attention in this implementation:

Is it ok that this implementation breaks the previous undocumented behavior of mode=0, 1, or 2 returning the correlation for mode='valid', 'same', or 'full' respectively? If not, how should maxlag input be passed?
Should the checks that are included in convolve() be added to correlate()?
Is it ok that I used NPY_MAX_INT to represent requests for a default argument in the C code?
Do I need to check for "default arguments" at all, given that the calling functions should now always call _pyarray_correlate with the appropriate number of arguments? I kind of feel there is some legacy behavior I should be maintaining, but I'm not 100% sure.
Is the error message in _pyarray_correlate "PyErr_SetString(PyExc_ValueError, "mode must be 0, 1, 2, or 3")" still valuable? What should it be replaced with?
[...]

Issue #5954, opened 2015-06-09
How to limit cross correlation window width in Numpy?


bringingheavendown:
I am learning numpy/scipy, coming from a MATLAB background. The xcorr function in Matlab has an optional argument "maxlag" that limits the lag range from –maxlag to maxlag. This is very useful if you are looking at the cross-correlation between two very long time series but are only interested in the correlation within a certain time range. The performance increases are enormous considering that cross-correlation is incredibly expensive to compute.

What is troubling me is that numpy.correlate does not have a maxlag feature. This means that even if I only want to see correlations between two time series with lags between -100 and +100 ms, for example, it will still calculate the correlation for every lag between -20000 and +20000 ms (which is the length of the time series). This (theoretically) gives a 200x performance hit! Is it possible that I could contribute this feature?

I have introduced this question as a [scipy issue](https://github.com/scipy/scipy/issues/4940) and on the [scipy-dev list](http://mail.scipy.org/pipermail/scipy-dev/2015-June/020757.html).  It seems the best place to start is with numpy.correlate, so that is what I am requesting.  I have done a [simple implementation](https://gist.github.com/bringingheavendown/b4ce18aa007118e4e084) which gives 50x speedup under [my conditions](https://github.com/scipy/scipy/issues/4940#issuecomment-110187847). 

This is my first experience with contributing to open-source software, so any pointers are appreciated. 

Other issues related to correlate functions include [ENH: Fold fftconvolve into convolve/correlate functions as a parameter #2651](https://github.com/scipy/scipy/issues/2651), [Use FFT in np.correlate/convolve? (Trac #1260) #1858](https://github.com/numpy/numpy/issues/1858), and [normalized cross-correlation (Trac #1714) #2310](https://github.com/numpy/numpy/issues/2310).

seberg:
Out of curiosity, how do matlab (maybe R?) handle rescaling effects due to different values when the max lag is not much smaller then the amount of data?
I think I often use `np.correlate(a, b[-max_lag:max_lag], mode='valid')`, to use always the same amount of data (and as much of the same value as possible). Maybe I am just crazy here, but wondering.
Not sure I opened an issue, but I think some function hidden in matplotlib had this kind of thing, and when maxlag got larger, I was annoyed by rescaling effects IIRC.

bringingheavendown:
Not sure what you mean by rescaling effects.  

[...]

PR #31261, opened 2026-04-16
ENH: add max_lag and lags parameters to correlate/convolve

HoniSanders:
## PR summary
This PR adds `max_lag` and `lags` keyword parameters to `np.correlate()` and `np.convolve()`, plus a companion function `correlation_lags()` that returns the corresponding lag indices for the relevant mode/lags. 

What was troubling me is that `numpy.correlate()` does not have a max_lag feature. This means that even if I only want to see correlations between two time series with lags between -100 and +100 ms, for example, it will still calculate the correlation for every lag between -20000 and +20000 ms (which is the length of the time series).  

This PR revives #5978, which I opened back in 2015 and never finished pushing through. I had introduced this question as issues on numpy and scipy: https://github.com/numpy/numpy/issues/5954, https://github.com/scipy/scipy/issues/4940 and on the [scipy-dev list](http://mail.scipy.org/pipermail/scipy-dev/2015-June/020757.html). It also got attention on [stackoverflow](https://stackoverflow.com/questions/30677241/how-to-limit-cross-correlation-window-width-in-numpy/47893831#47893831) at the time.  The new issue that this resolves is https://github.com/numpy/numpy/issues/17286.  

Apologies for the very long downtime. I'm excited to be able to come back to this and finally get this much-needed feature included.  
Also, I plan on squashing the commits in this branch and naming the single commit to be the same as the name of this PR.  I am expecting the review process to require more commits so pushed my branch as is.  I hope this is ok.  

### Proposed API

```python
np.correlate(a, v, mode=..., *, max_lag=None, lags=None)
np.convolve (a, v, mode=..., *, max_lag=None, lags=None)
np.correlation_lags(a_len, v_len, mode=..., *, max_lag=None, lags=None)
```

#### Parameter Design

- **max_lag=M** (int): symmetric inclusive window [-M, M] (2M+1 lags). Matches MATLAB's xcorr(x, y, M) convention.
- **lags=**: a range, a slice with explicit start/stop, or a 1-D integer array_like containing an arithmetic progression.

Both keyword-only, mutually exclusive. Default mode auto-resolves to 'lags' when either is supplied.

### Major file changes

- **Python** (numpy/_core/numeric.py): new max_lag/lags kwargs on `correlate` and `convolve`; helper functions `_lags_from_max_lag`, `_lags_from_lags`, `_lags_from_mode`, companion function `correlation_lags`.
- **C** (numpy/_core/src/multiarray/multiarraymodule.c): new `array_correlate_lags` private C function; _pyarray_correlate implements non-mode lags and is now the single normalization site (handles array swap, negative-step normalization, output reversal internally — callers pass any valid form). 

### Benchmark

For a user interested in a small number of lags around 0, there is a huge speedup:

| size1 | size2 | full + slice | `max_lag=5` | speedup |
|---|---|---|---|---|
| 1000 | 100 | 47.2 µs | 11.1 µs | 4.3× |
| 1000 | 1000 | 572 µs | 20.2 µs | 28× |
| 100000 | 100 | 4.09 ms | 14.0 µs | **290×** |
| 100000 | 1000 | 60.0 ms | 23.4 µs | **2500×** |

(From `benchmarks/benchmarks/bench_core.py::CorrConvLags`, included in this PR.)

## First time committer introduction
Hi!  I have used NumPy for more than a decade now, both directly, and as a key supporter of the whole scientific python ecosystem (e.g. pandas, pytorch, etc).  When I started working on this in 2015, I was only using NumPy for the first time, at the time to build simulations of biological neural networks that were more performant and open source than Matlab, which I had previously been using.  This was a feature that would have helped me quite a bit at the time, as I was cross-correlating very long time series, but only needed a relatively small window of time lags around 0.  I have wanted to come back to it because of the continued interest of others even though I myself no longer need this specific functionality.  

## AI Disclosure
Claude Code is what finally allowed me to finally bring this PR to completion.  

I used Claude as a pair-programming assistant for: 
- chasing down bugs (in particular it helped me find a buffer overflow that the earlier PR attempt had introduced in _pyarray_correlate); 
- proposing a more intuitive argument list for the Python API that I approved; 
- refactoring the vector inversion logic so the C function would be the single normalization site;
- proposing a more complete structure of the test cases (geometry/equivalence/dtype groupings) -- the expected values in the tests were computed by hand or by running and sub-slicing the pre-existing implementation. 
- writing some of the docstrings (Parameters/Returns wording)


ngoldbaum:
Thanks for the PR, I just enabled CI. If it fails you may need to enable CI on your fork to get CI passing in a timely manner.

I don't have time to go through the backlog of discussion about this and am not registering an opinion about whether we want to add this or not. That said, generally NumPy is pretty conservative about adding new keywords and functions, since the API surface is already very large and complicated.

One of the rules we have is that new features require sending a message to the numpy-discussion mailing list with a proposal to add the new feature. That gives the wider contributor base who don't actively follow github a chance to weigh in.

We're also generally conservative about adding new things to the C API, so your proposal to add a new C API item needs some justification. Keep in mind that basically no one will be able to use the NumPy 2.6 C API until perhaps 2028 due to projects keeping support for older NumPy releases. Is it really worth it to add the new C API? You'll need to explain why we should add the function.

Would you mind sending such a message to the mailing list about this proposal?

HoniSanders:
Thanks.  I did post about this on the numpy-discussion list at the time (see https://mail.python.org/archives/list/numpy-discussion@python.org/thread/JMXNOCFQRFJYISFKSRI7MRL7GPK5ZD4S/#JMXNOCFQRFJYISFKSRI7MRL7GPK5ZD4S and especially https://mail.python.org/archives/list/numpy-discussion@python.org/thread/6FC27XPOD4SBH3KJ7VY5DDAPS5Q46GPO/#6FC27XPOD4SBH3KJ7VY5DDAPS5Q46GPO where several people chimed in saying this would be useful).  I figured that was sufficient, but can re-post since it's been a while.  You can also see several people hopping in to support the feature on the previous PR and closed issue.

As for the C API question, I believe it needs a new C API function because the core vector calculation implementation needs to be in C and the existing arguments are not sufficient to support the new implementation.   Feel free to let me know if there is another way for the python function to call a modified version of the existing C functions.

ngoldbaum:
> I figured that was sufficient, but can re-post since it's been a while. You can also see several people hopping in to support the feature on the previous PR and closed issue.

Ah, I didn't search the archives.  I'll take a look at those discussions and let you know if I think a new mailing list ping is warranted.

> As for the C API question, I believe it needs a new C API function because the core vector calculation implementation needs to be in C and the existing arguments are not sufficient to support the new implementation. 

But why does the new function need to be publicly exposed in the C API? Can't NumPy just call an internal C function?

> Feel free to let me know if there is another way for the python function to call a modified version of the existing C functions.

You could not expose it publicly and instead just use the new C implementation internally.

HoniSanders:
oh sure, I didn't realize limiting it to internal use was an option.  I don't know who uses the public C API.  I guess the value of having the public C API function would just be to expose the functionality to whoever uses the C API as long as we are building the functionality anyway.  But I certainly wouldn't want the question of the C API to delay the inclusion of the functionality in the python API

ngoldbaum:
Oh great, good it was just a misconception. We definitely add features to the Python API without adding equivalent features in the public C API.

ngoldbaum:
I just took a look and given that it’s been more than ten years since the mailing list discussion, I do think a new ping is warranted. When you send your message, please include context about prior discussions and how this proposal differs from earlier proposals, if it does.

Also, I haven’t checked but it would also be good to see what other array libraries do and what the array API standard says about this case. If adding this makes us more conformant with the array API or otherwise makes it easier to write code that is agnostic to the underlying array API, that would also argue towards including this.

However, if this is an entirely new feature that only NumPy would implement, that makes it a harder sell.

Also please feel free to come to a NumPy community meeting to chat face-to-face about this over Zoom:

https://scientific-python.org/calendars/

HoniSanders:
Ok, I will send an email to the list, and I will try to attend this week's zoom.  

What do you mean by "other array libraries" and "array API standard"?

rkern:
[Array API Standard](https://data-apis.org/array-api/latest/). In particular, note the [Stakeholders](https://data-apis.org/array-api/latest/purpose_and_scope.html#stakeholders) which are those other array libraries.

`convolve` and `correlate` are not currently part of the standard, but those other libraries often do implement numpy functions that aren't in the standard yet. Changes to our implementation do have a "blast radius" in that it will either cause those libraries to do more work to follow numpy's choices or suffer greater divergence from numpy.

You may want to consider the option to implement more specialized functions in `scipy.signal` instead of modifying `np.convolve` and `np.correlate`. In particular, you could avoid needing `returns_lagvector`: we do not want to use this pattern of optional returns if we can at all avoid it. We will go to great lengths to avoid it.

HoniSanders:
Thanks for the pointer.  
As you mention, convolve and correlate are not part of that standard, so it shouldn't break anything.  The change I propose is backwards compatible so it will only cause more work for other libraries if they want to include this additional feature (which uses optional arguments).  In that case it will be less work for them, as they only have to pass through instead of implementing from scratch themselves.  I didn't find array libraries with conflicting implementations.  
I found several related implementation ideas:
-  Neural network libraries like pytorch and mxnet rely on a stride parameter for their convolution functions to avoid calculating convolutions at every lag (e.g. https://docs.pytorch.org/docs/stable/generated/torch.nn.modules.conv.Conv1d.html), similar to the stride entry in the range construction for the lags parameter proposed here.  
- Matplotlib already has a `maxlags` argument on its `xcorr` function:.  https://github.com/matplotlib/matplotlib/blob/dde076379dad6a51374625dbaa729a71958d4d88/lib/matplotlib/axes/_axes.pyi#L218.  It actually calls numpy.correlate('full') and just throws out the extra entries, for a huge amount of wasted calculations.  Numpy implementing this PR would allow matplotlib to immediately reap the performance benefits on their existing function parameters. 
- Scipy has a `correlation_lags` function to calculate the lags for a given correlation mode after the fact, similar to the returns_lagvector optional argument here.  
- The scipy people seemed to want this to be implemented here first and then pulled in over there when this was discussed at the time (https://github.com/scipy/scipy/issues/4940#issuecomment-110172721)

If you want, we can remove the returns_lagvector argument.  That was a suggestion proposed in the first round of getting feedback on this PR ten years ago.  We could have a separate function to calculate lags like scipy has if you prefer.

HoniSanders:
I pushed a commit to remove the returns_lagvector argument and instead have a companion function like scipy has.

rkern:
> As you mention, convolve and correlate are not part of that standard, so it shouldn't break anything. The change I propose is backwards compatible so it will only cause more work for other libraries if they want to include this additional feature (which uses optional arguments).

That's not quite the logic at play here. It's not like a dependency (e.g. scipy depends on numpy, if a new optional argument is added in a backwards-compatible fashion, that's fine because all of the existing usages of `np.convolve()` inside of scipy still work). Adding a new optional argument would still affect any of these libraries that provide a `convolve()` or `correlate()` that attempt to follow numpy's current signature. That's because if numpy users use the new arguments, that code will not be portable to those array implementations by swapping out `np` for `xp` (the placeholder name for the other-array-library namespace). So those other implementations still face the choice to do the work to add the new feature or else live with the growing discrepancy. There's a cost either way, and there's nothing about the optionality or backwards compatibility of the new feature that alleviates that cost.

numpy/_core/numeric.pyi
```diff
@@ -1048,44 +1049,88 @@ def flatnonzero(a: ArrayLike) -> _Array1D[np.intp]: ...
...
 
+type _LagsArg = int | tuple[int, int] | tuple[int, int, int] | range | slice | ArrayLike | None
+
 # keep in sync with `convolve` and `ma.core.correlate`
 @overload
 def correlate(  # noqa: UP047
-    a: _ArrayLike1D[_AnyNumericScalarT], v: _ArrayLike1D[_AnyNumericScalarT], mode: _CorrelateMode = "valid"
+    a: _ArrayLike1D[_AnyNumericScalarT], v: _ArrayLike1D[_AnyNumericScalarT],
+    mode: _CorrelateMode | L["lags"] = ..., *, maxlag: int | None = ...,
+    lags: _LagsArg = ...,
```
jorenham: nit: I think it'd help make it more readable if we'd place the params on separate lines

numpy/_core/numeric.pyi
```diff
@@ -1048,44 +1049,88 @@ def flatnonzero(a: ArrayLike) -> _Array1D[np.intp]: ...
 
 # NOTE: we ignore UP047 because inlining `_AnyScalarT` would result in a lot of code duplication
 
+type _LagsArg = int | tuple[int, int] | tuple[int, int, int] | range | slice | ArrayLike | None
```
jorenham: Can we use a more specific array-like type? `numpy._typing._ArrayLikeInt_co`, for example?

jorenham: nit: It can help to move the `| None` to the use-site, so that you can immediately see that it's the default.

numpy/_core/numeric.pyi
jorenham: And could you fill in the `...` defaults?

numpy/_core/numeric.py
```diff
@@ -713,12 +713,197 @@ def flatnonzero(a):
     return np.nonzero(np.ravel(a))[0]
 
 
-def _correlate_dispatcher(a, v, mode=None):
+_CorrModeDefault = object()
```
jorenham: We could use `numpy._globals._Novalue` instead

numpy/_core/numeric.py
```diff
@@ -713,12 +713,197 @@ def flatnonzero(a):
...
+    """
+    inverted = False
+    if alen < vlen:
+        alen, vlen = vlen, alen
+        inverted = True
+
+    if mode == 0:
+        m0, m1 = 0, alen - vlen + 1
+    elif mode == 1:
+        half = int(vlen / 2)
```
jorenham:
```suggestion
        half = vlen // 2
```

honi-at-simspace:
Thanks for the comments and the meeting today.  

With respect to downstream libraries wanting to match numpy's signatures, I realized that it would be trivial for other libraries to accept a new `lags/maxlag` keyword and just compute the same complete correlation they currently do and then just slice it at the end and return the sliced version.  That would keep the other libraries API consistent with numpy's very easily.  They can then decide later whether they want to actually implement the performant version of lag subsetting or not.  

Another point that came from the meeting today was with respect to interest in the feature.  The fact that matplotlib has a maxlag keyword in its cross correlation function (that just slices at the end without any performance improvement) is evidence that there is interest in this functionality, and the fact that they call numpy.correlate means that it would be extremely easy for them to update their call to immediately take benefit of the performance improvement.  
Additionally, I pointed out that there have been periodic comments on the previous closed PR expressing interest in the feature and asking when it would be implemented.  

tasks for me from the numpy meeting today:
- Remove PyArray_CorrelateLags from C API (done, see commits above)
- post on scipy https://discuss.scientific-python.org/c/contributor/12 to see what they think.  
- post on the Jax repo to see how they would feel about the numpy function signature changing and the lift necessary to match it.

HoniSanders:
Just a few other data points of other libraries having to do weird workarounds to deal with numpy.correlate being extremely inefficient for large arrays where only some lag values are relevant:
- https://github.com/bids-standard/pybids/pull/411 pybids downsampled (i.e. threw out some of) their source data in order to hack numpy.correlate into calculating fewer lags.  
- https://github.com/statsmodels/statsmodels/issues/4135 in statsmodels is solved with another [weird workaround](https://github.com/statsmodels/statsmodels/pull/4937/changes#diff-771ed9ae3d316dec0997427f2182d015dbfdaf6935335428bbf177ce1968bc72)

ngoldbaum:
Thanks! I think the statsmodels and matplotlib examples are probably sufficient to justify adding a `maxlags` keyword argument (following matplotlib).

@bashtage hope you don't mind the ping here, in case you want to weigh in from the statsmodels side.

I'd still like to get opinions from SciPy and Jax as well.

charris:
This is a common need, I am in favor.

honi-at-simspace:
Great.  I posted on scipy here: https://discuss.scientific-python.org/t/custom-lag-parameters-on-correlate-and-convolve-functions/2307

and jax here: https://github.com/jax-ml/jax/issues/37112. 

i also posted on matplotlib, and they said that they are not concerned with consistency with numpy on that function.  https://discourse.matplotlib.org/t/custom-lag-parameters-on-numpy-correlate-and-convolve-functions/26226

honi-at-simspace:
Does anyone have contacts at jax or scipy?  No one responded to either post.  

> I posted on scipy here: https://discuss.scientific-python.org/t/custom-lag-parameters-on-correlate-and-convolve-functions/2307

> and jax here: https://github.com/jax-ml/jax/issues/37112.

seberg:
@mhvk do you think this is useful for astropy?  I am a bit doubtful about common enough the use-cases at least for `lags` that shouldn't be using the FFT trick to begin with.
`maxlags` had the argument that matplotlib has it/uses it, which seems relevant (but should check with them).

mhvk:
@seberg - I don't know that it would be much use for astropy (the main reason for our `convolve` is to deal with missing data...), but I do think I would use this myself. On the keywords, I think you are right that the case for `maxlag` is stronger, as almost always one will want symmetric lags (and steps other than 1 would be rarer still). 

That said, a great advantage of being able to pass in lags is that you have those lags at hand right after, if, e.g., you want to plot the result. Note that I'm not in favour for introducing a dedicated function to return the lags, but it does point to a real problem, that it is not always all that obvious what the lags actually are. Obviously, this can (and should!) be solved by mentioning it in the docstring, but, heck, if you pass in `lags`, you don't have to worry about it at all. I do think that if one is allowed to pass in `lags`, they should just be any array of int; there is no reason to limit it to a slice.

Anyway, not sure whether or not to go with one, the other, or both...

Implementation-wise, I like that the main function in C now just takes lag information, but wonder if it isn't better to do the creation of that information from the keywords in python. I'd also tend not to change the C API interface at all, just change an internal function and call that from python. (I didn't even know the C version existed, and would agree with the random website I ran into when I googled it; https://runebook.dev/en/docs/numpy/reference/c-api/array/c.PyArray_Correlate2).

HoniSanders:
@ngoldbaum, can you approve the test workflows?

HoniSanders:
At the numpy meeting yesterday, it was decided to move forward as I had put sufficient effort into reaching out to other projects and the feature itself seemed useful.  

Design questions raised:
- I'd argue for including both `lags` and `max_lag`.  `max_lag` gives an easy entry for someone who wants a symmetric window around 0, which has been named as a useful case.  `lags` optionally gives additional control in case someone doesn't need the same precision on the output as on the input (lag stride >1) or wants a window centered around something other than 0.  Both are already implemented, so it seems more fully featured and less effort to include both.  
- After thinking about it, I'd argue for `max_lag` as the argument name.  It conforms to Python's snake case convention for compound names, which is something that numpy adheres to for arguments starting with max (e.g. [max_header_size](https://github.com/numpy/numpy/blob/4d711763f8c5b978abd88cea710147c14d802675/numpy/lib/_npyio_impl.py#L313) [max_rows](https://github.com/numpy/numpy/blob/4d711763f8c5b978abd88cea710147c14d802675/numpy/_core/src/multiarray/textreading/rows.c#L168) and [max_work](https://github.com/numpy/numpy/blob/4d711763f8c5b978abd88cea710147c14d802675/numpy/_core/multiarray.py#L1322)).  Numpy never uses max* without an underscore.  It makes sense for `max_lag` to be in singular since it is an int value, not a list.  This would parallel `np.linspace()` which uses `num` in the singular for an int argument that tells the function how many indices to generate.  Other packages are inconsistent in their naming and use `max_lag` (pmdarima) `maxlag` (Matlab) `maxlags` (matplotlib) nlags (statsmodels).

ngoldbaum:
I think the linter and stubtest failures are real.

honi-at-simspace:
@ngoldbaum, I think I fixed those errors.  can you re-approve the workflows?

HoniSanders:
Hi all, what are next steps? Is this ready to merge?

ngoldbaum:
@HoniSanders we just had a triage meeting where this PR came up. I'm going to do a final once-over today and either merge or request one last round of fixes. So far I haven't heard any pushback on this so I think it's ok to merge as far as process goes.

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -1132,26 +1132,45 @@ PyArray_MatrixProduct2(PyObject *op1, PyObject *op2, PyArrayObject* out)
...
+    npy_intp lag, tmplag, maxleft, maxright, phase2_end;
     npy_intp is1, is2, os;
     char *ip1, *ip2, *op;
     PyArray_DotFunc *dot;
+    int inverted;
 
     NPY_BEGIN_THREADS_DEF;
 
+    if (lag_step == 0) {
+        lag_step = 1;
```
ngoldbaum: should probably be  ValueError?

numpy/__init__.py
```diff
@@ -203,6 +203,7 @@
         copysign,
         copyto,
         correlate,
+        correlation_lags,
```
ngoldbaum: I'm confused, I thought we were adding a new keyword argument to `correlate`, not adding a whole new function to the NumPy API. I don't think we had any consensus to add a new function.

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -1213,40 +1239,69 @@ _pyarray_correlate(PyArrayObject *ap1, PyArrayObject *ap2, PyArray_Descr *typec,
...
-        dot(ip1, is1, ip2, is2, op, n, ret);
+
+    lag = min_lag;
+    if (lag < -n2+1) {
+        /* if min_lag is before any overlap between the vectors,
+         * then skip to first relevant lag
+         */
+        op += os*((-n2+1) - lag);
+        lag = -n2+1;
+    }
```
ngoldbaum, line 1256: Doesn't the amount to skip by assume `lag_step == 1`? I think that means e.g. this example:

```
a = np.arange(100, dtype=float)
v = np.arange(3, dtype=float)

result = np.convolve(a, v, lags=range(-50, 50, 3))
```

Will produce the wrong result. I get an array filled with zeros on this PR, which I think is wrong.

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -1310,19 +1373,80 @@ _pyarray_revert(PyArrayObject *ret)
...
+    default:
+        PyErr_SetString(PyExc_ValueError, "mode must be 0, 1, or 2");
+        return;
+    }
+
+    /* If we swapped to get n1 >= n2, transform the lags back to the
+     * caller's original orientation. */
+    if (inverted) {
+        i = m0;
+        i1 = (npy_intp)(npy_ceil((m1 - m0)/(float)s))*s;
```
ngoldbaum: Why are you casting `s` to float? That will lead to precision loss.

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -1162,48 +1181,55 @@ _pyarray_correlate(PyArrayObject *ap1, PyArrayObject *ap2, PyArray_Descr *typec,
...
-        break;
-    default:
-        PyErr_SetString(PyExc_ValueError, "mode must be 0, 1, or 2");
-        return NULL;
+    /* Normalize negative lag_step: reverse the range to use positive step.
+     * The output will need to be reversed before returning. */
+    if (lag_step < 0) {
+        inverted = 1;
+        i = min_lag;
+        i1 = (npy_intp)(npy_ceil((max_lag - min_lag)/(float)lag_step))*lag_step;
```
ngoldbaum: same here, why cast to float?

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -3063,6 +3200,105 @@ array_correlate2(PyObject *NPY_UNUSED(dummy),
...
+            {"v", NULL, &op2},
+            {"min_lag", &PyArray_IntpFromPyIntConverter, &min_lag},
+            {"max_lag", &PyArray_IntpFromPyIntConverter, &max_lag},
+            {"lag_step", &PyArray_IntpFromPyIntConverter, &lag_step},
+            {"|conjugate", &PyArray_BoolConverter, &conjugate}) < 0) {
+        return NULL;
+    }
+    if (min_lag == 0 && max_lag == 0 && lag_step == 0) {
+        /* if no lag parameters passed, use default: mode = 'valid' */
+        return PyArray_Correlate2(op1, op2, 0);
```
ngoldbaum: This seems fragile. Why not make this an error?

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -1213,40 +1239,69 @@ _pyarray_correlate(PyArrayObject *ap1, PyArrayObject *ap2, PyArray_Descr *typec,
...
+         */
+        op += os*((-n2+1) - lag);
+        lag = -n2+1;
+    }
+
+    /* Phase 1: lags where y is left of x, i.e. lag in [-(n2-1), 0).
+     * Overlap length is n2 + lag. */
+    maxleft = (0 < max_lag ? 0 : max_lag);
+    tmplag = lag;
+    for (lag = tmplag; lag < maxleft; lag+=lag_step) {
```
ngoldbaum:
```suggestion
    for (; lag < maxleft; lag+=lag_step) {
```

This pattern occurs in a few spots elsewhere and can be rewritten like this without defining an unnecessary temporary.

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -1213,40 +1239,69 @@ _pyarray_correlate(PyArrayObject *ap1, PyArrayObject *ap2, PyArray_Descr *typec,
...
-    }
-    else {
-        for (i = 0; i < (n1 - n2 + 1) && (!needs_pyapi || !PyErr_Occurred());
-             i++) {
-            dot(ip1, is1, ip2, is2, op, n, ret);
-            ip1 += is1;
+
+    /* Phase 2: lags where y entirely overlaps with x, i.e. lag in [0, n1-n2].
+     * Upper bound is the smaller of max_lag and the full-overlap boundary. */
+    n11 = n1;
```
ngoldbaum: why define this `n11` variable? Just use `n1`.

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -1213,40 +1239,69 @@ _pyarray_correlate(PyArrayObject *ap1, PyArrayObject *ap2, PyArray_Descr *typec,
 
     int needs_pyapi = PyDataType_FLAGCHK(PyArray_DESCR(ret), NPY_NEEDS_PYAPI);
     NPY_BEGIN_THREADS_DESCR(PyArray_DESCR(ret));
+    /* p is pointer to array start and s is stride */
+    /* i1 is x, i2, is y, o is answer */
```
ngoldbaum: The variables in this comment don't show up below. Remnant of an earlier draft?

numpy/_core/numeric.py
```diff
@@ -734,9 +919,21 @@ def correlate(a, v, mode='valid'):
...
+    max_lag : int, optional
+        Compute the cross-correlation at lags
+        ``-max_lag, -max_lag+1, ..., max_lag`` (a symmetric inclusive window
+        of ``2*max_lag+1`` lags around 0).
+        Mutually exclusive with ``lags``.
+    lags : range, slice, or 1-D array_like of int, optional
+        Explicit lag specification.  Accepts a Python ``range`` object, a
+        ``slice`` with explicit start and stop, or a 1-D array_like
+        containing an arithmetic progression of integer lag indices.
+        Mutually exclusive with ``max_lag``.
```
ngoldbaum: I thought we agreed on only adding one new keyword argument. Why add two?

numpy/_core/numeric.py
```diff
@@ -790,15 +997,39 @@ def correlate(a, v, mode='valid'):
...
+            a, v, lags_tuple[0], lags_tuple[1], lags_tuple[2])
+
+
+def _convolve_dispatcher(a, v, mode=None, *, max_lag=None, lags=None):
     return (a, v)
 
 
 @array_function_dispatch(_convolve_dispatcher)
-def convolve(a, v, mode='full'):
+def convolve(a, v, mode=_NoValue, *, max_lag=None, lags=None):
```
ngoldbaum: Why the change away from `_NoValue` for the default `mode`?

ngoldbaum (changes requested):
I spotted some significant issues (see below). I also don't think this PR agress with what we discussed at the community meeting: a single new keyword argument for `np.correlate` and `np.convolve`. This adds a new function and adds *two* new keyword arguments.

I know I said earlier that I was close to merging this but I'm no longer confident that we're on the same page on what you actually want to land, so I no longer think that.

numpy/__init__.py
HoniSanders: sorry for the confusion.  This is a companion function, not the function that calculates the cross-correlation.  The role of this companion function is to return the corresponding lag indices for the relevant mode/lags arguments.  This is based on the scipy companion function (https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.correlation_lags.html).  The reason I included this companion function is because @rkern did not like having a `return_lagvector` argument so I made it into a companion function instead.  If you want this removed altogether,  I can do that.

numpy/_core/src/multiarray/multiarraymodule.c
HoniSanders: it is a ValueError in _lags_from_lags in numeric.py.  So we can remove this

numpy/_core/src/multiarray/multiarraymodule.c
HoniSanders: wow, good eye!  i will add a new test for that when i fix that.

numpy/_core/src/multiarray/multiarraymodule.c
HoniSanders: good point, i will switch to integer ceiling division with no casting. that was a pattern i brought from python programming, not knowing better.

numpy/_core/src/multiarray/multiarraymodule.c
HoniSanders: just copying the pattern from the other spot.  lag_step is always 1 for all of these modes, so we can simplify the expression.

numpy/_core/src/multiarray/multiarraymodule.c
HoniSanders: ok

numpy/_core/src/multiarray/multiarraymodule.c
HoniSanders: yes, this was me making a note to myself as I was learning how the previous code worked.

numpy/_core/numeric.py
HoniSanders: This is a change *to* _NoValue for the default mode as suggested by @jorenham.  The reason not to have an explicit default value is that the default mode is different if a lag argument is passed (i.e. `mode=lags`) than if no lag arguments is passed (i.e. `mode='full'`)

numpy/_core/numeric.py
HoniSanders: I thought this was still a discussion that was undecided.  I gave my reasoning for including both [above](https://github.com/numpy/numpy/pull/31261#issuecomment-4402481626)
> I'd argue for including both lags and max_lag.  max_lag gives an easy entry for someone who wants a symmetric window around 0, which has been named as a useful case.  lags optionally gives additional control in case someone doesn't need the same temporal precision on the output as on the input for faster performance (lag stride >1) or wants a window centered around something other than 0 (e.g. if the temporal region of interest isn't at 0 lag). Both are already implemented, so it seems more fully featured and less effort to include both.

numpy/_core/src/multiarray/multiarraymodule.c
HoniSanders: This was from a previous iteration of the code when correlatelags was in the C API, so I had to check against invalid required parameters.  Now that it is not in the C API anymore, there's no situation where this would happen.  Without the check here, it will be caught in _pyarray_correlate which raises an error if lag_step=0.  So I will remove this.

numpy/_core/src/multiarray/multiarraymodule.c
HoniSanders: ok.  didn't know that one!

HoniSanders:
Thank you for your detailed review, @ngoldbaum! 

I responded to your comments.

In particular, I am open to removing `correlation_lags` if you think it clutters up the API too much.  I would argue more strongly for including both `max_lag` and `lags` arguments, as they serve different (real) purposes.

numpy/_core/numeric.py
seberg: I agree with Nathan. Yes, we need to change the default, but `mode=None` should be a perfectly good new default?
`_NoValue` serves a purpose for places where we can't use `None`, but in this case it's OK to support users passing `mode=None, max_lags=10` and we have no other reasons to have to worry about it here, I think.

We also don't need an explicit `mode='lags'` at all, since that is implicit by passing `max_lags=` already.

We need to document that `None` is the same as 'full' unless `max_lag` is passed.  There should also be an error raised if `max_lag` is passed and `mode is not None` (already the cas).

seberg:
> In particular, I am open to removing correlation_lags if you think it clutters up the API too much. I would argue more strongly for including both max_lag and lags arguments, as they serve different (real) purposes.

Please remove it and also please let's just drop `lags` for now.  We understand that you are in favor of both, but the situation is that we are pretty confident that `max_lag` seems useful enough and less so about the rest.  NumPy 2.5 will be branched very soon, so if we are going to discuss adding `lags` here all that is likely to happen at this point is that *both* miss NumPy 2.5.

numpy/_core/numeric.py
```diff
@@ -816,30 +1047,50 @@ def convolve(a, v, mode='full'):
...
+          the signal boundary have no effect. This corresponds with a lag tuple
+          of (0, N-M+1, 1) for N>M or (-M+N, 1, 1) for M>N.
+
+        'lags':
+          Mode 'lags' uses ``max_lag`` or ``lags`` to define the lags for
+          which to perform the convolution.
+          If ``max_lag`` or ``lags`` is provided and ``mode`` is not
+          specified, ``mode`` defaults to 'lags'.
+
+    max_lag : int, optional
```
bashtage: Is `max_lag` positive only?

numpy/_core/numeric.py
```diff
@@ -816,30 +1047,50 @@ def convolve(a, v, mode='full'):
...
+        Compute the convolution at lags ``-max_lag, -max_lag+1, ..., max_lag``
+        (a symmetric inclusive window of ``2*max_lag+1`` lags around 0).
+        Mutually exclusive with ``lags``.
+    lags : range, slice, or 1-D array_like of int, optional
```
bashtage: Why not just let `lags` accept a single int and behave like `max_lag`? Mutually exclusive inputs always strike me as challenging.

numpy/_core/numeric.pyi
```diff
@@ -1048,44 +1050,127 @@ def flatnonzero(a: ArrayLike) -> _Array1D[np.intp]: ...
...
+
 # keep in sync with `convolve` and `ma.core.correlate`
 @overload
 def correlate(  # noqa: UP047
-    a: _ArrayLike1D[_AnyNumericScalarT], v: _ArrayLike1D[_AnyNumericScalarT], mode: _CorrelateMode = "valid"
+    a: _ArrayLike1D[_AnyNumericScalarT],
+    v: _ArrayLike1D[_AnyNumericScalarT],
+    mode: _CorrelateMode | L["lags"] | _NoValueType = ...,
+    *,
+    max_lag: int | None = None,
```
bashtage: Are these type defs correct since `max_lag` and `lags` are mutually exclusive.  Shouldn't it be the last if `lags` is `_LagsArg` when `max_lag` must be either absent or `None`?

bashtage (commented):
The API changes feel overly complex to me, and I don't see much that would avoid adding a second keyword argument.

numpy/_core/numeric.py
mhvk: That would be my preference as well, allow `lags` to take a single int to behave like `max_lag`. I would not allow the slice and range just `int or 1-D array_like of int`.

To me, this is also the argument against having just `max_lag` - the name forces adding another keyword argument if one would like to have explicit lags later.

seberg: Hmmmm, that's a good point, but int hat case you are arguing strongly for just doing everything right now?

bashtage: I think avoiding an API expansion and then future change is best path.

seberg: Well, if you would want `lags=10` to mean what is now `max_lag=10` then clearly it's not nice to overload.  FWIW, the _other_ meaning for `lags=10` is to just return a *scalar* with the lag 10.
(I haven't quite made up my mind if I prefer the `lags=` overload, but I don't think I'll get to a strong opinion either way.)

bashtage: The second would be covered by `[10]`, while 10 would be `np.arange(-10,11,)`.

bashtage: I am probably biased because this is a pattern used in statsmodels for lags in statistical time series models, e.g, something like `ARIMA(y, lags=4)` (incl. 1,2,3,4) vs `ARIMA(y, lags=[1, 4])` (incl only 1 and 4)

HoniSanders: The funny thing is, this overload was my original design in the original PR, but somebody didn't like overloading arguments.  I would be glad to go with that.

HoniSanders:
It sounds like there is a preference for a single argument lags that can take an int or a slice (or possibly an arbitrary list of indices).  The argument for this is that it is extensible in the future without API changes. I can try to re-implement this with this pattern. I'm not sure I'm going to be able to get all the way to allowing arbitrary lists of indices, but I can certainly get to a single argument that can accept an int or a slice.

numpy/_core/numeric.py
honi-at-simspace: yes, that is verified in _lags_from_lags: Raise ValueError("lags integer must be non-negative").  I added a new test to test that requirement as well

honi-at-simspace:
I think I addressed all the review comments.  if there is a strong interest in accepting arbitrary lists, here is a possible approach:

```python
# _lags_from_lags: arithmetic progressions still take the fast tuple path;
# non-arithmetic arrays fall through and return the ndarray itself
step = int(arr[1] - arr[0])
if step != 0:
    expected = arr[0] + step * arange(arr.size)
    if (arr == expected).all():
        return (int(arr[0]), int(arr[-1]) + step, step)
return arr  # arbitrary — caller loops
```

```python
# in correlate / convolve
lags_spec = _lags_from_lags(lags)
if isinstance(lags_spec, tuple):
    return multiarray.correlatelags(a, v, *lags_spec)
return concatenate([
    multiarray.correlatelags(a, v, int(lag), int(lag) + 1, 1)
    for lag in lags_spec
])
```

This lets users do things like `np.correlate(a, v, lags=np.array([0, 1, 3, 7]))`. Arithmetic progressions (including non-unit step) still go through a single C call; only truly non-arithmetic arrays pay N calls.

stefanv:
Not to rehash the conversation above, but I am curious about the complexity implications of supporting only `lags=int` (`max_lags`) vs arbitrary `lags=[...]`. It seems like the prior should be a much simpler implementation. My thinking is, in terms of the API, that the max-lags idea is what most people will need (to limit computation), but once we've added `lags` (and its accompanying complexity) there's no going back.

honi-at-simspace:
Ok, so at the numpy community meeting, there was concern about supporting complexity.  

So, the suggestion is to use a `lags` argument name (that would allow for expanding functionality later by overload) but to limit the argument type to `int` and have the functionality be the same as the original `max_lag` argument.  

I will work on implementing this.

honi-at-simspace:
I have now limited np.correlate(lags=) to only accept ints

HoniSanders:
I realized one more example of a use case of non-unit strides.  Neural network convolution frequently uses non-unit strides, e.g. https://docs.pytorch.org/docs/stable/generated/torch.nn.modules.conv.Conv1d.html.  

But I don't need to re-open that discussion, and would be happy for this to be included as is. Just wanted to mention it because it occurred to me and seemed relevant.

HoniSanders:
@ngoldbaum @seberg @jorenham politely bumping this thread. Looking for a review. Thank you for your time.

numpy/_core/numeric.pyi
```diff
@@ -1048,44 +1048,106 @@ def flatnonzero(a: ArrayLike) -> _Array1D[np.intp]: ...
 
 # NOTE: we ignore UP047 because inlining `_AnyScalarT` would result in a lot of code duplication
 
+type _LagsArg = int | tuple[int, int] | tuple[int, int, int] | range | slice | _ArrayLikeInt_co
```
jorenham: `int` and `tuple[int, ...]` is already coveraged by `_ArrayLikeInt_co`, so they can be removed here (or if that wasn't intended, then `_ArrayLikeInt_co` should be replaced by a type that's a bit more narrow)

numpy/_core/numeric.pyi
```diff
@@ -1048,44 +1048,106 @@ def flatnonzero(a: ArrayLike) -> _Array1D[np.intp]: ...
...
+
 # keep in sync with `convolve` and `ma.core.correlate`
 @overload
 def correlate(  # noqa: UP047
-    a: _ArrayLike1D[_AnyNumericScalarT], v: _ArrayLike1D[_AnyNumericScalarT], mode: _CorrelateMode = "valid"
+    a: _ArrayLike1D[_AnyNumericScalarT],
+    v: _ArrayLike1D[_AnyNumericScalarT],
+    mode: _CorrelateMode | None = ...,
```
jorenham: Could you fill in the `= ...` here, and in the overloads below?

numpy/_core/numeric.pyi
HoniSanders: when i updated the lag function to only accept int arguments as requested above, i didn't update this signature.  I now updated it to only accept int and np.int.  thanks for catching that.

numpy/_core/numeric.pyi
HoniSanders: done

numpy/_core/numeric.py
```diff
@@ -78,7 +78,8 @@
     'argwhere', 'copyto', 'concatenate', 'lexsort', 'astype',
     'can_cast', 'promote_types', 'min_scalar_type',
     'result_type', 'isfortran', 'empty_like', 'zeros_like', 'ones_like',
-    'correlate', 'convolve', 'inner', 'dot', 'outer', 'vdot', 'roll',
+    'correlate', 'convolve', 'inner', 'dot',
+    'outer', 'vdot', 'roll',
```
jorenham: seems unnecessary

HoniSanders: fixed

HoniSanders:
I did a more thorough re-review since the issues @jorenham was bringing up were due to code changes being left over after their causes were reverted.  I have now fixed references to max_lag and lag tuples left over from the previous version that accepted more types of inputs.

PR #31264, opened 2026-04-17
ENH: Allow user-dtypes return views on `.conjugate()`

seberg:
This uses the fact that if something _doesn't_ define an imag, then `conjugate()` should be defined as a simple view (similar to `.real`).

This is a follow-up of a previous PR, where I didn't implement this isn't really backportable.

With this, QuadDType for example does return a view (and a complex new-style dtype can opt out this).

This _doesn't_ yet support for `.conjugate()` to return a view for complex numbers (which would make a lot of sense and could be added but take a bit more effort).

CC @SwayamInSync in case you have time to do the main review maybe, follow up to that other PR (but not backportable, IMO -- I really don't mind adding that `.conj()` could be a view eventually).

### DIFF
--- a/numpy/_core/src/multiarray/calculation.c
+++ b/numpy/_core/src/multiarray/calculation.c
@@ -841,7 +841,7 @@ PyArray_Conjugate(PyArrayObject *self, PyArrayObject *out)
         }
     }
     else {
-        if (!NPY_DT_is_numeric(NPY_DTYPE(PyArray_DESCR(self)))) {
+        if (!NPY_DT_is_numeric(PyArray_DTYPE(self))) {
             PyErr_SetString(PyExc_TypeError,
                             "cannot conjugate non-numeric dtype");
             return NULL;
--- a/numpy/_core/src/multiarray/calculation.c
+++ b/numpy/_core/src/multiarray/calculation.c
@@ -828,8 +828,8 @@ PyArray_Clip(PyArrayObject *self, PyObject *min, PyObject *max, PyArrayObject *o
 NPY_NO_EXPORT PyObject *
 PyArray_Conjugate(PyArrayObject *self, PyArrayObject *out)
 {
-    if (PyArray_ISCOMPLEX(self) || PyArray_ISOBJECT(self) ||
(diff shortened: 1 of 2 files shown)


numpy/_core/src/multiarray/calculation.c
```diff
@@ -841,12 +841,14 @@ PyArray_Conjugate(PyArrayObject *self, PyArrayObject *out)
         }
     }
     else {
-        PyArrayObject *ret;
-        if (!PyArray_ISNUMBER(self)) {
+        if (!NPY_DT_is_numeric(NPY_DTYPE(PyArray_DESCR(self)))) {
```
SwayamInSync: Just for consistency with the line above (831)
```suggestion
        if (!NPY_DT_is_numeric(PyArray_DTYPE(self))) {
```

SwayamInSync (commented):
LGTM, Thanks @seberg 
I agree making the ufunc path also return views might be good work, but at least with this PR user-defined behaviour will be consistent with builtins real types.

numpy/_core/src/multiarray/calculation.c
seberg: Good point (updated removing the unnecessary macro now)

seberg: Ah hah, thought that was the DType but it's not, so I guess it really doesn't matter except it's funny it diverges within the function. (This one is a static inline function, although I almost prefer the descr name in C to not confuse instance and class)

numpy/_core/src/multiarray/calculation.c
```diff
@@ -841,12 +841,14 @@ PyArray_Conjugate(PyArrayObject *self, PyArrayObject *out)
         }
     }
     else {
-        PyArrayObject *ret;
-        if (!PyArray_ISNUMBER(self)) {
+        if (!NPY_DT_is_numeric(PyArray_DTYPE(self))) {
```
seberg:
```suggestion
        if (!NPY_DT_is_numeric(NPY_DTYPE(PyArray_DTYPE(self)))) {
```

mhvk (approved):
Yes, makes sense, nice!

seberg:
Thanks for the reviews.

PR #31270, opened 2026-04-18
MAINT: Replace malloc/free with PyArray_malloc/free in buffer.c

DeineBruder:
### PR summary
This PR is an initial step toward addressing issue #28454 by phasing out direct malloc and free calls in the NumPy C core.

Using the system `malloc` bypasses Python's memory management and conflicts with Python's `mimalloc` allocator (default in Python 3.14+), which has been reported to cause segmentation faults and bus errors during initialization. This patch transitions the allocation to `PyArray_malloc`.

To avoid a massive, hard-to-review diff, I have constrained this initial PR to a proof-of-concept in `numpy/_core/src/multiarray/buffer.c`.

I compiled via `spin build`, then ran `spin test -t numpy/_core/tests`. My modifications didn't break anything (that wasn't already broken...)

If the maintainers agree with this approach, I will follow up by expanding this replacement pattern to the remaining files in `multiarray` and `common`.

I am a university student contributing to NumPy as part of an academic project focused on open-source development and system-level programming. I rely on NumPy for my coursework and wanted to tackle a low-level C maintenance task to better understand the library's internal architecture and memory management interface with CPython.

#### AI Disclosure
I used an LLM (Google Gemini) to help me in the installation, building and fork process, not in the thinking or coding part of the issue.

If there is anything I need to modify in this PR let me know.

### DIFF
--- a/numpy/_core/src/common/mem_overlap.c
+++ b/numpy/_core/src/common/mem_overlap.c
@@ -544,9 +544,9 @@ solve_diophantine(unsigned int n, diophantine_term_t *E, npy_int64 b,
         diophantine_term_t *Ep = NULL;
         npy_int64 *Epsilon = NULL, *Gamma = NULL;
 
-        Ep = malloc(n * sizeof(diophantine_term_t));
-        Epsilon = malloc(n * sizeof(npy_int64));
-        Gamma = malloc(n * sizeof(npy_int64));
+        Ep = PyMem_RawMalloc(n * sizeof(diophantine_term_t));
+        Epsilon = PyMem_RawMalloc(n * sizeof(npy_int64));
+        Gamma = PyMem_RawMalloc(n * sizeof(npy_int64));
         if (Ep == NULL || Epsilon == NULL || Gamma == NULL) {
             res = MEM_OVERLAP_ERROR;
         }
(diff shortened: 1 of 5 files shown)


ngoldbaum:
Hi! Thanks for looking at this.

I would personally rather get rid of `PyArray_malloc` and `PyArray_free` and instead just call `PyMem_RawMalloc` directly. Right now (since 2010!) we're defining some macros to add a wrapper name that is always just `PyMem_RawMalloc`.

I would also personally prefer if you did this refactor all in one PR. Big PRs that are focused on exactly one pattern are easier to review than a bunch of small PRs that fix the same pattern over and over again.

DeineBruder:
Hi @ngoldbaum, I have corrected what you pointed and continued the modification throughout the code.

There are still a few malloc in the code, those were cases in which I wasn't sure if we really want to change it (for instance simd or sorting algorithms).

Thanks for your review.

ngoldbaum (commented):
Sorry for taking a while to respond.

This wasn't quite what I was getting at. I wanted you to make a somewhat more invasive renaming in numpy internals. See https://github.com/DeineBruder/numpy/pull/1 for what I was trying to say.

ngoldbaum:
Closing in favor of #31503

PR #31271, opened 2026-04-19
MAINT: use new dtype in view creation rather than going through two views

mhvk:
EDITED since now follow-up rather than an alternative to #31234.

This PR lets `PyArray_View()` do a one-step view of both a possibly new subtype and a new dtype. To avoid duplicated checks on the dtype, I factored out that piece of the internal `_set_dtype`.

No AI was used.

### DIFF
--- a/numpy/_core/src/multiarray/common.c
+++ b/numpy/_core/src/multiarray/common.c
@@ -371,6 +371,103 @@ _may_have_objects(PyArray_Descr *dtype)
 }
 
 /*
+ * Check whether self can be viewed with the given dtype.
+ * If so, return a new reference to the dtype (possibly changed).
+ * If needed, also determine new dimensions and strides for the last axis.
+ * If no change is needed, newlastdim is set to -1.
+ */
+NPY_NO_EXPORT PyArray_Descr*
+_check_compatibility_with_new_dtype(
+    PyArrayObject *self, PyArray_Descr *type,
+    npy_intp *newlastdim, npy_intp *newlaststride)
+{
+    PyArray_Descr *dtype = PyArray_DESCR(self);
+    *newlastdim = -1;  /* By default, no change needed. */
+
+    /* Check that we are not reinterpreting memory containing Objects. */
+    if (_may_have_objects(dtype) || _may_have_objects(type)) {
+        if (npy_cache_import_runtime(
+                "numpy._core._internal", "_view_is_safe",
+                &npy_runtime_imports._view_is_safe) == -1) {
+            return NULL;
+        }
+        PyObject *safe = PyObject_CallFunctionObjArgs(
+                npy_runtime_imports._view_is_safe, dtype, type, NULL);
+        if (safe == NULL) {
+            return NULL;
+        }
+        Py_DECREF(safe);
+    }
+
+    if (type->elsize != dtype->elsize) {
+        /*
+         * Viewing as an unsized void implies a void dtype matching
+         * the size of the current dtype.
+         */
+        if (type->type_num == NPY_VOID && PyDataType_ISUNSIZED(type)) {
+            PyArray_Descr *newtype = PyArray_DescrNew(type);
+            if (newtype == NULL) {
+                return NULL;
+            }
+            newtype->elsize = dtype->elsize;
+            return newtype;
+        }
+        /*
+         * Otherwise, changing the size of the dtype results in a shape change,
+         * which we signal by setting newlastdim and newlaststride.
+         */
+        /* Check forbidden cases. */
+        if (PyDataType_HASSUBARRAY(type)) {
+            PyErr_SetString(PyExc_ValueError,
+                    "Changing the dtype to a subarray type is only supported "
+                    "if the total itemsize is unchanged");
+            return NULL;
+        }
+        int nd = PyArray_NDIM(self);
+        if (nd == 0) {
+            PyErr_SetString(PyExc_ValueError,
+                    "Changing the dtype of a 0d array is only supported "
+                    "if the itemsize is unchanged");
+            return NULL;
+        }
+        /* Resize on last axis only (we have nd>0 here). */
+        npy_intp lastdim = PyArray_DIMS(self)[nd-1];
+        if (lastdim != 1 && PyArray_SIZE(self) != 0 &&
+                PyArray_STRIDES(self)[nd-1] != dtype->elsize) {
+            PyErr_SetString(PyExc_ValueError,
+                    "To change to a dtype of a different size, the last axis "
+                    "must be contiguous");
+            return NULL;
+        }
+        if (type->elsize < dtype->elsize) {
+            /* If it is compatible, increase the size of the last axis. */
+            if (type->elsize == 0 || dtype->elsize % type->elsize != 0) {
+                PyErr_SetString(PyExc_ValueError,
+                        "When changing to a smaller dtype, its size must be a "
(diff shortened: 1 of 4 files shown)


seberg:
The intermediate view may not be a big deal, but yeah, it's definitely nice to skip it!

> but I thought I would get some opinions on whether this is the right direction at all before doing that. Though opening as draft for now

I like this in principle, but I think there are two subtleties, and unless we shrug them off (probably OK, but...), I am actually not sure there is a clean solution:
* I _think_ `PyArray_View()` diverges from `.view()` here, and I am not sure that it still works as well as it should with this change (without fixing `__array_finalize__` there can be regressions, e.g. `MaskedArray` itself could have one in theory.).
* It would be nice to keep `.dtype =` work with a DeprecationWarning to not break existing implementations.

Now, I can see calling `obj.view()` from `PyArray_View()` if it exists, while `arr.view()` would of course not and (try) call `.dtype =` (`arr.view()` would be the current implementation).
In that scenario, I think we get:
* `def view()` can be used to avoid the `.dtype =` setting (but it may be a bit inconvenient).
* If _delete_ `.view()` you can push all logic into `__array_finalize__`, which may make a lot of sense.  (But `MaskedArray` is interesting as it adds a kwarg, so we can't delete it.)

So, in my approach `_set_dtype` sticks around indefinitely, while in this `.view()` does, but may not be used.

With `_set_dtype` we could eventually deprecate and force transitioning to `__array_finalize__` only, but it's a two step thing. (I am not actually sure we can if we promise to call `.view()`, though.)

But too much text, not sure what it gives us.  To me using `_set_dtype` is still the easier choice probably. But if we either disregard the above problems a bit _or_ like the `PyArray_View()` calling `subclass.view()` as the final state for sure, then I think this is better.

mhvk:
@seberg - I've adapted this follow-up, where just one view is done for the "standard in the future" path. Note that I got carried away a bit in my reorganization of the code, which makes the difference view rather poor. You may want to just look at the file directly... (I do think it has become clearer as a result, in particular by just putting the path of the future first...)

seberg:
Nice, so this "just" re-organizes the double view.  I'll have a look, moves things around, but I think otherwise I'll probably just put it in unless I find an issue.

(I have a "more complete fix (TM)" for the deprecation, almost lined up, but might just put this in and rebase. it's a bit ridiculous and of course nothing is really 100%, but considering how important the `.dtype` assignment deprecation is in a sense...)

numpy/_core/src/multiarray/convert.c
```diff
@@ -612,82 +606,89 @@ PyArray_View(PyArrayObject *self, PyArray_Descr *type, PyTypeObject *pytype)
...
-            }
-            Py_DECREF(res);
+        NPY_ALLOC_WORKSPACE(newdims, npy_intp, 2 * 4, 2 * nd);
+        npy_intp *newstrides = newdims + nd;
+        int lastaxis = -1;
+        /* Check whether the type is compatible. */
+        Py_SETREF(type, _check_compatibility_with_new_dtype(
+                self, type, &lastaxis, &newdims[nd-1], &newstrides[nd-1]));
+        if (type == NULL) {
+            return NULL;
```
seberg, line 619: Ah, one little thing I am just spotting: This return fails to clean up the workspace.

Maybe use `NPY_DEFINE_WORKSPACE` + `NPY_INIT_WORKSPACE`.  Then we can just put the `npy_free_workspace` into a `goto finish;`?

(Although, probably not... just adding it on this path is good.)

seberg: Or just hardcode a stack allocation for `MAXDIMS * 2` space there, since `nd` is just limited/small enough, quite honestly.

mhvk:
MacOS / Accelerate warning would seem unrelated...

mhvk:
@seberg - sorry, looking at it again, I felt I could clean up the addition a bit more, so I did... Should now be truly ready for final review.

numpy/_core/src/multiarray/convert.c
mhvk: Now done in a path where one is guaranteed to pass by `free`

numpy/_core/src/multiarray/convert.c
```diff
@@ -612,82 +606,92 @@ PyArray_View(PyArrayObject *self, PyArray_Descr *type, PyTypeObject *pytype)
...
-            }
-            /* DEPRECATED 2026-04-13, NumPy 2.5 */
-            if (DEPRECATE(
-                    "numpy.ndarray.view() used a custom `dtype` setter "
-                    "to change the dtype of the view.  Subclasses should "
-                    "implement `_set_dtype` instead.") < 0) {
-                Py_CLEAR(ret);
-                goto finish;
-            }
+            NPY_ALLOC_WORKSPACE(newdims, npy_intp, 2 * 4, 2 * nd);
```
seberg, line 628:
```suggestion
            NPY_ALLOC_WORKSPACE(newdims, npy_intp, 2 * 4, 2 * nd);
            if (newdims == NULL) {
                goto finish;
            }
```

numpy/_core/src/multiarray/convert.c
```diff
@@ -612,82 +606,92 @@ PyArray_View(PyArrayObject *self, PyArray_Descr *type, PyTypeObject *pytype)
...
-        if (use_set_dtype) {
-            PyObject *res = PyObject_CallMethodOneArg(
-                    (PyObject *)ret,
-                    npy_interned_str._set_dtype, (PyObject *)type);
-            if (res == NULL) {
-                Py_CLEAR(ret);
-                goto finish;
-            }
-            Py_DECREF(res);
+        /* Take view with old or adjusted dims (steals refence to type) */
```
seberg:
```suggestion
        /* Take view with old or adjusted dims (steals reference to type) */
```

numpy/_core/src/multiarray/common.c
```diff
@@ -370,6 +370,103 @@ _may_have_objects(PyArray_Descr *dtype)
...
+
+    /* Check that we are not reinterpreting memory containing Objects. */
+    if (_may_have_objects(dtype) || _may_have_objects(type)) {
+        if (npy_cache_import_runtime(
+                "numpy._core._internal", "_view_is_safe",
+                &npy_runtime_imports._view_is_safe) == -1) {
+            return NULL;
+        }
+        PyObject *safe =PyObject_CallFunction(
+                npy_runtime_imports._view_is_safe, "OO", dtype, type);
```
seberg:
```suggestion
        PyObject *safe = PyObject_CallFunctionObjArgs(
                npy_runtime_imports._view_is_safe, dtype, type, NULL);
```
Just a nit, but I think CallFunctionObjArgs is nicer anyway.

seberg (approved):
A few small nits I'll just apply. Thanks.

numpy/_core/src/multiarray/getset.c
seberg, line 365:
```suggestion
    if (PyDataType_HASSUBARRAY(newtype)) {
        assert(newlastdim < 0);  // not allowed for subarrays
```
Feels a bit nerve-wrecking to me without that clarity.

seberg: (nvm, doesn't work in gh-ui somehow, I'll just push.)

seberg (approved):

mhvk:
Agreed with all changes, thanks!

PR #31273, opened 2026-04-19
ENH: ArrayMethod sorts for builtin dtypes

MaanasArora:
Adds new-style sorting and argsorting array-methods (#29737) to built-in NumPy dtypes (bools, integers, floats, dates; there are methods for string/unicode but not registered yet). This allows the new path in `PyArray_Sort` to be functional for these, which will enable writing descending sorts soon after this. ping @seberg @mhvk - thanks!

Happy to add descending sorts in this PR if it doesn't make this too hard to read, there are just a lot of (repetitive) lines with this direction. I tried reducing duplication, but seems unfortunately hard to do because of the C++ templating, and I actually somewhat prefer this so far just to keep the new/old sorts quite separate.

Still have to check the resolve descriptors setup and string/unicode quirks, but otherwise this is basically complete.

#### AI Disclosure

I used Claude to fill in the `*sort_loop` functions with the new signatures. I gave it the bool + existing code for each (arg)sort kind and let it generate the rest, then looked briefly but didn't find any inconsistencies, but will keep an eye.


numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -4600,6 +4729,72 @@ set_typeinfo(PyObject *dict)
...
+            .slots = sort_slots_@name@,
+            .flags = NPY_METH_NO_FLOATINGPOINT_ERRORS,
+    };
+    PyBoundArrayMethodObject *sort_method_@name@ = PyArrayMethod_FromSpec_int(
+            &sort_spec_@name@, 1);
+    if (sort_method_@name@ == NULL) {
+        return -1;
+    }
+    NPY_DT_SLOTS(dtypemeta)->sort_meth = sort_method_@name@->method;
+    Py_INCREF(sort_method_@name@->method);
```
MaanasArora, line 4772: Leaking, but causes a segfault when I decref...

MaanasArora:
Sorry, totally overcomplicated this! Closing this in favor of the far smaller alternative in #31278.

Issue #8627, opened 2017-02-16
Bug in np.histogram when min == max and values are >= 2**53


astrofrog:
The following call to ``histogram`` fails:

```
In [5]: np.histogram([2**53])
/Users/tom/miniconda3/envs/dev/lib/python3.6/site-packages/numpy/lib/function_base.py:566: RuntimeWarning: divide by zero encountered in double_scalars
  norm = bins / (mx - mn)
/Users/tom/miniconda3/envs/dev/lib/python3.6/site-packages/numpy/lib/function_base.py:591: RuntimeWarning: invalid value encountered in multiply
  tmp_a *= norm
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-5-41f77d7d86af> in <module>()
----> 1 np.histogram([2**53])

/Users/tom/miniconda3/envs/dev/lib/python3.6/site-packages/numpy/lib/function_base.py in histogram(a, bins, range, normed, weights, density)
    598             # The index computation is not guaranteed to give exactly
    599             # consistent results within ~1 ULP of the bin edges.
--> 600             decrement = tmp_a_data < bin_edges[indices]
    601             indices[decrement] -= 1
    602             # The last bin includes the right edge. The other bins do not.

IndexError: index -9223372036854775808 is out of bounds for axis 1 with size 11
```

whereas this succeeds:

```
In [6]: np.histogram([2**53-1])
Out[6]: 
(array([0, 0, 0, 0, 0, 0, 1, 0, 0, 0]),
 array([  9.00719925e+15,   9.00719925e+15,   9.00719925e+15,
          9.00719925e+15,   9.00719925e+15,   9.00719925e+15,
          9.00719925e+15,   9.00719925e+15,   9.00719925e+15,
          9.00719925e+15,   9.00719925e+15]))
```

The threshold for triggering the bug appears to be 2**53, but this only happens when the min of the array is the same as the max:

```
In [7]: np.histogram([2**53, 2**53])
/Users/tom/miniconda3/envs/dev/lib/python3.6/site-packages/numpy/lib/function_base.py:566: RuntimeWarning: divide by zero encountered in double_scalars
  norm = bins / (mx - mn)
/Users/tom/miniconda3/envs/dev/lib/python3.6/site-packages/numpy/lib/function_base.py:591: RuntimeWarning: invalid value encountered in multiply
  tmp_a *= norm
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
[...]

PR #31276, opened 2026-04-19
ENH: add string bin support for multivariate histograms

JosephMehdiyev:
### PR summary

fixes #20215
1) Adds multi dimensional bin width histogram algorithm(s)
    also generalizes existing bin width histogram algorithms to arrays (for dimensionality)
    majority of them will throw not implemented error when D>1
2) Generalizes some helper functions to multi dimensions
3) Adds string support for multi dimensional histograms
### Other comments
I am not content with "auto" choice but it should be practical enough.

#### AI Disclosure

~I used Claude as sanity check on my fixes. All modifications etc are done solely by my decisions and manually typed or copy pasted documentation from the related PR above.~
Claude was used extensively on some parts of the code, especially on `_get_bin_edges`. Other than that I do not remember to be honest.

### DIFF
--- a/numpy/lib/_histograms_impl.py
+++ b/numpy/lib/_histograms_impl.py
@@ -998,15 +998,23 @@ def histogramdd(sample, bins=10, range=None, density=None, weights=None):
         weights = np.asarray(weights)
 
     try:
-        M = len(bins)
         if isinstance(bins, str):
             raise TypeError
+        M = len(bins)
         if M != D:
-            raise ValueError(
-                'The dimension of bins must be equal to the dimension of the '
-                'sample x.')
+            if any(isinstance(b, str) for b in bins):
+                raise ValueError(
+                        'If string values are used in ArrayLike, dimension of bins must be equal'
+                        'to the dimension of the sample x.'
+                        )
+            if np.ndim(bins) <= 1:
+                    raise TypeError
+            else:
+                raise ValueError(
+                    'The dimension of bins must be equal to the dimension of the '
+                    'sample x.')
     except TypeError:
-        # bins is an integer or a string
+        # bins is an integer or a string or an ArrayLike of edges.
         bins = D * [bins]
 
     # normalize the range argument
--- a/numpy/lib/_histograms_impl.py
+++ b/numpy/lib/_histograms_impl.py
@@ -355,14 +355,15 @@ def _unsigned_subtract(a, b):
 
 def _get_bin_edges(a, bins, range, weights):
     """
-    Computes the bins used internally by `histogram`.
+    Computes the bins used internally by `histogram`, `histogramdd` and `histogram2d`.
 
     Parameters
     ==========
     a : ndarray
         Ravelled data array
     bins, range
-        Forwarded arguments from `histogram`.
+        Forwarded arguments from `histogram` or from current dimension of `histogramdd`
+        or `histogram2d`
     weights : ndarray, optional
         Ravelled weights array, or None
 
@@ -929,13 +930,17 @@ def histogramdd(sample, bins=10, range=None, density=None, weights=None):
 
         The first form should be preferred.
 
-    bins : sequence or int, optional
+    bins : sequence or int or str, optional
         The bin specification:
 
         * A sequence of arrays describing the monotonically increasing bin
           edges along each dimension.
         * The number of bins for each dimension (nx, ny, ... =bins)
         * The number of bins for all dimensions (nx=ny=...=bins).
+        * A sequence of strings defining the method to calculate the optimal
+          bin width for each dimension as defined by `histogram_bin_edges`.
+        * A string defining the methodused to calculate the
+          optimal bin width, as defined by `histogram_bin_edges`.
 
     range : sequence, optional
         A sequence of length D, each an optional (lower, upper) tuple giving
--- a/numpy/lib/_histograms_impl.py
+++ b/numpy/lib/_histograms_impl.py
@@ -997,20 +997,23 @@ def histogramdd(sample, bins=10, range=None, density=None, weights=None):
     if weights is not None:
         weights = np.asarray(weights)
 
-    try:
-        if isinstance(bins, str):
-            raise TypeError
-        M = len(bins)
-        if M != D:
-            if np.ndim(bins) <= 1:
-                raise TypeError
-            else:
-                raise ValueError(
-                    'The dimension of bins must be equal to the dimension of the '
-                    'sample x.')
-    except TypeError:
-        # bins is an integer or a string or an ArrayLike of edges.
+    if isinstance(bins, str):
+        # bins is a string
         bins = D * [bins]
+    else:
+        try:
+            M = len(bins)
+            if M != D:
+                if np.ndim(bins) == 1:
+                    # bins is an arraylike
+                    bins = D * [bins]
+                else:
+                    raise ValueError(
+                        'The dimension of bins must be equal to the dimension of the '
+                        'sample x.')
+        except TypeError:
+            # bins is an integer
+            bins = D * [bins]
 
     # normalize the range argument
     if range is None:
--- a/numpy/lib/_histograms_impl.py
+++ b/numpy/lib/_histograms_impl.py
@@ -980,11 +980,11 @@ def histogramdd(sample, bins=10, range=None, density=None, weights=None):
 
     try:
         # Sample is an ND-array.
-        N, D = sample.shape
+        _, D = sample.shape
     except (AttributeError, ValueError):
         # Sample is a sequence of 1D arrays.
         sample = np.atleast_2d(sample).T
-        N, D = sample.shape
+        _, D = sample.shape
 
     nbin = np.empty(D, np.intp)
     edges = D * [None]
@@ -994,12 +994,14 @@ def histogramdd(sample, bins=10, range=None, density=None, weights=None):
 
     try:
         M = len(bins)
+        if isinstance(bins, str):
+            raise TypeError
         if M != D:
             raise ValueError(
                 'The dimension of bins must be equal to the dimension of the '
                 'sample x.')
     except TypeError:
-        # bins is an integer
+        # bins is an integer or a string
         bins = D * [bins]
 
     # normalize the range argument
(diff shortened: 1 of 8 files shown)


numpy/lib/_histograms_impl.py
```diff
@@ -929,13 +930,17 @@ def histogramdd(sample, bins=10, range=None, density=None, weights=None):
 
         The first form should be preferred.
 
-    bins : sequence or int, optional
+    bins : sequence or int or str, optional
```
jorenham, line 933: don't forget to also update the stubs :)
https://github.com/numpy/numpy/blob/12e16ab322390059fffee12a153f10c92881f29d/numpy/lib/_histograms_impl.pyi#L243-L273

numpy/lib/_twodim_base_impl.py
```diff
@@ -705,18 +705,24 @@ def histogram2d(x, y, bins=10, range=None, density=None, weights=None):
     y : array_like, shape (N,)
         An array containing the y coordinates of the points to be
         histogrammed.
-    bins : int or array_like or [int, int] or [array, array], optional
+    bins : int or array_like or str or [int, int] or [array, array] or
```
jorenham, line 708: this should also be updated in the stubs:
https://github.com/numpy/numpy/blob/12e16ab322390059fffee12a153f10c92881f29d/numpy/lib/_twodim_base_impl.pyi#L223-L375

JosephMehdiyev:
Seems like I have focused on the old PR so much that I have missed so many things (other than stubs), please review after PR is ready for review

JosephMehdiyev:
@jorenham (afaik you are the author of these stubs)  is there a specific reason why overload stubs of `histogram2d` are more detailed (a couple of them even looks reduntant to me) and written differently than `histogram` and `histogramdd`? While I am here changing the file I might also rewrite it similar to `histogram` etc, unless these are intended


for example deleting this ~wont change things~ will succesfully run tests anyways (also true before PR changes)
```Python
@overload
def histogram2d(
    x: _ArrayLike1DNumber_co,
    y: _ArrayLike1DNumber_co,
    bins: _BinKind | Sequence[Sequence[int] | _BinKind],
    range: _ArrayLike2DFloat_co | None = None,
    density: bool | None = None,
    weights: _ArrayLike1DFloat_co | None = None,
) -> _Histogram2D[np.int_]: ...
```
since this below already handles that case
```Python
def histogram2d[ScalarT: _Number_co](
    x: _ArrayLike1DNumber_co,
    y: _ArrayLike1DNumber_co,
    bins: _BinKind | _ArrayLike1D[ScalarT] | Sequence[_ArrayLike1D[ScalarT] | _BinKind],
    range: _ArrayLike2DFloat_co | None = None,
    density: bool | None = None,
    weights: _ArrayLike1DFloat_co | None = None,
) -> _Histogram2D[ScalarT]: ...
```
I have no experience with .pyi files, I am mirroring my knowledge from .hpp and cpp templates FYI

jorenham:
> @jorenham (afaik you are the author of these stubs) is there a specific reason why overload stubs of `histogram2d` are more detailed (a couple of them even looks reduntant to me) and written differently than `histogram` and `histogramdd`? While I am here changing the file I might also rewrite it similar to `histogram` etc, unless these are intended
> 
> for example deleting this wont change things
> 
> ```python
> @overload
> def histogram2d(
>     x: _ArrayLike1DNumber_co,
>     y: _ArrayLike1DNumber_co,
>     bins: _BinKind | Sequence[Sequence[int] | _BinKind],
>     range: _ArrayLike2DFloat_co | None = None,
>     density: bool | None = None,
>     weights: _ArrayLike1DFloat_co | None = None,
> ) -> _Histogram2D[np.int_]: ...
> ```
> 
> since this below already handles that case
> 
> ```python
> def histogram2d[ScalarT: _Number_co](
>     x: _ArrayLike1DNumber_co,
>     y: _ArrayLike1DNumber_co,
>     bins: _BinKind | _ArrayLike1D[ScalarT] | Sequence[_ArrayLike1D[ScalarT] | _BinKind],
>     range: _ArrayLike2DFloat_co | None = None,
>     density: bool | None = None,
>     weights: _ArrayLike1DFloat_co | None = None,
> ) -> _Histogram2D[ScalarT]: ...
> ```
> 
> I have no experience with .pyi files, I am mirroring my knowledge from .hpp and cpp templates FYI

The difference here is in the `bins`, where the first overload here handles `int` sequences, and the second `ScalarT`. The important bit is that `ScalarT` only accept `np.number | np.bool`, and would therefore reject `int`.

So these overloads are distinct and non-overlapping. 

... or at least, that's what they *should* be. You added `_BinKind` to both, which makes them overlap in an incompatible way.

JosephMehdiyev:
Not sure the failing tests are PR related.
I did not do anything about https://github.com/numpy/numpy/issues/31296, let me know if I should add some documentation about it
PR should be ready to review now. I looked around the stuff I did and did not find any issues (hopefully I am right)

numpy/lib/_histograms_impl.py
```diff
@@ -993,13 +998,23 @@ def histogramdd(sample, bins=10, range=None, density=None, weights=None):
...
         M = len(bins)
         if M != D:
-            raise ValueError(
-                'The dimension of bins must be equal to the dimension of the '
-                'sample x.')
+            if any(isinstance(b, str) for b in bins):
+                raise ValueError(
+                        'If string values are used in ArrayLike, dimension of bins'
+                        'must be equal to the dimension of the sample x.'
+                        )
```
JosephMehdiyev: in second thought this specific warning seems unnecessary, i'll change it

JosephMehdiyev:
hey, can someone give a feedback for this PR?

numpy/lib/_histograms_impl.py
```diff
@@ -993,13 +998,18 @@ def histogramdd(sample, bins=10, range=None, density=None, weights=None):
...
         if M != D:
-            raise ValueError(
-                'The dimension of bins must be equal to the dimension of the '
-                'sample x.')
+            if np.ndim(bins) <= 1:
+                raise TypeError
+            else:
+                raise ValueError(
+                    'The dimension of bins must be equal to the dimension of the '
+                    'sample x.')
```
jorenham: You use a `TypeError` here to effictively "goto" the `except TypeError` on line 1012. Thhis doesn't seem like a good design pattern to me. So could you change it to use an `if` statement or something similar?

numpy/lib/_twodim_base_impl.pyi
```diff
@@ -242,7 +243,7 @@ def histogram2d[ScalarT: np.complexfloating](
 def histogram2d[ScalarT: np.inexact](
     x: _ArrayLike1D[ScalarT],
     y: _ArrayLike1D[ScalarT | _Int_co],
-    bins: int | Sequence[int] = 10,
+    bins: _BinKind | int | Sequence[int | _BinKind] = 10,
```
jorenham: nit: For consistency it's better to stick with either `_BinKind | int` or `int | _BinKind` (viz. the order)

JosephMehdiyev: done!

numpy/lib/_histograms_impl.py
JosephMehdiyev: I agree with you, I removed the psuedo "goto" stuff

JosephMehdiyev:
Other than minor fixes, is the ~fix~ PR  good? Any big issues? Would like to finish this PR at this point

jorenham:
I had Claude take a look at the stubs, and it actually found some real issues:

#### `complexfloating` overloads of `histogram2d` were NOT updated (bug)

_twodim_base_impl.pyi — the two overloads for `ScalarT: np.complexfloating` still have `bins: int | Sequence[int]`:

```python
@overload
def histogram2d[ScalarT: np.complexfloating](
    x: _ArrayLike1D[ScalarT],
    y: _ArrayLike1D[ScalarT | _Float_co],
    bins: int | Sequence[int] = 10,  # ← missing _BinKind
    ...
```

Since `np.complexfloating` is a subtype of `np.inexact`, the type checker matches these overloads *first* for complex arrays. If you call `np.histogram2d(complex_arr, complex_arr, bins="auto")`, it won't resolve to the `complexfloating` overload and will silently fall back to a less-specific one with `Any` in the return type — losing the `ScalarT` binding. These should also have `int | _BinKind | Sequence[int | _BinKind]`.

The same applies to the `Sequence[complex]` overload around _twodim_base_impl.pyi — it also doesn't include `_BinKind`.

#### `histogramdd` stubs don't precisely type per-dimension string bins

For `histogramdd`, the updated signature is `bins: _BinKind | SupportsIndex | ArrayLike`. This means `bins=['auto', 'fd']` (list of per-dimension estimators) is technically only covered through `ArrayLike`, which is semantically incorrect (it's for numeric arrays, not a list of method name strings). A more precise type would be:

```python
bins: _BinKind | SupportsIndex | ArrayLike | Sequence[_BinKind | SupportsIndex | ArrayLike] = 10
```

This is a relatively minor precision issue but worth noting, since a mixed list like `['auto', 5, np.array([0,1,2])]` is a valid input to `histogramdd` and wouldn't be well-typed.

#### Ordering inconsistency across files (minor, marked resolved)

The resolved review thread noted `int | _BinKind` vs `_BinKind | int`. Within _twodim_base_impl.pyi, `int | _BinKind` is now consistent. However, _histograms_impl.pyi uses `_BinKind | SupportsIndex | ...` (BinKind first). This cross-file inconsistency was not addressed.

JosephMehdiyev:
> Since np.complexfloating is a subtype of np.inexact, the type checker matches these overloads first for complex arrays. If you call np.histogram2d(complex_arr, complex_arr, bins="auto"), it won't resolve to the complexfloating overload and will silently fall back to a less-specific one with Any in the return type — losing the ScalarT binding. These should also have int | _BinKind | Sequence[int | _BinKind].

this was intentional by me, complex data does not behave nicely with string bins? I do not remember how I found this, I think i just looked at `hisstogram` implementation

I'll fix other minor things

JosephMehdiyev:
other than the stubs, the main fix should be OK. 
I keep forgetting about this PR, would be nice to finish it

jorenham (approved):
Typing-wise this looks fine now.

JosephMehdiyev:
I would appreciate if the main fix was also reviewed and conclude the PR

jorenham:
I understand, but it's probably a good idea to have a maintainer look at that bit who's actually familiar with these parts of these code (which I am not).

mattip:
I think the main fix (not the typing) is OK too. It seems like a nice cleanup of a bit of historical cruft. Anyone else want to take a look?

mattip (approved):

JosephMehdiyev:
Note that this will introduce some bugs that were already present in `histogram` such as https://github.com/numpy/numpy/issues/8627. Although I am planning to fix histogram related issues over the summer

JosephMehdiyev:
It would be nice to conclude this PR as I have a couple of other PRs continuing this work.
I already have another PR that fixes the bug I have mentioned above.

JosephMehdiyev:
Changes are because of https://github.com/numpy/numpy/issues/20215#issuecomment-4612608475
There are still a lot to do but the ~`stone`~ 'fd' and `scott` algorithms should work.

JosephMehdiyev:
stubs: The only things changed in stubs are that strings cannot be in array i.e "auto" is fine but not ["auto"] or ["auto", 2] and bins cannot be complex values. Complex values are unrelated to PR, but might as well fix it here.

1)  I could only generalize `fd` and `scott` algorithms as other algorithms do not have literature about N-D case. We could somewhat generalize others to D dimension by changing some variables to respect D, but it would be purely heuristic

2) `auto` is tricky as it is not possibe to use the existing 1-D auto.  `fd` or `scott` should be good enough for large dimensions because of the curse of dimensionality, but they may not work nicely on some cases in D=2 or D=3

3) I tried to read papers and implement other multivariate bin width algorithms, these are generally maximize some kind of likelihood function, but the problem is that it is computationally expensive and the end result is not practically useful in my testing

JosephMehdiyev:
cc @jorenham see the above short comment about the stub changes, be free to review whenever, (if) you want

I will continue to update the documentation, tests and clean up some code  too.

jorenham (approved):
the stub changes look fine

PR #31278, opened 2026-04-20
ENH: ArrayMethod sorting for built-in dtypes

MaanasArora:
Adds new-style sorting and argsorting array-methods (https://github.com/numpy/numpy/pull/29737) to built-in NumPy dtypes (bools, integers, floats, dates). This allows the new path in `PyArray_Sort` to be functional for these, which will enable writing descending sorts soon after this. Supersedes #31273 to be much leaner. ping @seberg @mhvk - thanks!

Still have to check the resolve descriptors setup and see if string/unicode can be added, but otherwise this is basically complete (modulo below refcount comment).

#### AI Disclosure

I used Claude to write the initial template `*_(arg)sort_loop` functions but I cleaned it to the point there is little left of it.

### DIFF
--- a/numpy/_core/src/multiarray/arraytypes.c.src
+++ b/numpy/_core/src/multiarray/arraytypes.c.src
@@ -4565,12 +4565,6 @@ argsort_resolve_descriptors(
  *         half, float, double, longdouble,
  *         cfloat, cdouble, clongdouble,
  *         datetime, timedelta#
- * #NAME = Bool,
- *         Byte, UByte, Short, UShort, Int, UInt,
- *         Long, ULong, LongLong, ULongLong,
- *         Half, Float, Double, LongDouble,
- *         CFloat, CDouble, CLongDouble,
- *         Datetime, Timedelta#
  * #rsort = 1*5, 0*15#
  */
 
--- a/numpy/_core/src/multiarray/arraytypes.c.src
+++ b/numpy/_core/src/multiarray/arraytypes.c.src
@@ -4527,11 +4527,8 @@ sort_resolve_descriptors(
     if (NPY_UNLIKELY(output_descrs[0] == NULL)) {
         return -1;
     }
-    output_descrs[1] = NPY_DT_CALL_ensure_canonical(input_descrs[1]);
-    if (NPY_UNLIKELY(output_descrs[1] == NULL)) {
-        Py_XDECREF(output_descrs[0]);
-        return -1;
-    }
+    Py_INCREF(output_descrs[0]);
+    output_descrs[1] = output_descrs[0];
 
     return method->casting;
 }
--- a/numpy/_core/src/multiarray/arraytypes.c.src
+++ b/numpy/_core/src/multiarray/arraytypes.c.src
@@ -4509,6 +4509,117 @@ static int
 }
 /**end repeat**/
 
+/*
+ *****************************************************************************
+ **                          SORTING ARRAYMETHODS                           **
+ *****************************************************************************
+ */
+
+static NPY_CASTING
+sort_resolve_descriptors(
+        PyArrayMethodObject *method,
+        PyArray_DTypeMeta *const *dtypes,
+        PyArray_Descr *const *input_descrs,
+        PyArray_Descr **output_descrs,
+        npy_intp *view_offset)
+{
+    output_descrs[0] = NPY_DT_CALL_ensure_canonical(input_descrs[0]);
+    if (NPY_UNLIKELY(output_descrs[0] == NULL)) {
+        return -1;
+    }
+    output_descrs[1] = NPY_DT_CALL_ensure_canonical(input_descrs[1]);
+    if (NPY_UNLIKELY(output_descrs[1] == NULL)) {
+        Py_XDECREF(output_descrs[0]);
+        return -1;
+    }
+
+    return method->casting;
+}
+
+static NPY_CASTING
+argsort_resolve_descriptors(
+        PyArrayMethodObject *method,
+        PyArray_DTypeMeta *const *dtypes,
+        PyArray_Descr *const *input_descrs,
+        PyArray_Descr **output_descrs,
+        npy_intp *view_offset)
+{
+    output_descrs[0] = NPY_DT_CALL_ensure_canonical(input_descrs[0]);
+    if (NPY_UNLIKELY(output_descrs[0] == NULL)) {
+        return -1;
+    }
+    output_descrs[1] = PyArray_DescrFromType(NPY_INTP);
+    if (NPY_UNLIKELY(output_descrs[1] == NULL)) {
+        Py_XDECREF(output_descrs[0]);
+        return -1;
+    }
+
+    return method->casting;
+}
+
+/**begin repeat
+ *
+ * #suff = bool,
+ *         byte, ubyte, short, ushort, int, uint,
+ *         long, ulong, longlong, ulonglong,
+ *         half, float, double, longdouble,
+ *         cfloat, cdouble, clongdouble,
+ *         datetime, timedelta#
+ * #NAME = Bool,
+ *         Byte, UByte, Short, UShort, Int, UInt,
+ *         Long, ULong, LongLong, ULongLong,
+ *         Half, Float, Double, LongDouble,
+ *         CFloat, CDouble, CLongDouble,
+ *         Datetime, Timedelta#
+ * #rsort = 1*5, 0*15#
+ */
+
+static int
+@suff@_sort_loop(PyArrayMethod_Context *context,
+        char *const data[], npy_intp const dimensions[],
+        npy_intp const strides[], NpyAuxData *NPY_UNUSED(auxdata))
+{
+    PyArrayMethod_SortParameters *params = (PyArrayMethod_SortParameters *)context->parameters;
+    switch (params->flags) {
+        case NPY_SORT_DEFAULT:
+            return quicksort_@suff@(data[0], dimensions[0], NULL);
+        case _NPY_SORT_HEAPSORT:
+        case NPY_SORT_STABLE:
+#if @rsort@
+            return radixsort_@suff@(data[0], dimensions[0], NULL);
+#else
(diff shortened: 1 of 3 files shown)


numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -4600,6 +4711,72 @@ set_typeinfo(PyObject *dict)
...
+            .slots = sort_slots_@name@,
+            .flags = NPY_METH_NO_FLOATINGPOINT_ERRORS,
+    };
+    PyBoundArrayMethodObject *sort_method_@name@ = PyArrayMethod_FromSpec_int(
+            &sort_spec_@name@, 1);
+    if (sort_method_@name@ == NULL) {
+        return -1;
+    }
+    NPY_DT_SLOTS(dtypemeta)->sort_meth = sort_method_@name@->method;
+    Py_INCREF(sort_method_@name@->method);
```
MaanasArora: For some reason, decrefing `sort_method_@name@` causes a segfault, but this is leaking otherwise...

numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -4509,6 +4509,111 @@ static int
...
+        npy_intp const strides[], NpyAuxData *NPY_UNUSED(auxdata))
+{
+    PyArrayMethod_SortParameters *params = (PyArrayMethod_SortParameters *)context->parameters;
+    switch (params->flags) {
+        case NPY_SORT_DEFAULT:
+            return aquicksort_@suff@(data[0], (npy_intp *)data[1], dimensions[0], NULL);
+        case _NPY_SORT_HEAPSORT:
+        case NPY_SORT_STABLE:
+#if @rsort@
+            return aradixsort_@suff@(data[0], (npy_intp *)data[1], dimensions[0], NULL);
```
seberg, line 4602: Just a quick question: we don't use radixsort for unstable as well?

numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -4509,6 +4509,111 @@ static int
...
+        return -1;
+    }
+    output_descrs[1] = PyArray_DescrFromType(NPY_INTP);
+    if (NPY_UNLIKELY(output_descrs[1] == NULL)) {
+        Py_XDECREF(output_descrs[0]);
+        return -1;
+    }
+
+    return method->casting;
+}
```
seberg, line 4555: Hmmm, I guess we these them for timedelta64/datetime64 unfortunately, because it is parametric?  (Just too bad, since this is all very simple in principle...)

numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -4509,6 +4509,111 @@ static int
...
+{
+    output_descrs[0] = NPY_DT_CALL_ensure_canonical(input_descrs[0]);
+    if (NPY_UNLIKELY(output_descrs[0] == NULL)) {
+        return -1;
+    }
+    output_descrs[1] = NPY_DT_CALL_ensure_canonical(input_descrs[1]);
+    if (NPY_UNLIKELY(output_descrs[1] == NULL)) {
+        Py_XDECREF(output_descrs[0]);
+        return -1;
+    }
```
seberg: You can just set the output to the input, since sorting must be in-place.

numpy/_core/src/multiarray/arraytypes.c.src
seberg: Rather cryptic, I agree.  The problem is that:
```
    if (PyType_Ready(&PyArrayMethod_Type) < 0) {
        return -1;
    }
    if (PyType_Ready(&PyBoundArrayMethod_Type) < 0) {
        return -1;
    }
```
needs to be called first.  Remind me: `object` and `void` already use this due to being the default fallback path, I guess?
(But I agree that they are different enough, either way.)

seberg (commented):
Thanks, small comments, mostly how to fix the decref issue.

mhvk (commented):
This looks nice! Main question I have is whether we can know these routines are actually used! You probably found out implicitly from errors... Can one test using the `RuntimeError`? Though it seems difficult to trigger in practice. So perhaps it can wait until we support descending...

numpy/_core/src/multiarray/arraytypes.c.src
MaanasArora: Right, don't seem to:

https://github.com/numpy/numpy/blob/e5cee5ed26a555256cc87c26ab6a5a2f9a43542f/numpy/_core/src/multiarray/arraytypes.c.src#L4162-L4177

numpy/_core/src/multiarray/arraytypes.c.src
MaanasArora: Unfortunate indeed, yes, it's because they're parametric. Given how often we're running into this, probably good to fix the issue somehow...

MaanasArora:
Thanks @mhvk! Yes I found out via errors indeed, and confirmed with `printf` statements (I know...) Do you mean a runtime error in `PyArray_Sort`? I think that could make sense, something that checks if one of these dtypes has made it past the array method path, it should be fast so we probably don't care about penalizing other dtypes. Going to try this if it makes sense!

numpy/_core/src/multiarray/arraytypes.c.src
MaanasArora: Good catch thanks! This is done.

numpy/_core/src/multiarray/arraytypes.c.src
MaanasArora: No I don't think so actually! `object` and `void` have a comparison function that is set here:

```
    (PyArray_CompareFunc*)@from@_compare,
```

So, they are using the old path. That said it shouldn't at all be hard to use `npy_default_sort_loop` to take in a new comparison function for these.

numpy/_core/src/multiarray/arraytypes.c.src
```diff
@@ -4509,6 +4509,108 @@ static int
...
+            return quicksort_@suff@(data[0], dimensions[0], NULL);
+        case _NPY_SORT_HEAPSORT:
+        case NPY_SORT_STABLE:
+#if @rsort@
+            return radixsort_@suff@(data[0], dimensions[0], NULL);
+#else
+            return timsort_@suff@(data[0], dimensions[0], NULL);
+#endif
+        default:
+            PyErr_Format(PyExc_RuntimeError,
```
mhvk, line 4585: I was wondering if it was possible to hit this `RuntimeError` -- but ideally from python. But that does seem tricky; having checked with print statements really is good enough!

numpy/_core/src/multiarray/arraytypes.c.src
mhvk: Hmm, if those use `@from@_compare`, can we then not do the same in #30415 (we discussed there somewhat at https://github.com/numpy/numpy/pull/30415#pullrequestreview-3960264084)

numpy/_core/src/multiarray/arraytypes.c.src
seberg: I would prefer:
```
Py_INCRFEF(output_descrs[0]);
output_descrs[1] = output_descrs[0];
```
shorter, and a bit clearer of what our loop requires (rather than what sort of course also provides).

numpy/_core/src/multiarray/arraytypes.c.src
seberg: Right, we could add that slot, but in the end, I suppose either way we'll be using a hard-coded path for those.  (We'll probably just want to change the signature of the internal sort function to re-align, but it is hard-coded either way.)
One could add a "uses compare slot" auto-fill, but I am not sure it is worth it.

numpy/_core/src/multiarray/arraytypes.c.src
MaanasArora: Is that not what we have?

numpy/_core/src/multiarray/arraytypes.c.src
seberg: Yeah, not sure we can reach it easily.  Although, once we follow up with reverse sorting, I think we'll have our proof at least?

numpy/_core/src/multiarray/arraytypes.c.src
seberg: sorry, nvm, I am not sure what I looked at or clicked to see the old version still.

numpy/_core/src/multiarray/arraytypes.c.src
MaanasArora: I think we can reach it by setting `descending=True` right? That should hit this?

Edit: sorry, I didn't recall we don't have it in Python yet for some reason! Yeah, that could be easier to test... no kwarg seems to hit this path right now in Python.

numpy/_core/src/multiarray/arraytypes.c.src
MaanasArora: Agreed, probably will hard-code those, so unclear whether filling the new slot with legacy sort is worth the tradeoff.

So for the DECREF... I guess this entire registration needed to be moved, so I added a function and called that after that block instead of within `set_typeinfo`, thanks!

seberg: That should work yeah.  If it is nicer, you can also just move the finalization of ArrayMethod_Type and BoundArrayMethod_Type to happen first, they don't have any dependencies and can happen much earlier.

MaanasArora: Ah okay, thanks! That does look nicer (less verbose), pushed!

seberg:
@MaanasArora happy to put this in if it helps with the other PR, just let me know.  I think we'll want to move much of this to C++ to avoid that `c.src -> C++` template transition (which is just really verbose).
(So in the end, maybe we can just go with the other PR directly and skip this also.)

MaanasArora:
@seberg Yeah I rebased the other PR and moved sort loop creation to C++, so we can just go with that one I think! I'll close this. Thanks!

PR #31282, opened 2026-04-20
DOC: update datetime documentation to explain that "NaT" also requires specific unit

riku-sakamoto:
### PR summary

This is a follow-up PR based on the comment https://github.com/numpy/numpy/pull/31213#issuecomment-4280003952

Thank you for pointing it out! @lucascolley 

#### AI Disclosure

I used ChatGPT to improve the English wording in this commit https://github.com/numpy/numpy/pull/31282/changes/3720a77eb84e11a49904fdf4866c8299385bdf28 .

### DIFF
--- a/doc/source/reference/arrays.datetime.rst
+++ b/doc/source/reference/arrays.datetime.rst
@@ -95,8 +95,8 @@ as ``now``.
 
     NAT (not a time):
 
-    >>> np.datetime64('nat')
-    np.datetime64('NaT')
+    >>> np.datetime64('nat', 'D')
+    np.datetime64('NaT', 'D')
 
     The current time (UTC, default second precision):
 
@@ -117,6 +117,11 @@ as ``now``.
 When creating an array of datetimes from a string, it is still possible
 to automatically select the unit from the inputs, by using the
 datetime type with generic units.
+
+.. deprecated:: 2.5
+  The generic units of `timedelta64` are deprecated since NumPy 2.5 and
+  will raise an error in the future. With this changes, `datetime64`'s ``NaT`` is now required to have an explicit time unit for consistency with other datetime and timedelta operations.
+
 
 .. admonition:: Example
 
@@ -218,8 +223,8 @@ data type also accepts the string "NAT" in place of the number for a "Not A Time
     >>> np.timedelta64(4, 'h')
     np.timedelta64(4,'h')
 
-    >>> np.timedelta64('nAt')
-    np.timedelta64('NaT')
+    >>> np.timedelta64('nAt', 'D')
+    np.timedelta64('NaT', 'D')
 
 Datetimes and Timedeltas work together to provide ways for
 simple datetime calculations.
@@ -245,11 +250,11 @@ simple datetime calculations.
     >>> np.timedelta64(1,'W') % np.timedelta64(10,'D')
     np.timedelta64(7,'D')
 
-    >>> np.datetime64('nat') - np.datetime64('2009-01-01')
+    >>> np.datetime64('nat', 'D') - np.datetime64('2009-01-01')
     np.timedelta64('NaT','D')
 
-    >>> np.datetime64('2009-01-01') + np.timedelta64('nat')
-    np.datetime64('NaT')
+    >>> np.datetime64('2009-01-01') + np.timedelta64('nat', 'D')
+    np.datetime64('NaT', 'D')
 
 There are two Timedelta units ('Y', years and 'M', months) which are treated
 specially, because how much time they represent changes depending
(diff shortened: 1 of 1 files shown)


doc/source/reference/arrays.datetime.rst
```diff
@@ -686,3 +691,12 @@ The straight forward way is to replace the generic unit with a specific time uni
 
     >>> arr = np.ones(5, dtype='m8[s]')
     >>> np.testing.assert_allclose(arr, np.timedelta64(1, "s"), atol=np.timedelta64(0, "s"))
+
+    Previously, `datetime64` allowed `NaT` without specifying a time unit. 
+    With the deprecation of generic units, `NaT` is now required to have an explicit time unit for consistency with other datetime and timedelta operations.
+
```
charris, line 705: Be nice to break the long lines here and elsewhere. You need hard linebreaks.

riku-sakamoto: Thank you for the review! I've updated them accordingly.

doc/source/reference/arrays.datetime.rst
```diff
@@ -118,6 +118,11 @@ When creating an array of datetimes from a string, it is still possible
 to automatically select the unit from the inputs, by using the
 datetime type with generic units.
 
+.. deprecated:: 2.5
+  The generic units of `timedelta64` were deprecated in NumPy 2.5 and
+  will raise an error in the future. With this change, ``NaT`` in `datetime64` is now required to have an explicit time unit.
```
charris: You missed this one.

doc/source/reference/arrays.datetime.rst
```diff
@@ -686,3 +697,13 @@ The straight forward way is to replace the generic unit with a specific time uni
 
     >>> arr = np.ones(5, dtype='m8[s]')
     >>> np.testing.assert_allclose(arr, np.timedelta64(1, "s"), atol=np.timedelta64(0, "s"))
+
+    Previously, `datetime64` allowed `NaT` without specifying a time unit. 
+    With the deprecation of generic units, `NaT` is now required to have an explicit time unit
+    for consistency with other datetime and timedelta operations.
+
+    >>> np.datetime64("NAT")
+    DeprecationWarning: The 'generic' unit for NumPy timedelta is deprecated, and will raise an error in the future. This includes implicit conversion of bare integers (e.g. `+ 1`).Please use a specific unit instead.
```
charris, line 707: AI suggests you can break this like so:
```
>>> np.datetime64("NAT")
DeprecationWarning: The 'generic' unit for NumPy timedelta is deprecated, ... Please use a specific unit instead.
```
I don't know if that works. We should probably fix the warning text.

doc/source/reference/arrays.datetime.rst
riku-sakamoto: Oh, sorry. Thank you! I've updated it.

doc/source/reference/arrays.datetime.rst
riku-sakamoto: I left it as is because this is inside a code snippet.
When I tried to break the lines, the documentation ended up like below, so the line breaks appear as is.

Would you prefer line breaks even inside a code snippet, or should I keep it?

image

charris: AI suggested shortening the output line using elipses `...` for missing text.  I don't know if it works, but maybe worth a try to find out. No, I wouldn't break the line.

ngoldbaum:
Do the tests for the deprecation capture these cases?

riku-sakamoto:
@ngoldbaum 

Yes, I think so. The tests below capture the deprecation warnings

https://github.com/numpy/numpy/blob/388bd75829842d8230147f57eba5f86bb3ad6c3c/numpy/_core/tests/test_deprecations.py#L439-L441

ngoldbaum (approved):
LGTM, thanks for following through on this @riku-sakamoto

doc/source/reference/arrays.datetime.rst
riku-sakamoto: Thanks for the clarification.

I tried it locally, but I’m not sure if it improves the situation because it still doesn't fit one page width
Anyway, if you prefer, I am happy to push the changes.

image

charris: If it works, we can always shorten the line more. I assume it worked? That is good to know. I'll put this in and you can make another PR if the improvement looks good. I think we could break it right after the elipses. It is just an example, if it occurs in the wild the user will see the whole thing.

Strictly speaking, it would be better to put some line breaks in the emitted warning message.

charris:
Thanks @riku-sakamoto .

Issue #31284, opened 2026-04-20
BUG: blas_fpe_check() deadlocks on glibc < 2.34 when numpy is imported from within dlopen (MKL + OpenMP)


isuruf:
### Describe the issue:

`blas_fpe_check()` in `numpy/__init__.py` (added in #30102, backported to 2.3.5) executes `ones((20,20)) @ ones((20,20))` at import time. When numpy is imported from within a `dlopen()` context — e.g. a C extension's static constructor calling `PyImport_ImportModule("numpy")` — this causes a deadlock on glibc < 2.34 with MKL BLAS.

**Mechanism:**

1. Python loads a C extension via `dlopen()` — glibc acquires `dl_load_lock`
2. The extension's static constructor calls `PyImport_ImportModule("numpy")`
3. numpy's `__init__.py` runs `blas_fpe_check()` → `x @ x` → `cblas_dgemm`
4. MKL dispatches `dgemm` via OpenMP, spawning worker threads
5. Worker threads call `mkl_serv_load_fun()` → `dlsym()` → tries to acquire `dl_load_lock`
6. **Deadlock**: main thread holds `dl_load_lock` and waits at `__kmp_join_barrier` for workers; workers block on `dl_load_lock` held by main thread

On glibc ≥ 2.34, `dlopen`/`dlsym` use a recursive lock (`dl_load_lock2`) so this doesn't deadlock. On glibc < 2.34 (RHEL 8, CentOS 8, Amazon Linux 2, etc.), the lock is non-reentrant.

Prior to #30102, no BLAS computation happened at import time, so this was not an issue.

### Reproduce the code example:

```python
# Create environment with MKL-backed numpy and any C extension that
# imports numpy from a static constructor (csp does this):
conda create -n test python=3.11 csp "blas=*=mkl" "numpy=2.3.5"



# repro.py — deadlocks on glibc < 2.34
import importlib.util
import os
import sysconfig

site_packages = sysconfig.get_path("platlib")
so_path = os.path.join(site_packages, "csp", "lib", "_cspimpl.so")

# Loading _cspimpl.so via dlopen triggers its C++ static constructors,
# which call PyImport_ImportModule("numpy"). This worked before 2.3.5.
spec = importlib.util.spec_from_file_location("_cspimpl", so_path)
mod = importlib.util.module_from_spec(spec)
spec.loader.exec_module(mod)  # hangs forever
print("OK")


**Workaround** — pre-import numpy before loading the extension:

[...]

PR #31286, opened 2026-04-20
BUG: BLAS FPE check import-time deadlock on macOS ARM64

Breno-Favaretto:
### PR summary
This prevents BLAS from being executed during module import, which can lead to deadlocks when NumPy is imported from within a dlopen-triggered static constructor (OpenMP + MKL + loader lock interaction).

Behavior of BLAS execution is unchanged outside import time.

Related to: https://github.com/numpy/numpy/issues/31284

  

#### AI Disclosure
AI tools were used to assist with drafting and structuring the explanation, but the analysis, design, and implementation decisions were made and verified by the author.

### DIFF
--- a/numpy/__init__.py
+++ b/numpy/__init__.py
@@ -860,8 +860,9 @@ else:
                         "Spurious warnings given by blas but suppression not "
                         "set up on this platform. Please open a NumPy issue.",
                         UserWarning, stacklevel=2)
-
-    blas_fpe_check()
+    
+    if sys.platform == "darwin" and s.uname().machine == "arm64":        
+            blas_fpe_check()
     del blas_fpe_check
 
     def hugepage_setup():


rgommers (changes requested):
This contains a typo that makes this clearly incorrect (`s` -> `os`), and the guard is also not in the right place. The explanation is a very basic restatement of the actual explanation given by another contributor in the linked issue.

That's too many issues, so I'm going to close it straight away. @Breno-Favaretto please don't restate someone else's analysis within hours without adding anything insightful yourself, and then make a PR for something they're clearly working on. And test your work before submitting it.

Finally, the AI disclosure seems nonsensical and just copied from some other PR, or auto-generated. Please do not do that again.

Breno-Favaretto:
Thanks for the feedback @rgommers and that's fair.

I moved too quickly on this and should have validated the change more carefully before opening the PR. I also understand the concern about overlapping with existing work.

I'm still getting familiar with the project and will take more time to contribute more carefully going forward. Also, English is not my first language, so I may sometimes misinterpret things  I’ll be more careful with that as well.

PR #31288, opened 2026-04-20
DOC/TYP: remove outdated subsection on 0D arrays

jorenham:
Because of the recent shape-typing improvements, most functions now return a scalar type (instead of 0D arrays) when they'd return a scalar at runtime.

#### AI Disclosure
N/A


charris:
Thanks Joren.

PR #31290, opened 2026-04-20
BUG: Raise warning for integer arithmetic with timedelta64/datetime64 arrays

### LINKED ISSUE #31255: BUG: Arithmetic operations on timedelta64 arrays are not associative
### Describe the issue:

Arithmetic operations on timedelta64 arrays are not associative without any warning.
This issue is somewhat related to #28287, but I think it is a seperate issue.

Would it be better to raise a warning in such cases because using `generic` unit for timedelta64 scalars now raises a DeprecationWarning since gh-29619 ?

### Rationale

Type promotion rules for timedelta64 arrays are defined below.
These rules don't use `generic` unit explicitly, but it is implicitly assumed when the other operand is not a timedelta64 array. 

https://github.com/numpy/numpy/blob/4f74f0adcfd3637a48da0ba9bf3b1a7e088db843/numpy/_core/src/umath/ufunc_type_resolution.c#L744-L764

```python
import numpy as np

a = np.array([1], dtype="m8[s]")
b = np.array([1])
c = np.array([1], dtype="m8[ms]")

print(a + b + c)
# Out: [2001]

print(a + (b + c))
# Out: [1002]
```

### Error message:

```shell

```

2.5.0.dev0+git20260415.4f74f0a

### Runtime Environment:

```
[{'numpy_version': '2.5.0.dev0+git20260415.4f74f0a',
  'python': '3.12.3 (main, Jul 12 2025, 11:14:50) [GCC 11.4.0]',
  'uname': uname_result(system='Linux', node='pop-os', release='6.17.9-76061709-generic', version='#202511241048~1764704751~22.04~b24b425 SMP PREEMPT_DYNAMIC Tue D', machine='x86_64')},
                      'found': ['X86_V3'],
                      'not_found': ['X86_V4', 'AVX512_ICL']}},
 {'architecture': 'Haswell',
  'filepath': '/usr/lib/x86_64-linux-gnu/openblas-pthread/libopenblasp-r0.3.20.so',
  'internal_api': 'openblas',
  'num_threads': 14,
  'prefix': 'libopenblas',
  'user_api': 'blas',
  'version': '0.3.20'}]
```

I found this issue while working on gh-31213.
It may be a follow-up issue of my previous PR gh-29619, so I've opened this.

may12day:
### PR summary

Adds deprecation warning for arithmetic between integer arrays and timedelta64 arrays.
Same issue was found in datetime64, added deprecation there too. Happy to revert if out of scope.
Fixes #31255

Hi, was going through internals of numpy and saw the issue. Wanted to contribute and I'll be happy to receive feedback.

#### AI Disclosure

Took minimal help from Claude in writing test cases.

### DIFF
--- a/numpy/_core/tests/test_deprecations.py
+++ b/numpy/_core/tests/test_deprecations.py
@@ -486,7 +486,7 @@ class TestDeprecatedGenericTimedelta(_DeprecationTestCase):
     ):
         """Test that timedelta64 array + integer array triggers deprecation."""
         # timedelta op int
-        self.assert_deprecated(op, num=1, args=(timedelta_arr, int_arr))
+        self.assert_deprecated(op, num=None, args=(timedelta_arr, int_arr))
         # int op timedelta
         self.assert_deprecated(op, num=None, args=(int_arr, timedelta_arr))
 
@@ -504,7 +504,7 @@ class TestDeprecatedGenericTimedelta(_DeprecationTestCase):
     ):
         """Test that datetime64 array + integer array triggers deprecation."""
         # datetime op int
-        self.assert_deprecated(op, num=1, args=(datetime_arr, int_arr))
+        self.assert_deprecated(op, num=None, args=(datetime_arr, int_arr))
         # int op datetime
         if op == np.add:
             self.assert_deprecated(op, num=None, args=(int_arr, datetime_arr))
@@ -516,7 +516,7 @@ class TestDeprecatedGenericTimedelta(_DeprecationTestCase):
         c = np.array([1], dtype="m8[ms]")
 
         # Both intermediate operations should trigger warnings
-        self.assert_deprecated(np.add, num=1, args=(a, b))
+        self.assert_deprecated(np.add, num=None, args=(a, b))
         self.assert_deprecated(np.add, num=None, args=(b, c))
 
 
--- a/numpy/_core/tests/test_deprecations.py
+++ b/numpy/_core/tests/test_deprecations.py
@@ -461,7 +461,7 @@ class TestDeprecatedGenericTimedelta(_DeprecationTestCase):
     def test_raise_warning_for_operation_with_generic_unit(
         self, value: int, generic_value: int, op: Callable
     ):
-        self.assert_deprecated(op, num=None, args=(value, generic_value))
+        self.assert_deprecated(op, num=2, args=(value, generic_value))
 
     def test_raise_warning_for_default_constructor(self):
         self.assert_deprecated(lambda: np.timedelta64())
@@ -486,9 +486,9 @@ class TestDeprecatedGenericTimedelta(_DeprecationTestCase):
     ):
         """Test that timedelta64 array + integer array triggers deprecation."""
         # timedelta op int
-        self.assert_deprecated(op, num=None, args=(timedelta_arr, int_arr))
+        self.assert_deprecated(op, num=1, args=(timedelta_arr, int_arr))
         # int op timedelta
-        self.assert_deprecated(op, num=None, args=(int_arr, timedelta_arr))
+        self.assert_deprecated(op, num=1, args=(int_arr, timedelta_arr))
 
     @pytest.mark.parametrize('datetime_arr', [
         np.array(['2020-01-01', '2020-01-02'], dtype="M8[D]"),
@@ -504,10 +504,10 @@ class TestDeprecatedGenericTimedelta(_DeprecationTestCase):
     ):
         """Test that datetime64 array + integer array triggers deprecation."""
         # datetime op int
-        self.assert_deprecated(op, num=None, args=(datetime_arr, int_arr))
+        self.assert_deprecated(op, num=1, args=(datetime_arr, int_arr))
         # int op datetime
         if op == np.add:
-            self.assert_deprecated(op, num=None, args=(int_arr, datetime_arr))
+            self.assert_deprecated(op, num=1, args=(int_arr, datetime_arr))
 
     def test_non_associative_case_warns(self):
         """Verify the specific non-associative case from gh-31255 warns."""
@@ -516,8 +516,8 @@ class TestDeprecatedGenericTimedelta(_DeprecationTestCase):
         c = np.array([1], dtype="m8[ms]")
 
         # Both intermediate operations should trigger warnings
-        self.assert_deprecated(np.add, num=None, args=(a, b))
-        self.assert_deprecated(np.add, num=None, args=(b, c))
+        self.assert_deprecated(np.add, num=1, args=(a, b))
+        self.assert_deprecated(np.add, num=1, args=(b, c))
 
 
 class TestTriDeprecationWithNonInteger(_DeprecationTestCase):
--- a/numpy/_core/tests/test_deprecations.py
+++ b/numpy/_core/tests/test_deprecations.py
@@ -461,7 +461,7 @@ class TestDeprecatedGenericTimedelta(_DeprecationTestCase):
     def test_raise_warning_for_operation_with_generic_unit(
         self, value: int, generic_value: int, op: Callable
     ):
-        self.assert_deprecated(op, num=2, args=(value, generic_value))
+        self.assert_deprecated(op, num=1, args=(value, generic_value))
 
     def test_raise_warning_for_default_constructor(self):
         self.assert_deprecated(lambda: np.timedelta64())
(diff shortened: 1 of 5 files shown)


ngoldbaum:
Please run the test suite and make sure it passes locally. Feel free to turn on CI for your fork if you want to have it run without us pressing any approval buttons.

Ping @riku-sakamoto in case you want to take a look.

numpy/_core/tests/test_deprecations.py
```diff
@@ -437,6 +437,60 @@ def test_raise_warning_for_NAT_construction(self):
         self.assert_deprecated(lambda: np.datetime64(None))
 
 
+class TestDeprecatedIntegerTimedeltaArrayArithmetic(_DeprecationTestCase):
+    # Deprecation in NumPy 2.5, 2026-04
+    message = "The 'generic' unit for NumPy timedelta is deprecated"
```
riku-sakamoto: It seems that it might be more consistent to add test methods into `TestDeprecatedGenericTimedelta` since the same deprecation message is raised.

numpy/_core/tests/test_deprecations.py
```diff
@@ -437,6 +437,60 @@ def test_raise_warning_for_NAT_construction(self):
...
+    @pytest.mark.parametrize("op", [np.add, np.subtract])
+    def test_timedelta_array_integer_array_arithmetic(
+        self, timedelta_arr, int_arr, op
+    ):
+        """Test that timedelta64 array + integer array triggers deprecation."""
+        # timedelta + int (num=None allows any number of warnings)
+        self.assert_deprecated(op, num=None, args=(timedelta_arr, int_arr))
+        # int + timedelta (for add only, subtract is different)
+        if op == np.add:
+            self.assert_deprecated(op, num=None, args=(int_arr, timedelta_arr))
```
riku-sakamoto: I think we need to test both `add` and `subtract` operations. It doesn't seem that this `if` statement is necessary.

may12day:
Hi @ngoldbaum 
Whenever you have a second, could you approve the workflows to run?

may12day:
Hi @ngoldbaum
Would you mind triggering the workflows again? I’m expecting a fresh pass will resolve the previous test failure.

ngoldbaum:
@riku-sakamoto would you mind giving this another once-over?

@seberg it looks like this didn't quite make it in time before 2.6 got branched. Do you think we should backport this to ensure it goes out in 2.5 with the other unitless timedelta/datetime deprecations?

numpy/_core/tests/test_deprecations.py
```diff
@@ -430,7 +431,7 @@ def test_raise_warning_for_timedelta_with_generic_unit(self, value: int | str):
     def test_raise_warning_for_operation_with_generic_unit(
         self, value: int, generic_value: int, op: Callable
     ):
-        self.assert_deprecated(op, args=(value, generic_value))
+        self.assert_deprecated(op, num=None, args=(value, generic_value))
```
riku-sakamoto: Is it necessary to set `num=None` ? (this warning is raised multiple times? )

numpy/_core/tests/test_deprecations.py
```diff
@@ -440,6 +441,54 @@ def test_raise_warning_for_NAT_construction(self):
...
+    ])
+    @pytest.mark.parametrize("op", [np.add, np.subtract])
+    def test_datetime_array_integer_array_arithmetic(
+        self, datetime_arr, int_arr, op
+    ):
+        """Test that datetime64 array + integer array triggers deprecation."""
+        # datetime op int
+        self.assert_deprecated(op, num=None, args=(datetime_arr, int_arr))
+        # int op datetime
+        if op == np.add:
```
riku-sakamoto, line 509: this `if` statement does not seem necessary. We need to test both `add` and `substract`.

riku-sakamoto (commented):
Thanks for the update! I've commented several nits.

numpy/_core/tests/test_deprecations.py
may12day: Yes. For the operation `np.add(np.timedelta64(3, "s"), 5)` , it triggers two deprecation warnings :
1. existing scalar generic-unit deprecation ([gh-29619](https://github.com/numpy/numpy/pull/29619))
2. new array operation deprecation (current one)

Without `num=None`, the test fails which expects 1 warning but receiving 2

numpy/_core/tests/test_deprecations.py
may12day: I guess `int - datetime` is not a valid operation.

numpy/_core/tests/test_deprecations.py
riku-sakamoto: If so, using `num=2` here might be more robust.
Also, could we set `num=1` in other tests such as `test_timedelta_array_integer_array_arithmetic` ?

may12day: Makes sense. I'll update it.

may12day:
@riku-sakamoto Done with the changes. What's the next step now?

riku-sakamoto:
@may12day 
Thank you. My concerns are addressed now. I don't have any further comments.

may12day:
Hi @ngoldbaum 
Changes are approved by the @riku-sakamoto. Could you please take a look again?

seberg (approved):
Thanks, LGTM and @riku-sakamoto alrady had a look, too, sorry for losing track of this!
(My only tiny comment would be that avoiding the duplication a bit would be nice, but it's also fine).

> Do you think we should backport this to ensure it goes out in 2.5 with the other unitless timedelta/datetime deprecations?

Sorry, I think we could as long as it hits the rc1, @charris what do you think?  I suspect we need to take this a bit slow anyway, so I am not too bothered if it misses 2.5, though.

seberg:
That said, this will need some test fixups, since I now see that CI failures are real.

numpy/_core/tests/test_deprecations.py
```diff
@@ -460,7 +461,7 @@ def test_raise_warning_for_timedelta_with_generic_unit(self, value: int | str):
     def test_raise_warning_for_operation_with_generic_unit(
         self, value: int, generic_value: int, op: Callable
     ):
-        self.assert_deprecated(op, args=(value, generic_value))
+        self.assert_deprecated(op, num=2, args=(value, generic_value))
```
may12day: @riku-sakamoto 
After reviewing the failed testcases, this testcase fires only 1 warning. I'll update `num=1`

numpy/_core/tests/test_deprecations.py
```diff
@@ -470,6 +471,54 @@ def test_raise_warning_for_NAT_construction(self):
...
+    ])
+    @pytest.mark.parametrize("op", [np.add, np.subtract])
+    def test_timedelta_array_integer_array_arithmetic(
+        self, timedelta_arr, int_arr, op
+    ):
+        """Test that timedelta64 array + integer array triggers deprecation."""
+        # timedelta op int
+        self.assert_deprecated(op, num=1, args=(timedelta_arr, int_arr))
+        # int op timedelta
+        self.assert_deprecated(op, num=1, args=(int_arr, timedelta_arr))
```
may12day: @riku-sakamoto @seberg 
According to failed testcases :
1. 1 warning is generated when there is single element int array or timedelta/datetime array as a operand first. Like `np.add(np.array([1]), np.array([1,2,3], dtype="m8[s]"))`
2. 2 warnings are generated when multi element integer array is the first operand. Like `np.add(np.array([1,2,3]), np.array([1,2,3], dtype="m8[s]"))`

So, I'll go with `num=None` for `# int op timedelta`. Thoughts?

seberg: Does `num=None` work?  Then just go for it.  Or just figure out `num=...`.  I honestly don't think it matters much.

(The only thing that would matter is if we can see why there are two warnings exactly and avoid the second one.  But, I suspect the reason is the casting fast-path, which means there isn't probably just nothing to be done.)

may12day: Most likely, yes. Earlier, `num=None` was working fine, it passed all the testcases.

seberg:
OK, I don't love the double warning, but I am pretty sure it is very hard to avoid it (some of Peter's work to optimize things might be on this path).

So thanks everyone, let's put this in.  Since the pre-release is out now, I think we probably shouldn't backport.

may12day:
Thanks for the help @seberg @riku-sakamoto @ngoldbaum. Glad to get my first PR across the finish line.

PR #31293, opened 2026-04-21
DEP: Correct dtype in `__array_finalize__` if _set_dtype=None

### LINKED ISSUE #31281: BUG: a regression from gh-31234 seen in astropy (broadcasting error while creating views of a `Masked` array)
As reported downstream in https://github.com/astropy/astropy/issues/19579, the following (verbatim) astropy test case started failing against `main`

```py
import numpy as np
import pytest
from astropy.utils.masked import Masked

a = np.arange(6.0).reshape(2, 3)
ma = Masked(a, mask=np.array(
    [[True, False, False],
     [False, True, False]]
))

def test_viewing_as_new_dtype_not_implemented():
    dt = "f4"
    a.view(dt)
    with pytest.raises(NotImplementedError, match="different.*size"):
        ma.view(dt)
```
output
```
================================ FAILURES =================================
________________ test_viewing_as_new_dtype_not_implemented ________________

    def test_viewing_as_new_dtype_not_implemented():
        dt = "f4"
        a.view(dt)
        with pytest.raises(NotImplementedError, match="different.*size"):
>           ma.view(dt)

t.py:15: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
astropy/utils/masked/core.py:691: in view
    result = super().view(dtype, self._get_masked_cls(type))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
astropy/utils/masked/core.py:722: in __array_finalize__
    self._set_mask(getattr(obj, "_mask", False))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = MaskedNDArray([[   ———,    ———,    ———,    ———, 0.    ,    ———],
               [0.    ,    ———, 0.    ,    ———,    ———, 2.3125]],
              dtype=float32)
mask = array([[ True, False, False],
       [False,  True, False]])
copy = False

    def _set_mask(self, mask, copy=False):
        self_dtype = getattr(self, "dtype", None)
        mask_dtype = (
            np.ma.make_mask_descr(self_dtype)
            if self_dtype and self_dtype.names
            else np.dtype("?")
        )
        ma = np.asanyarray(mask, dtype=mask_dtype)
        if ma.shape != self.shape:
            # This will fail (correctly) if not broadcastable.
            self._mask = np.empty(self.shape, dtype=mask_dtype)
>           self._mask[...] = ma
            ^^^^^^^^^^^^^^^
E           ValueError: could not broadcast input array from shape (2,3) into shape (2,6)

astropy/utils/masked/core.py:296: ValueError
========================= short test summary info =========================
FAILED t.py::test_viewing_as_new_dtype_not_implemented - ValueError: could not broadcast input array from shape (2,3) into shap...
============================ 1 failed in 0.49s ============================
```

The regression bisects to https://github.com/numpy/numpy/pull/31234
The error seem dubious to me as I'd expect broadcasting to work on these shapes, especially since no such error was previously raised in earlier versions.

seberg:
This avoids using the correct dtype in __array_finalize__ *unless* `_set_dtype = None` is set on the class.
We also use this new path to make record arrays follow the main class deprecation correctly.

This means that `recarray` does live in the future.

CC @mhvk, hopefully the last iteration (unless the larger `view` change works out)...

(no AI tool use)

Closes gh-31281

### DIFF
--- a/doc/release/upcoming_changes/31293.deprecation.rst
+++ b/doc/release/upcoming_changes/31293.deprecation.rst
@@ -7,11 +7,12 @@ that do not see this ``DeprecationWarning`` may wish to update their code.
 
 A subclass that does any ``dtype`` specific logic (i.e. verifying the dtype
 in ``__array_finalize__`` or has a ``dtype`` property) should now:
+
 * Set ``_set_dtype = None`` in which case ``arr.view(dtype=new_dtype)``
   will call ``__array_finalize__`` with the new dtype, ensuring that
   any validation ``__array_finalize__`` will run is done.
-* Define ``_set_dtype`` as a function (calling ``ndarray._set_dtype()``
-  to avoid ``DeprecationWarnings`` as a quick fix.
+* Or, for a quick fix, define ``_set_dtype`` as a function (calling
+  ``ndarray._set_dtype()`` to avoid ``DeprecationWarnings``.
   (Future versions might migrate towards the ``_set_dtype = None`` path.)
 
 Ideally, follow NumPy's deprecation to prevent ``dtype`` mutation by users.
--- a/doc/release/upcoming_changes/31293.deprecation.rst
+++ b/doc/release/upcoming_changes/31293.deprecation.rst
@@ -1,0 +1,19 @@
+Custom ``dtype`` property and ``__array_finalize__``
+----------------------------------------------------
+Previously ``arr.view(dtype=new_dtype)`` called ``arr.dtype = new_dtype``
+also for subclasses (i.e. the attribute setting).
+This path is now deprecated and refined, meaning that even subclasses
+that do not see this ``DeprecationWarning`` may wish to update their code.
+
+A subclass that does any ``dtype`` specific logic (i.e. verifying the dtype
+in ``__array_finalize__`` or has a ``dtype`` property) should now:
+* Set ``_set_dtype = None`` in which case ``arr.view(dtype=new_dtype)``
+  will call ``__array_finalize__`` with the new dtype, ensuring that
+  any validation ``__array_finalize__`` will run is done.
+* Define ``_set_dtype`` as a function (calling ``ndarray._set_dtype()``
+  to avoid ``DeprecationWarnings`` as a quick fix.
+  (Future versions might migrate towards the ``_set_dtype = None`` path.)
+
+Ideally, follow NumPy's deprecation to prevent ``dtype`` mutation by users.
+The use of ``ndarray._set_dtype()`` may be necessary for some subclass
+finalization patterns, but should otherwise be avoided.
--- a/numpy/_core/records.py
+++ b/numpy/_core/records.py
@@ -402,11 +402,13 @@ class recarray(ndarray):
             )
         return self
 
+    _set_dtype = None  # __array_finalize__ can deal with dtype changes
+
     def __array_finalize__(self, obj):
-        if self.dtype.type is not record and self.dtype.names is not None:
+        if issubclass(self.dtype.type, nt.void) and self.dtype.names is not None:
             # if self.dtype is not np.record, invoke __setattr__ which will
             # convert it to a record if it is a void dtype.
-            self.dtype = self.dtype
+            ndarray._set_dtype(self, sb.dtype((record, self.dtype)))
 
     def __getattribute__(self, attr):
         # See if ndarray has this attr, and return it if so. (note that this
@@ -455,11 +457,7 @@ class recarray(ndarray):
 
         newattr = attr not in self.__dict__
         try:
-            if attr == 'dtype':
-                # gh-29244
-                ret = self._set_dtype(val)
-            else:
-                ret = object.__setattr__(self, attr, val)
+            ret = object.__setattr__(self, attr, val)
         except Exception:
             fielddict = ndarray.__getattribute__(self, 'dtype').fields or {}
             if attr not in fielddict:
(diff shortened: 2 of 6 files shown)


seberg:
Hmmmm, or should we even only allow `_set_dtype = None`, but implementing it as a function does seem more obvious... moving things into `__array_finalize__` should be possible, but might be tricky...

EDIT: So one thing of course is, that if `_set_dtype is None`, you cannot use `subclass._set_dtype()` to avoid the dtype attribute setting deprecation.  At some point maybe that is niche enough that it doesn't matter... But I guess it may be enough of a reason to need the function pattern...

numpy/_core/records.py
```diff
@@ -402,11 +402,13 @@ def __new__(cls, shape, dtype=None, buf=None, offset=0, strides=None,
             )
         return self
 
+    _set_dtype = None  # __array_finalize__ can deal with dtype changes
+
     def __array_finalize__(self, obj):
-        if self.dtype.type is not record and self.dtype.names is not None:
+        if issubclass(self.dtype.type, nt.void) and self.dtype.names is not None:
```
seberg: This is the logic used in `__setattr__`, basically the same I guess, but thought I'd just copy paste it.

seberg: EDIT: ~Actually, no, I'll just undo this to avoid creating a new dtype here if it isn't necessary.  After that `dtype.type` is always `record`, so that check should be enough.~

Made it to have _both_ checks.  The second one is needed for some ridiculous tests to pass, but the original one is definitely OK as a fast-path.

doc/release/upcoming_changes/31293.deprecation.rst
```diff
@@ -0,0 +1,19 @@
...
+also for subclasses (i.e. the attribute setting).
+This path is now deprecated and refined, meaning that even subclasses
+that do not see this ``DeprecationWarning`` may wish to update their code.
+
+A subclass that does any ``dtype`` specific logic (i.e. verifying the dtype
+in ``__array_finalize__`` or has a ``dtype`` property) should now:
+* Set ``_set_dtype = None`` in which case ``arr.view(dtype=new_dtype)``
+  will call ``__array_finalize__`` with the new dtype, ensuring that
+  any validation ``__array_finalize__`` will run is done.
+* Define ``_set_dtype`` as a function (calling ``ndarray._set_dtype()``
```
mhvk: How about, "Of, for a quick fix to avoid ``DeprecationWarning``, define ..."

numpy/_core/src/multiarray/convert.c
```diff
@@ -561,17 +561,22 @@ PyArray_View(PyArrayObject *self, PyArray_Descr *type, PyTypeObject *pytype)
...
+     * Changing dtype on a subclass.  We support 4 paths:
      *
      * 1. subclass overrides _set_dtype: create subclass view first,
      *    then call _set_dtype (subclass handles dtype change).
+     *    If _set_dtype is set to None, we use path 3 below.
      * 2. subclass overrides the dtype descriptor (e.g. property with
      *    setter): create subclass view first, use the setter, but
      *    emit a deprecation asking to implement _set_dtype instead.
-     * 3. Otherwise (including plain ndarray): create an ndarray base
+     * 3. If _set_dtype is None (or base-class): create an ndarray base
```
mhvk: I think "or base-class" is incorrect, otherwise how would one ever get to path 4?

numpy/_core/src/multiarray/convert.c
```diff
@@ -561,17 +561,22 @@ PyArray_View(PyArrayObject *self, PyArray_Descr *type, PyTypeObject *pytype)
...
      *    view, set dtype internally, then create the subclass view
      *    if needed.  __array_finalize__ sees the final dtype+shape.
+     * 4. Otherwise, call `__array_finalize__` with old dtype and forcibly
```
mhvk: Maybe you meant to write it here, like `If _set_dtype is NULL or base-class: ...`

numpy/_core/src/multiarray/convert.c
```diff
@@ -660,7 +670,12 @@ PyArray_View(PyArrayObject *self, PyArray_Descr *type, PyTypeObject *pytype)
         goto finish;
     }
 
-    if (subtype != &PyArray_Type) {
+    /*
+     * Path 3: `_set_dtype is None` and `ret` is a base-class array
+     * with correct dtype+shape, this will call `__array_finalize__`
+     * with the final dtype+shape.
+     */
+    if (use_dtype_in_finalize && subtype != &PyArray_Type) {
```
mhvk: `use_dtype_in_finalize` can only be set if it is a subclass, so second part can be omitted.

mhvk (approved):
This does seem a good path forward! My comments are minimal, mostly about comments, so will approve now.

p.s. I may still follow up with changing the dims & strides beforehand.

mhvk:
@seberg - the addition of `_set_dtype = None` to `np.recarray` breaks a `FITS_rec` subclass in astropy. This means older astropy cannot be used with numpy 2.5. Personally, I don't think this is a disaster -- the fix is easy and can be backported (and people should not expect to be able to run newest numpy with older astropy), but I've raised it to ask for others' opinion. See https://github.com/astropy/astropy/pull/19584

PR #31294, opened 2026-04-21
MAINT: fix implicit conversion warnings

### LINKED ISSUE #31540: MAINT: implicit conversion warnings in lowlevel_strided_loops.c.src
I get a lot of warnings from clang 22 while compiling numpy on macOS like:
```console
../numpy/_core/src/multiarray/lowlevel_strided_loops.c.src:988:32: warning: implicit conversion from 'long long' to 'npy_longdouble' (aka 'double') changes value from 9223372036854775807 to 9223372036854775808 [-Wimplicit-const-int-float-conversion]
  988 |    if (_TO_RTYPE1(src_value) > LLONG_MAX) {
      |                              ~ ^~~~~~~~~
/opt/homebrew/Cellar/llvm/22.1.3/lib/clang/22/include/limits.h:109:20: note: expanded from macro 'LLONG_MAX'
  109 | #define LLONG_MAX  __LONG_LONG_MAX__
      |                    ^~~~~~~~~~~~~~~~~
<built-in>:65:27: note: expanded from macro '__LONG_LONG_MAX__'
   65 | #define __LONG_LONG_MAX__ 9223372036854775807LL
```

I think the warning is harmless and it is nice to suppress the warning to reduce the noise while compiling.

kumaraditya303:
### PR summary

Fixes #31540

I get a lot of warnings from clang 22 and this PR fixes it:
```console
../numpy/_core/src/multiarray/lowlevel_strided_loops.c.src:988:32: warning: implicit conversion from 'long long' to 'npy_longdouble' (aka 'double') changes value from 9223372036854775807 to 9223372036854775808 [-Wimplicit-const-int-float-conversion]
  988 |    if (_TO_RTYPE1(src_value) > LLONG_MAX) {
      |                              ~ ^~~~~~~~~
/opt/homebrew/Cellar/llvm/22.1.3/lib/clang/22/include/limits.h:109:20: note: expanded from macro 'LLONG_MAX'
  109 | #define LLONG_MAX  __LONG_LONG_MAX__
      |                    ^~~~~~~~~~~~~~~~~
<built-in>:65:27: note: expanded from macro '__LONG_LONG_MAX__'
   65 | #define __LONG_LONG_MAX__ 9223372036854775807LL
```

No AI used.

### DIFF
--- a/numpy/_core/src/multiarray/lowlevel_strided_loops.c.src
+++ b/numpy/_core/src/multiarray/lowlevel_strided_loops.c.src
@@ -981,6 +981,7 @@ static GCC_CAST_OPT_LEVEL int
 #   pragma GCC diagnostic push
 #   ifdef __clang__
 #   pragma GCC diagnostic ignored "-Wtautological-constant-out-of-range-compare"
+#   pragma GCC diagnostic ignored "-Wimplicit-const-int-float-conversion"
 #   endif
 #   pragma GCC diagnostic ignored "-Wtautological-compare"
 #   endif


ngoldbaum:
I think the warning is flagging a real issue - maybe we should just fix it?

Failing that, can you open an issue about the problem the warning is flagging and then add a comment here referring to that issue?

ngoldbaum:
Looks like the fix isn’t quite that simple

ngoldbaum:
> Failing that, can you open an issue about the problem the warning is flagging and then add a comment here referring to that issue?

btw I'm also happy with doing this

seberg:
@mattip just in case you have a quicker thought.  Maybe ignoring the warning is even the best thing (so long the compiler doesn't say "this is UB, I'll do something random", I am pretty sure things are correct)...
In C++ I feel we might be able to do a trick like seeing which type is actually used for the comparison and cast to that, but I dunno that is nice.

(I.e. I _think_ the cast is only right if we cast an integer to a float. In all other cases `>` will do promotion for the comparison and that is necessary?)

mattip:
I think the problem is on the left side, not the right side, no?

seberg:
> I think the problem is on the left side, not the right side, no?

I don't _think_ that is the problem, we have something like a `double->long` cast there, which gives us: `double_value > LONG_MAX`.
(IIRC, RTYPE is just there to cast `half` to `float`?)

Normally all is fine:
* The left hand has a smaller range (e.g. many int->int casts) it is tautological but the compiler realizes that.
* Both are integer values: I think things work out (but didn't think it fully through again, maybe there is an other branch that deals with signed/unsigned comparison shenanigans)
* For float to integer casts: The imprecise casts here should be OK, I guess.  But we get the new warning here now...

The problem is, adding the cast fixes the last one, but breaks some of the first two.

kumaraditya303:
> btw I'm also happy with doing this

I opened #31540 and reverted the PR to ignore the warnings.

seberg:
Thanks, just adding the warning filter makes sense.  Really not sure there is a nice other pattern available, TBH.

PR #31297, opened 2026-04-21
BUG: prevent crash when setting array flags with non-ASCII strings

Breno-Favaretto:
### PR summary
Fix a crash when using non-ASCII keys in:
```py
arr.flags.__setitem__
```

As reported in:
https://gist.github.com/devdanzin/ccc2d9553ca1c90ab1835362ee21a40a#reproducer-2-arrflags-segfault-on-non-ascii-key  
(discussed in #31046)

Using non-ASCII characters as keys could lead to a segmentation fault due to
unsafe handling of Unicode input in the C implementation.

#### Reproducer

```python
import numpy as np

arr = np.array([1, 2, 3])
arr.flags["é"] = True
```
#### Previous behavior
```raw
Segmentation fault    (core dumped)
```

#### New behavior
```raw
KeyError: 'Unknown flag'
```

####

#### AI Disclosure
AI was used to help me understand the English better and to assist in writing this message. All code and debugging were done by me.

### DIFF
--- a/numpy/_core/src/multiarray/flagsobject.c
+++ b/numpy/_core/src/multiarray/flagsobject.c
@@ -586,6 +586,9 @@ arrayflags_setitem(PyArrayFlagsObject *self, PyObject *ind, PyObject *item)
     if (PyUnicode_Check(ind)) {
         PyObject *tmp_str;
         tmp_str = PyUnicode_AsASCIIString(ind);
+        if (tmp_str == NULL){
+            goto fail;
+        }
         key = PyBytes_AS_STRING(tmp_str);
         n = PyBytes_GET_SIZE(tmp_str);
         if (n > 16) n = 16;


numpy/_core/src/multiarray/flagsobject.c
```diff
@@ -586,6 +586,9 @@ arrayflags_setitem(PyArrayFlagsObject *self, PyObject *ind, PyObject *item)
     if (PyUnicode_Check(ind)) {
         PyObject *tmp_str;
         tmp_str = PyUnicode_AsASCIIString(ind);
+        if (tmp_str == NULL){
```
jorenham, line 589: nitpick:

```suggestion
        if (tmp_str == NULL) {
```

jorenham:
oh, actually, this looks like it's a duplicate of #31257

Breno-Favaretto:
Ah, got it — thanks for the reference!

PR #31298, opened 2026-04-22
MAINT: Update dependabot configuration

charris:
Dependabot is not updating pytest even though pytest 9.0.3 was released April 7. Grok suggests this is a known dependabot problem with multiple files pinning the same dependency and suggests using groups to help it along. The same problem doesn't appear for actions, just the pip ecosystem.

[skip azp] [skip actions] [skip cirrus]

#### AI Disclosure

Grok read the numpy requirements files and suggested the fix.  This is a direct copy.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,7 @@ updates:
     schedule:
       interval: daily
     cooldown:
-      default-days: 7
+      default-days: 7     # optional
     commit-message:
       prefix: "MAINT"
     labels:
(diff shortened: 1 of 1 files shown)


charris:
This won't affect anything until it is merged, at which point dependabot should run.

EDIT: I think one single PR will show up, rather than a bunch of small ones, which could be good, or bad, depending on how many updates fail. I think that can be controlled by using multiple groups if needed. Typing requirements come to mind.

mattip:
Why did you remove the cooldowns? I think they would partially mitigate @charris ’s concerns about failed updates. Personally i would prefer a single pr to the multiple prs we now get.

charris:
I was somewhat hesitant to introduce the cool downs in the first place, especially for actions where I would want rapid response to cibuildwheel, they haven't been useful yet. That said, they haven't been bad either, I am completely neutral on the issue. I'll put them back until they actually cause trouble.

charris:
> Personally i would prefer a single pr to the multiple prs we now get.

If this actually does that, we could also add groups to the actions updates.

mattip:
Thanks @charris

PR #31300, opened 2026-04-22
DOC: wrong underline length and closing backtick

Carreau:
All in title

[skip ci]
[ci skip]

#### AI Disclosure

No AI tools used

### DIFF
--- a/numpy/f2py/crackfortran.py
+++ b/numpy/f2py/crackfortran.py
@@ -98,7 +98,7 @@ Usage:
      (see also fortran type declaration statement formats below)
 
 Fortran 90 type declaration statement format (F77 is subset of F90)
-====================================================================
+===================================================================
 (Main source: IBM XL Fortran 5.1 Language Reference Manual)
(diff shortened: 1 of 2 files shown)


numpy/f2py/crackfortran.py
```diff
@@ -98,7 +98,7 @@
      (see also fortran type declaration statement formats below)
 
 Fortran 90 type declaration statement format (F77 is subset of F90)
-====================================================================
+===================================================================
```
mattip: While here, let’s shorten the headline and move the second clause to be the first sentence of the text

Carreau: Done.

seberg:
Btw. is there an easy way to enforce our sphinx build to error out on this type of thing?  It's not like it happens a lot, but otherwise this will always be a recurring little thing.

Carreau:
Not that I know of, this was not emitted by sphinx.

PR #31302, opened 2026-04-22
ENH: Create a path for legacy dtypes to transition to new DType API

seberg:
This adds and importantly *backports* a way for legacy dtypes to transition to the new API.
I.e. with one caveat I found for now (array dtype printing prints as `np.array([1.0], dtype=repr(dtype))` and the repr now shouldn't be `bfloat16` for an `ml_dtype` bfloat16), this allows transitioning a package like `ml_dtypes` over.

Sure... For now it effectively just keep using some old code paths, but:
* It allows us to seriously start tthinking to deprecate old registration paths (and importantly casts)!
* It allows `ml_dtypes` to use new features they should want:
  * New-style `common_dtype` allows to fix most promotions, e.g. with Python integers (but also between their own dtypes).
  * `np.finfo` support (NumPy 2.4+).
  * ... I am missing something probably, plus anything else we add in future versions can be used if the runtime NumPy is new enough (with some manual header vendoring).

I am still thinking about how to test this and possibly adding some more sanity checks.
But: I have tested this with `ml_dtypes` and especially if the larger google tests run with that; that is presumably more thorough than anything we can do.

Yes, the backport is a bit of a hack (and in one place hinges on a very fun little implementation detail), but the important thing is that it works...

---

I think this is a good way to move forward, otherwise any single small issue in NumPy can be a regression for `ml_dtypes` and while for e.g. `quaddtype` that type of thing isn't all that bad (fix it, require newer NumPy version), for `ml_dtypes` it prevents a smooth transition.

This enables that transition, and if we are a bit agressive, dtypes are niche enough that I would be up for deprecating the old API functions soon unless some larger issue pops up.

(Did use a bit of claude to get started on the code.)

### DIFF
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -1243,6 +1243,19 @@ User-defined data types
         The struct is not a valid Python object, so do not use ``Py_DECREF``
         on it.
 
+        **Transitioning to the new DType AP**
+
+        Users currently using the old DType API may not be able to easily
+        transition due to NumPy issues that would cause regressions.
+        While new DTypes may reasonably limit themselves to newer NumPy
+        versions, existing DTypes cannot do so.
+
+        You can transition to the new DType API by using
+        the `NPY_DT_legacy_descriptor_proto` slot when registering.
+        This slot is available when compiling with NumPy 2.5 and provides
+        support for NumPy 2.0 and higher (it can also be vendored to compile
+        on older versions of NumPy).
+
     Register a data-type as a new user-defined data type for
     arrays. The type must have most of its entries filled in. This is
     not always checked and errors can produce segfaults. In
@@ -1277,6 +1290,10 @@ User-defined data types
     *totype*. Any old casting function is over-written. A ``0`` is
     returned on success or a ``-1`` on failure.
 
+    .. note::
+        This function will eventually be deprecated. Please migrate to the new
+        DType API for casts.  See `NPY_DT_legacy_descriptor_proto` for details.
+
     .. c:type:: PyArray_VectorUnaryFunc
 
         The function pointer type for low-level casting functions.
@@ -1289,6 +1306,12 @@ User-defined data types
     *scalar* = :c:data:`NPY_NOSCALAR` to register that an array of data-type
     *descr* can be cast safely to a data-type whose type_number is
     *totype*. The return value is 0 on success or -1 on failure.
+
+    .. note::
+        This function will eventually be deprecated. Please migrate to the new
+        DType API for casts.  See `NPY_DT_legacy_descriptor_proto` for details.
+        The new DType API provides more flexibility and speed even for dtypes
+        compatible with the legacy API.
 
 
 Special functions for NPY_OBJECT
@@ -3571,6 +3594,33 @@ Slot IDs and API Function Typedefs
 These IDs correspond to slots in the DType API and are used to identify
 implementations of each slot from the items of the ``slots`` array
 member of ``PyArrayDTypeMeta_Spec`` struct.
+
+.. c:macro:: NPY_DT_legacy_descriptor_proto
+
+   Compatibility slot to transition legacy dtypes to the new DType API.
+   Existing legacy dtypes that currently use ``PyArray_RegisterDataType``
+   should use this to transition to the new DType API.
+   The value of this slot is the ``PyArray_DescrProto`` struct.
+
+   This slot allows an *existing* legacy DType to use the new DType API
+   without breaking backwards compatibility (or with very minor changes).
+   I.e. NumPy will still consider this a "legacy" DType and use old code paths
+   where applicable, but use new features as they become available.
+
+   If used, this slot is required to be the first slot.
+
+    .. versionadded:: 2.5 (available starting NumPy 2.0)
+      This feature is added in NumPy 2.5, but is backported to be compatible
+      with NumPy 2.0+.  You can vendor this backport from ``npy_2_compat.h``
+      if you wish to compile with older NumPy versions.
+
+   .. note::
+      This slot exists for DTypes that currently use ``PyArray_RegisterDataType``
+      allowing them to transition when otherwise regressions would block
+      them from doing so.
+      It is a path for deprecating ``PyArray_RegisterDataType`` and
+      ``PyArray_RegisterCastFunc`` and ``PyArray_RegisterCanCast`` and further
+      slots in the future.
 
 .. c:macro:: NPY_DT_discover_descr_from_pyobject
 
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -3565,6 +3565,11 @@ Also see :ref:`dtypemeta` for documentation on ``PyArray_DTypeMeta`` and
  the examples in the ``numpy-user-dtypes`` repository for usage with both
  parametric and non-parametric data types.
 
+ .. note::
+
+    Custom DType registration is expected to happen during module import.
+    The registration API mutates global state and is not thread-safe.
+
 .. _dtype-flags:
 
 Flags
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -3613,7 +3613,10 @@ member of ``PyArrayDTypeMeta_Spec`` struct.
    I.e. NumPy will still consider this a "legacy" DType and use old code paths
    where applicable, but use new features as they become available.
 
-   If used, this slot is required to be the first slot.
+   If used, this slot is required to be the first slot.  The ``ArrFuncs``
+   fields are copied, but you may also set them via the slots below.
+   (This allows backporting e.g. sorts, although NumPy 2.4+ does support
+   a new way to implement sorts.)
 
    .. versionadded:: 2.5
 
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -1243,15 +1243,15 @@ User-defined data types
         The struct is not a valid Python object, so do not use ``Py_DECREF``
         on it.
 
-        **Transitioning to the new DType AP**
+        **Transitioning to the new DType API**
 
         Users currently using the old DType API may not be able to easily
         transition due to NumPy issues that would cause regressions.
         While new DTypes may reasonably limit themselves to newer NumPy
         versions, existing DTypes cannot do so.
 
-        You can transition to the new DType API by using
-        the `NPY_DT_legacy_descriptor_proto` slot when registering.
+        You can transition to the new DType API by using the
+        :c:macro:`NPY_DT_legacy_descriptor_proto` slot when registering.
         This slot is available when compiling with NumPy 2.5 and provides
         support for NumPy 2.0 and higher (it can also be vendored to compile
         on older versions of NumPy).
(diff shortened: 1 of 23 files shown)


doc/source/reference/c-api/array.rst
```diff
@@ -1243,6 +1243,19 @@ User-defined data types
         The struct is not a valid Python object, so do not use ``Py_DECREF``
         on it.
 
+        **Transitioning to the new DType AP**
```
ngoldbaum:
```suggestion
        **Transitioning to the new DType API**
```

ngoldbaum (commented):
I did a brief once-over and plan to do a more detailed pass later. I think I'm OK with this and the implementation ends up not being too bad.

That said, I'm a little nervous about merging this with zero tests.

Besides adding `ml_dtypes` as a test dependency, you could also add an example legacy user dtype that sets itself up using this API in the numpy-user-dtypes repository. We could then add numpy-user-dtypes as a git submodule of the numpy repository and use it in tests. I can see other situations where that would be useful.

Of course that doesn't straightforwardly test using this API from different numpy versions, but it at least lets us catch when we break dtypes using this API.

Failing that, I also think you might be able to use the `npy_rational` test datatype that's included in NumPy or maybe another user dtype that shares a lot of code with `npy_rational`.

seberg:
> Failing that, I also think you might be able to use the `npy_rational` test datatype that's included in NumPy

This was something I was definitely planning on doing.  I am not sure I am quite convinced of adding `ml_dtypes` yet, as it wouldn't even test this specific path (until it is used) at which point it is tested there (luckily `ml_dtypes` does test against NumPy dev).
Of course we could do it anyway, even if it doesn't actually test the additional stuff here...

MaanasArora:
I went through this in some detail and it all makes sense to me. I like the idea of being able to set up deprecations one by one with this and, given the backport constraint, this is probably as not hacky as it gets.

(Sorry if this is too far ahead but), would it make sense to use a legacy float type upgraded this way alongside quaddtype to comparatively test some of the trickier things, such as casting with python types? We can then compare/validate new-style paths in either case!

numpy/_core/src/multiarray/dtypemeta.c
```diff
@@ -324,8 +372,89 @@ dtypemeta_initialize_struct_from_spec(
...
+        descr->byteorder = legacy_proto->byteorder;
+        descr->_former_flags = 0;
+        descr->flags = legacy_proto->flags;
+        descr->elsize = legacy_proto->elsize;
+        descr->alignment = legacy_proto->alignment;
+        descr->metadata = NULL;
+        descr->hash = -1;
+        descr->reserved_null[0] = NULL;
+        descr->reserved_null[1] = NULL;
+        descr->subarray = legacy_proto->subarray;
```
ngoldbaum: Claude suggest that this alias can lead to reference count corruption if the realloc below fails. You could either grow `userdescrs` with realloc first or I think you could make this alias a `Py_NewRef` and adjust reference counts elsewhere.

numpy/_core/src/multiarray/dtypemeta.c
```diff
@@ -324,8 +372,89 @@ dtypemeta_initialize_struct_from_spec(
...
+
+        _PyArray_LegacyDescr **tmp = realloc(userdescrs,
+                (NPY_NUMUSERTYPES + 1) * sizeof(void *));
+        if (tmp == NULL) {
+            Py_DECREF(descr);
+            PyErr_NoMemory();
+            return -1;
+        }
+        userdescrs = tmp;
+        userdescrs[NPY_NUMUSERTYPES++] = descr;
```
ngoldbaum, line 434: This increment and the userdescrs realloc aren't thread-safe.

We should probably document somewhere that this whole API is intended to be used at import time and is not thread-safe.

numpy/_core/src/multiarray/public_dtype_api.c
```diff
@@ -75,12 +79,14 @@ PyArrayInitDTypeMeta_FromSpec(
         return -1;
     }
 
-    if (NPY_DT_SLOTS(DType)->setitem == NULL
-            || NPY_DT_SLOTS(DType)->getitem == NULL) {
-        PyErr_SetString(PyExc_RuntimeError,
-                "A DType must provide a getitem/setitem (there may be an "
-                "exception here in the future if no scalar type is provided)");
-        return -1;
+    if (!NPY_DT_is_legacy(DType)) {
```
ngoldbaum: This change won't be available at runtime on NumPy 2.0-2.4, so users will see different runtime behavior if they're running via the backport in the headers.

numpy/_core/include/numpy/npy_2_compat.h
```diff
@@ -249,6 +249,148 @@ DESCR_ACCESSOR(TYPEOBJ, typeobj, PyTypeObject *, 0)
...
+        if (slot->slot >= bad_offset && slot->slot < bad_offset + 30) {
+            slot->slot += good_offset - bad_offset;
+        }
+        slot++;
+    }
+#endif
+
+    if (proto == NULL || PyArray_RUNTIME_VERSION >= NPY_2_5_API_VERSION) {
+        return _PyArrayInitDTypeMeta_FromSpec(DType, spec);
+    }
```
ngoldbaum, line 315: It probably makes sense to move this check and early return before the `#if NPY_TARGET_VERSION < NPY_2_4_API_VERSION`, so calling this function and triggering an error doesn't mutate the `slots` state, but only on backports.

numpy/_core/src/multiarray/dtypemeta.c
```diff
@@ -222,7 +234,43 @@ dtypemeta_initialize_struct_from_spec(
...
+    PyArray_DescrProto *legacy_proto = NULL;
     PyType_Slot *spec_slot = spec->slots;
+    if (spec_slot != NULL
+            && spec_slot->slot == NPY_DT_legacy_descriptor_proto) {
+        legacy_proto = (PyArray_DescrProto *)spec_slot->pfunc;
+        spec_slot++;  /* consumed; the slot loop starts after this one */
+
+        DType->flags |= NPY_DT_LEGACY;
+        ((PyTypeObject *)DType)->tp_basicsize = sizeof(_PyArray_LegacyDescr);
+        ((PyTypeObject *)DType)->tp_new = (newfunc)legacy_dtype_default_new;
```
ngoldbaum, line 254: I think you need to call `PyType_Modified` now here: https://docs.python.org/3/c-api/type.html#c.PyType_Modified

Also probably worth adding a comment explaining why it's safe to do this.

numpy/_core/src/multiarray/dtypemeta.c
```diff
@@ -324,8 +372,89 @@ dtypemeta_initialize_struct_from_spec(
...
+        }
+
+        int typenum = NPY_USERDEF + NPY_NUMUSERTYPES;
+        if (typenum >= NPY_VSTRING) {
+            PyErr_SetString(PyExc_ValueError,
+                    "Too many user defined dtypes registered");
+            return -1;
+        }
+
+        _PyArray_LegacyDescr *descr = PyObject_Malloc(
```
ngoldbaum: This whole section of code is mostly identical to this code: https://github.com/numpy/numpy/blob/4d711763f8c5b978abd88cea710147c14d802675/numpy/_core/src/multiarray/usertypes.c#L271

Consider refactoring both into a single function exposed in `usertypes.h`.

ngoldbaum (commented):
Overall looking good but there are some issues I spotted.

IMO it's worth setting up a CI job that *builds* a test extension with NumPy 2.0 (and maybe the last stable release too?) and then runs it using numpy `main` to test the backport code paths.

numpy/_core/src/multiarray/public_dtype_api.c
seberg: Ah, I forgot to remove this.  I think it's fine to force setting these, even if it might be a bit duplicated.

numpy/_core/include/numpy/npy_2_compat.h
seberg: Unfortunately, one of the earlier PRs added a bug, this mutation is actually an important bug-fix for something like quaddtype in principle.

We didn't notice that, because in practice for new DTypes older versions of NumPy had always enough quirks that there was little reason in trying to make it work, but this is needed as a best bet unfortunately.

(it was a terrible slip indeed...)

numpy/_core/src/multiarray/dtypemeta.c
seberg: Thanks, I didn't know about this. I have no clue if this is needed for tp_new or not, but sounds like a good thing to add for sure!

seberg:
OK, I think I addressed those comments.  I also realized that I should have copied more slots over in both paths.  The rational tests should (mostly) cover that.
I added a test with `ml_dtypes`, I think the most interesting thing is to compile with the newst version and *run* on the old one, since the backport is only available by default there.
`ml_dtypes` can and probably should compile with older versions if they wish (by copying the compat header), but it'll be identical in practice to the compiling with the newest version.

(Of course right now, that test does very little... the rational tests are the interesting ones, but we would have to factor those out and run seperately, anyway, I may have to iterate on the CI test... this was added a bit on a whim right now)

EDIT: I'll probably do one more iteration tomorrow or so.

numpy/_core/include/numpy/npy_2_compat.h
ngoldbaum: Can you add a comment explaining that this is intentional so a future reader doesn't get confused? Even better if the comment links to the issue about the bug or PR fixing the issue.

numpy/_core/include/numpy/npy_2_compat.h
```diff
@@ -249,6 +249,178 @@ DESCR_ACCESSOR(TYPEOBJ, typeobj, PyTypeObject *, 0)
...
+
+#define _PyArrayInitDTypeMeta_FromSpec \
+    (*(int (*)(PyArray_DTypeMeta *, PyArrayDTypeMeta_Spec *))PyArray_API[362])
+#undef PyArrayInitDTypeMeta_FromSpec
+
+static inline int PyArrayInitDTypeMeta_FromSpec(
+        PyArray_DTypeMeta *DType, PyArrayDTypeMeta_Spec *spec)
+{
+    PyArray_DescrProto *proto = NULL;
+    if (spec->slots[0].slot == NPY_DT_legacy_descriptor_proto) {
```
ngoldbaum: claude points out that you added NULL checks for the `slots` in this PR but didn't add any NULL checks here. On purpose or an oversight? Seems harmless to add them.

ngoldbaum (approved):
All LGTM! Just one more minor issue but I'm going to approve. Thank you for adding all the new tests.

numpy/_core/include/numpy/npy_2_compat.h
seberg: Eh, that other NULL check is useless, but I guess we can give a nice error (and allow it here).

seberg: FWIW, made it so that passing NULL is the same as passing no slots (that is completely useless for now but seems like reasonable behavior).

numpy/_core/include/numpy/npy_2_compat.h
seberg: Expanded the earlier comment a little.

seberg:
Thanks for the review, going to put this in then.  I'll be testing this once more with `ml_dtypes` in the next week or two to make sure it works well.

PR #31304, opened 2026-04-22
Remove LICENSE workaround from pixi recipes

crusaderky:
The workaround is no longer needed after https://github.com/prefix-dev/pixi/pull/5776

This PR has also been tested downstream in tested downstream in https://github.com/crusaderky/cpython-pixi/pull/1


charris:
Thanks @crusaderky .

PR #31307, opened 2026-04-22
MAINT: Use Python 3.12 for dependabot pip updates.

charris:
Seems pytest has dropped support for older python versions. Make dependabot use a newer version. This is only needed for the pip ecosystem. The suggestion going forward is to always use NumPy's lowest supported version. 

#### AI Disclosure

Grok suggested this fix after looking at the dependabot log.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,6 +16,7 @@ updates:
     directory: /requirements
     schedule:
       interval: daily
+    python-version: "3.12"  # Dependencies are requiring newer python
     cooldown:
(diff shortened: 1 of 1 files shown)


charris:
The default Python version for dependabot is something like 3.9.

mattip:
Thanks @charris

charris:
Darn, Grok lied, that line is not supported by dependabot. The possible workarounds are complicated and uncertain,  so I think manual updates are the way to go.  I am going to revert this and do a manual update.

PR #31309, opened 2026-04-22
Fix recarray._set_dtype being None instead of callable

### LINKED ISSUE #31295: BUG: `recarray` doesn't have a proper `_set_dtype` method ?
### Describe the issue:

I'm trying to get rid of newly found deprecation warnings about `dtype` attribute setters in astropy.
There's one occurence in particular where I don't think we can use a view, so the next best thing would be `_set_dtype`, but it looks like it's not properly set on record arrays (it exists but it's a `None`)

```
self = <astropy.io.fits.hdu.table.BinTableHDU object at 0x10bf5f380>
data = rec.array([([ -939524096,           0],), ([ -637534208,  -939524096],),
           ([ -637534208, -1576992768],), ([ ...     ([-1090519040, -1157431040],), ([ -704643072,  2047082752],)],
          dtype=[('COMPRESSED_DATA', '<i4', (2,))])

    def _init_tbdata(self, data):
        columns = self.columns

        print(type(data))
>       data._set_dtype(data.dtype.newbyteorder(">"))
E       TypeError: 'NoneType' object is not callable

astropy/io/fits/hdu/table.py:212: TypeError
```

```python
import numpy as np
a = np.array([1], dtype="float64").view(np.rec.recarray)
a._set_dtype("float64")
```

### Error message:

```shell
  File "/Users/clm/dev/gh/neutrinoceros/astropy-compatibility-checks/numpy-dev/_astropy/t.py", line 3, in <module>
    a._set_dtype("float64")
    ~~~~~~~~~~~~^^^^^^^^^^^
TypeError: 'NoneType' object is not callable
```

```
2.5.0.dev0+git20260421.388bd75
3.14.3 (main, Feb  4 2026, 01:51:49) [Clang 21.1.4 ]
```

### Runtime Environment:

_No response_

_No response_

armorbreak001:
## Summary

The `recarray` class assigns `_set_dtype = None` as a class attribute (line 405), with the comment that `__array_finalize__` handles dtype changes. However, external code such as astropy calls `_set_dtype` directly on recarray instances, which raises `TypeError: 'NoneType' object is not callable`.

## Change

Replace the `None` assignment with a proper method that wraps void dtypes with the record type before delegating to `ndarray._set_dtype`. This mirrors the existing logic in `__array_finalize__` (lines 414-417) and is consistent with how `MaskedArray` handles the same pattern (`numpy/ma/core.py:3491-3492`).

## Test case (from issue)

```python
import numpy as np
a = np.array([1], dtype="float64").view(np.rec.recarray)
a._set_dtype("float64")  # previously: TypeError
```

Closes #31295

### DIFF
--- a/numpy/_core/records.py
+++ b/numpy/_core/records.py
@@ -402,7 +402,13 @@ class recarray(ndarray):
             )
         return self
 
-    _set_dtype = None  # __array_finalize__ can deal with dtype changes
+    def _set_dtype(self, dtype):
+        """Set the dtype, wrapping in record type if needed."""
+        if (dtype.type is not record and
+                issubclass(dtype.type, nt.void) and
+                dtype.names is not None):
+            dtype = sb.dtype((record, dtype))
+        ndarray._set_dtype(self, dtype)
 
     def __array_finalize__(self, obj):
         if (self.dtype.type is not record and


jorenham:

PR #31312, opened 2026-04-22
MAINT: Update some requirements.

charris:
Dependabot was not able to update these because it runs Python 3.9 and these packages have dropped support for that version. Dependabot might work if it read the pyproject.toml file to get the Python version, but then it would scan everything, and I am not ready for that experiment.

This might be a bit aggressive, we will see.

#### AI Disclosure

I used Grok to find the latest releases of a list of dependencies.

### DIFF
--- a/requirements/typing_requirements.txt
+++ b/requirements/typing_requirements.txt
@@ -2,5 +2,5 @@
 
 -r test_requirements.txt
 
-mypy==1.20.0
+mypy==1.20.2
 pyrefly==0.61.0
--- a/environment.yml
+++ b/environment.yml
@@ -24,7 +24,7 @@ dependencies:
   - pytest-xdist
   - hypothesis
   # For type annotations
-  - mypy=1.20.0
+  - mypy==1.20.2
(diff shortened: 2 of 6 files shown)


andyfaff:
The pkgconf is pinned to post1, post2 has issues.

charris:
> The pkgconf is pinned to post1, post2 has issues.

Fixed.

requirements/typing_requirements.txt
```diff
@@ -2,5 +2,5 @@
 
 -r test_requirements.txt
 
-mypy==1.20.0
+mypy==1.20.2
```
jorenham, line 5: this should probably also be updated in `environment.yml` then:
https://github.com/numpy/numpy/blob/9546c7fe4d81a0a02ae224aa60a3c493db2ba92a/environment.yml#L27

charris: Done. The `=` is a typo, that is why I didn't see it.

jorenham (approved):

PR #31277, opened 2026-04-19
DOC/TYP: Improved typing docs for ``ndarray``

jorenham:
These docs were outdated and did not describe how  `numpy.ndarray` can be used in a static typing context.

#### AI Disclosure
N/A

PR #31313, opened 2026-04-22
DOC/TYP: Move ``ndarray`` typing docs to a more appropriate section

jorenham:
I just realized that the `ndarray` typing doc changes in #31277 made it so that this subsection no longer makes a lot of sense under "Differences from the runtime NumPy API" section, so this moves it to the "API" section.

#### AI Disclosure

N/A


jorenham:
nice PR number, btw

jorenham:
Oh wait, the TOC doesn't look right:
image

I'll try something else

charris:
Thanks Joren.

PR #23302, opened 2023-02-28
DEP: deprecate `np.round_`; add `round`/`min`/`max` to the docs

rgommers:
In addition to adding  `np.round`, `np.min` and `np.max` to the html docs, this PR makes those the preferred aliases - as discussed/agreed in gh-13877.    

Closes gh-22617
Closes gh-19717
Closes gh-13877

PR #31315, opened 2026-04-23
DOC: use base min and max rather than aliases amin and amax in see-also

edbennett:
### PR summary

Currently, the documentation for `numpy.max` has a see-also to `numpy.amin`, which only states that it's an alias of `numpy.min`, so you have to click through a second time to see the actual detail. Similarly in reverse from `numpy.min` to `numpy.amax` to `numpy.max`.

This seems to be left over from when `numpy.amax` and `numpy.amin` were the standard, documented forms. This PR changes these so that the intermediary page on the alias is skipped, and you go straight from the `numpy.max` documentation to the `numpy.min` documentation, similarly to before #23302 was merged.

I am a research software engineer using Numpy heavily both directly and indirectly in my work in lattice quantum field theory. In this instance however I was teaching a [Software Carpentry lesson](https://swcarpentry.github.io/python-novice-inflammation) that makes use of both `amax` and `max`, and was looking up the reasoning behind the aliases, when I found the circuitous cross-references that this PR addresses.

#### AI Disclosure

No AI tools used.

### DIFF
--- a/numpy/_core/fromnumeric.py
+++ b/numpy/_core/fromnumeric.py
@@ -3058,7 +3058,7 @@ def max(a, axis=None, out=None, keepdims=np._NoValue, initial=np._NoValue,
 
     See Also
     --------
-    amin :
+    min :
         The minimum value of an array along a given axis, propagating any NaNs.
     nanmax :
         The maximum value of an array along a given axis, ignoring any NaNs.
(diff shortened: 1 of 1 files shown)


ngoldbaum:
Thanks! Please do keep on sending in documentation cleanups like this, if you feel like it.

edbennett:
Shall do, thanks for merging so quickly!

PR #31316, opened 2026-04-23
DEP: Narrow dtype deprecation subclass handling even further

### LINKED ISSUE #31295: BUG: `recarray` doesn't have a proper `_set_dtype` method ?

seberg:
This goes to a bit extreme lengths, but avoids regressions for weird subclasses.
Is it pretty? No, but the worst of it is just a transition and most subclasses shouldn't implement `dtype` attributes hopefully...

So hopefully this is mainly a hack to make astropy not fail on newer NumPy right away and maybe one more project...

Closes gh-31295 (but does not adjust what is reported in the title; `_set_dtype` should be used as a last resort by subclasses mainly; I'll consider it OK to break weirder things here as one can work around it.  What it *does* prevent is a regression: you will get the DeprecationWarning only.)

CC @mhvk I am tempted to go with this, even though it is rather extreme...  Deprecating `.dtype` attribute setting is a big step, so I am willing to go through extreme lengths to make sure we don't stumble over something.

### DIFF
--- a/numpy/_core/src/multiarray/convert.c
+++ b/numpy/_core/src/multiarray/convert.c
@@ -526,6 +526,127 @@ PyArray_NewCopy(PyArrayObject *obj, NPY_ORDER order)
     return (PyObject *)ret;
 }
 
+
+static int
+get_optional_set_dtype_and_dtype(
+    PyTypeObject *subtype,
+    PyObject **_set_dtype, PyObject **sub_dtype)
+{
+    if (PyObject_GetOptionalAttr(
+            (PyObject *)subtype, npy_interned_str._set_dtype,
+            _set_dtype) < 0) {
+        return -1;
+    }
+    if (PyObject_GetOptionalAttr(
+            (PyObject *)subtype, npy_interned_str.dtype, sub_dtype) < 0) {
+        Py_XDECREF(*_set_dtype);
+        return -1;
+    }
+    return 0;
+}
+
+
+/* Pick how a view() dtype change is propagated for `subtype`.
+ *
+ * Sets (at most) one of the three output flags; all three zero means
+ * the legacy in-place path (subclass can't observe).
+ *
+ *   use_dtype_in_finalize  -- `_set_dtype = None`: class wants
+ *                             __array_finalize__ to see the final dtype.
+ *   use_set_dtype          -- subclass `_set_dtype` method: call it
+ *                             as an adjust hook after viewing.
+ *   use_dtype_prop         -- subclass `dtype` descriptor wins over
+ *                             `_set_dtype`: call the setter (deprecated).
+ *
+ * Walk the MRO and see whether `_set_dtype` or `dtype` was overridden
+ * more specifically: whichever's resolved value diverges from what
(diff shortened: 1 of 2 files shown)


mhvk (approved):
@seberg - I pushed some small changes, mostly to make the default case be what we want for the future. I also split out getting both attributes for brevity. I think this looks good and confirm that in astropy it now gives only deprecation warnings.

numpy/_core/src/multiarray/convert.c
```diff
@@ -526,6 +526,126 @@ PyArray_NewCopy(PyArrayObject *obj, NPY_ORDER order)
...
+    PyObject **_set_dtype, PyObject **sub_dtype)
+{
+    if (PyObject_GetOptionalAttr(
+            (PyObject *)subtype, npy_interned_str._set_dtype,
+            _set_dtype) < 0) {
+        return -1;
+    }
+    if (PyObject_GetOptionalAttr(
+            (PyObject *)subtype, npy_interned_str.dtype, sub_dtype) < 0) {
+        return -1;
```
seberg, line 543:
```suggestion
        Py_XDECREF(_set_dtype);
        return -1;
```
But the refactor itself looks very nice.  Will wait a at least until tomorrow with merging though.

mhvk: Oops, yes, now done.

seberg:
Oh, I hand't merged this yet... Time to get this in if nobody else had an opinion.  Hopefully this is the last of it...

PR #31317, opened 2026-04-23
MAINT: simplify view implementations in MaskedArray and recarray.

mhvk:
These are possible after the changes in `ndarray.view()` in #31234 (and taken from the initial version of #31271). Also used the occasion to do `super().view` rather than calling `ndarray.view` (which predates the introduction of `super` in python...).

No AI used.

### DIFF
--- a/numpy/ma/core.py
+++ b/numpy/ma/core.py
@@ -3256,28 +3256,12 @@ class MaskedArray(ndarray):
         results.
         """
 
-        if dtype is None:
-            if type is None:
-                output = ndarray.view(self)
-            else:
-                output = ndarray.view(self, type)
-        elif type is None:
-            try:
-                if issubclass(dtype, ndarray):
-                    output = ndarray.view(self, dtype)
-                    dtype = None
-                else:
-                    output = ndarray.view(self, dtype)
-            except TypeError:
-                output = ndarray.view(self, dtype)
-        else:
-            output = ndarray.view(self, dtype, type)
+        if type is None and (isinstance(dtype, builtins.type)
+                             and issubclass(dtype, ndarray)):
+            type = dtype
+            dtype = None
 
-        # also make the mask be a view (so attr changes to the view's
-        # mask do no affect original object's mask)
-        # (especially important to avoid affecting np.masked singleton)
-        if getmask(output) is not nomask:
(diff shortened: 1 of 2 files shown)


numpy/ma/core.py
```diff
@@ -3256,28 +3256,12 @@ def view(self, dtype=None, type=None, fill_value=None):
...
-            except TypeError:
-                output = ndarray.view(self, dtype)
-        else:
-            output = ndarray.view(self, dtype, type)
+        if type is None and (isinstance(dtype, builtins.type)
+                             and issubclass(dtype, ndarray)):
+            type = dtype
+            dtype = None
 
-        # also make the mask be a view (so attr changes to the view's
```
mhvk, line 3276: This protection of the mask is not actually needed, as `__array_finalize__` already ensures it is a view (in fact, a double view given the reshape).

numpy/ma/core.py
```diff
@@ -3256,28 +3256,12 @@ def view(self, dtype=None, type=None, fill_value=None):
...
-        # mask do no affect original object's mask)
-        # (especially important to avoid affecting np.masked singleton)
-        if getmask(output) is not nomask:
-            output._mask = output._mask.view()
+        output = super().view(*[a for a in (dtype, type) if a is not None])
```
seberg, line 3264: Hahaha, I didn't want to believe that this dance is necessary, but it is...
(Might be nice to fix the C-side to just accept `None` as "unchanged", but doesn't matter really.)

numpy/ma/core.py
seberg: Looks good yeah. I suppose in some earlier version maybe dtype attribute setting didn't ensure this part. But I agree, this looks safe now.

seberg (approved):
LGTM and much cleaner/nicer.

Maybe briefly, I am missing a bit that the ability to refactor like this is new, the only bigger part to me seems to rely on `super().view()` in `mrecords`?

mhvk:
I agree I'm actually not 100% sure it is new, I wrote it to deal with the errors introduced by having `dtype` being passed on to `__array_finalize__`. Anyway, it is a lot shorter for `MaskedRecord` at least. Just as a last sanity check, I tested against astropy, and everything passes. So will merge...

PR #31319, opened 2026-04-24
BUG: fix `np.ones` to raise ValueError on creation for generic datetime64

andrewgtgt:
### PR summary
Fixes #30903 where `np.ones` does not raise ValueError for generic datetime64 on creation.
Since the type-checking pipeline is skipped for code like `np.ones(1, "M8")` from the example provided in the bug report, this fix runs a check in get_legacy_dtype_cast_function to stop any generic datetime64 arrays other than NaT, throwing a ValueError if found.

This fix passes most tests, including the simple regression test I made for this bug that previously failed. However, this fix failed a few tests when I ran them all locally with `spin test`, and these tests appear to rely on the functionality of not throwing an error at that point. For example, an error is thrown on line 1541 here:

https://github.com/numpy/numpy/blob/e87a3c5fef53f5e1256f5d6f50561592da82992a/numpy/_core/tests/test_nditer.py#L1540-L1546

And here are other such lines in tests causing errors:

> numpy/_core/tests/test_multiarray.py::TestMethods::test_dot_out_mem_overlap
```python
a = np.random.rand(3, 3).astype(dtype)
```
> numpy/_core/tests/test_multiarray.py::TestMethods::test_argpartition_out_of_range
> numpy/_core/tests/test_multiarray.py::TestMethods::test_partition_out_of_range
```python
d = np.arange(10).astype(dtype=dtype)
```
> numpy/_core/tests/test_ufunc.py::test_ufunc_types
```python
args = [np.ones((3, 3), t) for t in inp]
```
> numpy/_core/tests/test_casting_unittests.py::TestCasting::test_numeric_to_times
```python
cast._simple_strided_call((arr1, arr2))
```

Are the lines in these tests directly at odds with the change in behavior requested by the bug fix, or is there perhaps a better way a fix for this could be implemented that would avoid changing the tests? Any advice on how to proceed would be appreciated!

Hello, I'm Andrew, a university student looking to contribute to open source for experience and as part of my education. This is my first contribution and I am not yet that familiar with GitHub, so I welcome any and all feedback, not limited to just this issue.

#### AI Disclosure
GPT-5.3 was used to reason about the pipeline of how datetime64 arrays were created to compare locations at which to implement a fix. GPT-5.3 was also used to explain the datatypes at this location, and provide valid ways to check them to ensure only only invalid generic units would raise the desired error.


seberg:
Thanks, @riku-sakamoto might have some thoughts.  I would prefer to just include this in the generic deprecation warning we have now.
That said, I would need to look at it once more to recall the setup (i.e. why isn't it already included?).

(Since it isn't yet... it seems that there is still a bit more ground to cover, but maybe it's OK if that misses 2.5)

numpy/_core/src/multiarray/dtype_transfer.c
```diff
@@ -1270,6 +1270,21 @@ get_legacy_dtype_cast_function(
...
+            get_datetime_metadata_from_dtype(dst_dtype);
+            
+        if (meta != NULL && meta->base == NPY_FR_GENERIC) {
+            PyErr_SetString(PyExc_ValueError,
+                "Cannot convert a NumPy datetime value "
+                "other than NaT with generic units");
+            return NPY_FAIL;
+        }
+    }
+
```
riku-sakamoto: If my understanding is correct, `get_legacy_dtype_cast_function` is the method to create "cast" functions between two dtypes. So, when we want to avoid any unitless datetime dtypes, modifying `create_datetime_dtype` seems more straightforward.

https://github.com/numpy/numpy/blob/fd34ba1bda921815c3be1654a84f627f85278f2c/numpy/_core/src/multiarray/datetime.c#L589-L615

If so, we can fix others like `np.empty(1, dtype="M8")` as well. (In your current implementation, it does not raise any warnings and errors.)

riku-sakamoto (commented):
Thanks! @andrewgtgt 

I am still wondering and investigating why the following path is not called when `copyto` or `empty` is called with unitless datetime dtypes. Do you happen to know why?

https://github.com/numpy/numpy/blob/fd34ba1bda921815c3be1654a84f627f85278f2c/numpy/_core/src/multiarray/datetime.c#L2384-L2390


> Are the lines in these tests directly at odds with the change in behavior requested by the bug fix, or is there perhaps a better way a fix for this could be implemented that would avoid changing the tests?

It seems that changing these tests is the only way to fix the problem.

andrewgtgt:
> I am still wondering and investigating why the following path is not called when `copyto` or `empty` is called with unitless datetime dtypes. Do you happen to know why?

No, but it looks like the entire `convert_pyobject_to_datetime` function is not called for `np.ones(1, "M8")`.

#31054 claimed that "`ones()` and `ones_like()` fill the allocated array using `multiarray.copyto(a, 1, casting='unsafe')`. The raw Python integer 1 is copied directly into memory, bypassing NumPy's datetime type-checking pipeline entirely." And it seemed like it was going through some legacy path or something instead.

> If my understanding is correct, `get_legacy_dtype_cast_function` is the method to create "cast" functions between two dtypes. So, when we want to avoid any unitless datetime dtypes, modifying `create_datetime_dtype` seems more straightforward.

I agree, this is a more appropriate location for the check. The last commit moves the check to `create_datetime_dtype`, and now it passes cases including `np.empty(1, "M8")` that verify a ValueError is thrown. However, I think it isn't possible at this stage to discriminate whether or not `NaT` will be assigned, so that would also be disallowed with the newer implementation.

andrewgtgt:
Closing this PR as I am not sure how to cleanly throw a ValueError in the creation phase for generic datetime64, while also not doing so for NaT, as requested by the bug fix.

If someone thinks this would be valuable and desired functionality, feel free to continue based off of this discussion and/or take your own stab at it.

PR #31320, opened 2026-04-24
BUG: fix memory leak in np.zeros when fill-zero loop raises

MaartenBaert:
### PR summary

This PR fixes a memory leak in np.zeros that occurs when the fill-zero loop raises an exception (which can happen for custom user-defined dtypes). It does this by setting NPY_ARRAY_OWNDATA earlier to ensure the data will be deallocated, combined with changes to mem_handler cleanup to avoid a double-decref.

I'm not entirely sure whether this is the most appropriate way to fix the leak, but it does seem to work correctly based on testing with `tracemalloc`.

#### AI Disclosure

The bug was detected through testing with `tracemalloc`, then located and fixed by Claude Opus 4.7.
Full AI-generated description: https://gist.github.com/MaartenBaert/f3eadd280f84ef6896586025f59788ec

### DIFF
--- a/numpy/_core/src/multiarray/arrayobject.c
+++ b/numpy/_core/src/multiarray/arrayobject.c
@@ -449,10 +449,10 @@ _clear_array_attributes(PyArrayObject *self, npy_bool unraisable)
                 nbytes = 1;
             }
             PyDataMem_UserFREE(fa->data, nbytes, fa->mem_handler);
-            Py_CLEAR(fa->mem_handler);
         }
         fa->data = NULL;
     }
+    Py_CLEAR(fa->mem_handler);
 
     /* must match allocation in PyArray_NewFromDescr */
     npy_free_cache_dim(fa->dimensions, 2 * fa->nd);
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -906,6 +906,13 @@ PyArray_NewFromDescr_int(
             raise_memory_error(fa->nd, fa->dimensions, descr);
             goto fail;
         }
+        /*
+         * Set fa->data and NPY_ARRAY_OWNDATA immediately after allocation so
+         * that the fail path (Py_DECREF(fa) -> dealloc) frees the buffer if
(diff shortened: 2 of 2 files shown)


numpy/_core/src/multiarray/arrayobject.c
```diff
@@ -449,10 +449,10 @@ _clear_array_attributes(PyArrayObject *self, npy_bool unraisable)
                 nbytes = 1;
             }
             PyDataMem_UserFREE(fa->data, nbytes, fa->mem_handler);
-            Py_CLEAR(fa->mem_handler);
         }
         fa->data = NULL;
     }
+    Py_CLEAR(fa->mem_handler);
```
seberg, line 455:
```suggestion
    Py_XCLEAR(fa->mem_handler);
```
I guess data can be NULL but `mem_handler` already set (or OWNDATA not).  Makes all sense, but `mem_handler` can clearly be NULL as well.

(Needs to be applied, but I can just do that.)

numpy/_core/src/multiarray/ctors.c
```diff
@@ -998,7 +1003,6 @@ PyArray_NewFromDescr_int(
 
  fail:
     NPY_traverse_info_xfree(&fill_zero_info);
-    Py_XDECREF(fa->mem_handler);
```
seberg, line 1001: Hah, "fun" that the old logic worked out here because the owndata flag wasn't set yet and `mem_handler` was never deallocated.

seberg (approved):
Thanks a lot, looks good to me!  (Will hold off merging a moment.)

I guess adding a test that fails during the zero filling is annoying, but your DType tested this (maybe accidentally, so?).
(Our CI now has leak sanitizer running even, but we have no dtype that can be made to fail unless we write one just for this...)

numpy/_core/src/multiarray/arrayobject.c
seberg: Sorry... silly me, there is no XCLEAR, clear always includes a NULL check.

seberg:
OK, let's put this in, thanks again.

PR #31324, opened 2026-04-24
MAINT: Refactor subarray handling in _set_dtype, NewFromDescr, and views

mhvk:
A further follow-up of the way views are done and dtypes are set, specifically targeting dtypes with subarrays.

Subarrays were handled very similarly in `_set_dtype` and `NewFromDescr_int`, so the first commit factors out the common code. The second commit then calls that from the new dtype checking routine, so it also gets used in `.view()`.

@seberg - this probably wasn't worth the nearly a day I spent on it, but it does seem to make things cleaner, and hopefully it makes sense to you too...

No AI used.

### DIFF
--- a/numpy/_core/src/multiarray/common.c
+++ b/numpy/_core/src/multiarray/common.c
@@ -18,6 +18,7 @@
 #include "get_attr_string.h"
 #include "mem_overlap.h"
 #include "array_coercion.h"
+#include "alloc.h"  /* for npy_alloc_cache_dim */
 
 /*
  * The casting to use for implicit assignment operations resulting from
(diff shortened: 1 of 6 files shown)


numpy/_core/src/multiarray/common.c
```diff
@@ -370,19 +371,95 @@ _may_have_objects(PyArray_Descr *dtype)
...
 
+/*
+ * Get the number of dimensions of a subarray.  If added to nd,
+ * it exceeds NPY_MAXDIMS, return -1 with an exception set.
+ */
+static inline int
+_get_subarray_ndim(const PyArray_Descr *descr, const int nd)
+{
+    PyObject *shape = PyDataType_SUBARRAY(descr)->shape;
+    assert(shape);
```
mhvk, line 382: It would be for follow-up, but I wonder if by now we cannot simply assume `shape` is a tuple (or deprecate non-tuple?)

seberg: Yeah, I think that is fine. It seems ridiculous to think that a user-dtype can reasonably set `subarray`.  And it also seems unlikely that there isn't some code path that relies on it anyway.
(Fields bad enough, but I know there is one user-dtype out there that uses it to create a custom `object` dtype.)

numpy/_core/src/multiarray/getset.c
```diff
@@ -354,55 +354,23 @@ array_descr_set_internal(PyArrayObject *self, PyObject *arg)
...
-        Py_DECREF(temp);
-    }
-    else if (newlastdim >= 0) {
-        int lastaxis = PyArray_NDIM(self) - 1;
-        PyArray_DIMS(self)[lastaxis] = newlastdim;
-        PyArray_STRIDES(self)[lastaxis] = newlaststride;
+        ((PyArrayObject_fields *)self)->nd = newnd;
+        ((PyArrayObject_fields *)self)->dimensions = newdims;
+        ((PyArrayObject_fields *)self)->strides = newstrides;
     }
```
seberg, line 374: Might want to add an `else {assert(newdims == NULL);}`.  subarrays cannot have a an empty shape, so it's impossible to reach a case where `newdims` isn't NULL. But it isn't obvious.

(Had an agent have a look as a quick thing, and it noticed this nitpicky thing)

seberg (approved):
LGTM, had an agent look as these things are getting good about finding subtle things.  It noted that this changed:
```
np.zeros((), dtype=("f8", (3, 4)), order="F").flags
```
To be C rather than F-contiguous.

That is actually *correct* and a super niche bug-fix, I think.  If you like, we could add a test for it, but also happy if not.

numpy/_core/src/multiarray/common.c
mhvk: I had a quick check and there is a [test](https://github.com/numpy/numpy/blob/d4e5fafa28f4198c535e64ec18584f9ffd180d4d/numpy/_core/tests/test_multiarray.py#L4803-L4819) that checks the integer shape, so better for another time...

seberg: It's certainly supported as user input, but the created dtype will still contain a tuple there.

mhvk:
> LGTM, had an agent look as these things are getting good about finding subtle things. It noted that this changed:
> 
> ```
> np.zeros((), dtype=("f8", (3, 4)), order="F").flags
> ```
> 
> To be C rather than F-contiguous.
> 
> That is actually _correct_ and a super niche bug-fix, I think. If you like, we could add a test for it, but also happy if not.

Hmm, I think this means I actually introduced a bug(let). On master,
```
data = ([[1, 2, 3], [4, 5, 6]])
arr = np.full((), data, dtype=("f8", (2, 3)), order="F")
arr.flags
# C_CONTIGUOUS : False
# F_CONTIGUOUS : True
arr.strides
# (8, 16)
```
so it gives an actual fortran-order array, as requested.

But with this PR, it would change, even though for any non-empty shape it remains fortran-ordered. So, I better fix it...

numpy/_core/src/multiarray/getset.c
mhvk: OK, done (though note that `newdims` gets set to the old dims in this case, which is what I now assert).

numpy/_core/src/multiarray/common.c
mhvk: Yes, but that test checks a binary blob, and was initially added at the time `shape` was always made a tuple (see https://github.com/numpy/numpy/commit/b12429704242314703797ed2076d3036d19327eb). So, I worry that it is to ensure backward compatibility (but possibly that can be done more surgically?).

seberg:
Not sure that I think it strictly _should_ give an F-order array, but as that is what we always had happy to just keep it and test this (although I would be fine with changing it in the future).
(But diverging for 0-D is no good of course, which I guess was the previous state.)

From my side, this is still good (although now I am worried if my `assert` thought was bad because without assertion a compiler might warn about an empty branch...)

mhvk:
> From my side, this is still good (although now I am worried if my `assert` thought was bad because without assertion a compiler might warn about an empty branch...)

Compilers seem to be happy enough... Shall we put it in?

seberg:
Yeah, sure, thanks.  If they are not, easy follow up...

PR #31331, opened 2026-04-25
BLD: use batched C99 function detection to speed up configure stage of build

rgommers:
This reintroduces the `feature_detection_math.h` and `feature_detection_cmath.h` that we had in the distutils build (files reworked but adapted the idea from the 1.25.x branch - improved to avoid undefs and pragmas). It speeds up the configure stage significantly, especially on Windows where these checks are quite slow. On macOS the relative speedup is ~30%.

The other benefit is to significantly reduce the length of the build log printed to stdout, so it's easier to find things in it.

Note that the `_math.h` header in 1.25.x had `float` and `long double` signatures, but those were never used, so they're left out here.

The most significant change from 1.25.x is to use:
```C
static const funcptr funcs[] = { (funcptr)sin, ... };
```
to take a static address of each function, which forces the linker to resolve each function in the batch check with `cc.links`.

I also tried using batch checks for other groups of functions (e.g., `optional_file_funcs`) like in 1.25.x, but decided that the extra complexity in `_core/meson.build` wasn't worth it, so ended up reverting those changes. 

#### AI Disclosure

I used Claude Opus 4.7 to implement most of these changes, starting by adapting the headers from the 1.25.x branch, then getting rid of the `undef`'s and `pragma`s, and then implementing the batched checks. I then manually cleaned up some of the code and comments in the process of reviewing the code.

### DIFF
--- a/numpy/_core/feature_detection_cmath.h
+++ b/numpy/_core/feature_detection_cmath.h
@@ -1,0 +1,95 @@
+/*
+ * Prototypes for C99 complex-math functions probed as a batch by meson.build.
+ *
+ * For MSVC we must include <complex.h> to pick up the non-standard
+ * _Fcomplex/_Dcomplex/_Lcomplex typedefs, since MSVC doesn't support the C99 `_Complex`
+ * keyword directly. For every other compiler we avoid including <complex.h> so that
+ * macro-based declarations (notably musl's `#define crealf(x) ((float)(x))`) don't
+ * conflict with the plain prototypes below.
+ */
+
+#if defined(_MSC_VER) && !defined(__INTEL_COMPILER)
+#include <complex.h>
+typedef _Fcomplex cfloat;
+typedef _Dcomplex cdouble;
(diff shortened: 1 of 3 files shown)


rgommers:
The one failure is unrelated, that is due to a new MKL release yesterday.

I checked some CI logs: the batched checks work on Linux, macOS and Windows (as expected) and fail on FreeBSD (also as expected, because it's missing some complex routines). From [this CI log](https://cirrus-ci.com/task/4881297832673280?logs=build#L162):
```
  Checking if "mandatory libm functions (batched)" with dependency -lm links: YES
  Checking if "C99 complex funcs (batched)" with dependency -lm links: NO
  Checking for function "cabs" with dependency -lm: YES
  Checking for function "cabsf" with dependency -lm: YES
```
that shows that the fallback path works as it should.

seberg (approved):
LGTM, and nice to speed up! In case someone else who knows this better wants to have a look, will wait for a bit, but feel free to just merge.

PR #31332, opened 2026-04-26
ENH: allow larger than C int sized structured dtypes

### LINKED ISSUE #31308: ENH: extremely large single field size dtype support
I suspect this is already known, but I'm not seeing this clearly expressed in open/closed issues? For context the reproducer below came up in an in person discussion with @braxtoncuneo at a meeting at Vanderbildt University, where he was explaining that this limitation was causing some friction in the context of passing data from GPU to CPU (working with NumPy in numba specifically). If this were possible to do, they could store all of their state in a single "struct" and it would make things easier.

I'm not making a comment on whether this should be supported, but passing along some feedback from researchers in case the core team feedback might be helpful.

The single field size case is problematic:

```python
>>> kind = [("x", np.float64, 2**28)]
>>> kind_dtype = np.dtype(kind)
  File "<python-input-22>", line 1, in <module>
    kind_dtype = np.dtype(kind)
ValueError: invalid shape in fixed-type tuple: dtype size in bytes must fit into a C int.
```

The aggregated size case is "fine" it seems, even when it exceeds the sum total size of the reproducer above:

```python
>>> kind = [("x", np.float64, 2**28 - 1), ("y", np.float64, 1000)]
>>> kind_dtype = np.dtype(kind)
```

tylerjereddy:
* Fixes gh-30315 
* Fixes gh-31308

* ~Although the original failing regression test does pass here, it is not even close to safe to do this yet even though the full testsuite passes locally with `test_pickling_large` re-enabled.~

~I spent a few hours debugging on this one--I'll add a few initial inline self-review comments to start...~

#### AI Disclosure
~No AI tools used~ Sebastian did use AI support to review/improve the reach of the work in the code base, since many places use i.e,. `elsize` and so on. Tyler used Claude Opus 5 to select its top two concerns after carefully reviewing the PR and Sebastian's comments.

### DIFF
--- a/numpy/_core/src/multiarray/array_assign_array.c
+++ b/numpy/_core/src/multiarray/array_assign_array.c
@@ -42,7 +42,10 @@ copycast_isaligned(int ndim, npy_intp const *shape,
     int aligned;
     int big_aln, small_aln;
 
-    int uint_aln = npy_uint_alignment(dtype->elsize);
+    int uint_aln = 0;
+    if (dtype->elsize <= NPY_MAX_INT) {
+        uint_aln = npy_uint_alignment((int)dtype->elsize);
+    }
     int true_aln = dtype->alignment;
 
     /* uint alignment can be 0, meaning not uint alignable */
--- a/numpy/_core/src/multiarray/convert_datatype.c
+++ b/numpy/_core/src/multiarray/convert_datatype.c
@@ -2220,7 +2220,14 @@ get_byteswap_loop(
     PyArray_Descr *const *descrs = context->descriptors;
     assert(descrs[0]->kind == descrs[1]->kind);
     assert(descrs[0]->elsize == descrs[1]->elsize);
-    int itemsize = descrs[0]->elsize;
+    if (descrs[0]->elsize > NPY_MAX_INT) {
+        // TODO: add a regression test for >INT_MAX byte-swap itemsize support.
+        PyErr_SetString(PyExc_TypeError,
+                "byte-swapping with itemsize larger than INT_MAX "
+                "is not supported");
+        return -1;
+    }
+    int itemsize = (int)descrs[0]->elsize;
     *flags = NPY_METH_NO_FLOATINGPOINT_ERRORS;
     *out_transferdata = NULL;
     if (descrs[0]->kind == 'c') {
@@ -2958,20 +2965,16 @@ structured_to_nonstructured_resolve_descriptors(
 
     /* Void dtypes always do the full cast. */
     if (given_descrs[1] == NULL) {
+        /*
+         * Reject string casts, historically these worked only for empty arrays.
+         * (One could go via object and hex printing, but...)
+         */
+        if (dtypes[1]->type_num == NPY_STRING || dtypes[1]->type_num == NPY_UNICODE) {
+            return -1;
+        }
         loop_descrs[1] = NPY_DT_CALL_default_descr(dtypes[1]);
         if (loop_descrs[1] == NULL) {
             return -1;
-        }
-        /*
-         * Special case strings here, it should be useless (and only actually
-         * work for empty arrays).  Possibly this should simply raise for
-         * all parametric DTypes.
-         */
-        if (dtypes[1]->type_num == NPY_STRING) {
-            loop_descrs[1]->elsize = given_descrs[0]->elsize;
-        }
-        else if (dtypes[1]->type_num == NPY_UNICODE) {
-            loop_descrs[1]->elsize = given_descrs[0]->elsize * 4;
         }
     }
     else {
--- a/numpy/_core/src/multiarray/convert_datatype.c
+++ b/numpy/_core/src/multiarray/convert_datatype.c
@@ -2331,7 +2331,14 @@ get_byteswap_loop(
     PyArray_Descr *const *descrs = context->descriptors;
     assert(descrs[0]->kind == descrs[1]->kind);
     assert(descrs[0]->elsize == descrs[1]->elsize);
-    int itemsize = descrs[0]->elsize;
+    if (descrs[0]->elsize > NPY_MAX_INT) {
+        // TODO: add a regression test for >INT_MAX byte-swap itemsize support.
+        PyErr_SetString(PyExc_TypeError,
+                "byte-swapping with itemsize larger than INT_MAX "
+                "is not supported");
+        return -1;
+    }
+    int itemsize = (int)descrs[0]->elsize;
     *flags = NPY_METH_NO_FLOATINGPOINT_ERRORS;
     *out_transferdata = NULL;
     if (descrs[0]->kind == 'c') {
@@ -3070,20 +3077,16 @@ structured_to_nonstructured_resolve_descriptors(
 
     /* Void dtypes always do the full cast. */
     if (given_descrs[1] == NULL) {
+        /*
+         * Reject string casts, historically these worked only for empty arrays.
+         * (One could go via object and hex printing, but...)
+         */
+        if (dtypes[1]->type_num == NPY_STRING || dtypes[1]->type_num == NPY_UNICODE) {
+            return -1;
+        }
         loop_descrs[1] = NPY_DT_CALL_default_descr(dtypes[1]);
         if (loop_descrs[1] == NULL) {
             return -1;
-        }
-        /*
-         * Special case strings here, it should be useless (and only actually
-         * work for empty arrays).  Possibly this should simply raise for
-         * all parametric DTypes.
-         */
-        if (dtypes[1]->type_num == NPY_STRING) {
-            loop_descrs[1]->elsize = given_descrs[0]->elsize;
-        }
-        else if (dtypes[1]->type_num == NPY_UNICODE) {
-            loop_descrs[1]->elsize = given_descrs[0]->elsize * 4;
         }
     }
     else {
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -3460,7 +3460,7 @@ array_fromfile_binary(FILE *fp, PyArray_Descr *dtype, npy_intp num, size_t *nrea
 {
     PyArrayObject *r;
     npy_off_t start, numbytes;
-    int elsize;
+    npy_intp elsize;
 
     if (num < 0) {
         int fail = 0;
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -1204,6 +1204,11 @@ _dtype_from_buffer_3118(PyObject *memoryview)
          * TODO: void would make more sense here, as it wouldn't null
          *       terminate.
          */
+        if (view->itemsize > NPY_MAX_INT) {
+            PyErr_SetString(PyExc_ValueError,
+                    "buffer itemsize is too large for a string dtype");
+            return NULL;
+        }
         descr = PyArray_DescrNewFromType(NPY_STRING);
         if (descr == NULL) {
             return NULL;
@@ -3719,7 +3724,7 @@ PyArray_FromBuffer(PyObject *buf, PyArray_Descr *type,
     Py_buffer view;
     Py_ssize_t ts;
     npy_intp s, n;
-    int itemsize;
+    npy_intp itemsize;
     int writeable = 1;
 
     if (type == NULL) {
@@ -3806,7 +3811,7 @@ PyArray_FromBuffer(PyObject *buf, PyArray_Descr *type,
         n = s/itemsize;
     }
     else {
-        if (s < n*itemsize) {
+        if (itemsize != 0 && n > s / itemsize) {
             PyErr_SetString(PyExc_ValueError,
                             "buffer is smaller than requested"\
                             " size");
@@ -3856,7 +3861,7 @@ NPY_NO_EXPORT PyObject *
 PyArray_FromString(char *data, npy_intp slen, PyArray_Descr *dtype,
                    npy_intp num, char *sep)
 {
-    int itemsize;
+    npy_intp itemsize;
     PyArrayObject *ret;
     npy_bool binary;
 
@@ -3894,7 +3899,7 @@ PyArray_FromString(char *data, npy_intp slen, PyArray_Descr *dtype,
             num = slen/itemsize;
         }
         else {
-            if (slen < num*itemsize) {
+            if (num > slen / itemsize) {
                 PyErr_SetString(PyExc_ValueError,
                                 "string is smaller than " \
                                 "requested size");
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -1207,6 +1207,11 @@ _dtype_from_buffer_3118(PyObject *memoryview)
          * TODO: void would make more sense here, as it wouldn't null
          *       terminate.
          */
+        if (view->itemsize > NPY_MAX_INT) {
+            PyErr_SetString(PyExc_ValueError,
+                    "buffer itemsize is too large for a string dtype");
+            return NULL;
+        }
         descr = PyArray_DescrNewFromType(NPY_STRING);
         if (descr == NULL) {
             return NULL;
@@ -3728,7 +3733,7 @@ PyArray_FromBuffer(PyObject *buf, PyArray_Descr *type,
     Py_buffer view;
     Py_ssize_t ts;
     npy_intp s, n;
-    int itemsize;
+    npy_intp itemsize;
     int writeable = 1;
 
     if (type == NULL) {
@@ -3815,7 +3820,7 @@ PyArray_FromBuffer(PyObject *buf, PyArray_Descr *type,
         n = s/itemsize;
     }
     else {
-        if (s < n*itemsize) {
+        if (itemsize != 0 && n > s / itemsize) {
             PyErr_SetString(PyExc_ValueError,
                             "buffer is smaller than requested"\
                             " size");
@@ -3865,7 +3870,7 @@ NPY_NO_EXPORT PyObject *
 PyArray_FromString(char *data, npy_intp slen, PyArray_Descr *dtype,
                    npy_intp num, char *sep)
 {
-    int itemsize;
+    npy_intp itemsize;
     PyArrayObject *ret;
     npy_bool binary;
 
@@ -3903,7 +3908,7 @@ PyArray_FromString(char *data, npy_intp slen, PyArray_Descr *dtype,
             num = slen/itemsize;
         }
         else {
-            if (slen < num*itemsize) {
+            if (num > slen / itemsize) {
                 PyErr_SetString(PyExc_ValueError,
                                 "string is smaller than " \
                                 "requested size");
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -3469,7 +3469,7 @@ array_fromfile_binary(FILE *fp, PyArray_Descr *dtype, npy_intp num, size_t *nrea
 {
     PyArrayObject *r;
     npy_off_t start, numbytes;
-    int elsize;
+    npy_intp elsize;
 
     if (num < 0) {
         int fail = 0;
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -324,8 +324,8 @@ _convert_from_tuple(PyObject *obj, int align)
             overflowed = 1;
         }
         else {
-            overflowed = npy_mul_with_overflow_longlong(
-                &nbytes, type->elsize, (npy_intp) items);
+            overflowed = npy_mul_sizes_with_overflow(
+                &nbytes, type->elsize, items);
         }
         if (overflowed) {
             PyErr_SetString(PyExc_ValueError,
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -2723,7 +2723,8 @@ arraydescr_reduce(PyArray_Descr *self, PyObject *NPY_UNUSED(args))
     PyObject *ret, *mod, *obj;
     PyObject *state;
     char endian;
-    int elsize, alignment;
+    npy_intp elsize;
+    int alignment;
 
     ret = PyTuple_New(3);
     if (ret == NULL) {
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -1140,7 +1140,7 @@ _convert_from_dict(PyObject *obj, int align)
                 Py_DECREF(ind);
                 goto fail;
             }
-            long offset = PyArray_PyIntAsInt(off);
+            npy_intp offset = PyArray_PyIntAsIntp(off);
             if (error_converting(offset)) {
                 Py_DECREF(off);
                 Py_DECREF(tup);
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -316,12 +316,6 @@ _convert_from_tuple(PyObject *obj, int align)
                                 "dimension smaller then zero.");
                 goto fail;
             }
-            if (shape.ptr[i] > NPY_MAX_INTP) {
-                PyErr_SetString(PyExc_ValueError,
-                                "invalid shape in fixed-type tuple: "
-                                "dimension does not fit into a C int.");
-                goto fail;
-            }
         }
         npy_intp items = PyArray_OverflowMultiplyList(shape.ptr, shape.len);
         int overflowed;
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -337,7 +337,7 @@ _convert_from_tuple(PyObject *obj, int align)
         if (newdescr == NULL) {
             goto fail;
         }
-        newdescr->elsize = (long long)nbytes;
+        newdescr->elsize = nbytes;
         newdescr->subarray = PyArray_malloc(sizeof(PyArray_ArrayDescr));
         if (newdescr->subarray == NULL) {
             Py_DECREF(newdescr);
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -316,7 +316,7 @@ _convert_from_tuple(PyObject *obj, int align)
                                 "dimension smaller then zero.");
                 goto fail;
             }
-            if (shape.ptr[i] > NPY_MAX_INT) {
+            if (shape.ptr[i] > NPY_MAX_INTP) {
                 PyErr_SetString(PyExc_ValueError,
                                 "invalid shape in fixed-type tuple: "
                                 "dimension does not fit into a C int.");
@@ -325,13 +325,13 @@ _convert_from_tuple(PyObject *obj, int align)
         }
         npy_intp items = PyArray_OverflowMultiplyList(shape.ptr, shape.len);
         int overflowed;
-        int nbytes;
-        if (items < 0 || items > NPY_MAX_INT) {
+        long long nbytes;
+        if (items < 0 || items > NPY_MAX_INTP) {
             overflowed = 1;
         }
         else {
-            overflowed = npy_mul_with_overflow_int(
-                &nbytes, type->elsize, (int) items);
+            overflowed = npy_mul_with_overflow_longlong(
+                &nbytes, type->elsize, (npy_intp) items);
         }
         if (overflowed) {
             PyErr_SetString(PyExc_ValueError,
@@ -343,7 +343,7 @@ _convert_from_tuple(PyObject *obj, int align)
         if (newdescr == NULL) {
             goto fail;
         }
-        newdescr->elsize = nbytes;
+        newdescr->elsize = (int)nbytes;
         newdescr->subarray = PyArray_malloc(sizeof(PyArray_ArrayDescr));
         if (newdescr->subarray == NULL) {
             Py_DECREF(newdescr);
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -364,7 +364,7 @@ _convert_from_tuple(PyObject *obj, int align)
         }
         for (int i=0; i < shape.len; i++) {
             PyTuple_SET_ITEM(newdescr->subarray->shape, i,
-                             PyLong_FromLong((long)shape.ptr[i]));
+                             PyLong_FromSsize_t(shape.ptr[i]));
 
             if (PyTuple_GET_ITEM(newdescr->subarray->shape, i) == NULL) {
                 Py_DECREF(newdescr);
@@ -521,7 +521,7 @@ _convert_from_array_descr(PyObject *obj, int align)
             goto fail;
         }
         PyTuple_SET_ITEM(tup, 0, (PyObject *)conv);
-        PyTuple_SET_ITEM(tup, 1, PyLong_FromLong((long) totalsize));
+        PyTuple_SET_ITEM(tup, 1, PyLong_FromSsize_t(totalsize));
 
         /*
          * Title can be "meta-data".  Only insert it
@@ -642,7 +642,7 @@ _convert_from_list(PyObject *obj, int align)
             }
             maxalign = PyArray_MAX(maxalign, _align);
         }
-        PyObject *size_obj = PyLong_FromLong((long) totalsize);
+        PyObject *size_obj = PyLong_FromSsize_t(totalsize);
         if (!size_obj) {
             Py_DECREF(conv);
             goto fail;
@@ -1156,7 +1156,7 @@ _convert_from_dict(PyObject *obj, int align)
                 goto fail;
             }
 
-            PyTuple_SET_ITEM(tup, 1, PyLong_FromLong(offset));
+            PyTuple_SET_ITEM(tup, 1, PyLong_FromSsize_t(offset));
             /* Flag whether the fields are specified out of order */
             if (offset < totalsize) {
                 has_out_of_order_fields = 1;
@@ -1180,7 +1180,7 @@ _convert_from_dict(PyObject *obj, int align)
             if (align && _align > 1) {
                 totalsize = NPY_NEXT_ALIGNED_OFFSET(totalsize, _align);
             }
-            PyTuple_SET_ITEM(tup, 1, PyLong_FromLong(totalsize));
+            PyTuple_SET_ITEM(tup, 1, PyLong_FromSsize_t(totalsize));
             totalsize += newdescr->elsize;
         }
         if (len == 3) {
@@ -1797,7 +1797,7 @@ _convert_from_str(PyObject *obj, int align)
     }
 
     int check_num = NPY_NOTYPE + 10;
-    int elsize = 0;
+    npy_intp elsize = 0;
     /* A typecode like 'd' */
     if (len == 1) {
         /* Python byte string characters are unsigned */
@@ -1810,7 +1810,7 @@ _convert_from_str(PyObject *obj, int align)
 
         /* Attempt to parse the integer, make sure it's the rest of the string */
         errno = 0;
-        long result = strtol(type + 1, &typeend, 10);
+        npy_intp result = strtol(type + 1, &typeend, 10);
         npy_bool some_parsing_happened = !(type == typeend);
         npy_bool entire_string_consumed = *typeend == '\0';
         npy_bool parsing_succeeded =
@@ -1820,7 +1820,7 @@ _convert_from_str(PyObject *obj, int align)
             goto fail;
         }
 
-        elsize = (int)result;
+        elsize = result;
 
 
         if (parsing_succeeded && typeend - type == len) {
@@ -2820,7 +2820,7 @@ arraydescr_reduce(PyArray_Descr *self, PyObject *NPY_UNUSED(args))
         elsize = -1;
         alignment = -1;
     }
-    PyTuple_SET_ITEM(state, 5, PyLong_FromLong(elsize));
+    PyTuple_SET_ITEM(state, 5, PyLong_FromSsize_t(elsize));
     PyTuple_SET_ITEM(state, 6, PyLong_FromLong(alignment));
     PyTuple_SET_ITEM(state, 7, PyLong_FromUnsignedLongLong(
             self->flags & ~NPY_NOT_TRIVIALLY_COPYABLE));
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -343,7 +343,7 @@ _convert_from_tuple(PyObject *obj, int align)
         if (newdescr == NULL) {
             goto fail;
         }
-        newdescr->elsize = (int)nbytes;
+        newdescr->elsize = (long long)nbytes;
         newdescr->subarray = PyArray_malloc(sizeof(PyArray_ArrayDescr));
         if (newdescr->subarray == NULL) {
             Py_DECREF(newdescr);
@@ -410,7 +410,7 @@ _convert_from_array_descr(PyObject *obj, int align)
     /* Types with fields need the Python C API for field access */
     npy_uint64 dtypeflags = NPY_NEEDS_PYAPI;
     int maxalign = 1;
-    int totalsize = 0;
+    npy_intp totalsize = 0;
     PyObject *fields = PyDict_New();
     if (!fields) {
         Py_DECREF(nameslist);
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -1095,7 +1095,7 @@ _convert_from_dict(PyObject *obj, int align)
 
     /* Types with fields need the Python C API for field access */
     npy_uint64 dtypeflags = NPY_NEEDS_PYAPI;
-    int totalsize = 0;
+    npy_intp totalsize = 0;
     int maxalign = 1;
     int has_out_of_order_fields = 0;
     for (int i = 0; i < n; i++) {
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -627,7 +627,7 @@ _convert_from_list(PyObject *obj, int align)
     /* Types with fields need the Python C API for field access */
     npy_uint64 dtypeflags = NPY_NEEDS_PYAPI;
     int maxalign = 1;
-    int totalsize = 0;
+    npy_intp totalsize = 0;
     for (int i = 0; i < n; i++) {
         PyArray_Descr *conv = _convert_from_any(
                 PyList_GET_ITEM(obj, i), align); // noqa: borrowed-ref OK
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -319,8 +319,8 @@ _convert_from_tuple(PyObject *obj, int align)
         }
         npy_intp items = PyArray_OverflowMultiplyList(shape.ptr, shape.len);
         int overflowed;
-        long long nbytes;
-        if (items < 0 || items > NPY_MAX_INTP) {
+        npy_intp nbytes;
+        if (items < 0) {
             overflowed = 1;
         }
         else {
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -258,7 +258,7 @@ _convert_from_tuple(PyObject *obj, int align)
      */
     if (PyDataType_ISUNSIZED(type)) {
         /* interpret next item as a typesize */
-        int itemsize = PyArray_PyIntAsInt(PyTuple_GET_ITEM(obj,1));
+        npy_intp itemsize = PyArray_PyIntAsIntp(PyTuple_GET_ITEM(obj,1));
         if (type->type_num == NPY_UNICODE) {
             if (itemsize > NPY_MAX_INT / 4) {
                 itemsize = -1;
@@ -266,6 +266,9 @@ _convert_from_tuple(PyObject *obj, int align)
             else {
                 itemsize *= 4;
             }
+        }
+        else if (type->type_num == NPY_STRING && itemsize > NPY_MAX_INT) {
+            itemsize = -1;
         }
         if (itemsize < 0) {
             /* Error may or may not be set by PyIntAsInt. */
@@ -915,7 +918,7 @@ _validate_object_field_overlap(_PyArray_LegacyDescr *dtype)
     PyObject *names, *fields, *key, *tup, *title;
     Py_ssize_t i, j, names_size;
     PyArray_Descr *fld_dtype, *fld2_dtype;
-    int fld_offset, fld2_offset;
+    npy_intp fld_offset, fld2_offset;
 
     /* Get some properties from the dtype */
     names = dtype->names;
@@ -935,7 +938,7 @@ _validate_object_field_overlap(_PyArray_LegacyDescr *dtype)
             }
             return -1;
         }
-        if (!PyArg_ParseTuple(tup, "Oi|O", &fld_dtype, &fld_offset, &title)) {
+        if (!PyArg_ParseTuple(tup, "On|O", &fld_dtype, &fld_offset, &title)) {
             return -1;
         }
 
@@ -955,7 +958,7 @@ _validate_object_field_overlap(_PyArray_LegacyDescr *dtype)
                         }
                         return -1;
                     }
-                    if (!PyArg_ParseTuple(tup, "Oi|O", &fld2_dtype,
+                    if (!PyArg_ParseTuple(tup, "On|O", &fld2_dtype,
                                                 &fld2_offset, &title)) {
                         return -1;
                     }
@@ -1149,8 +1152,9 @@ _convert_from_dict(PyObject *obj, int align)
             }
             Py_DECREF(off);
             if (offset < 0) {
-                PyErr_Format(PyExc_ValueError, "offset %ld cannot be negative",
-                             offset);
+                PyErr_Format(PyExc_ValueError,
+                             "offset %zd cannot be negative",
+                             (Py_ssize_t)offset);
                 Py_DECREF(tup);
                 Py_DECREF(ind);
                 goto fail;
@@ -1164,10 +1168,10 @@ _convert_from_dict(PyObject *obj, int align)
             /* If align=True, enforce field alignment */
             if (align && offset % newdescr->alignment != 0) {
                 PyErr_Format(PyExc_ValueError,
-                        "offset %ld for NumPy dtype with fields is "
+                        "offset %zd for NumPy dtype with fields is "
                         "not divisible by the field alignment %d "
                         "with align=True",
-                        offset, newdescr->alignment);
+                        (Py_ssize_t)offset, newdescr->alignment);
                 Py_DECREF(ind);
                 Py_DECREF(tup);
                 goto fail;
@@ -1285,7 +1289,7 @@ _convert_from_dict(PyObject *obj, int align)
     if (tmp == NULL) {
         PyErr_Clear();
     } else {
-        int itemsize = (int)PyArray_PyIntAsInt(tmp);
+        npy_intp itemsize = PyArray_PyIntAsIntp(tmp);
         Py_DECREF(tmp);
         if (error_converting(itemsize)) {
             Py_DECREF(new);
@@ -1294,9 +1298,9 @@ _convert_from_dict(PyObject *obj, int align)
         /* Make sure the itemsize isn't made too small */
         if (itemsize < new->elsize) {
             PyErr_Format(PyExc_ValueError,
-                    "NumPy dtype descriptor requires %d bytes, "
-                    "cannot override to smaller itemsize of %d",
-                    new->elsize, itemsize);
+                    "NumPy dtype descriptor requires %zd bytes, "
+                    "cannot override to smaller itemsize of %zd",
+                    (Py_ssize_t)new->elsize, (Py_ssize_t)itemsize);
             Py_DECREF(new);
             goto fail;
         }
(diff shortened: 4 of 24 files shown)


numpy/_core/src/multiarray/descriptor.c
```diff
@@ -343,7 +343,7 @@ _convert_from_tuple(PyObject *obj, int align)
         if (newdescr == NULL) {
             goto fail;
         }
-        newdescr->elsize = nbytes;
+        newdescr->elsize = (int)nbytes;
```
tylerjereddy: This is obviously categorically wrong for what is trying to be accomplished here, and yet using `npy_intp` or `long long` currently causes other test failures.

`git grep` analysis does show a lot of places with i.e., `int itemsize` for example, so maybe that's not too surprising.

numpy/_core/tests/test_dtype.py
```diff
@@ -2049,3 +2043,15 @@ def test_signature_dtypes_classes(self, typename: str):
...
+def test_gh_31308():
+    kind = [("x", np.float64, 2 ** 28)]
+    kind_dtype = np.dtype(kind)
+
+
+@pytest.mark.xfail(run=False)
+def test_gh_31308_materialized():
+    kind = [("x", np.float64, 2 ** 28)]
+    kind_dtype = np.dtype(kind)
+    rec_arr = np.array((1,), dtype=kind_dtype)
```
tylerjereddy, line 2235: I spent several hours `printf` debugging why this materialization test fails but the `dtype` construction above doesn't error out. But the problem is that the regression test above is too crude, so I just wasted hours chasing the materialization of a broken descr...

numpy/_core/tests/test_dtype.py
```diff
@@ -2049,3 +2043,15 @@ def test_signature_dtypes_classes(self, typename: str):
 
         params_actual = set(sig.parameters)
         assert params_actual == params_expect
+
+
+def test_gh_31308():
+    kind = [("x", np.float64, 2 ** 28)]
+    kind_dtype = np.dtype(kind)
```
tylerjereddy, line 2074: Simply making this original regression test pass, and turning the one from Sebastian (`test_pickling_large`) back on and having that pass, isn't even close to safe.

For example, here `print(kind_dtype.itemsize)` is `-2147483648`. I should fix that before chasing down the array creation side of the story (which may resolve itself if this test is sufficiently thorough since `npy_intp` is already on the descr struct...).

Mostly a caution for myself when I come back to this to avoid wasting time again. I should really be quite thorough and making sure as many aspects of this dtype are correct as possible before venturing into the array creation proper.

numpy/_core/src/multiarray/descriptor.c
tylerjereddy: Ok, this is slightly better now--I switched it to `long long` and both of my new regression tests are now passing locally on ARM Mac. I believe it should be `npy_intp` and I suspect that will work "ok," but I'll need to check another time since I spent another several hours debugging.

The current state of the PR is that `test_pickling_large` is the only test failure locally on ARM Mac--I'll circle back to that one, probably with another round of debug printing and looking for the wrong use of `int` in some places.

numpy/_core/tests/test_dtype.py
tylerjereddy: The materialization test is now passing just fine locally for a larger than C int sized recarray dtype. I did guard it with a memory requirement decorator.

numpy/_core/tests/test_dtype.py
tylerjereddy: The regression test has been improved somewhat and now passes--it verifies that `itemsize` is set appropriately for this particular case.

I'll probably try to expand this regression test to some parametrized cases with different expected `itemsize` values I think, otherwise it is too risky to turn this on. Some things to test probably include multiple fields that exceed the sum of C int size when combined, different types, etc.

That said, when I do circle back here, I think I should prioritize getting `test_pickling_large` passing first, then expand the current tests.

numpy/_core/tests/test_dtype.py
tylerjereddy: (this test may also need a 64-bit machine guard?)

numpy/_core/src/multiarray/descriptor.c
```diff
@@ -316,7 +316,7 @@ _convert_from_tuple(PyObject *obj, int align)
                                 "dimension smaller then zero.");
                 goto fail;
             }
-            if (shape.ptr[i] > NPY_MAX_INT) {
+            if (shape.ptr[i] > NPY_MAX_INTP) {
```
seberg: I think you can just delete this now.  The conversion guarantees `intp` compatibility already.

numpy/_core/src/multiarray/descriptor.c
```diff
@@ -325,13 +325,13 @@ _convert_from_tuple(PyObject *obj, int align)
         }
         npy_intp items = PyArray_OverflowMultiplyList(shape.ptr, shape.len);
         int overflowed;
-        int nbytes;
-        if (items < 0 || items > NPY_MAX_INT) {
+        long long nbytes;
+        if (items < 0 || items > NPY_MAX_INTP) {
```
seberg: Agree that this should be `intp`, and this check is only needed for `< 0` to catch errors now.

numpy/_core/tests/test_dtype.py
```diff
@@ -737,13 +737,8 @@ def test_shape_matches_ndim(self):
...
     def test_shape_invalid(self):
         # Check that the shape is valid.
-        max_int = np.iinfo(np.intc).max
         max_intp = np.iinfo(np.intp).max
         # Too large values (the datatype is part of this)
-        assert_raises(ValueError, np.dtype, [('a', 'f4', max_int // 4 + 1)])
-        assert_raises(ValueError, np.dtype, [('a', 'f4', max_int + 1)])
-        assert_raises(ValueError, np.dtype, [('a', 'f4', (max_int, 2))])
-        # Takes a different code path (fails earlier:
         assert_raises(ValueError, np.dtype, [('a', 'f4', max_intp + 1)])
```
seberg, line 746: Would be nice to keep one that overflows later, e.g. `max_intp // 4 + 1`.

seberg (commented):
Just some quick comments.  Would be good to check if there might be other `(int)` casts are definitions (but I can see even asking some LLM to search the code-base).
I guess the pickling path is already fixed, so there likely isn't much, but I am not sure (some string paths or maybe the hashing).

numpy/_core/tests/test_dtype.py
```diff
@@ -1270,8 +1265,8 @@ def check_pickling(self, dtype):
             assert_equal(x, y)
             assert_equal(x[0], y[0])
 
+    @pytest.mark.slow()
```
tylerjereddy: `test_pickling_large` (and the rest of the full testsuite) now pass locally on this branch, thanks to a small type fix in `arraydescr_reduce`.

A few notes here:
- on an ARM M3 Max mac laptop, that test takes a long time to pass, more than 6 minutes: `1 passed in 402.71s (0:06:42)`; I think that means that the `slow()` marker doesn't cut it, and I guess NumPy doesn't have `xslow`? What do you want me to do here then? Do we really need to iterate over 2 x > C int cases x 5 pickle protocol versions? If so, how should we selectively run this only once in a while? Even a single protocol version is over a minute to run.
- should we also stack in a `requires_memory` decoration here? I believe these arrays are pretty massive like in the other test I added that materializes such arrays.

numpy/_core/src/multiarray/descriptor.c
tylerjereddy: Yep, purged in the most recent commit after confirming full testsuite passes locally.

numpy/_core/src/multiarray/descriptor.c
tylerjereddy: I'll push this up shortly, note this new compilation warning with that change (tests still pass)--can we live with that or need to add a function to handle this? (I believe the `npy_mul_with_overflow_longlong` func I switched to using here is `c.src` templated func, so may be annoying...)

```
../numpy/_core/src/multiarray/descriptor.c:328:17: warning: incompatible pointer types passing 'npy_intp *' (aka 'long *') to parameter of type 'npy_longlong *' (aka 'long long *') [-Wincompatible-pointer-types]
  328 |                 &nbytes, type->elsize, (npy_intp) items);
      |                 ^~~~~~~
../numpy/_core/src/common/templ_common.h.src:29:47: note: passing argument to parameter 'r' here
   29 | npy_mul_with_overflow_longlong(npy_longlong * r, npy_longlong a, npy_longlong b)
```

seberg: We have a `npy_mul_sizes_with_overflow` for non-negative `npy_intp` inputs.

numpy/_core/tests/test_dtype.py
tylerjereddy: Good point, I did `f8` with `//8 + 1` and `f4` with `//4 + 1`. Both pass (raise as expected).

numpy/_core/src/multiarray/descriptor.c
tylerjereddy: Ah thanks, pushed that change up after checking tests still pass, warning is gone now.

tylerjereddy:
Alright, most of the comments have been addressed now I think. Beyond some of the still-open discussions above, I wonder if I'm missing some important routes to structured dtype construction--my testing here mostly focused on the "tuple" specification route, but perhaps there are others that need testing/shims.

I guess that's related to my self-review rquirement for more thorough tests.

I'll have to circle back another day/time for that though.

numpy/_core/tests/test_dtype.py
seberg: Maybe just skip the "dtype is functional" test, or make the array empty `np.zeros(0, dtype=dtype)`.
I am not 100% sure if that covers everything, but it seems fair to me to do that.

There isn't really much to check anyway, the only thing I can think of is whether `dtype.flags == pickled.flags`, because the `dtype == pickled` should check almost everything else (metadata is already checked).

seberg:
> for more thorough tests.

I guess creation wise, it is probably if `1000000i,100000i` style also has overflow checks (I am not even sure that takes a different path, though).
Otherwise, I guess the biggest thing would be some actual usage tests, such as a cast/promotion.  Maybe one that explodes the size `result_type("<large_num>?,<large_num>", "<large_num>c,<large_num>?")` which goes to `<large_num>c,<large_num>c` (larger than either inputs).
It may be interesting to actually try one of those for a real cast, but of course that needs a lot of memory and is very slow.
(Just throwing it out there, not sure we need quite all of these.)

tylerjereddy:
>  it is probably if 1000000i,100000i style also has overflow checks

I parametrized `test_gh_31308()` to include `"2147483648i,2147483648i"` construction, which produces an `itemsize` of `0`, so that test case does indeed fail for now. I'll need to investigate.

I thought I saw a dictionary code path too.. but anyway one thing at a time.

seberg:
> I thought I saw a dictionary code path too.. but anyway one thing at a time.

Ah, yeah... `dict(names=["a", ...], formats=["20000f"])` and that can also include `offsets=[2**31-100]`, another path that could overflow, I guess.
(On the up-side, you are seriously cleaning up these code paths...)

tylerjereddy:
> "2147483648i,2147483648i"

That test case is now passing as well, after latest commit. Let me check the dictionary case next.

tylerjereddy:
> dict(names=["a", ...], formats=["20000f"])

Yes, that approach also overflows on `main` with large enough input--I've pushed up a new test case and small patch that allows it to pass.

tylerjereddy:
> offsets=[2**31-100]

An offset smaller than a C int was "ok," even if the field element size was already larger than a C int, but an offset larger than a C int would error out. Test cases and a small patch to fix/allow that have been pushed up.

I also had to resolve some merge conflicts, guess there is some activity in the descriptor source.

I'll continue to skip the CI for now.

numpy/_core/tests/test_dtype.py
tylerjereddy: The latest commit adds several more/complex structured dtype test cases to `test_gh_31308()`, and they all pass.

Since this is a self-review comment thread, I'll mark it resolved.

numpy/_core/tests/test_dtype.py
tylerjereddy: I left the initial materialization in, `y = np.zeros(3, dtype=pickled)`, but gave `check_pickling` a default-on argument that optionally runs the actually assertion on the materialized arrays, which I've turned off for the `test_pickling_large` calls.

I believe this is probably robust, but may have some risk, if some operating system did actually fully allocate the RAM on a `zeros` call, but I believe that's rare? CI may tells us--I just flushed CI for this time.

With that change, the test would no longer fail without the source patches here though, so I added an extra check that  `assert_equal(pickled.itemsize, dtype.itemsize)`, which does fail without the work in this PR. I was slightly surprised that `assert_equal(pickled.descr, dtype.descr)` wasn't sufficient to enforce proper reconstitution of itemsize, since that reads as "they have the same descriptors," but apparently it doesn't...

This also allowed me to remove the `slow` marker.

tylerjereddy: Alright, CI is failing on the first try, `IS_64BIT` may not be a sufficient guard for some Windows cases?

`test_gh_31308` needs a 64-bit guard. Some of the CI failures are not related--bit messy.

seberg: There is one MKL job generally failing right now, just ignore that one.

tylerjereddy: The latest commit suppresses CI again for now. There is one remaining failure that I'm going to need to debug directly on a Windows box, below the fold. The new `itemsize` check that I added to `check_pickling` seems to show `dtype = np.dtype(f"({2**31},)i")` getting set to `dtype(('<i4', (-2147483648,)))`, independent of serialization. The `intp` size on that Windows CI runner should be just fine since the test is guarded by `IS_64BIT`, which comes from `np.dtype(np.intp).itemsize`. `git grep -E -i "int elsize"` still has several hits on this branch, but way easier to debug directly on Win machine, so I'll have to circle back.


<details>

```
>           assert_equal(pickled.itemsize, dtype.itemsize)
E           AssertionError: 
E           Items are not equal:
E            ACTUAL: 0
E            DESIRED: 8589934592

arr_assert = False
buf        = b'cnumpy\ndtype\np0\n(VV0\np1\nI00\nI01\ntp2\nRp3\n(I3\nV|\np4\n(g0\n(Vi4\np5\nI00\nI01\ntp6\nRp7\n(I3\nV<\np8\nNNNI-1\nI-1\nI0\ntp9\nb(I-2147483648\ntp10\ntp11\nNNI0\nI4\nI0\ntp12\nb.'
dtype      = dtype(('<i4', (-2147483648,)))
pickled    = dtype(('<i4', (-2147483648,)))
```

</details>

seberg: You can probably get away without trying a windows box.  If this diverges on windows, you can be sure there is some use of `long` along the way.  Either a C `long` or `PyLong_FromLong` or so call somewhere along the branch.
And just grepping sees things like:
```
PyObject *size_obj = PyLong_FromLong((long) totalsize);
```
There are a lot more in that file... Got to just replace them all with FromSsizet (or whatever it is).  And most likely that will be enough.

(There might be corresponding `PyLong_AsLong` somewhere, but from a grep those seem more likely to be fixed than the reverse.)

tylerjereddy: Ok, the CI is passing apart from two clearly unrelated failures.

I did end up having to use a Windows box because fork CI debugging was too painful. I also added an additional assertion to one of my new tests (`test_gh_31308()`), since it should have picked up the descr problem that was causing the pickle test to fail (it now does).

seberg:
Thanks, this looks all good now.  I dunno, TBH, an agent might be simplest to find and maybe also just fix these... Although a grep for `elsize` and `ITEMSIZE` may be sufficient (also field iteration/names).
While for some paths it's totally fine (because we know we have simple numeric types), there are still a bunch of paths that are wrong, e.g. (but there are more):
* Use of `Oi` tuple unpacking in `descriptor.c` for the `offset`.
* Use of integers for field iterations `names` in `VOID_setitem/getitem/copyswap`...
* A few more `int.*elsize` definitions, many of which reachable by void dtypes.
  (Yes, a few of these are already bugs, since user DTypes were already allowed to be larger in practice.)

tylerjereddy:
I don't mind working on those. I wish we had an option for i.e., the preferred LLM to just complain about it directly in code review and for i.e., you to curate those complaints to the actual items I need to address, if we're going to lean on agent checks. Obviously I noted a few times above that I could `git grep` and find problems that still existed, although it is far more useful for me to identify actual reproducers/test cases for problems, rather than assuming that every `git grep` blemish needs to be addressed.

tylerjereddy:
Some downstream exploration of simplifications enabled by this branch is being explored by the scientists at https://github.com/mcdc-project/mcdc/pull/419.

tylerjereddy:
I pushed in a commit to test/fix the `tofile`/`fromfile` control flow, which had one mistyped `elsize`. I don't really like this though--the regression test didn't fail before the source patch--it looks like my other fixes allow the typecasting to work out in the end so that one feels more like a "should fix," but wasn't actually needed really.

The regression test is also extremely slow, though it will basically never run with `@requires_memory(free_bytes=2e9)`.

I'll keep chipping away at these, though I don't like the ones that don't have an explicit reproducer that fails even if we "know" we need to change a type.

numpy/_core/tests/test_records.py
```diff
@@ -108,6 +110,17 @@ def test_recarray_fromfile(self):
...
+    @pytest.mark.skipif(not IS_64BIT, reason="test requires 64-bit system")
+    @requires_memory(free_bytes=2e9)
+    def test_recarray_fromfile_massive(self, tmpdir):
+        kind = [("x", np.float64, 2 ** 28)]
+        kind_dtype = np.dtype(kind)
+        rec_arr = np.array((1,), dtype=kind_dtype)
+        with tmpdir.as_cwd():
+            rec_arr.tofile("f.data")
+            actual = np.fromfile("f.data", dtype=kind_dtype)
+            assert actual.itemsize == 2 ** 28 * 8
```
seberg: Ohh, fun.  OTOH, the error here would be that we are not reading everything (i.e. the last bit of the result not being -1).  But I think there is a fun little thing happening here:
* This is a signed int overflow `2**31 -> -2**31`
* Cast to `size_t` for reading, it'll actually just go back to `2**31` as that is unsigned.

So, no problem until we reach 2**32 at which point the error would be reading nothing at all.

Possible we could employ a funny trick here: Just try to read 1 element of a 2**32+1 sized dtype from a short file (not empty with the +1).
With `fromfile(...., count=1)` that _should_ try to read one element, fail and return an empty array.  But with the bug I think it'll actually return an array of length 1.
(That said, clear that is more of a regression test designed for the specific error in this code...)

tylerjereddy: Playing with this a bit, something really strange/bad happens somewhere--I can retrieve items from the reconstituted array quickly, but attempting to perform assertions on even those individual elements is incredibly slow.

For example,

```python
item = actual["x"][0][1].item() # fast
assert item > 0 # prohibitively slow
```

It is only slow if it is destined to fail (!), with or without the `npy_intp` patch applied. How strange, I wonder if something is getting corrupted, or if I'm missing something obvious getting expanded too large (`item` prints out innocently enough as either `0.0` or `1.0` depending on application of the patch). I spent a few hours poking at that because it was so strange looking. It will hang on the mismatch of the `tobytes` patterns as well, which might suggest a corruption somewhere I suppose (since the same bytes pattern comparison is fine in raw interpreter).

I didn't find any advantage to using `count=1`. Maybe I misunderstood, but our problem is that we have a single massive "item" so it doesn't cut that down/discretize? The test works just fine with it, but runtime/result is the same. Perhaps you meant reading back in with a separate smaller dtype itemsize and looking at a slice of the data that way, not sure.

For now, I've pushed in a revision to `test_recarray_fromfile_massive()` such that it uses a larger itemsize (that part of your suggestion was indeed needed), and passes within 1-10 seconds locally with the previous patch applied to `array_fromfile_binary`. Without that small patch, the test will basically hang for reasons I don't fully understand. I suppose that's at least a bit closer to what I want re: a test that doesn't really reasonably pass without the patch, though still a bit annoying.

seberg: > Maybe I misunderstood, but our problem is that we have a single massive "item" so it doesn't cut that down/discretize?

I think I just skipped over that it seemed to me that we pass `count=1` it'll assume it can read 1 item, while otherwise it checks how big the file actually is.
I.e. my thought was we can just put a few bytes into the file, but NumPy will incorrectly attempt to read an actual element.

---

> assert item > 0 # prohibitively slow

In general, or in `pytest`?  Because if it is in pytest, I could imagine it wants to do something like print `actual` or so?

tylerjereddy: Ah yeah, thanks, looks like suppressing the traceback allows the failure scenario to happen in a second or two (when type change is reverted): `spin test -t numpy/_core/tests/test_records.py::TestFromrecords::test_recarray_fromfile_massive -- --tb=no`

That's annoying that it doesn't auto truncate/summarize.

This also "works" though is not very nice:

```diff
--- a/numpy/_core/tests/test_records.py
+++ b/numpy/_core/tests/test_records.py
@@ -122,7 +122,9 @@ def test_recarray_fromfile_massive(self, tmpdir):
             actual = np.fromfile("f.data", dtype=kind_dtype)
             assert actual.itemsize == 2 ** 29 * 8
             item = actual["x"][0][1]
-            assert_allclose(item, 1)
+            if item != 1:
+                # avoid hang from pytest traceback dumping massive array
+                pytest.fail("fromfile elsize error", pytrace=False)
```

I'll leave this one alone for now--seems to be getting closer to something sensible perhaps...

tylerjereddy:
This case fails to materialize:

`kind = [('x', <class 'numpy.float64'>, 268435456), ('y', <class 'numpy.float64'>, 1)]`

with `OverflowError: signed integer is greater than maximum`. I've adjusted `test_gh_31308_materialized` to include that failing case, but haven't succeeding in patching it yet.

tylerjereddy:
To resolve https://github.com/numpy/numpy/pull/31332#issuecomment-4434946154, I noticed that while this construction failed to materialize: `[("x", np.float64, 2 ** 28), ("y", np.float64, 1)]`

this one, its "mirror image," succeeded: ` [("x", np.float64, 1), ("y", np.float64, 2 ** 28)]`

A tuple parsing overflow in CPython with `"Oi|O"` (via `dtype_traversal.c`) gets triggered for large offsets in the failing case, but in the mirrored case you can slip by since the final offset is still small before parsing the larger field in.

I pushed in a small patch for that, allowing that new test case to pass. That took many hours to solve though.

tylerjereddy:
The `isnative` check could overflow on a large structured dtype--regression testing was adjusted to catch that along with a small source patch allowing it to pass.

tylerjereddy:
To test the case of overflow in `_descr_find_object`, we'll need a way to force `pickle` protocol `2` on read-in. Using Python `3.12` locally, protocol `3` is the lowest selected for all options in `check_pickling` via `test_pickling_large`, because writing to a protocol below `3` does not guarantee reading back in at the same protocol, and there's no knob for that on read-in. I'll just note that here for now.

In the meanwhile I adjusted regression testing to fail for calling `np.argmax` on large structured dtypes when there is an `_ArrayMemoryError` (with busted `elsize`) instead of `TypeError`, then pushed in a small patch to restore the usual behavior (the expected `TypeError`). Note that this will basically cause `pytest` to hang on the print when it fails. That's going to be a common issue here, but I'll write the tests as "normal" for now.

tylerjereddy:
`ndarray.getfield` could overflow when working with these large structured dtypes--a regression test and fix for it have been pushed in. Note that this required a function signature change in `cython-30.pxd`. (I also noticed that incremental rebuilds don't seem to detect changes in the `pxd`, need a clean rebuild).

seberg:
> ndarray.getfield could overflow when working with these large structured dtypes

Good find, but we can't change the public C-API :(.  I may be tempted to just drop this one and live with it failing (because it's silly API anyway, IMO).

tylerjereddy:
Ok, I pushed in the mirror change for `ndarray.setfield` already--that has the same drawback. We could use `xfail` but that's kind of pointless. Perhaps more useful would be erroring out with a helpful error message for `getfield/setfield`. That's probably annoying with INT/INTP max checks or whatever, maybe I'll circle back to that based on what you decide.

(if interception is even possible)

seberg:
> That's probably annoying with INT/INTP max checks or whatever, maybe I'll circle back to that based on what you decide.

What we could do is change everything except the public API... I just find getfield/setfield such weird API that I wouldn't mind just deprecating it... and because of that I am not sure I care any way (since currently it'll just fail cryptically but not crash hard).

Btw. I started to do some fixes with a bit of help from an agent, but I can hold off a bit more.  (I have to review it, it might have gone a bit further than needed on the string side.)

tylerjereddy:
> I started to do some fixes with a bit of help from an agent, but I can hold off a bit more

No need to hold off. I'm probably not competitive vs. you + an agent if it is finding legit stuff.

I did push in one more commit with a regression test + fix for `np.loadtxt` with a `dtype` with a larger than C `int` sized `offset`, that fails before/passes after the source patch. That touched `field_types.h`, but perhaps that is allowed.

If you want to just push stuff in that's fine by me, or if you want to list off tests that you think it is important to make pass, etc., either is fine with me. I'm just spending a few hours here and there, and SciPy release duties will pull me away too.

seberg:
OK, I pushed some with somewhat shady tests[^1], but they work (e.g. the pickle one works I think).  This also now allows `V1234892843934` style to be large, but for now rejects it for string/unicode.
I removed the ability to cast unstructured void to string/unicode (this never worked except for *empty* arrays).  Large string sizes need to be audited, but it felt too much here...

*Right now, we still need to revert the public API `PyArray_(Get|Set)Field` changes.*  Could modify it internally (or just test for the error and don't care because it's silly API anyway -- which is my preference).

I need to do another pass at some point, but we should probably just go ahead with this soonish.

[^1]: E.g. the refcounting tests rely on `arr.flat =` setting which seems mostly a bug, but the reality is that those functions are likely unused otherwise. The pickle test does some weird setup (the agent wrote it), but I guess it should catch the thing, so...  Otherwise not everything is covered (mostly the "too large strings" added).

tylerjereddy:
> Right now, we still need to revert the public API PyArray_(Get|Set)Field changes

I pushed in a commit that does that, but I didn't actually have to back out my regression tests for `getfield`/`setfield` with large structured types, so I guess the public API change wasn't needed, at least not for the tested Python codepaths.

I've never even used `getfield`/`setfield` myself, so I'm not really bothered if the preference is to reverse the integer size shims associated with them and test for the original errors.

I fixed up a few small linter issues in that commit as well. I haven't done a pass over the LLM shims myself.

tylerjereddy:
I rebased on latest `main` and force pushed, which was a little messy after the past few months. I needed one fix to a usage of `npy_runtime_imports`  locally and then tests/linting passed locally on ARM mac at least.

I suspect I should go through the AI additions from Sebastian and make sure things look "ok" in the diff, maybe check if the AI-drafted tests fail on `main` and stuff like that.

tylerjereddy:
When I checkout ("transplant") the testing changes here on latest NumPy `main` branch I see the 30 failures summarized below the fold (with `NPY_AVAILABLE_MEM=32GB` set).

<details>

```
FAILED numpy/_core/tests/test_dtype.py::TestPickling::test_pickling_large - ValueError: invalid shape in fixed-type tuple: dimension does not fit into a C int.
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308[kind0-2147483648] - ValueError: invalid shape in fixed-type tuple: dtype size in bytes must fit into a C int.
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308[kind1-2147483648] - AssertionError: assert -2147483648 == 2147483648
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308[kind2-2147483648] - AssertionError: assert -2147483648 == 2147483648
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308[kind3-2147483648] - AssertionError: assert -2147483648 == 2147483648
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308[2147483648i,2147483648i-17179869184] - ValueError: invalid shape in fixed-type tuple: dimension does not fit into a C int.
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308[2147483648f,2147483648f-17179869184] - ValueError: invalid shape in fixed-type tuple: dimension does not fit into a C int.
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308[2147483648d,2147483648d-34359738368] - ValueError: invalid shape in fixed-type tuple: dimension does not fit into a C int.
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308[2b,2147483648b,2f,4i-2147483674] - ValueError: invalid shape in fixed-type tuple: dimension does not fit into a C int.
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308[kind8-8589934592] - ValueError: invalid shape in fixed-type tuple: dimension does not fit into a C int.
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308[kind9-8589934593] - ValueError: invalid shape in fixed-type tuple: dimension does not fit into a C int.
FAILED numpy/_core/tests/test_casting_unittests.py::TestCasting::test_void_to_string_special_case - Failed: DID NOT RAISE TypeError
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308[kind10-10737418140] - ValueError: invalid shape in fixed-type tuple: dimension does not fit into a C int.
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308[kind11-10737418240] - ValueError: invalid shape in fixed-type tuple: dimension does not fit into a C int.
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308[kind12-4294967413] - ValueError: invalid shape in fixed-type tuple: dimension does not fit into a C int.
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308_result_type_keeps_large_offsets - ValueError: invalid shape in fixed-type tuple: dtype size in bytes must fit into a C int.
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308_dict_itemsize_override_large - ValueError: integer won't fit into a C int
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308_array_itemsize_getter_large_dtype - ValueError: invalid shape in fixed-type tuple: dtype size in bytes must fit into a C int.
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308_binary_imports_reject_too_small_large_itemsize - ValueError: integer won't fit into a C int
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308_out_of_order_object_fields_large_offsets - ValueError: integer won't fit into a C int
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308_copyto_large_offsets - ValueError: integer won't fit into a C int
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308_negative_offset_error_message - AssertionError: Regex pattern did not match.
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308_legacy_setstate_keeps_object_flag - ValueError: integer won't fit into a C int
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308_materialized[val0-kind0-268435456] - ValueError: invalid shape in fixed-type tuple: dtype size in bytes must fit into a C int.
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308_materialized[val1-kind1-268435456] - ValueError: invalid shape in fixed-type tuple: dtype size in bytes must fit into a C int.
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308_deepcopy_and_scalar_object_large_offsets - ValueError: invalid shape in fixed-type tuple: dimension does not fit into a C int.
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308_getfield - ValueError: invalid shape in fixed-type tuple: dtype size in bytes must fit into a C int.
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308_setfield - ValueError: invalid shape in fixed-type tuple: dtype size in bytes must fit into a C int.
FAILED numpy/_core/tests/test_dtype.py::test_gh_31308_loadtxt - ValueError: integer won't fit into a C int
FAILED numpy/_core/tests/test_records.py::TestFromrecords::test_recarray_fromfile_massive - ValueError: invalid shape in fixed-type tuple: dtype size in bytes must fit into a C int.
```

</details>

From some manual analysis of tests that were added or modified in this PR, only the following two don't fail when transplanted to current `main`:
- `test_shape_invalid`
- `test_gh_31308_huge_void_scalars`

The `test_shape_invalid` changes are for ensuring a `ValueError` is raised, so that's probably "ok," and there's a discussion with Sebastian about the changes there above. `test_gh_31308_huge_void_scalars` has a `slow` marker, so I just needed to run that separately with `full` set and it also fails on `main`.

So, that's fairly sensible I think. Probably next things I should do are read the full diff carefully again and maybe a bit more `git grep` analysis on this branch for other improperly typed ints related to structured dtype sizing.

seberg:
Thanks, yeah, that sounds good.  I should probably have a brief look over as well, but the danger here is mostly that we are missing a place rather than any of the fixes being wrong. And of course it's annoying that we might not cover quite all fixes with tests.

Either way, we need to be a bit pragmatic here I guess, so please ping me if you are content so we push it forward!

numpy/_core/src/multiarray/convert_datatype.c
```diff
@@ -2331,7 +2331,14 @@ get_byteswap_loop(
...
     assert(descrs[0]->kind == descrs[1]->kind);
     assert(descrs[0]->elsize == descrs[1]->elsize);
-    int itemsize = descrs[0]->elsize;
+    if (descrs[0]->elsize > NPY_MAX_INT) {
+        // TODO: add a regression test for >INT_MAX byte-swap itemsize support.
+        PyErr_SetString(PyExc_TypeError,
+                "byte-swapping with itemsize larger than INT_MAX "
+                "is not supported");
+        return -1;
+    }
```
tylerjereddy, line 2340: As the `TODO` implies I suppose, I can delete this block and full tests pass on the branch. I'll look into regression testing this one if it isn't too crazy to get right.

tylerjereddy: I looked at this for a while, maybe this one was added by AI, because it doesn't look like it is possible for the `get_byteswap_loop()` function to ever pick up an `elsize` that is large from a structured dtype with a huge field. Perhaps the byteswapping is happening in fixed sized chunks or something like that?

Either way, I was able to write a regression test that captures a `TypeError` in these scenarios, but it required me to hoist the `elsize` checking farther up in the control flow to `PyArray_Byteswap()`. I won't push that in until confirmation we want this, but the full testsuite also passed with that diff below the fold.

Should probably also delete the error handling out of `get_byteswap_loop()`, perhaps add a brief comment on why we don't allow this in the source? Is it because it would take forever (if I remove the `TypeError` in `PyArray_Byteswap()` I quickly run out of patience on the large structured dtype used below the fold.. seems to take a long time)?

<details>

```diff
diff --git a/numpy/_core/src/multiarray/methods.c b/numpy/_core/src/multiarray/methods.c
index c24f26237c..aea99ab62e 100644
--- a/numpy/_core/src/multiarray/methods.c
+++ b/numpy/_core/src/multiarray/methods.c
@@ -534,6 +534,12 @@ PyArray_Byteswap(PyArrayObject *self, npy_bool inplace)
     PyArray_CopySwapNFunc *copyswapn;
     PyArrayIterObject *it;
 
+    if (PyArray_DESCR(self)->elsize > NPY_MAX_INT) {
+        PyErr_SetString(PyExc_TypeError,
+                "byte-swapping with itemsize larger than INT_MAX "
+                "is not supported");
+        return NULL;
+    }
     if (inplace && PyArray_FailUnlessWriteable(self, "array to be byte-swapped") < 0) {
         return NULL;
     }
diff --git a/numpy/_core/tests/test_records.py b/numpy/_core/tests/test_records.py
index a3e05463a8..6976c5cfa5 100644
--- a/numpy/_core/tests/test_records.py
+++ b/numpy/_core/tests/test_records.py
@@ -138,6 +138,15 @@ def test_recarray_fromfile_massive(self, tmpdir):
             item = actual["x"][0][1]
             assert_allclose(item, 1)
 
+    @pytest.mark.skipif(not IS_64BIT, reason="test requires 64-bit system")
+    @requires_memory(free_bytes=4.3e9)
+    def test_recarray_massive_byteswap_err(self):
+        kind = [("x", np.float64, 2 ** 29)]
+        kind_dtype = np.dtype(kind)
+        rec_arr = np.array((1,), dtype=kind_dtype)
+        with pytest.raises(TypeError, match="itemsize larger than"):
+            rec_arr.byteswap(inplace=True)
+
     def test_recarray_from_obj(self):
         count = 10
         a = np.zeros(count, dtype='O')
```

</details>

seberg: You are right, this is un-testable, because this loop is used only for numeric casts and numeric casts never have an arbitrarily large itemsize.
(Could even just make it an `assert()  // numeric casts have small itemsize` or so.)

tylerjereddy: Ok, I've pushed up the change set I described then. Skipping CI for now, as self-review continues.

numpy/_core/src/multiarray/array_assign_array.c
```diff
@@ -42,7 +42,10 @@ copycast_isaligned(int ndim, npy_intp const *shape,
     int aligned;
     int big_aln, small_aln;
 
-    int uint_aln = npy_uint_alignment(dtype->elsize);
+    int uint_aln = 0;
+    if (dtype->elsize <= NPY_MAX_INT) {
+        uint_aln = npy_uint_alignment((int)dtype->elsize);
+    }
```
tylerjereddy, line 48: The full testsuite passed locally on ARM Mac if I revert this change.

numpy/_core/src/multiarray/convert_datatype.c
```diff
@@ -2331,7 +2331,7 @@ get_byteswap_loop(
     PyArray_Descr *const *descrs = context->descriptors;
     assert(descrs[0]->kind == descrs[1]->kind);
     assert(descrs[0]->elsize == descrs[1]->elsize);
-    int itemsize = descrs[0]->elsize;
+    int itemsize = (int)descrs[0]->elsize;
```
tylerjereddy: The full testsuite passes locally on ARM Mac when I revert this typecast.

As discussed at https://github.com/numpy/numpy/pull/31332#discussion_r3992519644, I believe we don't really need it.

numpy/_core/src/multiarray/ctors.c
```diff
@@ -1207,6 +1207,11 @@ _dtype_from_buffer_3118(PyObject *memoryview)
          * TODO: void would make more sense here, as it wouldn't null
          *       terminate.
          */
+        if (view->itemsize > NPY_MAX_INT) {
+            PyErr_SetString(PyExc_ValueError,
+                    "buffer itemsize is too large for a string dtype");
+            return NULL;
+        }
```
tylerjereddy, line 1214: Full testsuite passes locally if I purge this handling.

numpy/_core/src/multiarray/ctors.c
```diff
@@ -3903,7 +3908,7 @@ PyArray_FromString(char *data, npy_intp slen, PyArray_Descr *dtype,
             num = slen/itemsize;
         }
         else {
-            if (slen < num*itemsize) {
+            if (num > slen / itemsize) {
```
tylerjereddy, line 3911: Full testsuite passes locally on ARM Mac after localized reversion:

<details>

```diff
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -3908,7 +3908,7 @@ PyArray_FromString(char *data, npy_intp slen, PyArray_Descr *dtype,
             num = slen/itemsize;
         }
         else {
-            if (num > slen / itemsize) {
+            if (slen < num * itemsize) {
```

</details>

numpy/_core/src/multiarray/descriptor.c
```diff
@@ -2903,7 +2906,7 @@ arraydescr_reduce(PyArray_Descr *self, PyObject *NPY_UNUSED(args))
         elsize = -1;
         alignment = -1;
     }
-    PyTuple_SET_ITEM(state, 5, PyLong_FromLong(elsize));
+    PyTuple_SET_ITEM(state, 5, PyLong_FromSsize_t(elsize));
```
tylerjereddy, line 2909: Full testsuite passed locally on ARM Mac when reverting to `PyLong_FromLong` here.

numpy/_core/src/multiarray/dtypemeta.c
```diff
@@ -820,7 +820,7 @@ void_ensure_canonical(_PyArray_LegacyDescr *self)
                         totalsize, field_descr->alignment);
                 maxalign = PyArray_MAX(maxalign, field_descr->alignment);
             }
-            PyObject *offset_obj = PyLong_FromLong(totalsize);
+            PyObject *offset_obj = PyLong_FromSsize_t(totalsize);
```
tylerjereddy, line 823: Full testsuite passed locally on ARM mac when reverting to `PyLong_FromLong`

numpy/_core/src/multiarray/nditer_api.c
```diff
@@ -2296,7 +2296,7 @@ npyiter_clear_buffers(NpyIter *iter)
         /* Buffer cannot be re-used (not that we should ever try!) */
         op_itflags[iop] &= ~NPY_OP_ITFLAG_BUF_REUSABLE;
 
-        int itemsize = dtypes[iop]->elsize;
+        npy_intp itemsize = dtypes[iop]->elsize;
```
tylerjereddy, line 2299: Full testsuite passes locally on ARM Mac when reverting back to `int` here.

numpy/_core/src/multiarray/scalarapi.c
```diff
@@ -469,7 +469,7 @@ PyArray_Scalar(void *data, PyArray_Descr *descr, PyObject *base)
     void *destptr;
     PyArray_CopySwapFunc *copyswap;
     int type_num;
-    int itemsize;
+    npy_intp itemsize;
```
tylerjereddy, line 472: Full testsuite passes locally on ARM Mac when reverting to plain `int`.

numpy/_core/src/multiarray/scalartypes.c.src
```diff
@@ -1709,11 +1708,11 @@ gentype_itemsize_get(PyObject *self, void *NPY_UNUSED(ignored))
 {
     PyArray_Descr *typecode;
     PyObject *ret;
-    int elsize;
+    npy_intp elsize;
 
     typecode = PyArray_DescrFromScalar(self);
     elsize = typecode->elsize;
-    ret = PyLong_FromLong((long) elsize);
+    ret = PyArray_PyIntFromIntp(elsize);
```
tylerjereddy, line 1733: full testsuite passes locally when reverting back to `ret = PyLong_FromLong((long) elsize);`

numpy/_core/src/multiarray/usertypes.c
```diff
@@ -73,9 +73,9 @@ _append_new(int **p_types, int insert)
 static npy_bool
 _default_nonzero(void *ip, void *arr)
 {
-    int elsize = PyArray_ITEMSIZE(arr);
+    npy_intp elsize = PyArray_ITEMSIZE(arr);
     char *ptr = ip;
-    while (elsize--) {
+    while (elsize-- > 0) {
```
tylerjereddy: Does the tractability for "truthiness" of an `npy_intp` vs. a plain `int` really differ enough to require this? Was able to pass full testsuite locally without it?

Looks like more of a style change than directly related to supporting large structured dtypes?

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -3215,7 +3215,7 @@ PyUFunc_Accumulate(PyUFuncObject *ufunc, PyArrayObject *arr, PyArrayObject *out,
                 }
             }
             else {
-                memmove(dataptr_copy[2], dataptr_copy[1], itemsize);
+                memmove(dataptr_copy[2], dataptr_copy[1], (size_t)itemsize);
```
tylerjereddy, line 3209: I don't think the tests can fail if the typecast is reverted because `memmove` will do it implicitly anyway (locally, tests are fine when reverted).

In this case, we might prevent compiler warnings and stuff like that though.

numpy/_core/tests/test_casting_unittests.py
```diff
@@ -693,11 +693,11 @@ def test_unicode_byteswapped_cast(self, order1, order2):
...
-        # Cover a small special case in void to string casting that could
-        # probably just as well be turned into an error (compare
-        # `test_object_to_parametric_internal_error` below).
-        assert np.array([], dtype="V5").astype("S").dtype.itemsize == 5
-        assert np.array([], dtype="V5").astype("U").dtype.itemsize == 4 * 5
+        # These used to succeed for empty arrays, but now we reject them:
+        with pytest.raises(TypeError, match="cannot cast"):
+            np.array([], dtype="V5").astype("S")  # was S5
+        with pytest.raises(TypeError, match="cannot cast"):
+            np.array([], dtype="V5").astype("U")  # was U5
```
tylerjereddy, line 700: I admit, I don't fully understand why we need to mess with this when messing with structured dtype field size limits.

tylerjereddy (commented):
I did a round of "adversarial" self-review where I mostly looked at the testsuite consequences of removing various shims and error handling.

I think we already discussed that hitting all cases with tests isn't going to happen, but it was a way to force me to read through again, and probably a more active NumPy dev  might help prioritize ones for which we really do want a test added for.

I didn't notice anything too wild when reading through, though I admit that the string/void handling changes, and their connection to the structured field size limit changes, are a bit over my head.

Maybe I'll do the `git grep` checking next, then circle back if/when folks pick some cases that I really should write a test for.

numpy/_core/src/multiarray/array_assign_array.c
seberg: some of this code is weird, but I think it's much much better just allow intp in `npy_uint_alignment()`.

EDIT: But no, you can't test this, it could affect speed but crashing only on niche platforms or in UBSAN.

numpy/_core/src/multiarray/descriptor.c
seberg: This (and most/all other converters) are a windows only issue.

numpy/_core/tests/test_casting_unittests.py
seberg: Surely we don't, but I'll guess that without this we would need another check on some branch (and this refactored it away).

seberg (commented):
Put a 👍 if I just agree, just tiny comments seeing this coming through.
Maybe this means you are close to happy? :)

numpy/_core/tests/test_dtype.py
tylerjereddy: `spin test -t numpy/_core/tests/test_dtype.py::TestPickling::test_pickling_large` now passes in `0.03` seconds locally and I believe I've made test improvements in response to Sebastian's comments here. I suppose this started as self-review/criticism anyway, and I believe this thread can probably be resolved.

numpy/_core/tests/test_records.py
tylerjereddy: The test in question does seem to be working properly, `spin test -t numpy/_core/tests/test_records.py::TestFromrecords::test_recarray_fromfile_massive` passes in about 1.5 seconds locally. That is a bit slow perhaps--I'll leave it to Sebastian to decide if this thread is resolved or not.

numpy/_core/src/multiarray/convert_datatype.c
tylerjereddy: Pretty confident we can resolve this review thread now, and it is a self-review anyway (I'm just reading through and checking for what's left)

tylerjereddy:
Now the `PyArray_PythonPyIntFromInt` changes from gh-32585 are causing me some pain here in local work, I don't want regular ints in some of those cases..

tylerjereddy:
Alright, I handled the merge conflicts following gh-32585. The new fastcall changes caused me to lose quite a bit of time because `PyArray_IntpFromPyIntConverter` doesn't live anywhere near the more common `PyArray_PythonPyIntFromInt` in the codebase and I was starting to think I'd have to write my own converter for `npy_parse_arguments`.

As a result, I didn't do an additional `git grep` analysis for `elsize` and so on yet, but I did do a sweep over the outstanding review comments and I believe they've been addressed for the most part (and a decent chunk of them are just self-review comments).

tylerjereddy:
I did a full read through of the diff again--I didn't see anything egregious. The LLM-written tests are a bit hard to read sometimes, but you already noted that.

I asked Claude Opus 5 to look through the review comments and diff against `main` and pick its top two concerns, which I've pasted below the fold. Do either of the top two concerns (bounds checking or `.flat[...]` read path) look serious enough to fix here? Apart from that, I'd argue this probably is starting to look like a step forward.

<details>

```
Concern 1: the bounds- and extent-checking family still takes int itemsize

  This is the one I'd prioritize, because the failures are memory-safety failures rather than clean errors, and one is reachable from pure Python with a 16-byte
  allocation.

  array_new — numpy/_core/src/multiarray/arrayobject.c:1157 still has int itemsize = descr->elsize;, so the "buffer is too small" check at arrayobject.c:1221
  multiplies a truncated itemsize:

  np.frombuffer(count=1): ValueError: buffer is smaller than requested size
  np.ndarray(buffer=)   : ACCEPTED -> nbytes=4294967296 from a 16-byte buffer

  The PR fixed PyArray_FromBuffer but not the np.ndarray(shape, dtype, buffer=...) constructor next to it. Reading a['pad'][0, 2**32-1] on that array reads ~4 GiB past
  a 16-byte bytearray and returns quietly. Line 1218 (shape=(-1,)) also divides by the truncated itemsize — on arm64 that silently yields (0,); on x86-64 integer
  division by zero traps, so that's a SIGFPE there.

  offset_bounds_from_strides(const int itemsize, const int nd, ...) — numpy/_core/src/common/mem_overlap.c:660 feeds get_array_memory_extents → solve_may_share_memory.
  With itemsize = 2**31 truncating to INT_MIN, the extent goes backwards:

  elsize: 2147483648
  shares_memory(a, a[::-1]) -> False (expected True)
  shares_memory(a, a)       -> False (expected True)
  a[:] = a[::-1]  ->  a['m'] = [2 2]   expected [2 1]

  That last line is silent data corruption: array_assign_array.c:384 takes PyArray_HASFIELDS(dst) && arrays_overlap(src, dst) to decide whether to make the defensive
  temporary copy, the overlap check says "no", and the copy is skipped. The same extent computation guards nditer_constr.c:3172/3181, mapping.c:789/2000, and
  item_selection.c:303/434/728.

  Also in this family: PyArray_CheckStrides(int elsize, ...) (arrayobject.c:1091, called from getset.c:195 for arr.strides = ...) and PyArray_ElementStrides
  (arrayobject.c:1047, strides[i] % itemsize → division by zero when elsize % 2**32 == 0, reached via NPY_ARRAY_ELEMENTSTRIDES at ctors.c:1830).

  Both PyArray_CheckStrides (slot 105) and PyArray_ElementStrides (slot 119) are public C-API, so given Sebastian's ruling on PyArray_GetField the signatures can't
  move — this wants either call-site guards (the pattern already used for npy_uint_alignment in array_assign_array.c) or npy_intp internal helpers that the public
  wrappers delegate to. offset_bounds_from_strides is NPY_VISIBILITY_HIDDEN, so that one can just become npy_intp.

  Concern 2: the .flat[...] read path was missed while the write path was fixed

  The PR converted iter_ass_sub_Bool (iterators.c:693) and iter_ass_sub_int (iterators.c:724) to npy_intp itemsize, but left their read-side siblings in the same file:
  iter_subscript_Bool (iterators.c:360) and iter_subscript_int (iterators.c:435).

  elsize: 2147483648
  flat[[1,0]]['m'] = [2 0]   expected [2 1]
  flat[bool]['m']  = [1 0]   expected [1 2]

  Beyond the wrong values, optr += itemsize (iterators.c:418, :501) advances the output pointer by INT_MIN, so every item after the first is written ~2 GiB before the
  start of the freshly allocated result buffer — an out-of-bounds write that happened not to fault on this run. Narrow fix: npy_intp at those two declarations.

  Smaller items also worth a look

  - npy_uint_alignment(int itemsize) (common.h:288) — the PR guards exactly one of nine call sites. array_assign_scalar.c:52/55/148/151/270,
    dtype_transfer.c:3395/3399, and array_assign.c:148 (IsUintAligned, used almost everywhere) still convert npy_intp → int implicitly, and a truncated value that
    lands on 1/2/4/8/16 yields a false "uint-aligned" claim. Sebastian already suggested taking intp in the function itself; doing that removes the special case in
    array_assign_array.c too.
  - arraytypes.c.src:2377 — int subitemsize in VOID_copyswapn is the exact sibling of the VOID_copyswap line the PR did convert (:2471 in the diff). Only reachable via
    a nested structured subarray base, but truncation gives either a silent early return (at :2399) or descr->elsize / subitemsize division by zero at :2418.
  - buffer.c:275 — new_offset = PyLong_AsLong(offset_obj) for field offsets. long is 32-bit on Windows, so memoryview(arr) on a large structured array will raise
    OverflowError there while working on Linux/macOS — the same Windows-only divergence class that cost time earlier in this PR with PyLong_FromLong. Wants
    PyLong_AsSsize_t. Separately, the padding loop at :290 appends one 'x' per byte of gap, so a multi-GB field offset builds a multi-GB format string.
  - _multiarray_tests.c.src:150 — npy_int itemsize = PyArray_ITEMSIZE(itx->ao); (test helper only).

  I checked each of the above against the source but only ran the targeted reproducers above, not the full test suite. Nothing was pushed and no issues or PRs were
  opened.
```

</details>

seberg:
Nice! I think it has a point (but maybe you already changed some as one was in your own review):
* `npy_uint_alignment` and `offset_bounds_from_strides` should just change their input type (that removes the guard on one `npy_uint_alignment` call).
* The `.flat[]` concern also sounds like simply changing a local C variable to `npy_intp itemsize`, might as well push that one I think.
* `arraytypes.c.src:2377` sounds also like a clear fix, although I guess I am not quite sure if there could be overflows later.
* The `buffer.c` one I don't care, it's an OK limitation.  The test narrowing might be nice to update, but also OK if it is a small inconsistency.

Either way, agreed. Maybe we can push those fixes and then put it in? If you prefer can also reverse that, put it in then follow up (and maybe do one more pass).

(I am slightly wondering if we should expand the "security" note that general structured dtypes shouldn't jut be accepted a bit in this direction, but that isn't for here.)

tylerjereddy:
Ok, I pushed in a commit for the first 3 points after checking the diff/lint/testing. The diff is just a bunch of `int`->`npy_intp` changes and one guard removal, which sounds consistent with your assessment.

For the concern about `VOID_copyswapn` overflow, Opus 5 indicated it was not possible anymore after the changes here (if you believe that.. though it has been pretty good lately..).

PR #31341, opened 2026-04-27
BUG: fix np.interp with empty arrays to return empty array

### LINKED ISSUE #30316: BUG: np.interp([], [], []) fails, when it could return an empty array

robrui:
Fixes #30316

When `np.interp([], [], [])` is called with all empty inputs, it now returns an empty array of the appropriate dtype instead of raising `ValueError: array of sample points is empty`.

**Changes:**
- Added check in `interp()` that detects when all of `x`, `xp`, and `fp` are empty arrays and returns an empty array
- Added regression test `test_empty_arrays` in `TestInterp`

**Note:** `np.interp(non_empty, [], [])` (where x is non-empty but xp is empty) still raises ValueError as before.

### DIFF
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -1672,6 +1672,10 @@ def interp(x, xp, fp, left=None, right=None, period=None):
         xp = np.concatenate((xp[-1:] - period, xp, xp[0:1] + period))
         fp = np.concatenate((fp[-1:], fp, fp[0:1]))
 
+    # Handle empty inputs: when all arrays are empty, return empty array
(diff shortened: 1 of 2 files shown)


jorenham:
We prefer to talk to humans. You ignored the PR template. This is also a duplicate of #31339

PR #31347, opened 2026-04-28
BUG: `np.linalg.svd(..., hermitian=True)` returns non-unitary `vh`

EarlMilktea:
### PR summary

Currently `np.linalg.svd` can return non-unitary `vh` due to sign 0 appearing when input array is singular.
This PR fixes the issue.

```python
>>> import numpy as np
>>> res = np.linalg.svd(np.array([[1, 0], [0, 0]]), hermitian=True)
>>> res.Vh
array([[1., 0.],
       [0., 0.]])
```

#### AI Disclosure

I've used ChatGPT to ensure that my PR is following the guideline. No codes are generated by AI.


EarlMilktea:
MEMO: I suspect two CI failures are not related to this PR.

numpy/linalg/_linalg.py
```diff
@@ -1810,6 +1810,8 @@ def svd(a, full_matrices=True, compute_uv=True, hermitian=False):
         if compute_uv:
             s, u = eigh(a)
             sgn = sign(s)
+            # avoid zero sign
+            sgn = np.where(sgn == 0, 1, sgn)
```
seberg: Thanks, can we assume that -0 isn't a thing and just use `signbit` instead of `sign`?

EarlMilktea: I personally prefer my approach to make the result predictable...

jorenham: well, there's a performance penalty with this approach

seberg: Yeah, but if we aren't sure (I dunno yet), there is already sorting, etc. going on, so it shouldn't actually matter (`where` just always feels slightly awkward to me, I guess).

seberg: Anyway, I think it's probably good. But @WarrenWeckesser might have a quick thought and know this a bit better.

EarlMilktea: Thank you all for comments. Let me know if there are any updates.

EarlMilktea: @WarrenWeckesser Could you kindly review this PR for me? Thank you in advance!

numpy/linalg/tests/test_linalg.py
```diff
@@ -745,6 +745,12 @@ def hermitian(mat):
 class TestSVDHermitian(SVDHermitianCases, SVDBaseTests):
     hermitian = True
 
+    def test_singular(self):
+        x = np.array([[1, 0], [0, 0]])
+        u, _, vh = linalg.svd(x, hermitian=self.hermitian)
+        assert_almost_equal(u @ u.T.conj(), np.eye(2))
```
WarrenWeckesser: The use of `assert_almost_equal` is discouraged; see the note in the [docstring](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html).  Replace these with `assert_allclose(...)`.

Also, it is generally a good idea to always be explicit about the desired relative tolerance when doing a floating point comparison.  The default for `assert_allclose()` is `rtol=1e-7`, but we should expect much better than that here.  Just how small to make it is a judgement call, but `rtol=1e-14` seems reasonable. Note that the default `atol` is 0, so the test will require that the off-diagonals of the product are exactly 0.  For this 2x2 matrix, that is probably fine.

numpy/linalg/_linalg.py
WarrenWeckesser: I agree with @seberg: `where` is awkward. It is also a bit slow in this case.  I recommend the one-liner `sgn = np.copysign(1.0, s)`, which is simpler and faster than using `where`. (Admittedly, the performance of the sign calculation is not a big part of the total time for this function, but that doesn't mean the performance difference should be ignored.)

WarrenWeckesser (changes requested):
Thanks @EarlMilktea.  The fix looks correct, but I have some suggested changes in the implementation.

numpy/linalg/tests/test_linalg.py
EarlMilktea: I've replaced them with `assert_allclose` and set `rtol` explicitly.

WarrenWeckesser (approved):
Looks good to me.

PR #31348, opened 2026-04-28
ENH: Add armv7l wheel builds (manylinux)

Lauszus:
### PR summary

This PR adds support for building armv7l (Raspberry Pi) wheels. 

Addresses the feature request in gh-31351. Please also see my corresponding numpy release PR: https://github.com/numpy/numpy-release/pull/45

My name is Kristian and currently work at https://candela.com/ where we use numpy for various Python applications. Some of these applications are running on a Raspberry Pi compute module 4. For now we have been building numpy from source in CI by extending the official manylinux Docker images.

I thought it would be nice to build it upstream instead to speed up the CI, but also I think it will benefit a lot of people using Raspberry Pi for various hobby projects.

#### AI Disclosure

I used Claude Sonnet 4.6 to fork and push the commit, but I told it exactly which lines to change.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -39,6 +39,8 @@ jobs:
           - [ubuntu-22.04, musllinux_x86_64, ""]
           - [ubuntu-22.04-arm, manylinux_aarch64, ""]
           - [ubuntu-22.04-arm, musllinux_aarch64, ""]
+          - [ubuntu-22.04-arm, manylinux_armv7l, ""]
+          - [ubuntu-22.04-arm, musllinux_armv7l, ""]
           - [macos-15-intel, macosx_x86_64, openblas]
           - [macos-14, macosx_arm64, openblas]
           - [windows-2022, win_amd64, ""]
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -106,6 +106,7 @@ jobs:
         uses: pypa/cibuildwheel@8d2b08b68458a16aeb24b64e68a09ab1c8e82084  # v3.4.1
         env:
           CIBW_BUILD: ${{ matrix.python }}-${{ matrix.buildplat[1] }}
+          CIBW_ARCHS: ${{ endsWith(matrix.buildplat[1], 'armv7l') && 'armv7l' || 'auto' }}
 
       - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -40,7 +40,6 @@ jobs:
           - [ubuntu-22.04-arm, manylinux_aarch64, ""]
           - [ubuntu-22.04-arm, musllinux_aarch64, ""]
           - [ubuntu-22.04-arm, manylinux_armv7l, ""]
-          - [ubuntu-22.04-arm, musllinux_armv7l, ""]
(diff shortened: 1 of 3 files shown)


andyfaff:
I'd note that the RPi5 units, and the CM5 modules, can both use the manylinux_aarch64 wheels that are currently on PyPI.

andyfaff:
I'm pretty sure that CM4 and RPi4 would also be able to use the manylinux_aarch64 wheels as the tech specs state "Broadcom BCM2711, Quad core Cortex-A72 (ARM v8) 64-bit SoC @ 1.8GHz"

Lauszus:
> I'm pretty sure that CM4 and RPi4 would also be able to use the manylinux_aarch64 wheels as the tech specs state "Broadcom BCM2711, Quad core Cortex-A72 (ARM v8) 64-bit SoC @ 1.8GHz"

Yes that is true, but we are stuck at 32-bit and cannot easily upgrade, as they are already deployed

Lauszus:
I have now opened up an issue: https://github.com/numpy/numpy/issues/31351

rgommers:
@Lauszus did you test the new CI jobs on your fork, and if so can you link to a passing log? Note that it's normal to develop/debug CI jobs on your fork first, because you can run just the relevant job, rather than triggering O(75) jobs on a PR.

rgommers:
It looks like for regular armv7l CI coverage, we already have that:

https://github.com/numpy/numpy/blob/3bea241a0fbf5610f876dc54d4caf7380f75471e/.github/workflows/linux.yml#L170

It was just a little harder to dig up because of the `armhf` vs. `armv7l` naming, but it's the same platform - the former is the Debian/Ubuntu naming for the specific 32-bit ABI. 

So my read is that this doesn't increase coverage, unless we want to consider armv7l on Musl - but does anyone use that, and does it turn up any new issues now that it's run for the first time (see my question about results above)?

Lauszus:
> It looks like for regular armv7l CI coverage, we already have that:
> 
> https://github.com/numpy/numpy/blob/3bea241a0fbf5610f876dc54d4caf7380f75471e/.github/workflows/linux.yml#L170
> 
> It was just a little harder to dig up because of the `armhf` vs. `armv7l` naming, but it's the same platform - the former is the Debian/Ubuntu naming for the specific 32-bit ABI.
> 
> So my read is that this doesn't increase coverage, unless we want to consider armv7l on Musl - but does anyone use that, and does it turn up any new issues now that it's run for the first time (see my question about results above)?

Yes does not increase coverage, as you said it's already supported (I have been compiling my own wheels for some time now).

The problem with piwheels is that they compile the wheels on an actual Raspberry Pi and only for the default Python version for each OS they support (Trixie, Bookworm and Bullseye), so if you are not using the default Python version or their OS, then the wheels won't work.

I agree on dropping the musl wheels, as I don't think anyone actually uses that.

Would you consider shipping armv7l manylinux wheels?

Lauszus:
I have now updated the PR to only build manylinux wheels

rgommers:
@andyfaff I'm inclined to merge this PR and remove the `armhf` job at https://github.com/numpy/numpy/blob/3bea241a0fbf5610f876dc54d4caf7380f75471e/.github/workflows/linux.yml#L170, given that this is much easier to maintain, and provides the same coverage. WDYT?

> Would you consider shipping armv7l manylinux wheels?

I think the response in https://github.com/numpy/numpy/issues/31351#issuecomment-4413461158 is final (unless another maintainer disagrees, but that seems unlikely), so we don't plan to upload wheels to PyPI.

PR #31352, opened 2026-04-28
MAINT: Simplify C++ sorting tags

seberg:
Manaas sorting PR added a huge amount of duplication, but while this all might have made sense with C++11, the whole pattern is a ridiculous amount of code-duplication with C++17.

This cleans it up, I used claude with guidance/follow-up, not sure if the `_type` pattern is best, but it works well and only changes the `_tag` header for now, which is nice.

For the next step this should make things rather simple: We can introduce a `cmp<bool reverse>` template function (at worst per kind), I think.

(It also removed the unused `_SWAP` macros, the most complicated change is moving the half helpers into the `numpy_tag.h` file, which is nicer, IMO.)

(Heavy use of claude, as it is mechanical replacement in a sense. -- It first gave me silly macros to make it shorter...)

### DIFF
--- a/numpy/_core/src/common/numpy_tag.h
+++ b/numpy/_core/src/common/numpy_tag.h
@@ -1,259 +1,224 @@
-#ifndef _NPY_COMMON_TAG_H_
-#define _NPY_COMMON_TAG_H_
+#ifndef NUMPY_CORE_SRC_COMMON_NUMPY_TAG_H_
+#define NUMPY_CORE_SRC_COMMON_NUMPY_TAG_H_
 
-#include "../npysort/npysort_common.h"
+#include "numpy/ndarraytypes.h"
+#include "numpy/npy_common.h"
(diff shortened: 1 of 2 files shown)


seberg:
Would be nice to get this in fairly quickly, so that gh-31345 can be rebased on top of it and avoid bloating it even more.

MaanasArora:
Very nice, huge deletions! I did wonder why the integer/float compares or less equals were duplicated, when basically identical... it might have been C++ 11, but for less equal in particular a base template could have cleared a lot up anyway! Sorry, I just didn't realize this was legacy I guess.

Yeah, a bool argument to the template makes sense, we would probably still need greater functions for float at least, because of NaNs. But this should definitely make things much smaller, thanks a lot!

charris:
Wow, it's almost readable :) Thanks Sebastian.

charris (approved):

PR #31357, opened 2026-04-28
BUG: fix memory leak in np.zeros when fill-zero loop raises (#31320)

charris:
Backport of #31320.

BUG: fix memory leak in np.zeros when fill-zero loop raises

### PR summary

This PR fixes a memory leak in np.zeros that occurs when the fill-zero loop raises an exception (which can happen for custom user-defined dtypes). It does this by setting NPY_ARRAY_OWNDATA earlier to ensure the data will be deallocated, combined with changes to mem_handler cleanup to avoid a double-decref.

I'm not entirely sure whether this is the most appropriate way to fix the leak, but it does seem to work correctly based on testing with `tracemalloc`.

#### AI Disclosure

The bug was detected through testing with `tracemalloc`, then located and fixed by Claude Opus 4.7.
Full AI-generated description: https://gist.github.com/MaartenBaert/f3eadd280f84ef6896586025f59788ec

### DIFF
--- a/numpy/_core/src/multiarray/arrayobject.c
+++ b/numpy/_core/src/multiarray/arrayobject.c
@@ -442,9 +442,9 @@ array_dealloc(PyArrayObject *self)
                 nbytes = 1;
             }
             PyDataMem_UserFREE(fa->data, nbytes, fa->mem_handler);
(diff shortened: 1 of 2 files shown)


charris:
Blas failures are unrelated.

PR #31360, opened 2026-04-29
BUG: Fix signed overflow issue in npy_gcd for INT_MIN on s390x

### LINKED ISSUE #31359: BUG: np.gcd returns negative result for INT_MIN input on s390x
### Describe the issue:

np.gcd returns a negative value when one of the inputs is the minimum value of a signed integer type on s390x. The signed-to-unsigned absolute value conversion in npy_gcd (npy_math_internal.h.src) uses `a < 0 ? -a : a`, which is undefined behavior when `a = INT_MIN`.

Found it when running tests on a RHEL9/CentOS Stream 9 machine with Python3.14 and GCC 11.5. I wasn't able to reproduce it on other archs or even later GCC versions.

```

___________________ TestRationalFunctions.test_gcd_overflow ____________________

self = <test_umath.TestRationalFunctions object at 0x3ff66a32690>

    def test_gcd_overflow(self):
        for dtype in (np.int32, np.int64):
            # verify that we don't overflow when taking abs(x)
            # not relevant for lcm, where the result is unrepresentable anyway
            a = dtype(np.iinfo(dtype).min)  # negative power of two
            q = -(a // 4)
>           assert_equal(np.gcd(a,  q * 3), q)
E           AssertionError: 
E           Items are not equal:
E            ACTUAL: np.int32(-536870912)
E            DESIRED: np.int32(536870912)

a          = np.int32(-2147483648)
dtype      = <class 'numpy.int32'>
q          = np.int32(536870912)
self       = <test_umath.TestRationalFunctions object at 0x3ff66a32690>
```
This seems to be fixed when using -fwrapv

```python
import numpy as np
a = np.int32(np.iinfo(np.int32).min)
q = -(a // 4)
print(np.gcd(a, q * 3))

That should produce 536870912, instead it produces -536870912
```

### Error message:

```shell

```

Numpy 2.3.4 (but also reproduced on latest versions
Python 3.14
GCC 11.5.0
CFLAGS: -O2 -march=z14

### Runtime Environment:

[{'numpy_version': '2.3.4',
  'python': '3.14.3 (main, Feb  4 2026, 00:00:00) [GCC 11.5.0 20240719 (Red '
            'Hat 11.5.0-14)]',
  'uname': uname_result(system='Linux', node='removed', release='5.14.0-697.el9.s390x', version='#1 SMP Wed Apr 22 08:43:20 UTC 2026', machine='s390x')},
 {'simd_extensions': {'baseline': [],
                      'found': ['VX', 'VXE', 'VXE2'],
                      'not_found': []}},
 {'available_backends': ['NETLIB', 'OPENBLAS-OPENMP'],
  'current_backend': 'OPENBLAS-OPENMP',
  'filepath': '/usr/lib64/libflexiblas.so.3.0',
  'internal_api': 'flexiblas',
  'loaded_backends': ['OPENBLAS-OPENMP'],
  'num_threads': 4,
  'prefix': 'libflexiblas',
  'user_api': 'blas',
  'version': '3.0.4'},
 {'architecture': 'Z14',
  'filepath': '/usr/lib64/libopenblaso-r0.3.29.so',
  'internal_api': 'openblas',
  'num_threads': 4,
  'prefix': 'libopenblas',
  'threading_layer': 'openmp',
  'user_api': 'blas',
  'version': '0.3.29'},
 {'filepath': '/usr/lib64/libgomp.so.1.0.0',
  'internal_api': 'openmp',
  'num_threads': 4,
  'prefix': 'libgomp',
  'user_api': 'openmp',
  'version': None}]

_No response_

AnkitAhlawat7742:
### PR summary

This PR fixes signed overflow issue  in npy_gcd when one of the inputs is the minimum value of a signed integer type (INT_MIN) on s390x architecture.

issue was the expression a < 0 ? -a : a in npy_math_internal.h.src causes undefined behavior when a = INT_MIN because negating INT_MIN overflows a signed integer on s390x. i see it is working fine with other arch so fix only provided to s390x 

#### AI Disclosure
No AI tools were used in identifying, reproducing, or fixing this bug. The bug was identified by running below code after updating the code ..before the code update i see the mismatch in expected and actual result which i informed in issue #31359
```
python3 -c "
import numpy as np
a = np.int32(np.iinfo(np.int32).min)
q = -(a // 4)
result = np.gcd(a, q * 3)
print('Result  :',result)
print('Expected: 536870912')
print('Status  :','PASS' if result == np.int32(536870912) else 'FAIL')
"
Result  : 536870912
Expected: 536870912
Status  : PASS
```

### DIFF
--- a/numpy/_core/tests/test_umath.py
+++ b/numpy/_core/tests/test_umath.py
@@ -4300,10 +4300,14 @@ class TestRationalFunctions:
         # verify that we don't overflow when taking abs(x) for INT_MIN
         # this was undefined behavior that manifested on s390x with GCC 11.5
         for dtype in (np.int8, np.int16, np.int32, np.int64):
-            a = dtype(np.iinfo(dtype).min)  # negative power of two
+            a = dtype(np.iinfo(dtype).min)  # INT_MIN
             q = -(a // 4)
+            # Test with INT_MIN as first argument
             assert_equal(np.gcd(a,  q * 3), q)
             assert_equal(np.gcd(a, -q * 3), q)
+            # Test with INT_MIN as second argument
+            assert_equal(np.gcd(q * 3,  a), q)
+            assert_equal(np.gcd(-q * 3, a), q)
 
     def test_decimal(self):
         from decimal import Decimal
--- a/numpy/_core/tests/test_umath.py
+++ b/numpy/_core/tests/test_umath.py
@@ -4297,9 +4297,9 @@ class TestRationalFunctions:
         assert_equal(np.lcm(a, b), 10 * big)
 
     def test_gcd_overflow(self):
-        for dtype in (np.int32, np.int64):
-            # verify that we don't overflow when taking abs(x)
-            # not relevant for lcm, where the result is unrepresentable anyway
+        # verify that we don't overflow when taking abs(x) for INT_MIN
+        # this was undefined behavior that manifested on s390x with GCC 11.5
+        for dtype in (np.int8, np.int16, np.int32, np.int64):
             a = dtype(np.iinfo(dtype).min)  # negative power of two
             q = -(a // 4)
             assert_equal(np.gcd(a,  q * 3), q)
--- a/numpy/_core/src/npymath/npy_math_internal.h.src
+++ b/numpy/_core/src/npymath/npy_math_internal.h.src
@@ -650,12 +650,24 @@ npy_lcm@c@(@type@ a, @type@ b)
  *
  * #type = (npy_int, npy_long, npy_longlong)*2#
  * #c = (,l,ll)*2#
+ * #utype = (npy_uint, npy_ulong, npy_ulonglong)*2#
  * #func=gcd*3,lcm*3#
  */
 NPY_INPLACE @type@
 npy_@func@@c@(@type@ a, @type@ b)
 {
+    #if defined(__s390__) || defined(__s390x__)
+    /*
+     * On s390x, negating INT_MIN results in a negative value due to how
+     * the architecture handles signed overflow. Cast to unsigned first
+     * to avoid this issue.git issue #31359
+     */
+    @utype@ a_abs = a < 0 ? (@utype@)(0) - (@utype@)a : (@utype@)a;
+    @utype@ b_abs = b < 0 ? (@utype@)(0) - (@utype@)b : (@utype@)b;
+    return (@type@)npy_@func@u@c@(a_abs, b_abs);
+#else
     return npy_@func@u@c@(a < 0 ? -a : a, b < 0 ? -b : b);
+#endif
 }
 /**end repeat**/
 
--- a/numpy/_core/src/npymath/npy_math_internal.h.src
+++ b/numpy/_core/src/npymath/npy_math_internal.h.src
@@ -656,18 +656,15 @@ npy_lcm@c@(@type@ a, @type@ b)
 NPY_INPLACE @type@
 npy_@func@@c@(@type@ a, @type@ b)
 {
-    #if defined(__s390__) || defined(__s390x__)
     /*
-     * On s390x, negating INT_MIN results in a negative value due to how
-     * the architecture handles signed overflow. Cast to unsigned first
-     * to avoid this issue.git issue #31359
+     * Cast to unsigned to avoid overflow when negating minimum signed value.
+     * For negative values, cast to unsigned first, then apply negation in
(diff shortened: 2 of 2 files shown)


stratakis:
Since this is undefined behavior, shouldn't it be fixed for all the platforms, even if it manifests only on s390x?

seberg:
I'd agree, compilers presumably are smart enough to do whatever is best on the platform, so we can just keep the code identical.  But if we bother to guarantee this, we should also add a test. (Ideally for all relevant combinations here.)

AnkitAhlawat7742:
Thank you for the quick review!
@stratakis @seberg — Agreed! Since this is undefined behavior, I have updated the fix to apply universally across all platforms rather than targeting s390x specifically. 
As suggested by @seberg, I have also extended the existing test_gcd_overflow test to cover all relevant signed integer type combinations (int8, int16, int32, int64) to prevent any future regressions.

AnkitAhlawat7742:
Hi @seberg ,
I noticed the CI failure in the "BLAS tests (Linux)" job is unrelated to my changes. It is failing due to a missing BLAS library in the GitHub Actions runner environment (No BLAS library detected).
Can you please help me here

seberg:
Just ignore it, MKL had an update or so and now it's failing everywhere, we need to see what is actually wrong (but I am not sure anyone yet did).

numpy/_core/tests/test_umath.py
```diff
@@ -4297,9 +4297,9 @@ def test_lcm_overflow(self):
         assert_equal(np.lcm(a, b), 10 * big)
 
     def test_gcd_overflow(self):
-        for dtype in (np.int32, np.int64):
-            # verify that we don't overflow when taking abs(x)
-            # not relevant for lcm, where the result is unrepresentable anyway
+        # verify that we don't overflow when taking abs(x) for INT_MIN
+        # this was undefined behavior that manifested on s390x with GCC 11.5
+        for dtype in (np.int8, np.int16, np.int32, np.int64):
```
seberg, line 4302: > that manifested on s390x with GCC 11.5

Ah, so we already have a test that but it only manifests with old GCC?  Do we need a test for `q == INT_MIN` as well, since you are fixing it?

AnkitAhlawat7742: Yes, the existing test `test_gcd_overflow` was already there but it only caught the bug on old GCC (11.5) on `s390x`. On other platforms like `macOS ARM64 or x86 `with newer GCC, the test passed even before the fix.

Regarding q == INT_MIN — I investigated and found even with the current fix, `gcd(INT_MIN, INT_MIN)` still returns a negative value because `abs(INT_MIN)` cannot be represented in the same signed type. 

For example i have tested math.gcd(-128, -128) returns 128 but `np.gcd` returns -128 for int8. Should I open a separate issue to track this?

I have also added test cases where INT_MIN is passed as the second argument to verify commutativity — ensuring the fix works regardless of argument order.

seberg: Yeah, I was just surprised since we have s390x CI.  So it seems like the issue almost went away by itself as gcc 11.5 is pretty old (not that I mind the change, seems safe enough).

AnkitAhlawat7742: That makes sense! 
Since the fix is safe and correct for all platforms, it is better to have it. future compiler updates or optimization level changes could still trigger it. Happy to make any further adjustments if needed!

stratakis: If you do an objdump on `numpy/_core/_multiarray_umath.cpython-314-s390x-linux-gnu.so`  you can see the bug triggered on this specific configuration with GCC 11.5 on s390x:

lpr  %r2,%r4   # overflows INT_MIN so it stays negative
dsgfr %r0,%r2  # signed division with negative value which propagates further, aka our bug

When using -fwrapv:
lpr  %r5,%r3 # Still overflow, INT_MIN negative
llgfr %r4,%r5 # zero-extend so the value is treated as unsigned
dlgr %r2,%r4 # unsigned division which produces the correct result

So with -fwrapv the issue is "resolved" in a way, but the original UB is still there.

The problem is in the lpr, which instead of computing the absolute value, it propagates a negative one.

Haven't disassembled with the patch from the PR, but I think that should resolve it.

seberg:
Thanks @stratakis and @AnkitAhlawat7742, let's just get it in.  Simple enough, and if we already have a test for it good to just avoid the UB probably.

PR #31365, opened 2026-04-30
Handle Python scalars consistently in np.astype

mncrftfrcnm:

I use numpy as a machine learner, and data-analytic. I also use it for a lot of other tasks, but primarily for ML. 
I was interested in reviewing the code of numpy, and how it works, and i saw a bug in astype.

So, the issue i found was in the astype function. This PR update 'np.astype':, so python scalar inputs are also handled, instead of ending with error 'AttributeError'
>>> import numpy as np
>>> np.astype(np.int64(1), np.float64)
np.float64(1.0)
>>> np.astype(1, np.float64)
  File "C:\Users\vov\anaconda3\Lib\site-packages\numpy\_core\numeric.py", line 2730, in astype
    return x.astype(dtype, copy=copy)
           ^^^^^^^^
AttributeError: 'int' object has no attribute 'astype'
>>>

So, as python integers do not have a .astype attribute, in my edits, if the integer is scalar, it is converted by the np.asarray , which does not raise an issue, and makes the python-native scalars work fine

#### AI Disclosure

No AI tools used

### DIFF
--- a/numpy/_core/numeric.py
+++ b/numpy/_core/numeric.py
@@ -2673,7 +2673,10 @@ def astype(x, dtype, /, *, copy=True, device=None):
             'Device not understood. Only "cpu" is allowed, but received:'
             f' {device}'
         )
+    if isscalar(x):
+        return np.asarray(x).astype(dtype, copy=copy)
     return x.astype(dtype, copy=copy)
+
 
 
 inf = PINF
--- a/numpy/_core/numeric.py
+++ b/numpy/_core/numeric.py
@@ -2674,7 +2674,7 @@ def astype(x, dtype, /, *, copy=True, device=None):
             f' {device}'
         )
     if isscalar(x):
-        return np.asarray(x).astype(dtype, copy=copy)
+        return np.asarray(x).astype(dtype, copy=copy)[()]
     return x.astype(dtype, copy=copy)
 
 
--- a/numpy/_core/numeric.py
+++ b/numpy/_core/numeric.py
@@ -2678,7 +2678,6 @@ def astype(x, dtype, /, *, copy=True, device=None):
     return x.astype(dtype, copy=copy)
 
 
-
 inf = PINF
 nan = NAN
 False_ = nt.bool(False)


numpy/_core/numeric.py
```diff
@@ -2673,9 +2673,12 @@ def astype(x, dtype, /, *, copy=True, device=None):
             'Device not understood. Only "cpu" is allowed, but received:'
             f' {device}'
         )
+    if isscalar(x):
+        return np.asarray(x).astype(dtype, copy=copy)
```
jorenham: Scalars should remain scalars, but this changes scalars to (0-d) arrays

```suggestion
        return np.asarray(x).astype(dtype, copy=copy)[()]
```

numpy/_core/numeric.py
```diff
@@ -2673,9 +2673,12 @@ def astype(x, dtype, /, *, copy=True, device=None):
...
     return x.astype(dtype, copy=copy)
 
 
+
```
jorenham: unrelated
```suggestion
```

seberg:
Not sure we should do this, since this function really exists only for the Array API, I would say, so I am not sure there is a reason to broaden it up beyond such use.

mncrftfrcnm:
> Not sure we should do this, since this function really exists only for the Array API, I would say, so I am not sure there is a reason to broaden it up beyond such use.

but still, it may be an issue?

seberg:
To me it isn't an issue until someone asks from it from a usability perspective outside of the original context.  But if other maintainers prefer to do coerction to array here because NumPy usually does, sure...

mncrftfrcnm:
> To me it isn't an issue until someone asks from it from a usability perspective outside of the original context. But if other maintainers prefer to do coerction to array here because NumPy usually does, sure...

That makes sense, and i also don't think anyone would often use this outside of what it is intended for.
How about returning a typeError(like for the rest issues) if the python int is passed, instead of attributeError?

(raise TypeError(
            "Input should be a NumPy array or scalar. "
            f"It is a {type(x)} instead."
        ))

jorenham:
> To me it isn't an issue until someone asks from it from a usability perspective outside of the original context.

I agree. In the stubs it's even stricter currently, because it only accepts `ndarray`, so no scalars at all:

https://github.com/numpy/numpy/blob/5cba3b4aa6c4847c6595d7e0a8e9f2f077296c79/numpy/_core/numeric.pyi#L1349-L1366

> How about returning a typeError(like for the rest issues) if the python int is passed, instead of attributeError?

Sounds good to me

mncrftfrcnm:
> > To me it isn't an issue until someone asks from it from a usability perspective outside of the original context.
> 
> I agree. In the stubs it's even stricter currently, because it only accepts `ndarray`, so no scalars at all:
> 
> https://github.com/numpy/numpy/blob/5cba3b4aa6c4847c6595d7e0a8e9f2f077296c79/numpy/_core/numeric.pyi#L1349-L1366
> 
> > How about returning a typeError(like for the rest issues) if the python int is passed, instead of attributeError?
> 
> Sounds good to me

So, should I create a new branch/PR, where the code will return attribute error for python-native numbers, or should I make changes on this one?

jorenham:
> So, should I create a new branch/PR, where the code will return attribute error for python-native numbers, or should I make changes on this one?

Whatever you prefer

mncrftfrcnm:
ok, i will create a new branch with PR

PR #31369, opened 2026-05-01
MAINT: Change the Scalar check in astype to only accept np objects

mncrftfrcnm:
### PR summary

So, the issue i found was in the astype function. This PR updates 'np.astype':, so python scalar inputs return attribute error, instead of ending with error 'AttributeError', which is not entirely correct

here is what the fixed version looks like, instead of attribute error
(error message edited a bit, so only the final error message is shown, omiting all other details, like: 'Traceback (most recent call last):
  File "<stdin>", line 1, in <module>')

```text
>>> import numpy

>>> print(numpy.astype(1, numpy.float64))
TypeError: Input should be a NumPy array or a NumPy scalar. It is a <class 'int'> instead.

>>> print(numpy.astype(1.0, numpy.float64))
TypeError: Input should be a NumPy array or a NumPy scalar. It is a <class 'float'> instead.

>>> print(numpy.astype(True, numpy.float64))
TypeError: Input should be a NumPy array or a NumPy  scalar. It is a <class 'bool'> instead.

>>> print(numpy.astype(numpy.int64(1), numpy.float64))
1.0

>>> print(numpy.astype(numpy.float64(1.0), numpy.float64))
1.0
>>> print(numpy.astype(numpy.datetime64("2024-01-01"), numpy.datetime64))
2024-01-01
>>> print(numpy.astype(numpy.array([1, 2, 3]), numpy.float64))
[1. 2. 3.]
```

here is what original original function output looked like:
```text
>>> import numpy as np
>>> np.astype(1, np.float64)
AttributeError: 'int' object has no attribute 'astype'
```

I use numpy as a machine learner, and data-analytic. I also use it for a lot of other tasks, but primarily for ML.
I was interested in reviewing the code of numpy, and how it works, and i saw a bug in astype.
#### AI Disclosure

No AI tools used

### DIFF
--- a/numpy/_core/numeric.py
+++ b/numpy/_core/numeric.py
@@ -2665,7 +2665,7 @@ def astype(x, dtype, /, *, copy=True, device=None):
     """
     if not isinstance(x, (np.ndarray, np.generic)):
         raise TypeError(
-            "Input should be a NumPy array or scalar. "
+            "Input should be a NumPy array or a NumPy scalar. "
             f"It is a {type(x)} instead."
         )
     if device is not None and device != "cpu":
--- a/numpy/_core/numeric.py
+++ b/numpy/_core/numeric.py
@@ -2663,7 +2663,7 @@ def astype(x, dtype, /, *, copy=True, device=None):
     True
 
     """
-    if not (isinstance(x, np.ndarray) or isscalar(x)):
+    if not isinstance(x, (np.ndarray, np.generic)):
         raise TypeError(
             "Input should be a NumPy array or scalar. "
             f"It is a {type(x)} instead."
--- a/numpy/_core/numeric.py
+++ b/numpy/_core/numeric.py
@@ -2663,9 +2663,9 @@ def astype(x, dtype, /, *, copy=True, device=None):
     True
 
     """
-    if not isinstance(x, (np.ndarray, np.generic)):
+    if not (isinstance(x, np.ndarray) or isscalar(x)):
         raise TypeError(
-            "Input should be a NumPy array or a NumPy scalar. "
+            "Input should be a NumPy array or scalar. "
             f"It is a {type(x)} instead."
         )
     if device is not None and device != "cpu":
@@ -2673,7 +2673,14 @@ def astype(x, dtype, /, *, copy=True, device=None):
             'Device not understood. Only "cpu" is allowed, but received:'
             f' {device}'
         )
-    return x.astype(dtype, copy=copy)
+    try:
+        return x.astype(dtype, copy=copy)
+    except AttributeError:
+        raise TypeError(
+            "Input should be a NumPy array or scalar."
+            f"It is a {type(x)} instead. Python scalars are not supported. "
+        )
+
 
 
 inf = PINF
--- a/numpy/_core/numeric.py
+++ b/numpy/_core/numeric.py
@@ -2677,7 +2677,7 @@ def astype(x, dtype, /, *, copy=True, device=None):
         return x.astype(dtype, copy=copy)
     except AttributeError:
         raise TypeError(
-            "Input should be a NumPy array or scalar."
+            "Input should be a NumPy array, scalar, or other object with an `astype` implementation."
             f"It is a {type(x)} instead. Python scalars are not supported. "
         )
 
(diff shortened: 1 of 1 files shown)


ngoldbaum:
I don’t think changing the astype implementation is correct, instead we should do a try/except (or the equivalent in C if the the getattr is happening in C) and raise the appropriate error at the site where it happens.

Generally it’s best to open issues about bugs you find before sending in a PR with a fix, particularly for projects you’re new to the internals of.

mncrftfrcnm:
> I don’t think changing the astype implementation is correct, instead we should do a try/except (or the equivalent in C if the the getattr is happening in C) and raise the appropriate error at the site where it happens.
> 
> Generally it’s best to open issues about bugs you find before sending in a PR with a fix, particularly for projects you’re new to the internals of.

So, should i do smt like this:
```text
    if not (isinstance(x, np.ndarray) or isscalar(x)):
        raise TypeError(
            "Input should be a NumPy array or scalar. "
            f"It is a {type(x)} instead."
        )
    if device is not None and device != "cpu":
        raise ValueError(
            'Device not understood. Only "cpu" is allowed, but received:'
            f' {device}'
        )
    try:
        return x.astype(dtype, copy=copy)
    except AttributeError:
        raise TypeError(
            "Input should be a NumPy array or scalar."
            f"It is a {type(x)} instead. Python scalars are not supported. "
        )

```

mncrftfrcnm:
> I don’t think changing the astype implementation is correct, instead we should do a try/except (or the equivalent in C if the the getattr is happening in C) and raise the appropriate error at the site where it happens.
> 
> Generally it’s best to open issues about bugs you find before sending in a PR with a fix, particularly for projects you’re new to the internals of.

Also, if I did everything correctly, it will only omit the python native scalars, returning proper error, which is what the previous reviewer asked for in the other PR, where I implemented python-native scalar handling(here's link to that PR: https://github.com/numpy/numpy/pull/31365 )

mncrftfrcnm:
> I don’t think changing the astype implementation is correct, instead we should do a try/except (or the equivalent in C if the the getattr is happening in C) and raise the appropriate error at the site where it happens.
> 
> Generally it’s best to open issues about bugs you find before sending in a PR with a fix, particularly for projects you’re new to the internals of.

ok, i've changed the way it handles attribute error to try_except

mncrftfrcnm:
@ngoldbaum

numpy/_core/numeric.py
```diff
@@ -2673,7 +2673,14 @@ def astype(x, dtype, /, *, copy=True, device=None):
             'Device not understood. Only "cpu" is allowed, but received:'
             f' {device}'
         )
-    return x.astype(dtype, copy=copy)
+    try:
+        return x.astype(dtype, copy=copy)
+    except AttributeError:
+        raise TypeError(
+            "Input should be a NumPy array or scalar."
```
ngoldbaum:
```suggestion
            "Input should be a NumPy array, scalar, or other object with an `astype` implementation."
```

numpy/_core/numeric.py
```diff
@@ -2673,7 +2673,14 @@ def astype(x, dtype, /, *, copy=True, device=None):
...
+            f"It is a {type(x)} instead. Python scalars are not supported. "
```
ngoldbaum:
```suggestion
            f"It is a {type(x)} instead."
```

IMO "python scalars" is confusing. There's also no guarantee that the operand is e.g. a PyInt, so there's a chance it's not even a helpful suggestion.

ngoldbaum (commented):
I think this needs a release note maybe because someone might be relying on this function raising `AttributeError`. I'm honestly a little on-the-fence about changing it at all, since it should be pretty obvious what the issue is even with the current error message and `AttributeError`.

ngoldbaum:
@seberg what do you think about this change? does it need a release note if we do change it?

seberg:
Seems OK to me without a release note, in practice nobody should rely on it.  But yeah, either way seems fine.
Honestly, I am not sure if I think this is very useful, and if we add a typecheck effectively, should we just check for `isinstance(obj, (np.ndarray, np.generic))`?!

But I don't mind it.

mncrftfrcnm:
> @seberg what do you think about this change? does it need a release note if we do change it?


so, what should i do?

mncrftfrcnm:
> @seberg what do you think about this change? does it need a release note if we do change it?

so, what should I do?

mncrftfrcnm:
@ngoldbaum what should I do?

mncrftfrcnm:
@seberg @ngoldbaum what should I change?

PR #31371, opened 2026-05-01
BUG: Avoid optimization causing astropy regression

### LINKED ISSUE #31363: BUG: crash (segfault) bisected to gh-29270
### Describe the issue:

seen on astropy's test suite, failures started a bit over 24h ago, which I believe corresponds to a nightly release being dropped.
I could reproduce this crash seemingly deterministically on a linux machine (but couldn't reprod on macos), and bisected the crash to [0f79a3a](https://github.com/numpy/numpy/commit/0f79a3a0921873c3d7a85c7b3d7cf1411b7d2143) (which is the merge commit for gh-29270)

(I'm opening from a different computer than the one I can reproduce from, I'll complete the report as an edit)

xref: https://github.com/astropy/astropy/issues/19621

```shell
pytest astropy/uncertainty
```

### Error message:

```shell
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7ca9adf in unlink_chunk (p=p@entry=0x39e9ee0, av=0x7ffff7e03ac0 <main_arena>) at ./malloc/malloc.c:1610
warning: 1610	./malloc/malloc.c: No such file or directory
```

Python 3.14.3
numpy main (as of 0f79a3a0921873c3d7a85c7b3d7cf1411b7d2143)

### Runtime Environment:

```
[{'numpy_version': '2.5.0.dev0+git20260428.0f79a3a',
  'python': '3.14.3 (main, Feb  3 2026, 22:52:18) [Clang 21.1.4 ]',
  'uname': uname_result(system='Linux', node='beelzebot', release='6.17.0-22-generic', version='#22~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Mar 26 15:25:54 UTC 2', machine='x86_64')},
                      'found': ['X86_V3'],
                      'not_found': ['X86_V4', 'AVX512_ICL', 'AVX512_SPR']}},
None
```

_No response_

seberg:
Arguably, I think astropy is relying on things it shouldn't rely on, but let's not break them.
I am considering adding a dict option here to enable the optimization on this path (i.e. allow setting the flag explicitly unless a subfield disables it).

But if astropy wants this in quickly to unblock CI, this should be fine.

Closes gh-31363

### DIFF
--- a/numpy/_core/tests/test_dtype.py
+++ b/numpy/_core/tests/test_dtype.py
@@ -628,6 +628,21 @@ class TestRecord:
         assert arr.dtype.hasobject  # but claims to contain objects
         del arr  # the deletion failed previously.
 
+    def test_structured_out_of_range(self):
+        # This test should be able to fail as an optimization, but astropy
+        # relied on the non-optimized behavior.
+        # A future version could force astropy to either fix the code or
+        # introduce a kwarg to disable the optimization default. But for now
+        # we avoid the astropy regression.
+        # The following dtype cannot be assumed to just include padding
+        dtype = np.dtype(dict(names=["a"], formats=["i4"], itemsize=8))
+        raw_arr = np.zeros(10, dtype="i8")
+        arr1 = raw_arr.view(dtype)
+        arr2 = np.full(10, -1, dtype="i8").view(dtype)
+        # This should only fill in the i4 field of `dtype`:
+        arr1[...] = arr2
+        assert (raw_arr.view("i4")[1::2] == 0).all()
+
 
 class TestSubarray:
     def test_single_subarray(self):
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -1315,6 +1315,13 @@ _convert_from_dict(PyObject *obj, int align)
             Py_DECREF(new);
             goto fail;
         }
+        if (itemsize != new->elsize) {
+            /*
+             * astropy relied on "holes" at the end not being used until this
+             * is fixed we cannot optimize the path as a full dtype copy.
+             */
+            new->flags |= NPY_NOT_TRIVIALLY_COPYABLE;
+        }
         /* Set the itemsize */
         new->elsize = itemsize;
     }


charris:
close/reopen

charris:
Thanks Sebastian. The debug test seems unrelated.

numpy/_core/tests/test_dtype.py
```diff
@@ -628,6 +628,21 @@ def test_nonstructured_with_object(self):
...
+        # relied on the non-optimized behavior.
+        # A future version could force astropy to either fix the code or
+        # introduce a kwarg to disable the optimization default. But for now
+        # we avoid the astropy regression.
+        # The following dtype cannot be assumed to just include padding
+        dtype = np.dtype(dict(names=["a"], formats=["i4"], itemsize=8))
+        raw_arr = np.zeros(10, dtype="i8")
+        arr1 = raw_arr.view(dtype)
+        arr2 = np.full(10, -1, dtype="i8").view(dtype)
+        # This should only fill in the i4 field of `dtype`:
```
mhvk, line 642: Thanks for putting this test in. I'm not quite sure it is the same as what we do in `uncertainty` - will need to think! And in a way odd, because if the `dtype` were not identical, I think there is an argument to be made to disallow the optimization of also copying the extra 4 bytes, as there is no way to know what those entail. 

But in the end, my use in astropy of a dtype as a way to effectively get different strides was perhaps not great... Unfortunately, it is pretty baked in, but I'll see if there is a way to make the dtype either fail `is_dtype_struct_simple_unaligned_layout` or make the copying safe for the cases that it passes.

mhvk: Interestingly, this particularly case would work if in the dtype definition one passed in `offsets=[0]`. For the astropy case, this also seems to work -- it effectively signals that one should do the alignment check and that one fails.

So, possibly that is another solution here? To run the alignment check even if `offsets=NULL`?

(Sorry, but too tired to think through now; maybe more tomorrow.)

seberg: Not sure I am quite following, I don't think astropy uses offsets (but passing offsets can be common, I think even for cases where it isn't strictly needed maybe).

The test shows of course that if you use `arr.view(dtype)` then in a sense it depends on what the previous dtype was whether or not the view is actually safe w.r.t. to optimized assignment...
In that sense, the correct way might be to disable the flag on `view` (in theory unless we know no new "holes" are created).

Hmmmpf, tricky. For new arrays it is all straight forward, for views it seems a bit like it's maybe all tricky?

PR #31376, opened 2026-05-03
MAINT: clarify why a structured dtype might not be trivially copyable

mhvk:
A follow-up on gh-31371, partially to have a place to discuss that is not a closed issue or PR: after thinking a bit more, I think the solution in gh-31371 is actually the right one: if you pass in `itemsize` and that leads to holes, then just like for `offsets`, one should assume those holes are on purpose and thus  the whole item cannot be safely copied. 

This PR is in partially to discuss; all it really does is change comments to make the logic explicit, and bring the tests for non-trivially-copyable to one place--there is no functional change.

EDIT: Note that even with this limitation, we have solved the original issue, which was about slowdowns for a regular structured array with no offsets or itemsize provided.

No AI used.

### DIFF
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -1315,24 +1315,19 @@ _convert_from_dict(PyObject *obj, int align)
             Py_DECREF(new);
             goto fail;
         }
-        if (itemsize != new->elsize) {
-            /*
-             * astropy relied on "holes" at the end not being used until this
-             * is fixed we cannot optimize the path as a full dtype copy.
-             */
-            new->flags |= NPY_NOT_TRIVIALLY_COPYABLE;
-        }
         /* Set the itemsize */
         new->elsize = itemsize;
     }
 
     /*
-     * Mark as not trivially copyable only if explicit offsets were provided
-     * and the layout has holes. Alignment padding (without explicit offsets)
-     * is safe to overwrite with memcpy.
+     * Check if anything prevents using memcpy for whole items for this dtype,
+     * i.e., whether there are any holes unrelated to alignment padding
+     * (since those holes might be used to avoid accessing/overwriting stuff).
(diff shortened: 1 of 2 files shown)


seberg:
There is a real use-case to use `itemsize=` for custom alignment (which is where I came from), similar to `align=True` (but you might want more).
Either way, there is probably no point in trying to push for it being OK and I just need to think about a `"trivially_copyable": True` or so (in this dict path)...

mhvk:
In principle, since `align=False` by default, `align=True` could have more implications than just C alignment. But would be good to have a concrete example. If it is similarly rare to astropy's usage, and one really wanted fast copy, one could just make the structure properly opague with `.view(f"V{dtype.itemsize}")`. Or one could explicitly fill the structure with padding.

seberg:
> could have more implications than just C alignment. But would be good to have a concrete example.

This comes from accelerators mostly (in principle also SIMD).  In some cases you want to align struct members to more than their itemsize. A main example is complex numbers which e.g. GPUs align to the full itemsize to optimize copies (something we do also with that weird uintalignment/alignment split...).

I.e. a bit like `align=16` but I am not really convinced that actually modifying the stored alignment is good, since there are two concepts of alignments: The one typical NumPy/CPU loops need at least, and the one users want especially when nesting structs...

Anyway, PR looks good, let's put it in, thanks!

mhvk:
OK, makes sense. But at least the original slowdown is now addressed! Also, not clear to me that those special-cased alignments should live on general dtypes; as you noted elsewhere, a view may have different requirements than a newly created array... Anyway, we can deal with that when the need arises...

seberg:
Yeah, views are very tricky... In principle, I guess it would make sense that mark a dtype as "not trivially copyable" when creating a new view, although that means creating a new dtype...
That might actually work quite generally in practice, but I think one can also move to such scheme in some future.

PR #31378, opened 2026-05-03
ENH: raise OverflowError on datetime64/timedelta64 arithmetic overflow

jbrockmendel:
Claude wrote this, I reviewed it manually.  cc @seberg


numpy/_core/tests/test_datetime.py
```diff
@@ -1093,6 +1093,101 @@ def test_cast_overflow_safe_unit_conversion(self):
...
+        assert ok_dt == np.datetime64(big, "s")
+        ok_td = np.timedelta64(big - 1, "s") + np.timedelta64(1, "s")
+        assert ok_td == np.timedelta64(big, "s")
+
+        # Array-shaped inputs (exercise the actual ufunc loop, not just scalar).
+        arr = np.array([0, big - 1, -big + 1], dtype="timedelta64[s]")
+        with pytest.raises(OverflowError, match="Overflow"):
+            arr + np.timedelta64(2, "s")
+        with pytest.raises(OverflowError, match="Overflow"):
+            arr - np.timedelta64(2, "s")
```
tylerjereddy: Maybe it is a nit, but this test is quite long, and doing a lot of things. It may make sense to wait for feedback from NumPy devs on the C code first, but my intuition is that it might be nice to split this test into separate tests a bit.

For example, typing out `with pytest.raises(OverflowError, match="Overflow")` 10 separate times for various permutations of binary operations might be split out to a parametrized test with i.e., (`x`, `y`, `operator`).

numpy/_core/src/umath/loops.c.src
```diff
@@ -908,7 +958,14 @@ TIMEDELTA_mq_m_multiply(char **args, npy_intp const *dimensions, npy_intp const
             *((npy_timedelta *)op1) = NPY_DATETIME_NAT;
         }
         else {
-            *((npy_timedelta *)op1) = in1 * in2;
+            char overflow = 0;
+            const npy_int64 result = safe_mul(in1, in2, &overflow);
```
ngoldbaum, line 964: You didn't create this problem, but I notice reading over this that the `safe_mul` implementation is susceptible to UB via signed int overflow, since it returns `a*b` unconditionally. It should only ever do the multiplication operation if it knows that it won't overflow, otherwise UB happens. The discarded result is probably fine but best to not use "safe" primitives that have UB.

There is a builtin overflow-checking signed multiplication function provided by most compilers we can use instead, but we maybe need to keep the fallbacks anyway? https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html

numpy/_core/src/umath/loops.c.src
```diff
@@ -817,7 +825,14 @@ DATETIME_Mm_M_add(char **args, npy_intp const *dimensions, npy_intp const *steps
             *((npy_datetime *)op1) = NPY_DATETIME_NAT;
         }
         else {
-            *((npy_datetime *)op1) = in1 + in2;
+            char overflow = 0;
+            const npy_int64 result = safe_add(in1, in2, &overflow);
+            if (overflow || result == NPY_DATETIME_NAT) {
+                npy_gil_error(PyExc_OverflowError,
+                        "Overflow in datetime64 addition");
```
ngoldbaum: minor nit: here and elsewhere where you generate overflow errors, it would be nicer to include the type information context you have, so here you could say "Overflow in datetime64 + timedelta64 addition".

numpy/_core/tests/test_datetime.py
```diff
@@ -1093,6 +1093,116 @@ def test_cast_overflow_safe_unit_conversion(self):
...
+        big = np.iinfo(np.int64).max
+
+        ok_dt = np.datetime64(big - 1, "s") + np.timedelta64(1, "s")
+        assert ok_dt == np.datetime64(big, "s")
+        ok_td = np.timedelta64(big - 1, "s") + np.timedelta64(1, "s")
+        assert ok_td == np.timedelta64(big, "s")
+
+        small = np.timedelta64(3, "s")
+        assert small * np.int64(7) == np.timedelta64(21, "s")
+        assert np.int64(7) * small == np.timedelta64(21, "s")
```
ngoldbaum, line 1266: Can you generate a C coverage report and scrutinize the C code you changed in this PR to make sure all code paths are covered, including the new error paths? You can generate a C coverage report with `spin test --gcov --gcov-format=html` if you have gcov installed. You can probably ask Claude to help you run the coverage tests and fill in coverage gaps. I asked for a review of this code and got the following comments, I didn't actually verify any of the claims from Claude:

<details>

1. safe_add overflow that does not land on NaT — none of the current add tests isolate this branch. Add e.g.:
  # INT64_MAX + 2 wraps to INT64_MIN + 1 (valid timedelta), so only safe_add can catch it
  with pytest.raises(OverflowError, match="Overflow"):
      np.timedelta64(big, "s") + np.timedelta64(2, "s")
  1. (Note: np.datetime64(big, "s") - dt_neg already isolates safe_sub — that test wraps to -3, not NaT — so subtract is fine.)
  2. Same for safe_mul — every multiply overflow test lands on NaT. Add:
  with pytest.raises(OverflowError, match="Overflow"):
      np.timedelta64(big, "s") * np.int64(2)
  3. Negative-side overflow in add (a < 0 && b < NPY_MIN_INT64 - a branch of safe_add):
  with pytest.raises(OverflowError, match="Overflow"):
      np.timedelta64(-big, "s") + np.timedelta64(-2, "s")
  4. Negative multiplier and INT64_MIN multiplier:
  with pytest.raises(OverflowError, match="Overflow"):
      np.timedelta64(big, "s") * np.int64(-2)
  with pytest.raises(OverflowError, match="Overflow"):
      np.timedelta64(2, "s") * np.int64(np.iinfo(np.int64).min)
  5. NaT short-circuit beats overflow check — a regression guard:
  assert np.isnat(np.datetime64('NaT', 's') + np.timedelta64(big, 's'))
  6. Result-equals-NaT in the array path. Scalar coverage is good, but the strided loop should also raise, not wrap:
  arr = np.array([0, -big], dtype="timedelta64[s]")
  with pytest.raises(OverflowError, match="Overflow"):
      arr + np.timedelta64(-1, "s")
  7. Asymmetric array coverage. Array tests currently exercise only mm_m_add, mm_m_subtract, mq_m_multiply. Five of the eight loops have no
  array-path coverage:
    - DATETIME_Mm_M_add (arr_M + arr_m)
    - DATETIME_mM_M_add (arr_m + arr_M)
    - DATETIME_Mm_M_subtract (arr_M - arr_m)
    - DATETIME_MM_m_subtract (arr_M - arr_M)
    - TIMEDELTA_qm_m_multiply (int_arr * arr_m)

  The comment claims the array case "exercises the actual strided ufunc loop," but that's only true for the loops that are exercised. At least one
  array test per loop would close the gap.

  Nothing here is a correctness blocker — the implementation is solid. The added-test items above mostly tighten the safety net rather than expose
  bugs I can see by inspection.

</details>

numpy/_core/src/umath/loops.c.src
```diff
@@ -807,6 +808,13 @@ NPY_NO_EXPORT void
 
 /**end repeat**/
 
+/*
+ * Overflow-checked arithmetic for datetime64/timedelta64.
+ *
+ * NPY_DATETIME_NAT == NPY_MIN_INT64.  A valid arithmetic result that
+ * happens to equal NPY_MIN_INT64 would be silently misinterpreted as
+ * NaT, so we treat it as overflow as well.
+ */
```
ngoldbaum, line 819: This comment applies equally to all the other loops you modified. Consider writing it a little more generically or add some test so the reader knows it doesn't just apply to the next loop

ngoldbaum (commented):
Overall looks good. A few comments inline.

seberg:
Sorry, a fly-by question (the answer is probably just "nah").  Have you considered whether we should just return `NaT` and set an "overflow" flag (if possible, which is most hardware) rather than creating a hard error?

jbrockmendel:
> Sorry, a fly-by question (the answer is probably just "nah"). Have you considered whether we should just return NaT and set an "overflow" flag (if possible, which is most hardware) rather than creating a hard error?

Short answer: nah.  Longer answer: i could imagine this if either a) we had a second sentinel analogous to "inf" or b) it was behind a np.errstate flag.

seberg:
> or b) it was behind a np.errstate flag.

OK, just wanted to bring it up in case you/others had thoughts. We can do that, but unless we hook it into the existing floats it doesn't matter (not great).  So if the default would be an error (which seems fine to me, under the assumption that is a really weird niche thing, though) then this is good.  (We _can_ do that, but it's involved.)

jbrockmendel:
FWIW raising on overflow has been the pandas behavior for a while and we've never gotten any complaints about it.

numpy/_core/src/umath/loops.c.src
ngoldbaum: I created https://github.com/numpy/numpy/pull/31396 to address this.

ngoldbaum (approved):
LGTM!

seberg (approved):
Since nobody voiced concerns about raising overflows more, let's get this in, thanks.

(I wonder slightly if we should make the NaT handling post-processing now to avoid the NaT checks on the fast-path, but only a potential follow up.)

Issue #24280, opened 2023-07-28
Cirrus CI free usage going away - job runtime issues & credits


rgommers:
This is getting kinda annoying: `[skip cirrus]` is broken and the wheel build and other CI jobs are running way too often. It was probably broken by the addition of other logic like always triggering on PRs with build-related label.

charris:
What do you want to happen with the build label? I use it when I want to test wheel builds. It is a bit annoying, as simply adding a label will retrigger the builds.

rgommers:
Labels are for indicating what a PR or issue is about, so I add it to build-related PRs. And running wheel builds by default is fine, but the `[skip cirrus]` should supercede that (just like it does for `[skip actions]` on GHA).

charris:
> but the [skip cirrus] should supercede

Fair enough, but some labels do trigger wheel builds, so they aren't only about what the PR or issue is about. Not sure how we could change that unless we want to add a `[build wheels]` option.

rgommers:
Why `[skip cirrus]` doesn't work is explained by https://github.com/numpy/numpy/blob/7fc72776b972bfbfdb909e4b15feb0308cf8adba/.cirrus.star#L23-L25

It looks like it's not the build label though, we run wheel builds on Cirrus always. E.g. look at this doc-only PR with no labels: gh-24277. This wastes a huge amount of compute time.

charris:
Cirrus is a bit different. One problem is that it cannot be manually run, I think it needs something in the *.yml file for that. I also think (only) wheels are built and tested, I'm not sure how a build can be made without that. It is on my problem list, but not annoying enough to spend time on it. Maybe @andyfaff has some thoughts.

andyfaff:
.cirrus.star is triggered to run with a lot of GH events on the numpy/numpy repo, e.g. if there are commits to PRs, commits to branches, PRs are opened, labels are attached to PRs, tags are pushed, merges, etc.

Here is the current logic for numpy's cirrus CI, as contained in .cirrus.star:

- only run jobs only on the `numpy/numpy` repo.
- (do a wheels build if a 'nightly' cron job is requested)
- for all other triggers (and there can be a lot of them) the configuration script looks at the commit message of the SHA, provided by CIRRUS_CHANGE_IN_REPO, for the event. If there is `[skip cirrus]` or `[skip ci]` in the message then don't run any CI.
- otherwise do the wheel build and the macosx_arm64 runs.

[...]

PR #31380, opened 2026-05-04
CI: remove Cirrus CI FreeBSD job

rgommers:
Cirrus CI got acquired and will cease operations on June 1st. There is no good alternative for FreeBSD, and this job isn't critical. So, we go back to the status quo from before, with no direct CI coverage.

This was also commented on at https://github.com/numpy/numpy/issues/24280#issuecomment-4363264545.

#### AI Disclosure

No AI usage.


DimitriPapadopoulos:
For the sake of completeness, [VM Actions](https://github.com/vmactions) provide the GitHub action [freebsd-vm](https://github.com/vmactions/freebsd-vm) to set up a FreeBSD virtual machine based on QEMU. It could simplify maitenance, but I don't have prior experience with it.

DimitriPapadopoulos:
In addition to removing `.cirrus.star`, an administrator will have to revoke Cirrus CI integration:
* Settings → Integrations → GitHub Apps → Cirrus CI
* Settings → Applications → Authorized OAuth Apps → Cirrus CI

rgommers:
> For the sake of completeness, [VM Actions](https://github.com/vmactions) provide the GitHub action [freebsd-vm](https://github.com/vmactions/freebsd-vm) to set up a FreeBSD virtual machine based on QEMU.

Thanks. And yes, I know - I investigated all the options, and there are no good/equivalent ones. It's not worth it to run under QEMU, way too slow.

charris:
Thanks Ralf.

charris:
I've removed Cirrus from GitHub Apps, but I don't see Authorized OAuth Apps anywhere.

DimitriPapadopoulos:
I think Authorized OAuth Apps are organisation-level settings only. I obviously don't have access to Numpy to verify. I just wanted to point out that Cirrus CI access should be revoked/uninstalled under organisation settings and repository settings.

rgommers:
That's all gone, we're good here. Thanks both!

PR #31388, opened 2026-05-05
CI: newer MKL uses so.3

### LINKED ISSUE #31386: MKL 2026.0.0 does not get detected
We do not pin mkl-dev in the [BLAS + MKL](https://github.com/numpy/numpy/blob/27b26e9e3f95e8af3611891ccd0ff53f3169cde2/.github/workflows/linux_blas.yml#L300) runs. They started failing when `pip install` started picking up `mkl-2026.0.0` instead of [mkl-2025.3.1](https://github.com/numpy/numpy/actions/runs/24842653061/job/72720771104#step:4:113). Is this a known problem with meson and mkl-2026?

mattip:
Fixes #31386

Newer MKL package uses so.3 suffix, symlink still required.

Tested locally. No AI was used.

### DIFF
--- a/.github/workflows/linux_blas.yml
+++ b/.github/workflows/linux_blas.yml
@@ -313,6 +313,7 @@ jobs:
         # add the expected .so -> .so.2 symlinks to fix linking
         cd ..
         for i in $( ls libmkl*.so.2 ); do ln -s $i ${i%.*}; done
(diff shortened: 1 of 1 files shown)


rgommers (approved):
Thanks @mattip! LGTM and fixes a failing CI job, so in it goes.

PR #31394, opened 2026-05-06
Add Geometric Mean

siaulich:
### PR summary
This PR implements the [geometric mean](https://en.wikipedia.org/wiki/Geometric_mean). The function is so far only added as plain function. If interest exists, I can add `numpy.nangmean`.

The method is added to `numpy.__all__`, documented in the API-reference and covered by tests.

Hi, I am Siemen Aulich. Currently finishing my master's in physics with a focus of machine-learning for high-energy physics. During an analysis of a machine-learning model, I wanted to use a geometric mean. I realised, that this function does not exist in `numpy`. 

#### AI Disclosure
AI-tools (GitHub copilot) was used to create the outline for the test. The numerics for the test, however, where computed using `scipy`

### DIFF
--- a/doc/source/reference/routines.statistics.rst
+++ b/doc/source/reference/routines.statistics.rst
@@ -22,6 +22,7 @@ Averages and variances
 .. autosummary::
    :toctree: generated/
 
+   gmean
    median
    average
    mean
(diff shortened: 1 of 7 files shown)


jorenham:
New feature proposals should be discussed in the mailing list first.

PR #31396, opened 2026-05-06
BUG: Avoid UB in safe_[add,sub,mul] helpers

ngoldbaum:
Over in https://github.com/numpy/numpy/pull/31378/changes#r3190759945, I noticed that these helpers can trigger UB, since the result is always computed even if overflow might happen. A sufficiently aggressive compiler might optimize functionality away if it detects the undefined behavior.

Clang and gcc both have intrinsics to handle this, which we use if available. I had to add new probes for the add and sub builtin intrinsics. For the fallback cases, I added early returns to avoid triggering overflow.

I used Claude Opus 4.7 to help with this.


charris (approved):

charris:
Thanks Nathan. I marked this for backport as it seems straight forward.

PR #31399, opened 2026-05-08
TYP: `_NestedSequence` type parameter default to work around a mypy issue

jorenham:
### PR summary

This works around false positive `var-annotated`  mypy errors in certain situations, that's currently blocking a fix for the typeshed stdlib stubs.

x-ref: https://github.com/python/typeshed/pull/15472#issuecomment-4406093095

#### AI Disclosure
N/A

PR #31404, opened 2026-05-08
TYP: `_NestedSequence` type parameter default to work around a mypy issue (#31399)

charris:
Backport of #31399.

### PR summary

This works around false positive `var-annotated`  mypy errors in certain situations, that's currently blocking a fix for the typeshed stdlib stubs.

x-ref: https://github.com/python/typeshed/pull/15472#issuecomment-4406093095

#### AI Disclosure
N/A

### DIFF
--- a/numpy/_typing/_nested_sequence.py
+++ b/numpy/_typing/_nested_sequence.py
@@ -1,13 +1,19 @@
 """A module containing the `_NestedSequence` protocol."""
 
-from typing import TYPE_CHECKING, Any, Protocol, TypeVar, runtime_checkable
+from typing import TYPE_CHECKING, Any, Protocol, runtime_checkable
 
 if TYPE_CHECKING:
     from collections.abc import Iterator
+    from typing_extensions import TypeVar
+
+    _T_co = TypeVar("_T_co", covariant=True, default=Any)
+else:
+    from typing import TypeVar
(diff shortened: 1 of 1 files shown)


numpy/_typing/_nested_sequence.py
```diff
@@ -4,6 +4,14 @@
 
 if TYPE_CHECKING:
     from collections.abc import Iterator
+    from typing_extensions import TypeVar
+
+    _T_co = TypeVar("_T_co", covariant=True, default=Any)
+else:
+    from typing import TypeVar
+
+    _T_co = TypeVar("_T_co", covariant=True)
```
jorenham, line 13: There's another `_T_co` on line 18 that's now shadowing this one

numpy/_typing/_nested_sequence.py
```diff
@@ -4,6 +4,14 @@
 
 if TYPE_CHECKING:
     from collections.abc import Iterator
+    from typing_extensions import TypeVar
+
+    _T_co = TypeVar("_T_co", covariant=True, default=Any)
+else:
+    from typing import TypeVar
```
jorenham, line 11: I'm not sure if it matters, but the `TypeVar` import at line 3 is now no longer needed

PR #31412, opened 2026-05-12
BUG: `np.ones` does not raise ValueError for generic datetime64

riku-sakamoto:
Closes gh-30903
(Follow up attempt to #31319)

### PR summary

This PR sets the descriptor for the Python scalar `1` in `array_copyto` when the destination dtype is `datetime64`.

With this change, `np.ones(..., dtype="M8")` raises an error immediately, because the fill value `1` is cast to a `datetime64` scalar with a generic unit.

#### Root cause

When `np.ones` is called with `dtype="M8"`, the Python integer `1` is eventually passed to the C cast function below. Since the C representation of `npy_datetime` is the same as `npy_int64`, this cast currently succeeds without checking whether the datetime unit is valid.

As a result, the invalid value is only detected later, for example when the array is passed to another function that checks the validity of the datetime unit, such as when the array is printed.

https://github.com/numpy/numpy/blob/c36307a4241a3cee9a806044787e6ebd3e7b2c03/numpy/_core/src/multiarray/arraytypes.c.src#L1300-L1345

#### Alternative

Another alternative is to create new casting rule for `datetime64`. (see below). But, this seems a bit too much for this issue, so I chose the current one. If preferred, I can implement the alternative one.

https://github.com/numpy/numpy/blob/0adb7ef2609525f0c3ab91fe2914cc700d4311cc/numpy/_core/src/multiarray/datetime.c#L4304-L4324

#### AI Disclosure

ChatGPT is used to polish the description of this PR, but I have not used AI to write the code for this PR.

### DIFF
--- a/numpy/_core/tests/test_nditer.py
+++ b/numpy/_core/tests/test_nditer.py
@@ -1531,10 +1531,12 @@ def test_iter_copy():
 @pytest.mark.parametrize("loop_dtype", np.typecodes["All"])
 @pytest.mark.filterwarnings(
     "ignore::numpy.exceptions.ComplexWarning",
-    "ignore::DeprecationWarning",
 )
 def test_iter_copy_casts(dtype, loop_dtype):
     # Ensure the dtype is never flexible:
+    if dtype.lower() == "m":
+        dtype = dtype + "8[D]"
+
     if loop_dtype.lower() == "m":
         loop_dtype = loop_dtype + "[ms]"
     elif np.dtype(loop_dtype).itemsize == 0:
--- a/numpy/_core/tests/test_nditer.py
+++ b/numpy/_core/tests/test_nditer.py
@@ -1533,12 +1533,15 @@ def test_iter_copy():
     "ignore::numpy.exceptions.ComplexWarning",
 )
 def test_iter_copy_casts(dtype, loop_dtype):
-    # Ensure the dtype is never flexible:
     if dtype.lower() == "m":
         dtype = dtype + "8[D]"
 
+    is_datetimelike = False
     if loop_dtype.lower() == "m":
-        loop_dtype = loop_dtype + "[ms]"
+        loop_dtype = loop_dtype + "8[ms]"
+        is_datetimelike = True
+
+    # Ensure the dtype is never flexible:
     elif np.dtype(loop_dtype).itemsize == 0:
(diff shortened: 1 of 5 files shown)


ngoldbaum:
Sorry for dropping this. I'm a little confused by the tests, since you're not testing the new behavior for `np.ones`.

riku-sakamoto:
@ngoldbaum 
Thank you for pointing it out!
Sorry, my bad. I've added a test to check that `np.ones` fails when using generic datetime.

numpy/_core/tests/test_nditer.py
```diff
@@ -1531,10 +1531,12 @@ def test_iter_copy():
...
     "ignore::numpy.exceptions.ComplexWarning",
-    "ignore::DeprecationWarning",
 )
 def test_iter_copy_casts(dtype, loop_dtype):
     # Ensure the dtype is never flexible:
+    if dtype.lower() == "m":
+        dtype = dtype + "8[D]"
+
     if loop_dtype.lower() == "m":
         loop_dtype = loop_dtype + "[ms]"
```
seberg:
```suggestion
        loop_dtype = loop_dtype + "8[ms]"
```
Its caught too broadly below, so we are not noticing here that it is incorrect...

seberg (approved):
Thanks, I think we can go with this.

For downstream impact, I guess this should only really be relevant for test-suites and at least larger ones would already have seen the new deprecation warning.

It may be nice to add a test for `copyto()` as well, since that is where you are actually changing things (which seems fine)?

numpy/_core/tests/test_nditer.py
ngoldbaum: I just pushed a commit that implements this and fixes some fallout that causes an assert below to fail. I also changed the bare `return` below to use a `pytest.xfail` to make it clearer that we're intentionally skipping the test.

ngoldbaum: Hope you don't mind the assist to land this @riku-sakamoto!

riku-sakamoto: No problem! Thanks for the assist.

seberg:
Thanks for the follow-ups, let's get this in then!

riku-sakamoto:
Thank you for the review! @seberg @ngoldbaum

PR #31413, opened 2026-05-12
DOC: Add forgotten ``register_dlpack_dtype`` to docs

seberg:
Forgot to add this, and I think there was a brief universal CI failure so missed it (although... maybe I just _thought_ that also affected circle CI and it didn't, don't recall)

### DIFF
--- a/doc/source/reference/routines.dtypes.rst
+++ b/doc/source/reference/routines.dtypes.rst
@@ -71,3 +71,11 @@ Others
 .. attribute:: ObjectDType
                VoidDType
 
+
+Routines for DType authors
+--------------------------
+
+.. autosummary::
+   :toctree: generated/
+
+   register_dlpack_dtype


mattip:
LGTM. The docstring says 
> While an error is raised, the export is registered even on error.

So it succeeds but errors? I think that is a bit unusual, no?

mattip:
Thanks @seberg

seberg:
> So it succeeds but errors? I think that is a bit unusual, no?

Yeah, it happens only in the case where you have two dtypes that want to register for the same DLPack dtype, though.  In that case `dtype -> dlpack` succeeds, but the `dlpack -> dtype` part cannot succeed (it would change what someone else registered), so only the second part fails.
(I did this, because I thought that was OK, maybe even nice, it does make sense to allow this half registration.  It also just simplified thread-safety, TBH.)

PR #31416, opened 2026-05-12
MAINT: ensure `_core.sctypes` lists are deterministically ordered

### LINKED ISSUE #31414: BUG: pytest-xdist collection failure in test_constructor_signatures due to non-deterministic sctypes set ordering
Describe the issue:

`TestSignature::test_constructor_signatures` is parametrized with `*sctypes["others"]`, but `sctypes["others"]` is a `set`. Sets have
non-deterministic iteration order, so xdist workers may collect `bytes_` and `str_` in different sequences and abort before any test runs.

The failure is intermittent: it depends on how xdist distributes work across workers. I see it consistently on machines with 32+ cores (riscv64 RISE runners).

Reproduce the code example:

```python
python -c "import sys; import numpy; sys.exit(not numpy.test(label='full', extra_argv=['-n=auto']))"
```

Error message:

```
ERROR gw0 - Different tests were collected between gw3 and gw0. The difference is:
 _core/tests/test_scalar_methods.py::TestSignature::test_constructor_signatures[clongdouble]
-_core/tests/test_scalar_methods.py::TestSignature::test_constructor_signatures[bytes_]
 _core/tests/test_scalar_methods.py::TestSignature::test_constructor_signatures[void]
 _core/tests/test_scalar_methods.py::TestSignature::test_constructor_signatures[str_]
+_core/tests/test_scalar_methods.py::TestSignature::test_constructor_signatures[bytes_]
1 error in 128.69s (0:02:08)
```

Python and NumPy Versions:

NumPy 2.5.0.dev0+git20260504.c9b6249, Python 3.12 (cp312 manylinux_riscv64 wheel CI)

Root cause:

`sctypes["others"]` is a `set` (`numpy/_core/_type_aliases.py`). Unpacking it directly with `*sctypes["others"]` in the parametrize decorator gives different ordering per worker.

Suggested fix:

Sort the set before unpacking (open to other approaches if `__name__` is not the right sort key):

```python
@pytest.mark.parametrize(
    "sctype",
    [
        *sctypes["int"],
        *sctypes["uint"],
        *sctypes["float"],
        *sctypes["complex"],
        *sorted(sctypes["others"], key=lambda t: t.__name__),
        np.datetime64,
        np.timedelta64,
    ],
)
```

First observed in PR #30995 (riscv64 wheel builds on RISE native runners).

rgommers:
The incomplete sorting before turning sets into lists caused issues with `pytest-xdist` not seeing the same test splits and therefore erroring out immediately (see gh-31414).

While this seems to happen only on RISC-V, the problem isn't actually RISC-V specific beyond it materializing there, possibly due to different implementation of sorting behavior when tie-breaking on the first sort key. That shouldn't be relied on though, this should be considered a (minor) bug in NumPy.

Closes gh-31414

#### AI Disclosure

Claude Opus 4.7 used to double-check the diagnosis. Fix implemented by hand.

### DIFF
--- a/numpy/_core/_type_aliases.py
+++ b/numpy/_core/_type_aliases.py
@@ -121,10 +121,10 @@ for type_info in typeinfo.values():
 
     del type_info, concrete_type
 
-# sort sctype groups by bitsize
+# sort sctype groups by bitsize, then by name (sort needs to be deterministic)
 for sctype_key in sctypes.keys():
     sctype_list = list(sctypes[sctype_key])
-    sctype_list.sort(key=lambda x: dtype(x).itemsize)
+    sctype_list.sort(key=lambda x: (dtype(x).itemsize, x.__name__))
     sctypes[sctype_key] = sctype_list
 
     del sctype_key, sctype_list


rgommers:
The one test failure is a floating-point warning from `np.dot`, that's unrelated.

mattip:
Let’s give it a try. Thanks @rgommers

PR #31418, opened 2026-05-12
TST: suppress shape deprecation warning in test_recarray_conflict_fields

eendebakpt:
### PR summary

The recarray field-conflict fallback exercises `arr.shape = ...` to trigger the field setter; the shape deprecation (#29536) made this emit a DeprecationWarning that surfaced in wheel build logs.

Noted in https://github.com/numpy/numpy/pull/29575#issuecomment-4433151108

#### AI Disclosure

AI was used to search through commit history and draft a fix.

### DIFF
--- a/numpy/_core/tests/test_records.py
+++ b/numpy/_core/tests/test_records.py
@@ -271,7 +271,9 @@ class TestFromrecords:
         assert_(type(ra.mean) is type(ra.var))
         ra = ra.reshape((1, 3))
         assert_(ra.shape == (1, 3))
-        ra.shape = ['A', 'B', 'C']
+        # gh-29536: setting .shape on an ndarray is deprecated
+        with pytest.warns(DeprecationWarning, match="Setting the shape"):
+            ra.shape = ['A', 'B', 'C']
         assert_array_equal(ra['shape'], [['A', 'B', 'C']])
         ra.field = 5
         assert_array_equal(ra['field'], [[5, 5, 5]])


mhvk (commented):
This looks fine, but I'm confused: how was this not seen in testing earlier?

eendebakpt:
> This looks fine, but I'm confused: how was this not seen in testing earlier?

From Claude: The warning is being silently swallowed. `pytest.ini` sets `filterwarnings = error`, which turns the `DeprecationWarning` into an exception. But `recarray.__setattr__` has a broad `except Exception:` (the field-name fallback path that routes `ra.shape = ...` to `setfield('shape', ...)` when the direct attribute set raises), and it catches the converted warning right along with the legitimate `TypeError` we expect for an invalid shape. So the test passes under pytest CI. The warning only leaks in wheel-build runs where `filterwarnings = error` isn't applied.

charris (approved):

seberg:
One thing we could actually do to fix it slightly cleaner would be to only give the deprecation warning on a successful reshape.
But `recarray` using `try/except` is so weird, and shape assignment warning makes at least some sense on failure as well, that I am also happy with this.

> This looks fine, but I'm confused: how was this not seen in testing earlier?

Right, at dev time we turn all warnings to errors, which is _usually_ the best thing, except for such weird `try/except` clauses.

mhvk:
OK, thanks for the explanation! My sense would be to put this in as is. If we do anything more, I think it would be to narrow down the exception caught in shape setting.

seberg:
Yeah, let's just get it in, thanks.  Not warning on something that errors anyway makes sense if it is related, but it isn't really here and `recarray` is just weird...

PR #31422, opened 2026-05-13
DOC: Fix indented table

Carreau:

When indented after a line that ends in a colon, RST parses as a definition list, the table being the definition.

This does not appear to be what we want.

see https://numpy.org/devdocs/dev/alignment.html

Built locally, this remove the bod text and dedent the table as expected.

#### AI Disclosure

No AI tools used

### DIFF
--- a/doc/source/dev/alignment.rst
+++ b/doc/source/dev/alignment.rst
@@ -41,13 +41,14 @@ alignment of 4 and "uint" alignment of 8 (equal to the true alignment of
 ``uint64``).
 
 Some cases where uint and true alignment are different (default GCC Linux):
-   ======   =========   ========    ========
-   arch     type        true-aln    uint-aln
-   ======   =========   ========    ========
-   x86_64   complex64          4           8
-   x86_64   float128          16           8
-   x86      float96            4          \-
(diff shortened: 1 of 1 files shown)


Carreau:
Question in the same file a bit below.

> 4. Strided copy code: Here, "uint alignment" is used instead.  If the itemsize
   of an array is equal to 1, 2, 4, 8 or 16 bytes and the array is uint
   aligned then instead NumPy will do ``*(uintN*)dst) = *(uintN*)src)`` for
   appropriate N. Otherwise, NumPy copies by doing ``memcpy(dst, src, N)``.

The unbalanced parentheses in ``*(uintN*)dst) = *(uintN*)src)`` seem wrong. should it read ``*(uintN*)dst = *(uintN*)src`` ? I can add it to this PR as well.

WarrenWeckesser (approved):
Looks good, thanks.

WarrenWeckesser:
Your suggested fix for the mismatched parentheses looks good too.

Carreau:
Ok, I removed the two lonely closing parentheses.

Carreau:
Thanks

PR #31433, opened 2026-05-14
BUG: Fix np.abs(nan) returning negative NaN on wasm32/Pyodide

### LINKED ISSUE #31421: BUG: np.abs(np.nan) on Pyodide WASM returns -np.nan
### Describe the issue:

`np.abs(np.nan)` should return a positive NaN but seems to return a negative one.

```python
# https://pyodide.org/en/0.29.4/console.html

import numpy as np

print(np.signbit(np.nan))  # prints False, as expected
print(np.signbit(-np.nan))  # prints True, as expected

print(np.signbit(np.abs(np.nan)))  # prints True?!
print(np.signbit(np.abs(-np.nan)))  # prints False, as expected
```

### Error message:

```shell

```

2.2.5
3.13.2 (main, May  7 2026, 08:28:34) [Clang 21.0.0git (https:/github.com/llvm/llvm-project 2f05451198e2f222ec66cec489

### Runtime Environment:

[{'numpy_version': '2.2.5',
  'python': '3.13.2 (main, May  7 2026, 08:28:34) [Clang 21.0.0git '
            '(https:/github.com/llvm/llvm-project 2f05451198e2f222ec66cec489',
  'uname': uname_result(system='Emscripten', node='emscripten', release='4.0.9', version='#1', machine='wasm32')},
 {'simd_extensions': {'baseline': [], 'found': [], 'not_found': []}}]

I was writing some modulo functions for different division functions (floor, ceil, trunc, rounded, Euclidean), created a testsuite, and also ran that in CI on Pyodide. While everything worked on native, the test fails in Pyodide.

I don't know where this bug originates, the chain could go from numpy to emscripten to clang's builtin. However, the WASM spec states that fabs(+-nan) = +nan

AnkitAhlawat7742:
Fix: #31421

### PR summary
This PR fixes a bug where np.signbit(np.abs(np.nan)) incorrectly returns True instead of False.

 ### problem 
 
 The `LONGDOUBLE_absolute` function in `numpy/_core/src/umath/loops.c.src` used `tmp + 0 to clear -0.0,` but this operation corrupted the sign bit of `NaN` values, causing np.signbit(np.abs(np.nan)) to return True.

### Testing 

Manually run the new test case and run all the `numpy/_core/tests/test_umath.py ` file test cases.

### DIFF
--- a/numpy/_core/tests/test_umath.py
+++ b/numpy/_core/tests/test_umath.py
@@ -4890,16 +4890,18 @@ def test_pos_nan():
     """Check np.nan is a positive nan."""
     assert_(np.signbit(np.nan) == 0)
 def test_abs_nan_signbit():
-    """#31421 abs(nan) preserves positive sign bit correctly for longdouble."""
-    pos_nan = np.longdouble(np.nan)
-    assert_(np.signbit(np.abs(pos_nan)) == 0,
-            "abs(+nan) should have positive sign for longdouble")
+    """#31421 abs(nan) preserves positive sign bit correctly."""
+    for dtype in [np.float32, np.float64, np.longdouble]:
+        pos_nan = dtype(np.nan)
+        assert_(np.signbit(np.abs(pos_nan)) == 0,
+                f"abs(+nan) should have positive sign for {dtype.__name__}")
 
-    neg_nan = np.longdouble(-np.nan)
-    assert_(np.signbit(np.abs(neg_nan)) == 0,
-            "abs(-nan) should have positive sign for longdouble")
+        neg_nan = dtype(-np.nan)
+        assert_(np.signbit(np.abs(neg_nan)) == 0,
+                f"abs(-nan) should have positive sign for {dtype.__name__}")
 
-    arr = np.array([np.nan, -np.nan], dtype=np.longdouble)
+    # Test with array
+    arr = np.array([np.nan, -np.nan])
     result = np.signbit(np.abs(arr))
     assert_array_equal(result, [False, False],
                       "abs of NaN array should have all positive signs")
--- a/numpy/_core/tests/test_umath.py
+++ b/numpy/_core/tests/test_umath.py
@@ -4889,6 +4889,21 @@ def test_nextafter_vs_spacing():
 def test_pos_nan():
     """Check np.nan is a positive nan."""
     assert_(np.signbit(np.nan) == 0)
+def test_abs_nan_signbit():
+    """#31421 abs(nan) preserves positive sign bit correctly."""
+    for dtype in [np.float16, np.float32, np.float64, np.longdouble]:
+        pos_nan = dtype(np.nan)
+        assert_(np.signbit(np.abs(pos_nan)) == 0,
+                f"abs(+nan) should have positive sign for {dtype.__name__}")
+
+        neg_nan = dtype(-np.nan)
+        assert_(np.signbit(np.abs(neg_nan)) == 0,
+                f"abs(-nan) should have positive sign for {dtype.__name__}")
+    arr = np.array([np.nan, -np.nan])
+    result = np.signbit(np.abs(arr))
+    assert_array_equal(result, [False, False],
+                      "abs of NaN array should have all positive signs")
+
 
 def test_reduceat():
     """Test bug in reduceat when structured arrays are not copied."""
--- a/numpy/_core/tests/test_umath.py
+++ b/numpy/_core/tests/test_umath.py
@@ -4890,16 +4890,16 @@ def test_pos_nan():
     """Check np.nan is a positive nan."""
     assert_(np.signbit(np.nan) == 0)
 def test_abs_nan_signbit():
-    """#31421 abs(nan) preserves positive sign bit correctly."""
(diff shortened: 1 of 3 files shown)


charris:
I want to know why adding 0.0 corrupted the nan sign, any ideas? Was it possible that the nan was already negative?

AnkitAhlawat7742:
HI @charris ,

The bug occurs because `nan > 0` always returns `false`so the code takes the else branch: `tmp = -in1` as per the existing code 
For positive NaN:
- `tmp = -(+nan)` = `-nan` (flips the sign bit)
- `tmp + 0` = `-nan` (the `+0` doesn't fix NaN sign bits)
- Result: negative NaN 

So `+ 0` was meant to clear `-0.0`, but it doesn't correct NaN sign bits so i checked `npy_fabsl()` fixes this issue  because it's designed to handle NaN
Hope i debug the issue in right direction

juntyr:
https://github.com/numpy/numpy/issues/31421#issuecomment-4450401840

> Hi [@juntyr](https://github.com/juntyr) , Did you also observe this issue with `float32` and `float64` types on wasm32/Pyodide? I noticed the same `tmp + 0` pattern exists in those implementations as well. Need to check with reviewer that , can we create separate PR for other types (float32`,`float64`)...if reviewer agree with this solution

I noticed the problem on WASM with `float32`, `float64`, and `float128`; but surprisingly `float16` works fine. Could the PR be amended to fix those as well?

seberg:
`float16` uses a custom bit-wise implementation for all of these.  But yea, I guess this code predates the `abs` being mandatory functions.  If the float abs also uses comparison, I think we might as well change it.
(Not sure what to think if wasm `abs` might just doesn't flip the signbit for NaN...)

AnkitAhlawat7742:
Thank you for the feedback! I have amended the PR to fix float32 and float64 as well.they had the same` a > 0 ? a : -a `pattern in` loops_unary_fp.dispatch.c.src` and modified the test case accordingly 
@seberg please review it once again and let me know if any modification needed

AnkitAhlawat7742:
HI @seberg ,
Please review the PR ..Happy to make any changes if needed!

numpy/_core/tests/test_umath.py
```diff
@@ -4893,6 +4889,23 @@ def test_nextafter_vs_spacing():
 def test_pos_nan():
     """Check np.nan is a positive nan."""
     assert_(np.signbit(np.nan) == 0)
+def test_abs_nan_signbit():
+    """#31421 abs(nan) preserves positive sign bit correctly."""
+    for dtype in [np.float32, np.float64, np.longdouble]:
```
juntyr: can we also test np.float16 here to guard against a future regression

AnkitAhlawat7742: Agreed ..!
Thanks for the suggestion 
i have updated the test case for `float16`

numpy/_core/tests/test_umath.py
```diff
@@ -4893,6 +4889,23 @@ def test_nextafter_vs_spacing():
 def test_pos_nan():
     """Check np.nan is a positive nan."""
     assert_(np.signbit(np.nan) == 0)
+def test_abs_nan_signbit():
```
tylerjereddy: maybe at least have an empty line above the function def

numpy/_core/tests/test_umath.py
```diff
@@ -4893,6 +4889,23 @@ def test_nextafter_vs_spacing():
...
+    """#31421 abs(nan) preserves positive sign bit correctly."""
+    for dtype in [np.float16, np.float32, np.float64, np.longdouble]:
```
tylerjereddy: could parametrize for better feedback, although array test below would then make sense to split off

numpy/_core/tests/test_umath.py
```diff
@@ -4893,6 +4889,23 @@ def test_nextafter_vs_spacing():
...
+        pos_nan = dtype(np.nan)
+        assert_(np.signbit(np.abs(pos_nan)) == 0,
+                f"abs(+nan) should have positive sign for {dtype.__name__}")
+
+        neg_nan = dtype(-np.nan)
+        assert_(np.signbit(np.abs(neg_nan)) == 0,
```
tylerjereddy: Should we start preferring the "modern" plain `assert` for new tests? I guess that's a minor point

numpy/_core/tests/test_umath.py
AnkitAhlawat7742: Done

numpy/_core/tests/test_umath.py
AnkitAhlawat7742: Done ..!!

numpy/_core/tests/test_umath.py
AnkitAhlawat7742: Thanks for the feedback ..Test case has been updated accordingly ,I have parametrized the test for better feedback, split the array test into a separate function.

seberg (approved):
Thanks, LGTM, let's put it in, since C99 guarantees `abs` being defined.

I am not sure that C99 actually guarantees the sign of NaN (or that we should), but in practice I don't see anything plausibly diverging.

PR #31437, opened 2026-05-14
BUG:  `dtype.metadata` lost for zero-sized void dtypes in array creation functions

### LINKED ISSUE #31436: BUG:  `dtype.metadata` lost for zero-sized void dtypes in array creation functions
### Describe the issue:

In NumPy 2.x, `dtype.metadata` is silently dropped when creating arrays with zero-sized void dtypes (`np.dtype(np.void, 0, metadata={...})`) via `np.zeros()`, `np.empty()`, and similar array creation functions.

This is a regression from NumPy 1.x behavior where metadata was preserved for all dtype kinds.

### Key observations

1. **Only zero-sized void dtypes are affected.** Regular numeric dtypes (float64, int32, etc.) and structured dtypes correctly preserve metadata through all creation functions.

2. **`np.ndarray()` (direct constructor) works correctly** — it preserves metadata even for zero-sized void dtypes.

3. **`np.array()` and `np.asarray()` also work** — only `np.zeros()`, `np.empty()`, `np.ones()`, and `np.full()` drop the metadata.

4. **`.view(dtype)` preserves metadata** and can be used as a workaround.

### Workaround

Use `np.ndarray(shape, dtype=dt)` instead of `np.zeros(shape, dtype=dt)`, or apply `.view(dt)` after creation:

```python
# Workaround 1: use np.ndarray directly
a = np.ndarray((5,), dtype=dt)

# Workaround 2: create then view
a = np.zeros(5, dtype=np.dtype(np.void, 0)).view(dt)
```

### Related

This likely stems from the internal dtype canonicalization path in NumPy 2.x that normalizes zero-sized void dtypes without carrying over the metadata field.

```python
import numpy as np

# Create a zero-sized void dtype with metadata
dt = np.dtype(np.void, 0, metadata={'type': 'struct_marker'})
print(f"dtype created: {dt}, metadata: {dt.metadata}")
# Output: dtype created: |V0, metadata: {'type': 'struct_marker'}

# Create an array using np.zeros — metadata is LOST
a = np.zeros(5, dtype=dt)
print(f"np.zeros dtype: {a.dtype}, metadata: {a.dtype.metadata}")
# Output: np.zeros dtype: |V0, metadata: None  ← BUG

# Create an array using np.empty — metadata is LOST
b = np.empty(5, dtype=dt)
print(f"np.empty dtype: {b.dtype}, metadata: {b.dtype.metadata}")
# Output: np.empty dtype: |V0, metadata: None  ← BUG

# --- Workaround: np.ndarray() preserves metadata ---
c = np.ndarray((5,), dtype=dt)
print(f"np.ndarray dtype: {c.dtype}, metadata: {c.dtype.metadata}")
# Output: np.ndarray dtype: |V0, metadata: {'type': 'struct_marker'}  ← OK
```

### Error message:

```shell
#### Expected behavior

`np.zeros(n, dtype=dt).dtype.metadata` should equal `dt.metadata` for all dtype kinds, including zero-sized void dtypes.

#### Actual behavior

`np.zeros(n, dtype=dt).dtype.metadata` returns `None` when `dt` is a zero-sized void dtype with metadata.
```

- **NumPy**: 2.2.6 (also reported to affect earlier 2.x versions)
- **Python**: 3.11.2
- **OS**: Linux (Debian)

### Runtime Environment:

_No response_

Zero-sized void dtypes with metadata are used as semantic type tags in serialization libraries (e.g., to mark struct boundaries, type annotations, etc.). The metadata carries information like `{'type': 'struct'}` that cannot be represented any other way since the dtype has zero data size.

moaedy:
### PR summary

`descr_is_legacy_parametric_instance` in `numpy/_core/src/multiarray/descriptor.c` treated any unsized legacy dtype (e.g. `|V0`) as a bare prototype, so `PyArray_ExtractDTypeAndDescriptor` dropped the caller's descriptor and `PyArray_Zeros_int` / `PyArray_Empty_int` rebuilt one from the DType singleton, discarding the original `metadata` dict. The fix treats a descriptor with a non-null `metadata` field as already-concrete, so the user-supplied descriptor (and its metadata) survives `np.zeros`, `np.empty`, `np.ones`, and `np.full`.

Long-time NumPy user interested in the dtype system; happy to contribute a small fix to descriptor handling.

#### AI Disclosure

Used Claude to draft the patch and the regression test. All changes reviewed by hand before commit.

Closes #31436

### DIFF
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -1420,6 +1420,16 @@ descr_is_legacy_parametric_instance(PyArray_Descr *descr,
         return 0;
     }
 
+    /*
+     * A descriptor with user-attached metadata is not a bare prototype:
+     * the metadata is explicit state that must round-trip through array
+     * creation, so keep it rather than collapsing to the DType singleton
+     * (gh-31436).
+     */
+    if (((_PyArray_LegacyDescr *)descr)->metadata != NULL) {
+        return 0;
+    }
+
     if (PyDataType_ISUNSIZED(descr)) {
         return 1;
     }
--- a/numpy/_core/tests/test_dtype.py
+++ b/numpy/_core/tests/test_dtype.py
@@ -927,6 +927,20 @@ class TestMetadata:
         d = np.dtype((np.void, np.dtype('i4,i4', metadata={'datum': 1})))
         assert_(d.metadata == {'datum': 1})
 
+    def test_metadata_preserved_zero_sized_void(self):
+        # gh-31436: array creation functions dropped metadata for unsized
+        # legacy dtypes (notably zero-sized void) because the descriptor was
+        # treated as a bare prototype and rebuilt from the DType singleton.
+        dt = np.dtype('V0', metadata={'type': 'struct_marker'})
+        assert dt.metadata == {'type': 'struct_marker'}
+        for arr in (np.zeros(5, dtype=dt),
+                    np.empty(5, dtype=dt),
+                    np.ones(5, dtype=dt),
+                    np.full(5, b'', dtype=dt)):
+            assert arr.dtype == dt
+            assert arr.dtype.metadata == {'type': 'struct_marker'}
+
+
 class TestString:
     def test_complex_dtype_str(self):
         dt = np.dtype([('top', [('tiles', ('>f4', (64, 64)), (1,)),


seberg:
Hmmmm, interesting, thanks!  Yeah, we should improve this in a few places and front-load `"V" -> type(dtype("V"))`, but this is fine as such.

I think we can actually improve this (I may have written this earlier), to add `UNSIZED(...) && NPY_DTYPE(descr)->singleton == descr`.  That way we don't focus on metadata, which isn't the interesting point as such.  That will also make your `np.dtype((np.void, 0))` be a valid workaround.
(And it would make sense to make sure that `np.dtype("V0")` creates a new dtype instance just so that this but at that point I should dive into the deeper fix and just do that now maybe.)

PR #24214, opened 2023-07-19
ENH: Create helper for conversion to arrays

seberg:
Had to give this a start to be able to experiment.  It does solve at least some problems (e.g. `np.select()` use-case and it looks like it would fix and/or simplify linalg code).

The main thing is that we often convert multiple arrays, and then want something like:
* `a, b, c = np.asarray(a), np.asarray(b), np.asarray(c)` and then `dtype = np.result_type(a, b, c)` often with an additional `0.0`.  That is a pretty common pattern, NEP 50 makes it tricky if you wish to support weak scalars, because you need to call `result_type` with the scalar information intact (before the `asarray` or manually undo).  (For example, this happens in `np.select()`)
  * Maybe libraries can disregard that, but for NumPy it's hard, since you want many functions to be ufunc-like, so if ufuncs do it, others probably should be able to/try to.
* Maybe mainly internally we sometimes need the `__array_wrap__` for subclass support, although it seems consistently used for single input functions (outside ufuncs).
  * Additional motif:  A clean wrapping pattern seems required for making scalar logic sane.
* It  may be useful to know if an input was a scalar or not (not a 0-D array).  You need to ask NumPy about it, but it discovered at conversion time.  So it is an information that is weird to tag onto an array (probably?) but figuring it out later would require duplicating work.

First, I wanted to just make a single function.  But with multiple things above that may be interesting to return, and e.g. `result_type()` might error and I am not sure we always need it.

So, is it insane to create a rich object (rather than a complicated return value and many kwargs):
```
# better names of course, currently in `np.core._multiarray_umath.ArrayCreationHelper`
arrays = creation_helper(*inputs)

# e.g. a linspace could do something like this:
dtype = arrays.common_dtype(ensure_inexact=True)

arrays.arrays  # arrays as tuple (probably allow arrays[0])

# Apply the typical wrapping if needed
# (should cache wrap, or fetch it early on rather, maybe just
# make it a method which applies the wrapping `arrays.wrap(result)`).
if arrays.wrap is not None:
    return arrays.wrap(result)
return result 
```
(I can see `a, b, info = helper(a, b)` as a nice pattern; OTOH it makes including casting kwarg heavy since you somewhat need to include it in the first step.)

Further things might include:
* Casting to a given or the common dtype.  (maybe `arrays.to_common_dtype(ensure_inexact=True)`)
* I think @mdhaber was hoping to include broadcasting.  Not sure, since it is one call away, but also seems fine to add.

PR #31447, opened 2026-05-15
BUG: pass DType class (not instance) to ufunc in `_unsigned_subtract`

SwayamInSync:
## Summary

`np.histogram` raises `TypeError: Cannot pass a new user DType instance to the dtype or signature arguments of ufuncs. Pass the DType class instead.` for any non-legacy user DType e.g. `numpy-quaddtype`

Root cause: `_unsigned_subtract` in `numpy/lib/_histograms_impl.py` passes `dtype=dt` (a DType **instance** from `np.result_type`) to `np.subtract`. For legacy/built-in DTypes the dispatcher silently converts instance to class, but for new user DTypes the dispatcher rejects instances by design (see`ufunc_object.c::_get_dtype`, comment at `ufunc_object.c:3954-3958`).

This is the same anti-pattern that #25201 (commit 04444cdff3, "MAINT: Use new converter to fix promotions and avoid manual wrapping") cleaned up across `function_base.py`, `fromnumeric.py`, `_function_base_impl.py`, etc. but `_histograms_impl.py` was missed. `linspace`, for example, already does `dtype=type(dt)`  with an explicit comment.

## AI Disclosure:
None

cc: @seberg

### DIFF
--- a/numpy/_core/tests/test_custom_dtypes.py
+++ b/numpy/_core/tests/test_custom_dtypes.py
@@ -195,6 +195,26 @@ class TestSFloat:
         with pytest.raises(TypeError):
             np.add(a, a, out=c, casting="safe")
 
+    def test_sfloat_histogram_passes_dtype_class(self):
+        # Regression test for gh-31447: ``_unsigned_subtract`` used to
+        # forward a DType instance (from ``np.result_type``) to
+        # ``np.subtract``, which the ufunc dispatcher rejects for
+        # non-legacy user DTypes with "Cannot pass a new user DType
+        # instance ...". SF has no registered subtract loop, so the
+        # call still raises after the fix, but with the expected
+        # missing-loop error instead of the instance-rejection error.
+        # ``_unsigned_subtract`` is exercised directly because SF also
+        # lacks ``minimum``/``maximum`` loops, so a full
+        # ``np.histogram(arr)`` call fails earlier without ever
+        # reaching the fixed code path.
+        from numpy.lib._histograms_impl import _unsigned_subtract
+        a = np.array([3.]).view(SF(1.))
+        b = np.array([1.]).view(SF(1.))
+        try:
+            _unsigned_subtract(a, b)
+        except TypeError as e:
+            assert "Cannot pass a new user DType instance" not in str(e)
+
     @pytest.mark.parametrize("ufunc",
             [np.logical_and, np.logical_or, np.logical_xor])
     def test_logical_ufuncs_casts_to_bool(self, ufunc):
--- a/numpy/_core/tests/test_custom_dtypes.py
+++ b/numpy/_core/tests/test_custom_dtypes.py
@@ -196,24 +196,24 @@ class TestSFloat:
             np.add(a, a, out=c, casting="safe")
 
     def test_sfloat_histogram_passes_dtype_class(self):
-        # Regression test for gh-31447: ``_unsigned_subtract`` used to
-        # forward a DType instance (from ``np.result_type``) to
-        # ``np.subtract``, which the ufunc dispatcher rejects for
-        # non-legacy user DTypes with "Cannot pass a new user DType
-        # instance ...". SF has no registered subtract loop, so the
-        # call still raises after the fix, but with the expected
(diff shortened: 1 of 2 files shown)


jorenham:
Wait... `dtype` parameters can also accept `type[np.dtype]`? Because `numpy.typing.DTypeLike` currently doesn't accept that :sweat_smile:

SwayamInSync:
yeah I noticed it here
https://github.com/numpy/numpy/blob/79b033101a38a0f1fa2c6b94380f31041f74da55/numpy/_core/src/umath/ufunc_object.c#L3964

So maybe `type[np.dtype]` got matched with the following? 👀 
https://github.com/numpy/numpy/blob/79b033101a38a0f1fa2c6b94380f31041f74da55/numpy/_typing/_dtype_like.py#L101

jorenham:
> So maybe `type[np.dtype]` got matched with the following? 👀
> 
> https://github.com/numpy/numpy/blob/79b033101a38a0f1fa2c6b94380f31041f74da55/numpy/_typing/_dtype_like.py#L101


Ah yea of course; the wildcard `type` type will accept dtype types (not a typo).

The same can't be said for the generic `numpy._typing._DTypeLike[_ScalarT: np.generic]` type though...

ngoldbaum:
Thanks @SwayamInSync. I'm OK to put this in as-is, but it'd be nice to see a regression test as well. Check out `numpy/_core/tests/test_custom_dtypes.py`, which has tests based on an internal tests-only scaled-float dtype. I think all you'd need to do is add a test that calls `np.histogram` on a scaled float array.

Feel free to add more tests there if you'd like for the similar patterns fixed in https://github.com/numpy/numpy/pull/24214.

SwayamInSync:
Added the test, actually `SF` arrays does not implement `minimum/maximum` so now a different error occurs, hence picked to test with single element arrays on `_unsigned_subtract` directly (also docunmented in comments in code)

numpy/_core/tests/test_custom_dtypes.py
```diff
@@ -195,6 +195,26 @@ def test_addition_cast_safety(self):
...
+        # lacks ``minimum``/``maximum`` loops, so a full
+        # ``np.histogram(arr)`` call fails earlier without ever
+        # reaching the fixed code path.
+        from numpy.lib._histograms_impl import _unsigned_subtract
+        a = np.array([3.]).view(SF(1.))
+        b = np.array([1.]).view(SF(1.))
+        try:
+            _unsigned_subtract(a, b)
+        except TypeError as e:
+            assert "Cannot pass a new user DType instance" not in str(e)
```
seberg: I would be slightly happier with a positive match using `pytest.raises(...)`.  Because we might change the error message there eventually and then the test becomes meaningless. OTOH, if we add a subtract loop or so we can adapt the test just fine.

SwayamInSync: Sure, I can add that but still have to use `match` argument there, as in this case the new error we get is `"ufunc 'subtract' did not contain a loop"` instead of `"Cannot pass a new user DType instance"`
So a general `pytest.raises(TypeError)` will not give any hint that this test belong to current issue of histogram fix.

ngoldbaum (approved):
A little annoying that the scaled float test dtype isn't quite adequate for this but LGTM.

SwayamInSync:
Should we consider replacing scaled float with `quaddtype` (both are float) and quaddtype keep getting updated upstream so it will be less hectic as maintaining a custom-dtype just for testing?

ngoldbaum:
I'm all for adding quaddtype as a test dependency if that would be helpful. There are some circular dependency concerns that may be annoying to deal with but I personally think we can handle any extra coordination needed around that.

Probably best to wait until 2.6 is branched next week though.

seberg:
> I'm all for adding quaddtype as a test dependency if that would be helpful.

We already have one test IIRC, the problem is that with older (maybe current) versions you have to do the `quaddtype` version check without even importing (as that would cause the NumPy test suite to fail).

But, here this is a bit awkward, but seems fine.

PR #31448, opened 2026-05-16
BUG: support () indexing for string scalars

### LINKED ISSUE #31435: BUG: `numpy.str_` and `numpy.bytes_` scalars do not support `[()]` indexing despite having `shape == ()`
### Describe the issue:

`numpy.str_` and `numpy.bytes_` are 0-dimensional numpy scalars (`shape == ()`, `ndim == 0`), yet indexing them with `[()]` (empty tuple, which is the standard way to extract the value from a 0-d array/scalar) raises a `TypeError`. This is inconsistent with all other numpy scalar types where `scalar[()]` works correctly.

The issue appears to be that Python's `str.__getitem__` / `bytes.__getitem__` takes priority over numpy's scalar indexing protocol, interpreting `()` as a tuple index into the string/bytes sequence rather than as a 0-d array index.

### Expected behavior

`numpy.str_('hello')[()]` should return `'hello'` (the Python `str` value), consistent with how `numpy.int64(5)[()]` returns `5`.

Since `numpy.str_` reports `shape == ()` and `ndim == 0`, it should support the standard 0-d indexing protocol that all other numpy scalars support. Code that generically extracts values from 0-d numpy scalars via `val[()]` breaks specifically for string and bytes scalars.

### Workaround

Use `.item()` instead of `[()]`:

```python
np.str_('hello').item()   # => 'hello'  (works)
np.bytes_(b'hello').item()  # => b'hello'  (works)
```

`.item()` works consistently across all numpy scalar types.

```python
import numpy as np

# All other numpy scalar types support [()] indexing:
assert np.int64(5)[()] == 5
assert np.float64(3.14)[()] == 3.14
assert np.bool_(True)[()] == True
assert np.complex128(1+2j)[()] == (1+2j)

# numpy.str_ and numpy.bytes_ do NOT, despite having shape == ():
s = np.str_('hello')
print(f"type: {type(s)}, shape: {s.shape}, ndim: {s.ndim}")  # shape=(), ndim=0

s[()]  # TypeError: string indices must be integers, not 'tuple'

b = np.bytes_(b'hello')
print(f"type: {type(b)}, shape: {b.shape}, ndim: {b.ndim}")  # shape=(), ndim=0

b[()]  # TypeError: byte indices must be integers or slices, not tuple
```

### Error message:

```shell



```

```
numpy version: 2.2.6
Python 3.11.2 (main, Aug 26 2024, 07:20:54) [GCC 12.2.0]
```

### Runtime Environment:

_No response_

_No response_

AriGood:
### PR summary

Zero-dimensional arrays and NumPy scalar types are documented to support the empty-tuple index `x[()]` as a scalar index that returns the scalar value. However, `np.str_` and `np.bytes_` scalars currently raise `TypeError` when indexed with `scalar[()]`, because the index is handled by the underlying Python `str`/`bytes` mapping implementation instead of NumPy’s scalar indexing behavior.

This change makes `np.str_` and `np.bytes_` scalars treat `scalar[()]` as a scalar index and return the scalar value, consistent with other 0‑d NumPy scalars. All other indices are still delegated to the Python `str`/`bytes` mapping implementation to preserve existing behavior.

In addition, it adds regression tests in
`numpy/_core/tests/test_scalarmath.py` that assert:

- `np.str_("abc").shape == ()` and `np.str_("abc")[()]` returns the same
  scalar value.
- `np.bytes_(b"abc").shape == ()` and `np.bytes_(b"abc")[()]` returns
  the same scalar value.

Fixes gh-31435.

Hi, this is my first contribution to NumPy. I use NumPy a lot in my own projects, so I wanted to help improve a library I use a lot and start getting involved in open source development.
#### AI Disclosure

ChatGPT was used to help make sure this PR is following guidelines. No AI code was used.

### DIFF
--- a/numpy/_core/src/multiarray/scalartypes.c.src
+++ b/numpy/_core/src/multiarray/scalartypes.c.src
@@ -4240,6 +4240,40 @@ gen_arrtype_subscript(PyObject *self, PyObject *key)
 }
 
 
+/*
+ * String-like scalars support empty-tuple indexing like other NumPy scalars.
+ * This returns the scalar itself for scalar[()], while all other indexing
+ * is delegated to the underlying Python bytes/str mapping implementation.
+ */
+static PyObject *
+string_arrtype_subscript(PyObject *self, PyObject *key)
+{
+    if (PyTuple_CheckExact(key) && PyTuple_GET_SIZE(key) == 0) {
+        Py_INCREF(self);
+        return self;
+    }
+    return PyBytes_Type.tp_as_mapping->mp_subscript(self, key);
+}
+
+static PyObject *
+unicode_arrtype_subscript(PyObject *self, PyObject *key)
+{
+    if (PyTuple_CheckExact(key) && PyTuple_GET_SIZE(key) == 0) {
+        Py_INCREF(self);
+        return self;
+    }
+    return PyUnicode_Type.tp_as_mapping->mp_subscript(self, key);
+}
+
+static PyMappingMethods string_arrtype_as_mapping = {
+    .mp_subscript = (binaryfunc)string_arrtype_subscript,
+};
+
+static PyMappingMethods unicode_arrtype_as_mapping = {
+    .mp_subscript = (binaryfunc)unicode_arrtype_subscript,
+};
+
+
 /**begin repeat
  * #Name = Bool,
  *         Byte, Short, Int, Long, LongLong,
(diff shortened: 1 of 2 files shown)


seberg:
Thanks, but I'll close this until the discussion on the issue is even somewhat close to a decision that this makes sense.

PR #31453, opened 2026-05-17
BUG: Fix regression in ``arr.conj()``

seberg:
I'll make these helpers static-inline functions on main. Unfortunately, this doesn't just cause wrong promotion for bool->uint8 in `arr.conj()`, but it also leads to all other non-complex types returning unnecessary copies which seems like a pretty unfortunate regression, even if it might not be very visible in practice.

I am suspecting we may need a 2.4.6 for this :(

(No AI)


charris:
Thanks Sebastian, seem simple enough.

PR #31454, opened 2026-05-17
MAINT: Add tests for conj and make NPY_DT_is_ helpers static inline

seberg:
Most of these used the flags, but that can go wrong because descrs also have flags :(.
This makes them static inline, for the `NPY_DT_is_user_defined` that is irrelevant, but just change it anyway (both structs have the identical type number).

(No AI)

### DIFF
--- a/numpy/_core/src/multiarray/multiarraymodule.c
+++ b/numpy/_core/src/multiarray/multiarraymodule.c
@@ -3851,7 +3851,7 @@ format_longfloat(PyObject *NPY_UNUSED(dummy), PyObject *args, PyObject *kwds)
  */
 static int _is_user_defined_string_array(PyArrayObject* array)
 {
-    if (NPY_DT_is_user_defined(PyArray_DESCR(array))) {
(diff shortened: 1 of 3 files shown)


numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -3851,7 +3851,7 @@ format_longfloat(PyObject *NPY_UNUSED(dummy), PyObject *args, PyObject *kwds)
  */
 static int _is_user_defined_string_array(PyArrayObject* array)
 {
-    if (NPY_DT_is_user_defined(PyArray_DESCR(array))) {
+    if (NPY_DT_is_user_defined(NPY_DTYPE(PyArray_DESCR(array)))) {
```
seberg, line 3854: Just to be double clear: This is a non-change, both DType and descr work here OK as they both should have the same `type_num`.  But the macro-name suggests DType, so let's use that.
If anything may should be slightly safer.

charris:
Thanks Sebastian.

PR #31455, opened 2026-05-17
added __complex__ to scalars

### LINKED ISSUE #27305: BUG: scalars missing several methods for array api compat
### Describe the issue:

I keep getting stuck trying to get tests in https://github.com/data-apis/array-api-tests to pass and the final errors are often due to arrays having turned into scalars and the tests not expecting that, (https://data-apis.org/array-api/draft/API_specification/array_object.html)

Reading #26850 it seems the intention is that the scalars should support the array api spec?

```python
import numpy as np
import array_api_strict as xps


# Some examples:
xps.mean(xps.asarray(4, dtype=xps.float32)).__iadd__(1)
np.mean(np.asarray(4, dtype=np.float32)).__iadd__(1) # AttributeError: 'numpy.float32' object has no attribute '__iadd__'

xps.mean(xps.asarray(4, dtype=xps.float32)).__complex__()
np.mean(np.asarray(4, dtype=np.float32)).__complex__() # AttributeError: 'numpy.float32' object has no attribute '__complex__'
```

### Error message:

_No response_

import sys, numpy; print(numpy.__version__); print(sys.version)
2.1.0
3.12.2 | packaged by conda-forge | (main, Feb 16 2024, 20:42:31) [MSC v.1937 64 bit (AMD64)]

### Runtime Environment:

_No response_

### Context for the issue:

_No response_

AriGood:
### PR summary

  
NumPy scalars now implement `__complex__()` consistently across real and complex dtypes, aligning them with Python’s numeric protocol. Real scalars share a new helper that converts the value via float(self) and returns a Python complex with zero imaginary part, complex scalars return a Python complex with matching real and imaginary parts. 

This fixes #27305 

#### AI Disclosure

Copilot was used to help orient me to scalartypes.c.src. No AI code was used.

### DIFF
--- a/numpy/_core/src/multiarray/scalartypes.c.src
+++ b/numpy/_core/src/multiarray/scalartypes.c.src
@@ -2080,6 +2080,7 @@ gentype_getarray(PyObject *scalar, PyObject *args)
 }
 
 static char doc_sc_wraparray[] = "__array_wrap__ implementation for scalar types";
+static char doc_scalar_complex[] = "__complex__($self, /)\n--\n\ncomplex(self)";
 
 /*
  * __array_wrap__ for scalars, returning a scalar if possible.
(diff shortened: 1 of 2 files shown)


jorenham:
We don't allow copilot reviews here.

jorenham:
Also, as you can read in the issue, there's no consensus on whether we actually want to do this.

AriGood:
Sorry about the review. 
It was still listed as a milestone (at least the reduced scope of complex is) so I thought it was at least worth making a PR in case you wanted to implement it. Feel free to close it if you don't think it is relevant.

seberg:
Thanks, but given the discussion on the issue, I think I'll close the PR for now.  If typing can get confused, the arguments in favor of this need to be strengthened, IMO.

PR #31458, opened 2026-05-18
BUG: Do not forward descending unless explicitly passed

pllim:
### PR summary

My interpretation of https://github.com/numpy/numpy/pull/31345#issuecomment-4478938827 from @seberg 

👋 from Astropy to fix downstream failure (https://github.com/astropy/astropy/issues/19760)

#### AI Disclosure

I hate AI and also too cheap to pay for one.

### DIFF
--- a/numpy/_core/fromnumeric.py
+++ b/numpy/_core/fromnumeric.py
@@ -939,7 +939,7 @@ def _sort_dispatcher(
 
 
 @array_function_dispatch(_sort_dispatcher)
-def sort(a, axis=-1, kind=None, order=None, *, stable=None, descending=None):
+def sort(a, axis=-1, kind=None, order=None, *, stable=None, descending=np._NoValue):
     """
     Return a sorted copy of an array.
 
@@ -1099,7 +1099,7 @@ def sort(a, axis=-1, kind=None, order=None, *, stable=None, descending=None):
     else:
         a = asanyarray(a).copy(order="K")
     # Sanitize for backward-compatibility
-    if descending:
+    if descending is not np._NoValue:
         a.sort(axis=axis, kind=kind, order=order, stable=stable, descending=descending)
     else:
         a.sort(axis=axis, kind=kind, order=order, stable=stable)
@@ -1113,7 +1113,7 @@ def _argsort_dispatcher(
 
 
 @array_function_dispatch(_argsort_dispatcher)
-def argsort(a, axis=-1, kind=None, order=None, *, stable=None, descending=None):
+def argsort(a, axis=-1, kind=None, order=None, *, stable=None, descending=np._NoValue):
     """
     Returns the indices that would sort an array.
 
@@ -1233,7 +1233,7 @@ def argsort(a, axis=-1, kind=None, order=None, *, stable=None, descending=None):
 
     """
     # Sanitize for backward-compatibility
-    if descending:
+    if descending is not np._NoValue:
         return _wrapfunc(
             a,
             "argsort",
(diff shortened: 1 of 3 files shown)


numpy/_core/fromnumeric.py
```diff
@@ -1098,7 +1098,12 @@ def sort(a, axis=-1, kind=None, order=None, *, stable=None, descending=None):
         axis = -1
     else:
         a = asanyarray(a).copy(order="K")
-    a.sort(axis=axis, kind=kind, order=order, stable=stable, descending=descending)
+    # Sanitize for backward-compatibility
+    if descending:
+        extra_kw = {"descending": descending}
+    else:
+        extra_kw = {}
+    a.sort(axis=axis, kind=kind, order=order, stable=stable, **extra_kw)
```
seberg: Awesome, thanks @pllim.  I would be tempted to just duplicating the `return a.sort(...)` (to avoid creating a dict just to unpack it immediately).
A test would be great, but I am also fine if not. (Needs a small dummy object -- I guess in `test_multiarray` although that file is annoying large.)

(We sometimes use `np._NoValue`, but I think in this case it should be fine to just not pass on if the user passes `descending=False` explicitly.)

Both are somewhat nits though, I admit :).

We probably should do the same change to `argsort()`.  (It uses `wrapfunc`, but that is just because it doesn't work in-place mainly, I guess.  We can do just the same and duplicate.)

pllim: Thanks for the detailed feedback! I implemented the functional code changes, hopefully as you requested.

However, I noticed there are already a bunch of `test_sort_descending` tests in `numpy/_core/tests/test_multiarray.py`. I am not sure how to write another test that would trigger this error without introducing that class from astropy. 😬

MaanasArora: Thank you for fixing this @pllim! About the test, I think you could create a small dummy class (does not need to have any real functionality or mirror astropy) that implements `argsort` with the right (kw)args, then write a test that instantiates an object and checks there is no error when calling `np.argsort(a, descending=False)`.

jorisvandenbossche:
For the `argsort` case (just dispatching to the object, and not going through `__array_func__`), you _could_ add a test with a small dummy class with just an `argsort` method with the old signature. I don't know how useful that would be, though (or how long numpy would like to keep the back compat for adding the `descending` keyword)

seberg:
I pushed a test, but someone should have a quick look (verified that it failed for both sort and argsort before).
Was worth it already for noticing the `dark-descent` branch name ;).

numpy/_core/fromnumeric.py
```diff
@@ -1098,7 +1098,11 @@ def sort(a, axis=-1, kind=None, order=None, *, stable=None, descending=None):
         axis = -1
     else:
         a = asanyarray(a).copy(order="K")
-    a.sort(axis=axis, kind=kind, order=order, stable=stable, descending=descending)
+    # Sanitize for backward-compatibility
+    if descending:
```
mhvk: In other code, we would set the argument to `np._NoValue` and pass on the argument if it is not that (see, e.g., `argmax` in the same file). I think that is better, because then if someone explicitly passes in `descending=False`, they get the correct error that the argument does not exist.

seberg: Hmmmm, yeah, let's go with this.  I somewhat thought that here the default may be unambiguous enough, but really... it is also for `keepdims=` probably and safer.

pllim:
Thank you for adding the test here, @seberg ! Given I am unfamiliar with your testing setup, it is more elegant than what I could have come up with. Is there anything else you need from me? If not, I'll let you do your thing. And thanks for appreciating the branch name.

seberg:
OK, changed to `np._NoValue`.

mhvk (approved):
Looks good, thanks both!!

pllim:
Hmm CI is failing. Is that related?

pllim:
Thank you for cleaning up my PR and bringing it over the finish line! I appreciate the timely response.

PR #31461, opened 2026-05-18
BUG: Handle nan values on histogram's slow path.

### LINKED ISSUE #28730: BUG: histogram error with nan objects
### Describe the issue:

If np.histogram is given an array of dtype object and containing nans, the histogram is incorrect

```python
import numpy as np
print(np.histogram(np.asarray([0., 1., np.nan], dtype=object), bins=np.arange(0, 5))[0])
print(np.histogram(np.asarray([0., 1., np.nan], dtype=float), bins=np.arange(0, 5))[0])
```

### Error message:

```shell
No error
```

Python 3.13.1
Numpy 2.2.2

### Runtime Environment:

_No response_

### Context for the issue:

_No response_

JosephMehdiyev:
### PR summary
Fixes https://github.com/numpy/numpy/issues/28730
Contrary to histogramdd (which is another problem I think), histogram uses a different algorithm that uses `_search_sorted_inclusive` on its slow path where nan values work ambigiously depending on the dtype. This is a small workaround to fix it basically.

~I am still not sure if this fix is proper, I will play around the PR until its ready~

#### AI Disclosure

I have used claude to review the code. The changes are not AI written.

### DIFF
--- a/numpy/lib/_histograms_impl.py
+++ b/numpy/lib/_histograms_impl.py
@@ -876,13 +876,23 @@ def histogram(a, bins=10, range=None, density=None, weights=None):
         cum_n = np.zeros(bin_edges.shape, ntype)
         if weights is None:
             for i in _range(0, len(a), BLOCK):
-                sa = np.sort(a[i:i + BLOCK])
+                tmp_a = a[i:i + BLOCK]
+                # nan != nan
+                not_nan = tmp_a == tmp_a
+                if not all(not_nan):
+                    tmp_a = tmp_a[not_nan]
+                sa = np.sort(tmp_a)
                 cum_n += _search_sorted_inclusive(sa, bin_edges)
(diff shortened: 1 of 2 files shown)


JosephMehdiyev:
Maybe an additional warning about the data containing nan values and how the histogram handles it would be nice addition? Otherwise it might be ambigious?

seberg:
If we go with https://github.com/numpy/numpy/pull/31431, I think we'll get this for free so I am not sure we should go the manual route here.

JosephMehdiyev:
#31431 surely would be a better fix (didn't know that PR existed), although I tried to test that branch and the issue is still there

seberg:
Ah, right.  That changes sorting but this would (also?) need `searchsorted` to take incomparable values into account.

PR #31469, opened 2026-05-19
DOC: clarify np.correlate definition and add proper docstring

### LINKED ISSUE #20090: BUG: numpy.correlate result does not match the documentation
### Describe the issue:

The result of 1D np.correlate has the order reversed compared to what would be expected from the API documentation.

```python
import numpy as np

# From Numpy documentation for nump.correlate in
# https://numpy.org/doc/stable/reference/generated/numpy.correlate.html:
#
# c_{av}[k] = sum_n a[n+k] * conj(v[n])
#
# Thus if a = [1,1] and v = [1,2,3,4,5,6] we would expect c_{av}[0] = 3
# (I.e., c_{av}[0] = a[0]*v[0] + a[1]*v[1]).
#
# But: in Numpy 1.17.4, Python 3.8.10, Linux Mint 20.2, the result is 11, and
# the full output of the code below is
#
# a :  [1 1]
# v :  [1 2 3 4 5 6]
# c_{av} :  [11  9  7  5  3]

a   = np.array ([1,1]);
v   = np.arange (1,7);
cav = np.correlate (a,v, mode='valid');

print ("a      : ", a)
print ("v      : ", v)
print ("c_{av} : ", cav);
```

### Error message:

_No response_

1.17.4 3.8.10 (default, Sep 28 2021, 16:10:42) 
[GCC 9.3.0]

lalitha2410:
### PR summary
Closes #20090. The existing docstring for np.correlate was a single line that didn't explain which definition of cross-correlation NumPy implements. This PR adds a proper docstring with the formula, a note about the alternate definition, parameters, return value, and examples.

Hi, I'm Lalitha, a computer science student. I use NumPy regularly and came across this issue while studying signal processing concepts. I wanted to make my first open source contribution and this seemed like a good place to start.

### AI Disclosure
Used an AI assistant for minor help with formatting and verifying example outputs. The fix and understanding of the issue is my own.

### DIFF
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -1846,7 +1846,56 @@ add_newdoc('numpy._core.multiarray', 'from_dlpack',
     """)
 
 add_newdoc('numpy._core.multiarray', 'correlate',
-    """cross_correlate(a,v, mode=0)""")
+    """
+    correlate(a, v, mode='valid')
+
+    Cross-correlation of two 1-dimensional sequences.
+
+    This function computes the correlation as commonly defined in signal
+    processing contexts::
+
+        c[k] = sum_n a[n+k] * conj(v[n])
+
+    Note that this definition is not universal. Another common definition
+    swaps the roles of the two sequences::
+
+        c[k] = sum_n a[n] * conj(v[n+k])
+
+    which is related to the first by ``c[k] = c[-k]`` (i.e., the result is
+    reversed). NumPy uses the first definition above.
+
+    Parameters
+    ----------
+    a : array_like
+        First input sequence.
+    v : array_like
+        Second input sequence.
+    mode : {'valid', 'full', 'same'}, optional
+        Refer to the `convolve` docstring for full descriptions of the modes.
+        Default is 'valid'.
+
+    Returns
+    -------
+    out : ndarray
+        Discrete cross-correlation of `a` and `v`.
+
+    See Also
+    --------
+    convolve : Discrete, linear convolution of two sequences.
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> np.correlate([1, 2, 3], [0, 1, 0.5])
+    array([3.5])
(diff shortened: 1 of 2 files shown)


seberg:
Thanks, but what you changed is an implementation detail for which the docstring doesn't matter much: it is not public API.
You want to look for the wrapping `def correlate` function and modify it's docstring to work on the issue.

lalitha2410:
Thanks for pointing that out! I've updated the PR to edit the correct file (numpy/_core/numeric.py). Added a line clarifying which definition NumPy uses. Let me know if you'd like any further changes.

numpy/_core/_add_newdocs.py
```diff
@@ -1846,7 +1846,56 @@
...
+    >>> np.correlate([1, 2, 3], [0, 1, 0.5])
+    array([3.5])
+
+    >>> np.correlate([1, 2, 3], [0, 1, 0.5], 'full')
+    array([0.5,  2. ,  3.5,  3. ,  0. ])
+
+    >>> np.correlate([1, 2], [1, 2, 3])
+    array([8, 5])
+
+    """)
```
ngoldbaum: I don't think we should add a detailed docstring for this purely internal function.

numpy/_core/numeric.py
```diff
@@ -757,6 +757,8 @@ def correlate(a, v, mode='valid'):
     .. math:: c'_k = \sum_n a_{n} \cdot \overline{v_{n+k}}
 
     which is related to :math:`c_k` by :math:`c'_k = c_{-k}`.
+    
+    NumPy uses the first definition above, where `a` is shifted by `k`.
```
ngoldbaum: These should be double backticks because this is restructured text, not markdown

lalitha2410: Thanks! Fixed the backticks to double backticks for RST formatting!

mattip:
Superseded by #32533

PR #31473, opened 2026-05-20
Setup compilation cache and use it for building the wheel

luhenry:
### PR summary

On slower platforms like linux-riscv64, it's very beneficial to use ccache to speed up the build. It also benefits other platforms as well.


luhenry:
Validating it works at https://github.com/luhenry/numpy/actions/workflows/wheels.yml?query=branch%3Accache-1

jorenham:

PR #31476, opened 2026-05-20
ENH: Use radixsort also for descending sorts

seberg:
This was skipped in the original PR to keep it a bit more minimal. I opted for just changing the bin discovery as I couldn't measure a difference for sorting a large int16 array.
(One could also reverse the bins later, but this gives the smaller code diff.)

The unused float path is deleted. I doubt we'll have appetite to re-add it soon, so I just deleted it.

The bot thought there was no test with duplicate values, so this adds one for bool+integers.

Used an agent a bit, but the change is so simple...


MaanasArora (approved):
Great, thanks for working on this and the nice new test! I went through in detail and this looks all good to me.

I asked Claude to find issues and it gave me nonsense, which is a good sign :)

ngoldbaum:
Looks good to me!

PR #31479, opened 2026-05-21
TYP: datetime.{[date]time,timedelta} also ArrayLike

scivision:
### PR summary

The Python types for time and duration:
```
      datetime.datetime
      datetime.date
      datetime.timedelta
```
have long been usable as an input to several Numpy NDarray
initializers including ndarray, asarray, atleast_1d, and similar.
However, numpy.typing didn't recognize them as ArraryLike till this PR.

#### AI Disclosure
No AI tools used

### DIFF
--- a/numpy/_typing/_array_like.py
+++ b/numpy/_typing/_array_like.py
@@ -1,3 +1,4 @@
+import datetime
 from collections.abc import Buffer, Callable, Collection
 from typing import TYPE_CHECKING, Any, Protocol, runtime_checkable
 
@@ -52,7 +53,11 @@ type _DualArrayLike[DTypeT: np.dtype, BuiltinT] = (
     | _NestedSequence[BuiltinT]
 )
 
-type ArrayLike = Buffer | _DualArrayLike[np.dtype, complex | bytes | str]
+type ArrayLike = Buffer | _DualArrayLike[np.dtype, complex | bytes |
+                                         datetime.datetime |
+                                         datetime.timedelta |
+                                         datetime.date |
+                                         str]
 
 # `ArrayLike<X>_co`: array-like objects that can be coerced into `X`
 # given the casting rules `same_kind`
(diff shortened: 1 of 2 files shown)


jorenham:
Thanks, I'll soon have a look. 

But note that we don't allow agentic AI here, including copilot

jorenham (commented):
Hmm, I don't think that we should do this:

```
In [1]: np.array(datetime.now())
Out[1]: array(datetime.datetime(2026, 5, 21, 10, 46, 8, 278330), dtype=object)

In [2]: np.array([datetime.now()])
Out[2]: array([datetime.datetime(2026, 5, 21, 10, 46, 25, 131859)], dtype=object)
```

Because as you can see here, numpy doesn't have any special treatmeant for `datetime`; the type is `numpy.object_`, like it would have if you'd pass it any arbitrary, **non-array-like**, Python object.

You might think that adding this at least couldn't hurt, but that's not the case either. Because if you take a look at the mypy_primer output, you see that the current error messages because significantly more complicated because of this. This will make type-checking errors even more to debug, and it's already difficult enough at the moment.

jorenham:
Mypy_primer also shows that error that this "resolves" is one in freqtrade, specifically, in [this line](https://github.com/freqtrade/freqtrade/blob/0d84e80eb77864d39c1212c008fc3aec7fe40d2d/freqtrade/optimize/backtesting.py#L430), where a potential `datetime` is passed to `numpy.isnan`.
However, this mypy error is correct, because passing a `datetime` to `numpy.isnan` will raise an error at runtime:

```
In [1]: np.isnan(datetime.now())
Cell In[1], line 1
----> 1 np.isnan(datetime.now())

TypeError: ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

In summary; this hurts type-safety, increases complexity, and doesn't solve the problem that you're suggesting it would solve.

So thanks for your contribution, but I'm going to close this.

scivision:
OK thank you!

PR #31480, opened 2026-05-21
BENCH: Update sort benchmarks to use new-style args and smaller dtypes

MaanasArora:
Updates the benchmarks to use the new-style sort kwargs, `stable` and `descending`, instead of the older `kind`, aligned with a [comment](https://github.com/numpy/numpy/pull/31345#issuecomment-4435933886) from @charris. This adds support for descending benchmarks and could be useful as we add dtypes such as `object`. cc @seberg

I also added a small class to allow for benchmarking smaller sized arrays. I'm not sure if I went too far with that or whether it is actually necessary, but mainly was curious as I noticed several benchmarks (even for `int16` and `float16`) were being skipped for certain array types, and I remember for `radixsort` I was curious about short/bool performance but they weren't included. Completely open to remove/adjust this to reduce complexity.

I can also post a benchmarks run if needed and this seems to be a good direction.

#### AI Disclosure
No AI tools used.

### DIFF
--- a/benchmarks/benchmarks/bench_function_base.py
+++ b/benchmarks/benchmarks/bench_function_base.py
@@ -280,12 +280,17 @@ class Sort(Benchmark):
         # Using np.sort(...) instead of arr.sort(...) because it makes a copy.
         # This is important because the data is prepared once per benchmark, but
         # used across multiple runs.
-        np.sort(self.arr, stable=stable, descending=descending)
+        if descending:
+            np.sort(self.arr, stable=stable, descending=True)
+        else:
+            # for backward compatibility to NumPy 2.0
+            np.sort(self.arr, stable=stable)
 
     def time_argsort(self, stable, descending, dtype, array_type):
         if descending:
             np.argsort(self.arr, stable=stable, descending=True)
         else:
+            # for backward compatibility to NumPy 2.0
             np.argsort(self.arr, stable=stable)
 
 
--- a/benchmarks/benchmarks/bench_function_base.py
+++ b/benchmarks/benchmarks/bench_function_base.py
@@ -181,7 +181,6 @@ class SortGenerator:
         """
         arr = SortGenerator.ordered_range(size, dtype=dtype)
         rnd = np.random.RandomState(1792364059)
-        np.random.shuffle(arr)
         rnd.shuffle(arr)
         return arr
 
--- a/benchmarks/benchmarks/bench_function_base.py
+++ b/benchmarks/benchmarks/bench_function_base.py
@@ -212,6 +212,83 @@ class SortGenerator:
         return np.array(b)
 
 
+class SortGeneratorSmall:
+    # This is a separate class as for certain small dtypes we
+    # need to repeat elements to get the desired array size.
+
+    # The size of the unsorted area in the "random unsorted area"
+    # benchmarks
+    AREA_SIZE = 100
+    # The size of the "partially ordered" sub-arrays
+    BUBBLE_SIZE = 100
+
+    limits = {
+        'bool': (0, 2),
+        'uint8': (0, 256),
+        'int8': (-128, 128),
+        'int16': (-32768, 32768),
+        'float16': (-1000, 1000),
+    }
+
+    @staticmethod
+    @memoize
+    def arange(dtype):
+        return np.arange(*SortGeneratorSmall.limits[dtype], dtype=dtype)
+
+    @staticmethod
+    @memoize
+    def random(size, dtype, rnd):
+        """
+        Returns a randomly-shuffled array.
+        """
+        arange = SortGeneratorSmall.arange(dtype)
+        rnd = np.random.RandomState(1792364059)
+        arr = rnd.choice(arange, size=size, replace=True)
+        return arr
+
+    @staticmethod
+    @memoize
+    def ordered(size, dtype, rnd):
+        """
+        Returns an ordered array.
+        """
+        arange = SortGeneratorSmall.arange(dtype)
+        return np.repeat(arange, size // arange.size + 1)[:size]
+
+    @staticmethod
+    @memoize
+    def reversed(size, dtype, rnd):
+        """
+        Returns an array that's in descending order.
+        """
+        arange = SortGeneratorSmall.arange(dtype)
+        return np.repeat(arange[::-1], size // arange.size + 1)[:size]
+
+    @staticmethod
+    @memoize
+    def uniform(size, dtype, rnd):
+        """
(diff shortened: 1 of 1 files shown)


MaanasArora:
Sorry, totally overcomplicated this before, pushed a change merging the two classes - seems much more palatable to allow small dtypes now!

benchmarks/benchmarks/bench_function_base.py
```diff
@@ -233,25 +265,25 @@ class Sort(Benchmark):
...
         # Using np.sort(...) instead of arr.sort(...) because it makes a copy.
         # This is important because the data is prepared once per benchmark, but
         # used across multiple runs.
-        np.sort(self.arr, kind=kind)
+        np.sort(self.arr, stable=stable, descending=descending)
 
-    def time_argsort(self, kind, dtype, array_type):
-        np.argsort(self.arr, kind=kind)
+    def time_argsort(self, stable, descending, dtype, array_type):
+        np.argsort(self.arr, stable=stable, descending=descending)
```
seberg:
```suggestion
        if descending:
            np.argsort(self.arr, stable=stable, descending=True)
        else:
            np.argsort(self.arr, stable=stable)
```
I am not actually sure we actually should care, but if we do it this way we can compare ascending sorts back in time to NumPy 2.0?

seberg (approved):
Thanks for fixing this!  LGTM, one small suggestion to try and make it that this also works for older NumPy versions, even if I am not sure how much we need to care about it.

benchmarks/benchmarks/bench_function_base.py
```diff
@@ -221,8 +241,20 @@ class Sort(Benchmark):
...
+            'float64',
+            'int64',
+            'float32',
+            'uint32',
+            'int32',
+            'int16',
+            'float16',
+            'uint8',
+            'int8',
+            'bool',
```
seberg, line 246: Should we just add `object` here right now?  Feels like we might as well.  (We could add more, but object is the most interesting one.)

MaanasArora: Would be nice, but it doesn't support `descending` right now, so the benchmarks would fail. Unless we special case, perhaps in the other PR?

benchmarks/benchmarks/bench_function_base.py
MaanasArora: Applied, thanks!

MaanasArora:
Thank you for reviewing, applied your suggestion for both sort and argsort!

benchmarks/benchmarks/bench_function_base.py
seberg: Ah, right... nvm, easier to follow up there then.

benchmarks/benchmarks/bench_function_base.py
```diff
@@ -221,8 +241,20 @@ class Sort(Benchmark):
     params = [
         # In NumPy 1.17 and newer, 'merge' can be one of several
         # stable sorts, it isn't necessarily merge sort.
```
seberg:
```suggestion
```

seberg (approved):

benchmarks/benchmarks/bench_function_base.py
```diff
@@ -171,22 +191,22 @@ def ordered(size, dtype, rnd):
...
-            raise SkipNotImplemented("Cannot construct arange for this size.")
+        if dtype not in SortGenerator.small_limits:
+            try:
+                with np.errstate(over="raise"):
+                    res = np.dtype(dtype).type(size - 1)
+            except (OverflowError, FloatingPointError):
+                raise SkipNotImplemented("Cannot construct arange for this size.")
 
-        return np.arange(size - 1, -1, -1, dtype=dtype)
+        return SortGenerator.ordered_range(size, dtype=dtype)[::-1]
```
seberg, line 201:
```suggestion
        return SortGenerator.ordered_range(size, dtype=dtype)[::-1].copy()
```
ooops, should have gone through things more careful, so one more suggestion to apply :).

I guess the bubbles are sorted always ascending... but probably not too bad.

MaanasArora: Ah yeah, sorry missed this, thanks!

True about the bubbles - perhaps worth reversing the final array at call site if descending?

seberg: Yeah, sure, why not... I guess passing on `descending` into the prep functions is trivial after all.

seberg: Well, actually if you just swap the whole array after the setup (which makes most sense), we should move the copy there :).

MaanasArora: Oh yes sorry, I pushed a bug! Fixing...

MaanasArora: Nvm, was confused... ok, I moved the copy after setup, hopefully this works!

benchmarks/benchmarks/bench_function_base.py
```diff
@@ -171,22 +191,22 @@ def ordered(size, dtype, rnd):
...
-            with np.errstate(over="raise"):
-                res = dtype.type(size - 1)
-        except (OverflowError, FloatingPointError):
-            raise SkipNotImplemented("Cannot construct arange for this size.")
+        if dtype not in SortGenerator.small_limits:
+            try:
+                with np.errstate(over="raise"):
+                    res = np.dtype(dtype).type(size - 1)
+            except (OverflowError, FloatingPointError):
+                raise SkipNotImplemented("Cannot construct arange for this size.")
```
MaanasArora:
```suggestion
```
Hmm, seems like we might not actually need this anymore? (Which is nice...)

benchmarks/benchmarks/bench_function_base.py
```diff
@@ -153,13 +153,33 @@ class SortGenerator:
...
     @staticmethod
     @memoize
     def random(size, dtype, rnd):
         """
         Returns a randomly-shuffled array.
         """
-        arr = np.arange(size, dtype=dtype)
+        arr = SortGenerator.ordered_range(size, dtype=dtype)
         rnd = np.random.RandomState(1792364059)
         np.random.shuffle(arr)
```
MaanasArora: Not sure if this existing code, duplicate, unfrozen shuffle and redeclaring rnd is intentional?

seberg:
```suggestion
```
fair to assume someone wanted to stabilize the run and got it wrong.

seberg:
Nice, thanks for the cleanups, especially!

PR #31481, opened 2026-05-21
DOCS: rewrite a sentence on the basics section

realihorrud:
### PR summary

There was an error, at least it seemed to me, you can see the changes they're pretty straightforward. To cut a long story short, there was some async with the sentence above the example. 

Then, hello, everyone. First of all, thanks for all of the contributors of NumPy. Second of all I use it daily, because I'm switching to Machine Learning / DS role, even though I have already worked with PHP and all around it, but Python community is much powerful (no doubts) :-)


MaanasArora:
Sorry, I don't think this is correct - the array in question does have two axes of different sizes. Unless you mean to propose some other change, we may just close this.

realihorrud:
@MaanasArora , yeah, I just reviewed and chances are you're right, that first axis (rows) has length of 2 and the second axis (columns) has length of 3. Indeed, the documentation was correct, then I think we can close this PR, thanks for clarifying it!

PR #31482, opened 2026-05-21
TYP: Fix ``ma.[as[any]]array`` fallback overloads

### LINKED ISSUE #31451: TYP: unexpected `var-annotated` error with masked array
### Describe the issue:

`mypy` raises `var-annotated` on `x1`, but `x2` seems fine.

- x-ref: https://github.com/python/typeshed/pull/15472#issuecomment-4470315371
- cc @jorenham found this while debugging the `list.{count, index, remove}` PR in typeshed
- This may possibly be related to mypy's long standing issue with bidirectional inference (https://github.com/python/mypy/issues/19304)

```python
import numpy as np

y: list[float] = [-1, -1, 10, 10.1, 10.2]
x1 = np.ma.array(y, mask=[True, True, False, False, False])
x2 = np.ma.array([-1, -1, 10, 10.1, 10.2], mask=[True, True, False, False, False])
```

### Error message:

```shell
tmp.py:4: error: Need type annotation for "x1"  [var-annotated]
```

- Python 3.14.5
- numpy               2.4.5

- mypy 2.2.0+dev.11008008752aa10000a6230322fc4ebe3eb2b456 (compiled: no)
- mypy 2.1.0 (compiled: yes)

_No response_

jorenham:
### PR summary

Fixes #31451

#### AI Disclosure

N/A

### DIFF
--- a/numpy/ma/core.pyi
+++ b/numpy/ma/core.pyi
@@ -2874,7 +2874,7 @@ def array[ScalarT: np.generic](
     ndmin: int = 0,
 ) -> _MaskedArray[ScalarT]: ...
 @overload
-def array[ScalarT: np.generic](
+def array(
     data: object,
     dtype: DTypeLike | None = None,
     copy: bool = False,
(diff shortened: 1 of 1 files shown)


seberg (approved):
Thanks, can't say I understand what was wrong, but this makes sense and matches `def array()`.

seberg:
Marking for backporting, seems like a simple bug fix to me and we are in the pre-2.5 phase anyway.

PR #31484, opened 2026-05-21
Improve error messages for invalid norm orders in linalg.norm

goingforstudying-ctrl:
Improve error messages for invalid norm orders in linalg.norm

Add valid options to error messages when users pass invalid string values for the ord parameter in numpy.linalg.norm for both vectors and matrices.

Before:
- Invalid norm order 'invalid' for vectors
- Invalid norm order for matrices.

After:
- Invalid norm order 'invalid' for vectors. Valid string values for vectors are: None, 'fro', 'f', 'nuc'.
- Invalid norm order 'invalid' for matrices. Valid string values for matrices are: None, 'fro', 'f', 'nuc'.

This helps users understand what values are accepted without needing to consult the documentation.


jorenham:

PR #31491, opened 2026-05-23
ENH: handle masked arrays on histogram.

### LINKED ISSUE #10019: histogram doesn't always work with masked arrays
numpy.histogram doesn't always work with masked arrays where the masked data is np.nans.

```
import numpy as np
x=np.random.randn(100)
x[50]=np.nan
xm=np.ma.masked_array(x, mask=np.isnan(x))

result=np.histogram(xm,bins=10)
print(result[0])

result=np.histogram(xm,bins=range(-3,3))
print(result[0])
```
The last case doesn't work for me 

The problem seems to be that `asarray `on line 647 in master returns an unmasked array with nans and that a.min() and a.max() returns nan on line 662.

changing `asarray` to `asanyarray` works for me but I don't know if it has any other implications.

I also read that histogram might not be designed to work with masked arrays

JosephMehdiyev:
Add logic to handle masked data and masked weights on histogram.
fixes https://github.com/numpy/numpy/issues/10019
### PR summary

This PR makes weights and the input data to be possibly passed with masked arrays. 
If both the weights and input are masked, it will combine the masked data as I feel like this is more natural. It will not raise errors if masks are not matched.
### Other comments
note that this PR does not add masked support for histogramdd. I will open other PRs for it (hopefully).
#### AI Disclosure

the lines 297-302 was generated by Claude (because of lazyness)
tests were partially written by claude
Claude was also used to review the code

### DIFF
--- a/numpy/lib/_histograms_impl.py
+++ b/numpy/lib/_histograms_impl.py
@@ -277,7 +277,7 @@ _hist_bin_selectors = {'stone': _hist_bin_stone,
 
 def _ravel_and_check_weights(a, weights):
     """ Check a and weights have matching shapes, and ravel both """
-    a = np.asarray(a)
+    a = np.asanyarray(a)
 
     # Ensure that the array is a "subtractable" dtype
     if a.dtype == np.bool:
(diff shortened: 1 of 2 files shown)


JosephMehdiyev:
hmmmm, something is very wrong about CI

PR #26650, opened 2024-06-09
BUG: add exception to searchsorted incompatible inputs

luxedo:
Closes https://github.com/numpy/numpy/issues/24032 by throwing an exception when `searchsorted` receives incompatible types

PR #31495, opened 2026-05-24
BUG: searchsorted rejects incompatible dtype pairs (gh-24032)

eendebakpt:
### PR summary
Without an explicit check, ``np.searchsorted([1, 2, 3], ['1'])`` and similar mismatched-type calls were silently promoted to a string dtype and returned wrong indices. Mirror ``np.less.resolve_dtypes`` in C inside ``PyArray_SearchSorted`` to raise ``TypeError`` for unsupported pairings, with a fast-path skip for same-DType and number/number pairs.

This is an alternative to #26650.

Benchmarks:
```
=== main  (bf8007a59e (parent of fix branch)) ===
numpy 2.5.0.dev0+git20260524.06eb78b
np.searchsorted (Python wrapper):
  small int+int ndarrays (3+2)          :  1004.8 ns
  medium int+int (1000+3)               :  1036.0 ns
  array+list (10+2)                     :  1396.0 ns
  list+list (3+2)                       :  2434.7 ns
  float+int (3+2)                       :  1238.1 ns
  str+str                               :  1244.9 ns
  int+scalar                            :  1153.5 ns
arr.searchsorted (no wrapper):
  small int+int (3+2)                   :   356.7 ns
  medium int+int (1000+3)               :   374.6 ns

=== branch  (06eb78babc fix/searchsorted-incompatible-types-c) ===
numpy 2.5.0.dev0+git20260524.06eb78b
np.searchsorted (Python wrapper):
  small int+int ndarrays (3+2)          :   998.1 ns
  medium int+int (1000+3)               :  1029.7 ns
  array+list (10+2)                     :  1394.2 ns
  list+list (3+2)                       :  2438.4 ns
  float+int (3+2)                       :  1234.5 ns
  str+str                               :  1229.0 ns
  int+scalar                            :  1158.8 ns
arr.searchsorted (no wrapper):
  small int+int (3+2)                   :   355.5 ns
  medium int+int (1000+3)               :   372.9 ns
```
So performance neutral.

#### AI Disclosure

Claude was used to implement for first C version. Additional prompts have been used to add the fast path and reduce the diff vs main.

### DIFF
--- a/benchmarks/benchmarks/bench_searchsorted.py
+++ b/benchmarks/benchmarks/bench_searchsorted.py
@@ -33,3 +33,17 @@ class SearchSorted(Benchmark):
 
     def time_searchsorted(self, array_size, n_queries, query_order, use_sorter, seed):
         np.searchsorted(self.arr, self.queries, sorter=self.sorter)
+
+
+class SearchSortedListNeedle(Benchmark):
+    # gh-31495: the needle is a python list, so the runtime is dominated by
+    # dtype discovery/conversion of `v`, which must only happen once.
+    params = [[1_000, 1_000_000]]
+    param_names = ['n_queries']
+
+    def setup(self, n_queries):
+        self.arr = np.arange(100)
+        self.queries = list(range(n_queries))
+
+    def time_searchsorted_list(self, n_queries):
+        self.arr.searchsorted(self.queries)
--- a/benchmarks/benchmarks/bench_searchsorted.py
+++ b/benchmarks/benchmarks/bench_searchsorted.py
@@ -33,17 +33,3 @@ class SearchSorted(Benchmark):
 
     def time_searchsorted(self, array_size, n_queries, query_order, use_sorter, seed):
         np.searchsorted(self.arr, self.queries, sorter=self.sorter)
-
-
-class SearchSortedListNeedle(Benchmark):
-    # gh-31495: the needle is a python list, so the runtime is dominated by
-    # dtype discovery/conversion of `v`, which must only happen once.
-    params = [[1_000, 1_000_000]]
-    param_names = ['n_queries']
-
-    def setup(self, n_queries):
-        self.arr = np.arange(100)
-        self.queries = list(range(n_queries))
-
-    def time_searchsorted_list(self, n_queries):
-        self.arr.searchsorted(self.queries)
--- a/doc/release/upcoming_changes/31495.change.rst
+++ b/doc/release/upcoming_changes/31495.change.rst
@@ -1,0 +1,9 @@
+``searchsorted`` rejects mixed string/non-string inputs
+-------------------------------------------------------
+`numpy.searchsorted` (and `numpy.ndarray.searchsorted`) now raises a
+``TypeError`` when one input is a string array and the other a non-string
+array, for example ``np.searchsorted([1, 2, 3], ["1"])``.  Previously both
+inputs were silently promoted to a string dtype, so the values were compared
+as strings and incorrect indices were returned.  Pairings that promote to a
+non-legacy-string dtype (such as ``StringDType`` or ``object``) are
+unaffected.
--- a/numpy/_core/src/multiarray/item_selection.c
+++ b/numpy/_core/src/multiarray/item_selection.c
@@ -2105,27 +2105,34 @@ PyArray_SearchSorted(PyArrayObject *op1, PyObject *op2,
     PyArray_ArgBinSearchFunc *argbinsearch = NULL;
     NPY_BEGIN_THREADS_DEF;
 
-    /* Discover op2's natural dtype, then promote.  Reject the specific
-     * string-vs-non-string mixture that would otherwise silently promote
-     * to a string dtype and return wrong indices (gh-24032). */
+    /* Reject the string-vs-non-string mixture that would otherwise silently
+     * promote to a string dtype and return wrong indices (gh-24032).  The
+     * check is on the dtypes, so it is applied independently of `v`'s size:
+     * an explicitly-typed but empty `v` (e.g. an empty integer array searched
+     * in a string `a`) is rejected just like a non-empty one, matching how
+     * comparison ufuncs such as ``np.less`` reject the same dtype pairing.
+     *
+     * The only exemption is when no dtype can be inferred from `v` at all
+     * (dtype_v == NULL, e.g. an empty list): such a `v` carries no type intent
+     * and always yields an empty result, so it must keep working. */
     PyArray_Descr *dtype_a = PyArray_DESCR(op1);
     PyArray_Descr *dtype_v = NULL;
     if (PyArray_DTypeFromObject(op2, NPY_MAXDIMS, &dtype_v) < 0) {
         return NULL;
     }
-    if (dtype_v == NULL) {
-        dtype_v = PyArray_DescrFromType(NPY_DEFAULT_TYPE);
-    }
-    if (PyTypeNum_ISSTRING(dtype_a->type_num)
-            != PyTypeNum_ISSTRING(dtype_v->type_num)) {
+    if (dtype_v != NULL
+            && PyTypeNum_ISSTRING(dtype_a->type_num)
+                != PyTypeNum_ISSTRING(dtype_v->type_num)) {
         PyErr_Format(PyExc_TypeError,
                 "Incompatible types for searching: a (%S) and v (%S)",
                 dtype_a, dtype_v);
         Py_DECREF(dtype_v);
         return NULL;
     }
-    dtype = PyArray_PromoteTypes(dtype_a, dtype_v);
-    Py_DECREF(dtype_v);
+    Py_XDECREF(dtype_v);
+
+    /* Find common type (unchanged historical behavior for the accepted cases) */
+    dtype = PyArray_DescrFromObject((PyObject *)op2, dtype_a);
     if (dtype == NULL) {
         return NULL;
     }
--- a/numpy/_core/src/multiarray/item_selection.c
+++ b/numpy/_core/src/multiarray/item_selection.c
@@ -1,4 +1,5 @@
 #define NPY_NO_DEPRECATED_API NPY_API_VERSION
+#define _UMATHMODULE
 #define _MULTIARRAYMODULE
 
 #define PY_SSIZE_T_CLEAN
@@ -8,6 +9,7 @@
 
 #include "numpy/arrayobject.h"
 #include "numpy/arrayscalars.h"
+#include "numpy/ufuncobject.h"
 
 #include "numpy/npy_math.h"
 #include "numpy/npy_cpu.h"
@@ -17,6 +19,7 @@
 
 
 #include "npy_static_data.h"
+#include "npy_import.h"
 #include "common.h"
 #include "dtype_transfer.h"
 #include "dtypemeta.h"
@@ -33,6 +36,7 @@
 #include "alloc.h"
 #include "arraytypes.h"
 #include "array_coercion.h"
+#include "dispatching.h"
 #include "simd/simd.h"
 
 static NPY_GCC_OPT_3 inline int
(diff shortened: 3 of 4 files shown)


charris:
Needs rebase. @ngoldbaum Any chance this was fixed as a byproduct of your recent PR?

seberg:
No, there was no recent fix around this, although if we introduce a new arraymethod (or ideally full gufunc) based approach here, that might give us a path forward.

Thanks Pieter for pushing it into C.  I can't say I love using the `less` promoter though (it makes some sense in Python, but to me seems awkward in C; promotion to string is just a terrible odd-one-out).
So, I think it may be saner to just reject the string promotion explicitly.  In principle, it would be nice to use a common helper in `PyUFunc_SimpleBinaryComparisonTypeResolver`, but I am not sure it is worthwhile `np.less` supports mixed int/uint paths as well there (searchsorted doesn't and currently goes to float64, which is technically wrong but maybe a different issue).

numpy/_core/tests/test_multiarray.py
```diff
@@ -3314,6 +3314,13 @@ class A(np.ndarray):
         assert_(not isinstance(a.searchsorted(b, 'left', s), A))
         assert_(not isinstance(a.searchsorted(b, 'right', s), A))
 
+    def test_searchsorted_incompatible_inputs(self):
+        # gh-24032: incompatible types must raise rather than silently
+        # promote to a string dtype and return wrong indices.
+        assert_raises(TypeError, np.searchsorted, [1, 2, 3], ["1"])
+        assert_raises(TypeError, np.searchsorted, ["1", "2", "3"], [2])
+        assert_raises(TypeError, np.searchsorted, [b"1", b"2", b"3"], 1)
```
seberg: Thanks for the update! Could you add a test that:
```
searchsorted(np.array(["a", "b"], dtype="T"), ["b"])
```
still works?  Because I _suspect_ this would make it fail.
(Another thing that might looks like it would fail here is object dtype array with the second needle array being a string one.)

My guess is: Promote first and then reject things only if the promoted type ended up with the NumPy string.
We may even want to limit this to NumPy input dtypes that aren't string (in theory a non-NumPy string dtype could correctly promote to string, although at that point we are very theoretical and need to cut corners one way or another...).

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -2105,8 +2105,27 @@ PyArray_SearchSorted(PyArrayObject *op1, PyObject *op2,
...
+    PyArray_Descr *dtype_a = PyArray_DESCR(op1);
+    PyArray_Descr *dtype_v = NULL;
+    if (PyArray_DTypeFromObject(op2, NPY_MAXDIMS, &dtype_v) < 0) {
+        return NULL;
+    }
+    if (dtype_v == NULL) {
+        dtype_v = PyArray_DescrFromType(NPY_DEFAULT_TYPE);
+    }
+    if (PyTypeNum_ISSTRING(dtype_a->type_num)
+            != PyTypeNum_ISSTRING(dtype_v->type_num)) {
```
KRRT7: This check rejects empty `v` inputs for string arrays because dtype discovery falls back to a non-string dtype even though no comparisons are needed. For example, `np.searchsorted(["a", "b"], [])` and `np.searchsorted([b"a", b"b"], [])` currently return an empty integer array, but this branch raises `TypeError` (`v` is discovered as `float64`). The check should be skipped for empty `v` or delayed until after `op2` has been converted so empty inputs preserve the existing behavior.

KRRT7 (commented):
This introduces a compatibility regression for empty search values.

ngoldbaum:
I asked Iason to look this over because he's interested in working on `searchsorted` and this seems related and landable.

numpy/_core/tests/test_multiarray.py
ikrommyd, line 1: Quick fly by comment. I'll look at the C changes later today. `with pytest.raises` is preferable over `assert_raises` even though a lot of old code still uses it. I'd also add a `match` in the raises assertions to ensure no other type error coming from somewhere else is ever raised.

numpy/_core/tests/test_multiarray.py
seberg: Since I got a ping on this PR, as a note, I don't think this has been addressed yet?

(Once that is, I don't have any further bigger picture thoughts here, it looks all good.)

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -2105,8 +2105,34 @@ PyArray_SearchSorted(PyArrayObject *op1, PyObject *op2,
...
+        PyErr_Format(PyExc_TypeError,
+                "Incompatible types for searching: a (%S) and v (%S)",
+                dtype_a, dtype_v);
+        Py_DECREF(dtype_v);
+        return NULL;
+    }
+    Py_XDECREF(dtype_v);
+
+    /* Find common type (unchanged historical behavior for the accepted cases) */
+    dtype = PyArray_DescrFromObject((PyObject *)op2, dtype_a);
```
ikrommyd: `PyArray_DescrFromObject` is now called twice unconditionally. I'm seeing a 50% slowdown for example on something like this
```python
In [1]: import numpy as np
   ...: big = list(range(1_000_000))
   ...: a = np.arange(100)
   ...: %timeit a.searchsorted(big)
35.9 ms ± 281 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
I'm not sure if that's important or not. I think you can do something like this to avoid the double unconditional call.
```suggestion
    if (PyArray_DTypeFromObject(op2, NPY_MAXDIMS, &dtype_v) < 0) {
        return NULL;
    }
    if (dtype_v != NULL) {
        if (PyTypeNum_ISSTRING(dtype_a->type_num)
                != PyTypeNum_ISSTRING(dtype_v->type_num)) {
            PyErr_Format(PyExc_TypeError,
                    "Incompatible types for searching: a (%S) and v (%S)",
                    dtype_a, dtype_v);
            Py_DECREF(dtype_v);
            return NULL;
        }
        /* Find common type; reuse the discovered dtype rather than scanning
         * `v` a second time with PyArray_DescrFromObject. */
        dtype = PyArray_PromoteTypes(dtype_a, dtype_v);
        Py_DECREF(dtype_v);
    }
    else {
        /* No dtype inferable from `v` (e.g. empty list): old promotion path */
        dtype = PyArray_DescrFromObject((PyObject *)op2, dtype_a);
    }
```

ikrommyd (commented):
Overall looking good I think. I have one suggestion regarding a performance regression for large lists for example here. It also needs a release note.

numpy/_core/src/multiarray/item_selection.c
seberg: Sorry, to repeat, but I think it might get lost easily.  Maybe I am just missing it, but I think my comment from above still applies: You probably need the promotion/`ResultType` first.
To make sure `np.array(["1", "2", "3"], dtype="T").searchsorted("2")` works.

eendebakpt: > `PyArray_DescrFromObject` is now called twice unconditionally. I'm seeing a 50% slowdown for example on something like this

Good point. I updated the code and added an ASV benchmark. Output:
```
main (2.6.0.dev0+git20260811.5799ed4):


· Discovering benchmarks
· Running 4 total benchmarks (1 commits * 1 environments * 4 benchmarks)
[ 0.00%] ·· Benchmarking existing-pyC__envs_env313_Scripts_python.exe
[12.50%] ··· Running (bench_searchsorted.SearchSortedListNeedle.time_searchsorted_list--)....
[62.50%] ··· ...earchSortedListNeedle.time_searchsorted_list                ok
[62.50%] ··· =========== ============
              n_queries              
             ----------- ------------
                 1000      52.8±3μs  
               1000000    54.6±0.3ms 
             =========== ============

[75.00%] ··· ...dScalarNeedle.time_searchsorted_scalar_float       1.35±0.05μs
[87.50%] ··· ...tedScalarNeedle.time_searchsorted_scalar_int          950±50ns
[100.00%] ··· ...tedScalarNeedle.time_searchsorted_scalar_str       1.00±0.04μs
PR #31495 (2.5.0.dev0+git20260804.43368b0):


· Discovering benchmarks
· Running 4 total benchmarks (1 commits * 1 environments * 4 benchmarks)
[ 0.00%] ·· Benchmarking existing-pyC__envs_env313_Scripts_python.exe
[12.50%] ··· Running (bench_searchsorted.SearchSortedListNeedle.time_searchsorted_list--)....
[62.50%] ··· ...earchSortedListNeedle.time_searchsorted_list                ok
[62.50%] ··· =========== ============
              n_queries              
             ----------- ------------
                 1000     47.5±0.2μs 
               1000000    51.5±0.8ms 
             =========== ============

[75.00%] ··· ...dScalarNeedle.time_searchsorted_scalar_float       1.30±0.05μs
[87.50%] ··· ...tedScalarNeedle.time_searchsorted_scalar_int           900±0ns
[100.00%] ··· ...tedScalarNeedle.time_searchsorted_scalar_str          1.10±0μs
```
So the regression has been addressed.

numpy/_core/tests/test_multiarray.py
eendebakpt: Thanks for pinging. I added the case as a regression test to `test_searchsorted_string_promotions_still_work`

benchmarks/benchmarks/bench_searchsorted.py
ikrommyd, line 1: Some maintainer should weigh in on this one but I personally wouldn't have added this benchmark to numpy main. I would have only used it for personal benchmarking. I don't know what the policy/usual attitude is towards benchmarks.

ikrommyd (commented):
Overall looks good to me. I only have a comment about the benchmark (not sure about it). The PR is still labeled as "backport candidate" but this is a small behavior change. Arguably it's a bug fix but it still changes what searchsorted accepts and doesn't accept so I don't know if it should be backported. I'm pasting below only two minor edge case comments found by fable:
1. The old path (`handle_promotion` inside discovery) fell back to object on *any* promotion error except FutureWarning; the new code only catches `DTypePromotionError`. The divergence is reachable only with an exotic user DType whose `common_dtype` raises something else, and propagating that error is arguably more correct anyway.
2. Promotion grouping changed: old folded `dtype_a` element-by-element, new promotes `dtype_a` with the pre-promoted `v` dtype. NumPy promotion isn't perfectly associative (e.g. `uint16` haystack with `[np.int8(1), np.float16(2)]` used to search as float64, now float32), but results in my battery were identical, and the new grouping matches what `np.less(a, np.asarray(v))` would resolve — arguably more consistent.

doc/release/upcoming_changes/31495.change.rst
```diff
@@ -0,0 +1,9 @@
+``searchsorted`` rejects mixed string/non-string inputs
+-------------------------------------------------------
+`numpy.searchsorted` (and `numpy.ndarray.searchsorted`) now raises a
+``TypeError`` when one input is a string array and the other a non-string
+array, for example ``np.searchsorted([1, 2, 3], ["1"])``.  Previously both
+inputs were silently promoted to a string dtype, so the values were compared
+as strings and incorrect indices were returned.  Pairings that promote to a
+non-legacy-string dtype (such as ``StringDType`` or ``object``) are
+unaffected.
```
ikrommyd, line 9: This could be a bit shorter. It's too big imo for such an issue.

benchmarks/benchmarks/bench_searchsorted.py
seberg: Yeah, I think I'll agree this time.  Pretty niche thing, so maybe we can avoid bloating the benchmarks for it.

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -2105,11 +2105,66 @@ PyArray_SearchSorted(PyArrayObject *op1, PyObject *op2,
...
+        if (dtype == NULL) {
+            /*
+             * The dtype discovery machinery falls back to object on
+             * promotion failure (e.g. datetime64 vs. string); mirror that
+             * so these pairings behave exactly as before.
+             */
+            if (PyErr_ExceptionMatches(npy_static_pydata.DTypePromotionError)) {
+                PyErr_Clear();
+                dtype = PyArray_DescrFromType(NPY_OBJECT);
+            }
```
seberg, line 2141: Puh, OK... Can keep it for this PR, but this seems to me like we should just propagate that error int he future.  It's also a `TypeError` which we should end up usually with here.  I can think of crazy cases that should fail, like:
```
np.searchsorted(np.ones(3, dtype="i,i"), np.ones(3, dtype="i,"))
```
but they seems so crazy, I am not sure it even makes sense to go through deprecation.  Promoting everything to `object` isn't something ufuncs do silently normally, and neither should `searchsorted` really.

seberg (commented):
Thanks also @eendebakpt LGTM as well.  I think I agree with Ianos to just skip the new benchmark this time.

PR #31503, opened 2026-05-25
ENH: use PyMem_RawMalloc instead of system allocator for better performance on free-threading

kumaraditya303:
### PR summary

Fixes #28454

This PR replaces usage of system allocator `malloc` with `PyMem_RawMalloc` and friends for better performance on free-threading where mimalloc is used as the underlying allocator.

#### AI Disclosure

I used Claude to find all the places where the replacement makes sense.

### DIFF
--- a/doc/release/upcoming_changes/31503.change.rst
+++ b/doc/release/upcoming_changes/31503.change.rst
@@ -1,4 +1,4 @@
 NumPy's internal memory allocations now use ``PyMem_RawMalloc``
 ---------------------------------------------------------------
 NumPy's internal memory allocations now use ``PyMem_RawMalloc`` instead of
-``malloc``and can be tracked by ``tracemalloc``.
+``malloc`` and can be tracked by ``tracemalloc``.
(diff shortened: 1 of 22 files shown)


ngoldbaum (approved):
Overall looks good. I'm approving because this is mergeable as-is but I also have some small followup suggestions:

* Can you add a release note fragment that we are using `PyMem_RawMalloc` for internal allocators that this will show up in `tracemalloc` statistics?
* Claude points out that you missed the uses in `_simd_convert.inc`:
    https://github.com/numpy/numpy/blob/cf16ec13dc456c3debccc46578a4f2535be2ea04/numpy/_core/src/_simd/_simd_convert.inc#L63

@charris are you OK with backporting this to the 2.5 branch with the release note? I don't think this is dangerous to backport and it'd be nice to include it with all the other free-threading scaling improvements for 2.5.

ngoldbaum:
I'm going to merge this to make sure it gets included in the 2.5.0rc1. Thanks @kumaraditya303!

PR #24018, opened 2023-06-22
ENH: Use Highway's VQSort on AArch64

Mousius:
Introduces Highway, a SIMD library from Google. Highway has it's own dispatch mechanism and SIMD flavour (which adds padding) which may be a good replacement for NumPy intrinsics. Highway also has its own set of vectorised Math routines we could bolt on.

For this patch, I've integrated the VQSort algorithm only on AArch64 so as not to impact other architectures. The VQSort algorithms performance is comparable to the AVX512 algorithms, according to this report: https://github.com/Voultapher/sort-research-rs/blob/main/writeup/intel_avx512/text.md

Performance improvements in the NumPy benchmark suite are as follows:
```
       before           after         ratio
     [f4b52d53]       [f08058d4]
     <main-setuppy>       <highway-sort>
+     53.0±0.08μs        221±0.4μs     4.18  bench_function_base.Sort.time_sort('quick', 'int64', ('ordered',))
+      84.8±0.3μs        222±0.3μs     2.62  bench_function_base.Sort.time_sort('quick', 'int64', ('reversed',))
+      89.0±0.2μs        197±0.2μs     2.21  bench_function_base.Sort.time_sort('quick', 'float64', ('ordered',))
+     51.5±0.01μs       80.0±0.1μs     1.55  bench_function_base.Sort.time_sort('quick', 'uint32', ('ordered',))
+     52.0±0.07μs       80.3±0.2μs     1.55  bench_function_base.Sort.time_sort('quick', 'int32', ('ordered',))
+       148±0.3μs        197±0.2μs     1.33  bench_function_base.Sort.time_sort('quick', 'float64', ('reversed',))
+       471±0.2μs        598±0.5μs     1.27  bench_function_base.Sort.time_argsort('heap', 'float64', ('ordered',))
+       536±0.7μs        644±0.8μs     1.20  bench_function_base.Sort.time_argsort('heap', 'float64', ('reversed',))
+     39.8±0.02μs      47.4±0.05μs     1.19  bench_function_base.Sort.time_argsort('merge', 'int32', ('sorted_block', 1000))
+     39.8±0.01μs      47.4±0.02μs     1.19  bench_function_base.Sort.time_argsort('merge', 'uint32', ('sorted_block', 1000))
+     40.1±0.01μs      47.6±0.04μs     1.19  bench_function_base.Sort.time_argsort('merge', 'int64', ('sorted_block', 1000))
[...]

PR #31504, opened 2026-05-25
ENH: wire Highway VQSelect for np.partition on aarch64/ppc64

dmatth1:
## Summary

  `np.partition` on non-x86 platforms (aarch64, ppc64) has always fallen back to
  scalar introselect. `np.sort` already uses Highway's VQSort on these targets
  (gh-24018); this wires the matching `VQSelectStatic` into `np.partition`.

  x86/x86-64 already has partition acceleration via Intel's `x86-simd-sort`; this
  is purely the non-x86 counterpart.

  ## What changed

  - `quickselect_dispatch` now tries Highway VQSelect on non-x86 before falling
    back to scalar introselect, mirroring the existing x86/non-x86 split in
    `quicksort.hpp`.
  - One dispatch file (`highway_qselect.dispatch.cpp`) covers all widths
    (int16/uint16/int32/uint32/int64/uint64/float32/float64), registered for
    `ASIMD` (aarch64 NEON) and `VSX2` (ppc64). 16-bit ints need only base ASIMD,
    so no separate ASIMDHP target.
  - float16 is excluded — `HWY_HAVE_FLOAT16` gating adds complexity and float16
    partition is uncommon; scalar introselect remains correct there.
  - A minimum-size threshold (`HWY_QSELECT_MIN_N = 16384`) keeps scalar
    introselect for L1-resident arrays where dispatch overhead dominates.
  - Removes the dead `DISABLE_HIGHWAY_OPTIMIZATION` macro.

  `np.argpartition` is **not** covered: VQSelect is keys-only and can't maintain a
  parallel index array.

  ## Threshold rationale

  A size sweep on M1 Pro (N=64–65536, kth=N/2) showed scalar introselect's
  random-pivot variance dominates below 16384 — at N=8192 int64 regresses to
  0.28× and float32 to 0.70×. At N=16384 every type reliably wins (1.2× for
  int64, up to 11× for 16-bit), so a single threshold is used for all widths.

  ## Benchmarks

  M1 Pro NEON, numpy 2.4.4 scalar vs this branch (ns/elem):
  
  | dtype   | n    | k   | scalar | vqselect | speedup |
  |---------|------|-----|--------|----------|---------|
  | float32 | 100K | 10  | 5.0    | 2.3      | 2.2×    |
  | float32 | 1M   | 100 | 6.8    | 2.6      | 2.6×    |
  | float32 | 10M  | 100 | 8.9    | 2.7      | 3.3×    |
  | float64 | 100K | 10  | 8.0    | 3.6      | 2.2×    |
  | float64 | 1M   | 100 | 9.2    | 4.5      | 2.0×    |
  | int32   | 100K | 10  | 6.3    | 1.3      | 4.9×    |
  | int32   | 1M   | 100 | 5.8    | 1.5      | 3.8×    |
  | int64   | 1M   | 100 | 5.9    | 2.9      | 2.0×    |

  16-bit types reach 8–12×. Below the 16384 threshold, parity (falls back to
  scalar). ppc64/VSX2 is wired but unmeasured.

  ## Testing

  New `np.partition` correctness tests across all dispatched dtypes, kth
  positions (min/median/max), threshold boundaries, and NaN ordering, plus
  on-target (`_native`) tests that run on aarch64/ppc64. Verified locally on
  Apple Silicon (aarch64): full build clean, new tests + the existing
  partition/sort suite pass (12707 passed).

### DIFF
--- a/doc/release/upcoming_changes/31504.performance.rst
+++ b/doc/release/upcoming_changes/31504.performance.rst
@@ -1,0 +1,26 @@
+``np.partition`` is now faster on aarch64 and ppc64
+----------------------------------------------------
+`numpy.partition` on non-x86 platforms (aarch64, ppc64) now uses
+Highway's ``VQSelectStatic`` instead of the scalar quickselect.
+Applies to arrays with 16 384 or more elements of ``int16``,
+``uint16``, ``int32``, ``uint32``, ``int64``, ``uint64``, ``float32``,
+or ``float64`` dtype.
+
+Measured speedups vs scalar introselect on Apple M1 Pro (NEON, aarch64),
+arrays of 100k–1M elements, kth = N/2:
+
+* ``float32``: 2.2–5.0×
+* ``float64``: 1.8–2.7×
+* ``int32``: 6.4×
+* ``uint32``: 4.2–7.0×
+* ``int64``: 2.4–2.9×
+* ``uint64``: 2.5–2.9×
+* ``int16``: 9.1–11.6×
+* ``uint16``: 8.1–12.3×
+
+Speedup varies with array size and kth position; the lower end of each
+range is for N=1M and the upper for N=100k (or vice versa for float32).
+ppc64 (VSX2) benchmarks have not yet been measured.
+
+`numpy.argpartition` is not affected; Highway VQSelect operates on a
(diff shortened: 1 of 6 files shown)


jorenham:

Issue #31424, opened 2026-05-13
Mid-term follow-ups for sorts


seberg:
We now have a new setup for sorts which allows us to have `descending=True`.  This is interesting also for partial (arg)-sorts and lexsort.
So there are a few things that would be nice to make progress on (not for 2.5 but 2.6):
* Use new sorts for `lexsort` and ideally implement descending
* The new setup can be broadened to partitioning in a very similar vain (I guess it may be slightly awkward that we have the array size and the partition size, but we can pass those just like in a gufunc).  This will make partitioning nicely available to user-dtypes *and* allow `descending=True`.
  * `descending=True` is important as a stepping stone to `top_k` if we want to implement it via partitioning.  Since we would want the smallest or largest values but pushing NaNs to the back.

After this, I hope we can clean up a lot of old code paths :).

PR #31511, opened 2026-05-26
ENH: Descending partitioning

MaanasArora:
Draft to add a `descending` kwarg to `np.partition` and `np.argpartition`, analogous to the sorting functions, thus addresses part of #31424. Not sure if the `SELECTKIND` changes constitute an ABI or C-API break, but analogous to the sorts as well. cc @seberg, thanks!

Note that this does not create new-style ArrayMethod registration analogous to the sorts, but rather updates the (apparently fully internal) `partition_map` and continues using the old-style `PartitionFunc`. (In fact, the `PyArray_(Arg)Partition` functions are unchanged!) Hopefully this is minimal enough and we can just add new-style ArrayMethods (and thus enable user-dtypes to use partitions) in follow-up? (This would then also unblock work for `top_k`.)

Still needs docs, tests, and a release note!

#### AI Disclosure
No AI used.

### DIFF
--- a/doc/release/upcoming_changes/31511.new_feature.rst
+++ b/doc/release/upcoming_changes/31511.new_feature.rst
@@ -1,5 +1,5 @@
 New `descending` keyword argument for ``numpy.partition`` and ``numpy.argpartition``
---------------------------------------------------------------------------
+------------------------------------------------------------------------------------
 Users can now pass the `descending=True` keyword argument to ``numpy.partition`` and
 ``numpy.argpartition`` to partition and argpartition arrays in descending order.
 NaN values, if present, are partitioned to the end of the array in both ascending and
--- a/doc/release/upcoming_changes/31511.new_feature.rst
+++ b/doc/release/upcoming_changes/31511.new_feature.rst
@@ -1,0 +1,9 @@
+New `descending` keyword argument for ``numpy.partition`` and ``numpy.argpartition``
+--------------------------------------------------------------------------
+Users can now pass the `descending=True` keyword argument to ``numpy.partition`` and
+``numpy.argpartition`` to partition and argpartition arrays in descending order.
+NaN values, if present, are partitioned to the end of the array in both ascending and
+descending sorts. This feature is available for all built-in dtypes except
+`string`, `unicode`, `void`, `object`, and `generic`. Note that SIMD optimizations
+for partitioning are currently not available for descending order, so performance
+may be slower.
--- a/doc/release/upcoming_changes/31511.new_feature.rst
+++ b/doc/release/upcoming_changes/31511.new_feature.rst
@@ -1,9 +1,9 @@
-New `descending` keyword argument for ``numpy.partition`` and ``numpy.argpartition``
-------------------------------------------------------------------------------------
-Users can now pass the `descending=True` keyword argument to ``numpy.partition`` and
-``numpy.argpartition`` to partition and argpartition arrays in descending order.
+New ``descending`` keyword argument for `numpy.partition` and `numpy.argpartition`
+----------------------------------------------------------------------------------
+Users can now pass the ``descending=True`` keyword argument to `numpy.partition` and
+`numpy.argpartition` to partition and argpartition arrays in descending order.
 NaN values, if present, are partitioned to the end of the array in both ascending and
 descending sorts. This feature is available for all built-in dtypes except
-`string`, `unicode`, `void`, `object`, and `generic`. Note that SIMD optimizations
-for partitioning are currently not available for descending order, so performance
-may be slower.
+``string``, ``unicode``, ``void``, ``object``, and ``generic``. Note that SIMD
+optimizations for partitioning are currently not available for descending order,
+so performance may be slower.
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -3318,14 +3318,13 @@ add_newdoc('numpy._core.multiarray', 'ndarray', ('partition',
     ----------
     kth : int or sequence of ints
         Element index to partition by. The k-th value of the array will
-        be in the position it would be in a sorted array, all elements that
-        sorted array, all elements that are less than this element (or
-        greater if `descending` is True) will be moved before it, and
-        all elements that are greater than or equal to this element
-        (or less than or equal if `descending` is True) will be moved after it.
-        The order of all elements within each partition is undefined. If
-        provided with a sequence of k-th it will partition all elements
-        indexed by k-th of them into their sorted position at once.
+        be in the position it would be in a sorted array, all elements
+        that are less than this element (or greater if `descending` is True)
+        will be moved before it, and all elements that are greater than or
+        equal to this element (or less than or equal if `descending` is True)
+        will be moved after it. The order of all elements within each partition
+        is undefined. If provided with a sequence of k-th it will partition all
+        elements indexed by k-th of them into their sorted position at once.
 
         .. deprecated:: 1.22.0
             Passing booleans as index is deprecated.
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -3341,14 +3341,14 @@ add_newdoc('numpy._core.multiarray', 'ndarray', ('partition',
         but unspecified fields will still be used, in the order in which
         they come up in the dtype, to break ties.
     descending : bool, optional
-        Partition order. If ``True``, the left partition would contain
-        values less than or equal to the k-th element and the right partition
-        will contain values greater than or equal to the k-th element.
+        Partition order. If ``True``, the left partition will contain
+        values greater than the k-th element and the right partition
+        will contain values less than or equal to the k-th element.
         If ``False`` or ``None``, the left partition will contain values
-        less than or equal to the k-th element and the right partition
-        will contain values greater than or equal to the k-th element. Values
-        that are NaN are partitioned towards the end of the array regardless of
-        order. Default: ``None``.
+        less than the k-th element and the right partition will contain values
+        greater than or equal to the k-th element. Values that are NaN
+        are partitioned towards the end of the array regardless of order.
+        Default: ``None``.
 
         .. versionadded:: 2.6.0
 
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -3316,8 +3316,6 @@ add_newdoc('numpy._core.multiarray', 'ndarray', ('partition',
 
     Parameters
     ----------
-    a : array_like
-        Array to be sorted.
     kth : int or sequence of ints
         Element index to partition by. The k-th value of the array will
         be in the position it would be in a sorted array, all elements that
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -3307,7 +3307,7 @@ add_newdoc('numpy._core.multiarray', 'ndarray', ('partition',
 
     a.partition(kth, axis=-1, kind='introselect', order=None, descending=None)
 
-    Partially sorts the array in such a way that the value of the element in k-th
+    Partially sorts the array in such a way that the value of the element in the k-th
     position is in the position it would be in a sorted array. In the output array,
     all elements that would be to the left of the k-th element in a sorted array are
     located to the left of this element and all that would be to the right are located
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -3284,10 +3284,10 @@ add_newdoc('numpy._core.multiarray', 'ndarray', ('dot',
 
 add_newdoc('numpy._core.multiarray', 'ndarray', ('argpartition',
     """
-    argpartition($self, kth, /, axis=-1, kind='introselect', order=None)
+    argpartition($self, kth, /, axis=-1, kind='introselect', order=None, descending=None)
     --
 
-    a.argpartition(kth, axis=-1, kind='introselect', order=None)
+    a.argpartition(kth, axis=-1, kind='introselect', order=None, descending=None)
 
     Returns the indices that would partition this array.
 
@@ -3302,27 +3302,30 @@ add_newdoc('numpy._core.multiarray', 'ndarray', ('argpartition',
 
 add_newdoc('numpy._core.multiarray', 'ndarray', ('partition',
     """
-    partition($self, kth, /, axis=-1, kind='introselect', order=None)
+    partition($self, kth, /, axis=-1, kind='introselect', order=None, descending=None)
     --
 
-    a.partition(kth, axis=-1, kind='introselect', order=None)
+    a.partition(kth, axis=-1, kind='introselect', order=None, descending=None)
 
-    Partially sorts the elements in the array in such a way that the value of
-    the element in k-th position is in the position it would be in a sorted
-    array. In the output array, all elements smaller than the k-th element
-    are located to the left of this element and all equal or greater are
-    located to its right. The ordering of the elements in the two partitions
-    on the either side of the k-th element in the output array is undefined.
+    Partially sorts the array in such a way that the value of the element in k-th
+    position is in the position it would be in a sorted array. In the output array,
+    all elements that would be to the left of the k-th element in a sorted array are
+    located to the left of this element and all that would be to the right are located
+    to its right. The ordering of the elements in the two partitions on the either side
+    of the k-th element in the output array is undefined.
 
     Parameters
     ----------
+    a : array_like
+        Array to be sorted.
     kth : int or sequence of ints
-        Element index to partition by. The kth element value will be in its
-        final sorted position and all smaller elements will be moved before it
-        and all equal or greater elements behind it.
-        The order of all elements in the partitions is undefined.
-        If provided with a sequence of kth it will partition all elements
-        indexed by kth of them into their sorted position at once.
+        Element index to partition by. The k-th value of the array will
+        be in the position it would be in a sorted array, all elements that
+        would be sorted before it will be moved before it, and all elements
+        that would be sorted after it will be moved after it. The order of
+        all elements within the partitions is undefined. If provided with a
+        sequence of k-th it will partition all elements indexed by k-th of
+        them into their sorted position at once.
 
         .. deprecated:: 1.22.0
             Passing booleans as index is deprecated.
@@ -3337,6 +3340,17 @@ add_newdoc('numpy._core.multiarray', 'ndarray', ('partition',
         be specified as a string, and not all fields need to be specified,
         but unspecified fields will still be used, in the order in which
         they come up in the dtype, to break ties.
+    descending : bool, optional
+        Partition order. If ``True``, the left partition would contain
+        values less than or equal to the k-th element and the right partition
+        will contain values greater than or equal to the k-th element.
+        If ``False`` or ``None``, the left partition will contain values
+        less than or equal to the k-th element and the right partition
+        will contain values greater than or equal to the k-th element. Values
+        that are NaN are partitioned towards the end of the array regardless of
+        order. Default: ``None``.
+
+        .. versionadded:: 2.6.0
 
     See Also
     --------
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -3319,11 +3319,13 @@ add_newdoc('numpy._core.multiarray', 'ndarray', ('partition',
     kth : int or sequence of ints
         Element index to partition by. The k-th value of the array will
         be in the position it would be in a sorted array, all elements that
-        would be sorted before it will be moved before it, and all elements
-        that would be sorted after it will be moved after it. The order of
-        all elements within the partitions is undefined. If provided with a
-        sequence of k-th it will partition all elements indexed by k-th of
-        them into their sorted position at once.
+        sorted array, all elements that are less than this element (or
+        greater if `descending` is True) will be moved before it, and
+        all elements that are greater than or equal to this element
+        (or less than or equal if `descending` is True) will be moved after it.
+        The order of all elements within each partition is undefined. If
+        provided with a sequence of k-th it will partition all elements
+        indexed by k-th of them into their sorted position at once.
 
         .. deprecated:: 1.22.0
             Passing booleans as index is deprecated.
@@ -3331,7 +3333,9 @@ add_newdoc('numpy._core.multiarray', 'ndarray', ('partition',
         Axis along which to sort. Default is -1, which means sort along the
         last axis.
     kind : {'introselect'}, optional
-        Selection algorithm. Default is 'introselect'.
+        NumPy currently offers only one selection algorithm, 'introselect',
+        and this parameter provides no additional functionality. Default
+        is ``None``.
     order : str or list of str, optional
         When `a` is an array with fields defined, this argument specifies
         which fields to compare first, second, etc. A single field can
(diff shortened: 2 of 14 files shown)


numpy/_core/include/numpy/ndarraytypes.h
```diff
@@ -197,9 +197,13 @@ typedef enum {
 
 
 typedef enum {
-        NPY_INTROSELECT=0
+        _NPY_SELECT_UNDEFINED = -1,
+        NPY_INTROSELECT = 0,
+        // new style names
+        NPY_SELECT_DEFAULT = 0,
+        NPY_SELECT_DESCENDING = 1,
```
seberg: Hmmm, I wonder if we should just also make it 4, so that nsorts and nselects might just be identical (no point right now, I guess).

seberg (commented):
> Note that this does not create new-style ArrayMethod registration analogous to the sorts, but rather updates the (apparently fully internal) partition_map and continues using the old-style PartitionFunc

I would like to introduce the new ArrayMethod approach, but currently this is all internal so I am very happy to start with this since it also splits things up into smaller PRs.
(Plus, may help to work on top-k in parallel.)

But yeah, needs tests and docs for sure :).

(Otherwise, I think kind+descending should be fine, except that kind is only a single option anyway, so it's a bit funny and we should maybe just have stable/unstable as for the sorts, even if that is even less interesting here probably.
)

numpy/_core/include/numpy/ndarraytypes.h
MaanasArora: Thanks, yeah, right now the kind is selected by its position in the array... not great, but would go away when we put in ArrayMethods anyway!

MaanasArora:
> I would like to introduce the new ArrayMethod approach, but currently this is all internal so I am very happy to start with this since it also splits things up into smaller PRs.

Thanks for reviewing! Yes, this seems to be a good way to break the PRs, and I definitely will follow-up very soon with ArrayMethods.

I added tests, will add docs soon! I found a bug where dtypes were silently falling to (default) sort even _with_ descending, so I disabled that (`TypeError`s now). But I realized strings and dates even actually don't use introselect; they fell back to sorting, which can easily be replicated for descending with ArrayMethods. I'm trying to see if dates can actually be conveniently mapped to the existing introselect (given we have the nice new comparisons), but have just disabled the tests for this push. If that doesn't work, can keep them disabled, and strings should probably be disabled for this PR anyway (the introselect would have to duplicated for strings otherwise I think, which can be done in follow-up).

_Edit_: I was able to route datetime/timedelta to the existing introselect, and the tests are also enabled now!

> we should maybe just have stable/unstable as for the sorts, even if that is even less interesting here probably.

Agreed having stable could be nice for consistency! Maybe would be nice to add another select algorithm but stable, from which we can fallback to stable sorts.

MaanasArora:
Added docs, release note, and fixed up the `ma` stubs to align! I'll think a bit more about the angle for the docs, but this should be ready to review, thanks!

numpy/_core/tests/test_multiarray.py
```diff
@@ -9253,7 +9435,7 @@ def test_no_suboffsets(self):
         try:
             import _testbuffer
         except ImportError:
-            raise pytest.skip("_testbuffer is not available")
+            raise pytest.mark.skip("_testbuffer is not available")
```
```suggestion
            pytest.skip("_testbuffer is not available")
```
Also above, `pytest.skip()` raises the `SkipException`.  Although, in the first, we might as well use a `@pytest.mark.skipif()` decorator and in this one a `_testbuffer = pytest.importskip("_testbuffer")`

(I am not clear why this is suddenly showing up, but so be it.)

numpy/ma/core.pyi
```diff
@@ -2615,35 +2617,35 @@ class MaskedArray(ndarray[_ShapeT_co, _DTypeT_co]):
...
     ) -> MaskedArray[tuple[int], np.dtype[intp]]: ...
     @overload  # axis: index (default)
     def argpartition(
         self,
         kth: _ArrayLikeInt,
         /,
         axis: SupportsIndex = -1,
-        kind: _PartitionKind = "introselect",
-        order: None = None,
+        kind: _PartitionKind | None = None,        order: None = None,
```
seberg: wooops :).

numpy/_core/tests/test_multiarray.py
```diff
@@ -3065,6 +3057,196 @@ def test_argsort_descending_string(self, dtype, stable, descending):
...
+        self._test_argpartition_descending(a, k, nan, descending)
+
+    @pytest.mark.parametrize('dtype', [np.str_, np.bytes_, np.object_])
+    def test_partition_and_argpartition_raise_on_descending(self, dtype):
+        a = np.arange(10, dtype=np.float64).astype(dtype)
+
+        with assert_raises(TypeError, msg="type does not support descending partition"):
+            np.partition(a, 5, descending=True)
+        with assert_raises(TypeError, msg="type does not support descending partition"):
+            np.argpartition(a, 5, descending=True)
```
seberg, line 3310: OK, yeah, would be nice to work on that, but we can follow up.

numpy/_core/src/multiarray/methods.c
```diff
@@ -1499,20 +1516,37 @@ array_argpartition(PyArrayObject *self,
...
             {"|kind", &PyArray_SelectkindConverter, &sortkind},
-            {"|order", NULL, &order}) < 0) {
+            {"|order", NULL, &order},
+            {"|descending", &PyArray_OptionalBoolConverter, &descending}) < 0) {
         return NULL;
     }
+
+    if (sortkind == _NPY_SELECT_UNDEFINED) {
+        // keywords only if sortkind not passed
+        sortkind = descending > 0? NPY_SELECT_DESCENDING: NPY_SELECT_DEFAULT;
```
seberg:
```suggestion
        sortkind = descending > 0 ? NPY_SELECT_DESCENDING : NPY_SELECT_DEFAULT;
```
(but just silly nit)

numpy/_core/tests/test_multiarray.py
```diff
@@ -2842,15 +2834,15 @@ def test_sort_descending_unsigned(self, dtype, stable, descending):
...
+        self._test_sort_descending_nan(a, np.nan, stable, descending)
 
     @pytest.mark.parametrize("dtype", [np.complex64, np.complex128, np.clongdouble])
     @pytest.mark.parametrize("stable", [True, False])
     @pytest.mark.parametrize("descending", [True, False])
     def test_sort_descending_complex(self, dtype, stable, descending):
         a = np.arange(-50, 51, dtype=dtype) + 1j * np.arange(-50, 51, dtype=dtype)
         self._test_sort_descending_nonan(a, stable, descending)
-        self._test_sort_descending_nan(a, stable, descending)
+        self._test_sort_descending_nan(a, np.nan + 1j * np.nan, stable, descending)
```
```suggestion
        self._test_sort_descending_nan(a, complex(np.nan, np.nan), stable, descending)
```
(Doesn't matter here, feel free to ignore.  It only matters for things like `complex(0, np.nan)`; but I like the passing in NaN refactor.)

numpy/_core/tests/test_multiarray.py
```diff
@@ -3065,6 +3057,196 @@ def test_argsort_descending_string(self, dtype, stable, descending):
...
+    @pytest.mark.parametrize('descending', [True, False])
+    @pytest.mark.parametrize('nan', [np.nan, None])
+    def test_argpartition_descending_floats(self, dtype, k, descending, nan):
+        a = np.linspace(-50, 50, 101, dtype=dtype)
+        self._test_argpartition_descending(a, k, nan, descending)
+
+    @pytest.mark.parametrize('dtype', [np.complex64, np.complex128, np.clongdouble])
+    @pytest.mark.parametrize('k', [2, 15, 50, 80])
+    @pytest.mark.parametrize('descending', [True, False])
+    @pytest.mark.parametrize('nan', [np.nan + 1j * np.nan, None])
```
seberg, line 3278: Should we test different kinds of NaNs for complex?  (Although, I don't care too much, and if I guess mixing them would be the interesting part.)

numpy/_core/fromnumeric.py
```diff
@@ -725,34 +725,36 @@ def matrix_transpose(x, /):
...
 
 
-def _partition_dispatcher(a, kth, axis=None, kind=None, order=None):
+def _partition_dispatcher(a, kth, axis=None, kind=None, order=None, descending=None):
     return (a,)
 
 
 @array_function_dispatch(_partition_dispatcher)
-def partition(a, kth, axis=-1, kind='introselect', order=None):
+def partition(a, kth, axis=-1, kind=None, order=None, descending=np._NoValue):
```
seberg: I like changing the default to `None`, but I am wondering about the astropy problem.  In this case, I suspect astropy is fine, though (it should just forward it back to NumPy).

That said, if we check briefly that pandas and astropy don't run into this, I think we can do it.  We _could_ actually use `np._NoValue` here as well.

numpy/_core/_add_newdocs.py
```diff
@@ -3302,27 +3302,30 @@
...
+    position is in the position it would be in a sorted array. In the output array,
+    all elements that would be to the left of the k-th element in a sorted array are
+    located to the left of this element and all that would be to the right are located
+    to its right. The ordering of the elements in the two partitions on the either side
+    of the k-th element in the output array is undefined.
 
     Parameters
     ----------
+    a : array_like
+        Array to be sorted.
```
seberg:
```suggestion
```
Hmmm, fun that we have a copy for partition and not argpartition, but let's just keep it.  This is the method though.

numpy/_core/src/multiarray/methods.c
```diff
@@ -1499,20 +1516,37 @@ array_argpartition(PyArrayObject *self,
...
+    int descending = -1;
     NPY_PREPARE_ARGPARSER;
 
     if (npy_parse_arguments("argpartition", args, len_args, kwnames,
             {"kth", NULL, &kthobj},
             {"|axis", &PyArray_AxisConverter, &axis},
             {"|kind", &PyArray_SelectkindConverter, &sortkind},
-            {"|order", NULL, &order}) < 0) {
+            {"|order", NULL, &order},
+            {"|descending", &PyArray_OptionalBoolConverter, &descending}) < 0) {
```
seberg:
```suggestion
            {"$descending", &PyArray_OptionalBoolConverter, &descending}) < 0) {
```
Probably might as well make it kwonly here as well.

numpy/_core/src/multiarray/methods.c
```diff
@@ -1499,20 +1516,37 @@ array_argpartition(PyArrayObject *self,
...
     }
+
+    if (sortkind == _NPY_SELECT_UNDEFINED) {
+        // keywords only if sortkind not passed
+        sortkind = descending > 0? NPY_SELECT_DESCENDING: NPY_SELECT_DEFAULT;
+    }
+    else {
+        // Check that no keywords are used
+        if (descending != -1) {
+            PyErr_SetString(PyExc_ValueError,
```
seberg, line 1543: I think I might go with a TypeError, but doesn't matter much.  TBH, I am a bit unsure if this is helpful?  `kind` is currently only introselect which seems OK, we may want to deprecate it eventually (with sort's `kind`).

numpy/_core/fromnumeric.py
```diff
@@ -876,6 +896,17 @@ def argpartition(a, kth, axis=-1, kind='introselect', order=None):
         field can be specified as a string, and not all fields need be
         specified, but unspecified fields will still be used, in the
         order in which they come up in the dtype, to break ties.
+    descending : bool, optional
+        Partition order. If ``True``, the left partition would contain
```
seberg: I wonder if we should just start with "Sort order" (feels like it may be simpler).

numpy/_core/fromnumeric.py
```diff
@@ -876,6 +896,17 @@ def argpartition(a, kth, axis=-1, kind='introselect', order=None):
...
+        indices of elements less than or equal to the k-th element and the
+        right partition will contain indices of elements greater than or
+        equal to the k-th element. If ``False`` or ``None``, the left partition
+        will contain indices of elements less than or equal to the k-th element
```
seberg: Whoops, this is reversed.  But I am wondering if we can make it a bit shorter/closer to the sort docs and maybe even just say "partitioned in ascending/descending order" and rely on the introduction documentation to be clear enough on it?

numpy/_core/_add_newdocs.py
```diff
@@ -3302,27 +3302,30 @@
...
-    a.partition(kth, axis=-1, kind='introselect', order=None)
+    a.partition(kth, axis=-1, kind='introselect', order=None, descending=None)
 
-    Partially sorts the elements in the array in such a way that the value of
-    the element in k-th position is in the position it would be in a sorted
-    array. In the output array, all elements smaller than the k-th element
-    are located to the left of this element and all equal or greater are
-    located to its right. The ordering of the elements in the two partitions
-    on the either side of the k-th element in the output array is undefined.
+    Partially sorts the array in such a way that the value of the element in k-th
```
seberg:
```suggestion
    Partially sorts the array in such a way that the value of the element in the k-th
```
I wonder if we should remove "the value of".  Since I think we can just say k-th element and the other docs do (but this was there before).

seberg (approved):
LGTM, thanks.  A few smaller things, but most comments are just observations/thoughts really.
I might think about the docs a bit more.  I wonder if it is clearer if we just keep the old text with `smaller/greater` even.
(And if the clarification at the end like "Sorting the partitions split by k-th individually would yield a fully sorted array." might be helpful.)

numpy/_core/tests/test_multiarray.py
MaanasArora: Thanks, this is done! (For the first one, not sure if the decorator would work, since there an argument is part of the condition)

numpy/ma/core.pyi
MaanasArora: Sorry, fixed!

numpy/_core/src/multiarray/methods.c
MaanasArora: Fixed, thanks!

numpy/_core/tests/test_multiarray.py
MaanasArora: Yeah, mixing would be nice! I should probably add an analogous test to `test_sort_descending_complex_lexorder`, let me try!

numpy/_core/_add_newdocs.py
MaanasArora: Sorry, fixed, thanks!

numpy/_core/src/multiarray/methods.c
MaanasArora: Done.

numpy/_core/src/multiarray/methods.c
MaanasArora: Yeah, might not be very helpful really. I was mirroring the process for sorts, as @charris preferred, but it is true there is really no mixing confusion here - I can remove the check if you like!

numpy/_core/fromnumeric.py
MaanasArora: Yeah, that's better, this is done - thanks!

numpy/_core/fromnumeric.py
MaanasArora: Sorry, the reversal is fixed! Shortening the docs would be nice. I think the issue is really that this is moved from the `kth` docs to here, and it's not natural. Perhaps we can move this to k-th, but it is not really relevant there. Maybe just like the original, a single smaller/greater wording is better, but it's not really correct for descending?

numpy/_core/_add_newdocs.py
MaanasArora: Article fixed, thanks! Yeah, I think we should remove it (missed in this push sorry). Let me do that.

numpy/_core/fromnumeric.py
MaanasArora: Thanks, would running the test suite for these projects with this replacement be enough? Or perhaps we should build with this, really, because of the C changes?

Edit: sorry, was confused, these are independent Python modules :) let me just run the test suite with spin-built numpy!

MaanasArora:
Sorry, silly errors, thanks for the thorough review! Aside from the docs, I guess the big thing is checking astropy and pandas are fine with the default change for `kind`, I'll get started on that! (A complex nan mixing test would also be nice.)

numpy/_core/fromnumeric.py
MaanasArora: OK, so for astropy, it actually seems `MaskedNDArray` (if that is the main concern) does not quite even use `np.partition`!? It falls back to a full sort:

```
    def argpartition(self, kth, axis=-1, kind="introselect", order=None):
        # TODO: should be possible to do this faster than with a full argsort!
        return self.argsort(axis=axis, order=order)

    def partition(self, kth, axis=-1, kind="introselect", order=None):
        # TODO: should be possible to do this faster than with a full argsort!
        return self.sort(axis=axis, order=None)
```

The test suite passed, except for two unrelated datetime/timedelta deprecation failures (generic). Let me double check if I have everything right though!

MaanasArora:
Pushed another iteration of the docs, not sure if it turned out much better with the parentheses, but the `descending` entry is simpler at least.

For the `kind` interop issue: Astropy refers to partitioning in its `MaskedNDArray` class, but doesn't actually seem to invoke `np.(arg)partition`, just full sorts, so the argument isn't being passed I think. Pandas seems to only use a `partition` method in the context of string partitioning, so perhaps not relevant at all there either? (I'm a bit surprised, though.) Test suites passed for both projects with dev NumPy installation on this branch.

seberg:
Cool, I am happy to not worry too much about it then.  Although, I think it may still make sense to just go with `np._NoValue` to not pass it on to the method at all (rather than `None`) -- will probably need to build a kwarg dict, but so be it...
Not sure if that is slightly wrong for the descending sorts, but should't the types also use NoValue?

MaanasArora:
Thanks, I switched to `np._noValue`, and used a dict for the kwargs! Also, added a test for complex with nan mixing.

> Not sure if that is slightly wrong for the descending sorts, but should't the types also use NoValue?

Yes, that seems right! Going to make that change, thanks. (`np._NoValue` itself doesn't seem to occur anywhere in the types, but `_NoValueType` does!)

numpy/_core/include/numpy/ndarraytypes.h
seberg: Right, but this is public API, so we can't defer this if that is what you mean?  (and if that means changing the other code, so be it).

numpy/_core/src/multiarray/methods.c
seberg: I don't see a reason but if sorts have it, then I don't mind just keeping it.

numpy/_core/_add_newdocs.py
```diff
@@ -3302,41 +3302,53 @@
...
     kth : int or sequence of ints
-        Element index to partition by. The kth element value will be in its
-        final sorted position and all smaller elements will be moved before it
-        and all equal or greater elements behind it.
-        The order of all elements in the partitions is undefined.
-        If provided with a sequence of kth it will partition all elements
-        indexed by kth of them into their sorted position at once.
+        Element index to partition by. The k-th value of the array will
+        be in the position it would be in a sorted array, all elements that
+        sorted array, all elements that are less than this element (or
```
seberg: Docs got a little bit garbled here. "... that sorted array".

numpy/_core/tests/test_multiarray.py
```diff
@@ -3065,6 +3057,229 @@ def test_argsort_descending_string(self, dtype, stable, descending):
...
+    @pytest.mark.parametrize('random_seed', [0, 1])
+    def test_partition_descending_complex_lexorder(
+        self, dtype, k, descending, random_seed
+    ):
+        arange = np.tile(np.arange(25, dtype=dtype), 4)
+        nans = np.full(100, np.nan + 1j * np.nan, dtype=dtype)
+
+        no_nans = arange + 1j * arange
+        im_nans = arange + 1j * nans
+        re_nans = nans + 1j * arange
```
seberg: My agent actually noticed that this doesn't do what it says it does (maybe not new), this was the reason why I mentioned `complex(..., ...)` in a sense.

Just replacing `1j * nans` with `complex(0, np.nan)` should do the trick (and the last `nans` just with `np.nan` or `complex(np.nan, 0)` just to be obvious)

seberg (commented):
One small round of nits, but we should get this done.  I think the new docs read quite well so happy there.
The things around `None` default, etc. mirrors sorts, so I don't want to think about the best approach `_NoValue` use is good now.

My agent had a few tiny test suggestions, but I don't think any of them need to be fixed particularly.

numpy/_core/tests/test_multiarray.py
MaanasArora: Ah yeah, that's right, thanks - I fixed this, but this actually unearthed a bad bug with the complex comparators (they were always a bit broken, but `descending` exposed it more)... I'm going to make a PR to fix that (as it probably should be backported), so I didn't update the sorting test here, but doing so in the other PR. Sorry it got missed in the descending PR.

numpy/_core/_add_newdocs.py
MaanasArora: Fixed, thanks!

numpy/_core/include/numpy/ndarraytypes.h
MaanasArora: Right, yeah, we should probably sort this out just now. I've made the enum effectively identical and changed the other code (I assume we're not actually adding the kwarg yet.) Not sure what to do with the `NPY_NSELECTS`...

doc/release/upcoming_changes/31511.new_feature.rst
```diff
@@ -0,0 +1,9 @@
+New `descending` keyword argument for ``numpy.partition`` and ``numpy.argpartition``
```
seberg:
```suggestion
New ``descending`` keyword argument for ``numpy.partition`` and ``numpy.argpartition``
```
also below `string`, etc. single ticks mean that it should be a link. `numpy.partition` can be single ticks though, because they can link to the actual functions :).

numpy/_core/include/numpy/ndarraytypes.h
```diff
@@ -197,9 +197,14 @@ typedef enum {
 
 
 typedef enum {
-        NPY_INTROSELECT=0
+        _NPY_SELECT_UNDEFINED = -1,
+        NPY_INTROSELECT = 0,
+        // new style names
+        NPY_SELECT_DEFAULT = 0,
+        NPY_SELECT_STABLE = 2,
```
seberg, line 204: Sorry, maybe this whole "make it identical" wasn't the best idea.  But it's also fine. However, we should make sure that we reject unsupported values in `get_partition_func` which I don't think is currently the case (could do both and just delete the stable for now, but I do think we need to reject things).

seberg (commented):
Thanks, one more comment on the kind.  I am not quite sure if we should write `descending=None` or NoValue (in the docs), but we support None now, and it reads maybe better so I am happy with it.

I think the rest is all good.  It would be nice to change it to just use the correct full sort (rather than the cmp function), but I think that is a good follow-up.

So with that one narrowing down, I think we can go ahead here.

doc/release/upcoming_changes/31511.new_feature.rst
MaanasArora: Ah sorry, I kept switching these! Thank you, this is done.

numpy/_core/include/numpy/ndarraytypes.h
MaanasArora: Thanks, yeah, I just switched to a switch statement like we have for sorts, seems more explicit and clearer hopefully! I'm happy to delete stable if you like.

seberg (approved):
Thanks, I tweaked the docs for `NPY_SELECTKIND` a bit since I noticed we forgot that.
But that is a tiny change in practice, so I think I'll just merge later today.

PR #31514, opened 2026-05-26
DOC: inline-role to code, and item present twice in See-also

Carreau:
The two inline roles appear as unresolved-links in emphases instead of code.

#### AI Disclosure

No AI tools used


rossbar (approved):
Thanks @Carreau !

PR #31515, opened 2026-05-26
TYP: widen einsum subscripts param to accept complex arrays in subscript-free calling convention

sarveshvetrivel:
### PR summary

The `subscripts` parameter in every `np.einsum` overload was typed as `str | _ArrayLikeInt_co`. In the subscript-free calling convention (e.g. `np.einsum(a, [0,1], b, [1,0], [0])`), the first argument is the data array itself — which can be float or complex, not just integer-like. This caused false-positive `[arg-type]` mypy errors for valid code.

The fix widens `subscripts` to `str | _ArrayLikeComplex_co` across all 13 overloads, since `_ArrayLikeComplex_co` is a superset covering bool, uint, int, float, and complex arrays. A reveal test for the subscript-free convention with `float64` and `complex128` arrays is also added.

I'm a student who uses NumPy for numerical computing work. I came across this issue while using `np.einsum` with complex arrays in a project and hitting the false-positive mypy error myself. 

#### AI Disclosure

Claude (Anthropic) was used to help navigate the codebase and identify the relevant overloads. The diagnosis, fix, and submission were done by me. All changed code was reviewed and understood before committing.

### DIFF
--- a/numpy/_core/einsumfunc.pyi
+++ b/numpy/_core/einsumfunc.pyi
@@ -34,7 +34,7 @@ type _CastingUnsafe = Literal["unsafe"]
 # Something like `is_scalar = bool(__subscripts.partition("->")[-1])`
 @overload
 def einsum(
-    subscripts: str | _ArrayLikeInt_co,
+    subscripts: str | _ArrayLikeComplex_co,
     /,
     *operands: _ArrayLikeBool_co,
     out: None = None,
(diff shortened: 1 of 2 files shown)


jorenham (approved):
Hmm yea good catch. This fix seems fine to me. 

And now that I look at these `einsum` stubs, it's a pretty big mess; most overloads are redundant, return `Any`, and are missing defaults... Improving / cleaning that up is very much out of scope here, but it would make for a nice follow-up.

PR #31517, opened 2026-05-26
MAINT/TYP: bump `pyrefly` to `1.0.0`

jorenham:
The upcoming Pyrefly release that includes the fix for the ignored (false positive) errors here is taking a lot longer than expected. With this, dependabot should be able do its thing again.

ref: https://github.com/numpy/numpy/pull/31516#issuecomment-4549909868

#### AI Disclosure
no ai


charris:
Note that we also have a week wait period after release. I plan on back porting the updates when they are available.

charris:
Thanks Joren.

PR #31518, opened 2026-05-27
Descending partition

Ramshetty02:
**# Summary**

Adds a `descending` keyword argument to `np.partition`
and `np.argpartition`, analogous to the sorting APIs.

This enables descending-order partition selection without
requiring users to reverse arrays manually.

**## Changes**

* Added `descending` kwarg support to:

np.partition
np.argpartition
Added tests for integer arrays
Verified tests pass locally

**#Testing**
bash
python -m pytest numpy/_core/tests/test_multiarray.py

Local result:

**```text**
17325 passed, 4 skipped
```


jorenham:

PR #9910, opened 2017-10-23
ENH: push histogram calculations to compiled_base

theodoregoetz:
This is a resubmittal of PR #9627  (my sincere apologies for the gitmess)

# Faster numpy.histogram() and histogramdd()

The **fast-histogram** python package demonstrates that histogramming real data onto a continuous range with evenly-spaced bins, can be made significantly faster. I took this idea of pushing the bin calculation and the filling of the histogram array into a C function and implemented it in **NumPy**.

For details, please my gist concerning this patch:
https://gist.github.com/theodoregoetz/10d2351421689bf2660b4f2fca350e6e

PR #31521, opened 2026-05-28
ENH: add cpp based uniform histogram calculation for optimization

JosephMehdiyev:
Related: https://github.com/numpy/numpy/pull/9910
### PR summary
This PR makes the fast path of the histogram significantly fast. The current benchmark vs main:
```
| Change   | Before [fefe4f5f] <main>   | After [fa366b8a] <hist-C>   |   Ratio | Benchmark (Parameter)                               |
|----------|----------------------------|-----------------------------|---------|-----------------------------------------------------|
| -        | 61.4±1μs                   | 40.4±0.7μs                  |    0.66 | bench_function_base.Histogram1D.time_small_coverage |
| -        | 883±40μs                   | 170±0.9μs                   |    0.19 | bench_function_base.Histogram1D.time_fine_binning   |
| -        | 861±40μs                   | 156±1μs                     |    0.18 | bench_function_base.Histogram1D.time_full_coverage  |
```

1) Introduces a optimization  that were first discussed in https://github.com/numpy/numpy/pull/9910
2) Fixes a (unknown?) bug where `histogram` would not go fast path for long double
3) adds new histogram files (which I am not sure if it is a good idea). this is mainly because it is possible to make histogram way more faster than this PR and histogram is a important function so I think it deserves its own file?

### Other comments
I want to discuss this with someone experienced before I refine, implement the logic further into multidimensional case (for histogramdd). I tried to make the code look like other C files as much as possible (not sure if its smart thing to do)

I also feel like histogram deserves a couple of more benchmarks (especially for the slow path)

#### AI Disclosure
Claude was used as an AI tool.
Some parts (e.g template bodies) are written by me, some are recycled from the old python code, some are  recycled from old PR (e.g error handling, numpy API etc) and some are written by AI  (especially the weight type parts) and reviewed manually.
AI was also used extensively on Numpy API/Cpython specific parts, in parallel I have read some documentation and checked if I used them correctly, so be warned

### DIFF
--- a/numpy/_core/meson.build
+++ b/numpy/_core/meson.build
@@ -1206,6 +1206,7 @@ src_multiarray = multiarray_gen_headers + [
   'src/multiarray/flagsobject.c',
   'src/multiarray/getset.c',
   'src/multiarray/hashdescr.c',
+  'src/multiarray/histogram.cpp',
   'src/multiarray/item_selection.c',
   'src/multiarray/iterators.c',
   'src/multiarray/legacy_dtype_implementation.c',
--- a/numpy/_core/src/multiarray/histogram.cpp
+++ b/numpy/_core/src/multiarray/histogram.cpp
@@ -1,0 +1,205 @@
+#define NPY_NO_DEPRECATED_API NPY_API_VERSION
+#define _MULTIARRAYMODULE
+
+#define PY_SSIZE_T_CLEAN
+#include <Python.h>
+
+#include "numpy/arrayobject.h"
+#include "numpy/npy_math.h"
+extern "C" {
+#include "npy_argparse.h"
+}
+#include "common.h"
+
+#include "histogram.h"
+
+static inline void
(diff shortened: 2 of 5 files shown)


JosephMehdiyev:
Hey @seberg, can we discuss about this PR in your free time if possible? You had a comment on the old PR. Thanks!
(PR is not ready to merge, I will still change it from draft)

JosephMehdiyev:
In the second thought, I do not think the users would care about faster histogram (neither the maintainers would deem it important?). Be free to close this PR if someone thinks it is not worth it

seberg:
Sorry, I probably looked at it and was unsure about adding C++ code and then never came back (too many PRs).
That said: Generally, kind of speedups are worthwhile, we just have to make sure to balance it with maintainability (but on first sight, I guess this is swapping out a fast-path, so it seems to so terrible) a bit and histograms are a bit tricky...

You could try the mailing list also, maybe someone jumps at it to have a look (I might, but I don't want to make promises).

Issue #31249, opened 2026-04-15
BUG: meson backend breaks two-stage callback-based F2PY builds that work with distutils


xianrui5891:
### Describe the issue:

In a **two-stage** F2PY callback workflow, `--backend distutils` works for us but `--backend meson` produces an extension that fails to import.

The workflow is:

1. write a callback-facing Fortran interface first,
2. let deeper host-side Fortran routines call that interface and complete the host-side logic around it,
3. build that host-side logic first as a library,
4. expose a Python-facing entry point together with the callback-facing interface to F2PY,
5. register Python callbacks through the final F2PY-generated layer,
6. let the prebuilt host-side Fortran logic invoke those callbacks later during execution.

This is a small reproducer for a broader integration pattern where Python provides a control layer around a larger scientific or engineering Fortran program.

In the variants we tested, the meson-built result ends up failing because required symbols cannot be resolved from the produced extension or from the prebuilt host library it depends on.

Related docs issue:

- [#31248](https://github.com/numpy/numpy/issues/31248)

### Reproduce the code example:

The files below correspond directly to the workflow above:

1. `python_interface.f90` defines the callback-facing interface,
2. `fortran_main.f90` is deeper host-side logic that calls that interface,
3. `fortran_main_plug_no_register.f90` is the Python-facing entry point,
4. `python_driver.py` is the Python side that registers the callbacks and calls the wrapped entry point.

```fortran
! python_interface.f90
module python_interface

public python_interface_a, python_interface_b

contains

subroutine python_interface_a()
    implicit none
    !f2py intent(callback, hide) python_a
    external python_a

    print *, "in python_interface_a, calling python_a..."
    call python_a()
end subroutine python_interface_a

subroutine python_interface_b()
    implicit none
    !f2py intent(callback, hide) python_b
    external python_b

    print *, "in python_interface_b, calling python_b..."
    call python_b()
end subroutine python_interface_b

end module python_interface
```

```fortran
! fortran_main.f90
module main_mod

use python_interface

public main

contains
subroutine main
    print *, "in main, calling python_interface_a then python_interface_b..."
    call python_interface_a()
    call python_interface_b()
end subroutine main

end module main_mod
```

```fortran
! fortran_main_plug_no_register.f90
module plug

use main_mod

public call_main

contains
subroutine call_main
    call main
end subroutine call_main

end module plug
```

```python
# python_driver.py
import test_module_attr

saved = {}

def python_a():
    saved["value"] = "state from python_a"
[...]

Issue #31248, opened 2026-04-15
DOC: Clarify and document a two-stage F2PY callback workflow for larger Fortran programs


xianrui5891:
### Issue with current documentation:

I checked the current F2PY callback and signature-file documentation:

- https://numpy.org/doc/stable/f2py/python-usage.html
- https://numpy.org/doc/stable/f2py/signature-file.html
- https://numpy.org/doc/stable/f2py/usage.html

These docs show that callbacks work, including module-attribute callbacks and indirect Fortran calls.

What still seems unclear to me is whether users are expected to rely on the following two-stage workflow for larger Fortran programs:

1. write a callback-facing Fortran interface first,
2. let deeper host-side Fortran routines call that interface and complete the host-side logic around it,
3. build that host-side Fortran logic first as a library,
4. expose a Python-facing entry point together with the callback-facing interface to F2PY,
5. register Python callbacks through the final F2PY-generated layer,
6. let the prebuilt host-side Fortran logic invoke those callbacks later during execution.

From the current docs, it is not clear whether this is considered a supported F2PY pattern or just something that happens to work in some cases.

More specifically, it is not clear whether users are expected to rely on a setup where:

1. callback registration happens at the final Python/F2PY boundary,
2. callback invocation happens later in deeper Fortran routines,
3. the callback-facing interface is involved both in the host-library build and in the final F2PY wrapping step.

### Idea or request for content:

It would be very helpful to add a small advanced example, or even a short explicit note, showing a two-stage callback workflow for a larger Fortran program.

More specifically, I would like the docs to clarify:

1. whether the two-stage callback workflow above is intended to be supported,
2. whether the callback-facing interface can legitimately participate both in the host-library build and in the final F2PY wrapping step,
3. whether only the Python-facing entry point and callback-facing interface need to be exposed during `.pyf` generation, while the deeper host-side Fortran logic remains outside that step,
4. whether this is a reasonable pattern for integrating Python with larger scientific or engineering Fortran programs.

A small example would be especially useful here. The current callback examples are enough to show the basic mechanism, but they do not really show how this scales to a larger host program with its own existing build structure.

This also seems more relevant now than it used to be. In AI for Science and related workflows, users increasingly want to connect Python AI/ML libraries to existing native Fortran scientific programs while preserving the original Fortran implementation and build structure.

[...]

PR #31523, opened 2026-05-28
DOC: clarify two stage callback workflow in F2PY

Haris-bin-shakeel:
### PR summary

This PR clarifies an existing F2PY documentation gap around callback usage in larger Fortran projects.

It adds a concise description of a two stage callback workflow in which:

- Python callback registration happens at the final Python F2PY boundary
- callback invocation can occur later in deeper Fortran routines
- callback-facing interface can be used in both the host-library build and the final F2PY wrapping step depending on how the Fortran codebase is structured and linked
- only boundary routines and callback interfaces need to be exposed in pyf generation, while deeper host logic can remain linked externally

The goal is to improve discoverability and reduce ambiguity without changing runtime behavior or introducing new backend guarantees.

Related issues:
- gh-31248
- context only: gh-31249

I use NumPy for ML and scientific workflows, and I am trying to contribute long term to core scientific Python tooling. I picked this issue because it improves clarity for users integrating Python callbacks with larger existing Fortran codebases, which is a practical pattern in real projects.

#### AI Disclosure

I used AI assistance to help draft and refine documentation text and PR wording. I reviewed and edited all final content manually before submission. No runtime code changes were introduced.

### DIFF
--- a/doc/source/f2py/python-usage.rst
+++ b/doc/source/f2py/python-usage.rst
@@ -244,6 +244,23 @@ In Python:
 .. literalinclude:: ./code/results/extcallback_session.dat
   :language: python
 
+Two-stage callback workflows
+----------------------------
+
+For larger Fortran codebases, it is common to keep most host-side logic in a
+prebuilt Fortran library and expose only the Python-facing entry points through
+F2PY. A callback-facing interface can still be used in this setup:
+
+* Python callback registration happens at the final Python/F2PY boundary.
+* Callback invocation may happen later in deeper Fortran routines.
+* The callback-facing interface can participate in both the host-library build
+  and the final F2PY wrapping step.
+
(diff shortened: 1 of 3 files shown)


HaoZeke (changes requested):
@Haris-bin-shakeel thanks for this, but I'm curious, do you have any examples where you use the callbacks in F2PY? :)

For instance, AFAIK we only do callbacks for wrapped routines, so a deeper Fortran routine that never appears in the signature has no trampoline, so neither module assignment nor an unrelated argument can reach it.

Haris-bin-shakeel:
Thanks for the review @HaoZeke , you're right, I had the constraint imprecise.

I've revised the section to explicitly state that F2PY only generates a callback trampoline for routines that appear in the wrapping step, and referenced the existing extcallback.f example as a concrete demonstration of this pattern.

Please let me know if further changes are needed.

PR #31541, opened 2026-05-30
DOC: Document Cython and Numba as ways of generating ufuncs

da-woods:
### PR summary

The ufunc C API is a little involved, so I think it's worth documenting a couple of the alternatives to generate fast ufuncs automatically. (There's already some reference to Cython and similar in the Numpy documentation in the preceding https://numpy.org/doc/stable/user/c-info.python-as-glue.html section so I don't think I'm introducing anything new here).

Examples are stolen from the respective documentation - I'm just aiming for a couple of quick examples and a link to the external library.

#### AI Disclosure

No AI tools used.

### DIFF
--- a/doc/source/user/c-info.ufunc-tutorial.rst
+++ b/doc/source/user/c-info.ufunc-tutorial.rst
@@ -887,7 +887,7 @@ Numba
 -----
 
 The `Numba vectorize decorator
-<https://numba.pydata.org/numba-doc/dev/user/vectorize.html>`_ provides
+<https://numba.readthedocs.io/en/stable/user/vectorize.html>`_ provides
 a quick way to compile a subset of Python to an accelerated function that
 behaves like a ufunc. It can be used without arguments to support any
 numeric type (compiled on a "just in time" basis as you use them):
--- a/doc/source/user/c-info.ufunc-tutorial.rst
+++ b/doc/source/user/c-info.ufunc-tutorial.rst
@@ -874,3 +874,82 @@ Sample usage::
     >>> npufunc.add_triplet(a, a)
     array([(2,  4,  6), (8, 10, 12)],
           dtype=[('f0', '<u8'), ('f1', '<u8'), ('f2', '<u8')])
+
+Alternatives to writing your own ufunc
+======================================
+
+While most people like the speed benefits that come from writing a ufunc
+using the Numpy C API, not everyone enjoys writing C code. Fortunately
+there are a few alternatives that can create ufuncs with C-like speed
+but without having to understand the Numpy C API.
+
+Numba
+-----
+
+The `Numba vectorize decorator
+<https://numba.pydata.org/numba-doc/dev/user/vectorize.html>`_ provides
+a quick way to compile a subset of Python to an accelerated function that
+behaves like a ufunc. It can be used without arguments to support any
+numeric type (compiled on a "just in time" basis as you use them):
+
+    .. code-block:: python
+
+        from numba import vectorize
+
+        @vectorize
+        def f(x, y):
+            return x * y
+
+or with arguments to support only specific numeric types (compiled when
+you define the function):
+
+    .. code-block:: python
+
+        @vectorize([int32(int32, int32),
+                    int64(int64, int64),
+                    float32(float32, float32),
+                    float64(float64, float64)])
+        def f(x, y):
+            return x + y
+
+In addition to simple ufuncs, Numba can also be used to generate
+generalized ufuncs in a very similar way.
+
+Numba is a run-time dependency (i.e. users of your library will need to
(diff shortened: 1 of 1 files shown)


doc/source/user/c-info.ufunc-tutorial.rst
```diff
@@ -874,3 +874,82 @@ Sample usage::
...
+While most people like the speed benefits that come from writing a ufunc
+using the Numpy C API, not everyone enjoys writing C code. Fortunately
+there are a few alternatives that can create ufuncs with C-like speed
+but without having to understand the Numpy C API.
+
+Numba
+-----
+
+The `Numba vectorize decorator
+<https://numba.pydata.org/numba-doc/dev/user/vectorize.html>`_ provides
```
jorenham: Those are the old docs

```suggestion
<https://numba.readthedocs.io/en/stable/user/vectorize.html>`_ provides
```

doc/source/user/c-info.ufunc-tutorial.rst
```diff
@@ -874,3 +874,82 @@ Sample usage::
...
+
+or with arguments to support only specific numeric types (compiled when
+you define the function):
+
+    .. code-block:: python
+
+        @vectorize([int32(int32, int32),
+                    int64(int64, int64),
+                    float32(float32, float32),
+                    float64(float64, float64)])
```
jorenham: I think it'd help to highlight that these are `numba` scalar types, not `numpy` ones, by adding explicit exports for them

da-woods: Done - thanks

mattip (approved):
LGTM and renders well, thanks!

At some point we could consider using intersphinx, but I think that is overkill for a single link to each project.

doc/source/user/c-info.ufunc-tutorial.rst
```diff
@@ -874,3 +874,84 @@ Sample usage::
...
+``cython.numeric`` to support the most widely available standard C integer,
+floating and complex numeric types).
+
+Cython functions need to be compiled ahead of time (and thus Cython is
+a compile-time dependency, but users of your function do not need to
+have Cython installed themselves). Numpy will be both a compile-time and
+a runtime dependency for your compiled module, since Cython itself uses
+the ufunc C API described earlier in this section.
+
+As of May 2026, only simple ufuncs are supported and generic ufuncs are
```
ev-br: Is "generic ufuncs" a typo for generalized?

da-woods: Thanks - it is.

mhvk (commented):
Thanks, overall this is great! I have one bigger and one smaller comment, with the bigger one hopefully little work, and the smaller one OK to ignore.

Bigger: Since numba does not actually create ufuncs, it is really an alternative mechanism to create a compiled function rather than an alternative way to create a ufunc (which this page is about). I would thus swap the order, i.e., present numba after cython. Note that whether something is a true ufunc can be important for developers (probably the people most likely to read this page), since a cython-created ufunc can be used with other array types via `__array_ufunc__` (and may thus "just work" with, e.g., dask), while for a numba generated function one would have to go through much more trouble. It may be worth mentioning this explicitly.

Smaller: the text above deals with a `logit` function. I think it might be nicer to do the same function in the examples as well. Personally, I'm also curious whether there is any speed difference between the three methods, but it is fine to leave that up to the reader!

da-woods:
Thanks for the comments @mhvk.

For a bit if context, I put Numba in to try to avoid this being overly promotional for Cython, and because I stole the idea from them in the first place. But I'm by no means a Numba expert.

I hadn't realised that the "not a real ufunc" was so practically significant. I'll have little investigate myself and add a bit more detail there. 

I can definitely change it so they all do the same thing. That would allow easier comparison.

I'd expect the Cython vs direct Numpy speed to be pretty similar (in that ultimately it's the same implementation and the Cython code is pretty close to the "standard reference" implementation in the Numpy docs). Numba could be different though. I'll try to get some numbers but don't want to go down that rabbit-hole too much

da-woods:
I had a go with putting a pandas series through a Numba ufunc, and a custom class where I defined `__array_ufunc__` and they both seem to work.  So it looks it me like they support that protocol and that may not be a huge concern from a user point of view.

da-woods:
FWIW the speed is basically identical for all 3. If you don't turn off the exception checks then Cython is ~5% slower.

mhvk (approved):
Thanks for looking into the details - now the text is even more helpful!!

mhvk:
@jorenham - if you are happy too, shall we get this in?

jorenham:
> @jorenham - if you are happy too, shall we get this in?

I'm always happy! (unless I step on a Lego, that is)

PR #31545, opened 2026-05-31
API: Make np.squeeze always return an ndarray for scalar inputs

spandan11106:
### PR summary
Closes #30109.
Currently, `np.squeeze` behaves inconsistently with scalar-like inputs:
```python
>>> np.squeeze(1)           # Python int → 0d array
array(1)
>>> np.squeeze(np.array(1)) # 0d ndarray → 0d array
array(1)
>>> np.squeeze(np.int_(1))  # np.generic scalar → scalar (unexpected behavior)
np.int64(1)
```
The issue is that `np.generic` subclasses have a `.squeeze()` method that returns `self` (the scalar itself). The current `squeeze` implementation in `fromnumeric.py` tries `a.squeeze` first, which succeeds for `np.generic` scalars and returns the scalar unchanged. This contradicts the docstring, which states:
> *"Note that if all axes are squeezed, the result is a 0d array and not a scalar."*
This PR converts `np.generic` inputs to 0d arrays using ellipsis indexing `a[...]` before invoking `a.squeeze`.
#### Changes:
* **Runtime Fix**: Converts `np.generic` inputs to 0d arrays in `numpy/_core/fromnumeric.py`.
* **Type Stubs**: Updates `numpy/_core/fromnumeric.pyi` to remove the scalar-returning overload of `squeeze` (so it falls back to returning `NDArray[ScalarT]`).
* **Tests**: Adds tests in `numpy/_core/tests/test_multiarray.py` checking various scalar types, `axis=0`, and Python scalars.
* **Release Note**: Adds a compatibility news fragment in `doc/release/upcoming_changes/27709.compatibility.rst`.
***
Hi everyone! I am Spandan. I use NumPy for development/data science work, and noticed the inconsistent behavior when squeezing NumPy scalars compared to Python scalars. I wanted to contribute to the repository to clean up this edge case.
***
#### AI Disclosure
Antigravity  was used to assist in researching the issue and drafting the test cases.

### DIFF
--- a/doc/release/upcoming_changes/27709.compatibility.rst
+++ b/doc/release/upcoming_changes/27709.compatibility.rst
@@ -1,0 +1,3 @@
+``np.squeeze`` now always returns an ``ndarray`` for scalar inputs
+-----------------------------------------------------------------
+``np.squeeze`` now always returns a `numpy.ndarray` (specifically a 0-dimensional array), even when the input is a NumPy scalar (a subclass of `numpy.generic`), rather than returning the scalar itself.
--- a/numpy/_core/fromnumeric.py
+++ b/numpy/_core/fromnumeric.py
@@ -1691,6 +1691,8 @@ def squeeze(a, axis=None):
     1234
 
     """
+    if isinstance(a, np.generic):
+        a = a[...]
     try:
         squeeze = a.squeeze
     except AttributeError:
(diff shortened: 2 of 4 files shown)


jorenham:
We don't allow automatic copilot reviews here.

PR #7798, opened 2016-07-02
ENH: properly account for trailing padding in PEP3118

ahaldane:
This changes how numpy treats trailing padding in the PEP3118 buffer interface, to fix #7797.

Previously, dtypes with trailing padding would have their trailing padding truncated when converted to  memoryview using pep3118. Now trailing padding is maintained.

Then, I also needed to modify the pep3118-to-array conversion to stop explicitly adding void fields for the trailing padding. This way code of the form `np.array(memoryview(arr))` will give back the original array even if there is trailing padding.

One thing I still need to think about is what to do for "aligned" types. The pep3118 doc is not clear what "alignment" means. It appears that `ix` and `ixxx` are supposed to mean the same thing: a structure of 8 bytes containing an integer in the first 4 bytes and 4 bytes of padding, but I am not 100% sure that is right.

PR #31547, opened 2026-06-01
BUG: account for PEP 3118 trailing padding

### LINKED ISSUE #7797: Array from memoryview fails if itemsize is set explicitly
Here's converting an array to a memoryview and back (Python 3) where itemsize equals the offsets of the last item plus its size:

```python
>>> d1 = np.dtype({'formats': ['u1'], 'offsets': [0], 'names': ['x']})
>>> a1 = np.empty(0, d1)
>>> memoryview(a1).format
'T{B:x:}'
>>> memoryview(a1).itemsize
1
>>> np.array(memoryview(a1))
array([], 
      dtype=[('x', 'u1')])
```

If we try to do the same where itemsize is bigger, it fails:

```python
>>> d2 = np.dtype({'formats': ['u1'], 'offsets': [0], 'names': ['x'], 'itemsize': 4})
>>> a2 = np.empty(0, d2)
>>> memoryview(a2).format
'T{B:x:}'  # shouldn't this be 'T{B:x:3x}'?
>>> memoryview(a2).itemsize
4
>>> n
# so far so good...
# however:
>>> np.array(memoryview(a2))
RuntimeWarning: Item size computed from the PEP 3118 buffer format string does not match the actual item size.
NotImplementedError: memoryview: unsupported format T{B:x:}
```

This seems quite wrong.

Looking at the code where it fails, `_descriptor_from_pep3118_format` only accepts a format string and not the itemsize, so either the generated format string is wrong and should explicitly contain the trailing bytes, or the above function should accept the itemsize provided by the buffer.

May be related: #6361

cjc0013:
Closes #7797.

This revives and updates the old #7798 approach for current main. Structured dtypes with trailing padding now include that padding in their generated PEP 3118 buffer format, so `array -> memoryview -> array` can round-trip without converting the padding into an explicit unnamed void field.

Changes:
- emit compact explicit padding runs in structured buffer formats, including trailing padding
- parse trailing padding outside `T{}` without implicit final alignment, matching the `struct` docs
- preserve aligned trailing padding inside `T{}`
- support unnamed repeat-0 PEP 3118 elements as alignment padding
- add focused regression coverage and a release note

Validation:
- `spin build -- -Dallow-noblas=true`
- `spin test -v -t numpy/_core/tests/test_multiarray.py::TestPEP3118Dtype`
  - 10 passed
- `spin test -v -t numpy/_core/tests/test_multiarray.py::TestNewBufferProtocol`
  - 29 passed, 1 skipped (`_testbuffer` unavailable)
- Direct local reproduction:
  - `memoryview(arr).format == "T{B:x:3x}"`
  - `memoryview(arr).itemsize == 4`
  - `np.array(memoryview(arr)).dtype == arr.dtype`

### DIFF
--- a/doc/release/upcoming_changes/31547.improvement.rst
+++ b/doc/release/upcoming_changes/31547.improvement.rst
@@ -1,0 +1,5 @@
+PEP 3118 structured dtype trailing padding now round-trips
+-----------------------------------------------------------
+Structured dtypes with trailing padding now expose that padding in their
+PEP 3118 buffer format and round-trip through ``memoryview`` without adding
+an explicit unnamed void field.
--- a/numpy/_core/_internal.py
+++ b/numpy/_core/_internal.py
@@ -628,6 +628,8 @@ def _dtype_from_pep3118(spec):
     return dtype
 
 def __dtype_from_pep3118(stream, is_subdtype):
+    # numpy interprets PEP 3118 formats which include named fields as
+    # structured dtypes, even if not enclosed by "T{}".
     field_spec = {
         'names': [],
         'formats': [],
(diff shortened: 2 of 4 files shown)


jorenham:

PR #31550, opened 2026-06-02
TYP: Annotate untyped attributes and dunder methods

jorenham:
I found a couple of missing stub definitions by using `pyrefly coverage report`. For some reason those weren't reported by stubtest or pyright's type-coverage tool. But according to Pyrefly (and rightly so IMO), we need these attribute and dunder methods typed in order for us to claim that we indeed have 100% type coverage.

In the 2.5.0 relnotes we claim that numpy now has 100% coverage, so we should probably backport this 😅.

#### AI Disclosure
No AI


charris:
Thanks Joren.

Issue #24115, opened 2023-07-04
ENH: use `PYTHONWARNDEFAULTENCODING=1` to fix optional encoding warnings in Python 3.10+


Zac-HD:
Under Python 3.11, though for some reason I haven't investigated _not_ 3.10,

```python
import numpy.testing
```
triggers the following `EncodingWarning`
```
$ python -X warn_default_encoding demo.py
numpy/testing/_private/utils.py:1251: EncodingWarning: 'encoding' argument not specified.
  output = subprocess.run(cmd, capture_output=True, text=True)
```

It would be great if Numpy added this to CI and fixed however many errors that turns up; with Pytest 7.4.0 and later running clean it's now reasonably practical to do so.  

You may want a two-stage approach where you fix all the current cases in Numpy first (like https://github.com/HypothesisWorks/hypothesis/pull/3619), and then revisit enforcing it in CI once upstream projects have cleaned up their own warnings (like https://github.com/HypothesisWorks/hypothesis/pull/3691, where I hit this issue).

mattip:
I see. We need to add `encoding="utf-8"` a bit everywhere to make that warning go away, and remove it for python 3.15+, according to [PEP 686](https://peps.python.org/pep-0686/). That warning was added for python3.11, so one the 3.11+ CI runs should add the env variable.

Zac-HD:
The warning exists from Python 3.10, though there were some related small changes in 3.11.  

We don't actually need to remove the encoding arguments in 3.15 - as I understand it Python will then default to using uft-8 by default, but end-users can still disable that so libraries should probably keep specifying encoding when we know which of  utf-8 or locale is preferred.

Searchingdays:
Can this issue be closed , the puremagic encoding having been applied, or what further work might be needed here?

petsuter:
The encoding is still missing here: https://github.com/numpy/numpy/blob/425f631926c3ae6c817774b4e079ffdf207e3907/numpy/testing/_private/utils.py#L1482 triggering warnings in pytest when `PYTHONWARNDEFAULTENCODING=1` is set.

DV-AG:
Since this issue is still open, is there any remaining work that I could contribute to, particularly around enabling PYTHONWARNDEFAULTENCODING=1 in CI?

shoumikchakravarty-dev:
Covered all the subprocess.run(text=True) call sites in the repo. Turning on PYTHONWARNDEFAULTENCODING=1 in CI so regressions get caught is pending

[...]

PR #31551, opened 2026-06-02
MAINT: specify encoding in numpy/testing/_private/utils.py

Lawson-Darrow:
Several `open()` and `subprocess.run()` calls in `numpy/testing/_private/utils.py` omit the `encoding` argument, which triggers `EncodingWarning` under `PYTHONWARNDEFAULTENCODING=1` (Python 3.10+) — for example when downstream projects run their test suites with that flag set.

The most recent report on gh-24115 points specifically at the `subprocess.run(cmd, capture_output=True, text=True)` call here; this PR also covers the sibling `open()` calls in the same module. All of these read ASCII data (from `/proc/<pid>/stat`, `/proc/meminfo`) or the output of `lscpu`, so `utf-8` is the correct explicit encoding.

```python
$ python -X warn_default_encoding -W error::EncodingWarning ...
EncodingWarning: 'encoding' argument not specified
```

This is a scoped step toward gh-24115 (the `numpy/testing` portion); the umbrella issue also covers adding the flag in CI, which can follow once the remaining modules are clean.

- [x] Pure mechanical change (adds `encoding='utf-8'`), no behavior change for the ASCII content these calls read.

### DIFF
--- a/numpy/testing/_private/utils.py
+++ b/numpy/testing/_private/utils.py
@@ -172,7 +172,7 @@ elif sys.platform[:5] == 'linux':
         """
         _proc_pid_stat = _proc_pid_stat or f'/proc/{os.getpid()}/stat'
         try:
-            with open(_proc_pid_stat) as f:
+            with open(_proc_pid_stat, encoding='utf-8') as f:
                 l = f.readline().split(' ')
             return int(l[22])
         except Exception:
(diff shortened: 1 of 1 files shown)


jorenham:

Lawson-Darrow:
Sorry about that! Happy to reopen manually and follow the template per the policy if that's alright with you!

jorenham:
Feel free to open a new PR

PR #31552, opened 2026-06-02
DOC: Add element-wise repeat example to numpy.repeat

lciti:
The current examples in the numpy.repeat docstring only show cases where broadcasting occurs.

This new example demonstrates the less obvious case where repeats is a sequence with the same length as the input array, allowing each element to be repeated a different number of times without broadcasting.

### PR summary
See above.

I am an academic and a long-time numpy user (approx 2008).
Numpy is part of my research toolboxes in machine learning and biomedical engineering.

#### AI Disclosure
AI was used to refine the wording of the commit message.


seberg:
Thanks, the other examples might also help from a refresher (and especially explanation with `axis=None`), but adding this example seems like an improvement.

PR #31554, opened 2026-06-02
MAINT: reflect msvc lower bound in error message

h-vetinari:
This was missed in #30489, which raised
```diff
-  if not cc.version().version_compare('>=19.20')
+  if not cc.version().version_compare('>=19.35')
```
without updating the corresponding error message; VS2019 corresponds to the 14.2 toolset and 19.20 compiler; VS2022 corresponds to 14.3 toolset and 19.3x compiler

CC @kumaraditya303


seberg:
Visual studio versions always confuse me, but looks all correct, thanks!

PR #31557, opened 2026-06-03
BUG: Fix order of real and imaginary NaNs when sorting complex in descending order

MaanasArora:
Fixes a bug from #31345 where complex numbers with only imaginary nans get sorted _before_ those with only real nans in descending sorts, when keeping in with our NaN semantics they should be sorted _after_ (real nans are the primary key). This should probably be backported to 2.5, cc @seberg @charris

The longstanding `less` implementation is a bit confusing, but I didn't refactor/simplify and kept this change minimal. It would be good to confirm this is the correct behavior, though, as the `greater` is a bit hard to read (hopefully less so now):

1. if greater, only push to right if the imaginary part is strictly NaN but the other is not,
2. if less, only push to right if the _other_ imaginary part is strictly NaN but _this_ one is not,
3. If real parts are equal or _both_ are NaN, break the tie by float-like comparison of imaginary,
4. Failing all (i.e., only _one_ of the real parts is NaN), push to the right only if first is NaN.

So for the real part at least:

* If both reals are non-NaN -> branch 1-2
* If exactly one real is NaN -> branch 4
* If both reals are NaN -> branch 3

Sorry, verbose, but hopefully that clears things, thank you for reviewing!

#### AI Disclosure
No AI tools used.

### DIFF
--- a/numpy/_core/src/common/numpy_tag.hpp
+++ b/numpy/_core/src/common/numpy_tag.hpp
@@ -146,16 +146,16 @@ struct complex_type : complex_tag {
     {
         const auto ra = creal(a), rb = creal(b);
         const auto ia = cimag(a), ib = cimag(b);
-        if (ra > rb || (ra == ra && rb != rb)) {
+        if (ra > rb && (ra == ra && rb == rb)) {
             return ia == ia || ib != ib;
         }
-        if (ra < rb || (ra != ra && rb == rb)) {
+        if (ra < rb && (ra == ra && rb == rb)) {
             return ib != ib && ia == ia;
         }
         if (ra == rb || (ra != ra && rb != rb)) {
             return ia > ib || (ib != ib && ia == ia);
         }
-        return ra != ra;
+        return rb != rb;
     }
 };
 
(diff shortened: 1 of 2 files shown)


numpy/_core/src/common/numpy_tag.hpp
```diff
@@ -146,16 +146,16 @@ struct complex_type : complex_tag {
     {
         const auto ra = creal(a), rb = creal(b);
         const auto ia = cimag(a), ib = cimag(b);
-        if (ra > rb || (ra == ra && rb != rb)) {
+        if (ra > rb && (ra == ra && rb == rb)) {
```
seberg:
```suggestion
        if (ra > rb) {
```
guaranteed if comparison is true.

numpy/_core/src/common/numpy_tag.hpp
```diff
@@ -146,16 +146,16 @@ struct complex_type : complex_tag {
...
             return ia == ia || ib != ib;
         }
-        if (ra < rb || (ra != ra && rb == rb)) {
+        if (ra < rb && (ra == ra && rb == rb)) {
```
seberg:
```suggestion
        if (ra < rb) {
```
Same here.

seberg (approved):
Thanks, can simplify it more and align with the first, but I'll just apply that. Not sure what kind of merge mess caused this, but good to have the test fixed.

PR #31558, opened 2026-06-03
ENH: pixi-packages: use `skip-pyc-compilation` for `tsan-freethreading`

lucascolley:


lucascolley:
cc @ngoldbaum @crusaderky

ngoldbaum:
Hmm, ~30 minutes is still the slowest CI job by a good margin 🫤

lucascolley:
> Hmm, ~30 minutes is still the slowest CI job by a good margin 🫤

looks like ubuntu-arm and macOS are faster, perhaps we can merge if we skip ubuntu?

lucascolley:
let me see if this is any faster with adding `skip_pyc_compilation` for the CPython package too before doing that

lucascolley:
@ngoldbaum I've got it down to 24 mins, how close is that to acceptable?

ngoldbaum:
Looking at a recent CI run, the three slowest builds were:

* macOS x86-64 conda (3.12) - 35:36
* Build wheel cp312-manylinux_riscv64 - 26:24
* riscv64 - 22:40

So it would be slow, but not the slowest or even the second-slowest. So on that account it's probably OK to go in.

Can you explain again what the marginal benefit of having the CI run going is? I've never actually noticed these pixi CI jobs failing because of issues in the pixi setups so I'm not sure how often they'll pick up bugs that other CI jobs would miss.

@rgommers you usually have good intuition about this - what do you say about enabling this CI job?

lucascolley:
> Can you explain again what the marginal benefit of having the CI run going is? I've never actually noticed these pixi CI jobs failing because of issues in the pixi setups so I'm not sure how often they'll pick up bugs that other CI jobs would miss.

admittedly, the use of enabling it right now when the number of downstream users of these packages is O(0) (unless @crusaderky has been making industrious use of them?) is negligible. However, in a future where SciPy et al. are using these packages, I think it is significantly preferable that issues hit NumPy CI before they hit downstream.

As for what they will pick up that other jobs won't, I suppose it is problems specific to NumPy's interaction with the conda-forge compiler toolchains and these special build variants. Maybe the coverage elsewhere is good enough ™️ but I still feel like this is nice to have before we introduce these packages into SciPy's workspace.

lucascolley:
draft pending merge of:
- https://github.com/python/cpython/pull/152318 
- https://github.com/python/cpython/pull/152314

rgommers:
In general, 24 minutes seems fine to enable, and if it becomes a problem we reserve the right to change our minds. It's in the right ball park I'd say for a TSan job.

lucascolley:
not blocking here, but https://github.com/python/cpython/pull/152385 is also pretty exciting

Issue #14104, opened 2019-07-24
Broken dtype hasobject check


petrovlesha:
I am trying to convert integer columns from structured array to unstructured. But it fails at `_view_is_safe` func, which just check `hasobject` property that is not changed. It worked on 1.13.3 but broken after update to 1.16.0.

### Reproducing code example:

```python
import numpy as np
array = np.ones((100,), dtype=[('user', np.object), ('item', np.float), ('value', np.float)])
array[['item','value']].view((np.float, 2))
```
or
```python
from np.lib.recfunctions import structured_to_unstructured
structured_to_unstructured(array[['item','value']])

```

### Error message:

```
  File "/home/petrovalex/.virtualenvs/np/local/lib/python2.7/site-packages/numpy/core/_internal.py", line 494, in _view_is_safe
    raise TypeError("Cannot change data-type for object array.")
TypeError: Cannot change data-type for object array.
```

### Numpy/Python version information:
`('1.16.4', '2.7.3 (default, Oct 26 2016, 21:01:49) \n[GCC 4.6.3]')`

seberg:
Oh, interesting, it seems that the `.hasobject` flag is not cleared/rechecked when the field selection happens. That is indeed an issue, surprising nobody noticed before. It probably is a regression in 1.16, so should try to fix it sooner rather than later.

eric-wieser:
I think this is a feature, field selection of arrays container objects result in padding that contains objects

seberg:
Hmmm, that is right. In principle the padded type should not touch the other parts. But, it is possible to remove the padded areas again using `view`, so we cannot just allow this.

However, once a `copy` is made the `.hasobject` flag can and should be safely cleared. This is all pretty tricky stuff :/.

PR #31559, opened 2026-06-03
BUG: support object fields in structured_to_unstructured

### LINKED ISSUE #21990: BUG: `recfunctions.structured_to_unstructured` does not work with object fields
### Describe the issue:

As mentioned in #14104, multi-field views of structured arrays with object fields are forbidden because they could leave objects hidden in the padding. `recfunctions.structured_to_unstructured` does not have this problem because it always uses all the fields, yet it's forbidden from operating on arrays with object fields, even when doing a copy.

```python
from numpy.lib import recfunctions
import numpy as np

recfunctions.structured_to_unstructured(np.empty(0, 'O,O'), copy=True)
```

### Error message:

```shell
Input In [367], in <cell line: 4>()
      1 from numpy.lib import recfunctions
      2 import numpy as np
----> 4 recfunctions.structured_to_unstructured(np.empty(0, 'O,O'), copy=True)

File <__array_function__ internals>:180, in structured_to_unstructured(*args, **kwargs)

File ~/Documents/Scuola/lsqfitgp/Repository/lsqfitgp/pythonvenv/lib/python3.10/site-packages/numpy/lib/recfunctions.py:981, in structured_to_unstructured(arr, dtype, copy, casting)
    979 with suppress_warnings() as sup:  # until 1.16 (gh-12447)
    980     sup.filter(FutureWarning, "Numpy has detected")
--> 981     arr = arr.view(flattened_fields)
    983 # next cast to a packed format with all fields converted to new dtype
    984 packed_fields = np.dtype({'names': names,
    985                           'formats': [(out_dtype, dt.shape) for dt in dts]})

File ~/Documents/Scuola/lsqfitgp/Repository/lsqfitgp/pythonvenv/lib/python3.10/site-packages/numpy/core/_internal.py:494, in _view_is_safe(oldtype, newtype)
    491     return
    493 if newtype.hasobject or oldtype.hasobject:
--> 494     raise TypeError("Cannot change data-type for object array.")
    495 return

TypeError: Cannot change data-type for object array.
```

1.22.4 3.10.2 (v3.10.2:a58ebcc701, Jan 13 2022, 14:50:16) [Clang 13.0.0 (clang-1300.0.29.30)]

Lawson-Darrow:
### PR summary

Fixes gh-21990. `recfunctions.structured_to_unstructured` reinterprets the input
with a view (`arr.view(flattened_fields)`), which NumPy forbids for object dtypes
because objects could be hidden in field padding (gh-14104). But this function
always uses every field, so a copy is always safe.

When the dtype contains objects, the result is now built by casting each leaf
field to the output dtype and stacking them along a new trailing axis. This
matches the view path's column order (subarray elements outer, nested fields
inner) and per-field casting, and always returns a copy. Plain, subarray, nested,
structured-subarray, and zero-length-subarray object fields are handled; a new
regression test covers them and the existing test_recfunctions suite still passes.

#### AI Disclosure

This PR was prepared with AI assistance: the implementation and tests were drafted
using Claude Code (Anthropic), and the diff was additionally reviewed with OpenAI
Codex. I have reviewed and understand the change, validated it against the existing
test_recfunctions suite and a cross-check of the copy path against the canonical
view path, take full responsibility for it, and am submitting and conducting all
interactions personally.

### DIFF
--- a/numpy/lib/recfunctions.py
+++ b/numpy/lib/recfunctions.py
@@ -932,6 +932,27 @@ def _common_stride(offsets, counts, itemsize):
     return stride
 
 
+def _flat_field_arrays(arr):
+    """Yield each scalar leaf field of a structured array, in the same order
+    used by the view-based ``structured_to_unstructured`` path: fields
+    left-to-right, subarray elements outer, nested base fields inner. Each
+    yielded array has the same shape as ``arr``."""
+    for name in arr.dtype.names:
+        field = arr[name]
+        subshape = field.shape[arr.ndim:]
+        if subshape:
+            for idx in np.ndindex(subshape):
+                elem = field[(Ellipsis, *idx)]
+                if elem.dtype.names is not None:
+                    yield from _flat_field_arrays(elem)
+                else:
+                    yield elem
+        elif field.dtype.names is not None:
+            yield from _flat_field_arrays(field)
+        else:
+            yield field
+
+
 def _structured_to_unstructured_dispatcher(arr, dtype=None, copy=None,
                                            casting=None):
     return (arr,)
@@ -952,7 +973,8 @@ def structured_to_unstructured(arr, dtype=None, copy=False, casting='unsafe'):
     Parameters
     ----------
     arr : ndarray
-       Structured array or dtype to convert. Cannot contain object datatype.
+       Structured array or dtype to convert. If it contains object fields a
+       copy is always returned, since a view is not possible in that case.
     dtype : dtype, optional
        The dtype of the output unstructured array.
     copy : bool, optional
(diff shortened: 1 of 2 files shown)


seberg:
Thanks, but not sure I am convinced: If you go to this much trouble, you could also do so in the check that `view` does probably. And that would just make the normal path work _without_ the need for a copy.

Lawson-Darrow:
Good point, I looked into it and that does seem cleaner and avoids the copy. 

The blanket rejection lives in `_view_is_safe`, which raises whenever either dtype `hasobject`. In `structured_to_unstructured` the `flattened_fields` dtype keeps every object field at the same byte offset as the original. So no object pointer slot is gained, lost, or moved which is the real hazard behind gh-14104.

So the goal would be to allow the view when both dtypes have the same object pointer slots at the same byte offsets, and keep raising otherwise. The normal view path would then work for object fields here so the copy special case should not be needed. There would also be a small docstring update since the current text says object dtype is unsupported.

That's a broader change than this PR given it touches view safety for all object-containing dtypes, so I wanted to confirm that is the scope you would prefer before reworking. If so, I can move the fix into `_view_is_safe` with tests for the safe and unsafe object-slot cases.

seberg:
Unless it get's out of hand/complicated, yes I think that is the better fix.

Lawson-Darrow:
I started on the `_view_is_safe` relaxation but hit a wall so flagging before going further. 

Allowing same object offset views does not make `structured_to_unstructured`'s normal path work for object fields because that path goes through `arr.view(np.uint8)` to do its byte level stride trick (recfunctions.py near line 1079). Viewing object pointers as raw bytes is the gh-14104 hazard itself so that step raises and should keep raising regardless of the `_view_is_safe` rule.

A no-copy view is only possible for the narrow case of an all-object contiguous struct where you can skip the byte trick and view straight to an object dtype with a trailing axis. Arbitrary structured dtypes with mixed fields, padding, custom offsets, or subarrays still need the byte machinery which objects cannot use.

So for this fix I think the copy is the simpler correct option which keeps it from getting out of hand as you mentioned. The `_view_is_safe` relaxation might be reasonable on its own but it would not remove the copy here. 

Happy to add a narrow all-object no-copy fast path as a separate change if you think it is worth it.

Issue #29003, opened 2025-05-19
BUG: np.percentile fails with internal overflow when using float16 input on large arrays


friedlerido:
### Describe the issue:

When using np.percentile or np.nanpercentile on a large float16 array from real data (≈9 million elements), I get:
ValueError: kth(=-9223372036845518721) out of bounds (9257087)

This happens even for low percentiles like 20 or 50, and is resolved by casting the array to float64.

⚠️ Expected Behavior:
Either: NumPy automatically upcasts float16 to float64
Or: raises a UserWarning if float16 is passed to percentile

### Reproduce the code example:

```python
# This does not happen with random arrays of similar shape, but does happen consistently  #with my real-world data:
arr = np.load("example.npy")  # I can share this privately if needed
print(arr.dtype)  # float16
print(arr.shape)  # (9257087,)

np.percentile(arr, 50)  # → raises ValueError

# ✅ This works:
np.percentile(arr.astype(np.float64), 50)  # works fine
```

### Error message:

```shell
/opt/venv/lib/python3.11/site-packages/numpy/lib/_function_base_impl.py:107: RuntimeWarning: overflow encountered in cast
  get_virtual_index=lambda n, quantiles: (n - 1) * quantiles,
/opt/venv/lib/python3.11/site-packages/numpy/lib/_function_base_impl.py:4750: RuntimeWarning: overflow encountered in cast
  indexes_above_bounds = virtual_indexes >= valid_values_count - 1
/opt/venv/lib/python3.11/site-packages/numpy/lib/_function_base_impl.py:4655: RuntimeWarning: invalid value encountered in multiply
  lerp_interpolation = asanyarray(add(a, diff_b_a * t, out=out))
/opt/venv/lib/python3.11/site-packages/numpy/lib/_function_base_impl.py:4656: RuntimeWarning: invalid value encountered in scalar multiply
  subtract(b, diff_b_a * (1 - t), out=lerp_interpolation, where=t >= 0.5,
np.float16(nan)
```

### Python and NumPy Versions:

2.2.4
3.11.3 (main, May 23 2023, 13:34:03) [GCC 10.2.1 20210110]

### Runtime Environment:

[{'numpy_version': '2.2.4',
  'python': '3.11.3 (main, May 23 2023, 13:34:03) [GCC 10.2.1 20210110]',
[...]

PR #31560, opened 2026-06-04
DOC: fix median docstrings around float16/float32 inputs

### LINKED ISSUE #31486: BUG/DOC: `numpy.ma.median` return dtype does not match documented behavior
### Describe the issue:

According to the documentation for [`ma.median`](https://numpy.org/doc/stable/reference/generated/numpy.ma.median.html#numpy-ma-median): Return data-type is `float64` for integers and floats smaller than `float64`, or the input data-type, otherwise.

However, in the following example, an input array with dtype `float16` produces a result with dtype `float16` rather than `float64`.

I have observed a similar mismatch with the documentation in: `numpy.nanpercentile`, `numpy.nanquantile`,  and `numpy.quantile`. I would be happy to provide minimal reproducing examples for them as well if helpful. 

Could this be an outdate on the documentation? 

Related to: #29003

```python
import numpy as np
import numpy.ma as ma

arr = np.array([0, 1, 2, 3], dtype=np.float16)
m = ma.median(arr)

print("value:", m)
print("dtype:", np.asarray(m).dtype)
```

### Error message:

```shell
dtype: float16 --> This result does not appear to match the documented return dtype behavior.
```

2.4.3

### Runtime Environment:

_No response_

Before reporting it, we reviewed the code, documentation and checked the latest development version of Numpy to confirm it.

Atharva2012:
**PR summary**
np.ma.median does not promote float16 and float32 inputs to float64 as documented. The docstring states the return type should be float64 for floats smaller than float64, but the result stays in the input dtype instead.
The bug is in _median in numpy/ma/extras.py. Two code paths,  the 1D path and the N-D path, use np.true_divide without first casting to the correct output dtype. The fix computes out_dtype = np.result_type(asorted.dtype, np.float64) and casts before dividing in both paths.

**UPDATE**: After reviewer feedback, it was found that both np.median and np.ma.median return the same dtype as the input consistently. The issue is in the docstring, not the code. This PR now corrects the docstring in numpy/ma/extras.py to accurately reflect the actual behavior.
  
 Closes #31486 Related to: #29003 

**First time committer introduction**
I am Atharva, I use NumPy for academic projects, but it's my first time contributing to this open source project. I kept the change small and focused on a single well-defined bug.

** AI Disclosure**
AI was used to reword the PR description in more suitable language.

### DIFF
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -3948,10 +3948,9 @@ def median(a, axis=None, out=None, overwrite_input=False, keepdims=False):
     Returns
     -------
     median : ndarray
-        A new array holding the result. Return data-type is "float64" 
-        for integer input. For floating-point input, the return data-type 
-        is the same as the input data-type. If `out` is specified, that 
-        array is returned instead.
+        A new array holding the result. Return data-type is ``float64``
+        for integer input, or the input data-type, otherwise. If `out` is
+        specified, that array is returned instead.
 
     See Also
     --------
@@ -4146,10 +4145,9 @@ def percentile(a,
         is a scalar. If multiple percentiles are given, first axis of
         the result corresponds to the percentiles. The other axes are
         the axes that remain after the reduction of `a`. If the input
-        contains integers or floats smaller than ``float64``, the output
-        data-type is ``float64``. Otherwise, the output data-type is the
-        same as that of the input. If `out` is specified, that array is
-        returned instead.
+        contains integers, the output data-type is ``float64``. Otherwise,
+        the output data-type is the same as that of the input. If `out` 
+        is specified, that array is returned instead.
 
     See Also
     --------
@@ -4349,10 +4347,9 @@ def quantile(a,
         is a scalar. If multiple probability levels are given, first axis
         of the result corresponds to the quantiles. The other axes are
         the axes that remain after the reduction of `a`. If the input
-        contains integers or floats smaller than ``float64``, the output
-        data-type is ``float64``. Otherwise, the output data-type is the
-        same as that of the input. If `out` is specified, that array is
-        returned instead.
+        contains integers, the output data-type is ``float64``. Otherwise,
+        the output data-type is the same as that of the input. If `out` is
+        specified, that array is returned instead.
 
     See Also
     --------
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -3948,11 +3948,10 @@ def median(a, axis=None, out=None, overwrite_input=False, keepdims=False):
     Returns
     -------
     median : ndarray
-        A new array holding the result. If the input contains integers
-        or floats smaller than ``float64``, then the output data-type is
-        ``np.float64``.  Otherwise, the data-type of the output is the
-        same as that of the input. If `out` is specified, that array is
-        returned instead.
+        A new array holding the result. Return data-type is "float64" 
+        for integer input. For floating-point input, the return data-type 
+        is the same as the input data-type. If `out` is specified, that 
+        array is returned instead.
 
     See Also
     --------
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -3952,7 +3952,7 @@ def median(a, axis=None, out=None, overwrite_input=False, keepdims=False):
         the output data-type is ``float64``. Otherwise, the output data-type
         is the same as that of the input. If `out` is specified, that array is
         returned instead.
-        
+
     See Also
     --------
     mean, percentile
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -3948,10 +3948,11 @@ def median(a, axis=None, out=None, overwrite_input=False, keepdims=False):
     Returns
     -------
     median : ndarray
-        A new array holding the result. Return data-type is ``float64``
-        for integer input, or the input data-type, otherwise. If `out` is
-        specified, that array is returned instead.
-
+        A new array holding the result. If the input contains integers,
+        the output data-type is ``float64``. Otherwise, the output data-type
+        is the same as that of the input. If `out` is specified, that array is
+        returned instead.
+        
     See Also
     --------
     mean, percentile
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -4147,7 +4147,7 @@ def percentile(a,
         the result corresponds to the percentiles. The other axes are
         the axes that remain after the reduction of `a`. If the input
         contains integers, the output data-type is ``float64``. Otherwise,
-        the output data-type is the same as that of the input. If `out` 
+        the output data-type is the same as that of the input. If `out`
         is specified, that array is returned instead.
 
     See Also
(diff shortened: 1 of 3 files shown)


Atharva2012:
@seberg This is my first PR for this project. Could use your opinion.

seberg:
Sorry, but this is the wrong approach.  You should make sure that `ma.median` mirrors normal `median` and if you check that you'll see that if anything this seems like a documentation issue.
(Although the code could call `mean()` for clarity, although it doesn't matter, the NumPy code does so mainly for some `astropy` compat stuff that I doubt is super important here.)

Atharva2012:
@seberg Thank you for the feedback! After checking, both np.median and np.ma.median return the same dtype as the input consistently, so this is a documentation issue. I've updated the docstring to accurately reflect the actual behavior.

numpy/lib/_function_base_impl.py
```diff
@@ -3948,11 +3948,10 @@ def median(a, axis=None, out=None, overwrite_input=False, keepdims=False):
     Returns
     -------
     median : ndarray
-        A new array holding the result. If the input contains integers
-        or floats smaller than ``float64``, then the output data-type is
-        ``np.float64``.  Otherwise, the data-type of the output is the
-        same as that of the input. If `out` is specified, that array is
-        returned instead.
+        A new array holding the result. Return data-type is "float64" 
+        for integer input. For floating-point input, the return data-type
```
MaanasArora: This excludes any other dtypes than integers and floats, right? If this is correct, I think it makes more sense to special-case integers yes, but not floats (say "otherwise" for same dtype as in the original.)

Atharva2012: @MaanasArora Thankyou for your input, I've made the changes and will push them shortly for review

Atharva2012:
@seberg Following the discussion here, and all the related pull request, I've went ahead and modified the percentile, quantile, nanpercentile and nanquantile docstrings to reflect the actual behavior, as they're mentioned in the original issue.

@MaanasArora Thankyou for your suggestion, I've modified the wording in the docstring so that it's easier to understand.

numpy/lib/_function_base_impl.py
```diff
@@ -3948,11 +3948,9 @@ def median(a, axis=None, out=None, overwrite_input=False, keepdims=False):
     Returns
     -------
     median : ndarray
-        A new array holding the result. If the input contains integers
-        or floats smaller than ``float64``, then the output data-type is
-        ``np.float64``.  Otherwise, the data-type of the output is the
-        same as that of the input. If `out` is specified, that array is
-        returned instead.
+        A new array holding the result. Return data-type is ``float64``
+        for integer input, or the input data-type, otherwise. If `out` is
```
MaanasArora:
```suggestion
        A new array holding the result. For integer input, the output
        data-type is ``float64``. Otherwise, the output data-type
        is the same as that of the input. If `out` is specified, that
        array is returned instead.
```

This is worded differently from the other functions.

numpy/ma/extras.py
```diff
@@ -710,8 +710,9 @@ def median(a, axis=None, out=None, overwrite_input=False, keepdims=False):
     median : ndarray
         A new array holding the result is returned unless out is
         specified, in which case a reference to out is returned.
-        Return data-type is `float64` for integers and floats smaller than
-        `float64`, or the input data-type, otherwise.
+        Return data-type is "float64" for integer input. For 
+        floating-point input, the return data-type is the same as 
+        the input data-type.
```
MaanasArora:
```suggestion
        For integer input, the output data-type is ``float64``.
        Otherwise, the output data-type is the same as that
        of the input.
```

MaanasArora (approved):
Thank you, just a wording suggestion for consistency (`np.median` reads a bit unnaturally), otherwise looks good to me. I think this is correct looking at the [underlying logic](https://github.com/numpy/numpy/blob/1916cfb1625c479dea2f01112559274c49c77ab8/numpy/_core/_methods.py#L125-L144).

numpy/lib/_function_base_impl.py
Atharva2012: Thanks for the suggestion! I think I'll keep the original phrasing here since it conveys the same information a bit more directly and concisely. But happy to discuss further if you feel strongly otherwise.

MaanasArora: That's fine, but then I think this should be the same for the other functions (is there a reason why it isn't)? I pulled this phrasing from the other functions. (I don't think the `input data-type, otherwise` is the right tone, but otherwise this is OK.)

numpy/ma/extras.py
```diff
@@ -710,9 +710,9 @@ def median(a, axis=None, out=None, overwrite_input=False, keepdims=False):
     median : ndarray
         A new array holding the result is returned unless out is
         specified, in which case a reference to out is returned.
-        Return data-type is `float64` for integers and floats smaller than
-        `float64`, or the input data-type, otherwise.
-
+        If the input contains integers, the output data-type is ``float64``.
+        Otherwise, the output data-type is the same as that of the input.
+        If `out` is specified, that array is returned instead.
     See Also
```
MaanasArora, line 717: nit: missing newline

numpy/lib/_function_base_impl.py
```diff
@@ -3948,12 +3948,11 @@ def median(a, axis=None, out=None, overwrite_input=False, keepdims=False):
...
-        A new array holding the result. If the input contains integers
-        or floats smaller than ``float64``, then the output data-type is
-        ``np.float64``.  Otherwise, the data-type of the output is the
-        same as that of the input. If `out` is specified, that array is
+        A new array holding the result. If the input contains integers,
+        the output data-type is ``float64``. Otherwise, the output data-type
+        is the same as that of the input. If `out` is specified, that array is
         returned instead.
-
+
```
MaanasArora: nit: extra whitespace

MaanasArora (commented):
Thank you, just linting issues now!

Atharva2012:
@seberg Did my best to address the original issue here, do let me know if anything's missing or needs polishing. Thanks.

MaanasArora:
@Atharva2012 Sorry, just noticed we are still missing `nanmedian`, could you please change that too. Also could you fix the linting issue (there is trailing whitespace in CI). Thanks!

Atharva2012:
Hey @MaanasArora , I fixed the nanmedian function, but could you help me find where the extra trailing whitespace is?

Edit: Got it, fixing it now

MaanasArora (approved):
Thanks @Atharva2012, I just pushed a minor grammar fix. This seems to be a sensible clarification. (N.B. This is actually not fully technically correct for e.g. `quantile(method="nearest")`, but this is fine for defaults so I don't think we need to include that, though open to other opinions).

I'm not sure if we need a test to check the dtype behavior actually, we might, so I'll wait to see if someone has an opinion on that, but otherwise looks good to me now.

Atharva2012:
@MaanasArora Thank you for the review and the grammar fix! Glad the clarification makes sense. Happy to wait on the test question and make any additional changes if needed.

MaanasArora:
LGTM, thanks. The edge-case with non-interpolating methods (`closest_observation`, `nearest` and the other `linear` variations) not promoting integers really seems niche and may be confusing in this section, so I'll just merge as-is.

PR #31562, opened 2026-06-04
BUG: For new-style wrapped legacy dtypes use old array printing path

seberg:
For such dtypes let's keep things mostly as things were at least on NumPy 2.5+.  In practice we may want to make things a nice more flexible here for all user-dtypes.

This doesn't change anything otherwise, the `__repr__` test was guaranteed equivalent to the new check except for this new path.

(No AI use)

### DIFF
--- a/numpy/_core/arrayprint.py
+++ b/numpy/_core/arrayprint.py
@@ -1576,7 +1576,7 @@ def dtype_short_repr(dtype):
     >>> dt = np.int64([1, 2]).dtype
     >>> assert eval(dtype_short_repr(dt)) == dt
     """
-    if type(dtype).__repr__ != np.dtype.__repr__:
+    if not type(dtype)._legacy:
         # TODO: Custom repr for user DTypes, logic should likely move.
         return repr(dtype)
     if dtype.names is not None:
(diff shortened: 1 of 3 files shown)


numpy/_core/arrayprint.py
```diff
@@ -1576,7 +1576,7 @@ def dtype_short_repr(dtype):
     >>> dt = np.int64([1, 2]).dtype
     >>> assert eval(dtype_short_repr(dt)) == dt
     """
-    if type(dtype).__repr__ != np.dtype.__repr__:
+    if not type(dtype)._legacy:
```
jorenham, line 1579: Hmm, maybe we should add this as `_legacy: ClassVar[bool]` to the `dtype` stubs

seberg: Maybe, so long it doesn't make it official that we guarantee it as public API.  We also have `_numeric` and `_abstract` right now.

MaanasArora (approved):
Looks good to me! Conservative behavior makes sense for wrapped dtypes, and I think the logic here is also cleaner until we move.

ngoldbaum:
I agree, in it goes. Thanks @seberg!

PR #31563, opened 2026-06-04
BUG,ENH: Relax finfo to be easier accessible for all user dtypes

seberg:
This does two small tweaks:
1. Remove the requirement for `inexact` scalar derivation (because I had not realized that this is problematic for bfloat16, etc.).
   * This makes the error message very slightly nicer for non-float finfo inputs, I expect that is OK. (This means we require all finfo fields to be filled in right now which quaddtype does as the only user, we could omit the decimal precision or so if someone asks for it.)
2. For complex dtypes rather than hard-code ours, assume that `arr.imag` succeeds and use `arr.imag.dtype` when it is a view to find the "real" dtype. After that, we assume that `finfo` filling will succeed.

Together (with the "new dtype backporting") this will allow `ml_dtypes` to support `np.finfo()` fully.
(It is very annoying to test it here, but we run `ml_dtypes` tests now and they will test it as soon as we start using it there.)

(No AI use.)

CC @SwayamInSync since you were involved in the first part.  I don't think requiring all "finfo" fields that we currently have in C to be filled should be a problem, right?
(If that changes, there is always the work-around to fill it with NaN in practice, probably.)

(I find the `obj2sctype` annoying and potentially fishy even for quaddtype since it is slightly parametric(?), but that part I don't dare to touch here.)

---

FWIW, I would like to consider this a bug-fix and backport.  The only change should be the error message change in practice.

### DIFF
--- a/numpy/_core/getlimits.py
+++ b/numpy/_core/getlimits.py
@@ -212,7 +212,8 @@ class finfo:
         if obj is not None:
             return obj
         dtypes = [dtype]
-        newdtype = ntypes.obj2sctype(dtype)
+        # Call result_type to normalize to e.g. native byte-order:
+        newdtype = numeric.result_type(dtype)
         if newdtype is not dtype:
             dtypes.append(newdtype)
             dtype = newdtype
@@ -220,7 +221,9 @@ class finfo:
         obj = cls._finfo_cache.get(dtype)
         if obj is not None:
             return obj
-        if not issubclass(dtype, numeric.floating):
+
+        sctype = newdtype.type
+        if sctype is not None and not issubclass(sctype, numeric.floating):
             newdtype = _convert_to_float_if_complex(dtype)
             if newdtype is not dtype:
                 # dtype changed, for example from complex128 to float64
--- a/numpy/_core/getlimits.py
+++ b/numpy/_core/getlimits.py
@@ -10,7 +10,7 @@ from functools import cached_property
 from numpy._utils import set_module
 
 from . import numeric, numerictypes as ntypes
-from ._multiarray_umath import _populate_finfo_constants
+from ._multiarray_umath import _finfo_get_realdtype, _populate_finfo_constants
 
 
 def _fr0(a):
@@ -25,31 +25,6 @@ def _fr1(a):
     if a.size == 1:
         a = a.reshape(())
     return a
-
-
-def _convert_to_float_if_complex(dtype):
-    # At this point, we assume the array may be a complex array for which we
-    # need to find the real dtype counterpart.
-    # In C, we could short-circuit but this is cached so create an array and
-    # check arr.real and arr.imag.
-    # A complex compsed of (real, imag) will:
-    # - return a view for both arr.real and arr.imag
-    #   (non-complex will not do this)
-    # - return the identical dtype for both arr.real and arr.imag
-    #   (might reject theoretical crazy complex dtypes with mixed precision)
-    # This may be over-careful, but let's err on the safe side.
-    try:
-        arr = numeric.empty(1, dtype=dtype)
-        imag_part = arr.imag
-        real_part = arr.real
-    except Exception:
-        # if array creation or arr.imag fails, assume it's not a complex.
-        pass
-    else:
-        if (imag_part.base is real_part.base is arr
-                and imag_part.dtype == real_part.dtype):
-            return imag_part.dtype
-    return dtype
 
 
 # Parameters for creating MachAr / MachAr-like objects
@@ -224,7 +199,7 @@ class finfo:
 
         sctype = newdtype.type
         if sctype is not None and not issubclass(sctype, numeric.floating):
-            newdtype = _convert_to_float_if_complex(dtype)
+            newdtype = _finfo_get_realdtype(dtype)
             if newdtype is not dtype:
                 # dtype changed, for example from complex128 to float64
                 dtypes.append(newdtype)
--- a/numpy/_core/getlimits.py
+++ b/numpy/_core/getlimits.py
@@ -27,11 +27,30 @@ def _fr1(a):
     return a
 
 
-_convert_to_float = {
-    ntypes.csingle: ntypes.single,
-    ntypes.complex128: ntypes.float64,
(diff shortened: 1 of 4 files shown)


numpy/_core/getlimits.py
```diff
@@ -27,11 +27,30 @@ def _fr1(a):
...
+    # A complex compsed of (real, imag) will:
+    # - return a view for both arr.real and arr.imag
+    #   (non-complex will not do this)
+    # - return the identical dtype for both arr.real and arr.imag
+    #   (might reject theoretical crazy complex dtypes with mixed precision)
+    # This may be over-careful, but let's err on the safe side.
+    try:
+        arr = numeric.empty(1, dtype=dtype)
+        imag_part = arr.imag
+        real_part = arr.real
```
seberg: To be clear, my first plan for this function was to guard it with the AbstractDType hierarchy, but since that is stalled behind heap-types, etc. and the `.imag` path is good and used either way, that guard felt a bit unnecessary.
(`ml_dtypes` cannot inherit from the scalar hierarchy easily, I think because it would mean inheriting the scalar methods and they make wrong assumptions.)

SwayamInSync:
> I don't think requiring all "finfo" fields that we currently have in C to be filled should be a problem, right?

Yes, no problem, we return 1 for all six now-mandatory float constants (and the four int ones), so quaddtype is unaffected.
And yeah there bf16 is inherited from generic, I think it'll be okay to remove the finfo from dtype class hierarchy dependency and make a developer choice.

SwayamInSync:
> (I find the obj2sctype annoying and potentially fishy even for quaddtype since it is slightly parametric(?), but that part I don't dare to touch here.)

I didn't thought about it earlier but yeah that logic routes the input to default scalar type. I think we can leave `obj2sctype` as it is (it does what it meant) and add a caching path in `finfo.new` that preserve the backend info? (out of scope for this PR, but worth discussing)

seberg:
Actually, pushed to use `np.result_type(dtype)` rather than the current code instead.  Basically, by this time we have a dtype instance, so the only thing left to do is to normalize away the byte-order.
`np.result_type()` does that better than `obj2sctype`, so we can just do that.  And while not needed, it seems rather OK to just do it here.

numpy/_core/src/multiarray/multiarraymodule.c
ngoldbaum: Shouldn't there be a new error here to match the one you added in the other branch?

numpy/_core/getlimits.py
```diff
@@ -27,11 +27,30 @@ def _fr1(a):
...
+    except Exception:
+        # if array creation or arr.imag fails, assume it's not a complex.
+        pass
```
ngoldbaum: I'm not a fan of this `Except: pass` construct since it can make it really hard to debug a situation where a developers triggers the exception. They would need to spelunk through the NumPy implementation to understand what is happening.

Wouldn't it be better to do something like `imag_part = getattr(arr, "imag", None)` and then handle the case when `imag_part` is None?

I would also probably want to know that I passed in a type that has dtype that isn't usable to create an array scalar rather than have the `ValueError` silently swallowed.

numpy/_core/tests/test_getlimits.py
```diff
@@ -66,7 +66,19 @@ def test_basic(self):
...
+    def test_byteorder(self, dt1, dt2):
+        # finfo should normalize to native byte-order.
+        assert_finfo_equal(finfo(dt1), finfo(dt2))
+
+    @pytest.mark.parametrize('dt', [np.int8, "V3", "S3", "f,f"])
+    def test_rejects_others(self, dt):
+        dtype = np.dtype(dt)
+        with pytest.raises(ValueError,
+                match=r"data type .* not compatible with finfo"):
+            finfo(dtype)
```
ngoldbaum, line 83: Maybe a test using the legacy rational dtype that exercises the error path only allowing new user dtypes?

ngoldbaum (commented):
Should the new `.real` and `.imag` heuristic be documented somewhere?

I also spotted a couple issues inline. I used an AI model to help with the review.

numpy/_core/getlimits.py
seberg: > Wouldn't it be better to do something like imag_part = getattr(arr, "imag", None) and then handle the case when imag_part is None?

Yeah, I hate `except Exception:` too, but wasn't sure if a `TypeError` is broad enough.
`getattr` could effectively call a ufunc.  Now, when it does, it's definitely not a compatible `complex`!
But in Python I can't know that, so the `try/except` is to avoid lowering into C (which is the alternative).

I think it should be fairly graceful here (because it almost certainly just raises "dtype not compatible with finfo"), but if you dislike it enough... I guess we can make a "try to fetch the real dtype" C level function.

numpy/_core/src/multiarray/multiarraymodule.c
seberg: oops, good point.  Yeah, should error out just the same way here also (i.e. for now require everything to be filled)

numpy/_core/getlimits.py
seberg: OK, I moved it to C, it's not too bad.

numpy/_core/tests/test_getlimits.py
seberg: The path isn't strictly limited to new user dtypes and rational isn't special, but I added it and object and StringDType as well here (object actually adds real coverage since it isn't view compatible).

ngoldbaum (approved):
It'd still be nice to get a documentation mention somewhere that the `.imag`/`.real` heuristic exists, but I'm happy for that to be a followup.

seberg:
Ooops, just forgot about the docs. Added two sentences to the constants docs about finfo with a side-note for complex (I don't think this belongs in `finfo`, so while not very discoverable it feels like a decent spot).

charris:
Going to push this off to 2.5.1.

seberg:
Argg, I shoudl have/could have just merged this, sorry I thought Nathan would see the docs update briefly.
I guess it's fine for backport, though since it has no effect on our dtypes anyway.

PR #31565, opened 2026-06-04
BUG: avoid SystemError in np.dot object loop when an error is already set

### LINKED ISSUE #29688: BUG: `SystemError` from np.dot when given certain incompatible types.
### Describe the issue:

This was observed in SciPy: https://github.com/scipy/scipy/issues/23561. That was with numpy 2.2.6, but the problem occurs in the main development branch too.

Here's a simple reproducer:

```
In [52]: np.__version__
Out[52]: '2.4.0.dev0+git20250905.8339bce'

In [53]: a = np.zeros((2, 2), dtype=complex)

In [54]: b = np.array([[np.timedelta64(123, 's')], [np.timedelta64(321, 's')]], dtype=object)

In [55]: np.dot(a, b)
UFuncTypeError                            Traceback (most recent call last)
UFuncTypeError: ufunc 'multiply' cannot use operands with types dtype('complex128') and dtype('<m8[s]')

SystemError                               Traceback (most recent call last)
Cell In[55], line 1
----> 1 np.dot(a, b)

SystemError: <built-in method __subclasscheck__ of numpy._DTypeMeta object at 0x7936970ce500> returned a result with an exception set
```
The `TypeError` looks reasonable, but the `SystemError` does not.

```python
See above.
```

### Error message:

```shell
See above.
```

2.4.0.dev0+git20250905.8339bce
3.13.7 (main, Sep  4 2025, 21:56:53) [GCC 11.4.0]

### Runtime Environment:

_No response_

### Context for the issue:

_No response_

Lawson-Darrow:
### PR summary
`np.dot` on certain incompatible types raises a confusing `SystemError` instead of the real `TypeError` (gh-29688):

```python
a = np.zeros((2, 2), dtype=complex)
b = np.array([[np.timedelta64(123, "s")], [np.timedelta64(321, "s")]], dtype=object)
np.dot(a, b)   # SystemError: ... returned a result with an exception set
```

`OBJECT_dot` has a void return and can't signal failure, so on an unsupported element multiply it sets the exception and returns. But `matrixproduct2` re-enters `OBJECT_dot` for later output slots with the exception still pending, and running Python C-API calls with an error set is what surfaces as the `SystemError`. Per @seberg's suggestion in the issue, this checks `PyErr_Occurred()` at the top of `OBJECT_dot` and bails, preserving the original error; the repro then raises the expected `TypeError`. Adds a regression test.

@seberg asked to confirm the guard is unnoticeable (it is one `PyErr_Occurred()` per call). Object-dtype `np.dot`, before vs after the guard:

| n | before | after |
|------|---------------|---------------|
| 1 | 0.613 us/call | 0.630 us/call |
| 8 | 4.391 us/call | 4.401 us/call |
| 1000 | 567.7 us/call | 563.9 us/call |

Differences are within run-to-run noise.

Closes #29688.

#### AI Disclosure
AI assisted with this PR. Tools: Claude Code (Anthropic) and OpenAI Codex. They helped locate the missing `PyErr_Occurred()` check in `OBJECT_dot` and draft the fix, the regression test, and this PR text. I reviewed and verified the change, built NumPy locally to confirm the repro is fixed and the test passes, ran the before/after benchmark, understand it, and can explain it.

### DIFF
--- a/numpy/_core/tests/test_multiarray.py
+++ b/numpy/_core/tests/test_multiarray.py
@@ -3889,6 +3889,14 @@ class TestMethods:
         assert_raises(TypeError, np.dot, c, A)
         assert_raises(TypeError, np.dot, A, c)
 
+    def test_dot_object_unsupported_multiply(self):
+        # gh-29688: an unsupported object multiply must raise the real error
+        # (TypeError), not a SystemError.
+        a = np.zeros((2, 2), dtype=complex)
+        b = np.array([[np.timedelta64(123, "s")],
+                      [np.timedelta64(321, "s")]], dtype=object)
+        assert_raises(TypeError, np.dot, a, b)
+
     def test_dot_out_mem_overlap(self):
         np.random.seed(1)
 
--- a/numpy/_core/src/multiarray/arraytypes.c.src
+++ b/numpy/_core/src/multiarray/arraytypes.c.src
@@ -3646,6 +3646,14 @@ OBJECT_dot(char *ip1, npy_intp is1, char *ip2, npy_intp is2, char *op, npy_intp 
     npy_intp i;
     PyObject *tmp1, *tmp2, *tmp = NULL;
     PyObject **tmp3;
+    /*
+     * np.dot can re-enter this loop with an error already set from an earlier
+     * failed element operation; bail out so we don't run with a pending
+     * exception and raise a confusing SystemError. See gh-29688.
+     */
+    if (PyErr_Occurred()) {
+        return;
+    }
     for (i = 0; i < n; i++, ip1 += is1, ip2 += is2) {
         if ((*((PyObject **)ip1) == NULL) || (*((PyObject **)ip2) == NULL)) {
             tmp1 = Py_False;


seberg:
I don't expect a noticable difference, but what did you time there exactly?

Lawson-Darrow:
I timed `np.dot(a, b)` where `a` and `b` are 1-D object arrays of `Fraction` values, at n = 1, 8, and 1000, before and after the guard on the same build (baseline is the same tree with the guard reverted). The guard runs once per `OBJECT_dot` call, not per element, so the cost is per-call. Even at n = 1 the two were within run-to-run noise (0.613 vs 0.630 us/call), and at n = 1000 they matched within noise. Happy to rerun a different way if you prefer.

numpy/_core/tests/test_multiarray.py
```diff
@@ -3889,6 +3889,14 @@ def test_dot_type_mismatch(self):
         assert_raises(TypeError, np.dot, c, A)
         assert_raises(TypeError, np.dot, A, c)
 
+    def test_dot_object_unsupported_multiply(self):
+        # gh-29688: an unsupported object multiply must raise the real error
+        # (TypeError), not a SystemError.
+        a = np.zeros((2, 2), dtype=complex)
+        b = np.array([[np.timedelta64(123, "s")],
+                      [np.timedelta64(321, "s")]], dtype=object)
+        assert_raises(TypeError, np.dot, a, b)
```
seberg, line 3898: FWIW, `with pytest.raises()` is nicer.

seberg (approved):
Thanks, let's just do this then.

FWIW, fractions are rather slow objects, but even with 10x faster objects I guess it doesn't matter much.

PR #31566, opened 2026-06-04
BUG: accept NumPy integer scalars in datetime64

### LINKED ISSUE #10004: datetime64/timedelta64 do not accept numpy integer types
Example:
```python
>>> np.datetime64(1478476800, "s")
numpy.datetime64('2016-11-07T00:00:00')

>>> np.datetime64(np.int64(1478476800), "s")
ValueError: Could not convert object to NumPy datetime

>>> np.timedelta64(1, "s")
numpy.timedelta64(1,'s')

>>> np.timedelta64(np.int64(1), "s")
ValueError: Could not convert object to NumPy timedelta
```

Ijtihed:
### PR summary

Closes gh-10004.

This fixes construction of `datetime64` scalars from NumPy integer scalars when an explicit datetime unit is provided. So `np.datetime64(np.int64(1478476800), "s")` now matches `np.datetime64(1478476800, "s")`.

The change mirrors the existing Python integer constructor behavior and preserves the current rule that integer-based `datetime64` construction requires an explicit unit. Construction without a unit or with `"generic"` still raises theexplicit-unit `ValueError`.

Tests cover signed and unsigned NumPy integer scalar construction, missing/generic unit errors and unsigned overflow behavior

#### AI Disclosure

I've used codex to run through my code and make sure I've missed nothing stupid

### DIFF
--- a/numpy/_core/src/multiarray/datetime.c
+++ b/numpy/_core/src/multiarray/datetime.c
@@ -2402,6 +2402,19 @@ convert_pyobject_to_datetime(PyArray_DatetimeMetaData *meta, PyObject *obj,
         }
         return 0;
     }
+    /* Do no conversion on NumPy integer scalars */
+    else if (PyArray_IsScalar(obj, Integer)) {
+        if (meta->base == NPY_FR_ERROR || meta->base == NPY_FR_GENERIC) {
+            PyErr_SetString(PyExc_ValueError, "Converting an integer to a "
+                            "NumPy datetime requires a specified unit");
+            return -1;
+        }
+        *out = PyLong_AsLongLong(obj);
+        if (error_converting(*out)) {
+            return -1;
+        }
+        return 0;
+    }
     /* Datetime scalar */
     else if (PyArray_IsScalar(obj, Datetime)) {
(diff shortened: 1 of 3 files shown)


numpy/_core/src/multiarray/datetime.c
```diff
@@ -2402,6 +2402,19 @@ convert_pyobject_to_datetime(PyArray_DatetimeMetaData *meta, PyObject *obj,
         }
         return 0;
     }
+    /* Do no conversion on NumPy integer scalars */
+    else if (PyArray_IsScalar(obj, Integer)) {
```
seberg: Hmmmpf, it would be nice to just use `operator.index`... But unfortunately, I guess that would have to happen quite late.
Either way, why duplicate all of the code below, though?

But even then, I find it a bit funny that this way, I don't think we support `np.array(0)` (i.e. 0-D integer arrays).

Ijtihed: I think its a good point on all three and were oversights from me

I collapsed it into one branch using PyNumber_Index and also went ahead and added 0-D integer array support

I also added testing forfor construction, missing unit, and overflow since I didnt have them initially. thanks!

PR #31573, opened 2026-06-05
PERF, BUG: speed up conversion of Python datetime/date to datetime64

eendebakpt:
`NpyDatetime_ConvertPyDateTimeToDatetimeStruct` is the per-element worker behind casting object arrays of Python datetimes to `datetime64`, e.g. `np.asarray(list_of_datetimes).astype("M8[us]")` (the floor of matplotlib's `date2num`, and relevant to pandas/xarray too).

This PR makes two complementary changes:

- **Fast path:** for genuine `datetime.date` / `datetime.datetime` objects
  (including subclasses such as pandas `Timestamp`), read the fields straight from
  the CPython C struct via the datetime C-API macros 
- **Cheaper attribute access** for the remaining cases: the duck-typed fallback and
  the per-element `tzinfo` lookup now use a single `PyObject_GetOptionalAttr` on
  pre-interned attribute names (added to `npy_interned_str`). The `tzinfo`
  lookup runs for every datetime

Benchmark

| cast | before | after | speedup |
| --- | --- | --- | --- |
| `datetime.datetime` → `M8[us]` | ~1130 ns | ~52 ns | ~21x |
| `datetime.date` → `M8[D]` | ~490 ns | ~38 ns | ~13x |

<details>
<summary>Benchmark script and notes</summary>

```python
import time, datetime, numpy as np

def best(fn, R=20000, trials=9):
    fn()
    ts = []
    for _ in range(trials):
        s = time.perf_counter()
        for _ in range(R):
            fn()
        ts.append(time.perf_counter() - s)
    return min(ts) / R

N = 100
dts = [datetime.datetime(2000, 1, 1) + datetime.timedelta(seconds=i * 1234567)
       for i in range(N)]
a = np.asarray(dts)
dates = [datetime.date(2000, 1, 1) + datetime.timedelta(days=i) for i in range(N)]
ad = np.asarray(dates)
base = np.arange(1000.0)  # unrelated op, sanity-check for binary-layout noise

print("dt   ->M8[us]: %6.1f ns/elem" % (best(lambda: a.astype('M8[us]')) / N * 1e9))
print("date ->M8[D] : %6.1f ns/elem" % (best(lambda: ad.astype('M8[D]')) / N * 1e9))
print("baseline sum : %6.1f ns"      % (best(lambda: base.sum()) * 1e9))
```

</details>

#### AI Disclosure

Claude code was used for profiling and implementing the selection of performance improvements selected.

### DIFF
--- a/numpy/_core/src/multiarray/datetime.c
+++ b/numpy/_core/src/multiarray/datetime.c
@@ -2233,7 +2233,8 @@ validate_and_return:
                 }
 
             /* The utcoffset function should return a timedelta */
-            PyObject *offset = PyObject_CallMethod(tzinfo, "utcoffset", "O", obj);
+            PyObject *offset = PyObject_CallMethodOneArg(
+                    tzinfo, npy_interned_str.utcoffset, obj);
             Py_DECREF(tzinfo);
             if (offset == NULL) {
                 return -1;
@@ -2243,8 +2244,8 @@ validate_and_return:
              * The timedelta should have a function "total_seconds"
              * which contains the value we want.
              */
-            PyObject *total_seconds = PyObject_CallMethod(
-                    offset, "total_seconds", "");
+            PyObject *total_seconds = PyObject_CallMethodNoArgs(
+                    offset, npy_interned_str.total_seconds);
             Py_DECREF(offset);
             if (total_seconds == NULL) {
                 return -1;
--- a/numpy/_core/src/multiarray/datetime.c
+++ b/numpy/_core/src/multiarray/datetime.c
@@ -2079,10 +2079,10 @@ pydatetime_attr_to_long(PyObject *obj, PyObject *name, long *value)
  * Tests for and converts a Python datetime.datetime or datetime.date
  * object into a NumPy npy_datetimestruct.
  *
- * While the C API has PyDate_* and PyDateTime_* functions, the following
- * implementation just asks for attributes, and thus supports
- * datetime duck typing. The tzinfo time zone conversion would require
- * this style of access anyway.
+ * Exact datetime.date / datetime.datetime objects are read directly from
+ * the C struct via the datetime C-API.  Other objects (including
+ * date/datetime subclasses) are queried for the attributes instead, and
+ * thus support datetime duck typing.
  *
  * 'out_bestunit' gives a suggested unit based on whether the object
  *      was a datetime.date or datetime.datetime object.
@@ -2098,7 +2098,6 @@ NpyDatetime_ConvertPyDateTimeToDatetimeStruct(
         PyObject *obj, npy_datetimestruct *out, NPY_DATETIMEUNIT *out_bestunit,
         int apply_tzinfo)
 {
-    PyObject *tmp;
     int isleap;
     int has_time;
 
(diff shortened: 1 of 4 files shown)


numpy/_core/src/multiarray/datetime.c
```diff
@@ -2264,15 +2291,15 @@ NpyDatetime_ConvertPyDateTimeToDatetimeStruct(
 
 invalid_date:
     PyErr_Format(PyExc_ValueError,
-            "Invalid date (%" NPY_INT64_FMT ",%" NPY_INT32_FMT ",%" NPY_INT32_FMT ") when converting to NumPy datetime",
```
eendebakpt, line 2278: This was already incorrect on main and apparently an unused path. We could consider dropping support for the duck-type datetimes.

numpy/_core/src/multiarray/datetime.c
```diff
@@ -2077,55 +2100,108 @@ NpyDatetime_ConvertPyDateTimeToDatetimeStruct(
...
-    /* Need at least year/month/day attributes */
-    if (!PyObject_HasAttrString(obj, "year") ||
-            !PyObject_HasAttrString(obj, "month") ||
-            !PyObject_HasAttrString(obj, "day")) {
+    /*
+     * Fast path for genuine datetime.date / datetime.datetime objects
+     * (including subclasses such as pandas Timestamp): read the fields
+     * straight from the C struct via the datetime C-API macros.
+     */
+    if (PyDate_Check(obj)) {
```
KRRT7: This fast path also applies to `datetime.date` / `datetime.datetime` subclasses. That changes existing behavior for subclasses that override `year`, `month`, `day`, `hour`, etc.: previously conversion went through Python attribute access, while this now reads the base C struct directly. If subclass attribute semantics should remain supported, this should use exact checks for the fast path and leave subclasses on the attribute path. Otherwise the behavior change needs to be explicit and covered.

KRRT7 (commented):
There are two compatibility regressions in the current implementation that should be addressed before merging.

numpy/_core/src/multiarray/datetime.c
```diff
@@ -2077,55 +2100,108 @@ NpyDatetime_ConvertPyDateTimeToDatetimeStruct(
...
         return -1;
     }
-    out->month = PyLong_AsLong(tmp);
-    if (error_converting(out->month)) {
-        Py_DECREF(tmp);
+    if (found == 0) {
+        has_time = 0;
+        goto validate_and_return;
+    }
+    out->hour = val;
```
KRRT7: The fallback now writes partial time fields before it knows whether all time attributes exist. If `hour` exists but `minute`, `second`, or `microsecond` is missing, the object still resolves as a date for unit inference, but explicit-unit conversions can preserve the partial time. The previous code checked all four time attributes before reading any of them, so these objects behaved as dates with zeroed time fields. The fallback should either pre-check the full time-attribute set or clear any partially written time fields when `has_time` becomes false.

numpy/_core/src/multiarray/datetime.c
eendebakpt: I think this is fine, but to be on the safe side I restricted to exact date and datetime objects. (the pandas TimeSeries wil fall through to the normal path)

numpy/_core/src/multiarray/datetime.c
```diff
@@ -2204,20 +2214,18 @@ NpyDatetime_ConvertPyDateTimeToDatetimeStruct(
     }
 
     /* Apply the time zone offset if it exists */
-    if (apply_tzinfo && PyObject_HasAttrString(obj, "tzinfo")) {
-        tmp = PyObject_GetAttrString(obj, "tzinfo");
-        if (tmp == NULL) {
+    if (apply_tzinfo) {
+        found = PyObject_GetOptionalAttr(obj, npy_interned_str.tzinfo, &tmp);
```
kumaraditya303: You can access the tzinfo as well using C API for the exact type.

eendebakpt: Indeed. Here I also added a fallback for subclasses (even though one could argue subclasses changing this behavior are on their own).

numpy/_core/src/multiarray/datetime.c
```diff
@@ -2204,55 +2213,60 @@ NpyDatetime_ConvertPyDateTimeToDatetimeStruct(
...
             if (PyErr_WarnEx(PyExc_UserWarning,
                 "no explicit representation of timezones available for np.datetime64",
                 1) < 0) {
+                    Py_DECREF(tzinfo);
                     return -1;
                 }
 
             /* The utcoffset function should return a timedelta */
-            offset = PyObject_CallMethod(tmp, "utcoffset", "O", obj);
+            PyObject *offset = PyObject_CallMethod(tzinfo, "utcoffset", "O", obj);
```
kumaraditya303: You can use cached interned string for `utcoffset` and CallMethodOneArg

numpy/_core/src/multiarray/datetime.c
```diff
@@ -2204,55 +2213,60 @@ NpyDatetime_ConvertPyDateTimeToDatetimeStruct(
...
                 return -1;
             }
-            Py_DECREF(tmp);
 
             /*
              * The timedelta should have a function "total_seconds"
              * which contains the value we want.
              */
-            tmp = PyObject_CallMethod(offset, "total_seconds", "");
+            PyObject *total_seconds = PyObject_CallMethod(
```
kumaraditya303: use callmethodnoarg

kumaraditya303 (approved):

kumaraditya303:
Thanks.

PR #31574, opened 2026-06-06
ENH: Support the `dot` family for user-dtypes

SwayamInSync:
### PR summary

This PR adds the `dot` and `correlate` fallback methods for the user-dtypes. It uses the `multiply` and `matmul` operations to delegate the operations as per the input shape.

closes #30793 

#### AI Disclosure
None


numpy/_core/numeric.py
```diff
@@ -1219,6 +1219,96 @@ def tensordot(a, b, axes=2):
...
+        )
+
+
+def _dot_contract(a, b, axis_a, axis_b):
+    notin_a = [k for k in range(a.ndim) if k != axis_a]
+    notin_b = [k for k in range(b.ndim) if k != axis_b]
+    n = a.shape[axis_a]
+    m = math.prod(a.shape[ax] for ax in notin_a)
+    p = math.prod(b.shape[ax] for ax in notin_b)
+    at = ascontiguousarray(a.transpose(notin_a + [axis_a]).reshape(m, n))
```
SwayamInSync: `np.dot` must make `out=` as C-style array, so I think doing `ascontiguousarray` might be OK?
Ideally the matmul/BLAS should handle any layout (for perfomance cases we can revert back to let dtype's BLAS handle this)

> If we revert then, this will require new release from numpy-quaddtype (PR https://github.com/numpy/numpy-quaddtype/pull/108 is already landed)

SwayamInSync:
Test failures seems independent from this PR

numpy/_core/numeric.py
```diff
@@ -1219,6 +1219,96 @@ def tensordot(a, b, axes=2):
...
+        _dot_check_out(out, res_shape, res_dtype)
+    if a.ndim == 0 or b.ndim == 0:
+        # 0-D: scalar multiply
+        res = np.multiply(a, b)
+    elif a.shape[-1] == 0 or 0 in res_shape:
+        # empty contraction or empty result: all-zeros, no matmul
+        res = np.zeros(res_shape, dtype=res_dtype)
+    elif a.ndim == 1 and b.ndim == 1:
+        # 1-D x 1-D: inner product
+        res = np.matmul(ascontiguousarray(a), ascontiguousarray(b))
```
mhvk: Here, you could just do `return  np.matmul(a, b, out=out)` - by default, output is contiguous.

numpy/_core/numeric.py
mhvk: By default, any ufunc produces a C-contiguous array, so I don't think you'd need to do anything.

numpy/_core/numeric.py
```diff
@@ -1219,6 +1219,96 @@ def tensordot(a, b, axes=2):
...
+
+def _dot_contract(a, b, axis_a, axis_b):
+    notin_a = [k for k in range(a.ndim) if k != axis_a]
+    notin_b = [k for k in range(b.ndim) if k != axis_b]
+    n = a.shape[axis_a]
+    m = math.prod(a.shape[ax] for ax in notin_a)
+    p = math.prod(b.shape[ax] for ax in notin_b)
+    at = ascontiguousarray(a.transpose(notin_a + [axis_a]).reshape(m, n))
+    bt = ascontiguousarray(b.transpose([axis_b] + notin_b).reshape(n, p))
+    res = np.matmul(at, bt)
```
mhvk: Are you sure the detailed handling of the axes is necessary? Remember that with `np.matmul`, like with all gufuncs, you have the option of explicitly passing in the axes that are contracted; see https://numpy.org/doc/stable/reference/ufuncs.html#optional-keyword-arguments

numpy/_core/numeric.py
```diff
@@ -1219,6 +1219,96 @@ def tensordot(a, b, axes=2):
...
+    return np.matmul(a, b)
+
+
+def _correlate_fallback(a, v, mode):
+    """Forward correlation via ``matmul`` for dtypes without a legacy dotfunc.
+
+    ``_pyarray_correlate`` already handles empty checks, complex conjugation
+    and the ``len(a) < len(v)`` swap/reverse, so ``a`` is the longer operand.
+    """
+    a = ascontiguousarray(np.asarray(a).ravel())
```
mhvk: Did you check all these `ascontiguousarray` calls are in fact necessary? I rather doubt it, since `matmul` produces a contiguous array.

numpy/_core/numeric.py
```diff
@@ -1219,6 +1219,96 @@ def tensordot(a, b, axes=2):
...
+    v = ascontiguousarray(np.asarray(v).ravel())
+    res_dtype = np.result_type(a.dtype, v.dtype)
+    n1, n2 = a.size, v.size
+    pad = zeros(n2 - 1, dtype=res_dtype)
+    apad = concatenate([pad, a, pad])
+    win = apad[arange(n1 + n2 - 1)[:, None] + arange(n2)]
+    full = np.matmul(win, v)
```
mhvk: I think you need `vecdot` here to ensure complex numbers are properly conjugated (do add at test! also for `dot`, where complex conjugation is not done [incorrectly, but the fallback needs to mimic that mistake...])

mhvk (commented):
Some comments in-line. Functionally, the one about complex conjugation for `correlate` is the most important, but the others should help simplify the implementation. In particular, please check carefully whether all those `ascontiguousarray` calls are in fact necessary (by having tests with non-contigous arrays).

SwayamInSync:
> By default, any ufunc produces a C-contiguous array, so I don't think you'd need to do anything.

@mhvk the element-wise ones does not, they preserve the order. So the case for (0-D x n-D) (just a multiply) will not go to C-contig order.

For BLAS ops yes it will always, I kept this earlier because in quaddtype the PR hasn't merged yet and in case any dtype did not conform with this.
But I agree, this should be a bug from dtype side then. For the above discussed case, I will keep and remove as in-general call.

numpy/_core/tests/test_multiarray.py
```diff
@@ -8348,6 +8348,105 @@ def test_3d_tensor(self):
...
+    ])
+    def test_dot_fallback_matches_dot(self, sa, sb):
+        from numpy._core.numeric import _dot_fallback
+        rng = np.random.default_rng(1)
+        a = rng.integers(-4, 5, sa).astype(np.float64)
+        b = rng.integers(-4, 5, sb).astype(np.float64)
+        ref = np.dot(a, b)
+        got = _dot_fallback(a, b)
+        assert_array_equal(got, ref)
+        assert got.shape == np.shape(ref)
```
mhvk: You can get the shape check by setting `strict=True` in `assert_array_equal`.

Also, I'm actually not 100% sure we should care about contiguity, but if we do, we should test it to ensure we don't regress, by adding something like `assert got.strides == ref.strides` (same in other tests).

SwayamInSync:
In quaddtype the `vecdot` and non-C-layout input support is not merged yet (The PRs are there)
So those tests will fail here.

I will try to fast-forward the work there to get the support merge nicely, and maybe should we add quaddtype's test dependecy to direct fetch and build from repo? @seberg

SwayamInSync:
Also I tried, `axes=` only selects the contracted core dims, but matmul still broadcasts the remaining dims whereas `np.dot` takes their outer product (a-batch then b-batch). Concretely: 
`dot((2,3,4),(6,4,5)) => (2,3,6,5)`, but `matmul(..., axes=...)` raises `ValueError` trying to broadcast `(2,3)` against `(6,)` and `dot((2,3),(4,3,5)) => (2,4,5)` while matmul gives `(4,2,5)` (wrong, no error). 
So the transpose/reshape into a single 2-D matmul is needed to express the outer-product contraction

seberg:
> I will try to fast-forward the work there to get the support merge nicely, and maybe should we add quaddtype's test dependecy to direct fetch and build from repo?

I already have one or two tests with quaddtype, but I am not sure it is installed yet.  The problem is that that even importing quaddtype can break some other tests.  So you need a bit of a tedious skipping for the quaddtype version (without actually importing it).

But, I think that's fine... we can factor that out into a little private helper.  (I think you should be able to find that hack if you search, quickly.)

SwayamInSync:
> But, I think that's fine... we can factor that out into a little private helper. (I think you should be able to find that hack if you search, quickly.)

Done, added in this PR as it wasn't that disturbing the context

SwayamInSync:
This is ready for 2nd round of review.
Will be happy to discuss design choices or if I missed some case.

numpy/_core/numeric.py
seberg: Unfortunately, for `ndim > 2` dot has a funny behavior that the result shape is `a.shape + b.shape` minus the reduction axes.
I.e. it doesn't actually do a proper broadcasting/stacked operation, it is the same as `tensordot(..., axes=[-1, -2])`.

However, `tensordot` actually does this precise dance already.  So we *could* consider being a bit annoying and just rejecting it for this path (we _should_ deprecate it on the other path I guess) with a note of "please use `tensordot` instead".

There is another good argument for this, the code has:
```
#if defined(HAVE_CBLAS)
    if (PyArray_NDIM(ap1) <= 2 && PyArray_NDIM(ap2) <= 2 &&
```
so we don't even use BLAS for this case, which is ridiculous and unless arrays are small users are better of to use the clearer `tensordot()` anyway.

(If we do that this might be simple enough to just move it to C.)

numpy/_core/tests/_quaddtype.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def importorskip_quaddtype():
+    from importlib.metadata import PackageNotFoundError, version
+    try:
+        installed = version("numpy_quaddtype")
+    except PackageNotFoundError:
+        pytest.skip("numpy_quaddtype is not installed")
+    if _pep440.Version(installed) < _pep440.Version(MIN_VERSION):
+        pytest.skip(f"numpy_quaddtype >= {MIN_VERSION} is required")
```
seberg, line 15: Did you mean to put `MIN_VERSION` to 1.1 since I think we are at 1.0 (and we need to reject it).
I guess we can make it customizable if needed.  The current min-version is vital because even importing is wrong, but after 1.1 is out that isn't the case anymore.

numpy/_core/numeric.py
```diff
@@ -1219,6 +1219,106 @@ def tensordot(a, b, axes=2):
...
+        raise ValueError("first array argument cannot be empty")
+    if v.size == 0:
+        raise ValueError("second array argument cannot be empty")
+    inverted = a.size < v.size
+    if inverted:
+        a, v = v, a
+    n1, n2 = a.size, v.size
+    pad = zeros(n2 - 1, dtype=np.result_type(a.dtype, v.dtype))
+    apad = concatenate([pad, a, pad])
+    win = apad[arange(n1 + n2 - 1)[:, None] + arange(n2)]
```
seberg: This can be a `sliding_window_view` I think.  I think `matmul` will be OK with core dimensions that have self overlap (if not, we could pass `axes` to make it a non-core dimension).  Although that is of course not BLAS compatible (for our code, I think it might even go back into `einsum`).

Otherwise, you would bloat memory usage like crazy here.  (I suspect we can do this, but I am slightly unsure and i tmay make sense to split the PR into two, since I think `dot` is a bit clearer.)

numpy/_core/tests/_quaddtype.py
SwayamInSync: Yeah that also works, just weird habit of having even jumps for sub-versions :)

numpy/_core/numeric.py
SwayamInSync: I am too concerned about the memory bloat here + `sliding_window_view` user-dtypes don't support.

Right to split this, I will do another PR for correlate and with the current and we can make further changes there

numpy/_core/numeric.py
SwayamInSync: fallback's `_dot_contract` is the tensordot-style `reshape=>matmul`, faster than legacy `N-D dot`, so rejecting it isn't a perf win, it's purely about not carrying the deprecation-worthy N-D dot semantics. I'm fine either way.

numpy/_core/numeric.py
seberg: > `sliding_window_view` user-dtypes don't support.

Ah, that is true and annoying :/.  Could possibly not go via `__array_interface__` there to make it work, but I am not sure that is better than trying a fix for `__array_interface__` (heck even adding a `dtype` field).

numpy/_core/numeric.py
seberg: Not sure I am following precisely, but I am leaning to just reject it with a "use tensordot" message.  Let's ping @mhvk, though.
(That would mean we should aim for a deprecation as a tendency, I guess.)

ngoldbaum:
We talked a little bit about this at the triage meeting and I'm in favor of not implementing the weird behavior of `np.dot`.

SwayamInSync:
I see, so I think we are set to deprecate the last case with warning "use tensordot"? Then we also need to do this for legacy path too right.

ngoldbaum:
> Then we also need to do this for legacy path too right.

Probably but it doesn't need to happen here.

SwayamInSync:
Apologies for the delay, the weird behaviour of np.dot with `a.ndim > 2 or b.ndim > 2` is gone and errors to use tensordot.
We might note to apply this for the builtin dtypes too in a follow-up.

SwayamInSync:
`np.dot` scalar-case is BLAS dependent, BLAS builds return C-contig while non-BLAS builds preserve the layout :)

SwayamInSync:
Failures seems independent of the work.
@seberg @mhvk @ngoldbaum this is ready to another review round

ngoldbaum:
I spotted a couple more issues that are worth addressing.

Types that don't have `dotfunc` currently raise `ValueError` on `np.dot` and now it raises `UfuncTypeError`.

Also there's a little weirdness around how mixed-dimensionality arrays are rejected:

```python
import numpy as np
from numpy._core.numeric import _dot_fallback

rng = np.random.default_rng(0)

def arr(*shape):
    return rng.integers(-4, 5, shape).astype(np.float64)

def dot_equals_matmul(a, b):
    ref = np.dot(a, b)
    try:
        m = np.matmul(a, b)
    except ValueError:
        return False
    return m.shape == ref.shape and np.array_equal(m, ref)

def fallback(a, b):
    try:
        _dot_fallback(a, b)
        return "ALLOWS"
    except ValueError:
        return "REJECTS"

cases = [
    (arr(2, 3, 4), arr(4)),
    (arr(2, 3, 4), arr(4, 5)),
    (arr(4), arr(6, 4, 5)),
    (arr(3, 4), arr(6, 4, 5)),     # outer product over batch axes: truly weird
]
for a, b in cases:
    same = "dot==matmul" if dot_equals_matmul(a, b) else "dot!=matmul"
    print(f"dot {str(a.shape):>10} x {str(b.shape):<10} {same}  ->  "
          f"fallback {fallback(a, b)}")
```

On your PR currently, this prints:

```
dot  (2, 3, 4) x (4,)       dot==matmul  ->  fallback ALLOWS
dot  (2, 3, 4) x (4, 5)     dot==matmul  ->  fallback REJECTS
dot       (4,) x (6, 4, 5)  dot==matmul  ->  fallback REJECTS
dot     (3, 4) x (6, 4, 5)  dot!=matmul  ->  fallback REJECTS
```

The fallback rejects cases where the result of `np.dot` is the same as `np.matmul`.  The minimal condition for the weird outer-product semantics is `a.ndim >= 2` and `b.ndim >= 3`, not 2, I think?

seberg:
>  The minimal condition for the weird outer-product semantics is a.ndim >= 2 and b.ndim >= 3, not 2, I think?

Sounds right, I am not sure if I care too much if we reject a bit more than needed.  But this rule seems simple enough.

I would ask you to try and clean-up the leftovers from the old approach though.  Output shape checking should be unnecessary, the ufunc machinery should take care of it.  And at that point going through Python should be more hassle than gain, I expect.

SwayamInSync:
Both are fixed

> Types that don't have dotfund currently raise ValueError on np.dot and now it raises UfuncTypeError.

Its because of using `matmul` I now catch and re-raise the similar to dot one

numpy/_core/tests/test_multiarray.py
```diff
@@ -8348,6 +8348,145 @@ def test_3d_tensor(self):
...
+    # when available, end-to-end through a real new-style DType (quaddtype).
+
+    @pytest.mark.parametrize("sa,sb", [
+        ((), ()), ((), (3, 4)), ((4, 3), ()), ((5,), (5,)), ((3,), (3, 4)),
+        ((4, 3), (3,)), ((4, 3), (3, 5)), ((2, 3, 4), (4,)),
+        ((2, 3, 4), (4, 5)), ((4,), (2, 4, 5)),
+        ((0, 3), (3, 4)), ((4, 0), (0, 5)),
+    ])
+    def test_dot_fallback_matches_dot(self, sa, sb):
+        from numpy._core.numeric import _dot_fallback
```
mhvk: Why not import these on top (here and below)?

numpy/_core/tests/test_multiarray.py
```diff
@@ -8348,6 +8348,145 @@ def test_3d_tensor(self):
...
+        assert _dot_fallback(a, b) == np.dot(a, b)
+
+    def test_dot_fallback_out(self):
+        from numpy._core.numeric import _dot_fallback
+        a = np.arange(12, dtype=np.float64).reshape(4, 3)
+        b = np.arange(15, dtype=np.float64).reshape(3, 5)
+        out = np.empty((4, 5), dtype=np.float64)
+        ret = _dot_fallback(a, b, out=out)
+        assert ret is out
+        assert_array_equal(out, np.dot(a, b))
```
mhvk: Elsewhere, you have `strict=True`, might as well do it here too.

numpy/_core/tests/test_multiarray.py
```diff
@@ -8348,6 +8348,145 @@ def test_3d_tensor(self):
...
+    def test_quaddtype_dot_family(self):
+        from numpy._core.tests._quaddtype import importorskip_quaddtype
+        numpy_quaddtype = importorskip_quaddtype()
+        qd = numpy_quaddtype.QuadPrecDType()
+
+        def q(arr):
+            return np.array(arr, dtype=qd)
+
+        def f(arr):
+            arr = np.asarray(arr)
```
mhvk: I'm confused, why not just `np.asarray(arr).astype(float)`?

numpy/_core/tests/test_multiarray.py
```diff
@@ -8348,6 +8348,145 @@ def test_3d_tensor(self):
...
+        a = rng.integers(-4, 5, (2, 3)).astype(np.float64)
+        b = rng.integers(-4, 5, (2, 3)).astype(np.float64)
+        assert _vdot_fallback(a, b) == np.vdot(a, b)
+        ac = (a + 1j * b).astype(np.complex128)
+        bc = (b - 1j * a).astype(np.complex128)
+        assert _vdot_fallback(ac, bc) == np.vdot(ac, bc)
+
+    def test_quaddtype_dot_family(self):
+        from numpy._core.tests._quaddtype import importorskip_quaddtype
+        numpy_quaddtype = importorskip_quaddtype()
```
mhvk, line 8440: there is no complex quaddtype yet, right?

numpy/_core/numeric.py
```diff
@@ -1219,6 +1220,31 @@ def tensordot(a, b, axes=2):
     return res.reshape(olda + oldb)
 
 
+def _dot_fallback(a, b, out=None):
+    """``dot`` via ``matmul``/``multiply`` for dtypes without a legacy dotfunc."""
+    a = np.asarray(a)
```
mhvk: Do use `np.asanyarray` here; no need to block subclasses. (One could even consider not casting at all, since `multiply` or `matmul` will do that anyway, and that might allow supporting, say, dask arrays with custom dtypes too, but that would mean defining `a_ndim = np.ndim(a)` and `b_ndim = np.ndim(b)` and using those in the dimension checks below.)

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -996,6 +996,17 @@ PyArray_MatrixProduct2(PyObject *op1, PyObject *op2, PyArrayObject* out)
     }
     Py_SETREF(typec, NPY_DT_CALL_ensure_canonical(typec));
 
+    if (typec != NULL && PyDataType_GetArrFuncs(typec)->dotfunc == NULL) {
```
mhvk: I don't think `typec` can possibly be NULL here (it is INCREF'd below this branch after all).

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -2626,6 +2637,16 @@ array_vdot(PyObject *NPY_UNUSED(dummy), PyObject *const *args, Py_ssize_t len_ar
     }
    Py_SETREF(type, NPY_DT_CALL_ensure_canonical(type));
 
+    if (type != NULL && PyDataType_GetArrFuncs(type)->dotfunc == NULL) {
```
mhvk: same here, `type` cannot be NULL here.

numpy/_core/numeric.py
```diff
@@ -1219,6 +1220,31 @@ def tensordot(a, b, axes=2):
...
+        raise ValueError(
+            "'dot' does not support the stacked outer-product semantics of "
+            "'a.ndim >= 2 and b.ndim >= 3' for user-defined dtypes; "
+            "use 'numpy.tensordot' instead."
+        )
+    op = np.multiply if a.ndim == 0 or b.ndim == 0 else np.matmul
+    try:
+        return op(a, b, out=out)
+    except _UFuncNoLoopError:
+        raise ValueError("dot not available for this type") from None
```
mhvk, line 1243: Can we add a note like "# Keep the error the same as it was before we had support for dot for user types."

numpy/_core/numeric.py
```diff
@@ -1219,6 +1220,31 @@ def tensordot(a, b, axes=2):
...
+
+
+def _vdot_fallback(a, b):
+    try:
+        return np.vecdot(np.asarray(a).ravel(), np.asarray(b).ravel())
```
mhvk: Here, change to `np.ravel(a), np.ravel(b)` to preserve subclasses (and possibly things like dask arrays).

numpy/_core/numeric.py
```diff
@@ -1219,6 +1220,31 @@ def tensordot(a, b, axes=2):
...
+    except _UFuncNoLoopError:
+        raise ValueError("dot not available for this type") from None
```
mhvk: SHould this be `"vdot ..."`

Also, again, a comment might be helpful to future maintainers who wonder why the error is not simply raised.

mhvk (commented):
Looks nice! Mostly nitpicks, except the request to try always to use `np.asanyarray` instead of `np.asarray` to make life easier for subclasses.

numpy/_core/numeric.py
SwayamInSync: done, it original error says "function ..."

SwayamInSync:
@seberg @mhvk @ngoldbaum 
Let me know if any point need to be addressed here, I think after this we can proceed to fix correlate in a follow-up

seberg:
IMO, at this point, going through Python adds as much complexity as it resolves, but if others don't mind...

SwayamInSync:
> IMO, at this point, going through Python adds as much complexity as it resolves, but if others don't mind...

Make sense, I will wait for the final thoughts then.

numpy/_core/tests/_quaddtype.py
mhvk: I think this comment still needs to be addressed...

mhvk (commented):
I had another look and think it is fine, though I somewhat agree with @seberg that the python overrides are sufficiently short one could consider just doing it in C. On the other hand, it is easier for people to understand where things go wrong if they can just look at the python code...

Anyway, I'm fine either way, just don't forget to update MIN_VERSION!

numpy/_core/tests/_quaddtype.py
seberg: I guess if the next release is 1.2 then it doesn't matter.  Although, I don't really see why numpy-quaddtype should use a different version scheme (distinguising odd from even versions) than NumPy.

SwayamInSync:
@mhvk @seberg updated

- `MIN_VERSION` is `1.1`
- moved the `importorskip_quaddtype` imports to the top of the test files.

On the C vs Python thing: I tried porting both fallbacks to C to see how it actually looks. The catch is the tests. Nothing in numpy's own test suite can reach the fallback through `np.dot`:

- builtin dtypes that have a `dotfunc` never hit it
- the ones without (`datetime64`, `S`, `U`, `V`) have no `matmul` loop either, so only the error path
- `rational` has a `dotfunc`
- `_ScaledFloatTestDType` has `multiply` but no `matmul`, and adding one through `sfloat_add_wrapping_loop` segfaults, that helper doesn't handle gufuncs

So either way I need a private helper for the tests to call, in Python it is `numpy._core.numeric._dot_fallback`, in C it would be two entries in the multiarray method table. That is why I am not fully convinced the port buys much, plus `np.ndim`/`np.ravel` are also needs to be put as a C function, and the test hook stays in both cases. 
After #31992 it is also not a clean win on the module state side, Python version needs the two `NPY_RUNTIME_IMPORTS_FIELDS` entries and the C one would need `F(vecdot)` in `NPY_N_OPS_FIELDS`, so one field list either way. I have the C version ready and passing, will be happy to push in case.

SwayamInSync:
merge conflict and found a regression on stringdtype (since we prevented the casting to common type).
Will fix and update here

SwayamInSync:
Pushed: main merged (conflict cleared) and the regression fixed.

The dotfunc check sat above dot's 0-D block, so DTypes without a dotfunc skipped the cast to the common dtype and reached multiply's string×int loop
```python
np.dot(np.array('ab'), 3, out=np.empty((), 'U6')) 
#  returned `'ababab' instead of raising. 
```

Moved it below the 0-D block (e45f5884db), so 0-D stays on dot's own path.

That left the multiply branch in `_dot_fallback` unreachable, so it's dropped (a5145db7af), the fallback is matmul-only now.
Behaviour is identical to 2.5.3 for every dtype without a dotfunc, so this only adds behaviour where dot used to raise.

SwayamInSync:
An old bug on main is exposed by the new tests added. In `PyArray_MatrixProduct2`, the 0-D branch returns without releasing `typec`:
For float64 the descriptor is a static builtin, so the leak never showed. `test_dot_scalar_operands_bypass_fallback` calls `np.dot` with 0-D string and `m8[s]` operands, and those descriptors are newly allocated each call, so they leak for real. 

I reproduced it locally: `np.dot(td, td)` on a `0-D m8[s]` array leaked one descriptor reference per call (1000 leaked after 1000 calls).

Fixed it in [631a826](https://github.com/numpy/numpy/pull/31574/commits/631a826596b8675f7ce48715a9722a5e65477529)

Other CI failure is unrelated

numpy/_core/tests/_quaddtype.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+
+from numpy._utils import _pep440
+
+MIN_VERSION = "1.1"
```
ngoldbaum, line 5: I guess we have a bit of a chicken-egg problem in that this release doesn't exist yet. Is there any chance you can make the test work (with xfails maybe?) against 1.0 so can be pretty confident it works? Right now this whole test is dead code.

numpy/_core/numeric.py
```diff
@@ -1231,6 +1232,34 @@ def tensordot(a, b, axes=2):
...
+    # keeps subclasses (and array-likes such as dask arrays) intact.  ``dot``
+    # handles 0-D itself, so only ``matmul`` is needed.
+    if np.ndim(a) >= 2 and np.ndim(b) >= 3:
+        raise ValueError(
+            "'dot' does not support the stacked outer-product semantics of "
+            "'a.ndim >= 2 and b.ndim >= 3' for user-defined dtypes; "
+            "use 'numpy.tensordot' instead."
+        )
+    try:
+        return np.matmul(a, b, out=out)
```
ngoldbaum, line 1247: should this specify `order="C"`? Otherwise stuff like this happens:

```python
import numpy as np
from numpy._core.numeric import _dot_fallback

a = np.ones((2, 3, 4, 5)).swapaxes(0, 1)
b = np.ones((5, 6))

expected = np.dot(a, b)
actual = _dot_fallback(a, b)

print(expected.flags.c_contiguous, actual.flags.c_contiguous)
# True False
```

doc/release/upcoming_changes/31574.improvement.rst
```diff
@@ -0,0 +1,15 @@
+``dot``, ``vdot`` and ``inner`` now work for user-defined DTypes
+----------------------------------------------------------------
+`numpy.dot`, `numpy.vdot` and `numpy.inner` (and thus also
+`numpy.linalg.multi_dot`) previously required the legacy ``dotfunc`` "array
+function" slot, which new-style user-defined DTypes cannot provide.  For such
+DTypes NumPy now falls back to the `numpy.matmul` ufunc, and to
+`numpy.vecdot` for `numpy.vdot`, so the operations work as long as the DType
+implements those.
+
+The stacked-array behaviour of `numpy.dot`, which for ``a.ndim >= 2`` and
```
ngoldbaum, line 10: It's not just `np.dot` though. Lots of other code paths go through `MatrixProduct2`. Maybe you could say "behaviour of `numpy.dot`, `numpy.inner`, `numpy.tensordot`, and other code paths like `numpy.cov` that rely on dot products"

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -1042,8 +1043,32 @@ PyArray_MatrixProduct2(PyObject *op1, PyObject *op2, PyArrayObject* out)
...
+        Py_DECREF(ap2);
+        Py_DECREF(typec);
+        if (npy_cache_import_runtime(
+                "numpy._core.numeric", "_dot_fallback",
+                &_npy_module_state->runtime_imports._dot_fallback) == -1) {
+            return NULL;
+        }
+        return PyObject_CallFunctionObjArgs(
+                _npy_module_state->runtime_imports._dot_fallback, op1, op2,
+                out != NULL ? (PyObject *)out : Py_None, NULL);
```
ngoldbaum, line 1070: Is there any way we can avoid so many repeated calls to `__array__`?

```python
import numpy as np
from numpy_quaddtype import QuadPrecDType

class ArrayLike:
    ndim = 2

    def __init__(self):
        self.data = np.ones((2, 2), dtype=QuadPrecDType())
        self.calls = 0

    def __array__(self, dtype=None, copy=None):
        self.calls += 1
        return np.array(self.data, dtype=dtype, copy=copy)

for fn in (np.matmul, np.dot):
    a, b = ArrayLike(), ArrayLike()
    fn(a, b)
    print(fn.__name__, a.calls, b.calls)
```

This prints on your PR

```
matmul 1 1
dot 3 3
```

Ideally we could avoid calling `__array__` repeatedly. If that ends up being annoying, is there any way to cap it at 2?

numpy/_core/numeric.py
ngoldbaum: Matmul and the ufunc machinery have different `out` compatibility rules. I'm not sure whether this is worth worrying about, but there's also this incosistency:

```python
import numpy as np
from numpy._core.numeric import _dot_fallback

a, b = np.ones((2, 3)), np.ones((3, 4))
outputs = {
    "extra batch": np.empty((5, 2, 4)),
    "Fortran": np.empty((2, 4), order="F"),
    "float32": np.empty((2, 4), dtype=np.float32),
    "strided": np.empty((2, 8))[:, ::2],
}

for label, out in outputs.items():
    for fn in (np.dot, _dot_fallback):
        try:
            result = fn(a, b, out=out)
            print(label, fn.__name__, "accepted", result.shape, result.dtype)
        except ValueError:
            print(label, fn.__name__, "ValueError")
```

On this PR I get:

```
extra batch dot ValueError
extra batch _dot_fallback accepted (5, 2, 4) float64
Fortran dot ValueError
Fortran _dot_fallback accepted (2, 4) float64
float32 dot ValueError
float32 _dot_fallback accepted (2, 4) float32
strided dot ValueError
strided _dot_fallback accepted (2, 4) float64
```

e.g. the "real" `dot` implementation rejects all these `out` shapes, dtypes, C/F order, and strides

ngoldbaum (commented):
I asked an AI model to look this over and it spotted some corner cases and one issue with the tests, see below. I also have a suggestion for the release note, although it's already getting pretty long so if you can come up with a one-sentence version that's even better.

numpy/_core/tests/_quaddtype.py
SwayamInSync: Right now v1.0 of numpy-quaddtype has the bug of registering broad promoter/loops and importing it can interfere with other unrelated tests (those issues are fixed upstream and planned to be out in v1.1) so technically we cannot even import it in the current released version.

So this guard is technically saving us from multiple issues (indeed a bad practice) hence I will make sure that next release of quaddtype rightly solves such problems

numpy/_core/numeric.py
SwayamInSync: Fixed, the fallback now passes order="C".
`dot c_contiguous True | fallback c_contiguous True`

and regarding dot's `out=` contract, earlier silently depends on the dtype. The fallback now validates out itself (ndim, C-contiguity, dtype) and reuses dot's own error message, so all four of your cases match:

```
extra batch  dot=ValueError  fallback=ValueError
Fortran      dot=ValueError  fallback=ValueError
float32      dot=ValueError  fallback=ValueError
strided      dot=ValueError  fallback=ValueError
```

I checked this with other possible permutation/combinations and all matches with the usual `dot` behaviour

numpy/_core/src/multiarray/multiarraymodule.c
SwayamInSync: Fixed. Down to cap of 2. The C side now hands `_dot_fallback` the already-converted `ap1/ap2` instead of the original objects, so the third conversion is gone.

```
matmul 1 1
dot    2 2
```

Subclasses still survive, and it removes a promotion inconsistency as a side effect, since the fallback now uses dot's common dtype rather than matmul's own.

doc/release/upcoming_changes/31574.improvement.rst
SwayamInSync: Fixed

numpy/_core/tests/_quaddtype.py
ngoldbaum: OK, then what do you think about deleting these dead tests in this PR and then re-adding them after you do a numpy-quaddtype release and bump the pin here?

numpy/_core/numeric.py
```diff
@@ -1231,6 +1232,40 @@ def tensordot(a, b, axes=2):
...
+    if a_ndim >= 2 and b_ndim >= 3:
+        raise ValueError(
+            "'dot' does not support the stacked outer-product semantics of "
+            "'a.ndim >= 2 and b.ndim >= 3' for user-defined dtypes; "
+            "use 'numpy.tensordot' instead."
+        )
+    if out is not None:
+        expected_ndim = max(a_ndim + b_ndim - 2, 0)
+        if (out.ndim != expected_ndim or not out.flags.c_contiguous
+                or out.dtype != np.result_type(a, b)):
```
ngoldbaum, line 1247: This still isn't quite right: it should also flag an incorrect output shape with the same number of dimensions and an un-aligned but C-contiguous output.

Rather than duplicating the rules in Python, can you instead extract the output validation from `new_array_for_sum` into a private C helper that this function can call?

ngoldbaum:
I think this is getting very close! I added a milestone and the triage review label to hopefully attract one more careful code review pass from a maintainer.

numpy/_core/tests/_quaddtype.py
SwayamInSync: Yeah good to me.

numpy/_core/numeric.py
seberg: Do you have an opinion on this?

I feel like we either give up and pass `out` as allocated and checked by `dot()`. Or we don't worry about any of this and just use the gufunc machinery as is (as it is the "gold standard" and what `dot` could transition to as well).

Here, we actually check that `out.dtype` is identical to the `result_type`, but we don't enforce it in the ufunc when `out=None`, where it isn't _strictly_ guaranteed.  At that point we are almost at `casting="no"` into the ufunc, but also that I would prefer not to.

(I feel like all of this is churn for fairly weird corner cases -- yes, I don't like the divergence much either, but...)

numpy/_core/tests/_quaddtype.py
seberg: I dunno, I honestly prefer putting this in with the dead tests (that work and were tested locally), then with a bunch of unit-tests for the `_dot_fallback()`.

(I don't like such unit-tests if it's avoidable. For one they don't age well if we delete `_dot_fallback()` and then the person who does that work has to now do the double work of also fixing all tests.)

Of course that relies on actually getting that quaddtype release out very soon (and this bug feels like almost enough of a reason to do so?)

numpy/_core/numeric.py
seberg: FWIW, when it comes to the `order="C"` that seems reasonable, because it is also a reasonable idea if we want to rewire things (unfortunately!).

In general, I don't want us to start to guarantee order, but maybe `dot` is special enough (to at least not rush changing it).

seberg (commented):
I think it's very close, but now that we have all of those additional check, I am honestly a bit confused.
I don't care a whole lot if the new path is slightly more relaxed here (heck, I could ignore the `order="C"` part).

About the tests... Serious question: Can we maybe just get that release out?  The fix for this issue alone feels like it may well warrent it? :)  (no idea how much work it is, but also Python 3.15 support, etc. is upcoming so it would be timely.)

PR #31575, opened 2026-06-06
BUG: Pass DType class `np.average` to support `weights` argument for user-dtypes

SwayamInSync:
### PR summary

Same fix as #31447

Discovered at https://github.com/numpy/numpy-quaddtype/issues/106

#### AI Disclosure
None

### DIFF
--- a/numpy/lib/tests/test_function_base.py
+++ b/numpy/lib/tests/test_function_base.py
@@ -538,6 +538,21 @@ class TestAverage:
         a = np.array([Fraction(1, 5), Fraction(3, 5)])
         assert_equal(np.average(a), Fraction(2, 5))
 
+    def test_new_style_user_dtype(self):
+        # average forwarded a DType instance (from result_type) to the
+        # weight sum / multiply ufuncs, which reject instances of non-legacy
+        # user DTypes.  It must pass the DType class instead.
+        numpy_quaddtype = pytest.importorskip("numpy_quaddtype")
+        qd = numpy_quaddtype.QuadPrecDType()
+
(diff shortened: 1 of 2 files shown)


numpy/lib/tests/test_function_base.py
```diff
@@ -538,6 +538,21 @@ def test_average_class_without_dtype(self):
...
+    def test_new_style_user_dtype(self):
+        # average forwarded a DType instance (from result_type) to the
+        # weight sum / multiply ufuncs, which reject instances of non-legacy
+        # user DTypes.  It must pass the DType class instead.
+        numpy_quaddtype = pytest.importorskip("numpy_quaddtype")
+        qd = numpy_quaddtype.QuadPrecDType()
+
+        a = np.array([1, 2, 3, 4], dtype=qd)
+        w = np.array([4, 3, 2, 1], dtype=qd)
+        assert float(np.average(a)) == 2.5
```
mhvk, line 550: I think one should test that the returned array has the right data type, with, e.g.,
```
assert_array_equal(np.average(a), np.array(2.5, dtype=qd), strict=True)
```

Note that it might also be good to make sure the precision is in fact kept, e.g., by writing
```
a = np.array([1, 2, 3, 4], dtype=qd) + np.array(1e-20, dtype=dq)
w = np.array([4, 3, 2, 1], dtype=qd) - np.array(1e-21, dtype=dq)
```
(and obviously adjusting the result accordingly)

seberg:
This needs the quaddtype test helper from gh-31574.  That looks like the reason for the CI failures here (it's very cryptic, but quaddtype had some bugs that caused cryptic crazy issues :)).

Issue #21300, opened 2022-04-05
ENH: Ensure F2PY generates Python-C stable API code


HaoZeke:
### Proposed new feature or change:

The [C API Stability](https://docs.python.org/3/c-api/stable.html#stable) document defines a subset of functions which are (to some extent) compatible across python versions. Thankfully most of the API functions used in `fortranobject.{c,h}` are already in the stable API.

It would be good to ensure this strictly though. A solution would be effectively to ensure that all C extensions compile with `Py_LIMITED_API` set to the same lowest Python version supported by NumPy itself. [1]

In practice, there is a slight performance hit from not using functions outside the stable API, but more importantly, the reference counting is not always the same between these functions (e.g. The sequence protocol defines an "unsafe" `Fast_GET_ITEM` which borrows a reference instead of `GetItem` which returns a new reference). This means that some care may need to be taken to ensure equivalence.

[1] Also to make sure that none of the "unstable" API functions are used

m-newhauser:
Does this still need working on?

HaoZeke:
> Does this still need working on?

Yes but it's tricky since f2py tests are still less than comprehensive and also at the same time very low priority... I'd not recommend picking this up in the near future.

PR #31576, opened 2026-06-06
ENH: Build and testing infrastructure for f2py Stable ABI

da-woods:
### PR summary

This is essential step 0 to support #21300.

First, it allows you to try to build f2py modules in the Limited API (by passing the right arguments to meson). Note that this will almost certainly result in compilation failure. That's a future problem. I haven't documented the command-line option because I don't intend real people to use it right now.

Second, it parameterizes all the f2py tests so that they can be run with the limited API set (for all `Py_LIMITED_API` values between 3.12 and the current version). These extra parameterized tests are immediately skipped since they are of course broken.

My assumption is that the Limited API/Stable ABI can be supported by checking `Py_LIMITED_API` at compile-time and won't require f2py to generate different code (this is how Cython handles it for example). Therefore the sole job is to pass it to meson but none of the code-generator bits need to know. That assumption might turn out to be wrong but dictates how I've started for now.

#### AI Disclosure

No AI tools used


numpy/f2py/tests/test_array_from_pyobj.py
```diff
@@ -25,18 +26,24 @@ def get_testdir():
     testroot = Path(__file__).resolve().parent / "src"
     return testroot / "array_from_pyobj"
 
-def setup_module():
+@pytest.fixture(scope="module", autouse=True)
+def setup_module_fixture(f2py_limited_api_module):
```
da-woods, line 30: For this (and also the `setup_method` fixtures) I had to modify them because the `setup_module`/`setup_method` infrastructure doesn't support parameters.

Instead use a regular fixture and `autouse` to force it to be run. That way it can depend on the parameterized limited_api

numpy/conftest.py
```diff
@@ -66,6 +66,11 @@ def pytest_configure(config):
         "leaks_references: Tests that are known to leak references.")
     config.addinivalue_line("markers",
         "slow: Tests that are very slow.")
+    # f2py_stable_abi_good is expected to go away if/when limited API is actually working
+    config.addinivalue_line("markers",
+        "f2py_stable_abi_good: Tests known to work in the Stable ABI")
+    config.addinivalue_line("markers",
+        "f2py_stable_abi_bad: Tests unlikely to ever work in the Stable ABI")
```
da-woods: Having a good and bad marker is a little ugly. I tried to add an extra option to `pytest_addoption` to let you force the tests on but `pytest_addoption` seems to be skipped when called from spin.

So my idea is that that `f2py_stable_abi_good` is a temporary tool for gradually enabling tests and testing them locally.

`f2py_stable_abi_bad` is a permanent tool for labelling stuff that's never expected to work.

But I'm prepared to believe there's a better way.

PR #31577, opened 2026-06-06
BUG: Limited API fix for INCREF/DECREF in Numpy headers

da-woods:
### PR summary

In the Limited API, `Py_INCREF` (and similar) don't contain a cast and thus the object has to be a `PyObject*`.

I'm not completely sure if npy_2_compat is intended to be Limited API compatible but it's easy enough to adjust anyway.

#### AI Disclosure

No AI tools used


charris:
Thanks @da-woods .

PR #31578, opened 2026-06-06
ENH: Make f2py use multiphase init

da-woods:
### PR summary

Adapt f2py to use multiphase init (and on Python 3.15+, to use PEP 793 and 820 init). This is mainly useful for future Limited API support (although mutliphase init is generally "preferred" and so arguably better anyway).

The main complication was the usercode that gets injected into the module init function. This wasn't tested (so I added a test) but user code with error handling requires `return NULL` to be valid, so I moved it to a separate function that keeps that interface.

One potential issue to flag is if people are doing embedding type things and manually calling the `PyInit_` function. This isn't recommended and isn't an insurmountable problem, but will break that case.

#### AI Disclosure

No AI tools used


numpy/f2py/rules.py
```diff
@@ -239,26 +239,18 @@
...
-    "#modulename#",
-    NULL,
-    -1,
-    f2py_module_methods,
-    NULL,
-    NULL,
-    NULL,
-    NULL
-};
+static PyObject *f2py_module_interface_usercode(PyObject *m, PyObject *d) {
```
da-woods, line 242: I've assumed here that `m` and `d` are the function-level variables that people might reasonably use. And that there's no reason for them to use `tmp`, `i` and `s` from the init function.

seberg: Yeah, sounds fine to me, wrapping it into it's own function is saner anyway.  However, I would lean towards adding/keeping the:
```
#modulename#_module = m;
```
alias here, which is a nice name.  (`#modulename#_error` seems rather unnecessary to me, though, and I doubt there is a point to `f2py_routine_defs` or so.)

numpy/f2py/rules.py
```diff
@@ -282,27 +274,74 @@
...
     if (PyType_Ready(&PyFortran_Type) < 0) {
-        return NULL;
+        return -1;
     }
 
-    return m;
+    return 0;
 }
+
+#ifndef Py_TARGET_ABI3T
```
seberg, line 295: I guess we could even just not add it for `PY_VERSION_HEX < 0x030F0000`?  But maybe best to keep things side-by-side (I am not sure, and it isn't like it matters).

seberg (commented):
Thanks, LGTM, modulo that one nit that maybe we should include the `..._module` alias.

Do you care for this being backported to 2.5, or are f2py extensions a bit "out there" and we don't mind to wait for 2.6 for supporting `Py_TARGET_ABI3T` for them?

(It seems fine to backport generally, but f2py is a bit annoying and the rc is already out for 2.5.)

da-woods:
> Do you care for this being backported to 2.5, [...]

There's a bunch of other stuff to be done to make it work in abi3t. This is essentially just one independent step. If you've got a release candidate already then I'd save this for 2.6

numpy/f2py/rules.py
da-woods: > However, I would lean towards adding/keeping the:
>
> ```
> #modulename#_module = m;
> ```

I think that's a global so should be available generally. I'll check properly later though.

seberg: Ohh, right. Nvm, I missed that it is a global (but it also means that it'll keep working either way).

numpy/f2py/rules.py
da-woods: That logic here was to keep `PyInit_...` working just in case people were doing things with embedding and `PyImport_AppendInittab`. It probably isn't many people, but their code will keep working as long as the  `PyInit_...` function as widely as possible. (There's no issue defining both `PyInit_...` and the newer `PyModExport_...` except for a little bit of extra memory).

I agree it doesn't really matter though so happy to change this either way.

numpy/f2py/rules.py
```diff
@@ -298,7 +298,7 @@
 #ifdef Py_GIL_DISABLED
     {Py_mod_gil, (void*)#gil_used#},
 #endif
-#if (defined(Py_LIMITED_API) && Py_LIMITED_API >= 0x030C0000) || PY_VERSION_HEX >= 0x030C0000
+#if (defined(Py_LIMITED_API) && Py_LIMITED_API >= 0x030C0000) || (!defined(Py_LIMITED_API) && PY_VERSION_HEX >= 0x030C0000)
```
seberg, line 301: Heh, thought none of this uses any API calls and would work always, but doesn't matter.

EDIT: :facepalm: of course `Py_mod_multiple_interpreters` must be supported by the version (I guess the newer version had a "ignore if you don't know" thingy maybe).

seberg (approved):
Thanks @da-woods, this seems safe to me and now is a good time to get the stone rolling for 2.6.

Let's get this in, hopefully it helps with some of your other PRs there.

numpy/f2py/rules.py
da-woods: Thinking about it more, I don't think you support any versions less than 3.12 anyway. So the check is kind of pointless but harmless.

seberg: Technically, downstream could support more then we do, though.  And this get's compiled into the downstream one.

PR #31580, opened 2026-06-07
DOC: clarify ufunc.at multidimensional indexing

### LINKED ISSUE #31579: DOC: ufunc.add behaves differently with indices as tuple and list/array

It comes fairly non-intuitive to me that the `ufunc.at` operates very differently when indices are provided either as array or list, or as a tuple. 

In first two cases, apparently they are taken as 1D indices (and applied multiple times to entire rows). Taking transpose of the array has no effect.

In the third case, when a tuple is provided, they are used as expected, i.e., as 2D indices.

```Python
In [1]: ar = np.zeros([3,3]); np.add.at(ar, np.array([[0,1,2],[0,1,2]]), 1); print(ar)
[[2. 2. 2.]
 [2. 2. 2.]
 [2. 2. 2.]]

In [2]: ar = np.zeros([3,3]); np.add.at(ar, [[0,1,2],[0,1,2]], 1); print(ar)
[[2. 2. 2.]
 [2. 2. 2.]
 [2. 2. 2.]]

In [3]: ar = np.zeros([3,3]); np.add.at(ar, ([0,1,2],[0,1,2]), 1); print(ar)
[[1. 0. 0.]
 [0. 1. 0.]
 [0. 0. 1.]]

```

This is quite an insidious difference in behaviour, particularly because:
* The numerically different result has same data type and shape in all cases, and is given without any explicit warning. Thus it might only manifest itself later in other parts of Python code not working properly.
* Many functions in Python ecosystem use duck typing and strive behave consistently for compound data types. Accordingly, many people, including me, seem to think about tuples just as an immutable, more efficient alternative to a list. Replacing one with another may cause errors sometimes, but should not redefine function behaviour.
* Many others may stick to numpy arrays everywhere for the sake of consistency. They may never get the idea that they should convert their np.array indices to a tuple. 

* I believe this **difference should be described in the docs** https://numpy.org/doc/stable/reference/generated/numpy.ufunc.at.html (at least in the Examples section)
* As it is seemingly unlikely that anybody would index 1D arrays by a list-of-lists containing the 1D indices, the **function should throw a warning about potentially undesired result**. Anybody wanting exactly this behaviour can easily suppress this warning by explicitly concatenating the 1D indices, i.e. using `[[0,1,2, 0,1,2]]` or equivalently `np.hstack([[0,1,2], [0,1,2]])` instead of `[[0,1,2],[0,1,2]]` in the above examples.

muff-c:
## Summary

This adds a small `ufunc.at` example showing how tuple indices select across dimensions, while a list of lists is treated as indices for one dimension.

The goal is just to make the existing behavior easier to spot from the docs.

## Validation

- `python -m py_compile numpy/_core/_add_newdocs.py`
- Ran the added example locally against installed NumPy 2.3.3
- `git diff --check`

Closes #31579

### DIFF
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -5791,6 +5791,23 @@ add_newdoc('numpy._core', 'ufunc', ('at',
     >>> a
     array([2, 4, 3, 4])
 
+    When indexing a multidimensional array, a tuple separates indexes
+    for different dimensions. A list or array is treated as indexes for
+    a single dimension:
+
+    >>> a = np.zeros((3, 3), dtype=int)
+    >>> np.add.at(a, ([0, 1, 2], [0, 1, 2]), 1)
+    >>> a
+    array([[1, 0, 0],
+           [0, 1, 0],
+           [0, 0, 1]])
+    >>> a = np.zeros((3, 3), dtype=int)
+    >>> np.add.at(a, [[0, 1, 2], [0, 1, 2]], 1)
+    >>> a
+    array([[2, 2, 2],
+           [2, 2, 2],
+           [2, 2, 2]])
+
     """))
 
 add_newdoc('numpy._core', 'ufunc', ('resolve_dtypes',


jorenham:

muff-c:
Sorry, you're right. I missed the PR template. I'll leave this closed.

PR #31582, opened 2026-06-08
CI: Add 3.15(t) CI for Windows and MacOS

ngoldbaum:
Over in the numpy-release repo, we [ran into a failing test](https://github.com/numpy/numpy-release/issues/43#issuecomment-4649318074) that only happens on GIL-enabled Python 3.15 on Windows.

This updates the MacOS and Windows CI to use 3.15(t) instead of 3.14(t). It also disables the test causing issues in numpy-release.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -119,7 +119,7 @@ jobs:
         build_runner:
           - [ macos-15-intel, "macos_x86_64" ]
           - [ macos-14, "macos_arm64" ]
-        version: ["3.12", "3.14t"]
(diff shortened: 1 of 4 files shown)


rgommers (approved):
LGTM, thanks @ngoldbaum. This could also use either a code comment or (probably better) an issue to track that that test needs to stop defining `DPy_TARGET_ABI3T` once Meson is updated - and we probably want that before the release in August with wheels for 3.15

ngoldbaum:
Thanks for the review, Ralf! I'm going to merge this and open an issue that links to the code that needs to be updated.

PR #31584, opened 2026-06-08
MAINT: Use Python C API functions for f2py error formatting

da-woods:
### PR summary

My original interest here was to eliminate the non-limited API access to `->tp_name` (and thus it's kind of part of https://github.com/numpy/numpy/issues/21300). However, looking at what I was changing, it seemed like a large amount of C string handling could be removed in the process (and because the result inevitably ends up in Python strings, I don't think this is any less efficient).

The main effect is to eliminate a bunch of fixed-size string buffers and the possibility for overflowing those.

Hopefully most of the output should be identical. There's one place where I've updated a test because the output is slightly different and the old version really didn't make sense to me.

#### AI Disclosure

All code and text was written by me. I used copilot to find a silly mistake I'd made that was causing it to crash (because I couldn't get `gdb` and `spin` to cooperate here) and to sanity check what I'd done.

### DIFF
--- a/numpy/f2py/cfuncs.py
+++ b/numpy/f2py/cfuncs.py
@@ -1400,7 +1400,7 @@ create_cb_arglist(PyObject* fun, PyTupleObject* xa , const int maxnofargs,
             if (!tp_name_obj) goto capi_fail;
             tp_name = PyUnicode_AsUTF8AndSize(tp_name_obj, NULL);
             if (tp_name == NULL) {
-                Py_DECREF(tp_name_obj); goto capi_fail
+                Py_DECREF(tp_name_obj); goto capi_fail;
             }
 #else
             tp_name = Py_TYPE(fun)->tp_name;
--- a/numpy/f2py/cfuncs.py
+++ b/numpy/f2py/cfuncs.py
@@ -534,17 +534,13 @@ cppmacros['CHECKARRAY'] = """
 cppmacros['CHECKSTRING'] = """
 #define CHECKSTRING(check,tcheck,name,show,var)\\
     if (!(check)) {\\
-        char errstring[256];\\
-        snprintf(errstring, sizeof(errstring), \"%s: \"show, \"(\"tcheck\") failed for \"name, slen(var), var);\\
-        PyErr_SetString(#modulename#_error, errstring);\\
+        PyErr_Format(#modulename#_error, \"(\"tcheck\") failed for \"name\": \"show, slen(var), var);\\
         /*goto capi_fail;*/\\
     } else """
 cppmacros['CHECKSCALAR'] = """
 #define CHECKSCALAR(check,tcheck,name,show,var)\\
     if (!(check)) {\\
-        char errstring[256];\\
-        snprintf(errstring, sizeof(errstring), \"%s: \"show, \"(\"tcheck\") failed for \"name, var);\\
-        PyErr_SetString(#modulename#_error,errstring);\\
+        PyErr_Format(#modulename#_error, \"(\"tcheck\") failed for \"name\": \"show, var);\\
         /*goto capi_fail;*/\\
     } else """
 # cppmacros['CHECKDIMS']="""
@@ -840,9 +836,7 @@ character_from_pyobj(character* v, PyObject *obj, const char *errmess) {
         }
     }
     {
-        /* TODO: This error (and most other) error handling needs cleaning. */
-        char mess[F2PY_MESSAGE_BUFFER_SIZE];
-        strcpy(mess, errmess);
+        PyObject *describe_obj;
         PyObject* err = PyErr_Occurred();
         if (err == NULL) {
             err = PyExc_TypeError;
@@ -852,11 +846,11 @@ character_from_pyobj(character* v, PyObject *obj, const char *errmess) {
             Py_INCREF(err);
             PyErr_Clear();
         }
-        size_t len = strlen(mess);
-        snprintf(mess + len, F2PY_MESSAGE_BUFFER_SIZE - len,
-                " -- expected str|bytes|sequence-of-str-or-bytes, got ");
-        f2py_describe(obj, mess + strlen(mess));
-        PyErr_SetString(err, mess);
+        describe_obj = f2py_describe_obj(obj);
+        if (describe_obj) {
+            PyErr_Format(err, "%s -- expected str|bytes|sequence-of-str-or-bytes, got %U", errmess, describe_obj);
+            Py_DECREF(describe_obj);
+        }
         Py_DECREF(err);
     }
     return 0;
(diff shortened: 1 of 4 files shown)


numpy/f2py/src/fortranobject.c
```diff
@@ -17,6 +17,20 @@ extern "C" {
...
+    #define DECREF_TP_NAME(tp_name)
+#elif Py_LIMITED_API >= 0x030D0000
+    #define GET_TP_NAME(obj) obj
+    #define TP_NAME_FMT "%T"
+    #define DECREF_TP_NAME(tp_name)
+#else
+    #define GET_TP_NAME(obj) PyType_GetQualName(Py_TYPE(obj));
+    #define TP_NAME_FMT "%S"
+    #define DECREF_TP_NAME(tp_name) Py_DECREF(tp_name)
+#endif
```
seberg, line 32: Heh, fun definition.  I guess `PyType_GetQualName` is safe (enough) even with an error in progress.

numpy/f2py/src/fortranobject.c
```diff
@@ -222,140 +236,59 @@ fortran_dealloc(PyFortranObject *fp)
...
-        size -= sizeof(notalloc);
-    }
-
-    return p - buf;
+    out = PyUnicode_FromFormat(
+        "array(%" NPY_INTP_FMT "%U)%s",
+        def.dims.d[0], ranks_str,
+        def.data == NULL ? ", not allocated" : "");
+    end:
+    Py_DECREF(ranks_str);
```
seberg:
```suggestion
    Py_XDECREF(ranks_str);
```
unless you are more careful above.

numpy/f2py/src/fortranobject.c
```diff
@@ -999,15 +952,21 @@ ndarray_from_pyobj(const int type_num,
...
+                    elsize, (npy_intp)PyArray_ITEMSIZE(arr));
+            } else {
+                too_small_str = PyUnicode_FromStringAndSize(NULL, 0);
+            }
+            if (too_small_str) {
+                PyErr_Format(
+                    PyExc_ValueError,
+                    "failed to initialize intent(cache) array%s%U",
+                    PyArray_ISONESEGMENT(arr) ? "" : " -- input must be in one segment",
+                    too_small_str);
```
seberg, line 967: Might as well also inline this one?  But also OK not to.

numpy/f2py/src/fortranobject.c
```diff
@@ -1371,41 +1349,41 @@ copy_ND_array(const PyArrayObject *arr, PyArrayObject *out)
...
+            if (!dim) {
+                Py_DECREF(dims_str);
+                goto end;
+            }
+            PyUnicode_AppendAndDel(&dims_str, dim);
+            if (!dims_str) goto end;
+        }
+        result = PyUnicode_FromFormat(
+            "(%U)-%c%" NPY_INTP_FMT "-%s",
+            dims_str, PyArray_DESCR(arr)->kind, PyArray_ITEMSIZE(arr), tp_name);
```
seberg:
```suggestion
            dims_str, PyArray_DESCR(arr)->kind, PyArray_ITEMSIZE(arr), tp_name);
        Py_DECREF(dims_str);
```

numpy/f2py/src/fortranobject.c
```diff
@@ -222,140 +236,59 @@ fortran_dealloc(PyFortranObject *fp)
...
-            p += n;
-            size -= n;
+            formatted_def = PyUnicode_FromStringAndSize("scalar", 6);
         }
+        if (!formatted_def) return NULL;
+        d = PyArray_DescrFromType(def.type);
+        if (d) {
+            s = PyUnicode_FromFormat("%s : '%c'-%U\n", def.name, d->type, formatted_def);
+        }
+        Py_DECREF(d);
```
seberg:
```suggestion
            Py_DECREF(d);
        }
```

numpy/f2py/src/fortranobject.c
```diff
@@ -1371,41 +1349,41 @@ copy_ND_array(const PyArrayObject *arr, PyArrayObject *out)
...
+            PyObject *dim = PyUnicode_FromFormat("%s%" NPY_INTP_FMT ",", i ? " ": "", PyArray_DIM(arr, i));
+            if (!dim) {
+                Py_DECREF(dims_str);
+                goto end;
+            }
+            PyUnicode_AppendAndDel(&dims_str, dim);
+            if (!dims_str) goto end;
+        }
+        result = PyUnicode_FromFormat(
+            "(%U)-%c%" NPY_INTP_FMT "-%s",
```
seberg:
```suggestion
            "(%U)-%c%" NPY_INTP_FMT "-" TP_NAME_FMT,
```
I guess as soon as someone actually tries to run tests with the limited API we would have noticed this :).

numpy/f2py/cfuncs.py
```diff
@@ -1391,10 +1390,29 @@ def errmess(s: str) -> None:
...
+#if defined(Py_LIMITED_API)
+        PyObject *tp_name_obj=NULL;
+#endif
+        if (fun != NULL) {
+#if defined(Py_LIMITED_API)
+            tp_name_obj = PyType_GetFullyQualifiedName(Py_TYPE(fun));
+            if (!tp_name_obj) goto capi_fail;
+            tp_name = PyUnicode_AsUTF8AndSize(tp_name_obj, NULL);
+            if (tp_name == NULL) {
+                Py_DECREF(tp_name_obj); goto capi_fail
```
seberg:
```suggestion
                Py_DECREF(tp_name_obj); goto capi_fail;
```
Same, shouldn't even have compiled with limited API.

seberg (approved):
Annoying to review, but really nice fixes!  Found a couple of issues (in part with an agent), I guess many of them would have been flushed out as soon as one actually tried to compile for it.

Anyway, LGTM, modulo those little fixups.

numpy/f2py/src/fortranobject.c
da-woods: Not quite sure what your proposing here.

I think `too_small_str` was separate because it needs formatting itself if that's what you're saying.

Or is it to just duplicate the whole thing into the "two small" and "not too small" cases above?

numpy/f2py/src/fortranobject.c
da-woods: This is just adapted from Cython where it seems to work fine. In principle it wouldn't be too hard to save/restore the error around `PyType_GetQualName` but doesn't seem worth it given that it works.

da-woods:
> I guess many of them would have been flushed out as soon as one actually tried to compile for it.

Yeah - it a bit hard to test all the limited API code until a sufficiently big chunk of it compiles. So I'm expecting a small "clean up the stuff I got wrong" change at the end.

numpy/f2py/src/fortranobject.c
seberg: yeah, nvm...  (I think the order changed a bit or so, then I stumbled and missed the obvious: that one needs formatting itself)

I guess that might even be possible since Python will stop parsing `va_args`, but it's silly.

seberg:
Cool, let me just get this in then thanks. Since I think I commented on all awkward parts.  The `tp_name` dance is fine FWIW, I would love to have something more elegant, but it isn't bad (and fairly limited).

(Now that I commented and it's still running loongarch, I'll be brutal, har.)

PR #31586, opened 2026-06-09
guard isdigit calls against non-ascii in str_to_int parsers

jmestwa-coder:
loadtxt parses integer fields from a Py_UCS4 buffer, but the digit checks hand raw code points to ctype isdigit():
- str_to_int64 calls isdigit(*p) and isdigit(d) on Py_UCS4 values
- str_to_uint64 does the same in its digit loop
- isdigit is only defined for unsigned char or EOF, so a code point above 255 (e.g. U+0100 in an int column) indexes the ctype table out of bounds on glibc
gated all five checks behind an ascii-range helper.

### DIFF
--- a/numpy/_core/src/multiarray/textreading/str_to_int.h
+++ b/numpy/_core/src/multiarray/textreading/str_to_int.h
@@ -16,6 +16,20 @@
  *
  * The actual functions are defined using macro templating below.
  */
+
+/*
+ * ctype isdigit() is only defined for inputs representable as an unsigned char
+ * (or EOF).  These parsers run over Py_UCS4 code points which can exceed that
+ * range, so a non-ASCII code point would index the ctype table out of bounds.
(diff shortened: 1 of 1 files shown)


jorenham:

jmestwa-coder:
> We prefer to talk to humans. You ignored the PR template.

Hey @jorenham I'm sorry about the template  I forgot it and thats on me I'll do it properly next time but honestly your reply kind of hurt I am a real person and these days I dont even know what im doing half the time im just trying to learn and help where i can and i really thought i found something worth fixing the isdigit calls getting raw Py_UCS4 code points can go out of bounds on glibc maybe im wrong i dont know im still figuring this all out it just felt rough having it closed in a few hours with "we prefer to talk to humans" when i am one a quick "please follow the template and reopen" wouldve meant alot anyway im sorry again i didnt mean to waste your time and thanks for everything you do for numpy

PR #31594, opened 2026-06-10
BUG: ppc64le: define NPY_HAVE_VSX when only higher VSX levels are enabled

### LINKED ISSUE #31593: BUG: ppc64le build fails with cpu-baseline=none - VSX dispatch targets miss NPY_HAVE_VSX (vec_load_len implicit declaration)
### Describe the issue:

On ppc64le (little-endian POWER), building NumPy with `-Dcpu-baseline=none`
  fails to compile. The VSX dispatch-target sources error out with a cascade of:

  ../numpy/_core/src/common/simd/vec/memory.h:63:12: error: implicit declaration
    of function 'vec_load_len' [-Wimplicit-function-declaration]

  plus `vec_vsx_ld`, `vec_xl`, `vec_vsx_st`, `vec_xst`, `vec_extract`, and many
  `cannot convert a value of type 'int' to vector type ... which has different
  size` follow-ons.

  ### Root cause:

  `vec_load_len` is an **s390x/VX (zarch)** intrinsic, not a Power one. The SIMD
  code in `numpy/_core/src/common/simd/vec/memory.h` selects the right branch via
  `NPY_HAVE_VSX`:

  ```c
  #ifdef NPY_HAVE_VSX
      return vec_promote(*npyv__ptr2u64(ptr), 0);   // Power
  #else // VX
      return vec_load_len((const unsigned long long*)ptr, 7);  // s390x
  #endif

  NPY_HAVE_VSX is the generic "any VSX level" flag, and meson_cpu/main_config.h.in
  only includes <altivec.h> under #ifdef NPY_HAVE_VSX.

  With cpu-baseline=none, VSX2/VSX3/VSX4 are enabled only as dispatch targets.
  Each dispatch target is compiled in isolation with just its own
  -DNPY_HAVE_VSX2 (etc.) and not the implied -DNPY_HAVE_VSX. So on a Power
  build the #else // VX branch is wrongly taken and <altivec.h> is never
  included.

  The actual failing compile command (note -DNPY_HAVE_VSX2 but no
  -DNPY_HAVE_VSX):

  gcc ... -O3 -DNPY_HAVE_VSX2 ... -mcpu=power8 -DNPY_MTARGETS_CURRENT=VSX2 ...
      -c numpy/_core/.../loops_arithmetic.dispatch.c

  This is why default CI builds don't hit it: the default ppc64le baseline is VSX2,
  which defines NPY_HAVE_VSX project-wide via add_project_arguments, so dispatch
  targets inherit it. It only surfaces when the baseline doesn't already include VSX
  (e.g. distributors that set cpu-baseline=none for portable binaries — NixOS does
  this)

```python
Minimal standalone reproducer (no NumPy build needed), mirroring npyv__loadl:

  #ifdef NPY_HAVE_VSX
  #  include <altivec.h>
  #endif
  typedef __vector unsigned long long npyv_u64;
  static npyv_u64 npyv__loadl(const void *ptr) {
  #ifdef NPY_HAVE_VSX
      return vec_promote(*(const unsigned long long*)ptr, 0);
  #else /* VX */
      return vec_load_len((const unsigned long long*)ptr, 7);
  #endif
  }
  int main(void){ unsigned long long x[2]={1,2}; npyv__loadl(x); return 0; }

  # fails — simulates a VSX2 dispatch target:
  $ gcc -O3 -mcpu=power8 -DNPY_HAVE_VSX2 -c repro.c
  repro.c: error: implicit declaration of function 'vec_load_len' ...

  # passes — with the implied base flag also defined:
  $ gcc -O3 -mcpu=power8 -DNPY_HAVE_VSX2 -DNPY_HAVE_VSX -c repro.c

  Or, full build: pip install . -Csetup-args=-Dcpu-baseline=none on ppc64le.
```

### Error message:

```shell
../numpy/_core/src/common/simd/vec/memory.h: In function 'npyv__loadl':
  ../numpy/_core/src/common/simd/vec/memory.h:63:12: error: implicit declaration
    of function 'vec_load_len' [-Wimplicit-function-declaration]
     63 |     return vec_load_len((const unsigned long long*)ptr, 7);
        |            ^~~~~~~~~~~~
  ... (vec_vsx_ld, vec_xl, vec_vsx_st, vec_xst, vec_extract, and
  "cannot convert a value of type 'int' to vector type" cascades)
```

NumPy v2.4.4 (also present on main). Python 3.13.

### Runtime Environment:

ppc64le (little-endian, POWER9), Linux. GCC 15.2 and GCC 16.1 both reproduce.
  Build configured with -Dcpu-baseline=none -Dcpu-dispatch=max.

_No response_

supr:
Fixes #31593.

## Problem

On ppc64le, building with `-Dcpu-baseline=none` fails: the VSX dispatch-target
sources hit `implicit declaration of function 'vec_load_len'` (an s390x/VX
intrinsic) and a cascade of undeclared `vec_*` errors, because `<altivec.h>`
is never included.

`NPY_HAVE_VSX` is used throughout the Power SIMD headers as the generic
"any VSX" flag — both to gate the `<altivec.h>` include in
`meson_cpu/main_config.h.in` and to select the Power (vs. s390x/VX) code path
in `simd/vec/*.h`. VSX2/VSX3/VSX4 all imply VSX.

But with an empty baseline, VSX2/3/4 are enabled only as **dispatch targets**,
each compiled in isolation with just its own `NPY_HAVE_VSX{2,3,4}` macro — not
the implied `NPY_HAVE_VSX`. The failing compile line shows `-DNPY_HAVE_VSX2`
with no `-DNPY_HAVE_VSX`. The Power code then falls into the `#else // VX`
branch and `<altivec.h>` is skipped.

Default CI builds don't catch this because the standard ppc64le baseline is
VSX2, which defines `NPY_HAVE_VSX` project-wide; dispatch targets inherit it.

## Fix

Normalize the flag in `meson_cpu/main_config.h.in`: define `NPY_HAVE_VSX`
whenever any of `NPY_HAVE_VSX2/VSX3/VSX4` is defined, before the `<altivec.h>`
include and the SIMD headers consume it.

```c
#if !defined(@P@HAVE_VSX) && (defined(@P@HAVE_VSX2) || defined(@P@HAVE_VSX3) || defined(@P@HAVE_VSX4))
    #define @P@HAVE_VSX 1
#endif
```

The guard is self-gating: it's a no-op on non-Power targets and on any build
that already defines `NPY_HAVE_VSX` (e.g. the default VSX2 baseline). It encodes
the existing implication (VSX2 contains VSX) that the headers already assume.

## Testing

- Minimal reproducer compiles cleanly with the fix; errors without it.
- Full build with `-Dcpu-baseline=none -Dcpu-dispatch=max` on ppc64le (POWER9,
  GCC 15.2) now succeeds; test suite passes (47899 passed, the only failure being
  an unrelated ppc64 long-double string-formatting test,
  `test_ppc64_ibm_double_double128`).

## Alternative considered

Making the meson `features` module propagate implied-feature macros to every
dispatch target generically (which would also help x86 SSE/AVX chains). That's a
larger, riskier change touching the build module; this header-level normalization
is minimal and targeted. Happy to pursue the broader fix if maintainers prefer.

### DIFF
--- a/meson_cpu/main_config.h.in
+++ b/meson_cpu/main_config.h.in
@@ -364,6 +364,20 @@
     #include <x86intrin.h>
 #endif
 
+/*
+ * VSX2/VSX3/VSX4 all imply the base VSX feature on Power. When NumPy is built
+ * with an empty baseline (cpu-baseline=none) and these levels are enabled only
+ * as dispatch targets, each target is compiled with just its own
+ * @P@HAVE_VSX{2,3,4} macro, not the implied @P@HAVE_VSX. The SIMD headers below
+ * (and the <altivec.h> include) gate on @P@HAVE_VSX as the generic any-VSX
+ * flag; without it, Power code wrongly takes the VX (s390x) branch and the
+ * AltiVec/VSX intrinsics are never declared, breaking the build. Normalize by
+ * defining @P@HAVE_VSX whenever any higher VSX level is present.
+ */
+#if !defined(@P@HAVE_VSX) && (defined(@P@HAVE_VSX2) || defined(@P@HAVE_VSX3) || defined(@P@HAVE_VSX4))
+    #define @P@HAVE_VSX 1
+#endif
+
 #if (defined(@P@HAVE_VSX) || defined(@P@HAVE_VX)) && !defined(__cplusplus) && defined(bool)
     /*
      * "altivec.h" header contains the definitions(bool, vector, pixel),


jorenham:

PR #31596, opened 2026-06-10
DOC: fix typo in `PyUFunc_FromFuncAndData` call in "Writing your own ufunc" docs

ikrommyd:
### PR summary
Just a small typo I noticed while learning about ufuncs and gufuncs. I'm pretty sure the `ntypes` argument should be 3 there as the ufunc is defined for 3 dtypes.

https://github.com/numpy/numpy/blob/d0bc7b7fa0473af73080d604f4c266293700b64a/doc/source/user/c-info.ufunc-tutorial.rst?plain=1#L560-L567

#### AI Disclosure
N / A

### DIFF
--- a/doc/source/user/c-info.ufunc-tutorial.rst
+++ b/doc/source/user/c-info.ufunc-tutorial.rst
@@ -590,7 +590,7 @@ is the primary thing that must be changed to create your own ufunc.
                 return NULL;
             }
 
-            logit = PyUFunc_FromFuncAndData(funcs, NULL, types, 4, 1, 1,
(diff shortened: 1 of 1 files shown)


ikrommyd:
Darn it, I should have probably added commands to skip all actions in the commit message :(

I plan to improve the documentation there later on when I learn more and also plan to write a similar page for gufuncs because I couldn't find any docs on how to write gufuncs (only the API docs).

seberg (approved):
Thanks, clearly right, not going to wonder about whether we can make a test for this ;).

ikrommyd:
Probably we could invent a test where this module seg faults haha. Thanks!

PR #31598, opened 2026-06-10
ENH: Make PyFortranObject a heap type

da-woods:
### PR summary

... and also adapt it to work as an opaque object.

The main reason for this is to support the Limited API (#21300). I don't think there's much value in keeping it as a static type outside the Limited API - it isn't possible to inherit from it so there shouldn't be many user-observable differences.

I'm treating "fortranobject.h" as semi-public. It'll mainly be consumed by f2py modules but presumably people could manually link to it and use it. With that in mind I tried to keep `PyFortran_Type` working via a macro.

Additionally, I spotted that it really should be a GC type (because it contains a dictionary with arbitrary contents) and so added that as well.

#### AI Disclosure

~No AI usage~ Edit: AI tracked down the location of `PyDataType_TYPE` for me


numpy/f2py/src/fortranobject.c
```diff
@@ -617,17 +654,55 @@ static PyMethodDef fortran_methods[] = {
...
+    {Py_tp_repr, (void*)fortran_repr},
+    {Py_tp_methods, fortran_methods},
+    {Py_tp_call, (void*)fortran_call},
+    {Py_tp_clear, (void*)fortran_clear},
+    {Py_tp_traverse, (void*)fortran_traverse},
+    {0, NULL}
 };
 
+static PyType_Spec fortran_spec = {
+    .name = "<f2py internal>.fortran",
```
da-woods, line 604: The representation of the type name is a small observable change (although I don't think an important one).

If I just leave it as "fortran" then there's a deprecation warning on import `"builtin type %.200s has no __module__ attribute"`.  Setting it to anything else means it's observable if you do `print(type(some_fortran_object))`.

The solution involves either a custom metaclass as a base type, or a custom class as a descriptor for `__module__`. We did the metaclass in Cython because the module name was a bit important for pickle-ability, but I don't believe that's a concern here.

So my view is that this doesn't matter, but it is an observable change.

numpy/f2py/src/fortranobject.c
```diff
@@ -283,9 +317,9 @@ fortran_doc(FortranDataDef def)
         if (!formatted_def) return NULL;
         d = PyArray_DescrFromType(def.type);
         if (d) {
-            s = PyUnicode_FromFormat("%s : '%c'-%U\n", def.name, d->type, formatted_def);
-            Py_DECREF(d);
-        }
+            s = PyUnicode_FromFormat("%s : '%c'-%U\n", def.name, _PyDataType_GET_ITEM_DATA(d)->type, formatted_def);
```
kumaraditya303: I think this can use `PyDataType_TYPE` instead of direct access which is discouraged.

da-woods: Thanks. I'd somehow failed to find these macros (I think because they're generated somewhere rather than defined in a header)

numpy/f2py/src/fortranobject.c
```diff
@@ -9,6 +9,18 @@ extern "C" {
 #include <stdlib.h>
 #include <string.h>
 
+#if !defined(Py_TARGET_ABI3T) && \
+    NPY_TARGET_VERSION < NPY_2_5_API_VERSION
+// We only use these macros for ABI3t support, so define the
+// simple "cast" version if needed to maximize compatibility with older Numpy.
+#ifndef (PyDataType_TYPE)
+#define PyDataType_TYPE (((PyArray_Descr_fields *)(descr))->type)
```
kumaraditya303: I think these accessor macros exists on older versions so this fallback shouldn't be needed, maybe I am missing something?

I am pretty sure `PyArray_Descr_fields` doesn't exists on older version though.

da-woods: I've definitely messed something up. The version I have locally has `#ifndef PyDataType_TYPE` without brackets and I've managed to push a version with brackets...

Will investigate...

da-woods: I've removed these definitions.

They're definitely documented as being added in Numpy 2.5. They aren't conditional on `NPY_TARGET_VERSION` so they should work with any set target version provided that the headers are up-to-date. I don't know if there's an expectation that F2Py will compile with older headers but I assume it isn't likely to happen in practice.

The `ifndef` wouldn't have worked anyway because they're functions rather than macros.

PR #31599, opened 2026-06-10
ENH: Rationalize f2py callback check to just require callable

da-woods:
### PR summary

... rather than specific types of callable. This means that we can eliminate the PyFortranType_Check1 (which is hard to do in the Limited API) because it's just caught by callable.

Also:
* Add some extra tests (which were already working before the changes, but good to test anyway)
* Get rid of some dead py2 code-paths (`im_func`)

#### AI Disclosure

No AI usage.

### DIFF
--- a/numpy/f2py/cfuncs.py
+++ b/numpy/f2py/cfuncs.py
@@ -1343,6 +1343,35 @@ create_cb_arglist(PyObject* fun, PyTupleObject* xa , const int maxnofargs,
     Py_ssize_t tot, opt, ext, siz, i, di = 0;
     CFUNCSMESS(\"create_cb_arglist\\n\");
     tot=opt=ext=siz=0;
+    /* Check if the function is callable */
+    if (!fun || !(PyCallable_Check(fun) || F2PyCapsule_Check(fun))) {
+#ifdef Py_LIMITED_API
+        PyObject *tp_name_obj = NULL;
+#endif
+        const char *tp_name = "NULL";
+        if (fun != NULL) {
+#ifndef Py_LIMITED_API
+            tp_name = Py_TYPE(fun)->tp_name;
+#else
+            tp_name_obj = PyType_GetQualName(Py_TYPE(fun));
+            if (!tp_name_obj) goto capi_fail;
+            tp_name = PyUnicode_AsUTF8AndSize(tp_name_obj, NULL);
+            if (!tp_name) {
+                Py_DECREF(tp_name_obj);
+                goto capi_fail;
+            }
+#endif
+        }
+        fprintf(
+            stderr,
+            \"Call-back argument must be a Python-callable|f2py-function \"
+            \"but got %s.\\n\", tp_name);
+#ifdef Py_LIMITED_API
+        Py_XDECREF(tp_name_obj);
+#endif
+        goto capi_fail;
+    }
+
     /* Get the total number of arguments */
     if (PyFunction_Check(fun)) {
         tmp_fun = fun;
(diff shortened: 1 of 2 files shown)


numpy/f2py/cfuncs.py
```diff
@@ -1343,39 +1343,52 @@ def errmess(s: str) -> None:
...
-                if (PyCFunction_Check(fun)) {
-                    /* In case the function has a co_argcount */
-                    di = 0;
-                }
-                if (xa != NULL)
-                    tot += PyTuple_Size((PyObject *)xa);
-            }
-            Py_XDECREF(tmp);
-        }
-        else if (PyFortran_Check(fun) || PyFortran_Check1(fun)) {
```
da-woods, line 1372: Really the motivation is just getting rid of this line to avoid adapting `PyFortran_Check1` for the Limited API.

It would definitely be possible to adapt the check though for a less intrusive fix. But the cleanup felt like an improvement.

numpy/f2py/cfuncs.py
```diff
@@ -1342,39 +1342,52 @@ def errmess(s: str) -> None:
...
+
     /* Get the total number of arguments */
     if (PyFunction_Check(fun)) {
         tmp_fun = fun;
         Py_INCREF(tmp_fun);
     }
     else {
         di = 1;
-        if (PyObject_HasAttrString(fun,\"im_func\")) {
-            tmp_fun = PyObject_GetAttrString(fun,\"im_func\");
```
seberg, line 1353: Very strange... I would guess that this stripped the bound `self`, and when `im_func` was removed in Python 3 that behavior just changed!?

(Oh well, not worth digging into, I agree it's clearly OK to remove this.)

numpy/f2py/cfuncs.py
```diff
@@ -1342,39 +1342,52 @@ def errmess(s: str) -> None:
...
-                    /* In case the function has a co_argcount */
-                    di = 0;
-                }
-                if (xa != NULL)
-                    tot += PyTuple_Size((PyObject *)xa);
-            }
-            Py_XDECREF(tmp);
-        }
-        else if (PyFortran_Check(fun) || PyFortran_Check1(fun)) {
+        if (PyObject_HasAttrString(fun,\"__call__\")) {
```
seberg, line 1381: The old code had: `tmp = PyObject_GetAttrString(fun, "__call__")` here in an attempt to unpack callable classes I think.

Not unpacking, means that we can't fetch `.__code__.co_argcount` below for a callable class instance:
```
class A:
    def __call__(self, b):
        return b

a = A()
a.__call__.__code__.co_argcount
```
That said, because `im_func` doesn't exist since Python 3, this all is off by one anyway I would think (thinking the `self` argument is an actual argument).

I dunno, it all seems a mess, I am not sure what is better, but I _think_ it changed?  My suspicion is actually that in all cases f2py callbacks were just utterly broken since Python 3 for any method or callable class and for `staticmethod`s things are fine either way (because `a.static_method` returns the python function object right away)...

seberg (commented):
Thanks, I find it a bit confusing (this seems like an overcomplicated hot mess to guess the number of arguments).

Would be tempted to just try with `inspect.<stuff>` but at this point...
Either way, one question/note, I should have a look at the test, but I suspect it's good.

numpy/f2py/cfuncs.py
da-woods: I had to dig out Python2 to test this, but as far as I can tell, `fun.im_func.__code__` has exactly the same `co_argcount` as `fun.__code__` so I think using `im_func` was probably always a bit pointless but harmless.

numpy/f2py/cfuncs.py
da-woods: I think you're right that I've broken the arg-count check for the arg count for a callable class.

I'll have a look at it and either put that back, or see if I can replace it with `inspect`. (And add a test-case for it).

The arg-count check is a bit weak I think in that it only validates the minimum number of arguments and not the maximum. I'm inclined not to change that here though

da-woods: The checks of classes with `__call__` don't end up doing any arg-count validation on either main or on this branch.
So at least I haven't broken anything that's isn't already broken.

I suspect the line:

```
if (PyCFunction_Check(fun)) {
    /* In case the function has a co_argcount */
    di = 0;
}
```

was mean to be `PyFunction_Check` (although I'm not sure if that's a complete fix or not).

I'm some of the way through trying to do it via inspect. Which seems like it should be better, but doesn't work yet...

seberg: Honestly, I am happy so long we are pretty confident we are not creating a regression by accident...

numpy/f2py/cfuncs.py
da-woods: The other incidental thing that I spotted while having another look is that `PyFortran_Type` has an automatically generated `__call__` so I don't think this path would ever get used anyway.

numpy/f2py/cfuncs.py
da-woods: I think my current opinion is to do `inspect` as a follow-up. It's a more thorough check, and also makes it possible to add a few extra bits of validation (which I think is a good thing). But that does come with a higher risk of regression.

PR #31600, opened 2026-06-10
ENH: Add SIMD acceleration for the Elbrus2000 (e2k) arch

arrowd:
e2k implements MMX, SSE and AVX via compiler intrinsics, which allows it to
piggy-back on the X86 code paths for SIMD.

The test suite is mostly green and I'm working on fixing the remaining failures. They weren't numpy-specific so far, but rather bugs in the compiler, Meson (see https://github.com/numpy/meson/pull/26 ) and other tools.

I'm working on a company that develops an RPM-based Linux distribution for LoongArch and Elbrus2000 arches. The distribution ships the numpy package and I'm working on making the build work and the test suite to pass.

#### AI Disclosure
No AI tools were used.

### DIFF
--- a/meson.options
+++ b/meson.options
@@ -47,6 +47,7 @@ option('test-simd', type: 'array',
          'NEON', 'ASIMD',
          'VX', 'VXE', 'VXE2',
          'LSX',
+         'E2K_SSE', 'E2K_AVX'
        ],
        description: 'CPU SIMD feature sets to be tested by the NumPy SIMD test module')
 option('test-simd-args', type: 'string', value: '',
--- a/meson_cpu/e2k/meson.build
+++ b/meson_cpu/e2k/meson.build
@@ -1,0 +1,27 @@
+source_root = meson.project_source_root()
+current_dir = meson.current_source_dir()
+mod_features = import('features')
+
+E2K_SSE = mod_features.new(
+  'E2K_SSE', 1, args: ['-msse', '-msse2', '-msse3', '-mssse3', '-msse4.1',
+                       '-msse4.2', '-mpopcnt', '-mfpmath=sse',
+                       '-DHWY_WANT_SSE4', '-DHWY_DISABLE_PCLMUL_AES'],
+  # Adds compiler definitions `NPY_HAVE_SSE*`
+  group: ['SSE', 'SSE2', 'SSE3', 'SSSE3', 'SSE41', 'SSE42', 'POPCNT'],
+  detect: 'E2K_SSE',
+  # reuse the x86 features testing code
+  test_code: files(current_dir + '/../x86/test_x86_v2.c')[0],
+)
+E2K_AVX = mod_features.new(
+  'E2K_AVX', 10, args: ['-mavx', '-mavx2', '-mfma', '-mbmi', '-mbmi2',
+                        '-mlzcnt', '-mf16c'],
+  implies: E2K_SSE,
+  # Adds compiler definitions `NPY_HAVE_SSE*`
+  group: ['AVX', 'AVX2', 'FMA3', 'BMI', 'BMI2', 'LZCNT', 'F16C'],
+  detect: 'E2K_AVX',
+  test_code: files(current_dir + '/../x86/test_x86_v3.c')[0],
+)
+
+E2K_FEATURES = {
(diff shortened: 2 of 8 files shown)


seiko2plus:
I don't have too much information about the `e2k` arch, but since it provides the exact `x86` SIMD extensions, then we don't need any new `meson` config. You can just treat it as `x86` by just mimicking the CPU family in here:
https://github.com/numpy/numpy/blob/37d5d57f0037a1c28e9055b1448e187cd3e73176/meson_cpu/meson.build#L96
and let the `meson` compile time tests detect the availability of the compiler support. Even though supporting new architectures is always welcome, but we're supposed to have CI support for it. Also, as for the ongoing move towards `Google Highway`, we also need to be backed by `Highway`. My guess is they wouldn't mind supporting it either according to their policy:
`
Our policy is that unless otherwise specified, targets will remain supported as long as they can be (cross-)compiled with currently supported Clang or GCC, and tested using QEMU. If the target can be compiled with LLVM trunk and tested using our version of QEMU without extra flags, then it is eligible for inclusion in our continuous testing infrastructure. Otherwise, the target will be manually tested before releases with selected versions/configurations of Clang and GCC.
`

arrowd:
> I don't have too much information about the e2k arch, but since it provides the exact x86 SIMD extensions, then we don't need any new meson config.

I can't just reuse the meson code from `meson_cpu/x86/meson.build`, because e2k misses some feature that being tested there. Specifically, it doesn't have `-msahf` in `X86_V2` and `-mmovbe` in `X86_V3`. So instead of adding conditions to x86 code I think it is cleaner to introduce `e2k/meson.build`.

> Also, as for the ongoing move towards Google Highway, we also need to be backed by Highway.

I'll look into that, thanks for the pointer.

seiko2plus:
> I can't just reuse the meson code from meson_cpu/x86/meson.build, because e2k misses some feature that being tested there. Specifically, it doesn't have -msahf in X86_V2 and -mmovbe in X86_V3. So instead of adding conditions to x86 code I think it is cleaner to introduce e2k/meson.build.

Makes sense, but you also can't redefine the x86 group features either; that will affect x86 too.

Also help me understand the CPU runtime detection code:
```C
    npy__cpu_have[NPY_CPU_FEATURE_X86_V2] = 1;
    npy__cpu_have[NPY_CPU_FEATURE_X86_V3] = 1;
    
```

Does the e2k arch support `SSE`/`AVX` as hardware baseline? If yes, then the default baseline should be `v3` and there is no need for adding the `v2` group.

Here is my suggestion, create a new meson feature with a proper name that matches the cpu line.
```meson
E2K_XXX = mod_features.new(
  'E2K_XXX', 1, args: ['-msse', '-msse2', '-msse3', '-mssse3', '-msse4.1', '-msse4.2',
                      '-mpopcnt', ...] # rest of all flags
  # Adds compiler definitions `NPY_HAVE_*` 
  group: ['SSE', 'SSE2', 'SSE3', 'SSSE3', 'SSE41', 'SSE42', 'POPCNT', 'AVX', 'AVX2', 'FMA3', 'BMI', 'BMI2', 'LZCNT', 'F16C'],
  detect: 'E2K_XXX',
  # reuse the x86 features
  test_code: files(current_dir + '/../x86/test_x86_v3.c')[0], 
)
```

Make sure you provide the group name within dispatch-able sources:
https://github.com/numpy/numpy/blob/b3033ed754fb14f91742fc50a249f8765f2a230b/numpy/_core/meson.build#L988-L1058


And, we need CI support too :).

arrowd:
> Does the e2k arch support SSE/AVX as hardware baseline?

It supports SSE/AVX at the intrinsic level, but on older CPUs AVX intrinsics expand into less efficient instrution packs than on newer ones. If I understand it right, having separate V2 and V3 features allows the end user to disable V3 and avoid getting inefficient code.

But now you mentioned that, I suppose we have to dynamically figure out the CPU we're running on in that function?

> And, we need CI support too :).

That'd be a bit problematic for now. It needs https://github.com/numpy/meson/pull/26 , then some fixes on the compiler, QEMU and even glibc parts, then wrapping it up into a GHA action. Not to mention that currently a lot of numpy tests fail on e2k and require further debugging.

PR #31603, opened 2026-06-11
MAINT: Replace CTLF with LF.

charris:
Don't know how those got here. @da-woods I suspect they came with
your work on f2py. AI suggests AI as the root source.

#### AI Disclosure

No AI was used except how to show the line endings of the diff.


charris:
Self merging because it is causing trouble.

da-woods:
Sorry about that.

Almost certainly did come from my changes, but not sure how the got there I'm afraid - on my side it had been traditionally typed into a Linux editor.

jorenham:
> AI suggests AI as the root source.

:rofl:

charris:
> AI suggests AI as the root source.

Grok blames Copilot. 

If it happens again we may want to run checks. It is tricky, as git checks out the files with LF, and they then show up in `git status` as modified. They can't be stashed or fixes pull down from github. I had to commit them, git then removed that commit when the fix was pulled. I suppose I could have changed my git settings temporarily.

da-woods:
In my defence, I think those lines come from the "apply suggestions from code review" commit.

I Sebastian said he ran it through some level of AI review (as did I, although at a different stage...) but I don't know how he generated the suggestion comments.

But it may well be worth guarding again if it happens with commonly used workflows.

charris:
> "apply suggestions from code review" 

I'll bet that is where they came from. Grok suggested something like that was how it could happen. Maybe we need to guide copilot. @seberg Do you recall how you generated the suggestions?

EDIT: Specifically, Grok suggested cut-and-paste from AI generated snippets.

seberg:
Yeah, it clearly came from applying suggestion from code review...

But while, yes, some of these came from an AI run... I know for sure that I wrote at least most (I believe all) of them by hand directly in the browser?!
So I really can't think why this happened here when I do suggestions all the time? Or is the code littered with similar things and we just don't notice?

(And it happened for _all_ of them here, not just a one-off bad edit)

charris:
> Or is the code littered with similar things

I checked, the only other file with CTLF was `doc/make.bat`, which is allowed by our configuration.

charris:
For future reference `git grep -I --name-only $'\r' -- .`  will search for the CR.

seberg:
Well, absolutely beats me, but so long it's a one-off...  I can't think what might have been weird on my side... But I would think that if anything github should normalize it correctly...
The most likely thing to me feels like a github fluke where it normalized it explicitly incorrectly, but also that is rather insane :).

charris:
> But I would think that if anything github should normalize it correctly

Agree.

andyfaff:
How about adding `*.py text=auto eol=lf` and `*.c text=auto eol=lf` to `.gitattributes`? Will this add a lot of churn to the repo, or are most of the files normalised already?

charris:
As of now, they are all normalized.

charris:
AI suggest adding `* text=auto` at the top of the `.gitattributes` file. Currently we specify `text` explicitly for a bunch of file types, apparently the code suggestions might not be one of those file types.  We aren't the first to have this problem. I'm not sure that would actually work, but git stores text internally with lf endings, so it might help to auto detect text.

andyfaff:
> AI suggest adding * text=auto at the top of the .gitattributes file

I think for all the files types we want to be terminated by LF we need to have `eol=lf`

charris:
> we need to have eol=lf

Text is always ended with lf internally.  You have to make a distinction between git's internal storage, and how things are checked out. A bare git repo has data, but shows no files. When the data is exposed in the workspace it may have different line endings depending on the platform and configuration.

PR #31604, opened 2026-06-11
MAINT: Skip tests marked thread_unsafe on free threaded python.

charris:
When running on Python 3.15t in particular, pytest will sometimes run tests marked thread_unsafe and we see DeprecationWarning raised. This also skips two other tests that don't raise warnings.

Not testing the deprecation warnings on free threaded python should be fine, they will be gone soon enough. The other tests should probably be reenabled once pytest gets fixed for free threading.

#### AI Disclosure

AI suggested the IS_FREE_THREADED detection code.


numpy/testing/tests/test_utils.py
```diff
@@ -1150,7 +1151,12 @@ def test_strict(self):
 
 @pytest.mark.filterwarnings(
     "ignore:.*NumPy warning suppression and assertion utilities are deprecated"
-    ".*:DeprecationWarning")
+    ".*:DeprecationWarning"
+    )
+@pytest.mark.skipif(
+    IS_FREE_THREADED,
+    reason="Unreliable under free-threaded Python + pytest-run-parallel"
```
seberg, line 1158: Hmmmm, a bit puzzling and I can't reproduce.  `pytest-run-parallel` shouldn't run these in parallel, so I don't think it should have anything to do with it (I am not 100% sure it doesn't start a new thread, but I don't think so).
Basically, the tests should run identically with and without pytest-run-parallel (but surrounding tests maybe not).

That said, I suspect this is good.  My guess would be that `suppress_warnings` and the more complex free-threaded warning handling in Python somehow interfere...  (And I am not motivated to dig into _how_ precisely, `suppress_warnings` does some things that were valid at the time but I am not shocked if it is just bad now.)
Luckily `suppress_warnings` is long obsoleted and also deprecated, so while I suspect the comment may not be quite spot on, it seems all fine to me.

seberg (approved):
Thanks, LGTM... I wouldn't mind understanding a bit better what you are seeing (I don't see it) and maybe explain that `suppress_warning()` and free-threaded seem to hate each other.  But also OK if not.

numpy/testing/tests/test_utils.py
charris: It turned up in the wheel builds after @andyfaff added 3.15 support.
See https://github.com/numpy/numpy-release/actions/runs/27314091369/job/80690701458. It doesn't always happen, but 2 out of 4 isn't bad.

seberg: Ahh, does it ever fail then or just print out?  Definitely looks like this the warning filter works to the extend of not making the `DeprecationWarning`s an error but not enough to actually hide them.
(I'll wager something to do with everyone patching `warnings.showwarnings()` or whatever it was.)
`--parallel-threads=0` doesn't change that, but then all adds up, this warning spam seems to exist on all free-threaded versions, I guess.

Either way, I guess that means it's fine to skip these test even if it may just be to avoid the spurious `DeprecationWarning` captures from pytest (for whatever reason).

---

That test failed with a MemoryError maybe really too much memory though? That is a bit strange considering that it is marked as thread unsafe.
Something about GC kicking in late maybe?!

<details>

```
  ================================== FAILURES ===================================
  ____________ TestRandomDist.test_dirichlet_moderately_small_alpha _____________
  [gw0] win32 -- Python 3.15.0 C:\Users\runneradmin\AppData\Local\Temp\cibw-run-c1u5370t\cp315t-win32\venv-test\Scripts\python.exe
  
  self = <numpy.random.tests.test_generator_mt19937.TestRandomDist object at 0x0F9F3F38>
  
      @pytest.mark.slow
      @pytest.mark.thread_unsafe(reason="crashes with low memory")
      def test_dirichlet_moderately_small_alpha(self):
          # Use alpha.max() < 0.1 to trigger stick breaking code path
          alpha = np.array([0.02, 0.04, 0.03])
          exact_mean = alpha / alpha.sum()
          random = Generator(MT19937(self.seed))
  >       sample = random.dirichlet(alpha, size=20000000)
  
  alpha      = array([0.02, 0.04, 0.03])
  exact_mean = array([0.22222222, 0.44444444, 0.33333333])
  random     = Generator(MT19937) at 0xD0EBA80
  self       = <numpy.random.tests.test_generator_mt19937.TestRandomDist object at 0x0F9F3F38>
  
  ..\venv-test\Lib\site-packages\numpy\random\tests\test_generator_mt19937.py:1275: 
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
  
  >   ???
  E   numpy._core._exceptions._ArrayMemoryError: Unable to allocate 458. MiB for an array with shape (20000000, 3) and data type float64
```

ngoldbaum: > Something about GC kicking in late maybe?!

Could be, the free-threaded GC happens with a delay. It's a stop-the-world GC and needs to be coordinated across threads.

ngoldbaum (commented):
Pytest-run-parallel has an option to skip thread-unsafe tests. It's not necessary to modify NumPy internals at all: https://github.com/Quansight-Labs/pytest-run-parallel#features

Pass `--skip-thread-unsafe=True` to the pytest invocation.

That said, I intentionally set up the pytest-run-parallel CI job to not do that, because otherwise thread-unsafe tests will never run *at all* on 3.14t. I would rather not add the skip and instead I would prefer to fix whatever issue you're running into in pytest-run-parallel.

In the future I'd appreciate it if you could file an issue, ping my github handle, and wait for me to take a look before opening a PR that simply disables things for free-threading.

ngoldbaum:
Oh I see, this has nothing to do with pytest-run-parallel at all.

I definitely don't want to *always* skip a bunch of tests on the free-threaded build. Let me try to reproduce this issue so we can understand it better.

ngoldbaum:
This happens on win32 - weren't we going to disable win32 wheels anyway?

seberg:
@ngoldbaum we need to figure out the memory error.  For the spurious warnings, I don't think its worth an issue... Something is confused, but its *deprecated* `suppress_warnings()` _plus_ pytest warning capture _plus_ free-threading.

I.e. there is no pytest-run-parallel issue, we can just chose to silence it by skipping these tests (which seems OK to me).

That said, this shouldn't fix the the `MemoryError`.  Not sure what is best there, maybe we can just add a `gc.collect()` even @ngoldbaum?!

ngoldbaum:
> That said, this shouldn't fix the the MemoryError. Not sure what is best there, maybe we can just add a gc.collect() even @ngoldbaum?!

Let me see if I can reproduce this with a win32 Python install on my Windows box...

charris:
> This happens on win32 - weren't we going to disable win32 wheels anyway?

Yes, but not on 2.5.x, and we want to continue to support it, just not post wheels for it -- Tier 3.

ngoldbaum:
I opened https://github.com/numpy/numpy-release/pull/52 with an attempt at a workaround for the memory issue.

It could be due to memory fragmentation and not necessarily an OOM per se.

ngoldbaum:
Closing in favor of #31607.

PR #31607, opened 2026-06-11
TST: add 3.15t win32 CI and fix issues seen in that build

ngoldbaum:
Adds win32 3.15t CI and fixes issues seen in the numpy-release automation. I also enabled "full" tests using the same test command as the windows wheel builder.

The Cython issues were reported upstream in https://github.com/cython/cython/issues/7742 and will be fixed in the next Cython release.

I adjusted tests that run meson to use a new helper that embeds debug info into the pytest failure message. That way debug output is no longer swallowed by pytest-xdist and we'll see errors in CI.

Fixed an issue in the cython tests that led to file paths longer than MAX_PATH so I updated the tests to copy the build tree to generate shorter paths.

Disabled `test_dirichlet_moderately_small_alpha` on 32 bit builds. 458 MiB is almost 10% of a 32 bit address space and on a test runner with 4 workers it's possible or even likely for memory to get sufficiently fragmented such that this test triggers OOM errors. mimalloc and the free-threaded build might make this situation worse?

ping @seberg @charris since this got reasonably involved.

I used Claude Fable 5 to help with this PR.


charris:
You may want to disable most of the other tests just to speed things up.

ngoldbaum:
It looks like there are a bunch of other issues related to Cython that the wheel builder job doesn't hit. I'll have to dig in to see what's happening there...

ngoldbaum:
https://github.com/cython/cython/issues/7742 is the cause of the cython crashes

charris:
This looks pretty straight forward. Because we will be supporting 3.15 in 2.5.x, it would probably make things easier if we backported these fixups when they are ready, we don't need to create wheels for them.

charris:
Note to self: backport Cython update when it happens.

charris:
I note that the DeprecationWarnings are still being captured. Any plans for that?

ngoldbaum:
That is a different bug in how those warnings are generated that requires tricky C fixes, I'll open a separate PR with what Claude came up with.

ngoldbaum:
I opened https://github.com/numpy/numpy/pull/31617 for the datetime/timedelta warnings.

numpy/testing/_private/utils.py
```diff
@@ -2868,3 +2868,21 @@ def wrapper(*args, **kwargs):
                         "deep recursion")
         return func(*args, **kwargs)
     return wrapper
+
+
+def run_meson(args, build_dir):
```
seberg: I like the idea of adding a helper, we have a few places that work around it, but even assigning to locals, `pytest` may not print enough to make out exactly.

How about renaming it, dunno maybe `check_subprocess()` and not including the `"meson"` so that we can use it for all such tests.

numpy/random/tests/test_generator_mt19937.py
```diff
@@ -1267,6 +1268,10 @@ def test_dirichlet_small_alpha(self):
 
     @pytest.mark.slow
     @pytest.mark.thread_unsafe(reason="crashes with low memory")
+    @pytest.mark.skipif(
+        not IS_64BIT,
+        reason="the 458 MiB sample array may not fit in a 32-bit "
+               "address space fragmented by earlier tests")
```
seberg, line 1274: Sounds fine, thanks, doesn't seem worth skipping only if allocation fails for this.

numpy/_core/tests/test_limited_api.py
```diff
@@ -38,8 +40,17 @@ def install_temp(tmpdir_factory):
     if IS_WASM:
         pytest.skip("No subprocess")
 
-    srcdir = os.path.join(os.path.dirname(__file__), 'examples', 'limited_api')
-    build_dir = tmpdir_factory.mktemp("limited_api") / "build"
+    # Build against a copy of the sources placed next to the build dir:
+    # meson refers to sources via paths relative to the build dir, and on
+    # Windows the unnormalized cwd + `..` chain joining the deeply nested
+    # pytest tmp dir and site-packages can exceed MAX_PATH, failing the
+    # compile with "Cannot open source file".
```
seberg, line 47: Fun when we actually pass the absolute dir, I guess.

.github/workflows/windows.yml
```diff
@@ -116,10 +118,10 @@ jobs:
         run: |
           python -m pip install -r requirements/test_requirements.txt
 
-      - name: Run test suite (fast)
+      - name: Run test suite (full)
         run: |
           cd tools
-          python -m pytest --pyargs numpy -m "not slow" -n2 --timeout=600 --durations=10
+          python -c "import sys; import numpy; sys.exit(not numpy.test(label='full', extra_argv=['-n=auto', '--durations=10', '--timeout=600']))"
```
seberg: This change seems unnecessary?  I get it that the `cibw_test_command` uses it but diverging the way we run them here feels confusing.

numpy/_core/tests/examples/cython/meson.build
```diff
@@ -15,6 +15,14 @@ if cy.version().version_compare('>=3.1.0')
   cython_args += ['-Xfreethreading_compatible=True']
 endif
 
+extra_c_args = []
+if host_machine.system() == 'windows' and host_machine.cpu_family() == 'x86'
+  # Cython's constant immortalization passes UINT32_MAX through a Py_ssize_t,
+  # which aborts module init on 32-bit free-threaded builds when asserts are
+  # enabled. See https://github.com/cython/cython/issues/7742.
+  extra_c_args += ['-DCYTHON_IMMORTAL_CONSTANTS=0']
```
seberg, line 24: Could mention that it should be fixed by Cython 3.2.6, but doesn't matter much.

seberg (commented):
Seems good, but maybe Chuck wants to look at something.  I think it is worth making that helper not meson specific even if we mostly/only use it for meson subprocess runs right now.

But generally looks all good and basically ready to merge.

charris:
Let's get this in. There are ways to avoid long, long lines in yml, but we can live with a few.

charris:
Thanks Nathan.

Issue #24859, opened 2023-10-04
ENH: dtype API needs an API slot for a single-element copy


ngoldbaum:
Currently, users will see seg faults if a user-defined dtype is passed somewhere where `copyswap` is used internally to copy array elements.

If we require that user-defined dtypes implement a single-element copy operation we will be able to implement a default copyswap implementation for user-defined dtypes which can be used where copyswap is used internally in NumPy.

seberg:
There are two slightly different use-cases here.  In most cases, where we need a single element copy, we actually need to copy many, which means preparation is useful/necessary.  But in some we may not need it.

The question is how we deal with allowing preparation (noting that for the majority of dtypes the preparation would be at best specializing for short copies or alignedness and it would be good if NumPy just provides that copy function for those without references at least.).

seberg:
We should do this, but probably not a disaster if it's not in 2.0, I think.

ngoldbaum:
https://github.com/numpy/numpy/issues/30795 has some more motivating examples for how this is still annoying

PR #28206, opened 2025-01-21
BUG: Remove unnecessary copying and casting from out array in choose

MaanasArora:
Resolves #22237.

This PR does not copy the existing `out` array in `np.choose`, instead creating the new output array from scratch and copying it back manually.

Thank you for reviewing!

PR #31609, opened 2026-06-11
BUG: fix StringDType distinct-allocator bugs and add tests

ngoldbaum:
Fixes bugs when multiple distinct stringdtype arrays are passed to `np.choose`, `np.put`, `np.putmask`, `np.place`, flatiter reads, and a bug that could lead to unsafe views being created for StringDType and object arrays.

Also exposes a new C helper function `_is_safe_view_cast` which takes `view_offset` into account and uses it in the Python helper functions for view safety. It only uses this function for StringDType. We could use it more generally but that might lead to breakage, so just fixing StringDType seemed like a practical first pass.

The C changes all work out to fixing the same couple patterns: legacy paths that used copyswap (e.g. #24859, which this might obviate the need for?) or code paths that assumed all input arrays have the same descriptor instance.

Also adds tests for functionality that is fixed here or is poorly tested currently. It makes use of a new helper function that allocates two distinct arrays of strings with allocations that live in the arena buffer to trigger these bugs.

I milestoned this for 2.5.0 and marked it for backport because these are all bugs that can cause segfaults.

I used Claude Fable 5 to work on this, including for the implementation and new tests.

### DIFF
--- a/numpy/_core/_internal.py
+++ b/numpy/_core/_internal.py
@@ -463,15 +463,6 @@ def _promote_fields(dt1, dt2):
     return res
 
 
-def _check_stringdtype_view(oldtype, newtype):
-    # equal StringDType instances can still own separate allocators
-    if oldtype.kind == "T" and not _is_view_safe_cast(oldtype, newtype):
-        raise TypeError(
-            "Cannot view StringDType array data through a different dtype "
-            "instance; use astype to copy the data to the new dtype "
-            "instance")
-
-
 def _getfield_is_safe(oldtype, newtype, offset):
     """ Checks safety of getfield for object arrays.
 
@@ -494,8 +485,7 @@ def _getfield_is_safe(oldtype, newtype, offset):
 
     """
     if newtype.hasobject or oldtype.hasobject:
-        if offset == 0 and newtype == oldtype:
-            _check_stringdtype_view(oldtype, newtype)
+        if offset == 0 and _is_view_safe_cast(oldtype, newtype):
             return
         if oldtype.names is not None:
             for name in oldtype.names:
@@ -525,10 +515,10 @@ def _view_is_safe(oldtype, newtype):
 
     """
 
-    # if the types are equivalent, there is no problem.
-    # for example: dtype((np.record, 'i4,i4')) == dtype((np.void, 'i4,i4'))
-    if oldtype == newtype:
-        _check_stringdtype_view(oldtype, newtype)
+    # more precise than ``oldtype == newtype``: e.g. dtype((np.record, 'i4,i4'))
+    # views safely as dtype((np.void, 'i4,i4')), while two equal StringDType
+    # instances with separate allocators do not
+    if _is_view_safe_cast(oldtype, newtype):
         return
 
     if newtype.hasobject or oldtype.hasobject:
--- a/numpy/_core/_internal.py
+++ b/numpy/_core/_internal.py
@@ -13,6 +13,7 @@ import warnings
 from numpy import _NoValue
 from numpy.exceptions import DTypePromotionError
 
+from ._multiarray_umath import _is_view_safe_cast
 from .multiarray import StringDType, array, dtype, promote_types
 
 try:
@@ -462,6 +463,15 @@ def _promote_fields(dt1, dt2):
     return res
 
 
+def _check_stringdtype_view(oldtype, newtype):
+    # equal StringDType instances can still own separate allocators
+    if oldtype.kind == "T" and not _is_view_safe_cast(oldtype, newtype):
+        raise TypeError(
+            "Cannot view StringDType array data through a different dtype "
+            "instance; use astype to copy the data to the new dtype "
+            "instance")
+
+
 def _getfield_is_safe(oldtype, newtype, offset):
     """ Checks safety of getfield for object arrays.
 
@@ -485,6 +495,7 @@ def _getfield_is_safe(oldtype, newtype, offset):
     """
     if newtype.hasobject or oldtype.hasobject:
         if offset == 0 and newtype == oldtype:
+            _check_stringdtype_view(oldtype, newtype)
             return
         if oldtype.names is not None:
             for name in oldtype.names:
@@ -517,6 +528,7 @@ def _view_is_safe(oldtype, newtype):
     # if the types are equivalent, there is no problem.
     # for example: dtype((np.record, 'i4,i4')) == dtype((np.void, 'i4,i4'))
     if oldtype == newtype:
+        _check_stringdtype_view(oldtype, newtype)
         return
 
     if newtype.hasobject or oldtype.hasobject:
--- a/numpy/_core/src/multiarray/compiled_base.c
+++ b/numpy/_core/src/multiarray/compiled_base.c
@@ -393,17 +393,18 @@ arr_place(PyObject *NPY_UNUSED(self), PyObject *args, PyObject *kwdict)
     src = PyArray_DATA(values);
     j = 0;
 
-    if (PyDataType_REFCHK(PyArray_DESCR(array))) {
-        /* dtypes with references may not define copyswap */
+    copyswap = PyDataType_GetArrFuncs(PyArray_DESCR(array))->copyswap;
+    if (copyswap == NULL) {
         NPY_cast_info cast_info;
         NPY_ARRAYMETHOD_FLAGS flags;
         const npy_intp one = 1;
-        const npy_intp strides[2] = {chunk, chunk};
+        const npy_intp elsize = chunk;
+        const npy_intp strides[2] = {elsize, elsize};
 
         NPY_cast_info_init(&cast_info);
         if (PyArray_GetDTypeTransferFunction(
                 PyArray_ISALIGNED(values) && PyArray_ISALIGNED(array),
-                chunk, chunk,
+                strides[0], strides[1],
                 PyArray_DESCR(values), PyArray_DESCR(array), 0,
                 &cast_info, &flags) < 0) {
             goto fail;
@@ -432,7 +433,6 @@ arr_place(PyObject *NPY_UNUSED(self), PyObject *args, PyObject *kwdict)
         NPY_cast_info_xfree(&cast_info);
     }
     else {
-        copyswap = PyDataType_GetArrFuncs(PyArray_DESCR(array))->copyswap;
         NPY_BEGIN_THREADS_DESCR(PyArray_DESCR(array));
         for (i = 0; i < ni; i++) {
             if (mask_data[i]) {
--- a/numpy/_core/src/multiarray/compiled_base.c
+++ b/numpy/_core/src/multiarray/compiled_base.c
@@ -16,6 +16,7 @@
 #include "ctors.h"
 #include "common.h"
 #include "dtypemeta.h"
+#include "dtype_transfer.h"
 #include "simd/simd.h"
 
 #include <string.h>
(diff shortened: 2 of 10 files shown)


ngoldbaum:
that debug interpreter test crash looks real

numpy/_core/src/multiarray/convert_datatype.c
```diff
@@ -648,6 +648,30 @@ PyArray_SafeCast(PyArray_Descr *type1, PyArray_Descr *type2,
 }
 
 
+/*
+ * Python-level helper answering whether reinterpreting data of dtype *from*
+ * as dtype *to* is a view. Unlike descriptor equality or `np.can_cast` with
+ * "no" casting, this distinguishes equivalent descriptor instances with
+ * separate internal state (e.g. StringDType allocators).
```
seberg: Nice helper!

ngoldbaum:
I think this is ready for review now. @seberg @MaanasArora @mhvk - hope you don't mind the ping. The C diff isn't so huge, most of the changes are new tests.

numpy/_core/src/multiarray/item_selection.c
MaanasArora, line 1040: For the `PyArray_Choose` fix here, I noticed we actually discussed this bug in https://github.com/numpy/numpy/pull/28206#discussion_r1944607085. Is it better to fix the `PyArray_ConvertToCommonType` function directly to convert all arrays to use an identical descriptor? Then I think we will lose the complexity below, as we only have one descriptor to cast from!

seberg: (have to catch up with actual review still; we forgot to open a follow-up issue back then, I guess.)

Theoretically, but it would require copying all the inputs even if they have the right dtypes already basically; and it would be a bit strange to force that (i.e. the point normally is for all arrays to have their own dtype for `StringDType`).

But without that, each casts needs the identity dtype passed in and that requires one cast per array.

MaanasArora: OK yes, might be a bit strange to copy and then cast separately later, thanks.

ngoldbaum: > have to catch up with actual review still; we forgot to open a follow-up issue back then, I guess.

Sorry for dumping a bunch of code on you on a Friday. Take your time!

numpy/_core/src/multiarray/compiled_base.c
```diff
@@ -392,19 +393,59 @@ arr_place(PyObject *NPY_UNUSED(self), PyObject *args, PyObject *kwdict)
...
     j = 0;
 
-    copyswap = PyDataType_GetArrFuncs(PyArray_DESCR(array))->copyswap;
-    NPY_BEGIN_THREADS_DESCR(PyArray_DESCR(array));
-    for (i = 0; i < ni; i++) {
-        if (mask_data[i]) {
-            if (j >= nv) {
-                j = 0;
+    if (PyDataType_REFCHK(PyArray_DESCR(array))) {
+        /* dtypes with references may not define copyswap */
```
seberg: Please change this to an explicit check for `->copyswap` being NULL, while it may be true in practice, I don't think the comment really adds up.

numpy/_core/src/multiarray/compiled_base.c
```diff
@@ -392,19 +393,59 @@ arr_place(PyObject *NPY_UNUSED(self), PyObject *args, PyObject *kwdict)
...
+        NPY_cast_info cast_info;
+        NPY_ARRAYMETHOD_FLAGS flags;
+        const npy_intp one = 1;
+        const npy_intp strides[2] = {chunk, chunk};
```
seberg: This is better as elsize (zero is also OK, but elsize is probably better in practice to hit the contiguous loops).
`chunk` just doesn't make sense here at all.

numpy/_core/src/multiarray/compiled_base.c
```diff
@@ -392,19 +393,59 @@ arr_place(PyObject *NPY_UNUSED(self), PyObject *args, PyObject *kwdict)
...
+
+        NPY_cast_info_init(&cast_info);
+        if (PyArray_GetDTypeTransferFunction(
+                PyArray_ISALIGNED(values) && PyArray_ISALIGNED(array),
+                chunk, chunk,
```
seberg: Should be identical to strides above, just use:

```suggestion
                strides[0], strides[1],
```

numpy/_core/src/multiarray/convert_datatype.c
```diff
@@ -648,6 +648,30 @@ PyArray_SafeCast(PyArray_Descr *type1, PyArray_Descr *type2,
...
+ */
+NPY_NO_EXPORT PyObject *
+_is_view_safe_cast(PyObject *NPY_UNUSED(module), PyObject *args)
```
seberg: Thanks, I really like this helper!  I do have a bit of a dislike for `PyArg_ParseTuple` (it's just silly slow API...), but I don't mind keeping it.

numpy/_core/src/multiarray/getset.c
```diff
@@ -671,26 +674,42 @@ array_flat_set(PyArrayObject *self, PyObject *val, void *NPY_UNUSED(ignored))
         retval = 0;
         goto exit;
     }
-    swap = PyArray_ISNOTSWAPPED(self) != PyArray_ISNOTSWAPPED(arr);
-    copyswap = PyDataType_GetArrFuncs(PyArray_DESCR(self))->copyswap;
     if (PyDataType_REFCHK(PyArray_DESCR(self))) {
+        /* dtypes with references may not define copyswap */
```
seberg: Again, check explicitly for copyswap.

Effectively it means that StringDType shouldn't implement it even if we allowed that (it's far better off with this pattern due to lock handling) but that seems reasonable.

Now... what is completely crazy here is the whole manual `PyArray_Item_XDECREF` nonsense, which looks like `copyswap()` would handle it just fine for objects (and structured dtypes).

For structured dtypes the new path may well be better (for larger arrays).  So I would be fine with keeping the REFCHCK _additionally_ to `copyswap == NULL`.

(Another interesting thing, but let's not act on it, I guess... This function should release the GIL for larger inputs but doesn't seem to.)

numpy/_core/src/multiarray/getset.c
```diff
@@ -671,26 +674,42 @@ array_flat_set(PyArrayObject *self, PyObject *val, void *NPY_UNUSED(ignored))
...
+        NPY_cast_info cast_info;
+        NPY_ARRAYMETHOD_FLAGS transfer_flags = 0;
+        npy_intp one = 1;
```
seberg, line 683: Yeah, would be cute to maybe just _always_ use this path but iterate all but one axes (like a ufunc).
But not here for backporting, this is already an annoying amount of complexity.

numpy/_core/src/multiarray/item_selection.c
seberg, line 752: Just to note that I am not insane about the copyswap thing above.  This path is different, because we assume that `memcpy` is sufficient (copyswap isn't even used).

(There is a new flag for optimizing structured dtypes that we should possibly check here as well, `NOT_TRIVIALLY_COPYABLE` but not this PR really... Won't complain if you add it, it's basically an extremely niche fix on structured dtypes.)

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -1119,15 +1122,27 @@ PyArray_Choose(PyArrayObject *ip, PyObject *op, PyArrayObject *out,
...
-        int is_aligned = IsUintAligned(obj);
         PyArray_Descr *obj_dtype = PyArray_DESCR(obj);
-        PyArray_GetDTypeTransferFunction(
-                    is_aligned,
-                    dtype->elsize,
-                    obj_dtype->elsize,
-                    dtype,
-                    obj_dtype, 0, &cast_info,
-                    &transfer_flags);
+        cast_infos = PyMem_New(NPY_cast_info, n);
```
seberg: Could use `NPY_ALLOC_WORKSPACE` or... just use `NPY_MAXARGS` the MultiIter is limited to that anyway (although of course it isn't explitily from here).

(I don't care though, I doubt micro-optimizations matter much for `choose`.)

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -1119,15 +1122,27 @@ PyArray_Choose(PyArrayObject *ip, PyObject *op, PyArrayObject *out,
...
+        }
+        for (i = 0; i < n; i++) {
+            int is_aligned = IsUintAligned(obj) && IsUintAligned(mps[i]);
+            if (PyArray_GetDTypeTransferFunction(
+                        is_aligned,
+                        PyArray_DESCR(mps[i])->elsize,
+                        obj_dtype->elsize,
+                        PyArray_DESCR(mps[i]),
+                        obj_dtype, 0, &cast_infos[i],
+                        &transfer_flags) < 0) {
```
seberg, line 1141: We *could* release the GIL here based on `transfer_flags` (although, I suppose while it _can't_ matter, in principle we should check all transfer flags).

This doesn't matter for objects, but it does matter a bit for strings.

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -3420,6 +3421,7 @@ PyArray_Where(PyObject *condition, PyObject *x, PyObject *y)
                         if (y_cast_info.func(
                                 &y_cast_info.context, args, &one,
                                 y_strides, y_cast_info.auxdata) < 0) {
+                            NPY_END_THREADS;
```
seberg: Seems nicer to me to just put it into `fail:` here (requires moving the `THREADS_DEF` to the start of the function).

numpy/_core/_internal.py
```diff
@@ -462,6 +463,15 @@ def _promote_fields(dt1, dt2):
     return res
 
 
+def _check_stringdtype_view(oldtype, newtype):
+    # equal StringDType instances can still own separate allocators
+    if oldtype.kind == "T" and not _is_view_safe_cast(oldtype, newtype):
+        raise TypeError(
+            "Cannot view StringDType array data through a different dtype "
+            "instance; use astype to copy the data to the new dtype "
+            "instance")
```
seberg: If you care for the `StringDType` error that is fine, but I don't think you need to guard for it here, in theory `StringDType` isn't special.
Maybe `_is_view_safe_cast` is not the best name here, but that function can completely replace the `newtype == oldtype`, I think?

seberg (commented):
Thanks, a few comments.  In some places I think you/the bot thought the RefCheck was a pattern to generalize, but it is used in a few places due to bare `memcpy` use not to avoid `copyswap` (although it may be fine to do so _additionally_).

Otherwise a few small nits, many take-it-or-leave it. I do also think the view safety check should be used generally if we create it, there is nothing special about `StringDType` (and it can short-circuit the `==`, I think.

numpy/_core/src/multiarray/convert_datatype.c
ngoldbaum: Changed it to `METH_FASTCALL`.

ngoldbaum:
I think I applied all your suggestions in the last commit. If you could give it another look I'd appreciate it.

seberg (approved):
Thanks, only had a brief look through the new changes, but looks all good.  (If you feel like it and haven't done so, I might have thrown an agent once more but don't have it set up here right now.)

Please go ahead and merge!

ngoldbaum:
In it goes, thanks for taking a look and giving me feedback Sebastian!

PR #31610, opened 2026-06-12
MAINT: add gitattribute to make all line endings LF

andyfaff:
### PR summary
Specify gitattribute to make line endings for text files \n. This should help prevent things like #31603 being necessary.

#### AI Disclosure
No AI used.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,9 @@
+# We want all text files to be /n
+# Specific settings below this can still override it
+# e.g.
+# *.bat   text eol=crlf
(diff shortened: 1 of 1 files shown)


ngoldbaum:
Seems fine to me and I think the cost of doing this is zero given we already set the correct line-endings for .cmd and .bat below this addition. I also explicitly checked that doing `git add --renormalize .` in this branch is a no-op.

If this happens to cause issues with anyone working on Windows we can re-think it.

Thanks @andyfaff.

PR #31614, opened 2026-06-12
ENH: New-style ArrayMethods for partitioning

MaanasArora:
### PR summary
Draft PR to move partitions and argpartitions to use new ArrayMethods that can be registered in the same way as sorts. The `kth` array is passed too, so `nin=2` (the second being of INTP type). That is really the only major change from the sort/argsort implementation, I think!

I didn't port the actual built-in implementations to keep the PR small. I also didn't use the "real" sorts for fallbacks, so the logic is a a bit weirdly asymmetric right now - I think I'll fix that as I iterate in this PR! I did export to user-dtypes (was trivial, the only place where to test this right now I guess), and added basic `sfloat` tests!

This is a bit drafty, so not really ready for review line-by-line, but general feedback is welcome, thank you!

#### AI Disclosure
No AI tools used except for line completion with GitHub Copilot.


numpy/_core/src/multiarray/item_selection.c
```diff
@@ -1714,11 +1751,78 @@ PyArray_Partition(PyArrayObject *op, PyArrayObject * ktharray, int axis,
...
+            &context, 1, 0, strides, &strided_loop, &auxdata, &method_flags) < 0) {
+                ret = -1;
+                goto fail;
+        }
+    }
+    else {
+        part = get_partition_func(PyArray_TYPE(op), which);
+    }
+
+    if (method == NULL && part == NULL) {
```
seberg, line 1798: Can't we just replace this whole thing with a call to `PyArray_Sort()` (or something slightly lower?).
Feels like not worth the trouble to duplicate sort logic here (even if it might make some errors or so a bit nicer).

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -1714,11 +1751,78 @@ PyArray_Partition(PyArrayObject *op, PyArrayObject * ktharray, int axis,
...
+            loop_descrs[0]->elsize, loop_descrs[1]->elsize, loop_descrs[2]->elsize};
+
+        if (method->get_strided_loop(
+            &context, 1, 0, strides, &strided_loop, &auxdata, &method_flags) < 0) {
+                ret = -1;
+                goto fail;
+        }
+    }
+    else {
+        part = get_partition_func(PyArray_TYPE(op), which);
```
seberg, line 1795: Question: do we really still need this?  For sorts we do, because it is public API for the time being.
However, for partitioning it is not so we should be able to just make sure to cover all partitioning paths we support (without sorts) via the new array-methods and delete this.

seberg (commented):
Two small quick comments, but seems like this is coming together nicely.

numpy/_core/src/multiarray/item_selection.c
MaanasArora: Thanks, yeah, I was actually thinking of adding a helper simply because coupling the logic doesn't seem very nice, but honestly this is really sorting so might as well call `PyArray_Sort` for now.

numpy/_core/src/multiarray/item_selection.c
MaanasArora: Yes, we only need this in this PR because the partitions aren't registered for built-in types yet, so I was thinking of removing them when they are! I thought that could be follow-up (the registration).

doc/source/reference/c-api/array.rst
```diff
@@ -2171,6 +2171,36 @@ and not set any of the other loop slots.
 These specs can be registered using :c:func:`PyUFunc_AddLoopsFromSpecs`
 along with other ufunc loops.
 
+Partitioning and Argpartitioning
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Similarly to sorting and argsorting, partitioning and argpartitioning methods
+can be registered using the ArrayMethod API. This is done by adding an
+ArrayMethod spec with the name ``"partition"`` or ``"argpartition"`` respectively.
```
MaanasArora, line 2179: Probably not worth "advertising" that sorting will be used as a fallback?

numpy/_core/src/multiarray/item_selection.c
MaanasArora: Thanks, this is done! Looks much better.

MaanasArora:
Added a release note, I think this is ready for review now!

doc/source/reference/c-api/array.rst
seberg: Sounds OK either way, so whatever you like.

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -1720,36 +1755,68 @@ PyArray_Partition(PyArrayObject *op, PyArrayObject * ktharray, int axis,
...
-    part = get_partition_func(PyArray_TYPE(op), which);
-    if (part == NULL) {
-        /* Use sorting, slower but equivalent */
-        if ((PyDataType_GetArrFuncs(PyArray_DESCR(op))->compare)
-            && !(which & NPY_SELECT_DESCENDING)) { // TODO: descending sorts for partition
-            sort = npy_quicksort;
-        }
-        else if (which & NPY_SELECT_DESCENDING) {
+
+    method = NPY_DT_SLOTS(NPY_DTYPE(PyArray_DESCR(op)))->part_meth;
```
ngoldbaum, line 1765: I think `partition_prep_kth_array` has to happen before this and then a few lines below you should be using `PyArray_DESCR(kthval)`. Otherwise you miss a cast to intp in the metadata for the next `resolve_descriptors` call.

That was suggested by an AI model and it might be wrong but it seems plausible to me.

Ditto for the same pattern in argpartition.

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -1356,18 +1356,32 @@ _new_sortlike(PyArrayObject *op, int axis, PyArray_SortFunc *sort,
...
-            npy_intp pivots[NPY_MAX_PIVOT_STACK];
-            npy_intp npiv = 0;
-            npy_intp i;
-            for (i = 0; i < nkth; ++i) {
-                ret = part(bufptr, N, kth[i], pivots, &npiv, nkth, op);
+            if (strided_loop != NULL) {
+                char *const data[3] = {bufptr, (char *)kth, bufptr};
+                npy_intp strides[3] = {elsize, sizeof(npy_intp), elsize};
+                npy_intp dimensions[2] = {N, nkth};
+                ret = strided_loop(context, data, dimensions, strides, NULL);
```
ngoldbaum: The last argument is NULL but you have an auxdata in the arguments for the loop function. Shouldn't that get passed through?

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -1776,24 +1850,43 @@ PyArray_ArgPartition(PyArrayObject *op, PyArrayObject *ktharray, int axis,
...
-        if ((PyDataType_GetArrFuncs(PyArray_DESCR(op))->compare) &&
-            !(which & NPY_SELECT_DESCENDING)) { // TODO: descending sorts for partition
-            argsort = npy_aquicksort;
-        }
-        else if (which & NPY_SELECT_DESCENDING) {
+    method = NPY_DT_SLOTS(NPY_DTYPE(PyArray_DESCR(op)))->argpart_meth;
+    if (method != NULL) {
+        PyArray_Descr *descr = PyArray_DESCR(op);
+        PyArray_Descr *kdescr = PyArray_DESCR(ktharray);
+        PyArray_Descr *odescr = PyArray_DescrFromType(NPY_INTP);
```
ngoldbaum: Technically this could fail (e.g. OOM).

numpy/_core/src/multiarray/item_selection.c
ngoldbaum: This error path and other error paths below this don't clean up the resources you're newly allocating in the code path above (e.g. the auxdata, the `odescr`, maybe others).

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -1720,36 +1755,68 @@ PyArray_Partition(PyArrayObject *op, PyArrayObject * ktharray, int axis,
...
         }
     }
+    else {
+        part = get_partition_func(PyArray_TYPE(op), which);
+    }
 
     /* Process ktharray even if using sorting to do bounds checking */
     kthrvl = partition_prep_kth_array(ktharray, op, axis);
     if (kthrvl == NULL) {
         return -1;
```
ngoldbaum: shouldn't this be e.g. `goto fail` instead? Assuming the suggestion I gave above rearranging this doesn't change that.

numpy/_core/src/umath/_scaled_float_dtype.c
```diff
@@ -969,6 +970,127 @@ sfloat_argsort_resolve_descriptors(
...
+    Py_INCREF(loop_descrs[0]);
+
+    loop_descrs[1] = given_descrs[1];
+    Py_INCREF(loop_descrs[1]);
+
+    loop_descrs[2] = PyArray_DescrFromType(NPY_INTP);
+    if (loop_descrs[2] == NULL) {
+        return -1;
+    }
+    Py_INCREF(loop_descrs[2]);
```
ngoldbaum:
```suggestion
```

`PyArray_DescrFromType` already returns a new reference, right?

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -253,6 +259,16 @@ PyUFunc_AddLoopsFromSpecs(PyUFunc_LoopSlot *slots)
                 goto finish;
             }
         }
+        else if (ufunc == npy_interned_str.partition) {
+            if (set_static_method<&NPY_DType_Slots::part_meth, false>(slot->spec) < 0) {
```
ngoldbaum, line 308: Does it maybe make sense to do some basic validation here to avoid a runtime segfault later? Things like checking that the spec sets e.g. `nin` and `nout` correctly? Also for argpartition.

numpy/_core/src/multiarray/item_selection.c
MaanasArora: I _think_ this might be fine either way because we require that `kthval` is INTP, in other words, `resolve_descriptors` should always output INTP. If it doesn't, we don't handle it either way actually; without the ufunc machinery there is no cast to `loop_descrs` and the prep will always cast to INTP regardless of what the loop expects. (This is actually mirrored for the actual input as well, `given_descrs[0]`, which could technically be incorrectly different from `loop_descrs[0]`.)

Not quite sure which would be more correct! The only reason I can think of passing the original `kthval` is if we want to prepare the cast differently or error on wrong descriptor, which I might try soon, but I've just adopted your suggestion for now at least, thanks!

numpy/_core/src/multiarray/item_selection.c
MaanasArora: I'm sorry, I copied this directly for the sorts, from which it looks this is missing as well for the last release cycle! I've fixed this here and will make a backport PR for the sorts ASAP. Apologies for the oversight!

numpy/_core/src/umath/_scaled_float_dtype.c
MaanasArora: Yes, fixed, sorry!

numpy/_core/src/multiarray/item_selection.c
MaanasArora: I think rearranging with your suggestion above fixed this, thanks!

numpy/_core/src/multiarray/item_selection.c
MaanasArora: Same here.

numpy/_core/src/umath/dispatching.cpp
MaanasArora: I think so, thanks! I'll add some for the sorts as well.

numpy/_core/src/multiarray/item_selection.c
MaanasArora: Fixed, thanks!

seberg: It actually can't for builtin types, we use it without a check quite a lot for those. But either way is good.

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -1807,12 +1863,71 @@ PyArray_ArgPartition(PyArrayObject *op, PyArrayObject *ktharray, int axis,
         return NULL;
     }
 
-    ret = _new_argsortlike(op2, axis, argsort, NULL, NULL, NULL, NULL, argpart,
-                           PyArray_DATA(kthrvl), PyArray_SIZE(kthrvl));
+    method = NPY_DT_SLOTS(NPY_DTYPE(PyArray_DESCR(op)))->argpart_meth;
+    if (method != NULL) {
+        PyArray_Descr *descr = PyArray_DESCR(op);
+        PyArray_Descr *kdescr = PyArray_DESCR(ktharray);
+        PyArray_Descr *odescr = PyArray_DescrFromType(NPY_INTP);
```
ngoldbaum: I think you're still leaking `odescr` in all the error paths and the main path below.

numpy/_core/src/multiarray/item_selection.c
ngoldbaum: You can trigger an abort in a debug build of CPython with `np.partition(sfloat_arr, np.int32(1))` and it's because the issue I described above isn't completely fixed. You should be using `PyArray_DESCR(kthval)` instead of `ktharray`. In this code `ktharray` is the *user's* descriptor, not the int32 descriptor that numpy preps for sfloat data after casting.

The same issue also appears at line 1869 below.

numpy/_core/src/umath/dispatching.cpp
ngoldbaum: Did the validation I requested above get added? I don't see it here or elsewhere where it could be added to get some extra validation for this.

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -1720,36 +1755,69 @@ PyArray_Partition(PyArrayObject *op, PyArrayObject * ktharray, int axis,
...
+
+        PyArray_DTypeMeta *dtypes[3] = {dt, kdt, dt};
+        PyArray_Descr *given_descrs[3] = {descr, kdescr, descr};
+        // Partition cannot be a view, so view offset is unused
+        npy_intp view_offset = 0;
+
+        if (method->resolve_descriptors(
+            method, dtypes, given_descrs, loop_descrs, &view_offset) < 0) {
+            PyErr_SetString(PyExc_TypeError,
+                            "unable to resolve descriptors for partition");
```
ngoldbaum: Doesn't `resolve_descriptors` already set its own error? You'd be clobbering that one here. I actually think this can cause deeper issues because you don't explicitly reset the error state before doing this.

Ditto in other `resolve_descriptors` error paths in this PR.

numpy/_core/src/multiarray/item_selection.c
ngoldbaum: You should also do error checking for this function, which can error and return `NULL`.

numpy/_core/src/umath/_scaled_float_dtype.c
```diff
@@ -969,6 +970,126 @@ sfloat_argsort_resolve_descriptors(
...
+    
+    loop_descrs[0] = given_descrs[0];
+    Py_INCREF(loop_descrs[0]);
+
+    loop_descrs[1] = given_descrs[1];
+    Py_INCREF(loop_descrs[1]);
+
+    loop_descrs[2] = PyArray_DescrFromType(NPY_INTP);
+    if (loop_descrs[2] == NULL) {
+        return -1;
```
ngoldbaum: this leaks the increfs you did just above

numpy/_core/tests/test_multiarray.py
```diff
@@ -3118,6 +3126,11 @@ def _test_partition_descending(self, a, k, nan, descending):
         before.sort(descending=descending)
         after.sort(descending=descending)
 
+        part = np.partition(a, k, descending=descending)
```
ngoldbaum: copy/paste error? this isn't used.

doc/source/reference/c-api/array.rst
```diff
@@ -2171,6 +2171,36 @@ and not set any of the other loop slots.
...
+inplace, hence we enforce  that ``data[0] == data[2]``. Argpartitioning
+returns a new array of indices, so the output must be of ``NPY_INTP`` type.
+
+The ``context`` passed to the loop contains the ``parameters`` field which
+for these operations is a ``PyArrayMethod_PartitionParameters *`` struct. This
+struct contains a ``flags`` field which is a bitwise OR of ``NPY_SELECTKIND``
+values indicating the kind of partition to perform (that is, whether it is a
+descending partition). If the strided loop depends on the flags, a good way
+to deal with this is to define :c:macro:`NPY_METH_get_loop`, and not set any
+of the other loop slots.
```
ngoldbaum: Shouldn't this mension that `dimensions[0]` is `N` and `dimensions[1]` is `nkth`? It should probably also talk a little about multi-kth and mention that data[1] is a contiguous npy_intp array.

doc/source/reference/c-api/array.rst
```diff
@@ -2171,6 +2171,36 @@ and not set any of the other loop slots.
...
+Partitioning and Argpartitioning
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Similarly to sorting and argsorting, partitioning and argpartitioning methods
+can be registered using the ArrayMethod API. This is done by adding an
+ArrayMethod spec with the name ``"partition"`` or ``"argpartition"`` respectively.
+The spec must have ``nin=2`` and ``nout=1`` for both partition and argpartition,
+where the first input ``data[0]`` is the array to partition and the second input
+``data[1]`` is the kth array of indices to partition by. Partitioning is
+inplace, hence we enforce  that ``data[0] == data[2]``. Argpartitioning
```
ngoldbaum:
```suggestion
inplace, hence we enforce that ``data[0] == data[2]``. Argpartitioning
```

doc/release/upcoming_changes/31614.new_feature.rst
```diff
@@ -0,0 +1,10 @@
+DType partitioning and argpartitioning supports the ArrayMethod API
+-------------------------------------------------------------------
+User-defined dtypes can now implement custom partitioning and argpartitioning
+using the ArrayMethod API in a fashion similar to sorting and argsorting.
+These methods are used by `numpy.partition` and `numpy.argpartition` when called
+with arrays of the user-defined dtype.
+
+The partitioning and argpartitioning methods are registered by passing the
+arraymethod specs that implement the operations to the `PyUFunc_AddLoopsFromSpecs`
+function. See the ``ArrayMethod`` API documentation for details.
```
ngoldbaum: Maybe mention that descending sorts on `str`/`bytes`/`object` works now?

ngoldbaum (commented):
I think some of the issues I pointed out in my last review pass weren't fixed. I also spotted a couple other minor issues. Also total nit but you should probably also configure your editor so code you add doesn't add trailing whitespace (like line 1896 in item_selection.c and a few spots elsewhere).

Also just as a general comment, maybe it makes sense to pass in `kth`/`nkth` via the `PyArrayMethod_PartitionParameters` struct? That would keep `nin=1` and more closely mirror the structure of sort and argsort. I'll note that a lot of the bugs I'm pointing out in these reviews are related to that difference in structure so making things more uniform would also avoid those bugs.

numpy/_core/src/multiarray/item_selection.c
MaanasArora: I'm not sure, do you mean a Python error? I couldn't find one, but if a partition function is not found (i.e. NULL return), `PyArray_(Arg)Sort` is used.

numpy/_core/src/multiarray/item_selection.c
MaanasArora: Apologies, what I was saying doesn't really make sense for this PR - to be clearer I was imagining that we allow to pass the user's dtype to the array method, and since we assume the output is always INTP, we can safely rely on the internal cast, to produce more ufunc-like behavior. But that doesn't make sense as we are moving the casting apparatus / responsibility to `resolve_descriptors`, which would need a new `NPY_INTP` descr instantiation, removal of that assert, as well as more changes to casting returns (instead of `NPY_NO_CASTING`). If it is ever interesting this could be a refactor, but I don't think it makes sense anymore. I am now using `kthval`, sorry again.

numpy/_core/src/umath/dispatching.cpp
MaanasArora: Sorry, I forgot to push this - I have pushed checks for `nin, nout` and `dtypes` now including for sorts.

numpy/_core/src/multiarray/item_selection.c
MaanasArora: Sorry, yes, I removed these type errors, including from the sorts (assuming we wouldn't backport).

numpy/_core/src/umath/_scaled_float_dtype.c
MaanasArora: Fixed, thanks!

numpy/_core/tests/test_multiarray.py
MaanasArora: Yes, sorry, this is fixed now!

doc/source/reference/c-api/array.rst
MaanasArora: Thanks, this is added now!

doc/release/upcoming_changes/31614.new_feature.rst
MaanasArora: If you mean descending partitioning I have updated the previous release note (as it is also for 2.6) to remove the unsupported dtypes. If it's not good to update existing release notes I can revert and add a new `improvement` note.

For `object` descending sorting a note was included in #31431, strings already sorted descending.

MaanasArora:
Thank you for your thorough review, and sorry about the last sloppy iteration. I think I've addressed all of your comments now.

> Also just as a general comment, maybe it makes sense to pass in kth/nkth via the PyArrayMethod_PartitionParameters struct? That would keep nin=1 and more closely mirror the structure of sort and argsort. I'll note that a lot of the bugs I'm pointing out in these reviews are related to that difference in structure so making things more uniform would also avoid those bugs.

Makes sense, that would mirror the sort structure better indeed! I think there are benefits of keeping `kth` as an array argument (gufunc-like) though? It's an array, so seems more natural for the user API (i.e. no new abstraction). I also think it might prevent divergence, for example, `searchsorted` in the same space would probably take two arrays. And we might even get to use gufunc-machinery more, e.g. for casting or validation, but I might be wrong...

numpy/_core/src/umath/_scaled_float_dtype.c
seberg: FWIW, for builtin types `PyArray_DescrFromType(NPY_INTP);` can't fail ;).

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -1720,36 +1755,67 @@ PyArray_Partition(PyArrayObject *op, PyArrayObject * ktharray, int axis,
...
+    if (method != NULL) {
+        PyArray_Descr *descr = PyArray_DESCR(op);
+        PyArray_Descr *kdescr = PyArray_DESCR(kthrvl);
+        PyArray_DTypeMeta *dt = NPY_DTYPE(descr);
+        PyArray_DTypeMeta *kdt = NPY_DTYPE(kdescr);
+
+        PyArray_DTypeMeta *dtypes[3] = {dt, kdt, dt};
+        PyArray_Descr *given_descrs[3] = {descr, kdescr, descr};
+        // Partition cannot be a view, so view offset is unused
+        npy_intp view_offset = 0;
```
seberg:
```suggestion
        npy_intp view_offset = NPY_MIN_INTP;
```
You are right, we ignore it, but let's initialize it to "can't use" out of principle.

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -1720,36 +1755,67 @@ PyArray_Partition(PyArrayObject *op, PyArrayObject * ktharray, int axis,
...
+
+        if (method->resolve_descriptors(
+            method, dtypes, given_descrs, loop_descrs, &view_offset) < 0) {
```
seberg:
```suggestion
                method, dtypes, given_descrs, loop_descrs, &view_offset) < 0) {
```

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -1702,9 +1730,16 @@ NPY_NO_EXPORT int
...
+    PyArray_Descr *loop_descrs[3];
+    NpyAuxData *auxdata = NULL;
+    NPY_ARRAYMETHOD_FLAGS method_flags = 0;
+
     PyArrayObject *kthrvl;
-    PyArray_PartitionFunc *part;
-    PyArray_SortFunc *sort;
+    PyArray_PartitionFunc *part = NULL;
     int n = PyArray_NDIM(op);
     int ret;
```
seberg: NIT:
```suggestion
    int ret = -1;
```
Just feels a bit safer to me (and can remove it below).

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -1720,36 +1755,67 @@ PyArray_Partition(PyArrayObject *op, PyArrayObject * ktharray, int axis,
...
+
+        PyArray_DTypeMeta *dtypes[3] = {dt, kdt, dt};
+        PyArray_Descr *given_descrs[3] = {descr, kdescr, descr};
+        // Partition cannot be a view, so view offset is unused
+        npy_intp view_offset = 0;
+
+        if (method->resolve_descriptors(
+            method, dtypes, given_descrs, loop_descrs, &view_offset) < 0) {
+            Py_DECREF(kthrvl);
+            return -1;
```
seberg: A bit tricky that this can't be a `goto` for cleanup, but it looks right so fine to keep.

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -1807,12 +1862,71 @@ PyArray_ArgPartition(PyArrayObject *op, PyArrayObject *ktharray, int axis,
...
+        PyArray_Descr *given_descrs[3] = { descr, kdescr, odescr };
+        // Partition cannot be a view, so view offset is unused
+        npy_intp view_offset = 0;
+
+        if (method->resolve_descriptors(
+            method, dtypes, given_descrs, loop_descrs, &view_offset) < 0) {
+            Py_DECREF(kthrvl);
+            Py_DECREF(op2);
+            Py_DECREF(odescr);
+            return NULL;
```
seberg: Unlike the above, I guess this one may actually get nicer with a goto that is smart enough to work here.

numpy/_core/src/umath/_scaled_float_dtype.c
```diff
@@ -969,6 +970,128 @@ sfloat_argsort_resolve_descriptors(
...
+    assert(given_descrs[1]->type_num == NPY_INTP);
+
+    loop_descrs[0] = given_descrs[0];
+    Py_INCREF(loop_descrs[0]);
+
+    loop_descrs[1] = given_descrs[1];
+    Py_INCREF(loop_descrs[1]);
+
+    loop_descrs[2] = loop_descrs[0];
+    Py_INCREF(loop_descrs[2]);
```
seberg, line 1024: I guess for a downstream user, I would prefer if they write `loop_descrs[1] = PyArray_DescrFromType(NPY_INTP)` to basically say "I expect this".
(Which doesn't matter here at all though, just right general pattern for ufuncs.)


(But I suppose most user dtypes aren't parametric so the default will just kick in either way.)

numpy/_core/src/umath/dispatching.cpp
```diff
@@ -244,15 +250,89 @@ PyUFunc_AddLoopsFromSpecs(PyUFunc_LoopSlot *slots)
...
+                goto finish;
+            }
+        }
+        else if (ufunc == npy_interned_str.argpartition) {
+            if (slot->spec->nin != 2 || slot->spec->nout != 1) {
+                PyErr_Format(PyExc_ValueError,
+                        "Argpartition method spec must have nin=2 and nout=1, got %d and %d",
+                        slot->spec->nin, slot->spec->nout);
+                goto finish;
```
seberg, line 317: NIT: At least this one can be moved into `set_static_method`, I think.  But I would be tempted to see if we can't wrangle the rest there as well to make it shorter.
(If the second part doesn't work, I admit I am not sure the first part is worth it.)

numpy/_core/src/multiarray/item_selection.c
seberg: Yeah, this is already correct as is.  The function is internal and can't fail it returns `NULL` for "not defined".

seberg (commented):
Thanks, I think this looks pretty much ready and if another maintainer things it's good and wants to merge, don't wait for me.

I made a few small comments, but just small things, I haven't checked the tests super carefully but it looks all fine and with the `SFloat` implementation we cover the new paths well.

(NOTE: I did a human only pass, I would probably do an AI pass for correctness once more, but as Nathan already did that, I doubt it would spot significant things right now.)

numpy/_core/src/umath/dispatching.cpp
MaanasArora: Hmm, I don't think the second part would work without a switch case against the method, so not sure if that is worth it, or the first part alone is worth it...

numpy/_core/src/umath/_scaled_float_dtype.c
MaanasArora: I did this actually, just to keep it clean for reference, thanks! Removed the error fallbacks as well, since these can't error, and three of those would be annoying :)

numpy/_core/src/multiarray/item_selection.c
MaanasArora: I moved this and also below to a `goto` that seems smarter (just checking the descriptors instead of the context itself), it seems fine, and cleaner if it is, thanks!

numpy/_core/src/multiarray/item_selection.c
MaanasArora: Done!

numpy/_core/src/multiarray/item_selection.c
MaanasArora: Multiple indents like this fixed.

numpy/_core/src/multiarray/item_selection.c
MaanasArora: Done, thanks!

seberg (approved):
Pushed a silly change, but looks good to me, thanks.

@ngoldbaum I'll probably just merge this very soon (I know you are busy, so if you see this later, don't hesitate to point out follow-ups you may like.)

numpy/_core/src/multiarray/item_selection.c
```diff
@@ -1720,36 +1755,65 @@ PyArray_Partition(PyArrayObject *op, PyArrayObject * ktharray, int axis,
...
+    if (method != NULL) {
+        PyArray_Descr *descr = PyArray_DESCR(op);
+        PyArray_Descr *kdescr = PyArray_DESCR(kthrvl);
+        PyArray_DTypeMeta *dt = NPY_DTYPE(descr);
+        PyArray_DTypeMeta *kdt = NPY_DTYPE(kdescr);
+
+        PyArray_DTypeMeta *dtypes[3] = {dt, kdt, dt};
+        PyArray_Descr *given_descrs[3] = {descr, kdescr, descr};
+        // Partition cannot be a view, so view offset is unused
+        npy_intp view_offset = NPY_MIN_INT;
```
ngoldbaum: shouldn't this be `MIN_INTP`? Also on line 1880.

ngoldbaum (approved):

numpy/_core/src/multiarray/item_selection.c
MaanasArora: Yes thanks, this is fixed.

ngoldbaum:
Thanks @MaanasArora

PR #31615, opened 2026-06-12
BUG: fix incorrect alignment handling and error handling in several code paths

ngoldbaum:
Fixes two error paths that are trivially incorrect for OOM scenarios:

* `arr_place` uses `PyArray_ResolveWritebackIfCopy` in an error path, but it probably makes more sense to use `PyArray_DiscardWritebackIfCopy`, as suggested in the docstring for that function: https://numpy.org/doc/2.4/reference/c-api/array.html#c.PyArray_DiscardWritebackIfCopy
* `PyArray_PutTo` and `PyArray_PutMask` fail to check if `PyArray_FromArray` fails.

Fixes an incorrect assumption in the flatiter assignment code - the result array might be a user-supplied unaligned array.

Fixes a logic error in boolean indexing, where if an error is raised during an iteration, the iteration is not halted and it's possible the error could be silently discarded.

Adds tests for the latter two cases.

I used Claude Fable to help out with this. This was spotted during work on #31609 and touches the same functions but the issues are unrelated to StringDType, so I'm issuing this PR separately.

### DIFF
--- a/numpy/_core/src/multiarray/item_selection.c
+++ b/numpy/_core/src/multiarray/item_selection.c
@@ -439,6 +439,9 @@ PyArray_PutTo(PyArrayObject *self, PyObject* values0, PyObject *indices0,
         Py_INCREF(PyArray_DESCR(self));
         obj = (PyArrayObject *)PyArray_FromArray(self,
                                                  PyArray_DESCR(self), flags);
+        if (obj == NULL) {
+            goto fail;
+        }
         copied = 1;
         assert(self != obj);
         self = obj;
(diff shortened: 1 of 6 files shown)


numpy/_core/src/multiarray/item_selection.c
```diff
@@ -439,6 +439,9 @@ PyArray_PutTo(PyArrayObject *self, PyObject* values0, PyObject *indices0,
         Py_INCREF(PyArray_DESCR(self));
         obj = (PyArrayObject *)PyArray_FromArray(self,
                                                  PyArray_DESCR(self), flags);
+        if (obj == NULL) {
+            goto fail;
+        }
```
seberg, line 444: ufff...

seberg (approved):
Thanks, looks all like straight forward and good fixes.  Ouch on some of them...
I wonder if the `PyArray_DiscardWritebackIfCopy` might be wrong elsewhere but nvm.

Tests look about as good as they can be nicely (e.g. subscript can in practice never fail here).

PR #31620, opened 2026-06-13
BUG: fix linspace returning NaN for equal infinite endpoints (gh-26699)

### LINKED ISSUE #26699: linspace(inf, inf, N) outputs NaN
### Describe the issue:

I would expect `linspace(X, X, N)` to produce an array with N copies of X. This works for finite values, but not for ±inf. This is an edge case that probably doesn't affect many people, but I don't think there is any ambiguity about the expected output.

```python
import numpy as np
np.linspace(np.inf, np.inf, 4)
```

### Error message:

```shell
/tmp/np/lib/python3.12/site-packages/numpy/_core/function_base.py:145: RuntimeWarning: invalid value encountered in subtract
  delta = np.subtract(stop, start, dtype=type(dt))
array([nan, nan, nan, inf])
```

2.0.0rc2
3.12.3 (main, Apr 10 2024, 05:33:47) [GCC 13.2.0]

### Runtime Environment:

[{'numpy_version': '2.0.0rc2',
  'python': '3.12.3 (main, Apr 10 2024, 05:33:47) [GCC 13.2.0]',
  'uname': uname_result(system='Linux', node='hippo', release='6.7.12-amd64', version='#1 SMP PREEMPT_DYNAMIC Debian 6.7.12-1 (2024-04-24)', machine='x86_64')},
                      'found': ['SSSE3',
                                'SSE41',
                                'POPCNT',
                                'SSE42',
                                'AVX',
                                'F16C',
                                'FMA3',
                                'AVX2'],
                      'not_found': ['AVX512F',
                                    'AVX512CD',
                                    'AVX512_KNL',
                                    'AVX512_KNM',
                                    'AVX512_SKX',
                                    'AVX512_CLX',
                                    'AVX512_CNL',
                                    'AVX512_ICL']}},
 {'architecture': 'Haswell',
  'filepath': '/tmp/np/lib/python3.12/site-packages/numpy.libs/libscipy_openblas64_-99b71e71.so',
  'internal_api': 'openblas',
  'num_threads': 16,
  'user_api': 'blas',
  'version': '0.3.27'}]

### Context for the issue:

_No response_

Ijtihed:
### PR summary

Fixes gh-26699.
  
np.linspace(np.inf, np.inf, N) returned [nan nan ... nan inf] instead of [inf inf ... inf]. The root cause is delta = stop - start producing nan via inf - inf which propagates through all subsequent multiplications,only the last element was rescued by the y[-1] = stop assignment

Fix: replace any nan in delta with zero wherever start == stop. The nan == nan is False identity ensures NaN inputs are unaffected in this case

#### AI Disclosure

Used Claude to make sure I didn't miss anything

### DIFF
--- a/numpy/_core/function_base.py
+++ b/numpy/_core/function_base.py
@@ -137,8 +137,19 @@ def linspace(start, stop, num=50, endpoint=True, retstep=False, dtype=None,
     else:
         integer_dtype = _nx.issubdtype(dtype, _nx.integer)
 
-    # Use `dtype=type(dt)` to enforce a floating point evaluation:
-    delta = np.subtract(stop, start, dtype=type(dt))
+    # Use `dtype=type(dt)` to enforce a floating point evaluation.
+    # Equal endpoints (including equal infinities) must produce a zero step,
+    # so skip the subtraction there: `inf - inf` would otherwise yield a
+    # spurious nan and an "invalid value" warning.  Using `where=` avoids the
+    # bad element entirely rather than suppressing warnings globally, so
+    # genuine invalid operations (e.g. mixed infinities) still warn.
+    equal = start == stop
+    if equal.ndim == 0:
+        delta = dt.type(0) if equal else np.subtract(
+            stop, start, dtype=type(dt))
+    else:
+        delta = np.subtract(stop, start, dtype=type(dt), where=~equal,
+                            out=np.zeros(equal.shape, dtype=type(dt)))
     y = _nx.arange(
         0, num, dtype=dt, device=device
     ).reshape((-1,) + (1,) * ndim(delta))
--- a/doc/release/upcoming_changes/31620.improvement.rst
+++ b/doc/release/upcoming_changes/31620.improvement.rst
@@ -1,0 +1,4 @@
+``np.linspace`` no longer returns NaN for equal infinite endpoints
+------------------------------------------------------------------
+``np.linspace(np.inf, np.inf, N)`` (and ``-np.inf``) now correctly returns
+an array filled with ``inf`` instead of ``nan``.
--- a/numpy/_core/tests/test_function_base.py
+++ b/numpy/_core/tests/test_function_base.py
@@ -1,5 +1,6 @@
 import platform
 import sys
+import warnings
 
 import pytest
 
@@ -21,6 +22,7 @@ from numpy import (
 from numpy._core import sctypes
 from numpy._core.function_base import add_newdoc
 from numpy.testing import (
+    IS_WASM,
     assert_,
     assert_allclose,
     assert_array_equal,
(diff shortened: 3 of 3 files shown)


numpy/_core/function_base.py
```diff
@@ -138,39 +138,43 @@ def linspace(start, stop, num=50, endpoint=True, retstep=False, dtype=None,
...
-        _mult_inplace = _nx.isscalar(delta)
-        step = delta / div
-        any_step_zero = (
-            step == 0 if _mult_inplace else _nx.asanyarray(step == 0).any())
-        if any_step_zero:
-            # Special handling for denormal numbers, gh-5437
-            y /= div
-            if _mult_inplace:
-                y *= delta
+    with np.errstate(invalid='ignore'):
```
MaanasArora: I think the `RuntimeWarning` which happens here with `start=stop=inf` is real, so maybe this isn't needed (even if we fix the actual output), or we should make it more focused to the error. In any case, we don't want to ignore errors for _all_ the lines below!

MaanasArora (commented):
Thanks! This solves the specific issue, but I'm wondering about the performance cost where this new code is not invoked. I'm not sure if the corner case is bad enough to need this (and a `lerp` ufunc would be much better as stated in the issue!)

numpy/_core/function_base.py
Ijtihed: any(isnan) short-circuits on the first element so for the common finite-scalar path the overhead is negligible. I agree a lerp ufunc is the right long-term fix but this is the minimum correct approach at the py level for now

errstate now wraps only the np.subtract call and all downstream arithmetic is back to normal, whoops

numpy/_core/function_base.py
```diff
@@ -152,25 +155,30 @@ def linspace(start, stop, num=50, endpoint=True, retstep=False, dtype=None,
...
         any_step_zero = (
             step == 0 if _mult_inplace else _nx.asanyarray(step == 0).any())
-        if any_step_zero:
-            # Special handling for denormal numbers, gh-5437
-            y /= div
-            if _mult_inplace:
-                y *= delta
+        # Use errstate to suppress warnings from 0*inf or 0*nan that can arise
+        # when step or delta is infinite (e.g. linspace(inf, -inf, N)).
+        with np.errstate(invalid='ignore'):
```
seberg: Fly by comment: but what is this there for?  NaN input don't produce new warnings and I don't really see where `inf` should originate from here (sensibly).

Ijtihed: just got  home and went over the code again

`0 * nan` in the comment was wrong since I just found out NaNs propagate silently under IEEE 754 and do not raise an invalid signal. The `errstate` is still needed for `0 * inf`: for example, `linspace(inf, -inf, N)` gives `delta = -inf - inf = -inf`, so `step = -inf`, and `y[0] * step` becomes `0 * (-inf)`, which raises the invalid signal. I updated the comment accordingly.

Ijtihed:
I noticed that the existing test for `linspace(inf, -inf)` was nearly a no-op, that `-inf` to `inf` had no test coverage, and that NaN input behavior was undocumented. I fixed all 3 in latest push.

edit: also added complex infinity tests

numpy/_core/function_base.py
seberg: Yeah... But `linspace(inf, -inf)` actually creates NaN values and _should_ warn.

(I guess `np.linspace(inf, -inf)` might warn more than once which is annoying but dunno if it's worth the trouble, does it?  -- it may also not since once you got a NaN, you don't get more warnings.)

Ijtihed: yeah you're right again. I removed all the extra errstate wrappers so the warnings from multiply and add propagate normally now

Ijtihed:
removed the extra errstate wrappers, linspace(inf, -inf, N) now warns from the multiply and add as it should. also dropped the isnan guard since where(start == stop, 0, delta) already returns delta unchanged when there's no NaN to fix because ig the check was redundant

numpy/_core/function_base.py
```diff
@@ -138,7 +138,9 @@ def linspace(start, stop, num=50, endpoint=True, retstep=False, dtype=None,
         integer_dtype = _nx.issubdtype(dtype, _nx.integer)
 
     # Use `dtype=type(dt)` to enforce a floating point evaluation:
-    delta = np.subtract(stop, start, dtype=type(dt))
+    with np.errstate(invalid='ignore'):
+        delta = np.subtract(stop, start, dtype=type(dt))
+        delta = _nx.where(start == stop, _nx.zeros_like(delta), delta)
```
MaanasArora: I see you removed the `isnan` check, might as well go without it? Because for the common finite scalar case, we don't get two calls, and even for arrays isnan actually penalizes the no-nan case more than the nan case, I think.

Also, `_nx.where` goes through every element, maybe we could use `delta[start == stop]` or similar?

Edit: my AI agent suggested guarding with:

```python
eq = (start == stop)
if _nx.any(eq):
```

I think this might make sense given this is an uncommon path!

Ijtihed: I dropped _nx.where and zeros_like. Scalar path uses type(delta)(0), array path uses delta[start == stop] = 0 in-place. i allso split the mixed-endpoint warns check into its own @pytest.mark.skipif(IS_WASM, ...) method to fix the Pyodide CI failure

Ijtihed: > Edit: my AI agent suggested guarding with:
> 
> ```python
> eq = (start == stop)
> if _nx.any(eq):
> ```
> 
> I think this might make sense given this is an uncommon path!

I believe both paths end up O(n) on the same boolean array so _nx.any() just swaps the indexed-assignment scan for a reduction scan. what do you think? Is it worth adding the lines anyhow?

MaanasArora: > I believe both paths end up O(n) on the same boolean array so _nx.any()

Ah sorry, you're right, I was confused. Thanks for the changes.

Ijtihed: Alright cool. Thx! I think this is ready now

Ijtihed:
@MaanasArora :)

MaanasArora:
Thanks, this seems to have the least performance overhead now, but I'll still defer to someone more familiar with this area to evaluate if this should go in (not very familiar with this code, sorry!)

numpy/_core/function_base.py
mattip: This seems like an antipattern to me. I am not sure it is threadsafe. So far we only use `with errstate` in helper functions like `isclose` and printing, I am not sure we want to start using it in core functions. Can we do the more specific `if np.isinf()` instead?

Ijtihed: > This seems like an antipattern to me. I am not sure it is threadsafe. So far we only use `with errstate` in helper functions like `isclose` and printing, I am not sure we want to start using it in core functions. Can we do the more specific `if np.isinf()` instead?

thank you! I changed it accordingly

mattip:
Thanks @Ijtihed

PR #31624, opened 2026-06-13
BUG: Fix thread safety issue in f2py allocatable arrays

da-woods:
### PR summary

There were two separate issues:
1. The target data was being passed as a global variable. This means that even assigning individual attributes from different threads is unsafe. Fixed this by passing it through the generated Fortran function call instead.
2. Second (less important), use critical sections to guard the parts where PyFortranObject internals are modified, so that assigning the same attribute from different threads won't corrupt the structure.

I haven't tried to guard anything except the internals of f2py for thread-safety. There's no real point trying to lock the contents of these variables since Fortran itself can access them without the lock.

Also add some tests for allocatable arrays (although not specifically a thread safety test).

This will generate some small merge conflicts with https://github.com/numpy/numpy/pull/31598 but nothing difficult to fix.

#### AI Disclosure

No AI tools used.

### DIFF
--- a/numpy/f2py/src/fortranobject.c
+++ b/numpy/f2py/src/fortranobject.c
@@ -291,21 +291,44 @@ fortran_doc(FortranDataDef def)
     }
 }
 
-static FortranDataDef *save_def; /* save pointer of an allocatable array */
 static void
-set_data(char *d, npy_intp *f)
+set_data(char *d, npy_intp *f, char **out_data)
 {           /* callback from Fortran */
     if (*f) /* In fortran f=allocated(d) */
-        save_def->data = d;
+        *out_data = d;
     else
-        save_def->data = NULL;
+        *out_data = NULL;
     /* printf("set_data: d=%p,f=%d\n",d,*f); */
+}
+
+static PyObject *
+fortran_getattr_allocatable(PyFortranObject *fp, FortranDataDef *fp_defs_i) {
+    int k, flag;
+    for (k = 0; k < fp_defs_i->rank; ++k) fp_defs_i->dims.d[k] = -1;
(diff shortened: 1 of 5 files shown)


numpy/f2py/src/fortranobject.c
```diff
@@ -291,21 +291,44 @@ fortran_doc(FortranDataDef def)
...
     else
-        save_def->data = NULL;
+        *out_data = NULL;
     /* printf("set_data: d=%p,f=%d\n",d,*f); */
 }
 
+static PyObject *
+fortran_getattr_allocatable(PyFortranObject *fp, FortranDataDef *fp_defs_i) {
+    int k, flag;
+    for (k = 0; k < fp_defs_i->rank; ++k) fp_defs_i->dims.d[k] = -1;
```
da-woods, line 307: Just moved from `fortran_getattr` with indentation changed, `fp->defs[i].` replaced with `fp_defs_i->` and a commented out `Py_INCREF` line removed.

numpy/f2py/src/fortranobject.c
```diff
@@ -397,10 +408,70 @@ fortran_getattr(PyFortranObject *fp, char *name)
...
 static int
-fortran_setattr(PyFortranObject *fp, char *name, PyObject *v)
+fortran_setattr_array(FortranDataDef *fp_defs_i, PyObject *v)
 {
-    int i, j, flag;
+    int k, flag;
+    npy_intp dims[F2PY_MAX_DIMS];
     PyArrayObject *arr = NULL;
+
+    if (fp_defs_i->func != NULL) { /* is allocatable array */
```
da-woods, line 417: Again, this is all a pretty straightforward move

PR #31626, opened 2026-06-13
DOC: Prevent arrow ligatures in generalized ufunc signatures fix #31625

### LINKED ISSUE #31625: BUG: Generalized ufunc docs render `->` in confusing way
### Describe the issue:

It's possible that this may related to fonts I have installed or something like that. I'm not completely sure, but it's possible that others won't experience this.

On the development docs (but not the stable docs), the arrows render as full arrows for me:

Image

For contrast, here's the same image on the stable docs

Image

They appear to copy-paste as separate `-` and `>` characters, but it's a little misleading because it implies that you need a special unicode character.

Bug in docs rendering so not applicable.

Go to https://numpy.org/doc/stable/reference/c-api/generalized-ufuncs.html and https://numpy.org/devdocs/reference/c-api/generalized-ufuncs.html in a browser.

### Error message:

```shell
Not applicable
```

Not applicable.

### Runtime Environment:

Firefox and Windows

_No response_

-----

Edit: In hindsight I should have used the "documentation" template

nishidhx:
DOC: Prevent arrow ligatures in generalized ufunc signatures

Bug: #31625 

Disable font ligatures for inline code literals so signatures
such as ()->() render literally instead of displaying a Unicode
arrow glyph in browsers using ligature-enabled fonts.

Investigatation: 
```js
$0.textContent
// '(),()->()'
```

image

- The generated HTML contains the literal text ()->() ($0.textContent returns (),()->()). The issue appears to be visual only. 
- The code element uses a font stack including Cascadia Code and has font-variant-ligatures: normal, which causes it to render -> as the ligature →. 
- Disabling ligatures on the code element restores the expected appearance.

Before: 
image

After:
image

### DIFF
--- a/doc/source/_static/numpy.css
+++ b/doc/source/_static/numpy.css
@@ -104,3 +104,8 @@ div.admonition-legacy>.admonition-title {
 .try_examples_outer_iframe {
   margin-top: 0.4em;
 }
+
+code.literal,
+code.literal .pre {
+    font-variant-ligatures: none;
+}


charris:
Thanks @nishidhx .

PR #31623, opened 2026-06-13
TST: Use pytest.warns as ignore filters+suppress_warnings interfere

seberg:
I am not sure why, it it seems that maybe when used multiple times (and actually capturing?) `suppress_warnings` somehow disrupts the pytest "ignore" filter *only* on free-threaded builds.
(That part is maybe not so weird, since free-threaded builds do more things to make warning filters thread-safe I think.)

I would think it has something to do with messing with `__warning_registery__`, i.e. after the first captured warning we clear it or so and then the second one pytest is in a bad state.

(I feel it's a waste of time to truly understand, as this should be fine)

Either way, it seems just asserting the warning is enough to fix it. One of the tests didn't need it but I added it for symmetry. I omitted `TestWarns`, for some reason it seems unaffected (it only uses `assert_warns()` once per test with actual capture.

**Large diff, but just whitespace, because `pytest.warns` isn't a decorator.**

CC @charris and @ngoldbaum alternative to Chuck's PR that doesn't quite skip them fully but seems to work fine to hide these warnings on free-threaded.

(No AI)

PR #31628, opened 2026-06-13
BUG: fix suppress_warnings when context-aware warnings is on

ngoldbaum:
Followup for https://github.com/numpy/numpy/pull/31623, fixing the root cause for the spurious warnings.

Ping @nascheme who wrote the new [context-aware warning](https://docs.python.org/3/library/warnings.html#concurrent-safety-of-context-managers) implementation in CPython and may want to see a case where that breaks assumptions in ecosystem code.

Note that this function is deprecated but IMO it's worth it to fix this anyway.

### DIFF
--- a/numpy/testing/_private/utils.py
+++ b/numpy/testing/_private/utils.py
@@ -2478,8 +2478,8 @@ class suppress_warnings:
             raise RuntimeError("cannot enter suppress_warnings twice.")
 
         self._orig_show = warnings.showwarning
-        self._filters = warnings.filters
-        warnings.filters = self._filters[:]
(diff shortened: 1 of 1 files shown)


seberg (approved):
OK, sure, makes sense, we could even delete `clear_registries` completely now (which was the only reason for this whole context).

Thanks, let's just put this in I guess, this copied `warning.catch_warnings` and that had to change (which is part of why we deprecated it, the other being that it guarded against things long unnecessary).

(While I think that we could delete others, `catch_warnings` will restore `filters` just fine, so this is safe on any Python version.)

PR #31629, opened 2026-06-13
TYP: Fix ``ma.[as[any]]array`` fallback overloads (#31482)

### LINKED ISSUE #31451: TYP: unexpected `var-annotated` error with masked array

charris:
Backport of #31482.

### PR summary

Fixes #31451

#### AI Disclosure

N/A

### DIFF
--- a/numpy/ma/core.pyi
+++ b/numpy/ma/core.pyi
@@ -2880,7 +2880,7 @@ def array[ScalarT: np.generic](
     ndmin: int = 0,
 ) -> _MaskedArray[ScalarT]: ...
 @overload
-def array[ScalarT: np.generic](
+def array(
     data: object,
     dtype: DTypeLike | None = None,
     copy: bool = False,
@@ -2892,7 +2892,7 @@ def array[ScalarT: np.generic](
     shrink: bool = True,
     subok: bool = True,
     ndmin: int = 0,
-) -> _MaskedArray[ScalarT]: ...
+) -> _MaskedArray[Any]: ...
 
 # keep in sync with `array`
 @overload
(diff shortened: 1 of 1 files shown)


jorenham:
Windows CI failure seems unrelated:

```console
$ 'D:\a\numpy\numpy\.venv\Scripts\python.exe' -P -m pytest 'numpy\typing'
ImportError while loading conftest 'D:\a\numpy\numpy\build-install\Lib\site-packages\numpy\conftest.py'.
numpy\__init__.py:109: in <module>
    from numpy.__config__ import show_config
numpy\__config__.py:4: in <module>
    from numpy._core._multiarray_umath import (
numpy\_core\__init__.py:24: in <module>
    from . import multiarray
numpy\_core\multiarray.py:11: in <module>
    from . import _multiarray_umath, overrides
E   RuntimeWarning: overflow encountered in cast
```

charris:
> Windows CI failure seems unrelated:

CI seems to be having a breakdown.

charris:
I think it is today's release of pytest 9.1.0. Lots of
```
ERROR numpy/_core/tests/test_array_coercion.py::TestStringDiscovery - pytest.PytestRemovedIn10Warning: Passing a non-Collection iterable to param...
```

charris:
I opened a pytest issue: https://github.com/pytest-dev/pytest/issues/14592

charris:
The problem on the numpy side is:
```
uv pip install -r /tmp/build_requirements.txt pytest pytest-xdist hypothesis pytest-timeout
```
Should use requirements where pytest is pinned.

PR #31636, opened 2026-06-14
MAINT, TST: Pin pytest to 9.0.3

charris:
The recent release of pytest 9.1.0 broke some tests where it was not pinned. This pins it in the maintenance/2.5.x branch until it gets fixed in main.

No AI used.


charris:
@jorenham I'm thinking the mypy problem may be due to a visual studio update. The latest version was released May 27, 2026.

seberg (approved):
Sounds fine, I guess we could do this in main as well.  Will install a few more things but I don't think that matters.

FWIW, the new deprecation warning looks like just forcing us to add a `list()` around usages...

charris:
>  I guess we could do this in main as well

Probably should just to avoid these surprises. There was a reason pytest was pinned :)

PR #31640, opened 2026-06-14
MAINT: Clean up use for PyObject_* APIs when PyMem_* would do

seberg:
We were historically a bit sloppy using PyObject_* allocation API for many non-object cases.
This cleans up the rest of those completely, I did change one RawMalloc to not be raw, because the GIL is held and we shouldn't use `RawMalloc` either for this type of allocation.

So this is just a small general improvement (I dunno how bad it is to use the other for non-objects, but I think the Python docs could be clearer about it if it is now on free-threaded).

The original motivation is weird and maybe doesn't matter: The 3.15b2 free-threaded Ubuntu 22.04 builds of the setup-python action fail with segfault for ``PyObject_Realloc(0, ...)``. On the build `PyMem_Realloc()` works fine, so it "solves" the issue although it may be a real issue with the build.

(Of course I am not sure that this is a build issue or a Python issue, I had opened a setup-python issue though.)

(No AI use, just double checked with it I didn't miss anything)

---

I am tempted to backport, just in case the setup-python issue isn't a one-off and because it is trivial.

CC @ngoldbaum in case you are interested about the crash issue.  I suppose @kumaraditya303 could have a brief look to review.

### DIFF
--- a/numpy/_core/src/multiarray/buffer.c
+++ b/numpy/_core/src/multiarray/buffer.c
@@ -66,7 +66,7 @@ _append_char(_tmp_string_t *s, char c)
         char *p;
         size_t to_alloc = (s->allocated == 0) ? INIT_SIZE : (2 * s->allocated);
 
-        p = PyObject_Realloc(s->s, to_alloc);
+        p = PyMem_Realloc(s->s, to_alloc);
         if (p == NULL) {
             PyErr_SetString(PyExc_MemoryError, "memory allocation failed");
             return -1;
@@ -473,7 +473,7 @@ _buffer_info_new(PyObject *obj, int flags)
     int err = 0;
 
     if (PyArray_IsScalar(obj, Void)) {
-        info = PyObject_Malloc(sizeof(_buffer_info_t));
+        info = PyMem_Malloc(sizeof(_buffer_info_t));
         if (info == NULL) {
             PyErr_NoMemory();
             goto fail;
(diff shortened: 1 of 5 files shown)


numpy/_core/src/multiarray/buffer.c
```diff
@@ -66,7 +66,7 @@ _append_char(_tmp_string_t *s, char c)
         char *p;
         size_t to_alloc = (s->allocated == 0) ? INIT_SIZE : (2 * s->allocated);
 
-        p = PyObject_Realloc(s->s, to_alloc);
+        p = PyMem_Realloc(s->s, to_alloc);
```
seberg, line 69: This change (probably) would make numpy work on that broken(?) setup-python build...

kumaraditya303:
> The 3.15b2 free-threaded Ubuntu 22.04 builds of the setup-python action fail with segfault for PyObject_Realloc(0, ...). On the build PyMem_Realloc() works fine, so it "solves" the issue although it may be a real issue with the build.

I think that crash might be because of https://github.com/python/cpython/issues/151297

seberg:
Yap, that must be it, then nothing wrong with setup-python build, they just got unlucky.

ngoldbaum:
I just merged with `main` to get a fresh test run but I think this is all correct changes so I'm going to merge it regardless of whether this was to work around an upstream bug.

ngoldbaum:
Thanks @seberg!

seberg:
It's all fixed up so we can backport but also really don't need to.

PR #31642, opened 2026-06-14
TST, TYP: Pin to windows-2022 for typecheck on windows.

charris:
The mypy problems on windows-latest follow from the upgrade to windows-2025. It may get fixed upstream at some point but for now pin to windows-2022.

#### AI Disclosure

AI suggested the fix.

### DIFF
--- a/.github/workflows/typecheck.yml
+++ b/.github/workflows/typecheck.yml
@@ -22,7 +22,7 @@ on:
       - '**.rst'
       - '.circlecl/**'
       - '.devcontainer/**'
-      - '.spin/**'
+      - '.spin/LICENSE'
       - 'benchmarks/**'
       - 'branding/**'
       - 'doc/**'
@@ -54,7 +54,7 @@ jobs:
         os_python:
           - [macos-latest, '3.14']
           - [ubuntu-latest, '3.13']
-          - [windows-latest, '3.12']
+          - [windows-2022, '3.12']
     steps:
     - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
       with:


jorenham:
I wonder which typing test actually caused this warning. Because it might be easier/cleaner to prevent it from occurring in the first place

jorenham:
I'm at a backgammon tournament right now, but I could take a look once I'm home if you want

charris:
@jorenham It looks like the typing tests didn't run. Is there something in the tags that suppressed it?

jorenham:
There's a path trigger in the workflow; that's probably it

jorenham:
> I'm at a backgammon tournament right now, but I could take a look once I'm home if you want

Hm yea I don't think this is caused by any specific test, or anything specific to typing for that matter. It looks more like it's a MSVC thing, so suppressing this warning is probably indeed the way to go here

charris:
typecheck ignores the spin path, mind if I change that? `cmds.py` is a python file.

jorenham:
> typecheck ignores the spin path, mind if I change that? `cmds.py` is a python file.

Oh yea of course go ahead

charris:
Different error:
```
numpy\typing\tests\test_runtime.py:89: in <module>
    "_SupportsArray": (_npt._SupportsArray, np.arange(10)),
                                            ^^^^^^^^^^^^^
E   ValueError: arange: cannot compute length
        Any        = typing.Any
        NamedTuple = <function NamedTuple at 0x000001CA11391300>
        Self       = typing.Self
        TYPES      = {'ArrayLike': TypeTup(typ=collections.abc.Buffer | _DualArrayLike[numpy.dtype, complex | bytes | str], args=(<class 'c...umpy.ndarray[_AnyShape, numpy.dtype[ScalarT]], args=(_AnyShape, numpy.dtype[ScalarT]), origin=<class 'numpy.ndarray'>)}
```
Not sure what that is about.

jorenham:
According to Claude "it's the same broken-floating-point problem on that Windows runner, just showing a second face":

> ## What that error is about
> 
> **It's not a bug in `test_runtime.py` — it's the same broken-floating-point problem on that Windows runner, just showing a second face.**
> 
> The lines at [test_runtime.py:89-90](numpy/typing/tests/test_runtime.py#L88-L92) (`np.arange(10)` at module scope) are old and unchanged on this branch — they only started failing because your `.spin/cmds.py` path-trigger fix made the typing tests actually run again on the updated Windows CI image.
> 
> ### Why `np.arange(10)` of all things
> 
> `np.arange` computes its output length in floating point. The path is:
> 
> `PyArray_ArangeObj` → [`_calc_length`](numpy/_core/src/multiarray/ctors.c#L3180) → [`_arange_safe_ceil_to_intp`](numpy/_core/src/multiarray/ctors.c#L3068):
> 
> ```c
> ivalue = npy_ceil(value);          // value = (stop - start) / step
> if (npy_isnan(ivalue)) {
>     PyErr_SetString(PyExc_ValueError, "arange: cannot compute length");
> ```
> 
> For `np.arange(10)` that's `ceil((10 - 0) / 1)` = `ceil(10.0)`. The `"cannot compute length"` ValueError fires **only when `npy_ceil` returns NaN**, i.e. when `value` is NaN. For a trivial `10.0 / 1.0` to come out NaN, the hardware FP unit itself has to be producing wrong results. So this is a genuine C-level exception caused by a misbehaving FP environment, not a Python/test issue.
> 
> ### Why it's the same root cause as the original warning
> 
> The PR is suppressing `RuntimeWarning: overflow encountered in cast`. That message is generated by numpy's FP-status-flag reporter in [extobj.c:407](numpy/_core/src/umath/extobj.c#L407) (`"%s encountered in %s"` → `"overflow"` in `"cast"`) — it fires when a cast operation finds the hardware overflow flag set. And [pytest.ini:7-8](pytest.ini#L7-L8) has `filterwarnings = error`, so that warning becomes a hard failure.
> 
> So both symptoms are downstream of one thing — the FP unit / control register on the new GitHub Windows image (the VS/MSVC update charris mentions) is in a bad state:
> - a stale/spurious **overflow** flag → `RuntimeWarning: overflow encountered in cast`
> - finite math returning **NaN** → `ValueError: arange: cannot compute length`
> 
> Your own conclusion in the thread ("it's an MSVC thing") is correct; this is just the second manifestation.
> 
> ### Practical implication for the PR
> 
> The added `-W ignore:overflow encountered in cast:RuntimeWarning` **will not catch this one**, for two reasons:
> 1. It's a `ValueError` raised from C, not a `RuntimeWarning` — `-W` can't touch it.
> 2. It happens at **import/collection** time (evaluating the module-level `PROTOCOLS` dict), before any test or warning filter for the case is relevant.
> 
> In other words, the warning-suppression band-aid addresses the mypy-job overflow symptom but the regular Windows test job will still blow up at collection as long as that runner's FP environment is broken. The real fix is environmental (pin/avoid the broken Windows image, or wait for the upstream runner image to be corrected) — there's nothing to fix in the typing tests themselves.
>

jorenham:
this seems to do the trick; thanks

seberg:
Wow, the new error points to something being very wrong, though?!  For a division to return a NaN it would have to be a division by zero or inf/inf.
Given the previous overflow maybe `inf` is actually more likely.  But for that particular paths it even looks like everything were Python objects up until that points (besides a float -> double conversion)?

Should we report this somewhere, although I'm not exactly sure where, might it even be related to uv?!

charris:
I'm not convinced that it is all the windows upgrade, there are other tests running under windows-latest and they are doing fine. Likely some combination of factors is involved.

PR #31643, opened 2026-06-14
TST/TYP: Fix mypy test failures with pytest 9.1.0

jorenham:
This Fixes the recent mypy test CI failures caused by pytest 9.1.0 new warnings about a generator function being used in `@pytest.mark.parametrize`.

With this, `spin mypy` will no longer fail when using `pytest==9.1.0`.

#### AI Disclosure
No AI

### DIFF
--- a/numpy/typing/tests/test_typing.py
+++ b/numpy/typing/tests/test_typing.py
@@ -30,8 +30,6 @@ else:
     NO_MYPY = False
 
 if TYPE_CHECKING:
-    from collections.abc import Iterator
-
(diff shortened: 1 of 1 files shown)


jorenham:
Hmmmm this seems to also be an issue in other tests, not just the mypy ones.

jorenham:
Ok nevermind; I'm confusing two separate CI issues. This pytest 9.1 issue is unrelated to typing, so I'll close this and address that as a whole. That Windows CI mypy test (overflow warning) problem is a separate issue, I'll investigate that a more. Sorry about the noise.

PR #31644, opened 2026-06-14
TST: Fix new pytest 9.1.0 warnings

jorenham:
Pytest 9.1.0 (https://github.com/pytest-dev/pytest/releases/tag/9.1.0) deprecated using non-collection iterators (generators and such) to parametrize tests. IMO that's a good call, so instead of ignoring those deprecation warnings, I updated the tests to use collection types (lists usually) for parametric tests.

There was also another new warning about class-scoped instance method fixtures now being deprecated that this fixes.

This should fix (at least) some of the recent CI test failures we've been seeing.

#### AI Disclosure
No AI

### DIFF
--- a/numpy/_core/tests/test_umath.py
+++ b/numpy/_core/tests/test_umath.py
@@ -479,31 +479,30 @@ class TestDivision:
         assert_equal(x % 100, [5, 10, 90, 0, 95, 90, 10, 0, 80])
 
     @pytest.mark.skipif(IS_WASM, reason="fp errors don't work in wasm")
-    @pytest.mark.parametrize("dtype,ex_val", itertools.product(
-        sctypes['int'] + sctypes['uint'], (
-            (
-                # dividend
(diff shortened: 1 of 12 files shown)


numpy/_core/tests/test_umath.py
```diff
@@ -479,31 +479,30 @@ def test_division_int(self):
...
+            "range(15)"
+        ),
+        (
+            # dividend
+            "np.arange(fo.min, fo.min+lsize).astype(dtype),"
+            # divisors
+            "np.arange(lsize//-2, lsize//2).astype(dtype),"
+            # scalar divisors
+            "range(fo.min, fo.min + 15)"
+        ), (
```
charris, line 499: Style? New line for `(`.

jorenham: Yea it's pretty inconsistent, but it's the same as it was, just dedented.

charris (approved):

charris:
Thanks Joren.

PR #31647, opened 2026-06-15
MAINT, TST: Pin unpinned pytest usage

charris:
The recent release of pytest 9.1.0 broke tests where it was not pinned. This uses the requirements files in those tests so that next time there is a problem it won't be a surprise. 

No AI was used.


charris:
This may also expose problems that were not caught before.

charris:
Yep, lots of new problems. Lets split this into two prs, one for the pin, the other for updating the pin. That will make a backport of fixes easier.

charris:
Looks like pixi decided it was a good day to go down.

charris:
The pixi problem is due to a version update with known breaking changes. It is unrelated.

charris:
Pixi fix has propagated.

ngoldbaum:
Thanks @charris.

PR #31657, opened 2026-06-16
BUG: linspace raises ValueError for non-finite values with integer dtype

### LINKED ISSUE #31651: BUG: linspace with infinite start/stop and integer dtype silently returns INT_MIN
### Describe the issue:

calling `np.linspace` with `np.inf` (or `-np.inf`) endpoints and an integer `dtype` silently returns `INT_MIN` (`-9223372036854775808`) for every element instead of raising an error. Only a `RuntimeWarning` is emitted an d no exception is raised so the wrong result propagates silently.

I believe the root cause is that the float64 intermediate array ends up containing non-finite values (`inf` or `NaN`, depending on the input combination) before the final cast. Casting a non-finite float to `int64` is undefined behavior in C; on x86-64 this consistently yields `INT_MIN`. NumPy does not validate that `start`/`stop` are finite when `integer_dtype=True` before performing this cast.

Related: #26699 and #17155 .

```python
import numpy as np

np.linspace(np.inf, np.inf, 5, dtype=int)
# array([-9223372036854775808, -9223372036854775808, -9223372036854775808,
#        -9223372036854775808, -9223372036854775808])

np.linspace(np.inf, -np.inf, 5, dtype=int)
# array([-9223372036854775808, -9223372036854775808, -9223372036854775808,
#        -9223372036854775808, -9223372036854775808])

```
Expected: `ValueError` (infinite values cannot be represented as integers).

### Error message:

```shell
RuntimeWarning: invalid value encountered in cast
  y = conv.wrap(y.astype(dtype, copy=False))


No exception is raised.
```

2.3.3
3.13.0 (tags/v3.13.0:60403a5, Oct  7 2024, 09:38:07) [MSC v.1941 64 bit (AMD64)]

### Runtime Environment:

_No response_

found while auditing edge cases in linspace with integer dtypes. Silent data corruption with no exception, only a RuntimeWarning which means callers using np.errstate(invalid='ignore') or suppressed warnings would receive INT_MIN values with no indication anything went wrong

kaizeenn:
## Summary

Closes #31651.

`np.linspace` with infinite (or NaN) endpoints and an integer `dtype` silently returns `INT_MIN` for every element instead of raising an error:

```python
np.linspace(np.inf, np.inf, 5, dtype=int)
# array([-9223372036854775808, ...])  ← silent data corruption
```

## Root cause

After the float64 intermediate array is built, `floor()` is called and then `astype(dtype)` converts to the requested integer type. Casting a non-finite float to integer is undefined behaviour in C; on x86-64 it consistently yields `INT_MIN`. NumPy emits a `RuntimeWarning` but does not raise, so callers using `np.errstate(invalid='ignore')` or suppressed warnings receive corrupt data silently.

## Fix

Add a finiteness check immediately after `floor()` (inside the `if integer_dtype:` branch) that raises `ValueError` with a descriptive message before the `astype()` call. Float-dtype paths are completely unaffected.

```python
if integer_dtype:
    if not _nx.isfinite(y).all():
        raise ValueError(
            "Cannot convert non-finite values (inf, -inf, or NaN) to "
            "integer dtype. Use a floating-point dtype or ensure that "
            "start and stop are finite."
        )
    _nx.floor(y, out=y)
```

## Tests

Added `TestLinspace.test_integer_dtype_non_finite_raises` covering:
- `+inf` to `+inf`
- `+inf` to `-inf`
- `-inf` to finite
- `NaN` start
- Regression: normal integer-dtype `linspace(0, 10, 5, dtype=int)` still works

### DIFF
--- a/numpy/_core/function_base.py
+++ b/numpy/_core/function_base.py
@@ -179,6 +179,12 @@ def linspace(start, stop, num=50, endpoint=True, retstep=False, dtype=None,
         y = _nx.moveaxis(y, 0, axis)
 
     if integer_dtype:
+        if not _nx.isfinite(y).all():
+            raise ValueError(
+                "Cannot convert non-finite values (inf, -inf, or NaN) to "
+                "integer dtype. Use a floating-point dtype or ensure that "
+                "start and stop are finite."
+            )
         _nx.floor(y, out=y)
 
     y = conv.wrap(y.astype(dtype, copy=False))
--- a/numpy/_core/tests/test_function_base.py
+++ b/numpy/_core/tests/test_function_base.py
@@ -482,6 +482,22 @@ class TestLinspace:
         y = linspace(start, stop, 3)
         assert_array_equal(y, array([[0.0, 1.0], [1.0, 1.0], [2.0, 1.0]]))
 
+    def test_integer_dtype_non_finite_raises(self):
+        # gh-31651: non-finite start/stop with integer dtype must raise
+        # ValueError instead of silently returning INT_MIN.
+        with pytest.raises(ValueError, match="non-finite"):
+            linspace(np.inf, np.inf, 5, dtype=int)
+        with pytest.raises(ValueError, match="non-finite"):
+            linspace(np.inf, -np.inf, 5, dtype=int)
+        with pytest.raises(ValueError, match="non-finite"):
+            linspace(-np.inf, 0, 3, dtype=int)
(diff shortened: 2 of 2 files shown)


seberg:
Thanks, but I'll close this.  I think this needs a decision on the issue but mainly it doesn't make sense to do in such a narrow way.  Either you do some strict checking for any out-of-bound or there is no point infinities are not special.

PR #26666, opened 2024-06-12
WIP: top_k draft implementation

JuliaPoo:
This is a follow up of the previous discussions at #15128. I've changed the `mode` keyword (as in previous discussion) into a `largest` bool flag. This follows API such as that from [torch.topk](https://pytorch.org/docs/stable/generated/torch.topk.html).

The API implemented here is:

```py
def top_k(
    a: ArrayLike,
    k: int,
    /,
    *,
    axis: None | int = ...,
    largest: bool = ...,
) -> tuple[NDArray[Any], NDArray[intp]]: ...
```

Carrying forward from the previous discussion, a parameter that might be useful is `sorted`. This is implemented in `torch.topk`, and follows from previous work at #19117.

PR #31659, opened 2026-06-16
ENH: Add `top_k` based on partitioning

MaanasArora:
### PR summary
Adds a `top_k` function to NumPy to take the largest/smallest elements from the array. This is taken from the previous implementation in #26666 but enhanced and simplified with descending argpartitions and a `sorted` parameter. With the new descending argpartitions, we no longer need manual handling of the `largest` parameter. 

Since argpartitions are always unstable, I use unstable sorts as well (as we can't make it stable easily and with good performance anyway.) The tests are from the original implementation, except I parametrized a test for `sorted`. Thanks for reviewing!

#### AI Disclosure
No AI tools used.


ngoldbaum:
It looks like there were NumPy mailing list discussions about this feature back in [2021](https://mail.python.org/archives/list/numpy-discussion@python.org/thread/F4P5UVTAKRJJ3OORI6UOWFSUEE5CNTSC/#UWULPEEVC4SRHVW37MSXJJKZL6YUSUFU) and [2022](https://mail.python.org/archives/list/numpy-discussion@python.org/thread/TCRBG6VOHHFMV3CWR3AFTVZK4JTVAG2K/#2IVDQ3AFZLGL6C3WJBHQXSHUOSQKTUP2). Per our policy, [you're supposed to ping the mailing list](https://numpy.org/doc/stable/dev/development_workflow.html#get-the-mailing-list-s-opinion) if you want to add a new feature or API surface, which this does.

Just wanted to raise that first before doing a deeper code review.

numpy/_core/fromnumeric.py
```diff
@@ -200,6 +201,114 @@ def take(a, indices, axis=None, out=None, mode='raise'):
...
+    if k <= 0:
+        raise ValueError(f'k(={k}) provided must be positive.')
+
+    arr = np.asanyarray(a)
+
+    if axis is None:
+        arr = arr.ravel()
+        axis = 0
+    elif axis <= 0:
+        axis = axis % arr.ndim
```
ngoldbaum: We have a [`normalize_axis_index`](https://github.com/numpy/numpy/blob/f5c9a42bfc2531fe0a8ac80946878a67c7a00e9a/numpy/_core/src/multiarray/multiarraymodule.c#L4330-L4350) helper for this. Grep for uses elsewhere in NumPy for examples.

numpy/_core/fromnumeric.py
```diff
@@ -200,6 +201,114 @@ def take(a, indices, axis=None, out=None, mode='raise'):
...
+
+    indices = np.argpartition(arr, k - 1, axis=axis, descending=largest)
```
ngoldbaum: You probably have to add a fallback for types that don't yet support descending sorts:

```
In [1]: np.top_k(["a", "b"], 1)
File ~/Documents/numpy/build-install/usr/lib/python3.14t/site-packages/numpy/_core/fromnumeric.py:55, in _wrapfunc(obj, method, *args, **kwds)
     54 try:
---> 55     return bound(*args, **kwds)
     56 except TypeError:
     57     # A TypeError occurs if the object does have such a method in its
     58     # class, but its signature is not identical to that of NumPy's. This
   (...)     62     # Call _wrapit from within the except clause to ensure a potential
     63     # exception has a traceback chain.

TypeError: type does not support descending partition

During handling of the above exception, another exception occurred:

Cell In[1], line 1
----> 1 np.top_k(["a", "b"], 1)

File ~/Documents/numpy/build-install/usr/lib/python3.14t/site-packages/numpy/_core/fromnumeric.py:297, in top_k(a, k, axis, largest, sorted)
    294 elif axis <= 0:
    295     axis = axis % arr.ndim
--> 297 indices = np.argpartition(arr, k - 1, axis=axis, descending=largest)
    299 slice_ = (np.s_[:],) * axis + (np.s_[:k],)
    300 indices = indices[slice_]

File ~/Documents/numpy/build-install/usr/lib/python3.14t/site-packages/numpy/_core/fromnumeric.py:1083, in argpartition(a, kth, axis, kind, order, descending)
   1080 if kind is not np._NoValue:
   1081     kwargs['kind'] = kind
-> 1083 return _wrapfunc(a, "argpartition", kth, axis=axis, order=order, **kwargs)

File ~/Documents/numpy/build-install/usr/lib/python3.14t/site-packages/numpy/_core/fromnumeric.py:64, in _wrapfunc(obj, method, *args, **kwds)
     55     return bound(*args, **kwds)
     56 except TypeError:
     57     # A TypeError occurs if the object does have such a method in its
     58     # class, but its signature is not identical to that of NumPy's. This
   (...)     62     # Call _wrapit from within the except clause to ensure a potential
     63     # exception has a traceback chain.
---> 64     return _wrapit(obj, method, *args, **kwds)

File ~/Documents/numpy/build-install/usr/lib/python3.14t/site-packages/numpy/_core/fromnumeric.py:44, in _wrapit(obj, method, *args, **kwds)
     41 # As this already tried the method, subok is maybe quite reasonable here
     42 # but this follows what was done before. TODO: revisit this.
     43 arr, = conv.as_arrays(subok=False)
---> 44 result = getattr(arr, method)(*args, **kwds)
     46 return conv.wrap(result, to_scalar=False)

TypeError: type does not support descending partition
```

numpy/_core/fromnumeric.py
```diff
@@ -200,6 +201,114 @@ def take(a, indices, axis=None, out=None, mode='raise'):
...
 
+def _top_k_dispatcher(a, k, /, *, axis=-1, largest=True, sorted=True):
+    return (a,)
+
+
+@array_function_dispatch(_top_k_dispatcher)
+def top_k(a, k, /, *, axis=-1, largest=True, sorted=True):
+    """
+    Returns the ``k`` largest/smallest elements and corresponding
+    indices along the given ``axis``.
```
ngoldbaum: How about "Return the ``k`` largest or smallest elements and their indices along an axis".

Keeping the first sentence as short as possible helps, since this shows up in e.g. `help()`.

numpy/_core/fromnumeric.py
```diff
@@ -200,6 +201,114 @@ def take(a, indices, axis=None, out=None, mode='raise'):
...
+    sorted: bool, optional
+        If True, the top ``k`` elements are returned in sorted order.
+        If False, sorted order is not guaranteed. The default is True.
+
+    Returns
+    -------
+    tuple_of_array: tuple
+        The output tuple of ``(topk_values, topk_indices)``, where
+        ``topk_values`` are the top ``k`` values and ``topk_indices``
+        are the corresponding indices.
```
ngoldbaum: Can this document the shape of the output array explicitly?

numpy/_core/fromnumeric.py
```diff
@@ -200,6 +201,114 @@ def take(a, indices, axis=None, out=None, mode='raise'):
...
+
+    Notes
+    -----
+    The returned indices are not guaranteed to be stable, i.e., the order
+    of the returned for any duplicate values is not guaranteed to be the
```
ngoldbaum: I think you accidentally a word: "the order of the returned for any duplicate"

numpy/_core/fromnumeric.py
```diff
@@ -200,6 +201,114 @@ def take(a, indices, axis=None, out=None, mode='raise'):
...
+    same as their order in the input array. This is the case regardless
+    of the value of the ``sorted`` parameter.
+
```
ngoldbaum, line 272: Probably needs to mention what happens with dtypes you haven't added the necessary infrastructure for?

numpy/_core/fromnumeric.py
```diff
@@ -200,6 +201,114 @@ def take(a, indices, axis=None, out=None, mode='raise'):
...
+    See Also
+    --------
+    argpartition : Indirect partition.
+    sort : Full sorting.
+
+    Examples
+    --------
+    >>> a = np.array([[1,2,3,4,5], [5,4,3,2,1], [3,4,5,1,2]])
```
ngoldbaum: Can you make this be a slightly smaller array so the output below is a little more compressed?

numpy/_core/fromnumeric.py
```diff
@@ -200,6 +201,114 @@ def take(a, indices, axis=None, out=None, mode='raise'):
...
+    (array([[5, 4, 5, 4, 5],
+            [3, 4, 3, 2, 2]]),
+     array([[1, 1, 2, 0, 0],
+            [2, 2, 1, 1, 2]]))
+    >>> a.flatten()
+    array([1, 2, 3, 4, 5, 5, 4, 3, 2, 1, 3, 4, 5, 1, 2])
+    >>> np.top_k(a, 2, axis=None)
+    (array([5, 5]), array([4, 5]))
+    >>> np.top_k(np.array([1., 2., 3., np.nan]), 2)
+    (array([3., 2.]), array([2, 1]))
```
ngoldbaum, line 297: Maybe an example with `largest=False`? You could maybe delete the `a.flatten()`? Not sure how useful that is.

ngoldbaum (commented):
I have one concrete code suggestion, a question about how to handle unsupported dtypes, and a bunch of documentation nits.

Overall this looks great though and IMO this is a good idea, we just need to dot our i's and cross our t's for the process on adding new API surface.

numpy/_core/fromnumeric.py
MaanasArora: Thanks, all in-built types except `void` and `generic` will support descending partitioning after #31614 - in fact, descending partitioning will fallback to sorting for all dtypes that support a sort but not partition (in the built-in dtypes, strings and objects), though we should clearly add an actual selection algorithm as well for these. So, I don't think we need a fallback?

MaanasArora:
Thank you for the review! I will ping the mailing list right away and address the comments after, but just answered about unsupported types.

rgommers:
> It looks like there were NumPy mailing list discussions about this feature back in [2021](https://mail.python.org/archives/list/numpy-discussion@python.org/thread/F4P5UVTAKRJJ3OORI6UOWFSUEE5CNTSC/#UWULPEEVC4SRHVW37MSXJJKZL6YUSUFU) and [2022](https://mail.python.org/archives/list/numpy-discussion@python.org/thread/TCRBG6VOHHFMV3CWR3AFTVZK4JTVAG2K/#2IVDQ3AFZLGL6C3WJBHQXSHUOSQKTUP2). Per our policy, [you're supposed to ping the mailing list](https://numpy.org/doc/stable/dev/development_workflow.html#get-the-mailing-list-s-opinion) if you want to add a new feature or API surface, which this does.

This was already settled, it was just blocked on implementation. See gh-15128, which both links to the previous mailing list discussion and summarizes the result of it.

rgommers:
To elaborate: a fresh ping on the mailing list that this is now close to being merged is a good idea, but we're not starting from scratch here, it's more an FYI than an open "let's redesign the API from scratch".

ngoldbaum:
Agreed, thanks for that extra historical context about this.

numpy/_core/fromnumeric.py
MaanasArora: Not quite sure how I would put it, given argsorts and sorts for descending didn't have it either!

MaanasArora:
Thanks, the mailing list post is up, and I think I've addressed most of the comments!

numpy/_core/fromnumeric.pyi
```diff
@@ -200,6 +201,16 @@ def take[ArrayT: np.ndarray](
...
 
+def top_k(
+    a: ArrayLike,
+    k: int,
+    /,
+    *,
+    axis: None | int = ...,
+    largest: bool = ...,
+    sorted: bool = ...,
+) -> tuple[NDArray[Any], NDArray[intp]]: ...
```
jorenham, line 212:
```suggestion
def top_k(
    a: ArrayLike,
    k: int,
    /,
    *,
    axis: int | None = -1,
    largest: bool = True,
    sorted: bool = True,
) -> tuple[NDArray[Any], NDArray[intp]]: ...
```

There's still room for improvement here, but that'd require adding a bunch of overloads, and it's a bit unreasonable to ask that of you here haha. So let's leave that for a follow-up; this is fine for now.

numpy/_core/fromnumeric.py
seberg: At https://github.com/data-apis/array-api/pull/722 there was also the `mode="largest"` rather than boolean.  I am not sure I prefer the string but we should bring it up as well there (and otherwise ignore it here for the moment).

However, one thing I very much prefer about the string is that `largest=True` is the default (nobody will write it out, `top` sounds like largest already) but `top_k(..., largest=False)` seems like a weird way to spell "smallest".

(Sorry, just noting here, I'll also post there.  For now we can ignore and it doesn't matter much where we land, but it seems an OK thing to give a chance for bikeshedding.)

MaanasArora:
Thanks everyone! I've updated this PR according to the consensus in the Array API PR (https://github.com/data-apis/array-api/pull/722):

- There is no flattening behavior, `axis=None` fails explicitly.
- `mode` is now a string (`Literal["largest", "smallest"]`) that defaults to `largest`. This maps to the previous boolean; behavior is unchanged otherwise.

This should be ready for another round of review!

numpy/_core/tests/test_multiarray.py
```diff
@@ -3959,6 +3959,76 @@ def test_argpartition_gh5524(self, kth_dtype):
...
+            ValueError,
+            r'mode\(="invalid"\) must be either "largest" or "smallest".'
+        ):
+            np.top_k(a, 2, mode="invalid")
+
+        with assert_raises_regex(
+            ValueError,
+            r"axis=None is not supported. Please provide a valid axis."
+        ):
+            np.top_k(a, 2, axis=None)
```
ikrommyd: I think it's preferable to use to use `with pytest.raises` instead of these  `assert_raises/assert_raises_regex` functions even though a lot of old code currently does (Sebastian commented that to another PR I made recently)

numpy/_core/fromnumeric.py
```diff
@@ -200,6 +201,111 @@ def take(a, indices, axis=None, out=None, mode='raise'):
...
 
+def _top_k_dispatcher(a, k, /, *, axis=-1, mode="largest", sorted=True):
+    return (a,)
+
+
+@array_function_dispatch(_top_k_dispatcher)
+def top_k(a, k, /, *, axis=-1, mode="largest", sorted=True):
+    """
+    Returns the ``k`` largest or smallest elements and their indices
+    indices along an axis.
```
ikrommyd: Duplicate "indices"
```suggestion
    Returns the ``k`` largest or smallest elements and their
    indices along an axis.
```

numpy/_core/fromnumeric.pyi
```diff
@@ -200,6 +201,16 @@ def take[ArrayT: np.ndarray](
     mode: _ModeKind = "raise",
 ) -> ArrayT: ...
 
+def top_k(
+    a: ArrayLike,
+    k: int,
+    /,
+    *,
+    axis: int | None = -1,
```
ikrommyd: `axis=None` is rejected at runtime and the docstring says `axis: int, optional`.

doc/release/upcoming_changes/31659.new_function.rst
```diff
@@ -0,0 +1,6 @@
+New function `numpy.top_k`
+----------------------------
+
+A new function ``np.top_k(array, k, axis=..., mode=..., sorted=...)``
```
ikrommyd: Trailing whitespace found with `git diff --check main...`
```suggestion
A new function ``np.top_k(array, k, axis=..., mode=..., sorted=...)``
```

ikrommyd (commented):
I'm wondering what we want to do here for things like this which will fail for mode="largest" and will pass for mode="smallest".
```py
>>> a = np.array(["a", "c", "b"])
>>> np.top_k(a, 2)
  File "/Users/iason/repos/numpy/build-install/usr/lib/python3.15/site-packages/numpy/_core/fromnumeric.py", line 55, in _wrapfunc
    return bound(*args, **kwds)
TypeError: type does not support descending partition

During handling of the above exception, another exception occurred:

  File "<python-input-10>", line 1, in <module>
    np.top_k(a, 2)
    ~~~~~~~~^^^^^^
  File "/Users/iason/repos/numpy/build-install/usr/lib/python3.15/site-packages/numpy/_core/fromnumeric.py", line 294, in top_k
    indices = np.argpartition(arr, k - 1, axis=axis, descending=largest)
  File "/Users/iason/repos/numpy/build-install/usr/lib/python3.15/site-packages/numpy/_core/fromnumeric.py", line 1080, in argpartition
    return _wrapfunc(a, "argpartition", kth, axis=axis, order=order, **kwargs)
  File "/Users/iason/repos/numpy/build-install/usr/lib/python3.15/site-packages/numpy/_core/fromnumeric.py", line 64, in _wrapfunc
    return _wrapit(obj, method, *args, **kwds)
  File "/Users/iason/repos/numpy/build-install/usr/lib/python3.15/site-packages/numpy/_core/fromnumeric.py", line 44, in _wrapit
    result = getattr(arr, method)(*args, **kwds)
TypeError: type does not support descending partition
>>> np.top_k(a, 2, mode="smallest")
(array(['a', 'b'], dtype='<U1'), array([0, 2]))
```
However, sorting and argsorting works fine with this array for ascending and descending sorts. Should `top_k` add a fallback with `np.argsort` perhaps, or guard that dtypes for now, or just let it error like it does until `argpartition` supports descending for such dtypes?

numpy/_core/fromnumeric.pyi
MaanasArora: Thanks for catching, this is fixed!

numpy/_core/tests/test_multiarray.py
MaanasArora: Done, thank you!

MaanasArora:
Thank you for the review! I think I have addressed all of your comments.

> However, sorting and argsorting works fine with this array for ascending and descending sorts. Should top_k add a fallback with np.argsort perhaps, or guard that dtypes for now, or just let it error like it does until argpartition supports descending for such dtypes?

After #31614, fallback to argsorts will naturally happen for any dtype that uses new-style sorting in `PyArray_ArgPartition` itself (that includes basically all built-in NumPy dtypes), so I don't believe we need another fallback here. User dtypes that won't support new-style sorting will obviously also not support descending sorts, so this cuts well theoretically I think. However Nathan also did bring this up above, so I'm OK to add a temporary fallback until that PR is merged if we deem it necessary, or we can discuss it more if needed.

ikrommyd:
Ah thanks I missed that discussion. Yeah either a fallback or a better error than `TypeError: type does not support descending partition` (which is raised internally from the `argpartition` call). If that PR is coming in quickly, I'd argue that a fallback is probably not worth the effort and a better error will do just fine here. But the maintainers should decide on that :)

seberg:
I dunno, I am comfortable with not having a perfect error message for these possible niche cases (and yeah, that other PR we should just make sure to get in).
I have two small thoughts although no strong enough opinion to care:
* I somewhat feel `top_k(arr, 0)` makes perfect sense.  I don't feel strongly about it, though. Mainly because it doesn't just drop out of course since `partition(arr, -1)` has a different meaning.  `partition(arr, [])` could work to get the error but also a bit annoying.  (Which is probably why it is rejected right now.)
* We could allow `axis=None`, but I am +/-0 on it.  OTOH, NumPy really allows it everywhere, on the other hand, maybe it's time to "discourage" it by just omitting it here.
  (Array API has no opinion about this for NumPy as it is cleanly defined.)

MaanasArora:
> I am comfortable with not having a perfect error message for these possible niche cases

I agree, after the other PR the case is niche indeed (user dtypes with legacy sorting), so I'd just use the existing error.

> I somewhat feel top_k(arr, 0) makes perfect sense.

Agreed as well, committed in the Array API PR as well now. I checked that `partition(arr, [])` doesn't actually seem to work unfortunately (`TypeError: Partition index must be integer`), so I just pushed a change that uses `kth=0` instead (not very nice, admittedly, though at least the slicing handles it implicitly)...

> (Array API has no opinion about this for NumPy as it is cleanly defined.)

Hmm, my impression was that the Array API _disallows_ this? If so that clears it I guess, otherwise I don't have a strong opinion either way (the reason I might not be very keen to include it despite precedence might just be that the function definition/description is pretty top-down, so good excuse :))

seberg:
> partition(arr, [])

You'd need to use `np.array([], dtype=np.intp)` :/.  Unfortunately, using `0` is also not strictly right I think because `np.top_k([], 0)` makes sense as well.

> Hmm, my impression was that the Array API disallows this? 

No, but it's honestly hard to say because it is never quite clear who the audience is (the library author or the end-user -- that is pragmatic, but leads to these confusions):  The end-user *cannot rely on it being defined*, but if there is language suggesting that array implementers must raise an error for `axis=None` then that would most likely just be a mistake there.

MaanasArora:
> You'd need to use np.array([], dtype=np.intp)

Ah sorry, right, pushed this thanks!

> The end-user cannot rely on it being defined

You're right, that makes sense. I don't have a strong opinion then!

ikrommyd:
Yeah since sorting like `np.argsort` allow `axis=None` with flattening, I think it makes sense to allow it here too.

seberg:
Sorry for not looking at it for a bit: basically think it is all fine but forgot to merge.  I do have the small ask: since nobody else weighed in on `axis=None`, I think we should just support the ravel behavior, TBH.  (I also feel the test would be nicer to split up, but I don't care).

Nothing too blocking, so if someone else wants to push it through quicker, fine.

ngoldbaum:
OK, in that case let’s merge this and then @MaanasArora can open a followup issue for the `axis=None` behavior.

ngoldbaum:
Thanks @MaanasArora!

MaanasArora:
Thanks for the reviews everyone! Follow-up coming soon...

PR #31660, opened 2026-06-16
TYP: Fix ``flatiter.__next__`` return type for ``object_`` and `StringDType`

jorenham:
### PR summary
While working on https://github.com/librosa/librosa/pull/2052, a new mypy error popped up, complaining that the iterated values of `ndarray.flat` for an `object_` array aren't subscriptable, because `np.object_` does not support `__getitem__`. In reality, `np.object_` values are pure fiction, so it `flatiter.__next__` should instead return `Any` for `object_` arrays.

This also special-cases `flatiter.__next__`  for `StringDType` arrays, which have no numpy scalar type, and would yield `builtins.str` instead.

#### AI Disclosure
No AI


numpy/_core/multiarray.pyi
```diff
@@ -1726,7 +1726,16 @@ class flatiter(Generic[_ArrayT_co]):
...
+
+    # we need to be careful not to yield fictional `np.object_` values
+    @overload
+    def __next__(self: flatiter[NDArray[np.object_]], /) -> Any: ...
+    @overload  # we can't use `np.generic` because it includes `np.object_`
+    def __next__[ScalarT: np.bool | np.number | np.flexible | np.datetime64 | np.timedelta64](
+        self: flatiter[NDArray[ScalarT]], /
+    ) -> ScalarT: ...
+    @overload  # `StringDType` has no scalar type
+    def __next__(self: flatiter[np.ndarray[Any, np.dtypes.StringDType]], /) -> str: ...
```
ngoldbaum, line 1738: Do you want to care about the fact that accessing a StringDType array can return the `na_object` as well?

```
>>> import numpy as np
>>> flt = np.array(["x", None], dtype=np.dtypes.StringDType(na_object=None)).flat
>>> flt[0]
'x'
>>> flt[1]
>>> flt[1] is None
True
```

It looks like this pattern occurs elsewhere (`ndarray.item`, `ndarray.__iter__`, `ndarray.tolist`, and `flatiter.__getitem__(int)` to name a few) so feel free to disregard this for now and do a fix focused on that later.

jorenham: > Do you want to care about the fact that accessing a StringDType array can return the `na_object` as well?

Part of me does, but I'm also a bit worried that doing so will set a precedent where we'll have to add special-casing for this all over the place. So I'm not sure if that's the way we want to go, given that the stubs are already a big mess.
But on the other hand, the stubs are *already* a big mess, so the ∂mess/∂code probably won't be very high.

In the stubs `StringDType` is already generic on the na object type, so at least that part is already possible: https://github.com/numpy/numpy/blob/8906919c2f37dcdd11f4d61298486c5261fb87b9/numpy/dtypes.pyi#L585

charris:
Thanks Joren.

PR #31661, opened 2026-06-16
BUG: raise error when casting from StringDType to longdouble with trailing null

ngoldbaum:
Fixes an issue in the StringDType to longdouble cast, where:

* We failed to check for an allocation failure
* We used `NumPyOS_ascii_strtold` but did not check all possible error cases for that function. In particular, if the parsed longdouble ends before the end of the string, like with the test case that has a trailing NULL.

This fixes that by improving the error checking. Also adds test cases for converting various invalid strings to built-in numeric dtypes.


seberg (approved):
Thanks, nice and to the point!  The new order also looks a bit clearer and tests are fine.

seberg:
Ooops, you got a conflict though...

PR #31662, opened 2026-06-16
BUG: NpyString comparison should use memcmp instead of strncmp

ngoldbaum:
strncmp compares *C strings*, and stops comparisons after hitting an embedded null. We should be using memcmp for our strings, which can contain embedded nulls.

Fixes that and adds tests for comparisons and conversions 

I used AI assistance to identify and help fix the bug.

Ping @mhvk since you were involved with the design here. C is hard!

### DIFF
--- a/doc/release/upcoming_changes/31662.improvement.rst
+++ b/doc/release/upcoming_changes/31662.improvement.rst
@@ -1,0 +1,1 @@
+StringDType comparisons now correctly handle embedded NULL bytes.
--- a/doc/release/upcoming_changes/31662.improvement.rst
+++ b/doc/release/upcoming_changes/31662.improvement.rst
@@ -1,1 +1,1 @@
-StringDType comparisons now correctly handle embedded NULL bytes.
+* StringDType comparisons now correctly handle embedded NULL bytes.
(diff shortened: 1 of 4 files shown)


doc/release/upcoming_changes/31662.improvement.rst
```diff
@@ -0,0 +1 @@
+StringDType comparisons now correctly handle embedded NULL bytes.
```
seberg:
```suggestion
* StringDType comparisons now correctly handle embedded NULL bytes.
```
I know, I know...

numpy/_core/tests/test_stringdtype.py
```diff
@@ -166,6 +168,67 @@ def test_null_roundtripping():
...
+
+
+@pytest.mark.parametrize("dtype", [object, "U{}", "S{}", "V{}"])
+def test_embedded_null_string_like_casts(dtype):
+    strings = ["a\0b", "\0leading", "multi\0null\0inside"]
+    arr = np.array(strings, dtype="T")
+    if dtype is object:
+        def dtype_func(size): return object
+    else:
+        def dtype_func(size): return dtype.format(size)
```
seberg: Honestly, I would just use a clearly large enough size `"S20"` or so. (but also fine without)

seberg (approved):
LGTM, can merge if you like with that nit applied (or wait for Marten).

mhvk (approved):
Looks good!

PR #31663, opened 2026-06-16
BUG: fix casts from StringDType to complex float

ngoldbaum:
This was clearly a copy/paste bug and the tests were written in such a way that they didn't notice the bug.

This was spotted using an AI model.


seberg (approved):
Thanks feel free to just merge.  If you feel like it otherwise: There seems little reason to not just modify the earlier test rather than adding a new one.
(But dunno, likely a matter of taste anyway.)

ngoldbaum:
> If you feel like it otherwise: There seems little reason to not just modify the earlier test rather than adding a new one.

I tried that first but it breaks the xfails on that test. This seemed easier than figuring that out.

PR #31664, opened 2026-06-16
MAINT: Update pytest pins to 9.1.0

charris:
This exposes 9.1.0 problems that were not yet fixed. Then we fix them.

Grok suggested the emscripten fix, but frankly, Grok wasn't all the helpful.

### DIFF
--- a/.github/workflows/emscripten.yml
+++ b/.github/workflows/emscripten.yml
@@ -40,4 +40,3 @@ jobs:
         env:
           CIBW_PLATFORM: pyodide
           CIBW_BUILD: cp312-*
-
--- a/.github/workflows/emscripten.yml
+++ b/.github/workflows/emscripten.yml
@@ -40,3 +40,4 @@ jobs:
         env:
           CIBW_PLATFORM: pyodide
           CIBW_BUILD: cp312-*
+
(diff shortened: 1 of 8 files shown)


charris:
Both `Linux tests / i686, cp312, full ` and `macOS tests / macOS x86-64 conda (3.12)` have a ton of errors like:
```
ERROR numpy/f2py/tests/test_array_from_pyobj.py::TestSharedMemory::test_inplace[CLONGDOUBLE] - AssertionError
```
Both run the f2py tests, I assume there is a 9.1.0 problem with the shared memory tests.

The emscripten failure is a permissions problem in clearing the pytest cache, we probably should disable the pytest caches and ignore the matrixlib deprecations.

charris:
NVM, the failing Fortran tests had a fixture that was causing the problem.

charris:
OK, I think this is ready for review. The pyodide warning suppression could be made more precise, but I think it is fine. Those PendingDeprecationsWarning messages have been showing up forever. They are supressed in _pytesttester for the other wheel tests.

.github/workflows/emscripten.yml
```diff
@@ -40,3 +40,4 @@ jobs:
         env:
           CIBW_PLATFORM: pyodide
           CIBW_BUILD: cp312-*
+
```
ngoldbaum:
```suggestion
```

ngoldbaum (approved):
LGTM! I was having trouble with the github suggestion UI giving a valid diff suggestion so I just went ahead and pushed a fixup commit for some minor style issues.

ngoldbaum:
Let's go ahead and merge this since CI already passed on the previous commit and most CI has already passed now.

ngoldbaum:
Thanks @charris!

PR #31665, opened 2026-06-16
BUG: fix StringDType embedded and trailing NULL handling in several operations

ngoldbaum:
Several StringDType ufunc operations (`str_len`, `find`, `count`, `rstrip`, `endswith`, `slice`) didn't handle trailing and/or embedded nulls correctly, see new tests for specific examples.


seberg (approved):
Thanks, looks all correct to me.  The other strings (where `buf` is used) are also correct since the buffer discovers the size.

PR #31666, opened 2026-06-17
TST: Avoid sctypeDict in tests as e.g. ml_dtypes may mutate it

seberg:
Using the typecodes may not be the nicest but is a common pattern and gives the exact same types otherwise.

Without this, if you import `ml_dtypes` these two tests fail. This can happen if:
* There is a chain of dependencies through some plugins.
* We ever start testing with `ml_dtypes` ourselves when installed.

(no AI)

### DIFF
--- a/numpy/_core/tests/test_multiarray.py
+++ b/numpy/_core/tests/test_multiarray.py
@@ -9210,7 +9210,7 @@ class TestNewBufferProtocol:
         self._check_roundtrip(x)
 
     def test_roundtrip_single_types(self):
-        for typ in np._core.sctypeDict.values():
+        for typ in np.typecodes["All"]:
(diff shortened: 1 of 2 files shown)


ngoldbaum:
I think you're hitting the issues fixed by https://github.com/numpy/numpy/pull/31664 in one of your CI jobs. I'm going to review that PR to unblock CI here then review this one.

ngoldbaum (approved):
LGTM! Looks like both cases have been stable for years and until we come up with something better `np.typecodes` is the API for doing stuff like this if you only want to consider NumPy types.

ngoldbaum:
Rebasing on main should make the conda failure go away.

ngoldbaum:
Only failures are upstream network issues now, thanks!

PR #31667, opened 2026-06-17
DOC: Update C extension tutorial to use PyInit_{name} and PyModuleDef API and fix missing PyArrayObject cast

prathamhole14:
### PR summary

I was going through the https://numpy.org/doc/stable/user/c-info.ufunc-tutorial.html and
https://numpy.org/doc/stable/user/c-info.how-to-extend.html#

I noticed some inconsistencies, and I think it's because the "How to Extend" page is outdated. I've tried to bring it up to date and make it consistent with the newer documentation.

Please take a look and let me know if it looks correct.

#### AI Disclosure

No

### DIFF
--- a/doc/source/user/c-info.how-to-extend.rst
+++ b/doc/source/user/c-info.how-to-extend.rst
@@ -53,27 +53,44 @@ Required subroutine
 
 There is exactly one function that must be defined in your C-code in
 order for Python to use it as an extension module. The function must
-be called init{name} where {name} is the name of the module from
+be called ``PyInit_{name}`` where ``{name}`` is the name of the module from
 Python. This function must be declared so that it is visible to code
 outside of the routine. Besides adding the methods and constants you
 desire, this subroutine must also contain calls like ``import_array()``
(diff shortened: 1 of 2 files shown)


doc/source/user/c-info.how-to-extend.rst
```diff
@@ -53,35 +53,52 @@ Required subroutine
...
+    PyInit_{name}(void)
     {
-       (void)Py_InitModule({name}, mymethods);
+       PyObject *m;
        import_array();
+       m = PyModule_Create(&moduledef);
+       if (!m) {
+           return NULL;
+       }
+       return m;
```
da-woods: The error check is a little pointless here and could just be `return PyModule_Create(&moduledef);` (although I guess might be worth spelling out if people are going to add additional actions after creating the module).

prathamhole14:
Thank you for the review, @da-woods. Please take another look at the updates.

mattip:
This is an improvement, if more is needed let's do it in follow-up PRs. Thanks @prathamhole14

PR #31669, opened 2026-06-17
Backport 31644

charris:
Backport of #31644.

Pytest 9.1.0 (https://github.com/pytest-dev/pytest/releases/tag/9.1.0) deprecated using non-collection iterators (generators and such) to parametrize tests. IMO that's a good call, so instead of ignoring those deprecation warnings, I updated the tests to use collection types (lists usually) for parametric tests.

There was also another new warning about class-scoped instance method fixtures now being deprecated that this fixes.

This should fix (at least) some of the recent CI test failures we've been seeing.

#### AI Disclosure
No AI

### DIFF
--- a/numpy/_core/tests/test_array_coercion.py
+++ b/numpy/_core/tests/test_array_coercion.py
@@ -16,16 +16,15 @@ from numpy.testing import IS_64BIT, assert_array_equal
 
 
 def arraylikes():
-    """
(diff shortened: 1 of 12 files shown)


charris:
I've merged these fixes into the PR pinning to pytest 9.1.0 in order to fix the failing macos test, so closing.

PR #31670, opened 2026-06-18
BUG: Use `auxdata` in (arg)sort `strided_loop`s

MaanasArora:
This was somehow missed back in #29737, sorry :/ I've marked this for backport as it's a clear bug in the user dtype implementation for slots.

#### AI Disclosure
No AI tools used.


seberg (approved):
Thanks, clearly right.  I'll accept that a test for this is tedious since it is pretty much unused in practice.

PR #31672, opened 2026-06-18
BUG: raise error when casting from StringDType to longdouble with trailing null (#31661)

charris:
Backport of #31661.

Fixes an issue in the StringDType to longdouble cast, where:

* We failed to check for an allocation failure
* We used `NumPyOS_ascii_strtold` but did not check all possible error cases for that function. In particular, if the parsed longdouble ends before the end of the string, like with the test case that has a trailing NULL.

This fixes that by improving the error checking. Also adds test cases for converting various invalid strings to built-in numeric dtypes.


charris:
@ngoldbaum I note that 3.15t hung on the first try here.

PR #31682, opened 2026-06-18
BUG: avoid deadlocks using NpyString API

ngoldbaum:
I'm working on a design for a new ByteStringDType and have come across some design issues with StringDType. In particular, it turns out that there are a number of places we're susceptible to possible deadlocks with CPython or between StringDType instances.

For `NpyString_acquire_allocators` and `NpyString_release_allocators`, the fix is to sort the mutexes by address before locking. I added tests for ufuncs that need 2, 3, and 4 allocators as well as one that uses a duplicate allocator that all deadlock on `main` and don't with the fixes here.

There are also a few more fixes:

* StringDType setitem was holding an allocator lock while allowing arbitrary Python code to execute. I added a reentrant example in a test that deadlocks right now.
* There are also issues in several casts with lock discipline and not reliably separating sections of code that hole the allocator lock or use the CPython C API.
* Python 3.12 using `PyThread_type_lock` can deadlock with the GIL, so I added a code that checks if the GIL is held before trying to acquire the lock. This is the same deadlock-avoiding behavior as PyMutex.

There are still places that use the C API with the allocator lock held but I'm pretty sure there are no reentrancy issues in those places.

This was all done heavily with AI assistance. Please let me know if you spot anything you think is fishy because of that.


numpy/_core/tests/test_multithreading.py
```diff
@@ -466,3 +470,189 @@ def func(b):
...
+    ``reason`` if the child is killed.
+
+    """
+    source = "\n".join(
+        textwrap.dedent(inspect.getsource(helper)) for helper in helpers
+    )
+    source += "\n" + textwrap.dedent(inspect.getsource(workload))
+    script = f"{source}\n{workload.__name__}(*{args!r})\n"
+    try:
+        run_subprocess([sys.executable, "-c", script], timeout=timeout)
```
kumaraditya303: maybe use `faulthandler.dump_traceback_later` with timeout so that it dumps the traceback in case of deadlock?

kumaraditya303:
I think we should harden the error paths as well to not call APIs which require the GIL, I think it is fine as long as there is no re-entrancy but that relies on a Python 3.12+ behavior that gc is only triggered at eval breaker points.

numpy/_core/src/multiarray/stringdtype/casts.cpp
```diff
@@ -1984,7 +1989,56 @@ string_to_bytes(PyArrayMethod_Context *context, char *const data[],
...
+                (Py_ssize_t)invalid_ascii_pos,
+                (Py_ssize_t)(invalid_ascii_pos + 1),
+                "ordinal not in range(128)");
+
+        if (exc == NULL) {
+            Py_DECREF(str);
+            return -1;
+        }
+
+        PyErr_SetObject(PyExceptionInstance_Class(exc), exc);
```
kumaraditya303: I think you can use `PyErr_SetRaisedException` which is preferred on newer versions.

ngoldbaum:
> I think we should harden the error paths as well to not call APIs which require the GIL

I looked at this today and fixing this will require a pretty substantial second pass. I think I'd like to do it in a followup since this is already pretty big.

I asked Sebastian to look this over during the triage meeting.

numpy/_core/src/multiarray/stringdtype/casts.cpp
seberg: This API is irrelevant, since there is no attached traceback yet.

numpy/_core/tests/test_multithreading.py
```diff
@@ -466,3 +470,226 @@ def func(b):
             x.__buffer__(flags[i % 2])
 
     run_threaded(func, max_workers=8, pass_barrier=True)
+
+
+def assert_no_deadlock(workload, *, args=(), helpers=(), timeout=30,
```
seberg, line 483: Part of me does wonder if we shouldn't just use `@pytest.mark.timeout(...)`.
Of course that doesn't:
* Gracefully give a traceback (without exiting the whole test run at least, which it does on windows or with `mode="thread"`).
* Allow the `STALL` logic to not wait the whole `timeout` (I guess that may be if that is important for local debugging of complex tests)

(I mean, for all I know a few other multi-threading tests in theory also check for deadlocks :)).

numpy/_core/tests/test_stringdtype.py
```diff
@@ -892,6 +892,91 @@ def test_float_nan_cast_na_object():
...
+
+
+@pytest.mark.parametrize("typename", ["complex64", "complex128", "clongdouble"])
+@pytest.mark.parametrize("na_object", [np.nan, pd_NA],
+                         ids=["np.nan", "pandas.NA"])
+def test_cfloat_to_string_nan_na_not_missing(typename, na_object):
+    # complex inputs never map to a NaN-like na_object on cast -- a
+    # real NaN sentinel has no complex equivalent -- so NaN components are
+    # stringified rather than stored as the missing value. This matches both the
+    # historical behavior and stringdtype_setitem.
```
seberg: Hmmmm, I find this confusing.  So we are saying that float NaN is clearly a NaN-like object and we don't care if that gets converted into a `pandas.NA`.

But you are arguing that complex NaN doesn't fit into a set that contains both float NaN and `pandas.NA`?

I guess you are just testing existing behavior, but are we fixing the float path here as a bug fix?

numpy/_core/tests/test_stringdtype.py
```diff
@@ -892,6 +892,91 @@ def test_float_nan_cast_na_object():
...
+                                      "longdouble"])
+@pytest.mark.parametrize("na_object", [np.nan, float("nan"), pd_NA],
+                         ids=["np.nan", "float('nan')", "pandas.NA"])
+def test_float_to_string_nan_na_consistent(typename, na_object):
+    dt = StringDType(na_object=na_object)
+    arr = np.array([1.5, np.nan, -2.0], dtype=typename).astype(dt)
+    assert arr[0] == "1.5"
+    assert arr[2] == "-2.0"
+    # the NaN element is the missing value (returned by identity), not "nan"
+    assert arr[1] is na_object
```
seberg: I don't want to block it if you have a good reason, but it isn't obvious to me that this is a reasonable backport, feels like it could regress someone.

Did this show up in an actual use-case?

numpy/_core/src/multiarray/stringdtype/static_string.c
```diff
@@ -328,23 +358,41 @@ NpyString_acquire_allocators(size_t n_descriptors,
...
+    }
+
+    uintptr_t last_locked_addr = 0;
+    int have_last_locked_addr = 0;
+
+    for (;;) {
+        npy_string_allocator *next_allocator = NULL;
+        uintptr_t next_addr = UINTPTR_MAX;
+
+        for (size_t i=0; i<n_descriptors; i++) {
```
seberg:
```suggestion
        for (size_t i = 0; i < n_descriptors; i++) {
```
(feel free to ignore)

numpy/_core/src/multiarray/stringdtype/casts.cpp
```diff
@@ -1984,7 +1989,56 @@ string_to_bytes(PyArrayMethod_Context *context, char *const data[],
...
+        Py_DECREF(exc);
+        Py_DECREF(str);
+    }
```
seberg: OK, let's keep this.  Part of me thinks that there *must* be a nicer pattern; this "set up temporaries to use after the `goto`" feels really strange...  Maybe just in-line release + helper function, but I dunno...
(An inline release feels like it would only be nice with some C++ scoped helper that knows it has been cleaned up so it doesn't clean up twice...)

EDIT: I suppose I am thinking of a RAII helper that has an `.unlock()` and could even be passed on.  Although, it feels like there might be an even nicer approach, but I don't see it right now...
(Right now this is a half-hearted fix in one place in a sense, although, I have to say that re-entrance seems like an unlikely problem.)

seberg (commented):
Before approving it, I think I have to repeat the one question: This PR doesn't just avoid deadlocks in the API, it *also* changes behavior around NA objects.
I don't care too much about doing it in one PR in principle, but unlike the deadlock fix, changing that isn't an obvious backport to me.

That whole "test deadlocks" looks right. It feels overly clever/engineered, but fine, if it seems to add enough to you then OK.

numpy/_core/tests/test_stringdtype.py
ngoldbaum: I'll extract this into its own PR. IMO it's a bugfix and the old behavior is just wrong but you're right that it's conceivably a breaking change.

numpy/_core/src/multiarray/stringdtype/casts.cpp
ngoldbaum: I refactored it to use RAII helpers for cleanup and it's a lot nicer. Thanks for the suggestion.

numpy/_core/tests/test_stringdtype.py
ngoldbaum: Let's deal with this in a subsequent PR.

numpy/_core/tests/test_multithreading.py
ngoldbaum: I'd rather not make pytest-timeout triggering after 10 minutes so load-bearing, particularly for a test that exists solely to detect if a deadlock starts happening again. The test helper is fancy, sure, but it seems to work and it's a nice helper for future work on avoiding deadlocks.

ngoldbaum:
@seberg I responded to all your comments I think, could you give this another look?

numpy/_core/tests/test_multithreading.py
seberg: Sorry, but I am confused.  I don't care too much, I don't love the idea of killing the test suite on deadlock (I just also am not sure I love the rather complex test harness).
But there is no 10 minute delay, I was asking about a per-test timeout guard which is also 30 seconds so relatively similar to what you go there.

seberg (approved):
Thanks for splitting and doing the release thing!  LGTM, I think you like your test harness, so let's not worry about whether we absolutely need or not. It's local to our tests after all.

PR #31687, opened 2026-06-19
MAINT: Prefer `PyTuple_Pack` over `Py_BuildValue`

hpkfft:
### PR summary

In a micro-benchmark that builds the tuple `(0, 1)` from two existing PyObjects, `PyTuple_Pack(2, zero, one)` is twice as fast as `Py_BuildValue("(OO)", zero, one)`, with runtimes of 17.7 nsec and 36.1 nsec per loop, respectively, as measured by `timeit` with Python 3.13.5 on x86_64 GNU/Linux.  (It makes sense that it's faster not having to parse a format string.)

So, I thought, why not use it in NumPy?

#### AI Disclosure

No AI tools used

### DIFF
--- a/numpy/_core/src/multiarray/npy_static_data.c
+++ b/numpy/_core/src/multiarray/npy_static_data.c
@@ -180,8 +180,7 @@ initialize_static_globals(void)
         return -1;
     }
 
-    npy_static_pydata.kwnames_is_copy =
-            Py_BuildValue("(O)", npy_interned_str.copy);
+    npy_static_pydata.kwnames_is_copy = PyTuple_Pack(1, npy_interned_str.copy);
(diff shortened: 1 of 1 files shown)


hpkfft:
The current version of [Intel SDE](https://www.intel.com/content/www/us/en/developer/articles/tool/software-development-emulator.html) is 10.8.

seberg (approved):
Thanks, I'll put it in because I have a very slight preference for this API.  But this code has *literally* zero impact on performance, so that is the only reason.

In fact in a place where performance impact matters a huge amount, even `PyTuple_Pack(...)` isn't as fast because `vargs` handling is complex packing/unpacking.

PR #31688, opened 2026-06-19
BUG: Fix signed integer overflow in datetime.c

### LINKED ISSUE #31686: BUG: Signed integer overflow in datetime.c:days_to_yearsdays on near-minimum np.datetime64
### Describe the issue:

There is a signed integer overflow in `numpy/_core/src/multiarray/datetime.c` when converting a `datetime64` object near `INT64_MIN` to a Python object via `.item()`.

Specifically, on this line: 

https://github.com/numpy/numpy/blob/7dfe67ef29c771c127130e1b72e243b7ff804360/numpy/_core/src/multiarray/datetime.c#L171

```python
import numpy as np

info = np.iinfo(np.int64)
dt64 = np.datetime64(info.min + 1, "D")

dt64.item()
# ../../numpy/_core/src/multiarray/datetime.c:171:31: runtime error: signed integer overflow:
# -9223372036854775807 - 10957 cannot be represented in type 'npy_int64' (aka 'long')
```

### Error message:

```shell
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../../numpy/_core/src/common/cblasfuncs.c:548:9 
../../numpy/_core/src/multiarray/datetime.c:171:31: runtime error: signed integer overflow: -9223372036854775807 - 10957 cannot be represented in type 'npy_int64' (aka 'long')
    #0 0x7b5de65e29e7 in days_to_yearsdays /home/alvaro/numpy/build/cp313/../../numpy/_core/src/multiarray/datetime.c:171:31
    #1 0x7b5de65a9411 in set_datetimestruct_days /home/alvaro/numpy/build/cp313/../../numpy/_core/src/multiarray/datetime.c:265:17
    #2 0x7b5de65adefc in NpyDatetime_ConvertDatetime64ToDatetimeStruct /home/alvaro/numpy/build/cp313/../../numpy/_core/src/multiarray/datetime.c:486:13
    #3 0x7b5de65ccaf1 in convert_datetime_to_pyobject /home/alvaro/numpy/build/cp313/../../numpy/_core/src/multiarray/datetime.c:2863:9
    #4 0x7b5de5afd18d in DATETIME_getitem /home/alvaro/numpy/build/cp313/../../numpy/_core/src/multiarray/arraytypes.c.src:1182:12
    #5 0x7b5de672bf39 in PyArray_GETITEM /home/alvaro/numpy/build/cp313/../../numpy/_core/src/multiarray/dtypemeta.h:304:12
    #6 0x7b5de672bdd7 in PyArray_MultiIndexGetItem /home/alvaro/numpy/build/cp313/../../numpy/_core/src/multiarray/item_selection.c:3161:12
    #7 0x7b5de67c9f15 in array_toscalar /home/alvaro/numpy/build/cp313/../../numpy/_core/src/multiarray/methods.c:748:12
    #8 0x00000181878e in cfunction_call (/home/alvaro/.local/share/uv/python/cpython-3.13.9-linux-x86_64-gnu/bin/python3.13+0x181878e)
    #9 0x000001851a76 in _PyObject_Call (/home/alvaro/.local/share/uv/python/cpython-3.13.9-linux-x86_64-gnu/bin/python3.13+0x1851a76)
    #10 0x7b5de61ef314 in gentype_generic_method /home/alvaro/numpy/build/cp313/../../numpy/_core/src/multiarray/scalartypes.c.src:143:15
    #11 0x7b5de61eae2f in gentype_item /home/alvaro/numpy/build/cp313/../../numpy/_core/src/multiarray/scalartypes.c.src:2123:12
    #12 0x000001ad27c1 in method_vectorcall_VARARGS.llvm.16033483530146946230.warm (/home/alvaro/.local/share/uv/python/cpython-3.13.9-linux-x86_64-gnu/bin/python3.13+0x1ad27c1)
    #13 0x000001823185 in _PyEval_EvalFrameDefault (/home/alvaro/.local/share/uv/python/cpython-3.13.9-linux-x86_64-gnu/bin/python3.13+0x1823185)
    #14 0x0000018aa113 in PyEval_EvalCode (/home/alvaro/.local/share/uv/python/cpython-3.13.9-linux-x86_64-gnu/bin/python3.13+0x18aa113)
    #15 0x0000018a5239 in run_eval_code_obj (/home/alvaro/.local/share/uv/python/cpython-3.13.9-linux-x86_64-gnu/bin/python3.13+0x18a5239)
    #16 0x0000018dc182 in run_mod.llvm.12468226955389376524 (/home/alvaro/.local/share/uv/python/cpython-3.13.9-linux-x86_64-gnu/bin/python3.13+0x18dc182)
    #17 0x000001a08446 in pyrun_file (/home/alvaro/.local/share/uv/python/cpython-3.13.9-linux-x86_64-gnu/bin/python3.13+0x1a08446)
    #18 0x000001a07b36 in _PyRun_SimpleFileObject (/home/alvaro/.local/share/uv/python/cpython-3.13.9-linux-x86_64-gnu/bin/python3.13+0x1a07b36)
    #19 0x000001a0764f in _PyRun_AnyFileObject (/home/alvaro/.local/share/uv/python/cpython-3.13.9-linux-x86_64-gnu/bin/python3.13+0x1a0764f)
    #20 0x000001a075f5 in pymain_run_file_obj (/home/alvaro/.local/share/uv/python/cpython-3.13.9-linux-x86_64-gnu/bin/python3.13+0x1a075f5)
    #21 0x000001a074d9 in pymain_run_file (/home/alvaro/.local/share/uv/python/cpython-3.13.9-linux-x86_64-gnu/bin/python3.13+0x1a074d9)
    #22 0x000001997e91 in Py_RunMain (/home/alvaro/.local/share/uv/python/cpython-3.13.9-linux-x86_64-gnu/bin/python3.13+0x1997e91)
    #23 0x000001997b44 in pymain_main.llvm.12760885333831575200 (/home/alvaro/.local/share/uv/python/cpython-3.13.9-linux-x86_64-gnu/bin/python3.13+0x1997b44)
    #24 0x00000199792c in main (/home/alvaro/.local/share/uv/python/cpython-3.13.9-linux-x86_64-gnu/bin/python3.13+0x199792c)
    #25 0x7f5dfbef3680 in __libc_start_call_main (/lib64/libc.so.6+0x3680) (BuildId: 18472003bbf1c5f098a09b5016b9b8bd4c7c59f0)
    #26 0x7f5dfbef3797 in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x3797) (BuildId: 18472003bbf1c5f098a09b5016b9b8bd4c7c59f0)
    #27 0x0000019fff68 in _start (/home/alvaro/.local/share/uv/python/cpython-3.13.9-linux-x86_64-gnu/bin/python3.13+0x19fff68)

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../../numpy/_core/src/multiarray/datetime.c:171:31
```

2.6.0.dev0+git20260618.7dfe67e
3.13.9 (main, Oct 28 2025, 12:10:42) [Clang 20.1.4 ]

### Runtime Environment:

_No response_

This was originally found in https://github.com/pandas-dev/pandas/pull/65831.

AnkitAhlawat7742:
### PR summary

PR fixes a signed integer overflow in `numpy/_core/src/multiarray/datetime.c `that occurs when converting datetime64 values near INT64_MIN to Python objects via `.item().the days_to_yearsdays() `function performs `(*days_) - (365*30 + 7)` which causes undefined behavior when `*days_ is near INT64_MIN` . this solution will Reordered arithmetic operations to avoid overflow

NA

### DIFF
--- a/numpy/_core/src/multiarray/datetime.c
+++ b/numpy/_core/src/multiarray/datetime.c
@@ -167,12 +167,27 @@ static npy_int64
 days_to_yearsdays(npy_int64 *days_)
 {
     const npy_int64 days_per_400years = (400*365 + 100 - 4 + 1);
+    const npy_int64 days_offset = (365*30 + 7);
     /* Adjust so it's relative to the year 2000 (divisible by 400) */
-    npy_int64 days = (*days_) - (365*30 + 7);
+    npy_int64 days;
     npy_int64 year;
 
+    if (*days_ < NPY_MIN_INT64 + days_offset) {
+        /* Extract 400-year cycles first to reduce magnitude */
+        days = *days_;
+        year = 400 * extract_unit_64(&days, days_per_400years);
+        days -= days_offset;
+        if (days < 0) {
+            days += days_per_400years;
+            year -= 400;
+        }
+    }
+    else {
+        days = (*days_) - days_offset;
+        year = 400 * extract_unit_64(&days, days_per_400years);
+    }
+
     /* Break down the 400 year cycle to get the year and day within the year */
-    year = 400 * extract_unit_64(&days, days_per_400years);
 
     /* Work out the year/day within the 400 year cycle */
     if (days >= 366) {
--- a/numpy/_core/src/multiarray/datetime.c
+++ b/numpy/_core/src/multiarray/datetime.c
@@ -174,13 +174,9 @@ days_to_yearsdays(npy_int64 *days_)
 
     if (*days_ < NPY_MIN_INT64 + days_offset) {
         /* Extract 400-year cycles first to reduce magnitude */
-        days = *days_;
+        days = *days_ + (days_per_400years - days_offset);
         year = 400 * extract_unit_64(&days, days_per_400years);
-        days -= days_offset;
-        if (days < 0) {
-            days += days_per_400years;
-            year -= 400;
-        }
+        year -= 400;
     }
     else {
         days = (*days_) - days_offset;
--- a/numpy/_core/tests/test_datetime.py
+++ b/numpy/_core/tests/test_datetime.py
@@ -1268,18 +1268,12 @@ class TestDateTime:
     def test_datetime64_item_int64_min_edge_case(self):
         info = np.iinfo(np.int64)
 
-        # Test INT64_MIN + 1
-        dt64 = np.datetime64(info.min + 1, "D")
-        result = dt64.item()
-        # For dates outside Python's datetime range (year 1-9999),
-        # NumPy returns the raw integer value
-        assert isinstance(result, (int, datetime.date))
-
-        # Test a few more extreme negative values
         for offset in [1, 2, 100, 1000, 10000]:
             dt64 = np.datetime64(info.min + offset, "D")
             result = dt64.item()
-            assert isinstance(result, (int, datetime.date))
+            # expected integer value
+            assert result == info.min + offset
+            assert isinstance(result, int)
 
     def test_pyobject_roundtrip(self):
         # All datetime types should be able to roundtrip through object
--- a/numpy/_core/tests/test_datetime.py
+++ b/numpy/_core/tests/test_datetime.py
@@ -1265,6 +1265,22 @@ class TestDateTime:
         assert small * np.int64(7) == np.timedelta64(21, "s")
         assert np.int64(7) * small == np.timedelta64(21, "s")
 
+    def test_datetime64_item_int64_min_edge_case(self):
+        info = np.iinfo(np.int64)
+
+        # Test INT64_MIN + 1
+        dt64 = np.datetime64(info.min + 1, "D")
+        result = dt64.item()
+        # For dates outside Python's datetime range (year 1-9999),
+        # NumPy returns the raw integer value
+        assert isinstance(result, (int, datetime.date))
+
+        # Test a few more extreme negative values
+        for offset in [1, 2, 100, 1000, 10000]:
+            dt64 = np.datetime64(info.min + offset, "D")
+            result = dt64.item()
+            assert isinstance(result, (int, datetime.date))
+
     def test_pyobject_roundtrip(self):
         # All datetime types should be able to roundtrip through object
         a = np.array([0, 0, 0, 0, 0, 0, 0, 0, 0,


numpy/_core/src/multiarray/datetime.c
```diff
@@ -167,12 +167,27 @@ static npy_int64
...
     const npy_int64 days_per_400years = (400*365 + 100 - 4 + 1);
+    const npy_int64 days_offset = (365*30 + 7);
     /* Adjust so it's relative to the year 2000 (divisible by 400) */
-    npy_int64 days = (*days_) - (365*30 + 7);
+    npy_int64 days;
     npy_int64 year;
 
+    if (*days_ < NPY_MIN_INT64 + days_offset) {
+        /* Extract 400-year cycles first to reduce magnitude */
+        days = *days_;
```
MaanasArora: I think you could just add `days_per_400years` here, then we lose the `days < 0` check (just always `year -= 400`). Since `days_` is so negative this should be safe.

To make this read even simpler, you could add `days_per_400years - days_offset`, then you don't need to subtract from days below.

numpy/_core/src/multiarray/datetime.c
```diff
@@ -167,12 +167,27 @@ static npy_int64
...
+        year = 400 * extract_unit_64(&days, days_per_400years);
+        days -= days_offset;
+        if (days < 0) {
+            days += days_per_400years;
+            year -= 400;
+        }
+    }
+    else {
```
MaanasArora, line 181: Hmm, if you simplify as above, we only have the identical  `extract_unit_64` line here; it might be nice to somehow lose this else branch. It could work if we get rid of `year -= 400` by initializing `year = 0` then `-400`. Just a nit though!

Edit: sorry, I don't think we should actually do this, feel free to ignore.

numpy/_core/tests/test_datetime.py
```diff
@@ -1265,6 +1265,22 @@ def test_arithmetic_valid_boundary(self):
         assert small * np.int64(7) == np.timedelta64(21, "s")
         assert np.int64(7) * small == np.timedelta64(21, "s")
 
+    def test_datetime64_item_int64_min_edge_case(self):
+        info = np.iinfo(np.int64)
+
+        # Test INT64_MIN + 1
+        dt64 = np.datetime64(info.min + 1, "D")
```
MaanasArora: This can be folded into the below for loop, already is there actually. (Then you could parametrize if you want, the test is not really that big though.)

numpy/_core/tests/test_datetime.py
```diff
@@ -1265,6 +1265,22 @@ def test_arithmetic_valid_boundary(self):
...
+        result = dt64.item()
+        # For dates outside Python's datetime range (year 1-9999),
+        # NumPy returns the raw integer value
+        assert isinstance(result, (int, datetime.date))
```
MaanasArora: Not sure this is helpful, given those are the only two plausible dtypes and the segfault tested would happen above. For correctness it would be much better to check the actual values equal if `int` at least. Not sure, is there any reason we are including `datetime.date` anyway? These all seem to be small enough values that we should expect `int`?

MaanasArora (commented):
Thanks, this would solve the issue I think! Just a refactor suggestion and should improve the test. That said, I actually couldn't reproduce this bug on main, which could be because C behavior with `iinfo` depends on the environment. Could you check if you can reproduce and if the test fails on main for you? (Fine if not I guess, but then we should maybe try to test this where it does...)

numpy/_core/src/multiarray/datetime.c
AnkitAhlawat7742: Thanks for the review. 
I agree, and I've updated the code accordingly.

numpy/_core/tests/test_datetime.py
AnkitAhlawat7742: Agree , Test case has been updated .

AnkitAhlawat7742:
> Thanks, this would solve the issue I think! Just a refactor suggestion and should improve the test. That said, I actually couldn't reproduce this bug on main, which could be because `iinfo` depends on the environment. Could you check if you can reproduce and if the test fails on main for you? (Fine if not I guess, but then we should maybe try to test this where it does...)

Thanks for the review ...!!
For testing, I've informed the bug reporter(seems like it is happening in panda's side ) to verify it from their side as well, so we can confirm that the issue has been fully resolved.

numpy/_core/src/multiarray/datetime.c
```diff
@@ -167,12 +167,23 @@ static npy_int64
...
 {
     const npy_int64 days_per_400years = (400*365 + 100 - 4 + 1);
+    const npy_int64 days_offset = (365*30 + 7);
     /* Adjust so it's relative to the year 2000 (divisible by 400) */
-    npy_int64 days = (*days_) - (365*30 + 7);
+    npy_int64 days;
     npy_int64 year;
 
+    if (*days_ < NPY_MIN_INT64 + days_offset) {
+        /* Extract 400-year cycles first to reduce magnitude */
```
MaanasArora, line 176:
```suggestion
        /* Add 400-year cycle to ensure days does not overflow */
```

Seems clearer!

MaanasArora (approved):
Thanks, sorry, my previous comment was a bit misleading - this should be tested with UBSan, in which case this _does_ fail on main for me, so I think CI will catch it and the test looks good. Would be nice to check with pandas but aside from fixing the comment this looks good to me!

Edit: sorry, UBSan actually doesn't run on CI it seems. Not sure here, I'll wait for maintainer review for other opinions on testing.

numpy/_core/src/multiarray/datetime.c
```diff
@@ -167,12 +167,23 @@ static npy_int64
 days_to_yearsdays(npy_int64 *days_)
 {
     const npy_int64 days_per_400years = (400*365 + 100 - 4 + 1);
+    const npy_int64 days_offset = (365*30 + 7);
     /* Adjust so it's relative to the year 2000 (divisible by 400) */
```
ngoldbaum, line 171: this comment is orphaned now, move it above the if statement?

ngoldbaum (commented):
Thanks, I agree that this fixes the bug.

However, doesn't the exact same bug happen elsewhere?

For example:

https://github.com/numpy/numpy/blob/ec56e828ca347ab14088413e23135fd983e53bb6/numpy/_core/src/multiarray/datetime.c#L313
https://github.com/numpy/numpy/blob/ec56e828ca347ab14088413e23135fd983e53bb6/numpy/_core/src/multiarray/datetime.c#L317
https://github.com/numpy/numpy/blob/ec56e828ca347ab14088413e23135fd983e53bb6/numpy/_core/src/multiarray/datetime.c#L330
https://github.com/numpy/numpy/blob/ec56e828ca347ab14088413e23135fd983e53bb6/numpy/_core/src/multiarray/datetime.c#L337-L399

Can't all of these mathematical operations on signed integers possibly overflow?

I would much rather we solve this generically rather than a one-off fix for this one particular issue.

Another way to fix this is to do intermediate calculations in a higher precision integer. We have a 128 bit integer implementation in NumPy for this purpose. That said, only the `year` is int64, everything else is `int32` so you could cast to int64 cheaply for those cases.

> Edit: sorry, UBSan actually doesn't run on CI it seems. Not sure here, I'll wait for maintainer review for other opinions on testing.

Yes it does: https://github.com/numpy/numpy/actions/runs/27960070587/job/82738842115?pr=31688

MaanasArora:
> Yes it does:

Ah sorry again, I just noticed that the test in this PR passes on main even on the UBSAN run (I accidentally opened https://github.com/numpy/numpy/pull/31713 and was then surprised). So we may need a stronger test. But I suppose we should be taking the generic approach.

Alvaro-Kothe:
> Yes it does: https://github.com/numpy/numpy/actions/runs/27960070587/job/82738842115?pr=31688

It seems that NumPy suppresses signed integer overflow in CI.

https://github.com/numpy/numpy/blob/ccfa0572e3fb8e2e59e82122460b7c87dff77c73/.github/workflows/compiler_sanitizers.yml#L82

https://github.com/numpy/numpy/blob/ccfa0572e3fb8e2e59e82122460b7c87dff77c73/tools/ci/ubsan_suppressions_arm64.txt#L8

ngoldbaum:
Yeah, we want to eventually delete all those suppressions by fixing these bugs!

AnkitAhlawat7742:
Thank you for the feedback, @ngoldbaum and  @MaanasArora 
 can we take this fix for the issue which is raised by @Alvaro-Kothe ? and I can create a follow-up PR for a comprehensive overflow 
what you guys say ?

ngoldbaum:
Please go ahead. Also for future notice, you don't need to ask to take up PRs like that. It just causes delays.

PR #31695, opened 2026-06-19
MAINT: Fix emscripten wheel build.

charris:
Emscripten currently fails to build wheels using
cibuildwheel >= 4.0.0.

AI suggested possible fixes.


charris:
Putting this in because we need it for newer cibuildwheel. I killed loongarch because it was stuck compiling uv. We need to fix that.

PR #31696, opened 2026-06-19
MAINT: Avoid compiling uv for loongarch

charris:
The uv package is pip installed, add `--only-binary` flag. That will allow upgrades, but only when they are
compiled. Rust compilation is very slow in emulation. The reason this is needed now is that uv had a new
release today 6/19/2026.

Used AI to discuss the problem and check the loongarch pip repo.


charris:
Self merging. This saves about 2 hours of compile time.

PR #31699, opened 2026-06-20
BUG: fix out array leak in reduceat and accumulate when dtype resolution fails

### LINKED ISSUE #31691: BUG: out reference leak in PyUFunc_Accumulate and PyUFunc_Reduceat on type resolution failure
## Bug Description

In both `PyUFunc_Accumulate` and `PyUFunc_Reduceat`, the `out` argument is INCREF'd early in the function, but when `reducelike_promote_and_resolve` fails immediately after, the code does `return NULL` directly — bypassing the `fail` label where `Py_XDECREF(out)` is located. This leaks the reference to `out`.

## Location

1. `numpy/_core/src/umath/ufunc_object.c`, `PyUFunc_Accumulate`, lines ~2745-2753
2. `numpy/_core/src/umath/ufunc_object.c`, `PyUFunc_Reduceat`, lines ~3168-3176

## Code

Both functions follow the same pattern:

```c
/* Take a reference to out for later returning */
Py_XINCREF(out);                          // line ~2745 / ~3168

PyArray_Descr *descrs[3];
PyArrayMethodObject *ufuncimpl = reducelike_promote_and_resolve(ufunc,
        arr, out, signature, NPY_TRUE, descrs, NPY_UNSAFE_CASTING,
        "accumulate");                     // or "reduceat"
if (ufuncimpl == NULL) {
    return NULL;                           // BUG: out's INCREF never balanced
}
```

Further down the function, the `fail:` label correctly does `Py_XDECREF(out)`, but the early `return NULL` bypasses it.

## Impact

Every time `accumulate()` or `reduceat()` is called with an `out` argument and type resolution fails (e.g., incompatible dtypes), the `out` array leaks a reference. Over repeated calls, this prevents the array from being garbage collected.

## Suggested Fix

Change `return NULL` to `goto fail` in both locations:

```c
if (ufuncimpl == NULL) {
    goto fail;
}
```

This ensures the `fail` label's cleanup runs, which includes `Py_XDECREF(out)`.

## Environment

- NumPy: current main branch
- Affected: `numpy.ufunc.accumulate()` and `numpy.ufunc.reduceat()` with `out` argument on type resolution failure

ikrommyd:
Closes https://github.com/numpy/numpy/issues/31691

### PR summary
In both of those cases `out` reference count is incremented before `reducelike_promote_and_resolve` is called but if the dtype resolution fails, it's not decremented. One can see it with a snippet like the following:
```python
>>> import sys, numpy as np
... arr = np.arange(3)
... out = np.empty(3, dtype="U5")
...
... before = sys.getrefcount(out)
... for _ in range(10000):
...     try:
...         np.add.accumulate(arr, out=out)
...     except Exception:
...         pass
... print(sys.getrefcount(out) - before)
...
10000 # with the fix, we get 0 here
```
Let me know if you think the above should be added as a test. I found it a trivial fix and not worthy of a specialized test personally.

N/A

#### AI Disclosure
No AI

### DIFF
--- a/numpy/_core/src/umath/ufunc_object.c
+++ b/numpy/_core/src/umath/ufunc_object.c
@@ -2749,6 +2749,7 @@ PyUFunc_Accumulate(PyUFuncObject *ufunc, PyArrayObject *arr, PyArrayObject *out,
             arr, out, signature, NPY_TRUE, descrs, NPY_UNSAFE_CASTING,
             "accumulate");
     if (ufuncimpl == NULL) {
+        Py_XDECREF(out);
         return NULL;
     }
 
@@ -3172,6 +3173,7 @@ PyUFunc_Reduceat(PyUFuncObject *ufunc, PyArrayObject *arr, PyArrayObject *ind,
             arr, out, signature, NPY_TRUE, descrs, NPY_UNSAFE_CASTING,
             "reduceat");
     if (ufuncimpl == NULL) {
+        Py_XDECREF(out);
         return NULL;
     }
 
--- a/numpy/_core/tests/test_ufunc.py
+++ b/numpy/_core/tests/test_ufunc.py
@@ -2733,6 +2733,18 @@ class TestUfunc:
         with pytest.raises(ValueError, match="(shape|size)"):
             np.add.accumulate(arr, out=out)
 
+    def test_reduceat_and_accumulate_out_dtype_resolution_failure(self):
+        # gh-31691: the out= error path leaked a reference to out when the
+        # ufunc dtype resolution failed (no matching loop for the out dtype).
+        arr = np.arange(3)
+        out = np.empty(3, dtype="U5")  # no add loop resolves to this
+
+        with pytest.raises(np._core._exceptions._UFuncNoLoopError):
+            np.add.reduceat(arr, [0, 1, 2], out=out)
+
+        with pytest.raises(np._core._exceptions._UFuncNoLoopError):
+            np.add.accumulate(arr, out=out)
+
     @pytest.mark.parametrize('out_shape',
                              [(), (1,), (3,), (1, 1), (1, 3), (4, 3)])
     @pytest.mark.parametrize('keepdims', [True, False])


MaanasArora (approved):
Thanks, LGTM! (All other returns goto `fail` or `finish`.)

Pretty sure we don't need a test for this as well!

ikrommyd:
Yeah what one can do is also `PyArray_Descr *descrs[3] = {NULL, NULL, NULL};` and `goto fail;` but that breaks the pattern that everywhere else in this file the descriptor array is left uninitialized before passed to the dtype resolver so I liked the current pattern more.

MaanasArora:
The descriptor array is set up for the ufunc branch, so I agree, this seems more natural!

seberg:
I agree we shouldn't be too strict about tests here, but it should be pretty easy to write one that hits this, right?
(I wouldn't worry about actually testing for the refcount leak, I guess for strings it may even get tested by the leak sanitizer/etc.  Mainly ensure we have coverage.)

But if not agreed, we can also just put it in (and I will in a bit).  This bug leak looks very theoretical (i.e. not going to happen in production code).

seberg (approved):

ikrommyd:
Sorry but I didn't understand what you meant with your comment here. The first sentence feels like is being contradicted by "I wouldn't worry about actually testing for the refcount leak" to me. What would you want to test here? The ref count is the leak no? Are you talking about a test that would just be caught by a sanitizer instead of asserting on the actual ref count number? I feel like the snippet in the PR description would be the test.

seberg:
I guess I would prefer to be sure to cover the error path at least.  Testing for the leak is nice, but I am more willing to ignore it (but if you test with strings failing here, then we should already be testing for leaks implicitly).

(But yes, I am only saying that because I don't want us to slip too casually into the "hard to test, let's skip it" territory.)

MaanasArora:
There are a few different bugs related to this, but this one seems only about the leak right @seberg? There doesn't seem to be any other failure to detect...

seberg:
Sorry, yeah, I should have looked closer.  Any test that covers `out=` and fails would work (it doesn't have to check refcounts explicitly).
But also yeah, we do have few but real tests _without_ `out=` passed in.

Anyway, if you add one I am happy, if not I'll put it in tomorrow.

ikrommyd:
LMK if what I added is what you had in mind. Indeed I wouldn't assert on the ref count either cause I think I've had cases in the past where a change in python version causes ref count assertions to fail in tests.

Edit: the linting failure is not in code I'm touching here? What?

seberg (approved):
Thanks, sorry for being obnoxious about adding tests :).

The lint failure, I'll ignore it: Not quite sure but the cython lint may be running process wide and maybe they had a release or so?  (We'll have to fix it if it shows up generally, though.)

ikrommyd:
https://github.com/numpy/numpy/pull/31711 I think Joren's on it.

ikrommyd:
@seberg since you added the backport candidate label in https://github.com/numpy/numpy/pull/31731, I guess it can go here too no?

Issue #31678, opened 2026-06-18
BUG: typecode reference leak in array_setstate on 4 error paths (methods.c)


Fuyugithub:
## Bug Description

In `numpy/_core/src/multiarray/methods.c`, after `Py_INCREF(typecode)` at line 2100, four early `return NULL` paths skip the matching `Py_DECREF(typecode)`, leaking the reference.

## Location

`numpy/_core/src/multiarray/methods.c:2100-2131`

## Code

```c
Py_INCREF(typecode);        // line 2100: refcount incremented

if (nd < 0) {
    return NULL;            // line 2103: typecode LEAKED
}
// ...
if (dimensions[i] < 0) {
    return NULL;            // line 2118: typecode LEAKED
}
// ...
if (overflow) {
    PyErr_NoMemory();
    return NULL;            // line 2123: typecode LEAKED
}
// ...
if (!PyList_Check(rawdata)) {
    return NULL;            // line 2131: typecode LEAKED
}
```

## Impact

Every failed ndarray unpickling operation leaks the `typecode` reference. This affects `pickle.loads()` / `numpy.loads()` on corrupted or malformed ndarray pickle data.

## Suggested Fix

Replace all early `return NULL` with `goto fail` where `fail` does `Py_DECREF(typecode)`:

```c
fail:
    Py_DECREF(typecode);
    return NULL;
```

## Environment

- NumPy version: current main
- Affected: `ndarray.__setstate__()` (unpickling)

ngoldbaum:
Fixed by #31701

PR #31701, opened 2026-06-20
BUG: Refactor error handling in array_setstate to prevent typecode leak

whyvineet:
### PR summary
This fixes the reference-count imbalance reported in #31678 

#### AI Disclosure
No AI tools used

### DIFF
--- a/numpy/_core/src/multiarray/methods.c
+++ b/numpy/_core/src/multiarray/methods.c
@@ -2100,7 +2100,7 @@ array_setstate(PyArrayObject *self, PyObject *args)
     Py_INCREF(typecode);
     nd = PyArray_IntpFromSequence(shape, dimensions, NPY_MAXDIMS);
     if (nd < 0) {
-        return NULL;
+        goto fail_typecode;
     }
     /*
      * We should do two things here:
@@ -2115,12 +2115,13 @@ array_setstate(PyArrayObject *self, PyObject *args)
         if (dimensions[i] < 0) {
             PyErr_SetString(PyExc_TypeError,
                     "impossible dimension while unpickling array");
-            return NULL;
+            goto fail_typecode;
         }
         overflowed = npy_mul_sizes_with_overflow(
                 &nbytes, nbytes, dimensions[i]);
         if (overflowed) {
-            return PyErr_NoMemory();
+            PyErr_NoMemory();
+            goto fail_typecode;
         }
     }
 
@@ -2128,7 +2129,7 @@ array_setstate(PyArrayObject *self, PyObject *args)
         if (!PyList_Check(rawdata)) {
             PyErr_SetString(PyExc_TypeError,
                             "object pickle not returning list");
-            return NULL;
+            goto fail_typecode;
         }
     }
     else {
@@ -2149,20 +2150,17 @@ array_setstate(PyArrayObject *self, PyObject *args)
         if (!PyBytes_Check(rawdata)) {
             PyErr_SetString(PyExc_TypeError,
                             "pickle not returning string");
-            Py_DECREF(rawdata);
-            return NULL;
+            goto fail_rawdata;
         }
 
         if (PyBytes_AsStringAndSize(rawdata, &datastr, &len) < 0) {
-            Py_DECREF(rawdata);
-            return NULL;
+            goto fail_rawdata;
         }
 
         if (len != nbytes) {
             PyErr_SetString(PyExc_ValueError,
                     "buffer size does not match array size");
-            Py_DECREF(rawdata);
-            return NULL;
+            goto fail_rawdata;
         }
     }
     /*
@@ -2286,6 +2284,12 @@ array_setstate(PyArrayObject *self, PyObject *args)
     PyArray_UpdateFlags(self, NPY_ARRAY_UPDATE_ALL);
 
     Py_RETURN_NONE;
+
+fail_rawdata:
+    Py_DECREF(rawdata);
+fail_typecode:
+    Py_DECREF(typecode);
+    return NULL;
 }
 
 /*NUMPY_API*/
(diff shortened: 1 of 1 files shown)


numpy/_core/src/multiarray/methods.c
da-woods: This seems like it might still be leaking `typecode`

da-woods: and here?

da-woods: and here?

da-woods (commented):
This is a fairly complicated function so it isn't too easy to follow the flow and all the ownership. But I find it easier to reason about if you have fewer exit points and fewer paths through.

That typically means that an owned reference should either be `NULL` or set. References shouldn't change between being owned or unowned (like `rawdata` does).

So you'd have something like

```
PyObject *result = NULL, *typecode = NULL, *rawdata = NULL;

// ... code goes here ...

result = final_thing_that_creates_the_result();
end:
Py_XDECREF(typecode);  // either NULL, or owned, XDECREF is happy both ways
Py_XDECREF(rawdata);
return result; // either NULL or set...
```

and then you have one exit point and everything goes through that. Every return should go through the `end` (except possibly right at the start of the function before anything is owned).

This probably requires taking ownership of both `typecode` and `rawdata` with a `Py_INCREF` immediately after `PyArg_ParseTuple`.

whyvineet:
I agree @da-woods; the current `goto` chain does not handle the shared failure paths very well. Using a single exit point for cleanup would make ownership easier to understand.

From what I understand, the goal is to keep local references consistent, either owned or `NULL`, and send all error paths through a shared cleanup block with `Py_XDECREF`. This avoids switching between borrowed and owned references in the function.

I'll take a look at restructuring the ownership and cleanup logic along those lines.

numpy/_core/src/multiarray/methods.c
```diff
@@ -2220,40 +2222,45 @@ array_setstate(PyArrayObject *self, PyObject *args)
...
         else {
-            /* The handlers should never be called in this case */
+            /*
+             * Direct pointer path: PyArray_SetBaseObject steals the rawdata
+             * reference. NULL rawdata afterward so Py_XDECREF at the end
+             * label does not double-decref a stolen ref.
+             */
             fa->mem_handler = NULL;
             fa->data = datastr;
             if (PyArray_SetBaseObject(self, rawdata) < 0) {
```
da-woods: Most reference steeling functions (including `PyArray_SetBaseObject` it looks like) steal a reference on both success and failure to be consistent

numpy/_core/src/multiarray/methods.c
```diff
@@ -2222,40 +2222,45 @@ array_setstate(PyArrayObject *self, PyObject *args)
                 if (!(PyArray_ISEXTENDED(self) ||
                       PyArray_DESCR(self)->metadata ||
                       PyDataType_C_METADATA(PyArray_DESCR(self)))) {
-                    fa->descr = PyArray_DescrFromType(
+                    PyArray_Descr *swapped = PyArray_DescrFromType(
                                     PyArray_DESCR(self)->type_num);
+                    Py_DECREF(fa->descr);
+                    fa->descr = swapped;
```
da-woods: You might be better doing `Py_SETREF(fa->descr, swapped)` (because can in principle cause arbitrary code to execute so you're leaving it in an potentially invalid state while that happens).

I don't think it can actually happen here, but I think it's a good idea anyway

numpy/_core/src/multiarray/methods.c
```diff
@@ -2222,40 +2222,45 @@ array_setstate(PyArrayObject *self, PyObject *args)
...
+                    if (swapped->byteorder == NPY_BIG) {
+                        swapped->byteorder = NPY_LITTLE;
                     }
-                    else if (PyArray_DESCR(self)->byteorder == NPY_LITTLE) {
-                        PyArray_DESCR(self)->byteorder = NPY_BIG;
+                    else if (swapped->byteorder == NPY_LITTLE) {
+                        swapped->byteorder = NPY_BIG;
                     }
+                    Py_DECREF(fa->descr);
+                    fa->descr = swapped;
```
da-woods: And here with `Py_SETREF`

da-woods (commented):
A few small comments (only one of which is a real error I think) but to me that looks much more likely to work right.

numpy/_core/src/multiarray/methods.c
whyvineet: Okay... I’ll switch this assignment.

numpy/_core/src/multiarray/methods.c
whyvineet: Here as well...

numpy/_core/src/multiarray/methods.c
whyvineet: I’ll recheck the ownership contract here... and adjust the cleanup accordingly

whyvineet:
I looked at how `PyArray_SetBaseObject()` is implemented in arrayobject.c. This function takes ownership of the reference and decreases it if there is a failure, so the local cleanup does not need to handle that reference after the call.

I changed the code to transfer ownership before calling `PyArray_SetBaseObject()` by clearing the local `rawdata` reference. This keeps the single-exit cleanup safe.

Could you take another look, @da-woods? Does this match what you had in mind?

da-woods:
> I changed the code to transfer ownership before calling `PyArray_SetBaseObject()` by clearing the local `rawdata` reference. This keeps the single-exit cleanup safe.
> 
> Could you take another look, @da-woods? Does this match what you had in mind?

Yes that looks right to me.

seberg (approved):
Thanks @whyvineet and especially also @da-woods for reviewing.  This is a bit of a dark corner, the fix is I think in a sense meaningless (you would need a bad pickle to hit it and at that point anything is possible).

(We use `finish:` more often than `end:` but I don't feel like style nitpicking, let's get it in.)

PR #31702, opened 2026-06-20
ENH: Add np.EinsumExpression

### LINKED ISSUE #28394: ENH: Compiled einsum path support

`einsum_path` pre-analyzes the best contraction strategy for an Einstein sum; fine. `einsum` can accept that contraction strategy; also fine. I imagine that this works well in the case that there are very few, very large calls to `einsum`.

Where this breaks down is the case where the input is smaller or the calls are more frequent, making the overhead of the Python side of `einsum` dominant when `optimize` is not `False`. This effect can get really bad and easily overwhelm any benefit of the planned `einsum_path` contraction.

Another issue is that the planned contraction implies a specific subscript string, but that string needs to be passed again to `einsum`; this API allows for a mismatch between the subscript and optimize arguments.

I can imagine a few solutions, but the one that's currently my favourite is a simple loop evaluation -> AST -> compiled set of calls to `tensordot` and `c_einsum`. In my testing this solves all of the problems I described above.

For a vile patch-hack that demonstrates the concept and does function correctly but is probably not the best way to implement this (and also does not have `out` support):

```python
import ast
import logging
import typing

import numpy as np

logger = logging.getLogger('contractions')

if typing.TYPE_CHECKING:
    class Contract(typing.Protocol):
        def __call__(self, *operands: np.ndarray) -> np.ndarray: ...


def build_contraction(
    name: str, subscripts: str, *shapes: tuple[int, ...],
    optimize: typing.Literal['greedy', 'optimal'] = 'greedy',
) -> 'Contract':
    """
    This is a wrapper for the numpy.einsum and numpy.einsum_path pair of functions to de-duplicate
    some parameters. It represents an Einstein tensor contraction that has been pre-planned to
    perform as well as possible.

    If we simply call einsum() with no optimization, the performance is "good" because it skips
    straight to a naive but efficient c_einsum.

    If we call einsum_path and then pass the calculated contraction to einsum(), the performance is
    awful because there's a lot of overhead in the contraction stage loops. The only way to
    alleviate this is to have einsum() run the contraction once to sort out its stage logic and
    essentially cache that procedure so that actual runs simply make calls to c_einsum() or
    tensordot().
    """

    def get_index(a: np.ndarray | object) -> int:
        # simple .index() does not work when `a` is an ndarray
        return next(i for i, x in enumerate(operands) if x is a)

    def c_einsum(intermediate_subscripts: str, *intermediate_operands: np.ndarray, **kwargs) -> object:
        """
        This is a patched-in c_einsum that accepts intermediate arguments from the real einsum()
        but, rather than calling the actual sum, constructs a planned call for our compiled
        contraction.
        """
        if kwargs:
            raise NotImplementedError()
        # xi = c_einsum(intermediate_subscripts, *operands)
        body.append(ast.Assign(
            targets=[ast.Name(id=f'x{len(operands)}', ctx=ast.Store())],
            value=ast.Call(
                func=ast.Name(id='c_einsum', ctx=ast.Load()),
                args=[ast.Constant(value=intermediate_subscripts)]
                + [
                    ast.Name(id=f'x{get_index(o)}', ctx=ast.Load())
                    for o in intermediate_operands
                ],
            ),
        ))
        # This is a placeholder sentinel only. It's used to identify inputs to subsequent einsum or tensordot calls.
        out = object()
        operands.append(out)
        return out

    def tensordot(a: np.ndarray, b: np.ndarray, axes: tuple[tuple[int, ...], ...]) -> object:
        """
        This is a patched-in tensordot that accepts intermediate arguments from the real einsum()
        but, rather than calling the actual dot, constructs a planned call for our compiled
        contraction.
        """
        # xi = tensordot(a, b, axes=axes)
        body.append(ast.Assign(
            targets=[ast.Name(id=f'x{len(operands)}', ctx=ast.Store())],
            value=ast.Call(
                func=ast.Name(id='tensordot', ctx=ast.Load()),
                args=[
                    ast.Name(id=f'x{get_index(a)}', ctx=ast.Load()),
                    ast.Name(id=f'x{get_index(b)}', ctx=ast.Load()),
                ],
                keywords=[ast.keyword(arg='axes', value=ast.Constant(value=axes))],
            ),
        ))
        # This is a placeholder sentinel only. It's used to identify inputs to subsequent einsum or tensordot calls.
        out = object()
        operands.append(out)
        return out

    # These are needed for einsum_path() and einsum() to perform planning, but the actual content
    # doesn't matter, only the shape.
    fake_arrays = [np.empty(shape) for shape in shapes]

    # Build and describe the optimized contraction path plan.
    contraction, desc = np.einsum_path(subscripts, *fake_arrays, optimize=optimize)
    logger.debug('%s contraction: \n%s', name, desc)

    # This will be mutated by our monkeypatched functions and assumes that each element is
    # accessible by an indexed-like variable, i.e. x0, x1... in this list's order.
    operands: list[np.ndarray | object] = list(fake_arrays)

    # AST statements in the function body; will be mutated by the monkeypatched functions
    body = []
    # Preserve the old numerical backend functions
    old_c_einsum = np._core.einsumfunc.c_einsum
    old_tensordot = np._core.einsumfunc.tensordot
    try:
        # Monkeypatch in our substitute functions
        np._core.einsumfunc.c_einsum = c_einsum
        np._core.einsumfunc.tensordot = tensordot
        # Run the real einsum() with fake data and fake numerical backend.
        np.einsum(subscripts, *fake_arrays, optimize=contraction)
    finally:
        # Restore the old numerical backend functions
        np._core.einsumfunc.c_einsum = old_c_einsum
        np._core.einsumfunc.tensordot = old_tensordot

    # The AST function representation; will always have the same name
    func = ast.FunctionDef(
        name='contraction',
        args=ast.arguments(
            args=[ast.arg(arg=f'x{i}') for i in range(len(shapes))],
        ),
        body=body + [ast.Return(
            value=ast.Name(id=f'x{len(operands) - 1}', ctx=ast.Load()),
        )],
    )

    ast.fix_missing_locations(func)  # Add line numbers
    code = compile(                  # Compile to an in-memory anonymous module
        source=ast.Module(body=[func]), filename='<compiled tensor contraction>',
        mode='exec', flags=0, dont_inherit=1, optimize=2,
    )
    globs = {                        # Globals accessible to the module
        '__builtins__': {},          # no built-ins
        'c_einsum': old_c_einsum,    # real numerical backends
        'tensordot': old_tensordot,
    }
    eval(code, globs)            # Evaluate the code, putting the function in globs
    return globs['contraction']  # Return the function reference
```

Ijtihed:
### PR summary

Closes #28394

Added `np.EinsumExpression` which is a callable class that pre-compiles einsum contraction paths. On repeated calls with the same shapes it skips subscript parsing and path optimization entirely

```python
expr = np.EinsumExpression('ij,jk,kl->il', (10, 20), (20, 5), (5, 10))
for _ in range(1000):
    result = expr(a, b, c)  # reuses pre-computed path
```

This uses `as_strided` with zero strides for path computation to avoid large allocations. Supports `out`, `dtype`, `order`, `casting` kwargs. Includes single-step optimization that uses `c_einsum` directly when the contraction has only one step.

### Benchmark 

| Benchmark | `einsum(optimize='greedy')` | `EinsumExpression` | Speedup |
|---|---|---|---|
| matmul 4x4 | 26.3 us | 6.1 us | 4.33x |
| trace 4x4 | 27.3 us | 8.2 us | 3.32x |
| chain 20x30x40x20 | 151.5 us | 71.5 us | 2.12x |
| 5-operand chain | 190.5 us | 42.1 us | 4.52x |

Speedup mainly from eliminating repeated `einsum_path` computation and subscript parsing. Largest gains on small arrays and many-operand contractions where Python overhead matters the most

I can add release notes after this moves forward

#### AI Disclosure

Double checking my work + found some edge cases :)


jorenham:

Ijtihed:
Its a draft pr man I just opened it to continue work. It wasnt ready for review yet @jorenham

Ijtihed:
Thanks! Sorry for the confusion lol. I saw someone else do it this way. If this is not alright please let me know as well.

jorenham:
We indeed prefer contributors testing and developing locally before opening a branch. Because for one thing, us maintainers get a gh notification every time a commit gets pushed. And then there's also the CI costs.

Ijtihed:
> We indeed prefer contributors testing and developing locally before opening a branch. Because for one thing, us maintainers get a gh notification every time a commit gets pushed. And then there's also the CI costs.

Ah okay. No worries then Ill close the PR. I didnt know that so ill reopen once its ready. Thanks!

Ijtihed:
@jorenham ready for review now, thanks for the patience!

numpy/_core/einsumfunc.pyi
```diff
@@ -177,3 +177,88 @@ def einsum_path(
     optimize: _OptimizeKind = "greedy",
     einsum_call: Literal[False] = False,
 ) -> tuple[list[Any], str]: ...
+
+class EinsumExpression:
+    __slots__: ClassVar[tuple[str, ...]]
```
jorenham: `__slots__` should actually match the runtime slots, so without the `ClassVar` or `tuple[str, ...]`

numpy/_core/einsumfunc.pyi
```diff
@@ -177,3 +177,88 @@ def einsum_path(
...
+    def __init__(
+        self,
+        subscripts: str,
+        /,
+        *shapes: tuple[int, ...],
+        optimize: Literal[True, "greedy", "optimal"] = "greedy",
+    ) -> None: ...
+    def __repr__(self) -> str: ...
```
jorenham: no need for `__repr__`; it already exists on `builtins.object`
```suggestion
```

numpy/_core/einsumfunc.pyi
```diff
@@ -177,3 +177,88 @@ def einsum_path(
...
+    ) -> Any: ...
+    @overload
+    def __call__[OutT: NDArray[np.bool | np.number]](
+        self,
+        *operands: Any,
+        out: OutT,
+        casting: _CastingUnsafe,
+        dtype: _DTypeLikeObject | None = ...,
+        order: _OrderKACF = ...,
+    ) -> OutT: ...
```
jorenham, line 214: could you fill in the parameter detaults, instead of `= ...`?

jorenham: Can we narrow `*operands` down to `ArrayLike`, or would that not be broad enough?

jorenham: We only need one `out: OutT` overload, the last one for example.

Ijtihed: > Can we narrow `*operands` down to `ArrayLike`, or would that not be broad enough?

the operands go through asanyarray() so I agree ArrayLike would be the correct bound. I matched the einsum overload pattern ie _ArrayLikeComplex_co / _ArrayLikeObject_co for safe casting and Any for unsafe but I am indifferent about this

Ijtihed:
Should all be fixed now as requested. Thank you! :D

numpy/_core/einsumfunc.pyi
jorenham: `einsum` isn't the best example TBH :sweat_smile:

numpy/_core/einsumfunc.pyi
```diff
@@ -177,3 +177,56 @@ def einsum_path(
...
+    ) -> Any: ...
+    @overload
+    def __call__(
+        self,
+        *operands: Any,
+        casting: _CastingUnsafe,
+        dtype: _DTypeLikeComplex_co | _DTypeLikeObject | None = None,
+        out: None = None,
+        order: _OrderKACF = "K",
+    ) -> Any: ...
```
jorenham, line 205: can we narrow the return types to `ndarray` or `NDArray` here?

jorenham: Also, the first two overloads are redundant right now: the third overload is currently broader than the first two combined, and all three return the same `Any` type.

Ijtihed: > can we narrow the return types to `ndarray` or `NDArray` here?

It can return a scalar for 0-d results (e.g. `EinsumExpression('i,i->', (3,), (3,))` returns `np.float64`) since `c_einsum` calls `PyArray_Return`. Its the saame as unoptimized `einsum` so I don't know if `NDArray` would be fully accurate, but I trust your judgment more than mine here lol

jorenham: If it can return scalar than `Any` would indeed be better. I was just basing this on the docstring, which says it returns `ndarray`

Ijtihed: I opened a PR to fix the docstring asw + collapsed the tests in latest push. Thanks!

ngoldbaum:
This is a new feature and as such it needs to be brought to the mailing list. You'll also need to find a regular NumPy contributor who is interested in this. Maybe try pinging the linked issue? It doesn't look like you've participated in that issue so far. Normally when one adds a big new feature and tries to contribute a bunch of code to a project like NumPy, it's best to ask first in an issue if it makes sense to do the work at all.

Finally, it's not clear to me why this feature needs to live in NumPy. Why can't you put up a package on PyPI that includes the Python code you want to add?

Ijtihed:
> This is a new feature and as such it needs to be brought to the mailing list. 

@ngoldbaum oh wow I seriously didn't know it worked like that. It makes sense though and I should've looked at that beforehand. I also should've talked in the issue and I might've jumped the gun a bit.

> Finally, it's not clear to me why this feature needs to live in NumPy. Why can't you put up a package on PyPI that includes the Python code you want to add?

The closest existing thing ik for this is opt_einsum.contract_expression hwich does this closely. however I thought the overhead being eliminated is intrinsic to the repo's own einsum Python layer and doing it like this would of course not need any third party dependency. I again in hindsight realize that's a discussable design decision (whether it's worth implementing or not) and I should've done that.

Should I open a thread (?) in the mailing list now or just discuss in the issue first? thanks man and sorry for the hassle!!

ngoldbaum:
No hassle. Maybe comment on the issue for now to see if that interest anyone that was participating in the discussion last year.

benchmarks/benchmarks/bench_linalg.py
```diff
@@ -273,3 +273,141 @@ def setup(self, configuration):
...
         return np.matmul(self.a1, self.a2)
+
+
+class EinsumExpressionOverhead(Benchmark):
+    """Benchmark EinsumExpression vs einsum for repeated calls."""
+
+    def setup(self):
+        # Small 2-op matmul
+        self.a_small = np.arange(16.0).reshape(4, 4)
+        self.b_small = np.arange(16.0).reshape(4, 4)
```
reinderien, line 284: `self.a_small.copy()`, etc. here and elsewhere. Or perhaps even a shallow copy `self.b_small = self.a_small`; I don't know whether it would make a perf difference

benchmarks/benchmarks/bench_linalg.py
```diff
@@ -273,3 +273,141 @@ def setup(self, configuration):
...
+        self.expr_small = np.EinsumExpression(
+            'ij,jk->ik', (4, 4), (4, 4))
```
reinderien, line 286: These should take `self.a_small.shape`, etc. rather than re-stating the shape inline

numpy/_core/tests/test_einsum.py
```diff
@@ -1366,3 +1367,254 @@ def inner(*inputs, out):
 
     # At with GROWINNER 11 decimals succeed (larger will be less)
     assert_almost_equal(res, value, decimal=15)
+
+
+class TestEinsumExpression:
+    def test_basic_matmul(self):
+        a = np.random.rand(4, 5)
+        b = np.random.rand(5, 3)
```
reinderien, line 1375: Is it safe to run a non-deterministic test like this? I hardly think so. If you're going to run a fuzz test, (1) set a constant seed, and (2) iterate for a while (I guess that's what `test_repeated_calls_same_result` does). Or if you _don't_ want a fuzz test, then don't use `rand` at all.

numpy/_core/tests/test_einsum.py
```diff
@@ -1366,3 +1367,254 @@ def inner(*inputs, out):
...
+        expr = np.EinsumExpression('ij,jk->ik', (4, 5), (5, 3))
+        result = expr(a, b)
+        expected = np.einsum('ij,jk->ik', a, b)
+        assert_allclose(result, expected)
```
reinderien, line 1379: We can do better than a floating-point product and `allclose`. If the inputs are integral, the comparison can be exact. This can replace, or ideally supplement, the existing tests.

numpy/_core/einsumfunc.py
```diff
@@ -1645,3 +1647,192 @@ def einsum(*operands, out=None, optimize=False, **kwargs):
...
+    ----------
+    subscripts : str
+        Specifies the subscripts for summation as a comma-separated list of
+        subscript labels. An explicit output can be specified with '->'.
+    *shapes : tuple of int
+        The shapes of the operands. One shape tuple per operand.
+    optimize : {True, 'greedy', 'optimal'}, optional
+        Choose the type of path.
+
+        * if True defaults to the 'greedy' algorithm
```
reinderien, line 1677: If we're here anyway, what's the point of including `True`? Why not just constrain to a `Literal['greedy', 'optimal']`?

numpy/_core/einsumfunc.py
```diff
@@ -1645,3 +1647,192 @@ def einsum(*operands, out=None, optimize=False, **kwargs):
...
+                raise TypeError(
+                    f"shape {i} must be a tuple, got {type(shape).__name__}")
+
+        if optimize is False or optimize is None:
+            raise ValueError(
+                "optimize must be True, 'greedy', or 'optimal' for "
+                "EinsumExpression; use np.einsum directly for "
+                "unoptimized contractions")
+        if optimize is True:
+            optimize = 'greedy'
```
reinderien, line 1735: This should all just be deleted. Your `in` check below is sufficient.

numpy/_core/einsumfunc.py
```diff
@@ -1645,3 +1647,192 @@ def einsum(*operands, out=None, optimize=False, **kwargs):
...
+        from numpy.lib._stride_tricks_impl import as_strided
+        fake_operands = []
+        for shape in shapes:
+            buf = asanyarray([0.0])
+            fake = as_strided(
+                buf, shape=shape, strides=(0,) * len(shape))
+            fake_operands.append(fake)
+
+        _, contraction_list = einsum_path(
+            subscripts, *fake_operands, optimize=optimize, einsum_call=True)
```
reinderien, line 1757: I guess as a first step this is fine, but requiring fake operands at all is a bad API for `einsum_path`. I don't have a good understanding of how the Numpy API is designed to make breaking changes - maybe offering an alternative implementation of `einsum_path`, or doing argument switching - but writing a good wrapper to a bad function is a little flimsy.

numpy/_core/einsumfunc.py
```diff
@@ -1645,3 +1647,192 @@ def einsum(*operands, out=None, optimize=False, **kwargs):
...
+        -------
+        output : ndarray
+            The result of the einsum contraction.
+        """
+        if len(operands) != self._num_operands:
+            raise ValueError(
+                f"expected {self._num_operands} operands, "
+                f"got {len(operands)}")
+
+        valid_kwargs = ['dtype', 'order', 'casting']
```
reinderien, line 1788: ...Why? Why not just delete `**kwargs` and write out these three arguments instead?

reinderien:
> Finally, it's not clear to me why this feature needs to live in NumPy. Why can't you put up a package on PyPI that includes the Python code you want to add?

If it's _only_ code added, then perhaps. But I really consider the 2.4 implementation to be faulty. I don't know how Numpy manages breaking changes, and I expect that it would be difficult to convince the community to adopt a breaking change, but there is a middle ground between the current PR that papers over the problem and the maximalist version that completely rewrites einsum. Different modes to einsum/einsum_path can be introduced via argument switching.

ngoldbaum:
> I don't know how Numpy manages breaking changes, and I expect that it would be difficult to convince the community to adopt a breaking change, but there is a middle ground between the current PR that papers over the problem and the maximalist version that completely rewrites einsum. Different modes to einsum/einsum_path can be introduced via argument switching.

It'd probably have to start with a mailing list message that has an actionable proposal. If you can, please also include some background info and links to prior discussions for people like me who are unfamiliar with the issues you'd like to see fixed.

I don't have a lot  of context about design decisions in einsum so I can't say what the best path forward is.

Ijtihed:
Please tag me @reinderien or link it here if you do. I can do it also. I'm still quite interested in this :)

Ijtihed:
I'll close this for now

Issue #29071, opened 2025-05-27
ENH: Use BLAS in np.einsum when optimize=False


lysnikolaou:
### Proposed new feature or change:

As we discussed in the NumPy community meeting some time ago, we want to propose that BLAS is used in `np.einsum`, even if `optimize=False`. To make it clear, this is not changing the default of `optimize` (it still is going to be false), just changing the underlying behavior _when it is false_. I've done a lot of benchmarking to understand the performance implications of this. It all boils down to this:

- I've verified, as mentioned in the meeting, that our handwritten version of `einsum` in C (`PyArray_EinsteinSum`) actually outperforms BLAS by quite a distance when calling `np.einsum` on relatively small arrays. I've written benchmarks that use 1-D, 2-D and 3-D arrays of different sizes. Here's the actual results:

``` 
| Change   | Before [6fb8dc25] <main>   | After [e0e97b04] <use-blas-even-if-nooptimize>   |   Ratio | Benchmark (Parameter)                                         |
|----------|----------------------------|--------------------------------------------------|---------|---------------------------------------------------------------|
| +        | 1.68±0.1μs                 | 2.84±0.04μs                                      |    1.69 | bench_linalg.EinsumNoOptimize.time_einsum_one_dim_small       |
| +        | 10.3±0.2μs                 | 12.0±0.1μs                                       |    1.17 | bench_linalg.EinsumNoOptimize.time_einsum_three_dim_small     |
| +        | 9.04±0.1μs                 | 10.5±0.1μs                                       |    1.16 | bench_linalg.EinsumNoOptimize.time_einsum_two_dim_small       |
| +        | 16.4±0.1μs                 | 18.6±0.2μs                                       |    1.14 | bench_linalg.EinsumNoOptimize.time_einsum_two_three_dim_small |
| -        | 81.6±0.7μs                 | 32.1±0.7μs                                       |    0.39 | bench_linalg.EinsumNoOptimize.time_einsum_one_dim_big         |
| -        | 547±7μs                    | 42.5±2μs                                         |    0.08 | bench_linalg.EinsumNoOptimize.time_einsum_three_dim_big       |
| -        | 1.65±0.01ms                | 90.0±1μs                                         |    0.05 | bench_linalg.EinsumNoOptimize.time_einsum_two_three_dim_big   |
| -        | 4.73±0.2ms                 | 170±4μs                                          |    0.04 | bench_linalg.EinsumNoOptimize.time_einsum_two_dim_big         |
[...]

PR #31705, opened 2026-06-21
DOC: fix einsum return type docstring

Ijtihed:
`einsum` docstring says it returns `ndarray` but `c_einsum` calls `PyArray_Return` which converts 0-d results to scalars (e.g `einsum('i,i->', a, b)` returns `np.float64`).

Came up during review of #31702 ([comment](https://github.com/numpy/numpy/pull/31702#discussion_r2162353983)).

#### AI Disclosure

Drafted with Claude.

### DIFF
--- a/numpy/_core/einsumfunc.py
+++ b/numpy/_core/einsumfunc.py
@@ -1300,8 +1300,20 @@ def einsum(*operands, out=None, optimize=False, **kwargs):
 
     Returns
     -------
-    output : ndarray
+    output : ndarray or scalar
         The calculation based on the Einstein summation convention.
+
+        An output with one or more indices is always returned as an
+        ndarray. When the output has no indices (e.g. ``'i,i->'``) the
+        result is 0-dimensional and the exact type depends on the
+        arguments:
+
+        - If `out` is given, it is returned unchanged (an ndarray).
+        - With the default ``optimize=False``, a scalar (e.g.
+          ``np.float64``) is returned rather than a 0-d array.
+        - With ``optimize`` enabled the 0-d result may be either a scalar
+          or a 0-d ndarray depending on the contraction path, and should
+          not be relied upon. Pass `out` to guarantee an ndarray.
 
     See Also
     --------


numpy/_core/einsumfunc.py
```diff
@@ -1300,8 +1300,10 @@ def einsum(*operands, out=None, optimize=False, **kwargs):
 
     Returns
     -------
-    output : ndarray
+    output : ndarray or scalar
         The calculation based on the Einstein summation convention.
+        When the output subscript has no indices (e.g. ``'i,i->'``),
+        a scalar is returned instead of an ndarray.
```
jorenham: Hmm, it turns out that it's not as simple as that:

```
In [1]: a = b = np.arange(4)

In [2]: np.einsum('i,i->', a, b)
Out[2]: np.int64(14)

In [3]: np.einsum('i,i->', a, b, optimize=True)
Out[3]: array(14)

In [4]: np.einsum('i,i->', a, b, optimize='greedy')
Out[4]: array(14)

In [5]: np.einsum('i,i->', a, b, optimize='optimal')
Out[5]: array(14)

In [6]: np.einsum('i,i->', a, b, out=np.empty(()))
Out[6]: array(14.)
```

Ijtihed: yeah it's path-dependent it seems. after some slightly deeper checking only the default is stable so `optimize=False` and no `out` and every index-less output returns a scalar.

how should I update the docstring in this case? It seems this (BLAS for optimize=false) gh-29071 would make it so it always returns array. but I think this should anyway for sure be more documented rather than just a todo line somewhere in the file :_)

ngoldbaum: I would try to fully describe the semantics as it actually works. It's OK if it's complicated as long as it's accurate.

Ijtihed: Added

numpy/_core/einsumfunc.py
```diff
@@ -1300,9 +1300,27 @@ def einsum(*operands, out=None, optimize=False, **kwargs):
...
+        ndarray. When the output has no indices (e.g. ``'i,i->'``) the
+        result is 0-dimensional and the exact type depends on the
+        arguments:
+
+        - If `out` is given, it is returned unchanged (an ndarray).
+        - With the default ``optimize=False``, a scalar (e.g.
+          ``np.float64``) is returned rather than a 0-d array.
+        - With ``optimize`` enabled, the result may be a scalar or a 0-d
+          ndarray depending on the optimized contraction path, and should
+          not be relied upon. A 0-d ndarray is produced only when the
```
ngoldbaum: I'd probably delete everything after the period on this line.

@jorenham what do you think of this text, besides my comment?

jorenham: I like the overall 3-point structure here.

In the third point there's a lot of details now about the specific `opimize` mechanics, though. To me that seems like a bit much; we're talking about just the `output` param, after all. I think one or two sentences should suffice here. 
Claude suggests: "With ``optimize`` enabled the 0-d result may be either a scalar or a 0-d `ndarray` depending on the contraction path, and should not be relied upon. Pass ``out`` to guarantee an `ndarray`."

Ijtihed: > "With `optimize` enabled the 0-d result may be either a scalar or a 0-d `ndarray` depending on the contraction path, and should not be relied upon. Pass `out` to guarantee an `ndarray`.

Should I just switch it to this?

jorenham: Fine by me; but don't forget to update the AI disclosure if you do

Ijtihed: Done for both

jorenham (approved):

ngoldbaum:
Thanks @ljtihed. For future reference you can add annotations to your commit message to skip the github actions CI for docs-only PRs. Our doc builds and doctests run on CircleCI.

https://numpy.org/doc/stable/dev/development_workflow.html#commands-to-skip-continuous-integration

PR #31709, opened 2026-06-22
MAINT/TYP: bump ``mypy`` to ``2.1.0``

jorenham:
It seems like none of the behavioral mypy changes affect numpy (locally, at least), but there have been quite a lot of performance improvements according to the [relnotes](https://mypy.readthedocs.io/en/stable/changelog.html#mypy-2-1), which is always welcome.

No AI

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -24,7 +24,7 @@ dependencies:
   - pytest-xdist
   - hypothesis
   # For type annotations
-  - mypy==1.20.2
+  - mypy==2.1.0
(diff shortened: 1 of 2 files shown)


jorenham:
Seems like there has been a new cython-lint release that's causing the lint CI job to fail, cc @MarcoGorelli 

edit: I opened https://github.com/numpy/numpy/pull/31711 to fix this.

charris:
Thanks Joren.

PR #31710, opened 2026-06-22
MAINT/TYP: bump ``pyrefly`` to ``1.1.1``

jorenham:
Some false positives have been resolved, which allowed us to remove 9 now-unused `# pyrefly: ignore` comments.

Another cool thing is that pyrefly 1.1 comes with a new `pyrefly coverage check` feature. I wrote this as a better and more flexible alternative to `pyright --verifytypes`, which we're currently using in CI to prevent type-coverage backslides. So my plan is to follow this up with a PR that replaces `pyright --verifytypes`  with `pyrefly coverage check`, which will also allow us to get rid of the required pyright wrapper script to run it. I also expect that it will be a lot faster.

relnotes:  
- https://github.com/facebook/pyrefly/releases/tag/1.1.0
- https://github.com/facebook/pyrefly/releases/tag/1.1.1

No AI


charris:
Needs rebase.

jorenham:
> Needs rebase.

strange; I didn't even touch the same lines :thinking:

charris:
Thanks Joren.

PR #31716, opened 2026-06-22
TYP: specialized unary ``inexact | object_`` endo-ufuncs

jorenham:
This is the first in a series of PRs where I'll be specializing individual (groups of) ufuncs so that they have proper typing support. Because currently, all ufuncs are (roughly speaking) annotated in the same "one size first all" way, causing all shape-typing and dtype information to be lost when they're used. 

I think this is the biggest typing-hole in numpy at the moment, but unfortunately it's going to be a lot of work to specialize each of the ufuncs. 

After experimenting (a lot) with this, I found that this "type-check-only subtyping" approach is the way to go (most robust, maintainable, and minimal differences between type-checkers). I've successfully been using this approach in scipy-stubs to specialize the `scipy.special` ufuncs, and so far I've had no issues with it, so it's a battle tested approach in that sense.

This is the first batch of ufuncs that tackles the 21 (excluding aliases) unary ufuncs with floating/complex/object input- and output-types (which is what I meant with "endo-ufuncs" in the title). As can be seen in https://gist.github.com/jorenham/d977c63cc53ad4f00878d37ee42ab0e8, it's the largest group of ufuncs sharing the same signature, so that's why I started with these ones. 

---

**TLDR**-ish;

For those that prefer code over my ramblings (so probabably everyone here), here's a quick little demonstration (using `pyrefly check`):

<details>
<summary>code for repro</summary>

```py
from typing import reveal_type

import numpy as np

reveal_type(np.sin(3.14))
reveal_type(np.sin([1, 2, 3]))
reveal_type(np.sin(np.eye(3, dtype=np.float64)))
```

</details>

*Before:*

(shape-type and dtype don't propagate)

```
 INFO revealed type: Any [reveal-type]
 --> ufunc_demo.py:5:12
  |
5 | reveal_type(np.sin(3.14))
  |            --------------
  |
 INFO revealed type: ndarray [reveal-type]
 --> ufunc_demo.py:6:12
  |
6 | reveal_type(np.sin([1, 2, 3]))
  |            -------------------
  |
 INFO revealed type: ndarray [reveal-type]
 --> ufunc_demo.py:7:12
  |
7 | reveal_type(np.sin(np.eye(3, dtype=np.float64)))
  |            -------------------------------------
  |
```

*After:*

```
 INFO revealed type: float64 [reveal-type]
 --> ufunc_demo.py:5:12
  |
5 | reveal_type(np.sin(3.14))
  |            --------------
  |
 INFO revealed type: ndarray[tuple[int], dtype[float64]] [reveal-type]
 --> ufunc_demo.py:6:12
  |
6 | reveal_type(np.sin([1, 2, 3]))
  |            -------------------
  |
 INFO revealed type: ndarray[tuple[int, int], dtype[float64]] [reveal-type]
 --> ufunc_demo.py:7:12
  |
7 | reveal_type(np.sin(np.eye(3, dtype=np.float64)))
  |            -------------------------------------
  |
```

You'd see the same results if you'd replace `sin` here with any of

- `a[rc]cos`
- `a[rc]cosh`
- `a[rc]sin`
- `a[rc]sinh`
- `a[rc]tan`
- `a[rc]tanh`
- `cos`
- `cosh`
- `exp`
- `exp2`
- `expm1`
- `log`
- `log10`
- `log1p`
- `log2`
- `rint`
- `sin`
- `sinh`
- `sqrt`
- `tan`
- `tanh`

---

Fun fact: Before mypy 1.20, stubtest would get mad if you'd try to do this. But this was "fixed" (or improved, depending on how you look at it) in https://github.com/python/mypy/pull/20352 over half a year ago, with this specific use-case in mind. It's always nice to see a long-term plan come together :)

---

No AI (wouldn't have worked anyway).

### DIFF
--- a/numpy/__init__.pyi
+++ b/numpy/__init__.pyi
@@ -390,6 +390,35 @@ from numpy._core.shape_base import (
     vstack,
     unstack,
 )
+from numpy._core.umath import (
+    arccos,
+    arccos as acos,
+    arccosh,
+    arccosh as acosh,
+    arcsin,
+    arcsin as asin,
+    arcsinh,
+    arcsinh as asinh,
+    arctan,
+    arctan as atan,
+    arctanh,
+    arctanh as atanh,
+    cos,
+    cosh,
+    exp,
+    exp2,
+    expm1,
+    log,
+    log10,
+    log1p,
+    log2,
+    rint,
+    sin,
+    sinh,
+    sqrt,
+    tan,
+    tanh,
+)
 
 from ._expired_attrs_2_0 import __expired_attributes__ as __expired_attributes__
 from ._globals import _CopyMode as _CopyMode
@@ -7545,13 +7574,7 @@ class ufunc:
 # Parameters: `__name__`, `ntypes` and `identity`
 absolute: _UFunc_Nin1_Nout1[L["absolute"], L[20], None]
 add: _UFunc_Nin2_Nout1[L["add"], L[22], L[0]]
-arccos: _UFunc_Nin1_Nout1[L["arccos"], L[8], None]
-arccosh: _UFunc_Nin1_Nout1[L["arccosh"], L[8], None]
-arcsin: _UFunc_Nin1_Nout1[L["arcsin"], L[8], None]
-arcsinh: _UFunc_Nin1_Nout1[L["arcsinh"], L[8], None]
 arctan2: _UFunc_Nin2_Nout1[L["arctan2"], L[5], None]
-arctan: _UFunc_Nin1_Nout1[L["arctan"], L[8], None]
-arctanh: _UFunc_Nin1_Nout1[L["arctanh"], L[8], None]
 bitwise_and: _UFunc_Nin2_Nout1[L["bitwise_and"], L[12], L[-1]]
 bitwise_count: _UFunc_Nin1_Nout1[L["bitwise_count"], L[11], None]
 bitwise_or: _UFunc_Nin2_Nout1[L["bitwise_or"], L[12], L[0]]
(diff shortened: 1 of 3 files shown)


jorenham:
**colour:**

- `colour/models/rgb/transfer_functions/rimm_romm_rgb.py:536` ([src](https://github.com/colour-science/colour/blob/develop/colour/models/rgb/transfer_functions/rimm_romm_rgb.py#L536)): this is an issue in colour itself, where it directly compares an `ArrayLike` using `<=`, which is indeed type-unsafe. So this error is a good thing.
- `colour/colorimetry/luminance.py:384` ([src](https://github.com/colour-science/colour/blob/c5bc9ea8e205733d83fc42c2f0976de787ccd2a4/colour/colorimetry/luminance.py#L384-L389)): Similarly, they're dividing by an `ArrayLike`, which is type-unsafe, so mypy rightly complains.
- `colour/colorimetry/luminance.py:452` ([src](https://github.com/colour-science/colour/blob/c5bc9ea8e205733d83fc42c2f0976de787ccd2a4/colour/colorimetry/luminance.py#L451-L457)): Same as above.
- `colour/models/hdr_ipt.py:142-146` ([src](https://github.com/colour-science/colour/blob/c5bc9ea8e205733d83fc42c2f0976de787ccd2a4/colour/models/hdr_ipt.py#L139-L146)): A `float` is modified in-place by using `*=` and `/=` with an `ndarray`. This is type-safe, but mypy doesn't like in-place type modifications like this. Regardless of whether mypy's behavior is good or bad, it's safe to conclude that this mypy error isn't caused by a regression in numpy.
- `colour/models/hdr_cie_lab.py:139-143` ([src](https://github.com/colour-science/colour/blob/c5bc9ea8e205733d83fc42c2f0976de787ccd2a4/colour/models/hdr_cie_lab.py#L138-L143)): same as above
- `colour/quality/cqs.py:600`: casting in Python is not much different from a `# type: ignore` (I could rant about this for hours, grumble...), so that this cast is now redundant is a good thing: an error has been resolved.

**(py)spark:**

- `python/pyspark/statcounter.py:32` ([src](https://github.com/apache/spark/blob/18d6d218e6b05c746a55021aa8883405445f1fe6/python/pyspark/statcounter.py#L27-L32)): mypy doesn't seem to widen this type to a union. Nothing we can help.

**optuna:**

- `optuna/samplers/nsgaii/_crossovers/_undx.py:85`  ([src](https://github.com/optuna/optuna/blob/243c156d9dda5093fa4358c6923fb22cd721fd04/optuna/samplers/nsgaii/_crossovers/_undx.py#L82-L85)): the result of `np.sqrt(n)`, which is now inferred as `float64`, is being assigned to an invariant `float` attribute, which is type-unsafe. Mypy is correct to report an error here, and this is not a numpy regression.

**dedupe:**

- `dedupe/canopy_index.py:71` ([src](https://github.com/dedupeio/dedupe/blob/3f61e79102910bd355e920a2df7e44c14c9cb247/dedupe/canopy_index.py#L66-L71)): mypy doesn't support correlated branch analysis, at least not when type-narrowing with `is None`, so mypy thinks that `weight: float | None`, which would indeed be type-unsafe to pass to `float.__pow__`. This seems unrelated to numpy, and I don't know why this error was hidden before, tbh.

---

None of the new errors numpy regressions or anything we could work around even. There are several (previously ignored or casted) typing issues that are now resolved.

jorenham:
@KelSolaar, are you able to live with these new `colour` errors?

thomasmansencal:
No issue on our end, we will fix our annotations accordingly!

charris:
Thanks Joren.

PR #31718, opened 2026-06-22
CI/TYP: Use Pyrefly to enforce 100% type coverage

jorenham:
From https://github.com/numpy/numpy/pull/31710:

> Another cool thing is that pyrefly 1.1 comes with a new `pyrefly coverage check` feature. I wrote this as a better and more flexible alternative to `pyright --verifytypes`, which we're currently using in CI to prevent type-coverage backslides. So my plan is to follow this up with a PR that replaces `pyright --verifytypes`  with `pyrefly coverage check`, which will also allow us to get rid of the required pyright wrapper script to run it. I also expect that it will be a lot faster.

So this replaces the (based)pyright type completeness check with `pyrefly coverage check`, because it simply does a better job at measuring type coverage.

The default coverage threshold of `pyrefly coverage check` is set to 100%, which NumPy has, so there's no need to manually specify a `--fail-under=100` or something.

Relevant Pyrefly docs: https://pyrefly.org/en/docs/report/

---

No AI


charris:
Thanks Joren.

PR #31723, opened 2026-06-23
MAINT: Update meson to match main

charris:
It is best to keep main and the stable branch in sink to avoid accidental changes.

No AI.


seberg:
Thanks Chuck.

(I restarted the ASAN+UBSAN job, it was running 3h and hadn't gone anywhere).

PR #31726, opened 2026-06-23
MAINT: Rename subroutine for crackfortran tests

tiesjan:
### PR summary

When importing `numpy` version 2.4.4 through Nix, its store path (path the package lives in) resulted in `/nix/store/r7i744barr4p4vr65110rh4n60w6p93i-python3.14-numpy-2.4.4`. This happens to contain the word `bar`: `...r7i744--bar--r4p4vr...`. Unfortunately, this makes one of the tests fail when building the Nix package locally. This is because the test checks for `bar` in a Fortran module(?) path, since that is a subroutine's name in that Fortran module.

By making the subroutine's name more unique, a collision with a Nix store path is less likely to happen.

I work as a software engineering lead for @channable and this package is being imported as a dependency as well as for some calculations for our service. The reason for contributing to your project is to ensure the collision with a Nix store path outlined above is less likely to happen again.

#### AI Disclosure

No AI tools used.

### DIFF
--- a/numpy/f2py/tests/src/crackfortran/gh23879.f90
+++ b/numpy/f2py/tests/src/crackfortran/gh23879.f90
@@ -5,14 +5,14 @@ module gh23879
 
  contains
 
-    subroutine foo(a, b)
+    subroutine foo42bar(a, b)
        integer, intent(in) :: a
        integer, intent(out) :: b
        b = a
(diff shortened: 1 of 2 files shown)


charris:
Thanks @tiesjan .

PR #31727, opened 2026-06-23
TST: use setup-sde instead of curl to get SDE binaries

### LINKED ISSUE #31591: CI: Intel SDE downloads are failing
Just opening to have an issue, but currently the two jobs:
`intel_sde_avx512` and `intel_sde_spr` are both failing because intels download server started to reject `curl`...

Not sure if there is a nice work-around (to get around whatever weird check intel has, or an alternative place to fetch it from).
Even something like this: https://github.com/petarpetrovt/setup-sde except that I am always not so sure about adding "random" actions.

That is e.g. here: https://github.com/numpy/numpy/actions/runs/27258956378/job/80500546634?pr=31584

which uses:
```
  SDE_URL="https://downloadmirror.intel.com/859732/sde-external-9.58.0-2025-06-16-lin.tar.xz"
  SDE_SHA256="f849acecad4c9b108259c643b2688fd65c35723cd23368abe5dd64b917cc18c0"
  curl -o /tmp/sde.tar.xz "$SDE_URL"
```

ngoldbaum:
Fixes #31591

### DIFF
--- a/.github/workflows/linux_simd.yml
+++ b/.github/workflows/linux_simd.yml
@@ -1,8 +1,7 @@
 name: Linux SIMD tests
 
-# To update Intel SDE (not handled by Dependabot): download new version from
-# https://www.intel.com/content/www/us/en/developer/articles/tool/software-development-emulator.html
-# and update SDE_URL and SDE_SHA256 (curl -sL <url> | sha256sum)
+# To update the Intel SDE version (not handled by Dependabot setup-sde updates):
+# update the sdeVersion inputs in the Intel SDE jobs below.
 #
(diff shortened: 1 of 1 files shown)


ngoldbaum:
It seems to be working. @andyfaff does this look OK to you?

ngoldbaum:
Also ping @petarpetrovt - we're thinking about using `setup-sde` in NumPy's CI.

charris:
Thanks Nathan.

r-devulap:
Following your lead @ngoldbaum, thanks for fixing :)

PR #31730, opened 2026-06-23
TYP: specialized unary ``floating | object_`` endo-ufuncs

jorenham:
Here's part two of the typed ufunc series, following #31716. This time, featuring `cbrt`, `fabs`, `deg2rad`, `rad2deg`, `degrees`, `radians`, which all share the same (static) type signature (see https://gist.github.com/jorenham/d977c63cc53ad4f00878d37ee42ab0e8 for details).

These changes are very similar to the ones in #31716, except that here, complex dtypes *aren't* supported; only the four `floating` and `object_` dtypes. So the before/after will be pretty much the same as for #31716.

---

No AI.

### DIFF
--- a/numpy/__init__.pyi
+++ b/numpy/__init__.pyi
@@ -403,15 +403,21 @@ from numpy._core.umath import (
     arctan as atan,
     arctanh,
     arctanh as atanh,
+    cbrt,
     cos,
     cosh,
+    deg2rad,
+    degrees,
     exp,
     exp2,
     expm1,
+    fabs,
     log,
     log10,
     log1p,
     log2,
+    rad2deg,
+    radians,
     rint,
     sin,
     sinh,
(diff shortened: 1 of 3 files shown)


jorenham:
All new mypy primer for `colour` seem to be cases where `numpy.typing.ArrayLike` is passed directly into one of these ufuncs, e.g. `np.radians(phi)` where `phi: ArrayLike` ([src](https://github.com/colour-science/colour/blob/c5bc9ea8e205733d83fc42c2f0976de787ccd2a4/colour/geometry/ellipse.py#L207)). This is indeed type-unsafe, because `ArrayLike` also accepts types that these ufuncs don't accept such as `str_` arrays and raw `builtins.bytes` objects. So mypy is right to complain here. The type-safe approach would be to guard against this by doing `a = np.array(...); assert a.dtype.kind in "bfO"`, `np.array(..., dtype=np.float64)` or something similar.

charris:
Thanks Joren.

PR #31731, opened 2026-06-23
BUG: fix leak in reductions when a ufunc override errors or is not found

### LINKED ISSUE #31677: BUG: Reference leak in PyUFunc_GenericReduction on CheckOverride error
## Bug Description

In `numpy/_core/src/umath/ufunc_object.c`, when `PyUFunc_CheckOverride` fails in `PyUFunc_GenericReduction`, the code returns NULL directly instead of jumping to the cleanup label, leaking `full_args.in` and `full_args.out`.

## Location

`numpy/_core/src/umath/ufunc_object.c:3777-3780`

## Code

```c
int errval = PyUFunc_CheckOverride(ufunc, _reduce_type[operation],
        full_args.in, full_args.out, wheremask_obj, args, len_args, kwnames, &override);
if (errval) {
    return NULL;    // BUG: should be `goto fail` — leaks full_args.in/out
}
```

## Impact

When `reduce`/`accumulate`/`reduceat` encounters a `__array_ufunc__` override error, the `full_args.in` and `full_args.out` tuple references are leaked. This happens every time an override is attempted on an object that raises an exception in its `__array_ufunc__` implementation.

## Suggested Fix

```c
if (errval) {
    goto fail;  // cleanup path that DECREFs full_args
}
```

## Environment

- NumPy version: current main
- Affected: `np.add.reduce()`, `np.accumulate()` etc. with `__array_ufunc__` overrides that raise

ikrommyd:
### PR summary
Closes https://github.com/numpy/numpy/issues/31677

Just another thing I noticed while reading reduction code and realised there's an open issue for it already. 
One can see this memory leak by doing something like the following and memory profiling that (with memray for example) and will notice the memory is just growing.
```py
import numpy as np

class Array:
    data = np.array([1,2,3])
    def __array_ufunc__(ufunc, method, *inputs, **kwargs):
        return NotImplemented

arr = Array()

if __name__ == "__main__":
    for _ in range(1_000_000):
        try:
            np.add.reduce(arr)
        except Exception as e:
            pass
```
I think `goto fail` is probably nicer than
```c
Py_XDECREF(full_args.in);
Py_XDECREF(full_args.out);
```
and I believe it's safe. Only the `else if (override)` branch right bellow cannot use `goto fail` as it needs to return the override and not NULL.

I don't think there is a way to have a test regarding the ref count in the python side because `full_args` is created internally in the function in C. But we could have a test that is just the above without the loop and a sanitizer would probably capture the leak. Let me know if you want that. cc @seberg because you wanted a similar test in https://github.com/numpy/numpy/pull/31699. 

N/A

#### AI Disclosure
No AI

### DIFF
--- a/numpy/_core/tests/test_umath.py
+++ b/numpy/_core/tests/test_umath.py
@@ -3954,6 +3954,21 @@ class TestSpecialMethods:
         with assert_raises_regex(TypeError, fnmatch.translate(msg)):
             np.add(A(), object(), out=1)
 
+    def test_ufunc_override_reduction_not_implemented_leak(self):
+        # gh-31731: the reduction override error path leaked references to
+        # the internal full_args.in/out tuples when __array_ufunc__ returned
+        # NotImplemented (or raised). No refcount assertion is made here; a
+        # leak sanitizer build is expected to catch the regression.
+
+        class A:
+            def __array_ufunc__(self, *args, **kwargs):
+                return NotImplemented
+
+        a = A()
+        assert_raises(TypeError, np.add.reduce, a)
+        assert_raises(TypeError, np.add.accumulate, a)
+        assert_raises(TypeError, np.add.reduceat, a, [0])
+
     def test_ufunc_override_disabled(self):
 
         class OptOut:
--- a/numpy/_core/tests/test_umath.py
+++ b/numpy/_core/tests/test_umath.py
@@ -3955,7 +3955,7 @@ class TestSpecialMethods:
             np.add(A(), object(), out=1)
 
     def test_ufunc_override_reduction_not_implemented_leak(self):
-        # gh-31731: the reduction override error path leaked references to
+        # gh-31677: the reduction override error path leaked references to
         # the internal full_args.in/out tuples when __array_ufunc__ returned
         # NotImplemented (or raised). No refcount assertion is made here; a
         # leak sanitizer build is expected to catch the regression.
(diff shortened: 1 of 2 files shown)


ikrommyd:
Just pasting the results of profiling the snippet above with memray (with native traces too) on main:
image
image
With this PR, it becomes flat:
image

seberg:
Yeah, just out of principle. Seems like a simple test to add and I expect leak sanitizer or so will find it even without a loop.

ikrommyd:
> Yeah, just out of principle. Seems like a simple test to add and I expect leak sanitizer or so will find it even without a loop.

Done

ikrommyd:
The BLAS tests failure should be unrelated

numpy/_core/tests/test_umath.py
```diff
@@ -3954,6 +3954,21 @@ def __array_ufunc__(self, *args, **kwargs):
...
+        # leak sanitizer build is expected to catch the regression.
+
+        class A:
+            def __array_ufunc__(self, *args, **kwargs):
+                return NotImplemented
+
+        a = A()
+        assert_raises(TypeError, np.add.reduce, a)
+        assert_raises(TypeError, np.add.accumulate, a)
+        assert_raises(TypeError, np.add.reduceat, a, [0])
```
seberg, line 3970: N.B. I generally prefer `with pytest.raises()` even if it a lot of old tests use `assert_raises`.

seberg (approved):
Thanks, for adding the test, good enough, let's put it in.

numpy/_core/tests/test_umath.py
ikrommyd: Ah thanks. Noted for the future. I intuitively tend to follow existing patterns and I literally just copied it from the tests right above it

PR #31732, opened 2026-06-23
TYP: Fix incorrect dtype inference of ``asarray([])``

jorenham:
There's a regression in 2.5.0 that caused `np.asarray([])`  to be inferred as a `bool` dtype, rather than `float64`. The fix is to move the relevant `float64` overloads before the `bool` ones (1d and 2d cases).

`np.array` and the other array constructors don't seem to have this issue; only `np.asarray`.

---

No AI.


numpy/typing/tests/data/reveal/array_constructors.pyi
```diff
@@ -92,6 +92,9 @@ assert_type(np.asarray(A, dtype="c16"), npt.NDArray[Any])
 assert_type(np.asarray(_f32_1d), _Array1D[np.float32])
 assert_type(np.asarray(_f32_1d, dtype=np.float64), _Array1D[np.float64])
 assert_type(np.asarray(_f32_1d, dtype="f8"), _Array1D[Any])
+assert_type(np.asarray([]), _Array1D[np.float64])
+assert_type(np.asarray([[]]), _Array2D[np.float64])
+assert_type(np.asarray(_py_b_1d), _Array1D[np.bool_])
```
tylerjereddy, line 97: is this accidentally duplicating the line below or?

jorenham: oh yea woops

jorenham:
The mypy primer errors for jax seem unrelated (smells more like a mypy issue to me): https://github.com/jax-ml/jax/blob/41804d74a5f4387681ab08e1513b770776960bed/jax/_src/numpy/lax_numpy.py#L3144-L3150

charris:
Thanks Joren.

PR #31739, opened 2026-06-24
BUG: fix ma.innerproduct for object-dtype inputs

### LINKED ISSUE #31736: BUG: `np.ma.innerproduct` throws error with object arrays
### Describe the issue:

On the behavior observed in the examples below,` np.ma.innerproduct` appears to mishandle object-dtype arrays. For the same inputs, `np.inner` returns the a scalar result, while `np.ma.innerproduct` raises an internal `AttributeError`.

Similar to #31019, I report it separately from #31735 because it affects a different API and fails with an exception. 

```python
import numpy as np
a = np.array([1,2,3], dtype=object)
b = np.array([0,1,0], dtype=object)

np.inner(a, b)           # -> 2
np.ma.innerproduct(a, b) # -> AttributeError
```

### Error message:

```shell
AttributeError: 'int' object has no attribute 'view'
```

2.6.0.dev0+git20260623.4357ad1

### Runtime Environment:

_No response_

Before reporting it, we reviewed the code, documentation and checked the latest development version of Numpy to confirm this observation.

Respirayson:
### PR summary

fix `np.ma.innerproduct` for object-dtype arrays.

previously, calling `np.ma.innerproduct` on object-dtype arrays could raise an `AttributeError ('int' object has no attribute 'view')` even though the equivalent `np.inner` call completed successfully and returned the expected scalar result.

update to correctly handle object-dtype scalar results and add a regression test

fixed #31736 

hi! im rayson, a quant developer who uses np regularly for data analysis, research, and personal projects.

im interested in contributing to open source software and was looking for issues that would be a good fit for a first contribution. I came across this bug, investigated the root cause, and worked on a fix together with a regression test.

#### AI Disclosure

No AI

### DIFF
--- a/numpy/ma/core.py
+++ b/numpy/ma/core.py
@@ -8343,7 +8343,10 @@ def inner(a, b):
         fa = fa.reshape((1,))
     if fb.ndim == 0:
         fb = fb.reshape((1,))
-    return np.inner(fa, fb).view(MaskedArray)
+    d = np.inner(fa, fb)
+    if np.ndim(d) == 0:
+        d = np.asarray(d)
+    return d.view(MaskedArray)
 
 
 inner.__doc__ = doc_note(np.inner.__doc__,
--- a/numpy/ma/core.py
+++ b/numpy/ma/core.py
@@ -1126,7 +1126,7 @@ class _MaskedBinaryOperation(_MaskedUFunc):
             tr = self.f.reduce(t, axis, dtype=dtype)
             mr = umath.logical_and.reduce(m, axis)
         
-        if np.ndim(tr) == 0:
+        if not isinstance(tr, (np.ndarray, np.generic)):
             tr = np.asarray(tr)
 
         if not tr.shape:
@@ -5179,7 +5179,7 @@ class MaskedArray(ndarray):
         if m is nomask:
             result = super().trace(offset=offset, axis1=axis1, axis2=axis2,
                                    out=out)
-            if np.ndim(result) == 0:
+            if not isinstance(result, (np.ndarray, np.generic)):
                 result = np.asarray(result)
             return result.astype(dtype)
         else:
@@ -5974,7 +5974,7 @@ class MaskedArray(ndarray):
             result = self.filled(fill_value).min(
                 axis=axis, out=out, **kwargs
             )
-            if np.ndim(result) == 0:
+            if not isinstance(result, (np.ndarray, np.generic)):
                 result = np.asarray(result)
             result = result.view(type(self))
             if result.ndim:
@@ -6083,7 +6083,7 @@ class MaskedArray(ndarray):
             result = self.filled(fill_value).max(
                 axis=axis, out=out, **kwargs
             )
-            if np.ndim(result) == 0:
+            if not isinstance(result, (np.ndarray, np.generic)):
                 result = np.asarray(result)
             result = result.view(type(self))
             if result.ndim:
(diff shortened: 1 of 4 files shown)


numpy/ma/core.py
```diff
@@ -8321,7 +8321,10 @@ def inner(a, b):
         fa = fa.reshape((1,))
     if fb.ndim == 0:
         fb = fb.reshape((1,))
-    return np.inner(fa, fb).view(MaskedArray)
+    d = np.inner(fa, fb)
+    if np.ndim(d) == 0:
+        d = np.asarray(d)
```
ngoldbaum, line 8361: Do we have special cases like this elsewhere in `np.ma` for object dtype? Or do we have the same bug in other functions?

Note that both object dtype and MaskedArray and somewhat unloved features and historically when we make changes in either subsystem we end up breaking other things unrelated to the fix.

shipitdev (commented):
Building on @ngoldbaum's concern, I wonder if we should just run a quick script feeding object arrays into the other linear algebra functions in `np.ma` to see what else breaks. 

If it's just an isolated bug here, this fix is good to go. If the whole module is making that same `.view()` assumption, we might need to handle it all at once. What do you guys think?

Respirayson:
> Building on @ngoldbaum's concern, I wonder if we should just run a quick script feeding object arrays into the other linear algebra functions in `np.ma` to see what else breaks.
> 
> If it's just an isolated bug here, this fix is good to go. If the whole module is making that same `.view()` assumption, we might need to handle it all at once. What do you guys think?

thanks for the suggestion! I wrote a small script to probe the public `np.ma` functions with object-dtype inputs and compare the behavior against the equivalent float64 inputs.

So it looks like `innerproduct` isn't an isolated case. I found several other functions (inner, max, min, ptp, and some others) that appear to fail for the same reason that a scalar result from np is treated as if it were an ndarray.

I'm verifying each of these now to ensure they're genuinely the same issue and ill put together follow-up fixes afterwards

numpy/ma/tests/test_core.py
```diff
@@ -4176,6 +4210,15 @@ def test_ptp(self):
         assert_equal(mX.ptp(0), cols)
         assert_equal(mX.ptp(1), rows)
 
+    def test_ptp_object(self):
```
shipitdev, line 4213: ptp() subs for self.min() - self.max() that's seems to resolve why there is no code diff about this test since max and min are already covered. tests seems a good addition.

numpy/ma/core.py
```diff
@@ -5945,7 +5950,11 @@ def min(self, axis=None, out=None, fill_value=None, keepdims=np._NoValue):
         # No explicit output
         if out is None:
             result = self.filled(fill_value).min(
-                axis=axis, out=out, **kwargs).view(type(self))
+                axis=axis, out=out, **kwargs
+            )
+            if np.ndim(result) == 0:
```
shipitdev: I did some digging, wanted to make sure this stays in line with numpy being built for speed. Looking closer at your approach it does a bit of unnecessary memory allocation, since np.asarray() runs even when the  result already supports everything we need. Leaning on Python's own isinstance and numpy's own np.generic type turns out to be a faster option while still giving the same correct result for the actual object-dtype bug So

Instead of:
```python
    if np.ndim(result) == 0:
        result = np.asarray(result)
```

what about:
```python
    if not isinstance(result, (np.ndarray, np.generic)):
        result = np.asarray(result)
```

shipitdev: This same np.ndim(x) == 0 pattern was added in five places in this diff (reduce, trace, min, max, inner), so this swap would need to go into all five here.

shipitdev (commented):
looks technically sound, although i have a few comments i'm going to drop below.

numpy/ma/core.py
Respirayson: Thanks for digging into the performance side and that makes sense! I've updated the code to use the `isinstance` check in all the five places in the diff. I also applied the same fix to `np.dot` as well since it was using the identical pattern.

PR #31741, opened 2026-06-24
TYP: specialized ``ceil``, ``floor``, ``trunc``, and ``invert`` ufuncs

jorenham:
Following #31730, this adds specialized typing support for the `bool | integer | object_` and `bool | integer | floating | object_` unary endo-ufuncs; `ceil`, `floor`, `trunc`, and `invert`.

No AI.

### DIFF
--- a/numpy/__init__.pyi
+++ b/numpy/__init__.pyi
@@ -404,6 +404,7 @@ from numpy._core.umath import (
     arctanh,
     arctanh as atanh,
     cbrt,
+    ceil,
     cos,
     cosh,
     deg2rad,
@@ -412,6 +413,7 @@ from numpy._core.umath import (
     exp2,
     expm1,
     fabs,
+    floor,
     log,
     log10,
     log1p,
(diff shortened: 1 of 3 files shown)


jorenham:
The new `jax` primer errors ([src](https://github.com/jax-ml/jax/blob/258fccb7ae1acbcd962e82548f057afac17a9892/jax/_src/numpy/lax_numpy.py#L5967-L5980)) are correct, because passing `ArrayLike` to `np.ceil` is indeed type-unsafe. Previously, this error was hidden, so this isn't a numpy regression.


The new primer error for `colour` ([src](https://github.com/colour-science/colour/blob/develop/colour/models/rgb/cylindrical.py#L192-L210)) seems to be caused by an inaccurate return type of their internal `as_float_array` function, which `np.floor` now no longer erases, leading to this mypy error for `np.choose` down the line. Not a numpy regression.

charris:
Thanks Joren.

PR #31742, opened 2026-06-24
TYP: specialized ``conj[ugate]``, ``reciprocal``, and ``square`` ufuncs

jorenham:
Following #31741, here's another batch of ufunc specializations, targeting the unary endo-ufuncs on `number | object_`; `conj[ugate]`, `reciprocal`, and `square`. 

Boolean input, although not in the type signature, required additional overloads, since boolean input results in `int8` output, so it couldn't be "absorbed" in the `int_` overloads like with the other specialized ufuncs (`sin` for example).

I also snuck in a tiny fix, so that obscure array-likes that aren't captured by any of the other overloads, will also be accepted if no `dtype` or `out` param is required. This is probably pretty rare in the wild (e.g. no related mymy primer errors), but it's an easy fix, so we might as well.

---

No AI.

### DIFF
--- a/numpy/__init__.pyi
+++ b/numpy/__init__.pyi
@@ -405,6 +405,8 @@ from numpy._core.umath import (
     arctanh as atanh,
     cbrt,
     ceil,
+    conjugate,
+    conjugate as conj,
     cos,
     cosh,
     deg2rad,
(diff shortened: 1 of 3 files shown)


charris:
Thanks Joren.

PR #31745, opened 2026-06-24
BUG: fix vdot returning None for empty object arrays

### LINKED ISSUE #31735: BUG: `vdot` of object arrays along zero-length is `None`
### Describe the issue:

Following the bug discussed in #31019 for `vecdot`, `np.vdot` appears to show similar behavior for zero-length object arrays: it returns `None` instead of a zero-like scalar result.

For comparison, with the `dtype=float`, the result is 0.0.  

```python
import numpy as np

cases = [
    np.empty((0,), dtype=object),
    np.empty((1,0), dtype=object),
    np.empty((0,1), dtype=object),
    np.empty((0,0), dtype=object),
]

for x in cases:
    print(x.shape, np.vdot(x, x))
```

### Error message:

```shell
Output: 

(0,) None
(1, 0) None
(0, 1) None
(0, 0) None
```

2.6.0.dev0+git20260623.4357ad1

Before reporting it, we reviewed the code, documentation and checked the latest development version of Numpy to confirm this observation.

MeGaurav4:
np.vdot on zero-length object arrays returns None instead of 0,
matching the same bug that was fixed in vecdot (#31033).

Add a zero-length guard in OBJECT_vdot to return PyLong_FromLong(0),
consistent with the numeric type behavior.

Closes #31735.

### DIFF
--- a/numpy/_core/src/multiarray/vdot.c
+++ b/numpy/_core/src/multiarray/vdot.c
@@ -145,6 +145,16 @@ OBJECT_vdot(char *ip1, npy_intp is1, char *ip2, npy_intp is2, char *op, npy_intp
     npy_intp i;
     PyObject *tmp0, *tmp1, *tmp2, *tmp = NULL;
     PyObject **tmp3;
+
+    if (n == 0) {
+        PyObject *zero = PyLong_FromLong(0);
+        if (zero == NULL) {
+            return;
+        }
+        Py_XSETREF(*((PyObject **)op), zero);
+        return;
+    }
+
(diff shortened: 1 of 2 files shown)


jorenham:
You ignored the PR template, so I'm going to assume that this PR has been made with agentic AI. We don't allow that here, because we prefer to talk to humans.

MeGaurav4:
@jorenham My apologies for skipping the PR template — that was my mistake. You're right that agentic AI was involved in drafting this PR, and I should have disclosed that upfront and followed the template. I'll make sure to do so in future contributions. Thanks for your time reviewing.

PR #31746, opened 2026-06-24
BUG: restrict weak scalar handling in resolve_dtypes to input operands

### LINKED ISSUE #31734: BUG: `ufunc.resolve_dtypes` segfaults
I was trying to figure out how to use `ufunc.resolve_dtypes`. So as per usual, I threw away the manual and instead just tried a couple of guesses.

First I tried passing `1` to `np.sin.resolve_dtypes`, that didn't work:

```
In [1]: np.sin.resolve_dtypes(1)
Cell In[1], line 1
----> 1 np.sin.resolve_dtypes(1)

TypeError: resolve_dtypes: The dtypes must be a tuple of `ufunc.nargs` length.
```

The error message seems helpful though; it must be a tuple (of size 2 in this case), so I tried with `(1, 1)` next:

```
In [2]: np.sin.resolve_dtypes((1, 1))
Cell In[2], line 1
----> 1 np.sin.resolve_dtypes((1, 1))

TypeError: Provided dtype must be a valid NumPy dtype, int, float, complex, or None.
```

Hmm, that didn't work either...
It mentions it accepts "dtype", I thought I'd try it with int *types*, i.e. `(int, int)`:

```
In [3]: np.sin.resolve_dtypes((int, int))
[1]    1121063 segmentation fault (core dumped)  ipython
```

😶 

---

I now realize that this works as intended with e.g. `resolve_dtypes((float, float))`.

This segfault clearly shouldn't have happened though; it should've raised a `TypeError` with a another helpful error message, like it did when I passed `(1, 1)` and `1`.

Diamagnetic:
### PR summary
This PR fixes a SegFault in `ufunc.resolve_dtypes` for calls like: `np.sin.resolve_dtypes((int, int))`

The bug:
- `int`/`float`/`complex` were treated as weak scalar markers in all operand positions.
- This means for output operands as well. This is invalid for those. And, it results in a segfault.

Fix:
- Restrict weak-scalar handling to input operands only.
- For output operands, builtin int/float/complex are interpreted as concrete NumPy dtypes.

Result:
- `np.sin.resolve_dtypes((int, int))` now raises a normal casting error (`UFuncOutputCastingError`) instead of crashing.
- Valid weak input behavior remains unchanged (e.g. `np.sin.resolve_dtypes((int, None))`).

Tests:
- Updated unary weak-scalar resolve_dtypes test coverage.

Hi all, I am new to the NumPy community. I use NumPy mainly in my data science, and graph theory projects. I wanted to make contributions to open source, and my procrastination finally ended recently. I was thinking of contributing to something I use regularly, and that's how I chose NumPy (I use it regularly). 

### AI Disclosure
- I used ChatGPT Go and OpenAI Codex (gpt-5.3-codex medium effort) to help with debugging and fixing the issue.
- I manually created the development environment, built numpy from source, reviewed, edited, and validated all code and tests.

Closes #31734

### DIFF
--- a/numpy/_core/src/umath/ufunc_object.c
+++ b/numpy/_core/src/umath/ufunc_object.c
@@ -6423,17 +6423,23 @@ py_resolve_dtypes_generic(PyUFuncObject *ufunc, npy_bool return_context,
             DTypes[i] = NPY_DTYPE(descr);
             Py_INCREF(DTypes[i]);
         }
-         /* Explicitly allow int, float, and complex for the "weak" types. */
+        /* Explicitly allow int, float, and complex for the "weak" types. */
         else if (descr_obj == (PyObject *)&PyLong_Type) {
             descr = PyArray_DescrFromType(NPY_INTP);
             dummy_arrays[i] = (PyArrayObject *)PyArray_Empty(0, NULL, descr, 0);
             if (dummy_arrays[i] == NULL) {
                 goto finish;
             }
-            PyArray_ENABLEFLAGS(dummy_arrays[i], NPY_ARRAY_WAS_PYTHON_INT);
-            Py_INCREF(&PyArray_PyLongDType);
-            DTypes[i] = &PyArray_PyLongDType;
-            promoting_pyscalars = NPY_TRUE;
+            if (i < ufunc->nin) {
+                PyArray_ENABLEFLAGS(dummy_arrays[i], NPY_ARRAY_WAS_PYTHON_INT);
+                Py_INCREF(&PyArray_PyLongDType);
+                DTypes[i] = &PyArray_PyLongDType;
+                promoting_pyscalars = NPY_TRUE;
+            }
+            else {
+                DTypes[i] = NPY_DTYPE(PyArray_DESCR(dummy_arrays[i]));
+                Py_INCREF(DTypes[i]);
+            }
         }
         else if (descr_obj == (PyObject *)&PyFloat_Type) {
             descr = PyArray_DescrFromType(NPY_DOUBLE);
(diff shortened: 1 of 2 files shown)


Diamagnetic:
Just following up on this PR in case it got buried. I would really appreciate a review when possible. Thanks in advance!

MaanasArora:
Thanks, this seems to be in the right place but I don't think the fix is right; a `UFuncOutputCastingError` is clearly the wrong error. Python types and concrete dtypes can't be interchanged like this.

Instead if the descriptor is an output descriptor, we should just skip the weak-scalar branches and then error (a nicer else-if error branch if `i < ufunc->nin` may be needed).

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -6423,39 +6423,57 @@ py_resolve_dtypes_generic(PyUFuncObject *ufunc, npy_bool return_context,
             DTypes[i] = NPY_DTYPE(descr);
             Py_INCREF(DTypes[i]);
         }
-         /* Explicitly allow int, float, and complex for the "weak" types. */
+        /* Explicitly allow int, float, and complex for the "weak" types. */
         else if (descr_obj == (PyObject *)&PyLong_Type) {
```
MaanasArora:
```suggestion
        else if (descr_obj == (PyObject *)&PyLong_Type && i < ufunc->nin) {
```

This would make sense as output descriptors can't use weak-scalar handling, then you would remove the new if-else branches, and maybe add a better error for input descriptors specifically.

Same for the other branches.

numpy/_core/tests/test_ufunc.py
```diff
@@ -3253,6 +3253,12 @@ def test_resolve_dtypes_basic(self):
         with pytest.raises(TypeError):
             np.add.resolve_dtypes((i4, f4, None), casting="no")
 
+    def test_resolve_dtypes_unary_weak_scalar(self):
+        assert np.sin.resolve_dtypes((int, None)) == (
+            np.dtype("f8"), np.dtype("f8"))
+        with pytest.raises(TypeError):
```
MaanasArora: We should add a regex match here for the error, as TypeError is very broad!

numpy/_core/src/umath/ufunc_object.c
MaanasArora, line 6473: Thanks, looks good now! We should probably add an else-if branch here, because the error below (the `int, float, complex` part) isn't exactly applicable for output descriptors anymore.

Diamagnetic: Thank you for the suggestion! I added an else-if branch for handling the output descriptor error.

numpy/_core/tests/test_ufunc.py
```diff
@@ -3253,6 +3253,12 @@ def test_resolve_dtypes_basic(self):
         with pytest.raises(TypeError):
             np.add.resolve_dtypes((i4, f4, None), casting="no")
 
+    def test_resolve_dtypes_unary_weak_scalar(self):
+        assert np.sin.resolve_dtypes((int, None)) == (
+            np.dtype("f8"), np.dtype("f8"))
+        with pytest.raises(TypeError, match=r".*[Oo]utput"):
```
MaanasArora: Thanks for the changes! Minor, but can we do an substring/exact match for the error here? That would be much more robust I think! (I could've been clearer when I said regex, I really just meant `match=`.)

Diamagnetic: Oh, now I get what you really meant. Thanks. I updated my code to have the exact match.

MaanasArora (approved):
Awesome thank you, looks good to me! Will put this in after CI passes.

PR #31747, opened 2026-06-24
ENH: Implement MaskedArray.tofile()

abolfazl-moeini:
`MaskedArray.tofile()` was not implemented and always raised `NotImplementedError`. That also meant `np.save()` failed on masked arrays, since it ends up calling `tofile()`.

This delegates to `filled(fill_value).tofile()`, same as `tobytes()`. Masked entries are written using the fill value; the mask itself is not stored in the file.

Updated the stub in `core.pyi` and replaced the old test that only checked for `NotImplementedError`.

### DIFF
--- a/numpy/ma/core.py
+++ b/numpy/ma/core.py
@@ -6403,20 +6403,52 @@ class MaskedArray(ndarray):
         """
         return self.filled(fill_value).tobytes(order=order)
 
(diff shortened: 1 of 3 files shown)


jorenham:
You ignored the PR template, so I'm going to assume that this PR has been made with agentic AI. We don't allow that here, because we prefer to talk to humans.

PR #31749, opened 2026-06-25
TYP: specialized ``positive``, ``negative``, ``sign``, and ``abs[olute]`` ufuncs

jorenham:
This one is pretty much the same story as #31742 and the other ufunc typing PRs. 

One new thing is that `abs[olute]` isn't an enfo-ufunc, because it maps complex floats to real floats, so I threw some additional overloads at those.

Oh and for `sign` there's also the `datetime64 -> float64` mapping that required some special casing (i.e. more overloads).

---

No AI.


jorenham:
The primer errors for `colour` look like the same ones as before: `ArrayLike` is being passed to these ufuncs, which is type-unsafe. So not a numpy regression.

Aniketsy:
i was just going through changes to understand, and i didn't understand what does this comment means `# bBhHiIlLqQefdgmFDGO => bBhHiIlLqQefdgFDGO, where m => d`  and got curious could you please explain ( i'm just asking for my understanding) thanks!

jorenham:
> i was just going through changes to understand, and i didn't understand what does this comment means `# bBhHiIlLqQefdgmFDGO => bBhHiIlLqQefdgFDGO, where m => d` and got curious could you please explain ( i'm just asking for my understanding) thanks!

That's the dtype signature in terms of the dtype char codes, obtained from the `ufunc.types`.

Aniketsy:
> > i was just going through changes to understand, and i didn't understand what does this comment means `# bBhHiIlLqQefdgmFDGO => bBhHiIlLqQefdgFDGO, where m => d` and got curious could you please explain ( i'm just asking for my understanding) thanks!
> 
> That's the dtype signature in terms of the dtype char codes, obtained from the `ufunc.types`.

got it thanks :)

charris:
Thanks Joren.

PR #31753, opened 2026-06-25
TST: Fix failing riscv64 tests.

charris:
The problem was an expired key.

AI suggested the fix.


charris:
Self merging, the problem still turns up.

PR #31755, opened 2026-06-25
ENH: Expand abstract DType hierarchy

seberg:
Introduces ``NumericAbstractDType`` and a fuller numeric "kind" tree mirroring the array-API hierarchy:

    NumericAbstractDType
        IntegerAbstractDType
            SignedIntegerAbstractDType
            UnsignedIntegerAbstractDType
        InexactAbstractDType
            FloatAbstractDType
            ComplexAbstractDType

The new abstract DTypes are exposed on ``numpy.dtypes`` and the legacy concrete DTypes inherit from them so ``isinstance``/``issubclass`` reflect the kind hierarchy.

Adds ``register()`` on abstract DTypes for legacy/user-defined DTypes to opt into a kind, with custom ``__subclasscheck__`` / ``__instancecheck__`` on the metaclass consulting the registered cache and the ``NPY_DT_NUMERIC`` flag.

``np.isdtype`` now uses and accepts DType classes (including these abstracts), although passing a non-instance as first argument is deprecated as I think it is a bit wonky.
(Do allow the scalar types as they map OK, but... really that shouldn't be the normal use-case, I think.)

(I used a bit of AI to implement some larger patterns, and some of the tests are; I'll go through once more myself as well.)


jorenham:
I don't think it's possible to make this type-safe; because there's no mechanism to "tell" static-type checkers that some registered custom dtype should be assignable to the relevant absract dtype if that custom dtype isn't an actual subclass of it.

This is why using the `register` method from `abc` types is often discouraged, and is also why the static types of `collections.abc` and stdlib subclasses thereof are such a mess today.

The abstract dtypes themselves seem like a good idea to me though, it's just the `register` and `__isinstance__` magic that I'm against.

jorenham:
Naming-wise, should `FloatAbstractDType` -> `FloatingAbstractDType`, and `ComplexAbstractDType` -> `ComplexFloatingAbstractDType`, for consistency with `np.floating` / `np.complexfloating`? Similarly it would be more consistent to call `NumericAbstractDType` -> `NumberAbstractDType` so that it matches `np.number` I think

seberg:
> The abstract dtypes themselves seem like a good idea to me though, it's just the register and __isinstance__ magic that I'm against.

Hmmm, maybe we don't need it strictly, although it is nice to tape over things for old-style dtypes.  But... I hope to move `ml_dtypes` so maybe that doesn't actually matter too much if you think `register` is too super annoying.

One nice thing about registering (in the future) is that it allows easy creation for custom "superclasses" that also match NumPy if you want to and that can be very nice for promotion logic (i.e. to spell out `Int64DType|MyDType`).  But... maybe we can spell that in a different way nicely, and then I am not sure it matters too much to allow `.register()`.
(I somewhat like it to be the same as the `abc`, but I guess ABC predate typing, etc...)

> ComplexFloatingAbstractDType

Yeah, I would prefer that, we just already ahve it in C... Maybe we can just add an alias there for now...

seberg:
Updated this and renamed the Python side for now only.  `.register()` is removed but for now the Numeric flag is special-cased for subclasscheck (i.e. quaddtype uses it).

(So maybe I am not 100% sure if there isn't some follow-up, but I think this should be ready -- although I should probably go through once more myself due to that back and forth.)

jorenham (commented):
don't forget about the `numpy.dtypes` stubs :)
https://github.com/numpy/numpy/blob/7e1f94485495485c5cc3a408ab9e945940f1b91f/numpy/dtypes.pyi

jorenham:
> don't forget about the `numpy.dtypes` stubs :) [`7e1f944`/numpy/dtypes.pyi](https://github.com/numpy/numpy/blob/7e1f94485495485c5cc3a408ab9e945940f1b91f/numpy/dtypes.pyi)

(or, if you prefer, I could also write them)

seberg:
Sorry that this was not updated for a while.  @jorenham I now pushed an update, the types are pretty much added by claude, they seemed reasonable, but can you take a look?

I changed things now to:
* Adopted your names, it makes sense to stick to the old ones in Python... (without even thinking much about what is nicer...)
* Rip out `register` as much as I like it, I guess we should be able to get away without it mostly.  This simplifies things a lot of course.
* For now I removed the fast-path to check the flag
* There is a fun thing where our `_is_numeric` flag is true for booleans but it's excluded from `NumberAbstractDType`.  We may want to move away from that at some point.

numpy/dtypes.pyi
```diff
@@ -125,6 +132,65 @@ class _NBit[AlignmentT: int, ItemSizeT: int]:
...
+class _FloatingDType[ScalarT: np.floating](  # type: ignore[misc]
+    _LiteralDType[ScalarT],
+    FloatingAbstractDType[ScalarT],
+): ...
+
+@type_check_only
+class _ComplexFloatingDType[ScalarT: np.complexfloating](  # type: ignore[misc]
+    _LiteralDType[ScalarT],
+    ComplexFloatingAbstractDType[ScalarT],
+): ...
```
jorenham, line 192: You also need some `# pyrefly:ignore[invalid-inheritance]` here to get the Pyrefly CI job (https://github.com/numpy/numpy/actions/runs/34870635499/job/104065361599?pr=31755) green:

```
ERROR Cannot extend final class `SignedIntegerAbstractDType` [invalid-inheritance]
   --> numpy/dtypes.pyi:173:5
    |
173 |     SignedIntegerAbstractDType[ScalarT],
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
ERROR Cannot extend final class `UnsignedIntegerAbstractDType` [invalid-inheritance]
   --> numpy/dtypes.pyi:179:5
    |
179 |     UnsignedIntegerAbstractDType[ScalarT],
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
ERROR Cannot extend final class `FloatingAbstractDType` [invalid-inheritance]
   --> numpy/dtypes.pyi:185:5
    |
185 |     FloatingAbstractDType[ScalarT],
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
ERROR Cannot extend final class `ComplexFloatingAbstractDType` [invalid-inheritance]
   --> numpy/dtypes.pyi:191:5
    |
191 |     ComplexFloatingAbstractDType[ScalarT],
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
 INFO 4 errors (407 suppressed)
```

numpy/dtypes.pyi
```diff
@@ -125,6 +132,65 @@ class _NBit[AlignmentT: int, ItemSizeT: int]:
...
+#
+# These mirror the numeric "kind" hierarchy of the scalar types.  They cannot
+# be instantiated and are meant for `isinstance`/`issubclass` checks and as the
+# `kind` argument of `numpy.isdtype`.
+#
+# NOTE: Unlike at runtime they are generic in their scalar type, so that the
+# concrete DTypes below can inherit from both `_LiteralDType[ScalarT]` and the
+# matching abstract DType without passing `numpy.dtype` conflicting type
+# arguments.  Subscripting them is supported at runtime as well, since
+# `numpy.dtype.__class_getitem__` is inherited.
```
jorenham, line 145: I don't think this prose is necessary.

numpy/dtypes.pyi
```diff
@@ -125,6 +132,65 @@ class _NBit[AlignmentT: int, ItemSizeT: int]:
...
+@final
+class InexactAbstractDType[ScalarT: np.inexact](NumberAbstractDType[ScalarT]): ...  # type: ignore[misc]
+
+@final
+class FloatingAbstractDType[ScalarT: np.floating](InexactAbstractDType[ScalarT]): ...  # type: ignore[misc]
+
+@final
+class ComplexFloatingAbstractDType[ScalarT: np.complexfloating](InexactAbstractDType[ScalarT]): ...  # type: ignore[misc]
+
+# Helper bases tying the concrete DTypes to their abstract "kind" (typing-only):
```
jorenham, line 168: this seems kinda obvious to me looking at the code below, so let's remove this

numpy/dtypes.pyi
jorenham, line 192: Also, maybe call these e.g. `_SignedIntegerLiteralDType`, so that it's obvious from the name that these are intersections of the abstract dtype types and the `_LiteralDType` helper type?

numpy/typing/tests/data/pass/dtype.py
jorenham, line 1: These added tests mostly test type-narrowing behavior of type-checkers themselves, not the stub definitions. 

Also, `assert_type` tests like these are conventionally placed in `typing/tests/data/reveal`.

But TBH, I don't think these need type-tests at all: there are no complicated overloads or anything, just static declarations, so there's really nothing interesting to test 🤷

jorenham (commented):
Looks a lot better now; nice.

I left some typing-related comments, most of them about stuff that can be removed.

PR #31756, opened 2026-06-25
BUG: restore ndarray.conjugate() for legacy user-defined dtypes

### LINKED ISSUE #31754: BUG: ndarray.conjugate() regression for legacy user-defined dtypes in 2.5.0
### Describe the issue:

`ndarray.conjugate()` (and `.conj()`) raises `TypeError: cannot conjugate non-numeric dtype` for legacy (pre-NEP-42) user-defined dtypes in NumPy 2.5.0. It worked in 2.4.x and earlier. This affects any dtype registered via `PyArray_RegisterDataType`, including ones that register their own `conjugate` ufunc loop. The `np.conjugate(arr)` *ufunc* is unaffected — only the `ndarray.conjugate()`/`.conj()` *method* regressed.

The example below uses NumPy's own in-tree `rational` test dtype, so it needs no third-party packages:

- **2.4.6:** `a.conjugate()` → `[1 2 3]` (no-op, correct for this real-valued dtype)
- **2.5.0:** `a.conjugate()` → `TypeError: cannot conjugate non-numeric dtype`

A real-world case is [numpy-quaternion](https://github.com/moble/quaternion), whose dtype registers a genuine `conjugate` ufunc loop; `arr.conjugate()` used to compute the quaternion conjugate via that loop and now errors (moble/quaternion#270).

**Root cause:** #31264 / commit b6bc683 ("ENH: Allow user-dtypes return views on `.conjugate()`") rewrote the dispatch in
`PyArray_Conjugate` (`numpy/_core/src/multiarray/calculation.c`):

```diff
-    if (PyArray_ISCOMPLEX(self) || PyArray_ISOBJECT(self) ||
-            PyArray_ISUSERDEF(self) || !NPY_DT_is_legacy(PyArray_DESCR(self))) {
+    if (NPY_DT_SLOTS(NPY_DTYPE(PyArray_DTYPE(self)))->imag_meth != NULL) {
```

The old condition routed user-defined dtypes (`PyArray_ISUSERDEF`) to the `conjugate` ufunc. The new one routes only dtypes whose DType defines an `imag_meth` slot. Legacy user dtypes have no `imag_meth` (a NEP-42/43 slot) and are never flagged `NPY_DT_NUMERIC` (for legacy descriptors that flag is set only when `PyTypeNum_ISNUMBER(descr->type_num)`, always false for `type_num >= NPY_USERDEF`), so they fall through to the new `!NPY_DT_is_numeric(...)` branch and error — even when a `conjugate` loop is registered.

(For clarity: the `"cannot conjugate non-numeric dtype"` *message* predates this and came from the expired no-op-fallthrough deprecation; that expiry was fine, since in 2.4.x the `PyArray_ISUSERDEF` carve-out kept user dtypes off that path. The regression is specifically the removal of that carve-out in #31264, which doesn't appear to have been intended to change legacy-dtype behavior.)

**Expected behavior:** as in 2.4.x — dispatch to the registered `conjugate` ufunc loop if one exists, otherwise return a view (real-valued no-op). Legacy user dtypes should not unconditionally error.

```python
import numpy as np
from numpy._core._rational_tests import rational

a = np.array([1, 2, 3], dtype=rational)
print("dtype:", a.dtype, "type_num:", a.dtype.num)  # rational 256
print("np.conjugate (ufunc) works:", np.conjugate(a))  # [1 2 3]
print(a.conjugate())  # TypeError on 2.5.0, was [1 2 3] on 2.4.6
```

### Error message:

```shell
  File "<python-input-0>", line 7, in <module>
    print(a.conjugate())  # TypeError on 2.5.0, was [1 2 3] on 2.4.6
          ~~~~~~~~~~~^^
TypeError: cannot conjugate non-numeric dtype
```

2.5.0
3.14.6 | packaged by conda-forge | (main, Jun 12 2026, 08:47:56) [Clang 20.1.8 ]

### Runtime Environment:

[{'numpy_version': '2.4.6',
  'python': '3.14.6 | packaged by conda-forge | (main, Jun 12 2026, 08:47:56) '
            '[Clang 20.1.8 ]',
  'uname': uname_result(system='Darwin', node='Beasty', release='23.6.0', version='Darwin Kernel Version 23.6.0: Mon Jul 29 21:13:04 PDT 2024; root:xnu-10063.141.2~1/RELEASE_ARM64_T6020', machine='arm64')},
                      'not_found': ['ASIMDFHM']}},
 {'ignore_floating_point_errors_in_matmul': True},
 {'architecture': 'VORTEX',
  'filepath': '/Users/boyle/.continuum/mambaforge/envs/science/lib/libopenblas.0.dylib',
  'internal_api': 'openblas',
  'num_threads': 12,
  'prefix': 'libopenblas',
  'threading_layer': 'openmp',
  'user_api': 'blas',
  'version': '0.3.33'},
 {'filepath': '/Users/boyle/.continuum/mambaforge/envs/science/lib/libomp.dylib',
  'internal_api': 'openmp',
  'num_threads': 12,
  'prefix': 'libomp',
  'user_api': 'openmp',
  'version': None}]

Found via a downstream bug report against the quaternion package (moble/quaternion#270): a user's array-conjugation code that worked on numpy <= 2.4.6 began raising TypeError on 2.5.0. The same regression reproduces on NumPy's own in-tree `rational` dtype, so it affects any legacy user-defined dtype, not just quaternion.

moble:
### PR summary

#31264 rewrote PyArray_Conjugate to dispatch to the conjugate ufunc only when the DType defines an `imag_meth` slot, dropping the explicit `PyArray_ISUSERDEF` route. Legacy user dtypes have no `imag_meth` and are not flagged numeric, so `arr.conjugate()` began raising "cannot conjugate non-numeric dtype" in 2.5.0, even when a conjugate ufunc loop is registered (e.g. numpy-quaternion).

Restore the legacy user-dtype route so these dtypes dispatch to the ufunc as before, while leaving the new-style view/numeric behavior intact.

Closes #31754

#### AI Disclosure
I used Claude to find where the original bug came from, and to check through the docs and commit history to convince myself that this was the right approach.

### DIFF
--- a/numpy/_core/src/multiarray/calculation.c
+++ b/numpy/_core/src/multiarray/calculation.c
@@ -828,8 +828,16 @@ PyArray_Clip(PyArrayObject *self, PyObject *min, PyObject *max, PyArrayObject *o
 NPY_NO_EXPORT PyObject *
 PyArray_Conjugate(PyArrayObject *self, PyArrayObject *out)
 {
-    if (NPY_DT_SLOTS(NPY_DTYPE(PyArray_DTYPE(self)))->imag_meth != NULL) {
-        /* The dtype has `arr.imag` so `conjugate` must exist (or error) */
+    if (NPY_DT_SLOTS(NPY_DTYPE(PyArray_DTYPE(self)))->imag_meth != NULL
+            || PyArray_ISUSERDEF(self)) {
+        /*
+         * The dtype has `arr.imag` so `conjugate` must exist (or error).
+         * Legacy user-defined dtypes (`type_num >= NPY_USERDEF`) have no
+         * `imag_meth` and are not flagged numeric, but they may register a
+         * `conjugate` ufunc loop (e.g., quaternion), so dispatch to the ufunc
+         * for them as well, preserving the pre-2.5 behavior.  (New-style
+         * user dtypes have `type_num == -1` and are unaffected by this.)
+         */
         if (out == NULL) {
             return PyArray_GenericUnaryFunction(self,
                                                 n_ops.conjugate);
--- a/numpy/_core/tests/test_multiarray.py
+++ b/numpy/_core/tests/test_multiarray.py
@@ -4535,6 +4535,14 @@ class TestMethods:
         with pytest.raises(TypeError, match="cannot conjugate non-numeric dtype"):
             a.conj()
 
+    def test_conjugate_legacy_user_dtype(self):
+        # gh-31754 : legacy user-defined dtypes are not flagged numeric and have
+        # no `.imag`, but `conjugate` must still dispatch to the ufunc (which
+        # falls back to a cast for `rational`) rather than raising.
+        a = np.array([1, 2, 3], dtype=rational)
+        assert_array_equal(a.conjugate(), np.conjugate(a))
+        assert_array_equal(a.conjugate(), [1, 2, 3])
+
     def test_conjugate_out(self):
         # Minimal test for the out argument being passed on correctly
         # NOTE: The ability to pass `out` is currently undocumented!


moble:
Actually, I notice that that the version appears to be 2.6.0-dev.  Could this be backported to 2.5, if relevant?

numpy/_core/src/multiarray/calculation.c
```diff
@@ -828,8 +828,16 @@ PyArray_Clip(PyArrayObject *self, PyObject *min, PyObject *max, PyArrayObject *o
...
+    if (NPY_DT_SLOTS(NPY_DTYPE(PyArray_DTYPE(self)))->imag_meth != NULL
+            || PyArray_ISUSERDEF(self)) {
+        /*
+         * The dtype has `arr.imag` so `conjugate` must exist (or error).
+         * Legacy user-defined dtypes (`type_num >= NPY_USERDEF`) have no
+         * `imag_meth` and are not flagged numeric, but they may register a
+         * `conjugate` ufunc loop (e.g., quaternion), so dispatch to the ufunc
+         * for them as well, preserving the pre-2.5 behavior.  (New-style
+         * user dtypes have `type_num == -1` and are unaffected by this.)
+         */
```
ngoldbaum, line 840: I prefer not leaving comments behind that memorialize old bugs - can you cut this down some?

ngoldbaum (commented):
Seems OK to me but @seberg would know better than me if there are subtle issues.

I added a backport-candidate label to signal that this should be backported on merge.

seberg:
I don't really think this is the best approach: This way, we lose a window of opportunity to start returning views even when `PyArray_ISUSERDEF` is true which is possible when using the new API but to create a legacy-style DType (for compat reasons).

ngoldbaum:
We're just going to close this, please open a new pull request if you're still interested in working on this.

PR #31759, opened 2026-06-25
BUG: aligned access for the descriptor fields struct on 32bit abi3t

ngoldbaum:
Over in https://github.com/PyO3/rust-numpy/pull/556 work is ongoing to add abi3t support for crates that depend on PyO3 and rust-numpy.

There is [a failing test run](https://github.com/PyO3/rust-numpy/actions/runs/27867012031/job/82473307007?pr=556) for win32 3.15t, in a test that accesses the `metadata` field for a datetime dtype instance.

I was able to track this down to the uint64 `flags` field on the `PyArray_Descr_fields` and `_PyArray_LegacyDescr_fields` structs.

On a 32 bit free-threaded build, `sizeof(PyObject)` is 20 bytes. That places the `flags` field at a 4-byte aligned address instead of an 8-byte aligned address, the compiler inserts some padding bytes, and suddenly the fields access function is returning addresses at the wrong offset for all fields *after* the `flags` entry.

The fix I landed on was to define the `flags` as two 32 bit integers. Each gets read individually, and the compiler inserts no more packing bytes into the struct. Inside the `flags` access macro, we re-assemble the lo and hi parts of the 64 bit int.

Also adds a static assert in case a future change to `sizeof(PyObject)` breaks our assumptions here. Also adds a regression test to access the datetime metadata field, like the rust-numpy test did.

I used Claude to help identify and fix this bug.

### DIFF
--- a/numpy/_core/include/numpy/ndarraytypes.h
+++ b/numpy/_core/include/numpy/ndarraytypes.h
@@ -616,6 +616,23 @@ typedef struct {
                                 NPY_ITEM_IS_POINTER | NPY_ITEM_REFCOUNT | \
                                 NPY_NEEDS_INIT | NPY_NEEDS_PYAPI)
 
+/*
+ * Under Py_TARGET_ABI3T these field structs are overlaid at
+ * `(char *)obj + sizeof(PyObject)`, which is only pointer-aligned on 32-bit
+ * free-threaded builds (there sizeof(PyObject) is 20 bytes). An 8-byte
+ * `npy_uint64 flags` member would be over-aligned and packing bytes would shift
+ * every later field, so expose it as two 32-bit halves.
+ */
+#if defined(Py_TARGET_ABI3T)
+    #if NPY_BYTE_ORDER == NPY_BIG_ENDIAN
+        #define NPY_DT_FLAGS_FIELD npy_uint32 _flags_hi; npy_uint32 _flags_lo;
+    #else
+        #define NPY_DT_FLAGS_FIELD npy_uint32 _flags_lo; npy_uint32 _flags_hi;
+    #endif
+#else
+    #define NPY_DT_FLAGS_FIELD npy_uint64 flags;
+#endif
+
 #if NPY_FEATURE_VERSION >= NPY_2_0_API_VERSION
 /*
  * Public version of the Descriptor struct as of 2.x
@@ -647,7 +664,7 @@ typedef struct _PyArray_Descr_fields {
         /* number representing this type */
         int type_num;
         /* Space for dtype instance specific flags. */
-        npy_uint64 flags;
+        NPY_DT_FLAGS_FIELD
         /* element size (itemsize) for this type */
         npy_intp elsize;
         /* alignment needed for this type */
@@ -708,7 +725,7 @@ typedef struct {
         char byteorder;
         char _former_flags;
         int type_num;
-        npy_uint64 flags;
+        NPY_DT_FLAGS_FIELD
         npy_intp elsize;
         npy_intp alignment;
         PyObject *metadata;
--- a/numpy/_core/include/numpy/npy_2_compat.h
+++ b/numpy/_core/include/numpy/npy_2_compat.h
@@ -151,7 +151,11 @@ PyArray_ImportNumPyAPI(void)
     static inline npy_uint64
     PyDataType_FLAGS(const PyArray_Descr *dtype)
     {
-    #if NPY_FEATURE_VERSION >= NPY_2_0_API_VERSION
+    #if defined(Py_TARGET_ABI3T)
+        /* flags is stored as two 32-bit halves here (see ndarraytypes.h). */
+        const PyArray_Descr_fields *_f = _PyDataType_GET_ITEM_DATA(dtype);
+        return ((npy_uint64)_f->_flags_hi << 32) | (npy_uint64)_f->_flags_lo;
+    #elif NPY_FEATURE_VERSION >= NPY_2_0_API_VERSION
         return _PyDataType_GET_ITEM_DATA(dtype)->flags;
     #else
         return (unsigned char)dtype->flags;  /* Need unsigned cast on 1.x */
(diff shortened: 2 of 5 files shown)


numpy/_core/include/numpy/ndarraytypes.h
```diff
@@ -616,6 +616,23 @@ typedef struct {
...
                                 NPY_NEEDS_INIT | NPY_NEEDS_PYAPI)
 
+/*
+ * Under Py_TARGET_ABI3T these field structs are overlaid at
+ * `(char *)obj + sizeof(PyObject)`, which is only pointer-aligned on 32-bit
+ * free-threaded builds (there sizeof(PyObject) is 20 bytes). An 8-byte
+ * `npy_uint64 flags` member would be over-aligned and packing bytes would shift
+ * every later field, so expose it as two 32-bit halves.
+ */
+#if defined(Py_TARGET_ABI3T)
```
ngoldbaum, line 626: Maybe this condition should also include e.g. `&& IS_32BIT`.

ngoldbaum: Hmm, on second thought I bet this doesn't work at all on a 32-bit GIL-enabled interpreter (which might run an extension built with `Py_TARGET_ABI3T`, which implies `Py_GIL_DISABLED`), since PyObject is 8 bytes there and there's no padding to correct for...

ngoldbaum: OK, I think what we actually need to fix 32 bit builds is *new explicit accessor API functions* for all of the fields after `flags` that might be impacted by this alignment/packing issue. We can handle all of this in NumPy, we just can't handle it in the single `GET_ITEM_DATA` API function we have now.

That also probably means 32bit builds abi3t can't be supported until NumPy 2.6 because we need API functions to fix this ☹️.

Ping @seberg.

ngoldbaum:
I'm going to close this PR because the approach is unsound and open a followup issue.

seberg:
> On a 32 bit free-threaded build, sizeof(PyObject) is 20 bytes. That places the flags field at a 4-byte aligned address instead of an 8-byte aligned address, the compiler inserts some padding bytes, and suddenly the fields access function is returning addresses at the wrong offset for all fields after the flags entry.

All of this sounds like a Python bug though?  Python has to guarantee `max_align_t` alignment of the actual fields.

EDIT: OK, it does guarantee for the whole object, but if `PyObject` isn't a multiple of that it seems to me we are in deep trouble.  At that point the whole `PyObject_GetTypeData()` function becomes useless for many objects.  And I _think_ the whole thing is broken, because `.tp_basicsize = -sizeof(my_struct)` is unclear that it'll do the right thing either, because it'll presumably **not** ensure padding to `max_align_t` there.

Or in other words: The only real solution to me seems for Python to ensure `max_align_t` size in a sense.  That may not sound great, but anything else means that these patterns becomes unusable without extreme care in 32bit.

da-woods:
> And I think the whole thing is broken, because `.tp_basicsize = -sizeof(my_struct)` is unclear that it'll do the right thing either, because it'll presumably not ensure padding to `max_align_t` there.

I believe that with will align to max_align_t (https://github.com/python/cpython/blob/285d96dd784620fe8afb02688abe72a916555061/Objects/typeobject.c#L5133 and https://github.com/python/cpython/blob/285d96dd784620fe8afb02688abe72a916555061/Objects/typeobject.c#L6088).

That makes `PyObject_GetTypeData` correct on an object declared with negative size (which is how it's intended to be used).

But I think it makes the assumption that

```
struct S{
   PyObject_HEAD
   int data1;
   int data2;
};
```

is safe to access through a

```
struct T{
    int data1;
    int data2;
};
```

pointer dodgy. (I think I did claim it was dodgy at the time, but failed to identify *this* as the reason).

So it really means that it's only safe if you consistently treat a `PyObject*` and opaque or non-opaque.

seberg:
Ah, thanks for the note @da-woods, right.  If Python does insert the extra alignment for such types, then a type defined via the opaque API can talk to others safely.

But it is still super annoying :(.  Sure, using accessors functions works, but it means that we are far worse off then stable ABI objects here.  I don't find it very enticing.

Part of me wonders if the right thing wouldn't be to just add the extra padding in NumPy if running on Python 3.15+.  That breaks all existing Python 3.15 downstream that are already build against NumPy 2.5 which are also:
* 32bit
* free-threaded
* (It doesn't break other builds, because for them nothing changes.)

But :shrug:, waiting for 3.16+ wouldn't be better, it's not like NumPy has a release for 3.15+ available yet, so in a sense we can do that?

Of course... We can also rip this all out and introduce a giant family of getter functions...  In theory this affects pretty much all our objects (in practice it _may_ not but for those others adding padding would also do nothing).

(EDIT: I dunno, maybe that is crazy and Cython might be a bit annoying about it as well if the size is Python dependent?  And I just really need to warm up to "accessors all the way" even if everything is designed to _not_ require this.
But I of course already find a per-object GetFields function an unnecessary hedge for a future where ABI flexibility reasons that I really don't begin to see being useful as it seems to assume opaque API as the only API and even then seems rather pointless to me.)

seberg:
> pointer dodgy. (I think I did claim it was dodgy at the time, but failed to identify this as the reason).

We were discussing it being dodgy because of aliasing rules mostly, IIRC.  If you still think that, then we would have to just stop this all.  For the alignment issue, manual alignment is at least conceivable (i.e. align the normal ABI with what an *opaque* API would do in practice for its objects as I said above).

da-woods:
> > pointer dodgy. (I think I did claim it was dodgy at the time, but failed to identify this as the reason).
> 
> We were discussing it being dodgy because of aliasing rules mostly, IIRC.

Yes indeed. I can't claim to have had an really useful advanced insight here. 

> Cython might be a bit annoying about it as well if the size is Python dependent?

For the moment I'm assuming that it's hard to mix opaque and non-opaque modules useful (or at least, they can only mix one way because we can always make things opaque). The only complication is the memoryview internal classes. I'll have to go back and check if they have this alignment issue on 32 bits. If not, it'll be luck rather than judgment.

PR #31760, opened 2026-06-25
TYP: specialized ``numpy.strings`` ufuncs

jorenham:
Following #31749, this adds specialized typing support for the (unary) `numpy.strings` ufuncs:

- `isdecimal`
- `isnumeric`
- `isalnum`
- `isalpha`
- `isdigit`
- `islower`
- `isspace`
- `istitle`
- `isupper`
- `str_len`

These are a bit simpler because each of them has a single output dtype (`bool` / `int_`). For some reason these also have `ufunc.identity` set (`False` / `0`), so I also annotated those as literal types (I thought those were only used by binary ufuncs? Oh well, it doesn't hurt I guess).

---

No AI


numpy/_core/defchararray.pyi
```diff
@@ -20,6 +20,19 @@ from numpy._typing import (
...
 
+from .strings import (
+    isalnum,
+    isalpha,
+    isdecimal,
+    isdigit,
+    islower,
+    isnumeric,
+    isspace,
+    istitle,
```
charris, line 31: That's an interesting one.

charris:
Thanks Joren.

numpy/_core/defchararray.pyi
jorenham: it's a better match with runtime this way 🤷 : https://github.com/jorenham/numpy/blob/15b9b4871d852ba19e151efbc2918ccf9e74bc43/numpy/_core/defchararray.py#L30-L35

PR #31763, opened 2026-06-26
MAINT: Enable E501 linting for numpy/_core/_add_newdocs.py

abheetsharma09:
Closes #28947 (partially)

This PR enables the Ruff E501 rule for `numpy/_core/_add_newdocs.py` by:
- Removing `E501` from the per-file-ignores list in `ruff.toml`
- Fixing all lines that exceeded the 88-character limit

All long docstrings have been split naturally, and long function signatures inside docstrings have been exempted with `# noqa: E501` to preserve documentation formatting.

### DIFF
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -5900,7 +5900,7 @@ add_newdoc('numpy._core', 'ufunc', ('resolve_dtypes',
 add_newdoc('numpy._core', 'ufunc', ('_resolve_dtypes_and_context',
     """
     _resolve_dtypes_and_context($self, dtypes, *, signature=None, 
-    casting=None, reduction=False) # noqa: E501
+                                                    casting=None, reduction=False) # noqa: E501
     --
 
     _resolve_dtypes_and_context(dtypes, *, signature=None, 
--- a/numpy/_core/_add_newdocs.py
+++ b/numpy/_core/_add_newdocs.py
@@ -391,7 +391,8 @@ add_newdoc('numpy._core', 'nditer',
     Here is an example function which operates like a "lambda" ufunc:
 
     >>> def luf(lamdaexpr, *args, **kwargs):
-    ...    '''luf(lambdaexpr, op1, ..., opn, out=None, order='K', casting='safe', buffersize=0)'''
+    ...    '''luf(lambdaexpr, op1, ..., opn, out=None, order='K', 
+    casting='safe', buffersize=0)'''
     ...    nargs = len(args)
     ...    op = (kwargs.get('out',None),) + args
     ...    it = np.nditer(op, ['buffered','external_loop'],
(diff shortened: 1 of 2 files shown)


numpy/_core/_add_newdocs.py
```diff
@@ -5881,10 +5899,12 @@ def _array_method_doc(name: str, params: str, doc: str) -> None:
 
 add_newdoc('numpy._core', 'ufunc', ('_resolve_dtypes_and_context',
     """
-    _resolve_dtypes_and_context($self, dtypes, *, signature=None, casting=None, reduction=False)
+    _resolve_dtypes_and_context($self, dtypes, *, signature=None, 
+    casting=None, reduction=False) # noqa: E501
```
eendebakpt:
```suggestion
    _resolve_dtypes_and_context($self, dtypes, *, signature=None, 
                                                    casting=None, reduction=False) # noqa: E501
```
(and similar for several other locations)

ruff.toml
```diff
@@ -81,7 +81,7 @@ ignore = [
 "bench_*.py" = ["B015", "B018", "RUF012"]
 "test*.py" = ["B015", "B018", "E201", "E714", "RUF012"]
 
-"numpy/_core/_add_newdocs.py" = ["E501"]
+"numpy/_core/_add_newdocs.py" = []
```
eendebakpt:
```suggestion
```

abheetsharma09:
hlo..there did i do something wrong??

jorenham:
> hlo..there did i do something wrong??

you forgot to fill in the required AI disclosure from the PR template

numpy/_core/_add_newdocs.py
```diff
@@ -5881,10 +5899,12 @@ def _array_method_doc(name: str, params: str, doc: str) -> None:
 
 add_newdoc('numpy._core', 'ufunc', ('_resolve_dtypes_and_context',
     """
-    _resolve_dtypes_and_context($self, dtypes, *, signature=None, casting=None, reduction=False)
+    _resolve_dtypes_and_context($self, dtypes, *, signature=None, 
+                                                    casting=None, reduction=False) # noqa: E501
```
jorenham, line 5903: this line should not be touched because it affects the ([argument clinic](https://docs.python.org/3.10/howto/clinic.html)) runtime type signature (`inspect.signature`)

numpy/_core/_add_newdocs.py
```diff
@@ -3284,7 +3290,8 @@
 
 add_newdoc('numpy._core.multiarray', 'ndarray', ('argpartition',
     """
-    argpartition($self, kth, /, axis=-1, kind='introselect', order=None, descending=None)
+    argpartition($self, kth, /, axis=-1,
+      kind='introselect', order=None, descending=None) # noqa: E501
```
jorenham, line 3294: this line should not be touched because it affects the ([argument clinic](https://docs.python.org/3.10/howto/clinic.html)) runtime type signature (`inspect.signature`)

jorenham (changes requested):
I'm not sure we should apply E501 to this file, because as I also said in the individual comments; the docstring formatting of the docstrings here has a real impact on the runtime behavior.

jorenham:
Thinking about it a bit more, I realized that it's impossible to ignore `E501` for the individual argument-clinic signatures without breaking them. So I don't see any way to enforce `E501` on this file without breaking a bunch of stuff. 

So thanks for your contribution, but I'm afraid I'm going to have to close this.

PR #31780, opened 2026-06-27
DOC: Remove Python 2 specific references from docs

abheetsharma09:
This PR removes outdated references to python2 from documentation and code comments, following the instructions in issue #22635. I confirmed locally that all instances outside of release notes and changelogs have been updated.

AI Usage Disclosure:
I used Claude to help me understand the contribution workflow. I made the changes myself and reviewed them before submitting.

### DIFF
--- a/doc/neps/nep-0012-missing-data.rst
+++ b/doc/neps/nep-0012-missing-data.rst
@@ -529,7 +529,7 @@ Some examples::
     >>> np.mean(a)
     NA(dtype='<f8')
     >>> np.mean(a, skipna=True)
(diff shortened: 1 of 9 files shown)


rkern:
Some of these should not have been changed for the reasons documented in the issue you cite as taking instruction from.
Some of the rest are just not useful (and missed the point of the issue: to remove references to the `python2` executable, not just all strings "python2", some of just referred correctly to Python 2).
Some have introduced new errors.
There are a couple here that are notionally correct, but not worth fixing up this PR to include them, IMO. It's such a low-priority issue.

PR #31781, opened 2026-06-27
ENH: Add np.median_sorted_arrays with O(log(min(m,n))) implementation

ManvithReddyyy:
This pull request proposes a new public API, `np.median_sorted_arrays`, for efficiently computing the median of two already sorted one-dimensional arrays without concatenating or merging them.

### PR Summary

This proposal introduces `np.median_sorted_arrays(a, b)`, which computes the median of two pre-sorted arrays using the binary-search partition algorithm.

The proposed implementation:

* Computes the median without concatenating, merging, or sorting the input arrays.
* Uses binary search on the smaller array.
* Achieves **O(log(min(m, n)))** time complexity.
* Uses **O(1)** additional space.
* Correctly handles both odd and even combined lengths.
* Supports integer and floating-point arrays.
* Handles duplicate values, negative values, and the case where one input array is empty.

The motivation for this proposal is to provide an efficient API for users who already have sorted data and wish to compute the median without the additional cost of merging or sorting.

I welcome feedback on whether this functionality would be an appropriate addition to NumPy's public API and on the proposed interface.

### First-time committer introduction

Hello! This is my first contribution to NumPy.

I use NumPy extensively for numerical computing, data processing, and algorithmic programming. While working on algorithms involving sorted arrays, I noticed there is currently no dedicated API for computing the median of two already sorted arrays using the optimal binary-search approach. I thought this could be a useful addition and wanted to contribute a proposal for discussion and review.

Thank you for taking the time to review this pull request.

### DIFF
--- a/benchmarks/benchmarks/bench_function_base.py
+++ b/benchmarks/benchmarks/bench_function_base.py
@@ -105,6 +105,24 @@ class Median(Benchmark):
         np.median(self.wide, axis=0)
 
 
+class MedianSortedArrays(Benchmark):
+    params = [
+        [(1000, 1000), (100000, 100000), (10, 100000), (1, 1000000)],
+    ]
+    param_names = ['sizes']
+
+    def setup(self, sizes):
+        m, n = sizes
+        self.a = np.sort(np.random.random(m))
+        self.b = np.sort(np.random.random(n))
+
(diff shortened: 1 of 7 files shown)


jorenham:

PR #31782, opened 2026-06-27
BUG: fix `np.insert` ignoring out of bounds negative indices in multi element obj

### LINKED ISSUE #31777: BUG: `np.insert` accepts negative out-of-bounds indices with multi-element obj
### Describe the issue:

`np.insert` does not consistently reject negative out-of-bounds indices when `obj` contains multiple indices. For an array of length 5, `-6` should raise `IndexError`, as it does for scalar and single-element `obj`, but `np.insert(a, [-6, 0], [9, 8])` returns a result instead.

```python
import numpy as np

a = np.arange(5)
print(f"a:              {a}")
print(f"insert [-6, 0]: {np.insert(a,[-6,0],[9,8])}") # should IndexError
```

### Error message:

```shell
Output:

a:              [0 1 2 3 4]
insert [-6, 0]: [0 8 1 2 3 4 9]
```

```
2.6.0.dev0+git20260626.80c6c35
3.14.4 (main, Apr 14 2026, 14:46:33) [Clang 22.1.3 ]
```

### Runtime Environment:

```
[{'numpy_version': '2.6.0.dev0+git20260626.80c6c35',
  'python': '3.14.4 (main, Apr 14 2026, 14:46:33) [Clang 22.1.3 ]',
  'uname': uname_result(system='Darwin', node='Mac-2.local', release='25.2.0', version='Darwin Kernel Version 25.2.0: Tue Nov 18 21:09:56 PST 2025; root:xnu-12377.61.12~1/RELEASE_ARM64_T6041', machine='arm64')},
                      'not_found': ['ASIMDFHM']}},
```

Found while reviewing `np.insert` overload implementation for Numba project and differential-testing it against NumPy.

Aniketsy:
Fixes #31777 
### PR summary
- Added a bounds check for multi element obj before negative index to catch out-of-bounds indices and raise IndexError.

#### AI Disclosure
- yes to verify fix

### DIFF
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -5566,8 +5566,9 @@ def insert(arr, obj, values, axis=None):
         indices = indices.astype(intp)
 
     if indices.size > 0 and (indices.min() < -N or indices.max() > N):
+        oob = indices.min() if indices.min() < -N else indices.max()
         raise IndexError(
-            f"index {indices[oob][0]} is out of bounds "
+            f"index {oob} is out of bounds "
             f"for axis {axis} with size {N}"
         )
     indices[indices < 0] += N
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -5565,8 +5565,10 @@ def insert(arr, obj, values, axis=None):
         # Can safely cast the empty list to intp
         indices = indices.astype(intp)
 
-    if indices.size > 0 and (indices.min() < -N or indices.max() > N):
-        oob = indices.min() if indices.min() < -N else indices.max()
+    min_idx = indices.min()
+    max_idx = indices.max()
+    if indices.size > 0 and (min_idx < -N or max_idx > N):
+        oob = min_idx if min_idx < -N else max_idx
         raise IndexError(
             f"index {oob} is out of bounds "
             f"for axis {axis} with size {N}"
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -5565,6 +5565,12 @@ def insert(arr, obj, values, axis=None):
         # Can safely cast the empty list to intp
         indices = indices.astype(intp)
 
+    oob = (indices < -N) | (indices > N)
+    if oob.any():
+        raise IndexError(
+            f"index {indices[oob][0]} is out of bounds "
+            f"for axis {axis} with size {N}"
+        )
     indices[indices < 0] += N
 
     numnew = len(indices)
--- a/numpy/lib/_function_base_impl.py
+++ b/numpy/lib/_function_base_impl.py
@@ -5565,14 +5565,15 @@ def insert(arr, obj, values, axis=None):
         # Can safely cast the empty list to intp
         indices = indices.astype(intp)
 
-    min_idx = indices.min()
-    max_idx = indices.max()
-    if indices.size > 0 and (min_idx < -N or max_idx > N):
-        oob = min_idx if min_idx < -N else max_idx
-        raise IndexError(
-            f"index {oob} is out of bounds "
-            f"for axis {axis} with size {N}"
-        )
+    if indices.size > 0:
+        min_idx = indices.min()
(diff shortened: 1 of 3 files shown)


numpy/lib/tests/test_function_base.py
swap357: can the test be folded into existing `test_index_out_of_bounds` ?

Aniketsy: thanks , yes it make sense

numpy/lib/_function_base_impl.py
```diff
@@ -5565,6 +5565,12 @@ def insert(arr, obj, values, axis=None):
         # Can safely cast the empty list to intp
         indices = indices.astype(intp)
 
+    oob = (indices < -N) | (indices > N)
+    if oob.any():
```
ngoldbaum:
```suggestion
    if indices.size > 0 and (indices.min() < -N or indices.max() > N)
```

This is in the fast path and `indices` might be a large array so it's worth being intentional about this change. Using reductions avoids allocating any temporary arrays. Integer array reductions are also optimized with SIMD.

I believe that the suggestion is equivalent to what you have above, please double-check me and make sure your test catches the `size==0` corner case.

Also since people may see new errors in previously "working" code, IMO this deserves a release note briefly explaining the change.

ngoldbaum: (btw @ikrommyd - look: another customer for a minmax reduction!)

Aniketsy: @ngoldbaum thanks for the improvement, i've applied suggestion and our test covers the corner case.

i'll add release note for this

numpy/lib/_function_base_impl.py
```diff
@@ -5565,6 +5565,11 @@ def insert(arr, obj, values, axis=None):
         # Can safely cast the empty list to intp
         indices = indices.astype(intp)
 
+    if indices.size > 0 and (indices.min() < -N or indices.max() > N)
```
Aniketsy:
```suggestion
    if indices.size > 0 and (indices.min() < -N or indices.max() > N):
```

numpy/lib/_function_base_impl.py
Aniketsy: ahh we missed `:` in our if condition 😀

numpy/lib/_function_base_impl.py
```diff
@@ -5565,6 +5565,12 @@ def insert(arr, obj, values, axis=None):
         # Can safely cast the empty list to intp
         indices = indices.astype(intp)
 
+    if indices.size > 0 and (indices.min() < -N or indices.max() > N):
+        oob = indices.min() if indices.min() < -N else indices.max()
```
MaanasArora: More lines, but I think we can skip recomputing these: maybe declare `min_idx, max_idx` above and then break and nest the if condition for the error?

Aniketsy: yes, i had this thought and this is valid improvement, but i thought, as this will be only computed when an error is being raised, which is rare, so i think we're fine.  what do you think ?

But still we can go with improvement 😀 .

Thanks for the review!

MaanasArora: I think I'd still go with it since I think it is also more readable! The if-statement is also quite long as is.

MaanasArora: Sorry, in case it wasn't clear, I meant I'd go with the suggestion :)

Aniketsy: Yes , i got it ... I'll update this shortly 😊. Thanks

ngoldbaum (approved):
Looks good! I’ll let Maanas merge if he agrees.

numpy/lib/_function_base_impl.py
ikrommyd: Just noticed this tag in my github notifications, and yeah while adding minmax all the parts of the codebase that calculate min and max separately should be refactored.

MaanasArora (approved):
Looks good, thanks @Aniketsy! Merging...

PR #31786, opened 2026-06-27
DOC: remove stale out reference from vstack, hstack  docstrings

shipitdev:
Small docstring fix for `vstack` and `hstack`.

Both list this under `dtype`:

    Cannot be provided together with `out`.

but neither function takes an `out` argument — confirmed below. The whole `dtype`/`casting` block (even the `.. versionadded:: 1.24` tag) is identical in both, which points to it being copied over from `stack`/`concatenate`. Those two do have `out`, so the line makes sense there; it just never got dropped when reused for `vstack`/`hstack`.

```python
>>> import numpy as np, inspect
>>> inspect.signature(np.vstack)
<Signature (tup, *, dtype=None, casting='same_kind')>
>>> inspect.signature(np.hstack)
<Signature (tup, *, dtype=None, casting='same_kind')>
>>> np.vstack(([1, 2], [3, 4]), out=None)
    ...
TypeError: vstack() got an unexpected keyword argument 'out'
```

Left `stack`/`concatenate` untouched since the sentence is accurate for them.

### AI Disclosure

Used an AI assistant to scan for stale/undocumented parameter mismatches across docstrings and to confirm this one via `inspect.signature` and the actual `TypeError` above. Reviewed and wrote up the fix manually.

### DIFF
--- a/numpy/_core/shape_base.py
+++ b/numpy/_core/shape_base.py
@@ -238,8 +238,7 @@ def vstack(tup, *, dtype=None, casting="same_kind"):
         each element along the zeroth axis is treated as a separate array.
 
     dtype : str or dtype
-        If provided, the destination array will have this dtype. Cannot be
-        provided together with `out`.
+        If provided, the destination array will have this dtype.
 
         .. versionadded:: 1.24
 
@@ -313,8 +312,7 @@ def hstack(tup, *, dtype=None, casting="same_kind"):
         each element along the zeroth axis is treated as a separate array.
 
     dtype : str or dtype
-        If provided, the destination array will have this dtype. Cannot be
-        provided together with `out`.
+        If provided, the destination array will have this dtype.
 
         .. versionadded:: 1.24
 


shipitdev:
gentle ping, The clang check timed out, but since this is just a doc fix I'm assuming it's a CI flake. Let me know if I need to do anything to push this forward!

eeshsaxena (commented):
Confirmed - neither `vstack` nor `hstack` takes an `out` argument, so removing that sentence is correct. Checking the signatures on 2.4.4:

```python
>>> import numpy as np, inspect
>>> [( "out" in inspect.signature(getattr(np, f)).parameters) for f in ("vstack","hstack","stack","concatenate")]
[False, False, True, True]
```

`vstack`/`hstack` are `(tup, *, dtype=None, casting='same_kind')` (no `out`), while `stack`/`concatenate` do have `out` - so as you note, the whole `dtype`/`casting` block including that line was copied from those two. Keeping just the first sentence of the `dtype` entry is the right call. Thanks for the fix.

charris:
Thanks @shipitdev .  For doc only fixes, append a line with `[skip azp] [skip actions]` at the bottom of the commit message.

PR #31791, opened 2026-06-29
ENH: Fast-path zero-width axes in np.pad

KRRT7:
### PR summary

Fast-path zero-width axes in `np.pad` for `mode="linear_ramp"` and the statistic modes `"maximum"`, `"mean"`, `"median"`, and `"minimum"`.

These modes currently do mode-specific work even for axes where the requested pad width is `(0, 0)`. This change skips that per-axis work inside those mode branches while preserving validation behavior for invalid mode-specific arguments.

The shortcut applies to both fully zero-width padding and mixed padding such as `((0, 0), (0, 0), (1, 1))`, where only the zero-width axes skip the expensive mode-specific work. Other `np.pad` modes stay on their existing paths.

Validation:
- `uv run pytest numpy/lib/tests/test_arraypad.py -q`
- `uv run ruff check numpy/lib/_arraypad_impl.py numpy/lib/tests/test_arraypad.py benchmarks/benchmarks/bench_lib.py`

Targeted native NumPy benchmark:

```bash
spin bench --compare upstream/main HEAD -t bench_lib.Pad
```

reported `PERFORMANCE INCREASED`.

Representative rows:

- `bench_lib.Pad.time_pad((4, 4, 4, 4), 0, "linear_ramp")`: `50.2+/-1us -> 5.20+/-0.07us`, ratio `0.10`
- `bench_lib.Pad.time_pad((1, 1, 1, 1, 1), 0, "linear_ramp")`: `73.1+/-20us -> 5.28+/-0.04us`, ratio `0.07`
- `bench_lib.Pad.time_pad((256, 128, 1), 0, "linear_ramp")`: `52.3+/-1us -> 9.12+/-0.04us`, ratio `0.17`
- `bench_lib.Pad.time_pad((4, 4, 4, 4), 0, "mean")`: `23.0+/-3us -> 4.80+/-0.02us`, ratio `0.21`
- `bench_lib.Pad.time_pad((1024, 1024), 0, "mean")`: `488+/-20us -> 158+/-10us`, ratio `0.32`

#### AI Disclosure

LLM / Harness: hybrid.


numpy/lib/_arraypad_impl.py
```diff
@@ -837,6 +837,22 @@ def pad(array, pad_width, mode='constant', **kwargs):
     stat_functions = {"maximum": np.amax, "minimum": np.amin,
                       "mean": np.mean, "median": np.median}
 
+    if mode == "linear_ramp" and not np.any(pad_width):
+        _as_pairs(kwargs.get("end_values", 0), array.ndim)
+        order = 'F' if array.flags.fnc else 'C'
+        return array.copy(order=order)
+
+    if mode in stat_functions and not np.any(pad_width):
```
ngoldbaum: You're calling `np.any(pad_width)` twice, which is unnecessary and may be significant for big enough `pad_width`. Is there perhaps a way to avoid the overhead of that in the fast path?

ngoldbaum:
> Validation:
>
> uv run pytest numpy/lib/tests/test_arraypad.py -q
> uv run ruff check numpy/lib/_arraypad_impl.py numpy/lib/tests/test_arraypad.py
> Local timings on a noisy machine show the targeted modes improving clearly. For example, on a small 3D array, linear_ramp went from about 41 us to 5.3 us, and statistic modes went from about 15-33 us to about 5.3-5.7 us.

For performance improvements we prefer to see benchmarks. It looks like there is a `Pad` benchmark already. Can you maybe add a case or two to that?

See here for more: https://numpy.org/devdocs/benchmarking.html.

It'd also be nice if you could point to a real-world case where this optimization is important.

KRRT7:
sorry, disregard that last push, I'm benchmarking it separately.

KRRT7:
For motivation: a lot of libraries build higher-level padding APIs on top of `np.pad` and pass NumPy’s modes through. xarray and xrft are good examples: they accept padding widths built from named dimensions or grid metadata, then delegate to NumPy.

In that kind of code, it’s pretty natural for a computed padding width to sometimes be zero.

Examples:
- https://github.com/pydata/xarray/blob/71cfb13ac7e05a8d37f63291436fa90820af9304/xarray/core/variable.py
- https://github.com/xgcm/xrft/blob/c74533fce02b7279ec5d61c8a97a26902728bbaf/xrft/padding.py

KRRT7:
CI is flaky,

numpy/lib/_arraypad_impl.py
```diff
@@ -837,6 +837,23 @@ def pad(array, pad_width, mode='constant', **kwargs):
     stat_functions = {"maximum": np.amax, "minimum": np.amin,
                       "mean": np.mean, "median": np.median}
 
+    zero_pad_width = not np.any(pad_width)
```
ngoldbaum: This unconditional `np.any` call in the fast path is making all other users of this function pay for the optimization you're adding, which isn't great.

When I asked an AI model to review this PR, it measured ~25% overhead for small arrays.

I bet a `spin bench` run would catch this. If you're doing optimization work, please run benchmarks locally and think about any regressions you see - they may be real!

Also, doesn't this mean people only get the optimization if this is a *truly* no-op pad operation? What about e.g. a pad like `((0,0),(0,0),(1,1))`?

Wouldn't it be better to combine the two `if` statements into something like this:

```
if mode == "linear_ramp" or mode in stat_functions:
    if not any(left or right for left, right in pad_width):
```

I'm pretty sure that also avoids the need to use `np.any` at all? It also means that you optimize cases that actually do need to do some work, which is probably more realistic for real operations that might show up in xarray.

ngoldbaum (commented):
I asked an AI model to review this PR and it identified some behavior changes that should be reconsidered or explicitly tested and called out in e.g. a release note:


> - **Stricter for empty arrays.** On main, the `array.size == 0` branch runs before the mode branches, so kwargs are never validated for empty inputs. The fast path validates first, so e.g. `np.pad(np.empty((0, 3)), 0, mode='mean', stat_length=-1)` or a wrongly-shaped `stat_length`/`end_values` now raises `ValueError` where main silently returned a copy. Arguably a fix, but it's an undocumented behavior change and should be intentional (and ideally tested).
> - **Laxer about values.** Errors that previously surfaced from actually doing the (useless) work now vanish on zero-width pads: object-dtype `mean`/`median` that raised `TypeError` now succeeds; `end_values='abc'` raised `DTypePromotionError`, now silently accepted; the `ComplexWarning` for complex `end_values` on a float array is gone. Defensible ("don't validate values for a no-op"), but worth a conscious decision since the same PR makes shape validation *stricter*.

numpy/lib/_arraypad_impl.py
KRRT7: Addressed in `3d39ce8fce`.

I removed the unconditional `np.any(pad_width)` check before mode dispatch. The shortcut now lives inside the `linear_ramp` and statistic branches and skips only axes whose `width_pair` is zero-width, so unrelated modes no longer pay for the check. This also handles mixed pad widths like `((0, 0), (0, 0), (1, 1))` by skipping the zero-width axes while still doing the needed work on padded axes.

I also moved this after the existing empty-array branch, and added tests for the empty-array behavior and `maximum`/`minimum` `stat_length=0` handling.

Validation:
- `uv run pytest numpy/lib/tests/test_arraypad.py -q`
- `uv run ruff check numpy/lib/_arraypad_impl.py numpy/lib/tests/test_arraypad.py benchmarks/benchmarks/bench_lib.py`
- `spin bench --compare upstream/main HEAD -t bench_lib.Pad` reported `PERFORMANCE INCREASED`.

KRRT7:
I updated the implementation to avoid the unconditional zero-width check before mode dispatch. The shortcut now only runs inside the expensive `linear_ramp` and statistic-mode branches, and skips work per-axis when that axis has `pad_width == (0, 0)`. That keeps unrelated modes on their existing path and also covers mixed pad widths like `((0, 0), (0, 0), (1, 1))`.

I reran the targeted native benchmark:

```bash
spin bench --compare upstream/main HEAD -t bench_lib.Pad
```

ASV reported `PERFORMANCE INCREASED`.

| Benchmark | Before | After | Ratio |
|---|---:|---:|---:|
| `Pad.time_pad((4, 4, 4, 4), 0, "linear_ramp")` | `50.2±1μs` | `5.20±0.07μs` | `0.10` |
| `Pad.time_pad((1, 1, 1, 1, 1), 0, "linear_ramp")` | `73.1±20μs` | `5.28±0.04μs` | `0.07` |
| `Pad.time_pad((256, 128, 1), 0, "linear_ramp")` | `52.3±1μs` | `9.12±0.04μs` | `0.17` |
| `Pad.time_pad((4, 4, 4, 4), 0, "mean")` | `23.0±3μs` | `4.80±0.02μs` | `0.21` |
| `Pad.time_pad((256, 128, 1), 0, "mean")` | `47.9±0.7μs` | `8.89±0.05μs` | `0.19` |
| `Pad.time_pad((1024, 1024), 0, "mean")` | `488±20μs` | `158±10μs` | `0.32` |
| `Pad.time_pad((4194304,), 0, "mean")` | `1.48±0.09ms` | `934±10μs` | `0.63` |

KRRT7:
I pushed a follow-up commit (`14a657c52f`) to preserve the old validation behavior for zero-width padding.

The fast path now still skips the expensive no-op padding work, but validates skipped zero-width axes first:

- `linear_ramp`: preserves invalid `end_values` errors and complex-to-real cast warnings.
- statistic modes: preserves `stat_length=0` behavior for max/min and dtype/object reduction errors.
- empty-array behavior remains on the existing `array.size == 0` path before the shortcut.

I also added regression tests for these cases.

Validation run:

- `uv run pytest numpy/lib/tests/test_arraypad.py -q`
  - `644 passed, 1 xpassed`
- `uv run ruff check numpy/lib/_arraypad_impl.py numpy/lib/tests/test_arraypad.py benchmarks/benchmarks/bench_lib.py`
  - passed

Targeted benchmark rerun:

`uv run spin bench --compare upstream/main HEAD -t bench_lib.Pad`

| case | main | PR | ratio |
|---|---:|---:|---:|
| `(256, 128, 1), 0, linear_ramp` | `53.0±0.9μs` | `31.8±0.9μs` | `0.60` |
| `(4, 4, 4, 4), 0, linear_ramp` | `50.8±1μs` | `38.8±0.8μs` | `0.76` |
| `(4194304,), 0, mean` | `1.61±0.06ms` | `911±40μs` | `0.57` |
| `(256, 128, 1), 0, mean` | `46.4±0.6μs` | `19.1±0.2μs` | `0.41` |

Preserving validation does reduce the original speedup somewhat, but the targeted zero-width cases still improve while keeping the previous error/warning behavior.

numpy/lib/_arraypad_impl.py
```diff
@@ -877,6 +909,9 @@ def pad(array, pad_width, mode='constant', **kwargs):
         end_values = kwargs.get("end_values", 0)
         end_values = _as_pairs(end_values, padded.ndim)
         for axis, width_pair, value_pair in zip(axes, pad_width, end_values):
+            if not any(width_pair):
+                _validate_zero_width_linear_ramp(padded, axis, value_pair)
+                continue
```
ngoldbaum, line 915: This validation runs on `padded`, but the non-shortcut path below operates on `roi`.
The difference matters: `_view_roi` restricts axes *after* the current one to the
original data area, while `padded` still contains the not-yet-written pad slots for
those axes. For object dtype, `_pad_simple`'s `np.empty` fills those slots with
`None`, so validation trips over them:

```python
>>> a = np.array([[1, 2], [3, 4]], dtype=object)
>>> np.pad(a, ((0, 0), (1, 1)), mode="linear_ramp")  # works on main
TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'
```

The trigger is a zero-width axis ordered before an axis with nonzero width;
`((1, 1), (0, 0))` is unaffected because earlier axes are already filled, which is
why the existing tests (all-zero widths) pass.

Passing `roi` instead fixes this — the `hasobject` branch of the helper then makes
exactly the call main makes today, so behavior parity holds by construction:

```suggestion
            roi = _view_roi(padded, original_area_slice, axis)
            if not any(width_pair):
                _validate_zero_width_linear_ramp(roi, axis, value_pair)
                continue
```

The stat branch below has the same problem.

numpy/lib/_arraypad_impl.py
```diff
@@ -294,6 +314,18 @@ def _get_stats(padded, axis, width_pair, length_pair, stat_func):
     return left_stat, right_stat
 
 
+def _validate_zero_width_stat(padded, axis, length_pair, stat_func):
```
ngoldbaum: A second, subtler problem with receiving the full `padded` array: for non-object
dtypes this reduces `padded[0, ..., 0]`, which is **uninitialized `np.empty`
memory** whenever a later axis has a nonzero left pad width (e.g.
`pad_width=((0, 0), (1, 1))`). The value is discarded, so results stay correct,
but the reduction still consumes uninitialized memory — I could produce a spurious
`RuntimeWarning: invalid value encountered in reduce` by poisoning the pad region
with signaling-NaN bit patterns, and sanitizers/valgrind would flag the read.

Switching the call sites to pass `roi` (see comments above) fixes this too:
`roi[0, ..., 0]` is always either original data or an already-written pad value.
Once that's done, please also rename the `padded` parameter to `roi` here and in
`_validate_zero_width_linear_ramp`, so it stays obvious that these helpers must
not be handed the raw padded array.

numpy/lib/tests/test_arraypad.py
```diff
@@ -118,6 +118,58 @@ def test_zero_padding_shortcuts(self, mode):
...
+        test = np.array(["a", "b"])
+        with pytest.raises(TypeError):
+            np.pad(test, 0, mode=mode)
+
+    @pytest.mark.parametrize("mode", ["maximum", "mean", "median", "minimum"])
+    def test_zero_padding_stat_modes_validate_object_dtype(self, mode):
+        test = np.array([object(), object()], dtype=object)
+        with pytest.raises(TypeError):
+            np.pad(test, 0, mode=mode)
+
```
ngoldbaum, line 172: The object-dtype tests here only cover all-zero pad widths, which is exactly the
case that *doesn't* exercise the bug above. Please add a mixed-width object test —
this currently raises `TypeError` on this branch for all five modes and passes
once validation uses `roi`:

```python
    @pytest.mark.parametrize(
        "mode", ["linear_ramp", "maximum", "mean", "median", "minimum"]
    )
    def test_zero_width_axis_object_mixed_widths(self, mode):
        # Zero-width axes must not feed not-yet-filled pad values (None
        # for object dtype) into the validation of later padded axes.
        test = np.array([[1, 2], [3, 4]], dtype=object)
        expected = np.pad(test.astype(float), ((0, 0), (1, 1)), mode=mode)
        result = np.pad(test, ((0, 0), (1, 1)), mode=mode)
        assert result.dtype == object
        assert_array_equal(result.astype(float), expected)
```

ngoldbaum (commented):
AI-assisted review, FYI, but this isn't yet mergeable

ngoldbaum (approved):

PR #31792, opened 2026-06-29
MAINT: pixi-packages: remove redundant `uv` dep

lucascolley:
this is now included by default when one does not specify `pip` as a dependency


charris:
Thanks @lucascolley .

lucascolley:
thanks Chuck

PR #31793, opened 2026-06-29
TYP: specialized unary predicate ufuncs

jorenham:
After #31761, this adds specialized type annotations to the unary predicate ufunc signatures (i.e. those that return `bool`).

With this, the only unary ufunc left is `bitwise_count`, which I'll tackle after this one.

---

No AI


charris:
Thanks Joren.

PR #31797, opened 2026-06-29
TYP: specialized ``bitwise_count`` ufunc

jorenham:
Following #31793, this adds specialized typing support for the `bitwise_count` ufunc. With this, all unary (1-in, 1-out) ufuncs have specialized type annotations. We now no longer need the type-check-only / "phantom" `_UFunc_Nin1_Nout1` subtype of `np.ufunc`, which this also removes.

---

No AI.


jorenham:
the linting errors are unrelated

update: I opened https://github.com/numpy/numpy/pull/31798 for this

jorenham:
@charris when merging this, could you merge with a merge commit instead of squashing? I'm working on a follow-up branch, which is based on this one, so squashing these commits will cause merge conflicts.

charris:
Thanks Joren.

PR #31748, opened 2026-06-25
MAINT, DOC: Update sphinx to 8.2.3.

charris:
This pins sphinx to version 8.2.3. Newer upstream dependencies need
a later sphinx version and we might as well deal with that.

~No AI.~ AI used to analyse back compatibility issues with sphinx 8.2.3.
and also provided some code snippets.

The big problem here was `See Also` as used in some documentation.  The
``SeeAlso`` class was removed from ``sphinx.directives.other``, the
``seealso`` directive is now registered internally, but the raw docutils
parser used in refguide_check became stricter about content formatting.
The fix was adding a lenient handler for ``seealso`` so the checker would
accept existing numpydoc-style "See Also" sections without forcing
indentation changes, which would hurt console readability.

The full Sphinx documentation build continues to work without that
modification.

[skip azp] [skip actions]

PR #31798, opened 2026-06-29
MAINT: fix ruff errors in ``tools/refguide_check.py``

jorenham:
The lint CI job apparently didn't run in #31748, so some ruff errors were missed that are now causing the lint CI job to fail on main (see e.g. https://github.com/numpy/numpy/actions/runs/28394673323/job/84130268695?pr=31797).

Nothing a `ruff check --fix` couldn't fix though.

---

No AI.


jorenham:
closing in favor of #31799

PR #31813, opened 2026-06-30
TYP: specialized binary comparison ufuncs

jorenham:
This specialized the binary comparison ufuncs: `[not_]equal`, `greater[_equal]`, and `less[_equal]`.

This is one of the easiest classes of the binary ufuncs (no dtype promotion to worry about), so I thought these would make for a good start. 
But as you can see "easiest" doensn't mean "easy"; it's still painfully complex. That's why I didn't add the (many) shape-typing overloads, leaving it as future work.

I loosely based this on the `scipy.special` ufunc stubs from scipy-stubs ([src](https://github.com/scipy/scipy-stubs/blob/991da7e5943cdeec3884760c90572559a51d1e29/scipy-stubs/special/_ufuncs.pyi#L470)). 

---

No AI.


charris:
Thanks Joren.

PR #31816, opened 2026-06-30
ENH: add mechanism to register reduction loops to ufuncs for multi-output reductions

ikrommyd:
### PR summary
This is adding the ability to add reduction loops to ufuncs as it was being recently discussed in the numpy community meetings. The current reductions are implemented using the binary ufunc loops which are 2 -> 1 (2 inputs 1 output) and by setting the output pointer equal to the first input pointer which makes it the accumulation location and not moving it. We generalize this logic for ufuncs that return N outputs. However the normal ufunc loop cannot be used in a similar way in this case. The trick is not possible. So we add the ability to add reduction loops to ufuncs that are N + 1 -> N loops while their nominal/forward loops are 2 -> N.

This is done via a single new function pointer on the array method called `get_reduction_loop` similar to `get_strided_loop`. The whole reduction machinery is then generalized to support loops that are N + 1 -> N instead of hardcoding the assumption that the loop is 2 -> 1. In the nominal binary ufunc case, everything collapses down to the old 2 -> 1 assumption. This generalization was the "easy" part. I kept it to `reduce` right now to minimize the diff for comments but can done for `reduceat` and `accumulate` in a similar manner. The "difficult" algorithmically part for me was the `reducelike_promote_and_resolve_multi` function mainly because since the reduction loops live on the array method via the `get_reduction_loop` function pointer, we have to resolve the dtypes for the "forward pass" loop, get the array method from ufunc object's `_loops` and then get the reduction loop from it. This potentially needs some discussion about how it's best to do so. I generally like the fact that I have to touch only a few files to implement this whole machinery here.

An example of how this machinery can be used by numpy or an extension is here: https://gist.github.com/ikrommyd/51dddb52e54c6f99aa86637b16a11d22
The PR/design should be reviewed in alignment with that.
Here I have a `minmax.cpp` extension with a `setup.py` that implements a `minmummaximum` ufunc and registers  reduction loops for it which would be `minmax`.  I also include a test file `test_minmax.py` in the gist.

This is obviously not ready, I'm just opening it as a draft for comments about the core ideas. We definitely need to consolidate the design and add better checks and error messages in a few places. Currently on my macos and linux machines both the numpy testing suite and `test_minmax.py` pass so I'm curious to see what the full CI does.

I plan to make a PR on my fork soon (next couple of days) that targets the PR branch that implements minmax as part of numpy using the reduction loop mechanism instead of having it in a separate extension.

This is towards https://github.com/numpy/numpy/issues/9836

N/A

#### AI Disclosure
Claude Opus 4.8 via the Claude Code harness has been used for A) querying around the numpy codebase as a fancier grep B) reviewing for logic or reference counting bugs and C) to write comments that I was too lazy to write myself. The whole design and ideas are my own and most of the C code is my own. The AI has mainly fixed my bugs and wrote code to register the reduction loops on the ufunc in my minmax.cpp extension in the gist.

cc @ngoldbaum @seberg please tag anyone else who might be interested in taking a look at the design.

P.S. ignore the changes in `numpy/_core/src/umath/loops_minmax.dispatch.c.src`, they are just for benchmarking purposes because my minmax.cpp extension has no SIMD right now.

### DIFF
--- a/doc/release/upcoming_changes/31816.c_api.rst
+++ b/doc/release/upcoming_changes/31816.c_api.rst
@@ -1,21 +1,6 @@
 New mechanism to register reduction loops to ufuncs for multi-output reductions
 -------------------------------------------------------------------------------
-``ArrayMethod`` implementations can now register a dedicated reduction loop via
-the ``NPY_METH_get_reduction_loop`` slot. This allows ufuncs with more than one
-output (a two-in/``N``-out elementwise loop) to support
-:meth:`~numpy.ufunc.reduce`, which previously only worked for single-output
-binary ufuncs. ``reduce`` still requires exactly two inputs. Only the output
-arity is generalized. The reduction loop has an
-``(N+1)``-in/``N``-out signature: it consumes the running accumulators
-together with one streamed input element and produces the updated
-accumulators. Such a loop can also register per-output reduction identities
-through the ``NPY_METH_get_multi_reduction_initials`` slot (the multi-output
-analogue of ``NPY_METH_get_reduction_initial``), used to seed empty reductions
-and reductions with a ``where=`` mask.
-:meth:`~numpy.ufunc.reduce` on such a ufunc returns a tuple with one array per
-output, its ``out`` argument accepts a tuple of output arrays (one per
-output), and its ``initial`` argument accepts either a single value
-(broadcast to every output) or a tuple with one value per output.
-
-See :ref:`c-api.reduction-loop-tutorial` for a worked example of registering a
-reduction loop on a custom ufunc.
+Ufuncs with more than one output can now support :meth:`~numpy.ufunc.reduce`
+by registering a dedicated reduction loop (and, optionally, per-output
+identities) on their ``ArrayMethod``. The reduction returns one array per
+output. See :ref:`c-api.reduction-loop-tutorial` for a worked example.
--- a/doc/release/upcoming_changes/31816.c_api.rst
+++ b/doc/release/upcoming_changes/31816.c_api.rst
@@ -1,0 +1,18 @@
+New mechanism to register reduction loops to ufuncs for multi-output reductions
+-------------------------------------------------------------------------------
+``ArrayMethod`` implementations can now register a dedicated reduction loop via
+the ``NPY_METH_get_reduction_loop`` slot. This allows ufuncs with more than one
+output (a two-in/``N``-out elementwise loop) to support
+:meth:`~numpy.ufunc.reduce`, something that was previously restricted to
+single-output binary ufuncs; ``reduce`` still requires exactly two inputs,
+only the output arity is generalized. The reduction loop has an
+``(N+1)``-in/``N``-out signature: it consumes the running accumulators
+together with one streamed input element and produces the updated
+accumulators.
+:meth:`~numpy.ufunc.reduce` on such a ufunc returns a tuple with one array per
+output, its ``out`` argument accepts a tuple of output arrays (one per
+output), and its ``initial`` argument accepts either a single value
+(broadcast to every output) or a tuple with one value per output.
+
+See :ref:`c-api.reduction-loop-tutorial` for a worked example of registering a
+reduction loop on a custom ufunc.
--- a/doc/release/upcoming_changes/31816.c_api.rst
+++ b/doc/release/upcoming_changes/31816.c_api.rst
@@ -3,12 +3,15 @@ New mechanism to register reduction loops to ufuncs for multi-output reductions
 ``ArrayMethod`` implementations can now register a dedicated reduction loop via
 the ``NPY_METH_get_reduction_loop`` slot. This allows ufuncs with more than one
 output (a two-in/``N``-out elementwise loop) to support
-:meth:`~numpy.ufunc.reduce`, something that was previously restricted to
-single-output binary ufuncs; ``reduce`` still requires exactly two inputs,
-only the output arity is generalized. The reduction loop has an
+:meth:`~numpy.ufunc.reduce`, which previously only worked for single-output
+binary ufuncs. ``reduce`` still requires exactly two inputs. Only the output
+arity is generalized. The reduction loop has an
 ``(N+1)``-in/``N``-out signature: it consumes the running accumulators
 together with one streamed input element and produces the updated
-accumulators.
+accumulators. Such a loop can also register per-output reduction identities
+through the ``NPY_METH_get_multi_reduction_initials`` slot (the multi-output
+analogue of ``NPY_METH_get_reduction_initial``), used to seed empty reductions
+and reductions with a ``where=`` mask.
 :meth:`~numpy.ufunc.reduce` on such a ufunc returns a tuple with one array per
 output, its ``out`` argument accepts a tuple of output arrays (one per
 output), and its ``initial`` argument accepts either a single value
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -1851,6 +1851,44 @@ the functions that must be implemented for each slot.
    initial value is correct, since NumPy may call this even when it is not
    strictly necessary to do so.
 
+.. c:macro:: NPY_METH_get_reduction_loop
+
+   .. versionadded:: 2.6
+
+   Registers a dedicated loop for use by :meth:`~numpy.ufunc.reduce`,
+   implemented as a :c:type:`PyArrayMethod_GetLoop` function (the same
+   typedef used for ``NPY_METH_get_loop``). This is required to reduce
+   ufuncs with more than one output, since the "forward" elementwise loop of
+   such a ufunc cannot be used as a reduction loop the way a single-output loop
+   can (by aliasing the output to the first input). Instead, the returned
+   :c:type:`PyArrayMethod_StridedLoop` must implement the reduction
+   directly, with an ``(nout + 1)``-in/``nout``-out signature: it takes the
+   current per-output accumulators followed by one streamed input element,
+   and writes the updated accumulators. That is, for a ufunc whose forward
+   loop has ``nout`` outputs, the *data*, *strides*, and descriptor arrays
+   passed to the reduction loop are laid out as::
+
+       [acc_0, ..., acc_{nout-1}, x, out_0, ..., out_{nout-1}]
+
+   where ``x`` is the streamed element being reduced in and each ``out_i``
+   is aliased to (and typically has stride 0 relative to) the matching
+   ``acc_i``.
+
+   If ``NPY_METH_get_reduction_loop`` is not set, :meth:`~numpy.ufunc.reduce`
+   falls back to ``NPY_METH_get_loop``/``NPY_METH_strided_loop``, which only
+   works for the traditional two-input/one-output case. Calling
+   :meth:`~numpy.ufunc.reduce` on a ufunc with more than one output whose
+   resolved ArrayMethod does not register a reduction loop raises a
+   :exc:`ValueError`. See :ref:`c-api.reduction-loop-tutorial` for a
+   worked example.
+
+   This slot only relaxes the *output* arity: :meth:`~numpy.ufunc.reduce`
+   (as well as :meth:`~numpy.ufunc.accumulate` and
+   :meth:`~numpy.ufunc.reduceat`) still requires the ufunc itself to have
+   exactly two inputs, independent of this slot. Calling any of them on a
+   ufunc with ``nin != 2`` raises a :exc:`ValueError` regardless of
+   ``nout`` or whether a reduction loop is registered.
+
 Flags
 ~~~~~
 
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -1889,6 +1889,33 @@ the functions that must be implemented for each slot.
    ufunc with ``nin != 2`` raises a :exc:`ValueError` regardless of
    ``nout`` or whether a reduction loop is registered.
 
+.. c:macro:: NPY_METH_get_multi_reduction_initials
+
+   .. versionadded:: 2.6
+
+   Registers the per-output reduction identity/initial values, implemented as
+   a :c:type:`PyArrayMethod_GetMultiReductionInitials` function. It is the
+   multi-output version of :c:macro:`NPY_METH_get_reduction_initial` and fills
+   one initial value per reduction output instead of a single one.
+   :meth:`~numpy.ufunc.reduce` uses it to seed the accumulators when the
+   reduction is empty or when a ``where=`` mask is given, for a ufunc whose
+   loop also registers a :c:macro:`NPY_METH_get_reduction_loop`.
+
+   Its signature matches :c:macro:`NPY_METH_get_reduction_initial`, except that
+   the final argument is ``void **initials``, an array of ``nout`` pointers,
+   one per reduction output, each pointing to the buffer to fill. The
+   *reduction_is_empty* argument and the -1/0/1 return value have the same
+   meaning as for :c:macro:`NPY_METH_get_reduction_initial`. A return of 1 must
+   fill every output.
+
+   A method may register at most one of
+   :c:macro:`NPY_METH_get_reduction_initial` and
+   ``NPY_METH_get_multi_reduction_initials``. The latter supports single-output
+   reductions too. For a ufunc with more than one output it must be paired with
+   a :c:macro:`NPY_METH_get_reduction_loop`, otherwise
+   :meth:`~numpy.ufunc.reduce` is unreachable and the identity would never be
+   used. See :ref:`c-api.reduction-loop-tutorial` for a worked example.
+
 Flags
 ~~~~~
 
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -1851,6 +1851,17 @@ the functions that must be implemented for each slot.
    initial value is correct, since NumPy may call this even when it is not
    strictly necessary to do so.
 
+.. c:type:: int (PyArrayMethod_GetMultiReductionInitials)( \
+        PyArrayMethod_Context *context, npy_bool reduction_is_empty, \
+        void **initials)
+
+   Multi-output version of :c:type:`PyArrayMethod_GetReductionInitial`, used to
+   query the per-output initial values for a reduction. It behaves the same
+   way, except that *initials* is an array of ``nout`` pointers, one per
+   reduction output, each pointing to the buffer to fill. The
+   *reduction_is_empty* argument and the -1, 0, or 1 return value have the same
+   meaning. A return of 1 must fill every output.
+
 .. c:macro:: NPY_METH_get_reduction_loop
 
    .. versionadded:: 2.6
@@ -1899,14 +1910,8 @@ the functions that must be implemented for each slot.
    one initial value per reduction output instead of a single one.
    :meth:`~numpy.ufunc.reduce` uses it to seed the accumulators when the
    reduction is empty or when a ``where=`` mask is given, for a ufunc whose
-   loop also registers a :c:macro:`NPY_METH_get_reduction_loop`.
-
-   Its signature matches :c:macro:`NPY_METH_get_reduction_initial`, except that
-   the final argument is ``void **initials``, an array of ``nout`` pointers,
-   one per reduction output, each pointing to the buffer to fill. The
-   *reduction_is_empty* argument and the -1/0/1 return value have the same
-   meaning as for :c:macro:`NPY_METH_get_reduction_initial`. A return of 1 must
-   fill every output.
+   loop also registers a :c:macro:`NPY_METH_get_reduction_loop`. See
+   :c:type:`PyArrayMethod_GetMultiReductionInitials` for the signature.
 
    A method may register at most one of
    :c:macro:`NPY_METH_get_reduction_initial` and
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -1879,7 +1879,7 @@ the functions that must be implemented for each slot.
    works for the traditional two-input/one-output case. Calling
    :meth:`~numpy.ufunc.reduce` on a ufunc with more than one output whose
    resolved ArrayMethod does not register a reduction loop raises a
-   :exc:`ValueError`. See :ref:`c-api.reduction-loop-tutorial` for a
+   :exc:`TypeError`. See :ref:`c-api.reduction-loop-tutorial` for a
    worked example.
 
    This slot only relaxes the *output* arity: :meth:`~numpy.ufunc.reduce`
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -1856,11 +1856,13 @@ the functions that must be implemented for each slot.
         void **initials)
 
    Multi-output version of :c:type:`PyArrayMethod_GetReductionInitial`, used to
-   query the per-output initial values for a reduction. It behaves the same
-   way, except that *initials* is an array of ``nout`` pointers, one per
-   reduction output, each pointing to the buffer to fill. The
-   *reduction_is_empty* argument and the -1, 0, or 1 return value have the same
-   meaning. A return of 1 must fill every output.
+   query the per-output initial values for a reduction. It behaves the same as
+   :c:type:`PyArrayMethod_GetReductionInitial`, except that *initials*
+   is an array of ``nout`` pointers, one per reduction output, each pointing
+   to the buffer to fill. The *reduction_is_empty* argument and the -1, 0, or 1
+   return value have the same meaning as :c:type:`PyArrayMethod_GetReductionInitial`.
+   A return of 1 indicates every initial value has been successfully initialized
+   with valid data.
 
 .. c:macro:: NPY_METH_get_reduction_loop
 
@@ -1871,7 +1873,8 @@ the functions that must be implemented for each slot.
    typedef used for ``NPY_METH_get_loop``). This is required to reduce
    ufuncs with more than one output, since the "forward" elementwise loop of
    such a ufunc cannot be used as a reduction loop the way a single-output loop
-   can (by aliasing the output to the first input). Instead, the returned
+   can (by pointing the output and the first input at the same memory, so that
+   the loop accumulates in place). Instead, the returned
    :c:type:`PyArrayMethod_StridedLoop` must implement the reduction
    directly, with an ``(nout + 1)``-in/``nout``-out signature: it takes the
    current per-output accumulators followed by one streamed input element,
@@ -1881,24 +1884,25 @@ the functions that must be implemented for each slot.
 
        [acc_0, ..., acc_{nout-1}, x, out_0, ..., out_{nout-1}]
 
-   where ``x`` is the streamed element being reduced in and each ``out_i``
-   is aliased to (and typically has stride 0 relative to) the matching
-   ``acc_i``.
+   where ``x`` is the streamed element being reduced in, and each ``out_i``
+   points at the same memory as the matching ``acc_i`` (and typically has a
+   stride of 0 relative to it).
 
    If ``NPY_METH_get_reduction_loop`` is not set, :meth:`~numpy.ufunc.reduce`
    falls back to ``NPY_METH_get_loop``/``NPY_METH_strided_loop``, which only
-   works for the traditional two-input/one-output case. Calling
+   works for the typical two-input/one-output case. Calling
    :meth:`~numpy.ufunc.reduce` on a ufunc with more than one output whose
    resolved ArrayMethod does not register a reduction loop raises a
    :exc:`TypeError`. See :ref:`c-api.reduction-loop-tutorial` for a
    worked example.
 
-   This slot only relaxes the *output* arity: :meth:`~numpy.ufunc.reduce`
-   (as well as :meth:`~numpy.ufunc.accumulate` and
-   :meth:`~numpy.ufunc.reduceat`) still requires the ufunc itself to have
-   exactly two inputs, independent of this slot. Calling any of them on a
-   ufunc with ``nin != 2`` raises a :exc:`ValueError` regardless of
-   ``nout`` or whether a reduction loop is registered.
+   Note that this slot only lifts the restriction on how many outputs a
+   ufunc may have. It does not change how many inputs a ufunc may have:
+   :meth:`~numpy.ufunc.reduce` (as well as :meth:`~numpy.ufunc.accumulate`
+   and :meth:`~numpy.ufunc.reduceat`) still only works on ufuncs that take
+   exactly two inputs, whether or not a reduction loop is registered.
+   Calling any of these methods on a ufunc with a number of inputs other
+   than two raises a :exc:`ValueError`.
 
 .. c:macro:: NPY_METH_get_multi_reduction_initials
 
--- a/doc/source/reference/c-api/array.rst
+++ b/doc/source/reference/c-api/array.rst
@@ -1888,6 +1888,15 @@ the functions that must be implemented for each slot.
    points at the same memory as the matching ``acc_i`` (and typically has a
    stride of 0 relative to it).
 
+   The *strides* argument passed to ``NPY_METH_get_reduction_loop`` itself at
+   setup time uses this same layout, so that ``strides[i]`` describes the
+   ``i``-th operand of the loop being requested. ``strides[nout]`` is the
+   stride of the streamed input, and each ``strides[nout + 1 + i]`` repeats
+   ``strides[i]``, because ``out_i`` and ``acc_i`` are the same buffer. The
+   accumulator strides are normally 0, since the reduction accumulates in
+   place. When a ``where=`` mask is used, one further entry at
+   ``strides[2 * nout + 1]`` holds the mask stride.
+
    If ``NPY_METH_get_reduction_loop`` is not set, :meth:`~numpy.ufunc.reduce`
    falls back to ``NPY_METH_get_loop``/``NPY_METH_strided_loop``, which only
    works for the typical two-input/one-output case. Calling
--- a/doc/source/user/basics.ufuncs.rst
+++ b/doc/source/user/basics.ufuncs.rst
@@ -76,7 +76,7 @@ so calling its methods raises an error::
    >>> np.divmod.reduce([1, 2, 3])
    Traceback (most recent call last):
        ...
-   ValueError: reduce only supported for functions returning a single value
+   ValueError: divmod.reduce is not supported: the resolved loop does not register a reduction loop
 
 The reduce-like methods all take an *axis* keyword, a *dtype*
 keyword, and an *out* keyword, and the arrays must all have dimension >= 1.
--- a/doc/source/user/basics.ufuncs.rst
+++ b/doc/source/user/basics.ufuncs.rst
@@ -65,7 +65,7 @@ Attempting to call these methods on other ufuncs will cause a
 :exc:`ValueError` (or, for :meth:`~numpy.ufunc.reduce` on a ufunc with more
 than one output, a :exc:`TypeError`), unless the ufunc's loop implementation
 registers a dedicated reduction loop, in which case :meth:`~numpy.ufunc.reduce`
-also works for ufuncs with more than one output; see
+also works for ufuncs with more than one output. See
 :doc:`c-info.reduction-loop-tutorial` for how to add one to a custom ufunc.
 
 For example, :func:`numpy.add` takes two inputs and returns one output,
--- a/doc/source/user/basics.ufuncs.rst
+++ b/doc/source/user/basics.ufuncs.rst
@@ -58,11 +58,14 @@ All ufuncs have 5 methods. 4 reduce-like methods
 (:meth:`~numpy.ufunc.reduce`, :meth:`~numpy.ufunc.accumulate`,
 :meth:`~numpy.ufunc.reduceat`, :meth:`~numpy.ufunc.outer`) and one
 for inplace operations (:meth:`~numpy.ufunc.at`).
-See :ref:`ufuncs.methods` for more. However, these methods only make sense on 
+See :ref:`ufuncs.methods` for more. However, these methods only make sense on
 ufuncs that take two input arguments and return one output argument (so-called
 "scalar" ufuncs since the inner loop operates on a single scalar value).
 Attempting to call these methods on other ufuncs will cause a
-:exc:`ValueError`.
+:exc:`ValueError`, unless the ufunc's loop implementation registers a
+dedicated reduction loop, in which case :meth:`~numpy.ufunc.reduce` (but not
+the other four methods) also works for ufuncs with more than one output; see
+:doc:`c-info.reduction-loop-tutorial` for how to add one to a custom ufunc.
 
 For example, :func:`numpy.add` takes two inputs and returns one output,
 so its methods work::
@@ -70,13 +73,18 @@ so its methods work::
    >>> np.add.reduce([1, 2, 3])
    6
 
-But :func:`numpy.divmod` returns two outputs (quotient and remainder),
-so calling its methods raises an error::
+But :func:`numpy.divmod` returns two outputs (quotient and remainder) and has no
+reduction loop registered, so calling its methods raises an error::
 
    >>> np.divmod.reduce([1, 2, 3])
    Traceback (most recent call last):
        ...
    ValueError: divmod.reduce is not supported: the resolved loop does not register a reduction loop
+
+For a ufunc with more than one output whose loop *does* register a reduction
+loop, :meth:`~numpy.ufunc.reduce` instead returns a tuple with one array per
+output. Its *initial* argument then also accepts either a single value,
+broadcast to every output, or a tuple with one value per output.
 
 The reduce-like methods all take an *axis* keyword, a *dtype*
 keyword, and an *out* keyword, and the arrays must all have dimension >= 1.
--- a/doc/source/user/basics.ufuncs.rst
+++ b/doc/source/user/basics.ufuncs.rst
@@ -62,9 +62,10 @@ See :ref:`ufuncs.methods` for more. However, these methods only make sense on
 ufuncs that take two input arguments and return one output argument (so-called
 "scalar" ufuncs since the inner loop operates on a single scalar value).
 Attempting to call these methods on other ufuncs will cause a
-:exc:`ValueError`, unless the ufunc's loop implementation registers a
-dedicated reduction loop, in which case :meth:`~numpy.ufunc.reduce` (but not
-the other four methods) also works for ufuncs with more than one output; see
+:exc:`ValueError` (or, for :meth:`~numpy.ufunc.reduce` on a ufunc with more
+than one output, a :exc:`TypeError`), unless the ufunc's loop implementation
+registers a dedicated reduction loop, in which case :meth:`~numpy.ufunc.reduce`
+also works for ufuncs with more than one output; see
 :doc:`c-info.reduction-loop-tutorial` for how to add one to a custom ufunc.
 
 For example, :func:`numpy.add` takes two inputs and returns one output,
--- a/doc/source/user/basics.ufuncs.rst
+++ b/doc/source/user/basics.ufuncs.rst
@@ -79,7 +79,7 @@ reduction loop registered, so calling its methods raises an error::
    >>> np.divmod.reduce([1, 2, 3])
    Traceback (most recent call last):
        ...
-   ValueError: divmod.reduce is not supported: the resolved loop does not register a reduction loop
+   TypeError: divmod.reduce is not supported: the resolved loop does not register a reduction loop
 
 For a ufunc with more than one output whose loop *does* register a reduction
 loop, :meth:`~numpy.ufunc.reduce` instead returns a tuple with one array per
--- a/doc/source/user/c-info.reduction-loop-tutorial.rst
+++ b/doc/source/user/c-info.reduction-loop-tutorial.rst
@@ -46,7 +46,7 @@ on a ufunc with ``nin != 2`` (regardless of ``nout``) raises a
 
 If a ufunc has more than one output and its resolved ``ArrayMethod`` does
 not register a reduction loop, calling ``.reduce()`` on it raises a
-:exc:`ValueError`, e.g. ``numpy.divmod.reduce`` fails since ``divmod``
+:exc:`TypeError`, e.g. ``numpy.divmod.reduce`` fails since ``divmod``
 has no reduction loop.
 
 Example: a two-output ``minimummaximum`` ufunc
--- a/doc/source/user/c-info.reduction-loop-tutorial.rst
+++ b/doc/source/user/c-info.reduction-loop-tutorial.rst
@@ -91,8 +91,8 @@ with ``/* BEGIN main computation */`` and ``/* END main computation */``.
          * minmax.c
          * A minimal ufunc for a single dtype ('f8') that computes the
          * running minimum and maximum in one pass, and registers a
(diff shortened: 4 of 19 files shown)


seberg:
CC @mhvk I would be very curious if you have some high-level thoughts on this?

mhvk:
@seberg - thanks for the ping!

@ikrommyd - this looks very nice indeed, and I like the general idea of having a new slot available to register reduction functions. Like you, I'm also pleased that in the end the changes are not very intrusive. 

On the PR itself, one thing I think is important is to have something in numpy that exercises the new function; possibly, your minmax, maybe just for integers or floats, could serve that purpose? You'll have seen that there are several `_core/src/umath/_*.c` test functions, which you could add too.

ikrommyd:
Thanks @mhvk! Yeah I was planning to open a separate PR targetting this PR's branch that implements minmax in numpy but it wouldn't be much different than what the gist already has. I mean one just registers the reduction loops using the slot and `PyUFunc_AddLoopFromSpec` which is how normally one adds ufunc loops in the new-style array method API. I'm just working on something else too the last couple of days so I got caught with that. Let me know if you have any other suggestions about what you'd like to see. I was also planning to do this in an external extension like user dtypes or quaddtype to show how such extensions would interact with this API. I was also considering writing a short NEP about this too once the design is consolidated just for documentation purposes.

ngoldbaum:
If Marten's OK with a minmax reduction landing as part of this PR, I'm fine with it too.

IMO a NEP isn't necessary to merge this, so it would be purely for documentation purposes. It would also add some process overhead that you might not want to deal with.

ikrommyd:
Yeah I wanted to keep the diff minimal for easier first look here on the design. We also need to touch reduceat and accumulate in a similar manner to reduce so that all 3 reduction-like ops work with this. Happy to add minmax + reduceat + accumulate changes here. I generally find smaller PRs easier to work with from the reviewers' perspective and then I would apply all the comments made here for reduce to accumulate and reduceat so what I originally had in mind was to have these and minmax as separate PRs but happy to adjust according to your liking :smiley:

mhvk:
All sounds good. I agree with the sense to stick to just `reduce` here, but include an example of some kind. I think your current `minmax` is perhaps more suitable as a simple test case, since it is not vectorized---I find it also really helpful to have those test examples...  But just suggestions, do follow what works best for you, this is a nice contribution as is!  One thing, as you were mentioning documentation, that definitely would need to be added, at least in the C API text.

doc/source/user/basics.ufuncs.rst
```diff
@@ -76,7 +76,7 @@ so calling its methods raises an error::
    >>> np.divmod.reduce([1, 2, 3])
    Traceback (most recent call last):
        ...
-   ValueError: reduce only supported for functions returning a single value
+   ValueError: divmod.reduce is not supported: the resolved loop does not register a reduction loop
```
ikrommyd: just wanted to make the docs ci pass, this whole section needs refactoring now that multi-output reductions are allowed

ikrommyd:
Okay since I have not seen any opposing feedback on the design here, I plan to turn this into a more proper PR this week with testing and documentation. As Marten suggested, I will add the minmax toy example as a C test.
Then people can properly review and the feedback will be considered in the separate PRs that will touch accumulate and reduceat in a similar manner.

ikrommyd:
I've added documentation updates and I took my minmax and turned it into a C test by adding a double loop only + promoter for python scalars. I didn't want to add full dtype support as that will be done in the actual minmax implementation that will follow. The linting failure is unrelated and is on a file I'm not touching here.

IMO this is ready for the first review pass. There may be some doc displaying issues cause I didn't check the website preview but will do that later. Gonna do a rebase to main too.

doc/source/user/c-info.reduction-loop-tutorial.rst
```diff
@@ -0,0 +1,395 @@
...
+This example is intentionally minimal and leaves out several things a
+real-world implementation, like the built-in :func:`numpy.add` or a fuller
+``minimummaximum``, would normally have:
+
+- The promoter always resolves to ``'f8'`` regardless of the input dtypes,
+  so e.g. integer input arrays are silently cast to ``float64`` rather than
+  keeping their own dtype. A multi-dtype ufunc would instead register a
+  loop per dtype and pick the common one, as in the built-in
+  :func:`numpy.minimum`/:func:`numpy.maximum`.
+- It has no :c:macro:`NPY_METH_get_reduction_initial`, so
```
seberg: And I guess unfortunately that would need a `char **initial` or so (I don't have a thought beyond adding that and maybe considering to actually do a deprecation of the current one). 
But yeah, we can follow up...

numpy/_core/src/umath/reduction.c
```diff
@@ -302,38 +317,68 @@ PyUFunc_ReduceWrapper(PyArrayMethod_Context *context,
...
+                }
             }
         }
         else {
             /*
              * Fetch initial from ArrayMethod, we pretend the reduction is
              * empty when the iteration is.  This may be wrong, but when it is,
              * we will not need the identity as the result is also empty.
+             * The identity seeds output 0 and is broadcast to the rest (the
+             * reduction outputs share a dtype).
```
seberg: Do they have to, strictly speaking, share a dtype, though?  (I know we brainstormed this, but I don't recall quite where it went, I thought it went towards "maybe not".  That said, I am OK if we for now assume that they do here and allow using `get_reduce_initial` in that case)

numpy/_core/src/umath/reduction.c
```diff
@@ -302,38 +317,68 @@ PyUFunc_ReduceWrapper(PyArrayMethod_Context *context,
...
-                initial_buf = NULL;
+                /* We have no initial value available, free buffers to indicate */
+                for (int i = 0; i < nout; i++) {
+                    PyMem_FREE(initial_buf[i]);
+                    initial_buf[i] = NULL;
+                }
+            }
+            else {
+                for (int i = 1; i < nout; i++) {
+                    memcpy(initial_buf[i], initial_buf[0], op_dtypes[i]->elsize);
```
seberg: This is fishy for refcounted types we can't just memcpy.  I dunno if it may just be pragmatic to call the function multiple times (and presume that it would be ridiculous if it wasn't idempotent w.r.t. to the `has_initial` return, maybe by just breaking when it is 0?).

numpy/_core/src/umath/reduction.c
```diff
@@ -429,27 +480,46 @@ PyUFunc_ReduceWrapper(PyArrayMethod_Context *context,
...
+    else {
+        ret = PyTuple_New(nout);
+        if (ret == NULL) {
+            goto fail;
+        }
+        for (int i = 0; i < nout; i++) {
+            PyArrayObject *res_i = NpyIter_GetOperandArray(iter)[i];
+            Py_INCREF(res_i);
+            PyTuple_SET_ITEM(ret, i, (PyObject *)res_i);
+        }
```
seberg, line 515: I think we have a helper (or maybe CPython by now) that creates a tuple from a C-array + size that you can use here.

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -2492,16 +2629,18 @@ reduce_loop(PyArrayMethod_Context *context,
...
-        dataptrs_copy[2] = dataptrs[0];
-        strides_copy[0] = strides[0];
-        strides_copy[1] = strides[1];
-        strides_copy[2] = strides[0];
+        /* Expand N+1 iterator operands into the (N+1)->N loop args */
+        for (int i = 0; i < nout; ++i) {
+            dataptrs_copy[i] = dataptrs[i];
+            strides_copy[i] = strides[i];
+            dataptrs_copy[nout + 1 + i] = dataptrs[i];
+            strides_copy[nout + 1 + i] = strides[i];
```
seberg, line 2654: Is there actually a reason to promise this part for a reduce loop?  I.e. `dataptrs_copy[nout + 1 + i] = dataptrs[i]` is always true, so reduce loops shouldn't even read it (I believe that also goes for your example).

I think the actual thing that we _must_ do is set:
```
dataptrs_copy[2] = dataptrs[0];
``` 
and even that we actually only need when `nout == 1` (because I we don't want to think about whether this is an explicit reduce loop or a normal loop that is being used as one.

Funnily, this might even be problematic in theory, because we may not check `NPY_MAXARGS` for being smaller than `nout * 2 + 1`.
(Might also be nice to avoid `NPY_MAXARGS` and use the "workspace" setup even if a bit tedious, but if it seems awkward in multiple places, I don't care.)

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -2273,6 +2273,135 @@ PyUFunc_GenericFunctionInternal(PyUFuncObject *ufunc,
...
+    for (int i = 0; i < fwd_nargs; i++) {
+        Py_XDECREF(operation_DTypes[i]);
+        Py_XDECREF(fwd_signature[i]);
+    }
+    if (res < 0) {
+        return NULL;
+    }
+
+    if (ufuncimpl->get_reduction_loop == NULL) {
+        PyErr_Format(PyExc_ValueError,
```
seberg: Should we just use a _UfuncNoLoop` error (which is a TypeError subclass)?  Although, I guess it _is_ a bit different.  Either way, I think it should be a  TypeError.

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -2273,6 +2273,135 @@ PyUFunc_GenericFunctionInternal(PyUFuncObject *ufunc,
...
+ *
+ * The outputs need not share a dtype: each `out_i` (and its aliased
+ * accumulator `acc_i`) takes the i-th resolved *forward output* descriptor,
+ * so they can differ from one another.  Only the single streamed input has a
+ * separate dtype, taken from the resolved forward *input* descriptor (the
+ * forward inputs all correspond to the array being reduced).  The sole
+ * cross-operand constraint is the structural one that `acc_i` aliases `out_i`.
+ */
+static PyArrayMethodObject *
+reducelike_promote_and_resolve_multi(PyUFuncObject *ufunc,
```
seberg, line 2298: I should re-read this once more eventually, but it looks not all that scary!  Hopefully we don't find weird corner cases :).

seberg (commented):
Overall, very nice, I have only some small comments.  I think that with the current minmax + tests using `initial=` we should have a fairly reasonable test coverage, which is also good (the no "identity" and identity logic part).
One thing we could make sure to test is fortran order (it shouldn't matter but that way we have `strides != 0` for the reduced array), object dtype could be nice, but I don't want to bloat this too much (i.e. the initial buffer bug).

numpy/_core/src/umath/ufunc_object.c
ikrommyd: Yeah if you have any ideas on the algorithm, please let me know. This is the the algorithm I came up with but I don't know much about dtype resolution edge cases.

numpy/_core/src/umath/reduction.c
ikrommyd: you mean `PyTuple_FromArray` probably

numpy/_core/src/umath/ufunc_object.c
ikrommyd: Yeah `_UFuncNoLoopError`'s good I think here. Or we make a new one `_UFuncNoReductionLoopError` :)

numpy/_core/src/umath/ufunc_object.c
ikrommyd: Ah I must have forgotten this. I wanted to add a check that `nout * 2 + 1` is not too big and raise an error instantly actually before anything is called in reduction code paths. If you literally need more than `NPY_MAXARGS` allows you to do a reduction, something is seriously wrong. I'm reading the comments a bit quickly as I'm at scipy so I will read the rest of this one later more carefully.

numpy/_core/src/umath/reduction.c
ikrommyd: Yeah this is a bug I think. Because the reduction loop may as well have different dtypes per output array in the same way you can have a ufunc forward loop have different dtypes per output array. But get_reduction_initial gives us a number that we blindly memcpy the bytes here for all outputs. Any suggestions? A new slot for multi-output reduction initials?

ikrommyd: Cause imagine you have a `sumprod` reduction and you want the initials to be 0 and 1 respectively. We should either allow default initials like that or disable initial entirely for multi-output reductions unless passed by the user with the `initial` arg.

seberg: >  A new slot for multi-output reduction initials?

Yeah, but dunno if we should here or just defer.  It seems pretty sane (and maybe obvious enough) that you just can't use this unless both the dtype and the value is identical for all.

Either way, enjoy SciPy for now!

ikrommyd: if we defer this from this PR, what do you suggest we do? Ignore it or raise an error if `get_reduction_initial` is set?

numpy/_core/src/umath/reduction.c
ikrommyd: But don't I still need to build the C array using `NpyIter_GetOperandArray` in a loop? So I thought doing `PyTuple_SET_ITEM` inside the same loop is best? Am I stupid and not seeing it?

ikrommyd:
Merge conflict is simple. It's just the new member in the array method struct from the loop caching PR. Will fix on the next rebase. Just don't wanna ruin the state right now.

numpy/_core/src/umath/reduction.c
seberg: You can also add it if you think that is best.  I guess it would be just not supporting initials but now that you say: I guess not supporting identities is likely very unsatisfying.

numpy/_core/src/umath/reduction.c
seberg: Why in a loop?  Why wouldn't `PyTuple_FromArray(nout, NpyIter_GetOperandArray(iter));` work?

ikrommyd: Ah yeah `PyTuple_FromArray((PyObject *const *)NpyIter_GetOperandArray(iter), nout);` to be precise will work. I'm stupid.

numpy/_core/src/umath/ufunc_object.c
ikrommyd: Switched this to a custom `TypeError`. The message that `raise_no_loop_found_error` gives did not feel right to me here. I prefer the custom "the resolved loop does not register a reduction loop" vs the generic "ufunc 'divmod' did not contain a loop with signature matching types" cause that doesn't specify that it is a reduction problem only. Or I can do a new `_UFuncNoReductionLoopError` if you prefer that.

numpy/_core/src/umath/ufunc_object.c
ikrommyd: I added a check during the spec validation that `meth->nout` is not greater than `(NPY_MAXARGS - 2) / 2` which comes from solving `2 * nout + 1 + 1 (for the mask) = NPY_MAXARGS`

numpy/_core/src/umath/reduction.c
ikrommyd: Yeah I think I'll add it in this PR. I'd like it if this one was complete and defer only `reduceat` and `accumulate` for immediate follow-up PRs.

ikrommyd: @seberg I added `get_multi_reduction_initials`, check the latest commits

ikrommyd:
I have implemented one more small `get_multi_reduction_initials` method on the array method to be able to set initial values to multi output reductions. Potentially we can deprecate the old `get_reduction_initial` and use this one which can definitely do the single output reductions too as Sebastian suggested considering earlier.

doc/release/upcoming_changes/31816.c_api.rst
```diff
@@ -0,0 +1,21 @@
...
+``(N+1)``-in/``N``-out signature: it consumes the running accumulators
+together with one streamed input element and produces the updated
+accumulators. Such a loop can also register per-output reduction identities
+through the ``NPY_METH_get_multi_reduction_initials`` slot (the multi-output
+analogue of ``NPY_METH_get_reduction_initial``), used to seed empty reductions
+and reductions with a ``where=`` mask.
+:meth:`~numpy.ufunc.reduce` on such a ufunc returns a tuple with one array per
+output, its ``out`` argument accepts a tuple of output arrays (one per
+output), and its ``initial`` argument accepts either a single value
+(broadcast to every output) or a tuple with one value per output.
```
ngoldbaum: This is far too detailed and technical for a release note. It's ok to gloss over stuff and only cover the most important things. Try to aim for a 2-3 line explanation, if you can.

doc/source/reference/c-api/array.rst
```diff
@@ -1851,6 +1851,76 @@ the functions that must be implemented for each slot.
    initial value is correct, since NumPy may call this even when it is not
    strictly necessary to do so.
 
+.. c:type:: int (PyArrayMethod_GetMultiReductionInitials)( \
+        PyArrayMethod_Context *context, npy_bool reduction_is_empty, \
+        void **initials)
+
+   Multi-output version of :c:type:`PyArrayMethod_GetReductionInitial`, used to
+   query the per-output initial values for a reduction. It behaves the same
+   way, except that *initials* is an array of ``nout`` pointers, one per
```
ngoldbaum: Same way as what? Maybe instead: "It behaves the same as ``PyArray_GetReductionInitial``"?

doc/source/reference/c-api/array.rst
```diff
@@ -1851,6 +1851,76 @@ the functions that must be implemented for each slot.
...
+   reduction output, each pointing to the buffer to fill. The
+   *reduction_is_empty* argument and the -1, 0, or 1 return value have the same
+   meaning. A return of 1 must fill every output.
```
ngoldbaum: similarly, "have the same meaning as ``PyArray_GetReductionInitial``". It's OK to repeat yourself IMO to make it extra clear exactly which function you're referring to.

ngoldbaum: Also maybe for the last sentence "A return value of 1 indicates every ``initial`` value has been initialized with valid data"

doc/source/reference/c-api/array.rst
```diff
@@ -1851,6 +1851,76 @@ the functions that must be implemented for each slot.
...
+   implemented as a :c:type:`PyArrayMethod_GetLoop` function (the same
+   typedef used for ``NPY_METH_get_loop``). This is required to reduce
+   ufuncs with more than one output, since the "forward" elementwise loop of
+   such a ufunc cannot be used as a reduction loop the way a single-output loop
+   can (by aliasing the output to the first input). Instead, the returned
+   :c:type:`PyArrayMethod_StridedLoop` must implement the reduction
+   directly, with an ``(nout + 1)``-in/``nout``-out signature: it takes the
+   current per-output accumulators followed by one streamed input element,
+   and writes the updated accumulators. That is, for a ufunc whose forward
+   loop has ``nout`` outputs, the *data*, *strides*, and descriptor arrays
```
ngoldbaum, line 1882: should "descriptor" also be italicized like the previous three terms?

doc/source/reference/c-api/array.rst
```diff
@@ -1851,6 +1851,76 @@ the functions that must be implemented for each slot.
...
+
+       [acc_0, ..., acc_{nout-1}, x, out_0, ..., out_{nout-1}]
+
+   where ``x`` is the streamed element being reduced in and each ``out_i``
+   is aliased to (and typically has stride 0 relative to) the matching
+   ``acc_i``.
+
+   If ``NPY_METH_get_reduction_loop`` is not set, :meth:`~numpy.ufunc.reduce`
+   falls back to ``NPY_METH_get_loop``/``NPY_METH_strided_loop``, which only
+   works for the traditional two-input/one-output case. Calling
```
ngoldbaum: maybe "typical" instead of "traditional". Unless Greeks have traditional practices around ufuncs 😜.

doc/source/reference/c-api/array.rst
```diff
@@ -1851,6 +1851,76 @@ the functions that must be implemented for each slot.
...
+   :meth:`~numpy.ufunc.reduce` on a ufunc with more than one output whose
+   resolved ArrayMethod does not register a reduction loop raises a
+   :exc:`TypeError`. See :ref:`c-api.reduction-loop-tutorial` for a
+   worked example.
+
+   This slot only relaxes the *output* arity: :meth:`~numpy.ufunc.reduce`
```
ngoldbaum: "relaxes the output arity" is far too technical, please rephrase in more words to make it clearer what you're saying.

doc/source/user/c-info.reduction-loop-tutorial.rst
```diff
@@ -0,0 +1,413 @@
...
+
+Background
+==========
+
+:meth:`~numpy.ufunc.reduce` (and the ``.sum()``/``.prod()``/... methods
+built on top of it for functions like :func:`numpy.add` and
+:func:`numpy.multiply`) historically only worked for ufuncs with exactly
+one output. The trick that makes this possible is that the ordinary
+two-input/one-output elementwise loop can be used as the reduction loop:
+if the output pointer is aliased to the first input pointer (with stride 0)
```
ngoldbaum: "aliased" is too technical for this discussion without a definition. Either define it explicitly or use different, simpler wording here and below.

doc/source/user/c-info.reduction-loop-tutorial.rst
```diff
@@ -0,0 +1,413 @@
...
+and left unmoved, the loop keeps accumulating into the same location as it
+walks over the array.
+
+That trick does not generalize to ufuncs with more than one output, because
```
ngoldbaum: maybe add "(like ``unwrap`` or ``modf``)"?

doc/source/user/c-info.reduction-loop-tutorial.rst
```diff
@@ -0,0 +1,413 @@
...
+data/stride layout.
+
+``reduce`` (like ``accumulate`` and ``reduceat``) still requires the ufunc's
+forward loop to have exactly two inputs as reducing folds an accumulator and
+one new element together at each step, which is only well-defined for a
+binary combining operation. This is unrelated to the number of outputs and
+was already true before multi-output reduction existed: calling ``reduce``
+on a ufunc with ``nin != 2`` (regardless of ``nout``) raises a
+:exc:`ValueError` ("... only supported for binary functions"). Only the
+*output* arity is generalized here.
```
ngoldbaum: This text could use a code example to illustrate what you're talking about, otherwise it's hard to read. I'd also give a pass over this whole document to reduce the number of run-on sentences in technical discussions.

doc/source/user/c-info.reduction-loop-tutorial.rst
```diff
@@ -0,0 +1,413 @@
...
+
+If a ufunc has more than one output and its resolved ``ArrayMethod`` does
+not register a reduction loop, calling ``.reduce()`` on it raises a
+:exc:`TypeError`, e.g. ``numpy.divmod.reduce`` fails since ``divmod``
+has no reduction loop.
+
+Example: a two-output ``minimummaximum`` ufunc
```
ngoldbaum, line 69: Since we're planning to actually add this feature to NumPy, maybe we could make this tutorial cover a different combination of reductions that we probably won't add? Maybe `sumvar` or something a little unusual?

Otherwise when we actually add the feature to NumPy, this tutorial is kinda weird: why would anyone do this? 

Alternatively we can remember when you add a real `minmax` reduction to NumPy to come back and add some text about how this is an illustrative example of how NumPy's implementation works.

doc/source/user/c-info.reduction-loop-tutorial.rst
```diff
@@ -0,0 +1,413 @@
...
+===============================================
+
+The example below defines a ``minimummaximum`` ufunc for the ``'f8'``
+(``double``) dtype only, computing the running minimum and maximum in a
+single reduction pass: ``minimummaximum.reduce(a)`` returns the pair
+``(a.min(), a.max())``, computed in one loop over ``a`` rather than in two.
+
+Because ``NPY_METH_get_reduction_loop`` is an ``ArrayMethod`` slot, the
+ufunc's loop must be registered through the new-style ``PyArrayMethod_Spec``
```
ngoldbaum:
```suggestion
ufunc's loop must be registered through the ``PyArrayMethod_Spec``
```

it's not that new anymore, take it or leave it though

doc/source/user/c-info.reduction-loop-tutorial.rst
```diff
@@ -0,0 +1,413 @@
...
+API (:c:func:`PyUFunc_AddLoopFromSpec`) rather than the legacy
+``PyUFunc_FromFuncAndData`` loop table used in the previous tutorial. The
```
ngoldbaum, line 80: This argues for updating the previous tutorial...

Not saying you need to do that though.

doc/source/user/c-info.reduction-loop-tutorial.rst
```diff
@@ -0,0 +1,413 @@
...
+            int res = PyUFunc_AddLoopFromSpec(minimummaximum, &spec);
+            Py_DECREF(double_descr);
+            if (res < 0 || register_minimummaximum_promoter(minimummaximum) < 0
+                    || PyModule_AddObject(m, "minimummaximum", minimummaximum) < 0) {
+                Py_XDECREF(minimummaximum);
+                Py_DECREF(m);
+                return NULL;
+            }
+            return m;
+        }
```
ngoldbaum, line 300: FWIW I'm not reviewing these code examples in detail

doc/source/user/c-info.reduction-loop-tutorial.rst
```diff
@@ -0,0 +1,413 @@
...
+      .. code-block:: python
+
+         from setuptools import setup, Extension
+         from numpy import get_include
+
+         minmax = Extension('minmax',
+                            sources=['minmax.c'],
+                            include_dirs=[get_include()])
+
+         setup(name='minmax', version='1.0', ext_modules=[minmax])
```
ngoldbaum, line 362: Instead of repeating this build boilerplate maybe we could extract it from this tutorial and the one above and then you could link in both to a general "how to set up packaging for projects using the NumPy C API" tutorial.

numpy/_core/src/umath/loops_minmax.dispatch.c.src
```diff
@@ -34,7 +34,7 @@
 
 // special optimization for fp scalars propagates NaNs
 // since there're no C99 support for it
-#ifndef NPY_DISABLE_OPTIMIZATION
+#if 0 // benchmark: disable manual scalar optimization, use naive ternary SCALAR_OP
```
ngoldbaum: this and all other changes to this file must be reverted before merging (commenting so we don't forget)

numpy/_core/src/multiarray/array_method.c
```diff
@@ -316,6 +324,46 @@ fill_arraymethod_from_slots(
...
+            meth->get_multi_reduction_initials != NULL) {
+        PyErr_Format(PyExc_ValueError,
+                "ArrayMethod cannot register both get_reduction_initial and "
+                "get_multi_reduction_initials. Use get_multi_reduction_initials "
+                "alone (it supports single-output reductions too). "
+                "(method: %s)",
+                spec->name);
+        return -1;
+    }
+    if (meth->nout > 1 && meth->get_reduction_initial != NULL) {
```
ngoldbaum:
```suggestion
    if (meth->nout > 1 && meth->get_reduction_initial != NULL && meth->get_reduction_loop != NULL) {
```

Since a single-output `get_reduction_initial` on a multi-output method is harmless as long as no reduction loop is registered.

You could also add an `assert(context->method->nout == 1)` inside `reduction_get_initial` as a defense-in-depth for this.

numpy/_core/src/multiarray/array_method.c
```diff
@@ -961,6 +1009,48 @@ PyArrayMethod_GetMaskedStridedLoop(
...
+                        sizeof(char *) * nargs);
+    if (data == NULL) {
+        PyErr_NoMemory();
+        return -1;
+    }
+    data->base.free = _masked_stridedloop_data_free;
+    data->base.clone = NULL;  /* not currently used */
+    data->unmasked_stridedloop = NULL;
+    data->nargs = nargs;
+
```
ngoldbaum, line 1042: Rather than using the supplied `fixed_strides` below, maybe it makes sense to copy it into a stack-allocated array, to make sure `fixed_strides` always has valid data in a user `get_reduction_loop` implementation?

This follows the same pattern used at line 2776 in `ufunc_object.c`.

numpy/_core/src/multiarray/array_method.h
```diff
@@ -74,9 +74,52 @@ typedef struct PyArrayMethodObject_tag {
...
+reduction_get_loop_func(PyArrayMethodObject *meth)
+{
+    return meth->get_reduction_loop != NULL ? meth->get_reduction_loop : meth->get_strided_loop;
+}
+
+
+/* Returns the reduction identity/initial value(s) from whichever slot
+ * `context->method` registered. Registration guarantees at most one is set. */
+static inline int
+reduction_get_initial(PyArrayMethod_Context *context,
```
ngoldbaum, line 107: My AI model spotted that you missed the call to `get_reduction_initial` in `wrapping_array_method.c` and suggests the following to fix that and avoid possible crashes from a NULL pointer dereference:

```diff
--- a/numpy/_core/src/umath/wrapping_array_method.c
+++ b/numpy/_core/src/umath/wrapping_array_method.c
@@ -199,14 +199,20 @@ wrapping_method_get_identity_function(
     int nin = context->method->nin, nout = context->method->nout;
     PyArray_DTypeMeta **dtypes = context->method->wrapped_dtypes;
 
     if (context->method->translate_given_descrs(
             nin, nout, dtypes, context->descriptors, orig_descrs) < 0) {
         return -1;
     }
-    int res = context->method->wrapped_meth->get_reduction_initial(
-            &orig_context, reduction_is_empty, item);
+    void *initials[1] = {item};
+    int res = reduction_get_initial(
+            &orig_context, reduction_is_empty, initials);
     for (int i = 0; i < nin + nout; i++) {
         Py_DECREF(orig_descrs[i]);
     }
     return res;
 }
```

numpy/_core/src/umath/reduction.c
```diff
@@ -343,13 +384,13 @@ PyUFunc_ReduceWrapper(PyArrayMethod_Context *context,
...
     if (wheremask != NULL) {
-        if (PyArrayMethod_GetMaskedStridedLoop(context,
+        if (PyArrayMethod_GetMaskedReductionLoop(context,
                 1, strideptr, &strided_loop, &auxdata, &flags) < 0) {
             goto fail;
         }
     }
     else {
-        if (context->method->get_strided_loop(context,
+        if (reduction_get_loop_func(context->method)(context,
```
ngoldbaum, line 411: same comment as elsewhere, define a stack-allocated `fixed_strides` array that's explicitly initialized up to `nout`. Look at `try_reduce_contiguous` for an example that does this correctly.

numpy/_core/src/umath/reduction.c
```diff
@@ -429,27 +476,41 @@ PyUFunc_ReduceWrapper(PyArrayMethod_Context *context,
...
+    }
+    if (nout == 1) {
+        Py_INCREF(result);
+        ret = (PyObject *)result;
+    }
+    else {
+        ret = PyTuple_FromArray((PyObject *const *)NpyIter_GetOperandArray(iter), nout);
+        if (ret == NULL) {
+            goto fail;
+        }
```
ngoldbaum:
```suggestion
        ret = PyTuple_New(nout);
        if (ret == NULL) {
            goto fail;
        }
        for (int i = 0; i < nout; i++) {
            /* iterator operand may be a writeback temporary (COPY_IF_OVERLAP) */
            PyArrayObject *res_i = out[i] != NULL ?
                    out[i] : NpyIter_GetOperandArray(iter)[i];
            Py_INCREF(res_i);
            PyTuple_SET_ITEM(ret, i, (PyObject *)res_i);
        }
```

This way you can prefer an `out[i]` if it's given instead of always generating a writeback temporary.

ngoldbaum: You could also add a test that does e.g. `mm.reduce(a, axis=0, out=(a[0], a[1]))`, right now on your PR this returns arrays that are not `a[0]` and `a[1]`.

numpy/_core/src/umath/reduction.c
```diff
@@ -302,38 +318,63 @@ PyUFunc_ReduceWrapper(PyArrayMethod_Context *context,
...
+                goto fail;
+            }
         }
         if (initial != NULL) {
-            /* must use user provided initial value */
-            if (PyArray_Pack(op_dtypes[0], initial_buf, initial) < 0) {
-                goto fail;
+            /* must use user provided initial value(s) */
+            for (int i = 0; i < nout; i++) {
+                PyObject *initial_i = initial_is_tuple ? PyTuple_GET_ITEM(initial, i) : initial;
```
ngoldbaum, line 354: I think you should add an explicit check for `None` here, like this:

```
                if (initial_is_tuple && initial_i == Py_None) {
                            "entries of a tuple 'initial' cannot be None");
                    goto fail;
                }
```

Otherwise `mm.reduce(a, initial=(None, 5.0))` would return `(nan, 5.0)` (that's what `PyArray_Pack` does with `None` for float arrays). Since `intial=None` means "explicitly no initial value" users would plausibly expect a per-output version of that. You could also implement a per-output version of the default `initial=None` behavior I guess, but that seems pretty complicated with no use-case and can always be added later.

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -2273,6 +2273,143 @@ PyUFunc_GenericFunctionInternal(PyUFuncObject *ufunc,
...
+     * outputs keep their individual dtypes), while the single streamed input
+     * takes the forward *input* descriptor.
+     *
+     * There is only one streamed operand, but the forward loop has `nin`
+     * inputs. They are all the array being reduced, resolved with the same
+     * DType, so a sane resolver gives them equivalent descriptors. We keep the
+     * first as the stream dtype (asserted below) and discard the rest.
+     */
+    for (int i = 1; i < nin; i++) {
+        assert(PyArray_EquivTypes(fwd_descrs[0], fwd_descrs[i]));
```
ngoldbaum: I'd make this a `TypeError` rather than an assert. Remember that asserts only show up in debug builds.

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -2456,21 +2599,23 @@ reduce_loop(PyArrayMethod_Context *context,
...
-                dataptrs_copy[1] = dataptrs[1];
-                dataptrs_copy[2] = dataptrs[0];
-                strides_copy[0] = strides[0];
-                strides_copy[1] = strides[1];
-                strides_copy[2] = strides[0];
+                /* Expand N+1 iterator operands into the (N+1)->N loop args */
+                for (int i = 0; i < nout; ++i) {
+                    dataptrs_copy[i] = dataptrs[i];
+                    strides_copy[i] = strides[i];
+                    dataptrs_copy[nout + 1 +i] = dataptrs[i];
```
ngoldbaum:
```suggestion
                    dataptrs_copy[nout + 1 + i] = dataptrs[i];
```

nit

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -2587,59 +2761,87 @@ try_reduce_contiguous(
     char *src = PyArray_BYTES(arr);
     if (!has_initial) {
         /*
-         * No identity available -- seed the accumulator with arr[0] and
+         * No identity available, so seed each accumulator with arr[0] and
          * reduce over arr[1:].
          */
-        memcpy(accum, src, descrs[1]->elsize);
+        for (int i = 0; i < nout; i++){
```
ngoldbaum:
```suggestion
        for (int i = 0; i < nout; i++) {
```

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -3747,31 +3950,23 @@ PyUFunc_GenericReduction(PyUFuncObject *ufunc,
...
-            }
+        if (out_obj != Py_None && _set_full_args_out(ufunc->nout, out_obj, &full_args) < 0) {
+            goto fail;
         }
     }
     else if (out_obj) {
         if (out_obj == Py_Ellipsis) {
             out_obj = NULL;
             return_scalar = NPY_FALSE;
         }
```
ngoldbaum, line 3975: I think you also want to add a check for `Py_None` here with an early `return 0`, otherwise explicitly passing `out=None` will trigger a TypeError about `out` needing to be a tuple.

numpy/_core/src/umath/_reduction_loop_tests.c
```diff
@@ -0,0 +1,247 @@
...
+    PyArray_DTypeMeta *dt = NPY_DTYPE(double_descr);
+    PyArray_DTypeMeta *dtypes[4] = {dt, dt, dt, dt};
+
+    PyType_Slot slots[] = {
+        {NPY_METH_strided_loop, (void *)&double_minimummaximum_loop},
+        {NPY_METH_get_reduction_loop, (void *)&minimummaximum_get_reduction_loop},
+        {NPY_METH_get_multi_reduction_initials, with_identity ?
+         (void *)&minimummaximum_get_multi_reduction_initials : NULL},
+        {0, NULL},
+    };
```
ngoldbaum:
```suggestion
    PyType_Slot slots[4] = {
        {NPY_METH_strided_loop, (void *)&double_minimummaximum_loop},
        {NPY_METH_get_reduction_loop, (void *)&minimummaximum_get_reduction_loop},
        {0, NULL},
        {0, NULL},
    };
    if (with_identity) {
        slots[2].slot = NPY_METH_get_multi_reduction_initials;
        slots[2].pfunc = (void *)&minimummaximum_get_multi_reduction_initials;
    }
```

A little clearer IMO and less chance of someone copying this as an example and relying on an implementation detail in the fill loop to avoid a NULL-pointer dereference.

numpy/_core/tests/test_reduction_loop.py
ngoldbaum, line 1: Some suggestions from an AI model for improved coverage in the tests:

> `mmi.reduce(a, where=mask)` without initial — the registered identity should seed fully-masked slots (I verified `mmi.reduce(a, axis=1, where=row_mask)` yields the per-output identities for masked-out rows).
> `out= arrays` overlapping the operand (e.g. `out=(a[0], a[1]) with axis=0)` asserting both values and returned identity — currently fails the identity half, see reduction.c comment.
> `out=None` and `out=(None, arr)` (mixed) — the latter works today and matches __call__'s scalar-conversion behavior for the None slot.
> `initial=(None, x)` — currently silently NaN, see reduction.c comment.
> an object-dtype (refcounted) multi-output loop to exercise the PyArray_ClearBuffer paths; nothing covers REFCHK descriptors through the new code.

IMO the object-dtype tests are worth adding because the REFCHK codepaths are nontrivial.

doc/source/user/basics.ufuncs.rst
```diff
@@ -58,25 +58,34 @@ All ufuncs have 5 methods. 4 reduce-like methods
...
 :meth:`~numpy.ufunc.reduceat`, :meth:`~numpy.ufunc.outer`) and one
 for inplace operations (:meth:`~numpy.ufunc.at`).
-See :ref:`ufuncs.methods` for more. However, these methods only make sense on 
+See :ref:`ufuncs.methods` for more. However, these methods only make sense on
 ufuncs that take two input arguments and return one output argument (so-called
 "scalar" ufuncs since the inner loop operates on a single scalar value).
 Attempting to call these methods on other ufuncs will cause a
-:exc:`ValueError`.
+:exc:`ValueError` (or, for :meth:`~numpy.ufunc.reduce` on a ufunc with more
+than one output, a :exc:`TypeError`), unless the ufunc's loop implementation
```
ngoldbaum, line 66: I think the `TypeError` detail here needs a release note?

Could we avoid the different error, given the backward compatibility concerns, or is that annoying to deal with?

I don't have a strong opinion about making the change though, just always prefer backward compatibility if we can reasonably avoid changing things.

doc/source/reference/c-api/array.rst
```diff
@@ -1851,6 +1851,76 @@ the functions that must be implemented for each slot.
...
+   current per-output accumulators followed by one streamed input element,
+   and writes the updated accumulators. That is, for a ufunc whose forward
+   loop has ``nout`` outputs, the *data*, *strides*, and descriptor arrays
+   passed to the reduction loop are laid out as::
+
+       [acc_0, ..., acc_{nout-1}, x, out_0, ..., out_{nout-1}]
+
+   where ``x`` is the streamed element being reduced in and each ``out_i``
+   is aliased to (and typically has stride 0 relative to) the matching
+   ``acc_i``.
```
ngoldbaum: Worth also saying explicitly what the *setup-time* `strides` argument of `get_reduction_loop` contains.

numpy/_core/src/umath/reduction.c
ngoldbaum: I'm marking this as resolved.

numpy/_core/src/umath/reduction.c
ngoldbaum: I'm marking this one as resolved as well.

numpy/_core/src/umath/ufunc_object.c
ngoldbaum, line 6724: Maybe add a comment here that we could maybe support reduction loops here with `nout != 1`, but it would take some work. You could return the resolved forward-loop descriptors maybe? So `resolve_dtypes((f8, f8, None, None), reduction=True)` would resolve the forward loop and return `(stream, stream, out_0, out_1)`. This is really only useful for something like numba that is trying to wrap NumPy. This corresponds to your passing `&dummy_arrays[0]` below - that would be relaxed at the same time.

ngoldbaum (commented):
Meaty review for a meaty PR!

I think this is well on its way but there are still a bunch of issues my AI-assisted review found, see below.

doc/source/reference/c-api/array.rst
ikrommyd: it's not everywhere else in this file so no I'd say. Only `descr` is cause it's a variable name I guess

doc/source/reference/c-api/array.rst
ikrommyd: my parents used to tell my stories about ufuncs 🤣

doc/source/user/c-info.reduction-loop-tutorial.rst
ikrommyd: I'll remove the word alias/aliasing and just explain what I mean in words

doc/source/user/c-info.reduction-loop-tutorial.rst
ikrommyd: It's mostly copied from the C test so that's fine

doc/source/user/c-info.reduction-loop-tutorial.rst
ikrommyd: I think I want to keep minmax here and add a text later when `minmax` is implemented properly. I have the following reasons

1. `minmax` is nice because the result makes sense with or without initial values since np.min([]) raises by default in numpy. If we added a `sumprod` for example, initial values of 0 and 1 respectively are more needed.
2. It's good that it is derived from the tests and we know that this example is well-tested.
3. It's a real use-case so showing in the future that "hey this is a mock implementation of something real in numpy" is nice IMO

doc/source/user/c-info.reduction-loop-tutorial.rst
ikrommyd: Yes I can do that right after this PR. We can have a ufunc creation tutorial that describes the legacy ufunc creation method and the array method spec way. Or it can be both in the same page too.

doc/source/user/c-info.reduction-loop-tutorial.rst
ikrommyd: I think I'd prefer to do that together with updating the existing ufunc tutorials in a separate PR. I feel like writing a new tutorial about the packaging is out of scope here.

doc/source/user/basics.ufuncs.rst
ikrommyd: I had it value error in the beginning but Sebastian said that it should be a type error earlier. I added a release note.

doc/source/reference/c-api/array.rst
ikrommyd: Added a small paragraph

doc/source/user/c-info.reduction-loop-tutorial.rst
ikrommyd: I believe I've done so and done some light overhaul too.

numpy/_core/src/multiarray/array_method.c
ikrommyd: Implemented a stride copying

numpy/_core/src/umath/reduction.c
ikrommyd: Added `fixed_strides` here too

numpy/_core/src/umath/ufunc_object.c
ikrommyd: Added such a comment and TODO

numpy/_core/src/umath/ufunc_object.c
ikrommyd: Yeah I thought we wouldn't encounter this actually so I wanted to make this debug build only but sure

numpy/_core/src/umath/reduction.c
ikrommyd: Added the error and test for it + mentioning this in the docs too

numpy/_core/src/umath/ufunc_object.c
ikrommyd: Indeed that's a regression from main. Fixed and added a test.

numpy/_core/src/umath/reduction.c
ikrommyd: Added a test too

numpy/_core/src/multiarray/array_method.h
ikrommyd: Indeed. I was not looking for effects of `get_reduction_initial` outside the files I was already editing.

numpy/_core/src/umath/loops_minmax.dispatch.c.src
ikrommyd: Restored it already

numpy/_core/tests/test_reduction_loop.py
ikrommyd: The cases of overlapping out=, out=None, initial=(None, x) are already covered from test cases I added for previous review comments.

ikrommyd: I added the rest. The object dtype loop in the C test is fully vibe coded tbh cause I was too lazy to write it but looks correct to me.

ikrommyd:
Whew that took a while to get done haha. I feel like I've addressed all your review comments @ngoldbaum!

ngoldbaum:
@ikrommyd OK, I think this is ready now. You'll need to clear the merge conflict and then I think I'll go ahead and merge.

While you're doing that, I also think this test would add some coverage for the object forward loop:

```
def test_object_forward(self):
    a = np.array([3, -7, 12], dtype=object)
    b = np.array([1, 9, -2], dtype=object)
    got_min, got_max = mm(a, b)
    assert got_min.tolist() == [1, -7, -2]
    assert got_max.tolist() == [3, 9, 12]
```

ikrommyd:
@ngoldbaum - rebased, fixed the conflict and added the test you suggested.

ngoldbaum:
OK, no one else weighed in since I suggested merging this yesterday, so in it goes!

If anyone would like to register any comments or suggestions, just ping Iason and he'll get to it on a second pass.

PR #31817, opened 2026-06-30
BUG: einsum: use object dtype's own additive identity in reductions

### LINKED ISSUE #29200: BUG: einsum doesn't work with str objects
### Describe the issue:

np.einsum incorrectly assumes every dtype's empty form is from 0. Str objects in the intermediate outputs should be initialized to `""`, and specifying `out=np.array(['',''])` doesn't help.

```python
import numpy as np
# Adapted from https://numpy.org/doc/stable/reference/generated/numpy.tensordot.html
a = np.array(range(1, 9))
a.shape = (2, 2, 2)
A = np.array(('a', 'b', 'c', 'd'), dtype=object)
A.shape = (2, 2)
np.tensordot(a, A) # Works

np.einsum("ijk,jk->i", a, A) # Doesn't work with strings

A=np.array([[1,10],[100,1000]])
np.einsum("ijk,jk->i", a, A) # Works only with numbers
```

### Error message:

```shell
File "<python-input-1>", line 9, in <module>
    np.einsum("ijk,jk->i", a, A) # Doesn't work with strings
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/run/user/1000/venv/lib/python3.13/site-packages/numpy/_core/einsumfunc.py", line 1423, in einsum
    return c_einsum(*operands, **kwargs)
TypeError: unsupported operand type(s) for +: 'int' and 'str'
```

2.3.0
3.13.3 (main, Apr  8 2025, 19:55:40) [GCC 14.2.0]

### Runtime Environment:

[{'numpy_version': '2.3.0',
  'python': '3.13.3 (main, Apr  8 2025, 19:55:40) [GCC 14.2.0]',
  'uname': uname_result(system='Linux', node='daniel-desktop3', release='6.15.0', version='#53 SMP PREEMPT_DYNAMIC Mon Jun  2 17:02:18 EDT 2025', machine='x86_64')},
                      'found': ['SSSE3',
                                'SSE41',
                                'POPCNT',
                                'SSE42',
                                'AVX',
                                'F16C',
                                'FMA3',
                                'AVX2'],
                      'not_found': ['AVX512F',
                                    'AVX512CD',
                                    'AVX512_KNL',
                                    'AVX512_KNM',
                                    'AVX512_SKX',
                                    'AVX512_CLX',
                                    'AVX512_CNL',
                                    'AVX512_ICL',
                                    'AVX512_SPR']}},
 {'architecture': 'Haswell',
  'filepath': '/run/user/1000/venv/lib/python3.13/site-packages/numpy.libs/libscipy_openblas64_-56d6093b.so',
  'internal_api': 'openblas',
  'num_threads': 24,
  'user_api': 'blas',
  'version': '0.3.29'}]

### Context for the issue:

After discovering einsum, I now think of basically everything else in terms of it. I was learning np.tensordot and working through its examples. I was able to convert the nested `for` loop into an einsum, but the string example failed to convert.

There is also [this StackOverflow question](https://stackoverflow.com/q/15606937/10477326) about other kinds of objects.

CodeBlackwell:
## Summary

Fixes gh-29200. `np.einsum` zero-initialised its reduction accumulator to the
integer `0` for every dtype. For object arrays whose elements are not
int-addable (`str`, `list`, custom objects), the first accumulation step
computed `0 + element` and raised `TypeError`; supplying an initialised `out=`
did not help because the output buffer was overwritten with `0` first.

## Fix

- Initialise the einsum output to `None` when the sum-of-products runs in
  **object** dtype. The decision keys on the loop dtype, so
  `einsum(..., dtype='f8', out=object_array)` still seeds `0` (otherwise the f8
  buffer would cast `None` to `nan`).
- Treat a `None`/`NULL` accumulator in the object sum-of-products as the
  identity — the first contribution wins, so each element's own additive
  identity is used instead of a hard `0`.
- Empty object reductions still collapse to `0` (matching `np.sum`/`np.dot`),
  guarded by a zero iteration size, which can never be the data-`None` case
  (where the loop ran).

## Behaviour

```python
>>> a = np.arange(1, 9).reshape(2, 2, 2)
>>> A = np.array(('a', 'b', 'c', 'd'), dtype=object).reshape(2, 2)
>>> np.einsum('ijk,jk->i', a, A)        # before: TypeError
array(['abbcccdddd', 'aaaaabbbbbbcccccccdddddddd'], dtype=object)
>>> np.tensordot(a, A)                  # already worked — now matches
array(['abbcccdddd', 'aaaaabbbbbbcccccccdddddddd'], dtype=object)
```

## Tests

Adds `test_einsum_object_dtype_non_numeric` covering string reductions,
`tensordot` parity, unchanged numeric-object reductions, `out=`
reinitialisation, and the empty-reduction identity. The full `test_einsum.py`
suite passes (79 tests).

### DIFF
--- a/numpy/_core/src/multiarray/einsum.cpp
+++ b/numpy/_core/src/multiarray/einsum.cpp
@@ -17,10 +17,9 @@
 #include <numpy/npy_common.h>
 #include <numpy/arrayobject.h>
 
-#include <array_assign.h>   //PyArray_AssignRawScalar
-
 #include <ctype.h>
 extern "C" {
+#include <array_assign.h>   //PyArray_AssignRawScalar
 #include "convert.h"
 #include "common.h"
 #include "ctors.h"
@@ -1055,10 +1054,28 @@ PyArray_EinsteinSum(char *subscripts, npy_intp nop,
         goto fail;
     }
 
-    /* Initialize the output to all zeros or None*/
+    /*
+     * Initialize the output: zeros for numeric loops, but None when the
+     * sum-of-products itself runs in object dtype. Object dtype has no universal
+     * additive zero — a hard int 0 made `0 + element` raise TypeError for
+     * str/list/etc. (gh-29200). None is a "no value yet" sentinel the object loop
+     * replaces with the first contribution. The decision keys on the LOOP dtype,
+     * not the output array's: `einsum(..., dtype='f8', out=object_array)` reduces
+     * in f8 and must still seed 0 (else the f8 buffer casts None to nan).
+     */
     ret = NpyIter_GetOperandArray(iter)[nop];
-    if (PyArray_AssignZero(ret, NULL) < 0) {
-        goto fail;
+    {
+        PyArray_Descr *loop_dtype = NpyIter_GetDescrArray(iter)[nop];
+        if (loop_dtype->type_num == NPY_OBJECT && PyArray_ISOBJECT(ret)) {
+            PyObject *none = Py_None;
+            if (PyArray_AssignRawScalar(ret, PyArray_DESCR(ret), (char *)&none,
+                                        NULL, NPY_SAFE_CASTING) < 0) {
+                goto fail;
+            }
+        }
+        else if (PyArray_AssignZero(ret, NULL) < 0) {
+            goto fail;
+        }
     }
 
     /***************************/
@@ -1161,6 +1178,22 @@ finish:
         ret = out;
     }
     Py_INCREF(ret);
+
+    /*
+     * An empty object reduction (a contracted axis of size 0, so the whole
+     * iteration is empty) never reaches the sum-of-products and keeps its None
+     * identity. Collapse that to 0 so empty object sums match np.dot/np.sum
+     * (gh-29200). einsum contracts the same axes for every output element, so a
+     * zero iteration size means ALL outputs are empty — never the data-None case,
+     * where the loop ran (iter size > 0) and already used real contributions.
+     */
+    if (PyArray_ISOBJECT(ret) && PyArray_SIZE(ret) > 0
+            && NpyIter_GetIterSize(iter) == 0) {
+        if (PyArray_AssignZero(ret, NULL) < 0) {
+            Py_DECREF(ret);
+            goto fail;
+        }
+    }
 
     NpyIter_Deallocate(iter);
     for (iop = 0; iop < nop; ++iop) {
--- a/numpy/_core/src/multiarray/einsum.cpp
+++ b/numpy/_core/src/multiarray/einsum.cpp
@@ -1055,13 +1055,12 @@ PyArray_EinsteinSum(char *subscripts, npy_intp nop,
     }
 
     /*
-     * Initialize the output: zeros for numeric loops, but None when the
-     * sum-of-products itself runs in object dtype. Object dtype has no universal
-     * additive zero — a hard int 0 made `0 + element` raise TypeError for
-     * str/list/etc. (gh-29200). None is a "no value yet" sentinel the object loop
-     * replaces with the first contribution. The decision keys on the LOOP dtype,
-     * not the output array's: `einsum(..., dtype='f8', out=object_array)` reduces
-     * in f8 and must still seed 0 (else the f8 buffer casts None to nan).
+     * Initialize the output: zeros for numeric loops, None for object-dtype
+     * loops. Object dtype has no universal additive identity, so the object
+     * sum-of-products loop treats None as "no value yet" and seeds the
+     * reduction with the first contribution. Key on the loop dtype, not the
+     * output array's dtype: einsum(..., dtype='f8', out=object_array) reduces
+     * in f8 and must seed 0 (the f8 buffer cannot hold None).
      */
     ret = NpyIter_GetOperandArray(iter)[nop];
     {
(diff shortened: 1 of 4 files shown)


charris:
This needs a release note. See `doc/release/upcoming_changes/` for examples.

CodeBlackwell:
I've added the requested release note

numpy/_core/src/multiarray/einsum.cpp
```diff
@@ -1162,6 +1179,22 @@ PyArray_EinsteinSum(char *subscripts, npy_intp nop,
...
     Py_INCREF(ret);
 
+    /*
+     * An empty object reduction (a contracted axis of size 0, so the whole
+     * iteration is empty) never reaches the sum-of-products and keeps its None
+     * identity. Collapse that to 0 so empty object sums match np.dot/np.sum
+     * (gh-29200). einsum contracts the same axes for every output element, so a
+     * zero iteration size means ALL outputs are empty — never the data-None case,
+     * where the loop ran (iter size > 0) and already used real contributions.
+     */
```
ngoldbaum, line 1188: IMO these sorts of comments from Claude that memorialize old bugs aren't very useful. Comments should communicate only information that is not obvious from reading the current code. IMO that doesn't include describing old versions of the code that were buggy. I really wish that was in the Claude system prompt, but someone at Anthropic really likes comments.

numpy/_core/src/multiarray/einsum_sumprod.c.src
```diff
@@ -1062,7 +1062,20 @@ static void
             }
         }
 
-        PyObject *sum = PyNumber_Add(*(PyObject **)dataptr[nop], prod);
+        PyObject *accum = *(PyObject **)dataptr[nop];
+        PyObject *sum;
+        if (accum == NULL || accum == Py_None) {
```
ngoldbaum, line 1067: I'm not sure whether there are corner cases where this isn't an object dtype.

numpy/_core/src/multiarray/einsum.cpp
CodeBlackwell: Fair enough, and agreed. Trimmed all three comment blocks to present-tense rationale in 1513fa4b (the gh-29200 refs and old-behavior narration are gone from the C sources; the history lives in the test and the release note). Comment-only change, einsum tests still pass.

numpy/_core/src/multiarray/einsum_sumprod.c.src
CodeBlackwell: Good question to press on. I traced it: there is no path where this loop runs against a non-object accumulator.

The dispatch and the None-seed key on the same dtype:

- `get_sum_of_products_function` (einsum_sumprod.c.src:1269) selects a loop purely by `type_num`, and every return indexes a table by `[type_num]` (:1281, :1299, :1308, :1320, :1324). The tables are declared `[NPY_NTYPES_LEGACY][...]` and must "match up with the type enum" (:1093), with the `object_sum_of_products_*` variants occupying exactly the `NPY_OBJECT` row. So an object loop is only ever selected when the loop dtype is object.
- The caller passes `NpyIter_GetDescrArray(iter)[0]->type_num` (einsum.cpp:1131-1133), and einsum forces a single loop dtype across all operands: with an explicit `dtype=` every operand including the output gets that descr (einsum.cpp:1002-1011), and with `dtype=NULL` the iterator is constructed with `NPY_ITER_COMMON_DTYPE` (einsum.cpp:1045-1046). So `[0]->type_num` (the dispatch key) always equals `[nop]->type_num` (the key the None-seeding checks). When this loop runs, `dataptr[nop]` is an object slot, buffered cases included; conversely `dtype='f8', out=object_array` reduces in f8, never reaches this loop, and gets the 0 seed.

The `accum == NULL` half follows this loop's existing convention for the operand reads ("convention is to treat nulls as None", :1051 and :1057).

One behavioral consequence worth flagging while we're here: because None doubles as the "no contribution yet" seed, a genuine `None` data element now acts as an additive identity, e.g. `np.einsum('i->', np.array([None, 'x'], dtype=object))` returns `'x'`, where `np.sum` on the same array raises `TypeError` from `None + 'x'`. Happy to add a note or a guard if you'd rather those stay consistent.

CodeBlackwell: Small addendum after double-checking myself — the invariant is stronger than dtype equality:

- With `NPY_ITER_COMMON_DTYPE`, the iterator constructor replaces every operand's loop descriptor with the same object ("Replacing all data types", nditer_constr.c:339-348, over all `nop+1` operands including the output). The explicit `dtype=` branch in einsum.cpp does the same. So the dispatch key and the seed key don't just have equal type_nums, they read the same descriptor pointer.
- The one exotic case I could construct (object inputs with a non-object `out=`) fails at iterator construction under safe casting, so it never reaches a loop at all.

ngoldbaum:
Are you copy/pasting output from an LLM in your comment responses to me? We prefer to speak with humans.

CodeBlackwell:
Sorry if that was a bit dense, admittedly yes formatted by LLM but not just zero shot and wall spam pasted. I Genuinely just wanted you to know I had triple check myself and what i had done to do so.

CodeBlackwell:
Just a friendly bump. Anything else needed from me on this one?

ngoldbaum:
Does this actually have a real use-case where a human ran into an error on a real workflow? Einsum internals are very complicated and there aren't any NumPy contributors who can properly review changes like this. I don't particularly want to make a change that might break other usecases for something very niche, particularly if the code was mostly AI-generated by someone with very little context about NumPy.

PR #31821, opened 2026-07-01
BUG: fix numpy datetime API to be compatible with older versions

### LINKED ISSUE #31814: BUG: extensions built against NumPy 2.5 segfault at runtime against NumPy 2.4 because get_(datetime|timedelta)64_value/_unit route through unguarded 2.5 C-API slots
### Describe the issue:

In pandas, we released a version built with NumPy 2.5 and users experienced a segfault when _run_ with NumPy <= 2.4 e.g. https://github.com/pandas-dev/pandas/issues/66086. 

The following agent analysis identified that https://github.com/numpy/numpy/pull/31091 updated `get_datetime64_value`, `get_timedelta64_value` and `get_datetime64_unit` to call `_PyDatetimeScalarObject_GetValue`, `_PyTimedeltaScalarObject_GetValue` and `_PyDatetimeScalarObject_GetMetadata` respectively, and with NumPy <=2.4 the [`PyArray_API` table is shorter](https://github.com/numpy/numpy/pull/31091/changes#diff-f62ae46232e805312be16c5cc3ff8b0e3fc606dbd929b8bc62459a1031584080R418-R421) so accessing these new implementations will segfault

-------

Commit 15a344dcce52a0e17d3d8f7af37f858f98a9f058 changed the Cython inline helpers in `numpy/__init__.cython-30.pxd` from direct struct-field access to calls through brand-new C-API table slots, and registered those slots without a MinVersion("2.5") guard. As a result, any extension that cimport numpy and calls these helpers while targeting an older NPY_FEATURE_VERSION builds a binary that calls `PyArray_API[378]`. Run against NumPy ≤ 2.4 (where the table is shorter), this reads past the end of the API table → garbage function pointer → hard crash (SIGBUS / SIGSEGV), with no Python-level traceback.

`numpy/__init__.cython-30.pxd` before #31091 (≤ 2.4):

```cython
cdef inline npy_timedelta get_timedelta64_value(object obj) noexcept nogil:
    return (<PyTimedeltaScalarObject*>obj).obval        # version-agnostic struct access
```

After #31091 (2.5):

```cython
cdef inline npy_timedelta get_timedelta64_value(object obj) noexcept nogil:
    return _PyTimedeltaScalarObject_GetValue(obj)       # -> PyArray_API[378], new in 2.5
```

The new slots in numpy/_core/code_generators/numpy_api.py:

```python
'_PyDatetimeScalarObject_GetMetadata':  (375,),   # note: no MinVersion
'_PyTimedeltaScalarObject_GetMetadata': (376,),
'_PyDatetimeScalarObject_GetValue':     (377,),
'_PyTimedeltaScalarObject_GetValue':    (378,),
```

```python
(A minimal, AI generated example)

# td_repro.pyx
# cython: language_level=3
cimport numpy as cnp

cnp.import_array()

def td_value(obj):
    """Return the int64 underlying a numpy.timedelta64 scalar."""
    return cnp.get_timedelta64_value(obj)

def dt_value(obj):
    """Return the int64 underlying a numpy.datetime64 scalar."""
    return cnp.get_datetime64_value(obj)

# setup.py
from setuptools import Extension, setup
import numpy as np
from Cython.Build import cythonize

ext = Extension(
    "td_repro",
    ["td_repro.pyx"],
    include_dirs=[np.get_include()],
    define_macros=[
        ("NPY_NO_DEPRECATED_API", "NPY_1_7_API_VERSION"),
        ("NPY_TARGET_VERSION", "NPY_1_21_API_VERSION"),
    ],
)
setup(name="td_repro", ext_modules=cythonize([ext], language_level="3"))

# run.py
import numpy as np
import td_repro
td = np.timedelta64(28, "D")
print("timedelta value:", td_repro.td_value(td))
dt = np.datetime64("2020-01-01", "ns")
print("datetime value:", td_repro.dt_value(dt))
```

### Error message:

```shell
(Run with NumPy 2.4)

Current thread's C stack trace (most recent call first):
  Binary file ".../bin/python3.14", at _Py_DumpStack+0x48 [0x10282ba54]
  Binary file ".../bin/python3.14", at faulthandler_dump_c_stack+0x58 [0x102840914]
  Binary file ".../bin/python3.14", at faulthandler_fatal_error+0x17c [0x102840854]
  Binary file "/usr/lib/system/libsystem_platform.dylib", at _sigtramp+0x38 [0x185463744]
  Binary file ".../td_repro.cpython-314-darwin.so", at __pyx_pw_8td_repro_1td_value+0x29c [0x102e5f2c8]
  Binary file ".../bin/python3.14", at _TAIL_CALL_CALL+0x13c [0x10276e390]
  Binary file ".../bin/python3.14", at _PyEval_Vector+0x2c8 [0x102765584]
  Binary file ".../bin/python3.14", at PyEval_EvalCode+0xb0 [0x1027651fc]
  Binary file ".../bin/python3.14", at run_mod+0xf0 [0x10280daac]
  Binary file ".../bin/python3.14", at pyrun_file+0x90 [0x10280d84c]
  Binary file ".../bin/python3.14", at _PyRun_SimpleFileObject+0x100 [0x10280d280]
  Binary file ".../bin/python3.14", at _PyRun_AnyFileObject+0x50 [0x10280cbc4]
  Binary file ".../bin/python3.14", at pymain_run_file_obj+0xb4 [0x10283d500]
  Binary file ".../bin/python3.14", at pymain_run_file+0x48 [0x10283cc84]
  Binary file ".../bin/python3.14", at Py_RunMain+0x634 [0x10283c420]
  Binary file ".../bin/python3.14", at pymain_main+0x1e0 [0x10283d820]
  Binary file ".../bin/python3.14", at main+0x3c [0x102572634]
  Binary file "/usr/lib/dyld", at start+0x1b50 [0x18509be00]
Extension modules: numpy._core._multiarray_umath, numpy.linalg._umath_linalg, td_repro (total: 3)
```

NumPy 2.5

### Runtime Environment:

_No response_

_No response_

kumaraditya303:
### PR summary

Fixes https://github.com/numpy/numpy/issues/31814

This PR fixes the pandas crash because cython definitons were using symbols which are only present on numpy 2.5 or above. This PR adds fallback just like other macros to directly access the fields on non-opaque so that it is compatible with numpy 2.4 or older. 

#### AI Disclosure

None used.

### DIFF
--- a/numpy/_core/code_generators/cversions.txt
+++ b/numpy/_core/code_generators/cversions.txt
@@ -87,4 +87,4 @@
 # Version 22 (NumPy 2.5.0) Opaque PyObject ABI support and
 # NPY_DT_legacy_descriptor_proto support.
 # Version 22 (NumPy 2.6.0) No change
-0x00000016 = cdfba10e4e01454262d2293c394cbed5
+0x00000016 = 83149cba850edb61bf48e1dd180f17a6
--- a/numpy/_core/code_generators/numpy_api.py
+++ b/numpy/_core/code_generators/numpy_api.py
@@ -415,10 +415,10 @@ multiarray_funcs_api = {
     '_PyDataType_GET_ITEM_DATA':               (372, MinVersion("2.5")),
     '_PyArrayMultiIter_GET_ITEM_DATA':         (373, MinVersion("2.5")),
     '_PyArrayNeighborhoodIter_GET_ITEM_DATA':  (374, MinVersion("2.5")),
-    '_PyDatetimeScalarObject_GetMetadata':         (375,),
-    '_PyTimedeltaScalarObject_GetMetadata':        (376,),
-    '_PyDatetimeScalarObject_GetValue':            (377,),
-    '_PyTimedeltaScalarObject_GetValue':           (378,),
+    '_PyDatetimeScalarObject_GetMetadata':     (375, MinVersion("2.5")),
+    '_PyTimedeltaScalarObject_GetMetadata':    (376, MinVersion("2.5")),
+    '_PyDatetimeScalarObject_GetValue':        (377, MinVersion("2.5")),
+    '_PyTimedeltaScalarObject_GetValue':       (378, MinVersion("2.5")),
 }
 
 ufunc_types_api = {
--- a/numpy/_core/include/numpy/ndarraytypes.h
+++ b/numpy/_core/include/numpy/ndarraytypes.h
@@ -1547,6 +1547,14 @@ typedef struct {
 #define _PyDataType_GET_ITEM_DATA(descr) ((PyArray_Descr_fields *)(descr))
 #undef _PyArray_LegacyDescr_GET_ITEM_DATA
 #define _PyArray_LegacyDescr_GET_ITEM_DATA(descr) ((_PyArray_LegacyDescr_fields *)(descr))
+#undef _PyDatetimeScalarObject_GetMetadata
+#define _PyDatetimeScalarObject_GetMetadata(self) ((PyDatetimeScalarObject *)self)->obmeta
+#undef _PyTimedeltaScalarObject_GetMetadata
+#define _PyTimedeltaScalarObject_GetMetadata(self) ((PyTimedeltaScalarObject *)self)->obmeta
+#undef _PyDatetimeScalarObject_GetValue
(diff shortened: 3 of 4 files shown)


seberg (approved):
These diverging is clearly wrong, so LGTM, thanks for the quick fix!

This must be follow up with the API names/general cleanup before we release Python 3.15 wheels, though.

If you want to do this here, happy as well, otherwise please go ahead and just self-merge if you like.

PR #31822, opened 2026-07-01
TST: add tests for stable ABI numpy extensions

kumaraditya303:
### PR summary

This PR adds tests for stable ABI extensions compiled with numpy C-API as well as cython API. 
`limited_api.c` is used as a template and parameterized by `MODULE_NAME` for every combination of Python abi3 version and NumPy target version. Same is done for `limited_api_cython` where cython extension is tested against every combination of Python abi3 version and NumPy target version. I've expanded tests to cover the datetime APIs which was broken recently to avoid such future breakages.

The failing MSVC test on x86 is because of https://github.com/numpy/numpy/issues/31762 which is now detected as a failure. 

#### AI Disclosure

I used Claude to work on this.


kumaraditya303:
Seems like gcc 9 which is used in the failing old_gcc CI does not support taking address of opaque pointer while de-referencing it. In Limited C-API `PyTypeObject` is opaque and previously the tests were empty so this was missed, we may need to skip this on such old gcc compilers. 


```console
rror: dereferencing pointer to incomplete type ‘PyTypeObject’ {aka ‘struct _typeobject’}
   1156 | #define PyTimedeltaArrType_Type (*(PyTypeObject *)PyArray_API[216])
        |                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  /opt/hostedtoolcache/Python/3.12.13/x64/include/python3.12/object.h:384:80: note: in definition of macro ‘PyObject_TypeCheck’
    384 | #  define PyObject_TypeCheck(ob, type) PyObject_TypeCheck(_PyObject_CAST(ob), (type))
        |                                                                                ^~~~
  limited_api_cython_3_9_npy2_0.abi3.so.p/limited_api_cython.pyx.c:5078:47: note: in expansion of macro ‘PyTimedeltaArrType_Type’
   5078 |   __pyx_r = PyObject_TypeCheck(__pyx_v_obj, (&PyTimedeltaArrType_Type));
        |                                               ^~~~~~~~~~~~~~~~~~~~~~~
```

numpy/_core/tests/examples/limited_api/meson.build
```diff
@@ -28,40 +28,43 @@ npy_path = run_command(py, [
...
+        modname = 'limited_api_cython_' + py_ver.replace('.', '_') + '_npy' + npy_ver.replace('.', '_')
+        py.extension_module(
+            modname,
+            'limited_api_cython.pyx',
+            install: false,
+            cython_args: ['--module-name', modname],
+            c_args: [
+              '-DNPY_NO_DEPRECATED_API=0',
+              '-DNPY_TARGET_VERSION=' + npy_macro,
+              '-DCYTHON_LIMITED_API=1',
```
da-woods:
```suggestion
```

You don't need this. Cython works it out from `Py_LIMITED_API`

ngoldbaum:
We talked about this at the community meeting and no one objected to dropping gcc9 from the CI matrix. I'm going to put in a PR to do that.

numpy/_core/tests/examples/limited_api/meson.build
ngoldbaum: We still have a minimum cython version of 3.0.6 and I think this *is* needed on Cython 3.0.X. That said, in practice I think we're using Cython 3.1 or newer everywhere and we could probably bump our minimum required Cython version.

https://github.com/numpy/numpy/blob/b689e13d6ec31bd5f34969aef69d934614a797e9/meson.build#L34-L36

numpy/_core/tests/test_limited_api.py
```diff
@@ -94,18 +228,18 @@ def install_temp(tmpdir_factory):
...
-    import limited_api1  # Earliest (3.6)  # noqa: F401
-    import limited_api2  # cython  # noqa: F401
-    import limited_api_latest  # Latest version (current Python)  # noqa: F401
 
 @pytest.mark.skipif(
     sys.version_info < (3, 15), reason="opaque PyObject requires Python 3.15+"
 )
+@pytest.mark.skipif(
+    not IS_64BIT, reason="opaque abi3t extensions are broken on 32-bit"
+)
```
ngoldbaum, line 229: if we merge this before #31771, #31771 can delete this, right?

numpy/_core/tests/test_limited_api.py
```diff
@@ -82,6 +83,139 @@ def install_temp(tmpdir_factory):
...
+@pytest.mark.xfail(
+    sysconfig.get_config_var("Py_DEBUG"),
+    reason=(
+        "Py_LIMITED_API is incompatible with Py_DEBUG, Py_TRACE_REFS, "
+        "and Py_REF_DEBUG"
+    ),
+)
+@pytest.mark.xfail(
+    NOGIL_BUILD,
+    reason="Py_GIL_DISABLED builds do not currently support abi3",
```
ngoldbaum:
```suggestion
    reason="Py_GIL_DISABLED builds do not support abi3",
```

They never will

numpy/_core/tests/test_limited_api.py
```diff
@@ -82,6 +83,139 @@ def install_temp(tmpdir_factory):
...
+)
+@pytest.mark.parametrize("module_name", limited_api_module_names())
+def test_limited_api_abi3(install_temp, module_name):
+    mod = pytest.importorskip(module_name)
```
ngoldbaum: A real ABI regression will often manifest as an ImportError. I don't think we should skip failed imports here, it should be surfaced as an error.

numpy/_core/tests/examples/limited_api/meson.build
da-woods: I'm not sure that one test requires an upgrade necessarily

I'd probably skip the test at least on Cython 3.0.x. If I remember right, the Limited API support became usable around Cython 3.1 so I wouldn't expect it to be a useful test before that.

numpy/_core/tests/examples/limited_api/limited_api.c
```diff
@@ -1,13 +1,28 @@
...
 
 #include <Python.h>
 #include <numpy/ndarraytypes.h>
 #include <numpy/arrayobject.h>
 #include <numpy/ufuncobject.h>
+#include <stddef.h>
+#include <string.h>
+
+
+#define MODULE_NAME_STR NPY_STRINGIFY(MODULE_NAME)
```
ngoldbaum: Shouldn't this be `NPY_TOSTRING`? I'm pretty sure the current behavior is to include the string literal `MODULE_NAME`.

Claude found this, it also thinks a series of coincidences cause CI to fail to break because of this.

numpy/_core/tests/examples/limited_api/meson.build
```diff
@@ -28,40 +28,42 @@ npy_path = run_command(py, [
...
-      'limited_api_latest',
-      'limited_api_latest.c',
-      c_args: [
-        '-DNPY_NO_DEPRECATED_API=NPY_1_21_API_VERSION',
-      ],
-      include_directories: [npy_include_path],
-      limited_api: py.language_version(),
-  )
+py_limited_api_versions = ['3.9', '3.10', '3.11', '3.12', '3.13', '3.14', '3.15']
+npy_target_versions = ['2.0', '2.1', '2.2', '2.3', '2.4', '2.5']
```
ngoldbaum: Please add comments that there is a paired set of variables in `test_limited_api.py` that need to be updated when these are.

Also add comments at line 180 of `test_limited_api.py` pointing to these variables in `meson.build`.

numpy/_core/tests/examples/limited_api/meson.build
```diff
@@ -28,40 +28,42 @@ npy_path = run_command(py, [
...
+            limited_api: py_ver,
+        )
+        modname = 'limited_api_cython_' + py_ver.replace('.', '_') + '_npy' + npy_ver.replace('.', '_')
+        py.extension_module(
+            modname,
+            'limited_api_cython.pyx',
+            install: false,
+            cython_args: ['--module-name', modname],
+            c_args: [
+              '-DNPY_NO_DEPRECATED_API=0',
```
ngoldbaum: Why not 1_21 like the old test? or maybe a newer version?

numpy/_core/tests/examples/limited_api/limited_api.c
```diff
@@ -1,13 +1,28 @@
...
+#ifndef Py_LIMITED_API
+#error "Py_LIMITED_API must be defined before including this file"
 #endif
 
 #include <Python.h>
 #include <numpy/ndarraytypes.h>
 #include <numpy/arrayobject.h>
 #include <numpy/ufuncobject.h>
+#include <stddef.h>
+#include <string.h>
```
ngoldbaum, line 14: Are these leftover from an earlier pass? I don't think you actually use anything from these headers.

numpy/_core/tests/examples/limited_api/limited_api.c
```diff
@@ -318,46 +333,87 @@ limited_api_opaque_multi_iter_nexti(PyObject *mod, PyObject *args)
...
+
+PyMODINIT_FUNC
+ENTRY_FUNC_NAME(void)
+{
+    import_array();
+    import_umath();
+    return PyModule_Create(&moduledef);
+}
+
+#endif
```
ngoldbaum: No trailing newline.

numpy/_core/tests/examples/limited_api/limited_api_cython.pyx
```diff
@@ -0,0 +1,150 @@
...
+    cdef char *data
+
+    if cnp.PyArray_SIZE(arr) == 0:
+        return 0
+
+    it = cnp.NpyIter_New(
+        arr,
+        cnp.NPY_ITER_READONLY | cnp.NPY_ITER_EXTERNAL_LOOP | cnp.NPY_ITER_REFS_OK,
+        cnp.NPY_KEEPORDER, cnp.NPY_NO_CASTING, <cnp.dtype>NULL)
+    iternext = cnp.NpyIter_GetIterNext(it, NULL)
```
ngoldbaum, line 34: No error checking on either of these function calls.

numpy/_core/tests/test_limited_api.py
```diff
@@ -94,18 +228,18 @@ def install_temp(tmpdir_factory):
     NOGIL_BUILD,
     reason="Py_GIL_DISABLED builds do not currently support abi3",
```
ngoldbaum:
```suggestion
    reason="Py_GIL_DISABLED builds do not support abi3",
```

numpy/_core/tests/test_limited_api.py
```diff
@@ -82,6 +83,139 @@ def install_temp(tmpdir_factory):
...
+    assert flags == td.dtype.flags
+    assert base == 8    # NPY_FR_ms
+    assert num == 10
+
+    # Non-datetime descriptors have no c_metadata and are rejected.
+    with pytest.raises(RuntimeError):
+        mod.datetime_metadata(np.arange(3))
+
+
+def _check_cython_module(mod):
```
ngoldbaum: This shares a bunch of code with `_check_c_module`. Can you unify these two helpers?

ngoldbaum (commented):
Overall looks pretty good. I left some comments inline.

numpy/_core/tests/test_limited_api.py
kumaraditya303: yes

numpy/_core/tests/test_limited_api.py
kumaraditya303: done

numpy/_core/tests/examples/limited_api/meson.build
kumaraditya303: done

numpy/_core/tests/examples/limited_api/limited_api.c
kumaraditya303: In limited API these headers are not included by Python.h and numpy requires them for `PyArray_ITER_RESET` which uses memset

numpy/_core/tests/examples/limited_api/limited_api.c
kumaraditya303: Good catch, fixed.

ngoldbaum:
I think if you rebase on or merge with main the error will go away because #31843 is merged.

numpy/_core/tests/examples/limited_api/meson.build
```diff
@@ -28,40 +28,45 @@ npy_path = run_command(py, [
...
+    if py.language_version().version_compare('>=' + py_ver)
+      foreach npy_ver : _NPY_TARGET_VERSIONS
+        modname = 'limited_api_' + py_ver.replace('.', '_') + '_npy' + npy_ver.replace('.', '_')
+        npy_macro = 'NPY_' + npy_ver.replace('.', '_') + '_API_VERSION'
+        py.extension_module(
+            modname,
+            'limited_api.c',
+            c_args: [
+              '-DMODULE_NAME=' + modname,
+              '-DNPY_TARGET_VERSION=' + npy_macro,
```
ngoldbaum: Shouldn't this also define NPY_NO_DEPRECATED_API?

numpy/_core/tests/test_limited_api.py
ngoldbaum: Doesn't this need to be bumped now as well?

numpy/_core/tests/test_limited_api.py
```diff
@@ -82,6 +84,123 @@ def install_temp(tmpdir_factory):
...
+        with pytest.raises(RuntimeError):
+            mod.datetime_metadata(np.arange(3))
+
+
+# Test limited API extension modules for all supported Python and NumPy versions
+# The _PY_ABI3_VERSIONS and _NPY_TARGET_VERSIONS lists should be kept in sync
+# with the lists defined in meson.build, and the test should be updated
+# if new versions are added here.
+_PY_ABI3_VERSIONS = ["3.9", "3.10", "3.11", "3.12", "3.13", "3.14", "3.15"]
+_NPY_TARGET_VERSIONS = ["2.0", "2.1", "2.2", "2.3", "2.4", "2.5"]
```
ngoldbaum: It may also make sense to add a case that doesn't define `NPY_TARGET_VERSION` at all, which the old test did.

numpy/_core/tests/examples/limited_api/limited_api_cython.pyx
```diff
@@ -0,0 +1,155 @@
...
+    while cnp.PyArray_ITER_NOTDONE(it):
+        total += (<double *>cnp.PyArray_ITER_DATA(it))[0]
+        cnp.PyArray_ITER_NEXT(it)
+    return total
+
+
+def iter_goto(cnp.ndarray arr, tuple coord):
+    """Get element at a coordinate using PyArray_ITER_GOTO."""
+    cdef int nd = cnp.PyArray_NDIM(arr)
+    cdef cnp.npy_intp destination[32]
```
ngoldbaum, line 91:
```suggestion
    cdef cnp.npy_intp destination[NPY_MAXDIMS_LEGACY_ITERS]
```

to better match the C test.

ngoldbaum (approved):
I did one more review pass and caught a couple minor issues and one suggestion about testing without `NPY_TARGET_VERSION` that I missed on the last pass. Otherwise I think this looks great so I'm approving. Please go ahead and merge this once you've addressed this last round of comments.

numpy/_core/tests/examples/limited_api/limited_api_cython.pyx
kumaraditya303: That macro is not available in Cython API, I'm sticking to defining it as constant for now.

numpy/_core/tests/test_limited_api.py
kumaraditya303: done

numpy/_core/tests/test_limited_api.py
kumaraditya303: bumped it to 3.1.0+

charris:
Missed the cython version check in `meson.build`. I'll fix that.

charris:
See #31875.

PR #31824, opened 2026-07-01
TYP: specialized ``logical_{and,or,xor}`` ufuncs

jorenham:
Similar to #31813, this adds specialized type signatures for 2-in, 1-out functions, but this time for the `logical_{and,or,xor}` ones, which, unlike the binary comparison ufuncs, also (and always) have `object_` output if one of the inputs has an `object_` dtype. The aggregation methods of these ones also accept all dtypes instead of just `bool`. These ones also have (boolean) identities, for which I used a generic type param to avoid duplication.

---

No AI.


jorenham:
The new primer errors for `colour` ([src](https://github.com/colour-science/colour/blob/c167aa6f56e8fbad99cd12abf4d43bf36ea2b20b/colour/geometry/intersection.py#L235)) are accurate and surface an actual typing issue in colour itself, so that's a win.

The `pandas` ([src](https://github.com/pandas-dev/pandas/blob/6e86ac5b43fbcf9f8d3c922c06a3fd9232ff0550/pandas/core/frame.py#L16069)) and `xarray` ([src](https://github.com/pydata/xarray/blob/098158d4b19863bcba3e448b8f5f5a8da87d610f/xarray/core/groupby.py#L544)) now-unused `# type: ignore` errors are both resolved false positives, so those are also wins.

charris:
Thanks Joren.

PR #31825, opened 2026-07-01
BUG/ENH: fix casting and setitem nan behavior between StringDType and float16/32

ngoldbaum:
### PR summary

This was originally part of #31682 but @seberg asked me to split it off. I also took his comments on that version about how I handled complex values into account.

Currently, we only support correctly parsing NaN values for float64 for casting and setitem. This leads to this (IMO) strange asymmetry:
 
```
>>> import numpy as np
>>> sdt = np.dtypes.StringDType(na_object=np.nan)
>>> arr = np.array([1, np.nan, 4.5], dtype=np.float32)
>>> arr.astype(sdt)
array(['1.0', 'nan', '4.5'], dtype=StringDType(na_object=nan))
>>> isinstance(arr.astype(sdt)[1], str)
True
>>> arr = np.array([1, np.nan, 4.5], dtype=np.float64)
>>> arr.astype(sdt)
array(['1.0', nan, '4.5'], dtype=StringDType(na_object=nan))
>>> arr.astype(sdt)[1] is np.nan
True
```

IMO this is just a straight-up bug in the implementation, which relies on coercing the value to a Python scalar, and fails to catch e.g. `np.float32('nan')` because float32 scalars aren't subtypes of PyFloat. That said, it is an observable behavior change.

We don't parse any complex-valued NaN as a NaN missing data value, leading to behavior like this:

```
>>> arr.astype(np.dtypes.StringDType(na_object=np.nan))
array(['(nan+0j)', 'nanj'], dtype=StringDType(na_object=nan))
```

This behavior is a bit more arguable to change, but under the principle that missing values should round-trip correctly via casting and that setitem should behave identically to casting, I also added behavior for complex values with `nan` real values to coerce to a null string if the na_object is NaN-like. This followes this existing behavior in NumPy:

```
>>> arr = np.array([complex(np.nan, 3.5), complex(0, np.nan), complex(3.5, np.nan), complex(np.nan, np.nan)])
>>> arr
array([nan+3.5j, 0. +nanj, 3.5+nanj, nan+nanj])
>>> arr.astype('float64')
<python-input-3>:1: ComplexWarning: Casting complex values to real discards the imaginary part
array([nan, 0. , 3.5, nan])
```

Note that the `ComplexWarning` happens unconditionally here for all complex values, even if the real part is zero.

I refactored StringDType internals to reflect this behavior. This refactoring included correcting some restructuring to reflect that the float to string casts might now produce null strings where has_nan_na was ignored before.

Added tests for this new behavior for floats and complex dtypes. Also adds a test that round-trips from float64 to float64 via complex, float64, and stringdtype, including all the permutations of that should round-trip.

#### AI Disclosure
An AI model initially noticed this issue. I also used an AI to brainstorm and review the code changes and tests.


seberg:
Let me ping a random pandas maintainer @mroeschke in the hope of a quick thought.

To be honest, to me it seems very strange to distinguish complex NaNs, probably _stanger_ than breaking the round-tripping of them.

ngoldbaum:
FWIW Pandas doesn’t use StringDType so I’m not sure why they’d care here. Or were you just trying to ask about what behavior makes sense?

seberg:
>  were you just trying to ask about what behavior makes sense?

Yes, mostly asking because I think pandas dev's have a better gut feeling for NaN/NA stuff then I do.

mroeschke:
> To be honest, to me it seems very strange to distinguish complex NaNs

I don't have much of a well informed opinion here so will tag in @jbrockmendel also.

My _gut_ says a `nan + nanj` value seems more definitely a NaN missing data value with `np.dtypes.StringDType(na_object=np.nan)` as any non-nan real or imaginary part could be significant?

Otherwise in pandas, for this "several behaviors are possible" cases, we're trying to move toward raising and telling the user to preprocess the data to a more predictable state for the operation

jbrockmendel:
I don't have a strong opinion on the exact API other than to aim for internal consistency.  np.isnan returns True when *either* component is NaN, so I'd suggest treating that as NaN-like across the board.

ngoldbaum:
OK, 3-against-1 on this one. I’ll go ahead and update this PR for the complex number case.

MaanasArora:
Thanks Nathan, I went through in detail and couldn't find any correctness issues. Neither did two AI agents (nothing real).

I agree we shouldn't treat real/imaginary parts differently for NaN handling, but a minor concern I have with using either nan as the check is whether we would lose information in casts... (A little wild, but I can somewhat imagine someone trying to render partial complex NaNs into strings and just ending with lost non-nan real and imaginary parts!) To be fair though, e.g., `real+nanj != real.nanj`, so I could see the case for consistency (even with `np.isnan`).

Was just curious if this mattered, but don't block on me!

ngoldbaum:
Sorry for the long delay here. I just pushed a commit implementing the suggestion to use the same rules as `np.isnan` for complex NaN.

numpy/_core/src/multiarray/stringdtype/casts.cpp
```diff
@@ -1208,6 +1235,66 @@ static PyType_Slot s2float_slots[] = {
...
+        if (value.real == -1.0 && PyErr_Occurred()) {
+            PyErr_Clear();  // out-of-double-range clongdouble: finite, not NaN
+            return 0;
+        }
+        if (!(npy_isnan(value.real) || npy_isnan(value.imag))) {
+            return 0;
+        }
+        if (npy_gil_warning(npy_static_pydata.ComplexWarning, 1,
+                            "Casting complex values to real discards "
+                            "the imaginary part") < 0) {
```
seberg: I would honestly drop this warning (here and below).  Yes, it's potentially lossy since we conflate real and complex part into a single "NA", but it isn't really clear to me and feels like it may just be annoying to deal with?

numpy/_core/src/multiarray/stringdtype/casts.cpp
```diff
@@ -1208,6 +1235,66 @@ static PyType_Slot s2float_slots[] = {
...
+    if (PyFloat_Check(obj) || PyArray_IsScalar(obj, Floating)) {
+        double value = PyFloat_AsDouble(obj);
+        if (value == -1.0 && PyErr_Occurred()) {
+            PyErr_Clear();  // out-of-double-range longdouble: finite, not NaN
+            return 0;
+        }
+        return float_is_nan_na<npy_double>(value) ? 1 : 0;
+    }
+    if (PyComplex_Check(obj) || PyArray_IsScalar(obj, ComplexFloating)) {
+        Py_complex value = PyComplex_AsCComplex(obj);
```
seberg, line 1362: My bot pointed out that we might overflow for longdoubles.  Happy to ignore though, unless it warns (and I am not sure it does), it ends up at the same place anyway.

seberg (approved):
Thanks, sorry about letting it hang, I am fine with basically just doing this definitely an improvement, it's all rather hard (what about quaddtype NaNs for example), but not sure we can fix this easily.

I am OK as is, although I would slightly lean to remove the `ComplexWarning`, so do with that what you think is right.

numpy/_core/src/multiarray/stringdtype/casts.cpp
ngoldbaum: Agreed, thanks for pointing this out

numpy/_core/src/multiarray/stringdtype/casts.cpp
ngoldbaum: I simplified the error handling a bit and I think that satisfies the concern here. I thought about adding a test for complex longdouble but... meh.

ngoldbaum:
Thanks for the review, let's bring this one in.

PR #31841, opened 2026-07-02
BLD: lower cython and meson-python minimum version and adjust dependabot config

ngoldbaum:
### PR summary

c.f. https://github.com/numpy/numpy/pull/31826#issuecomment-4860401938

Depdendabot incorrectly bumped these from versions that were set intentionally as minimum bounds by a human.

I added `increase-if-necessary` to prevent future pins of `>=` minimum bounds. This will still bump `==` pins.

https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/manage-your-dependency-security/controlling-dependencies-updated#defining-a-versioning-strategy

There were other incorrect bumps in the doc, release, and linter requirements but I'm ignoring those because they're less load-bearing than build requirements and I don't really see a problem with freezing the versions dependabot updated there.

#### AI Disclosure
No AI use


charris:
I updated some things in doc_requirements. I don't mind the updates as they allow us to test the dependencies as well as numpy. If you really don't want updates, ignore is probably the right thing to do.

ngoldbaum:
> I updated some things in doc_requirements.

IMO if you want dependabot to update these doc requirements you should make them be `==` versions. Want me to update the linter, doc, and release requirements to be `==`?

charris:
I'm not through with the docs yet, I want to test sphinx 9 to see if there are any new problems. I updated sphinx because some of the other doc dependencies couldn't update for the old version. What were you thinking of doing for the others?

charris:
The `==` do seem to get updated. Sphinx is actively ignored.

ngoldbaum:
> What were you thinking of doing for the others?

Just convert any `>=` bounds that have been updated recently to `==` bounds. IMO the build bounds really should be `>=`.

I'd rather not ignore these two in case we add more `>=` bounds in the future, which also probably shouldn't be updated automatically by dependabot.

rgommers (approved):
LGTM, thanks @ngoldbaum. Not sure if the discussion was done, so I won't merge.

The PRs to bump `>=` are obviously just nonsensical, they test nothing since the latest version gets installed either way. Dependabot should also be configured to group updates and bump less frequently, because there's 490 PRs by now and they're mostly just spam.

lint/release `==` pinning makes sense to me. docs requirements less so, unless it's actually problematic in CI - but we support a range of versions, and many contributors use that.

These requirements files should go soon by the way, they're obsolete now that `dependency-groups` in `pyproject.toml` is a thing. SciPy already removed them and it looks much better now.

ngoldbaum:
I went ahead and made dependabot updates weekly instead of daily and dropped all the explicit `>=` pins on the docs requirements, which should also reduce dependabot noise. I also made the linter and release gitpython dependency be `==` requirements.

@charris do you have any problems with these changes still?

charris (approved):

PR #31843, opened 2026-07-02
BLD: set minimum required gcc version to 10.3

### LINKED ISSUE #31828: BLD: update minimum required GCC to GCC 10
See https://github.com/numpy/numpy/pull/28102, which was the last time we bumped the minimum required GCC version 18 months ago.

According to https://scipy.github.io/devdocs/dev/toolchain.html#compilers, GCC 9 is still the minimum required for both NumPy and SciPy.

Over in https://github.com/numpy/numpy/pull/31822 we are hitting CI failures that happen only with GCC 9.

GCC 9 ships with Ubuntu 20.04 (six years old and EOL now) and is not needed on any of the manylinux images we use, they need GCC 10 or newer.

@rgommers what do you think? This isn't motivated by anything besides seeing CI breakage which we can work around. Then again, it does seem like a reasonable time to do this bump for a version of GCC that came out in 2019.

ngoldbaum:
### PR summary

Fixes #31828, see that issue for details.

#### AI Disclosure

No AI use.

### DIFF
--- a/.github/workflows/linux_simd.yml
+++ b/.github/workflows/linux_simd.yml
@@ -89,19 +89,11 @@ jobs:
       with:
         python-version: '3.12'
 
-    - name: Install GCC9/10
+    - name: Install GCC10
       run: |
         echo "deb http://archive.ubuntu.com/ubuntu focal main universe" | sudo tee /etc/apt/sources.list.d/focal.list
         sudo apt update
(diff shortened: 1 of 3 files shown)


rgommers (approved):
LGTM, thanks @ngoldbaum

PR #31845, opened 2026-07-02
PERF: Optimize reduction helper fast paths

KRRT7:
### PR summary

This PR optimizes a common Python-level fast path used by NumPy reductions.

The change makes `_wrapreduction` and `_wrapreduction_any_all` take their known optional reduction arguments explicitly and positionally instead of collecting them through internal `**kwargs`. This avoids creating and iterating over a generic keyword dict in these private helpers while preserving the existing forwarding behavior for explicit `keepdims`, `initial`, and `where` values.

The PR is intentionally scoped to this internal reduction-wrapper cleanup only. The separate `normalize_axis_tuple` fast path and the extra empty-`passkwargs` direct-call fast path were removed to keep the patch smaller and easier to review.

ASV results for the affected in-tree reduction benchmarks:

```bash
spin bench --compare upstream/main HEAD \
  -t bench_reduce.SmallReduction \
  -t bench_reduce.StatsReductions.time_min \
  -t bench_reduce.StatsReductions.time_max \
  -t bench_reduce.StatsReductions.time_prod
```

reported `PERFORMANCE INCREASED`.

Representative rows from the focused run:

- `bench_reduce.SmallReduction.time_sum(4)`: `802±40ns -> 530±10ns`, ratio `0.66`
- `bench_reduce.SmallReduction.time_any(100)`: `735±20ns -> 478±20ns`, ratio `0.65`
- `bench_reduce.SmallReduction2D.time_sum_axis_1`: `1.06±0.01μs -> 761±10ns`, ratio `0.72`
- `bench_reduce.StatsReductions.time_prod("int64")`: `816±9ns -> 519±8ns`, ratio `0.64`
- `bench_reduce.StatsReductions.time_min("uint64")`: `781±30ns -> 506±8ns`, ratio `0.65`
- `bench_reduce.StatsReductions.time_max("float64")`: `800±20ns -> 500±6ns`, ratio `0.62`

I am interested in NumPy performance work.

#### AI Disclosure

AI tools were used in preparing this PR.

Codex was used to run local tests and benchmarks, and help prepare draft PR text. A private tool was used to locate the issue, the code was reviewed by me before submission, and I am responsible for the final patch.

### DIFF
--- a/numpy/_core/fromnumeric.py
+++ b/numpy/_core/fromnumeric.py
@@ -64,6 +64,11 @@ def _wrapfunc(obj, method, *args, **kwds):
         return _wrapit(obj, method, *args, **kwds)
 
 
+# The positional-only signature and unrolled _NoValue checks (rather than
+# **kwargs with a dict comprehension) are deliberate: these helpers are on
+# the hot path of every reduction (sum, prod, min, max, any, all), and
+# avoiding the creation and iteration of a temporary kwargs dict measurably
+# reduces call overhead for small arrays.  See gh-31845.
 def _wrapreduction(obj, ufunc, method, axis, dtype, out,
                    keepdims=_NoValue, initial=_NoValue, where=_NoValue, /):
     passkwargs = {}
--- a/numpy/_core/fromnumeric.py
+++ b/numpy/_core/fromnumeric.py
@@ -64,8 +64,8 @@ def _wrapfunc(obj, method, *args, **kwds):
         return _wrapit(obj, method, *args, **kwds)
 
 
-def _wrapreduction(obj, ufunc, method, axis, dtype, out, *,
-                   keepdims=_NoValue, initial=_NoValue, where=_NoValue):
+def _wrapreduction(obj, ufunc, method, axis, dtype, out,
+                   keepdims=_NoValue, initial=_NoValue, where=_NoValue, /):
     passkwargs = {}
     if keepdims is not _NoValue:
         passkwargs["keepdims"] = keepdims
@@ -90,8 +90,8 @@ def _wrapreduction(obj, ufunc, method, axis, dtype, out, *,
     return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
 
 
-def _wrapreduction_any_all(obj, ufunc, method, axis, out, *,
-                           keepdims=_NoValue, where=_NoValue):
+def _wrapreduction_any_all(obj, ufunc, method, axis, out,
+                           keepdims=_NoValue, where=_NoValue, /):
     # Same as above function, but dtype is always bool (but never passed on)
     passkwargs = {}
     if keepdims is not _NoValue:
@@ -2497,8 +2497,7 @@ def sum(a, axis=None, dtype=None, out=None, keepdims=np._NoValue,
         )
 
     return _wrapreduction(
-        a, np.add, 'sum', axis, dtype, out,
-        keepdims=keepdims, initial=initial, where=where
+        a, np.add, 'sum', axis, dtype, out, keepdims, initial, where
     )
 
 
@@ -2611,7 +2610,7 @@ def any(a, axis=None, out=None, keepdims=np._NoValue, *, where=np._NoValue):
 
     """
     return _wrapreduction_any_all(a, np.logical_or, 'any', axis, out,
-                                  keepdims=keepdims, where=where)
+                                  keepdims, where)
 
 
 def _all_dispatcher(a, axis=None, out=None, keepdims=None, *,
@@ -2706,7 +2705,7 @@ def all(a, axis=None, out=None, keepdims=np._NoValue, *, where=np._NoValue):
 
     """
     return _wrapreduction_any_all(a, np.logical_and, 'all', axis, out,
-                                  keepdims=keepdims, where=where)
+                                  keepdims, where)
 
 
 def _cumulative_func(x, func, axis, dtype, out, include_initial):
@@ -3195,7 +3194,7 @@ def max(a, axis=None, out=None, keepdims=np._NoValue, initial=np._NoValue,
     5
     """
     return _wrapreduction(a, np.maximum, 'max', axis, None, out,
-                          keepdims=keepdims, initial=initial, where=where)
+                          keepdims, initial, where)
 
 
 @array_function_dispatch(_max_dispatcher)
@@ -3212,7 +3211,7 @@ def amax(a, axis=None, out=None, keepdims=np._NoValue, initial=np._NoValue,
     ndarray.max : equivalent method
     """
     return _wrapreduction(a, np.maximum, 'max', axis, None, out,
-                          keepdims=keepdims, initial=initial, where=where)
+                          keepdims, initial, where)
 
 
 def _min_dispatcher(a, axis=None, out=None, keepdims=None, initial=None,
@@ -3333,7 +3332,7 @@ def min(a, axis=None, out=None, keepdims=np._NoValue, initial=np._NoValue,
     6
     """
     return _wrapreduction(a, np.minimum, 'min', axis, None, out,
-                          keepdims=keepdims, initial=initial, where=where)
+                          keepdims, initial, where)
 
 
 @array_function_dispatch(_min_dispatcher)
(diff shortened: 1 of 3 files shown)


jorenham:

KRRT7:
@jorenham I've updated the body

ngoldbaum:
@ikrommyd you've been thinking about reductions a lot lately. What do you think about this? This sort of Python code change is trickier than it looks to review. From what you've seen of the reduction tests, do you think these code paths are well-tested enough to trust the test passing as a good enough signal that there aren't any behavior changes?

eendebakpt:
I have looked at the `_wrapreduction` before since it adds quite some overhead for small arrays, so this might be worthwhile.

However, there are 3 different optimizations in the PR: remove `**kwargs` from `_wrapreduction`/`_wrapreduction_any_all`, fast path in case `passkwargs` would turn out empty, and a fast path for `normalize_axis_tuple`.

 @KRRT7 Can you show benchmarks for the individual changes? In the first message I see "reported PERFORMANCE INCREASED", but not the individual results.

KRRT7:
@eendebakpt the tool I use, asv, by default only shows that, I'll post the individual results shortly.

ngoldbaum:
We have a `spin bench` command that wraps asv - it shows output for individual benchmarks.

KRRT7:
thanks for letting me know, I was using asv directly / manually.

ngoldbaum:
Please give https://numpy.org/devdocs/benchmarking.html a read if you haven't.

numpy/_core/fromnumeric.py
```diff
@@ -76,26 +86,44 @@ def _wrapreduction(obj, ufunc, method, axis, dtype, out, **kwargs):
...
                 return reduction(axis=axis, out=out, **passkwargs)
 
+    if passkwargs is None:
+        return ufunc.reduce(obj, axis, dtype, out)
     return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
 
 
-def _wrapreduction_any_all(obj, ufunc, method, axis, out, **kwargs):
+def _wrapreduction_any_all(obj, ufunc, method, axis, out, *,
+                           keepdims=_NoValue, where=_NoValue):
```
eendebakpt: We could make this positional only. That might be a bit faster

ikrommyd (commented):
Commenting since I was tagged by Nathan. From my recent experience, I think reductions are well-tested overall.

The first thing I noticed here is that now stray or unexpected kwargs will raise an error since  they are being turned into explicit keyword-only arguments instead of being passed through silently which I think is fine in general since this is supposed to be internal code. We just need to ensure there is no keyword argument here missing.

I don't think improving the performance of this is a bad idea in general however one should benchmark here and ensure that the performance improvement is worth it for a nominal size of arrays that reductions are usually called upon. @KRRT7 I'd personally make a `pyperf` benchmark that benchmarks exactly these code paths for arrays of nominal size.

My guess is that these code paths have near 100% coverage in the numpy test suite but the author can check with a tool or something too.

KRRT7:
for transparency, internally I have optimized quite a few paths significantly, and numpy itself end to end, I'm just trying to figure out how I can effectively open PRs and have them reviewed in a timely manner.

numpy/_core/fromnumeric.py
KRRT7: Done in fbfc622f66. I made both private helpers positional-only and changed the local call sites to pass the reduction sentinels positionally.

I re-ran the focused in-tree benchmark subset with `spin bench --compare upstream/main HEAD -t bench_reduce.SmallReduction -t bench_reduce.StatsReductions.time_min -t bench_reduce.StatsReductions.time_max -t bench_reduce.StatsReductions.time_prod`; it reports `PERFORMANCE INCREASED`.

KRRT7:
I reran the focused in-tree benchmark subset on the current PR head (`fbfc622f`) against `upstream/main` (`cd1cfb8e`):

```bash
spin bench --compare upstream/main HEAD \
  -t bench_reduce.SmallReduction \
  -t bench_reduce.StatsReductions.time_min \
  -t bench_reduce.StatsReductions.time_max \
  -t bench_reduce.StatsReductions.time_prod
```

ASV reported `PERFORMANCE INCREASED`.

Representative rows from the latest clean run:

- `bench_reduce.SmallReduction.time_any(100)`: `660±3ns -> 459±4ns`, ratio `0.70`
- `bench_reduce.SmallReduction.time_max(100)`: `754±5ns -> 483±5ns`, ratio `0.64`
- `bench_reduce.SmallReduction.time_sum(100)`: `763±7ns -> 493±8ns`, ratio `0.65`
- `bench_reduce.SmallReduction2D.time_sum_axis_1`: `1.03±0.01μs -> 722±6ns`, ratio `0.70`
- `bench_reduce.StatsReductions.time_prod('int64')`: `774±2ns -> 509±20ns`, ratio `0.66`
- `bench_reduce.StatsReductions.time_max('uint64')`: `768±9ns -> 499±2ns`, ratio `0.65`
- `bench_reduce.StatsReductions.time_min('bool_')`: `731±9ns -> 473±6ns`, ratio `0.65`

KRRT7:
I also reduced the scope of the PR after the earlier benchmark discussion.

The original version had three separate optimizations: the reduction wrapper keyword handling, an empty-`passkwargs` direct-call fast path, and a `normalize_axis_tuple` scalar-axis fast path. Those are all somewhat independent, so I removed the latter two to keep this PR focused on the smallest reduction-wrapper change with a clear benchmark signal.

The current diff only changes the private `_wrapreduction` / `_wrapreduction_any_all` helpers and their local call sites in `fromnumeric.py`. That should make the behavioral surface smaller and the patch easier to review.

eendebakpt (approved):
This is now I focused PR with a nice performance win. Inlining the `_wrapreduction` would help as well, but that is for a different PR.

There is also considerable overhead in the `array_function_dispatch`, I will open a separate PR for that.

numpy/_core/fromnumeric.py
```diff
@@ -83,10 +90,14 @@ def _wrapreduction(obj, ufunc, method, axis, dtype, out, **kwargs):
     return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
```
mhvk, line 95: It is annoying, but you can get a bit more of a speedup by doing
```
if passkwargs:
    return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
else:
    return ufunc.reduce(obj, axis, dtype, out)
```

Alternatively, and perhaps insane, but if we are really trying to optimize the most common cases (i.e., `keepdims` may be present, but `where` and `initial` likely not), then perhaps start with,
```
is_ndarray := type(obj) is np.ndarray
if is_ndarray and initial is _NoValue and where is _NoValue:
    # Fast path for two most common cases.
    if keepdims is _NoValue:
        return ufunc.reduce(obj, axis, dtype, out)
    else:
        return ufunc.reduce(obj, axis, dtype, out, keepdims=keepdims)

... [what you have, but use is_ndarray]
```

mhvk (approved):
Makes sense to me. One suggestion for even more speed-up (I *think*, to be tested!), but also OK to just go with what you have.

numpy/_core/fromnumeric.py
KRRT7: @mhvk I originally had more perf speedups in this PR but for now I've scoped it so that it get merged in quick. once it does I plan to open follow up PRs.

numpy/_core/fromnumeric.py
```diff
@@ -63,9 +64,15 @@ def _wrapfunc(obj, method, *args, **kwds):
         return _wrapit(obj, method, *args, **kwds)
 
 
-def _wrapreduction(obj, ufunc, method, axis, dtype, out, **kwargs):
-    passkwargs = {k: v for k, v in kwargs.items()
-                  if v is not np._NoValue}
+def _wrapreduction(obj, ufunc, method, axis, dtype, out,
```
ngoldbaum, line 72:
```suggestion
# The positional-only signature and unrolled _NoValue checks are deliberate: 
# these helpers are on the hot path of every reduction. See gh-31845.
def _wrapreduction(obj, ufunc, method, axis, dtype, out,
```

ngoldbaum (commented):
Can you add a release note about the performance improvement?

See doc/release/upcoming_changes/README.rst for instructions on how to do that.

ngoldbaum:
Thanks @KRRT7! I think you have a clearer idea of how we’d like to review this sort of thing. It also helps to focus on high-impact code paths like this one. Please feel free to keep working on these sorts of performance improvements.

I also wonder if there’s more overhead to remove by writing these helpers in C.

KRRT7:
> I also wonder if there’s more overhead to remove by writing these helpers in C.

most likely, but I'm not too familiar with C yet.

numpy/_core/fromnumeric.py
mhvk: OK, no problem. Fine to do it in a next step!

PR #31848, opened 2026-07-03
ENH: implement `np.unwrap` as a generalized ufunc

### LINKED ISSUE #9959: Make unwrap a ufunc
This would allow unwrap to work on subclasses of ndarrays. deg2rad, rad2deg (etc.) are all ufuncs, so it's not readily apparent why unwrap is not a ufunc.

ikrommyd:
### PR summary
Closes https://github.com/numpy/numpy/issues/9959

This PR implements `np.unwrap` as a generalized ufunc. We have to preserve the python wrapper to keep the API intact.
While making this PR I also found that unwrap had a docstring and typing issue. It's period argument can be an integer too and when both the input array and period are integers, the output dtype is also an integer. This was already shown in the docstring examples but not reflected in the types and typing stubs. I fixed both here.

The core of unwrap is now a generalized ufunc in C++ (signature ``(n),(),()->(n)``). The loop is implemented with discont argument always being a double as the function asks for it to be a python float.
As a result, unwrap now preserves the ndarray subclasses rather than always returning a base ndarray which is what the issue wanted at the time.

I chose not to add an object dtype C++ loop here as that is way too complicated for what it offers because I assume there is very little use of unwrap for object dtypes. We fall back to the old python implementation for object dtypes. I vibe coded an object dtype loop and it's 2x faster than the fallback on my machine but I think it's not worth having this complicated C++ code for this.

For integer/float dtypes, I see a nice 2x - 3x speed improvement on my machine from limited benchmarking and memory should be improved too as the fused loop avoids copies.

As unwrap had very limited testing, I added tests here that all pass on `main` except the `test_subclass_preserved` of course because until now unwrap did not preserve ndarray subclasses.

#### AI Disclosure
AI tools were used to help me translate the python implementation algorithm into the C++ loop. The documenting comments above the loop were also written by AI after I asked it (but I don't know if those are useful and can be removed).

### DIFF
--- a/doc/release/upcoming_changes/9959.improvement.rst
+++ b/doc/release/upcoming_changes/9959.improvement.rst
@@ -4,4 +4,6 @@ The core of ``numpy.unwrap`` is now a generalized ufunc in C++ (signature ``(n),
 covering the floating point and signed integer dtypes while performing the operation
 in a single pass without the intermediate arrays the previous Python implementation
 had to allocate. As a result, ``numpy.unwrap`` now preserves the ``ndarray`` subclasses
-rather than always returning a base ``ndarray``.
+rather than always returning a base ``ndarray`` in all cases except masked arrays.
+For masked arrays and their subclasses, the current unwrap algorithm is ill-defined
+so it unmasks the maskedarray and performs the operation on an ndarray instead.
--- a/doc/release/upcoming_changes/9959.improvement.rst
+++ b/doc/release/upcoming_changes/9959.improvement.rst
@@ -1,0 +1,8 @@
+``numpy.unwrap`` is now implement as a generalized ufunc and preserves array subclasses
+---------------------------------------------------------------------------------------
+The core of ``numpy.unwrap`` is now a generalized ufunc in C (signature ``(n),(),()->(n)``)
+covering floating point, signed integer, and object dtypes while performing the operation
+in a single pass without allocating intermediate arrays the previous Python implementation
+had to allocate. As a result, ``numpy.unwrap`` now dispatches through ``__array_ufunc__``
+and preserves the ``ndarray`` subclasses rather than always returning a base ``ndarray``.
+
--- a/doc/release/upcoming_changes/9959.improvement.rst
+++ b/doc/release/upcoming_changes/9959.improvement.rst
@@ -1,7 +1,7 @@
-``numpy.unwrap`` is now implement as a generalized ufunc and preserves array subclasses
----------------------------------------------------------------------------------------
+``numpy.unwrap`` is now implemented as a generalized ufunc and preserves array subclasses
+-----------------------------------------------------------------------------------------
 The core of ``numpy.unwrap`` is now a generalized ufunc in C++ (signature ``(n),(),()->(n)``)
-covering floating point, signed integer, and object dtypes while performing the operation
-in a single pass without allocating intermediate arrays the previous Python implementation
+covering the floating point and signed integer dtypes while performing the operation
+in a single pass without the intermediate arrays the previous Python implementation
 had to allocate. As a result, ``numpy.unwrap`` now preserves the ``ndarray`` subclasses
 rather than always returning a base ``ndarray``.
--- a/doc/release/upcoming_changes/9959.improvement.rst
+++ b/doc/release/upcoming_changes/9959.improvement.rst
@@ -1,6 +1,6 @@
 ``numpy.unwrap`` is now implement as a generalized ufunc and preserves array subclasses
 ---------------------------------------------------------------------------------------
-The core of ``numpy.unwrap`` is now a generalized ufunc in C (signature ``(n),(),()->(n)``)
+The core of ``numpy.unwrap`` is now a generalized ufunc in C++ (signature ``(n),(),()->(n)``)
 covering floating point, signed integer, and object dtypes while performing the operation
 in a single pass without allocating intermediate arrays the previous Python implementation
 had to allocate. As a result, ``numpy.unwrap`` now preserves the ``ndarray`` subclasses
--- a/doc/release/upcoming_changes/9959.improvement.rst
+++ b/doc/release/upcoming_changes/9959.improvement.rst
@@ -3,6 +3,5 @@
 The core of ``numpy.unwrap`` is now a generalized ufunc in C (signature ``(n),(),()->(n)``)
 covering floating point, signed integer, and object dtypes while performing the operation
 in a single pass without allocating intermediate arrays the previous Python implementation
-had to allocate. As a result, ``numpy.unwrap`` now dispatches through ``__array_ufunc__``
-and preserves the ``ndarray`` subclasses rather than always returning a base ``ndarray``.
-
+had to allocate. As a result, ``numpy.unwrap`` now preserves the ``ndarray`` subclasses
+rather than always returning a base ``ndarray``.
--- a/numpy/_core/code_generators/ufunc_docstrings.py
+++ b/numpy/_core/code_generators/ufunc_docstrings.py
@@ -3294,6 +3294,28 @@ add_newdoc('numpy._core.umath', '_ones_like',
 
     """)
 
+add_newdoc('numpy._core.umath', '_unwrap',
+    """
+    _unwrap(p, discont, period)
+
+    Generalized ufunc backing `numpy.unwrap`, with signature
+    ``(n),(),()->(n)``.
+
+    For each 1-D slice ``p`` along the core dimension, unwrap the values by
+    changing elements whose absolute difference from their predecessor exceeds
+    ``max(discont, period/2)`` to their `period`-complementary values, in a
+    single pass with no temporaries.  `discont` and `period` are scalars and
+    are expected to already share the common dtype of `p`; the public
+    `numpy.unwrap` wrapper fills their defaults, resolves that dtype and casts
+    them.  Loops are provided for the half, single, double and extended
+    floating-point dtypes, the signed integer dtypes and object.
+
+    See Also
+    --------
+    unwrap
+
+    """)
+
 add_newdoc('numpy._core.umath', 'power',
     """
     First array elements raised to powers from second array, element-wise.
--- a/numpy/_core/code_generators/ufunc_docstrings.py
+++ b/numpy/_core/code_generators/ufunc_docstrings.py
@@ -3303,12 +3303,10 @@ add_newdoc('numpy._core.umath', '_unwrap',
 
     For each 1-D slice ``p`` along the core dimension, unwrap the values by
     changing elements whose absolute difference from their predecessor exceeds
-    ``max(discont, period/2)`` to their `period`-complementary values, in a
-    single pass with no temporaries.  `discont` and `period` are scalars and
-    are expected to already share the common dtype of `p`; the public
-    `numpy.unwrap` wrapper fills their defaults, resolves that dtype and casts
-    them.  Loops are provided for the half, single, double and extended
-    floating-point dtypes, the signed integer dtypes and object.
+    ``max(discont, period/2)`` to their `period`-complementary values.
+    `discont` and `period` are scalars that are expected to already share the
+    common dtype of `p`. The public `numpy.unwrap` wrapper fills their
+    defaults, resolves that dtype and casts them.
 
     See Also
     --------
--- a/numpy/_core/code_generators/ufunc_docstrings.py
+++ b/numpy/_core/code_generators/ufunc_docstrings.py
@@ -3304,9 +3304,9 @@ add_newdoc('numpy._core.umath', '_unwrap',
     For each 1-D slice ``p`` along the core dimension, unwrap the values by
     changing elements whose absolute difference from their predecessor exceeds
     ``max(discont, period/2)`` to their `period`-complementary values.
-    `discont` and `period` are scalars that are expected to already share the
-    common dtype of `p`. The public `numpy.unwrap` wrapper fills their
-    defaults, resolves that dtype and casts them.
+    `period` is expected to already share `p`'s dtype, and `discont` its own
+    matching dtype. The public `numpy.unwrap` wrapper fills their defaults
+    and resolves and casts both before calling this.
 
     See Also
     --------
--- a/numpy/_core/src/umath/unwrap.cpp
+++ b/numpy/_core/src/umath/unwrap.cpp
@@ -86,19 +86,17 @@ floor_mod(T a, T b)
 }
 
 /*
- * One loop per real dtype. T is the storage/compute type; discont is read
- * as npy_longdouble only for the longdouble loop, npy_double otherwise.
- * npy_half has no native arithmetic, so it gets its own loop below.
+ * One loop per (T, D) pair. T is the storage/compute type; D is discont's
+ * type, either T itself (NEP 50 weak-scalar match) or a wider float type
+ * for explicitly-typed/cross-kind discont. npy_half has no native
+ * arithmetic, so it gets its own loop below.
  */
-template <typename T>
+template <typename T, typename D>
 static int
 unwrap_loop(PyArrayMethod_Context *NPY_UNUSED(context), char *const data[],
             npy_intp const dimensions[], npy_intp const strides[],
             NpyAuxData *NPY_UNUSED(auxdata))
 {
-    using D = std::conditional_t<std::is_same_v<T, npy_longdouble>,
-                                  npy_longdouble, npy_double>;
-
     npy_intp n_outer = dimensions[0];
     npy_intp n = dimensions[1];
     if (n == 0) {
@@ -162,6 +160,7 @@ unwrap_loop(PyArrayMethod_Context *NPY_UNUSED(context), char *const data[],
 static inline float h2f(npy_half h) { return npy_half_to_float(h); }
 static inline npy_half f2h(float f) { return npy_float_to_half(f); }
 
+template <typename D>
 static int
 unwrap_half_loop(PyArrayMethod_Context *NPY_UNUSED(context), char *const data[],
                  npy_intp const dimensions[], npy_intp const strides[],
@@ -181,7 +180,15 @@ unwrap_half_loop(PyArrayMethod_Context *NPY_UNUSED(context), char *const data[],
 
     for (npy_intp k = 0; k < n_outer; k++,
             p_o += s_p, disc_o += s_disc, per_o += s_per, out_o += s_out) {
-        npy_double discont = *(const npy_double *)disc_o;
+        /* npy_half discont is read through h2f, matching the NEP 50 weak-
+         * scalar case where discont is compared at half precision */
+        double discont;
+        if constexpr (std::is_same_v<D, npy_half>) {
+            discont = h2f(*(const npy_half *)disc_o);
+        }
+        else {
+            discont = *(const D *)disc_o;
+        }
         npy_half period = *(const npy_half *)per_o;
         npy_half interval_high = f2h(h2f(period) / 2.0f);
         npy_half interval_low = f2h(-h2f(interval_high));
@@ -218,7 +225,7 @@ unwrap_half_loop(PyArrayMethod_Context *NPY_UNUSED(context), char *const data[],
 
 static int
 add_unwrap_loop(PyObject *ufunc, int typenum,
-                PyArrayMethod_StridedLoop *loop, int discont_typenum = NPY_DOUBLE)
+                PyArrayMethod_StridedLoop *loop, int discont_typenum)
 {
     PyArray_DTypeMeta *dt = PyArray_DTypeFromTypeNum(typenum);
     if (dt == NULL) {
@@ -272,28 +279,37 @@ init_unwrap_ufunc(PyObject *umath)
     struct Loop {
         int typenum;
         PyArrayMethod_StridedLoop *loop;
+        int discont_typenum;
     };
 
+    /*
+     * float dtypes get two loops: discont at native precision (the NEP 50
+     * weak-scalar case, where the python wrapper matches discont to p's
+     * dtype) and at wider precision (explicitly-typed/cross-kind discont).
+     * Integer dtypes only get the wider loop -- NEP 50 native-int discont
+     * is not implemented, so the wrapper always widens for those.
+     */
     Loop loops[] = {
-        {NPY_HALF, unwrap_half_loop},
-        {NPY_FLOAT, unwrap_loop<npy_float>},
-        {NPY_DOUBLE, unwrap_loop<npy_double>},
-        {NPY_LONGDOUBLE, unwrap_loop<npy_longdouble>},
-        {NPY_BYTE, unwrap_loop<npy_byte>},
-        {NPY_SHORT, unwrap_loop<npy_short>},
-        {NPY_INT, unwrap_loop<npy_int>},
-        {NPY_LONG, unwrap_loop<npy_long>},
-        {NPY_LONGLONG, unwrap_loop<npy_longlong>},
+        {NPY_HALF, unwrap_half_loop<npy_half>, NPY_HALF},
+        {NPY_HALF, unwrap_half_loop<npy_double>, NPY_DOUBLE},
+        {NPY_FLOAT, unwrap_loop<npy_float, npy_float>, NPY_FLOAT},
+        {NPY_FLOAT, unwrap_loop<npy_float, npy_double>, NPY_DOUBLE},
+        {NPY_DOUBLE, unwrap_loop<npy_double, npy_double>, NPY_DOUBLE},
+        {NPY_LONGDOUBLE, unwrap_loop<npy_longdouble, npy_longdouble>,
+         NPY_LONGDOUBLE},
+        {NPY_BYTE, unwrap_loop<npy_byte, npy_double>, NPY_DOUBLE},
+        {NPY_SHORT, unwrap_loop<npy_short, npy_double>, NPY_DOUBLE},
+        {NPY_INT, unwrap_loop<npy_int, npy_double>, NPY_DOUBLE},
+        {NPY_LONG, unwrap_loop<npy_long, npy_double>, NPY_DOUBLE},
+        {NPY_LONGLONG, unwrap_loop<npy_longlong, npy_double>, NPY_DOUBLE},
         /* object dtype falls back to the python implementation instead */
         // {NPY_OBJECT, object_unwrap_loop},
     };
 
     int res = 0;
     for (const auto& entry : loops) {
-        int discont_typenum = entry.typenum == NPY_LONGDOUBLE
-                ? NPY_LONGDOUBLE : NPY_DOUBLE;
         if (add_unwrap_loop(ufunc, entry.typenum, entry.loop,
-                             discont_typenum) < 0) {
+                             entry.discont_typenum) < 0) {
             res = -1;
             break;
         }
--- a/numpy/_core/src/umath/unwrap.cpp
+++ b/numpy/_core/src/umath/unwrap.cpp
@@ -1,0 +1,293 @@
+/**
+ * This module provides the inner loops for the unwrap ufunc
+ */
+
+#define _UMATHMODULE
+#define _MULTIARRAYMODULE
+#define NPY_NO_DEPRECATED_API NPY_API_VERSION
+
+#define PY_SSIZE_T_CLEAN
+#include <Python.h>
+
+#include <type_traits>
+
+#include "numpy/halffloat.h"
+#include "numpy/ndarraytypes.h"
+#include "numpy/npy_common.h"
+#include "numpy/npy_math.h"
+#include "numpy/utils.h"
+#include "array_method.h"
+#include "dispatching.h"
+#include "dtypemeta.h"
+
+#include "unwrap.h"
+
+
+/*
+ * `unwrap` is a scan (each output depends on the running phase correction of
+ * the whole prefix along the core axis), so it cannot be a plain element-wise
+ * ufunc.  Each loop performs the diff / modular-wrap / cumulative-correction in
+ * a single pass over the core dimension, without allocating any temporaries.
+ *
+ * The operands are, in order, the values `p` (core `(n)`), the scalars
+ * `discont` and `period`, and the output (core `(n)`).  The Python wrapper in
+ * numpy/lib/_function_base_impl.py fills the `discont`/`period` defaults,
+ * resolves the common dtype and casts the scalars to it (rounding `discont` up
+ * for integer dtypes so that the `|dd| < discont` threshold keeps its meaning).
+ *
+ * gufunc strided-loop layout for signature (n),(),()->(n):
+ *   dimensions[0]      outer (broadcast) loop count
+ *   dimensions[1]      core length n
+ *   strides[0..3]      outer strides for p, discont, period, out
+ *   strides[4]         core stride of p
+ *   strides[5]         core stride of out
+ */
+
+/*
+ * floor-modulo matching numpy's `mod`, picked by overload on the npy floating
+ * types.  The long double overload is only defined when it is a distinct type
+ * from double (otherwise the double overload already covers npy_longdouble).
+ */
+static inline npy_float
+floor_mod(npy_float a, npy_float b)
+{
+    npy_float m;
+    npy_divmodf(a, b, &m);
+    return m;
+}
+static inline npy_double
+floor_mod(npy_double a, npy_double b)
+{
+    npy_double m;
+    npy_divmod(a, b, &m);
+    return m;
+}
+#if NPY_SIZEOF_LONGDOUBLE != NPY_SIZEOF_DOUBLE
+static inline npy_longdouble
+floor_mod(npy_longdouble a, npy_longdouble b)
+{
+    npy_longdouble m;
+    npy_divmodl(a, b, &m);
+    return m;
+}
+#endif
+
+/*
+ * One strided loop per real dtype.  `T` is both the storage and the compute
+ * type; the integer vs. floating-point parts of the algorithm (floor division
+ * for the interval, the even-period boundary, integer floor-modulo) select with
+ * `if constexpr` on `T`.  `discont` is only a threshold (never part of the wrap
+ * arithmetic), so it is always a plain double operand -- it keeps its value
+ * exactly and never drags the result dtype (e.g. integer input stays integer).
+ *
+ * `npy_half` has no native arithmetic, so it gets its own loop below that
+ * computes in float; every other real dtype uses this template.
+ */
+template <typename T>
+static int
+unwrap_loop(PyArrayMethod_Context *NPY_UNUSED(context), char *const data[],
+            npy_intp const dimensions[], npy_intp const strides[],
+            NpyAuxData *NPY_UNUSED(auxdata))
+{
+    npy_intp n_outer = dimensions[0];
+    npy_intp n = dimensions[1];
+    npy_intp s_p = strides[0], s_disc = strides[1];
+    npy_intp s_per = strides[2], s_out = strides[3];
+    npy_intp ip_step = strides[4];
+    npy_intp op_step = strides[5];
+
+    char *p_o = data[0], *disc_o = data[1], *per_o = data[2], *out_o = data[3];
+
+    for (npy_intp k = 0; k < n_outer; k++,
+            p_o += s_p, disc_o += s_disc, per_o += s_per, out_o += s_out) {
+        if (n <= 0) {
+            continue;
+        }
+        npy_double discont = *(const npy_double *)disc_o;
+        T period = *(const T *)per_o;
+        /* floor division for ints, exact half for floats */
+        T interval_high = period / 2;
+        T interval_low = -interval_high;
+        bool boundary_ambiguous;
+        if constexpr (std::is_integral_v<T>) {
+            /* only an even period has a representable +/-period/2 boundary */
+            boundary_ambiguous = (period % 2) == 0;
+        }
+        else {
+            boundary_ambiguous = true;
+        }
+
+        const char *ip = p_o;
+        char *op = out_o;
+        T prev = *(const T *)ip;
+        *(T *)op = prev;
+        T cum = 0;
+        for (npy_intp i = 1; i < n; i++) {
+            ip += ip_step;
+            op += op_step;
+            T cur = *(const T *)ip;
+            T dd = cur - prev;
+            T ddmod;
+            if constexpr (std::is_integral_v<T>) {
+                /* floor-modulo into [0, period); period > 0 so a single fixup */
+                T m = (T)((dd - interval_low) % period);
+                if (m < 0) {
+                    m += period;
+                }
+                ddmod = m + interval_low;
+            }
+            else {
+                ddmod = floor_mod(dd - interval_low, period) + interval_low;
+            }
+            if (boundary_ambiguous && ddmod == interval_low && dd > 0) {
+                ddmod = interval_high;
+            }
+            T corr = ddmod - dd;
+            T adiff = dd < 0 ? (T)-dd : dd;
+            if (adiff < discont) {
+                corr = 0;
+            }
+            cum += corr;
+            *(T *)op = (T)(cur + cum);
+            prev = cur;
+        }
+    }
+    return 0;
+}
+
+/* Half is stored as npy_half but has no native arithmetic, so it loads/stores
+ * npy_half and does the whole scan in float. */
+static int
+unwrap_half_loop(PyArrayMethod_Context *NPY_UNUSED(context), char *const data[],
+                 npy_intp const dimensions[], npy_intp const strides[],
+                 NpyAuxData *NPY_UNUSED(auxdata))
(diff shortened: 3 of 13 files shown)


numpy/lib/_function_base_impl.pyi
jorenham, line 1: Technically speaking, it should now be `unwrap: np.ufunc`, or rather an appropriate hand-crafted specialization of `np.ufunc` like the ones in https://github.com/numpy/numpy/blob/2450a2c28ca49377a47e451e545be76067ef1cd1/numpy/_core/umath.pyi.

But for now this'll be fine. I'll probably circle back to this as follow-up.

ikrommyd: Thanks, I'm not particularly good with python typing (not a heavy user of it) so just suggest away whatever you think is right :)

numpy/lib/_function_base_impl.py
```diff
@@ -1833,34 +1837,37 @@ def unwrap(p, discont=None, axis=-1, *, period=2 * pi):
...
-        # `ddmod[mask] == sign(dd[mask])*period/2`.
-        _nx.copyto(ddmod, interval_high,
-                   where=(ddmod == interval_low) & (dd > 0))
-    ph_correct = ddmod - dd
-    _nx.copyto(ph_correct, 0, where=abs(dd) < discont)
-    up = array(p, copy=True, dtype=dtype)
-    up[slice1] = p[slice1] + ph_correct.cumsum(axis)
-    return up
+    dtype = np.result_type(p, period)
+    if p.dtype == object or dtype.isbuiltin != 1:
```
ikrommyd: I'm not sure what I should do with user dtypes here. I tried it with quad dtype and it works now and it works on main too (which is the same code effectively) because I think the quad dtype registers loops for all the functions the python implementation calls. What does numpy have to gurantee here regarding user dtypes? `dtype.isbuiltin != 1` was my "hack" basically.

numpy/lib/_function_base_impl.pyi
jorenham: it's fine as-is :)

numpy/lib/_function_base_impl.py
ngoldbaum: I think it would be nicer to attempt to call the ufunc and only fall back if you get a UFuncLoopNotFoundError. The way you've written it right now I think quaddtype can never use the ufunc, even if it adds a loop.

ikrommyd: Ah yeah that makes sense

ikrommyd: And can `np.result_type(p, period)` error when user dtypes are involved?

ngoldbaum: I'm not sure offhand, I'd have to trace through the code.

ikrommyd:
FYI this is what one gets with the current unwrap testing on main. Some error messages are different here versus main and naturally on main subclasses are not returned. But I'd argue for the cases that unwrap shouldn't work anyways, the error messages are a bit better now instead of overflow errors and `TypeError: ufunc 'remainder' not supported`
```
FAILED numpy/lib/tests/test_function_base.py::TestUnwrap::test_subclass_preserved[float32] - AssertionError: assert False
FAILED numpy/lib/tests/test_function_base.py::TestUnwrap::test_subclass_preserved[float64] - AssertionError: assert False
FAILED numpy/lib/tests/test_function_base.py::TestUnwrap::test_subclass_preserved[int8] - AssertionError: assert False
FAILED numpy/lib/tests/test_function_base.py::TestUnwrap::test_subclass_preserved[int32] - AssertionError: assert False
FAILED numpy/lib/tests/test_function_base.py::TestUnwrap::test_subclass_preserved[int64] - AssertionError: assert False
FAILED numpy/lib/tests/test_function_base.py::TestUnwrap::test_subclass_preserved[bool] - AssertionError: assert False
FAILED numpy/lib/tests/test_function_base.py::TestUnwrap::test_subclass_preserved[object] - AssertionError: assert False
FAILED numpy/lib/tests/test_function_base.py::TestUnwrap::test_complex_raises - TypeError: ufunc 'remainder' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting r...
FAILED numpy/lib/tests/test_function_base.py::TestUnwrap::test_unsigned_integer_period_raises[uint8] - OverflowError: Python integer -4 out of bounds for uint8
FAILED numpy/lib/tests/test_function_base.py::TestUnwrap::test_unsigned_integer_period_raises[uint32] - OverflowError: Python integer -4 out of bounds for uint32
FAILED numpy/lib/tests/test_function_base.py::TestUnwrap::test_unsigned_integer_period_raises[uint64] - OverflowError: Python integer -4 out of bounds for uint64
```

numpy/lib/_function_base_impl.py
ikrommyd: It is now
```py
    try:
        return _unwrap(...) # try gufunc
    except np._core._exceptions._UFuncNoLoopError:
        if dtype.isbuiltin == 1 and p.dtype != object:
            raise # raise in the case that another error happened
        # continue with fallback python implementation
```
Let me know if you have better ideas.

ikrommyd:
IMO this should be ready for a review pass

numpy/_core/src/umath/unwrap.cpp
```diff
@@ -0,0 +1,299 @@
...
+    npy_divmodl(a, b, &m);
+    return m;
+}
+#endif
+/* integer floor-modulo into [0, b); the caller guarantees b > 0. */
+template <typename T>
+static inline std::enable_if_t<std::is_integral_v<T>, T>
+floor_mod(T a, T b)
+{
+    T m = a % b;
```
MaanasArora: My agent noticed this needs a zero-division guard for `period == 0` otherwise this will segfault due to zero division. I believe the older code would raise a `RuntimeWarning` and return 0 since it used `np.divmod`, so might make sense to preserve that behavior.

We should add a test for this too!

numpy/_core/code_generators/ufunc_docstrings.py
```diff
@@ -3294,6 +3294,28 @@ def add_newdoc(place, name, doc):
...
+    ``(n),(),()->(n)``.
+
+    For each 1-D slice ``p`` along the core dimension, unwrap the values by
+    changing elements whose absolute difference from their predecessor exceeds
+    ``max(discont, period/2)`` to their `period`-complementary values, in a
+    single pass with no temporaries.  `discont` and `period` are scalars and
+    are expected to already share the common dtype of `p`; the public
+    `numpy.unwrap` wrapper fills their defaults, resolves that dtype and casts
+    them.  Loops are provided for the half, single, double and extended
+    floating-point dtypes, the signed integer dtypes and object.
```
MaanasArora: I don't think there is an `object` loop?

doc/release/upcoming_changes/9959.improvement.rst
```diff
@@ -0,0 +1,7 @@
+``numpy.unwrap`` is now implement as a generalized ufunc and preserves array subclasses
+---------------------------------------------------------------------------------------
+The core of ``numpy.unwrap`` is now a generalized ufunc in C++ (signature ``(n),(),()->(n)``)
+covering floating point, signed integer, and object dtypes while performing the operation
```
MaanasArora: Same as in the docstring (no object loop)!

numpy/_core/code_generators/ufunc_docstrings.py
```diff
@@ -3294,6 +3294,28 @@ def add_newdoc(place, name, doc):
...
+    """
+    _unwrap(p, discont, period)
+
+    Generalized ufunc backing `numpy.unwrap`, with signature
+    ``(n),(),()->(n)``.
+
+    For each 1-D slice ``p`` along the core dimension, unwrap the values by
+    changing elements whose absolute difference from their predecessor exceeds
+    ``max(discont, period/2)`` to their `period`-complementary values, in a
+    single pass with no temporaries.  `discont` and `period` are scalars and
```
MaanasArora: Somewhat a nitpick, but this seems to be exposing some implementation details (single pass, no temporaries) I'm not sure we want in `umath`. It is also a bit verbose, should we constrain this to only the functional aspects?

MaanasArora (commented):
I didn't look very deeply, but will do so very soon - this looks nice, thanks! Just one bug, and mostly doc comments otherwise for now.

MaanasArora (commented):
I didn't look very deeply, but will do so very soon - this looks nice, thanks! Just one bug, and mostly doc comments otherwise for now.

numpy/_core/code_generators/ufunc_docstrings.py
ikrommyd: yeah forgotten from when I had one

doc/release/upcoming_changes/9959.improvement.rst
ikrommyd: yeah forgotten from when I had one

numpy/_core/code_generators/ufunc_docstrings.py
ikrommyd: I can reduce this. I still wanted to add a docstring similar to the private _ones_like right above it but yeah it doesn't need to be that technical indeed.

numpy/_core/src/umath/unwrap.cpp
ikrommyd: Yeah that's right, will fix. On main it does this:
```
In [2]: np.unwrap([1,2,3], period=0)
/home/iason/micromamba/envs/py3.14/lib/python3.14/site-packages/numpy/lib/_function_base_impl.py:1841: RuntimeWarning: divide by zero encountered in remainder
  ddmod = mod(dd - interval_low, period) + interval_low
Out[2]: array([1, 1, 1])
```

numpy/_core/src/umath/unwrap.cpp
```diff
@@ -0,0 +1,306 @@
...
+ */
+template <typename T>
+static inline std::enable_if_t<std::is_integral_v<T>, T>
+floor_mod(T a, T b)
+{
+    if (b == 0) {
+        npy_set_floatstatus_divbyzero();
+        return 0;
+    }
+    T m = a % b;
```
ngoldbaum: Another bug in this function:

```
p = np.array([0, 3, 6, 1, 4], dtype=np.int64)
np.unwrap(p, period=-8)
```

Old result:

```
[0, 3, 6, 9, 12]
```

In this PR:

```
[0, -5, -10, -15, -20]
```

ngoldbaum: Another issue is that modulo arithmetic means dealing with UB: https://eel.is/c++draft/expr.mul#4

Here's a little example for how to do this correctly: https://cmu-sei.github.io/secure-coding-standards/sei-cert-c-coding-standard/rules/integers-int/int33-c/#remainder

Maybe you want `npy_remainder`?

numpy/_core/src/umath/unwrap.cpp
```diff
@@ -0,0 +1,306 @@
...
+        const char *ip = p_o;
+        char *op = out_o;
+        T prev = *(const T *)ip;
+        *(T *)op = prev;
+        T cum = 0;
+        for (npy_intp i = 1; i < n; i++) {
+            ip += ip_step;
+            op += op_step;
+            T cur = *(const T *)ip;
+            T dd = cur - prev;
```
ngoldbaum, line 138: This also needs to use a checked subtraction function instead of raw subtraction; signed integer overflow is UB.

numpy/_core/src/umath/unwrap.cpp
```diff
@@ -0,0 +1,306 @@
...
+            /* wrap the raw diff into [interval_low, interval_low + period);
+             * floor_mod dispatches to the integer or floating overload on T. */
+            T ddmod = (T)(floor_mod((T)(dd - interval_low), period) + interval_low);
+            if (boundary_ambiguous && ddmod == interval_low && dd > 0) {
+                ddmod = interval_high;
+            }
+            T corr = ddmod - dd;
+            T adiff = dd < 0 ? (T)-dd : dd;
```
ngoldbaum: the negation here is possibly problematic for similar reasons to comments above. Either avoid using negation here or do a check for invalid cases first.

numpy/_core/src/umath/unwrap.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+        const char *ip = p_o;
+        char *op = out_o;
+        float prev = npy_half_to_float(*(const npy_half *)ip);
+        *(npy_half *)op = npy_float_to_half(prev);
+        float cum = 0;
+        for (npy_intp i = 1; i < n; i++) {
+            ip += ip_step;
+            op += op_step;
+            float cur = npy_half_to_float(*(const npy_half *)ip);
```
ngoldbaum: Why not do the entire operation in native float16?

Try, for example, this code, which shows that this loop produces slightly different results because of the different rounding behavior of doing the calculation in a float32 intermediate.

```
import numpy as np

p = np.array([-64.0, -50.1875], dtype=np.float16)
period = np.float16(7.3)

result = np.unwrap(p, period=period)

# Convert for unambiguous display; float16 formatting hides the exact value.
print(result.astype(np.float64))
print([hex(x) for x in result.view(np.uint16)])
```

before

```
[-64.   -64.75]
['0xd400', '0xd40c']
```

after

```
[-64.     -64.8125]
['0xd400', '0xd40d']
```

numpy/lib/_function_base_impl.py
```diff
@@ -1833,34 +1864,21 @@ def unwrap(p, discont=None, axis=-1, *, period=2 * pi):
...
-                   where=(ddmod == interval_low) & (dd > 0))
-    ph_correct = ddmod - dd
-    _nx.copyto(ph_correct, 0, where=abs(dd) < discont)
-    up = array(p, copy=True, dtype=dtype)
-    up[slice1] = p[slice1] + ph_correct.cumsum(axis)
-    return up
+    dtype = np.result_type(p, period)
+    try:
+        return _unwrap(p, discont, period,
+                       signature=(type(dtype), np.float64, type(dtype), type(dtype)),
```
ngoldbaum: hardcoding float64 here loses precision for longdouble on platforms where it's not float64

numpy/lib/_function_base_impl.py
```diff
@@ -1833,34 +1864,21 @@ def unwrap(p, discont=None, axis=-1, *, period=2 * pi):
...
+                       axis=axis)
+    except np._core._exceptions._UFuncNoLoopError:
+        # We could implement a gufunc loop for object dtype
+        # but it's not worth the burden of all the extra c++ code.
+        # We let object dtype go through the fallback.
+        if dtype.isbuiltin == 1 and p.dtype != object:
```
ngoldbaum: argh, my AI agent spotted an issue with this:

> This decides whether to fall back using the promoted result dtype rather than the input dtype. A user DType can promote with period to a built-in dtype but have no _unwrap loop; in that case this re-raises even though the old composite implementation may work through the DType’s registered component ufunc loops.
>Could the fallback decision be based on the exact loop signatures implemented here rather than dtype.isbuiltin? For example, re-raise only when the requested signature should have matched one of _unwrap’s explicitly registered built-in loops, and otherwise attempt the compatibility fallback.

numpy/lib/tests/test_function_base.py
```diff
@@ -2261,6 +2261,102 @@ def test_period(self):
...
+    def test_int_float_period_promotes(self, dtype):
+        # a float period turns an integer array into a float64 result
+        out = unwrap(np.array([0, 1, 2, -1, 0], dtype=dtype), period=2 * np.pi)
+        assert out.dtype == np.float64
+
+    @pytest.mark.parametrize("dtype", [np.float32, np.float64, np.int8, np.int32,
+                                       np.int64, np.bool_, object])
+    def test_subclass_preserved(self, dtype):
+        class MyArray(np.ndarray):
+            pass
```
ngoldbaum, line 2308: you should also add a test with a nontrivial `__array_ufunc__`. I bet there are examples elsewhere in the codebase. Maybe it happens automatically in the `__array_ufunc__` tests? In particular you want to see what happens with an implementation that returns `NotImplemented` for unimplemented ufuncs and one that actually calls an implementation.

ngoldbaum: While we're here and I know it sucks but we should probably think about MaskedArray too?

numpy/lib/tests/test_function_base.py
```diff
@@ -2261,6 +2261,102 @@ def test_period(self):
...
+    )
+    def test_period_zero(self, dtype, expected):
+        p = np.array([1, 2, 3], dtype=dtype)
+        if IS_WASM:
+            with np.errstate(divide="ignore", invalid="ignore"):
+                result = unwrap(p, period=0)
+        else:
+            with pytest.warns(RuntimeWarning):
+                result = unwrap(p, period=0)
+        assert_array_equal(result, np.array(expected, dtype=result.dtype))
```
ngoldbaum: I'm not saying you should actually do all of this, but certainly some of the edge case testing it points out is worth thinking about:

> The current tests mainly cover expected examples but miss several compatibility-sensitive cases:
> * positive and negative integer periods;
> * signed dtype extrema and INT_MIN;
> * period=-1 and period=0;
> * randomized float16 inputs;
> * long-double discontinuities;
> * non-native byte order;
> * active ndarray subclasses and custom DTypes.
> Randomized comparisons against the reference would give much stronger confidence that the fused loop is an optimization of the existing operation rather than a subtly different algorithm.

numpy/lib/tests/test_function_base.py
```diff
@@ -2261,6 +2261,102 @@ def test_period(self):
...
+        p = np.arange(5, dtype=np.complex128).astype(object)
+        with pytest.raises(TypeError, match="unsupported operand type"):
+            unwrap(p)
+
+    @pytest.mark.parametrize("dtype", [np.uint8, np.uint32, np.uint64])
+    def test_unsigned_integer_period_raises(self, dtype):
+        # unsigned + integer period needs the negative bound ``-period // 2``,
+        # which is out of range for the unsigned dtype
+        p = np.array([0, 3, 6, 1, 4], dtype=dtype)
+        with pytest.raises(np._core._exceptions._UFuncNoLoopError):
```
ngoldbaum, line 2362: On NumPy main:

```
p = np.array([0, 3, 6, 1, 4], dtype=np.uint8)
np.unwrap(p, period=8)
File ~/.pyenv/versions/3.14.5/lib/python3.14/site-packages/numpy/lib/_function_base_impl.py:1852, in unwrap(p, discont, axis, period)
   1850     boundary_ambiguous = True
   1851 interval_low = -interval_high
-> 1852 ddmod = mod(dd - interval_low, period) + interval_low
   1853 if boundary_ambiguous:
   1854     # for `mask = (abs(dd) == period/2)`, the above line made
   1855     # `ddmod[mask] == -period/2`. correct these such that
   1856     # `ddmod[mask] == sign(dd[mask])*period/2`.
   1857     _nx.copyto(ddmod, interval_high,
   1858                where=(ddmod == interval_low) & (dd > 0))

OverflowError: Python integer -4 out of bounds for uint8
```

In your PR:

```
File ~/Documents/numpy/build-install/usr/lib/python3.14/site-packages/numpy/lib/_function_base_impl.py:1872, in unwrap(p, discont, axis, period)
   1870 dtype = np.result_type(p, period)
   1871 try:
-> 1872     return _unwrap(p, discont, period,
   1873                    signature=(type(dtype), np.float64, type(dtype), type(dtype)),
   1874                    axis=axis)
   1875 except np._core._exceptions._UFuncNoLoopError:
   1876     # We could implement a gufunc loop for object dtype
   1877     # but it's not worth the burden of all the extra c++ code.
   1878     # We let object dtype go through the fallback.
   1879     if dtype.isbuiltin == 1 and p.dtype != object:

UFuncTypeError: ufunc '_unwrap' did not contain a loop with signature matching types (<class 'numpy.dtypes.UInt8DType'>, <class 'numpy.dtypes.Float64DType'>, <class 'numpy.dtypes.UInt8DType'>) -> <class 'numpy.dtypes.UInt8DType'>
```

Which is fair but it at least deserves a release note. But ideally a promoter would fire and you'd get the OverflowError via another pass.

ngoldbaum (changes requested):
OK, first off, I'm very happy with the performance improvements. As I'm sure you know, I'm very happy with doing this sort of thing to improve performance of NumPy in ways that are natural in the codebase.

Now given that, I think we should probably check in with the numpy-discussion mailing list before actually merging the PR. While this is mostly not user-visible (except via the performance and memory improvements) this does add a new ufunc to numpy, which downstream libraries are going to have to deal with. @mhvk can check in with an astropy hat on probably but there are lots of other smaller projects that this will impact as well. IMO it's worth it but it's worth pinging the mailing list to check in first.

I also did a review pass and noticed some issues that'll need to be addressed. I may spot more stuff on a further pass but hopefully we'll converge quickly after this first pass is addressed.

ikrommyd:
Thanks for the review! @ngoldbaum I'll address everything in the coming week as best as I can.
In general I was mostly trying to preserve behavior with main as much as I could but couldn't fully preserve the exceptions indeed. I guess because unwrap didn't really care too much about errors from how I read the imlementation and was like "yeah whatever the other functions raise". You also pointed out some real bugs in the ufunc code which I will fix obviously.

numpy/lib/tests/test_function_base.py
ikrommyd: I actually did like the newer error much better than the internal overflow error in `mod` which is why I never went for a promoter. The PR error tells the user more than this shouldn't work while the overflow error has a higher chance of being miss-interpreted as a bug. One can also wrap it in an even nicer error with a `raise ... from err`.

ikrommyd:
I have implemented most of the review comments btw I just haven't officially responsed because I am unsure about some things (like masked arrays). Unsure meaning I have doubts as to what the right thing to do is.

numpy/_core/src/umath/unwrap.cpp
```diff
@@ -0,0 +1,304 @@
...
+    npy_intp s_p = strides[0], s_disc = strides[1];
+    npy_intp s_per = strides[2], s_out = strides[3];
+    npy_intp ip_step = strides[4];
+    npy_intp op_step = strides[5];
+
+    char *p_o = data[0], *disc_o = data[1], *per_o = data[2], *out_o = data[3];
+
+    for (npy_intp k = 0; k < n_outer; k++,
+            p_o += s_p, disc_o += s_disc, per_o += s_per, out_o += s_out) {
+        if (n <= 0) {
```
mhvk: Can we ever get called with `n==0`? If so, no need for this to be in the loop; might as well `return 0` right on top for it.

numpy/lib/tests/test_function_base.py
mhvk: For the subclass, e.g., see `PhysicalQuantity2` in `test_functools` which I just made look more like astropy's `Quantity` in #31997. Though looking further down I think you've got it covered...

For MaskArray, I think you fall back now? That seems correct!

numpy/lib/_function_base_impl.py
```diff
@@ -1833,34 +1864,29 @@ def unwrap(p, discont=None, axis=-1, *, period=2 * pi):
...
-                   where=(ddmod == interval_low) & (dd > 0))
-    ph_correct = ddmod - dd
-    _nx.copyto(ph_correct, 0, where=abs(dd) < discont)
-    up = array(p, copy=True, dtype=dtype)
-    up[slice1] = p[slice1] + ph_correct.cumsum(axis)
-    return up
+    if np.ma.isMaskedArray(p):
+        # the gufunc's C loop has no concept of masking and would compute
+        # straight through masked values; the fallback's ops are mask-aware
+        return _unwrap_fallback(p, discont, period, axis)
```
mhvk: Ideally, the code would not have to do isinstance checks and this is done in `MaskedArray.__array_ufunc__`. But lacking that, no choice really...

numpy/_core/src/umath/unwrap.cpp
```diff
@@ -0,0 +1,304 @@
...
+            continue;
+        }
+        D discont = *(const D *)disc_o;
```
mhvk, line 113: This confuses me, why is `discont` not just the type (at least for floating point ones?)

mhvk (commented):
A bit of a fly-by review: looks very nice, no real comments any more, more little queries

ngoldbaum:
> I have implemented most of the review comments btw I just haven't officially responsed because I am unsure about some things (like masked arrays). Unsure meaning I have doubts as to what the right thing to do is.

IMO the behavior you have here is fine, it follows the existing behavior for `cumsum` and `cumprod`, for example.

The new masked array behavior probably deserves a mention in the release notes, with a `compatibility` tag. IMO it's OK to do a change like that since the old behavior was clearly not optimal for masked array users.

ikrommyd:
My problem with masked arrays is that neither approach is really correct.

If we go through `asarray` like main branch does (which is the same as calling the gufunc loop on them), the mask just gets dropped and `unwrap` computes over whatever garbage happens to be sitting in the masked slots as if it were real data. Obviously wrong.

If we go through the fallback instead, the mask survives, but `diff`/`cumsum` only know how to propagate masks positionally and they have no idea that `unwrap` needs the delta to the nearest *valid* sample. So a masked gap just gets treated as "no correction needed here," which can silently miss a real wrap and leave everything after the gap offset by a full period.

Ideally masked arrays should fully error here or have a proper implementation in the masked array module but we can't really raise an error now if it used to wrong just incorrectly right?

mhvk:
@ikrommyd - I think `MaskedArray` needs its own implementation, which, as you note, is tricky. But for *this* PR I think the choice you make, of just keeping the existing behaviour, is the proper one. Though probably good to add a TODO alongside the test that really one should do better!

ikrommyd:
@mhvk clarification: do you suggest keeping the behavior the same as main here which would drop the masked array completely or letting masked arrays go through the fallback which will now "respect" the mask but the result will be wrong too? I'd lean on the former to not break behavior but that breaks what the PR allows which is preserving ndarray subclasses.

mhvk:
@ikrommyd - I suggest whatever causes no change from the status quo for `MaskedArray` -- generally, regular numpy functions outside of `np.ma` are not safe for `MaskedArray`, so that is nothing new. (Of course, if we can make it safe, we should do that, but it does not have to be in this PR. Changing the way it breaks would not seem to be a good idea, though!)

mhvk:
p.s. Conversely, if you have a good idea how to unwrap MaskedArray, we should of course do that, but best in a new PR!

ikrommyd:
So what I can do for masked array which feels the most normal for me to do here is to actually call the gufunc loop on the array removing the mask to preserve the result exactly like main but rewrap it in a masked array with no masked elements to preserve the ndarray subclass. This feels the most normal to me and fits what this PR will add in the changelog which is that "unwrap now preserves ndarray subclasses".

numpy/_core/src/umath/unwrap.cpp
ikrommyd: The idea is to keep `discont` a double (or long double) and not let it be cast down to the dtype of the incoming array. We lose precision if we do so.

numpy/lib/_function_base_impl.py
ikrommyd: yeah I've left a todo to properly implement unwrap in the masked array module.

numpy/_core/src/umath/unwrap.cpp
ikrommyd: I pulled it out as an an early exit indeed

ikrommyd:
Damn it...I accidentally clicked on copilot :( Apologies for that. I didn't even realize I clicked it. Probably while scrolling.

ikrommyd:
@ngoldbaum So I think I like the current behavior for masked arrays the most. Regarding signed integers, I like the no loop error a lot more than promoter + overflow but if you insist I can add a promoter. I'd much rather keep the no loop error and document it in the release notes though. Finally for the UBs that you commented on, I checked usual ufunc loops like absolute, add, subtract etc and they do not seem to care at all about these UBs like signed integer subtraction so why should we add checks in this loop? (checks that also will hurt performance a tiny bit).

mhvk:
@ikrommyd - I don't quite agree we should behave as if the mask doesn't exist and then return `MaskedArray` -- this would signal that the mask has been taken into account. I think we should either properly support `MaskedArray` or not change behaviour at all.

numpy/_core/src/umath/unwrap.cpp
mhvk: But why do we care about precision loss for float? I can see some advantage for integers, since it allows being in between two, but for the floating types it seems weird that the loops are not simply `fff->f`, etc.

ikrommyd:
Yeah I understand your point of view Marten. I guess I did not like that the release note then is "unwrap preserves ndarray subclasses....oh but not masked array" 🤣. But yeah I see the point that when you return a masked array type, the user expects the operation to be properly done in masked array space.

ikrommyd:
@mhvk what are your thoughts on https://github.com/numpy/numpy/pull/31848/changes/9a45e8658a64c458cafed3ae1b1d2c30f98bb32e regarding the `discont` parameter? I asked claude to make it NEP-50 compartible and gave it the NEP.

ikrommyd:
Or I have one more idea. We get rid of the whole python promotion and passing `signature` in the ufunc and do the promotion with a promoter in c++. @ngoldbaum and @mhvk what do you think?

MaanasArora:
I agree with Marten it doesn't seem to make much sense to keep e.g. `discont` double if the input is `float32`; I don't think we're gaining any real precision as this is an elementwise comparison, and there is no error accumulation / propagation. I would simply cast it to the input array dtype (you could make an exception for integers to preserve boundary precision, though there could be better ways to handle the boundary). NEP 50 applies to promoting Python scalars especially when the dtype is "real" (ends up in a result for example) but this seems to be on the other end of a specific internal mechanic. In any case, I'm pretty sure this new loop registration is complicated and not right just to keep `discont`'s type.

numpy/lib/_function_base_impl.py
```diff
@@ -1833,34 +1864,36 @@ def unwrap(p, discont=None, axis=-1, *, period=2 * pi):
...
-        # `ddmod[mask] == -period/2`. correct these such that
-        # `ddmod[mask] == sign(dd[mask])*period/2`.
-        _nx.copyto(ddmod, interval_high,
-                   where=(ddmod == interval_low) & (dd > 0))
-    ph_correct = ddmod - dd
-    _nx.copyto(ph_correct, 0, where=abs(dd) < discont)
-    up = array(p, copy=True, dtype=dtype)
-    up[slice1] = p[slice1] + ph_correct.cumsum(axis)
-    return up
+    dtype = np.result_type(p, period)
```
ngoldbaum: Should this be inside the try/except so the fallback is triggered if this errors?

numpy/_core/src/npysort/x86-simd-sort
ngoldbaum, line 1: this shouldn't be updated in this PR

numpy/lib/tests/test_function_base.py
```diff
@@ -2261,6 +2285,255 @@ def test_period(self):
...
+        # judged smaller than discont and does get corrected
+        p = np.array([0, 3], dtype=np.float16)
+        period = 4
+        discont = 3 + 1e-4
+        assert_array_equal(unwrap(p, discont=discont, period=period), [0, -1])
+
+    def test_masked_array(self):
+        # the gufunc's C loop can't propagate masking through the scan the
+        # way the fallback's mask-aware ops do, so this must match the
+        # fallback exactly, not just "look reasonable"
```
ngoldbaum: this is stale, right? can you do a pass over the PR for stale comments?

ngoldbaum (commented):
I spotted a couple minor issues with a claude review pass. I think C promoters are generally what you want instead of implementing promotion logic in the Python wrapper, but I'll defer to Maanas and Sebastian's opinions.

numpy/_core/src/umath/unwrap.cpp
```diff
@@ -0,0 +1,320 @@
...
+    PyArrayMethod_Spec spec = {};
+    spec.name = "unwrap_loop";
+    spec.nin = 3;
+    spec.nout = 1;
+    spec.casting = NPY_NO_CASTING;
+    spec.flags = (NPY_ARRAYMETHOD_FLAGS)0;
+    spec.dtypes = dtypes;
+    spec.slots = slots;
+
+    if (PyUFunc_AddLoopFromSpec_int(ufunc, &spec, 1)) {
```
MaanasArora: nit: you could do `ret = PyUFunc_AddLoopFromSpec_int(...)` and return that, then the decrefs are merged.

numpy/_core/src/umath/unwrap.cpp
```diff
@@ -0,0 +1,320 @@
...
+        /* object dtype falls back to the python implementation instead */
+        // {NPY_OBJECT, object_unwrap_loop},
+    };
+
+    int res = 0;
+    for (const auto& entry : loops) {
+        if (add_unwrap_loop(ufunc, entry.typenum, entry.loop,
+                             entry.discont_typenum) < 0) {
+            res = -1;
+            break;
```
MaanasArora: If you want to break immediately (nice), very silly nit, but it could be nice to just remove `res` and decref manually, return -1. The break "breaks" the code a bit :) and there are only two cases really

MaanasArora (commented):
Just two fly-by nits... the `discont` handling still needs to be resolved of course.

For the promotion logic, I'm not sure if C is the right level. We only need to do what `PyArray_ResultType` does, but this is mostly just legacy behavior, not sure if it will be simpler enough as we handle non-builtins too. But maybe Sebastian has a better idea if it's worth it.

numpy/lib/tests/test_function_base.py
ikrommyd: yeah the whole PR needs a cleanup actually after we converge on the dtype resolution

numpy/_core/src/npysort/x86-simd-sort
ikrommyd: thanks for noticing, probably I staged it by accident while rebasing

numpy/lib/_function_base_impl.py
ikrommyd: well it still won't fail with a `_UFuncNoLoopError` right? Plus I wasn't expecting this to ever fail. What kind of failure might we have here that might be worth trying the fallback for?

ikrommyd:
Because I know very little about implementing promoters, I had claude generate me one to see the idea and pushed it in another branch. This is the commit: https://github.com/ikrommyd/numpy/commit/c32bc51ee34cd01e21035ce6730daffe47b04f8e. Do you think we should go into that direction?

numpy/_core/src/umath/unwrap.cpp
ikrommyd: should be done

numpy/_core/src/umath/unwrap.cpp
ikrommyd: should be done

MaanasArora:
I still don't think we should care about specializing precision for `discont`, and that branch just moves the same logic into C. I think that should probably be resolved first. To be clear, I think a C promoter is worth it if it is more simply correct (for both legacy behavior and user dtypes), not to solve that problem.

ikrommyd:
Main's output dtype is `result_type(p, period)` only, never `discont`, so folding `discont` into that promotion would change the return dtype based on a threshold value, which is worse than what we're fixing. But `discont`'s own comparison precision on main already follows NEP 50 (weak scalars downcast to `p`'s dtype, explicit ones promote), so a gufunc that always compares `discont` at a fixed type (double, or always `T`) diverges from main in one direction or the other.  For example if  `p=[0,3]` with float16 dtype, and `discont=3.0001`, main gives `[0,-1]`, while fixed-double gives `[0,3]`. That's why I think `discont` needs its own resolution, separate from `T`. Doing it in python or c++ does not matter much.

MaanasArora:
I agree casting to `T` is technically a behavior change, just not sure if we shouldn't make it. I also don't think it's going to affect weak scalars (as tested), just explicit ones.

But even right now, an explicit `longdouble` `discont` would still lose precision. Then we would need at least 3 more loops to bring this to its logical completion!

ikrommyd:
I'd be fine with casting to `T` for floats but we can't do it for integer `T`s

numpy/lib/_function_base_impl.py
```diff
@@ -1833,34 +1864,36 @@ def unwrap(p, discont=None, axis=-1, *, period=2 * pi):
...
+    # match NEP 50: a discont that would collapse to dtype anyway (the
+    # weak-scalar or same-dtype case) is compared at that native precision;
+    # anything wider (explicit stronger scalar, or cross-kind on an integer
+    # dtype) is compared at >=double
+    discont_type = np.longdouble if dtype.type is np.longdouble else np.float64
+    if _nx.issubdtype(dtype, _nx.floating) and np.result_type(dtype, discont) == dtype:
+        discont_type = dtype.type
```
ngoldbaum:
```suggestion
    discont_type = type(dtype) if floating else np.float64
```

I think this is all that's needed. And since it's a one-liner, let's not do the C++ promoter. What do you think?

numpy/_core/src/umath/unwrap.cpp
ngoldbaum: I think my other suggestion would resolve this thread.

numpy/lib/tests/test_function_base.py
ngoldbaum: I'm fine with changing the error, it just needs a release note.

numpy/_core/src/umath/unwrap.cpp
ngoldbaum: Eh, on second thought it's OK not to worry about this. We should ideally clean this up codebase-wide then enable UBSan checks.

numpy/lib/_function_base_impl.py
ngoldbaum: A user DType might raise a `DTypePromotionError`, which is a subclass of TypeError and then be caught by your try/except. IMO it makes sense to move it inside the try/except for a little extra robustness at no cost to us.

ngoldbaum (commented):
I think @MaanasArora and @ikrommyd have actually converged on a design, see my comment below. I think a one-liner is better than a C++ promoter.

numpy/lib/_function_base_impl.py
ngoldbaum: Suggested release note: an explicitly typed `discont` argument passed to the `unwrap` function wider than the result dtype is now compared at the result dtype rather than promoting. This may change the result of `np.unwrap` by ~1 ULP.

MaanasArora: Perfect yes, I think we've converged and this seems to be best way to do it, thanks!

numpy/_core/src/umath/unwrap.cpp
```diff
@@ -0,0 +1,314 @@
...
+floor_mod(T a, T b)
+{
+    if (b == 0) {
+        npy_set_floatstatus_divbyzero();
+        return 0;
+    }
+    if (b == -1) {
+        if (a == std::numeric_limits<T>::min()) {
+            npy_set_floatstatus_divbyzero();
+        }
```
ngoldbaum: I don't think it's correct to set a divide by zero error here (`x % -1` doesn't divide by zero). You only call this helper with signed integers anyway so you could also add a `&& std::is_signed_v<T>` to the `enable_if_t` predicate above, delete this stanza that sets the floating point status,  and completely sidestep this whole issue.

ikrommyd:
Is it a thing to do multiple release notes like an improvement + another one for behavior changes or should everything go into the same `‎doc/release/upcoming_changes/9959.improvement.rst`?

ngoldbaum:
Two different release notes are a thing, you can add a second one for the changes.

ikrommyd:
@ngoldbaum @MaanasArora I believe I have addressed the comments and have added the converged dtype logic in https://github.com/numpy/numpy/pull/31848/changes/86a2566e84edf5e4089e4ca41c7a1f6aef934152
The riscv CI failure is unrelated infrastructure problem.
Regarding Nathan's comments about UB with signed integers, I think these should probably all be fixed centrally in numpy at some point and not do anything here as no numpy additions/subtractions that I saw care about UB so far.

numpy/lib/_function_base_impl.py
ikrommyd: I believe I did that.

ngoldbaum (approved):
LGTM! I'd appreciate it if one of Maanas or Marten could look this over and hit the merge button if they approve.

Also maybe worth a ping to the mailing list as a head's up for downstream maintainers who try to be fully compatible with all NumPy ufuncs. That's pretty much just astropy (and Marten already knows) but it may be other people too.

MaanasArora (approved):
Looks good to me! Thanks for the changes, I think the dtype resolution comes together quite nicely. I didn't look very closely at the tests but they seem thorough. In it goes... thanks @ikrommyd for the nice improvement!

ikrommyd:
Thank y'all for the reviews! Cheers!

PR #31851, opened 2026-07-03
TYP: specialized ``ldexp`` ufunc

jorenham:
Following #31824, this adds the specialized typing stubs for the `np.ldexp` binary ufunc. The nice thing about this one is that the dtype promotions one epend on the first inoput, not the second one, so it's not as complicated to type as the other (all remaining) binary ufuncs.

---

No AI.


charris:
Thanks Joren.

PR #31853, opened 2026-07-03
CI: fix hangs on MacOS ASan CI

ngoldbaum:
### PR summary
This should (hopefully) fix hangs seen on the UBSan CI lately.

This is happening because of the phased rollout of macos-26 images: https://github.com/actions/runner-images/issues/14167

I'm pretty sure the problem is that LLVM 20 doesn't support macos-26. Let's see!

#### AI Disclosure
No AI use.

### DIFF
--- a/.github/workflows/compiler_sanitizers.yml
+++ b/.github/workflows/compiler_sanitizers.yml
@@ -33,7 +33,7 @@ jobs:
   clang_ASAN_UBSAN:
     # To enable this workflow on a fork, comment out:
     if: github.repository == 'numpy/numpy'
-    runs-on: macos-latest
+    runs-on: macos-26
     steps:
     - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
       with:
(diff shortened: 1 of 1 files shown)


ngoldbaum:
It does fix the hang: https://github.com/numpy/numpy/actions/runs/28679567181/job/85060098843

.github/workflows/compiler_sanitizers.yml
```diff
@@ -54,15 +54,15 @@ jobs:
...
+        brew install llvm
+        LLVM_PREFIX=$(brew --prefix llvm)
         echo CC="$LLVM_PREFIX/bin/clang" >> $GITHUB_ENV
         echo CXX="$LLVM_PREFIX/bin/clang++" >> $GITHUB_ENV
         echo LDFLAGS="-L$LLVM_PREFIX/lib" >> $GITHUB_ENV
         echo CPPFLAGS="-I$LLVM_PREFIX/include" >> $GITHUB_ENV
     - name: Build Python with address sanitizer
       run: |
-        CONFIGURE_OPTS="--with-address-sanitizer" pyenv install 3.14
+        CONFIGURE_OPTS="--with-address-sanitizer" pyenv install -v 3.14
```
ngoldbaum, line 65: I figure extra CI logging is worth having more context if something like this ever happens again in the future.

charris:
Thanks Nathan. Let's see how this goes, macos seems generally overloaded these days.

PR #31859, opened 2026-07-04
MNT: remove some obsolete string to bool workarounds

ikrommyd:
### PR summary
Removes a couple of obsolete string to bool workarounds and updates a comment and adds a tiny bit of testing.

N/A

#### AI Disclosure
No AI

### DIFF
--- a/numpy/_core/numeric.py
+++ b/numpy/_core/numeric.py
@@ -534,7 +534,15 @@ def count_nonzero(a, axis=None, *, keepdims=False):
     if axis is None and not keepdims:
         return multiarray.count_nonzero(a)
 
-    a_bool = asanyarray(a, dtype=np.bool)
+    a = asanyarray(a)
+
+    # This is a performance optimization for character dtypes
+    # TODO: this can be removed if the legacy fixed-width string dtypes are ever removed
+    if np.issubdtype(a.dtype, np.character):
+        a_bool = a != a.dtype.type()
+    else:
+        a_bool = a.astype(np.bool, copy=False)
+
     return a_bool.sum(axis=axis, dtype=np.intp, keepdims=keepdims)
 
 
--- a/numpy/_core/numeric.py
+++ b/numpy/_core/numeric.py
@@ -535,13 +535,7 @@ def count_nonzero(a, axis=None, *, keepdims=False):
         return multiarray.count_nonzero(a)
 
     a = asanyarray(a)
-
-    # TODO: this works around .astype(bool) not working properly (gh-9847)
-    if np.issubdtype(a.dtype, np.character):
-        a_bool = a != a.dtype.type()
-    else:
-        a_bool = a.astype(np.bool, copy=False)
-
+    a_bool = a.astype(np.bool, copy=False)
     return a_bool.sum(axis=axis, dtype=np.intp, keepdims=keepdims)
 
 
--- a/numpy/_core/numeric.py
+++ b/numpy/_core/numeric.py
@@ -534,8 +534,7 @@ def count_nonzero(a, axis=None, *, keepdims=False):
     if axis is None and not keepdims:
         return multiarray.count_nonzero(a)
 
-    a = asanyarray(a)
-    a_bool = a.astype(np.bool, copy=False)
+    a_bool = asanyarray(a, dtype=np.bool)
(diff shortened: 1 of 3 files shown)


numpy/_core/numeric.py
```diff
@@ -535,13 +535,7 @@ def count_nonzero(a, axis=None, *, keepdims=False):
...
 
     a = asanyarray(a)
-
-    # TODO: this works around .astype(bool) not working properly (gh-9847)
-    if np.issubdtype(a.dtype, np.character):
-        a_bool = a != a.dtype.type()
-    else:
-        a_bool = a.astype(np.bool, copy=False)
-
+    a_bool = a.astype(np.bool, copy=False)
```
eendebakpt: `asanyarray` has a `dtype` and `copy` argument. Can this cast be combined with the previous line?

ikrommyd: Ah yeah with `copy=None` which is the default in `asanyarray` it should theoretically be equivalent. Done.

numpy/_core/numeric.py
```diff
@@ -534,14 +534,7 @@ def count_nonzero(a, axis=None, *, keepdims=False):
     if axis is None and not keepdims:
         return multiarray.count_nonzero(a)
 
-    a = asanyarray(a)
-
-    # TODO: this works around .astype(bool) not working properly (gh-9847)
-    if np.issubdtype(a.dtype, np.character):
-        a_bool = a != a.dtype.type()
```
ngoldbaum, line 541: This is actually a pretty decent performance boost for `U` and `S` dtypes. Try running `np.count_nonzero` on a large array of characters, I see >10x performance regressions with this PR applied.

This is a big enough difference that IMO it's worth leaving behind the if statement and a comment explaining that this is a workaround for a dedicated 'S'/'U' -> bool cast loop. I don't think it's worth adding that loop since in the long term I want to completely replace 'S' and 'U' with StringDType and a new ByteSTringDType that both have fixed-width modes.

numpy/_core/tests/test_numeric.py
```diff
@@ -1840,6 +1840,15 @@ def assert_equal_w_dt(a, b, err_msg):
                 assert_equal(np.count_nonzero(m),
                              expected, err_msg=err_msg)
 
+    @pytest.mark.parametrize("dtype", ["S5", "U5"])
```
ngoldbaum: Might as well add `"T"` as well for completeness.

numpy/_core/tests/test_numeric.py
```diff
@@ -1840,6 +1840,15 @@ def assert_equal_w_dt(a, b, err_msg):
...
+    def test_count_nonzero_character_axis(self, dtype):
+        a = np.array([
+            ["", "0", " "],
+            ["1", "", "False"],
+        ], dtype=dtype)
+
+        assert_array_equal(np.count_nonzero(a, axis=1), [2, 2])
```
ngoldbaum, line 1850: Might as well test the path for an unspecified axis as well.

```suggestion
        assert_array_equal(np.count_nonzero(a, axis=1), [2, 2])
        assert np.count_nonzero(a) == 4
```

numpy/_core/numeric.py
ikrommyd: I did not consider that actually, I thought `astype` or `asarray` would do internal tricks. That effectively makes the whole PR useless though no? I mean, apart from the test changes.

ngoldbaum: Unfortunately yeah.

ikrommyd: Okay then I will see this in my own benchmark too and I can just remove the comment and keep the test changes cause it's not really a TODO if there's a performance reason for it. Does that sound good or should I just close it all together? I think uncommenting the `yield param(np.bool(0), id="bool")` in the tests and removing the TODO comment are still valid.

ngoldbaum: I'd still leave a comment explaining that this is for performance reasons and we can eventually remove it when/if we drop the legacy fixed-width string dtypes.

ikrommyd:
@ngoldbaum I think I addressed all your review comments.

ngoldbaum:
Thanks @ikrommyd!

PR #30049, opened 2025-10-22
BLD: update scipy-openblas, use -Dpkg_config_path

mattip:
Closes #29391 by patching OpenBLAS. Hopefully by the time 0.3.31 is released we can remove the patch.

Also use `spin build -- -Dpkg_config_path=` instead of setting `PKG_CONFIG_PATH` in the normal (non-wheel building) CI.

Issue #29391, opened 2025-07-16
BUG: np.dot produces incorrect results when run concurrently


BenLewis-Seequent:
### Describe the issue:

When using multiple threads that perform a `np.dot` operations on 2d arrays, where the second argument is transposed, the result of the dot product can be incorrect. The incorrect values that is produced seem to be valid results from another threads operation.

Each thread is performing the `np.dot` on it own independent arrays, so this shouldn't cause any threading problems.

This only seems to occur, if the arrays involved in the operation is big enough, for example a (50,000, 3) x (3, 3) operation doesn't cause this bug but a (100,000, 3) x (3, 3) operation does.

### Reproduce the code example:

```python
import threading

import numpy as np

# If set lower than ~50,000 the bug is not triggered
N_POINTS_PER_THREAD = 100000

# If True, the matrix is transposed before multiplication, this results in incorrect results
# If False, the matrix is used as is, which produces correct results
TRANSPOSE = True

unequal_condition = threading.Event()

def worker(thread_id):
    points = np.ones((N_POINTS_PER_THREAD, 3), dtype=np.float64) * thread_id
    matrix = np.identity(3, dtype=np.float64)
    result = np.dot(points, matrix.T if TRANSPOSE else matrix)
    if not np.all(result == thread_id):
        unequal_condition.set()
        print(f"Thread {thread_id} produced wrong results")


def run():
    num_threads = 10
    
    # Run worker threads
    threads = []
    for i in range(num_threads):
        t = threading.Thread(target=worker, args=(i,))

        threads.append(t)
        t.start()
    
    for t in threads:
        t.join()

    if unequal_condition.is_set():
        print("Unequal condition was set, indicating at least one thread produced wrong results.")
        return True
    else:
        print("All threads produced correct results.")
        return False

if __name__ == "__main__":
    num_runs = 5000
    for i in range(1, num_runs + 1):
        print(f"Run {i} of {num_runs}")
        if run():
            break
```

### Error message:

```shell
Example output: 
Run 1 of 5000
All threads produced correct results.
Run 2 of 5000
All threads produced correct results.
Run 3 of 5000
All threads produced correct results.
Run 4 of 5000
All threads produced correct results.
Run 5 of 5000
All threads produced correct results.
Run 6 of 5000
All threads produced correct results.
Run 7 of 5000
Thread 1 produced wrong results
Thread 3 produced wrong results
Unequal condition was set, indicating at least one thread produced wrong results.
```

[...]

PR #31860, opened 2026-07-04
BUG: Add thread-safe dot wrapper and regression tests for #29391

eeshsaxena:
# NumPy PR — Fix: `np.dot` race condition with transposed args (gh-29391)

**Addresses:** [numpy/numpy#29391](https://github.com/numpy/numpy/issues/29391)  
**Complementary to:** [numpy/numpy#30049](https://github.com/numpy/numpy/pull/30049)  
**Affects:** NumPy with OpenBLAS 0.3.28 / 0.3.29 (any platform)  
**Fixed in:** NumPy ≥ 2.3.5 (via PR #30049)

---

## Problem

When multiple Python threads call `np.dot(A, B.T)` on large arrays simultaneously,
some threads receive **silently incorrect results** — values computed for a different
thread's input.

```python
import threading, numpy as np

def worker(i, results):
    pts = np.ones((100_000, 3)) * i   # all values == i
    mat = np.eye(3)
    results[i] = np.dot(pts, mat.T)   # should be all-i, but sometimes isn't!

results = [None] * 10
threads = [threading.Thread(target=worker, args=(i, results)) for i in range(10)]
for t in threads: t.start()
for t in threads: t.join()

# On OpenBLAS 0.3.28/0.3.29: results[3] might contain values from results[7]
```

## Root Cause

The bug is in **OpenBLAS 0.3.28** (`driver/level3/level3_thread.c`),
introduced via OpenBLAS PR #4741 ("Enhancing Core Utilization in BLAS Calls").

OpenBLAS template-compiles `gemm_driver()` into four functions — one per
transpose combination: `NN`, `NT`, `TN`, `TT`. The `level3_lock` mutex that
is supposed to serialize concurrent GEMM calls was inadvertently made a
**function-local `static` variable**, giving each variant its own lock:

```c
// OpenBLAS 0.3.28 — THE BUG:
int gemm_driver_nt(...) {
    static pthread_mutex_t level3_lock = PTHREAD_MUTEX_INITIALIZER; // own lock!
    pthread_mutex_lock(&level3_lock);
    // ... access shared internal buffers ...
    pthread_mutex_unlock(&level3_lock);
}
// gemm_driver_nn has a DIFFERENT lock — no serialization with NT callers!
```

Result: Thread A (calling `np.dot(a, b)` → `NN`) and Thread B (calling
`np.dot(a, b.T)` → `NT`) acquire **different locks** and race on shared
internal OpenBLAS state → data corruption.

**Fixed in OpenBLAS 0.3.30** by moving `level3_lock` to file scope.
**Shipped in NumPy 2.3.5** via PR #30049 (updates bundled `scipy-openblas`).

## This PR Contributes

Since the core fix (updating vendored OpenBLAS) is already in PR #30049, this
PR adds the complementary test infrastructure and a user-facing utility:

### 1. `numpy/tests/test_dot_threading.py` — Regression Test Suite

A comprehensive pytest test suite that:

- **Regression test** (`TestDotTransposeRaceConditionRegression`): Detects the race
  condition on 100k-row arrays with 10 concurrent threads, marked `xfail` on
  affected OpenBLAS versions and expected to pass on ≥ 0.3.30.
- **Baseline tests** (`TestDotBaselineCorrectness`): Verify that non-transposed
  concurrent dot, small arrays, and single-threaded usage always work correctly.
- **Workaround tests** (`TestWorkaroundEffectiveness`): Validate that `threadpoolctl`,
  `threading.Lock`, and `OPENBLAS_NUM_THREADS=1` prevent the race.
- **Environment detection** (`TestOpenBLASVersionInfo`): Verify the version
  detection utility works across OpenBLAS detection strategies.

### 2. `numpy/_core/_blas_version_check.py` — Version Detection + Warning

Detects the installed OpenBLAS version using multiple strategies (NumPy config
API, ctypes CDLL `openblas_get_config`, library list parsing) and emits a
`RuntimeWarning` at most once per thread when a known-buggy version is in use.

```python
from numpy._core._blas_version_check import get_blas_threading_info
print(get_blas_threading_info())
# {'library': 'openblas', 'version': '0.3.29', 'is_threading_safe': False, ...}
```

### 3. `numpy/_thread_safe_dot.py` — Thread-safe Wrappers

Drop-in replacements for `np.dot`, `np.matmul`, `np.inner` that are safe
to call concurrently on all OpenBLAS versions. Uses `threadpoolctl` (preferred)
or `threading.Lock` (fallback) to prevent concurrent BLAS variant dispatch.

```python
from numpy._thread_safe_dot import safe_dot, configure

configure("threadpoolctl")  # or "lock"
result = safe_dot(A, B.T)   # always correct in threads
```

---

## Running the Tests

```bash
# Quick run (CI-friendly, ~10s)
python -m pytest numpy/tests/test_dot_threading.py -v

# Full run with slow tests (~2 min)
python -m pytest numpy/tests/test_dot_threading.py -v -m slow

# On affected OpenBLAS (0.3.28/0.3.29): xfail tests should show as XFAIL
# On fixed OpenBLAS (>= 0.3.30):        all tests should PASS
```

## Workarounds for Users on Affected Versions

```python
# Option 1: threadpoolctl (recommended, pip install threadpoolctl)
from threadpoolctl import threadpool_limits
with threadpool_limits(limits=1, user_api="blas"):
    result = np.dot(A, B.T)

# Option 2: environment variable (process-wide)
# OPENBLAS_NUM_THREADS=1 python your_script.py

# Option 3: threading.Lock
import threading
_lock = threading.Lock()
with _lock:
    result = np.dot(A, B.T)

# Option 4: safe_dot from this PR
from numpy._thread_safe_dot import safe_dot
result = safe_dot(A, B.T)
```

> **Note:** `np.ascontiguousarray(B.T)` is NOT a complete fix. It forces
> the `NN` GEMM variant but does not protect against races between other
> variant pairs (e.g., two concurrent `TT` calls from different threads).

---

## Files Changed

| File | Action |
|------|--------|
| `numpy/tests/test_dot_threading.py` | **NEW** — regression + workaround tests |
| `numpy/_core/_blas_version_check.py` | **NEW** — OpenBLAS version detection |
| `numpy/_thread_safe_dot.py` | **NEW** — thread-safe dot/matmul/inner |

### DIFF
--- a/numpy/_core/_blas_version_check.py
+++ b/numpy/_core/_blas_version_check.py
@@ -1,0 +1,268 @@
+"""
+OpenBLAS version detection and threading safety warnings for NumPy.
+
+This module detects whether the installed OpenBLAS version is affected by
+the thread-unsafe scratch buffer bug (fixed in OpenBLAS >= 0.3.30) and
+issues appropriate RuntimeWarnings so users can take action.
+
+Background
+----------
+OpenBLAS <= 0.3.29 contains a race condition in its multi-level blocked
+DGEMM implementation. When the computation is large enough to use the
+blocked kernel path (typically arrays with >= ~50k–100k rows), two threads
+can simultaneously read/write the same internal scratch buffer, producing
+silently incorrect results.
+
+The bug specifically affects ``np.dot(A, B.T)`` and similar operations
+where one argument is a non-contiguous (transposed) array, because only
+then does OpenBLAS need to allocate and use the scratch buffer.
+
+References
+----------
+- NumPy issue:  https://github.com/numpy/numpy/issues/29391
+- NumPy PR fix: https://github.com/numpy/numpy/pull/30049
+- OpenBLAS fix: https://github.com/OpenMathLib/OpenBLAS/pull/XXXX
+"""
+
+import re
+import threading
+import warnings
+
+
+# The first OpenBLAS version that contains the thread-safety fix.
+_SAFE_OPENBLAS_VERSION = (0, 3, 30)
+_SAFE_OPENBLAS_VERSION_STR = "0.3.30"
+
+# Per-thread flag to avoid spamming repeated warnings in hot loops.
+_warned_state = threading.local()
+
+
+# ---------------------------------------------------------------------------
+# Internal helpers
+# ---------------------------------------------------------------------------
+
+def _parse_version(version_str):
+    """
+    Parse a version string like "0.3.29" into a comparable integer tuple.
+
+    Returns
+    -------
+    tuple of int, e.g. (0, 3, 29), or None if parsing fails.
+    """
+    if not version_str or version_str == "unknown":
+        return None
+    match = re.match(r"(\d+)\.(\d+)(?:\.(\d+))?", version_str.strip())
+    if not match:
+        return None
+    major = int(match.group(1))
+    minor = int(match.group(2))
+    patch = int(match.group(3) or "0")
+    return (major, minor, patch)
+
+
+def _get_openblas_version():
+    """
+    Attempt to extract the OpenBLAS version from NumPy's build/runtime info.
+
+    Tries multiple strategies in order of reliability, because the NumPy
+    config API changed significantly between NumPy 1.x and 2.x.
+
+    Returns
+    -------
+    str or None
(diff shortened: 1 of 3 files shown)


jorenham:

PR #31862, opened 2026-07-04
BUG: fix `StringDType` coerce flag in binary ufunc promotion (#31861)

### LINKED ISSUE #31861: BUG: `StringDType(coerce=...)` is not preserved by binary ufuncs when operands differ
### Describe the issue:

**Brief description of bug**

Binary operations on `StringDType` arrays do not correctly combine the `coerce` setting from both operands.

When two arrays with different `coerce` settings are used in an operation such as `np.add`, the result dtype effectively inherits `coerce` from the **first** operand only.

The result should enable coercion only when **both** input dtypes have `coerce=True` (same rule as `np.promote_types` / `common_instance`).

- Current output

```python
StringDType()
True
```

- Expected output

```python
StringDType(coerce=False)
False
```

**Why its incorrect**

The result reports `coerce=True` even though one input has `coerce=False`. That allows assignments like `c[0] = object()` to succeed by stringifying the value. With `coerce=False` preserved, that assignment should raise:

```
ValueError: StringDType only allows string data when string coercion is disabled.
```

Note that behavior is order-dependent for ufuncs: `np.add(coerce=False, coerce=True)` happens to look correct, while `np.add(coerce=True, coerce=False)` does not which is consistent with only the first descriptor being used.

**Debugging hints**

`np.promote_types` already does the right thing both ways, because `common_instance` in `numpy/_core/src/multiarray/stringdtype/dtype.c` was fixed recently in https://github.com/numpy/numpy/pull/31846 (backport https://github.com/numpy/numpy/pull/31856) to use:

```c
dtype1->coerce && dtype2->coerce
```

The remaining bug is in the binary ufunc path: `binary_resolve_descriptors` in `numpy/_core/src/umath/stringdtype_ufuncs.cpp`, which still combines `coerce` incorrectly:

```diff
@@ -248,7 +248,7 @@ binary_resolve_descriptors(...)
 {
     PyArray_StringDTypeObject *descr1 = (PyArray_StringDTypeObject *)given_descrs[0];
     PyArray_StringDTypeObject *descr2 = (PyArray_StringDTypeObject *)given_descrs[1];
-    int out_coerce = descr1->coerce && descr1->coerce;
+    int out_coerce = descr1->coerce && descr2->coerce;
     PyObject *out_na_object = NULL;
```

So the result dtype depends only on the first input’s `coerce` flag (the `descr1 && descr1` typo), which is why `np.add` still misbehaves even though type promotion via `promote_types` is correct.

Applying that one-line change produces the expected result:

```python
StringDType(coerce=False)
False
```

and assigning a non-string value correctly raises

```
ValueError: StringDType only allows string data when string coercion is disabled.
```

If this is confirmed to be a bug, I'd be happy to contribute a fix

```python
import numpy as np
from numpy.dtypes import StringDType

a = np.array(["a"], dtype=StringDType(coerce=True))
b = np.array(["b"], dtype=StringDType(coerce=False))

c = np.add(a, b)

print(c.dtype)        # currently: StringDType()
print(c.dtype.coerce) # currently: True; expected: False

c[0] = object()       # currently succeeds (stringifies); should raise ValueError
```

### Error message:

```shell
No exception is raised.

If the resulting dtype preserved coerce=False, assigning a non-string value would raise:

ValueError: StringDType only allows string data when string coercion is disabled.
```

```
2.6.0.dev0+git20260625.67f37e4
3.14.3 (main, Apr  8 2026, 11:37:06) [Clang 14.0.3 (clang-1403.0.22.14.1)]
```

### Runtime Environment:

```
[{'numpy_version': '2.6.0.dev0+git20260625.67f37e4',
  'python': '3.14.3 (main, Apr  8 2026, 11:37:06) [Clang 14.0.3 '
            '(clang-1403.0.22.14.1)]',
  'uname': uname_result(system='Darwin', node='192.168.0.100', release='22.6.0', version='Darwin Kernel Version 22.6.0: Wed Jul  5 22:22:52 PDT 2023; root:xnu-8796.141.3~6/RELEASE_ARM64_T8103', machine='arm64')},
                      'not_found': ['ASIMDFHM']}},
```

_No response_

Ethan0456:
### PR summary

Fixes #31861.

Binary `StringDType` ufuncs resolved output `coerce` with `descr1->coerce && descr1->coerce` instead of combining both operands.

That made results order-dependent (e.g. `np.add` with `coerce=True` then `coerce=False` kept `coerce=True`) and allowed non-string assignment via implicit stringification.

This changes the combination to `descr1->coerce && descr2->coerce`, matching `common_instance` / `np.promote_types`, and adds a small regression test for both operand orders.

I have been using NumPy for some time now. I hit this while working with `StringDType` and `coerce`, and found two places with similar typo bug. I checked and found that one typo in `common_instance` has been fixed yesterday, so wanted to fix the remaining typo in the binary ufunc path.

#### AI Disclosure

I used Grok to help locate the bug in `binary_resolve_descriptors`, draft the regression test, and refine the issue/PR wording. I made and verified the fix with correctness checks, rebuilds and the test suite; final review and PR submission were done by me.

### DIFF
--- a/numpy/_core/tests/test_stringdtype.py
+++ b/numpy/_core/tests/test_stringdtype.py
@@ -1858,6 +1858,27 @@ def test_coerce_promotion_commutative():
     assert np.promote_types(dt_no_coerce, dt_no_coerce) == dt_no_coerce
 
 
+@pytest.mark.parametrize(
+    "coerce_a, coerce_b",
+    [(True, False), (False, True)],
+)
+def test_add_preserves_coerce_false(coerce_a, coerce_b):
+    # Mixing coerce=True and coerce=False under ufuncs should promote to
+    # coerce=False (same as np.promote_types), so assignment of non-strings
+    # remains strict.
+    a = np.array(["a"], dtype=StringDType(coerce=coerce_a))
+    b = np.array(["b"], dtype=StringDType(coerce=coerce_b))
+
+    c = np.add(a, b)
+    assert c.dtype == StringDType(coerce=False)
+    assert c.dtype.coerce is False
+    assert_array_equal(c, np.array(["ab"], dtype=StringDType(coerce=False)))
+
+    # Same error as a direct StringDType(coerce=False) array.
+    with pytest.raises(ValueError, match="string coercion is disabled"):
+        c[0] = object()
+
+
 def test_repeat(string_array):
     res = string_array.repeat(1000)
     # Create an empty array with expanded dimension, and fill it.  Then,
--- a/numpy/_core/tests/test_stringdtype.py
+++ b/numpy/_core/tests/test_stringdtype.py
@@ -1820,10 +1820,13 @@ def test_unset_na_coercion():
                      StringDType(na_object=None)]:
         if op_dtype is None:
             op = "2"
+            expected = StringDType(na_object=None)  # coerce defaults to True
         else:
             op = np.array("2", dtype=op_dtype)
+            expected = StringDType(na_object=None, coerce=op_dtype.coerce)
         res = arr + op
         assert_array_equal(res, ["hello2", "world2"])
+        assert res.dtype == expected
 
     # dtype instances with distinct explicitly set NA objects are incompatible
     for op_dtype in [StringDType(na_object=pd_NA), StringDType(na_object="")]:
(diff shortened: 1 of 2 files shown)


shipitdev (commented):
Thanks for the patch, @Ethan0456!

For a maintainer reviewing this (just laying out the PR simply for an easier merge):
I reviewed this, and it looks like whoever originally coded this just mistyped int out_coerce = descr1->coerce && descr1->coerce; when trying to resolve the output coerce flag. Because it checked descr1 twice and ignored descr2, binary_resolve_descriptors was returning the wrong output descriptor if the two arrays had different coerce flags.

The tests correctly isolate the bug, seem totally fine to me. Overall, this looks like a super easy merge to me! Thank you.

numpy/_core/tests/test_stringdtype.py
```diff
@@ -1858,6 +1858,27 @@ def test_coerce_promotion_commutative():
     assert np.promote_types(dt_no_coerce, dt_no_coerce) == dt_no_coerce
 
 
+@pytest.mark.parametrize(
+    "coerce_a, coerce_b",
+    [(True, False), (False, True)],
+)
+def test_add_preserves_coerce_false(coerce_a, coerce_b):
```
ngoldbaum: Instead of adding a new test, couldn't you simply add an assert to `test_unset_na_coercion`:

https://github.com/numpy/numpy/blob/b8853b27c1884002eefc55b8e3a87e7a1bac241f/numpy/_core/tests/test_stringdtype.py#L1813-L1826

Something like:

```
expected_coerce = op_dtype is None or op_dtype.coerce
assert res.dtype == StringDType(na_object=None, coerce=expected_coerce)
```

ngoldbaum (commented):
Thanks for catching this, I agree the fix is correct.

I left one comment about the new test, which I think is duplicative of existing tests.

AIs like to write new tests, it's good to prompt them to prefer expanding existing tests rather than to write new tests.

numpy/_core/tests/test_stringdtype.py
Ethan0456: Good catch. I have dropped the new test and extended `test_unset_na_coercion`; now each branch builds the expected result dtype (including `coerce`) before asserting it.

Ethan0456:
@ngoldbaum Addressed your review feedback in the latest commit, thanks!

ngoldbaum:
Thanks @Ethan0456!

PR #31867, opened 2026-07-05
ENH: Reduce __array_function__ dispatch overhead via tuple-spec in fromnumeric (#31866)

### LINKED ISSUE #31866: ENH: Reduce `__array_function__` dispatch overhead

Every call to a `@array_function_dispatch`-decorated function (e.g. `np.sum`, `np.prod`, `np.max`, `np.any`, `np.sort`, `np.reshape`, ...) currently pays the cost of the full NEP 18 dispatch machinery in `dispatcher_vectorcall` (`numpy/_core/src/multiarray/arrayfunction_override.c`) even when all
arguments are plain ndarrays and no override is possible. Below are two different approaches to reduce the overhead.

### Approach A — Skip dispatcher for exact-ndarray + basic-type args

Branch: <https://github.com/numpy/numpy/compare/main...eendebakpt:numpy:dispatcher_exact_ndarray_fastpath>

Fast-path check at the top of `dispatcher_vectorcall`: if every positional and keyword argument's type is either `PyArray_Type` or a *scalar* Python builtin (`int`, `float`, `bool`, etc.), skip the dispatcher entirely and call `default_impl` directly.

Benchmark (`bench_reduction_performance.py`, Python 3.14 release):

| Benchmark            | main      | branch    | ratio         |
|----------------------|----------:|----------:|--------------:|
| np.sum(array_4)      |    992 ns |    908 ns |  1.09× faster |
| np.sum(array_20)     |    990 ns |    915 ns |  1.08× faster |
| np.prod(array_20)    |    961 ns |    879 ns |  1.09× faster |
| np.max(array_20)     |    973 ns |    894 ns |  1.09× faster |
| np.sum(int_20)       |    992 ns |    904 ns |  1.10× faster |
| np.sum(10x10)        |   1.00 µs |    914 ns |  1.10× faster |
| np.sum(10x10,axis=0) |   1.55 µs |   1.51 µs |  1.03× faster |
| np.sum(10x10,axis=1) |   1.49 µs |   1.47 µs |  1.02× faster |
| np.any(bool_20)**  |  **834 ns** | **735 ns** | **1.14× faster |
| np.sum(2x2_float)    |    986 ns |    908 ns |  1.09× faster |
| np.sum(2x2_int)      |    989 ns |    907 ns |  1.09× faster |
| np.sin(array_20) †   |    323 ns |    322 ns | not sig.      |
| **Geometric mean**   |     ref   |           | **1.07× faster** |

† control: not `@array_function_dispatch`-decorated, so unchanged as expected.

### Approach B — Tuple-spec form of `@array_function_dispatch`

Branch: <https://github.com/numpy/numpy/compare/main...eendebakpt:numpy:fast_dispatcher>

Extends `@array_function_dispatch` to accept a tuple of argument names in addition to a callable dispatcher:

```python
# Before
def _sum_dispatcher(a, axis=None, dtype=None, out=None, ...): return (a, out)
@array_function_dispatch(_sum_dispatcher)
def sum(a, axis=None, ...): ...

# After
@array_function_dispatch(("a", "out"))
def sum(a, axis=None, ...): ...
```

At decoration time the arg positions are resolved against the impl's signature.  At call time, `dispatcher_vectorcall` extracts the relevant args directly from `args`/`kwnames` in C — no Python-level dispatcher call.  Then applies the same exact-ndarray + basic-type check as approach A.

Many reduction dispatchers are converted, but a few retain the legacy form.

Benchmark on the same script gives **1.04× geomean** (**update**: should be a bit more now) — smaller than approach A because the callable dispatcher call itself is only ~50 ns.  By working on the downstream `get_implementing_args_and_methods` / tuple-build work performance might be improved towards A.

## Discussion

Approach A and B are quite different. A is a much smaller change with larger performance impact in the happy path (e.g. exact numpy arrays are arguments, a few boolean or int flags). Approach B is conceptually cleaner and also faster in the case of array subclasses. 

The performance gain is not very high in the benchmarks here, but combined with other improvements (for example https://github.com/numpy/numpy/pull/31845) the (relative) gain is larger.

Pavan19047:
### Description

This PR implements **Approach B** described in #31866 to significantly reduce the overhead of NumPy's `__array_function__` dispatch mechanism.

Currently, every call to an `@array_function_dispatch`-decorated function pays the cost of a full C-to-Python roundtrip just to evaluate the dispatcher function (e.g., `_sum_dispatcher`), even when all arguments are standard ndarrays. 

This PR eliminates that overhead for most functions by allowing the decorator to accept a tuple of argument names instead of a callable Python function.

#### Key Changes:
1. **C-Level Dispatch Engine**: Augmented `PyArray_ArrayFunctionDispatcherObject` to extract relevant arguments directly from the C-level `args` and `kwnames` arrays. It also includes a **fast-path**: if all arguments are exact `ndarrays` or basic Python scalars, the dispatcher is bypassed entirely.
2. **Python Decorator**: `overrides.py` now accepts a tuple of strings (e.g., `@array_function_dispatch(("a", "out"))`). These are resolved to index/name pairs at decoration time using `inspect.signature`.
3. **Refactored `fromnumeric.py`**: Converted ~40 boilerplate dispatchers to the new tuple-spec form. (Dispatchers needing complex iteration, like `_choose_dispatcher`, were kept as standard callables; the C engine supports both).
4. **Tests**: Added a full suite of tests in `test_overrides.py` to cover tuple-spec extraction, keyword arguments, and subclass overriding.

All 34,000+ tests in the core module pass locally with zero regressions. 

### Related Issue
Closes #31866 

### Additional context
This PR focuses on `fromnumeric.py` as a robust proof-of-concept for the new engine. I plan to roll out this same tuple-spec refactoring to the remaining modules (`numeric.py`, `shape_base.py`, etc.) in subsequent PRs to completely eliminate the python-level dispatch overhead across the board.

### DIFF
--- a/numpy/_core/fromnumeric.py
+++ b/numpy/_core/fromnumeric.py
@@ -99,11 +99,7 @@ def _wrapreduction_any_all(obj, ufunc, method, axis, out, **kwargs):
     return ufunc.reduce(obj, axis, bool, out, **passkwargs)
 
 
-def _take_dispatcher(a, indices, axis=None, out=None, mode=None):
-    return (a, out)
-
-
-@array_function_dispatch(_take_dispatcher)
+@array_function_dispatch(("a", "out"))
 def take(a, indices, axis=None, out=None, mode='raise'):
     """
     Take elements from an array along an axis.
@@ -200,11 +196,7 @@ def take(a, indices, axis=None, out=None, mode='raise'):
     return _wrapfunc(a, 'take', indices, axis=axis, out=out, mode=mode)
 
 
-def _reshape_dispatcher(a, /, shape, order=None, *, copy=None):
-    return (a,)
-
-
-@array_function_dispatch(_reshape_dispatcher)
+@array_function_dispatch(("a",))
 def reshape(a, /, shape, order='C', *, copy=None):
     """
     Returns a reshaped ndarray without changing data.
@@ -430,11 +422,7 @@ def choose(a, choices, out=None, mode='raise'):
     return _wrapfunc(a, 'choose', choices, out=out, mode=mode)
 
 
-def _repeat_dispatcher(a, repeats, axis=None):
-    return (a,)
-
-
-@array_function_dispatch(_repeat_dispatcher)
+@array_function_dispatch(("a",))
 def repeat(a, repeats, axis=None):
     """
     Repeat each element of an array after themselves
@@ -483,11 +471,7 @@ def repeat(a, repeats, axis=None):
     return _wrapfunc(a, 'repeat', repeats, axis=axis)
 
 
-def _put_dispatcher(a, ind, v, mode=None):
-    return (a, ind, v)
-
-
-@array_function_dispatch(_put_dispatcher)
+@array_function_dispatch(("a", "ind", "v"))
 def put(a, ind, v, mode='raise'):
     """
     Replaces specified elements of an array with given values.
@@ -547,11 +531,7 @@ def put(a, ind, v, mode='raise'):
     return put(ind, v, mode=mode)
 
 
-def _swapaxes_dispatcher(a, axis1, axis2):
-    return (a,)
-
-
-@array_function_dispatch(_swapaxes_dispatcher)
+@array_function_dispatch(("a",))
 def swapaxes(a, axis1, axis2):
     """
     Interchange two axes of an array.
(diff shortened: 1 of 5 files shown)


mattip:
You ignored the PR template and our [AI policy](https://numpy.org/doc/stable/dev/ai_policy.html). Closing.

PR #31869, opened 2026-07-05
DOC: Correct backcompat sentence and reduce NumPy 2 docs

seberg:
We forgot to delete an early sentence claiming no backwards compatibility for NumPy headers.

I also added two footnotes that are admittedly a bit long but:
1. I think when a NumPy 3 happens, we should improve so that there are 2.x releases that already generate runtime code that is compatible with 3.x.
   I.e. whatever we change it is always API _removals_ that drive things (even if removals via replacements). A 2.x version _can_ be compatible easily, so long users would set a "don't use deprecated API" define.
2. I don't hesitate to vendor "future" NumPy API in downstream (e.g. `ml_dtypes`).  I think it is a pragmatic thing to do so long it is only done for old versions and guarded via `NPY_FEATURE_VERSION`.

Looking at this, I realized that the "version API" C-API docs are outdated and also the NumPy 2 transition only interesting in how to set the build vs. runtime version.
That could probably be re-organized a bit more, but it seemed OK to me.

(Used a bit of AI to get started on cleaning out some of the old code, it also wanted "backward" rather than "backwards", so...)

### DIFF
--- a/doc/source/dev/depending_on_numpy.rst
+++ b/doc/source/dev/depending_on_numpy.rst
@@ -28,11 +28,10 @@ least two releases. For more details, see :ref:`NEP23`.
 
 NumPy provides both a Python API and a C-API. The C-API can be accessed
 directly or through tools like Cython or f2py. If your package uses the
-C-API, it's important to understand NumPy's application binary interface
-(ABI) compatibility: NumPy's ABI is forward compatible but not backward
-compatible. This means that binaries compiled against an older version of
-NumPy will still work with newer versions, but binaries compiled against a
-newer version will not necessarily work with older ones.
+NumPy C-API this will generally be backwards compatible to all relevant older
+NumPy versions and forward compatible within the same major NumPy version.
+For more details, for example if you wish to use API added in newer NumPy versions,
+see :ref:`depending_on_numpy`.
 
 Modules can also be safely built against NumPy 2.0 or later in
 :ref:`CPython's abi3 mode <python:stable-abi>`, which allows
@@ -100,7 +99,8 @@ NumPy 1.25 will, when using defaults, expose a C-API compatible with NumPy
 1.19. (the exact version is set within NumPy-internal header files).
 
 NumPy is also forward compatible for all minor releases, but a major release
-will require recompilation (see NumPy 2.0-specific advice further down).
+is expected to require recompilation (see NumPy 2.0-specific advice further down).\
+[#major-transition]_
 
 The default behavior can be customized for example by adding::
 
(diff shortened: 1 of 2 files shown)


ngoldbaum:
The docs build failure is real

seberg:
> The docs build failure is real

Spinx is hard, fixed... (seems we use `c:macro` for the entries but `c:data` is used for linking everywhere.

mhvk (approved):
Changes look good to me, but I find this still somewhat confusing (it is the kind where when I read it, it seems to make sense, but then when I think about it an hour later, I'm confused again...), so probably best not to rely on my approval only!

doc/source/dev/depending_on_numpy.rst
```diff
@@ -132,6 +132,23 @@ For conda-forge packages, please see
...
+.. [#future-api] We do not provide a mechanism to compile a single extension
+    module that is compatible with old NumPy versions but uses new NumPy API
+    when running with a newer NumPy version.
+    This can be achieved via ``PyArray_RUNTIME_VERSION`` and manually
+    backported API, although you may not receive fixes from NumPy headers when
+    using a backport. We recommend only backporting when
+    ``NPY_FEATURE_VERSION`` is lower than the version at which the API was
+    introduced. That way you pick up official header definitions and fixes
+    once they are available. Even without bugs, future NumPy versions could
+    introduce incompatible changes eventually.
```
ngoldbaum, line 149: Is there a spot you can cross-link to explaining this process? Otherwise I'm not totally sure offhand what one needs to backport, so I doubt a random reader would either.

doc/source/dev/depending_on_numpy.rst
```diff
@@ -178,68 +195,32 @@ alongside other, newer packages
...
-*Example for a package using the NumPy C-API (via C/Cython/etc.) which wants to support
-NumPy 1.23.5 and up*:
+NumPy 2.0 ABI handling
+~~~~~~~~~~~~~~~~~~~~~~
+
+NumPy 2.0 changed the C ABI. The important rule for binary wheels is:
+
+1. Wheels built against NumPy 1.xx **will not work** with NumPy 2.0 or later.
+2. Wheels built against NumPy 2.x **will work** with NumPy 1.xx at runtime
+   (within the exposed feature subset).
```
ngoldbaum: Does "exposed feature subset" mean the overlap of feature surface between 1.x and 2.x? Something else? Please clarify a little here.

ngoldbaum (commented):
Overall looks good. I left a couple comments in spots I thought were confusing.

doc/source/dev/depending_on_numpy.rst
seberg: Honestly, no.  The process is copy pasting our headers (possibly including the generated API table slot) which *is* just really ugly.
(But honestly, not sure I would want `#define NPY_UNSAFE_EXPOSE_ALL_API` or so users could do that without the runtime check...)

If it feels too cryptic, I can also just delete it since I don't have a good idea to improve it much.

doc/source/dev/depending_on_numpy.rst
seberg: Tweaked it to mention `NPY_TARGET_VERSION` again and link back there.

ngoldbaum:
Thanks Sebastian!

Issue #27679, opened 2024-10-30
BUG: Inconsistent casting error for masked-arrays in-place operations


rhugonnet:
### Describe the issue:

The changes in scalar casting of 2.0 seem to affect NumPy arrays and masked-arrays differently for in-place operations, resulting in unexpected errors for masked-arrays. See example below (which only occurs in NumPy 2.X, but not on 1.26.4).
Maybe linked to https://github.com/numpy/numpy/issues/7991 and https://github.com/numpy/numpy/issues/27029.

### Reproduce the code example:

```python
import numpy as np
test = np.array([1, 2], dtype="uint8")
test += 5
test2 = np.ma.masked_array([1, 2], dtype="uint8")
test2 += 5
```

### Error message:

```shell
Traceback (most recent call last):
  File "/home/atom/miniconda3/envs/geoutils-dev/lib/python3.11/site-packages/IPython/core/interactiveshell.py", line 3577, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-5-14912cbe9744>", line 1, in <module>
    test2 += 5
  File "/home/atom/miniconda3/envs/geoutils-dev/lib/python3.11/site-packages/numpy/ma/core.py", line 4422, in __iadd__
    self._data.__iadd__(other_data)
numpy._core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'add' output from dtype('int64') to dtype('uint8') with casting rule 'same_kind'
```

### Python and NumPy Versions:

2.1.2
3.11.10 | packaged by conda-forge | (main, Oct 16 2024, 01:27:36) [GCC 13.3.0]

### Runtime Environment:

[{'numpy_version': '2.1.2',
  'python': '3.11.10 | packaged by conda-forge | (main, Oct 16 2024, 01:27:36) '
            '[GCC 13.3.0]',
  'uname': uname_result(system='Linux', node='pop-os', release='6.8.0-76060800daily20240311-generic', version='#202403110203~1711393930~22.04~331756a SMP PREEMPT_DYNAMIC Mon M', machine='x86_64')},
 {'simd_extensions': {'baseline': ['SSE', 'SSE2', 'SSE3'],
                      'found': ['SSSE3',
                                'SSE41',
                                'POPCNT',
                                'SSE42',
                                'AVX',
                                'F16C',
                                'FMA3',
                                'AVX2'],
                      'not_found': ['AVX512F',
                                    'AVX512CD',
[...]

Issue #27029, opened 2024-07-24
BUG: unexpected type promotion multiplying a masked array with an integer


gerritholl:
### Describe the issue:

Multiplying a masked array by 2 causes an unexpected type promotion from float32 to float64 or from int32 to int64.  The same does not happen for regular arrays.

### Reproduce the code example:

```python
import numpy as np

arr = np.array([0, 1], dtype="float32")
marr = np.ma.masked_array(arr)
print((arr * 2).dtype)
print((marr * 2).dtype)

arr = np.array([0, 1], dtype="int32")
marr = np.ma.masked_array(arr)
print((arr * 2).dtype)
print((marr * 2).dtype)
```

### Error message:

```shell
No traceback, but unexpected output:

float32
float64
int32
int64
```

### Python and NumPy Versions:

2.0.0
3.12.0 | packaged by conda-forge | (main, Oct  3 2023, 08:43:22) [GCC 12.3.0]

### Runtime Environment:

[{'numpy_version': '2.0.0',
  'python': '3.12.0 | packaged by conda-forge | (main, Oct  3 2023, 08:43:22) '
            '[GCC 12.3.0]',
  'uname': uname_result(system='Linux', node='oflws222', release='5.3.18-150300.59.106-default', version='#1 SMP Mon Dec 12 13:16:24 UTC 2022 (774239c)', machine='x86_64')},
 {'simd_extensions': {'baseline': ['SSE', 'SSE2', 'SSE3'],
                      'found': ['SSSE3',
                                'SSE41',
                                'POPCNT',
                                'SSE42',
                                'AVX',
                                'F16C',
                                'FMA3',
                                'AVX2'],
                      'not_found': ['AVX512F',
                                    'AVX512CD',
                                    'AVX512_KNL',
                                    'AVX512_KNM',
                                    'AVX512_SKX',
                                    'AVX512_CLX',
                                    'AVX512_CNL',
                                    'AVX512_ICL',
                                    'AVX512_SPR']}},
 {'architecture': 'Haswell',
  'filepath': '/data/gholl/mambaforge/envs/py312/lib/libopenblasp-r0.3.27.so',
  'internal_api': 'openblas',
  'num_threads': 8,
  'prefix': 'libopenblas',
  'threading_layer': 'pthreads',
  'user_api': 'blas',
  'version': '0.3.27'}]

### Context for the issue:

I'm not sure if I understand fully the new rules of type promotion, but I would expect consistency between regular arrays and masked arrays.

[...]

PR #31871, opened 2026-07-05
BUG: MaskedArray arithmetic with Python scalars ignores NEP 50 weak-type

### LINKED ISSUE #31868: BUG: `numpy.ma.MaskedArray` arithmetic with Python scalars differs from `ndarray`
### Describe the issue:

When you use binary arithmetic with Python scalars, `numpy.ma.MaskedArray` gives different results compared to `numpy.ndarray`.

For example, if you add a Python integer to an `int8` `ndarray`, the array keeps its dtype. But if you do the same with a `MaskedArray`, the result changes to `int64`. This also happens with other binary operations like exponentiation.

This occurs even when the `MaskedArray` has no masked elements, so the only difference between the two inputs is that one is an `ndarray` and the other is a `MaskedArray`.

Since `numpy.ma` is documented as a "nearly work-alike replacement" for NumPy and `MaskedArray` is a subclass of `ndarray`, I expected equivalent arithmetic operations to produce the same dtype and values when no elements are masked.

I confirmed this behavior on both NumPy 2.5.1 and the current development branch.

```python
import numpy as np
import numpy.ma as ma

arr = np.array([100], dtype=np.int8)
masked = ma.masked_array([100], dtype=np.int8)

print("ndarray:", arr + 50, (arr + 50).dtype) # ndarray: [-106] int8
print("MaskedArray:", masked + 50, (masked + 50).dtype) # MaskedArray: [150] int64

arr = np.array([2], dtype=np.int8)
masked = ma.masked_array([2], dtype=np.int8)

print("ndarray:", arr ** 10, (arr ** 10).dtype) # ndarray: [0] int8
print("MaskedArray:", masked ** 10, (masked ** 10).dtype) # MaskedArray: [1024] int64
```

### Error message:

```shell

```

2.5.1
3.15.0a6 (tags/v3.15.0a6:15b216f, Feb 11 2026, 12:54:03) [MSC v.1944 64 bit (AMD64)]

### Runtime Environment:

_No response_

_No response_

shipitdev:
Closes #31868

### Problem

MaskedArray arithmetic with Python scalars produces a different dtype
than the same operation on a plain ndarray:
```python
import numpy as np
import numpy.ma as ma

arr = np.array([100], dtype=np.int8)
masked = ma.masked_array([100], dtype=np.int8)

print((arr + 50).dtype)     # int8  (Expected, respects NEP 50)
print((masked + 50).dtype)  # int64 (Bug)
```

This was due to:
`_MaskedBinaryOperation.__call__` calls `getdata()` on both operands before calling the ufunc. For Python scalars, `getdata()` has no
`._data` to grab, so it falls back to `np.array(50)` which creates a 0-d int64 array. Under NEP 50 that's a "strong" type, so the ufunc
promotes int8 to int64. The standalone `power()` function has the same issue.

### Fix

Skip `getdata()` for Python scalars (`int`, `float`, `complex`) in the two call sites, passing them straight to the ufunc where NEP 50
treats them as weak types.

### Test & Enviornment

Added `test_scalar_arithmetic_preserves_dtype_nep50` covering int8,float32, four operators, reversed operand order, and numpy-scalar promotion.

* **OS:** Darwin 25.5.0 (arm64)
* **Python:** 3.14.0
* **NumPy:** 2.6.0.dev0+git20260706.7e381e7
```bash
$ spin test numpy/ma/tests/test_core.py -- -x -q
...
=========================== short test summary info ===========================
SKIPPED [1] numpy/testing/_private/utils.py:2692: Could not determine available memory...
XFAIL numpy/ma/tests/test_core.py::TestMaskedConstant::test_coercion_unicode - See gh-9750
XFAIL numpy/ma/tests/test_core.py::TestMaskedConstant::test_coercion_bytes - See gh-9750
4170 passed, 1 skipped, 2 xfailed in 2.37s
```

### AI Disclosure

AI was used only to strictly format this PR for maintainer readability and ease I also Used an AI assistant to help trace the call chain from `MaskedArray.__add__` through `_MaskedBinaryOperation.__call__` to
`getdata`, and to confirm the root cause.

The diagnosis, fix, and tests were written and verified manually.

### DIFF
--- a/numpy/ma/core.py
+++ b/numpy/ma/core.py
@@ -1064,8 +1064,10 @@ class _MaskedBinaryOperation(_MaskedUFunc):
         Execute the call behavior.
 
         """
-        # Get the data, as ndarray
-        (da, db) = (getdata(a), getdata(b))
+        # Get the data, as ndarray. Python scalars are passed through
+        # unchanged so that ufuncs see them as weak types under NEP 50
+        da = a if isinstance(a, (int, float, complex)) else getdata(a)
+        db = b if isinstance(b, (int, float, complex)) else getdata(b)
         # Get the result
         with np.errstate():
             np.seterr(divide='ignore', invalid='ignore')
@@ -7181,9 +7183,9 @@ def power(a, b, third=None):
     ma = getmask(a)
     mb = getmask(b)
     m = mask_or(ma, mb)
-    # Get the rawdata
-    fa = getdata(a)
-    fb = getdata(b)
+    # Get the rawdata, preserving Python scalars for NEP 50 (gh-31868)
+    fa = a if isinstance(a, (int, float, complex)) else getdata(a)
+    fb = b if isinstance(b, (int, float, complex)) else getdata(b)
     # Get the type of the result (so that we preserve subclasses)
     if isinstance(a, MaskedArray):
         basetype = type(a)
--- a/numpy/ma/tests/test_core.py
+++ b/numpy/ma/tests/test_core.py
@@ -2089,6 +2089,31 @@ class TestMaskedArrayArithmetic:
         assert_equal(test.mask, control.mask)
         assert_equal(a.mask, [0, 0, 0, 0, 1])
 
+    def test_scalar_arithmetic_preserves_dtype_nep50(self):
+        # MaskedArray arithmetic with Python scalars should follow NEP 50
+        marr = masked_array([100], dtype=np.int8)
+        arr = np.array([100], dtype=np.int8)
+
+        assert_equal((marr + 50).dtype, np.int8)
+        assert_equal((marr - 50).dtype, np.int8)
+        assert_equal((marr * 2).dtype, np.int8)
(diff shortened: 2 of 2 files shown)


shipitdev:
I first tried fixing this inside `getdata()` itself (returning Python scalars unchanged). That worked for normal arithmetic but when i ran the whole test_core i noticed it broke all in-place operators (`+=`, `-=`, etc.) . The call-site approach avoids this by only changing the path where scalars go to a ufunc.

shipitdev:
### Relation to prior work

This issue was previously reported in #27029 and #27679. A more comprehensive fix covering both regular and in-place operations exists in #28567. This PR addresses the non-in-place case only, using a call-site approach (bypassing `getdata()` for Python scalars in`_MaskedBinaryOperation.__call__` and `power()`) rather than modifying `getdata()` itself.

If a maintainer feels this is not relevant or conflicting with #28567. I have no problem abandoning this, However if the comprehensive pr stall for any longer, i feel that this narrower fix covers the most common user-facing case.

PR #31872, opened 2026-07-06
DOC: clarify copy semantics of BitGenerator state

### LINKED ISSUE #22337: DOC: clarify copy semantics of random state

In the discussion of the state of the random generators, most specifically https://numpy.org/doc/stable/reference/random/bit_generators/generated/numpy.random.PCG64.state.html#numpy.random.PCG64.state the semantics of the returned state are unclear.  Suppose something like this:
```python
state0 = myBitGenerator.state
# do stuff, including generating random numbers that use the generator
myBitGenerator.state = state0
```
If the `.state` accessor returns a view, or a direct reference to the internal state, `myBitGenerator.state = state0` is a no-op and does not reset the state to what it was.  And what the code should have done was to make a `copy` (deep? shallow?) of it when it first got it.

Say whether one needs to make a copy of the state, and what kind of copy, to reset it in the same session.  And/or say whether the thing you retrieve from `myBitGenerator.state` can change after you've retrieved it.

More generally, at the moment the story for state might differ for each type of `BitGenerator` and so there is no way to discuss them in general.  It would be helpful to be able to discuss them in general, and for them to have a common API.

I suppose another issue is whether state for the legacy `RandomState` could also be discussed at the same time.

Aastha204:
## Summary

Clarifies the copy semantics of `BitGenerator.state` in response to #22337.

- Adds a new **"Saving and restoring state"** section to the BitGenerators docs explaining that `.state` returns a **snapshot**, not a live view of the internal PRNG state
- Documents that subsequent random draws do **not** modify a previously saved state object, so no extra copy is needed for normal save/restore within a session
- Notes when `copy.deepcopy` may be needed (for BitGenerators whose state dict contains NumPy arrays: `MT19937`, `Philox`, `SFC64`)
- Adds `Notes` sections to the `state` docstrings for all BitGenerators (`BitGenerator`, `PCG64`, `PCG64DXSM`, `MT19937`, `Philox`, `SFC64`)
- Updates `RandomState.get_state` / `set_state` docs and adds a cross-link from the legacy random docs

Fixes #22337

## Test plan

- [x] Built docs locally with `spin docs` and verified the new "Saving and restoring state" section renders correctly
- [x] Verified cross-references from `PCG64.state` and other BitGenerator pages link to the new section
- [x] Confirmed no code logic was changed — documentation only

### DIFF
--- a/doc/source/reference/random/bit_generators/index.rst
+++ b/doc/source/reference/random/bit_generators/index.rst
@@ -160,6 +160,47 @@ by a tuple of elements. Here we use a base entropy value and an integer
 Note that the sequences produced by the latter method will be distinct from
 those constructed via `~SeedSequence.spawn`.
 
+.. _random-bit-generator-state:
+
+Saving and restoring state
+--------------------------
+
+All BitGenerators expose a ``state`` property that can be used to capture
+the current PRNG state and restore it later::
+
+    state = bit_generator.state
+    values_before = generator.random(5)
+    bit_generator.state = state
+    values_again = generator.random(5)
+    assert (values_before == values_again).all()
+
+The ``state`` getter returns a **snapshot** of the internal PRNG state. Each
+access returns a new ``dict``; continuing to draw random numbers from the
+generator does **not** update a previously retrieved state object.
+
+Setting ``bit_generator.state`` copies the values from the provided ``dict``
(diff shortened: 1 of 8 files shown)


mattip:
You ignored the PR template. We prefer to talk to humans.

PR #31873, opened 2026-07-06
TYP: fix specialized ufunc bool/int => float promotions

jorenham:
Previously, all ``integer`` and ``bool`` inputs were promoted to ``float64`` in the speciailzed float ufuncs. But in reality, `bool`/`[u]int8` promotes to `float16`, and `[u]int16` promotes to `float32`. 

```
In [1]: np.spacing(np.True_)
Out[1]: np.float16(0.000977)

In [2]: np.spacing(np.int8(1))
Out[2]: np.float16(0.000977)

In [3]: np.spacing(np.int16(1))
Out[3]: np.float32(1.1920929e-07)

In [4]: np.spacing(np.int32(1))
Out[4]: np.float64(2.220446049250313e-16)

In [5]: np.spacing(np.int64(1))
Out[5]: np.float64(2.220446049250313e-16)
```

The solution is, as per usual, to add even more overloads ( :tada: ), special-casing these three promotion rule classes.

---

No AI (although I indirectly realized that this was an issue because of https://github.com/scipy/scipy-stubs/issues/1823, so I suppose that indirectly Fable 5 helped me discover this issue).

### DIFF
--- a/numpy/_core/umath.pyi
+++ b/numpy/_core/umath.pyi
@@ -207,6 +207,10 @@ type _numeric = np.number | np.timedelta64
 type _time = np.datetime64 | np.timedelta64
 type _non_object = _to_number | _time | np.flexible  # np.generic - np.object_
 
+type _as_f64 = np.int64 | np.uint64 | np.int32 | np.uint32
+type _as_f32 = np.int16 | np.uint16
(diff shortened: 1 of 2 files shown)


charris:
Thanks Joren.

PR #31874, opened 2026-07-06
TYP: fix unit-dependent ``timedelta64``/``datetime64`` overloads

jorenham:
The `_TimeUnitSpec` type alias type parameter was unused, leading to incorrect inference for the ``timedelta64`` and ``datetime64`` constructors:

```
reveal_type(np.timedelta64(5, 'D').item())          # int
reveal_type(np.datetime64('2021-01-01', 's').item())  # int
reveal_type(np.datetime64(dt.date(2021, 1, 1), 'ns').item())  # date
int(np.timedelta64(5, 'D'))                          # accepted by checkers
```

Clearly, these are all wrong.

Luckily it's a pretty trivial fix.

---

Claude Fable 5 (high) discovered this bug. but no AI used for the fix.


charris:
Thanks Joren.

PR #31877, opened 2026-07-06
TYP: fix ``random.Generator.choice`` output type for int sequences

jorenham:
Both these cases were typed as `int` in the stubs:

```pycon
In [1]: rng.choice(3)
Out[1]: 0

In [2]: rng.choice([1,2,3])
Out[2]: np.int64(1)
```

The second (`rng.choice([1,2,3])`) case was wrong, and is now correctly annotated to return `np.int_` in the stubs as well.

This matters because `int` and `np.int_` are incompatible types.

---

### DIFF
--- a/numpy/random/_generator.pyi
+++ b/numpy/random/_generator.pyi
@@ -688,17 +688,28 @@ class Generator:
     def permuted[ArrayT: np.ndarray](self, /, x: ArrayLike, *, axis: int | None = None, out: ArrayT) -> ArrayT: ...
 
     #
-    @overload  # >=0d int, size=None (default)
+    @overload  # 0d int, size=None (default)
     def choice(
         self,
         /,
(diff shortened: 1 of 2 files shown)


rkern:
If you just omit that overload, does it not get caught by the `_ArrayLike[ScalarT]` that follows? Or am I expecting too much out of `_ArrayLike[ScalarT]`?

jorenham:
> If you just omit that overload, does it not get caught by the `_ArrayLike[ScalarT]` that follows? Or am I expecting too much out of `_ArrayLike[ScalarT]`?

`_ArrayLike[ScalarT]` only matches array-likes with a known *numpy* scalar type (`ScalarT: np.generic`), so it indeed won't match a sequence of builtin python `int`s.

charris:
Thanks Joren.

PR #31878, opened 2026-07-06
TYP: fix Pyright error for ``piecewise`` with all-scalar ``funclist``

jorenham:
Pyright rejected valid `np.piecewise` calls like `np.piecewise(x, [x < 0, x >= 0], [-1.0, 1.0])`. I'm not sure whether this is because pyright is overly strict, or because mypy and pyrefly are not strict enough.
But either way, I added an overload to ensure that pyright will no longer falsely reject calls with all-scalar `funclist`.

---


charris:
Thanks Joren.

PR #31879, opened 2026-07-06
TYP: fix ``str_.__getitem__`` and ``bytes_.__getitem__``

jorenham:
At runtme the `__getitem__` methods of `str_` and `bytes_` is inherited from `builtins.str` and `builtins.bytes` respectively. But because the stubs declare `__getitem__` on `generic` itself (for good reasons) and the multi-inheritance order, these `__getitem__` are shadowed by `generic` (stubs-only). 

```
In [1]: np.str_("spam")[0]
Out[1]: 's'

In [2]: np.str_("spam")[0:]
Out[2]: 'spam'

In [3]: np.bytes_(b"spam")[0]
Out[3]: 115

In [4]: np.bytes_(b"spam")[0:]
Out[4]: b'spam'

In [5]: np.str_("spam")[()]
Cell In[5], line 1
----> 1 np.str_("spam")[()]

In [6]: np.bytes_(b"spam")[()]
Cell In[6], line 1
----> 1 np.bytes_(b"spam")[()]

```

The fix is to override these `__getitem__` methods on `str_` and `bytes_` using the `__getitem__` typeshed signatures of `str` and `bytes`.

---


charris:
Thanks Joren.

PR #31880, opened 2026-07-06
TYP: fix ``std``/``var`` output type for complex input

jorenham:
The `np.std`/`np.var` functions and `ndarray.std`/`ndarray.var` methods incorrectly returned complex dtypes for complex input. Because at runtime that would result in real output:

```
In [1]: c = np.array([1j, 1])

In [2]: np.std(c)
Out[2]: np.float64(0.7071067811865476)

In [3]: c.std()
Out[3]: np.float64(0.7071067811865476)
```

Miraculously, the fix didn't require adding new overloads this time. It even got rid of two!

---


charris:
Thanks Joren.

PR #31881, opened 2026-07-06
TYP: fix ``random.RandomState.get_state`` conflicting overloads

jorenham:
Calling `random.RandomState.get_state()` (i.e. without passing `legacy`) incorrectly resulted in a `dict[str, Any]` output type because the first overload incorrectly has a `legacy: Literal[False] = False` default set. At runtime, it defaults to `True`.

According to Claude Fable 5 backporting this is a good idea because "state pickling is a common use of this API". I never use this API myself, so i'm not sure if that's true. So if you think it isn't, then just remove the backport-canditate label.

---


charris:
Thanks Joren.

PR #31882, opened 2026-07-06
TYP: fix swapped ``linalg.slogdet`` dtypes of ``sign``/``logabsdet``

jorenham:
In short: `linalg.slogdet(...).sign` can be complex and `logabsdet` can only be real, but it was the other way around in the stubs.

---


charris:
Thanks Joren.

PR #31883, opened 2026-07-06
TYP: fix ``datetime64`` unit conversion

jorenham:
The documented approach to convering the unit of a `np.datetime64` is by passing it to the constructor together with the desired new unit, e.g. `np.datetime64(my_dt, "s")`. 

The stubs incorrectly rejected this use-case; so here's the fix.

---


charris:
Thanks Joren.

PR #31884, opened 2026-07-06
BUG: centralized helper for output coerce and na_object in stringdtype operations

ngoldbaum:
### PR summary
Related to #31825 but issuing it as a separate PR because the two changes don't overlap at all.

We currently have custom logic for determining the correct `na_object` and `coerce` value when operations mix `StringDType` instances. In some cases this can even lead to crashes. The following script segfaults for me because the `na_object` in the fill character array is not being handled correctly.

```python
Python 3.14.5 (main, May 22 2026, 08:10:06) [Clang 21.0.0 (clang-2100.0.123.102)] on darwin
>>> import numpy as np
>>> from numpy.dtypes import StringDType
>>> print(np.__version__)
2.5.1
>>> arr = np.array(["a"], dtype="T")
>>> fill = np.array("*", dtype=StringDType(na_object="*"))
>>> print(np.strings.center(arr, 5, fill))
[1]    87011 segmentation fault  python
```

This centralizes the logic for handling the `na_object` and `coerce` argument in ufunc loops and in the comparison machinery into one set of shared helpers.

Also adds tests for this behavior, including the case that leads to a crash above.

#### AI Disclosure
I found this issue using an AI model. I also used an AI model to help with the tests and the approach for the fix.

### DIFF
--- a/numpy/_core/src/multiarray/stringdtype/dtype.c
+++ b/numpy/_core/src/multiarray/stringdtype/dtype.c
@@ -211,6 +211,62 @@ stringdtype_compatible_na(PyObject *na1, PyObject *na2, PyObject **out_na) {
     return 0;
 }
 
+// Combine the na_object and coerce attributes of the string descriptors in
+// descrs, ignoring descriptors of other dtypes. The na_objects must be
+// compatible, and coercion is only enabled if it is enabled for all string
+// descriptors. Returns -1 with an error set if two descriptors have
+// incompatible na_objects. out_na_object (a borrowed reference) and
+// out_coerce may be NULL if only the compatibility check is needed.
+NPY_NO_EXPORT int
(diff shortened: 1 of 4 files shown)


ngoldbaum:
@mhvk hope you don't mind a ping to look this over. #31825 as well if you have the bandwidth 🙈.

ngoldbaum:
@Ethan0456 you just put in a PR for a similar issue. Maybe you want to take a look at this?

Ethan0456:
@ngoldbaum LGTM! thanks for the clear repro. I reproduced the `center` segfault on `2.5.1`; fixed on this branch with `['**a**']`. The shared helpers match the intended rules, and scoping resolve to `nin` avoids walking `NULL` outs on ops like `center`. New mixed-instance tests look right and passed locally.

charris:
Thanks Nathan.

PR #31885, opened 2026-07-06
TYP: fix int promotion for ``linalg.trace``

jorenham:
Same story as #31876, but for `numpy.linalg.trace`.

---


charris:
Thanks Joren.

PR #31886, opened 2026-07-06
TYP: fix ``tri{l,u}`` shape-typing for 1d input

jorenham:
`np.tril` and `np.triu` return a 2d array for 1d input at runtime, but the stubs returned a 1d array in this case. 

---


charris:
Thanks Joren.

PR #31887, opened 2026-07-06
BUG: preserve the sign of zero in np.sign (gh-29485)

momomuchu:
Fixes #29485.

`np.sign` discarded the sign of a signed zero: `np.sign(np.float64(-0.0))` returned `+0.0` instead of `-0.0`, and likewise for float16/float32/longdouble and for each component of a complex value. The three `sign` loops in `_core/src/umath/loops.c.src` (the templated real loop, `HALF_sign`, and the complex loop) hard-coded a `0` result for a zero-magnitude input, dropping the sign bit.

Each of those branches now returns the input zero unchanged, so the sign bit is preserved. Non-zero results are unchanged, and for complex each component keeps its own sign, matching the values discussed in the issue.

On the Array API: the standard's `sign` special cases specify `+0` for a zero input, so preserving the sign is a deliberate deviation. I added the two affected conformance cases to `tools/ci/array-api-xfails.txt` with a comment, next to the existing deliberate-deviation entries.

Tests: signbit-based cases added to `TestSign` for the real and complex dtypes (a plain value comparison cannot see the bug since `-0.0 == 0.0`). Release note added under `doc/release/upcoming_changes/`.

Object dtype is left as-is per the discussion, and the separate `sign(complex(inf, nan))` behavior is out of scope here.


jorenham:

PR #31890, opened 2026-07-06
BUG: preserve the sign of zero in np.sign (gh-29485)

momomuchu:
### PR summary

Fixes #29485. This supersedes #31887. I opened that one from the `gh` CLI without realizing NumPy has a PR template, so I am reopening cleanly with the template and disclosure filled in.

`np.sign` was dropping the sign of a signed zero: `np.sign(-0.0)` returned `+0.0` instead of `-0.0`, for float16/float32/float64/longdouble and for each component of a complex value. The three sign loops in `_core/src/umath/loops.c.src` returned a literal `0` for a zero input, throwing away the sign bit. They now return the input zero itself, so the sign is preserved. Non-zero results are unchanged.

The Array API standard specifies `+0` for sign of a zero, so this is a deliberate deviation; the two affected conformance cases are in `tools/ci/array-api-xfails.txt` next to the other deviations, plus a release note. Tests use `signbit` since `-0.0 == 0.0` hides the bug. Object dtype is out of scope.

Hi, I'm Mohamed. I've used NumPy across a lot of projects for a long time, so it felt like a nice one to start contributing back on.

#### AI Disclosure

I use AI tooling in my workflow, and it was genuinely useful here, mostly for tracking down where in the codebase the bug actually came from. I went through the change, I understand it, and I own it.


tools/ci/array-api-xfails.txt
```diff
@@ -67,6 +67,10 @@ array_api_tests/test_special_cases.py::test_iop[__ifloordiv__(x1_i is -infinity
 array_api_tests/test_special_cases.py::test_iop[__ifloordiv__(isfinite(x1_i) and x1_i > 0 and x2_i is -infinity) -> -0]
 array_api_tests/test_special_cases.py::test_iop[__ifloordiv__(isfinite(x1_i) and x1_i < 0 and x2_i is +infinity) -> -0]
 
+# np.sign preserves the sign of zero (gh-29485); the Array API standard specifies +0.
+array_api_tests/test_special_cases.py::test_unary[sign((x_i is -0 or x_i == +0)) -> 0]
+array_api_tests/test_special_cases.py::test_unary[sign((real(x_i) is -0 or real(x_i) == +0) and (imag(x_i) is -0 or imag(x_i) == +0)) -> 0 + 0j]
```
ngoldbaum, line 72: I don't see any discussion about the array API in the issue and IMO this change is a non-starter.

momomuchu:
Right, not worth a conformance deviation. Closing.

PR #31893, opened 2026-07-07
ENH: Support DLPack for scalars by converting to 0-D array

### LINKED ISSUE #27137: BUG: array api incompatibility in output of reductions (sum/prod/min/max/...)
### Describe the issue:

reduction (e.g. xp.sum) are expected to produce arrays, but in numpy they produce scalars, that don't follow array api specifications (e.g. I can't use dlpack on those).

In numpy v1, np.array_api correctly returned arrays in this case

```python
import numpy as np
x = np.zeros([3, 4])

type(x.__array_namespace__().sum(x))
# outputs: numpy.float64 expected: numpy.ndarray
```
```


### Error message:

_No response_

2.0.1
3.11.9 (main, Apr  6 2024, 17:59:24) [GCC 11.4.0]


### Runtime Environment:

<irrelevant>

### Context for the issue:

_No response_

### PR DESCRIPTION
**Pull request by MaanasArora**
Closes #27137 by adding `__dlpack__` and `__dlpack_device__` methods to NumPy scalars. `scalar.__dlpack__` currently delegates to `ndarray.__dlpack__` by converting the scalar to a 0-D array. This is a very minimal solution - a copy is fairly natural in this context, so this workaround seemed natural as well. ping @seberg - thanks!

I have to check if we can add additional handling for the `copy=` kwarg - that seemed to be the main case where the array and scalar implementations might diverge in practice (we are _always_ copying). Perhaps with certain conditions we can add validation for `copy`. I also have to check if I missed some other nuances within the arguments, and needs a release note, but otherwise, this is ready for review!

#### AI Disclosure
I used an agent to search for DLPack's implementation in more detail, to check for errors in my code, and to write a few tests (the `independent_copy` one, mainly).

### DIFF
--- a/numpy/_core/src/multiarray/dlpack.c
+++ b/numpy/_core/src/multiarray/dlpack.c
@@ -568,6 +568,27 @@ array_dlpack_device(PyArrayObject *self, PyObject *NPY_UNUSED(args))
     return Py_BuildValue("ii", device.device_type, device.device_id);
 }
 
+// for scalar __dlpack__ we create a 0-dim array and call the array version
+NPY_NO_EXPORT PyObject *
+gentype_dlpack(PyObject *self,
+        PyObject *const *args, Py_ssize_t len_args, PyObject *kwnames)
+{
+    PyObject *array = PyArray_FromScalar(self, NULL);
+    if (array == NULL) {
+        return NULL;
+    }
+
+    PyObject *result = array_dlpack((PyArrayObject *)array, args, len_args, kwnames);
+    Py_DECREF(array);
+    return result;
+}
+
+NPY_NO_EXPORT PyObject *
+gentype_dlpack_device(PyObject *NPY_UNUSED(self), PyObject *NPY_UNUSED(args))
+{
+    return Py_BuildValue("ii", kDLCPU, 0);
+}
+
 NPY_NO_EXPORT PyObject *
 from_dlpack(PyObject *NPY_UNUSED(self),
         PyObject *const *args, Py_ssize_t len_args, PyObject *kwnames)
--- a/numpy/_core/src/multiarray/dlpack.c
+++ b/numpy/_core/src/multiarray/dlpack.c
@@ -482,7 +482,7 @@ device_converter(PyObject *obj, DLDevice *result_device)
 
 
 NPY_NO_EXPORT PyObject *
-array_dlpack_impl(PyArrayObject *self,
+array_dlpack_int(PyArrayObject *self,
     PyObject *stream, PyObject *max_version, DLDevice *result_device,
     NPY_COPYMODE copy_mode)
 {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by MaanasArora** [comment]
I Just pushed an update to reject `copy=False` explicitly and pass `NPY_COPY_IF_NEEDED` to the array delegate (to avoid a double copy). I also refactored to use the same implementation function for both, not sure if nice, so I can revert and use a more manual check if needed. (I wasn't entirely sure if we needed to call the array's Python function, but it seems _more_ complicated.) I think this should be ready for review now!

**Comment by ikrommyd** [comment]
Added myself to the reviewers here as its one of the things I wanted to review when I find the next time slot.

File: numpy/_core/src/multiarray/dlpack.c [hunk 1]
```diff
@@ -568,6 +578,55 @@ array_dlpack_device(PyArrayObject *self, PyObject *NPY_UNUSED(args))
...
+            {"$dl_device", &device_converter, &result_device},
+            {"$copy", &PyArray_CopyConverter, &copy_mode})) {
+        return NULL;
+    }
+    if (result_device.device_type != kDLCPU ||
+        result_device.device_id != 0) {
+        PyErr_SetString(PyExc_BufferError,
+            "Cannot export a scalar to DLPack on a non-CPU device.");
+        return NULL;
+    }
```
ikrommyd: I think this is dead code. `device_converter` should already guarantee that `device_type` will be `kDLCPU`. Unless you wanted it for a different error message.
```suggestion
```

numpy/_core/tests/test_dlpack.py
ikrommyd, line 1: I'd add testing for the `dl_device` argument and for the `copy=True` case which if I'm seeing correctly are currently missing.

ikrommyd (approved):
I think this is good and clean! Only left two minor comments.

ikrommyd:
@seberg based on your comment https://github.com/numpy/numpy/issues/27137#issuecomment-4934701129
I'd like to ask, should any of these work? Cause they don't right now. So thanks for reminding me to consider bfloat16 for example too.
```python
In [1]: import ml_dtypes

In [2]: ml_dtypes.bfloat16(4).__dlpack__()
BufferError                               Traceback (most recent call last)
Cell In[2], line 1
----> 1 ml_dtypes.bfloat16(4).__dlpack__()

BufferError: DLPack only supports signed/unsigned integers, float and complex dtypes (or dtypes registered by third-party packages).

In [3]: from numpy_quaddtype import QuadPrecDType, QuadPrecision

In [4]: QuadPrecision(4).__dlpack__()
BufferError                               Traceback (most recent call last)
Cell In[4], line 1
----> 1 QuadPrecision(4).__dlpack__()

BufferError: DLPack only supports signed/unsigned integers, float and complex dtypes (or dtypes registered by third-party packages).
```

MaanasArora:
#32029 supersedes and incorporates part of this solution now, so closing!

Issue #12389, opened 2018-11-14
performance of numpy.roll with empty axes or zero shift(s)


grlee77:
`numpy.roll` (and functions that use it such as `numpy.fft.fftshift`, `numpy.fft.ifftshift`) are inefficient when `shift=0` (or `axis=()`). In that case it would be possible to just return the input array rather than allocating a new empty array and then copying the old array into it. 

Obviously, the  downside to changing this would be potentially breaking existing code that relies on `numpy.roll` always returning a copy (although I don't see any guarantee to that effect in the docstring). 

I encountered this when using a "centered FFT" wrapper I had created for `numpy.fft.fftn` that had calls to `fftshift`/`ifftshift`, but where I sometimes passed `axes=()` when I wanted an uncentered FFT. I noticed that the uncentered case was not faster and this was due to this unecessary copy being made in `np.roll`.  It is not a big problem for me to avoid calling `fftshift` or `ifftshift` when the axes are empty in my centered FFT function, but thought this might be of more general interest.

If there is interest, I can make a PR with an implementation. If this is considered too much micro-optimization, feel free to close the issue.

### Reproducing code example:

```python
import numpy as np
r = np.empty((256, 256, 256))
%timeit np.roll(r, (), axis=())
```
`54.2 ms ± 3.78 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)`

The time is reduced to 10-20 **microseconds** if a copy is not made.

seberg:
I tend to be wary with such optimizations, they can lead to strange bugs in the opposite case where most of the time you have a copy and suddenly not. And then you have incorrect in-place behaviour that can silently break your code. There are only very few exceptions of this rule after all `ravel`/`reshape` being the most prominent ones (and frankly maybe even the only ones).

I personally would like to keep it at few exceptions. An interesting option could be an `out` argument, although in the non-trivial overlap case, we would have to be very careful.

PR #31894, opened 2026-07-07
PERF: fast-path no-op roll cases

KRRT7:
### PR summary

Fast-paths two no-op `np.roll` cases:

- return the input directly for `axis=()`
- avoid the generic slice-product roll loop when all normalized shifts are zero, while preserving copy semantics

Addresses gh-12389.

Benchmarks from `spin bench --compare main HEAD -t bench_manipulate.DimsManipulations`:

- `time_roll_empty_axes`: 59-64% faster
- `time_roll_zero_shift`: 16-18% faster

#### AI Disclosure

LLM / Harness: hybrid.

### DIFF
--- a/numpy/_core/numeric.py
+++ b/numpy/_core/numeric.py
@@ -1309,19 +1309,28 @@ def roll(a, shift, axis=None):
         if broadcasted.ndim > 1:
             raise ValueError(
                 "'shift' and 'axis' should be scalars or 1D sequences")
+        if len(axis) == 0:
+            return a
+
         shifts = dict.fromkeys(range(a.ndim), 0)
         for sh, ax in broadcasted:
             shifts[ax] += int(sh)
 
         rolls = [((slice(None), slice(None)),)] * a.ndim
+        roll_needed = False
         for ax, offset in shifts.items():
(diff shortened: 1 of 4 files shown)


numpy/_core/numeric.py
```diff
@@ -1309,19 +1309,28 @@ def roll(a, shift, axis=None):
...
         shifts = dict.fromkeys(range(a.ndim), 0)
         for sh, ax in broadcasted:
             shifts[ax] += int(sh)
 
         rolls = [((slice(None), slice(None)),)] * a.ndim
+        roll_needed = False
         for ax, offset in shifts.items():
             offset %= a.shape[ax] or 1  # If `a` is empty, nothing matters.
             if offset:
+                roll_needed = True
```
eendebakpt, line 1324: This adds a (tiny) bit of overhead for all cases. How important is the case where no roll is needed? Or where axis is empty?

ikrommyd: would show up as a slower benchmark if it was noticeable no?

KRRT7: either way, I have more PRs that build on top of this and recover whatever (imo, negligible) losses there might be.

PR #31896, opened 2026-07-07
TYP: reduce ufunc stubs spaghetti using shaped array type aliases

jorenham:
There was a lot of duplication in the shaped array type declarations in the ufuncs stubs and type-tests, so I introduced some type-aliases to clean that up.

---

No AI.


charris:
Thanks Joren.

PR #31897, opened 2026-07-07
PERF: fast-path large order diff

KRRT7:
### PR summary

Add a fast path for `np.diff` when `n` is at least the effective axis length. In that case the result is empty, so numeric and non-object dtypes can avoid repeated differencing of already-empty arrays. Object arrays keep the existing path so Python-level subtraction behavior is preserved.

ASV quick comparison against `main`:

```
bench_function_base.Diff.time_large_n((10,), 10):       21.0 us -> 16.5 us
bench_function_base.Diff.time_large_n((10, 10), 10):    41.3 us -> 13.9 us
bench_function_base.Diff.time_large_n((10,), 1000):      346 us -> 12.4 us
bench_function_base.Diff.time_large_n((10, 10), 1000):   481 us -> 15.0 us
```

#### AI Disclosure

LLM / Harness: hybrid.

### DIFF
--- a/benchmarks/benchmarks/bench_function_base.py
+++ b/benchmarks/benchmarks/bench_function_base.py
@@ -120,6 +120,20 @@ class Percentile(Benchmark):
         np.percentile(self.o, [25, 75])
 
 
+class Diff(Benchmark):
+    params = [
+        [(10,), (10, 10)],
+        [10, 1000],
+    ]
+    param_names = ["shape", "n"]
+
+    def setup(self, shape, n):
+        self.arr = np.arange(np.prod(shape)).reshape(shape)
+
+    def time_large_n(self, shape, n):
+        np.diff(self.arr, n=n)
+
+
 class Select(Benchmark):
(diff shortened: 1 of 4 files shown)


mattip:
This is the third such PR. If we accept all of them, what will be the overall impact on time to run the benchmarking suite? I am a bit hesitant to add so many new benchmarks. Also, like the other one, you need to benchmark the non-empty paths.

KRRT7:
> Also, like the other one, you need to benchmark the non-empty paths.

internally I run the end to end benchmarks, however I'll start posting all of them.


> what will be the overall impact on time to run the benchmarking suite?

overall shouldn't be much, but I can drop the benchmarks, but then in the future I'll be asked to provide benchmarks....

seberg:
Honestly, this type of proposal is a waste of maintainer time unless you can motivate a very real _actual_ use-case where it matters.
If you wish us to review relevant PRs, it would be good to be mindful and think about this before opening things.

PR #31903, opened 2026-07-07
TYP: specialized ``copysign``, ``heaviside``, ``logaddexp[2]``, and ``nextafter`` ufuncs

jorenham:
Following #31851, this specialized the `floating * floating -> floating`  ufuncs. As you can see, this is already quite a bit more involved than the previous binary ufuncs because the dtype promotions depend on both inputs simultaneously. Even without shape-typing support added yet the `__call__`  method here already counts 32 overloads :sweat_smile:. Hopefully the comments, type alias names, and the type-tests, help make it clear what's going on a bit better.

The remaining ufuncs will be even more complex (they accept more dtypes), so there's something to look forward to :P.

---

No AI.


charris:
Thanks Joren.

PR #31905, opened 2026-07-08
BUG: fix CPU feature env diagnostic buffer overruns

stratakis:
### PR summary

This is a follow up to gh-30877.

I've had on the back on my mind to verify and check some other issues I saw in the file and finally managed to get to it.

The original PR fixed a buffer overrun in CPU baseline validation. This PR fixes two similar issues in `npy__cpu_check_env`.

The two issues fixed:

- nonexistent features: an accepted maximum-length environment variable value could write the terminating NUL one byte past the `nexist` stack buffer
- unsupported dispatched features: repeated unsupported feature names could overflow the `notsupp` diagnostic buffer, which was sized from `NPY_WITH_CPU_DISPATCH` rather than from the accepted environment variable length.

Both fixes avoid the trailing-space-and-trim pattern when building the diagnostic strings.

Noting though that the fix for the `notsupp` buffer increases stack usage in `npy__cpu_check_env` because the buffer is now bounded by `NPY__MAX_VAR_LEN` instead of  `sizeof(NPY_WITH_CPU_DISPATCH) + 1`. On my x86_64 laptop, `-fstack-usage` reports the function growing from 2224 bytes to 3184 bytes, a 960 bytes increase. The function is non-recursive and only runs while parsing CPU feature environment variables during import, however if the size icrease is deemed unacceptable I can change the approach to something like heap allocation, it will require more code changes though.

These tests are mainly sanitizer regressions, in non-sanitized builds, the previous out-of-bounds writes might not crash.

#### AI Disclosure
I used codex with the gpt 5.5 model. Guided to create the tests, (hence the tests are AI generated) and to review the C code changes, also to verify any incompatibilities (where it verified the stack increase) and to create small reproducers with the same logic for me to actually verify the issues exist.

### DIFF
--- a/numpy/_core/src/common/npy_cpu_features.c
+++ b/numpy/_core/src/common/npy_cpu_features.c
@@ -308,8 +308,11 @@ npy__cpu_check_env(int disable, const char *env) {
         int feature_id = npy__cpu_dispatch_fid(feature);
         if (feature_id == 0) {
             int flen = strlen(feature);
+            if (nexist_cur != nexist) {
+                *nexist_cur++ = ' ';
+            }
             memcpy(nexist_cur, feature, flen);
-            nexist_cur[flen] = ' '; nexist_cur += flen + 1;
+            nexist_cur += flen;
             goto next;
         }
         // check if the feature supported by the running machine
@@ -335,7 +338,6 @@ npy__cpu_check_env(int disable, const char *env) {
 
     *nexist_cur = '\0';
     if (nexist[0] != '\0') {
-        *(nexist_cur-1) = '\0'; // trim the last space
         if (PyErr_WarnFormat(PyExc_ImportWarning, 1,
             "%sYou cannot %s CPU features (%s), since "
             "they are not part of the dispatched optimizations\n"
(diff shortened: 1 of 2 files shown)


numpy/_core/tests/test_cpu_features.py
```diff
@@ -335,6 +352,28 @@ def test_impossible_feature_enable(self):
...
+    def test_repeated_unavailable_feature(self, action):
+        """
+        Test that repeated unavailable features can be processed without
+        overflowing the diagnostic buffer.
+        """
+        if self.UNAVAILABLE_FEAT is None:
+            pytest.skip("There are no unavailable features to test with")
+
+        bad_feature = self.UNAVAILABLE_FEAT
+        feature_repeats = min(64, 1024 // (len(bad_feature) + 1))
```
ngoldbaum: Why cap at 64? If your fix is correct it should work with just `feature_repeats = 1024 // (len(bad_feature) + 1)`. We have ASan CI running so we'd even see a heap buffer overflow in our tests on the unpatched code.

numpy/_core/src/common/npy_cpu_features.c
ngoldbaum: Totally unrelated but while we're here, `var_len` is a `size_t`, so I'm pretty sure we want `%zu` here for the second argument instead of `%d`.

ngoldbaum (approved):
Just two very minor nits, otherwise this looks good to me. The increased stack size doesn't matter I think.

I'll probably merge this soon but let's ping @seiko2plus since he originally wrote this code.

stratakis:
Rebased to address the comments.

ngoldbaum:
Thanks @stratakis!

PR #31852, opened 2026-07-03
DOC: cross-reference np.ma.correlate from np.ma.convolve for propagate_mask

shoumikchakravarty-dev:
### PR summary
This is an improvement PR. I have added an Examples section to the documentation for np.ma.convolve showing the propagate_mask behavior (mirroring the existing examples on its sibling np.ma.correlate). Doc-only; doctest-verified locally.

### First time committer introduction
No. Previous Contribution https://github.com/numpy/numpy/pull/31802

#### AI Disclosure
An AI assistant (Anthropic's Claude, via the Claude desktop app) was used to locate a public function whose docstring lacked an Examples section, draft the example, and verify it passes as a doctest. I reviewed and understand each line; the change is documentation-only- a new Examples section added to np.ma.convolve. This PR summary and introduction were written by me.

PR #31802, opened 2026-06-30
DOC: fix duplicated-word and grammar typos in C-API reference

shoumikchakravarty-dev:
### PR summary
Removes duplicated-word and grammar typos in the C-API reference docs (array.rst, datetimes.rst) and a duplicated word in an error-message string (`numpy/_core/__init__.py`).

### First time committer introduction
Hello! I'm new to the NumPy community. I use Python in my project work and have been learning NumPy for some time. I've also served as a technical reviewer for programming books, including a Python title. I'd like to start spending time on open source contribution, and this is my first PR toward that.

#### AI Disclosure
An AI assistant (Anthropic's Claude, via the Claude desktop app) was used to scan
the documentation and source for duplicated-word and grammar typos and to suggest
the corrected wording. I reviewed and verified each change myself. No program logic
was AI-generated — the changes are limited to documentation text and one duplicated
word in an error-message string. This PR summary and my introduction were written by me.

PR #31907, opened 2026-07-08
DOC: add Examples section to np.ma.polyfit [skip actions]

shoumikchakravarty-dev:
### PR summary
Adds a proper Examples section to np.ma.polyfit and converts its docstring from the doc_note wrapper pattern to a self-contained standalone docstring, matching the style already used by np.ma.ediff1d, np.ma.convolve, and other functions in the module.

No. Previous contributions
#31802 
#31852 

#### AI Disclosure
An AI assistant (Anthropic's Claude, via the Claude desktop app) was used to locate np.ma.polyfit as a function whose docstring lacked masked-array-specific examples, draft the standalone docstring and three examples, and verify they pass as doctests

### DIFF
--- a/numpy/ma/extras.py
+++ b/numpy/ma/extras.py
@@ -2297,5 +2297,3 @@ def polyfit(x, y, deg, rcond=None, full=False, w=None, cov=False):
         return np.polyfit(x[not_m], y[not_m], deg, rcond, full, w, cov)
     else:
         return np.polyfit(x, y, deg, rcond, full, w, cov)
-
-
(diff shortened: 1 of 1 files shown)


mattip:
Nice. I like it that you only show the mask effect, without diving into what polyfit does since that is covered elsewhere. There is a linting error: too many new lines at the end of the file, since you removed a line but not the newlines before it.

shoumikchakravarty-dev:
@mattip - Good catch. I have updated the changes. please review further.

mattip:
Thanks @shoumikchakravarty-dev

PR #31908, opened 2026-07-08
Update x86-simd-sort subproject (5adb334 → fa944ef)

r-devulap:
### PR summary

Porting latest x86-simd-sort changes to NumPy: summary of relevant changes: 

- Add nans_last parameter to all sort/select APIs to control NaN  placement (end or start) independently of sort direction (https://github.com/numpy/x86-simd-sort/pull/235) 
- Add descending parameter to argselect across all backends. 

#### AI Disclosure
Copilot authored: https://github.com/numpy/x86-simd-sort/pull/231


MaanasArora (approved):
Great, thank you @r-devulap, clearly no issues here, except I'm not sure if this shouldn't be backported so we can also backport the performance fix for descending sorts? Let's check with @seberg (sorry for the ping!)

seberg:
I am fine with backporting it as well to consider it a perf bugfix, as it seems rather safe.  That said, also not too worried if we don't/decide not to.
But I guess this should be very safe as it would have to be a very fun bug in the simd-sort repo to affect us without even using the new paths :).

PR #31911, opened 2026-07-08
TYP: fix output dtype of ``bincount`` with ``weights``

jorenham:
If `weights` are passed to `bincount`, the returned dtype is `float64`, not `intp`:

```pycon
>>> np.bincount(np.array([0, 1, 1, 2]), weights=np.array([.5, 1, 1.5, 2])).dtype
dtype('float64')
>>> np.bincount(np.array([0, 1, 1, 2]), weights=np.array([1, 2, 3, 4])).dtype
dtype('float64')
```

This fixes the stubs to match this runtime behavior

---

(more stub fixes incoming, btw)


charris:
Thanks Joren.

PR #31912, opened 2026-07-08
TYP: fix return type of ``strings.[r]partition``

jorenham:
Unlike `np.char.[r]partition`, which returns a (single) array, `np.strings.[r]partition` returns a 3-tuple or arrays:

```pycon
>>> r = np.strings.partition(np.array(["a b", "c d"]), " ")
>>> type(r), len(r), type(r[0])
(<class 'tuple'>, 3, <class 'numpy.ndarray'>)
```

The docstring also confirms this.

This fixes the stubs, that previously return (single) arrays here, to match this runtime behavior.

---

(more stub fixes incoming, btw)


charris:
Thanks Joren.

PR #31914, opened 2026-07-08
TYP: ``rec.from{arrays,records}`` should accept ``names=`` without ``formats=``

jorenham:
Previously, passing `names`  without `formats` to `np.rec.fromarrays` and `np.red.fromrecords` was rejected, even though it's valid at runtime:

```
In [1]: np.rec.fromarrays([np.array([1., 2.]), np.array([3, 4])], names="a,b")
Out[1]: 
rec.array([(1., 3), (2., 4)],
          dtype=[('a', '<f8'), ('b', '<i8')])
```

---

(more stub fixes on the way)


charris:
Thanks Joren.

PR #31915, opened 2026-07-08
TYP: single-operand ``nditer`` should accept nested ``op_flags``

jorenham:
At runtime you can do

```
In [1]: np.nditer(np.array([1,2,3]), op_flags=[["readwrite"]])
Out[1]: <numpy.nditer at 0x72bd5f5ba790>
```

but the stubs rejected this.

---

(more stub fixes on the way)


charris:
Thanks Joren.

PR #31916, opened 2026-07-08
TYP: 2nd ``histogramdd`` return type should be ``list``, not ``tuple``

jorenham:
because

```
In [1]: tuple(map(type, np.histogramdd([1,2,3])))
Out[1]: (numpy.ndarray, list)
```
---


charris:
Thanks Joren.

PR #31917, opened 2026-07-08
TYP: fix ``complex_sort`` for builtin sequences

jorenham:
`complex_sort` accepts builtin scalar sequences at runtime :

```
In [1]: np.sort_complex([1,2])
Out[1]: array([1.+0.j, 2.+0.j])
```

But the stubs rejected this, so this adds an additional overload to handle this case.

---


charris:
Thanks Joren.

PR #31918, opened 2026-07-08
TYP: override ``ma.MaskedArray.astype`` to return ``MaskedArray``

jorenham:
`ma.MaskedArray.astype` used to be inherited from `ndarray.astype` in the stubs, causing it to return a `ndarray` kind instead of `MaskedArray`. The only way to fix this is by overriding `asarray` in `MaskedArray` (Python has no support for higher kinded typing and `typing.Self` cannot be used as a generic type).

---


charris:
Thanks Joren.

PR #31919, opened 2026-07-08
TYP: fix single-arg ``ma.where`` return type

jorenham:
Single-arg `np.ma.where` calls return a tuple of `ndarray`, not `MaskedArray`:

```
In [1]: np.ma.where(np.array([True, False]))
Out[1]: (array([0]),)

In [2]: np.ma.where(np.array([True, False]).view(np.ma.MaskedArray))
Out[2]: (array([0]),)
```

---


charris:
Thanks Joren.

PR #31920, opened 2026-07-08
TYP: fix ``ma.count_masked`` return type when ``out=None`` (default)

jorenham:
`np.ma.count_masked` with `axis=None` (the default) returns a bare `np.intp` at runtime, but the stubs incorrectly used `NDArray[np.intp]` as output type in that case.

---

(only a couple of stub fixes left now)


charris:
Thanks Joren.

PR #31921, opened 2026-07-08
TYP: Backport multiple typing fixes 2.

charris:
Backports of:

TYP: fix ``ma.count_masked`` return type when ``out=None`` (default) (#31920)
TYP: fix single-arg ``ma.where`` return type (#31919)
TYP: fix ``complex_sort`` for builtin sequences (#31917)
TYP: 2nd ``histogramdd`` return type should be ``list``, not ``tuple`` (#31916)
TYP: single-operand ``nditer`` should accept nested ``op_flags`` (#31915)
TYP: ``rec.from{arrays,records}`` should accept ``names=`` without ``formats=`` (#31914)
TYP: fix return type of ``strings.[r]partition`` (#31912)
TYP: fix output dtype of ``bincount`` with ``weights`` (#31911)

No AI.


jorenham:
Thanks Chuck

PR #31924, opened 2026-07-08
TYP: some fixes for the latest binary floating ufuncs specialization.

jorenham:
The specialized ufunc type introduced in #31903 had some issues:

1. The `builtin.float` overloads were too high up, causing `float64` (which sublcasses `float`) to also be caught, leading to incorrect output dtypes in some cases.
2. The `.outer` tests were missing (multi-line typo).
3. Not all 4 scalar/array promotor/promoteee overload combinations were present, only 2 of those. But since I did not want to double the 32 overloads that we already had, I instead decided to ignore pyright's warnings about incompatible overlapping overloads and "grouped" the scalar/array cases together. To my surprise, neither mypy, pyright, or pyrefly, seemed to care about this in practice: they all inferred exactly the same for each of the very many testcases. So I think we're fine to ignore this overlapping-overload warning here.

---

No AI.


charris:
Thanks Joren.

PR #31929, opened 2026-07-09
TYP: minor ``random`` typing fixes

jorenham:
This fixes:

-  ``random.RandomState.random_integers`` scalar input returns ``np.long``, not a builtin ``int``
- ``random.Generator.permuted`` doesn't preserve array subtypes

I don't think this is worth backporting (legacy/niche)

---

No AI used for writing any code. Claude Fable 5 spotted these issues.


rkern (approved):

charris:
Thanks Joren.

PR #31935, opened 2026-07-09
TYP: specialized ``a[rc]tan2`` and ``hypot`` ufuncs

jorenham:
These two ufuncs technically accept `object_` dtypes, but in that case it delegates to `arctan2`/`hypot` methods of the wrapped python objects, which in most cases will lead to an (IMO confusing) error:

```
In [1]: np.hypot(np.array(1, np.object_), 1)
AttributeError                            Traceback (most recent call last)
Cell In[1], line 1
----> 1 np.hypot(np.array(1, np.object_), 1)

AttributeError: 'int' object has no attribute 'hypot'
```

(it took me a while to realize that it wasn't `np` that was shadowed by an `int` here)

So because of this type-unsafety, I instead recycled the existing `floating * floating -> floating` ufunc specialization here. 

---

No AI.

### DIFF
--- a/numpy/__init__.pyi
+++ b/numpy/__init__.pyi
@@ -401,6 +401,8 @@ from numpy._core.umath import (
     arcsinh as asinh,
     arctan,
     arctan as atan,
+    arctan2,
+    arctan2 as atan2,
     arctanh,
     arctanh as atanh,
     bitwise_count,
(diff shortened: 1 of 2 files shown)


charris:
Thanks Joren.

PR #31937, opened 2026-07-09
DOC: document LinAlgError for incompatible dimensions in lstsq

shipitdev:
### PR Summary

I noticed that the `Raises` section for `np.linalg.lstsq` only documents the convergence failure. However, the function also raises a `LinAlgError` if `a` and `b` have mismatched first dimensions. 

This behavior is already actively checked by `test_incompatible_dims` in `test_linalg.py`, so this PR just updates the docstring to reflect the actual behavior.

**Quick reproducer**
```python
np.linalg.lstsq(np.ones((3, 2)), np.ones((4,)), rcond=None)
# LinAlgError: Incompatible dimensions
```

### AI Disclosure
-

### DIFF
--- a/numpy/linalg/_linalg.py
+++ b/numpy/linalg/_linalg.py
@@ -2442,7 +2442,6 @@ def lstsq(a, b, rcond=None):
     Raises
     ------
     LinAlgError
-        If computation does not converge.
 
     See Also
     --------


mattip:
Please use [skip actions] in your commit message for doc-only changes. It saves a lot of CI resources.

I don't think the intent is to list all the ways lstsq can raise. Maybe we should remove the section entirely, or somehow state that the list is incomplete in two words "(partial list)" or something like that.

shipitdev:
> Please use [skip actions] in your commit message for doc-only changes. It saves a lot of CI resources.

thanks for the tip i'll now use the skip actions on my doc changes commit to save resources.
> 
> I don't think the intent is to list all the ways lstsq can raise. Maybe we should remove the section entirely, or somehow state that the list is incomplete in two words "(partial list)" or something like that.

i understand, yes i think i'd agree about not wanting to clutter the docstring with every single edge case. I think we can easily cover your concern without literally writing `(partial list)` just by softening the language. What if we rephrased it to say "Commonly raised if the computation does not converge, or if a and b have incompatible first dimensions." ? what do you say ? i feel like that makes it clear the list isn't exhaustive while keeping it professional. Let me know if that works for you or if you'd still prefer i just nuke the section entirely.

numpy/linalg/_linalg.py
```diff
@@ -2442,7 +2442,8 @@ def lstsq(a, b, rcond=None):
     Raises
     ------
     LinAlgError
-        If computation does not converge.
+        If the least-squares computation does not converge, or if a and
+        b have incompatible first dimensions.
```
ngoldbaum: I agree with Matti that's it's clearer just to delete the explanation entirely. It's a linear algebra operation, it may raise a `LinAlgError` if something goes wrong.

shipitdev: alright, i'll remove the raises section and amend the last commit. thank you for the review.

ngoldbaum: I'd keep the section and just say it might raise LinAlgError with no further explanation. That is, just delete the sentence you changed.

shipitdev: ah you're right, i initially thought leaving the error without a description would actually trigger a numpydoc or spin validation error, which is why i went with the suggestion of removing the section entirely to be safe However, i just verified it locally and it actually is perfectly fine without the description ( just throws two completely unrelated warnings) I'll amend the commit to leave just the error name and push it up now. Thanks for your time here !

ngoldbaum:
Thanks @shipitdev

PR #31938, opened 2026-07-09
BUG: linspace with integer dtype silently returns out-of-range values

### LINKED ISSUE #31651: BUG: linspace with infinite start/stop and integer dtype silently returns INT_MIN

palios-taey:
### PR summary
`linspace` with an integer `dtype` casts the computed float64 values straight to the target integer with no bounds check. For non-finite endpoints, or a range that overflows the dtype, that cast is meaningless. On current main `np.linspace(np.inf, np.inf, 5, dtype=int)` returns `[0, 0, 0, 0, 9223372036854775807]` and `np.linspace(0, 1e30, 5, dtype=np.int32)` returns all `2147483647` — only a `RuntimeWarning`, no error, and the wrong values propagate silently.

This adds a representability check before the cast. Any computed value outside the target dtype's range raises `ValueError`; `NaN` and `inf` fail the same comparison and are rejected too. The bounds are `float(info.min)` and `float(info.max) + 1`, which are exact powers of two for every integer width, so the check is exact at the `2**63` boundary where a naive `y <= info.max` fails (float64 rounds `int64` max up to `2**63`). This is the general out-of-bound check from the gh-31651 discussion — not an `inf`-only special case, so finite overflow is rejected as well.

Closes #31651.

First contribution here. I use NumPy daily across ML-infrastructure work and hit this while generating integer index grids — a non-finite endpoint slipped through and produced silently-wrong indices instead of an error, which took a while to track down. Sending the fix the maintainers asked for in the gh-31651 thread.

#### AI Disclosure
This change was prepared with AI assistance — design exploration, the implementation, and the validation run were done with an AI coding assistant. I reviewed the diff and the test, ran the reasoning myself, and I take responsibility for the contribution.


numpy/_core/function_base.py
```diff
@@ -191,6 +192,15 @@ def linspace(start, stop, num=50, endpoint=True, retstep=False, dtype=None,
 
     if integer_dtype:
         _nx.floor(y, out=y)
+        info = iinfo(dtype)
+        lo = _nx.float64(info.min)
+        hi = _nx.float64(int(info.max) + 1)
+        if not ((y >= lo) & (y < hi)).all():
```
eendebakpt:
```suggestion
        if not (y.min() >= lo and y.max() < hi):
```
Might be faster. Could you do a quick microbenchmark to see the performance impact on `np.linspace`?

numpy/_core/tests/test_function_base.py
```diff
@@ -477,6 +477,30 @@ def test_round_negative(self):
         t = array([-1, -1, 0, 0, 1, 1, 2, 3], dtype=int)
         assert_array_equal(y, t)
 
+    def test_integer_dtype_representability(self):
```
eendebakpt, line 480: Could you also add a test for uint8?

palios-taey:
Thanks — adopted the `min()`/`max()` form, it is faster. One guard needed: `y.min()`/`y.max()` raise on an empty array, and `num=0` reaches this path, so I kept it as `if y.size and not (y.min() >= lo and y.max() < hi):` to leave the empty case a no-op (matching the old `.all()` behavior there).

Microbenchmark (median ns/op, `linspace(..., dtype=int)`):

| num | old `.all()` | `min/max` (guarded) |
|---|---|---|
| 50 | 8984 | 8217 |
| 10,000 | 29594 | 25694 |
| 1,000,000 | 4,322,419 | 3,346,763 |

Negligible at small sizes, ~22% faster at 1e6. Also added the uint8 tests (overflow raises, in-range `linspace(0, 255, 6, uint8)` unchanged). Pushed.

doc/release/upcoming_changes/31938.change.rst
eendebakpt, line 1: The number of the file should refer to the PR, not the issue

numpy/_core/tests/test_function_base.py
```diff
@@ -477,6 +477,33 @@ def test_round_negative(self):
...
```
eendebakpt, line 480: What happens to `linspace(0, 2**63 - 1, 2, dtype=np.int64)`?

doc/release/upcoming_changes/31938.change.rst
```diff
@@ -0,0 +1,8 @@
+New behavior for linspace with integer dtypes
```
eendebakpt, line 1: `logspace` and `geomspace` could be adjusted as well.

palios-taey:
Thanks — addressed all three:

**Release note** renamed to `31938.change.rst`.

**`linspace(0, 2**63 - 1, 2, dtype=np.int64)`** — with this PR it raises; on current main it returns `[0, 9223372036854775807]` with a `RuntimeWarning: invalid value encountered in cast`. The reason is that `2**63 - 1` isn't representable in float64 — `float64(2**63 - 1) == 2**63` — so the computed intermediate is `[0, 2**63]`, which is out of `int64` range. Main's cast saturates `2**63` to `INT64_MAX` (which here happens to equal the requested `2**63 - 1`); this PR sees the value at the exclusive upper bound and raises. Same for `2**63 - 2` (also rounds to `2**63`). The last exactly-representable integer is `2**53` — above that linspace-into-int is lossy regardless (e.g. `2**53 + 1` silently returns `2**53` on both main and this branch).

So this is a *representable-in-the-target-dtype* guard, consistent with the strict out-of-bound direction from gh-31657. That does mean it rejects top-boundary endpoints whose float64 image rounds up to `2**63`, even though main's saturating cast lands back on a plausible value with a warning. Two options at that boundary: (a) keep the strict raise — anything whose float image is out of range errors; or (b) soften so a value that saturates back into range is allowed. I lean (a): the saturation being 'correct' at `2**63 - 1` is coincidental (`2**63 - 2` saturates to `2**63 - 1` too, which is wrong), and (a) is the strict shape from gh-31657. But it's your call on the semantics — happy to switch to (b) if you'd prefer the boundary to pass.

**`logspace` / `geomspace`** — same silent/saturating out-of-range integer cast, but they don't route through linspace's guard; they generate in float and do their own final `astype`. Cleanest fix is a small shared representability helper applied before the integer cast in each, with tests for the negative/complex/unsigned cases `geomspace` allows. I'd suggest a separate PR to keep this one focused — say the word and I'll open it, or fold it in here if you'd rather.

seberg:
I am closing this, if anyone **actually cares for a real world scenario, then say so. Otherwise this whole discussion is a waste of maintainer time, quite honestly.**
(Issue templates now explicitly ask for it, maybe we'll add that to PR templates as well.)

If you actually care enough, please go to one of our meetings in person.

@eendebakpt if you actually think this makes sense, I'll reopen.

eendebakpt:
I have not encountered real-world issues with this. But in general I want things to be correct, so if there is a (relatively) easy fix for this I would be open for it.

I just checked and the examples given here do give a warning (no exception, just a warning). So `linspace` is not returning silently. If it would, then I think we should fix it for sure. Now it is a bit of a grey area.

seberg:
:shrug: the problem is maybe mostly that we have way too much of "maybe" PRs.  But to me it is also one of these: oh did you run into it in real code? plus the fact that it's not like many functions in NumPy waste CPU cycles to guard for this specific kind of thing.

PR #31940, opened 2026-07-09
PERF: avoid an unnecessary copy in np.take with out=

### LINKED ISSUE #28636: `numpy_take` slower when `out` is provided
### Describe the issue:

I've noticed this when trying to use the C-API `PyArray_TakeFrom` but it can be shown using python as well:

```python
import numpy as np
print(np.__version__)
# 2.2.4

x = np.arange(85, dtype=float)
idxs = np.repeat(np.arange(85), 110)
out = np.empty_like(np.take(x, idxs, out=None))
%timeit np.take(x, idxs, out=None)
%timeit np.take(x, idxs, out=out)
# 6.76 μs ± 106 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
# 8.99 μs ± 30.8 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```
With those sizes, I would expect the overhead to be mostly from copying the values, not checking whether the provided output is valid.

```python
NA
```

### Error message:

```shell
NA
```

numpy 2.2.4, python 3.12.8

### Runtime Environment:

_No response_

### Context for the issue:

_No response_

Ijtihed:
### PR summary

Fixes #28636.

`np.take(a, indices, out=out)` in the default `mode="raise"` was slower than the same call without `out`. `PyArray_TakeFrom` forced a defensive `ENSURECOPY` of `out` and wrote through a temporary that was copied back on success and only so `out` stayed unchanged if an out-of-bounds index raised

this removes that copy so in `raise` mode the result is written directly into `out` just like `mode="wrap"`/`"clip"` already do. The overlap-with-`self` copy (`arrays_overlap`) was kept so aliasing stays correct.

there **IS a behavior change on an out-of-bounds index** since `out` may now be partially written before `IndexError` is raised (previously it was left unchanged). `take` still raises. This proooobably warrants a release note and I'll write it after review

Before/after `out=`

- 1-D, 85 -> 9350 elems: `7.66us -> 4.90us` (1.56x)
- 1-D, 1000 -> 100000 elems: `99.5us -> 47.3us` (2.11x)
- axis 0, `(200, 64) -> (4000, 64)`: `214us -> 64.6us` (3.31x)

The no-`out` path is unchanged and the speedup grows with output size (since the removed copy is `O(n)`)

#### AI Disclosure

Used AI to double check my code + find any edge cases I might've missed :)

### DIFF
--- a/numpy/_core/tests/test_multiarray.py
+++ b/numpy/_core/tests/test_multiarray.py
@@ -6268,6 +6268,38 @@ class TestTake:
         ret = np.take(x, inds, out=out)
         assert ret is out
 
+    def test_out_matches_no_out(self):
+        # gh-28636: with out=, take now writes directly into out instead of
+        # through a temporary copy. The result must match a plain take.
+        rng = np.random.default_rng(1)
+        for dtype in (np.float64, np.int32, np.complex128, np.uint8):
+            x = (rng.random((4, 5, 6)) * 100).astype(dtype)
+            for axis in (0, 1, 2, -1):
+                idx = rng.integers(0, x.shape[axis], size=7)
+                expected = np.take(x, idx, axis=axis)
+                out = np.empty_like(expected)
+                ret = np.take(x, idx, axis=axis, out=out)
+                assert ret is out
+                assert_array_equal(out, expected)
+
+    def test_out_raise_still_raises(self):
+        # gh-28636: an out-of-bounds index must still raise with out=.
+        x = np.arange(10.0)
+        out = np.empty(3)
+        assert_raises(IndexError, np.take, x, [1, 20, 3], out=out)
+        assert_raises(IndexError, np.take, x, [-11, 1, 2], out=out)
+        # valid indices into the same out still work
+        assert_array_equal(np.take(x, [1, 2, 3], out=out), [1, 2, 3])
+
+    def test_out_object_identity(self):
+        # object dtype. references must be copied into out.
+        a, b, c = [1], [2], [3]
+        x = np.empty(3, dtype=object)
+        x[0], x[1], x[2] = a, b, c
+        out = np.empty(3, dtype=object)
+        np.take(x, [2, 0, 1], out=out)
+        assert out[0] is c and out[1] is a and out[2] is b
+
 
 class TestLexsort:
     @pytest.mark.parametrize('dtype', [
--- a/doc/release/upcoming_changes/31940.compatibility.rst
+++ b/doc/release/upcoming_changes/31940.compatibility.rst
@@ -1,0 +1,8 @@
+``np.take`` may modify ``out`` when an index is out of bounds
+-------------------------------------------------------------
+With the default ``mode="raise"``, `numpy.take` now writes its result
+directly into the ``out`` array instead of into a temporary copy. As a
+consequence, ``out`` may be partially modified when an out-of-bounds index
+raises an ``IndexError``; previously ``out`` was guaranteed to be left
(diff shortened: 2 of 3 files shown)


Ijtihed:
@mattip since you mentioned this I thought you might be interested :)

numpy/_core/tests/test_multiarray.py
ikrommyd, line 1: That might be a little bit of an excessive testing for this kind of tiny change IMO.

numpy/_core/tests/test_multiarray.py
```diff
@@ -6268,6 +6268,79 @@ def test_ret_is_out(self, shape):
...
+        filler = [0]
+        x = np.empty(3, dtype=object)
+        x[0], x[1], x[2] = [1], [2], [3]
+        out = np.empty(3, dtype=object)
+        out[0] = out[1] = out[2] = filler
+        gc.collect()
+        rc = sys.getrefcount(filler)  # local + 3 slots in out
+        np.take(x, [0, 1, 2], out=out)
+        # the three filler references held by out must have been released
+        assert sys.getrefcount(filler) == rc - 3
```
ikrommyd: I'm not sure if this can ever start failing but I had tests in the past in other libraries where asserting reference counts would break as python evolves cause they would change how reference counting works internally as free threaded python started being a thing. For that reason I try to just add tests that would just be detected by leak sanitizer per se under a sanitized build of cpython and numpy instead of asserting on ref counts directly. It may just be fine though.

ikrommyd (commented):
If the consensus is to just let `out` be modified, that is like basic guarantee in terms of c++ exception safety. I'd be tempted to add an additional message to the raised error in that case. Like if the error is raised and `out=` is set, you also say in the error message that `out` may be modified. cc @seberg @mattip what do you think? It can be done using by getting the exception that would be raised which is `IndexError: index 4 is out of bounds` and adding a note to it using the `add_note` method they have (numpy is already 3.12+). Something like
```c
if (npy_fasttake(
        dest, src, indices_data, n, m, max_item, nelem, chunk,
        clipmode, itemsize, needs_refcounting, src_descr, dst_descr,
        axis) < 0) {
    if (out != NULL && obj == out) {
        PyObject *exc = PyErr_GetRaisedException();
        if (exc != NULL) {
            PyObject *res = PyObject_CallMethod(exc, "add_note", "s",
                "the output array `out` may have been partially modified");
            if (res == NULL) {
                PyErr_Clear();
            }
            Py_XDECREF(res);
            PyErr_SetRaisedException(exc);
        }
    }
    goto fail;
}
```

seberg:
I like the add-note idea, although it applies to many operations (i.e. if we do that, I think it should probably be a central helper and maybe added to some other places).

FWIW, it seems OK to me, although I don't care too much either way (`mode=...` is an existing path to get the no-copy in practice).
But it does need a release note IMO as it relaxes an _intentional_ guarantee.

Ijtihed:
thanks both! I added a compatibility release note documenting that out can now be partially modified if an out of bounds index raises (the relaxed guarantee) then dropped the refcount-asserting test and trimmed the rest down to the essentials so there's no getrefcount asserts left. leak coverage can just be left to the sanitized CI

@ikrommyd @seberg I left add_note out for now since i agree it makes more sense as a central helper across the ops that can partially write out :)

Ijtihed:
> I like the add-note idea, although it applies to many operations (i.e. if we do that, I think it should probably be a central helper and maybe added to some other places).

I'd also be happy to open a pr for this too if you think it makes sense

PR #31941, opened 2026-07-09
TYP: specialized ``divide`` ufunc

jorenham:
This ufunc specialization for ``divide`` is similar to #31935, but also supports `complexfloating`, `object_`, and `timedelta64` dtypes. This one also has different promotion rules for `integer` and `bool`, which are always promoted to `float64`, independent of precision, which made things a bit ~easier~ less difficult.
But even so, I decided to not add specific `float16` and `[c]longdouble` promotion overloads here, because doing so would add another +8 overloads to `__call__` alone, which already counts 42 overloads now. Those dtypes are still accepted, but when combined with different input dtypes, they'll result in `Any` now (matching the current behavior, where `ufunc` always returns `Any` for basically everything).

---

No AI.


jorenham:
> Diff from [mypy_primer](https://github.com/hauntsaninja/mypy_primer?rgh-link-date=2026-07-09T20%3A48%3A36Z), showing the effect of this PR on type check results on a corpus of open source code:
> 
> ```diff
> dedupe (https://github.com/dedupeio/dedupe)
>   error: INTERNAL ERROR -- Please try using mypy master on GitHub:
> ```

See https://github.com/dedupeio/dedupe/pull/1229

charris:
Thanks Joren.

PR #31943, opened 2026-07-09
MAINT, ENH: Tuple-spec `array_function_dispatch` 

### LINKED ISSUE #31866: ENH: Reduce `__array_function__` dispatch overhead

eendebakpt:
We allow `@array_function_dispatch` to take a tuple of relevant-argument names instead of a callable dispatcher. The names are resolved to `(kwname, position)` pairs at decoration time. The tuple approach avoids the overhead of calling the dispatcher and creating an intermediate tuple. Fast paths are added for exact numpy arrays and python basic types.

Fixes https://github.com/numpy/numpy/issues/31866. Also see https://github.com/numpy/numpy/issues/31098.

About 260 trivial dispatchers are migrated (~600 lines removed); the callable form remains for sequence dispatchers 

Benchmark results (Python 3.14):
```
np.sum(array_4): Mean +- std dev: [bench_main_sq] 621 ns +- 5 ns -> [bench_branch_sq] 576 ns +- 4 ns: 1.08x faster
np.sum(array_20): Mean +- std dev: [bench_main_sq] 625 ns +- 4 ns -> [bench_branch_sq] 579 ns +- 3 ns: 1.08x faster
np.prod(array_20): Mean +- std dev: [bench_main_sq] 598 ns +- 3 ns -> [bench_branch_sq] 550 ns +- 7 ns: 1.09x faster
np.max(array_20): Mean +- std dev: [bench_main_sq] 591 ns +- 3 ns -> [bench_branch_sq] 560 ns +- 10 ns: 1.06x faster
np.sum(int_20): Mean +- std dev: [bench_main_sq] 628 ns +- 9 ns -> [bench_branch_sq] 585 ns +- 14 ns: 1.07x faster
np.sum(10x10): Mean +- std dev: [bench_main_sq] 635 ns +- 5 ns -> [bench_branch_sq] 587 ns +- 3 ns: 1.08x faster
np.sum(10x10,axis=0): Mean +- std dev: [bench_main_sq] 1.20 us +- 0.02 us -> [bench_branch_sq] 1.14 us +- 0.02 us: 1.05x faster
np.sum(10x10,axis=1): Mean +- std dev: [bench_main_sq] 1.14 us +- 0.02 us -> [bench_branch_sq] 1.10 us +- 0.02 us: 1.04x faster
np.any(bool_20): Mean +- std dev: [bench_main_sq] 572 ns +- 16 ns -> [bench_branch_sq] 516 ns +- 8 ns: 1.11x faster
np.sum(2x2_float): Mean +- std dev: [bench_main_sq] 628 ns +- 8 ns -> [bench_branch_sq] 575 ns +- 5 ns: 1.09x faster
np.sum(2x2_int): Mean +- std dev: [bench_main_sq] 626 ns +- 7 ns -> [bench_branch_sq] 584 ns +- 9 ns: 1.07x faster
np.sin(array_20): Mean +- std dev: [bench_main_sq] 371 ns +- 6 ns -> [bench_branch_sq] 376 ns +- 4 ns: 1.01x slower
np.sum(subclass_20): Mean +- std dev: [bench_main_sq] 972 ns +- 10 ns -> [bench_branch_sq] 945 ns +- 16 ns: 1.03x faster
np.transpose(10x10): Mean +- std dev: [bench_main_sq] 259 ns +- 2 ns -> [bench_branch_sq] 226 ns +- 1 ns: 1.15x faster
np.ravel(array_20): Mean +- std dev: [bench_main_sq] 233 ns +- 4 ns -> [bench_branch_sq] 196 ns +- 2 ns: 1.19x faster

Geometric mean: 1.08x faster
```

`np.sin` is the control (not dispatched through `@array_function_dispatch`).

<details>

```python
"""Micro-benchmark for np.sum/prod/max/any dispatcher overhead."""
import numpy as np
import pyperf


class MyArr(np.ndarray):
    pass


def make_setup():
    return {
        "array_4":    np.arange(4, dtype=np.float64),
        "array_20":   np.arange(20, dtype=np.float64),
        "int_20":     np.arange(20, dtype=np.int64),
        "bool_20":    np.ones(20, dtype=bool),
        "arr_10x10":  np.arange(100, dtype=np.float64).reshape(10, 10),
        "arr_2x2f":   np.arange(4, dtype=np.float64).reshape(2, 2),
        "arr_2x2i":   np.arange(4, dtype=np.int64).reshape(2, 2),
        "subclass_20":np.arange(20, dtype=np.float64).view(MyArr),
    }


BENCHES = [
    ("np.sum(array_4)",         "np.sum(a['array_4'])"),
    ("np.sum(array_20)",        "np.sum(a['array_20'])"),
    ("np.prod(array_20)",       "np.prod(a['array_20'])"),
    ("np.max(array_20)",        "np.max(a['array_20'])"),
    ("np.sum(int_20)",          "np.sum(a['int_20'])"),
    ("np.sum(10x10)",           "np.sum(a['arr_10x10'])"),
    ("np.sum(10x10,axis=0)",    "np.sum(a['arr_10x10'], axis=0)"),
    ("np.sum(10x10,axis=1)",    "np.sum(a['arr_10x10'], axis=1)"),
    ("np.any(bool_20)",         "np.any(a['bool_20'])"),
    ("np.sum(2x2_float)",       "np.sum(a['arr_2x2f'])"),
    ("np.sum(2x2_int)",         "np.sum(a['arr_2x2i'])"),
    ("np.sin(array_20)",        "np.sin(a['array_20'])"),
    ("np.sum(subclass_20)",     "np.sum(a['subclass_20'])"),
    # cheap view ops migrated in the mass tuple-spec conversion
    ("np.transpose(10x10)",     "np.transpose(a['arr_10x10'])"),
    ("np.ravel(array_20)",      "np.ravel(a['array_20'])"),
]


def main():
    runner = pyperf.Runner()
    runner.metadata['numpy_version'] = np.__version__
    a = make_setup()
    for name, expr in BENCHES:
        runner.timeit(name=name, stmt=expr, globals={'np': np, 'a': a})


if __name__ == '__main__':
    main()
```

</details>

#### AI Disclosure

Claude Opus 4.x was used to find optimizations for the dispatcher overhead. The corresponding issue contains an alternative. After selecting this option, Claude was used to refine the PR.

### DIFF
--- a/numpy/_core/overrides.py
+++ b/numpy/_core/overrides.py
@@ -1,6 +1,5 @@
 """Implementation of __array_function__ overrides from NEP-18."""
 import collections
-import enum
 import functools
 import inspect
 
@@ -9,17 +8,49 @@ from numpy._core._multiarray_umath import (
     _get_implementing_args,
     add_docstring,
 )
+from numpy._globals import _NoValue
 from numpy._utils import set_module  # noqa: F401
 from numpy._utils._inspect import getargspec
 
 ARRAY_FUNCTIONS = set()
 
-# Signature families used by the exact-ndarray reduction fast path.
-class _ReductionKind(enum.IntEnum):
-    # Keep in sync with the enum in arrayfunction_override.c
-    SUM_PROD = 1
-    MIN_MAX = 2
-    ANY_ALL = 3
+# ufunc.reduce positional argument order, used by the exact-ndarray
+# reduction fast path.  Public reduction functions (np.sum, np.any, ...)
+# use these same parameter names, which is how their signatures are
+# mapped onto a direct ufunc.reduce call in C.
+_REDUCE_SLOTS = {
+    "a": 0, "axis": 1, "dtype": 2, "out": 3,
+    "keepdims": 4, "initial": 5, "where": 6,
+}
+# Values for reduce arguments the caller did not pass (or passed as
+# np._NoValue).  Slot 0 (`a`) has no default: it is required.
+_REDUCE_DEFAULTS = (None, None, None, None, False, _NoValue, True)
+
+
+def _resolve_reduction_spec(implementation, ufunc, defaults_override):
+    """Map ``implementation``'s parameters onto ``ufunc.reduce`` arguments.
+
+    Returns ``(reduce_callable, slots, defaults)`` where ``slots[i]`` is the
+    reduce-argument index of the i-th parameter (aligned with the
+    ``param_names`` table from ``_resolve_relevant_arg_spec``) and
+    ``defaults`` holds the reduce-call value for every argument the caller
+    omitted.  The C dispatcher uses this to call ``ufunc.reduce`` directly
+    for exact-ndarray calls, skipping the Python wrapper entirely.
+    """
+    defaults = list(_REDUCE_DEFAULTS)
+    if defaults_override:
+        for name, value in defaults_override.items():
+            defaults[_REDUCE_SLOTS[name]] = value
+    slots = []
+    for name, param in inspect.signature(implementation).parameters.items():
+        if (name not in _REDUCE_SLOTS
+                or param.kind not in (param.POSITIONAL_OR_KEYWORD,
+                                      param.KEYWORD_ONLY)):
+            raise RuntimeError(
+                f"reduction fast path cannot map parameter {name!r} of "
+                f"{implementation.__qualname__} onto ufunc.reduce")
+        slots.append(_REDUCE_SLOTS[name])
+    return (ufunc.reduce, tuple(slots), tuple(defaults))
 
 
 array_function_like_doc = (
@@ -87,7 +118,9 @@ add_docstring(
     Returns
     -------
     Sequence of arguments with __array_function__ methods, in the order in
-    which they should be called.
+    which they should be called.  Returns an empty sequence when every
+    argument is an exact ``ndarray`` or a basic Python type (the caller
+    short-circuits to the default implementation in that case).
     """)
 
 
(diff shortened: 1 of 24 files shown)


numpy/_core/overrides.py
```diff
@@ -142,29 +191,46 @@ def array_function_dispatch(dispatcher=None, module=None, verify=True,
...
-                        f"{implementation} does not seem to comply.")
-
-        if docs_from_dispatcher and dispatcher.__doc__ is not None:
-            doc = inspect.cleandoc(dispatcher.__doc__)
-            add_docstring(implementation, doc)
-
-        public_api = _ArrayFunctionDispatcher(dispatcher, implementation)
+        if is_tuple_spec:
+            spec = _resolve_relevant_arg_spec(implementation, dispatcher)
+            public_api = _ArrayFunctionDispatcher(spec, implementation)
```
jorenham: did you verify that this didn't affect the `inspect.signature` of the decorated functions?

eendebakpt: I added a parameterized test `test_inspect_signature_matches_implementation` to check for this.

seberg:
Thanks, overall, I think this is good, I have two quick questions:
* Is that fast-path check you added really worthwhile?  We already should have a similar fast-path later on, no?  It feels like the wrong way to add it, TBH. (I.e. add a second fast-path, rather than making sure that the existing fast-path is fast.)
* So one thing that I think we do lose here is that you are only extracting arguments that we actually dispatch on.  This should save a bit of time, but may mean that we forward the call without checking the signature exactly?
  (I think in part, I am wondering whether it would make sense to re-use the full `npy_argparse` machinery here with a boolean pattern of which arguments are "relevant" -- the downside is that it is slower for irrelevant kwargs, the upside that we get the check.  But... maybe it is silly, because whatever we gain by simplifying the parsing itself, the setup likely adds more complexity; so to be worthwhile we maybe really need to want the signature correctness check.)

eendebakpt:
@seberg I rebased onto main, eliminating the recently added enums. For review: the first commit now contains the changes to the machinery, the second commits converts the remaining dispatchers where possible.

The fast path in `get_implementing_args_and_methods` with `n_safe` is indeed a bit odd, but needed for performance. If any argument is not an exact ndarray or python scalar, then we must return the ndarray arguments in the `implementing_args` (which involves storing data, increfs). The performance gain is not huge, the PR would still be worthwhile without this fast path. Maybe we could move the fast path into `get_implementing_args_and_methods` (which will be a intermediate in performance, and intermediate in code complexity). I can benchmark variations later.

For the converted dispatches that used the enum performance effect is roughly neutral, for converted dispatchers performance gain is in the 5 to 20% range, the override dispatch (where `__array_function__` is called) is also faster.

In a followup PR we can make more dispatched functions avoid the python paths. Prototype in https://github.com/numpy/numpy/commit/4e1aab9447b7a5bd9f1f81c7ecefbd479ad54110 gives

| case | base (gh-31943) | forward | speedup |
|---|---|---|---|
| `np.argmax(f)` | 919 ns | 399 ns | **2.3x** |
| `np.transpose(m)` | 361 ns | 174 ns | **2.1x** |
| `np.swapaxes(m, 0, 1)` | 421 ns | 202 ns | **2.1x** |
| `np.argsort(f, kind="stable")` | 1173 ns | 593 ns | **2.0x** |
| `np.argsort(f)` | 1176 ns | 607 ns | 1.9x |
| `np.ravel(m)` | 318 ns | 175 ns | 1.8x |
| `np.cumsum(a)` | 1525 ns | 856 ns | 1.8x |
| `np.searchsorted(sf, 5.5)` | 1144 ns | 649 ns | 1.8x |
| `np.argmax(m, axis=0, out=)` | 1289 ns | 699 ns | 1.8x |
| `np.argmax(m, axis=0)` | 1383 ns | 795 ns | 1.7x |
| `np.nonzero(a)` | 599 ns | 350 ns | 1.7x |
| `np.take(a, [1, 2])` | 1221 ns | 743 ns | 1.6x |
| `np.repeat(a, 2)` | 1138 ns | 724 ns | 1.6x |
| `np.reshape(a, (4, 5))` | 621 ns | 383 ns | 1.6x |
| `np.round(f, 2)` | 2421 ns | 1708 ns | 1.4x |
| `np.compress([T,F]*10, a)` | 2212 ns | 1567 ns | 1.4x |
| `np.sum(a)` *(control)* | 708 ns | 674 ns | ~1.0x |
| `np.sin(f)` *(control)* | 539 ns | 470 ns | ~1.0x |

<details>

```python
import sys
import timeit

import numpy as np

label = sys.argv[1] if len(sys.argv) > 1 else "run"

a = np.array(list(range(20)), dtype=np.int64)
f = np.array([0.5 * i for i in range(20)])
m = np.array([float(i) for i in range(100)]).reshape(10, 10)
sorted_f = np.array([float(i) for i in range(20)])
out_am = np.empty(10, dtype=np.intp)

CASES = {
    "reshape(a, (4, 5))": lambda: np.reshape(a, (4, 5)),
    "ravel(m)": lambda: np.ravel(m),
    "transpose(m)": lambda: np.transpose(m),
    "take(a, [1, 2])": lambda: np.take(a, [1, 2]),
    "argmax(f)": lambda: np.argmax(f),
    "argmax(m, axis=0)": lambda: np.argmax(m, axis=0),
    "argmax(m, axis=0, out=)": lambda: np.argmax(m, axis=0, out=out_am),
    "argsort(f)": lambda: np.argsort(f),
    "argsort(f, kind='stable')": lambda: np.argsort(f, kind="stable"),
    "searchsorted(sf, 5.5)": lambda: np.searchsorted(sorted_f, 5.5),
    "nonzero(a)": lambda: np.nonzero(a),
    "cumsum(a)": lambda: np.cumsum(a),
    "round(f, 2)": lambda: np.round(f, 2),
    "repeat(a, 2)": lambda: np.repeat(a, 2),
    "swapaxes(m, 0, 1)": lambda: np.swapaxes(m, 0, 1),
    "compress([T,F]*10, a)": lambda: np.compress([True, False] * 10, a),
    "sum(a) [control]": lambda: np.sum(a),
    "sin(f) [control]": lambda: np.sin(f),
}

best = {name: float("inf") for name in CASES}
for _ in range(15):
    for name, fn in CASES.items():
        t = timeit.timeit(fn, number=10_000) / 10_000
        best[name] = min(best[name], t)

print(f"### {label}  (numpy {np.__version__})")
for name in CASES:
    print(f"{name:28s} {best[name] * 1e9:8.1f} ns")
```

</details>

seberg:
OK, I don't think I'll manage to review in depth very soon, unfortunately.  The other branch looks like a neat idea.
I suppose doing the full argument parsing is noticeably slower probably (rather than validating only if we dispatch)?  I would find it neater if the fast-path was somehow integrate into the loop, just before the "have I seen this type" check, I guess.  But I can see that may turn out just as messy...

eendebakpt:
Merge conflicts. Closing for now in favor of https://github.com/numpy/numpy/pull/32523

PR #31944, opened 2026-07-09
CI/TYP: deselect ``dedupe`` in mypy_primer 

jorenham:
After the recent removal of the mypy plugin, mypy_primer is now complaining about an internal error in `dedupe`, which still uses that (deprecated) mypy plugin. For example: https://github.com/numpy/numpy/pull/31943#issuecomment-4930043495

So let's deslect `dedupe` until https://github.com/dedupeio/dedupe/pull/1229 gets merged.

I also updated mypy_primer to the latest commit, after going over the full [diff](https://github.com/hauntsaninja/mypy_primer/compare/05f73ec3d85bb4f55676f3c57f2c3e5136228977..ff3d9531335dad605d43fcc208c9c159bb09fa81) (which seemed fine).

---

No AI.

### DIFF
--- a/.github/workflows/mypy_primer.yml
+++ b/.github/workflows/mypy_primer.yml
@@ -58,10 +58,12 @@ jobs:
           echo ''
           cd ..
           # fail action if exit code isn't zero or one
+          # (dedupe is temporarily disabled: https://github.com/dedupeio/dedupe/pull/1229)
           (
             mypy_primer \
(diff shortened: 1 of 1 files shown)


jorenham:
mypy_primer still seems to work: https://github.com/numpy/numpy/actions/runs/29056284277/job/86248308646?pr=31944

charris:
Thanks Joren.

PR #31951, opened 2026-07-10
TYP: Avoid shadowed dtype annotations

jelle-openai:
### PR summary

gh-30754 fixed annotations shadowed by class members under
Python 3.14 annotation-scope rules. Later ndarray and MaskedArray
signatures reintroduced bare dtype references, which the ty type checker resolves to
the dtype property. As a result, it infers Unknown for operations
such as argmax(keepdims=True).

Use _dtype or np.dtype for the 53 affected annotations and bounds, and
move the _dtype alias below the dtype definition. This restores precise
ty inference without changing runtime behavior.

A follow-up could add ty alongside Pyrefly in the type-checking
workflow and check the 166 production .pyi files. With the small
existing baseline of final-class, override, parameter-default, and
type-argument diagnostics suppressed, invalid-type-form can remain
enabled and would catch regressions like this. If a maintainer agrees
that running ty in CI is valuable, I can work on this as a followup.

Follow-up to gh-30754.

This is @JelleZijlstra's work account. I have made many contributions to the Python ecosystem though as far as I recall I have not contributed to numpy before. See below for how I came to make this contribution.

#### AI Disclosure

I was having Codex look for cases where ty inferred less precise types than other type checkers. When it found some issues related to numpy, I had it prepare a fix inside numpy. I personally reviewed all code changes and wrote the final version of the PR body.

### DIFF
--- a/numpy/__init__.pyi
+++ b/numpy/__init__.pyi
@@ -850,10 +850,6 @@ type _BuiltinObjectLike = (
     | tuple[Any, ...] | list[Any] | set[Any] | frozenset[Any] | dict[Any, Any]
 )  # fmt: skip
 
-# Introduce an alias for `dtype` to avoid naming conflicts.
-# NOTE: This should _not_ be `Final[_]`, `_: TypeAlias`, or `type _`
-_dtype = dtype
-
 type _ByteOrderChar = L["<", ">", "=", "|"]
 # can be anything, is case-insensitive, and only the first character matters
 type _ByteOrder = L[
@@ -1716,6 +1712,10 @@ class dtype(Generic[_ScalarT_co], metaclass=_DTypeMeta):
     @property
     def type(self) -> py_type[_ScalarT_co]: ...
 
+# Introduce an alias for `dtype` to avoid naming conflicts.
+# NOTE: This should _not_ be `Final[_]`, `_: TypeAlias`, or `type _`
+_dtype = dtype
+
 @type_check_only
 class _ArrayOrScalarCommon:
     @property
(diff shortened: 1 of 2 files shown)


numpy/__init__.pyi
```diff
@@ -1717,6 +1712,10 @@ class dtype(Generic[_ScalarT_co], metaclass=_DTypeMeta):
     @property
     def type(self) -> py_type[_ScalarT_co]: ...
 
+# Introduce an alias for `dtype` to avoid naming conflicts.
+# NOTE: This should _not_ be `Final[_]`, `_: TypeAlias`, or `type _`
+_dtype = dtype
```
jorenham, line 1717: Did you move this because ty doesn't support forward references in stubs or something? (it's no problem; just wondering)

jorenham (approved):
Thanks Jelle!

Personally I think we should support as many type checkers as we reasonably can. I've been using ty in scipy-stubs for a while now, and even though there are still things it doesn't support (e.g. variadic type parameters), it has been smooth sailing so far. So as far as I'm concerned I'm all for adding ty as a CI job (as long as it doesn't require too many workarounds, that is).

numpy/__init__.pyi
jelle-openai: My Codex said something about this forward reference causing problems for ty. I didn't look too deeply as the change seemed harmless.

jorenham: might be good to know if that's the case then, because I believe there are other similar forward class references like this one elsewhere in the stubs

jelle-openai: Yeah, confirmed ty doesn't like forward references with plain assignments. I opened astral-sh/ty#3967.

I think it still makes sense to merge this change to make progress; it seems pretty harmless to move the alias around. In a followup PR I can run ty in CI and fix any similar issues elsewhere, or maybe ty will fix the issue first.

jorenham: Good to know; in it goes :)

PR #31956, opened 2026-07-10
CI: Run ty on NumPy's type stubs

jelle-openai:
### PR summary

Add a Python 3.14 ty job covering the production stubs and pin ty alongside the other typing tools. To get ty to pass, I disabled the `invalid-method-override` rule (it had >200 hits I think, not really fixable) and mostly `# ty: ignore`d the rest. Most issues found by ty were the same as ones already flagged by mypy and pyright. There are also a lot of `subclass-of-final-class`; perhaps it makes sense to disable that rule globally too?

#### AI Disclosure

The PR was fully written by Codex.


numpy/_typing/_ufunc.pyi
jorenham, line 1: can we disable `subclass-of-final-class` for the entire module here? Otherwise I'm also fine with disabling this globally, seeing as this isn't the only module that has a multiple `subclass-of-final-class` errors.

numpy/__init__.pyi
```diff
@@ -5032,7 +5032,7 @@ class ndarray(_ArrayOrScalarCommon, Generic[_ShapeT_co, _DTypeT_co]):
 # NOTE: while `np.generic` is not technically an instance of `ABCMeta`,
 # the `@abstractmethod` decorator is herein used to (forcefully) deny
 # the creation of `np.generic` instances.
-# The `# type: ignore` comments are necessary to silence mypy errors regarding
+# The mypy suppression comments are necessary to silence errors regarding
```
jorenham:
```suggestion
# The ignore comments are necessary to silence errors regarding
```

pyproject.toml
```diff
@@ -303,3 +303,15 @@ errors = { bad-override = "ignore" }
 [[tool.pyrefly.sub-config]]
 matches = "numpy/matrixlib/defmatrix.pyi"
 errors = { bad-override = "ignore" }
+
+[tool.ty.environment]
+python-version = "3.14"
```
jorenham: If we omit this, will ty automatically determine it from the environment? Because if so, that would have my preference.

pyproject.toml
```diff
@@ -303,3 +303,15 @@ errors = { bad-override = "ignore" }
...
+
+[tool.ty.src]
+include = ["numpy/*.pyi", "numpy/**/*.pyi"]
```
jorenham: can we simplify this to `["`numpy/**/*.pyi`"]`?

pyproject.toml
```diff
@@ -303,3 +303,15 @@ errors = { bad-override = "ignore" }
...
+exclude = ["numpy/typing/tests/**"]
+
+[tool.ty.rules]
+deprecated = "ignore"
+# NumPy intentionally models runtime-incompatible specialized overrides in its stubs.
```
jorenham: "accidentally" would probably be more accurate than "intentionally" haha. But either way, we can just get rid of this comment as far as I'm concerned.

```suggestion
```

numpy/_typing/_ufunc.pyi
jorenham: Ah sorry I missed that you already proposed this in the description:

> There are also a lot of subclass-of-final-class; perhaps it makes sense to disable that rule globally too?

So yea, let's

jelle-openai: Done, removed all the ignore comments for this rule too

pyproject.toml
jelle-openai: Yes. It now runs 3.12 instead as that's numpy's minimum version

pyproject.toml
jelle-openai: Maybe accidental at runtime but intentional that the stubs reflect the runtime :) . In any case, comment removed.

jorenham:
This is probably worth discussing a bit before merging, so I've added the "triage review" label. Would you be willing to come to a NumPy community meeting or triage meeting to @jelle-openai? See https://scientific-python.org/calendars/ for the data.

jorenham:
We discussed this at the community meeting just now, and we're slightly worried that this might hinder us in using typing features that ty doesn't support at the moment (e.g. variadic type parameters using PEP 695 syntax).
So I think it would help if you could explain the advantages of this at a triage meeting (there's one this time next week and then every two weeks) or a community meeting (next one is in two weeks this time).

jelle-openai:
Thanks, I probably won't be able to make it next week. I think TypeVarTuple support in ty is landing soon, but it might make sense to wait a bit until ty stabilizes a bit more.

jorenham:
Makes sense. I'll mark this as draft, and then we can un-draft (or whatever that's called) once ty is production ready :)

charris:
@jelle-openai Are you still interested in this.

jelle-openai:
Yes, I still think you should do this. I see there's a conflict though. Since I wrote this PR ty has gained TypeVarTuple support.

jorenham:
We could try rebasing and seeing how it goes. Ty has matured quite in the meantime, so unless any blocking issues turn up, I think we can go on ahead and start using it here.

PR #31959, opened 2026-07-11
MAINT: Use Py_CHARMASK for ctype calls on signed char in string parsers

isl-Ramzi:
### PR summary

The datetime, business-day and fromstring/fromfile parsers hand a plain `char` straight to `isdigit`/`isspace`/`tolower`, so any input byte above `0x7f` arrives as a negative value and the ctype call is undefined per the C standard, which requires the argument be representable as `unsigned char` or equal to `EOF`. These paths run on user data: `np.datetime64` and timedelta strings, `np.busdaycalendar(weekmask=...)`, the `np.fromstring`/`np.fromfile` separator, and rational string parsing. Each argument is wrapped in `Py_CHARMASK`, the same guard `numpyos.c` already uses for its ctype calls.

Reachable examples that currently drive a high byte into a bare ctype call:

- `np.datetime64('\xe9020-01-01')`
- `np.busdaycalendar(weekmask='\xe9on Tue')`
- `np.fromstring('1\xe92\xe93', sep='\xe9', dtype=int)`

Valid-input behaviour is unchanged; the datetime and multiarray test suites pass locally.

I use NumPy day to day for numerical and data-processing work. I came across this while reading through the string parsing code and noticed the ctype calls were being fed signed `char` directly, so I put together a fix that follows the `Py_CHARMASK` idiom already used elsewhere in the tree.

### DIFF
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -141,11 +141,11 @@ swab_separator(const char *sep)
         return NULL;
     }
     /* add space to front if there isn't one */
-    if (*sep != '\0' && !isspace(*sep)) {
+    if (*sep != '\0' && !isspace(Py_CHARMASK(*sep))) {
         *s = ' '; s++;
     }
     while (*sep != '\0') {
-        if (isspace(*sep)) {
+        if (isspace(Py_CHARMASK(*sep))) {
             if (skip_space) {
                 sep++;
             }
@@ -209,7 +209,7 @@ fromstr_skip_separator(char **s, const char *sep, const char *end)
         }
         else if (*sep == ' ') {
             /* whitespace wildcard */
-            if (!isspace(c)) {
+            if (!isspace(Py_CHARMASK(c))) {
                 sep++;
                 continue;
             }
--- a/numpy/_core/src/multiarray/datetime.c
+++ b/numpy/_core/src/multiarray/datetime.c
@@ -2585,9 +2585,9 @@ convert_pyobject_to_timedelta(PyArray_DatetimeMetaData *meta, PyObject *obj,
 
         /* Check for a NaT string */
         if (len <= 0 || (len == 3 &&
-                        tolower(str[0]) == 'n' &&
-                        tolower(str[1]) == 'a' &&
-                        tolower(str[2]) == 't')) {
+                        tolower(Py_CHARMASK(str[0])) == 'n' &&
+                        tolower(Py_CHARMASK(str[1])) == 'a' &&
+                        tolower(Py_CHARMASK(str[2])) == 't')) {
             *out = NPY_DATETIME_NAT;
             succeeded = 1;
         }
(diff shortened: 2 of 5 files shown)


mattip:
I think this is a bit of a code smell. We use char where we mean unsigned char. It would be better to fix at source, how noisy would that be?

isl-Ramzi:
I looked into fixing it at the type level. The catch is the char* originates outside these functions: the buffers come from the Python string APIs as char*, and the parser signatures themselves are char const* (NpyDatetime_ParseISO8601Datetime alone is called from dtype_transfer.c, datetime.c and stringdtype/casts.cpp). Pushing unsigned char down to the source means changing those signatures and casting at every call site, plus in the spots where the same chars feed strcmp, pointer arithmetic (*substr - '0') and assignments back into char* buffers. That ends up a much larger and more invasive diff than the ctype calls it's meant to guard.

Py_CHARMASK is the localized guard numpyos.c and scalartypes.c.src already use for exactly this, so this keeps it consistent with what's there. I'm happy to go the signature route if you'd prefer, but wanted to flag that it reaches well beyond the parsers.

Gooh456:
Read-through review (didn't build/run this — just read the diff and the surrounding files, no local test run).

The `Py_CHARMASK` fix looks correct and consistently applied everywhere it's used here. While checking whether the same "raw ctype call on an out-of-domain value" pattern shows up elsewhere, I found `numpy/_core/src/multiarray/textreading/str_to_int.h` (not touched by this PR) has the same underlying issue, arguably worse: `str_to_int64`/`str_to_uint64` call `isdigit(*p)` / `isdigit(d)` where `p` is `const Py_UCS4 *` and `d` holds a full Unicode codepoint (up to 0x10FFFF), not a `char`. That's not just the -128..-1 signed-char range this PR fixes — it's an arbitrary value going straight into a ctype lookup, undefined behavior for basically any non-ASCII input. Since these are the fast-path integer parsers behind `np.loadtxt`/`genfromtxt`, this looks reachable with ordinary UTF-8 content (an accented letter, CJK character, or emoji in a field numpy tries to parse as an int) rather than needing pathological input. `Py_CHARMASK` won't fix it since the value isn't a `char` to begin with — the same file already uses `Py_UNICODE_ISSPACE()` two lines above each `isdigit()` call for the same string, so `Py_UNICODE_ISDIGIT()` looks like the natural fix, matching the existing pattern in that file. Might be worth a follow-up, since it's the same bug class this PR is otherwise cleaning up.

isl-Ramzi:
Good catch, and you're right that it's the same bug class. It's a bit different in kind though: str_to_int.h is working on Py_UCS4 codepoints rather than char, so Py_CHARMASK wouldn't be the right tool there. Py_UNICODE_ISDIGIT is the natural fix, matching the Py_UNICODE_ISSPACE calls already sitting a couple lines above each isdigit. Since it's a separate file with a wider value range, I'd rather keep this PR focused on the char parsers and handle that as a follow-up so the two don't get tangled. Happy to open it once this one settles.

Gooh456:
Makes sense, that's a fair distinction — codepoints vs char really do need different tooling. Happy to keep this one scoped to the char parsers, I'll watch for the str_to_int.h follow-up.

PR #31960, opened 2026-07-11
BUG: Preserve f2py argv tokens containing spaces

### LINKED ISSUE #31958: BUG: f2py reparses tokenized argv entries containing spaces
﻿### Describe the issue:

F2PY receives already-tokenized command-line arguments from normal entry points such as `f2py` and `python -m numpy.f2py`, but one newer-option preprocessing path rebuilds that argv list as a string and then splits it on whitespace again.

In `numpy/f2py/f2py2e.py`, `get_newer_options()` currently does:

```python
iline = (' '.join(iline)).split()
```

That can destroy valid argv tokens that contain spaces. This matters for realistic Windows paths such as user profile directories, OneDrive for Business sync roots such as `OneDrive - <organization name>`, `Program Files`, include directories, and build directories. The issue is not that users pass an unquoted shell string; the shell/Python layer can already preserve quoted paths as one `sys.argv` element, and F2PY then splits that element internally.

This is reachable from normal user entry points:

```text
f2py / python -m numpy.f2py
  -> numpy.f2py.f2py2e:main()
  -> run_main(sys.argv[1:])        # non-compile/signature-generation path
  -> scaninputline(comline_list)
  -> get_newer_options(inputline)
```

The compile path also passes `sources = sys.argv[1:]` to `get_newer_options(sources)` later in `run_compile()`. There appear to be related argv-boundary risks in the Meson wrapper-generation path, where `run_compile()` constructs a command string with `' '.join(...)` and then calls `.split()` before `run_main()`.

```python
from numpy.f2py.f2py2e import get_newer_options

cases = [
    ["-I", r"C:\Users\First Last\include", r"C:\src\fib.f90"],
    ["--build-dir", r"C:\Users\First Last\f2py build", "-m", "fib", r"C:\src\fib.f90"],
    ["--freethreading-compatible", r"C:\Users\First Last\src\fib.f90"],
]

for case in cases:
    print("IN =", case)
    print("OUT=", get_newer_options(case))
```

On Windows, an additional path-list example is:

```python
from numpy.f2py.f2py2e import get_newer_options

print(get_newer_options([
    "--include-paths",
    r"C:\Users\First Last\include;D:\vendor headers",
    r"C:\src\fib.f90",
]))
```

### Error message:

For the first example, the include directory token is split at the space:

```shell
IN = ['-I', 'C:\\Users\\First Last\\include', 'C:\\src\\fib.f90']
OUT= (['C:\\Users\\First'], None, ['Last\\include', 'C:\\src\\fib.f90'])
```

For a build directory, the remainder passed to the legacy parser is also split:

```shell
IN = ['--build-dir', 'C:\\Users\\First Last\\f2py build', '-m', 'fib', 'C:\\src\\fib.f90']
OUT= ([], None, ['--build-dir', 'C:\\Users\\First', 'Last\\f2py', 'build', '-m', 'fib', 'C:\\src\\fib.f90'])
```

Expected behavior: already-tokenized argv entries should remain intact when passed to `argparse.parse_known_args()`. For example, `r"C:\Users\First Last\include"` should stay one token, not become `r"C:\Users\First"` plus `r"Last\include"`.

Observed while inspecting the current NumPy development checkout. The focused local reproduction used Python 3.12 in a conda environment and the repository source at `numpy/f2py/f2py2e.py`.

### Runtime Environment:

This is a command-line argument preprocessing issue and does not require compiled NumPy artifacts to reproduce. The local reproduction used a Windows environment, but the underlying token-boundary problem is platform-independent whenever an argv element contains whitespace.

I found this while reviewing F2PY option parsing after looking at nearby path-handling behavior. The practical impact is that users who correctly quote a path containing spaces can still have that path split inside F2PY after Python has already preserved it as one `sys.argv` entry.

This is most realistic on Windows, where normal paths may contain spaces, for example under `C:\Users\First Last\...`, `OneDrive - <organization name>\\...`, or `C:\\Program Files (x86)\\...`. The narrow behavior I would expect to preserve is already-tokenized argv entries; this does not require changing how F2PY parses arbitrary shell command strings.

reachsrinivaschennupati-tech:
## Summary

Fixes #31958.

`get_newer_options` receives an already-tokenized argv list from the normal f2py entry points, but it rebuilt that list as a string and split it again. That loses token boundaries for valid arguments containing spaces, such as Windows include paths, build directories, and source paths under user profile or OneDrive directories.

This PR passes the token list directly to `argparse.parse_known_args`, preserving already-tokenized argv entries while keeping the existing parsing behavior for `-I`, `--include-paths`, `--include_paths`, and `--freethreading-compatible`.

## Tests

- `python3 -m py_compile numpy/f2py/f2py2e.py numpy/f2py/tests/test_f2py2e.py`
- Focused parser harness against the patched `get_newer_options` definitions:
  - preserves `-I "C:\Users\First Last\include"`
  - preserves `--include-paths "C:\Users\First Last\include;D:\vendor headers"`
  - preserves `--build-dir "C:\Users\First Last\f2py build"` in the remaining argv while parsing `--freethreading-compatible`
- Verified the pre-fix `(' '.join(iline)).split()` behavior reproduces the reported token split.

### DIFF
--- a/numpy/f2py/f2py2e.py
+++ b/numpy/f2py/f2py2e.py
@@ -554,7 +554,6 @@ def f2py_parser():
     return parser
 
 def get_newer_options(iline):
-    iline = (' '.join(iline)).split()
     parser = f2py_parser()
     args, remain = parser.parse_known_args(iline)
     ipaths = args.include_paths
--- a/numpy/f2py/tests/test_f2py2e.py
+++ b/numpy/f2py/tests/test_f2py2e.py
@@ -754,6 +754,51 @@ def test_inclpath(monkeypatch):
     assert remain == []
 
 
+def test_get_newer_options_preserves_tokenized_arguments_with_spaces(monkeypatch):
+    """Preserve already-tokenized argv entries that contain spaces."""
+    monkeypatch.setattr(os, "pathsep", ";")
+
+    include_paths, ftcompat, remain = get_newer_options(
+        ["-I", r"C:\Users\First Last\include", r"C:\src\fib.f90"]
+    )
+
+    assert include_paths == [r"C:\Users\First Last\include"]
+    assert ftcompat is None
+    assert remain == [r"C:\src\fib.f90"]
+
+    include_paths, ftcompat, remain = get_newer_options(
+        [
+            "--include-paths",
+            r"C:\Users\First Last\include;D:\vendor headers",
+            r"C:\src\fib.f90",
+        ]
+    )
+
+    assert set(include_paths) == {
+        r"C:\Users\First Last\include",
+        r"D:\vendor headers",
+    }
+    assert ftcompat is None
+    assert remain == [r"C:\src\fib.f90"]
+
+    include_paths, ftcompat, remain = get_newer_options(
+        [
+            "--build-dir",
+            r"C:\Users\First Last\f2py build",
+            "--freethreading-compatible",
+            r"C:\Users\First Last\src\fib.f90",
+        ]
+    )
+
+    assert include_paths == []
+    assert ftcompat is True
+    assert remain == [
+        "--build-dir",
+        r"C:\Users\First Last\f2py build",
+        r"C:\Users\First Last\src\fib.f90",
+    ]
+
+
 def test_hlink():
     """Add to the include directories
 


jorenham:

reachsrinivaschennupati-tech:
@jorenham appologies, I have n't reviewed the AI policy, I used OpenAI Codex as an assistant to help inspect the issue and validate the local change. I reviewed the relevant code, made appropriate corrections, but used for commenting.

I will follow the AI policy going forward. Thanks for pointing this out.

PR #31961, opened 2026-07-11
BUG: make base_repr upper-bound error message track len(digits)

nikolauspschuetz:
`base_repr` derives its upper bound from the digit alphabet:

```python
digits = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ'
if base > len(digits):
    raise ValueError("Bases greater than 36 not handled in base_repr.")
```

The guard uses `len(digits)` specifically to avoid hardcoding the bound, but the error message re-hardcodes `36`. If `digits` is ever changed, the guard adapts while the message silently reports the stale number. This makes the message derive the bound the same way the guard does, via an f-string, so the two can't drift apart. The rendered text is identical today (`len(digits) == 36`) and behavior is unchanged.

`test_base_range` is tightened to assert the message (`assert_raises_regex(..., "Bases greater than 36")`) instead of just the exception type, pinning the reported bound.

### DIFF
--- a/numpy/_core/numeric.py
+++ b/numpy/_core/numeric.py
@@ -2145,7 +2145,9 @@ def base_repr(number, base=2, padding=0):
     """
     digits = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ'
     if base > len(digits):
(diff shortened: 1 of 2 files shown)


jorenham:

PR #31962, opened 2026-07-11
BUG: drop stray "ValueError:" prefix from einsum order error message

nikolauspschuetz:
`_parse_output_order` (reached from `np.einsum(..., order=..., optimize=True)`) raised a `ValueError` whose message string itself began with `"ValueError: "`:

```python
raise ValueError(
    "ValueError: order must be one of "
    f"'C', 'F', 'A', or 'K' (got '{order}')"
)
```

So an invalid order rendered as `ValueError: ValueError: order must be one of 'C', 'F', 'A', or 'K' (got 'Z')` — the class name embedded inside its own message, a copy-paste artifact. This drops the stray prefix. The enumerated set is unchanged and already correct (`K`, `C`, `F`, `A` are exactly the accepted orders).

A regression test (`test_invalid_output_order`) asserts the message with a start-anchored pattern (`^order must be one of ...`), which fails on the doubled prefix and passes once it's removed.

### DIFF
--- a/numpy/_core/einsumfunc.py
+++ b/numpy/_core/einsumfunc.py
@@ -1136,8 +1136,7 @@ def _parse_output_order(order, a_is_fcontig, b_is_fcontig):
             return "C"
     else:
         raise ValueError(
-            "ValueError: order must be one of "
-            f"'C', 'F', 'A', or 'K' (got '{order}')"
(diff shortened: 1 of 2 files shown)


jorenham:

PR #26634, opened 2024-06-06
BUG: Replace dots with underscores in f2py meson backend for lib identifiers

mathomp4:
As seen in #26623, if a library itself has a dot, e.g., `-lMAPL.base`, then the meson backend produces code like:
```meson
MAPL.base = declare_dependency(link_args : ['-lMAPL.base'])
```
But meson apparently sees `MAPL.base` as saying this is the `base` method in the `MAPL` type (or class?). Thus, it expects parentheses. 

So this PR uses a `replace()` call to replace all the `.` to `_` in the `lib` identifier. This has to be done both in the `declare_dependency` and in the `dependencies` bit of `py.extension_module`.

With this fix I get:
```meson
MAPL_constants = declare_dependency(link_args : ['-lMAPL.constants'])
MAPL_base = declare_dependency(link_args : ['-lMAPL.base'])
```

This seems to fix any issues in my testing, but my fear is cases I don't have. (Though perhaps libraries with dots is a rare case anyway and mainly a style used in the our code.)

Closes #26623 

---

NOTE: I tested this with Numpy 1.26.4, and I see main is now in v2 land. It would be nice to get this into a v1 tag as well if v2 is still a ways off. I could backport this to `maintenance/1.26.x` if desired.

PR #31963, opened 2026-07-11
BUG: avoid invalid Meson identifiers for f2py libraries

### LINKED ISSUE #31964: F2PY Meson backend generates invalid identifiers for some -l library names
### Describe the issue:

The f2py Meson backend currently derives Meson variable identifiers from `-l` library names. [#26634](https://github.com/numpy/numpy/pull/26634) fixed one instance of this by replacing dots with underscores, but other linker-valid library names can still produce invalid or ambiguous Meson identifiers.

For example, a library name such as `scalapack-openmpi` is valid as a linker argument (`-lscalapack-openmpi`) but is not valid as a Meson assignment target if used directly as the generated variable name:

```meson
scalapack-openmpi = declare_dependency(link_args : ['-lscalapack-openmpi'])
```

Meson rejects that because the assignment target must be an identifier. Similar problems can occur for names beginning with digits, and multiple different library names can collapse to the same underscore form if sanitization is not made unique.

A proposed fix is open in [#31963](https://github.com/numpy/numpy/pull/31963). It generates private indexed Meson identifiers while preserving the original `-l<name>` linker arguments.

A representative f2py invocation is:

```bash
f2py -c module.f90 -lscalapack-openmpi --backend meson
```

The relevant call chain is:

```text
f2py CLI args
  -> run_compile() extracts -l entries into libraries
  -> MesonBackend.write_meson_build()
  -> MesonTemplate.libraries_substitution()
  -> generated meson.build lib_declarations/lib_list
```

### Error message:

Meson can reject the generated `meson.build` with a parse error similar to:

```text
ParseException: Assignment target must be an id.
```

Observed on the current NumPy development branch while inspecting the generated f2py Meson backend output. This is a source-generation issue and does not depend on a completed NumPy build.

### Runtime Environment:

Windows local checkout; the affected generated Meson syntax is platform-independent.

VectorPeak:
### PR summary

Fixes #31964.

Fixes invalid Meson variable names generated by the f2py Meson backend for `-l` libraries whose names are valid linker library names but are not valid Meson identifiers.

#26634 handled dotted library names by replacing `.` with `_`, but other inputs can still break or become ambiguous when the raw library name is used as the Meson variable name. For example, a library such as `scalapack-openmpi` can lead to an invalid assignment target in the generated `meson.build`.

This PR separates the internal Meson variable name from the user-provided linker library name. The generated dependency variable now uses a private indexed identifier, while the linker argument still preserves the original value, e.g. `-lscalapack-openmpi` remains `-lscalapack-openmpi`.

The regression test parses the generated `meson.build` and asserts that the generated dependency declarations preserve the original `-l<name>` linker arguments, while the unsafe original library names are not used directly as Meson assignment targets. It covers dotted, hyphenated, realistic hyphenated, underscore/collision, and digit-leading library names:

- `foo.bar`
- `scalapack-openmpi`
- `foo-bar`
- `foo_bar`
- `1foo`

Validation performed locally:

- `git diff --check`
- `spin lint`
- A direct `MesonTemplate` parser check using `mesonbuild.mparser.Parser(...).parse()`, which completed with `parse-ok` and verified that generated dependency declarations preserve the original `-l...` arguments.

I also attempted the focused pytest command below, but this unbuilt local source checkout stops during NumPy test configuration before selected tests run because the generated `numpy.version` module is unavailable:

```bash
D:\ZXY\Dev\Miniforge3\envs\prw-numpy-py312\python.exe -m pytest numpy/f2py/tests/test_regression.py -q -k "meson_library_names or gh26623"
```

N/A.

#### AI Disclosure

Codex was used to investigate the failing f2py/Meson code path, draft the small code and regression-test change, run focused local validation, and draft this PR description. The diff and submission were reviewed by a human before opening the PR.

### DIFF
--- a/numpy/f2py/_backends/_meson.py
+++ b/numpy/f2py/_backends/_meson.py
@@ -11,6 +11,13 @@ from string import Template
 from ._backend import Backend
 
 
+def _meson_identifier(prefix: str, value: str, index: int) -> str:
+    safe_value = re.sub(r"[^a-zA-Z0-9_]", "_", value)
+    if not safe_value or safe_value[0].isdigit():
+        safe_value = f"_{safe_value}"
+    return f"{prefix}_{index}_{safe_value}"
+
+
 class MesonTemplate:
     """Template meson build file generation class."""
 
@@ -91,6 +98,11 @@ class MesonTemplate:
         )
 
     def libraries_substitution(self) -> None:
+        lib_names = [
+            (_meson_identifier("lib", lib, i), lib)
+            for i, lib in enumerate(self.libraries)
+        ]
+
         self.substitutions["lib_dir_declarations"] = "\n".join(
             [
                 f"lib_dir_{i} = declare_dependency(link_args : ['''-L{lib_dir}'''])"
@@ -100,13 +112,13 @@ class MesonTemplate:
 
         self.substitutions["lib_declarations"] = "\n".join(
             [
-                f"{lib.replace('.', '_')} = declare_dependency(link_args : ['-l{lib}'])"
-                for lib in self.libraries
+                f"{name} = declare_dependency(link_args : ['-l{lib}'])"
+                for name, lib in lib_names
             ]
         )
 
         self.substitutions["lib_list"] = f"\n{self.indent}".join(
-            [f"{self.indent}{lib.replace('.', '_')}," for lib in self.libraries]
+            [f"{self.indent}{name}," for name, _ in lib_names]
         )
         self.substitutions["lib_dir_list"] = f"\n{self.indent}".join(
             [f"{self.indent}lib_dir_{i}," for i in range(len(self.library_dirs))]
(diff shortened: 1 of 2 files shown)


HaoZeke (changes requested):
Thanks for this @VectorPeak. 
#26634 only handled dots so `scalapack-openmpi` wouldn't work.
Separating the `meson` id from the `-l` makes sense.

I'd like a regression test to demonstrate the linker still gets `-l` and then this should be done :)

VectorPeak:
Thanks @HaoZeke, I updated the regression test to check the generated Meson dependency declarations.

It now includes `scalapack-openmpi`, parses the generated `meson.build`, and verifies the original `-l<name>` link args are preserved while unsafe names are not emitted directly as Meson assignment targets.

HaoZeke (approved):
Thanks @VectorPeak, LGTM, in it goes

PR #31966, opened 2026-07-11
TST: skip limited API tests on incompatible ABIs

kumaraditya303:
### PR summary

Previously these tests where marked `xfail` whereas I think it makes more sense to skip it instead. 

#### AI Disclosure

None


rgommers (approved):
Agreed, not useful to run slow tests over and over when we know the answer won't change.

rgommers:
Thanks Kumar!

PR #31968, opened 2026-07-11
BUG: fix refcount leak on overlapping `copyto` with `where=False`

Ethan0456:
### PR summary

This PR fixes a reference leak in `PyArray_AssignArray` (used by `np.copyto` and related assign paths) when:

1. `src`/`dst` overlap, so an intermediate temp is allocated (`copied_src = 1`), and
2. a scalar boolean `where=False` triggers an early return that used to skip `Py_DECREF` of that temp.

Success and failure paths already released the temp; the early-return path did not. With object dtypes, filling the temp `INCREF`s every element, so those references (and the temp array) were leaked.

**Fix:**
`Py_DECREF(src)` on the `where=False` early return when `copied_src` is set. No change to assignment semantics (`where=False` still writes nothing).

**Test:**
Regression test in `test_api.py` that checks object refcounts after overlapping `np.copyto(..., where=False)`.

No release note (internal leak only; no user-visible API/behavior change).

I've been exploring the NumPy codebase recently, finding and fixing bugs. I had one PR merged last week and am continuing with small correctness fixes like this.

### AI Disclosure

AI was used to help locate and draft the fix. The bug, fix, and tests were reviewed and validated by a human, and this PR was prepared and submitted by human.

### DIFF
--- a/numpy/_core/src/multiarray/array_assign_array.c
+++ b/numpy/_core/src/multiarray/array_assign_array.c
@@ -440,7 +440,10 @@ PyArray_AssignArray(PyArrayObject *dst, PyArrayObject *src,
             wheremask = NULL;
         }
         else {
-            /* where=False copies nothing */
+            /* where=False copies nothing; still free any overlap temp */
+            if (copied_src) {
+                Py_DECREF(src);
+            }
             return 0;
         }
     }
--- a/numpy/_core/src/multiarray/array_assign_array.c
+++ b/numpy/_core/src/multiarray/array_assign_array.c
@@ -440,7 +440,7 @@ PyArray_AssignArray(PyArrayObject *dst, PyArrayObject *src,
             wheremask = NULL;
         }
         else {
-            /* where=False copies nothing; still free any overlap temp */
+            /* where=False copies nothing */
             if (copied_src) {
                 Py_DECREF(src);
             }
--- a/numpy/_core/tests/test_api.py
+++ b/numpy/_core/tests/test_api.py
@@ -446,26 +446,16 @@ def test_copyto():
     assert_raises(TypeError, np.copyto, [1, 2, 3], [2, 3, 4])
 
 
-@pytest.mark.skipif(not HAS_REFCOUNT, reason="Python lacks refcounts")
-def test_copyto_overlapping_where_false_no_refcount_leak():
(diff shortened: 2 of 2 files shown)


numpy/_core/tests/test_api.py
```diff
@@ -446,6 +446,28 @@ def test_copyto():
...
+            a[i, j] = object()
+
+    victim = a[0, 0]
+    before = sys.getrefcount(victim)
+    np.copyto(a, a[::-1, :], where=False)
+    after = sys.getrefcount(victim)
+    assert after == before, (
+        f"refcount leak after overlapping copyto where=False: "
+        f"before={before}, after={after}"
+    )
```
ikrommyd: I'd suggest not asserting on the refcount directly but adding a test that would just have the leak and could be caught by a leak sanitizer. See https://github.com/numpy/numpy/pull/31699 and https://github.com/numpy/numpy/pull/31731 for examples.

ikrommyd (changes requested):

numpy/_core/tests/test_api.py
Ethan0456: Thanks for the pointer. Dropped the `getrefcount` assert and switched to a sanitizer-style path test following #31699 and #31731

numpy/_core/src/multiarray/array_assign_array.c
```diff
@@ -440,7 +440,10 @@ PyArray_AssignArray(PyArrayObject *dst, PyArrayObject *src,
             wheremask = NULL;
         }
         else {
-            /* where=False copies nothing */
+            /* where=False copies nothing; still free any overlap temp */
```
kumaraditya303: nit: the code is self-explanatory, you don't need to change the comment

Ethan0456: Sure, I have restored the original comment.

kumaraditya303 (approved):
LGTM, thanks.

Ethan0456:
Hi @kumaraditya303, thanks for reviewing and approving this PR! Since the requested changes have been addressed and the CI is passing, I wanted to check if there's anything else needed before it can be merged. If not, I'd appreciate it if you could help move it forward when you have a chance. Thanks!

mattip:
Thanks @Ethan0456

PR #30496, opened 2025-12-21
ENH: Hash-based np.unique with support for return_index/return_inverse/return_counts

math-hiyoko:
### summary
This PR updates np.unique so that the hash-based implementation is used when any of
return_index, return_inverse, or return_counts is requested.
#28374 

### Benchmark observations
From local benchmarks, most cases look reasonable. However, for integer dtypes, the case where only return_counts=True (UniqueIntegers.time_unique_counts) is requested shows a noticeable slowdown compared to the existing sort-based implementation (up to ~2–5× in some cases, see table below).

### Possible follow-up improvement
~One potential improvement would be to keep using the original sort-based path specifically for the “integer + counts-only” case, while continuing to use the hash-based implementation for all other combinations (especially those requiring index and/or inverse support).~

~This would be a very lightweight change, as it only involves adjusting the relevant entry in unique_funcs, without modifying the core implementation.~

Update: The sort-based path is now explicitly used for integer dtypes when only return_counts is requested.
This is implemented via a small dispatch change (adjusting the corresponding entry in unique_funcs), without modifying the core implementation.

### Note on no-option integer case
I also observed a small regression for integer inputs even when no optional outputs are requested. This is likely due to the additional branching and function call overhead introduced by the new dispatch logic.

In absolute terms, the difference is on the order of a few milliseconds, so this seems acceptable, but I wanted to mention it explicitly for completeness.

close #28374 
close #30113

PR #31970, opened 2026-07-11
ENH: Optimize `np.unique` with adaptive thresholding for hashing vs sorting

### LINKED ISSUE #31969: PERF: `np.unique` has massive performance regression for large numeric arrays due to `_unique_hash`
#### Description
In NumPy 2.x, a hash-based unique algorithm (`_unique_hash` using C++ `std::unordered_set`) was introduced (PR #26018) to find unique values. However, for numeric arrays (e.g. `int64`, `int32`), this has introduced a massive performance regression at medium-to-large scales compared to sorting. Hashing a 1M element `int64` array takes over **712 ms**, whereas sorting-based unique (which is SIMD-accelerated via `x86-simd-sort` in NumPy 2.x) takes only **15.2 ms** (a **46x difference**!).

#### Why this happens
1. **Node Allocations**: `std::unordered_set` is a node-based container. Inserting 1,000,000 unique values incurs 1,000,000 heap memory allocations.
2. **Cache Locality**: Linked lists in the bucket array lead to significant CPU cache misses.
3. **No Vectorization**: Hashing/lookup cannot be vectorized, whereas sorting uses AVX-512/AVX2 registers to process multiple elements in parallel.

Even when there are heavy duplicates (e.g., 10 unique elements repeated 100,000 times each), sorting is still **3x faster** than hashing (15.8 ms vs 45.7 ms). Hashing only beats sorting at micro scales (e.g., `< 50` elements) where Python-level sorting setup overhead dominates.

#### Proposed Solution
Introduce adaptive thresholding: for numeric and fixed-width dtypes, route to the hash-based pathway only if the size is small (e.g., `< 100` elements). For larger sizes, bypass `_unique_hash` and use the sorting pathway. For variable-width string types (`StringDType`), we keep the hashing pathway at all sizes, as hashing remains faster than string comparisons.

rajat315315:
Closes: #31969 

#### Proposed Changes
This PR implements adaptive thresholding for `np.unique` to select the most efficient pathway (hashing vs sorting) based on the input dtype and size:
* **`StringDType`**: Always routed to `_unique_hash` as hashing variable-width strings remains faster than sorting.
* **All other dtypes** (numeric, complex, fixed-width unicode): Routed to `_unique_hash` only if the array size is **`< 100`** elements (where Python/C setup overhead of sorting dominates). For larger sizes, it directly uses the SIMD-accelerated sorting-based unique pathway.

#### Benchmarks (1,000,000 elements)

| Dtype / Scenario | Before (ms) | After (ms) | Speedup | Pathway Used |
| :--- | :---: | :---: | :---: | :---: |
| `int64` (All unique) | 712.6 ms | 15.2 ms | **46x** | Sorting (After) vs Hashing (Before) |
| `int64` (10 unique) | 45.7 ms | 15.8 ms | **2.9x** | Sorting (After) vs Hashing (Before) |
| `'U15'` (Distinct strings) | 99.4 ms | 12.2 ms | **8.1x** | Sorting (After) vs Hashing (Before) |
| `StringDType` (1M repeated) | 99.5 ms | 99.5 ms | **1.0x** | Hashing (Unchanged) |

#### Benchmarks (10 elements)

| Dtype | Before (us) | After (us) | Speedup | Pathway Used |
| :--- | :---: | :---: | :---: | :---: |
| `int64` | 6.0 us | 6.0 us | **1.0x** | Hashing (Unchanged) |

#### Verification
All set operation tests pass:
```bash
pytest numpy/lib/tests/test_arraysetops.py
```
**Output**: `117 passed in 0.73s`

I use it a lot for data chuking.

#### AI Disclosure

Used Gemini 3.1 Pro in benchmarking the results.

### DIFF
--- a/numpy/lib/_arraysetops_impl.py
+++ b/numpy/lib/_arraysetops_impl.py
@@ -370,7 +370,7 @@ def _unique1d(ar, return_index=False, return_inverse=False,
         # Hashing is faster for variable-width StringDType.
         # For other types (numeric, fixed-width string/unicode), sorting is
         # faster at larger scales, but hashing is faster at small scales (size < _UNIQUE_HASH_THRESHOLD).
-        is_vstring = getattr(ar.dtype, "__class__", None).__name__ == "StringDType"
+        is_vstring = isinstance(ar.dtype, np.dtypes.StringDType)
         if is_vstring or len(ar) < _UNIQUE_HASH_THRESHOLD:
             # First we convert the array to a numpy array, later we wrap it back
             # in case it was a subclass of numpy.ndarray.
--- a/numpy/lib/_arraysetops_impl.py
+++ b/numpy/lib/_arraysetops_impl.py
@@ -32,6 +32,8 @@ __all__ = [
     "unique_values"
 ]
 
+_UNIQUE_HASH_THRESHOLD = 100
+
 
 def _ediff1d_dispatcher(ary, to_end=None, to_begin=None):
     return (ary, to_end, to_begin)
@@ -367,16 +369,16 @@ def _unique1d(ar, return_index=False, return_inverse=False,
     if not optional_indices and not return_counts and not np.ma.is_masked(ar):
         # Hashing is faster for variable-width StringDType.
         # For other types (numeric, fixed-width string/unicode), sorting is
-        # faster at larger scales, but hashing is faster at small scales (size < 100).
+        # faster at larger scales, but hashing is faster at small scales (size < _UNIQUE_HASH_THRESHOLD).
         is_vstring = getattr(ar.dtype, "__class__", None).__name__ == "StringDType"
-        if is_vstring or len(ar) < 100:
+        if is_vstring or len(ar) < _UNIQUE_HASH_THRESHOLD:
             # First we convert the array to a numpy array, later we wrap it back
             # in case it was a subclass of numpy.ndarray.
             conv = _array_converter(ar)
             ar_, = conv
 
-            if (hash_unique := _unique_hash(ar_, equal_nan=equal_nan)) \
-                is not NotImplemented:
+            hash_unique = _unique_hash(ar_, equal_nan=equal_nan)
+            if hash_unique is not NotImplemented:
                 if sorted:
                     hash_unique.sort()
                 # We wrap the result back in case it was a subclass of numpy.ndarray.
--- a/numpy/lib/_arraysetops_impl.py
+++ b/numpy/lib/_arraysetops_impl.py
@@ -365,17 +365,22 @@ def _unique1d(ar, return_index=False, return_inverse=False,
 
     # masked arrays are not supported yet.
     if not optional_indices and not return_counts and not np.ma.is_masked(ar):
-        # First we convert the array to a numpy array, later we wrap it back
-        # in case it was a subclass of numpy.ndarray.
-        conv = _array_converter(ar)
-        ar_, = conv
+        # Hashing is faster for variable-width StringDType.
+        # For other types (numeric, fixed-width string/unicode), sorting is
+        # faster at larger scales, but hashing is faster at small scales (size < 100).
+        is_vstring = getattr(ar.dtype, "__class__", None).__name__ == "StringDType"
+        if is_vstring or len(ar) < 100:
+            # First we convert the array to a numpy array, later we wrap it back
+            # in case it was a subclass of numpy.ndarray.
+            conv = _array_converter(ar)
+            ar_, = conv
 
-        if (hash_unique := _unique_hash(ar_, equal_nan=equal_nan)) \
-            is not NotImplemented:
-            if sorted:
-                hash_unique.sort()
-            # We wrap the result back in case it was a subclass of numpy.ndarray.
-            return (conv.wrap(hash_unique),)
+            if (hash_unique := _unique_hash(ar_, equal_nan=equal_nan)) \
+                is not NotImplemented:
+                if sorted:
+                    hash_unique.sort()
+                # We wrap the result back in case it was a subclass of numpy.ndarray.
+                return (conv.wrap(hash_unique),)
 
     # If we don't use the hash map, we use the slower sorting method.
     if optional_indices:


numpy/lib/_arraysetops_impl.py
```diff
@@ -365,17 +365,22 @@ def _unique1d(ar, return_index=False, return_inverse=False,
...
-            is not NotImplemented:
-            if sorted:
-                hash_unique.sort()
-            # We wrap the result back in case it was a subclass of numpy.ndarray.
-            return (conv.wrap(hash_unique),)
+        # Hashing is faster for variable-width StringDType.
+        # For other types (numeric, fixed-width string/unicode), sorting is
+        # faster at larger scales, but hashing is faster at small scales (size < 100).
+        is_vstring = getattr(ar.dtype, "__class__", None).__name__ == "StringDType"
+        if is_vstring or len(ar) < 100:
```
ngoldbaum: This `100` should be an ALL_CAPS constant defined at module scope.

It's also not at all clear to me that 100 is correct for all OS/CPU/Compiler combinations - did you try benchmarking on more than one system?

numpy/lib/_arraysetops_impl.py
```diff
@@ -365,17 +365,22 @@ def _unique1d(ar, return_index=False, return_inverse=False,
...
-        if (hash_unique := _unique_hash(ar_, equal_nan=equal_nan)) \
-            is not NotImplemented:
-            if sorted:
-                hash_unique.sort()
-            # We wrap the result back in case it was a subclass of numpy.ndarray.
-            return (conv.wrap(hash_unique),)
+        # Hashing is faster for variable-width StringDType.
+        # For other types (numeric, fixed-width string/unicode), sorting is
+        # faster at larger scales, but hashing is faster at small scales (size < 100).
+        is_vstring = getattr(ar.dtype, "__class__", None).__name__ == "StringDType"
```
ngoldbaum: This is a weird way to do this, why not use `issubclass`?

numpy/lib/_arraysetops_impl.py
```diff
@@ -365,17 +365,22 @@ def _unique1d(ar, return_index=False, return_inverse=False,
...
+        if is_vstring or len(ar) < 100:
+            # First we convert the array to a numpy array, later we wrap it back
+            # in case it was a subclass of numpy.ndarray.
+            conv = _array_converter(ar)
+            ar_, = conv
+
+            if (hash_unique := _unique_hash(ar_, equal_nan=equal_nan)) \
+                is not NotImplemented:
```
ngoldbaum: I realize this was already there, but can you do this without the walrus operator? It's a little ugly with the line continuation

ngoldbaum (commented):
I worry a bit that the static threshold you're adding needs to be tunable or do something a little smarter. I also wonder if the performance difference depend on the number of unique values in the data or otherwise on the structure of the data.

Your PR description has benchmarks but it's not at all clear to me what the nature of the benchmarks you're running are.

rajat315315:
To make the global constant smarter, we can either initialize a dict for different dtypes.

```
_UNIQUE_HASH_THRESHOLDS = {
    'i': 100,  # integers
    'f': 120,  # floating points
    # ...
}
```

Or let numpy pick this value from Environment variable, which I think would make it too user dependent.

ptim0626:
I have tried to use the script below to see if I can observe a similar speed-up when the array's values are all unique:
```python
import time
import numpy as np

print(np)

rng = np.random.default_rng(721)
arr = np.arange(0, 1_000_000)
rng.shuffle(arr)

time_start = time.perf_counter()
ret = np.unique(arr)
time_elapsed = (time.perf_counter() - time_start)
print(f"{time_elapsed * 1000:5.3f} ms")
```
After introducing the threshold it is **42-46 ms**, and before the change was **143-150 ms** in my machine, which is about **3.5x** speed-up. 

For the 10-unique-element case:
```python
import time
import numpy as np

rng = np.random.default_rng(721)
arr = np.empty(1_000_000, dtype=int)
sz = arr.size

for k in range(11):
    st, ed = sz//10 * (k-1), sz//10 * k
    arr[st:ed] = rng.integers(0, 1_000_000)

time_start = time.perf_counter()
ret = np.unique(arr)
time_elapsed = (time.perf_counter() - time_start)
print(f"{time_elapsed * 1000:5.3f} ms")
```
They are both about **14-16 ms** on my machine (Apple M2), so I don't see a speed-up in this case. I will try to run the above two in a fresh env later to see if I get similar results.

~~Not sure if it is related to a hard-coded `HASH_TABLE_INITIAL_BUCKETS` [here](https://github.com/numpy/numpy/blob/main/numpy/_core/src/multiarray/unique.cpp#L35). The current threshold of 100 with 8-byte data type is below 1024, so some resizing may be happening if more than the current threshold.~~
EDIT: seems not related as they are pointers.

adrinjalali:
This would basically pretty much completely remove the benefit from hash based unique we see on the scikit-learn side, where we call unique on large arrays of low cardinality unique values.

So the threshold is not based on array size, rather on unique value count. This also brings out the discussion on #30867 and it's interesting to see its effect on this benchmark.

I ran a benchmark to compare sort vs hash and also used #30867 to compare how it does in the context of this discussion. Putting the benchmark code and what the two end up showing.

tl;dr; is that there's a slowdown in some cases, where there are many unique values. But I'm not sure if that case is where we have most unique calls. We could introduce a "method" of sort to control the backend algorithm used to calculate the hash, but from what I see here, hash based unique makes more sense to be the default.

Hope this helps.

-----------------

Note, bellow is all LLM generated.

Benchmark code: <details>

```py
"""Benchmark hash-based vs sort-based np.unique kernels; write results to JSON.

Context
-------
- numpy#31969 / PR numpy#31970: claims sorting beats the hash-based unique
  (introduced in numpy#26018) for large arrays, and adds a size threshold
  (< 100 elements -> hash, otherwise sort, StringDType always hash).
- PR numpy#30867 (on top of numpy#30496): replaces std::unordered_set with
  absl::flat_hash_set and extends the hash path to
  return_index/return_inverse/return_counts.

This script measures the two kernels directly (bypassing the Python dispatch)
so the same grid can be run under any build. Run it once per build:

    PYTHONPATH=<checkout>/build-install/usr/lib/python3.12/site-packages \
        python bench_np_unique.py [--full] [--out results.json]

The default output file is bench_results_<kernel>.json where <kernel> is
"std" (main / PR-31970 branch) or "absl" (PR-30867 branch), auto-detected.
Then merge the runs into a paste-ready report:

    python summarize_np_unique.py bench_results_std.json bench_results_absl.json
"""

import argparse
import json
import platform
import re
import subprocess
import sys
import time

import numpy as np
from numpy._core._multiarray_umath import _unique_hash
from numpy.lib import _arraysetops_impl as _impl

rng = np.random.default_rng(0)

# ----------------------------------------------------------------------
# build feature detection
# ----------------------------------------------------------------------
_probe = np.array([1, 2], dtype=np.int64)
try:
    # PR-30867 signature: returns (values, index, inverse, counts)
    _unique_hash(_probe, equal_nan=True, return_index=False,
                 return_inverse=False, return_counts=False)
    HASH_RETURNS_TUPLE = True
except TypeError:
    HASH_RETURNS_TUPLE = False

KERNEL = "absl" if HASH_RETURNS_TUPLE else "std"
HAS_31970_THRESHOLD = hasattr(_impl, "_UNIQUE_HASH_THRESHOLD")


def cpu_model():
    if platform.system() == "Linux":
        with open("/proc/cpuinfo") as f:
            m = re.search(r"model name\s*:\s*(.*)", f.read())
            if m:
                return m.group(1)
    try:
        return subprocess.run(["sysctl", "-n", "machdep.cpu.brand_string"],
                              capture_output=True, text=True).stdout.strip()
    except Exception:
        return platform.processor()


# ----------------------------------------------------------------------
# the two kernels, called directly
# ----------------------------------------------------------------------
def hash_unique(ar):
    """The hash kernel, unsorted output (np.unique_values semantics)."""
    if HASH_RETURNS_TUPLE:
        res = _unique_hash(ar, equal_nan=True, return_index=False,
                           return_inverse=False, return_counts=False)
        if res is NotImplemented:
            raise RuntimeError(f"hash NotImplemented for {ar.dtype}")
        return res[0]
    res = _unique_hash(ar, equal_nan=True)
    if res is NotImplemented:
        raise RuntimeError(f"hash NotImplemented for {ar.dtype}")
    return res


def sort_unique(ar):
    """The sort branch of _unique1d (flatten always copies, sort is inplace)."""
    ar = ar.flatten()
    ar.sort()
    mask = np.empty(ar.shape, dtype=np.bool_)
    mask[:1] = True
    mask[1:] = ar[1:] != ar[:-1]
    return ar[mask]


def hash_unique_inverse(ar):
    res = _unique_hash(ar, equal_nan=True, return_index=False,
                       return_inverse=True, return_counts=False)
    if res is NotImplemented:
        raise RuntimeError(f"hash NotImplemented for {ar.dtype}")
    return res[0], res[2]


def sort_unique_inverse(ar):
    """The argsort branch of _unique1d (what return_inverse costs today)."""
    ar = ar.flatten()
    perm = ar.argsort(kind="quicksort")
    aux = ar[perm]
    mask = np.empty(aux.shape, dtype=np.bool_)
    mask[:1] = True
    mask[1:] = aux[1:] != aux[:-1]
    imask = np.cumsum(mask) - 1
    inv_idx = np.empty(mask.shape, dtype=np.intp)
    inv_idx[perm] = imask
    return aux[mask], inv_idx


# ----------------------------------------------------------------------
# data + timing helpers
# ----------------------------------------------------------------------
def make_data(dtype, n, n_unique):
    if n_unique == "all":
        n_unique = n
    if dtype in ("int8", "int32", "int64"):
        base = np.arange(n_unique)
        vals = rng.choice(base, size=n) if n_unique < n else rng.permutation(base)
        return vals.astype(dtype)
    # random 12-char ascii strings
    pool = np.array(
        ["".join(chr(c) for c in row)
         for row in rng.integers(97, 123, size=(min(n_unique, n), 12))]
    )
    arr = pool[rng.integers(0, len(pool), size=n)]
    if dtype == "StringDType":
        return arr.astype(np.dtypes.StringDType())
    return arr.astype(dtype)


def timeit(fn, ar, repeats):
    best = float("inf")
    for _ in range(repeats):
        t0 = time.perf_counter()
        r = fn(ar)
        best = min(best, time.perf_counter() - t0)
    return best, r


def log(msg):
    print(msg, file=sys.stderr, flush=True)


def grid(section, dtypes, sizes, uniques):
    rows = []
    for dtype in dtypes:
        for n in sizes:
            for nu in uniques:
                if nu != "all" and nu > n:
                    continue
                ar = make_data(dtype, n, nu)
                repeats = 7 if n <= 1_000_000 else 3
                th, rh = timeit(hash_unique, ar, repeats)
                ts, rs = timeit(sort_unique, ar, repeats)
                assert len(rh) == len(rs)
                rows.append({"section": section, "dtype": dtype, "n": n,
                             "n_unique": nu, "hash_ms": th * 1e3,
                             "sort_ms": ts * 1e3})
                log(f"  {dtype} n={n:,} n_unique={nu}: "
                    f"hash {th*1e3:.2f} ms, sort {ts*1e3:.2f} ms")
    return rows


# ----------------------------------------------------------------------
# sections
# ----------------------------------------------------------------------
def sklearn_scenarios(full):
    n = 10_000_000 if full else 1_000_000
    classes = np.array([f"class_{i}" for i in range(100)])
    return n, [
        ("int64, 5 classes", rng.integers(0, 5, n)),
        ("int64, 100 classes", rng.integers(0, 100, n)),
        ("<U9, 5 classes", classes[rng.integers(0, 5, n)]),
        ("<U9, 100 classes", classes[rng.integers(0, 100, n)]),
    ]


def section_sklearn(full):
    rows = []
    n, scenarios = sklearn_scenarios(full)
    for name, y in scenarios:
        th, _ = timeit(hash_unique, y, 5)
        ts, _ = timeit(sort_unique, y, 5)
        te, _ = timeit(np.unique_values, y, 5)
        row = {"scenario": name, "n": n, "hash_ms": th * 1e3,
               "sort_ms": ts * 1e3, "unique_values_ms": te * 1e3}
        if HAS_31970_THRESHOLD:
            _impl._UNIQUE_HASH_THRESHOLD = 100
            t_pr, _ = timeit(np.unique_values, y, 5)
            _impl._UNIQUE_HASH_THRESHOLD = float("inf")
            t_old, _ = timeit(np.unique_values, y, 5)
            _impl._UNIQUE_HASH_THRESHOLD = 100
            row["pr31970_ms"] = t_pr * 1e3
            row["pre_pr31970_ms"] = t_old * 1e3
        rows.append(row)
        log(f"  sklearn e2e {name}: " + ", ".join(
            f"{k}={v:.1f}" for k, v in row.items() if k.endswith("_ms")))
    return rows


def section_inverse(full):
    if not HASH_RETURNS_TUPLE:
        return []
    rows = []
    n = 10_000_000 if full else 1_000_000
    scenarios = [
        ("int64, 5 classes", rng.integers(0, 5, n)),
        ("int64, 100 classes", rng.integers(0, 100, n)),
        ("int64, all unique", rng.permutation(n)),
    ]
    for name, y in scenarios:
        th, (vh, ih) = timeit(hash_unique_inverse, y, 3)
        ts, (vs, is_) = timeit(sort_unique_inverse, y, 3)
        assert len(vh) == len(vs)
        assert (vh[ih] == vs[is_]).all()
        rows.append({"scenario": name, "n": n, "hash_inverse_ms": th * 1e3,
                     "argsort_inverse_ms": ts * 1e3})
        log(f"  inverse {name}: hash {th*1e3:.1f} ms, argsort {ts*1e3:.1f} ms")
    return rows


if __name__ == "__main__":
    p = argparse.ArgumentParser()
    p.add_argument("--full", action="store_true",
                   help="also run the 10M-element grids (slower)")
    p.add_argument("--out", default=None,
                   help="output JSON path (default bench_results_<kernel>.json)")
    args = p.parse_args()
    out = args.out or f"bench_results_{KERNEL}.json"

    log(f"numpy {np.__version__}, hash kernel: {KERNEL}, writing to {out}")

    results = {
        "meta": {
            "numpy_version": np.__version__,
            "kernel": KERNEL,
            "kernel_desc": ("absl::flat_hash_set (PR 30867)" if KERNEL == "absl"
                            else "std::unordered_set (main / PR 31970)"),
            "has_pr31970_threshold": HAS_31970_THRESHOLD,
            "cpu": cpu_model(),
            "machine": platform.machine(),
            "full": args.full,
        },
    }

    log("section 1: cardinality regimes (int64)")
    sizes = [100_000, 1_000_000, 10_000_000] if args.full else [1_000_000]
    results["regimes"] = grid("regimes", ["int64"], sizes,
                              [2, 10, 100, 10_000, "all"])

    log("section 2: crossover sweep")
    results["crossover"] = grid(
        "crossover", ["int64"], [1_000_000],
        [1000, 10_000, 50_000, 100_000, 200_000, 500_000, 800_000, "all"])
    if args.full:
        results["crossover"] += grid(
            "crossover", ["int64"], [10_000_000],
            [100_000, 1_000_000, 5_000_000])

    log("section 3: strings")
    sizes = [1_000_000, 10_000_000] if args.full else [1_000_000]
    results["strings"] = grid("strings", ["U12", "StringDType"], sizes,
                              [10, 100, 10_000, "all"])

    log("section 4: sklearn-style end-to-end")
    results["sklearn"] = section_sklearn(args.full)

    log("section 5: return_inverse (absl kernel only)")
    results["inverse"] = section_inverse(args.full)

    with open(out, "w") as f:
        json.dump(results, f, indent=1)
    log(f"wrote {out}")
```

</details>

Summariser code: <details>

```py
"""Merge bench_np_unique.py result files into one paste-ready markdown report.

Usage:
    python summarize_np_unique.py bench_results_std.json bench_results_absl.json

Accepts one or more JSON files produced by bench_np_unique.py (one per numpy
build). With both the std (main / PR-31970 branch) and absl (PR-30867 branch)
runs it prints a combined comparison; with a single file it prints that run.
Output is GitHub-flavored markdown on stdout.
"""

import json
import sys


def load(paths):
    runs = {}
    for path in paths:
        with open(path) as f:
            data = json.load(f)
        runs[data["meta"]["kernel"]] = data
    return runs


def fmt(ms):
    if ms is None:
        return "-"
    return f"{ms:,.2f}" if ms < 100 else f"{ms:,.0f}"


def ratio(a, b):
    return "-" if a is None or b is None else f"{a / b:.2f}"


def key(row):
    return (row["dtype"], row["n"], str(row["n_unique"]))


def merged_rows(runs, section):
    """Yield (dtype, n, n_unique, sort_ms, {kernel: hash_ms}) preserving order."""
    order, sort_ms, hashes = [], {}, {}
    for kernel, data in runs.items():
        for row in data.get(section, []):
            k = key(row)
            if k not in hashes:
                order.append(k)
                hashes[k] = {}
            # sort code is identical in every build; keep the fastest run
            sort_ms[k] = min(row["sort_ms"], sort_ms.get(k, float("inf")))
            hashes[k][kernel] = row["hash_ms"]
    for k in order:
        yield k, sort_ms[k], hashes[k]


def kernel_table(runs, section):
    kernels = [k for k in ("std", "absl") if k in runs]
    heads = ["dtype", "n", "n_unique", "sort (ms)"]
    heads += [f"{k} hash (ms)" for k in kernels]
    heads += [f"{k} hash / sort" for k in kernels]
    heads += ["winner"]
    print("| " + " | ".join(heads) + " |")
    print("|" + "---|" * len(heads))
    for (dtype, n, nu), ts, hs in merged_rows(runs, section):
        cells = [dtype, f"{n:,}", nu, fmt(ts)]
        cells += [fmt(hs.get(k)) for k in kernels]
        cells += [ratio(hs.get(k), ts) for k in kernels]
        best = min([("sort", ts)] + [(f"{k} hash", hs[k]) for k in kernels],
                   key=lambda kv: kv[1])
        cells.append(f"**{best[0]}**")
        print("| " + " | ".join(cells) + " |")
    print()


def sort_control_check(runs):
    """The sort kernel is identical code in every build -> use it as a
    control for run-to-run (thermal, load) variation."""
    if len(runs) < 2:
        return
    per_key = {}
    for data in runs.values():
        for section in ("regimes", "crossover", "strings"):
            for row in data.get(section, []):
                per_key.setdefault((section,) + key(row), []).append(row["sort_ms"])
    devs = [max(v) / min(v) - 1 for v in per_key.values() if len(v) > 1]
    if devs:
        print(f"Cross-run control: the sort kernel is identical code in every "
              f"build; its timings agree across runs within "
              f"{100 * max(devs):.0f}% worst-case "
              f"(median {100 * sorted(devs)[len(devs) // 2]:.0f}%), so the "
              f"hash-kernel comparison below is not confounded by machine "
              f"state.\n")


def takeaways(runs):
    print("### Takeaways\n")
    for kernel in ("std", "absl"):
        if kernel not in runs:
            continue
        desc = runs[kernel]["meta"]["kernel_desc"]
        few, allu, cross = [], [], None
        for (dtype, n, nu), ts, hs in merged_rows({kernel: runs[kernel]},
                                                  "regimes"):
            if hs.get(kernel) is None:
                continue
            r = hs[kernel] / ts
            (allu if nu == "all" else few).append(r)
        for (dtype, n, nu), ts, hs in merged_rows({kernel: runs[kernel]},
                                                  "crossover"):
            if n == 1_000_000 and hs.get(kernel, 0) / ts > 1 and cross is None:
                cross = nu
        few_wins = [r for r in few if r < 1]
        print(f"- **{desc}**: in the low-cardinality regime (the "
              f"scikit-learn `classes` case) hash is up to "
              f"{1 / min(few):.1f}x faster than sort"
              f" (wins {len(few_wins)}/{len(few)} of the int64 grid cases);"
              f" in the all-unique worst case it is {max(allu):.1f}x slower."
              f" At n=1M the crossover is around {cross} unique values.")
    if "std" in runs and "absl" in runs:
        speedups = []
        for (dtype, n, nu), ts, hs in merged_rows(runs, "regimes"):
            if "std" in hs and "absl" in hs:
                speedups.append(hs["std"] / hs["absl"])
        print(f"- **absl vs std kernel** (int64 grid): absl is "
              f"{min(speedups):.1f}x-{max(speedups):.1f}x faster than the "
              f"current std::unordered_set across the grid.")
    print()


def main(paths):
    runs = load(paths)
    metas = list(runs.values())

    print("## np.unique: hash-based vs sort-based kernels\n")
    m0 = metas[0]["meta"]
    print(f"- CPU: {m0['cpu']} ({m0['machine']})")
    for data in runs.values():
        m = data["meta"]
        print(f"- run `{m['kernel']}`: numpy `{m['numpy_version']}`, "
              f"hash kernel = {m['kernel_desc']}")
    print("- `hash` = the `_unique_hash` kernel (unsorted output, "
          "`np.unique_values` semantics); `sort` = the flatten+sort+mask "
          "branch of `_unique1d`. Both called directly, bypassing dispatch.")
    print("- best of several repeats, shuffled data, fixed seed\n")
    sort_control_check(runs)
    takeaways(runs)

    print("### 1. int64: the decisive variable is cardinality, not size\n")
    kernel_table(runs, "regimes")

    print("### 2. Crossover: unique-ratio where sort overtakes hash\n")
    kernel_table(runs, "crossover")

    print("### 3. Strings (no SIMD sort exists for these)\n")
    kernel_table(runs, "strings")

    print("### 4. End-to-end scikit-learn-style calls: `np.unique_values(y)`\n")
    print("This is what `sklearn.utils.multiclass.type_of_target` and "
          "`sklearn.utils._encode` do on every classifier fit/validation.\n")
    kernels = [k for k in ("std", "absl") if k in runs]
    heads = (["scenario", "n", "sort kernel (ms)"]
             + [f"{k} hash kernel (ms)" for k in kernels])
    print("| " + " | ".join(heads) + " |")
    print("|" + "---|" * len(heads))
    scen = {}
    for k in kernels:
        for row in runs[k].get("sklearn", []):
            scen.setdefault((row["scenario"], row["n"]), {})[k] = row
    for (name, n), by_kernel in scen.items():
        any_row = next(iter(by_kernel.values()))
        cells = [name, f"{n:,}", fmt(any_row["sort_ms"])]
        cells += [fmt(by_kernel[k]["hash_ms"]) if k in by_kernel else "-"
                  for k in kernels]
        print("| " + " | ".join(cells) + " |")
    print()

    if "std" in runs and runs["std"]["meta"]["has_pr31970_threshold"]:
        rows = [r for r in runs["std"]["sklearn"] if "pr31970_ms" in r]
        if rows:
            print("PR-31970 dispatch effect on those calls "
                  "(threshold=100 vs pre-PR always-hash):\n")
            print("| scenario | n | PR 31970 (ms) | pre-PR (ms) | "
                  "PR slowdown |")
            print("|---|---|---|---|---|")
            for r in rows:
                print(f"| {r['scenario']} | {r['n']:,} "
                      f"| {fmt(r['pr31970_ms'])} | {fmt(r['pre_pr31970_ms'])} "
                      f"| {r['pr31970_ms'] / r['pre_pr31970_ms']:.2f}x |")
            print()

    if "absl" in runs and runs["absl"].get("inverse"):
        print("### 5. `return_inverse` (hash path added by PR 30496/30867): "
              "`LabelEncoder`'s exact call\n")
        print("| scenario | n | hash+inverse (ms) | argsort+inverse (ms) | "
              "hash / argsort |")
        print("|---|---|---|---|---|")
        for r in runs["absl"]["inverse"]:
            print(f"| {r['scenario']} | {r['n']:,} "
                  f"| {fmt(r['hash_inverse_ms'])} "
                  f"| {fmt(r['argsort_inverse_ms'])} "
                  f"| {r['hash_inverse_ms'] / r['argsort_inverse_ms']:.2f} |")
        print()


if __name__ == "__main__":
    if len(sys.argv) < 2:
        sys.exit(__doc__)
    main(sys.argv[1:])
```

</details>

Benchmark results:

<details>

## np.unique: hash-based vs sort-based kernels

- CPU: AMD Ryzen 9 PRO 7940HS w/ Radeon 780M Graphics (x86_64)
- run `std`: numpy `2.6.0.dev0+git20260719.8156dc3`, hash kernel = std::unordered_set (main / PR 31970)
- run `absl`: numpy `2.5.0.dev0+git20260224.1af7b45`, hash kernel = absl::flat_hash_set (PR 30867)
- `hash` = the `_unique_hash` kernel (unsorted output, `np.unique_values` semantics); `sort` = the flatten+sort+mask branch of `_unique1d`. Both called directly, bypassing dispatch.
- best of several repeats, shuffled data, fixed seed

Cross-run control: the sort kernel is identical code in every build; its timings agree across runs within 15% worst-case (median 3%), so the hash-kernel comparison below is not confounded by machine state.

### Takeaways

- **std::unordered_set (main / PR 31970)**: in the low-cardinality regime (the scikit-learn `classes` case) hash is up to 2.5x faster than sort (wins 9/12 of the int64 grid cases); in the all-unique worst case it is 11.8x slower. At n=1M the crossover is around 100000 unique values.
- **absl::flat_hash_set (PR 30867)**: in the low-cardinality regime (the scikit-learn `classes` case) hash is up to 5.4x faster than sort (wins 9/12 of the int64 grid cases); in the all-unique worst case it is 4.3x slower. At n=1M the crossover is around 800000 unique values.
- **absl vs std kernel** (int64 grid): absl is 1.2x-3.2x faster than the current std::unordered_set across the grid.

### 1. int64: the decisive variable is cardinality, not size

| dtype | n | n_unique | sort (ms) | std hash (ms) | absl hash (ms) | std hash / sort | absl hash / sort | winner |
|---|---|---|---|---|---|---|---|---|
| int64 | 100,000 | 2 | 0.44 | 1.37 | 1.18 | 3.12 | 2.69 | **sort** |
| int64 | 100,000 | 10 | 1.01 | 0.97 | 0.76 | 0.96 | 0.75 | **absl hash** |
| int64 | 100,000 | 100 | 2.00 | 1.02 | 0.72 | 0.51 | 0.36 | **absl hash** |
| int64 | 100,000 | 10000 | 2.85 | 2.13 | 1.15 | 0.75 | 0.41 | **absl hash** |
| int64 | 100,000 | all | 2.85 | 8.67 | 4.46 | 3.04 | 1.57 | **sort** |
| int64 | 1,000,000 | 2 | 3.35 | 9.88 | 7.38 | 2.95 | 2.21 | **sort** |
| int64 | 1,000,000 | 10 | 9.98 | 9.51 | 7.26 | 0.95 | 0.73 | **absl hash** |
| int64 | 1,000,000 | 100 | 19.38 | 9.87 | 7.26 | 0.51 | 0.37 | **absl hash** |
| int64 | 1,000,000 | 10000 | 38.10 | 16.89 | 8.09 | 0.44 | 0.21 | **absl hash** |
| int64 | 1,000,000 | all | 38.04 | 310 | 96.81 | 8.16 | 2.54 | **sort** |
| int64 | 10,000,000 | 2 | 68.14 | 96.32 | 73.88 | 1.41 | 1.08 | **sort** |
| int64 | 10,000,000 | 10 | 139 | 95.99 | 71.29 | 0.69 | 0.51 | **absl hash** |
| int64 | 10,000,000 | 100 | 229 | 97.26 | 74.67 | 0.42 | 0.33 | **absl hash** |
| int64 | 10,000,000 | 10000 | 418 | 167 | 80.04 | 0.40 | 0.19 | **absl hash** |
| int64 | 10,000,000 | all | 482 | 5,707 | 2,388 | 11.84 | 4.95 | **sort** |

### 2. Crossover: unique-ratio where sort overtakes hash

| dtype | n | n_unique | sort (ms) | std hash (ms) | absl hash (ms) | std hash / sort | absl hash / sort | winner |
|---|---|---|---|---|---|---|---|---|
| int64 | 1,000,000 | 1000 | 28.27 | 14.44 | 7.39 | 0.51 | 0.26 | **absl hash** |
| int64 | 1,000,000 | 10000 | 36.99 | 16.15 | 8.01 | 0.44 | 0.22 | **absl hash** |
| int64 | 1,000,000 | 50000 | 37.02 | 34.77 | 12.03 | 0.94 | 0.32 | **absl hash** |
| int64 | 1,000,000 | 100000 | 37.43 | 39.11 | 15.76 | 1.04 | 0.42 | **absl hash** |
| int64 | 1,000,000 | 200000 | 38.30 | 51.49 | 19.96 | 1.34 | 0.52 | **absl hash** |
| int64 | 1,000,000 | 500000 | 39.59 | 131 | 32.58 | 3.31 | 0.82 | **absl hash** |
| int64 | 1,000,000 | 800000 | 40.14 | 165 | 50.48 | 4.12 | 1.26 | **sort** |
| int64 | 1,000,000 | all | 37.30 | 258 | 85.91 | 6.91 | 2.30 | **sort** |
| int64 | 10,000,000 | 100000 | 507 | 407 | 147 | 0.80 | 0.29 | **absl hash** |
| int64 | 10,000,000 | 1000000 | 505 | 2,359 | 843 | 4.67 | 1.67 | **sort** |
| int64 | 10,000,000 | 5000000 | 492 | 4,043 | 1,393 | 8.21 | 2.83 | **sort** |

### 3. Strings (no SIMD sort exists for these)

| dtype | n | n_unique | sort (ms) | std hash (ms) | absl hash (ms) | std hash / sort | absl hash / sort | winner |
|---|---|---|---|---|---|---|---|---|
| U12 | 1,000,000 | 10 | 89.11 | 63.55 | 14.08 | 0.71 | 0.16 | **absl hash** |
| U12 | 1,000,000 | 100 | 98.96 | 63.93 | 14.11 | 0.65 | 0.14 | **absl hash** |
| U12 | 1,000,000 | 10000 | 107 | 83.77 | 19.97 | 0.78 | 0.19 | **absl hash** |
| U12 | 1,000,000 | all | 119 | 333 | 138 | 2.79 | 1.15 | **sort** |
| U12 | 10,000,000 | 10 | 1,174 | 642 | 143 | 0.55 | 0.12 | **absl hash** |
| U12 | 10,000,000 | 100 | 1,240 | 657 | 142 | 0.53 | 0.11 | **absl hash** |
| U12 | 10,000,000 | 10000 | 1,407 | 850 | 200 | 0.60 | 0.14 | **absl hash** |
| U12 | 10,000,000 | all | 1,658 | 7,270 | 2,374 | 4.38 | 1.43 | **sort** |
| StringDType | 1,000,000 | 10 | 164 | 16.49 | 14.92 | 0.10 | 0.09 | **absl hash** |
| StringDType | 1,000,000 | 100 | 178 | 17.62 | 14.82 | 0.10 | 0.08 | **absl hash** |
| StringDType | 1,000,000 | 10000 | 209 | 32.40 | 18.68 | 0.16 | 0.09 | **absl hash** |
| StringDType | 1,000,000 | all | 250 | 286 | 177 | 1.14 | 0.71 | **absl hash** |
| StringDType | 10,000,000 | 10 | 2,016 | 217 | 206 | 0.11 | 0.10 | **absl hash** |
| StringDType | 10,000,000 | 100 | 2,138 | 237 | 194 | 0.11 | 0.09 | **absl hash** |
| StringDType | 10,000,000 | 10000 | 2,456 | 388 | 248 | 0.16 | 0.10 | **absl hash** |
| StringDType | 10,000,000 | all | 3,020 | 6,233 | 3,435 | 2.06 | 1.14 | **sort** |

### 4. End-to-end scikit-learn-style calls: `np.unique_values(y)`

This is what `sklearn.utils.multiclass.type_of_target` and `sklearn.utils._encode` do on every classifier fit/validation.

| scenario | n | sort kernel (ms) | std hash kernel (ms) | absl hash kernel (ms) |
|---|---|---|---|---|
| int64, 5 classes | 10,000,000 | 103 | 97.92 | 71.15 |
| int64, 100 classes | 10,000,000 | 226 | 104 | 73.75 |
| <U9, 5 classes | 10,000,000 | 942 | 434 | 119 |
| <U9, 100 classes | 10,000,000 | 1,119 | 427 | 119 |

PR-31970 dispatch effect on those calls (threshold=100 vs pre-PR always-hash):

| scenario | n | PR 31970 (ms) | pre-PR (ms) | PR slowdown |
|---|---|---|---|---|
| int64, 5 classes | 10,000,000 | 103 | 134 | 0.77x |
| int64, 100 classes | 10,000,000 | 225 | 132 | 1.71x |
| <U9, 5 classes | 10,000,000 | 928 | 540 | 1.72x |
| <U9, 100 classes | 10,000,000 | 1,102 | 569 | 1.94x |

### 5. `return_inverse` (hash path added by PR 30496/30867): `LabelEncoder`'s exact call

| scenario | n | hash+inverse (ms) | argsort+inverse (ms) | hash / argsort |
|---|---|---|---|---|
| int64, 5 classes | 10,000,000 | 152 | 432 | 0.35 |
| int64, 100 classes | 10,000,000 | 156 | 1,050 | 0.15 |
| int64, all unique | 10,000,000 | 2,639 | 1,555 | 1.70 |

</details>

ngoldbaum:
@adrinjalali thanks for taking a look. Just a head’s up that copy/pasting large amounts of LLM output should be avoided. I’m guessing an LLM was involved anyway given the lengthy comments in your benchmarks.

One thing we could do is add a check to bail out from the hash-based unique if the number of unique values exceeds a threshold where we expect performance to be poor. That means no one pays the full cost of hashing and the worst-case slowdows would be a lot better.

Another option is to use a better hash function that more closely matches what absl does. #30496 has a bunch of discussion about this and several possible ways forward. Vendoring an implementation of a smarter hash function than fnv-1a is much more appealing than depending on absl in the NumPy build.

In either case, the approach in this PR is not the way to go.

If there isn’t an issue already about these worst-case slowdowns I’d appreciate it if one of you could open one and then a new PR can fix it.

I’m going to go ahead and close this PR, please open a new PR after opening a new issue and outlining an approach you’d like to take. If the future please don’t open a PR before doing that, particularly with LLM-gnerated code. LLMs tend to have poor codebase-wide insight and knowledge of previous discussions.

adrinjalali:
> Just a head’s up that copy/pasting large amounts of LLM output should be avoided. I’m guessing an LLM was involved anyway given the lengthy comments in your benchmarks.

Sure thing. I genuinely wonder what the right approach here would be though (I added an LLM disclaimer to my post anyway).

I posted the benchmark codes in case somebody wants to reproduce them, and posted my own take before the ones I posted within `<detail>` tags. It was a couple of hours of work, and I didn't want it to go to waste.

Anyhow, I'm glad it helped you settle on this particular PR.

PR #31975, opened 2026-07-12
BUG: fix swallowed cast error in fancy indexing assignment

### LINKED ISSUE #31974: BUG: cast error during buffered fancy-index assignment is swallowed (SystemError or silently leaked exception)
### Describe the issue:

If `arr[idx] = values` needs a buffered cast and the cast fails in a buffer chunk after the first one (array > 8192 elements), the error is set but the assignment reports success. Result: the statement silently "succeeds", `dst` is partially written, and the leaked exception surfaces at a later unrelated call (or as `SystemError: ... returned a result with an exception set` when it crosses a function-call boundary, e.g. inside `np.put_along_axis`).

Small arrays raise a clean `TypeError` for the same assignment, so behavior is size-dependent. Not object-specific: a failing string cast (`svals[9000] = "xx"` into an int array) leaks the same way.

Root cause: a cast failure during buffer refill makes `npyiter_buffered_iternext` return 0, indistinguishable from normal end of iteration (`nditer_templ.c.src` ~line 248). `mapiter_set` loops `do {...} while (mit->outer_next(...))` and only checks `PyErr_Occurred()` at the top of the loop body, so the error set by the final iternext is never seen; the function returns 0 (`lowlevel_strided_loops.c.src:2010-2157`), and `array_assign_subscript` (`mapping.c:2175`) reports success.

```python
import numpy as np

N = 20000  # > 8192 (one buffer chunk)
dst = np.zeros(N, dtype=int)
vals = np.arange(N, dtype=object)
vals[9000] = None  # in the second chunk

dst[np.arange(N)] = vals   # does NOT raise
print("got here")          # leaked TypeError detonates here or later
```

### Error message:

```shell
got here
  File "repro.py", line 9, in <module>
    print("got here")
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'

Same assignment via np.put_along_axis raises immediately instead, but as:
SystemError: <function put_along_axis at 0x...> returned a result with an exception set

dst has the first 8192 elements written, rest untouched. The partial write is
expected (see gh-31965 / gh-31940); this issue is only the swallowed exception.
```

NumPy 2.2.4, 2.5.1, and main @ 9ef40d32ec. Python 3.13.5, Linux x86_64 (Debian 13).

### Runtime Environment:

_No response_

Found while auditing partial-write behavior for gh-31965.

Ijtihed:
### PR summary

Fixes #31974.

`arr[idx] = vals` can swallow a cast error when the values are buffered and the cast fails while the iterator refills a buffer chunk (array bigger than one 8192-element chunk). the buffered iternext returns 0 on error which looks exactly like a normal end of iteration, so `mapiter_set`/`mapiter_get` returned success with the exception still set. depending on what runs next that's either a silent bogus success (the error surfaces at a later unrelated call) or `SystemError: ... returned a result with an exception set` (e.g. via `np.put_along_axis`).

this adds a `PyErr_Occurred()` check after the iteration loops in `mapiter_@name@`. small arrays already raise cleanly since first-chunk failures are caught at `NpyIter_Reset` so this just makes big arrays consistent. kept the check unconditional (not `needs_api`-gated) because non-object casts can fail too, e.g. a string -> int cast hits the same leak

#### AI Disclosure

Used Claude to double check my work and mae sure I didn't miss anything

### DIFF
--- a/numpy/_core/src/multiarray/lowlevel_strided_loops.c.src
+++ b/numpy/_core/src/multiarray/lowlevel_strided_loops.c.src
@@ -2154,6 +2154,11 @@ mapiter_@name@(
         }
 /**end repeat1**/
     }
+
+    /* Check if the above iteration ended with an error */
+    if (PyErr_Occurred()) {
+        return -1;
+    }
     return 0;
 }
 
--- a/numpy/_core/tests/test_indexing.py
+++ b/numpy/_core/tests/test_indexing.py
@@ -777,6 +777,17 @@ class TestFancyIndexingCast:
                      zero_array.__setitem__, bool_index, np.array([1j]))
         assert_equal(zero_array[0, 1], 0)
 
+    def test_fancy_assign_buffered_cast_error(self):
+        # gh-31974: a cast failure while refilling the iterator buffer
+        # (only possible when the array exceeds one buffer chunk) ended
+        # iteration silently, leaving the error set but unraised.
+        N = 20000
+        dst = np.zeros(N, dtype=int)
+        vals = np.arange(N, dtype=object)
+        vals[9000] = None
+        with pytest.raises(TypeError):
+            dst[np.arange(N)] = vals
+
 
 class TestFancyIndexingEquivalence:
     def test_object_assign(self):


numpy/_core/src/multiarray/lowlevel_strided_loops.c.src
```diff
@@ -2154,6 +2154,16 @@ mapiter_@name@(
         }
 /**end repeat1**/
     }
+
+    /*
+     * A buffered cast may fail inside `outer_next`/`subspace_next`
+     * (the buffered iternext), which then returns 0 exactly as on a
+     * normal end of iteration, so the loops above can finish "cleanly"
+     * with the error still set.  Check for it here.
+     */
```
ngoldbaum: This sort of comment should be maximum one-line. AI models like to memorialize bugs and if we add comments like this for a few years then the codebase will be riddled with references to old bugs for no benefit.

Ijtihed: > if we add comments like this for a few years then the codebase will be riddled with references to old bugs for no benefit.

ohhhh that makes sense. I will limit it to one line. thanks!

numpy/_core/src/multiarray/lowlevel_strided_loops.c.src
```diff
@@ -2154,6 +2154,11 @@ mapiter_@name@(
         }
 /**end repeat1**/
     }
+
+    /* A buffered cast failure returns 0 like a normal iteration end. */
```
kumaraditya303:
```suggestion
    /* Check if the above iteration ended with an error */
```

Ijtihed: thank you! yeah I think this is better :)

kumaraditya303 (approved):

ngoldbaum:
Thanks @Ijtihed

PR #31976, opened 2026-07-12
ENH: Use `PyModule_AddObjectRef` in `_umath_linalg`

prathamhole14:
### PR summary

In the [PyModule_GetDict](https://docs.python.org/3/c-api/module.html#c.PyModule_GetDict)'s docs it's mentioned:
`It is recommended extensions use other PyModule_* and PyObject_* functions rather than directly manipulate a module’s [__dict__].`

I also noticed `PyDict_SetItemString` can fail (returns -1)  [ref ](https://docs.python.org/3/c-api/dict.html#c.PyDict_SetItemString), but in the current code the return value was ignored. (Edited)

#### AI Disclosure

Final review with claude, it suggested to change `dictionary` word to `module` in `addUfuncs`

### DIFF
--- a/numpy/linalg/umath_linalg.cpp
+++ b/numpy/linalg/umath_linalg.cpp
@@ -4681,7 +4681,7 @@ GUFUNC_DESCRIPTOR_t gufunc_descriptors [] = {
 };
 
 static int
-addUfuncs(PyObject *dictionary) {
+addUfuncs(PyObject *module) {
     PyUFuncObject *f;
     int i;
     const int gufunc_count = sizeof(gufunc_descriptors)/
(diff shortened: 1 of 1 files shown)


mattip:
`PyDict_SetItemString` is part of the stable ABI, what motivated changing it for `PyModule_AddObjectRef`?

prathamhole14:
In the [PyModule_GetDict](https://docs.python.org/3/c-api/module.html#c.PyModule_GetDict)'s docs it's mentioned:
`It is recommended extensions use other PyModule_* and PyObject_* functions rather than directly manipulate a module’s [__dict__].`

I also noticed `PyDict_SetItemString` can fail (returns -1)  [ref ](https://docs.python.org/3/c-api/dict.html#c.PyDict_SetItemString), but in the current code the return value was ignored.

Yes, this is not a "mandatory" but since I was auditing `_umath_liang` I thought we should do this.

mattip:
Thanks @prathamhole14

PR #31978, opened 2026-07-12
BUG: stop object argmax/argmin on comparison errors

### LINKED ISSUE #31977: BUG: object dtype argmax/argmin swallows comparison errors and keeps writing to out
### Describe the issue:

`np.argmax`/`np.argmin` on an object array whose comparison raises does not stop at the error. The inner loop returns success with the exception set, the outer loop keeps calling comparisons for the remaining rows (each call now running with a pending exception, producing chained `SystemError`s), and keeps writing indices into the result. With `out=` the caller's array gets garbage indices written after the error; expected is a plain `ValueError: boom`.

Analogous ops all propagate the error cleanly (`np.sort`, `np.argsort`, builtin `max`, `a.max()`), argmax/argmin is the outlier.

Root cause (main @ 9ef40d32ec): `OBJECT_argmax` returns 0 (success) when `PyObject_RichCompareBool` fails (`numpy/_core/src/multiarray/arraytypes.c.src:3323`, argmin at `:3387`), and `_PyArray_ArgMinMaxCommon` ignores the argfunc return value entirely (`numpy/_core/src/multiarray/calculation.c:176-179`). Same failure pattern as gh-29688 (`np.dot` object loop, fixed in #31565), but here the ArgFunc returns int so proper propagation is possible: make the object argfuncs return -1 and check the return in the caller

```python
import numpy as np

class BadCmp:
    def __init__(self, v):
        self.v = v
    def __gt__(self, other):
        if self.v == 3 or other.v == 3:
            raise ValueError("boom")
        return self.v > other.v

a = np.array([[BadCmp(1), BadCmp(2)],
              [BadCmp(3), BadCmp(4)],
              [BadCmp(1), BadCmp(2)]], dtype=object)
out = np.full(3, 77, dtype=np.intp)
np.argmax(a, axis=1, out=out)
# np.argmin behaves identically via __lt__
```

### Error message:

```shell
ValueError: boom

SystemError: <function BadCmp.__gt__ at 0x...> returned a result with an exception set

SystemError: <built-in method argmax of numpy.ndarray object at 0x...> returned a result with an exception set

out afterwards is [1, 0, 0]: row 0 legitimate, rows 1 and 2 written after/despite the error.
```

NumPy 2.2.4, 2.5.1, and main @ 9ef40d32ec. Python 3.13.5, Linux x86_64 (Debian 13).

### Runtime Environment:

_No response_

Found while auditing behavior for partial writes for gh-31965

Ijtihed:
### PR summary

Fixes #31977.

`np.argmax`/`np.argmin` on object arrays kept going after a comparison raised so `OBJECT_argmax`/`OBJECT_argmin` returned 0 (success) when `PyObject_RichCompareBool` failed and `_PyArray_ArgMinMaxCommon` ignored the argfunc return value entirely so the outer loop kept calling comparisons with the exception already pending (chained `SystemError`s) and kept writing indices into `out` after the error.

This makes the object argfuncs return -1 on comparison failure and checks the return value in `_PyArray_ArgMinMaxCommon` which discards the pending writeback to `out` on error. same pattern as gh-29688 / #31565 (`np.dot` object loop) but here the ArgFunc returns int so the error can propagate properly instead of the early-exit workaround. rows after the failing one are no longer written and `np.sort`/`np.argsort` already behave this way

#### AI Disclosure

Used Claude to  double check my work and make sure i didn't miss anything

### DIFF
--- a/numpy/_core/src/multiarray/arraytypes.c.src
+++ b/numpy/_core/src/multiarray/arraytypes.c.src
@@ -3321,7 +3321,7 @@ OBJECT_argmax(PyObject **ip, npy_intp n, npy_intp *max_ind,
                 int greater_than = PyObject_RichCompareBool(val, mp, Py_GT);
 
                 if (greater_than < 0) {
-                    return 0;
+                    return -1;
                 }
                 if (greater_than) {
                     mp = val;
@@ -3385,7 +3385,7 @@ OBJECT_argmin(PyObject **ip, npy_intp n, npy_intp *min_ind,
                 int less_than = PyObject_RichCompareBool(val, mp, Py_LT);
 
                 if (less_than < 0) {
-                    return 0;
+                    return -1;
                 }
                 if (less_than) {
                     mp = val;
--- a/numpy/_core/src/multiarray/calculation.c
+++ b/numpy/_core/src/multiarray/calculation.c
@@ -174,7 +174,12 @@ _PyArray_ArgMinMaxCommon(PyArrayObject *op,
     n = PyArray_SIZE(ap)/m;
     rptr = (npy_intp *)PyArray_DATA(rp);
     for (ip = PyArray_DATA(ap), i = 0; i < n; i++, ip += elsize*m) {
-        arg_func(ip, m, rptr, ap);
+        if (arg_func(ip, m, rptr, ap) < 0) {
+            NPY_END_THREADS_DESCR(PyArray_DESCR(ap));
+            /* `rp` may be a WRITEBACKIFCOPY of `out`; do not write back */
+            PyArray_DiscardWritebackIfCopy(rp);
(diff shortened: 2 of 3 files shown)


mattip:
We are being flooded with AI-assisted PRs for made up issues that no one has seen in reality. Please refrain from submitting this kind of PR on an issue you yourself raised until the issue has been triaged and deemed worthy of fixing.

Ijtihed:
> until the issue has been triaged and deemed worthy of fixing.

makes sense. will close it for now then

whyAreYouLying:
That is why I made this comment in this PR 25432 (comments) [link](https://github.com/scipy/scipy/pull/25432#issuecomment-4815705460). cc @serge-sans-paille and @seberg

PR #31979, opened 2026-07-12
ENH: raise OverflowError on timedelta64 float multiply/divide overflow

jbrockmendel:
Human here: @seberg as with my other recent PRs, acknowledging this is mostly claude.  This is porting pandas-dev/pandas#65515

Claude below the line

-----

Follow-up to GH-31378, which made `datetime64`/`timedelta64` add, subtract, and integer-multiply raise `OverflowError` on int64 overflow instead of silently wrapping. This extends the same guarantee to the **float** multiply and divide loops, which previously cast the `float64` result straight to `int64` and silently saturated (or landed on the `NaT` sentinel) when the value exceeded int64 range.

### What changes

`TIMEDELTA_md_m_multiply` (`td * float`), `TIMEDELTA_dm_m_multiply` (`float * td`), and `TIMEDELTA_md_m_divide` (`td / float`, which also backs `td // float`) now raise `OverflowError` when the result would overflow int64:

```python
>>> np.timedelta64(2**62, "s") * 2.5      # was: silently wrapped
OverflowError: Overflow in timedelta64 * float64 multiplication
>>> np.timedelta64(2**62, "s") / 0.4      # was: silently wrapped
OverflowError: Overflow in timedelta64 / float64 division
```

A shared `timedelta_from_double` helper maps a `NaN` result to `NaT` and treats any magnitude `>= 2**63` (finite overflow **or** `±inf`) as overflow. The bound is `2**63` rather than `int64.max`, because `2**63 - 1` is not exactly representable in `float64` and rounds up to `2**63`; a result landing there would otherwise slip past the check and saturate on the cast (and `-2**63` is itself the `NaT` sentinel).

### Preserved (non-overflow) behavior

A `NaN` operand, `0 * inf`, division by zero, and division by infinity still yield `NaT` (or zero), matching the pre-existing convention:

```python
>>> np.timedelta64(5, "s") / 0.0          # NaT, not an error
>>> np.timedelta64(5, "s") / np.float64(np.inf)   # 0 seconds
>>> np.timedelta64(5, "s") * np.float64(np.nan)   # NaT
```

The one intentional behavior change to an existing test: `timedelta64(nonzero) * inf` now raises instead of returning `NaT`.


jorenham:
sorry but this goes against the AI policy

jbrockmendel:
Fair enough.  You closed while I was in the process of updating the comment with the ping to seberg acknowledging the AI use.

jorenham:
Ah ok, that changes things

jorenham:
I'm still not sure how I feel about Claude as co-author though.

ngoldbaum:
> I'm still not sure how I feel about Claude as co-author though.

IMO it's mostly advertising for anthropic but it shouldn't stop us from accepting the code. AI-generated code will get in either way.

ngoldbaum:
Also sorry for the drive-by comment, I think improvements like this are great and I'll try to get it reviewed.

numpy/_core/src/umath/loops.c.src
```diff
@@ -995,6 +995,37 @@ TIMEDELTA_qm_m_multiply(char **args, npy_intp const *dimensions, npy_intp const
...
+ */
+static inline int
+timedelta_from_double(double result, npy_timedelta *out)
+{
+    if (npy_isnan(result)) {
+        *out = NPY_DATETIME_NAT;
+        return 0;
+    }
+    if (npy_fabs(result) >= 9223372036854775808.0 /* 2**63 */) {
+        return 1;
```
seberg: NIT: I'd prolly use `-1` just because it's the more typical pattern, I think.  Maybe we should also in-line the error, just to avoid duplicating the message?

numpy/_core/src/umath/loops.c.src
```diff
@@ -1097,13 +1127,19 @@ TIMEDELTA_md_m_divide(char **args, npy_intp const *dimensions, npy_intp const *s
...
-                *((npy_timedelta *)op1) = (npy_timedelta)result;
-            }
-            else {
+            /* Compute first so the FPU still flags divide-by-zero as before. */
+            const double result = (double)in1 / in2;
+            if (in2 == 0.0) {
+                /* Division by exactly zero yields NaT, matching numpy's
+                 * convention.  Unlike a tiny nonzero divisor (whose infinite
+                 * quotient is an overflow), this is not treated as overflow. */
                 *((npy_timedelta *)op1) = NPY_DATETIME_NAT;
```
seberg, line 1129: Not sure I understand the comment and path.  If `in2` is zero, the result will be NaN and `timedelta_from_double` will do just fine.  Just delete it?

The old code had the extra branch to map infinities to NaN, but you want to change that to raise an error aggressively (debatable, but IIRC there were some previous PRs where this was effectively decided more generally for datetime).

numpy/_core/tests/test_datetime.py
```diff
@@ -1265,6 +1265,118 @@ def test_arithmetic_valid_boundary(self):
...
+        # would otherwise overflow (the NaT short-circuit runs first).
+        assert np.isnat(nat * 2.5)
+        assert np.isnat(2.5 * nat)
+        assert np.isnat(nat / 0.4)
+        assert np.isnat(nat * np.float64(np.inf))
+
+        # NaN operand yields NaT, not overflow.
+        assert np.isnat(td * np.float64(np.nan))
+        assert np.isnat(np.float64(np.nan) * td)
+        assert np.isnat(td / np.float64(np.nan))
```
seberg: With these AI tests I always feel they should be parametrized more (which I would appreciate if easy), but OK...

seberg (commented):
Thanks, a few nits that would be nice to address.  This tends to be a little bit more verbose then I like, but I tall looks good.

numpy/_core/src/umath/loops.c.src
jbrockmendel: i prefer NaT since that matches the behavior we recently merged into pandas.  How strongly do you feel about raising?

seberg: I dunno, I have a slight nudge towards NaT being fine.  But I thought we had some other place recently where the overflow was given an error (maybe converting units)?
There is an argument for both as NaT isn't as neat as inf, but...

(In that sense, if you prefer NaT, I guess that is also the "no change" and we should use it?)

jbrockmendel: we recently made overflows in `low_unit.astype(high_unit)` raise instead of silently wrap.  But I think the closer analogy is division by int zero and float zero.  division by int zero gives NaT, so I think division by float zero should do the same

seberg: I am not sure, but if we aren't changing behavior _and_ you prefer it, it seems clear enough to me that it is fine :).

jbrockmendel:
Updated, tried to trim verbosity.

numpy/_core/src/umath/loops.c.src
```diff
@@ -1173,13 +1193,16 @@ TIMEDELTA_md_m_divide(char **args, npy_intp const *dimensions, npy_intp const *s
...
-            else {
+            /* Divide first so the FPU still flags divide-by-zero.  A zero
+             * divisor gives NaT like the integer loops; a tiny nonzero one
+             * overflows. */
+            const double result = (double)in1 / in2;
+            if (in2 == 0.0) {
                 *((npy_timedelta *)op1) = NPY_DATETIME_NAT;
             }
+            else if (timedelta_from_double(result, (npy_timedelta *)op1) < 0) {
+                return;
```
ngoldbaum, line 1204: The caller of this function and the other one you changed currently doesn't check if a Python exception is set, so these error cases are never propagated to Python. For example:

```python
a = np.array([2**62], dtype="timedelta64[ns]")
a * np.full((2, 1), 1e300)
# SystemError
```

Currently all of these loops don't have any Python error checking because they don't have the `NPY_METH_REQUIRED_PYAPI` flag. I guess with the use of `npy_gil_error` you're trying to work around that but to do so you'd need a new flag for loops that need a Python thread state to set Python errors but otherwise don't need it. Or add the `NPY_METH_REQUIRED_PYAPI` flag, but then the loops will no longer release the GIL.

Another way to do it is to do a bigger refactoring to make these functions no longer be `void` and then you can do C error handling at the call sites and only set an exception once you've exited the loop.

PR #21817, opened 2022-06-21
ENH: Always fill object fields with None rather than NULL

seberg:
The intention here is to transition (or achieve?) a future where
NumPy always fills object fields with None rather than NULL.

Filling with NULL is nice (infinitely fast in some cases and easy),
but it also adds the trap of having to anticipate NULLs everywhere.
Most importantly, cython fails to do so (currently) so this is a
bug magnet.

At this point, ideally any fully created array is guaranteed to be
initialized with proper Python objects.  Downstream may break that,
so we should not allow it, but this does also add an assert to
many ufuncs to flush out further potential issues.

NULL initialization is (as of now) still relevant for new buffers.
That means that core functionality like `setitem`, or coping into
a buffer should still expect a NULL initialized value.

---

Closes gh-21738 (and some downstream issues)

We should maybe discuss this briefly, the alternative would be to put more pressure on downstream to fix their side (e.g. more aggressively keep the NULL filling), although I suspect we do have some bugs of our own.

Issue #7961, opened 2016-08-22
sqrt raises AttributeError on unexpected input


silvioricardoc:
``` python
>>> numpy.sqrt(9)
3.0
>>> numpy.sqrt(None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: sqrt
```

I expected a TypeError or ValueError, and spent quite some time trying to understand why `sqrt` was "missing"...

charris:
`None` is treated as an object, which is why a `sqrt` attribute is looked for. Perhaps the error message can be improved.

timhoffm:
Same problem with sympy integers:

~~~
>>> import numpy as np
>>> x = Integer(10)
>>> np.sqrt(x)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'Integer' object has no attribute 'sqrt'
~~~

Reported in https://github.com/matplotlib/matplotlib/issues/13106.

A better error message would really help to find out what's going on.

eric-wieser:
+1 from me on improving the error message. Raising `TypeError("Object loop of ufunc {} does not support object of type {} which has no {} method")` if the attribute is absent would be helpful.

I'm not sure if we should also try and check is the attribute is callable.

Perhaps chaining the exception would be useful here

PR #31988, opened 2026-07-13
ENH,  BUG: use faster calling conventions (vectorcall) in several hot paths

eendebakpt:
### PR summary
Replace `PyObject_CallMethod`/`PyObject_CallFunction` by the faster vectorcall based calling conventions (`PyObject_Vectorcall`, `PyObject_CallOneArg`, `PyObject_CallMethodNoArgs/OneArg`, `PyObject_CallNoArgs`) and interned strings. This saves temporary tuple creation, variadic argument parsing and repeated creation of temporary strings.

*This also fixes that tuple elements are _not_ unpacked as arguments: `np.floor(np.array((1,), dtype="O", ndmax=0))  # fails correctly now`*

In the same PR we address some bugs due to re-entrancy in object methods.

Benchmark results for a few selected cases (pyperf, Windows, MSVC):

| Benchmark                  | main    | PR                    |
|----------------------------|---------|-----------------------|
| frompyfunc_1in_1000        | 48.2 us | 37.7 us: 1.28x faster |
| frompyfunc_2in_1000        | 53.0 us | 40.7 us: 1.30x faster |
| conjugate_object_1000      | 119 us  | 54.5 us: 2.18x faster |
| floor_object_1000          | 126 us  | 95.2 us: 1.33x faster |
| gcd_object_1000            | 155 us  | 132 us: 1.18x faster  |
| datetime_as_string_tz_1000 | 1.60 ms | 1.47 ms: 1.09x faster |
| matrix_add_2x2             | 1.22 us | 1.16 us: 1.05x faster |
| Geometric mean             | (ref)   | 1.31x faster          |

<details>

```python
"""Benchmarks for the simpler-calling-convention changes.

Run with:
    python bench_call_conventions.py -o result.json
    python -m pyperf compare_to base.json new.json --table
"""
import datetime
import fractions

import pyperf
import numpy as np

N = 1000

obj_int = np.array([i + 10**30 for i in range(N)], dtype=object)
obj_int2 = np.array([(i + 7) * 3 + 10**30 for i in range(N)], dtype=object)
obj_complex = np.array([complex(i, -i) for i in range(N)], dtype=object)
obj_frac = np.array(
    [fractions.Fraction(3 * i + 1, 2) for i in range(N)], dtype=object
)

identity = np.frompyfunc(lambda x: x, 1, 1)
add2 = np.frompyfunc(lambda x, y: x, 2, 1)

dt_arr = np.array(["2020-05-17T12:00"] * N, dtype="M8[m]")
tzinfo = datetime.timezone(datetime.timedelta(hours=5, minutes=30))

mat = np.matrix([[1.0, 2.0], [3.0, 4.0]])


def bench_frompyfunc_1in(loops):
    t0 = pyperf.perf_counter()
    for _ in range(loops):
        identity(obj_int)
    return pyperf.perf_counter() - t0


def bench_frompyfunc_2in(loops):
    t0 = pyperf.perf_counter()
    for _ in range(loops):
        add2(obj_int, obj_int2)
    return pyperf.perf_counter() - t0


def bench_conjugate_object(loops):
    t0 = pyperf.perf_counter()
    for _ in range(loops):
        np.conjugate(obj_complex)
    return pyperf.perf_counter() - t0


def bench_floor_object(loops):
    t0 = pyperf.perf_counter()
    for _ in range(loops):
        np.floor(obj_frac)
    return pyperf.perf_counter() - t0


def bench_gcd_object(loops):
    t0 = pyperf.perf_counter()
    for _ in range(loops):
        np.gcd(obj_int, obj_int2)
    return pyperf.perf_counter() - t0


def bench_datetime_as_string_tz(loops):
    t0 = pyperf.perf_counter()
    for _ in range(loops):
        np.datetime_as_string(dt_arr, timezone=tzinfo)
    return pyperf.perf_counter() - t0


def bench_matrix_add(loops):
    m = mat
    t0 = pyperf.perf_counter()
    for _ in range(loops):
        m + m
    return pyperf.perf_counter() - t0


runner = pyperf.Runner()
runner.bench_time_func("frompyfunc_1in_1000", bench_frompyfunc_1in)
runner.bench_time_func("frompyfunc_2in_1000", bench_frompyfunc_2in)
runner.bench_time_func("conjugate_object_1000", bench_conjugate_object)
runner.bench_time_func("floor_object_1000", bench_floor_object)
runner.bench_time_func("gcd_object_1000", bench_gcd_object)
runner.bench_time_func("datetime_as_string_tz_1000", bench_datetime_as_string_tz)
runner.bench_time_func("matrix_add_2x2", bench_matrix_add)
```

</details>

#### AI Disclosure

Claude was prompted to find all inefficient python calls in a performance critical path (this means there are some cases remaining that can be converted if needed). The benchmark script was written by Claude as well.


KRRT7:
I think this PR introduces a borrowed-reference lifetime bug in the object `gcd` loop.

`npy_ObjectGCD` changed from `PyObject_CallFunction(..., "OO", i1, i2)` to `PyObject_Vectorcall(...)`. The old call built a temporary tuple and therefore kept strong references to both object-array operands while `math.gcd` ran. With vectorcall, `math.gcd` receives borrowed pointers. If the first operand's `__index__` re-entrantly clears the array slot holding the second operand, `math.gcd` can later dereference the freed second operand.

Reproducer on this branch:

```python
import faulthandler
import numpy as np

faulthandler.enable()
victim_array = None

class Killer:
    def __index__(self):
        global victim_array
        print("Killer.__index__: clearing victim array slot", flush=True)
        victim_array[0] = None
        print("Killer.__index__: cleared", flush=True)
        return 12

class Victim:
    def __index__(self):
        print("Victim.__index__", flush=True)
        return 8

    def __del__(self):
        print("Victim.__del__", flush=True)

left = np.empty(1, dtype=object)
left[0] = Killer()
victim_array = np.empty(1, dtype=object)
victim_array[0] = Victim()

print("about to call np.gcd", flush=True)
np.gcd(left, victim_array)
```

I ran this in a subprocess locally and got `returncode=-11` with a segfault. The output shows `Victim.__del__` before `Victim.__index__`, and the C stack includes:

```text
math_gcd
PyObject_Vectorcall
npy_ObjectGCD
PyUFunc_OO_O
```

As a control, keeping an extra Python reference to `victim_array[0]` before calling `np.gcd` makes the same script complete and return `[4]`, which matches the hypothesis that the old temporary tuple was providing the missing strong reference.

Suggested fix: keep strong refs around the vectorcall, at least for `npy_ObjectGCD`, e.g. `Py_INCREF(i1); Py_INCREF(i2); ... Py_DECREF(i2); Py_DECREF(i1);`, similar to the explicit argument protection already added in `PyUFunc_On_Om`.

eendebakpt:
> I think this PR introduces a borrowed-reference lifetime bug in the object `gcd` loop.

I agree. Similar bugs are already present on main, I am addressing those as well.

numpy/_core/src/umath/loops.c.src
```diff
@@ -215,37 +215,49 @@ NPY_NO_EXPORT void
...
+        return;
+    }
     UNARY_LOOP {
         PyObject *in1 = *(PyObject **)ip1;
         PyObject **out = (PyObject **)op1;
         /* We allow NULL, but try to guarantee non-NULL to downstream */
         assert(in1 != NULL);
         PyObject *ret, *func;
-        func = PyObject_GetAttrString(in1 ? in1 : Py_None, meth);
+        PyObject *obj = in1 ? in1 : Py_None;
```
seberg:
```suggestion
        PyObject *obj = Py_NewRef(in1 ? in1 : Py_None);
```

numpy/_core/src/umath/loops.c.src
```diff
@@ -215,37 +215,49 @@ NPY_NO_EXPORT void
...
+        /* hold obj: the attribute lookup can clear its array slot */
+        Py_INCREF(obj);
+        func = PyObject_GetAttr(obj, meth_name);
```
seberg, line 233: If we do this, should just go all the way and use `CallMethodNoArgs`, or is there some very strange reason that differs?  (I don't _think_ so, except that it should fast-path proper methods, and we use the `CallMethod` below.)

numpy/_core/src/umath/funcs.inc.src
```diff
@@ -156,33 +156,35 @@ npy_ObjectLogicalNot(PyObject *i1)
...
-                                 "O", obj);
+    return PyObject_CallOneArg(npy_static_pydata.math_trunc_func, obj);
 }
 
 static PyObject *
 npy_ObjectGCD(PyObject *i1, PyObject *i2)
 {
     PyObject *gcd = NULL;
 
+    /* hold i1 and i2: the calls below can clear their array slots */
```
seberg: Did this really change?  I can't say I like this here, it feels too magical to me. I.e. I would expect the "owner" to take care of it, which is the `OO_O` loop, instead we are guarding against a very indirectly borrowed reference being dropped due to recursion (not thread-safety, which we don't guarantee anyway).

I dunno, leaves me a bit unsatisfied, would it matter to just move this pattern into the main loop?

seberg (commented):
Generally looks like a fair and simple enough improvement, although the pattern for holding on to the references feels a bit annoying to me?

numpy/_core/src/umath/loops.c.src
eendebakpt: The changes so far are only to make the method safe for re-entrant calls. 

Looking at the code it seems we can change the PyObject_GetAttr/PyCallable_Check/PyObject_CallNoArgs into a single PyObject_CallMethodNoArgs. It might be a bit hard to keep the exact exception message though. Would it be okay to just return whatever error message PyObject_CallMethodNoArgs gives?

seberg: It seems OK unless you can think of a larger/more obvious change (such as a random attached function failing suddenly)?
But I _suspect_ that `CallMethod` is really just a fast-path that skips binding the method if it is one.

numpy/_core/src/umath/funcs.inc.src
eendebakpt: It was a pre-existing bug. Moving to the owner indeed seems nicer (I found other similar issue). There is one case with negative performance impact (np.absolute on object arrays containing small ints), but that needs the increfs to be safe as well.

numpy/_core/src/umath/loops.c.src
eendebakpt: Just calling `PyObject_CallMethodNoArgs` will get us back to https://github.com/numpy/numpy/issues/7961. We can call `PyObject_CallMethodNoArgs` and in case of failure inspect the exception type. Diff for that is https://github.com/numpy/numpy/commit/9774183086ca3fcf5e06f36ad48bde6f8e4a60e6

There would still be minor differences, e.g. 
```
import numpy as np
class A:
    sqrt = 10
    
x=A()
np.sin(x)
```
would give a different `TypeError`, but the main cases would be unchanged. To keep this PR focused I would prefer to leave it as is for now, but if you prefer to make the change that is fine wiht me as well.

seberg: Yeah, nevermind, I didn't really think of this weird chaining, annoying, but no point in polluting the PR with it then.

numpy/_core/src/umath/loops.c.src
```diff
@@ -282,11 +302,15 @@ PyUFunc_OOO_O(char **args, npy_intp const *dimensions, npy_intp const *steps, vo
...
         assert(in2 != NULL);
         assert(in3 != NULL);
-        PyObject *ret = f(
-            in1 ? in1 : Py_None,
-            in2 ? in2 : Py_None,
-            in3 ? in3 : Py_None
-        );
+        /* hold the operands: the call can run Python code that clears
+         * their array slots */
+        in1 = Py_NewRef(in1 ? in1 : Py_None);
```
eendebakpt, line 306: @seberg The ternary on `in1` is a bit odd since there is an assert `in1 != NULL` just above. I kept it because it was there historically (some defensive programming). I am fine with leaving as is, or removing the ternary.

numpy/_core/src/multiarray/calculation.c
```diff
@@ -642,9 +643,10 @@ PyArray_Round(PyArrayObject *a, int decimals, PyArrayObject *out)
         }
         if (decimals == 0) {
             if (out) {
-                return PyObject_CallFunction(n_ops.rint, "OO", a, out);
+                PyObject *args[2] = {(PyObject *)a, (PyObject *)out};
+                return PyObject_Vectorcall(n_ops.rint, args, 2, NULL);
             }
-            return PyObject_CallFunction(n_ops.rint, "O", a);
+            return PyObject_CallOneArg(n_ops.rint, (PyObject *)a);
```
mhvk, line 649: It is no big deal, but `PyObject_VectorCall` is part of the stable API, while `CallOneArg` is not, so maybe it makes sense to do `PyObject_VectorCall(n_ops.rint, (PyObject **)&a, 1, NULL);`?

numpy/_core/src/multiarray/datetime.c
```diff
@@ -2320,7 +2320,8 @@ get_tzoffset_from_pytzinfo(PyObject *timezone_obj, npy_datetimestruct *dts)
     }
 
     /* Convert the datetime from UTC to local time */
-    loc_dt = PyObject_CallMethod(dt, "astimezone", "O", timezone_obj);
+    loc_dt = PyObject_CallMethodOneArg(dt, npy_interned_str.astimezone,
```
mhvk, line 2323: Same here but `Py_ObjectVectorcallMethod`

numpy/_core/src/multiarray/vdot.c
```diff
@@ -147,11 +147,11 @@ OBJECT_vdot(char *ip1, npy_intp is1, char *ip2, npy_intp is2, char *op, npy_intp
     PyObject **tmp3;
     for (i = 0; i < n; i++, ip1 += is1, ip2 += is2) {
         if ((*((PyObject **)ip1) == NULL) || (*((PyObject **)ip2) == NULL)) {
-            tmp1 = Py_False;
-            Py_INCREF(Py_False);
+            tmp1 = Py_NewRef(Py_False);
         }
         else {
-            tmp0 = PyObject_CallMethod(*((PyObject **)ip1), "conjugate", NULL);
+            tmp0 = PyObject_CallMethodNoArgs(*((PyObject **)ip1),
```
mhvk, line 153: Possibly `PyObject_VectorcallMethod(*((PyObject **)ip1), npy_interned_str.conjugate, NULL, 0, NULL)` but have not checked this is allowed.

numpy/_core/src/umath/loops.c.src
```diff
@@ -215,37 +218,48 @@ NPY_NO_EXPORT void
...
             PyObject *exc, *val, *tb;
-            PyTypeObject *type = in1 ? Py_TYPE(in1) : Py_TYPE(Py_None);
+            PyTypeObject *type = Py_TYPE(obj);
             PyErr_Fetch(&exc, &val, &tb);
             PyErr_Format(PyExc_TypeError,
                          "loop of ufunc does not support argument %d of "
                          "type %s which has no callable %s method",
                          i, type->tp_name, meth);
             npy_PyErr_ChainExceptionsCause(exc, val, tb);
             Py_XDECREF(func);
```
mhvk: Unrelated to this PR, but why `XDECREF(func)` in a branch where `func == NULL`...

numpy/_core/src/umath/loops.c.src
mhvk: Yes, here and above that is pretty weird. The comment is also confusing "/* We allow NULL, but try to guarantee non-NULL to downstream */", especially right before a statement that explicitly does *not* allow NULL.

mhvk (approved):
This looks rather nice! I have some in-line comments about trying to use limited API stuff, but, really, we do not do elsewhere so it is no big deal.
Similarly, I am confused about the guards against NULL inputs, but arguably for this PR it is better to not change it.
Given that, approving!

numpy/_core/src/multiarray/calculation.c
eendebakpt: Thanks — you are right that `PyObject_Vectorcall` is stable ABI (since 3.12) while `PyObject_CallOneArg` is not. Some context though: numpy itself is not built against the limited API, and `PyObject_CallOneArg` is already used in a number of places in `_core/src` (e.g. `multiarray/methods.c`, `multiarray/descriptor.c`). `PyObject_CallOneArg` also passes `PY_VECTORCALL_ARGUMENTS_OFFSET` internally, which `PyObject_Vectorcall(..., (PyObject **)&a, 1, NULL)` cannot, so it can be marginally faster for callables that use the offset. Given that, I would slightly prefer to keep `PyObject_CallOneArg` here, but happy to switch if you feel the stable-ABI consistency is worth it.

mhvk: No, since we need to fix this in the end throughout numpy, and then can consider how we do it more generally, it is not worth it now if there is any doubt about it being equivalent.

numpy/_core/src/umath/loops.c.src
eendebakpt: For reference: code was added in https://github.com/numpy/numpy/pull/21817.

numpy/_core/src/umath/loops.c.src
eendebakpt: Unrelated, but indeed is not needed. Removed it.

numpy/_core/src/umath/loops.c.src
seberg: Well, the `assert` isn't in normal builds, it's an attempt to flush out places where NumPy doesn't correctly initialize new object arrays.
While being an `assert()` isn't in the final wheel/build so if downstream gets it wrong, we don't crash them.

So I think the comment adds up, you just have to remember that the `assert()` is effectively gone for any build that affects downstream.

seberg:
OK, I had another look through, one thing my agent found (aside from a few things I didn't care about) and it is actually right:
```
PyObject_CallFunction(f, "O", obj)
```
actually **unpacks** things if `obj` is a tuple, which is clearly utterly nonsense and a bug fix here:
```
np.floor(np.array((1,), dtype="O", ndmax=0))  # fails correctly now, floor((1,)) doesn't make sense.
```

I think Marten liked it and believe the comments were all addressed, so going to merge it (I edited the initial comment about this bug fix; it is arguably a larger change, but it also feels so super niche and clearly _correct_ that I am willing).
Note that the fix *doesn't* affect `np.frompyfunc()` which would worry me a lot more.

PR #31989, opened 2026-07-13
TYP: specialized ``gcd``/``lcm`` and binary bitwise ufuncs

jorenham:
This specialized the binary endo-ufuncs on `bool | integer | object_`:

- `gcd`
- `lcm`
- `[bitwise_]left_shift`
- `[bitwise_]right_shift`
- `bitwise_and`
- `bitwise_or`
- `bitwise_xor`

These first four technically don't accept `bool * bool` input (only the `bitwise_{and,or,xor}` do), but it was a lot easier to share the same specialized ufunc type for all of them.

Like #31941, I didn't add overloads for all possible integer x integer promotions, as that would require 100 or so overloads. Instead, I only special-cased the promotions of the most popular ones (according to github code search): `uint8`, `int32`, and `int64`. If the ufunc is called with some other dtype combo (e.g. `u16` x `i8`), then it'll select the relevant fallback overload and return  `Any` or `NDArray[Any]` (same as the current behavior).

---

No AI.


jorenham:
The new mypy_primer error for pandas ([src](https://github.com/pandas-dev/pandas/blob/d671ee2606c42eb0bde80ad44e73e3c1e9628829/pandas/core/dtypes/dtypes.py#L525)) is correct: the `bitwise_xor.reduce` return type isn't compatible with `builtins.int`, so mypy is right to complain here, and this is a typing issue in pandas.

The errors for static_frame can be avoided on our side, so gimme a minute to do that.

charris:
Thanks Joren.

PR #31991, opened 2026-07-14
TYP: disable ``bool``-only overloads of ``gcd``/``lcm``/``*_shift`` ufuncs

jorenham:
As I explained in #31989, I cheated a bit (for the sake of avoiding code duplication) by sharing the same ufunc specialization class for both the `integer | object_` and the `bool | integer | object_` classes of binary ufuncs. The problem is that the former (`gcd`/`lcm`/`{left,right}_shift`) doesn't accept all-`bool` input at runtime, whereas the latter (`bitwise_{and,or,xor}`) does. 

This adds a mechanism to selectively "disable" the `bool`-only overloads for the ufuncs that don't support this, by exploiting (contravariant) parametric polymorphism and `self` mathcing. 

Note that this requires using the legacy `TypeVar` instead of the new PEP 695 syntax because PEP 695 has no mechanism for specifying co-/contra-variance; it is instead "inferred" (theoretically variance inference is impossible, so this as a bad design decision, which is proved by this PR (grumble)).

Anyway, I think I'll try to build upon this bit of static typing acrobatics, by conditionally extending this specialized ufunc subtype with float-promotion overloads for e.g. `fmod`. If it'll work out, we'll be able to avoid a lot of code duplication.

At the moment it's still possible to pass all-`bool` arguments to the ufuncs, even though the relevant overloads are disabled. This is because of the *necessary* `__array_ufunc__` overloads at the end, which currently have no way of rejecting ufuncs based on its individual supported dtype domains. I have some ideas on how we can add those, but I'm not prioritizing this because most users don't mind false negatives (i.e. type-checkers not complaining avoid invalid calls), and care more about false positives (i.e. type-checkers complaining about valid calls).

---

No AI.


numpy/_core/umath.pyi
```diff
@@ -7411,11 +7412,11 @@ class _ufunc_21_divide(_ufunc_21[None]):  # type: ignore[misc]
         out: np.ndarray | None = None,
     ) -> OutT: ...
 
-# ?bBhHiIlLqQO, ?bBhHiIlLqQO => ?bBhHiIlLqQO
+# [?]bBhHiIlLqQO, [?]bBhHiIlLqQO => [?]bBhHiIlLqQO
 #
 # promotion table:
 #       b8   i8   u8  i16  u16  i32  u32  i64  u64  O
-# b8    b8   i8   u8  i16  u16  i32  u32  i64  u64  O
+# b8   [b8]  i8   u8  i16  u16  i32  u32  i64  u64  O
```
charris, line 7419: Out of curiousity, what does `[]` mean here.

charris:
Thanks Joren.

Issue #31930, opened 2026-07-09
SubTracker: Move static variables into a struct and access them via PyModule_GetState.


prathamhole14:
Long Term Effort Tracker: https://github.com/numpy/numpy/issues/31913

| Module | Status | Link |
| --- | --- | --- |
| `lapack_lite` | Done | #31928 |
| `_multiarray_umath` | In review | #31992 |
| `_umath_linalg` | No change needed | — |
| `_pocketfft_umath` | No change needed | — |
| `random` (9 Cython modules) | Out of scope | — |
| `_simd`, `_multiarray_tests` | No change needed | — |
| `_rational_tests` | Open | — |

### TODO

- [ ] State propagation in `_multiarray_umath` #31992 requires separate PRs 
- [ ] Limited API for `_multiarray_umath` depends on the above.

ngoldbaum:
For these subtracker issues I'd like to see a bit more detail about what you're planning to do in the description.

da-woods:
You probably already know this (in that I'm sure you've thought about it more than me), but I think you have some "fun" to come with `import_array()`.

I think you need to fix both sides of it, in that
1. The macro uses a C global `static void**` as your array API table (rather than a pointer in the module state).
2. Ultimately it'll need to point to per-interpreter types rather than static types.

prathamhole14:
Hi everyone!
This is a high level overview of how I plan to migrate to module-level state for the `_multiarray_umath` module.

1. Infrastructure set up containing these 5 structs and other scattered global variables (zero functional changes). 
       A : Add to the bottom of `npy_static_data.h` - No change header file import 
       B : Create a new module_state.h in multiarray - Clean in one single 
       I prefer option B but please let me know what you think. 
2. Migration PR for each struct `npy_interned_str_struct`, `npy_static_pydata_struct`, `npy_static_cdata_struct` ,`npy_runtime_imports_struct`, `npy_global_state_struct`.
  To keep the PRs for the larger structs easy to review, I plan to temporarily use a global pointer `_npy_module_state` to access the   module state. This avoids having to thread the state pointer through dozens of deep call chains in a single PR. We can remove the  transitional global pointer once all migrations are complete. 
Temporary global pointer like `multiarray_umath_state *_npy_module_state = NULL` 

4. Migration of other scattered globals just like structs.
This is high level of how I am planning to proceed, please let me know what you think then I will start creating PRs one by one. 

CC: @mattip @ngoldbaum @rgommers @da-woods

[...]

PR #31992, opened 2026-07-14
ENH: Migrate `_multiarray_umath` global variables to per-module state

prathamhole14:
### PR summary
Planner for this migration is this https://github.com/numpy/numpy/issues/31930#issuecomment-4958219779, but instead of multiple PRs there will be multiple commits in this PR itself. 

Related to: https://github.com/numpy/numpy/issues/31930

**Request to reviewers**
I am planning on having 7 major commits in this PR, first commit is for infra, then next 5 for all structs and the 7th is for scattered globals.  

So at each stage before I push next commit I would like to get an approval, so I can maintain clean history and this might also turn out useful in future if we want to go back/revert a particular stage or in debugging. 

CC: @mattip @ngoldbaum @rgommers 
#### AI Disclosure

Used AI (claude and gemini) for researching and finding out what I have to work on and writing comments (intentionally asked it to make comments verbose for now -- so they can be used as trackers and later I will make them concise) other than that most of the code, logic, and architecture were written by me.


mattip:
This looks correct to me so far. Do we have module teardown tests to make sure the GC paths are hit? You can try adding a temporary debug print statement in the `m_traverse`, `m_clear`, `m_free` functions and then run tests to make sure they are called somewhere.

prathamhole14:
>You can try adding a temporary debug print statement in the m_traverse, m_clear, m_free functions and then run tests to make sure they are called somewhere.

I Added temporary `fprintf(stderr, ...)` to all three GC hooks and ran `gc.collect()` test (and later used spin test as well) after importing NumPy. The output confirmed `m_traverse` is called repeatedly (as expected) but did not get an `m_clear` or `m_free`. Is this expected as they happen only after interpreter/subinterpreter shutdown?
```python
sys.stderr.flush()
n = gc.collect()
sys.stderr.flush()
```

Also I am ready with the next commit locally, please let me know when can I push it!

mattip:
Is the global state the only blocker for subinterpreter support or are there more nuances I am not aware of? Maybe we could add a xfailing test that adds a subinterpreter and shuts it down, I don't think it should segfault. Then in that test you could make sure it hits the code paths but still fails since we do not yet support subinterpreters.

Go ahead with the next commit as well.

numpy/_core/src/multiarray/compiled_base.c
```diff
@@ -1503,7 +1504,7 @@ arr_add_docstring(PyObject *NPY_UNUSED(dummy), PyObject *const *args, Py_ssize_t
 
     /* Don't add docstrings */
 #if PY_VERSION_HEX > 0x030b0000
-    if (npy_static_cdata.optimize > 1) {
+    if (_npy_module_state->static_cdata.optimize > 1) {
```
mattip: Hmm. Are you sure using the global static `_npy_module_state` is the best path forward here? This function, executed only at startup, would seem to be a good candidate for seeing how disruptive a migration to the `PyModule_GetState` model would be.

prathamhole14: I initially used the global here because I plan to completely eliminate `_npy_module_state` in a follow-up PR that handles the complex state propagation. But yeah, you're right for functions like this that already have easy access to the module object, I will use the `get_module_state()` wrapper (which calls `PyModule_GetState`) right now.

I will address the deeper functions that require threading the state pointer through the call stack in the next PR. Let me know what you think!

prathamhole14:
If someone could please "approve" [Migrate npy_static_cdata to module state](https://github.com/numpy/numpy/pull/31992/changes/ac1c949dfc000ea0138f4c0f6f17bca67bd6eaf4), then I will push the next migration commit afterwards thanks. 

Edit: Oops I had to push [Migrate npy_global_state to per-module state](https://github.com/numpy/numpy/pull/31992/changes/131f8e9aad214074e02be621fe9de4560c24e683) since I did some minor ref in [Migrate npy_static_cdata to module state](https://github.com/numpy/numpy/pull/31992/changes/cda4a46374922926d5620d3fce717bd71f99bd4a)

ngoldbaum:
Does this have performance overhead? I looked over the diff and didn’t see any spots where we call the function in a tight loop so I doubt there is, but I want to make sure you’re thinking about that.

prathamhole14:
```
| Change   | Before [941bc425] <main>   | After [131f8e9a] <multiarray_module_state>   |   Ratio | Benchmark (Parameter)                                                                    |
|----------|----------------------------|----------------------------------------------|---------|------------------------------------------------------------------------------------------|
| +        | 799±40μs                   | 986±3μs                                      |    1.23 | bench_core.CountNonzero.time_count_nonzero(2, 1000000, <class 'numpy.int64'>)            |
| +        | 104±0.9μs                  | 114±4μs                                      |    1.1  | bench_core.CountNonzero.time_count_nonzero(3, 1000000, <class 'bool'>)                   |
| +        | 586±4μs                    | 637±0.6μs                                    |    1.09 | bench_core.CountNonzero.time_count_nonzero_multi_axis(2, 1000000, <class 'bool'>)        |
| +        | 886±3μs                    | 963±5μs                                      |    1.09 | bench_core.CountNonzero.time_count_nonzero_multi_axis(3, 1000000, <class 'bool'>)        |
| +        | 298±0.5μs                  | 320±0.2μs                                    |    1.08 | bench_core.CountNonzero.time_count_nonzero_multi_axis(1, 1000000, <class 'bool'>)        |
| +        | 194±5μs                    | 208±0.3μs                                    |    1.07 | bench_core.CountNonzero.time_count_nonzero(3, 10000, <class 'object'>)                   |
| +        | 15.5±0.4ms                 | 16.6±0.1ms                                   |    1.07 | bench_core.CountNonzero.time_count_nonzero_multi_axis(3, 1000000, <class 'str'>)         |
| +        | 735±2ns                    | 787±6ns                                      |    1.07 | bench_core.StatsMethods.time_sum('float64', 100)                                         |
| +        | 363±0.4ns                  | 385±20ns                                     |    1.06 | bench_core.CountNonzero.time_count_nonzero(3, 100, <class 'bool'>)                       |
| +        | 880±3μs                    | 929±30μs                                     |    1.06 | bench_core.CountNonzero.time_count_nonzero_axis(3, 1000000, <class 'bool'>)              |
| +        | 8.61±0.02μs                | 9.13±0.08μs                                  |    1.06 | bench_core.CountNonzero.time_count_nonzero_multi_axis(2, 10000, <class 'bool'>)          |
| +        | 12.9±0.02μs                | 13.7±0.04μs                                  |    1.06 | bench_core.CountNonzero.time_count_nonzero_multi_axis(3, 10000, <class 'numpy.int8'>)    |
| +        | 804±7ns                    | 849±10ns                                     |    1.06 | bench_core.StatsMethods.time_prod('int64', 100)                                          |
| +        | 7.31±0.08μs                | 7.70±0.04μs                                  |    1.05 | bench_core.CountNonzero.time_count_nonzero_multi_axis(1, 10000, <class 'numpy.int32'>)   |
| +        | 11.7±0.02μs                | 12.3±0.01μs                                  |    1.05 | bench_core.CountNonzero.time_count_nonzero_multi_axis(3, 10000, <class 'bool'>)          |
| +        | 15.1±0.03μs                | 15.9±0.05μs                                  |    1.05 | bench_core.CountNonzero.time_count_nonzero_multi_axis(3, 10000, <class 'numpy.float32'>) |
| +        | 15.1±0.04μs                | 15.9±0.08μs                                  |    1.05 | bench_core.CountNonzero.time_count_nonzero_multi_axis(3, 10000, <class 'numpy.int32'>)   |
| +        | 735±7ns                    | 773±4ns                                      |    1.05 | bench_core.StatsMethods.time_max('float32', 100)                                         |
| +        | 1.08±0μs                   | 1.14±0.01μs                                  |    1.05 | bench_core.StatsMethods.time_prod('complex64', 100)                                      |
| -        | 1.32±0.04ms                | 1.25±0.03ms                                  |    0.95 | bench_core.CountNonzero.time_count_nonzero_multi_axis(3, 1000000, <class 'numpy.int8'>)  |
| -        | 416±3ns                    | 393±2ns                                      |    0.94 | bench_core.Core.time_array_l1                                                            |
| -        | 131±2μs                    | 124±2μs                                      |    0.94 | bench_core.CountNonzero.time_count_nonzero_multi_axis(1, 10000, <class 'object'>)        |
| -        | 389±0.7μs                  | 366±1μs                                      |    0.94 | bench_core.CountNonzero.time_count_nonzero_multi_axis(1, 1000000, <class 'numpy.int16'>) |
| -        | 1.58±0.04ms                | 1.48±0.03ms                                  |    0.94 | bench_core.CountNonzero.time_count_nonzero_multi_axis(3, 1000000, <class 'numpy.int16'>) |
| -        | 7.91±0.9μs                 | 7.46±0.01μs                                  |    0.94 | bench_core.NumPyChar.time_add_big_list_small_string                                      |
| -        | 6.91±0.6μs                 | 6.44±0.04μs                                  |    0.93 | bench_core.Core.time_diag_l100                                                           |
| -        | 347±0.2μs                  | 322±2μs                                      |    0.93 | bench_core.CountNonzero.time_count_nonzero_multi_axis(1, 1000000, <class 'numpy.int8'>)  |
| -        | 774±4μs                    | 723±7μs                                      |    0.93 | bench_core.CountNonzero.time_count_nonzero_multi_axis(2, 1000000, <class 'numpy.int8'>)  |
| -        | 395±3μs                    | 368±0.3μs                                    |    0.93 | bench_core.CountNonzero.time_count_nonzero_multi_axis(3, 10000, <class 'object'>)        |
| -        | 803±6μs                    | 736±1μs                                      |    0.92 | bench_core.CountNonzero.time_count_nonzero(3, 1000000, <class 'numpy.int32'>)            |
| -        | 143±10μs                   | 124±0.1μs                                    |    0.87 | bench_core.CountNonzero.time_count_nonzero_axis(1, 10000, <class 'object'>)              |
| -        | 40.4±3ms                   | 35.0±0.05ms                                  |    0.87 | bench_core.CountNonzero.time_count_nonzero_axis(3, 1000000, <class 'object'>)            |

PERFORMANCE DECREASED.
```

I ran `spin bench -c main -t bench_core.CountNonzero` comparing main against an empty dummy commit. Even with identical code, the results fluctuated by ±10%. (Note: These were run on my local PC).

Apologies for the delay! I ran into some major Ubuntu issues and had to switch to Debian XFCE. My SSH key for the qgpu servers is scheduled to be updated after 3 AM UTC today. After that I willl run the tests on the server to get a more reliable results.

ngoldbaum:
Why did you choose that benchmark to look at? Would you expect the changes in this PR to have a significant impact? If so, why? It’s not clear to me and I suspect you’re measuring noise.

prathamhole14:
I first ran `spin bench -c main -t bench_core` on the `multiarray_module_state` branch (results above) since I was not sure how much can I trust the benchmarks, that's why intentionally I chose `bench_core.CountNonzero` to find out how much is the noise because it showed the largest regressions (1.23x) and I also think my changes from this PR shouldn't affect `CountNonzero`.  

>Would you expect the changes in this PR to have a significant impact? 

Nope, I think it should be little to none till now atleast (this is my intuition - needs reliable testing)
> If so, why? It’s not clear to me and I suspect you’re measuring noise.

Yes I was measuring noise.  Here are the `bench_core.CountNonzero` benchmarks against dummy commit I did earlier:
Even with completely identical code, the results fluctuated by ±10% on my pc. 

```
| Change   | Before [941bc425] <main>   | After [c3e0325a] <test_benchmark_noise>   |   Ratio | Benchmark (Parameter)                                                             |
|----------|----------------------------|-------------------------------------------|---------|-----------------------------------------------------------------------------------|
| +        | 601±20μs                   | 664±30μs                                  |    1.1  | bench_core.CountNonzero.time_count_nonzero(3, 1000000, <class 'numpy.int32'>)     |
| +        | 57.6±0.3μs                 | 62.5±0.1μs                                |    1.08 | bench_core.CountNonzero.time_count_nonzero_multi_axis(3, 10000, <class 'str'>)    |
| +        | 6.00±0.09μs                | 6.39±0.3μs                                |    1.07 | bench_core.CountNonzero.time_count_nonzero_multi_axis(2, 100, <class 'object'>)   |
| -        | 56.0±2μs                   | 50.6±0.2μs                                |    0.9  | bench_core.CountNonzero.time_count_nonzero(3, 10000, <class 'str'>)               |
| -        | 305±10μs                   | 274±2μs                                   |    0.9  | bench_core.CountNonzero.time_count_nonzero_multi_axis(2, 10000, <class 'object'>) |

PERFORMANCE DECREASED.
```

prathamhole14:
https://github.com/numpy/numpy/issues/32056#issuecomment-5037772880 benchmarks till [Migrate npy_interned_str to per-module state](https://github.com/numpy/numpy/pull/31992/changes/5fffd0c35fdf8033b0e08a5470743a0370ffaeba)

prathamhole14:
Dug into the remaining scattered globals (`PyDataMem_DefaultHandler`,`userdescrs`/`NPY_NUMUSERTYPES`, and the `DType` singleton pointers in `dtypemeta.c`/`abstractdtypes.c`). These cannot be migrated right now since all of them are entries in `PyArray_API`, one literal array compiled into the `.so`, so there's exactly one copy per process. A module state field only fixes our internal copy, external code still reads the old shared slot directly.

Two categories:

1. `PyDataMem_DefaultHandler` and `NPY_NUMUSERTYPES` are raw variable slots, external code dereferences `PyArray_API[n]` as the variable's own memory. Making them real per interpreter means turning them into `accessor` functions instead, same pattern as `PyDataMem_SetHandler` and `GetHandler` already use. That's an ABI break per `numpy_api.py`'s own rules and needs a C API version change.

2. The `DType` pointers store the address of one canonical type object. Heap typing them fixes a different problem, one struct being mutated by multiple interpreters, not the table only being able to hold one address per slot. Same category as `PyArray_Type`.

Questions for you all:

- `PyDataMem_DefaultHandler` is basically read only after init. Is it fine to document as a permanent shared exception or do we want the ABI change?
- `NPY_NUMUSERTYPES` can't just be documented as shared, `PyArray_RegisterDataType` still mutates it live, so migrating `userdescrs` alone would `desync` registrations across interpreters. ABI fix, or do we document legacy user dtype registration as not subinterpreter safe?
- DType singletons, is shared type identity across interpreters acceptable? or do we want per interpreter type tables eventually?

None of this blocks what this PR already have, it only blocks the final `Py_mod_multiple_interpreters` flip and the `TestSubinterpreterTeardown` xfail. I am not sure if this is the right place to discuss this, maybe a separate issue?

prathamhole14 (commented):
Once this PR is merged I will put series of independent PR which will help in state propagation for each struct and other scattered globals.

ngoldbaum:
I'm not sure of the consequences of sharing the memory handler between interpreters.

I think for all of these you should just document them in e.g. a subinterpreter support tracking issue. These all sound like things we could conceivably fix with a NumPy 3.0-style ABI and API break but we'd have to first decide we want to do that and then figure out how to do that while avoiding downstream breakage as much as possible.

Since we don't yet support subinterpreters or limited API builds, we don't need to worry about possibly breaking any early experimental support you come up with as part of your internship.

prathamhole14:
Thank you for the review @ngoldbaum,  
> I think for all of these you should just document them in e.g. a subinterpreter support tracking issue.

There isn't a sub interpreters tracking issue yet, I will put one soon. 

Meanwhile Yes then I think this PR is merge ready, please let me know if there are any required changes. 

EDIT: In the meantime I will review this once myself -- I am afraid while fixing merge conflict If I have not messed things up.

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -5006,24 +5009,404 @@ set_flaginfo(PyObject *d)
...
+    Py_CLEAR(state->n_ops.real);
+    Py_CLEAR(state->n_ops.imag);
+
+    return 0;
+}
+
+static void
+multiarray_umath_free(void *m)
+{
+    /* Nothing to free yet; non-PyObject C data is trivially destructed */
```
ngoldbaum: What about the module object itself:

```suggestion
    multiarray_umath_clear((PyObject *)m);
    _npy_module_state = NULL;
```

numpy/_core/src/multiarray/number.h
```diff
@@ -45,7 +45,7 @@ typedef struct {
     PyObject *imag;
 } NumericOps;
 
-extern NPY_NO_EXPORT NumericOps n_ops;
+/* n_ops lives in multiarray_umath_state.n_ops (module_state.h) */
```
ngoldbaum: I'd prefer not leaving comments behind like this after this migration, people can use grep

numpy/_core/src/multiarray/number.h
ngoldbaum, line 3: `module_state.h` includes this header and that header is used in C++ translation units, so this header needs `extern "C"` blocks.

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -5438,11 +5826,14 @@ static struct PyModuleDef_Slot _multiarray_umath_slots[] = {
...
-    .m_methods = array_module_methods,
-    .m_slots = _multiarray_umath_slots,
+    .m_base     = PyModuleDef_HEAD_INIT,
+    .m_name     = "_multiarray_umath",
+    .m_size     = sizeof(multiarray_umath_state),
+    .m_methods  = array_module_methods,
+    .m_slots    = _multiarray_umath_slots,
+    .m_traverse = multiarray_umath_traverse,
+    .m_clear    = multiarray_umath_clear,
+    .m_free     = multiarray_umath_free,
```
ngoldbaum, line 5707: Doesn't there need to be some way to keep the module alive until interpreter shutdown (e.g. an unbalanced `Py_INCREF` during module init)? Otherwise `m_free` could actually be invoked and NumPy would leave behind live references to dead objects in the C API. Of course if we ever want to let NumPy support subinterpreters we'd need to fix that, but I think given your findings about the public C API, that also means we'd need to make a C API break, which implies a NumPy 3.0 release. So in other words, we probably won't be supporting subinterpreters anytime soon, so `m_free` and `m_clear` should probably never fire in-practice.

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -5006,24 +5009,404 @@ set_flaginfo(PyObject *d)
...
+    Py_CLEAR(state->n_ops.minimum);
+    Py_CLEAR(state->n_ops.rint);
+    Py_CLEAR(state->n_ops.conjugate);
+    Py_CLEAR(state->n_ops.matmul);
+    Py_CLEAR(state->n_ops.clip);
+    Py_CLEAR(state->n_ops.real);
+    Py_CLEAR(state->n_ops.imag);
+
+    return 0;
+}
```
ngoldbaum: these hand-written implementations will definitely go stale.

It'd be much better if this was checked/generated based on a statically defined (and checked in a test) list of static data.

numpy/_core/src/umath/extobj.c
```diff
@@ -118,8 +119,8 @@ fetch_curr_extobj_state(npy_extobj *extobj)
 {
     PyObject *capsule;
     if (PyContextVar_Get(
-            npy_static_pydata.npy_extobj_contextvar,
-            npy_static_pydata.default_extobj_capsule, &capsule) < 0) {
+            npy_get_module_state()->static_pydata.npy_extobj_contextvar,
+            npy_get_module_state()->static_pydata.default_extobj_capsule, &capsule) < 0) {
```
ngoldbaum: here any many times elsewhere you're calling the function twice, IMO it's better and clearer to bind a local variable and not rely on the compiler optimizing this for you

numpy/_core/src/multiarray/descriptor.c
```diff
@@ -140,16 +140,17 @@ _try_convert_from_dtype_attr(PyObject *obj)
...
 
     if (!PyArg_ParseTuple(args, "O:set_typeDict", &dict)) {
         return NULL;
     }
+    multiarray_umath_state *state = get_module_state(module);
     /* Decrement old reference (if any)*/
-    Py_XDECREF(typeDict);
-    typeDict = dict;
+    Py_XDECREF(state->typeDict);
+    state->typeDict = dict;
```
ngoldbaum: pre-existing but this should be `Py_XSETREF(state->typeDict, Py_NewRef(dict));`. You could even send in a separate PR for that fix so it lands in a reasonable amount of time and can be backported.

numpy/_core/tests/test_multiarray.py
```diff
@@ -11916,3 +11921,53 @@ def test_match_sequence_pattern_3d(self):
...
+    is declared in multiarraymodule.c).
+    """
+
+    @pytest.mark.xfail(
+        reason=(
+            "numpy._core._multiarray_umath has not yet opted in to "
+            "Py_MOD_MULTIPLE_INTERPRETERS_SUPPORTED global state "
+            "migration is still in progress."
+        )
+    )
```
ngoldbaum: I'm not a fan of adding new non-`strict` xfail tests. Especially ones like these where a future code change would make these pass and not some element of randomness.

https://docs.pytest.org/en/stable/reference/reference.html#pytest-mark-xfail

ngoldbaum (commented):
I asked an AI model to look this over and it created the following table of possibly performance-sensitive changes in this PR:

| Site | Frequency |
|---|---|
| `number.c` — every `nb_*` slot (`array_add`, `array_multiply`, in-place variants, `fast_scalar_power`) | every Python operator on an ndarray |
| `arrayobject.c` — `array_richcompare`, `_void_compare` | every `==`/`<`/… on an ndarray |
| `scalartypes.c.src` — `gentype_*` binops, `gentype_power`, `find_binary_operation_path` | every operator on a NumPy scalar — the most overhead-sensitive path NumPy has |
| `extobj.c:118` — `fetch_curr_extobj_state` (via `_get_bufsize_errmask`) | start of essentially **every ufunc call**; currently does *two* separate state fetches |
| `alloc.c` — `PyDataMem_GetHandler` (every new array buffer), `indicate_hugepages` (≥4 MiB allocs) | per array creation |
| `arrayfunction_override.c` — `get_array_function`, `is_default_array_function`, `try_reduction` | every `__array_function__`-dispatched public-API call |
| `methods.c` — `NPY_FORWARD_NDARRAY_METHOD` | every forwarded method (`.mean()`, `.sum()`, …) |
| `binop_override.h` — `binop_should_defer` | mixed-type binops only (same-type ops early-exit before the state access), and the adjacent `__array_ufunc__` attribute lookup dwarfs it |

I also have some inline comments, see below

So to better settle the performance overhead, I'd like you to try running the following benchmarks on an unshared machine before and after this PR:

* `bench_scalar`
* `bench_ufunc` (in particular the benchmarks using small arrays, to measure ufunc overhead)
* small-array binops in `bench_core`
* `bench_array_coercion`

If there is any measurable overhead, it should show up there. If you do find it, the fix is to lift the access via a function into a function-local variable so the function call doesn't happen repeatedly.

numpy/_core/src/multiarray/multiarraymodule.c
prathamhole14: Oh yeah right.

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -5036,6 +5115,19 @@ _multiarray_umath_exec(PyObject *m) {
     }
     module_loaded = 1;
 
+    /*
+     * State that is never torn down lives outside the module state: the
+     * static types, the PyArray_API table and the builtin descriptor and
+     * DType singletons reachable through it. Clearing the module state while
+     * those are still reachable would leave dead pointers in the public C
+     * API, so leak a reference to keep m_clear and m_free from running.
```
ngoldbaum: One might think that the ~150 `Py_VISIT` calls in `m_traverse` might also be unnecessary. I checked with an AI-generated gc benchmark and the cost is tiny so it's not worth worrying about.

numpy/_core/src/multiarray/module_state.h
```diff
@@ -0,0 +1,126 @@
...
+ * Use this only in internal functions deep in the call chain that cannot
+ * easily receive a module pointer yet. Prefer get_module_state() where
+ * a module pointer is available.
+ *
+ * FIXME: Remove all call sites once the full migration is complete.
+ */
+static inline multiarray_umath_state *
+npy_get_module_state(void)
+{
+    assert(_npy_module_state != NULL);
```
ngoldbaum: Is this assert really necessary? It's most of the cost of this function and numpy compiles in asserts by default. Not in releases, but still, I think debug builds should be performant too. This pointer is loaded once and never unset because the module free never runs. Seems unnecessary to me to check that fact every time we call this function.

Take a look at the disassembly in godbolt: https://godbolt.org/z/ofvf7GeG5. The assert is most of the instructions.

If you get rid of the assert, why not just leave `_npy_module_state` as a global?

ngoldbaum (commented):
Just one request below but otherwise I think this is all mechanically clean.

numpy/_core/src/multiarray/module_state.h
prathamhole14: Yeah, since it's not released in release build I had put it but yes it makes sense to get rid of it entirely, therefore `npy_get_module_state()` is just `_npy_module_state` -- replaced it in the latest commit.

Thanks for sharing the disassembly in godbolt -- I will now use it in my workflow :)

ngoldbaum (approved):
I'm approving but I'm going to leave this to someone else to give this a once-over before merging.

Thanks @prathamhole14!

numpy/_core/src/common/npy_cpu_dispatch.c
```diff
@@ -25,21 +27,28 @@ npy_cpu_dispatch_tracer_init(PyObject *mod)
     if (err != 0) {
         return -1;
     }
-    npy_static_pydata.cpu_dispatch_registry = reg_dict;
+    /*
+     * state->static_pydata.cpu_dispatch_registry is a separate owned
+     * reference from the one held by mod_dict above, since it is
+     * visited/cleared by the module's GC hooks.
+     */
+    Py_INCREF(reg_dict);
```
kumaraditya303: This is unnecessary, there is a decref above which can removed and ref can be stolen instead.

numpy/_core/src/common/npy_import.c
```diff
@@ -6,13 +6,25 @@
 #include <stdatomic.h>
 
 
-NPY_VISIBILITY_HIDDEN npy_runtime_imports_struct npy_runtime_imports;
+/*
+ * This mutex only guards the act of importing (i.e. writing a freshly
```
kumaraditya303: Can you make the comments shorter? It can simple as "Process global mutex to serialize first time import..."

numpy/_core/src/multiarray/module_state.h
```diff
@@ -0,0 +1,116 @@
...
+#include "multiarraymodule.h"     /* npy_global_state_struct */
+#include "number.h"               /* NumericOps */
+#include "module_state_fields.h"  /* NPY_*_FIELDS field lists */
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*
+ * Master module state struct.
```
kumaraditya303: Same, I'd like the comments to be minimal and you shouldn't explain things like how much memory Python allocates, in future CPython can allocate more because of alignment change and the comment would be incorrect.

numpy/_core/tests/test_multiarray.py
```diff
@@ -28,6 +28,11 @@
 
 import pytest
 
+try:
+    import _interpreters
```
kumaraditya303: The test shouldn't use private module, it should the public interpreter executor.

prathamhole14: I had used `_interpreters` to get coverage on 3.13 too since `concurrent.interpreters` only landed in `3.14`, so the it means the test skips below that. Do you still want me to use `concurrent.interpreters` or the current changes are fine?

kumaraditya303: IIRC there are bugs in implementation on 3.13, it is better to use 3.14+ interface and skip on 3.13.

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -5162,6 +5236,16 @@ _multiarray_umath_exec(PyObject *m) {
     }
     module_loaded = 1;
 
+    /*
+     * The static types and the PyArray_API table outlive the module, so
+     * clearing the module state would leave dead pointers in the public C
+     * API. Leak a reference to keep m_clear and m_free from running.
+     */
+    Py_INCREF(m);
```
kumaraditya303: I think it is bit fragile, if m_clear and m_free shouldn't run at all then they should be defined as no-op functions, otherwise an unbalanced ref count (which might a bug in user code) can trigger them.

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -5132,24 +5138,92 @@ set_flaginfo(PyObject *d)
...
     }
 
     return 0;
 }
 
 static int module_loaded = 0;
 
+static int
+multiarray_umath_traverse(PyObject *m, visitproc visit, void *arg)
+{
```
kumaraditya303, line 5199: For time being I think these should be no-op functions

numpy/_core/src/umath/ufunc_object.c
```diff
@@ -5966,7 +5972,8 @@ prepare_input_arguments_for_outer(PyObject *args, PyUFuncObject *ufunc)
     PyObject *tmp1 = PyTuple_GET_ITEM(args, 0);
     PyObject *tmp2 = PyTuple_GET_ITEM(args, 1);
 
-    int is_matrix = PyObject_IsInstance(tmp1, npy_runtime_imports.numpy_matrix);
+    int is_matrix = PyObject_IsInstance(
```
kumaraditya303: minor: I don't think the line will be too long even you keep it single line, that would also be easier to read IMO, same for others

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -5132,24 +5137,117 @@ set_flaginfo(PyObject *d)
...
     return 0;
 }
 
 static int module_loaded = 0;
 
+/*
+ * The module state is never torn down: the static types and the PyArray_API
+ * table keep using objects it holds. These slots are no-ops so an unbalanced
+ * Py_DECREF cannot clear it either.
+ */
```
kumaraditya303, line 5177: do `#if 1 ... #else ` instead of `#if 0` and add comment to 1

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -5162,6 +5260,15 @@ _multiarray_umath_exec(PyObject *m) {
     }
     module_loaded = 1;
 
+    /*
+     * Deallocating the module also frees the module state, so leak a
+     * reference to keep it valid for the life of the process.
+     */
+    Py_INCREF(m);
```
kumaraditya303: this is no longer needed

prathamhole14: The no-op slots prevent `m_clear` from wiping the fields but `module_dealloc` still calls `PyMem_Free (md_state)` when the last reference is gone which leaves `_npy_module_state` dangling. So if numpy is removed from `sys.modules` we may read freed memory later instead of raising the expected `module_loaded ImportError`

I added this @ngoldbaum request above https://github.com/numpy/numpy/pull/31992#discussion_r3677409075. Let me know if you are looking at something else that keeps the module alive.

kumaraditya303: In that case, we should probably mark the module objects as immortal or for time being allocate `_npy_module_state` statically for now until subinterpreters are truly supported in numpy.

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -5563,11 +5672,14 @@ static struct PyModuleDef_Slot _multiarray_umath_slots[] = {
...
 
 static struct PyModuleDef moduledef = {
-    .m_base = PyModuleDef_HEAD_INIT,
-    .m_name = "_multiarray_umath",
-    .m_size = 0,
-    .m_methods = array_module_methods,
-    .m_slots = _multiarray_umath_slots,
+    .m_base     = PyModuleDef_HEAD_INIT,
+    .m_name     = "_multiarray_umath",
+    .m_size     = sizeof(multiarray_umath_state),
```
kumaraditya303, line 5702: until subinterpreters are truly supported I suggest to **not** allocate module state on heap but statically i.e. keep `m_size = 0`

mattip: Why not allocate on the heap now? Is there a price?

numpy/_core/src/multiarray/multiarraymodule.c
prathamhole14: `PyUnstable_SetImmortal` cannot be used as it is python 3.15 only.

numpy/_core/src/multiarray/multiarraymodule.c
kumaraditya303: If the module gets deallocated then state would be freed, till now the state was statically allocated so there was no lifetime issues but now if an extension holds borrowed refs to these objects then after module gets deallocated it would point to freed memory. I'd advise against doing this until there is a clear direction of how numpy would expose its API with support for subinterpreters.

mattip: That is why there is an incref on line 5267 - to prevent the module refcnt from ever reaching 0.

prathamhole14: Yes I agree `Py_INCREF` covers that case and it also survives GC. I also checked with a minimal `m_size>0` module and without incref `m_free` ran and with it it never ran.

I feel this is safe, also if we are to `m_size = 0` here then we need to do the same for other per module state transitions I did earlier.

numpy/_core/src/multiarray/multiarraymodule.c
```diff
@@ -5162,6 +5261,15 @@ _multiarray_umath_exec(PyObject *m) {
...
     module_loaded = 1;
 
+    /*
+     * Deallocating the module also frees the module state, so leak a
+     * reference to keep it valid for the life of the process.
+     */
+    Py_INCREF(m);
+
+    multiarray_umath_state *state = get_module_state(m);
+    _npy_module_state = state;
```
mattip: In the error paths, nothing resets `_npy_module_state` to `NULL`. Also there is no `Py_DECREF(m)`. I think it would be cleaner if they were reset.

numpy/_core/src/multiarray/multiarraymodule.c
kumaraditya303: Ok, let's keep it as `Py_INCREF`

numpy/_core/src/multiarray/multiarraymodule.c
prathamhole14: Done i have moved the init body to `_multiarray_umath_exec_impl` and `_multiarray_umath_exec` is now wrapper that handles the incref and state assignment and if  the implementation fails it resets `_npy_module_state` and drops the reference.

kumaraditya303 (approved):

numpy/_core/src/multiarray/module_state.h
```diff
@@ -0,0 +1,76 @@
...
+    return (multiarray_umath_state *)state;
+}
+
+/*
+ * Only one state ever exists: the module opts out of subinterpreters and
+ * refuses a second load. Use only where no module pointer is reachable.
+ *
+ * FIXME: Remove once all access sites receive the module or state pointer.
+ */
+NPY_VISIBILITY_HIDDEN extern multiarray_umath_state *_npy_module_state;
```
seberg, line 70: (pretty unrelated to merging) So I guess we mentioned it yesterday a bit, but what is the plan here, how much of this can we fix up in a relatively quick followup and how much of it is blocked/hard to do?

numpy/_core/src/multiarray/module_state_fields.h
```diff
@@ -0,0 +1,212 @@
...
+ *
+ * interned_str.errmode_strings is an array, so it is handled separately at
+ * each use site.
+ */
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define NPY_INTERNED_STR_FIELDS(F) \
```
seberg, line 16: Likely it's just the way to do it, but my first reaction to seeing this is "there should be way to avoid duplicating all fields".

seberg (commented):
Just silly comments browsing through, don't worry about them.

I suppose until there is actually more than one global state this effectively does nothing (e.g. we can't remove the guards against reloading NumPy -- although I guess the full thing there may be harder and also require dealing with heap types, etc... seems like a very long tail to get somewhere).

(I am slightly wondering if we should just almost always use an early function level `state = ...` even if `...` is `_npy_module_state` for now. But it also doesn't matter here, more of a mid-term code style question.)

numpy/_core/src/multiarray/module_state.h
prathamhole14: I don't think this will take a lot of time, I can create 5 parallel PRs (or a single big one) for propagation and after this the FIXME can be addressed. But I am thinking if I should do limited api work first or the one I just mentioned since per module state is not requirement for LIMITED API (it is for subinterpreters).

mattip: I would like to see progress on limited API. Can we open a follow-on issue for the remaining fixup?

numpy/_core/src/multiarray/module_state_fields.h
prathamhole14: Let's address this later and merge this PR since so I can work on LIMITED API asynchronously :)

kumaraditya303:
I'll merge this so that @prathamhole14 can work on further limited API changes, the remaining fixups can be done in follow up PRs. Thanks all for reviews.

PR #31994, opened 2026-07-14
BUG: fix segfault in StringDType byteswap()

Yeshwanth-G:
``ndarray.byteswap()`` crashed for StringDType arrays because
  `PyArray_Byteswap` unconditionally called the legacy ``copyswapn`` ArrFuncs
  slot, which is NULL for StringDType -- i.e. a call through a NULL function
  pointer.

  StringDType stores its data in a byte-order independent representation, so
  byteswapping has no meaning. Skip the swap when ``copyswapn`` is NULL,
  making byteswap a no-op for such dtypes instead of crashing.

TODO: Format PR description properly.


jorenham:

PR #31997, opened 2026-07-14
BUG: fix subclass problems introduced in linspace by gh-31620

mhvk:
#31620 made some nice improvements to `np.linspace`'s handling of infinities which unfortunately broke astropy (https://github.com/astropy/astropy/issues/20043). This PR fixes them. Unfortunately, the fix was a little involved, so the diff is more than I had hoped; see inline comments. cc @Ijtihed, who wrote #31620.

It would be good to get this in before the next release...

No AI used.

### DIFF
--- a/numpy/_core/function_base.py
+++ b/numpy/_core/function_base.py
@@ -10,7 +10,7 @@ from numpy._core._multiarray_umath import _array_converter
 from numpy._core.multiarray import add_docstring
 
 from . import numeric as _nx
-from .numeric import asanyarray, nan, ndim, result_type
+from .numeric import asanyarray, nan, result_type
 
 __all__ = ['logspace', 'linspace', 'geomspace']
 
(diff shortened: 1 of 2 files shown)


numpy/_core/tests/test_function_base.py
```diff
@@ -67,6 +67,45 @@ def __rtruediv__(self, x):
 class PhysicalQuantity2(ndarray):
     __array_priority__ = 10
 
+    def __array_ufunc__(self, ufunc, method, *args, **kwargs):
```
mhvk, line 70: This and below are needed to make this mock Quantity subclass more like astropy's `Quantity`, e.g., not keeping subclass for results that are boolean, and requiring output to be the right subclass (the latter is what caused the failure with the previous implementation).

numpy/_core/tests/test_function_base.py
```diff
@@ -393,13 +432,17 @@ def test_physical_quantities(self):
 
     def test_subclass(self):
```
mhvk, line 433: Just expand the test coverage to ensure both scalar and multi-d start/stop paths are covered.

numpy/_core/function_base.py
```diff
@@ -137,29 +137,26 @@ def linspace(start, stop, num=50, endpoint=True, retstep=False, dtype=None,
...
-    if equal.ndim == 0:
-        delta = dt.type(0) if equal else np.subtract(
-            stop, start, dtype=type(dt))
-    else:
-        delta = np.subtract(stop, start, dtype=type(dt), where=~equal,
-                            out=np.zeros(equal.shape, dtype=type(dt)))
-    y = _nx.arange(
+    # Wrapping y ensures that ndarray subclasses like astropy's Quantity
+    # can override the further operations below. See gh-7142.
+    y = conv.wrap(_nx.arange(
```
mhvk, line 148: This is really the main change: go towards a subclass as soon as possible, rather than wrap at the end. In principle, one could also wrap `out` but this felt cleaner (and removed the need for wrapping `y` at the end).

numpy/_core/function_base.py
```diff
@@ -137,29 +137,26 @@ def linspace(start, stop, num=50, endpoint=True, retstep=False, dtype=None,
...
+    ).reshape((-1,) + (1,) * equal.ndim), to_scalar=False)
+    delta = np.zeros_like(y, shape=equal.shape, dtype=type(dt))
+    # Use `dtype=type(dt)` to enforce a floating point evaluation.
+    np.subtract(stop, start, dtype=type(dt), where=~equal, out=delta)
 
-    # In-place multiplication y *= delta/div is faster, but prevents
-    # the multiplicant from overriding what class is produced, and thus
-    # prevents, e.g. use of Quantities, see gh-7142. Hence, we multiply
-    # in place only for standard scalar types.
+    # In-place multiplication y *= delta/div is faster, but cannot work
```
mhvk, line 155: This "new" reason was already a major yet uncommented-on reason for why we needed to keep scalars separate.

mhvk (commented):
Some in-line comments, hopefully helpful for review.

ngoldbaum:
Can you fix the linter failure? @ljthead I'd like you to do a review pass please.

Ijtihed:
LGTM, I have no comments (just lint ig). I don't have much time to go super in depth into this

mhvk:
OK, thanks both! Linter issue should now be fixed (somehow, `spin lint` gave me all problems in the code base, so I gave up; perhaps I should have run it before making a commit or so...)

MaanasArora (approved):
Thanks @mhvk, looks great to me! This definitely fixes the issue - it improves the use of subclass behavior in the routine generally, which is correct for the other operations too. Also, very nice use of `where=`! Merging...

Edit: This should be backported I think clearly? I've added the label, please correct me if I'm wrong.

mhvk:
Thanks! But, no, this should not be backported, since it fixes #31620 (which introduced the nice `~where` trick) and that was only merged recently, so it is not yet in any released version. (The problem was found in astropy's testing on numpy-dev.)

MaanasArora:
Ah sorry, that's right, thanks for the correction!

PR #31998, opened 2026-07-14
ENH: include axis/size detail in tensordot shape-mismatch error

shipitdev:
### PR summary
Right now, when tensordot hits a shape mismatch, it just throws a generic `ValueError("shape-mismatch for sum")`. It doesn't tell you which axes failed or what the actual sizes were, which makes debugging a huge pain. I've seen a bunch of threads on the numpy mailing list and in unrelated downstream projects where users are completely stuck because the error gives them zero context.

This PR just updates the error message to actually include the axis and size details right when the mismatch is detected.

For example, instead of the generic error, it now raises:
`ValueError: shape-mismatch for sum: axis 1 of a (size 4) does not match axis 0 of b (size 5)`

Or if the axis counts themselves are off:
`ValueError: shape-mismatch for sum: a and b have a different number of axes to contract (2 vs 1, from axes=([0, 1], [0]))`

I didn't touch any of the underlying validation logic or behavior. It still sets the exact same flags and raises at the exact same point as before—it just formats a much more helpful string before it does

### AI Use
An LLM was used for research and understanding.

### DIFF
--- a/numpy/_core/numeric.py
+++ b/numpy/_core/numeric.py
@@ -1186,29 +1186,22 @@ def tensordot(a, b, axes=2):
     bs = b.shape
     ndb = b.ndim
     equal = True
-    msg = None
     if na != nb:
         equal = False
-        msg = (
-            f"shape-mismatch for sum: `a` and `b` have a different number "
-            f"of axes to contract ({na} vs {nb}, from axes = {axes!r})"
-        )
     else:
         for k in range(na):
             if as_[axes_a[k]] != bs[axes_b[k]]:
                 equal = False
-                msg = (
-                    f"shape-mismatch for sum: axis {axes_a[k]} of `a` "
-                    f"(size {as_[axes_a[k]]}) does not match axis "
-                    f"{axes_b[k]} of `b` (size {bs[axes_b[k]]})"
-                )
                 break
             if axes_a[k] < 0:
                 axes_a[k] += nda
             if axes_b[k] < 0:
                 axes_b[k] += ndb
     if not equal:
-        raise ValueError(msg)
+        raise ValueError(
+            f"shape-mismatch for tensordot: axes of `a` (shape {as_}) not "
+            f"compatible with axes of `b` (shape {bs})"
+        )
 
     # Move the axes to sum over to the end of "a"
     # and to the front of "b"
--- a/numpy/_core/numeric.py
+++ b/numpy/_core/numeric.py
@@ -1186,19 +1186,29 @@ def tensordot(a, b, axes=2):
     bs = b.shape
     ndb = b.ndim
     equal = True
+    msg = None
     if na != nb:
         equal = False
+        msg = (
+            f"shape-mismatch for sum: `a` and `b` have a different number "
+            f"of axes to contract ({na} vs {nb}, from axes = {axes!r})"
+        )
     else:
         for k in range(na):
             if as_[axes_a[k]] != bs[axes_b[k]]:
                 equal = False
+                msg = (
+                    f"shape-mismatch for sum: axis {axes_a[k]} of `a` "
+                    f"(size {as_[axes_a[k]]}) does not match axis "
+                    f"{axes_b[k]} of `b` (size {bs[axes_b[k]]})"
+                )
                 break
             if axes_a[k] < 0:
                 axes_a[k] += nda
             if axes_b[k] < 0:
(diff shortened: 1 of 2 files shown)


numpy/_core/numeric.py
```diff
@@ -1186,19 +1186,29 @@ def tensordot(a, b, axes=2):
...
+                    f"{axes_b[k]} of `b` (size {bs[axes_b[k]]})"
+                )
                 break
             if axes_a[k] < 0:
                 axes_a[k] += nda
             if axes_b[k] < 0:
                 axes_b[k] += ndb
     if not equal:
-        raise ValueError("shape-mismatch for sum")
+        raise ValueError(msg)
```
tylerjereddy: Although it may seem a little superficial to ask for tests for adjusted failure error messages, in this case it looks like the error path itself isn't flushed by tests (see below from coverage report on this branch when running full testsuite).

You might be able to get a bit more value out of the patch by testing the error path and enforcing the general structure of the error messages while you're adding that testing?

Image

shipitdev: You're right, didn't add tests for that path since this was just an error message update not new logic so I figured the existing coverage on the validation branch itself was enough. But fair, the message content itself isn't actually asserted anywhere so a regression there wouldn't get caught.

I'll add tests for both message branches and push a follow up commit shortly.

numpy/_core/numeric.py
```diff
@@ -1186,19 +1186,29 @@ def tensordot(a, b, axes=2):
     bs = b.shape
     ndb = b.ndim
     equal = True
+    msg = None
     if na != nb:
         equal = False
+        msg = (
+            f"shape-mismatch for sum: `a` and `b` have a different number "
```
eendebakpt: This the message write `tensordot` instead of `sum`?

numpy/_core/numeric.py
```diff
@@ -1186,19 +1186,29 @@ def tensordot(a, b, axes=2):
     bs = b.shape
     ndb = b.ndim
     equal = True
+    msg = None
```
eendebakpt:
```suggestion
```

numpy/_core/numeric.py
```diff
@@ -1186,19 +1186,29 @@ def tensordot(a, b, axes=2):
...
     if na != nb:
         equal = False
+        msg = (
```
eendebakpt: We are adding the `msg` now in two places. We could also add a single message at the point where we raise stating "shape-mismatch: axis {axes_a} not compatible with {axis_b} ..."

numpy/_core/numeric.py
shipitdev: i'd agree, sum i picked up from before this change the phrasing likely came from the internal math that tensordot in implemeted as a sum of products however it doesn't really help the user, i'll rename it

numpy/_core/numeric.py
shipitdev: On the duplication I could go either way: keep pinpointing the exact axis/size pair that failed (just cleaner, one spot) or go simpler something like:

raise ValueError(f"shape-mismatch for tensordot: axes {axes_a} of a not compatible with axes {axes_b} of b")

Shorter, but the user has to spot the mismatch themselves instead of being told directly. for the final call what do you prefer ?

eendebakpt: I think I prefer the short version, just mentioning the input. The axes_a and axes_b are modified inplace, so I would keep them out.

shipitdev:
Pushed. i've Renamed sum -> tensordot, and cut into a single message that references a/b's shapes instead of axes_a/axes_b. Updated the existing tests to match the new message too.

numpy/_core/numeric.py
```diff
@@ -1198,7 +1198,10 @@ def tensordot(a, b, axes=2):
             if axes_b[k] < 0:
                 axes_b[k] += ndb
     if not equal:
-        raise ValueError("shape-mismatch for sum")
+        raise ValueError(
```
Sanjays2402, line 1201: this fires for two different failures: unequal axis lengths, and na != nb (mismatched number of contracted axes, the test_axes_count_mismatch case). for the count mismatch the shapes aren't really the useful info, the axes lists are. and even in the length-mismatch case you print full shapes but never say which axis pair didn't line up, which is the thing the user actually needs. worth including axes_a/axes_b?

Issue #15317, opened 2020-01-12
Clean up internal callers of PyUnicode_AsASCIIString


eric-wieser:
Many of the C functions in numpy have code that is something like:
```C
PyObject *foo(PyObject *obj) {
    if (PyUnicode_Check(obj)) {
        /* accept unicode input */
        PyObject *obj_bytes = PyUnicode_AsASCIIString(obj);
        if (obj_bytes == NULL) {
             return NULL;
        }
        PyObject *ret = foo(obj_bytes);
        Py_DECREF(obj_bytes);
    }
    
    char *str = NULL;
    Py_ssize_t length = 0;
    if (PyBytes_AsStringAndSize(obj, &str, &length) < 0) {
        return NULL;
    }

    // work with bytes
}
```
This construction means they raise `UnicodeEncodeError` for some illegal inputs, which is often less useful than the message would normally be

It would be better to work with the unicode objects directly, by copying the approach taken in the C code changes in #15261, notably [these lines](https://github.com/numpy/numpy/commit/e83bd4658de5d7ffb6a460f68d2638c1b0a35068#diff-475dc2a28ab28625b0f02d7f006683abL1411-R1419) and [these lines](https://github.com/numpy/numpy/commit/e83bd4658de5d7ffb6a460f68d2638c1b0a35068#diff-475dc2a28ab28625b0f02d7f006683abL1497-R1502).

---

Spawned from discussion in https://github.com/numpy/numpy/pull/15261#issuecomment-573367589

PR #32000, opened 2026-07-15
MAINT: parse nditer flags as unicode directly (gh-15317)

nakasan617:
Part of gh-15317 (cleaning up internal callers of `PyUnicode_AsASCIIString`).

`NpyIter_GlobalFlagsConverter` and `NpyIter_OpFlagsConverter` encoded unicode flag strings to bytes with `PyUnicode_AsASCIIString` before inspecting them. For any non-ASCII input that raises a `UnicodeEncodeError`, which is less useful than numpy's own "unexpected flag" error.

This works with the unicode object directly via `PyUnicode_AsUTF8AndSize` (matching the idiom already used elsewhere in `multiarray`), so a non-ASCII flag now reports a `ValueError` naming the offending flag instead:

```python
>>> import numpy as np
>>> np.nditer(np.arange(3), flags=['\N{SNOWMAN}'])
ValueError: Unexpected iterator global flag "☃"
```

Kept intentionally small — this is one file / two twin converters. The remaining `PyUnicode_AsASCIIString` call sites from gh-15317 can follow as separate PRs (some, e.g. the `S`-dtype setitem and `tofile` paths, deserve their own discussion since ASCII encoding is arguably intentional there).

### Notes
- Behavior for valid ASCII/bytes flags is unchanged; only the error for non-ASCII input improves.
- Regression tests added: non-ASCII and non-string flags in `test_iter_flags_errors`, plus `test_iter_bytes_flags` pinning that bytes flags remain accepted (that branch was previously untested).

Draft while CI builds/runs the suite.

### DIFF
--- a/numpy/_core/src/multiarray/nditer_pywrap.c
+++ b/numpy/_core/src/multiarray/nditer_pywrap.c
@@ -112,7 +112,7 @@ NpyIter_GlobalFlagsConverter(PyObject *flags_in, npy_uint32 *flags)
     int iflags, nflags;
 
     PyObject *f;
-    char *str = NULL;
+    const char *str = NULL;
     Py_ssize_t length = 0;
     npy_uint32 flag;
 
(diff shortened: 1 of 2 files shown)


ngoldbaum:
I'm closing this pull request because you didn't read the PR template.

nakasan617:
sorry. I read and understood. will create one by myself shortly.

PR #32001, opened 2026-07-15
DOC: add Examples section to np.ma.view [skip actions]

shoumikchakravarty-dev:
### PR summary

MaskedArray.view in numpy/ma/core.py has Parameters, See Also, and Notes sections but no Examples. This PR adds three illustrative doctests showing:

1. The default view of a masked array (same dtype and mask preserved)
2. Using the fill_value parameter - unique to MaskedArray.view vs ndarray.view -  to set a custom fill value on the view without copying data
3. Viewing as a plain numpy.ndarray, which does not preserve the mask

Not a first-time contributor

#### AI Disclosure
Claude (Anthropic) was used to assist in drafting the docstring examples and verifying expected output values. All examples were reviewed and confirmed correct by running them manually. The final content and PR submission are my own work.

### DIFF
--- a/numpy/ma/core.py
+++ b/numpy/ma/core.py
@@ -3254,6 +3254,28 @@ class MaskedArray(ndarray):
         memory. Therefore if ``a`` is C-ordered versus fortran-ordered, versus
         defined as a slice or transpose, etc., the view may give different
         results.
+
+        Examples
+        --------
+        >>> import numpy as np
+        >>> a = np.ma.array([1.0, 2.0, 3.0], mask=[0, 1, 0])
+        >>> a
+        masked_array(data=[1.0, --, 3.0],
+                     mask=[False,  True, False],
+               fill_value=1e+20)
+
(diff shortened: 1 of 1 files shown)


mattip:
Not everything needs examples. What did you find surprising that warranted adding them?

Also: `[skip ci]` skips all ci. Use `[skip actions]` for documentation PRs to skip github actions while still building and testing the documentation on CircleCI.

shoumikchakravarty-dev:
On the examples, the two things I found non-obvious were 
(1) the fill_value parameter, it's unique to MaskedArray.view with no equivalent on ndarray.view, so readers coming from ndarray may not realize they can change the fill value in-place on the view; and
(2) that view(np.ndarray) silently drops the mask and exposes the underlying data, which is easy to stumble into accidentally.

I am happy to drop them if the preference is to keep the docstring minimal. Please advise. 

I have also updated for the [skip actions] instead of the [skip ci]

mattip:
Thanks @shoumikchakravarty-dev. Makes sense.

PR #32003, opened 2026-07-15
TYP: specialized ``fmod`` ufunc

jorenham:
Unlike the binary bitwise ufuncs from #31991 this one upcasts `bool` to `int8`, and also accepts `floating` dtypes.

In case anyone is counting; this one requires 50 overloads for just `__call__` (I'm trying to keep it under 64).

<details>
<summary>Promotion table:</summary>

```
      b8   i8   u8  i16  u16  i32  u32  i64  u64  f16  f32  f64  f80
b8    i8   i8   u8  i16  u16  i32  u32  i64  u64  f16  f32  f64  f80  
i8    i8   i8  i16  i16  i32  i32  i64  i64  f64  f16  f32  f64  f80  
u8    u8  i16   u8  i16  u16  i32  u32  i64  u64  f16  f32  f64  f80  
i16  i16  i16  i16  i16  i32  i32  i64  i64  f64  f32  f32  f64  f80  
u16  u16  i32  u16  i32  u16  i32  u32  i64  u64  f32  f32  f64  f80  
i32  i32  i32  i32  i32  i32  i32  i64  i64  f64  f64  f64  f64  f80  
u32  u32  i64  u32  i64  u32  i64  u32  i64  u64  f64  f64  f64  f80  
i64  i64  i64  i64  i64  i64  i64  i64  i64  f64  f64  f64  f64  f80  
u64  u64  f64  u64  f64  u64  f64  u64  f64  u64  f64  f64  f64  f80  
f16  f16  f16  f16  f32  f32  f64  f64  f64  f64  f16  f32  f64  f80  
f32  f32  f32  f32  f32  f32  f64  f64  f64  f64  f32  f32  f64  f80  
f64  f64  f64  f64  f64  f64  f64  f64  f64  f64  f64  f64  f64  f80  
f80  f80  f80  f80  f80  f80  f80  f80  f80  f80  f80  f80  f80  f80  
```

</details>

---

No AI.


numpy/_core/umath.pyi
```diff
@@ -9463,6 +9462,1467 @@ class _ufunc_21_bio(_ufunc_21[_IdT_co], Generic[_IdT_co, _ScalarT_contra]):  # t
...
     ) -> OutT: ...
 
+# bBhHiIlLqQefdgO, bBhHiIlLqQefdgO => bBhHiIlLqQefdgO
+#
+# Object dtypes aren't included here because this simply calls `.fmod(_)` on the
+# underlying Python object, which in practice will usually lead to a `TypeError`.
+#
+# Only the i64/i32/i8/u8 int promotion rules are implemented, building upon `_ufunc_21_bio`.
+@type_check_only
+class _ufunc_21_fmod(_ufunc_21[None]):  # type: ignore[misc]
```
eendebakpt, line 9472: Could we parameterize or generate the overloaded calls from a file?

jorenham: Parametrization can help in only some specific cases (we're using it for e.g. `bitwise_{and,or,xor}` and `lcm`/`gcd` to "toggle" the boolean overloads).

I've indeed thought about generating specialized `ufunc` types from the ufunc signatures, but I'm not sure if that'll help much, because there are often differences between ufuncs that don't really fit a common pattern. For example, `gcd(True, True)` will raise a `UFuncTypeError`, but `fmod(True, True)` returns `np.int8(0)`, and there's no way tell by looking at their `ufunc.types` which one will happen.
And for the more complicated ufuncs (like `fmod`), not all dtype promotion combos have overloads, because that would lead too 100+ overloads. This can become a big performance bottleneck for type-checkers.
Then there's also all of the awkward logistics that comes with code generation, which in my experience is a big price to pay.

Some hybrid code-gen approach might work though, using a DSL that transpiles to `pyi` stubs and supports something like type-macro's or re-usable type-mappings (something like per-parameter overloads so that we can avoid combinatorial explosions). I've given this a lot of thought, and realized that this will be **a lot** of work do do right (because there's also IDE support, mapping type-checker errors using source maps, performance), on top of also having to deal with the awkard codegen logistics. I might give this a shot one day, but I'll probably go all the way then, and write a "typescript for python (stubs)" with builtin typechecker, basically rewriting Python's type system from scratch without all of its design flaws. That'll be a multi-year project that could be difficult to get funding for though...

charris:
Thanks Joren.

PR #32011, opened 2026-07-15
CI: Fix PR labeler

scottshambaugh:
### PR summary
https://github.com/numpy/numpy/pull/30935 broke the PR labeler by trying to import `js-yaml` which is not in the include scope for this action. `continue-on-error` was set to true, so this was silently breaking.

Rather than add another dependency, this does the necessary parsing in a line of regex.

I also did a quick look at all the other jobs with `continue-on-error = true` and none look to be failing persistently.

#### AI Disclosure
Claude used to draft the regex, manually reviewed.


.github/workflows/labeler.yml
```diff
@@ -14,18 +14,26 @@ jobs:
...
+          const prefixToLabel = {};
+          for (const line of content.split('\n')) {
+            const match = line.match(/^\s*"([^"]+)"\s*:\s*"([^"]+)"\s*$/);
+            if (match) {
+              prefixToLabel[match[1]] = match[2];
+            }
+          }
+          if (Object.keys(prefixToLabel).length === 0) {
+            throw new Error('No entries parsed from pr-prefix-labeler.yml');
+          }
```
jorenham: or we just use json 😛

scottshambaugh: That's cleaner, switched! I don't see anywhere else that uses that datafile

charris:
I honestly can't tell if this is working or not, there is still a `js.index` error, and I have no idea were it is coming from, or if it is valid. So I am going to put this in with the proviso that it may not be working.

PR #32012, opened 2026-07-16
BUG: fix `np.fromiter` corruption when reusing a `StringDType` instance 

### LINKED ISSUE #32013: BUG: `np.fromiter` creates corrupted arrays when re-using StringDType instance
### Describe the issue:

When re-using the StringDType instance, `np.fromiter` creates corrupted arrays leading to flaky failures.

```python
Reproducer:

>>> np.__version__
'2.5.0'

>>> sd = np.dtypes.StringDType()

>>> a = ["a"*18, "b"*18]*10_000

>>> np.fromiter(iter(a), dtype=sd, count=len(a))
array(['aaaaaaaaaaaaaaaaaa', 'bbbbbbbbbbbbbbbbbb', 'aaaaaaaaaaaaaaaaaa',
       ..., 'bbbbbbbbbbbbbbbbbb', 'aaaaaaaaaaaaaaaaaa',
       'bbbbbbbbbbbbbbbbbb'], shape=(20000,), dtype=StringDType())

>>> np.fromiter(iter(a), dtype=sd, count=len(a)) # same snippet ran for the second time segfaults.
```

>>> import sys, numpy; print(numpy.__version__); print(sys.version)
2.5.0
3.14.4

Yeshwanth-G:
### PR summary

Fixes https://github.com/numpy/numpy/issues/32013

`np.fromiter` is buggy for StringDtype arrays. Specifically, when the dtype instance is already owned by someother array,
it creates corrupted structures since it re-uses the input dtype instance to create the output array.

This PR fixes this issue to avoid using the input dtype descriptor to write the output.

Reproducer which this PR fixes:

```
In [2]: sd = np.dtypes.StringDType()

In [3]: a = ["a"*18, "b"*18]*10_000

In [4]: np.fromiter(iter(a), dtype=sd, count=len(a))
Out[4]: 
array(['aaaaaaaaaaaaaaaaaa', 'bbbbbbbbbbbbbbbbbb', 'aaaaaaaaaaaaaaaaaa',
       ..., 'bbbbbbbbbbbbbbbbbb', 'aaaaaaaaaaaaaaaaaa',
       'bbbbbbbbbbbbbbbbbb'], shape=(20000,), dtype=StringDType())

In [5]: np.fromiter(iter(a), dtype=sd, count=len(a))
Out[5]: Fatal Python error: Segmentation fault
```

#### AI Disclosure

I’ve used Claude Opus to shepherd this along.

### DIFF
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -4000,6 +4000,19 @@ PyArray_FromIter(PyObject *obj, PyArray_Descr *dtype, npy_intp count)
     PyArray_Dims new_dims = {dims, PyArray_NDIM(ret)};
 
     char *item = PyArray_BYTES(ret);
+    /*
+     * PyArray_NewFromDescr may hand back an array whose descriptor is not
+     * `dtype`.  For subarray dtypes it uses the base descriptor, and packing
+     * must keep using the subarray `dtype` itself.  For parametric dtypes such
+     * as StringDType it may return an equivalent but distinct descriptor that
+     * owns its own storage (the string arena) -- this happens when the passed
+     * descriptor is already owned by another array, e.g. when the same dtype
+     * instance is reused.  In that case values must be packed with the array's
+     * actual descriptor, otherwise they are written into the wrong arena and
+     * the memory is freed twice on cleanup.
+     */
+    PyArray_Descr *pack_descr = PyDataType_HASSUBARRAY(dtype) ?
+                                dtype : PyArray_DESCR(ret);
     for (i = 0; i < count || count == -1; i++, item += elsize) {
         PyObject *value = PyIter_Next(iter);
         if (value == NULL) {
@@ -4028,7 +4041,7 @@ PyArray_FromIter(PyObject *obj, PyArray_Descr *dtype, npy_intp count)
             item = ((char *)PyArray_DATA(ret)) + i * elsize;
         }
 
-        if (PyArray_Pack(dtype, item, value) < 0) {
+        if (PyArray_Pack(pack_descr, item, value) < 0) {
             Py_DECREF(value);
             goto fail;
         }
--- a/numpy/_core/src/multiarray/ctors.c
+++ b/numpy/_core/src/multiarray/ctors.c
@@ -3984,11 +3984,7 @@ PyArray_FromIter(PyObject *obj, PyArray_Descr *dtype, npy_intp count)
 
     elsize = dtype->elsize;
 
-    /*
-     * Note that PyArray_DESCR(ret) may not match dtype.  There are exactly
-     * two cases where this can happen: empty strings/bytes/void (rejected
-     * above) and subarray dtypes (supported by sticking with `dtype`).
-     */
+
     Py_INCREF(dtype);
     ret = (PyArrayObject *)PyArray_NewFromDescr(&PyArray_Type, dtype, 1,
                                                 &elcount, NULL,NULL, 0, NULL);
@@ -4000,6 +3996,10 @@ PyArray_FromIter(PyObject *obj, PyArray_Descr *dtype, npy_intp count)
     PyArray_Dims new_dims = {dims, PyArray_NDIM(ret)};
 
     char *item = PyArray_BYTES(ret);
+
+    /*
+     * Note that PyArray_DESCR(ret) may not match dtype.
+     */
     PyArray_Descr *pack_descr = PyDataType_HASSUBARRAY(dtype) ? dtype : PyArray_DESCR(ret);
 
     for (i = 0; i < count || count == -1; i++, item += elsize) {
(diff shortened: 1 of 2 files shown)


numpy/_core/src/multiarray/ctors.c
```diff
@@ -4000,6 +4000,18 @@ PyArray_FromIter(PyObject *obj, PyArray_Descr *dtype, npy_intp count)
...
+    /*
+     * PyArray_NewFromDescr may hand back an array whose descriptor is not
+     * `dtype`.  For subarray dtypes we use the base descriptor `dtype` for packing.
+     * For parametric dtypes such as StringDType it may return an equivalent but
+     * distinct descriptor that owns its own storage (the string arena) -- this
+     * happens when the passed descriptor is already owned by another array,
+     * e.g. when the same dtype instance is reused.  In that case values must be
+     * packed with the array's actual descriptor, otherwise they are written into
+     * the wrong arena and the memory is freed twice on cleanup.
+     */
```
ngoldbaum, line 4002: IMO the code is self-documenting as you've updated it and this comment is unnecessary. Claude likes to insert unnecessarily long comments.

numpy/_core/src/multiarray/ctors.c
ngoldbaum: I think this comment should actually be on the line you're changing. And also you're fixing the issue it's talking about. Can you move this comment down to the line you're changing and delete "There are exactly two cases..." and everything after.

numpy/_core/tests/test_stringdtype.py
```diff
@@ -724,6 +724,17 @@ def test_fancy_indexing(string_list):
         assert_array_equal(sarr[ind], uarr[ind])
 
 
+def test_fromiter_reused_dtype():
+    # Reusing the same StringDType instance across fromiter
+    # calls used to write strings into the wrong arena create corrupted arrays.
```
ngoldbaum:
```suggestion
    # calls used to write strings into the wrong arena, creating corrupted arrays
```

ngoldbaum (approved):
LGTM, just two nits inline.

numpy/_core/src/multiarray/ctors.c
```diff
@@ -4000,6 +4000,8 @@ PyArray_FromIter(PyObject *obj, PyArray_Descr *dtype, npy_intp count)
     PyArray_Dims new_dims = {dims, PyArray_NDIM(ret)};
 
     char *item = PyArray_BYTES(ret);
+    PyArray_Descr *pack_descr = PyDataType_HASSUBARRAY(dtype) ? dtype : PyArray_DESCR(ret);
```
ngoldbaum, line 4003: Ping @seberg can you also take a look at this? You know a lot more about structured dtypes than I do.

seberg: Yeah, should make sense... But if we ever try to put a string dtype into a subarray one we'll get bitten.

I wish we'd just never allowed subarray dtypes here now :(

So I think this is OK, I could imagine that we could:
* create the array with `_NPY_ARRAY_ENSURE_DTYPE_IDENTITY`
* Then re-create it (basically) to get rid of the subarray part.

But while I guess we may need to do something like that when we allow nesting StringDTypes into structured dtypes (dtypes with a finalizer), until then... Yes, I think it is fine.

ngoldbaum: Hmm, well, we *do* actually support putting StringDType into a subarray:

```
>>> data = [("x" * 18, "y" * 18)] * 20
... dt = np.dtype((np.dtypes.StringDType(), 2))
... f1 = np.fromiter(iter(data), dtype=dt, count=len(data))
... f2 = np.fromiter(iter(data), dtype=dt, count=len(data))
... print(f2[0])
...
MemoryError: Failed to load string in StringDType getitem
```

ngoldbaum: I threw an AI model at this and it came up with this solution, on top of this PR branch:

https://gist.github.com/ngoldbaum/a74a9a8eaf9318626231194086ce5532

seberg: Hummm, wait... I would have sworn that:
```
a = np.zeros(10, [("a", (np.dtypes.StringDType(), 10))])
```
this isn't supported (and I think it isn't truly?).  For normal arrays what you have is fine, but yeah, this fix wouldn't actually help for your example...

I have to think about it, but as much as I don't like almost fixes, merging this may still be the right option.

Also because the easiest solution right now is likely to make that dtype construction fail.

ngoldbaum: Argh, you're right, we have this:

https://github.com/numpy/numpy/blob/f67f65a1ab0d0998896703d6f205912d6975f966/numpy/_core/src/multiarray/descriptor.c#L498-L502

But that doesn't catch descriptors like yours. So really this isn't supported at all. So this PR is the correct fix for now and then in the future if we ever want to *actually* worry about structured dtypes we can worry about it then.

@Yeshwanth-G sorry for that disgression - are you OK with updating your PR to address my original comments?

ngoldbaum (approved):
Thanks!

numpy/_core/src/multiarray/ctors.c
ngoldbaum: I opened https://github.com/numpy/numpy/pull/32027 as a followup.

Yeshwanth-G:
Thanks for the reviews here @ngoldbaum , @seberg !

PR #31928, opened 2026-07-09
ENH: Implement per-module state for lapack_lite to avoid global static states

prathamhole14:
### PR summary

Related to: https://github.com/numpy/numpy/issues/31930 and https://github.com/numpy/numpy/issues/32451

#### AI Disclosure

Only for some research and improving comments.

PR #32015, opened 2026-07-16
DEP: Deprecate `numpy.linalg.lapack_lite` module

prathamhole14:
### PR summary
As titled, based on: https://github.com/numpy/numpy/pull/31928#issuecomment-4989512472

Note: 
If a user simply `import numpy.linalg.lapack_lite` it will not give warning, it will warn only if used in test (run via pytest) or with `-W` flag with python3/python. 
 
CC: @seberg @mattip 

#### AI Disclosure

To write release note - it LGTM but someone please check it again. And to find the best way to write test for module dep.

### DIFF
--- a/numpy/linalg/lapack_lite.pyi
+++ b/numpy/linalg/lapack_lite.pyi
@@ -60,7 +60,12 @@ _ilp64: Final[bool] = ...
 
 class LapackError(Exception): ...
 
-@deprecated("The numpy.linalg.lapack_lite module is deprecated and will be removed in a future release. It is an internal implementation detail of numpy.linalg. Users that need LAPACK or BLAS routines should use scipy.linalg.lapack or scipy.linalg.blas instead.")
+@deprecated(
+    "The numpy.linalg.lapack_lite module is deprecated and will be removed in a future release. "
+    "It is an internal implementation detail of numpy.linalg. Users that need LAPACK or BLAS "
+    "routines should use scipy.linalg.lapack or scipy.linalg.blas instead. For users checking "
+    "BLAS/LAPACK library paths or configuration, use numpy.show_config() instead."
+)
 def dgelsd(
     m: int,
     n: int,
@@ -78,7 +83,12 @@ def dgelsd(
     info: int,
 ) -> _DGELSD: ...
 
-@deprecated("The numpy.linalg.lapack_lite module is deprecated and will be removed in a future release. It is an internal implementation detail of numpy.linalg. Users that need LAPACK or BLAS routines should use scipy.linalg.lapack or scipy.linalg.blas instead.")
+@deprecated(
+    "The numpy.linalg.lapack_lite module is deprecated and will be removed in a future release. "
+    "It is an internal implementation detail of numpy.linalg. Users that need LAPACK or BLAS "
+    "routines should use scipy.linalg.lapack or scipy.linalg.blas instead. For users checking "
+    "BLAS/LAPACK library paths or configuration, use numpy.show_config() instead."
+)
 def zgelsd(
     m: int,
     n: int,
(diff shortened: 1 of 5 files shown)


jorenham (commented):
it would be nice if we could slap some `@typing_extensions.deprecated` onto the functions in https://github.com/numpy/numpy/blob/cda64f0cf4584aa9e735d46ba27874296501333a/numpy/linalg/lapack_lite.pyii . That way static type-checkers will warn you if you use it at type-check-time.

prathamhole14:
Thanks for pointing that out I was unaware of `@typing_extensions.deprecated`.

jorenham (approved):
stub changes look good; thanks 👍

numpy/linalg/lapack_lite.pyi
```diff
@@ -127,6 +135,8 @@ def dorgqr(
     lwork: int,
     info: int,  # out
 ) -> _DORGQR: ...
+
+@deprecated("The numpy.linalg.lapack_lite module is deprecated and will be removed in a future release. It is an internal implementation detail of numpy.linalg.")
```
ngoldbaum: Can these deprecation warnings suggest what users should do instead?

See https://github.com/search?q=%22import+lapack_lite%22&type=code for a large number of real-world users of the `lapack_lite` module. Do any of the uses in my search hit these new warnings? What should they do when they see the warning.

mattip:
If there are people using this module for real code, then the whole premise of this issue and PR are less correct ( i.e. wrong) and we shouldn’t deprecate it.

numpy/linalg/lapack_lite.pyi
jorenham: If you filter out all of numpy's own sources in site-packages and stuff, then it's really not that much (152 hits):

https://github.com/search?q=%22import+lapack_lite%22++NOT+is%3Aarchived+NOT+path%3A%2Fnumpy%5C%2Flinalg%5C%2F%2F&type=code

seberg: Even there a lot seem just the numpy source.  It's a bit unclear I looked at a few:
* 2 were importing scipy, so I think they can use the scipy lapack exposure if they really want the "low-level" feel?
* one seemed to pull it in to load math so's but loading `numpy.linalg` should do that, but not sure this one is weird...
* one actually *patched* `numpy.linalg.lapack_lite` rather than importing it.
* One was some super outdated chunk of code, I wouldn't worry about.

Could scan more, but I would think it's fine dozens of users and it seems unlikely many don't have better alternatives (i.e. scipy).

That said, pointing to scipy for it's low-level lapack exposure probably makes sense?

numpy/linalg/tests/test_deprecations.py
```diff
@@ -17,3 +21,16 @@ def test_qr_mode_full_removed():
         np.linalg.qr(a, mode='economic')
     with pytest.raises(ValueError, match="deprecated in NumPy 1.8"):
         np.linalg.qr(a, mode='e')
+
+
+@pytest.mark.skipif(IS_WASM, reason="Cannot start subprocess")
+def test_lapack_lite_deprecation():
```
rgommers: This seems like a bit much, the other test should cover it. Subprocesses are slow, and don't work on multiple platforms - IS_WASM conflicts with the just merged Android support.

rgommers (commented):
Deprecating LGTM, this isn't meant to be public. One test should be removed before merging.

ngoldbaum (commented):
I asked an AI model to search for open-source uses of this module and it found two cases that aren't covered by the advice here:

* astra-toolbox/astra-toolbox: imports numpy.linalg.lapack_lite from C++ purely as a RTLD_DEEPBIND side-effect to stop MATLAB's MKL from clashing with numpy's LAPACK. It calls no routine at all.
* BLAS/Fortran introspection recipes — e.g. simomarsili/ndd uses `numpy.linalg.lapack_lite.__file__` to locate the .so and run ldd on it; the _ilp64 attribute is used elsewhere to detect 64-bit BLAS.

The former could use `numpy.linalg._umath_linalg` instead. I'm not sure if we need to explicitly document that though because it's an unusual thing to do. Maybe one of us could send in a PR to astra-toolbox and reference this PR?

The latter could use `np.show_config()` instead. We should probably explicitly suggest `np.show_confg()` in the deprecation warning for people who are using the module for introspection.

ngoldbaum (approved):

mattip:
Thanks @prathamhole14

PR #32021, opened 2026-07-16
TYP/CI: Fix ruff error in `_core.umath` stubs

jorenham:
The lint CI job is failing on main because of a ruff error in the stubs (mea culpa). Not sure why this wasn't caught before...

```console
$ spin lint
Running Ruff Check...
error[E222][*]: Multiple spaces after operator
    --> numpy/_core/umath.pyi:9352:21
     |
9350 |         *,
9351 |         axis: int = 0,
9352 |         dtype: str |  type,
     |                     ^^
9353 |         out: None = None,
9354 |     ) -> npt.NDArray[Any]: ...
     |
help: Replace with single space
9349 |         indices: tuple[int, ...],
9350 |         *,
9351 |         axis: int = 0,
     -         dtype: str |  type,
9352 +         dtype: str | type,
9353 |         out: None = None,
9354 |     ) -> npt.NDArray[Any]: ...
9355 |     @overload  # out=<given>

Found 1 error.
[*] 1 fixable with the `--fix` option.
```

---

No AI.

PR #32022, opened 2026-07-16
CI: Move the lint job to separate workflow

jorenham:
This way it's easier to also have the linter run on `.pyi`-only changes, so that we can avoid oopsies like #32021.

---

No AI.


charris (approved):

charris:
Thanks Joren.

PR #32024, opened 2026-07-16
MAINT/TYP: Bump ``mypy`` to ``2.3.0``

jorenham:
Apparently there was a regression in mypy 2.2.0 that causes it to ignore the type parameter defaults of generic types in certain cases, so I had to adjust the type-tests as a workaround.

Changelog: https://mypy.readthedocs.io/en/stable/changelog.html

---

No AI.


charris (approved):

charris:
Thanks Joren.

PR #32025, opened 2026-07-16
BUG: fix my fix of linspace for subclasses. Arguably simpler.

mhvk:
This is painful, but it turns out my fix in #31997 actually did not fix the astropy failures in https://github.com/astropy/astropy/issues/20043 -- or, rather, they changed the failure from trying to write quantities in a numpy array to trying to add incompatible units. Of course, I *had* tested my fix, but decided afterwards I could make it even nicer and did not test that "nicer" version. So, here is the slightly less nice one that does fix astropy and does also pass the new tests I introduced. Unfortunately, I'll really have to add units to the `PhysicalQuantity2` test class if I want to be able to distinguish between the two versions. But maybe this is good enough...

ping @MaanasArora

### DIFF
--- a/numpy/_core/function_base.py
+++ b/numpy/_core/function_base.py
@@ -143,20 +143,25 @@ def linspace(start, stop, num=50, endpoint=True, retstep=False, dtype=None,
     # bad element entirely rather than suppressing warnings globally, so
     # genuine invalid operations (e.g. mixed infinities) still warn.
     equal = start == stop
-    # Wrapping y ensures that ndarray subclasses like astropy's Quantity
-    # can override the further operations below. See gh-7142.
-    y = conv.wrap(_nx.arange(
-        0, num, dtype=dt, device=device
-    ).reshape((-1,) + (1,) * equal.ndim), to_scalar=False)
-    delta = np.zeros_like(y, shape=equal.shape, dtype=type(dt))
+    # Wrapping delta ensures that ndarray subclasses like astropy's Quantity
+    # can override the subtraction correctly below. See gh-7142.
+    delta = conv.wrap(
+        np.zeros(shape=equal.shape, dtype=type(dt)),
+        to_scalar=False
+    )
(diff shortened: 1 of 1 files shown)


numpy/_core/function_base.py
```diff
@@ -175,7 +180,7 @@ def linspace(start, stop, num=50, endpoint=True, retstep=False, dtype=None,
         # sequences with 0 items or 1 item with endpoint=True (i.e. div <= 0)
         # have an undefined step
         step = nan
-        # Multiply with delta out-of-place in case delta is not scalar.
+        # Multiply out-of-place in case delta is not scalar or a subclass.
```
MaanasArora, line 183: Is it a reasonable contract that `y * delta`, i.e. the (non-subclass) ndarray times a subclass, should always produce the (same) subclass? I suppose it should work here for astropy, but wondering how general it is... just a question, since we no longer wrap `y` itself.

MaanasArora (approved):
Thanks for the fix, looks good to me, just one clarifying question! This is indeed a bit tricky to fully test in NumPy...

numpy/_core/function_base.py
mhvk: I think the idea was that here the subclass has a *choice* to make `y` a subclass, while in-place it does not have that choice. So, effectively we allow the subclass to set itself up for the later `y += start`.

I guess in principle we could add another test where a subclass does not make `y` a subclass and that deals with in-place addtion, but since with this change we are effectively back to where we were before #31620, I think it is OK...

MaanasArora: OK makes sense! At least there is a choice for either case. You can always break things if you really try anyway :) Thanks!

MaanasArora:
Right, even the previous PR simply moved the wrapping a few operations up, I can't see an issue for well-behaved subclasses. I think this should be fine to go in. Thanks @mhvk!

PR #32027, opened 2026-07-17
BUG: reject structured DTypes that contain StringDType fields

ngoldbaum:
### PR summary

Currently we reject structured dtypes that contain StringDType in only one code path:

https://github.com/numpy/numpy/blob/f67f65a1ab0d0998896703d6f205912d6975f966/numpy/_core/src/multiarray/descriptor.c#L498-L502

This rejects dtypes like `np.dtype([("a", StringDType())])` but lets through dtypes like `[("a", (T, 10))]` or `"T,i4"`.

This PR closes those holes.

There's also a reference leak in the existing error path, which this fixes.

### Open questions

Unfortunately people may be depending on the old, broken support. I'm not sure if closing these holes should trigger deprecation warnings?

Alternatively we could try to actually support StringDType subarrays, https://github.com/numpy/numpy/pull/32012#discussion_r3601337804 might have a path towards that.

#### AI Disclosure
I used an AI model to identify the code paths that need new errors. It also spotted the reference leak.

### DIFF
--- a/numpy/_core/src/multiarray/descriptor.c
+++ b/numpy/_core/src/multiarray/descriptor.c
@@ -308,6 +308,18 @@ _convert_from_tuple(PyObject *obj, int align)
             return type;
         }
 
+        /*
+         * A subarray dtype is never attached to an array, so a base with
+         * per-instance state (a finalize slot, e.g. StringDType) could
+         * never be finalized and anything using the dtype would misbehave.
+         */
+        if (NPY_DT_SLOTS(NPY_DTYPE(type))->finalize_descr != NULL) {
+            PyErr_Format(PyExc_TypeError,
+                    "%s is not currently supported within subarray dtypes.",
+                    ((PyTypeObject *)NPY_DTYPE(type))->tp_name);
+            goto fail;
+        }
+
         /* validate and set shape */
         for (int i=0; i < shape.len; i++) {
             if (shape.ptr[i] < 0) {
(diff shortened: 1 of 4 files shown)


numpy/_core/src/multiarray/descriptor.c
```diff
@@ -388,6 +388,27 @@ _convert_from_tuple(PyObject *obj, int align)
...
+ * an exception set if rejected, 0 otherwise.
+ */
+static int
+_reject_stringdtype_field(PyArray_Descr *descr)
+{
+    while (PyDataType_HASSUBARRAY(descr)) {
+        descr = PyDataType_SUBARRAY(descr)->base;
+    }
+    if (PyObject_IsInstance((PyObject *)descr,
+                            (PyObject *)&PyArray_StringDType)) {
```
seberg: Thanks for the quick fix! Rather than this, can you check for it defining the "finalizer" slot?  That is (in principle) more generic, even if string dtype is the only one right now.
(But maybe this is important for BytesDType :) for example).

> Unfortunately people may be depending on the old, broken support. I'm not sure if closing these holes should trigger deprecation warnings?

Hmmmpf, yeah... I would prefer to just disable it and hope for the best, but can go either way.

Another thing that would make this simpler/nicer is if we also reject subarray dtype creation like this.  Then we don't have to iterate bases at all here.
(Honestly, maybe you already do here, but then we don't need to iterate.)

(It's not like subarrays work, they only "work", because you can't actually attach them to an array in practice...)

mhvk:
Not sure I have much to add, though in the end it would be nice to just support StringDType in structured arrays, just like we support object.

seberg (approved):
Thanks, for fixing this and converting to check for `finalize_descr`! looking I wonder if we could put this check to later to avoid duplicating, but with a helper it doesn't matter much.

Good to get this fixed!

PR #32029, opened 2026-07-17
ENH: Support DLPack for scalars by exporting read-only data

### LINKED ISSUE #27137: BUG: array api incompatibility in output of reductions (sum/prod/min/max/...)

MaanasArora:
Resolves #27137 by adding DLPack support to scalars. Unlike #31893, this does not convert to a dummy 0-D array, instead as suggested by @seberg, exposes a read-only reference to the scalar's underlying data (the same function used for the buffer protocol). As a result only `copy=False` is allowed, the shared memory is read-only and signaled as such, and therefore DLPack >=1 is supported.

I'm going to recheck some of the trickier things around argument parsing / defaults, but this should be mostly ready for review. Thanks!

#### AI Disclosure
I used an agent to find correctness issues and suggest missing tests.

### DIFF
--- a/doc/release/upcoming_changes/32029.new_feature.rst
+++ b/doc/release/upcoming_changes/32029.new_feature.rst
@@ -1,0 +1,6 @@
+Add DLPack support to NumPy scalars
+-----------------------------------
+``scalar.__dlpack__`` and ``scalar.__dlpack_device__`` methods have been added
+to NumPy scalars to allow exporting them to DLPack, similarly to ndarrays.
+Scalars can only be exported with read-only access (``__dlpack__(copy=False)``)
+and to DLPack version 1.0 or higher.
--- a/doc/release/upcoming_changes/32029.new_feature.rst
+++ b/doc/release/upcoming_changes/32029.new_feature.rst
@@ -2,5 +2,3 @@ Add DLPack support to NumPy scalars
 -----------------------------------
 ``scalar.__dlpack__`` and ``scalar.__dlpack_device__`` methods have been added
 to NumPy scalars to allow exporting them to DLPack, similarly to ndarrays.
-Scalars can only be exported with read-only access (``__dlpack__(copy=False)``)
-and to DLPack version 1.0 or higher.
--- a/numpy/_core/src/multiarray/dlpack.c
+++ b/numpy/_core/src/multiarray/dlpack.c
@@ -661,10 +661,15 @@ gentype_dlpack(PyObject *self,
     }
 
     if (copy_mode == NPY_COPY_ALWAYS) {
-        // Scalars cannot be copied
-        PyErr_SetString(PyExc_BufferError,
-            "Cannot export scalars with copy=True to DLPack.");
-        return NULL;
+        /* Export a fresh 0-d array that owns the copied data. */
+        PyArrayObject *arr = (PyArrayObject *)PyArray_FromScalar(self, NULL);
+        if (arr == NULL) {
+            return NULL;
+        }
+        PyObject *res = create_dlpack_capsule(
+                NULL, arr, major_version >= 1, &result_device, 1);
+        Py_DECREF(arr);
+        return res;
     }
 
     if (major_version < 1) {
@@ -675,8 +680,7 @@ gentype_dlpack(PyObject *self,
     }
 
     return create_dlpack_capsule(
-            self, NULL, major_version >= 1, &result_device,
-            copy_mode == NPY_COPY_ALWAYS);
+            self, NULL, major_version >= 1, &result_device, 0);
 }
 
 NPY_NO_EXPORT PyObject *
--- a/numpy/_core/src/multiarray/dlpack.c
+++ b/numpy/_core/src/multiarray/dlpack.c
@@ -396,7 +396,15 @@ fill_dl_tensor_information_from_scalar(
     DLTensor *dl_tensor, PyObject *self, DLDevice *result_device)
 {
     PyArray_Descr *dtype = PyArray_DescrFromScalar(self);
+    if (dtype == NULL) {
+        return -1;
+    }
+
     void *data = scalar_value(self, dtype);
+    if (data == NULL) {
+        Py_DECREF(dtype);
+        return -1;
+    }
 
     int ret = fill_dl_tensor_information(dl_tensor, result_device,
         dtype->elsize, 0, NULL, NULL, dtype, data);
@@ -546,7 +554,7 @@ array_dlpack(PyArrayObject *self,
             {"$stream", NULL, &stream},
             {"$max_version", NULL, &max_version},
             {"$dl_device", &device_converter, &result_device},
-            {"$copy", &PyArray_CopyConverter, &copy_mode})) {
+            {"$copy", &PyArray_CopyConverter, &copy_mode}) < 0) {
         return NULL;
     }
 
@@ -621,8 +629,7 @@ gentype_dlpack(PyObject *self,
     PyObject *stream = Py_None;
     PyObject *max_version = Py_None;
     NPY_COPYMODE copy_mode = NPY_COPY_IF_NEEDED;
-    /* For scalars, version >=1 is supported, so set default to 1. */
-    long major_version = 1;
+    long major_version = 0;
     /* We allow the user to request a result device in principle. */
     DLDevice result_device = {kDLCPU, 0};
 
@@ -631,7 +638,7 @@ gentype_dlpack(PyObject *self,
             {"$stream", NULL, &stream},
             {"$max_version", NULL, &max_version},
             {"$dl_device", &device_converter, &result_device},
-            {"$copy", &PyArray_CopyConverter, &copy_mode})) {
+            {"$copy", &PyArray_CopyConverter, &copy_mode}) < 0) {
         return NULL;
     }
 
@@ -667,21 +674,13 @@ gentype_dlpack(PyObject *self,
         return NULL;
     }
 
-    /*
-     * TODO: The versioned and non-versioned structs of DLPack are very
-     * similar but not ABI compatible so that the function called here requires
-     * branching (templating didn't seem worthwhile).
-     *
-     * Version 0 support should be deprecated in NumPy 2.1 and the branches
-     * can then be removed again.
-     */
     return create_dlpack_capsule(
             self, NULL, major_version >= 1, &result_device,
             copy_mode == NPY_COPY_ALWAYS);
 }
 
 NPY_NO_EXPORT PyObject *
-gentype_dlpack_device(PyObject *self, PyObject *NPY_UNUSED(args))
+gentype_dlpack_device(PyObject *NPY_UNUSED(self), PyObject *NPY_UNUSED(args))
 {
     return Py_BuildValue("ii", kDLCPU, 0);
 }
--- a/numpy/_core/src/multiarray/dlpack.c
+++ b/numpy/_core/src/multiarray/dlpack.c
@@ -106,7 +106,7 @@ array_dlpack_deleter(DLManagedTensorVersioned *self)
 
     PyGILState_STATE state = PyGILState_Ensure();
 
-    PyArrayObject *array = (PyArrayObject *)self->manager_ctx;
+    PyObject *array = self->manager_ctx;
     // This will also free the shape and strides as it's one allocation.
     PyMem_Free(self);
     Py_XDECREF(array);
@@ -127,25 +127,6 @@ array_dlpack_deleter_unversioned(DLManagedTensor *self)
     PyArrayObject *array = (PyArrayObject *)self->manager_ctx;
     PyMem_Free(self);
     Py_XDECREF(array);
-
-    PyGILState_Release(state);
-}
-
-/*
- * Deleter for a NumPy exported scalar dlpack DLManagedTensor(Versioned).
- */
-static void
-scalar_dlpack_deleter(DLManagedTensorVersioned *self)
-{
-    if (!Py_IsInitialized()) {
-        return;
-    }
-
-    PyGILState_STATE state = PyGILState_Ensure();
-
-    PyObject *scalar = (PyObject *)self->manager_ctx;
-    PyMem_Free(self);
-    Py_XDECREF(scalar);
 
     PyGILState_Release(state);
 }
@@ -454,12 +435,7 @@ create_dlpack_capsule(PyObject *scalar, PyArrayObject *array,
         DLManagedTensorVersioned *managed = (DLManagedTensorVersioned *)ptr;
         capsule_name = NPY_DLPACK_VERSIONED_CAPSULE_NAME;
         capsule_deleter = (PyCapsule_Destructor)dlpack_capsule_deleter;
-        if (array == NULL) {
-            managed->deleter = scalar_dlpack_deleter;
-        }
-        else {
-            managed->deleter = array_dlpack_deleter;
-        }
+        managed->deleter = array_dlpack_deleter;
         managed->manager_ctx = self;
 
         dl_tensor = &managed->dl_tensor;
--- a/numpy/_core/src/multiarray/dlpack.c
+++ b/numpy/_core/src/multiarray/dlpack.c
@@ -6,6 +6,7 @@
 
 #include "dlpack/dlpack.h"
 #include "numpy/arrayobject.h"
+#include "scalartypes.h"
 #include "npy_pycompat.h"
 #include "npy_argparse.h"
 #include "npy_dlpack.h"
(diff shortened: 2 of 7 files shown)


numpy/_core/src/multiarray/dlpack.c
```diff
@@ -130,6 +131,24 @@ array_dlpack_deleter_unversioned(DLManagedTensor *self)
     PyGILState_Release(state);
 }
 
+/*
+ * Deleter for a NumPy exported scalar dlpack DLManagedTensor(Versioned).
+ */
+static void
+scalar_dlpack_deleter(DLManagedTensorVersioned *self)
```
seberg: I think we can just re-use the array one, since both do exactly the same thing.

seberg (commented):
Thanks, just had a brief look, and the restructure looks nice, I don't think you need two deleters, but that is the only thing I noticed.  So when you are happy, let me know and I suspect it's basically ready.

numpy/_core/src/multiarray/dlpack.c
MaanasArora: Done, I just removed the `(PyArrayObject *)` cast, it doesn't seem to matter, thanks!

MaanasArora:
Thanks for looking! I briefly checked the argument parsing again and it seems right to me, so I think this is fine now for me as well.

numpy/_core/src/multiarray/dlpack.c
```diff
@@ -568,6 +614,78 @@ array_dlpack_device(PyArrayObject *self, PyObject *NPY_UNUSED(args))
...
 
+NPY_NO_EXPORT PyObject *
+gentype_dlpack(PyObject *self,
+        PyObject *const *args, Py_ssize_t len_args, PyObject *kwnames)
+{
+    PyObject *stream = Py_None;
+    PyObject *max_version = Py_None;
+    NPY_COPYMODE copy_mode = NPY_COPY_IF_NEEDED;
+    /* For scalars, version >=1 is supported, so set default to 1. */
+    long major_version = 1;
```
ngoldbaum: Shouldn't this be 0? Otherwise an exporter that doesn't have a version (i.e. one from before version 1.0) will be allowed through. If you set this to zero then the BufferError below will trigger in that case.

numpy/_core/src/multiarray/dlpack.c
```diff
@@ -366,19 +351,66 @@ fill_dl_tensor_information(
...
+
+
+/*
+ * Fill the dl_tensor struct from a `self` scalar.
+ */
+static int
+fill_dl_tensor_information_from_scalar(
+    DLTensor *dl_tensor, PyObject *self, DLDevice *result_device)
+{
+    PyArray_Descr *dtype = PyArray_DescrFromScalar(self);
```
ngoldbaum, line 398: This can fail for e.g. datetimes and needs an error check

numpy/_core/src/multiarray/dlpack.c
```diff
@@ -366,19 +351,66 @@ fill_dl_tensor_information(
...
+    void *data = scalar_value(self, dtype);
```
ngoldbaum, line 403: this can also return NULL for string scalars. Both of these are probably only in OOM paths. Still, it's cheap to add error checks and that makes things more robust.

numpy/_core/src/multiarray/dlpack.c
```diff
@@ -568,6 +614,78 @@ array_dlpack_device(PyArrayObject *self, PyObject *NPY_UNUSED(args))
...
+        PyErr_SetString(PyExc_ValueError,
+                "NumPy only supports stream=None.");
+        return NULL;
+    }
+
+    if (copy_mode == NPY_COPY_ALWAYS) {
+        // Scalars cannot be copied
+        PyErr_SetString(PyExc_BufferError,
+            "Cannot export scalars with copy=True to DLPack.");
+        return NULL;
```
ngoldbaum: What about (untested, from an AI model):

```
/* Export a fresh 0-d array that owns the copied data. */
    PyArrayObject *arr = (PyArrayObject *)PyArray_FromScalar(self, NULL);
    if (arr == NULL) {
        return NULL;
    }
    PyObject *res = create_dlpack_capsule(
            NULL, arr, major_version >= 1, &result_device, 1);
    Py_DECREF(arr);
    return res;
```

numpy/_core/src/multiarray/dlpack.c
```diff
@@ -568,6 +614,78 @@ array_dlpack_device(PyArrayObject *self, PyObject *NPY_UNUSED(args))
...
+    }
+
+    /*
+     * TODO: The versioned and non-versioned structs of DLPack are very
+     * similar but not ABI compatible so that the function called here requires
+     * branching (templating didn't seem worthwhile).
+     *
+     * Version 0 support should be deprecated in NumPy 2.1 and the branches
+     * can then be removed again.
+     */
```
ngoldbaum: copy/paste error? scalars are always exported as version 1.

numpy/_core/src/multiarray/dlpack.c
```diff
@@ -568,6 +614,78 @@ array_dlpack_device(PyArrayObject *self, PyObject *NPY_UNUSED(args))
...
+    return create_dlpack_capsule(
+            self, NULL, major_version >= 1, &result_device,
+            copy_mode == NPY_COPY_ALWAYS);
+}
+
+NPY_NO_EXPORT PyObject *
+gentype_dlpack_device(PyObject *self, PyObject *NPY_UNUSED(args))
```
ngoldbaum:
```suggestion
gentype_dlpack_device(PyObject *NPY_UNUSED(self), PyObject *NPY_UNUSED(args))
```

numpy/_core/src/multiarray/dlpack.c
```diff
@@ -568,6 +614,78 @@ array_dlpack_device(PyArrayObject *self, PyObject *NPY_UNUSED(args))
...
+    long major_version = 1;
+    /* We allow the user to request a result device in principle. */
+    DLDevice result_device = {kDLCPU, 0};
+
+    NPY_PREPARE_ARGPARSER;
+    if (npy_parse_arguments("__dlpack__", args, len_args, kwnames,
+            {"$stream", NULL, &stream},
+            {"$max_version", NULL, &max_version},
+            {"$dl_device", &device_converter, &result_device},
+            {"$copy", &PyArray_CopyConverter, &copy_mode})) {
```
ngoldbaum:
```suggestion
            {"$copy", &PyArray_CopyConverter, &copy_mode}) < 0) {
```

ngoldbaum (commented):
Overall looks great. I threw an AI at this and it spotted some (mostly minor) issues. The biggest problem is that unversioned exports will make it through your version < 1 check, I think.

This needs a release note. Also the new public methods you're adding need docstrings.

numpy/_core/src/multiarray/dlpack.c
MaanasArora: Done in the older code as well, thanks!

numpy/_core/src/multiarray/dlpack.c
MaanasArora: Indeed, I've removed the comment - hope that is fine, thanks!

numpy/_core/src/multiarray/dlpack.c
MaanasArora: Yes that would work indeed! #31893 does exactly this - do we want to allow both? It could be nice, not sure. In the issue @seberg seemed to recommend we limit to read-only bytes (as in buffer protocol). That looks to be sufficient for the soft promise in Array API (the issue motivation).

numpy/_core/src/multiarray/dlpack.c
MaanasArora: That's right yes, sorry! I've changed this to 0, and updated the tests to ensure we reject `max_version=None`.

MaanasArora:
Thanks for the review @ngoldbaum, I think I have addressed most of your comments!

The remaining one is about allowing `copy=True` - it could be nice adding that branch but it feels a little odd to rely on a 0-D array for copy and not for read-only. I'm tempted to use `PyArray_Scalar` to replicate the array path; now we cover scalars explicitly so we could really just use `PyArray_Scalar` or similar maybe?

Edit: it is still added surface, so I'd defer to Sebastian if he thinks it's worth it to add.

numpy/_core/src/multiarray/dlpack.c
seberg: I agree, arrays support `copy=True`, so scalars should as well and the easiest way is to create a new array the way that Nathan suggested (I am very sure that the second `NULL` means we get a copy, although in practice it doesn't matter for any relevant dtypes since dlpack doesn't support structured ones).

numpy/_core/src/multiarray/dlpack.c
```diff
@@ -568,6 +622,69 @@ array_dlpack_device(PyArrayObject *self, PyObject *NPY_UNUSED(args))
...
 }
 
+NPY_NO_EXPORT PyObject *
+gentype_dlpack(PyObject *self,
+        PyObject *const *args, Py_ssize_t len_args, PyObject *kwnames)
+{
+    PyObject *stream = Py_None;
+    PyObject *max_version = Py_None;
+    NPY_COPYMODE copy_mode = NPY_COPY_IF_NEEDED;
+    long major_version = 0;
```
seberg, line 635: Let's keep it for now to default to 0.  It would be a bit suprising but somewhat valid to default to version 1.
Since version 0 always fails right now it would make some sense.

Let's keep this for now, but I'll bring it up on the `nanobind` issue.  As much as I like saving 10%, I think nanobind is annoying for gambling here...

seberg (approved):
Thanks, LGTM, don't wait for me (if anyone else looks through).

We should allow the `copy=True` as Nathan suggested.

numpy/_core/src/multiarray/dlpack.c
MaanasArora: Thanks for clarifying! Then we should allow v0 for `copy=True` I suppose (it's a new array, so not read-only)? I've pushed an update!

numpy/_core/src/multiarray/dlpack.c
MaanasArora: Agreed, I think the fact that the defaults failed tripped me up. But now if we support `copy=True` with v0, it makes it slightly more surprising to have a v1 default maybe...

doc/release/upcoming_changes/32029.new_feature.rst
```diff
@@ -0,0 +1,6 @@
+Add DLPack support to NumPy scalars
+-----------------------------------
+``scalar.__dlpack__`` and ``scalar.__dlpack_device__`` methods have been added
+to NumPy scalars to allow exporting them to DLPack, similarly to ndarrays.
+Scalars can only be exported with read-only access (``__dlpack__(copy=False)``)
```
ngoldbaum: is this still true? Using this PR branch:

```
Preimported NumPy 2.6.0.dev0+git20260719.ea4a03d as np

In [1]: import numpy as np
   ...: x = np.float64(2)
   ...: np.from_dlpack(x, copy=True)
Out[1]: array(2.)
```

numpy/_core/src/multiarray/dlpack.c
```diff
@@ -568,6 +622,73 @@ array_dlpack_device(PyArrayObject *self, PyObject *NPY_UNUSED(args))
...
+    /* We allow the user to request a result device in principle. */
+    DLDevice result_device = {kDLCPU, 0};
+
+    NPY_PREPARE_ARGPARSER;
+    if (npy_parse_arguments("__dlpack__", args, len_args, kwnames,
```
ngoldbaum, line 640: I think this and `__dlpack_device__` still need docstrings

numpy/_core/src/multiarray/dlpack.c
```diff
@@ -568,6 +622,73 @@ array_dlpack_device(PyArrayObject *self, PyObject *NPY_UNUSED(args))
...
+        PyObject *res = create_dlpack_capsule(
+                NULL, arr, major_version >= 1, &result_device, 1);
+        Py_DECREF(arr);
+        return res;
+    }
+
+    if (major_version < 1) {
+        PyErr_SetString(PyExc_BufferError,
+            "Cannot export scalars since signalling readonly "
+            "is unsupported by DLPack (supported by newer DLPack version).");
```
ngoldbaum: this could maybe suggest passing `copy=True`?

ngoldbaum: also what about not raising if `copy==None`, where I think copies are allowed.

numpy/_core/src/multiarray/dlpack.c
```diff
@@ -366,19 +351,74 @@ fill_dl_tensor_information(
...
+    return ret;
+}
+
+
+static PyObject *
+create_dlpack_capsule(PyObject *scalar, PyArrayObject *array,
+    int versioned, DLDevice *result_device, int copied)
+{
+    int ndim = array == NULL ? 0 : PyArray_NDIM(array);
+    PyObject *self = array == NULL ? scalar : (PyObject *)array;
```
ngoldbaum, line 422:
```suggestion

    /* Scalars are exported read-only, which only versioned capsules can signal. */
    assert(array != NULL || versioned);
 
```

To catch future edits ever violate scalars taking the rule that scalars always export versioned capsules

numpy/_core/src/multiarray/dlpack.c
```diff
@@ -568,6 +622,73 @@ array_dlpack_device(PyArrayObject *self, PyObject *NPY_UNUSED(args))
...
+                "NumPy only supports stream=None.");
+        return NULL;
+    }
+
+    if (copy_mode == NPY_COPY_ALWAYS) {
+        /* Export a fresh 0-d array that owns the copied data. */
+        PyArrayObject *arr = (PyArrayObject *)PyArray_FromScalar(self, NULL);
+        if (arr == NULL) {
+            return NULL;
+        }
```
ngoldbaum, line 672:
```suggestion
        }
        if (PyArray_BASE(arr) != NULL) {
            /* void scalars without OWNDATA give a view; force a real copy */
            Py_SETREF(arr, (PyArrayObject *)PyArray_NewCopy(arr, NPY_KEEPORDER));
            if (arr == NULL) {
                return NULL;
            }
        }
```

Because `FromScalar` doesn't actually copy all void scalars in all cases:

https://github.com/numpy/numpy/blob/b5ee89fcd46e5f19db1f2cb55b1ecc9b51a98322/numpy/_core/src/multiarray/scalarapi.c#L222-L231

ngoldbaum (commented):
Getting closer, one more review pass.

numpy/_core/src/multiarray/dlpack.c
MaanasArora: Right but I don't think we support void for DLPack?

numpy/_core/src/multiarray/dlpack.c
MaanasArora: Both done, thanks - I tried to add a special case to avoid an error as such:

`    if (copy_mode != NPY_COPY_NEVER && major_version < 1 || copy_mode == NPY_COPY_ALWAYS) {`

A bit confusing, though `COPY_IF_NEEDED` inherently does allow us this flexibility, I suppose. The alternative is defaulting to copy even when read-only is possible, what do you think?

numpy/_core/src/multiarray/dlpack.c
MaanasArora: Added!

doc/release/upcoming_changes/32029.new_feature.rst
MaanasArora: Right not true anymore after the `copy=True` branch was added - removed, thanks!

numpy/_core/src/multiarray/dlpack.c
ngoldbaum: I think this is good!

numpy/_core/src/multiarray/dlpack.c
ngoldbaum: Ah good point. Maybe just as an assert that `PyArray_BASE(arr) == NULL` then? So future callers don't have issues, if other dtypes happen to break that assumption.

ngoldbaum (approved):
All that's left is quibbles I think, so I'm approving. Feel free to self-merge once you think it's ready.

numpy/_core/src/multiarray/dlpack.c
MaanasArora: Yeah that sounds good, added, thanks!

MaanasArora:
Thanks @ngoldbaum! I think this should be ready - merging!

PR #32030, opened 2026-07-17
BUG: add a special case for StringDType in np.isdtype

### LINKED ISSUE #27545: ENH: `numpy.isdtype` with custom dtypes

The implementation of `numpy.isdtype` currently checks against an explicit list of kind names, and another one of dtypes (`numpy._core._type_aliases.allTypes`), and raises for all other objects. While the description of the exception ("dtype argument must be a NumPy dtype") makes sense for  for dtypes created for other array API implementing types, I don't think this is  the case for custom `numpy` dtypes which are explicitly written for `numpy`.

Incidentally, checking the new `StringDType` also raises on `numpy=2.1.1` (and as far as I can tell would even fail on [current `main`](https://github.com/numpy/numpy/blob/6f63230b54bc9910c849e78922d015e607b6848f/numpy/_core/numerictypes.py#L368-L468)):
```python
In [1]: import numpy as np
   ...: 
   ...: np.isdtype(np.dtypes.StringDType(), "real floating")
Cell In[1], line 3
      1 import numpy as np
----> 3 np.isdtype(np.dtypes.StringDType(), "real floating")

File .../numpy/_core/numerictypes.py:425, in isdtype(dtype, kind)
    423     dtype = _preprocess_dtype(dtype)
    424 except _PreprocessDTypeError:
--> 425     raise TypeError(
    426         "dtype argument must be a NumPy dtype, "
    427         f"but it is a {type(dtype)}."
    428     ) from None
    430 input_kinds = kind if isinstance(kind, tuple) else (kind,)
    432 processed_kinds = set()

TypeError: dtype argument must be a NumPy dtype, but it is a <class 'numpy.dtypes.StringDType'>.
```

This came up in pydata/xarray#9568, where someone wanted to use `xarray` with the dtypes defined by `ml_types`, which eventually failed because of the `numpy.isdtype` call.

So my question is: should `numpy.isdtype` support checking custom `numpy` dtypes? And if so, should it always return `False`, or do we need some kind of mechanism to include custom dtypes in the categories defined by the kind names?

ngoldbaum:
### PR summary

Fixes #27545.

I don't see a cleaner way to fix this without just adding a special case in the error path. This function is only for built-in NumPy DTypes so it would be a bigger change to just make it return True if it's handed a dtype or dtypemeta instance, which is probably a more principled fix.

#### AI Disclosure

No AI use on this one.

### DIFF
--- a/numpy/_core/tests/test_numerictypes.py
+++ b/numpy/_core/tests/test_numerictypes.py
@@ -470,6 +470,32 @@ class TestIsDType:
         with assert_raises_regex(ValueError, r".*not a known kind name.*"):
             np.isdtype(np.int64, "int64")
 
+    def test_isdtype_string_dtype(self):
+        # gh-27545: StringDType has no NumPy scalar type, but it is a
+        # built-in dtype, so `isdtype` must accept it rather than raise
+        dt = np.dtypes.StringDType()
+        assert not np.isdtype(dt, "bool")
+        for kind in self.dtype_group_dict:
+            assert not np.isdtype(dt, kind)
+
+        # matches itself and the StringDType class, which stands in for
+        # the scalar type that identifies the other dtypes
+        assert np.isdtype(dt, dt)
+        assert np.isdtype(dt, np.dtypes.StringDType)
+        assert np.isdtype(np.dtypes.StringDType, dt)
+        assert np.isdtype(dt, ("numeric", np.dtypes.StringDType))
+        assert not np.isdtype(np.int64, dt)
+        assert not np.isdtype(dt, ("integral", np.int64))
+
+        # dtype parameters are ignored, like datetime64 units
+        assert np.isdtype(np.dtypes.StringDType(na_object=None), dt)
+        assert np.isdtype(dt, np.dtypes.StringDType(na_object=np.nan,
+                                                    coerce=False))
+
+        # fixed-width unicode strings are a different kind
+        assert not np.isdtype(dt, np.str_)
+        assert not np.isdtype(np.dtype("U8"), dt)
+
     def test_sctypes_complete(self):
         # issue 26439: int32/intc were masking each other on 32-bit builds
         assert np.int32 in sctypes['int']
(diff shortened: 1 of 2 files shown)


seberg:
Note that this would be superseded by gh-31755, but I need to file it (and make a decision if we expose `.register()` or not, but I am OK with just not doing it for now).

But as always, that doens't mean we can't just do this now :).

seberg (approved):
Seems simple enough, so happy if you put it in, just a small merge conflict for the other PR.

numpy/_core/tests/test_numerictypes.py
```diff
@@ -470,6 +470,32 @@ def test_isdtype_invalid_args(self):
...
+        assert np.isdtype(dt, np.dtypes.StringDType)
+        assert np.isdtype(np.dtypes.StringDType, dt)
+        assert np.isdtype(dt, ("numeric", np.dtypes.StringDType))
+        assert not np.isdtype(np.int64, dt)
+        assert not np.isdtype(dt, ("integral", np.int64))
+
+        # dtype parameters are ignored, like datetime64 units
+        assert np.isdtype(np.dtypes.StringDType(na_object=None), dt)
+        assert np.isdtype(dt, np.dtypes.StringDType(na_object=np.nan,
+                                                    coerce=False))
```
seberg, line 493: Ah one note: I think I would want to deprecate this.  It doesn't actually make much sense, because to ask it for a dtype instance as second argument, IMO.
(Even if it is convenient.)

ngoldbaum: Sure, I can add that.

seberg: If you like to do it for all inputs, but can also do it in the other PR, I dunno how nice it is without the changes around.

ngoldbaum: Ah in that case why don't you add the deprecation in the other PR. You have more context here than I do.

PR #32035, opened 2026-07-18
MAINT: fix ma.vander docstring to match its actual signature

### LINKED ISSUE #32033: DOC: numpy.ma.vander docstring documents unsupported `increasing` parameter

`numpy.ma.vander` reuses the `numpy.vander` docstring via `ma.doc_note`, but its implementation accepts fewer arguments than the inherited documentation describes.

https://github.com/numpy/numpy/blob/848a57d5b97fc3c17e7ff4f5be83d09991f52b22/numpy/ma/extras.py#L2243-L2255

Since the wrapper reuses the upstream docstring, the published documentation for `numpy.ma.vander` (https://numpy.org/doc/stable/reference/generated/numpy.ma.vander.html, and likewise on the current devdocs) describes the API as:

```text
vander(x, N=None, increasing=False)
```

including the `increasing` parameter and an example using:

```python
>>> np.vander(x, increasing=True)
```

However, the callable API of `numpy.ma.vander` is:

```python
>>> import inspect
>>> import numpy as np
>>> inspect.signature(np.ma.vander)
<Signature (x, n=None)>
```

Attempting to use the documented keyword argument raises a `TypeError`:

```python
>>> import numpy as np

>>> x = np.ma.array([1, 2, 3, 5])
>>> np.ma.vander(x, increasing=True)
...
TypeError: vander() got an unexpected keyword argument 'increasing'
```

As a result, the published documentation does not match the current implementation.

I’d like some guidance on the intended behavior before I open a PR, as there seem to be two reasonable options. 

1. Add support for the `increasing` parameter and forward it to `numpy.vander`, so the implementation matches the interface described in the inherited documentation.
2. Replace the inherited docstring with one specific to the wrapper, documenting only the arguments that `numpy.ma.vander` actually supports. 

I’m happy to work on whichever approach best fits the intended design of `numpy.ma.vander`.

Souptik96:
Closes #32033.

## Problem

`numpy.ma.vander` reuses `numpy.vander`'s docstring via `ma.doc_note(np.vander.__doc__, vander.__doc__)`, but `ma.vander`'s actual signature is `(x, n=None)` -- it does not implement the `increasing` parameter that the inherited docstring describes (and documents with an example). Following the published docs and calling `np.ma.vander(x, increasing=True)` raises `TypeError`.

## Fix

This resolves the mismatch by giving `ma.vander` its own accurate, self-contained docstring describing only the parameters it actually accepts (`x`, `n`), plus a `Notes` section describing the masked-value behavior. This is option 2 from the issue discussion (correct the docs to match the implementation, rather than adding `increasing` support to the implementation).

## Testing

Added `TestVander` to `numpy/ma/tests/test_extras.py`:
- `test_vander`: verifies `ma.vander` matches `np.vander` for unmasked input and zeroes masked rows.
- `test_vander_docstring_matches_signature`: asserts `increasing` is no longer mentioned in the docstring, and that passing it still raises `TypeError` (regression guard against re-introducing the mismatch).

Ran the full `numpy/ma/tests/test_extras.py` suite (105 tests) against the patched module in an isolated environment -- all pass, no regressions.

### DIFF
--- a/numpy/ma/extras.py
+++ b/numpy/ma/extras.py
@@ -2242,6 +2242,34 @@ def clump_masked(a):
 
 def vander(x, n=None):
     """
+    Generate a Vandermonde matrix, treating masked values as zero rows.
+
+    The columns of the output matrix are powers of the input vector,
+    with the `i`-th output column equal to the input vector raised
+    element-wise to the power of ``n - i - 1``.
+
+    Parameters
+    ----------
+    x : array_like
+        1-D input array. Masked values are set to zero in the
+        corresponding rows of the output.
+    n : int, optional
+        Number of columns in the output. If `n` is not specified, a
+        square array is returned (``n = len(x)``).
+
+    Returns
+    -------
+    out : ndarray
+        Vandermonde matrix. The first column is ``x^(n-1)``, the
+        second ``x^(n-2)`` and so forth. Rows corresponding to masked
+        values in `x` are set to zero.
+
+    See Also
+    --------
+    numpy.vander
+
+    Notes
+    -----
     Masked values in the input array result in rows of zeros.
 
     """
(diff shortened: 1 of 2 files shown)


mattip:

PR #25894, opened 2024-02-27
CI, BLD, TST: Re-enable Emscripten/Pyodide CI job for NumPy

agriyakhetarpal:
## Description

This PR supersedes gh-24603. It adds a CI job to test NumPy v2.0.0.dev0 against a Pyodide (wasm32) runtime. Some of the key changes here are:

1. A patch has been added to the `ci/tools/emscripten/` directory based on an upstream change to Pyodide at https://github.com/pyodide/pyodide/pull/4502. This patch ensures that the correct Meson build system (i.e., `vendored-meson`) is found during the build process.
2. A new requirements file has been created that contains pure-Python dependencies that can be installed inside a Pyodide virtual environment. It has been placed in the same directory as above.
3. Some ancillary files have been added to ensure that the WASM wheel is compiled without specialised CPU instructions (SIMD, etc.) and that the build procedure does not need to build against BLAS or LAPACK, which are currently unavailable on WebAssembly.
4. Various tests have been skipped, such as:
    - All `f2py` tests, since Fortran cannot run in WASM
    - Some tests for `np.where()` that require floating point exception support
    - Some tests related to the use of subprocesses to retrieve configurations for how NumPy has been built
    - A bug in Cython, i.e., https://github.com/cython/cython/issues/5411, has now been resolved. The `import cython` line was breaking the test discovery because Cython is not supported in Pyodide in-tree yet.
    - and so on

Thanks to @rgommers and the notes provided on gh-24603, all tests pass! Here's a workflow run from my fork where they can be observed: https://github.com/agriyakhetarpal/numpy/actions/runs/8065629246

PR #21895, opened 2022-07-01
ENH, CI: Add Emscripten to CI

hoodmane:
This is using a branch of Pyodide that finishes support for out of tree builds. This is blocked on Pyodide:
- [x] Merge changes from my branch into main Pyodide
- [x] Release Pyodide v0.21.0a3 (hopefully) with out of tree build support

Other than that, I think it is nearly ready.

PR #32036, opened 2026-07-18
ENH: Emulate floating-point exception flags on Emscripten to surface `LinAlgError`s

### LINKED ISSUE #4859: NumPy: `numpy.linalg.LinAlgError` not thrown
## 🐛 Bug

The following code:

```python
import numpy as np

A = np.array([[1,-2j],[2j,5]])
print(np.linalg.cholesky(A))

B = np.array([[1,3],[2,4]])
np.linalg.cholesky(B)
```

Produces an error on normal CPython, just like the [docs state it should](https://numpy.org/doc/stable/reference/generated/numpy.linalg.cholesky.html).

```pytb
[[1.+0.j 0.+0.j]
 [0.+2.j 1.+0.j]]
  File "/Users/henryschreiner/git/webpages/iscinumpy.gitlab.io/np.py", line 7, in <module>
    np.linalg.cholesky(B)
  File "/usr/local/lib/python3.12/site-packages/numpy/linalg/linalg.py", line 779, in cholesky
    r = gufunc(a, signature=signature, extobj=extobj)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/numpy/linalg/linalg.py", line 115, in _raise_linalgerror_nonposdef
    raise LinAlgError("Matrix is not positive definite")
numpy.linalg.LinAlgError: Matrix is not positive definite
```

Pyodide does not get the exception, though, and instead just sets NaN's.

### To Reproduce

On https://pyodide.org/en/stable/console.html

```pycon
>>> import numpy as np
>>> A = np.array([[1,-2j],[2j,5]])
>>> print(np.linalg.cholesky(A))
[[1.+0.j 0.+0.j]
 [0.+2.j 1.+0.j]]
>>> B = np.array([[1,3],[2,4]])
>>> np.linalg.cholesky(B)
array([[nan, nan],
       [nan, nan]])
```

The same code produces an error outside a pyodide.

### Expected behavior

It should throw `numpy.linalg.LinAlgError`.

### Additional context

See https://github.com/scikit-hep/iminuit/pull/995, this is breaking iminuit's test suite.

agriyakhetarpal:
### PR summary

This PR adds a software-based emulation of `fenv.h`'s exception flags for the Emscripten platform, as a means to surface software-level linalg errors in WASM. This is a partial solution, as it only works for errors NumPy raises itself through `npy_set_floatstatus_*` and `np::FloatStatus`, and not for the ones that would have to be raised by a hardware trap, which is a limitation of the WASM binary format. We rely on this here because Emscripten's `fenv.h` is a stub (https://github.com/emscripten-core/emscripten/blob/91f02346bdc3383f1de4a764275674d652c6e28b/system/lib/libc/musl/include/fenv.h#L8) and neither alters nor reports the status of the FP environment. I discussed this with @ngoldbaum recently, who pointed me to `npy_extobj_contextvar`, and I discussed it further with @seberg. We are at the EuroSciPy 2026 sprints!

The idea is that Emscripten lacks an FP status register, so the errors NumPy reports when explicitly calling `npy_set_floatstatus_*` are silently lost. This happens primarily in the `numpy.linalg` gufuncs, which flag LAPACK failures via `npy_set_floatstatus_invalid`, so `np.errstate` never sees them.

NumPy having its own `npy_fenv.h` is useful because any other places where `fenv.h` was being used pick up the flags too, which is what makes `FloatStatus`es work and recovers the half-float overflow and underflow reports in `half_private.hpp`. Note that my previous implementation in this PR did not do so and was limited to just the IEEE 754 floating-point standard functions in `ieee754.c.src`.

This works with the default symbol visibility, rather than with hidden visibility, because npymath is statically linked into every extension module. This software-based C variable "register" of sorts needs to be collapsed from per-module copies to be shared across modules, so that the linalg gufunc's writing in `_umath_linalg` is not lost with the FP-error check's read in `_multiarray_umath`. It is our responsibility to handle this software-level "register" because we don't have a shared hardware-level FP status register, as is available in a non-WASM build on a conventional platform.

We discussed this approach, and more or less came to the conclusion that the approach I have utilised here is better than the alternative approach to handle this, which is about exposing the LAPACK `info` parameter on a LAPACK failure and having a separate `_linalg_errstate` context manager for these errors.

Closes pyodide/pyodide#4859
See also WebAssembly/design#1384

#### AI Disclosure

Yes, I used Claude Code Sonnet 4.6 for socratic purposes, i.e., to ask questions and gain a rudimentary understanding of the NumPy gufunc infrastructure and how it handles error statuses. The code is written by me, a human, since it's just based on the nearby non-Emscripten code branch. I derived some help to make sure that my C syntax is apt, since I don't write it on a daily basis, rather barely, and thus am a bit rusty about the language :-)


numpy/_core/tests/test_umath.py
```diff
@@ -4450,7 +4441,6 @@ def test_it(self):
             assert_almost_equal(fz.real, fr, err_msg=f'real part {f}')
             assert_almost_equal(fz.imag, 0., err_msg=f'imag part {f}')
 
-    @pytest.mark.xfail(IS_WASM, reason="doesn't work")
```
agriyakhetarpal, line 4453: The `"doesn't work"` tests here were also xpassing when I was testing locally, but they pass for unrelated reasons.

agriyakhetarpal:
Ready for review, @seberg and @ngoldbaum! I added a better note about requiring default visibility in the PR description, which I hope is adequate, but PTAL and LMK if I should add a code comment or something. I tested this with the SciPy QMC sampler I was discussing, and it works nicely. This should also fix a bunch of cases of LinAlgErrors-not-raising in scikit-learn, statsmodels, iminuit, and Matplotlib when it makes it to a release.

numpy/_core/tests/test_umath.py
seberg: Hmmmm, did wasm make longdouble work now?  (Or maybe it is correctly just `double` on those builds now...)
These skips look very "longdouble is broken" to me :).

seberg:
Thanks, FWIW, this seems pragmatic and pretty good to me.  The visibility/linking makes sense but is also an odd one out: We don't really use this type of thing.

@mattip does this worry you to expose these `npy_get_floatstatus_*`, `npy_set_floatstatus_*` as public symbols (and rely on the fact that linalg is always imported after umath)?  This is even _only_ on wasm.
(I think you may have a more reliable gut feeling about this type of change; everything else here is very straight-forward in a sense after all.)

numpy/_core/tests/test_umath.py
agriyakhetarpal: Kind of; these xfails were added in gh-21895, when we used Emscripten 3.x. I forgot to unskip them in gh-25894 when I was first working on NumPy in Pyodide/WASM, at which point we were also on 3.x. We are at 5.0.3 in Pyodide now. But https://github.com/emscripten-core/emscripten/issues/22123 and https://github.com/pyodide/pyodide/issues/4881 are currently open (and we checked that they are valid), so I think the only logical explanation to me is that it's a combination of testing up to 15 decimal places (which aren't sensitive enough) and that those longdouble functions were just returning garbage or something previously.

Also note that `test_precisions_consistent` isn't actually validating quad longdouble, and passes at double tolerance.

mattip:
I wonder if this is the right level, if it wouldn't be better to try to fake the actual [floating-point routines and macros](https://en.cppreference.com/c/header/fenv) in `fenv.h`

agriyakhetarpal:
@seberg, I think the ordering for linalg/umath is purely cosmetic!

@mattip, yes, well, it does have to live somewhere, and if not NumPy, then Emscripten would have been the right place – but there is no consensus on https://github.com/emscripten-core/emscripten/issues/22005, mostly pushback, and https://github.com/emscripten-core/emscripten/issues/11086 can't really move without a proper WebAssembly design change. The exception macros I've added here previously did exist and were rather later removed on purpose in earlier versions of Emscripten back then: https://github.com/emscripten-core/emscripten/pull/11087, and, based on the conversation there, were likely intended to reduce the libc size.

That is why I think there is little to no benefit to Emscripten in re-adding FE_INVALID, FE_DIVBYZERO, FE_OVERFLOW, etc., because it still won't fix the issue. Probably the only benefit would be that we wouldn't need the default visibility hack that way. I'm not sure whether NumPy should have its own `fenv.h` that overrides the one provided by libc for these compiler cases, in case that's what you meant.

Please let me know in case I misunderstood this – thanks!

mattip:
I guess a single `_npy_wasm_floatstatus` global is not the end of the world. I was thinking that you could mock fenv.h for enscriptem here, inside numpy, but if all it takes is a single global, so be it. This of course goes against what we are trying to do elsewhere in moving away from static variables, but I think the whole floating point error status mechanism is not threadsafe.

agriyakhetarpal:
Agreed, it's not ideal in the abstract once you put it that way. At least this is under Emscripten only, so native builds won't gain a new global and the free-threading work should be unaffected, we don't intersect. Pyodide will remain single-threaded and keep the GIL enabled for a reasonable amount of time in the coming years, since adding `-pthreads` is a pretty gnarly ABI change for us.

agriyakhetarpal:
FYI, a0da728a8445c88768ab2037e551e023353ee19c is a commit that I just pushed that shows what a `npy_fenv` header can look like. This is basically a bit of juggling around the imports and macros everywhere, plus the Emscripten bits. There is still one mutable global here, but the nice thing is that it also allows `np::FloatStatus`es to work on Emscripten. My implementation before this commit only handled things at the `npy_*_floatstatus_*` layer, so we can think of this as one layer below that, in some sense of the world, as the half-float conversions in `half_private.hpp` start reporting as well.

Overall, doing it this way allows us to remove a skip for one more test in d338c7be9ac5dedb966e7d1cea84c2811707b8db: the float16 sum overflow in `numpy/_core/tests/test_ufunc.py::test_sum` because it uses `FloatStatus::RaiseOverflow()`, which works now.

Would this be a way forward to better address your concerns about this PR, @mattip? Also, re-requesting a review from @seberg since the diff changed slightly.

P.S. I also pushed another commit to unskip `numpy/_core/tests/test_datetime.py::test_timedelta_divmod_warnings` (it was already passing before this fenv change; I just forgot to unskip that one when unskipping the rest 😅)

mattip:
Yes this is what I was imagining. The diff is not all that bad, the testing changes are all unskipping existing tests, and the header reads quite clearly, thanks!

seberg:
Yeah, symbol visibility wise it is annoying, but seems OKish to me here and I am fine with it.  Is the indirection via `#define` macros useful for some reason?  (I'll guess it is though, e.g. because these are existing functons so we can't re-define them.)

Iif @mattip is happy enough, I am.  The only question would be that this _is_ kinda private right now, should we re-inforce that on the `npy_wasm_fenv_flags` naming (I am fine if not).

agriyakhetarpal:
> Is the indirection via `#define` macros useful for some reason? (I'll guess it is though, e.g. because these are existing functons so we can't re-define them.)

Yes, I considered another option, which is to define non-`static` `fetestexcept` and co., that would override those set by Emscripten's MUSL. However, that would introduce duplicate symbols and fail at link time.

> The only question would be that this _is_ kinda private right now, should we re-inforce that on the `npy_wasm_fenv_flags` naming (I am fine if not).

Right, as far as the default visibility is concerned, we could actually name it `npy__wasm_fenv_flags` instead (and same for the other `npy_wasm_fe*` symbols, but those are kinda harmless because they are `static inline`d). Would that work, @seberg?

ngoldbaum:
@agriyakhetarpal we talked a little about this at the triage meeting. We feel like if you're fine with committing to fix issues that come up, this is OK to merge. Is this ready?

agriyakhetarpal:
Hi @ngoldbaum (and those who were present at the triage meeting), thanks for discussing this! I can commit to fixing issues that come up with this part of the code, to the best of my ability, as I am with Emscripten- and Pyodide-specific problems overall. I confirm that this is ready. Thank you!

agriyakhetarpal:
I merged `main` just now, to see if any issues may have arisen (just in case)

ngoldbaum:
Thanks @agriyakhetarpal!

PR #32038, opened 2026-07-19
BUG: reference leak in `simd_sequence_from_iterable`

Ethan0456:
### PR summary

`simd_sequence_from_iterable` leaked the `PySequence_Fast` result when `simd_sequence_new` failed: the min-size path already `Py_DECREF`d, the OOM path did not. Add the missing `Py_DECREF`.

No new test; the leak only runs on allocation failure, which we cannot force from Python. Same approach as #31968 (no `getrefcount` asserts).

No release note (internal leak; no API change).

Exploring NumPy and submitting small correctness fixes (e.g. #31968).

### AI Disclosure

AI helped locate and draft the fix. Human-reviewed and submitted by a human.


ngoldbaum (approved):
LGTM.

PR #32039, opened 2026-07-19
BUG: broadcast the condition in np.ma.masked_where

### LINKED ISSUE #17605: Condition in numpy.ma.masked_where does not broadcast
The numpy.ma.masked_where does not broadcast the masking condition over the array to mask. 

### Reproducing code example:
```python
import numpy as np

condition = np.array([False, True, False])
a = np.random.randn(3, 100)
np.ma.masked_where(condition[:, np.newaxis], a) #throws IndexError

condition_br = np.broadcast_to(condition[:, np.newaxis], a.shape)
np.ma.masked_where(condition_br, a) #works as expected
```

1.19.1 3.8.5 (default, Sep  3 2020, 21:29:08) [MSC v.1916 64 bit (AMD64)]

vsaraikin:
Fixes gh-17605.

np.ma.masked_where raised IndexError whenever the condition's shape differed from the input, even when the condition was broadcastable against it. That is inconsistent with np.where and boolean indexing, which both broadcast the condition. So a column-shaped condition like (3, 1) against a (3, 2) array raised instead of masking the selected rows.

The change broadcasts the condition against the input with np.broadcast_to and only raises when the shapes are genuinely incompatible. The existing behaviour of combining the condition with the input's own mask is preserved.

Added test_masked_where_broadcast covering a column condition, a row condition, the mask-combining case, and a genuinely incompatible shape that still raises. It fails on main and passes with the change.

Happy to add a release note fragment once this has a PR number (added in a follow-up commit).

### DIFF
--- a/doc/release/upcoming_changes/32039.improvement.rst
+++ b/doc/release/upcoming_changes/32039.improvement.rst
@@ -1,0 +1,8 @@
+``np.ma.masked_where`` broadcasts the condition
+-----------------------------------------------
+`numpy.ma.masked_where` now broadcasts the ``condition`` against ``a`` instead
+of requiring their shapes to match exactly, matching the behaviour of
+`numpy.where` and boolean indexing. A broadcastable condition such as a column
+of shape ``(3, 1)`` against a ``(3, 2)`` input previously raised an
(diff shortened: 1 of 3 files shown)


jorenham:

